PR #7806, opened 2021-10-02
wolfssl: use for SHA256, MD4, MD5, and setting DES odd parity

haydenroche5:
Prior to this commit, OpenSSL could be used for all these functions, but
not wolfSSL. This commit makes it so wolfSSL will be used if USE_WOLFSSL
is defined.


bagder (approved):
Lovely!

bagder:
Thanks!

PR #7807, opened 2021-10-03
connection: tcp-maxseg option to configure MSS

freedge:
MTU clamping issues can be hard to investigate. We add an option
"tcp-maxseg" that allows to reduce the MSS sent by the client
in the TCP handshake. On Linux, trying to increase the MSS will
be ignored, MSS can be reduced down to 88.


bagder:
Wow, this is niche. Can you please elaborate on when you and users would use this option? tcp(7)'s `TCP_MAXSEG` section is not helping much...

Also, can't the `CURLOPT_SOCKOPTFUNCTION` be used for this and therefore the library won't need change?

Also note that the build fails in numerous CI builds.

freedge:
Hi! I had to troubleshoot a few MTU issues, basically client and server exchange MSS between each other, and each MSS is based on the MTU of their network interface. For example if both client and server use jumbo frames, both send a MSS of 8960 and they will both send TCP packets with 8960 byte payload. Chance is though, that there is a router in the middle that does not support jumbos, so 8960 byte packets will be dropped.

To solve this issue there is a path MTU discovery mechanism, where a router would see a payload too big and reply with an ICMP "too big" packet, there is another mechanism called clamping, where routers rewrite the MSS of the TCP handshake packets so it fits (so my example above is incorrect, while both client and server exchange a MSS of 8960, they would receive a much lower value like 1460, and send payloads of that size).

However both these are sometimes failing, so there is another workaround that can be applied on client side, through the use of iptables, however it requires the user to run a command as root.

When investigating such issue, for example you get a user complaining about an SSL handshake timeout, the first reply is usually, can you reproduce with curl? Then if you suspect a MTU issue, you can ask the user to run something like ```iptables -t mangle -A POSTROUTING -p tcp --tcp-flags SYN,RST SYN -o eth0 -j TCPMSS --set-mss 1460``` which is not always possible (eg: user not root, or from within a container) and feels overly complex. You can of course troubleshoot the issue with a couple of pings of various length (but only network guys find "ping" exciting, users enjoy much more a successful curl).

I will look into the CURLOPT_SOCKOPTFUNCTION thing. CI issue seems link to int/long conversion, strangely I missed the warning in my environment. I can look into it.

However if you believe this is too niche and there is no interest in having this feature I am fine with closing this pull request, no need to clutter curl with useless options :)

bagder:
I'm happy to see your willingness to improve curl, but yes, I think this particular command line option is so very specific and while it helped you for your problem this time, to me this feels like a too narrow functionality to put in there for everyone.

I'll leave this request open for a while longer anyway to allow others to chime in and tell us their views in case there's more interest or something else that I've not realized myself that would change my mind.

danielgustafsson:
Like I said on Twitter, my fear is that this is niche enough that we won't know when we break it and noone will find out until long after the fact. I like seeing cool stuff being proposed, but I think the long-term maintenance cost is prohibitive on this one.

PR #7808, opened 2021-10-03
Fix some deprecation issues with OpenSSL 3.0

MarcelRaad:
This should fix everything except openssl.c itself, which requires more work as some of the functions used there are deprecated without a replacement.

### DIFF
--- a/lib/curl_ntlm_core.c
+++ b/lib/curl_ntlm_core.c
@@ -49,7 +49,14 @@
      in NTLM type-3 messages.
  */
 
-#if defined(USE_OPENSSL) || defined(USE_WOLFSSL)
+#if defined(USE_OPENSSL)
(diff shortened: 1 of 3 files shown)


MarcelRaad:
I just noticed the bad naming in sha256.c, where I used the Curl_ prefix for TU-local/static functions and types. Any better idea that won't clash with the names in any external SSL backend?

bagder:
> Any better idea that won't clash with the names in any external SSL backend?

Static functions don't need to be `Curl_` prefixed at all. If you still want the functions to have a prefix, you're free to select one you like that's specific for that single source file. In some cases we've used `my_` then but there's no fixed rule.

bagder (approved):
Looks good to me. For some reason the zuul build doesn't show up now!

PR #7809, opened 2021-10-04
BINDINGS: Fix outdated links

nightlark:
I noticed a few outdated URLs when I was checking out the bindings:

- cpr, BBHTTP, and curl-rust now live under different user/org names on GitHub
- hbcurl for Harbour has renamed their master branch to main
- Eiffel package has moved from `room.eiffel.com` to `iron.eiffel.com`
- Haskell package link has moved
- libcurl-mono package is inaccessible, url has been updated to archive.org -- though the download does not work and I have been unable to find where the package might have moved to


bagder (approved):
:+1:

bagder:
Thanks!

Issue #7742, opened 2021-09-17
libcurl from a Windows dll sometimes incorrectly identifies presence of ALPN


wyattoday:
### I did this

Compile latest libcurl on Windows with SChannel (either static or dynamic library). Then include either static or dynamic libcurl from inside a dll that will be run in another host application. ALPN detction will now sometimes fail depending on whether the host application was compiled with the correct configured [Application Manifest file](https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests).

In this scenario *we* control the dll (which eventually calls libCurl functionality). However, we do not control the "host application". Meaning the functionality we call in curl could work fine or break depending on whether the host application was compiled and packaged correctly.

Also, this is not a hypothetical situation -- we've seen this in the field.

### I expected the following

libCurl should work the same regardless of whether the host application has a manifest configured correctly.

### curl/libcurl version

7.79.0

### operating system

Windows 8.1 or newer.

### Why does it fail?

Incorrect Windows Version detection is used. Or, rather, flaky version detection is used. This was a problem of Microsoft's making (they decided to make some version detection functions flaky).

### Solution

Use alternative version detection functions that *always* returns the correct version. I've already tested and made a PR with the solution. See https://github.com/curl/curl/pull/7727

jay:
I agree we should do something about this. I think libcurl should know the actual version and I don't know what disadvantage there would be to that.

PR #7727, opened 2021-09-15
Fix version checking in Windows (proper backwards and forwards compatible version checking)

wyattoday:
[VerifyVersionInfo](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-verifyversioninfoa) has some of the same problems as GetVersionEx(). Specifically, the problem that we're dealing with is inaccurate version reporting if the application linking libCurl *doesn't* explicitly have an application manifest saying it's compatible with a version greater than or equal to the version we're testing.

Or to put it in MSDN parlance:

> Windows 10:  VerifyVersionInfo returns false when called by applications that do not have a compatibility manifest for Windows 8.1 or Windows 10 if the lpVersionInfo parameter is set so that it specifies Windows 8.1 or Windows 10, even when the current operating system version is Windows 8.1 or Windows 10. Specifically, VerifyVersionInfo has the following behavior:
>
> - If the application has no manifest, VerifyVersionInfo behaves as if the operation system version is Windows 8 (6.2).
> - If the application has a manifest that contains the GUID that corresponds to Windows 8.1, VerifyVersionInfo behaves as if the operation system version is Windows 8.1 (6.3).
> - If the application has a manifest that contains the GUID that corresponds to Windows 10, VerifyVersionInfo behaves as if the operation system version is Windows 10 (10.0).

Why is this a problem? With executables we release, we have full control over the manifests, and thus this isn't a problem. With libraries we release that link to our libraries that use libCurl, suddenly the version detection is broken if the end-developer doesn't know what manifests are, why they're needed, and how to implement them correctly. In short: it becomes a big hassle where features magically work if run from one executable linking the library and suddenly don't work if run from another executable on the same machine linking the exact same library.

Right now this only affects ALPN. However, TLS 1.3 support has arrived in Windows Server 2022 (their LTS server product) and Windows 11. And to detect and use TLS 1.3 proper version checking needs to be implemented (including checking build numbers).

So... what do we do? Well, luckily there's one version function that doesn't trying to outsmart developers (yet) and still reports accurate version numbers regardless of manifests the calling executable may or may not have. Namely, [RtlGetVersion()](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlgetversion)

TLDR: this fixes ALPN detection and prepares for Window Server 2022 / Windows 11 detection for TLS 1.3 support (as well as TCP fast open support on Windows).

[...]

PR #7784, opened 2021-09-27
Add SChannel TLS 1.3 support

wyattoday:
Add SChannel TLS 1.3 support to curl / libcurl.

Has detection at runtime to enable / disable support depending on version of Windows.

Also, added a new define to disable optional code: `DISABLE_SCHANNEL_CLIENT_CERT` (reduces compiled code size if that feature is not needed).

A bit more work needs to be done before this is "merge-ready", but TLS 1.3 support is working.

---

This depends on this PR being merged: https://github.com/curl/curl/pull/7727

Remaining work: Implement "set_ssl_ciphers" for `SCH_CREDENTIALS`. This is in the inverse of the `SCHANNEL_CRED` behavior.

Namely, we're disabling ciphers rather than enabling them. This will require a bit of thinking. I'll work on this this week, but if anyone wants to jump in and write that function, I'll be happy to accept the help.

Once that is implemented, everything else should be good to go. But I'll happily accept input on any of it.

PR #7810, opened 2021-10-04
version_win32: use actual version instead of manifested version

jay:
- Use RtlVerifyVersionInfo instead of VerifyVersionInfo, when possible.

Later versions of Windows have normal version functions that compare and
return versions based on the way the application is manifested, instead
of the actual version of Windows the application is running on. We
prefer the actual version of Windows so we'll now call the Rtl variant
of version functions (RtlVerifyVersionInfo) which does a proper
comparison of the actual version.

Reported-by: Wyatt O'Day

Ref: https://github.com/curl/curl/pull/7727

Fixes https://github.com/curl/curl/issues/7742
Closes #xxxx


wyattoday:
Looks good to me. Just needs build-version support to properly support [my TLS 1.3 PR](https://github.com/curl/curl/pull/7784).

jay:
You're welcome to add a `const unsigned int buildNumber` param after this change lands. I'd like to keep this as simple as possible to move things along.

MarcelRaad (approved):
For the record, I agree with @vszakats that it's not libcurl's job to work around bugs in applications libcurl is directly or indirectly used in, and even if it does, there are many more libraries that don't. I don't have a strong opinion on that though, which is why I've been silent in https://github.com/curl/curl/pull/7727. And as this seems to have no drawbacks, and especially if it helps to land https://github.com/curl/curl/pull/7784 in a less controversial way, I'm fine with it. Thanks to both of you @jay and @wyattoday !

vszakats:
This looks good. With the concern that it's not libcurl's job to fix the bad design around Windows version/feature detection. Such practice will result in libcurl behaving unexpectedly and differently from other code that's designed to honor manifests and the official way of feature-detection. I'm don't think this is the correct direction, though with some luck, it will not cause (or seldom cause) problems in practice.

lib/version_win32.c
```diff
@@ -214,10 +226,23 @@ bool curlx_verify_windows_version(const unsigned int majorVersion,
...
+     possible. Note though the function signature and underlying fundamental
+     types are the same, the return values are different. */
+  if(pRtlVerifyVersionInfo) {
+    matched = !pRtlVerifyVersionInfo(&osver,
+      (VER_MAJORVERSION | VER_MINORVERSION |
+       VER_SERVICEPACKMAJOR | VER_SERVICEPACKMINOR),
+      cm);
+  }
+  else {
+    matched = !!VerifyVersionInfoW(&osver,
```
mback2k, line 241: Why double negation here?

wyattoday:
> You're welcome to add a `const unsigned int buildNumber` param after this change lands. I'd like to keep this as simple as possible to move things along.

Will do. As soon is this is merged I'll make a new PR.

lib/version_win32.c
jay: VerifyVersionInfoW returns BOOL nonzero for success which is, technically in MS world, not guaranteed to be 1. In order for matched to return a guaranteed bool true/false (ie 1/0) I did a double negation. I could instead do it like `else if(VerifyVersionInfoW... {matched = true} else matched = false;` but I think the double negation is cleaner

jay:
I've made a slight change to this, defining OUR_OSVERSIONINFOEXW to be certain the struct we use is always compatible with both VerifyVersionInfoW and RtlVerifyVersionInfo. More details can be seen in the separate commit which I will later squash in.

PR #7812, opened 2021-10-04
Fix a few issues on MidnightBSD

laffer1:


m4/curl-functions.m4
```diff
@@ -2050,6 +2050,10 @@ AC_DEFUN([CURL_CHECK_FUNC_GETADDRINFO], [
         dnl hpux 11.11 and newer
         tst_tsafe_getaddrinfo="yes"
         ;;
+      midnightbsd*)
+        dnl all MidnightBSD versions
```
danielgustafsson, line 2054: Why does it say "all versions" here while the hunk above checks for a version with `< 100000`? If that check is written such that it covers all current versions, then it seems highly likely to break. If that's the case, can it be written more defensively? If it's not the case, aren't these contradictory?

laffer1: There is a catch all below this that catches all BSDs with a value of no.  We don't want that.  Every release of MidnightBSD supports a thread safe getaddrinfo as it was forked from FreeBSD 6.1.

laffer1: These aren't the same file and have different functionality.

bagder (approved):

bagder:
Thanks!

PR #7822, opened 2021-10-06
cmake: add CURL_ENABLE_SSL option

nightlark:
I came across a project building libcurl as a subproject that was trying to set `SSL_ENABLED` to `OFF` to disable SSL support in libcurl. Other than `SSL_ENABLED` being set as a local variable in libcurl and basically overriding anything that they might have set in the cache, issues would have also been encountered because `CMAKE_USE_OPENSSL` defaults to `ON`.

I think adding an `ENABLE_SSL` option (maybe renamed to match any curl cmake option naming convention) that the various SSL backends also depend on would make it much easier for projects building libcurl as a subproject to disable ssl without needing to determine which `CMAKE_USE_*` variables for SSL backends are set, and more obvious what option needs to be turned ON/OFF.

By `ENABLE_SSL` defaulting to `ON`, the behavior of existing SSL related options should remain the same and avoid breaking any existing ci setups or build scripts.

Thoughts?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,29 +343,29 @@ endif()
 
 # check SSL libraries
 # TODO support GnuTLS
-option(ENABLE_SSL "Enable SSL support" ON)
+option(CURL_ENABLE_SSL "Enable SSL support" ON)
 if(CMAKE_USE_WINSSL)
   message(FATAL_ERROR "The cmake option CMAKE_USE_WINSSL was renamed to CMAKE_USE_SCHANNEL.")
 endif()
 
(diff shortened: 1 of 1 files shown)


jzakrzewski:
Please namespace the option. `ENABLE_SSL` is in my opinion too generic and may cause clashes with other projects. I myself have for example included MongoDB driver and curl in the same project and that alone would cause problems with this change.

nightlark:
Option name is now under the `CURL_` namespace.

MarcelRaad:
Merged now. Thank you!

PR #7825, opened 2021-10-07
Add missing goto error

qarmin:
This PR adds missing `goto error`
Without this, failure was not handled. `result` was set to `CURLE_OUT_OF_MEMORY`, but in line 811 it was changed to somethings else, so it was useless

Looks that was copy paste bug from:
https://github.com/curl/curl/blob/e8c8775eaa79fb93208c70a3c607e7e23222e4db/lib/c-hyper.c#L587-L591

### DIFF
--- a/lib/http_proxy.c
+++ b/lib/http_proxy.c
@@ -805,7 +805,7 @@ static CURLcode CONNECT(struct Curl_easy *data,
       if(hyper_request_set_uri(req, (uint8_t *)hostheader,
                                strlen(hostheader))) {
         failf(data, "error setting path");
-        result = CURLE_OUT_OF_MEMORY;
+        goto error;
       }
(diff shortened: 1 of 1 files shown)


jay:
This should be correct except that there is a bigger problem. There are a number of places in the hyper handling code where where goto error and result can be CURLE_OK. I think it should probably end like this

goto cleanup;
error:
if(result == CURLE_OK)
result = CURLE_OUT_OF_MEMORY;
cleanup:
....

@bagder

bagder:
> shouldn't the handshake and other things only be freed on error? does that code currently work?

They run lots of tests fine in the CI and I've done repeated torture tests, even if they are a bit more limited when hyper is used they haven't triggered problems recently.

PR #7826, opened 2021-10-07
Modify library order for openssl linking.

michael-afanasiev:
I've been having issues linking fully static openssl, libz, and libcrypto libraries to curl, with undefined libz symbols coming from libcrypto during configure. This small re-ordering of libraries in openssl.m4 fixes the issue. Happy for any feedback on whether or not this is the correct way to patch.


bagder:
Thanks!

PR #7530, opened 2021-08-03
tests/*server.pl: flush output before executing subprocess

mback2k:
Also avoid shell processes staying around by using exec. This may solve the random Windows CI issues.

PR #7833, opened 2021-10-08
tests/runtests.pl: use default post command delay of 100ms

mback2k:
Although flushing the output buffers reduced the flakiness
of the curl test suite on Windows already, some tests were
still randomly failing. This is another attempt to fix it.

Follow up to #7530


jay:
It will be interesting to know the result of this, however I don't think it's a good idea to actually put it in.

mback2k:
Closing in favor of https://github.com/curl/curl/pull/8516.

PR #7834, opened 2021-10-08
urlapi: URL decode host names

### LINKED ISSUE #7830: Percent-encoded host names in URLs are not decoded
### I did this

    $ curl https://%63url.se/
    curl: (6) Could not resolve host: %63url.se

The percent-encoding in the host name was not acknowledge but is used as-is! The [RFC 3986 section for host name](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) says it can be percent encoded!

### I expected the following

    $ curl https://%63url.se/
    [content from https://curl.se]

... since %63 would be decoded to 'c'.

### curl/libcurl version

7.79.1 and git master

### operating system

any

### credits

This flaw was identified by Noam Moshe,Sharon Brizinov, Raul Onitza-Klugman and Kirill Efimov

bagder:
The host name is stored decoded and is encoded when used to extract the
full URL.

As a bonus, setting the host name part with curl_url_set() no longer
accepts a name that contains space, CR of LF.

Test 1560 has been extended to verify.

Reported-by: Noam Moshe
Reported-by: Sharon Brizinov
Reported-by: Raul Onitza-Klugman
Reported-by: Kirill Efimov
Fixes #7830

### DIFF
--- a/docs/libcurl/curl_url_get.3
+++ b/docs/libcurl/curl_url_get.3
@@ -56,7 +56,7 @@ default port for the scheme.
 .IP CURLU_URLDECODE
 Asks \fIcurl_url_get(3)\fP to URL decode the contents before returning it. It
 will not attempt to decode the scheme, the port number or the full URL.
-
+´
 The query component will also get plus-to-space conversion as a bonus when
 this bit is set.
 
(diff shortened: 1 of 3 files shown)


bagder:
[A canonical URL host name dilemma](https://curl.se/mail/lib-2021-10/0005.html)

PR #7837, opened 2021-10-10
Actually append "-" to --range

verhovsky:
Because the block is missing the `else`, it always executes and overwrites the result of the `if` block before it. So curl has been warning users that it has fixed their incorrect `--range` it but not actually fixing it.


verhovsky:
Because the block is missing the `else`, it always executes, overwriting the result of the `if` block before it, so curl has been warning users that their `--range` is incorrect but not actually fixing it.

This was broken back in 2005 in https://github.com/curl/curl/commit/a93af43974c02c675443e285fa9d9710b2f0a7c9#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176L2363 

(I noticed this working on https://github.com/curlconverter/curlconverter/pull/303)

bagder:
Thanks for that, now I'm with you a 100%. I've made a test case locally that reproduces this problem and verifies your PR.

PR #7838, opened 2021-10-11
docs(templates): PR Template

gptkrsh:
Added PR Template


danielgustafsson:
> I think having a template that can guide people on their PR description is better than empty description PRs

You yourself opened this PR with just three words without a description, with just a single word difference from the commit message =)

Issue #7818, opened 2021-10-05
Ignored tests may cause CI job failures


jay:
### I did this

I have seen failures in the CI (any job which automatically builds and tests our code) where according to runtests output the only tests that failed are ignored tests, ie tests in parentheses (123). First reported [here](https://github.com/curl/curl/pull/7803#issuecomment-932594892).

### I expected the following

Ignored tests should not cause the CI to fail.

PR #7841, opened 2021-10-11
runtests: split out ignored tests

bagder:
Report ignore tests separately from the actual fails. And add temporary
debug output to help track the #7818 issue.

Ref #7818

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -6029,6 +6029,7 @@ if(azure_check_environment()) {
 #
 
 my $failed;
+my $failedign;
 my $testnum;
 my $ok=0;
 my $ign=0;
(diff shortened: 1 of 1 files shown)


bagder:
thanks!

bagder:
I think there's a problem in failing the CI job because of this: the person and people involved at looking into this kind of failures have no idea of what to do about a failure to kill a test server. The actual tested code is probably working perfectly fine!

mback2k:
I think differently about these two things being unrelated. A good example can be seen [here](https://dev.azure.com/daniel0244/curl/_build/results?buildId=7411&view=logs&j=2b73ab9a-a309-5990-a610-4b7ab5f23f7e&t=0ac3f3d5-d1b2-5832-4b65-a397e800db6f&l=1794). The test server still being alive causes the following test 1112 to fail because it expects the test server to have been started fresh (probably with a different configuration or because a previous test could cause issues with it, e.g. blocking it).

My goal would be to keep/reintroduce this behavior with a clear error message at the end and then work on avoiding these server unexpectedly alive in the future.

mback2k:
That is true, but how else can we work on improving the test suite? We could think about only using "-u" on the master branch?

PR #7842, opened 2021-10-12
Fix some problems with the HTTP3 documentation

ghost:
1. If writing to a system path if the command is not prefixed with `sudo` it will cause a permission denied error
2. The patched OpenSSL branch has been updated to `openssl-3.0.0+quic` to match upstream OpenSSL version.
3. We should not disable GnuTLS docs.
4. The `autoreconf` command should be `autoreconf -fi`

If you believe my changes are wrong, feel free to request changes.

### DIFF
--- a/docs/HTTP3.md
+++ b/docs/HTTP3.md
@@ -33,18 +33,18 @@ in the master branch using pull-requests, just like ordinary changes.
 
 Build (patched) OpenSSL
 
-     % git clone --depth 1 -b OpenSSL_1_1_1k+quic https://github.com/quictls/openssl
+     % git clone --depth 1 -b openssl-3.0.0+quic https://github.com/quictls/openssl
      % cd openssl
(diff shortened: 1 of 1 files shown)


bagder:
Since we encourage the use of an installation prefix for this instruction, I think it implies that the user can actually create files there. I think the addition of sudo on all the lines for make install is a complication that we can skip, especially since you add the note in the bottom about it anyway.

OpenSSL 3.0.0 puts the stuff in `lib64/` on 64 bit builds and not just in plain `lib/` which I suppose you also need to update? (and it makes things more complicated since now the path depends on the arch you build for)

bagder:
> `lib64`. That is for Linux not for all MacOS versions especially on Apple Silicon machines

My point: the instruction as it reads right now might fail on Linux x86-64 machines, which are quite commonly used!

PR #7844, opened 2021-10-12
aws-sigv4: fix signature when post data is in binary

theawless:
User sets the [post fields size](https://curl.se/libcurl/c/CURLOPT_POSTFIELDSIZE.html) for binary data. Hence, we should not be using `strlen` on it.
The code is very similar to the [one here in mqtt.c](https://github.com/curl/curl/blob/master/lib/mqtt.c#L484-L491).
I have performed basic sanity testing but do let me know if anything more is required.

Python code to create a binary test file with null character:
```
data = "testPo\0stData".encode()
assert(len(data), 13)
with open("testPostFile.bin", 'wb+') as file:
    file.write(data)
```
Build and installation:
```
./configure --with-secure-transport && make && make install
```
Curl command to test:
```
/usr/local/bin/curl -vvv --aws-sigv4 "testProviderOne:testProviderTwo:testRegion:testService" -u testKeyId:testSecretKey -X POST http://localhost:8080 --data-binary '@testPostFile.bin'
```
I put a log statement (for testing) to confirm the fix:
```
...

Post data str-len:6 post-field-size:13

...

> POST / HTTP/1.1
> Host: localhost:8080
> Authorization: TESTPROVIDERONE4-HMAC-SHA256 Credential=testKeyId/20211012/testRegion/testService/testproviderone4_request, SignedHeaders=host;x-testprovidertwo-date, Signature=14fced2b34bae056b9bc361e02e93525595977a28645f283d73999dfe8527248
> X-Testprovidertwo-Date: 20211012T101558Z
> User-Agent: curl/7.80.0-DEV
> Accept: */*
> Content-Length: 13
> Content-Type: application/x-www-form-urlencoded

...
```


theawless:
> Wait, what happens if you decide to post data using the read callback?

I think the read callback might not supported by sig-v4 because we need to calculate the signature on all of the post data and send that in the header.
For the CLI though, I noticed that both data and data-binary options setup the post fields so it works fine (after the fix).

> Have you looked into adding a test case that verifies this?

Let me try doing that.

bagder:
> I think the read callback might not supported by sig-v4 because we need to calculate the signature on all of the post data and send that in the header.

I suspected so. It doesn't feel like you have to fix that issue in this PR, but it feels like a) the code should then return some sort of error early instead of continuing and pretending it might actually work - and the limitation should probably be documented clearly.

bagder:
Thanks!

Issue #4939, opened 2020-02-17
A paused HTTP/2 stream doesn't actually stop the data


bagder:
### I did this

Write an application that makes multiple HTTP/2 streams over the same connection. Return `CURL_WRITEFUNC_PAUSE` from the write callback when you want to pause a stream.

This only stops the data from being delivered to the application. libcurl will then instead just buffer all the incoming data in memory in an ever-growing realloc()ed buffer until the stream is again unpaused  or shut down. This is not intended, expected or documented.

(Pausing a HTTP/2 transfer with `curl_easy_pause()` has in fact the exact same behavior.)

### I expected the following

Data to stop flowing over the wire for the stream until we unpause the stream again.

### curl/libcurl version

All

### operating system

All

bagder:
Here's [4939.c](https://github.com/curl/curl/files/4231898/4939.c.txt) which reproduces the problem.

jay:
Ref: https://curl.haxx.se/mail/lib-2020-02/0051.html

Can't we just make the window something more reasonable like 1MB or something?

bagder:
> Can't we just make the window something more reasonable like 1MB or something?

That would certainly make it a simpler fix. And yeah, maybe just do that for now and then possibly do something else in a future if users really need something more flexible.

jay:
Since I posted that I found #1102 for the history of why we added the large window. I tried that example with NGHTTP2_INITIAL_WINDOW_SIZE, 1MB and 10MB.

`time src/curl  https://upload.wikimedia.org/wikipedia/commons/d/d6/Seattle_7.jpg -v >/dev/null`

NGHTTP2_INITIAL_WINDOW_SIZE:
real	0m27.026s
user	0m0.328s
sys	0m0.848s

1MB:
real	0m6.774s
user	0m0.252s
sys	0m0.280s

10MB:
real	0m5.689s
user	0m0.240s
sys	0m0.400s

(1 << 30):
real	0m5.586s
user	0m0.220s
sys	0m0.488s

Repeated runs deviate by about a second.

Bandwidth and latency affect a lot and I wonder if whatever the alternative is if it's better to pursue that instead.

/cc @afrind

afrind:
I don't know the curl or nghttp2 code, but I assume that when application pauses reading from a stream, curl stops releasing flow control for it.  The peer is still allowed to send up to a full stream flow control window.  There isn't a good way to revoke stream flow control credit that's already granted.

[...]

PR #7848, opened 2021-10-12
http2: add option CURLMOPT_STREAM_WINDOW_SIZE to set stream window size

steini2000:
When pausing a stream the data is still transferred and uses CPU time. On low-end CPUs this can consume a lot of CPU and time until the 32MB buffer is filled. Add an option the set the stream window size. Default is still 32MB.
This is the option that was saved for another day in Issue #4939.


bagder:
This is continued (and stalled) in #8263. Closing this.

PR #7856, opened 2021-10-14
openssl: with OpenSSL 1.1.0+ a failed RAND_status means goaway

### LINKED ISSUE #7840: openssl provider hangs curl
I built `curl` using a current openssl (3.x/master) and activated a (third-party) openssl [provider](https://www.openssl.org/docs/manmaster/man7/provider.html) by adding to openssl.cnf as follows:
```
oqsprovider = oqsprovider_sect

[oqsprovider_sect]
activate = 1
```

I expected simple commands like `curl https://www.google.com` to keep working.

What happened instead was a silent hang. Even adding `-vvv` didn't reveal anything.

When looking into the problem I admit to have not heeded the explicit warning in the `openssl.cnf` file:

```
# If no providers are activated explicitly, the default one is activated implicitly.
# See man 7 OSSL_PROVIDER-default for more details.
#
# If you add a section explicitly activating any other provider(s), you most
# probably need to explicitly activate the default provider, otherwise it
# becomes unavailable in openssl.  As a consequence applications depending on
# OpenSSL may not work correctly which could lead to significant system
# problems including inability to remotely access the system."
```

So, when activating the default provider via 
```
[default_sect]
activate = 1
```

everything worked OK.

I consider this silent hang a bug -- but am unsure as to whether it is a `curl` or `openssl` bug. When testing the same config file using `openssl s_client`, though, a tell-tale error message was output and the command terminated:

```
> openssl s_client -connect www.google.com:443
C021F46C7F7F0000:error:0A0000A1:SSL routines:SSL_CTX_new_ex:library has no ciphers:ssl/ssl_lib.c:3251:
```

-> Thus the issue is created here: `curl` in my eyes should issue a similar error message and not hang.

I did this on `master`:

curl 7.80.0-DEV (x86_64-pc-linux-gnu) libcurl/7.80.0-DEV OpenSSL/3.1.0 zlib/1.2.11
Release-Date: [unreleased]
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets

on 
Linux T430s 5.4.0-87-generic #98~18.04.1-Ubuntu SMP Wed Sep 22 10:45:04 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

bagder:
One reason we know it can fail is if a provider is used that doesn't do
a proper job.

Reported-by: Michael Baentsch
Fixes #7840

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -451,6 +451,7 @@ static CURLcode ossl_seed(struct Curl_easy *data)
   }
 #ifdef HAVE_RANDOM_INIT_BY_DEFAULT
   /* with OpenSSL 1.1.0+, a failed RAND_status is a showstopper */
+  failf(data, "Insufficient randomness");
   return CURLE_SSL_CONNECT_ERROR;
 #else
 


bagder:
I still think it was mostly correct but I've changed it to instead say:

> One reason we know it can fail is if a provider is used that doesn't do a proper job or is wrongly configured.

PR #7866, opened 2021-10-17
test1185: verify checksrc

bagder:


bagder:
This is a start. This does not yet test all style violations it can detect.

Issue #7877, opened 2021-10-19
SChannel: mem-leaks on failed InitializeSecurityContext()


gvanem:
When building with `-DCURLDEBUG=1`, there are some leaks in `vtls/schannel.c`
on *abnormal* protocol handshake situations. One is caused by the failure [here](https://github.com/curl/curl/blob/master/lib/vtls/schannel.c#L1146).

Verified with this `<curl_root>\src\schannel-leak-test.bat` file:

```
@echo off
setlocal
set CURL_SSL_BACKEND=schannel
curl.exe -vL http://www.hoyle.me.uk/pi/
perl ..\tests\memanalyze.pl -v c:\temp\memdebug.curl
```

Running it results in:
```
...

* STATE: PROTOCONNECT => PROTOCONNECTING handle 0xc7a5aa0; line 2000 (connection #1)
* schannel: SSL/TLS connection with www.hoyle.me.uk port 443 (step 2/3)
* schannel: encrypted data got 494
* schannel: encrypted data buffer: offset 494 length 4096
* schannel: next InitializeSecurityContext failed: SEC_E_INVALID_TOKEN (0x80090308) -
  Tokenet som ble levert til funksjonen, er ugyldig == "The token supplied to the function is invalid ".

* multi_done
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
* The cache now contains 1 members
* Closing connection 1
* Expire cleared (transfer 0xc7a5aa0)

Leak detected: memory still allocated: 8232 bytes
At c7ebae8, there's 4096 bytes.
 allocated by vtls/schannel.c:1060
At 48a6aa0, there's 24 bytes.
 allocated by vtls/schannel.c:708
At c7e9438, there's 16 bytes.
 allocated by vtls/schannel.c:935
At c7eaad0, there's 4096 bytes.
 allocated by vtls/schannel.c:1048
```

And using `wget  http://www.hoyle.me.uk/pi/` (w/OpenSSL) gives:
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="https://www.hoyle.me.uk/pi/">here</a>.</p>
<hr>
<address>Apache/2.4.46 (Debian) Server at www.hoyle.me.uk Port 80</address>
</body></html>
...
SSL handshake failed.
[...]

PR #7878, opened 2021-10-19
SChannel: plug mem-leaks

gvanem:
My attempt for a fix to my reported issue [7877](https://github.com/curl/curl/issues/7877).

And fix the `403 Forbidden` link at the same time:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa375924.aspx

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -719,8 +719,6 @@ schannel_acquire_credential_handle(struct Curl_easy *data,
   }
   BACKEND->cred->refcount = 1;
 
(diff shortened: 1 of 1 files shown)


gvanem:
What's needed to get this PR merged?

jay:
> What's needed to get this PR merged?

https://github.com/curl/curl/pull/7878#discussion_r735167976

Your fix moves the verbose message out of the shutdown function and only calls it for the first socket. If there is a second socket (FTPS) then that message should be shown as well.

gvanem:
I give up on this PR. Someone else with more understanding of the VTLS should take over. Closing.

PR #2290, opened 2018-02-05
openssl: Don't add verify locations when verifypeer==0

pschlan:
When peer verification is disabled, calling SSL_CTX_load_verify_locations
is not necessary. Only call it when verification is enabled to save
resources and increase performance.

PR #7892, opened 2021-10-22
openssl: if verifypeer is not requested, skip the CA loading

bagder:
It was previously done mostly to show a match/non-match in the verbose
output even when verification was not asked for. This change skips the
loading of the CA certs unless verifypeer is set to save memory and CPU.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -3066,60 +3066,36 @@ static CURLcode ossl_connect_step1(struct Curl_easy *data,
     }
   }
 
+  if(verifypeer && !imported_native_ca && (ssl_cafile || ssl_capath)) {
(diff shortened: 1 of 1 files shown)


sjadhavar:
This commit is causing the problem in our application that uses libcurl.

Please see the old comment for more details on this.
https://github.com/curl/curl/pull/2290#issuecomment-378230863

IMO, we should revert this commit.

bagder:
I think a start would be to file an issue explaining the problem.

pschlan:
Interesting (and glad) to see this has been re-added after my previous commit has been reverted. After my PR in #2290, I've added a note in https://github.com/pschlan/curl/blob/master/docs/libcurl/opts/CURLOPT_SSL_VERIFYPEER.3#L61-L68 to document the expensive loading of the ca file despite VERIFYPEER being disabled. Should we remove this note now?

bagder:
Oh nice catch @pschlan, yes we should. You wanna do it?

Issue #7906, opened 2021-10-25
curl no longer able to access localhost running Go server


vkuznet:
### I did this
After recent update of curl via MacPorts I found that it can't access my localhost host running Go server. The output I see is the following:
```
*   Trying 127.0.0.1:8989...
* Connected to localhost (127.0.0.1) port 8989 (#0)
> GET /api HTTP/1.1
> Host: localhost:8989
> User-Agent: curl/7.78.0
> Accept: */*
>
```
After that it hangs forever. I don't see that request arrive to my server. 

### I expected the following
If I use system curl then everything is fine, e.g.
```
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8989 (#0)
> GET /api HTTP/1.1
> Host: localhost:8989
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
...
```

### curl/libcurl version
The curl I get from MacPorts has the following info
```
/opt/local/bin/curl -V
curl 7.78.0 (x86_64-apple-darwin20.6.0) libcurl/7.78.0 OpenSSL/1.1.1l zlib/1.2.11 zstd/1.5.0 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.2)
Release-Date: 2021-07-21
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB PSL SSL TLS-SRP UnixSockets zstd
```
while my system provides me curl with this version:
```
/usr/bin/curl -V
curl 7.64.1 (x86_64-apple-darwin20.0) libcurl/7.64.1 (SecureTransport) LibreSSL/2.8.3 zlib/1.2.11 nghttp2/1.41.0
Release-Date: 2019-03-27
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: AsynchDNS GSS-API HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM NTLM_WB SPNEGO SSL UnixSockets
```

### operating system
macOS 11.6
```
uname -a
[...]

PR #7907, opened 2021-10-26
connect: connect to an ipv6 address first (happy eyeballs)

jay:
If a host's address list contains addresses from both families
(IPv4 and IPv6) and any IP version is allowed then make the first
connection to an IPv6 address.

"Happy Eyeballs" RFC says if an AAAA record (IPv6) is available, it
should always be tried first.

Prior to this change curl made the first connection to whichever address
family was in the first entry.

For example, consider a host with two entries, the first 127.0.0.1 IPv4
and the second [::1] IPv6:

--resolve "localhost:443:127.0.0.1,[::1]"

Before:
*   Trying 127.0.0.1:443...
(curl waits 200 ms for 127.0.0.1 before trying ::1)

After:
* family0 == v6, family1 == v4
*   Trying ::1:443...
(curl waits 200 ms for ::1 before trying 127.0.0.1)

Ref: https://datatracker.ietf.org/doc/html/rfc8305#section-3

Bug: https://github.com/curl/curl/issues/7906
Reported-by: Valentin Kuznetsov

Closes #xxxx

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -569,6 +569,18 @@ static struct Curl_addrinfo *ainext(struct connectdata *conn,
   return ai;
 }
 
+/* count of entries that have same tempfamily[tempindex] */
+static int aicount(struct connectdata *conn, int tempindex)
+{
+  int n = 0;
+  struct Curl_addrinfo *ai = conn->tempaddr[tempindex];
+  for(; ai; ai = ai->ai_next) {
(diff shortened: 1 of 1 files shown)


bagder:
Right, as getaddrinfo orders the results in the preferred order this PR seems unnecessary. Can we close it then?

PR #7690, opened 2021-09-09
CI: align test flags in various config files

mback2k:
1. Remove redundant parameters -a -p from target test-nonflaky.
2. Disable testing of SCP protocol on native Windows environments.
3. Use Makefile target to run tests in autotools builds on AppVeyor.
4. Don't use -vc parameter which is reserved for debugging.

Replaces #7591

--

The other things mentioned in #7591 will be done separately.

PR #7591, opened 2021-08-19
CI: align make and test flags in various config files

mback2k:
1. Remove redundant parameters -a -p from target test-nonflaky.
2. Disable testing of SCP protocol for all Windows environments.
3. Use Makefile target to run tests in autotools builds on AppVeyor.
4. Make sure we also compile the examples on all touched CIs.
5. Don't use -vc parameter which is reserved for debugging.

PR #6535, opened 2021-01-26
curl: make --remote-time error message respect --silent

bagder:
... and provide a little more details in the error messages.

Fixes #6533

Issue #6079, opened 2020-10-14
Azure DevOps shows GCC 10 compilation errors due to type mismatch between curl_off_t typeof and format


mback2k:
### I did this
#5891 tried to fix this previously as I spotted this locally. Now that GCC 10 is also being used in the MSYS2 based Azure DevOps builds, we are bitten by this again in CI:

Logs taken from 32-bit and 64-bit builds of today's release:
```
2020-10-14T09:24:38.3859321Z tool_filetime.c: In function 'setfiletime':
2020-10-14T09:24:38.3860145Z tool_filetime.c:106:15: error: ISO C does not support the 'I' printf flag [-Werror=format=]
2020-10-14T09:24:38.3860990Z   106 |               "Failed to set filetime %" CURL_FORMAT_CURL_OFF_T
2020-10-14T09:24:38.3862159Z       |               ^~~~~~~~~~~~~~~~~~~~~~~~~~
2020-10-14T09:24:38.3863740Z tool_filetime.c:106:15: error: format '%d' expects argument of type 'int', but argument 3 has type 'curl_off_t' {aka 'long long int'} [-Werror=format=]
2020-10-14T09:24:38.3864867Z In file included from ../include/curl/curl.h:38,
2020-10-14T09:24:38.3865679Z                  from ../lib/curl_setup.h:152,
2020-10-14T09:24:38.3866111Z                  from tool_setup.h:36,
2020-10-14T09:24:38.3866514Z                  from tool_filetime.h:24,
2020-10-14T09:24:38.3866881Z                  from tool_filetime.c:22:
2020-10-14T09:24:38.3867258Z ../include/curl/system.h:199:42: note: format string is defined here
2020-10-14T09:24:38.3867656Z   199 | #  define CURL_FORMAT_CURL_OFF_T     "I64d"
2020-10-14T09:24:38.3868391Z tool_filetime.c:125:17: error: ISO C does not support the 'I' printf flag [-Werror=format=]
2020-10-14T09:24:38.3868916Z   125 |                 "Failed to set filetime %" CURL_FORMAT_CURL_OFF_T
2020-10-14T09:24:38.3869346Z       |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~
2020-10-14T09:24:38.3870193Z tool_filetime.c:125:17: error: format '%d' expects argument of type 'int', but argument 3 has type 'curl_off_t' {aka 'long long int'} [-Werror=format=]
[...]

PR #7922, opened 2021-10-29
CI: build examples for additional code verification

mback2k:
Some CIs already build them, let's do it on more of them.

Follow up to #7690
Replaces #7591

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -83,7 +83,7 @@ stages:
     - script: ./buildconf && ./configure --enable-warnings --enable-werror $(configure)
       displayName: 'configure $(name)'
 
-    - script: make V=1 && cd tests && make V=1
+    - script: make V=1 && make V=1 examples && cd tests && make V=1
       displayName: 'compile'
       env:
         MAKEFLAGS: "-j 2"
@@ -191,7 +191,7 @@ stages:
     - script: $(container_cmd) -l -c "cd $(echo '%cd%') && ./buildconf && ./configure $(configure)"
       displayName: 'configure $(name)'
 
-    - script: $(container_cmd) -l -c "cd $(echo '%cd%') && make V=1 && cd tests && make V=1"
+    - script: $(container_cmd) -l -c "cd $(echo '%cd%') && make V=1 && make V=1 examples && cd tests && make V=1"
       displayName: 'compile'
       env:
         MAKEFLAGS: "-j 2"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,7 @@ commands:
   build:
     steps:
       - run: make V=1
+      - run: make V=1 examples
 
   test:
     steps:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -57,7 +57,7 @@ freebsd_task:
     #  esac
     - ./configure --prefix="${HOME}"/install --enable-debug --with-openssl --with-libssh2 --with-brotli --with-gssapi --with-libidn2 --enable-manual --enable-ldap --enable-ldaps --with-librtmp --with-libpsl --with-nghttp2 || { tail -300 config.log; false; }
   compile_script:
-    - make V=1 && cd tests && make V=1
+    - make V=1 && make V=1 examples && cd tests && make V=1
   test_script:
     # blackhole?
     - sysctl net.inet.tcp.blackhole
@@ -122,7 +122,7 @@ windows_task:
   configure_script: |
     %container_cmd% -l -c "cd $(echo '%cd%') && ./buildconf && ./configure %configure%"
   compile_script: |
-    %container_cmd% -l -c "cd $(echo '%cd%') && make V=1 && cd tests && make V=1"
+    %container_cmd% -l -c "cd $(echo '%cd%') && make V=1 && make V=1 examples && cd tests && make V=1"
   install_script: |
     %container_cmd% -l -c "cd $(echo '%cd%') && make V=1 install && PATH=/usr/bin:/bin find . -type f -path '*/.libs/*.exe' -print -execdir mv -t .. {} \;"
   test_script: |
--- a/docs/examples/anyauthput.c
+++ b/docs/examples/anyauthput.c
@@ -31,6 +31,11 @@
 
 #include <curl/curl.h>
 
+#if defined(__GNUC__) && defined(__MINGW32__)
+/* GCC 10 on mingw has issues with this, disable */
+#pragma GCC diagnostic ignored "-Wformat"
+#endif
+
 #ifdef WIN32
 #  include <io.h>
 #  define READ_3RD_ARG unsigned int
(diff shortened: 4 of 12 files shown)


mback2k:
Is Circle CI actually running for PRs? I can't find it in the listing below or am I just blind?

mback2k:
Quoting @jay from https://github.com/curl/curl/pull/7591#issuecomment-903308993 for further discussion:

> > do you have any idea why the examples using `fprintf` cannot compile on the failed Azure DevOps builds?
> 
> It looks like warnings are treated as errors but I don't know why it's an error for just the examples. I found [this post](https://curl-library.cool.haxx.narkive.com/QL3dOZrx/warning-iso-c-does-not-support-the-i64-ms-printf-length-modifier) from you many years ago, and maybe we can get around it with -Wno-pedantic-ms-format

Thanks. I just checked and found that `-Wno-pedantic-ms-format` is actually being used to build the examples. The error message is also a little different this time:

```
2021-10-29T20:45:35.3132082Z depbase=`echo anyauthput.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
2021-10-29T20:45:35.3137364Z i686-w64-mingw32-gcc -DHAVE_CONFIG_H   -I../../include -DCURL_NO_OLDIES  -isystem C:/msys64/mingw32/include -isystem C:/msys64/mingw32/include -isystem C:/msys64/mingw32/include  -Werror-implicit-function-declaration -g -O0 -pedantic -Wall -W -Wpointer-arith -Wwrite-strings -Wunused -Wshadow -Winline -Wnested-externs -Wmissing-declarations -Wmissing-prototypes -Wno-long-long -Wbad-function-cast -Wfloat-equal -Wno-multichar -Wsign-compare -Wundef -Wno-format-nonliteral -Wendif-labels -Wstrict-prototypes -Wdeclaration-after-statement -Wold-style-definition -Wstrict-aliasing=3 -Wcast-align -Wtype-limits -Wold-style-declaration -Wmissing-parameter-type -Wempty-body -Wclobbered -Wignored-qualifiers -Wconversion -Wno-sign-conversion -Wvla -ftree-vrp -Wno-pedantic-ms-format -Wdouble-promotion -Wformat=2 -Warray-bounds=2 -Wshift-negative-value -Wshift-overflow=2 -Wnull-dereference -fdelete-null-pointer-checks -Wduplicated-cond -Wunused-const-variable -Wduplicated-branches -Wrestrict -Walloc-zero -Wformat-overflow=2 -Wformat-truncation=2 -Wimplicit-fallthrough=4 -Wno-system-headers -Wenum-conversion -Werror -pedantic-errors -MT anyauthput.o -MD -MP -MF $depbase.Tpo -c -o anyauthput.o anyauthput.c &&\
2021-10-29T20:45:35.3148223Z mv -f $depbase.Tpo $depbase.Po
```
```
2021-10-29T20:45:36.3328869Z anyauthput.c: In function 'read_callback':
2021-10-29T20:45:36.3329906Z anyauthput.c:93:19: error: ISO C does not support the 'I' printf flag [-Werror=format=]
2021-10-29T20:45:36.3330688Z    93 |   fprintf(stderr, "*** We read %" CURL_FORMAT_CURL_OFF_T
2021-10-29T20:45:36.3331692Z       |                   ^~~~~~~~~~~~~~~
2021-10-29T20:45:36.3332535Z anyauthput.c:93:19: error: format '%d' expects argument of type 'int', but argument 3 has type 'curl_off_t' {aka 'long long int'} [-Werror=format=]
2021-10-29T20:45:36.3334204Z In file included from ../../include/curl/curl.h:35,
2021-10-29T20:45:36.3334847Z                  from anyauthput.c:32:
2021-10-29T20:45:36.3335536Z ../../include/curl/system.h:199:42: note: format string is defined here
2021-10-29T20:45:36.3336240Z   199 | #  define CURL_FORMAT_CURL_OFF_T     "I64d"
2021-10-29T20:45:36.3949522Z cc1.exe: all warnings being treated as errors
2021-10-29T20:45:36.4188318Z make[2]: *** [Makefile:1813: anyauthput.o] Error 1
2021-10-29T20:45:36.4189142Z make[2]: *** Waiting for unfinished jobs....
2021-10-29T20:45:36.6744255Z make[2]: Leaving directory '/c/__w/1/s/docs/examples'
2021-10-29T20:45:36.6756852Z make[1]: *** [Makefile:1928: check-am] Error 2
2021-10-29T20:45:36.6759935Z make[1]: Leaving directory '/c/__w/1/s/docs/examples'
2021-10-29T20:45:36.6848015Z make: *** [Makefile:1741: examples] Error 2
2021-10-29T20:45:36.7881488Z ##[error]Cmd.exe exited with code '2'.
2021-10-29T20:45:36.8831887Z ##[section]Finishing: compile
```
So the error message mentions `'%d'` , but then points to a definition of `CURL_FORMAT_CURL_OFF_T` as `"I64d"`.

I think something is off with how the typedef's are done if used from the examples. This looks like a mismatch to me.

mback2k:
Another interesting find: the error message does not appear in the Cirrus Windows builds which are using the same Docker images as the Azure Windows builds, therefore use the identical toolchain. The only difference is that the Cirrus Windows builds are release and not debug builds.

jay:
This looks like #6079, which was solved by ignoring `-Wformat`. If it's a bug in gcc 10 then one of us should report it. I don't have mingw-w64 gcc 10 but I just tried mingw-w64 gcc 9 and I don't get the error unless I exclude `-Wno-pedantic-ms-format`.

mback2k:
Thanks @jay. Let's see what happens if I put the same workaround into the examples, although I am still wondering why the Cirrus CI builds are not affected without it.

It also looks like this is mostly affecting external code, because curl internal code mostly uses the `curlx_*` versions of `printf*`. For example `tool_writeout.c` uses curlx as a workaround while `tool_filetime.c` uses @bagder pragma-based workaround. I guess a follow up PR could be to make `tool_filetime.c` also use curlx instead, right?

mback2k:
Regarding the follow up PR mentioned above. I just [created a draft](https://github.com/curl/curl/pull/7941) one, since the pragma is now obsolete due to #6535.

mback2k:
@bagder I don't have the time now, but you may wanna investigate this error that shows up in AppVeyor for master and this PR:
```
3>c:\projects\curl\src\tool_operate.c(889) : error C2220: warning treated as error - no 'object' file generated
3>c:\projects\curl\src\tool_operate.c(889) : warning C4701: potentially uninitialized local variable 'per' use
```

mback2k:
@bagder please confirm with another approval if you are fine with adding the pragma to the relevant examples, thanks.

bagder (commented):
I think the added pragma lines are unfortunate since the examples are meant to contains as little "extra fluff" as possible. I assume you haven't figured out any other way to work around the problems you get without them? Can we remove/change the printf lines instead?

mback2k:
@bagder I tried to update the relevant `print`f lines by using `unsigned long` instead of `curl_off_t`.

mback2k:
Rebased again after @bagder simplified `progressfunc.c` with e75bc3ae3ff695d51350d9eeb20742ad0abcf8c3.

bagder (approved):

PR #7928, opened 2021-10-30
manpages: Replace acute accent by single quote

sergiodj:
Hello,

This issue was flagged by Debian's lintian while preparing the package
there.

One should use the \(aq sequence when printing a single quote on a
manpage.  curl's manpages are currently using \', which will render as
an acute accent.  This commit fixes it.

### DIFF
--- a/docs/cmdline-opts/cacert.d
+++ b/docs/cmdline-opts/cacert.d
@@ -16,7 +16,7 @@ set, and uses the given path as a path to a CA cert bundle. This option
 overrides that variable.
 
 The windows version of curl will automatically look for a CA certs file named
-\'curl-ca-bundle.crt\', either in the same directory as curl.exe, or in the
+\(aqcurl-ca-bundle.crt\(aq, either in the same directory as curl.exe, or in the
 Current Working Directory, or in any folder along your PATH.
 
 If curl is built against the NSS SSL library, the NSS PEM PKCS#11 module
--- a/docs/cmdline-opts/pinnedpubkey.d
+++ b/docs/cmdline-opts/pinnedpubkey.d
@@ -10,7 +10,7 @@ Added: 7.39.0
 Tells curl to use the specified public key file (or hashes) to verify the
 peer. This can be a path to a file which contains a single public key in PEM
 or DER format, or any number of base64 encoded sha256 hashes preceded by
-\'sha256//\' and separated by \';\'
+\(aqsha256//\(aq and separated by \(aq;\(aq
 
 When negotiating a TLS or SSL connection, the server sends a certificate
 indicating its identity. A public key is extracted from this certificate and
--- a/docs/cmdline-opts/proxy-pinnedpubkey.d
+++ b/docs/cmdline-opts/proxy-pinnedpubkey.d
@@ -10,7 +10,7 @@ Added: 7.59.0
 Tells curl to use the specified public key file (or hashes) to verify the
 proxy. This can be a path to a file which contains a single public key in PEM
 or DER format, or any number of base64 encoded sha256 hashes preceded by
-\'sha256//\' and separated by \';\'
+\(aqsha256//\(aq and separated by \(aq;\(aq
 
 When negotiating a TLS or SSL connection, the server sends a certificate
 indicating its identity. A public key is extracted from this certificate and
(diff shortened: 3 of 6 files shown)


bagder:
According to https://www.systutorials.com/docs/linux/man/7-groff/, the single quote is `\(cq`. Can you point to a reference for `\(aq` ?

I would like to make sure roffit supports this first before we merge this so that the man page still renders correctly on the website.

Also, I think maybe we can leave the quotes in the `*.d` files and instead have them replaced when generating the man page, like this:
~~~diff
--- a/docs/cmdline-opts/gen.pl
+++ b/docs/cmdline-opts/gen.pl
@@ -97,10 +97,13 @@ sub printdesc {
             }
         }
         # quote "bare" minuses in the output
         $d =~ s/( |\\fI|^)--/$1\\-\\-/g;
         $d =~ s/([ -]|\\fI|^)-/$1\\-/g;
+        # handle singe and double quotes
+        $d =~ s/\'/\\(aq/g;
+        $d =~ s/\"/\\(dq/g;
         print $d;
     }
 }
 
 sub seealso {

bagder:
I found a reference here: https://manpages.ubuntu.com/manpages/bionic/man7/groff.7.html

but also: `\(aq` isn't portable it seems? https://lists.gnu.org/archive/html/groff/2009-02/msg00013.html

sergiodj:
> I found a reference here:
> https://manpages.ubuntu.com/manpages/bionic/man7/groff.7.html
> 
> but also: `\(aq` isn't portable it seems?
> https://lists.gnu.org/archive/html/groff/2009-02/msg00013.html

Thanks for the reply, Daniel.

Yeah, it seems `\(aq` isn't portable, but then again it seems like
there is no portable way to represent single quotes.

I understand your concern about portability here, and I'm definitely
not an expert in this subject, but I *think* it should be somewhat
safe to assume that people will be using groff to generate the
manpage, and there we can be sure that `\(aq` will work.

sergiodj:
> According to https://www.systutorials.com/docs/linux/man/7-groff/, the
> single quote is `\(cq`. Can you point to a reference for `\(aq` ?
> 
> I would like to make sure roffit supports this first before we merge
> this so that the man page still renders correctly on the website.
> 
> Also, I think maybe we can leave the quotes in the `*.d` files and
> instead have them replaced when generating the man page, like this:
> ~~~diff --- a/docs/cmdline-opts/gen.pl +++ b/docs/cmdline-opts/gen.pl
> @@ -97,10 +97,13 @@ sub printdesc { } } # quote "bare" minuses in the
> output $d =~ s/( |\\fI|^)--/$1\\-\\-/g; $d =~ s/([
> -]|\\fI|^)-/$1\\-/g;
> +        # handle singe and double quotes
> +        $d =~ s/\'/\\(aq/g;
> +        $d =~ s/\"/\\(dq/g; print $d; } }
>  
>  sub seealso {

Thanks for the review.

I liked the idea of replacing these characters in the script that's
generating the manpage.  It had occurred to me while I was writing the
patch, but it was easier to just replace everything with `\(aq` TBH.

Let me know if you're planning to check this in and I'll gladly close
this PR.

Thanks.

sergiodj:
> Let me know if you're planning to check this in and I'll gladly close
> this PR.

Ah, just a quick comment: if you decide to go this route, we should
un-escape the single quotes (i.e., remove the backslashes) from the
`*.d` files.

bagder:
> I liked the idea of replacing these characters in the script that's generating the manpage. It had occurred to me while I was writing the patch, but it was easier to just replace everything with `\(aq` TBH.

Easier yes, but the texts become unreadable and harder to manage.

> Let me know if you're planning to check this in and I'll gladly close this PR.

The script only generates the curl.1 man page. All the .3 man pages are still written and provided as-is so we still need those addressed. And having `\(aq` instead of single quotes in there is problematic I think. I don't have any better fix in mind yet though...

>  if you decide to go this route, we should un-escape the single quotes (i.e., remove the backslashes) from the
*.d files IMO.

I agree.

bagder:
We only actually need to use `\(aq` if the single quote is the first character of a line, right?

bagder:
I merged #7933 and #7934 as alternatives to your PR. Thanks a lot for reporting. Is there now anything left for this PR to do?

sergiodj:
> I merged #7933 and #7934 as alternatives to your PR. Thanks a lot for reporting. Is there now anything left for this PR to do?

Thank you, Daniel.  Everything looks good, so I'm closing this PR.

PR #7929, opened 2021-10-31
ngtcp2: Specify the missing required callback functions

tatsuhiro-t:
This PR adds the missing ngtcp2 required callback functions.

cb_rand is meant not to fail, because the weak pseudo random number is just fine.
Curl_rand is a strong random data generator and may fail.  In this PR, I just fill 0 in case of failure.
You might not like it though.


bagder:
> Curl_rand is a strong random data generator and may fail. In this PR, I just fill 0 in case of failure. You might not like it though.

I don't really have any better suggestion. It seems highly unlikely to ever actually occur in a working situation.

bagder:
Thanks!

PR #7930, opened 2021-10-31
docs: reduce English contractions

bagder:
You're => You are
Hasn't => Has not
Doesn't => Does not
Don't => Do not
You'll => You will
etc

### DIFF
--- a/README
+++ b/README
@@ -40,7 +40,7 @@ GIT
 
     git clone https://github.com/curl/curl.git
 
-  (you'll get a directory named curl created, filled with the source code)
+  (you will get a directory named curl created, filled with the source code)
 
 SECURITY PROBLEMS
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ To download the very latest source from the Git server do this:
 
     git clone https://github.com/curl/curl.git
 
-(you'll get a directory named curl created, filled with the source code)
+(you will get a directory named curl created, filled with the source code)
 
 ## Security problems
 
--- a/docs/BUG-BOUNTY.md
+++ b/docs/BUG-BOUNTY.md
@@ -35,7 +35,7 @@ Check out the current award amounts at [https://hackerone.com/curl](https://hack
 # Who is eligible for a reward?
 
 Everyone and anyone who reports a security problem in a released curl version
-that hasn't already been reported can ask for a bounty.
+that has not already been reported can ask for a bounty.
 
 Vulnerabilities in features that are off by default and documented as
 experimental are not eligible for a reward.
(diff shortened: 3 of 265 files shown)


jay:
Why? Contractions are fine.

bagder:
To make the language less informal. To make it easier to search for phrases (due to less single-quotes).

jay:
I don't think this makes things any clearer, or any less clear. However apostrophes are common and used all the time in English. It seems overly burdensome to avoid them.

bagder:
Different sources say different things about contraction use in English. There's the (old?) school that says  contractions are only for informal texts and not for technical or academic texts. There's also several sources that insist that especially non-native English readers understand texts without contractions easier.

I checked, and my impression is that man pages in general (I checked random samples from my Debian install) tend to go easy on contractions.

If you feel strongly that contractions improve the texts, then I won't push for removing them.

bagder:
I fired off [this twitter poll](https://twitter.com/bagder/status/1455659677400829955), and it certainly shows that there's no general consensus...

navybrother23:
I know this PR is scoped to docs and is meant to "reduce" not to "remove", but I found one additional contraction in the root level README in case you wanted to add it to this PR

jay:
> If you feel strongly that contractions improve the texts, then I won't push for removing them.

I don't feel they're an improvement or a retreat. In other words I don't think this is better or worse than what's there now. If someone submitted `don't` or `do not` either is acceptable to me. If I have any objection it's to where this could be headed, with banning the apostrophe (ie banning contractions) by adding it to checksrc or the guidelines.

bagder:
This PR is explicitly about *reducing* contractions in the existing documentation. It's not about banning them and not even about removing them completely - the latter mostly because I'm sure we'll miss a few and I bet there are or can be situations where the language flow is improved by contractions even if most of them are removed.

bagder:
Screenshot 2021-11-04 at 08-34-45 Daniel 🥌 Stenberg on Twitter

... not sure what to do with this! :grinning:

jzakrzewski:
My personal taste is with contractions. However, from what I read on the Internet (that's a hell of an oracle, isn't it?) in formal writing contractions are recommended to be avoided.
From the other hand it all depends on how formal we want the documentation to be.
Now, when I read rfc7231, there are only few contractions and most of the "not"s are spelled out in full.

All in all I don't think it's worth the effort to change existing text just to avoid contractions. Except you already did it, so we can very well adopt it. I'm strongly against a hard requirement on not using the contractions at all, however.

Just my mustard to the topic ;)

bagder:
Thanks for your feedback! I'll leave this PR open for a few days more, but unless someone actually objects to it I will  proceed and merge.

PR #7938, opened 2021-11-01
If query replaces the long switch case with the same function

EinfachBeez:
I'm happy to be proven wrong, but the switch case does exactly the same thing as this short if query. Of course, this would also be possible via switch, but would not be as nice in my opinion.
e.g.
```
case 'a': case 'b': etc...
        return (char) in - ' ';
```

Even after short performance tests, this solution is faster than using the long switch case.

### DIFF
--- a/lib/strcase.c
+++ b/lib/strcase.c
@@ -36,60 +36,8 @@ char Curl_raw_toupper(char in)
   if(in >= 'a' && in <= 'z')
     return (char)('A' + in - 'a');
 #else
-  switch(in) {
-  case 'a':
-    return 'A';
-  case 'b':
-    return 'B';
-  case 'c':
-    return 'C';
-  case 'd':
-    return 'D';
-  case 'e':
-    return 'E';
-  case 'f':
-    return 'F';
-  case 'g':
-    return 'G';
-  case 'h':
-    return 'H';
-  case 'i':
-    return 'I';
-  case 'j':
-    return 'J';
-  case 'k':
-    return 'K';
-  case 'l':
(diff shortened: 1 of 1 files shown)


bagder:
"remember EBCDIC" as the comment says. Did you test this on EBCDIC or even with a system that needs/uses `CURL_DOES_CONVERSIONS`?

Timeox2k:
In fact, as far as I can tell from my experience, it should also improve performance and speed up the process in general.

The only thing I see is that the query (at least as it seems) exists several times in a row, but this can also be only an optical illusion by the commit.

bagder:
> it should also improve performance and speed up the process in general.

What's the "it" referring to in this sentence? Are you talking about the commit in an EBCDIC scenario?

Timeox2k:
By "it" I meant switching from switch to if. But it could also be that I am wrong. I know that an if else in a ladder is slower than a switch, but since this is only a single if it should be faster. But I am not sure how the compiler handles it in c.

bagder:
Have you looked at an EBCDIC character table and then looked back at this PR again?

EinfachBeez:
Yes I have noticed the problem. But the question is if an ascii to ebcdic converter would be worth it?

bagder:
I don't understand the question. Are you saying you want to break the functionality to make the code faster for a build setup you don't use yourself?

bagder:
The existing code *works with* EBCDIC, it also doesn't assume EBCDIC so it works for ascii platforms too that want the run-time conversions.

PR #7935, opened 2021-11-01
tests/smbserver.py: fix compatibility with impacket 0.9.23+

mback2k:
impacket now performs sanity checks if the requested and to
be served file path actually is inside the real share path.

Ref: https://github.com/SecureAuthCorp/impacket/pull/1066

Fixes #7924

PR #7602, opened 2021-08-21
Minor script cleanups

a1346054:

PR #7940, opened 2021-11-01
tests/*.py: support Python 2 again while preferring Python 3

mback2k:
This makes it possible to use the Python-based test servers
again in legacy environments which do not have modern Python.

Follow up to #7602 which broke Python 2 support
Follow up to #7935 which fixed impacket support

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -209,14 +209,14 @@ stages:
           name: 32-bit OpenSSL/libssh2
           container_img: ghcr.io/mback2k/curl-docker-winbuildenv/msys2-mingw32:ltsc2019
           container_cmd: C:\msys64\usr\bin\sh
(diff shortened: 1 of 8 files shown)


bagder:
Is this really worth the struggle? Isn't it better to make sure we just skip the tests that need python3 if that isn't available?

danielgustafsson:
> On 6 Dec 2021, at 16:51, Daniel Stenberg ***@***.***> wrote:

> Is this really worth the struggle? Isn't it better to make sure we just skip the tests that need python3 if that isn't available?

If we can enumerate the affected environments it might be easier to judge?

mback2k:
All the Windows builds on Azure and Cirrus CIs are affected.

bagder:
@mback2k did you have any particular platforms or targets in mind when you did this PR?

mback2k:
> All the Windows builds on Azure and Cirrus CIs are affected.

Not actually correct, at least the msys2-based builds already support Python 3. Therefore closing this in favor of #8544.

Issue #7942, opened 2021-11-01
Unexpected behaviour when combing --create-dirs and --etag-compare/--etag-save


Earnestly:
### I did this
With the following config file:
```
url https://www.youtube.com/feeds/videos.xml?channel_id=UCl9OJE9OpXui-gRsnWjSrlA
etag-save bar/etag
etag-compare bar/etag
output bar/xml
next
url https://www.youtube.com/feeds/videos.xml?channel_id=UC7_gcs09iThXybpVgjHZ_7g
etag-save foo/etag
etag-compare foo/etag
output foo/xml
```

I used:

```
% mkdir bar
% curl --create-dirs -ZK - < config
Warning: Failed to open foo/etag
curl: (23) Failed writing received data to disk/application
curl: (23) Failed writing received data to disk/application
```

### I expected the following

It appears that all multiplexed connections to youtube will fail if the parent directory of the `etag` file doesn't exist for one of the config sections.

The most ideal result would be that curl would create the leading directory for the etag file just as it would for `--output`.

### curl/libcurl version
```
curl 7.79.1 (x86_64-pc-linux-gnu) libcurl/7.79.1 OpenSSL/1.1.1l zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.46.0
Release-Date: 2021-09-22
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets zstd
```

### operating system
```
Linux teapot 5.14.14-arch1-1 #1 SMP PREEMPT Wed, 20 Oct 2021 21:35:18 +0000 x86_64 GNU/Linux
```

dfandrich:
--create-dirs is clearly documented to NOT create such directories, and I'm
personally hesitant to make it do so. There are a number of config files that
curl uses like this one, and in most cases they will exist before curl starts.
There are probably also people who rely on the current error behaviour to catch
typos, etc.  The difference between config files and data files is a big one,
as typically a single configuration is used to download multiple files, making
[...]

PR #7943, opened 2021-11-02
tool_operate: improve the error message when etags can't be saved

bagder:
Reported-by: Earnestly on github
Ref: #7942


bagder:
Replaced by #7945

PR #7945, opened 2021-11-02
tool_operate: don't make a failed etag save a fatal error

bagder:
When failing to create the output file for saving an etag, only fail
that particular single transfer and allow others to follow.

In a serial transfer setup, if no transfer at all is done due to them
all being skipped because of this error, curl will output an error
message and return exit code 26.

Reported-by: Earnestly on github
Ref: #7942


dfandrich:
Looks like this change is causing torture test failures in tests 343 344 345 347,

bagder:
Thanks, will fix!

PR #7948, opened 2021-11-03
vtls/rustls: update to compile with rustls-ffi tip

### LINKED ISSUE #7947: curl tip does not compile with rustls-ffi tip
Using rustls-ffi commit 60d9c2adcbe26cbcb3f6ee4c32cc79ca85c1b9c1, and curl commit 90e74206b98180b7a185d83d6d20071190e2df59, I get the following compile errors:

```
vtls/rustls.c: In function 'cr_init_backend':
vtls/rustls.c:312:3: error: implicit declaration of function 'rustls_client_config_builder_set_protocols'; did you mean 'rustls_client_config_builder_set_alpn_protocols'? [-Werror=implicit-function-declaration]
  312 |   rustls_client_config_builder_set_protocols(config_builder, alpn, 2);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |   rustls_client_config_builder_set_alpn_protocols
vtls/rustls.c:312:3: warning: nested extern declaration of 'rustls_client_config_builder_set_protocols' [-Wnested-externs]
vtls/rustls.c: In function 'cr_set_negotiated_alpn':
vtls/rustls.c:360:3: error: implicit declaration of function 'rustls_connection_get_alpn_protocol'; did you mean 'rustls_connection_alpn_protocol'? [-Werror=implicit-function-declaration]
  360 |   rustls_connection_get_alpn_protocol(rconn, &protocol, &len);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |   rustls_connection_alpn_protocol
vtls/rustls.c:360:3: warning: nested extern declaration of 'rustls_connection_get_alpn_protocol' [-Wnested-externs]
vtls/rustls.c: At top level:
vtls/rustls.c:553:3: warning: initialization of 'size_t (*)(char *, size_t)' {aka 'long unsigned int (*)(char *, long unsigned int)'} from incompatible pointer type 'struct rustls_str (*)(void)' [-Wincompatible-pointer-types]
  553 |   rustls_version,                  /* version */
      |   ^~~~~~~~~~~~~~
vtls/rustls.c:553:3: note: (near initialization for 'Curl_ssl_rustls.version')
```

I believe that these are due to a recent refactoring in rustls-ffi. I will submit a PR shortly, but wanted to create a tracking issue first.

I believe that we don't currently have a compatibility guarantee with rustls/rustls-ffi besides "you should target the most recent commit" so that's what I'll target as well.

CC @jsha

kevinburke:
Some method names were changed in a recent refactoring. We have no
compatibility guarantee with rustls currently, so rustls-ffi tip is
probably the best commit to target.

Fixes #7947.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -672,7 +672,7 @@ if test X"$want_hyper" != Xno; then
     LIB_HYPER="-lhyper -ldl -lpthread -lm"
     if test X"$want_hyper" != Xdefault; then
       CPP_HYPER=-I"$want_hyper_path/capi/include"
-      LD_HYPER="-L$want_hyper_path/target/debug"
+      LD_HYPER="-L$want_hyper_path/target/release -L$want_hyper_path/target/debug"
     fi
   fi
   if test -n "$LIB_HYPER"; then
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -165,12 +165,13 @@ cr_recv(struct Curl_easy *data, int sockindex,
       *err = CURLE_OK;
       return 0;
     }
-    else if(rresult != RUSTLS_RESULT_OK) {
+    else if(rresult != RUSTLS_RESULT_OK &&
+            rresult != RUSTLS_RESULT_PLAINTEXT_EMPTY) {
       failf(data, "error in rustls_connection_read");
       *err = CURLE_READ_ERROR;
       return -1;
     }
-    else if(n == 0) {
+    else if(n == 0 || rresult == RUSTLS_RESULT_PLAINTEXT_EMPTY) {
       /* rustls returns 0 from connection_read to mean "all currently
         available data has been read." If we bring in more ciphertext with
         read_tls, more plaintext will become available. So don't tell curl
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -544,6 +544,12 @@ cr_close(struct Curl_easy *data, struct connectdata *conn,
   }
 }
 
+static size_t cr_version(char *buffer, size_t size)
+{
+  struct rustls_str ver = rustls_version();
+  return msnprintf(buffer, size, "%s", ver.data);
+}
+
 const struct Curl_ssl Curl_ssl_rustls = {
   { CURLSSLBACKEND_RUSTLS, "rustls" },
   SSLSUPP_TLS13_CIPHERSUITES,      /* supports */
@@ -551,7 +557,7 @@ const struct Curl_ssl Curl_ssl_rustls = {
 
   Curl_none_init,                  /* init */
   Curl_none_cleanup,               /* cleanup */
-  rustls_version,                  /* version */
+  cr_version,                      /* version */
   Curl_none_check_cxn,             /* check_cxn */
   Curl_none_shutdown,              /* shutdown */
   cr_data_pending,                 /* data_pending */
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -27,7 +27,7 @@
 #include "curl_printf.h"
 
 #include <errno.h>
-#include <crustls.h>
+#include <rustls.h>
 
 #include "inet_pton.h"
 #include "urldata.h"
@@ -309,10 +309,10 @@ cr_init_backend(struct Curl_easy *data, struct connectdata *conn,
   config_builder = rustls_client_config_builder_new();
 #ifdef USE_HTTP2
   infof(data, "offering ALPN for HTTP/1.1 and HTTP/2");
-  rustls_client_config_builder_set_protocols(config_builder, alpn, 2);
+  rustls_client_config_builder_set_alpn_protocols(config_builder, alpn, 2);
 #else
   infof(data, "offering ALPN for HTTP/1.1 only");
-  rustls_client_config_builder_set_protocols(config_builder, alpn, 1);
+  rustls_client_config_builder_set_alpn_protocols(config_builder, alpn, 1);
 #endif
   if(!verifypeer) {
     rustls_client_config_builder_dangerous_set_certificate_verifier(
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -161,7 +161,7 @@ cr_recv(struct Curl_easy *data, int sockindex,
       (uint8_t *)plainbuf + plain_bytes_copied,
       plainlen - plain_bytes_copied,
       &n);
-    if(rresult == RUSTLS_RESULT_ALERT_CLOSE_NOTIFY) {
+    if(n == 0) {
       *err = CURLE_OK;
       return 0;
     }
@@ -171,11 +171,7 @@ cr_recv(struct Curl_easy *data, int sockindex,
       *err = CURLE_READ_ERROR;
       return -1;
     }
-    else if(n == 0 || rresult == RUSTLS_RESULT_PLAINTEXT_EMPTY) {
-      /* rustls returns 0 from connection_read to mean "all currently
-        available data has been read." If we bring in more ciphertext with
-        read_tls, more plaintext will become available. So don't tell curl
-        this is an EOF. Instead, say "come back later." */
+    else if(rresult == RUSTLS_RESULT_PLAINTEXT_EMPTY) {
       infof(data, "cr_recv got 0 bytes of plaintext");
       backend->data_pending = FALSE;
       break;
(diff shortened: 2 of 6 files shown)


bagder:
It would be good if `scripts/zuul/before_script.sh` was also updated to use the latest and thus verify changes like this!

kevinburke:
Jacob commented on #7947 that he'd prefer to wait until we do a rustls-ffi release so I think I'll wait on that and then update the RUSTLS_VERSION in zuul.d/jobs.yaml to match.

kevinburke:
Test 565 failed in AppVeyor, not clear that it's related to this patch.

bagder:
> Test 565 failed in AppVeyor, not clear that it's related to this patch.

It's not. The appveyor CI builds are a bit flaky.

bagder:
I get this error trying this PR with latest crustls:
~~~
  RUN      checksrc
vtls/rustls.c:553:3: error: initialization of ‘size_t (*)(char *, size_t)’ {aka ‘long unsigned int (*)(char *, long unsigned int)’} from incompatible pointer type ‘struct rustls_str (*)(void)’ [-Wincompatible-pointer-types]
  553 |   rustls_version,                  /* version */
      |   ^~~~~~~~~~~~~~
vtls/rustls.c:553:3: note: (near initialization for ‘Curl_ssl_rustls.version’)
~~~

The crustls header describing rustls_version is now clearly not accurate. [I filed an issue](https://github.com/rustls/rustls-ffi/issues/168).

bagder:
> Jacob commented on #7947 that he'd prefer to wait until we do a rustls-ffi release so I think I'll wait on that and then update the RUSTLS_VERSION in zuul.d/jobs.yaml to match.

But that puts us in a weird situation, right? If we merge this code we make it work with current crustls but the CI job will fail. Not ideal.

kevinburke:
> But that puts us in a weird situation, right? If we merge this code we make it work with current crustls but the CI job will fail. Not ideal.

I guess what I had in mind was wait on merging this PR until rustls-ffi 0.8.0 or whatever is released, then update zuul.d/jobs.yml as part of this PR, verify they both pass, then merge. Maybe I'm not understanding correctly.

jsha:
I'm on the same page as @kevinburke. I'm planning to do the rustls-ffi 0.8.0 release very soon - in the next few days - and then we should update this PR (or start a new one) such that docs/RUSTLS.md and scripts/zuul/before_script.sh get updated to use the new version.

bagder:
I'm fine with that approach if you are, given that you're the ones doing the most work on this. I'm just used to working with the bleeding edge of in-progress libraries in other cases (quiche, ngtcp2, hyper etc) so crustls sticks out a little here in that we need to stick to a fixed tag in order to build the current curl tip.

jsha:
>  I'm just used to working with the bleeding edge of in-progress libraries in other cases (quiche, ngtcp2, hyper etc) so crustls sticks out a little here in that we need to stick to a fixed tag in order to build the current curl tip.

Doesn't this mean when those libraries commit breaking changes to main, curl's integration tests immediately break?

There's value in sticking to a similar workflow here, but potentially breaking curl's integration tests by pushing to crustls main seems like a big disadvantage. Maybe I'm misunderstanding.

jsha:
[rustls-ffi v0.8.0 is now released](https://github.com/rustls/rustls-ffi/releases/tag/v0.8.0). Thanks for updating this branch, @kevinburke!

We still need to deal with these two items from the change log:

> Reading of plaintext from a rustls_connection. When the internal plaintext buffer is empty, rustls_connection_read will return RUSTLS_RESULT_PLAINTEXT_EMPTY. That means no more plaintext can be read until additional TLS bytes are ingested via rustls_connection_read_tls, and rustls_connection_process_new_packets is called. Previously this condition was indicated by returning RUSTLS_RESULT_OK with out_n set to 0.

> Handling of unclean close and the close_notify TLS alert. Mirroring upstream changes, a rustls_connection now tracks TCP closed state like so: rustls_connection_read_tls considers a 0-length read from its callback to mean "TCP stream was closed by peer." If that happens before the peer sent close_notify, rustls_connection_read will return RUSTLS_RESULT_UNEXPECTED_EOF once the available plaintext bytes are exhausted. This is useful to protect against truncation attacks. Note: some TLS implementations don't send close_notify. If you are already getting length information from your protocol (e.g. Content-Length in HTTP) you may choose to ignore UNEXPECTED_EOF so long as the number of plaintext bytes was as expected.

Here's the relevant portion of rustls.c: https://github.com/curl/curl/blob/b3e1ed316062b7984a4a29e7b599f9b8640c7749/lib/vtls/rustls.c#L159-L186

Essentially we want to change the cases for RUSTLS_RESULT_ALERT_CLOSE_NOTIFY and n == 0.

When n == 0, we want to treat that as a clean close and do what we currently do for RUSTLS_RESULT_ALERT_CLOSE_NOTIFY: set `*err` to CURLE_OK, and return 0.

When result == RUSTLS_RESULT_PLAINTEXT_EMPTY, we want to do what we currently do for n == 0: set `backend->data_pending = FALSE;` and break.

Because of the above-mentioned changes in the change log, we no longer expect to get RUSTLS_RESULT_ALERT_CLOSE_NOTIFY in this function (or ever), so we won't have a case specially handling that error anymore.

kevinburke:
> When result == RUSTLS_RESULT_PLAINTEXT_EMPTY, we want to do what we currently do for n == 0: set backend->data_pending = FALSE; and break.

I handled this case in #7950 but I will pull that in here. I will make the other changes as well

lib/vtls/rustls.c
```diff
@@ -161,20 +161,17 @@ cr_recv(struct Curl_easy *data, int sockindex,
       (uint8_t *)plainbuf + plain_bytes_copied,
       plainlen - plain_bytes_copied,
       &n);
-    if(rresult == RUSTLS_RESULT_ALERT_CLOSE_NOTIFY) {
+    if(n == 0) {
```
jsha, line 164: We discussed this a little out-of-band, but writing it down here:

The check for `n == 0` has to come last. If the result was not `RUSTLS_RESULT_OK`, n was not written to.

The check for `RUSTLS_RESULT_PLAINTEXT_EMPTY` should come first. That way you don't have to repeat it in two `if` stanzas.

kevinburke: OK, done - updated the code to match your new description.

kevinburke:
The AppVeyor test failure is on test 715, which I've seen fail on other PR's I've submitted, but not consistently.

configure.ac
```diff
@@ -672,7 +672,7 @@ if test X"$want_hyper" != Xno; then
     LIB_HYPER="-lhyper -ldl -lpthread -lm"
     if test X"$want_hyper" != Xdefault; then
       CPP_HYPER=-I"$want_hyper_path/capi/include"
-      LD_HYPER="-L$want_hyper_path/target/debug"
+      LD_HYPER="-L$want_hyper_path/target/release -L$want_hyper_path/target/debug"
```
jsha, line 675: This looks like an unrelated change to the config for hyper.

lib/vtls/rustls.c
```diff
@@ -161,23 +161,20 @@ cr_recv(struct Curl_easy *data, int sockindex,
...
     }
-    else if(n == 0) {
-      /* rustls returns 0 from connection_read to mean "all currently
-        available data has been read." If we bring in more ciphertext with
-        read_tls, more plaintext will become available. So don't tell curl
-        this is an EOF. Instead, say "come back later." */
-      infof(data, "cr_recv got 0 bytes of plaintext");
-      backend->data_pending = FALSE;
-      break;
+    else if(n == 0)
```
jsha, line 175: Missing a curly brace here:

```suggestion
    else if(n == 0) {
```

jsha:
Also, can you pull in the commit from #7956 that changes the expected signature for rustls_version? Otherwise this produces a warning on build, and also doesn't print the rustls version when you run `curl --version`.

lib/vtls/rustls.c
```diff
@@ -543,14 +540,20 @@ cr_close(struct Curl_easy *data, struct connectdata *conn,
   }
 }
 
+static size_t cr_version(char *buffer, size_t size)
+{
+  struct rustls_str ver = rustls_version();
+  return msnprintf(buffer, size, "%s", ver.data);
```
jsha, line 546: This should be:

```suggestion
  return msnprintf(buffer, size, "%.*s", (int)ver.len, ver.data);
```

configure.ac
jsha: Don't forget to revert this diff.

bagder (approved):

bagder:
Thanks!

PR #7950, opened 2021-11-03
vtls/rustls: handle RUSTLS_RESULT_PLAINTEXT_EMPTY

### LINKED ISSUE #7949: vtls/rustls: error on RUSTLS_RESULT_PLAINTEXT_EMPTY instead of handling
If I make a request to https://go.googlesource.com using tip + pull request #7948, I get the following error:

```
curl: (56) error in rustls_connection_read
```

Inspecting this location in the source code, it appears we are swallowing the error returned by Rust and instead passing along CURLE_READ_ERROR:

```c
    else if(rresult != RUSTLS_RESULT_OK) {
      failf(data, "error in rustls_connection_read");
      *err = CURLE_READ_ERROR;
      return -1;
    }
```

The actual error, when printed, is RUSTLS_RESULT_PLAINTEXT_EMPTY, which looks like the same case being handled in the next if block:

```c
    else if(n == 0) {
      /* rustls returns 0 from connection_read to mean "all currently
        available data has been read." If we bring in more ciphertext with
        read_tls, more plaintext will become available. So don't tell curl
        this is an EOF. Instead, say "come back later." */
      infof(data, "cr_recv got 0 bytes of plaintext");
      backend->data_pending = FALSE;
      break;
    }
```

Which suggests a fix, just pass RUSTLS_RESULT_PLAINTEXT_EMPTY to that second block instead of to the first one.

kevinburke:
Previously we'd return CURLE_READ_ERROR if we received this, instead
of triggering the error handling logic that's present in the next if
block down.

After this change, curl requests to https://go.googlesource.com using
HTTP/2 complete successfully.

Fixes #7949.

### DIFF
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -165,12 +165,13 @@ cr_recv(struct Curl_easy *data, int sockindex,
       *err = CURLE_OK;
       return 0;
     }
-    else if(rresult != RUSTLS_RESULT_OK) {
+    else if(rresult != RUSTLS_RESULT_OK &&
+            rresult != RUSTLS_RESULT_PLAINTEXT_EMPTY) {
       failf(data, "error in rustls_connection_read");
       *err = CURLE_READ_ERROR;
       return -1;
     }
-    else if(n == 0) {
+    else if(n == 0 || rresult == RUSTLS_RESULT_PLAINTEXT_EMPTY) {
       /* rustls returns 0 from connection_read to mean "all currently
         available data has been read." If we bring in more ciphertext with
         read_tls, more plaintext will become available. So don't tell curl


bagder:
`make checksrc`...
~~~
./vtls/rustls.c:168:87: warning: Longer than 79 columns (LONGLINE)
     else if(rresult != RUSTLS_RESULT_OK && rresult != RUSTLS_RESULT_PLAINTEXT_EMPTY) {

kevinburke:
Fixed checksrc issue

kevinburke:
The AppVeyor failure is on test 715 on one of the test suites. Here is the log for that test: https://gist.github.com/kevinburke/98590b1de7407eebacb0308f538e068d

I am not sure that this patch would have affected that code so it seems like a test flake; I can open a separate ticket if you like.

kevinburke:
Other failures seem to be focused on the "brew install" step but I don't see any additional logs or information about why they'd be failing. https://github.com/curl/curl/runs/4095721123?check_suite_focus=true

jay:
I've set some of the CI jobs to re-run

jsha:
The enum value RUSTLS_RESULT_PLAINTEXT_EMPTY is new in rustls-ffi main (which will become rustls-ffi 0.8.0 within a few days), so this should fail when built against curl master, which relies on rustls-ffi 0.7.2. I'm surprised this passes tests! Anyhow, I think this should be combined into #7948 or another change that upgrades to 0.8.0 and also updates the function calls and result code handling all in one go.

kevinburke:
Merged the changes here into #7948, let's review that.

PR #7951, opened 2021-11-03
zuul: build quiche with BoringSSL built separately

### LINKED ISSUE #7927: The quiche+boring CI builds fail on Zuul
### I did this

I checked recent CI builds. I noticed that both builds on zuul that build quiche (+boringssl) fail consistently. The build ends with:
~~~
  -- Build files have been written to: /home/zuul/quiche/target/release/build/quiche-986e736b297204f2/out/build
  running: "cmake" "--build" "." "--target" "bssl" "--config" "RelWithDebInfo" "--parallel" "8"

  --- stderr
  Unknown argument --parallel
  Unknown argument 8
  Usage: cmake --build <dir> [options] [-- [native-options]]
  Options:
    <dir>          = Project binary directory to be built.
    --target <tgt> = Build <tgt> instead of default targets.
                     May only be specified once.
    --config <cfg> = For multi-configuration tools, choose <cfg>.
    --clean-first  = Build target 'clean' first, then build.
                     (To clean only, use --target 'clean'.)
    --use-stderr   = Ignored.  Behavior is default in CMake >= 3.0.
    --             = Pass remaining options to the native tool.
  thread 'main' panicked at '
  command did not execute successfully, got: exit status: 1
~~~

[example failure](https://curl.zuul.vexxhost.dev/build/a31e48a911054be0b0d3af08fa5719d6)

I *think* this is the boringssl part? It looks like it uses a cmake option that the cmake version used here doesn't support.

### I expected the following

All quiche builds to be fine.

### curl/libcurl version

current git master

### operating system

These builds run on ubuntu-bionic.

bagder:
... in an attempt to avoid BoringSSL using cmake --parallel because
that's an option that doesn't exist in the cmake shipped in Ubuntu
Bionic, which is a system commonly used in our CI systems.

Fixes #7927

### DIFF
--- a/scripts/zuul/before_script.sh
+++ b/scripts/zuul/before_script.sh
@@ -120,9 +120,7 @@ if [ "$TRAVIS_OS_NAME" = linux -a "$QUICHE" ]; then
   curl https://sh.rustup.rs -sSf | sh -s -- -y
   source $HOME/.cargo/env
   cd $HOME/quiche
-  cargo build -v --release --features ffi,pkg-config-meta,qlog
-  mkdir -v deps/boringssl/src/lib
-  ln -vnf $(find target/release -name libcrypto.a -o -name libssl.a) deps/boringssl/src/lib/
+  QUICHE_BSSL_PATH=$HOME/boringssl cargo build -v --release --features ffi,pkg-config-meta,qlog
 fi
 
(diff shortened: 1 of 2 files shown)


bagder:
I was wrong, this fix might also work. Testing it out...

bagder:
This still didn't work but #7952 did. Going with that work-around for now.

PR #7955, opened 2021-11-03
.github: retry macos "brew install" command on failure

kevinburke:
Previously we saw errors attempting to run "brew install", see
https://github.com/curl/curl/runs/4095721123?check_suite_focus=true
for an example, since this command is idempotent, it is safe to run
again.


bagder (approved):

bagder:
Thanks!

PR #7956, opened 2021-11-04
vtls/rustls: adapt to the updated rustls_version proto

bagder:
Combine this with #7948

### DIFF
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -543,6 +543,13 @@ cr_close(struct Curl_easy *data, struct connectdata *conn,
   }
 }
 
+static size_t cr_version(char *buffer, size_t size)
+{
+  struct rustls_str ver = rustls_version();
+  /* the provided string isn't NUL-terminated */
+  return msnprintf(buffer, size, "%.*s", (int)ver.len, ver.data);
+}
+
 const struct Curl_ssl Curl_ssl_rustls = {
   { CURLSSLBACKEND_RUSTLS, "rustls" },
   SSLSUPP_TLS13_CIPHERSUITES,      /* supports */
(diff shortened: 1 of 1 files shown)


kevinburke:
LGTM

lib/vtls/rustls.c
```diff
@@ -543,14 +543,20 @@ cr_close(struct Curl_easy *data, struct connectdata *conn,
   }
 }
 
+static size_t cr_version(char *buffer, size_t size)
+{
+  struct rustls_str ver = rustls_version();
+  return msnprintf(buffer, size, "%s", ver.data);
```
jsha, line 549: This should use ``%.*s`, ver.len, ver.data`, since ver's len will probably be shorter than the buffer's size.

bagder: The header file says it is fine to treat it as a C string though, so shouldn't a plain %s write the same output?

jsha: The header comment for rustls_str says:

```
/**
 * A read-only view on a Rust `&str`. The contents are guaranteed to be valid
 * UTF-8. As an additional guarantee on top of Rust's normal UTF-8 guarantee,
 * a `rustls_str` is guaranteed not to contain internal NUL bytes, so it is
 * safe to interpolate into a C string or compare using strncmp. Keep in mind
 * that it is not NUL-terminated.
 *
 * The memory exposed is available as specified by the function
 * using this in its signature. For instance, when this is a parameter to a
 * callback, the lifetime will usually be the duration of the callback.
 * Functions that receive one of these must not dereference the data pointer
 * beyond the allowed lifetime.
 */
typedef struct rustls_str {
  const char *data;
  size_t len;
} rustls_str;
```

So it's safe to _interpolate_ into a C string, but not safe to use directly as a C string (which assumes NUL termination). I'm happy to improve the docs if I can make them clearer!

bagder: I think it was just a problem with this user not reading what it actually says... :open_mouth:

bagder:
Merged into #7948 already

PR #7960, opened 2021-11-04
ngtcp2: support latest QUIC TLS RFC9001

a-denoyelle:
QUIC Transport Parameters Extension has been changed between draft-29
and latest RFC9001. Most notably, its identifier has been updated from
0xffa5 to 0x0039. The version is selected through the QUIC TLS library
via the legacy codepoint.

Disable the usage of legacy codepoint in curl to switch to latest
RFC9001. This is required to be able to keep up with latest QUIC
implementations.

### DIFF
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -303,7 +303,7 @@ static int quic_init_ssl(struct quicsocket *qs)
 
   SSL_set_app_data(qs->ssl, qs);
   SSL_set_connect_state(qs->ssl);
-  SSL_set_quic_use_legacy_codepoint(qs->ssl, 1);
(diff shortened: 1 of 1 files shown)


bagder:
@tatsuhiro-t this should be just fine, right?

tatsuhiro-t:
It is fine.  I'd suggest that ALPN should also be updated to h3 if v1 transport_parameters codepoint is used.

bagder:
The RFC for QUIC v1 is already published though, while the HTTP/3 one isn't so shouldn't we strictly remain on h3-[draft] until the spec actually ships?

tatsuhiro-t:
Good point.  I think just continuing to use h3-29 would be fine.

bagder:
Thanks both of you!

PR #7962, opened 2021-11-04
man pages: fix backslash-n in examples

bagder:
... to be proper backslash-backslash-n sequences to render nicely in man
and on the website.

Reported-by: Sergey Markelov

Fixes https://github.com/curl/curl-www/issues/163


jay (approved):
follow up to 2415556

PR #7963, opened 2021-11-05
Turn the Location: into a terminal hyperlink

dfandrich:
I just learned about a new capability that terminals have gained in the last
few years, namely the ability to create clickable hyperlinks on the terminal
with arbitrary text.  Many terminals already turn URLs on screen into clickable
links but there are times when curl could supply links that aren't displayed on
screen.

This PR adds the ability for curl to create hyperlinks for Location: redirects.
There is not enough information for a terminal to do this automatically for
relative URLs, but curl has enough information to do it. This is enabled when
--styled-output is turned on and a Location: header is displayed. 

There are probably other cases when curl could enhance the output with links
like this, so this is mostly a proof of concept to see how easy and useful it
is. It turns out that this marginal feature takes a surprising amount of code
to accomplish due to CURLINFO_REDIRECT_URL being unavailable. I won't take it
personally if the end result is rejection this PR for a too-low
price/performance ratio.

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -47,6 +47,15 @@ static char *parse_filename(const char *ptr, size_t len);
    bold-off code (21) isn't supported everywhere - like in the mac
    Terminal. */
 #define BOLDOFF "\x1b[0m"
+/* OSC 8 hyperlink escape sequence */
+#define LINK "\x1b]8;;"
+#define LINKST "\x1b\\"
+#define LINKOFF LINK LINKST
+#endif
+
+#ifdef LINK
+static void write_linked_location(CURL *curl, const char *location,
+    size_t loclen, FILE *stream);
 #endif
 
 /*
@@ -204,7 +213,16 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
     if(value) {
       size_t namelen = value - ptr;
       fprintf(outs->stream, BOLD "%.*s" BOLDOFF ":", namelen, ptr);
+#ifndef LINK
       fwrite(&value[1], cb - namelen - 1, 1, outs->stream);
+#else
+      if(curl_strnequal("Location", ptr, namelen)) {
+        write_linked_location(per->curl, &value[1], cb - namelen - 1,
+            outs->stream);
+      }
+      else
+        fwrite(&value[1], cb - namelen - 1, 1, outs->stream);
+#endif
     }
     else
       /* not "handled", just show it */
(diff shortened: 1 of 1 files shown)


dfandrich:
Here is an example using vte.
hyperlink

src/tool_cb_hdr.c
```diff
@@ -204,7 +208,89 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
     if(value) {
       size_t namelen = value - ptr;
       fprintf(outs->stream, BOLD "%.*s" BOLDOFF ":", namelen, ptr);
+#ifndef LINK
       fwrite(&value[1], cb - namelen - 1, 1, outs->stream);
+#else
+      if(!strncasecmp("Location", ptr, namelen)) {
```
bagder, line 214: I think this function call needs to be replaced with `curl_strnequal` for portability.

bagder (commented):
I like it! I'm just a tad bit worried what this might do in terminals without support?

src/tool_cb_hdr.c
dfandrich: Done. I can never remember which of those replacement functions is available in libcurl and which in curl.

dfandrich:
I tried this on five different terminal that do not support OSC 8 hyperlinks, or at least don't do anything with them (Konsole, xterm, eterm, screen, Linux console), and all of them ignored the hyperlink and displayed the text as before. It's been a while since I've looked at the ANSI spec, but I'm guessing it's using some feature of the escape sequences that's designed to be ignored by terminals that don't know about this feature.

src/tool_cb_hdr.c
```diff
@@ -204,7 +208,89 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
...
+        if(u) {
+          curl_url_cleanup(u);
+          free(copyloc);
+          free(scheme);
+          free(finalurl);
+        }
+      }
+      else
+        fwrite(&value[1], cb - namelen - 1, 1, outs->stream);
+#endif
```
bagder, line 293: maybe rather put this whole block into its own function to make it easier to deal with when glancing the code.

bagder (approved):

dfandrich:
I've moved the new code into a new function, as requested, and fixed the memory cleanup to use the right curl_free function. I think it's ready to be submitted now.

Eliteghost19:
is something messed up

________________________________________

dfandrich:
There was a compiler warning in the non-enabled case that broke some builds. I've fixed that now.

PR #7966, opened 2021-11-05
aws_sigv4: fix headers computation for some unhandled cases.

outscale-mgo:
Up to this PR, canonical header, and signed header was hardcoded
in the aws-sigv4 signature.

Those patch handle canonical headers and signed headers creation
as it is explain here:
https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

The algorithm tell that signed and canonical must contain at last `host` and `x-amz-date`.

So we check whatever they are present in the curl http headers list.
If they are, we use the one enter by curl user, otherwise we generate them.
then we to lower, and remove space from each http headers plus `host` and `x-amz-date`,
then sort them all by alphabetical order.

This patch also fix a bug with host header, which was ignoring the port.

Because I now handle the port in the signature,
I had to hardcode the host in the tests, in which the port was random,
which was making impossible to had a signature in the tests data.


lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,125 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+                             const char *provider1_low,
+                             const char *provider1_mid,
+                             char **canonical_headers,
+                             char **signed_headers)
+{
+  char *date_hdr_key =  curl_maprintf("X-%s-Date", provider1_mid);
+  char *date_full_hdr = curl_maprintf("x-%s-date:%s",
+                                      provider1_low, timestamp);
+#ifdef DEBUGBUILD
+  char *force_host = getenv("CURL_FORCETIME");
```
bagder, line 85: Why is the environment variable to *force host* called force **time** ?

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,125 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+  int ret = CURLE_OUT_OF_MEMORY;
+  struct curl_slist *l;
+  int again = 1;
+
+  if(Curl_checkheaders(data, "Host")) {
+    head = NULL;
+  }
+#ifdef DEBUGBUILD
+  else if(force_host) {
+    head = curl_slist_append(NULL, "host:127.0.0.1");
```
bagder, line 97: what if this fails? (which they do even in debug builds when we run torture tests etc)

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,125 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+  }
+#endif
+  else {
+    char *full_host = data->state.aptr.host ? data->state.aptr.host :
+      curl_maprintf("host:%s", hostname);
```
bagder, line 102: No space after the colon? And what if `curl_maprintf()` returns NULL?

In what situations is `data->state.aptr.host` NULL when this is called? What then if there's a custom port number or an IPv6 address host name?

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,125 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+
+    head = curl_slist_append(NULL, full_host);
```
bagder, line 104: What if this returns NULL?

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,125 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+      memmove(tmp, tmp2, strlen(tmp2) + 1);
+    tmp = tmp + strlen(tmp) - 1;
+    while (*tmp == ' ' || *tmp == '\t' || *tmp == '\n' || *tmp == '\r') {
+      *tmp = 0;
+      --tmp;
+    }
+  }
+
+  if(!Curl_checkheaders(data, date_hdr_key))
+    curl_slist_append(head, date_full_hdr);
```
bagder, line 142: No return code check for a function that is fallible.

bagder:
Why does it have to sort the headers?

> I had to hardcode the host in the tests, in which the port was random, which was making impossible to had a signature in the tests data.

I think that's fine. Another way would be to replace the port number part only with a fixed number for the test run.

lib/http_aws_sigv4.c
outscale-mgo: TBH I really don't know if not having data->state.aptr.host is possible, I've made this ternary out of fear of having full_host NULL, which would be a bug.
Should I remove the condition ?

bagder: It is probably wise to keep it.

outscale-mgo: No the CanonicalHeaders must contain `Lowercase(HeaderName) + ':' + Trimall(HeaderValue) + '\n'`.
so no space before and after the Value.
I don't see why it wouldn't work with the current code with IPv6 ?
I guess I could try to add manually the port, if the ip is set using `conn->host.name`, though most case where a port is needed should be handle using `data->state.aptr.host`. 
My guess is that this would be too much code for handling a too niche corner case, maybe add a comment saying we don't handle port if host is set using `conn->host.name` is enough ?

outscale-mgo:
> Why does it have to sort the headers?

the "header" have to be sort in the canonical header and the signed header.
but if I'm correct, I don't change the existing headers order in  `data->set.headers`

outscale-mgo:
note that I still need to converts sequential spaces to a single space, in order to be conform to the algorithm.

outscale-mgo:
> 

should be done now

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,159 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+    if(!tmp_head)
+      goto fail;
+    head = tmp_head;
+  }
+
+  /* remove whitespace, and lowercase each headers */
+  for(l = head; l; l = l->next) {
+    char *tmp = l->data, *tmp2;
+
+    Curl_strntolower(l->data, l->data, strlen(l->data));
```
matwey, line 134: Sorry to bother. 

Here you lowercase the whole header while the specs require you to lowercase only the header name:

https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,159 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+    }
+
+  }
+
+  if(!Curl_checkheaders(data, date_hdr_key)) {
+    tmp_head = curl_slist_append(head, date_full_hdr);
+    if(!tmp_head)
+      goto fail;
+    tmp_head = head;
+  }
```
matwey, line 175: If `X-Amz-Date` header is missed from the request headers, but present in CanonicalHeaders, how could the remote endpoint validate the signature?

lib/http_aws_sigv4.c
matwey: Please note that `data->state.aptr.host` consists of `\r\n` on its end. You have to trim them in order not to create extra empty lines in CanonicalHeaders.

lib/http_aws_sigv4.c
outscale-mgo: well, I have this code:
```
  if(!Curl_checkheaders(data, date_hdr_key)) {
    tmp_head = curl_slist_append(head, date_full_hdr);
    if(!tmp_head)
      goto fail;
    tmp_head = head;
  }
```
As, if I remember correctly AWS say you need to have X-Amz-Date, I force the header if not present.

matwey: Sure, but first of all, headers in the request must be consistent with the headers in CanonicalHeaders. Here, you force the `X-Amz-Date` header only in CanonicalHeaders.

lib/http_aws_sigv4.c
```diff
@@ -66,6 +68,159 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+
+      if(tmp != tmp2) {
+        if(*tmp2 == 0)
+          /* remove spaces at the end */
+          *tmp = 0;
+        else {
+          *tmp++ = ' '; /* if tab convert to space */
+          if (*tmp == ' ' || *tmp == '\t')
+            memmove(tmp, tmp2, strlen(tmp2));
+        }
```
matwey, line 160: I may be wrong here, but the spec says:
```
Trim() Remove any leading or trailing white-space. 
```
and this code seems to do extra work on intermediate spaces. Do you know any valid examples with space-containing header values?

Please note, that trailing `\r\n` (from `data->state.aptr.host`) also could be trimmed here.

lib/http_aws_sigv4.c
outscale-mgo: yes your right, this code doesn't add it to the request header. (I mix up my code a little)
but the date is add to the request header by doing `data->state.aptr.userpwd = auth_headers` at the end of Curl_output_aws_sigv4.
as auth_headers contain `X-%s-Date`

lib/http_aws_sigv4.c
outscale-mgo: yes, your right, I'm presently working on fixing this, thanks for the feedback.

lib/http_aws_sigv4.c
matwey: You are right: `auth_headers` has `X-%s-Date`, but I don't think it is fully correct. Here we consider two cases: 1) user has provided `X-%s-Date` headers 2) user has not provided the header. In the former case we will end up with the request consisted of two `X-%s-Date` headers (one from the user and one from `auth_headers`) with (most probably) different values, that highly likely will lead to the signature verification failure at the server side.

What do you think about moving `X-%s-Date` from `auth_headers` to `make_headers`?

outscale-mgo: I think your right, also from my understand of 
```
For HTTP/1.1 requests, the host header must be included as a signed header. For HTTP/2 requests that include the :authority header instead of the host header, you must include the :authority header as a signed header. If you include a date or x-amz-date header, you must also include that header in the list of signed headers. 
```
if `date` is provide, I must not include `x-%s-date`, which is another corner case I don't handle yet.

I'm not sure I can move `auth_headers` into `make_headers` as it contain the whole signature, but I will not add `x-%s-date` all the time.

matwey: Section "Example: PUT Object" of the spec contains the request with both `Date` and `X-Amz-Date` headers at the same time.

lib/http_aws_sigv4.c
matwey: The `Host:` bug seems to be fixed now.

lib/http_aws_sigv4.c
matwey: Thanks, now the headers are lower-cased properly.

outscale-mgo: Thanks for the feedback :)

matwey:
@outscale-mgo sorry, it seems that this PR should be rebased onto current master, now it is conflicting.

lib/http_aws_sigv4.c
```diff
@@ -66,6 +70,198 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+
+        l->data = next->data;
+        next->data = tmp;
+        again = 1;
+      }
+    }
+  }
+
+  for(l = head; l; l = l->next) {
+    char *tmp = *canonical_headers;
```
outscale-mgo, line 235: note, that I'm presently refactoring this code to use dynbuf instead of curl_maprintf for this chunk.

lib/http_aws_sigv4.c
```diff
@@ -66,6 +70,198 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+  int again = 1;
+
+  if(!date_hdr_key || !date_full_hdr)
+    goto fail;
+
+  if(Curl_checkheaders(data, "Host")) {
+    head = NULL;
+  }
+#ifdef DEBUGBUILD
+  else if(force_host) {
```
matwey, line 110: Just for information. I am not insisting on anything.

Take a look how I overcame the random port issue using `CONNECT_TO` feature:  https://github.com/curl/curl/pull/8202/commits/fc296959cda6ea28d180256d1e68309da7b62ab1
I think this could have advantage that the real code path is tested in this case.

outscale-mgo:
hello,

any news on this @bagder :) ? 

Note that this PR should be prioritized over 7600 as encoding can be manually done. 
This PR enable to call some API that require V4 signature that can be done with current implementation.

lib/http_aws_sigv4.c
```diff
@@ -66,6 +70,188 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+       * and are readd after */
+      pos = strlen(full_host);
+      if(pos > 1 && full_host[pos - 1] == '\n') {
+        full_host[pos - 1] = 0;
+        if(pos > 2 && full_host[pos - 2] == '\r')
+          full_host[pos - 2] = 0;
+      }
+    }
+    else {
+      full_host = curl_maprintf("host:%s", hostname);
```
bagder, line 124: I'm troubled by this code adding its own `Host` header here. It doesn't use nearly the same logic nor does it use the same case or space as the code in http.c. It looks like trouble.

lib/http_aws_sigv4.c
```diff
@@ -66,6 +70,188 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
...
+    if(!tmp)
+      goto fail;
+    ++tmp;
+    while(*tmp == ' ' || *tmp == '\t')
+      ++tmp;
+    strncpy(timestamp, tmp, TIMESTAMP_SIZE - 1);
+    timestamp[TIMESTAMP_SIZE - 1] = 0;
+  }
+
+  /* reorder */
```
bagder, line 209: And by this you mean you alpha-sort them in a case sensitive manner, right?

Not *terribly* important but the reason libcurl sends Host: as the first header is this paragraph from [RFC 7230 section 3.2.2](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2):

>   it is good practice to send  header fields that contain control data first, such as Host on  requests and Date on responses, so that implementations can decide  when not to handle a message as early as possible.

bagder:
This PR isn't about "refactoring" as far as I can see it and I think using that title makes it a disservice. It rather seems to fix a few issues with sgv4 and add several test cases. It would be better if that was easier to spot.

lib/http_aws_sigv4.c
matwey: It is supposed to be part of so-called 'Canonical Headers'. We need them only to calculate their hash:
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
The server will check us by constructing 'Canonical Headers' on its side and calculating the hash. This means that on both sides 'Canonical Headers' must follow the same strict format and order.

lib/http_aws_sigv4.c
outscale-mgo: Yes,
Note that I don't reorder the header send by curl (if I do, then my code is buggy)
I create a temporary string that contain the apha sorted headers, so I can create an hash out of the string.

outscale-mgo:
> This PR isn't about "refactoring" as far as I can see it and I think using that title makes it a disservice. It rather seems to fix a few issues with sigv4 and add several test cases. It would be better if that was easier to spot.

Yes I apologize for the mistake.
I'm bad at naming, I did name this PR for what is does instead of what it's trying to do.
I'll try to improve.

hansonchar:
Is this PR ready to be merged or do the 3 failing checks have to pass as a precondition?

outscale-mgo:
> Is this PR ready to be merged or do the 3 failing checks have to pass as a precondition?

If those test need to pass, thanks for telling me, as on whose 3 failing test: 1 time out, and 2 test that fail are unrelated to AWS v4 signature. 
So it seems to me that the failure are not due to my code, but I can be wrong.

hansonchar:
Perhaps you can make a no op PR to see if it would pass all the tests, by
making, say, only changes to some comments?

That will help prove one way or the other.

bagder:
The tests are super useful.

But I think the code is written in a too complicated way. It abuses `tmp` for variable names all over, it has duplicated code, it does too many allocations, it uses too many (similarly named) variables in single functions with too big scope.

I think it smells badly of a strict focus to make it work but with no desire to make the code readable, efficient and maintainable.

To better illustrate what I am talking about, I've spent two hours on the code and I could remove **60** lines of code while maintaining the same functionality and if I may say so, increased readability and removed a whole bunch of memory allocations. See below.

<details>

~~~diff
From eb2341bc14ec8a01fb1fa4e022a8d431497fc34d Mon Sep 17 00:00:00 2001
Date: Tue, 23 Aug 2022 09:33:42 +0200
Subject: [PATCH] aws-sigv4: refactor

---
 lib/http_aws_sigv4.c    | 420 +++++++++++++++++-----------------------
 tests/libtest/lib1955.c |   2 +-
 2 files changed, 182 insertions(+), 240 deletions(-)

diff --git a/lib/http_aws_sigv4.c b/lib/http_aws_sigv4.c
index 20441cd1f..672f8b7b9 100644
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -51,11 +51,11 @@
     ret = Curl_hmacit(Curl_HMAC_SHA256,     \
                       (unsigned char *)k,   \
                       (unsigned int)kl,     \
                       (unsigned char *)d,   \
                       (unsigned int)dl, o); \
-    if(ret != CURLE_OK) {                   \
+    if(ret) {                               \
       goto fail;                            \
     }                                       \
   } while(0)
 
 #define TIMESTAMP_SIZE 17
@@ -77,29 +77,74 @@ static char *find_date_hdr(struct Curl_easy *data, const char *sig_hdr)
   if(tmp)
     return tmp;
   return Curl_checkheaders(data, STRCONST("Date"));
 }
 
+/* remove whitespace, and lowercase all headers */
+static void trim_headers(struct curl_slist *head)
+{
+  struct curl_slist *l;
+  for(l = head; l; l = l->next) {
+    char *value; /* to read from */
+    char *store;
+    size_t colon = strcspn(l->data, ":");
+    Curl_strntolower(l->data, l->data, colon);
+
+    value = &l->data[colon + 1];
+    store = value;
+
+    /* skip leading whitespace */
+    while(*value && ISSPACE(*value))
+      value++;
+
+    while(*value) {
+      int space = 0;
+      while(*value && ISSPACE(*value)) {
+        value++;
+        space++;
+      }
+      if(space) {
+        /* replace any number of consecutive whitespace with a single space,
+           unless at the end of the string, then nothing */
+        if(*value)
+          *store++ = ' ';
+      }
+      else
+        *store++ = *value++;
+    }
+    *store = 0; /* null terminate */
+  }
+}
+
 static CURLcode make_headers(struct Curl_easy *data,
                              const char *hostname,
                              char *timestamp,
-                             const char *provider1_low,
-                             const char *provider1_mid,
+                             char *provider1,
                              char **date_header,
                              struct dynbuf *canonical_headers,
                              struct dynbuf *signed_headers)
 {
-  char *date_hdr_key =  curl_maprintf("X-%s-Date", provider1_mid);
-  char *date_full_hdr = curl_maprintf("x-%s-date:%s",
-                                      provider1_low, timestamp);
+  char *date_hdr_key;
+  char *date_full_hdr;
   struct curl_slist *head = NULL;
   struct curl_slist *tmp_head = NULL;
   CURLcode ret = CURLE_OUT_OF_MEMORY;
   struct curl_slist *l;
   int again = 1;
   char *full_host = NULL;
 
+  /* provider1 mid */
+  Curl_strntolower(provider1, provider1, strlen(provider1));
+  provider1[0] = Curl_raw_toupper(provider1[0]);
+
+  date_hdr_key =  curl_maprintf("X-%s-Date", provider1);
+
+  /* provider1 lowercase */
+  Curl_strntolower(provider1, provider1, 1); /* first byte only */
+  date_full_hdr = curl_maprintf("x-%s-date:%s",
+                                provider1, timestamp);
+
   if(!date_hdr_key || !date_full_hdr)
     goto fail;
 
   if(Curl_checkheaders(data, STRCONST("Host"))) {
     head = NULL;
@@ -109,18 +154,13 @@ static CURLcode make_headers(struct Curl_easy *data,
       size_t pos;
 
       full_host = strdup(data->state.aptr.host);
       if(!full_host)
         goto fail;
-      /* remove /r/n as the separator for canonical request must be '\n'
-       * and are readd after */
-      pos = strlen(full_host);
-      if(pos > 1 && full_host[pos - 1] == '\n') {
-        full_host[pos - 1] = 0;
-        if(pos > 2 && full_host[pos - 2] == '\r')
-          full_host[pos - 2] = 0;
-      }
+      /* remove /r/n as the separator for canonical request must be '\n' */
+      pos = strcspn(full_host, "\n\r");
+      full_host[pos] = 0;
     }
     else {
       full_host = curl_maprintf("host:%s", hostname);
     }
 
@@ -140,71 +180,34 @@ static CURLcode make_headers(struct Curl_easy *data,
     if(!tmp_head)
       goto fail;
     head = tmp_head;
   }
 
-  /* remove whitespace, and lowercase each headers */
-  for(l = head; l; l = l->next) {
-    char *tmp = l->data, *tmp2;
-    size_t pos;
-
-    for(pos = 0; tmp[pos] && tmp[pos] != ':'; ++pos);
-    Curl_strntolower(l->data, l->data, pos);
-
-    tmp += pos;
-    if(!*tmp)
-      continue;
-    ++tmp;
-    tmp2 = tmp;
-    while(*tmp2 == ' ' || *tmp2 == '\t') {
-      ++tmp2;
-    }
-    if(tmp != tmp2)
-      memmove(tmp, tmp2, strlen(tmp2) + 1);
-
-    while(*tmp) {
-      tmp2 = tmp;
-      while(*tmp2 == ' ' || *tmp2 == '\t') {
-        ++tmp2;
-      }
-
-      if(tmp != tmp2) {
-        if(*tmp2 == 0)
-          /* remove spaces at the end */
-          *tmp = 0;
-        else {
-          *tmp++ = ' '; /* if tab convert to space */
-          if(*tmp == ' ' || *tmp == '\t')
-            memmove(tmp, tmp2, strlen(tmp2));
-        }
-        tmp = tmp2;
-      }
-      else {
-        ++tmp;
-      }
-    }
-  }
+  trim_headers(head);
 
   *date_header = find_date_hdr(data, date_hdr_key);
   if(!*date_header) {
     tmp_head = curl_slist_append(head, date_full_hdr);
     if(!tmp_head)
       goto fail;
     head = tmp_head;
     *date_header = curl_maprintf("%s: %s", date_hdr_key, timestamp);
   }
   else {
-    char *tmp;
+    char *value;
 
     *date_header = strdup(*date_header);
-    tmp = strchr(*date_header, ':');
-    if(!tmp)
+    if(!*date_header)
+      goto fail;
+
+    value = strchr(*date_header, ':');
+    if(!value)
       goto fail;
-    ++tmp;
-    while(*tmp == ' ' || *tmp == '\t')
-      ++tmp;
-    strncpy(timestamp, tmp, TIMESTAMP_SIZE - 1);
+    ++value;
+    while(ISSPACE(*value))
+      ++value;
+    strncpy(timestamp, value, TIMESTAMP_SIZE - 1);
     timestamp[TIMESTAMP_SIZE - 1] = 0;
   }
 
   /* alpha-sort in a case sensitive manner */
   for(again = 1; again;) {
@@ -250,60 +253,54 @@ fail:
   curl_slist_free_all(head);
 
   return ret;
 }
 
+/* maximum lenth for the aws sivg4 parts */
+#define MAX_SIGV4_LEN 64
+#define MAX_SIGV4_LEN_TXT "64"
+
 CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
 {
   CURLcode ret = CURLE_OUT_OF_MEMORY;
   size_t len;
-  const char *tmp0;
-  const char *tmp1;
-  char *provider0_low = NULL;
-  char *provider0_up = NULL;
-  char *provider1_low = NULL;
-  char *provider1_mid = NULL;
-  char *region = NULL;
-  char *service = NULL;
+  const char *arg;
+  char provider0[MAX_SIGV4_LEN + 1]="";
+  char provider1[MAX_SIGV4_LEN + 1]="";
+  char region[MAX_SIGV4_LEN + 1]="";
+  char service[MAX_SIGV4_LEN + 1]="";
   const char *hostname = conn->host.name;
-#ifdef DEBUGBUILD
-  char *force_timestamp;
-#endif
   time_t clock;
   struct tm tm;
   char timestamp[TIMESTAMP_SIZE];
   char date[9];
   struct dynbuf canonical_headers;
   struct dynbuf signed_headers;
   char *date_header = NULL;
-  Curl_HttpReq httpreq;
-  const char *method;
-  size_t post_data_len;
-  const char *post_data = data->set.postfields ? data->set.postfields : "";
+  const char *post_data = data->set.postfields;
+  size_t post_data_len = 0;
   unsigned char sha_hash[32];
   char sha_hex[65];
-  char *canonical_request = NULL;
-  char *request_type = NULL;
-  char *credential_scope = NULL;
-  char *str_to_sign = NULL;
+  char *canonical_request;
+  char *request_type;
+  char *credential_scope;
+  char *str_to_sign;
   const char *user = data->state.aptr.user ? data->state.aptr.user : "";
-  const char *passwd = data->state.aptr.passwd ? data->state.aptr.passwd : "";
   char *secret = NULL;
-  unsigned char tmp_sign0[32] = {0};
-  unsigned char tmp_sign1[32] = {0};
+  unsigned char sign0[32] = {0};
+  unsigned char sign1[32] = {0};
   char *auth_headers = NULL;
 
   DEBUGASSERT(!proxy);
   (void)proxy;
 
   if(Curl_checkheaders(data, STRCONST("Authorization"))) {
     /* Authorization already present, Bailing out */
     return CURLE_OK;
   }
 
-
   /* we init thoses buffers here, so goto fail will free initialized dynbuf */
   Curl_dyn_init(&canonical_headers, CURL_MAX_HTTP_HEADER);
   Curl_dyn_init(&signed_headers, CURL_MAX_HTTP_HEADER);
 
   /*
@@ -311,236 +308,187 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
    * Google and Outscale use the same OSC or GOOG,
    * but Amazon uses AWS and AMZ for header arguments.
    * AWS is the default because most of non-amazon providers
    * are still using aws:amz as a prefix.
    */
-  tmp0 = data->set.str[STRING_AWS_SIGV4] ?
+  arg = data->set.str[STRING_AWS_SIGV4] ?
     data->set.str[STRING_AWS_SIGV4] : "aws:amz";
-  tmp1 = strchr(tmp0, ':');
-  len = tmp1 ? (size_t)(tmp1 - tmp0) : strlen(tmp0);
-  if(len < 1) {
-    infof(data, "first provider can't be empty");
+
+  /* provider1[:provider2[:region[:service]]]
+
+     No string can be longer than N bytes of non-whitespace
+   */
+  (void)sscanf(arg, "%" MAX_SIGV4_LEN_TXT "[^:]"
+               ":%" MAX_SIGV4_LEN_TXT "[^:]"
+               ":%" MAX_SIGV4_LEN_TXT "[^:]"
+               ":%" MAX_SIGV4_LEN_TXT "s",
+               provider0, provider1, region, service);
+  if(!provider0[0]) {
+    failf(data, "first provider can't be empty");
     ret = CURLE_BAD_FUNCTION_ARGUMENT;
     goto fail;
   }
-  provider0_low = malloc(len + 1);
-  provider0_up = malloc(len + 1);
-  if(!provider0_low || !provider0_up) {
-    goto fail;
-  }
-  Curl_strntolower(provider0_low, tmp0, len);
-  provider0_low[len] = '\0';
-  Curl_strntoupper(provider0_up, tmp0, len);
-  provider0_up[len] = '\0';
-
-  if(tmp1) {
-    tmp0 = tmp1 + 1;
-    tmp1 = strchr(tmp0, ':');
-    len = tmp1 ? (size_t)(tmp1 - tmp0) : strlen(tmp0);
-    if(len < 1) {
-      infof(data, "second provider can't be empty");
-      ret = CURLE_BAD_FUNCTION_ARGUMENT;
-      goto fail;
-    }
-    provider1_low = malloc(len + 1);
-    provider1_mid = malloc(len + 1);
-    if(!provider1_low || !provider1_mid) {
-      goto fail;
-    }
-    Curl_strntolower(provider1_low, tmp0, len);
-    provider1_low[len] = '\0';
-    Curl_strntolower(provider1_mid, tmp0, len);
-    provider1_mid[0] = Curl_raw_toupper(provider1_mid[0]);
-    provider1_mid[len] = '\0';
-
-    if(tmp1) {
-      tmp0 = tmp1 + 1;
-      tmp1 = strchr(tmp0, ':');
-      len = tmp1 ? (size_t)(tmp1 - tmp0) : strlen(tmp0);
-      if(len < 1) {
-        infof(data, "region can't be empty");
-        ret = CURLE_BAD_FUNCTION_ARGUMENT;
-        goto fail;
-      }
-      region = Curl_memdup(tmp0, len + 1);
-      if(!region) {
-        goto fail;
-      }
-      region[len] = '\0';
-
-      if(tmp1) {
-        tmp0 = tmp1 + 1;
-        service = strdup(tmp0);
-        if(!service) {
-          goto fail;
-        }
-        if(strlen(service) < 1) {
-          infof(data, "service can't be empty");
-          ret = CURLE_BAD_FUNCTION_ARGUMENT;
-          goto fail;
-        }
-      }
-    }
-  }
-  else {
-    provider1_low = Curl_memdup(provider0_low, len + 1);
-    provider1_mid = Curl_memdup(provider0_low, len + 1);
-    if(!provider1_low || !provider1_mid) {
-      goto fail;
-    }
-    provider1_mid[0] = Curl_raw_toupper(provider1_mid[0]);
-  }
+  else if(!provider1[0])
+    strcpy(provider1, provider0);
 
-  if(!service) {
-    tmp0 = hostname;
-    tmp1 = strchr(tmp0, '.');
-    if(!tmp1) {
-      infof(data, "service missing in parameters or hostname");
+  if(!service[0]) {
+    char *hostdot = strchr(hostname, '.');
+    if(!hostdot) {
+      failf(data, "service missing in parameters and hostname");
       ret = CURLE_URL_MALFORMAT;
       goto fail;
     }
-    len = tmp1 - tmp0;
-    service = Curl_memdup(tmp0, len + 1);
-    if(!service) {
+    len = hostdot - hostname;
+    if(len > MAX_SIGV4_LEN) {
+      failf(data, "service too long in hostname");
+      ret = CURLE_URL_MALFORMAT;
       goto fail;
     }
+    strncpy(service, hostname, len);
     service[len] = '\0';
 
-    if(!region) {
-      tmp0 = tmp1 + 1;
-      tmp1 = strchr(tmp0, '.');
-      if(!tmp1) {
-        infof(data, "region missing in parameters or hostname");
+    if(!region[0]) {
+      const char *reg = hostdot + 1;
+      const char *hostreg = strchr(reg, '.');
+      if(!hostreg) {
+        failf(data, "region missing in parameters and hostname");
         ret = CURLE_URL_MALFORMAT;
         goto fail;
       }
-      len = tmp1 - tmp0;
-      region = Curl_memdup(tmp0, len + 1);
-      if(!region) {
+      len = hostreg - reg;
+      if(len > MAX_SIGV4_LEN) {
+        failf(data, "region too long in hostname");
+        ret = CURLE_URL_MALFORMAT;
         goto fail;
       }
+      strncpy(region, reg, len);
       region[len] = '\0';
     }
   }
 
 #ifdef DEBUGBUILD
-  force_timestamp = getenv("CURL_FORCETIME");
-  if(force_timestamp)
-    clock = 0;
-  else
-    time(&clock);
+  {
+    char *force_timestamp = getenv("CURL_FORCETIME");
+    if(force_timestamp)
+      clock = 0;
+    else
+      time(&clock);
+  }
 #else
   time(&clock);
 #endif
   ret = Curl_gmtime(clock, &tm);
-  if(ret != CURLE_OK) {
+  if(ret) {
     goto fail;
   }
-  ret = CURLE_OUT_OF_MEMORY;
   if(!strftime(timestamp, sizeof(timestamp), "%Y%m%dT%H%M%SZ", &tm)) {
+    ret = CURLE_OUT_OF_MEMORY;
     goto fail;
   }
 
-  if((ret = make_headers(data, hostname, timestamp,
-                         provider1_low, provider1_mid,
-                         &date_header,
-                         &canonical_headers,
-                         &signed_headers)) != CURLE_OK)
+  ret = make_headers(data, hostname, timestamp, provider1,
+                     &date_header, &canonical_headers, &signed_headers);
+  if(ret)
     goto fail;
   ret = CURLE_OUT_OF_MEMORY;
 
   memcpy(date, timestamp, sizeof(date));
   date[sizeof(date) - 1] = 0;
 
-  if(data->set.postfieldsize < 0)
-    post_data_len = strlen(post_data);
-  else
-    post_data_len = (size_t)data->set.postfieldsize;
+  if(post_data) {
+    if(data->set.postfieldsize < 0)
+      post_data_len = strlen(post_data);
+    else
+      post_data_len = (size_t)data->set.postfieldsize;
+  }
   if(Curl_sha256it(sha_hash, (const unsigned char *) post_data,
-                   post_data_len)) {
+                   post_data_len))
     goto fail;
-  }
 
   sha256_to_hex(sha_hex, sha_hash, sizeof(sha_hex));
 
-  Curl_http_method(data, conn, &method, &httpreq);
-
-  canonical_request =
-    curl_maprintf("%s\n" /* HTTPRequestMethod */
-                  "%s\n" /* CanonicalURI */
-                  "%s\n" /* CanonicalQueryString */
-                  "%s\n" /* CanonicalHeaders */
-                  "%s\n" /* SignedHeaders */
-                  "%s",  /* HashedRequestPayload in hex */
-                  method,
-                  data->state.up.path,
-                  data->state.up.query ? data->state.up.query : "",
-                  Curl_dyn_ptr(&canonical_headers),
-                  Curl_dyn_ptr(&signed_headers),
-                  sha_hex);
-  if(!canonical_request) {
-    goto fail;
+  {
+    Curl_HttpReq httpreq;
+    const char *method;
+
+    Curl_http_method(data, conn, &method, &httpreq);
+
+    canonical_request =
+      curl_maprintf("%s\n" /* HTTPRequestMethod */
+                    "%s\n" /* CanonicalURI */
+                    "%s\n" /* CanonicalQueryString */
+                    "%s\n" /* CanonicalHeaders */
+                    "%s\n" /* SignedHeaders */
+                    "%s",  /* HashedRequestPayload in hex */
+                    method,
+                    data->state.up.path,
+                    data->state.up.query ? data->state.up.query : "",
+                    Curl_dyn_ptr(&canonical_headers),
+                    Curl_dyn_ptr(&signed_headers),
+                    sha_hex);
+    if(!canonical_request)
+      goto fail;
   }
 
-  request_type = curl_maprintf("%s4_request", provider0_low);
-  if(!request_type) {
+  /* provider 0 lowercase */
+  Curl_strntolower(provider0, provider0, strlen(provider0));
+  request_type = curl_maprintf("%s4_request", provider0);
+  if(!request_type)
     goto fail;
-  }
 
   credential_scope = curl_maprintf("%s/%s/%s/%s",
                                    date, region, service, request_type);
-  if(!credential_scope) {
+  if(!credential_scope)
     goto fail;
-  }
 
   if(Curl_sha256it(sha_hash, (unsigned char *) canonical_request,
-                   strlen(canonical_request))) {
+                   strlen(canonical_request)))
     goto fail;
-  }
 
   sha256_to_hex(sha_hex, sha_hash, sizeof(sha_hex));
 
+  /* provider 0 uppercase */
+  Curl_strntoupper(provider0, provider0, strlen(provider0));
+
   /*
-   * Google allow to use rsa key instead of HMAC, so this code might change
-   * In the future, but for now we support only HMAC version
+   * Google allows using RSA key instead of HMAC, so this code might change
+   * in the future. For now we ony support HMAC.
    */
   str_to_sign = curl_maprintf("%s4-HMAC-SHA256\n" /* Algorithm */
                               "%s\n" /* RequestDateTime */
                               "%s\n" /* CredentialScope */
                               "%s",  /* HashedCanonicalRequest in hex */
-                              provider0_up,
+                              provider0,
                               timestamp,
                               credential_scope,
                               sha_hex);
   if(!str_to_sign) {
     goto fail;
   }
 
-  secret = curl_maprintf("%s4%s", provider0_up, passwd);
-  if(!secret) {
+  /* provider 0 uppercase */
+  secret = curl_maprintf("%s4%s", provider0,
+                         data->state.aptr.passwd ?
+                         data->state.aptr.passwd : "");
+  if(!secret)
     goto fail;
-  }
 
-  HMAC_SHA256(secret, strlen(secret),
-              date, strlen(date), tmp_sign0);
-  HMAC_SHA256(tmp_sign0, sizeof(tmp_sign0),
-              region, strlen(region), tmp_sign1);
-  HMAC_SHA256(tmp_sign1, sizeof(tmp_sign1),
-              service, strlen(service), tmp_sign0);
-  HMAC_SHA256(tmp_sign0, sizeof(tmp_sign0),
-              request_type, strlen(request_type), tmp_sign1);
-  HMAC_SHA256(tmp_sign1, sizeof(tmp_sign1),
-              str_to_sign, strlen(str_to_sign), tmp_sign0);
+  HMAC_SHA256(secret, strlen(secret), date, strlen(date), sign0);
+  HMAC_SHA256(sign0, sizeof(sign0), region, strlen(region), sign1);
+  HMAC_SHA256(sign1, sizeof(sign1), service, strlen(service), sign0);
+  HMAC_SHA256(sign0, sizeof(sign0), request_type, strlen(request_type), sign1);
+  HMAC_SHA256(sign1, sizeof(sign1), str_to_sign, strlen(str_to_sign), sign0);
 
-  sha256_to_hex(sha_hex, tmp_sign0, sizeof(sha_hex));
+  sha256_to_hex(sha_hex, sign0, sizeof(sha_hex));
 
+  /* provider 0 uppercase */
   auth_headers = curl_maprintf("Authorization: %s4-HMAC-SHA256 "
                                "Credential=%s/%s, "
                                "SignedHeaders=%s, "
                                "Signature=%s\r\n"
                                "%s\r\n",
-                               provider0_up,
+                               provider0,
                                user,
                                credential_scope,
                                sha_hex,
                                date_header);
@@ -552,16 +500,10 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
   data->state.aptr.userpwd = auth_headers;
   data->state.authhost.done = TRUE;
   ret = CURLE_OK;
 
 fail:
-  free(provider0_low);
-  free(provider0_up);
-  free(provider1_low);
-  free(provider1_mid);
-  free(region);
-  free(service);
   Curl_dyn_free(&canonical_headers);
   Curl_dyn_free(&signed_headers);
   free(canonical_request);
   free(request_type);
   free(credential_scope);
diff --git a/tests/libtest/lib1955.c b/tests/libtest/lib1955.c
index b3e1ec156..053c3e1ae 100644
--- a/tests/libtest/lib1955.c
+++ b/tests/libtest/lib1955.c
@@ -3,11 +3,11 @@
  *  Project                     ___| | | |  _ \| |
  *                             / __| | | | |_) | |
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
  * are also available at https://curl.haxx.se/docs/copyright.html.
  *
-- 
2.37.2

bagder (commented):
I proposed a larger patch

lib/http_aws_sigv4.c
```diff
@@ -44,67 +44,252 @@
...
+  if(Curl_checkheaders(data, STRCONST("Host"))) {
+    head = NULL;
+  }
+  else {
+    char full_host[FULL_HOST_LEN];
+
+    if(data->state.aptr.host) {
+      size_t pos;
+
+      strncpy(full_host, data->state.aptr.host, FULL_HOST_LEN);
```
bagder, line 169: This is arbitrarily cutting off the header, which then will make it fail. Shouldn't it rather return an error if the existing header is too long?

lib/http_aws_sigv4.c
```diff
@@ -44,67 +44,252 @@
...
+      full_host[FULL_HOST_LEN - 1] = 0;
+      /* remove /r/n as the separator for canonical request must be '\n' */
+      pos = strcspn(full_host, "\n\r");
+      full_host[pos] = 0;
+    }
+    else {
+      msnprintf(full_host, FULL_HOST_LEN, "host:%s", hostname);
+      full_host[FULL_HOST_LEN - 1] = 0;
```
bagder, line 177: msnprintf() null-terminates the string, no need to do it again.

lib/http_aws_sigv4.c
```diff
@@ -44,67 +44,252 @@
...
+
+/* FQDN + host: */
+#define FULL_HOST_LEN (255 + sizeof("host:"))
+
+/* string been x-PROVIDER-date:TIMESTAMP, I need +1 for ':' */
+#define DATE_FULL_HDR_LEN (DATE_HDR_KEY_LEN + TIMESTAMP_SIZE + 1)
+
+static CURLcode make_headers(struct Curl_easy *data,
+                             const char *hostname,
+                             char *timestamp,
```
bagder, line 135: maybe add a comment here explaining that the timestamp buffer needs to have a certain size or things will burn!

bagder:
also: **"This branch has conflicts that must be resolved"**

lib/http_aws_sigv4.c
outscale-mgo: Thanks for the review,
I've just add above the function declaration.
```
/* strlen(timestamp) need to be inferior to TIMESTAMP_SIZE */
```

lib/http_aws_sigv4.c
```diff
@@ -44,67 +44,259 @@
...
+
+#define DATE_HDR_KEY_LEN (MAX_SIGV4_LEN + sizeof("X--Date"))
+
+/* FQDN + host: */
+#define FULL_HOST_LEN (255 + sizeof("host:"))
+
+/* string been x-PROVIDER-date:TIMESTAMP, I need +1 for ':' */
+#define DATE_FULL_HDR_LEN (DATE_HDR_KEY_LEN + TIMESTAMP_SIZE + 1)
+
+/* strlen(timestamp) need to be inferior to TIMESTAMP_SIZE */
```
bagder, line 133: I propose:
~~~
/* sizeof(timestamp) needs to be at least TIMESTAMP_SIZE  bytes */

bagder (commented):
I think is is ready for merge. I could only find one nit, and that's in a comment.

lib/http_aws_sigv4.c
outscale-mgo: maybe it's my english lacking here, but doesn't "need to be at last", mean that strlen(timestamp) must be 17 or more ? which would mean an overflow ?
I've just patch to comment to say:
```
strlen(timestamp) should be 16, and having it bigger than TIMESTAMP_SIZE could cause an overflow
```
it's more tedious, but more precise I guess.

bagder: strlen() is the length of a string, not the size of a buffer so that is just plain wrong to describe how large a buffer is. Also, the word "inferior" does not mean shorter or smaller, it means "worse" or "less good", so it makes no sense there.

My version of the comment describes how large the buffer needs to be. It needs to be at least 17 bytes. Right?

outscale-mgo: I've change it to that
```C
/*
 * timestamp should point to a buffer of at last TIMESTAMP_SIZE bytes,
 * if it point to a buffer bigger than TIMESTAMP_SIZE bytes
 * code could overflow
 */
```

bagder: Explain this. How would it "overflow" if the buffer is larger than that? The code assumes 17 bytes, it does not touch the bytes beyond that does it?

outscale-mgo: Ok, overflow was wrong sorry (I thought I use strcpy, but I use `n` functions),
```C
msnprintf(date_full_hdr, DATE_FULL_HDR_LEN,
                "x-%s-date:%s", provider1, timestamp);
```
If timestamp is too big here, the signature will fail.

on the other hand timestamp been set like this:
```C
if(!strftime(timestamp, sizeof(timestamp), "%Y%m%dT%H%M%SZ", &tm)) {
``` 

I don't think it can be something else than 17. 
But if the code calling `make_headers` is modify, then timestamp could be too large for the buffer in which it is copy.

bagder: > If timestamp is too big here, the signature will fail.

Yes, but the comment speaks of the *buffer size* for `timestamp`, now you are talking about the string put into that buffer. They are not the same.

> if(!strftime(timestamp, sizeof(timestamp), "%Y%m%dT%H%M%SZ", &tm)) {

Right, and this code makes sure it doesn't overwrite the buffer. It is works perfectly fine with a larger one.

I claim your updated comment is wrong. The buffer needs to be *at least* TIMESTAMP_SIZE bytes big. If it is bigger, the code still works.

outscale-mgo: ok I've update the comment, thanks for the review.

bagder (approved):

Issue #7965, opened 2021-11-05
`curl --limit-rate -T` does not effectively limit upload transfer rate for small files


wnrph:
### I did this
```terminal
$ small_file=$(mktemp); echo "very small file" > $small_file; wc -c $small_file
16 /tmp/tmp.z6w5UAxAV6

$ curl -v -H Expect: --limit-rate 1 -T $small_file http://httpbin.org/put
*   Trying 54.156.165.4:80...
* TCP_NODELAY set
* Connected to httpbin.org (54.156.165.4) port 80 (#0)
> PUT /put HTTP/1.1
> Host: httpbin.org
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Length: 16
> 
* We are completely uploaded and fine
```
Curl produced the log output (including the last line) immediately. I also captured network traffic with tcpdump which confirmed that the file was sent immediately without limiting the rate.

I get the same result when using the `--no-buffer` option

### I expected the following
I expected curl to limit the rate at which bytes were transmitted. At one byte per second the header and a 16 byte file should take at least a minute or so to upload

### curl/libcurl version
```terminal
curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3
Release-Date: 2020-01-08
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS brotli GSS-API HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets
```
#### uname -a
Linux 5.4.0-88-generic #99-Ubuntu SMP Thu Sep 23 17:29:00 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
#### lsb_release
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal

bagder:
Correct, and I think the best "fix" for this is to enhance the documentation to set the expectations right.

[...]

PR #971, opened 2016-08-19
Improve speed limits, not based on average speeds anymore

jjk-jacky:
Speed limits (from CURLOPT_MAX_RECV_SPEED_LARGE & CURLOPT_MAX_SEND_SPEED_LARGE)
were applied simply by comparing limits with the cumulative average speed of the
entire transfer; While this might work at times with good/constant connections,
in other cases it can result to the limits simply being "ignored" for more than
"short bursts" (as told in man page).

Consider a download that goes on much slower than the limit for some time
(because bandwidth is used elsewhere, server is slow, whatever the reason), then
once things get better, curl would simply ignore the limit up until the average
speed (since the beginning of the transfer) reached the limit.
This could prove the limit useless to effectively avoid using the entire
bandwidth (at least for quite some time).

So instead, we now use a "moving starting point" as reference, and every time at
least as much as the limit as been transferred, we can reset this starting point
to the current position. This gets a good limiting effect that applies to the
"current speed" with instant reactivity (in case of sudden speed burst).

PR #7970, opened 2021-11-06
limit-rate.d: this is average over several seconds

bagder:


docs/cmdline-opts/limit-rate.d
```diff
@@ -17,6 +17,9 @@ Appending 'k' or 'K' will count the number as kilobytes, 'm' or 'M' makes it
 megabytes, while 'g' or 'G' makes it gigabytes. The suffixes (k, M, G, T, P)
 are 1024 based. For example 1k is 1024. Examples: 200K, 3m and 1G.
 
+The rate limiting logic works on averaging the transfer speed to no more than
+the set threshold over a period of multiple seconds.
```
jay, line 21: Ref: https://github.com/curl/curl/issues/7965

If the goal here is informing the user that rate limiting split second transfers has no effect then I think you should say that, such as "Rate limiting is averaged over several seconds. For shorter transfers it will have no effect."

bagder: Shorter than what? And it will/might have *some* effect depending on factors. I'm vague to avoid having to make specific claims that I can't back up.

jay: If it might have some effect then disregard.

wnrph: This information appears to be outdated. At least the send rate is already limited immediately after sending the header.

wnrph: The limiting logic also appears to have been changed such that it is no longer is based on averaging (see 4b86113f5ed66270364738a351ebb5bb23cd0882, #971)

bagder: I maintain my added sentence is still correct and might help users understand what to expect from this option.

wnrph, line 21: @fabiankeil What are your thoughts? Is this an accurate description of how the rate limiting works?

fabiankeil:
wnrph ***@***.***> wrote on 2021-11-09:

> @fabiankeil What are your thoughts? Is this an accurate description of how the rate limiting works?

I don't have time to look at the current code right
now but I consider the addition proposed by Daniel
an improvement.

For my Privoxy tests it would be great if I could
limit the rate for header data as well but I suppose
for most use cases it doesn't matter.

While the documentation could make the fact that
the limit isn't applied to the client header data
more obvious it could be argued that this is unrelated
to this pull request.

Fabian

PR #7992, opened 2021-11-11
CURLOPT_REDIRECT_STEP

fsbs:
This option tells libcurl to perform only a single transfer (like with CURLOPT_FOLLOWLOCATION *disabled*), but to preserve the redirect state, so that the next transfer will perform the next redirect step in the redirect chain. You can also think of it as a middle-ground between disabling or enabling CURLOPT_FOLLOWLOCATION: the user lets libcurl do its redirect magic, but performs each redirect step individually so that other things can be changed between redirects.

Example:

```c
CURLcode res;
char *url = "https://httpbingo.org/redirect/2";

curl_easy_setopt(curl, CURLOPT_REDIRECT_STEP, 1L);

do {
    printf("--- redirect step: %s\n", url);
    curl_easy_getinfo(curl, CURLINFO_REDIRECT_URL, &url);
} while(res == CURLE_OK && url != NULL);
```

* When the redirect chain reaches its end the next transfer will start from the beginning of the redirect chain.
* The user can use CURLINFO_REDIRECT_URL to see whether there's any redirect steps left or whether the end of the chain has been reached.
* The user can interrupt the redirect chain by another call to curl_easy_setopt() with CURLOPT_REDIRECT_STEP, which will reset the redirect state.
* The getinfo values are not reset between redirects but are instead updated, so that for example total time refers to the start of the redirect chain.
* When using the multi interface, the handle must be removed and re-added for each redirect step, just like with FOLLOWLOCATION disabled.

I realize this option deviates from the convention of each perform completely resetting the previous easy handle's state (data->state) in Curl_pretransfer(), always starting from the original request.

But I would find this option extremely useful so that, for example, the user can modify non-redirect-related options based on which URL is requested in each redirect step, or switch multi handles and share objects between redirects. One concrete example is Tor-like domain-based stream isolation, using different proxy credentials for each domain.

Without this option the user has to re-implement libcurl's redirect logic - which goes beyond just Curl_follow() and leads to a spiral of duplicating libcurl's other parts, such as "ignorebody" logic for instance, which further depends on duplicating libcurl's header parsing, etc. It gets very messy to attempt to duplicate all of these libcurl's internals with just easy setopt options since they don't map 1:1 to data->state. And it's not possible for the user to mess with the data->state directly because it gets reset in Curl_pretransfer().  For some things (like "ignorebody") callbacks need to be abused to change the behavior mid-transfer. This situation is unfortunate when you just want to be able to change CURLOPT_PROXY between redirects for instance and leave everything else to libcurl.

In short, this option is for those who are completely satisfied with how libcurl does redirects, but need to do other things between redirects.

I haven't written documentation and tests yet because I'd like to get some feedback if this option is acceptable at all.


bagder:
> I realize this option deviates from the convention of each perform completely resetting the previous easy handle's state (data->state) in Curl_pretransfer(), always starting from the original request.

This is my only real concern about this approach. It breaks the standard paradigm and in a rather subtle way too. I'm now wrestling with if I think that's a showstopper, fine or if there's a middle-ground somehow...

fsbs:
@badger My initial idea was a redirect callback that would be invoked right before the handle goes back to CONNECT state for a further redirect. In the callback the user could then change things up. After the user's callback returns, Curl_pretransfer() or a modified version of it would be called in a similar fashion, updating some fields while preserving redirect-related state.

But there are also problems with the callback approach: using *setopt() in callbacks in the middle of a perform breaks the paradigm in yet another way. And multi handle switching in the callback would be impossible - or even if made possible it would break yet another paradigm: removing an easy handle from a multi before MSGSENT should interrupt the whole request.

I think the current approach of the handle going to MSGSENT and instead preserving part of the previous `data->state` on next perform is still better.

Hypothetically the option could also be turned into a separate function instead, something like "curl_easy_perform_redirect()" and "curl_multi_add_handle_redirect()", which would internally set the same bit (`data->set.http_redirect_step`). This would avoid one easy option deviating from the paradigm of existing functions and make the API more explicit. However the use case here is limited to just one protocol, so I imagine that's even worse.

bagder (changes requested):
I believe that this new way of storing and using state in the handle in subsequent transfers works too different and doesn't follow the standard API paradigms. I don't think we should do it this way as I'm worried this will confuse users and lead to worse problems long-term.

danielgustafsson:
@fsbs are you going to pursue a changed patch responding to @bagder 's comments, or should we close this PR in the meantime?

bagder:
Closing. No responses and the current approach is not deemed a working approach for libcurl.

PR #8002, opened 2021-11-13
Add -lm to configure for rustls build.

jsha:
Note: The list of libraries that rustc tells us we need to include is
longer, but also includes some more platform-specific libraries that I
am not sure how to effectively incorporate. Adding just -lm seems to
solve an immediate problem, so I'm adding just that.

Related: https://github.com/rustls/rustls-ffi/pull/213


bagder (approved):

bagder:
Thanks!

PR #8005, opened 2021-11-13
configure: better diagnostics if hyper is built wrong

### LINKED ISSUE #8001: configure: detect an incorrectly-built Hyper
### I did this

In Hyper directory:

```
cargo build
```

In curl directory:

```
./configure --with-hyper=$HOME/rust/hyper/
```

### I expected the following

Success!

### I got the following

    configure: error: --with-hyper but hyper was not found. See docs/HYPER.md.

The problem is that I didn't build Hyper with the appropriate feature flags. The message is confusing because Hyper _did_ exist and was in the right place, but didn't have the correct symbols. We should add a check to configure that the file (libhyper.a) exists in the given directory. If the file exists but the build fails, we can specifically say "hyper was found but it was built with the wrong flags. See docs/HYPER.md"

/cc @kevinburke

bagder:
If hyper is indeed present in the specified directory but couldn't be
used to find the correct symbol, then offer a different error message to
better help the user understand the issue.

Sugggested-by: Jacob Hoffman-Andrews
Fixes #8001

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -703,7 +703,8 @@ if test X"$want_hyper" != Xno; then
     LIBS="$LIB_HYPER $LIBS"
 
     if test "x$cross_compiling" != "xyes"; then
-      DIR_HYPER=`echo $LD_HYPER | $SED -e 's/^-L//'`
+      dnl remove -L, separate with colon if more than one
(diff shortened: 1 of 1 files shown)


bagder:
@jsha something like this?

jsha:
Looks great!

PR #8008, opened 2021-11-13
OpenSSL lib path changed to root since OpenSSL_1_1_0-pre3

tsengjun:
For more details, please refer to this commit https://github.com/openssl/openssl/commit/777a288270cb4c3671f9095630415207d867396b


bagder:
Isn't that OpenSSL commit a change in their build system only? Does it really change how/where they *install* the library? We do after all pick up the lib in the installed location.

tsengjun:
From the code difference between branches, OpenSSL updated their build system since 1.1.0.
Since that time, libraries are not in the lib directory anymore. They introduced build.info to describe the new library building process. And put the libcrypto.lib, libssl.lib in its root folder.

tsengjun:
About the AppVeyor build, it is the same 1.1.1 branch. And the result is a success. I think it is because the build system change does not impact the installation of DLL. Only a static link will find the .lib files in that exact location. @bagder

bagder:
curl builds should use installed openssl paths, not the build paths, so why does it matter if they change where they put the lib in their build?

tsengjun:
@bagder Thank you for helping me to sort it out!
I missed the "nmake install_dev" step when building OpenSSL, which will copy all the .lib files into the target folder.

bagder:
Indeed! And then we don't need this PR, right?

tsengjun:
Yes, all problems are clear. Close this PR as not a problem.
@bagder Thank you for reviewing this PR!

PR #8022, opened 2021-11-16
tests/CI.md: add more information on CI environments

### LINKED ISSUE #8012: Documentation of existing CI behavior
Looking in the source code it's clear that there's a build runner that is supposed to be running the rustls tests, however, it's difficult to actually find that build because it's not linked in Github's checks UI. Further the fact that Github checks reports ~45 different tests made me think that I just wasn't looking hard enough and it must be there somewhere.

It turns out there is a whole other builder suite at https://curl.zuul.vexxhost.dev/builds?job_name=curl-debug-rustls (and maybe others?) that is running CI tests. While there are issues with that builder (#7522) we could probably do a better job of documenting the different CI jobs that are running currently.

I propose adding `tests/CI.md` with documentation on:

- The different locations that CI checks are run
- Which location runs which subset of tests
- Where you should look to verify your PR did not introduce regressions
- Who owns keys or knows how to set up CI (maybe the answer is just @bagder)

I'd also like to add a link in `zuul.d/jobs.yaml` with a link to the curl.zuul.vexxhost.dev page, just to make that link more explicit.

I would be happy to take a first stab at this.

kevinburke:
Fixes #8012.

### DIFF
--- a/tests/CI.md
+++ b/tests/CI.md
@@ -1,0 +1,103 @@
+# Continuous Integration for curl
+
+Curl runs in many different environments, so every change is run against a large
+number of test suites.
+
+Every pull request is verified for each of the following:
+
+ - ... it still builds, warning-free, on Linux and macOS, with both
+   clang and gcc
+ - ... it still builds fine on Windows with several MSVC versions
+ - ... it still builds with cmake on Linux, with gcc and clang
+ - ... it follows rudimentary code style rules
+ - ... the test suite still runs 100% fine
+ - ... the release tarball (the "dist") still works
+ - ... it builds fine in-tree as well as out-of-tree
+ - ... code coverage does not shrink drastically
+ - ... different TLS backends still compile and pass tests
+
+If the pull-request fails one of these tests, it will show up as a red X and
+you are expected to fix the problem. If you do not understand when the issue is
+or have other problems to fix the complaint, just ask and other project
+members will likely be able to help out.
+
+Consider the following table while looking at pull request failures:
+
+ | CI platform as shown in PR          | State  | What to look at next       |
+ | ----------------------------------- | ------ | -------------------------- |
+ | CI / codeql                         | stable | quality check results      |
+ | CI / fuzzing                        | stable | fuzzing results            |
+ | CI / macos ...                      | stable | all errors and failures    |
+ | Code scanning results / CodeQL      | stable | quality check results      |
+ | FreeBSD FreeBSD: ...                | stable | all errors and failures    |
+ | LGTM analysis: Python               | stable | new findings               |
+ | LGTM analysis:  C/C++               | stable | new findings               |
+ | buildbot/curl_winssl_ ...           | stable | all errors and failures    |
+ | continuous-integration/appveyor/pr  | stable | all errors and failures    |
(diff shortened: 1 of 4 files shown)


tests/CI.md
```diff
@@ -0,0 +1,101 @@
...
+### Azure
+
+The following tests are run in Microsoft Azure CI environment:
+
+- Ubuntu tests with a variety of different compilation options.
+- Windows tests with a variety of different compilation options.
+
+These are all configured in `.azure-pipelines.yml`.
+
+As of November 2021 Daniel Stenberg is the only person with administrator access
```
kevinburke, line 72: I'm not sure if this, and the lines below, are accurate, please correct me if they are not.

bagder: @mback2k is also a member of the curl "org" on Azure Pipelines. I think it could also be mentioned (for all the services) that additional admins/group members can be added on request when needed.

tests/CI.md
```diff
@@ -0,0 +1,103 @@
...
+BoringSSL, etc).
+- Curl compiled with different C compilers.
+
+As of November 2021, the tests run (sometimes) but do not run consistently and
+do not report results to the Github checks runner - you need to manually check
+for failures. See [#7522](https://github.com/curl/curl/issues/7522) for more
+information.
+
+As of November 2021 Daniel Stenberg is the only person with administrator access
+to the Zuul CI environment.
```
bagder, line 101: I'm not sure I have any administrator rights there...!

tests/CI.md
```diff
@@ -0,0 +1,103 @@
...
+
+These are configured in `zuul.d` and have test runners in `scripts/zuul`.
```
bagder, line 103: We're also using **Circle CI** and **Cirrus CI**...

tests/CI.md
```diff
@@ -0,0 +1,103 @@
...
+As of November 2021 Daniel Stenberg is the only person with administrator access
+to the Azure CI environment.
+
+### Appveyor
+
+Appveyor runs a variety of different Windows builds, with different compilation
+options.
+
+As of November 2021 Daniel Stenberg is the only person with administrator access
+to the Appveyor CI environment.
```
bagder, line 81: Additional "team members" in the "curl org" on Appveyor are: @mback2k, @jay, @vszakats, @dfandrich and @danielgustafsson

tests/CI.md
kevinburke: Oh, ok. I guess is there anything you need to log into, or configure, at curl.vexxhost.dev, or was that all set up for you? Who would you talk to about it?

tests/CI.md
kevinburke: Updated

tests/CI.md
kevinburke: Updated

tests/CI.md
kevinburke: Updated

tests/CI.md
bagder: It was setup for me/us.  @mnaser @albinvass did the job.

bagder:
Thanks!

PR #8023, opened 2021-11-16
Build curl http3 without shared libraries to prevent conflicts with existing lib curl

obonaventure:
When building curl, it is safer to disable the shared library to avoid conflicts with a shared version of libcurl that could already have been installed on the system. This patch disables the shared libraries when building curl that supports http3.

### DIFF
--- a/docs/HTTP3.md
+++ b/docs/HTTP3.md
@@ -130,7 +130,7 @@ Build curl:
      % git clone https://github.com/curl/curl
      % cd curl
      % autoreconf -fi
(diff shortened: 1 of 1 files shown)


bagder (commented):
I think this might be getting a little too much into the weeds and is speculating on what the user wants and can do. It's not really *safer* without shared libraries, you remove the risk that the user loads the wrong shared library in run-time. We need to put some trust in the user's ability as well and I don't want us to add build options that might trick the user into thinking that's necessary for building a HTTP/3-enabled curl.

PR #8028, opened 2021-11-17
configure: fix runtime-lib detection on macOS

bwncp:
With a non-standard installation of openssl we get this error:

    checking run-time libs availability... failed
    configure: error: one or more libs available at link-time are not available run-time. Libs used at link-time: -lnghttp2 -lssl -lcrypto -lssl -lcrypto -lz

There's already code to set `LD_LIBRARY_PATH` on Linux, so set `DYLD_LIBRARY_PATH` equivalent on macOS.


bagder (approved):
Nice approach! :+1:

bagder:
Thanks!

PR #8030, opened 2021-11-17
Don't set _USRDLL on a static Windows build

donny-dont:
The `_USRDLL` macro should only be set when building a `.dll`.


donny-dont:
An issue was reported at https://github.com/microsoft/vcpkg/issues/12763 and in vcpkg it was fixed in https://github.com/microsoft/vcpkg/pull/12800 but this version of the fix fits more with the surrounding CMake code.

bagder (approved):
Looks perfectly reasonable to me!

bagder:
Thanks!

PR #8035, opened 2021-11-19
curl: modified search for files in "homedir"

### LINKED ISSUE #8033: curl tries to read known_hosts from wrong location
### I did this
curl -u user:pass -T file.dat sftp://host.domain.tld/folder

### I expected the following

upload "file.dat" to host.domain.tld/folder

### what happened

although host.domain.tld is contained in known_hosts there's an error message:
curl: (60) SSL peer certificate or SSH remote key was not OK

AFAICT, curl tries to load known_hosts from %APPDATA%\\.ssh (user\\AppData\\Roaming\\.ssh), while it should look at %USERPROFILE%\\.ssh (user\\.ssh), which is user's actual home folder and is used by OpenSSH integrated in Windows as well.

BTW, would be great to have an example of how to get the hash for --hostpubmd5 and --hostpubsha256 in the documentation (ie, what data to hash exactly).

### curl/libcurl version

[curl -V output]
```
curl 7.80.0 (x86_64-pc-win32) libcurl/7.80.0 OpenSSL/3.0.0 (Schannel) zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.2 libssh2/1.10.0 nghttp2/1.46.0 libgsasl/1.10.0
Release-Date: 2021-11-10
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI TLS-SRP UnixSockets zstd
```

### operating system

Windows 10 21H1

bagder:
curl now iterates over all the environment variables
trying to access the file in question until it finds it. Last resort is
then getpwuid() if available. Previously it would first try to find a
home directory and if that was set, insist on checking only that
directory for the file.

The Windows specific checks are now done differently too and in this
order:

1 - %USERPROFILE%
2 - %APPDATA%
3 - %USERPROFILE%\\Application Data\\

The windows order is modified to match how the Windows 10 ssh tool works
when it searches for .ssh/known_hosts.

Reported-by: jeffrson on github
Fixes #8033


jzakrzewski:
Shouldn't some documentation also be updated?

bagder:
indeed!

bagder:
I note that we don't really document anywhere how it finds `known_hosts`, there's only documentation for how it finds `.curlrc`  - which on the other hand follows the same logic.

bagder:
gosh, thanks @jay!

bagder:
Test 1268 fails on Freebsd because it can't find a known_hosts. It shouldn't - but it still does in many cases so the test is a bit flaky.

I'm working on #8040 in the meantime to avoid trying to set the known_hosts unless SFTP or SCP is actually used.

bagder:
This looks broken on Windows according to the CI builds atm.

bagder:
@jay can I ask you to take a quick look to see what makes this go so horribly wrong on Windows?

jay:
I see a ton of failed tests on azure due to `protocol FAILED` which doesn't make sense. I set azure to re-run to see what happens.

bagder:
> so there may be something else going on, will investigate.

I interpreted that as a curl crash...

jay:
> I interpreted that as a curl crash...

I am trying to reproduce this locally without any success using msys2 mingw builds, but I don't have the same gcc settings.

jay:
I couldn't reproduce locally in msys2. All the segfault failures in the various CIs it seems are from msys builds.

AppVeyor has a lot of builds that are failing for an unrelated reason, "Build execution time has reached the maximum allowed time for your plan (60 minutes)."

bagder:
This drives me mad...

jay:
I found a configuration that reproduces this which initially did not. We have this for azure:
`./buildconf && ./configure --host=x86_64-w64-mingw32 --build=x86_64-w64-mingw32 --prefix=/mingw64 --enable-debug --enable-werror --enable-sspi --with-schannel --with-winidn --with-libssh2"`

I thought that would cross compile for 64-bit but actually it built 32-bit curl which I guess does not show the issue. I think --target is missing? That is probably because I did it from MINGW32 prompt not MINGW64. The latter builds a 64-bit curl and I can reproduce there, so I will look into it tonight.

jay:
The crash is caused by memory corruption that occurs after a double free. parseconfig calls findfile which `curl_free`'s the `home` pointer even though that may then be returned and `curl_free`'d again by parseconfig.

The crash was not reproducible in gdb so I sprinkled some (fprintf)(stderr,....  which did not show any output without (fflush)(stderr). I wonder if this is similar to what @mback2k saw in the CI where it isn't always writing out? That is weird and I don't think it's a bug in curl. stderr is supposed to write out immediately, so if there's a crash we still have the output.

A separate possible crash is calling free() instead of curl_free() on curl_maprintf.(). That is only an issue for a shared library libcurl on an OS that may load two copies of the C runtime, one for libcurl and one for the curl tool (and have separate memory management).

I think we should not allow functions to return memory that needs to be curl_free'd unless they're curl_ API functions. IMO it is too easy to lose track of. To that end I've modified findfile to return a regular heap allocated string to be free'd instead of curl_free'd.

curl_free is possible to fall into with aprintf,vaprintf which may be defined to curlx_maprintf,curlx_mvaprintf respectively.
https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/lib/curlx.h#L112-L113

bagder:
Thanks a lot @jay!

PR #8038, opened 2021-11-19
Remove rustls comment about checking handshaking

jsha:
The comment is incorrect in two ways:
 - It says the check needs to be last, but the check is actually first.
 - is_handshaking actually starts out true.
 
Prompted by https://github.com/rustls/rustls-ffi/issues/134.


bagder (approved):

bagder:
Thanks!

Issue #8041, opened 2021-11-22
 Issue 41258: curl:curl_fuzzer_gopher: Heap-buffer-overflow in seturl


bagder:
### I did this

This is a report filed by oss-fuzz ([here](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41258)), but with restricted permissions. I'm reposting parts of it here for greater visibility. Me and @cmeister2 have not managed to reproduce this locally.

The issue happens with the fuzz executable sets a URL that is very short and simple:
TLVHeader(type='CURLOPT_URL' (1), length=7, data=b'FI:/\x00\x01\x00')

It then points to a buffer overflow in seturl:
~~~
  | [Environment] ASAN_OPTIONS=alloc_dealloc_mismatch=0:allocator_may_return_null=1:allocator_release_to_os_interval_ms=500:allow_user_segv_handler=0:check_malloc_usable_size=0:detect_leaks=1:detect_odr_violation=0:detect_stack_use_after_return=1:fast_unwind_on_fatal=0:handle_abort=2:handle_segv=2:handle_sigbus=2:handle_sigfpe=2:handle_sigill=2:max_uar_stack_size_log=16:print_scariness=1:print_summary=1:print_suppressions=0:quarantine_size_mb=64:redzone=32:strict_memcmp=1:strip_path_prefix=/workspace/:symbolize=0:use_sigaltstack=1
-- | --
  | +----------------------------------------Release Build Stacktrace----------------------------------------+
  | Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds-afl_curl_d1a7f12cc2e5055727a9c66d5eca203f3c8f5a6c/revisions/curl_fuzzer_gopher /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash
  | Time ran: 0.030854463577270508
  |  
  | =================================================================
  | ==145209==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x604000038cb5 at pc 0x00000048a2b8 bp 0x7ffe9e73a330 sp 0x7ffe9e739af0
  | READ of size 9 at 0x604000038cb5 thread T0
  | SCARINESS: 24 (9-byte-read-heap-buffer-overflow)
  | #0 0x48a2b7 in strcpy /src/llvm-project/compiler-rt/lib/asan/asan_interceptors.cpp:424:5
[...]

PR #8043, opened 2021-11-22
urlapi: cleanup scheme parsing

icing:
Just a suggestion coming into my mind while thinking about #8041. Feel free to ignore.

`urlapi.c: Curl_is_absolute_url()`

  * always have a defined return in `buf`(if given)
  * only copy when scheme is detected

### DIFF
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -258,7 +258,7 @@ bool Curl_is_absolute_url(const char *url, char *buf, size_t buflen)
     if(buf) {
       buf[i] = 0;
       while(i--) {
-        buf[i] = TOLOWER(url[i]);
+        buf[i] = (char)TOLOWER(url[i]);
       }
     }
     return TRUE;
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -235,26 +235,31 @@ static void strcpy_url(char *output, const char *url, bool relative)
 bool Curl_is_absolute_url(const char *url, char *buf, size_t buflen)
 {
   size_t i;
+  char s;
+
+  if (buf && buflen)
+    buf[0] = 0; /* always leave a defined value in buf */
 #ifdef WIN32
   if(STARTS_WITH_DRIVE_PREFIX(url))
     return FALSE;
 #endif
-  for(i = 0; i < buflen && url[i]; ++i) {
-    char s = url[i];
-    if((s == ':') && (url[i + 1] == '/')) {
-      if(buf)
+  for(i = 0; i < buflen; ++i) {
+    if((s = url[i]) &&
+      (ISALNUM(s) || (s == '+') || (s == '-') || (s == '.') )) {
+      /* RFC 3986 3.1 explains:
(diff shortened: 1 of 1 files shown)


bagder:
It doesn't lowercase the string, which makes test 1560 fail: https://cirrus-ci.com/task/5019655648575488?logs=test#L2482

And when writing this, I also realize that the check on https://github.com/curl/curl/blob/18270893abdb19f0ca170c118f8a2847dbd304be/lib/urlapi.c#L820 should rather use plain strcmp...

icing:
made the change.

lib/urlapi.c
```diff
@@ -251,13 +251,15 @@ bool Curl_is_absolute_url(const char *url, char *buf, size_t buflen)
...
-        break;
+      break;
     }
   }
   if(i && (url[i] == ':') && (url[i + 1] == '/')) {
     if(buf) {
-      memcpy(buf, url, i);
       buf[i] = 0;
+      while(i--) {
+        buf[i] = TOLOWER(url[i]);
```
bagder, line 261: I think this needs the typecast from the original to avoid warnings:
~~~c
buf[i] = (char)TOLOWER(url[i]);

bagder:
I think this identifies another bug in the previous code as well. The third argument to `Curl_is_absolute_url` is confusingly used both as buffer length as well as maximum length acceptable for a scheme. I think maybe we should use it only as a buffer length and *always* use `MAX_SCHEME_LEN` as the max scheme length?

Either way, I fixed test 1560 by adding this patch:
~~~diff
diff --git a/lib/urlapi.c b/lib/urlapi.c
index 9005f040b..f1ec4713d 100644
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -816,11 +816,11 @@ static CURLUcode seturl(const char *url, CURLU *u, unsigned int flags)
     return CURLUE_OUT_OF_MEMORY;
 
   hostname = &path[urllen + 1];
   hostname[0] = 0;
 
-  if(Curl_is_absolute_url(url, schemebuf, sizeof(schemebuf))) {
+  if(Curl_is_absolute_url(url, schemebuf, sizeof(schemebuf) - 1)) {
     url_has_scheme = TRUE;
     schemelen = strlen(schemebuf);
   }
 
   /* handle the file: scheme */
@@ -1525,11 +1525,11 @@ CURLUcode curl_url_set(CURLU *u, CURLUPart what,
     CURLUcode result;
     char *oldurl;
     char *redired_url;
     CURLU *handle2;
 
-    if(Curl_is_absolute_url(part, NULL, MAX_SCHEME_LEN + 1)) {
+    if(Curl_is_absolute_url(part, NULL, MAX_SCHEME_LEN)) {
       handle2 = curl_url();
       if(!handle2)
         return CURLUE_OUT_OF_MEMORY;
       result = parseurl(part, handle2, flags);
       if(!result)

bagder:
On second thoughts, maybe just apply my fix and then work on that little snag in a separate PR.

bagder (approved):

bagder:
Thanks!

PR #8048, opened 2021-11-23
Update tags and clear up grammar and add info.

A-Yamout:


README.md
```diff
@@ -2,17 +2,15 @@
 
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/63/badge)](https://bestpractices.coreinfrastructure.org/projects/63)
 [![Coverity passed](https://scan.coverity.com/projects/curl/badge.svg)](https://scan.coverity.com/projects/curl)
-[![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/l1vv31029huhf4g4?svg=true)](https://ci.appveyor.com/project/curlorg/curl)
-[![Azure DevOps Build Status](https://dev.azure.com/daniel0244/curl/_apis/build/status/curl.curl?branchName=master)](https://dev.azure.com/daniel0244/curl/_build/latest?definitionId=1&branchName=master)
```
bagder, line 6: Why remove these?

README.md
```diff
@@ -2,17 +2,15 @@
...
 [![Backers on Open Collective](https://opencollective.com/curl/backers/badge.svg)](#backers)
 [![Sponsors on Open Collective](https://opencollective.com/curl/sponsors/badge.svg)](#sponsors)
 [![Language Grade: C/C++](https://img.shields.io/lgtm/grade/cpp/g/curl/curl.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/curl/curl/context:cpp)
 [![Codacy Badge](https://api.codacy.com/project/badge/Grade/d11483a0cc5c4ebd9da4ff9f7cd56690)](https://app.codacy.com/app/curl/curl)
 [![Fuzzing Status](https://oss-fuzz-build-logs.storage.googleapis.com/badges/curl.svg)](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=-opened&can=1&q=proj:curl)
 
-Curl is a command-line tool for transferring data specified with URL
-syntax. Find out how to use curl by reading [the curl.1 man
+Curl is one of the oldist command-line tool for transferring data specified with a URL
+syntax. It has been used by some of the smallist libraries to flying a helecopter on the moon. Find out how to use curl by reading [the curl.1 man
```
bagder, line 13: This is not correct English. I also don't think we need to say that it is "one of the oldest" tools. I'm not convinced that is always seen as a good thing in the mind of the readers.

README.md
A-Yamout: because they present a bad imagine as in they are failing and it looks neater as one line.

README.md
A-Yamout: Is most used a better term?

README.md
bagder: We could discuss the value of *any* of those silly badges. I figure the point of them is to either provide them all as a way to show project status in several different ways, or we can just remove them and have them replaced with fixed texts instead. The fact that our appveyor and azure CI builds often fail is a known problem we have that we want worked on and solved, not hidden.

README.md
bagder: I think we should leave the "selling" text for the curl.se website front page. I prefer to have the `README.md` to just inform in a neutral tone what curl and libcurl are and what they do.

PR #8052, opened 2021-11-24
cmake: fix leftover CURL_USE_SCHANNEL mistake

bagder:
CURL_USE_SCHANNEL is the proper new name.

Follow-up to 9108da2c26d

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,6 +352,16 @@ if(WIN32)
   check_library_exists_concat("winmm"  getch        HAVE_LIBWINMM)
 endif()
 
+foreach(_LIB GSSAPI OPENLDAP LIBSSH LIBSSH2 BEARSSL MBEDTLS NSS OPENSSL
+        SCHANNEL SECTRANSP WOLFSSL)
+  if(CMAKE_USE_${_LIB})
+    message(FATAL_ERROR "The option CMAKE_USE_${_LIB} is now CURL_USE_${_LIB}.")
+  endif()
+endforeach()
+if(CMAKE_USE_WINSSL)
+  message(FATAL_ERROR "The option CMAKE_USE_WINSSL was renamed to CURL_USE_SCHANNEL.")
+endif()
+
 # check SSL libraries
 # TODO support GnuTLS
 option(CURL_ENABLE_SSL "Enable SSL support" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,14 +352,20 @@ if(WIN32)
   check_library_exists_concat("winmm"  getch        HAVE_LIBWINMM)
 endif()
 
+set(CURL_RECONFIG_REQUIRED 0)
 foreach(_LIB GSSAPI OPENLDAP LIBSSH LIBSSH2 BEARSSL MBEDTLS NSS OPENSSL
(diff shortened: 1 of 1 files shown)


jay:
I think we should error on the outdated names so as not to cause unexpected builds, I think it would be something like this

~~~cmake
foreach(_LIB GSSAPI OPENLDAP LIBSSH LIBSSH2 BEARSSL MBEDTLS NSS OPENSSL SCHANNEL SECTRANSP WOLFSSL)
  if(CMAKE_USE_${_LIB})
    message(FATAL_ERROR "The option CMAKE_USE_${_LIB} was renamed to CURL_USE_${_LIB}.")
  endif()
endforeach()
if(CMAKE_USE_WINSSL)
  message(FATAL_ERROR "The option CMAKE_USE_WINSSL was renamed to CURL_USE_SCHANNEL.")
endif()
~~~
CMAKE_USE_WINSSL isn't in the loop because there is no CURL_USE_WINSSL

bagder:
> I think we should error on the outdated names so as not to cause unexpected builds, I think it would be something like this

I like that. Will amend.

jay:
It would be good to show all the errors at once rather than the user having to run it repeatedly. I just pushed a change that does that:

~~~
-- curl version=[7.81.0-DEV]
CMake Error at CMakeLists.txt:360 (message):
  The option CMAKE_USE_LIBSSH2 was renamed to CURL_USE_LIBSSH2.


CMake Error at CMakeLists.txt:360 (message):
  The option CMAKE_USE_OPENSSL was renamed to CURL_USE_OPENSSL.


CMake Error at CMakeLists.txt:365 (message):
  The option CMAKE_USE_WINSSL was renamed to CURL_USE_SCHANNEL.


CMake Error at CMakeLists.txt:368 (message):
  Reconfig required
~~~

Note for CMAKE_USE_ variables that were enabled by default and already cached from a prior build it will fail. For example CMAKE_USE_LIBSSH2 was on by default even if the user wasn't using it and would be cached, so I got that error for CMAKE_USE_LIBSSH2 even though I don't pass it as an option.

The variables can instead be removed from the cache like `unset(CMAKE_USE_${_LIB} CACHE)` however then the error would not appear a second time if the user didn't specify the variable on the command line. That could be good or bad. If the user is just running cmake without the options (ie just rerunning their build, if this is possible with cmake) then they may not see it a second time. Instead the change I pushed does not unset instead requiring a rebuild.

jzakrzewski:
I think this is the correct way (not touching the cache and letting the user take care of that). 
What I think is missing is a mechanism to suppress the errors. Since the options live in CMAKE_ "namespace" they can gain some meaning from CMake itself, in which curl will start spitting out spurious errors.

bagder:
@jzakrzewski don't you think it is enough if we add the warnings now and leave them in for a few releases and then remove them completely?

bagder:
We could even add a comment about a tentative removal date or something.

bagder:
Thanks @jay, I squashed it all into a single commit: yours.

jzakrzewski:
> @jzakrzewski don't you think it is enough if we add the warnings now and leave them in for a few releases and then remove them completely?

Not a fan, but I won't really object. Although there are people that do "big fast-forward upgrades", I think everybody that does that must be prepared for dealing with weird stuff happening.

PR #8054, opened 2021-11-24
openldap: handle connect phase with a state machine

monnerat:
This is the first step of my work on openldap.
This allows future implementation of explicit TLS and SASL authentication.

### DIFF
--- a/lib/openldap.c
+++ b/lib/openldap.c
@@ -70,6 +70,15 @@
  */
 /* #define CURL_OPENLDAP_DEBUG */
 
+/* Machine states. */
+typedef enum {
+  OLDAP_STOP,           /* Do nothing state, stops the state machine */
+  OLDAP_SSL,            /* Performing SSL handshake. */
+  OLDAP_BIND,           /* Simple bind reply. */
+  OLDAP_BINDV2,         /* Simple bind reply in protocol version 2. */
+  OLDAP_LAST            /* Never used */
+} ldapstate;
+
 #ifndef _LDAP_PVT_H
 extern int ldap_pvt_url_scheme2proto(const char *);
 extern int ldap_init_fd(ber_socket_t fd, int proto, const char *url,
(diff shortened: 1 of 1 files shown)


lib/openldap.c
```diff
@@ -205,8 +259,68 @@ static CURLcode oldap_setup_connection(struct Curl_easy *data,
...
+    passwd.bv_len = strlen(passwd.bv_val);
+  }
+
+  rc = ldap_sasl_bind(li->ld, binddn, LDAP_SASL_SIMPLE, &passwd,
+                      NULL, NULL, &li->msgid);
+  if(rc == LDAP_SUCCESS)
+    state(data, newstate);
+  else
+    result = oldap_map_error(rc, conn->bits.user_passwd?
+                                 CURLE_LOGIN_DENIED: CURLE_LDAP_CANNOT_BIND);
```
bagder, line 287: unexpected indent level ^

lib/openldap.c
```diff
@@ -205,8 +259,68 @@ static CURLcode oldap_setup_connection(struct Curl_easy *data,
...
+  return result;
+}
+
 #ifdef USE_SSL
 static Sockbuf_IO ldapsb_tls;
```
bagder, line 292: I know you didn't add this, but this is really bad isn't it? A static variable like this seems like a potential breaker of many things if we do many LDAP parallel transfers, right?

bagder (approved):

lib/openldap.c
bagder: Strike that. This struct is just full of function pointers and should be fine to reuse.

lib/openldap.c
monnerat: Fixed.

lib/openldap.c
monnerat: I don't like this very much either. But as you say, it should'nt harm.

bagder:
Thanks!

monnerat:
Thanks for pulling.

PR #8063, opened 2021-11-26
updated README (modified some sentences to clarify the understanding)

FahimFBA:
Modified the `README.md` file on the `master` branch.

Tried to clear some sentences so that they can provide the exact meaning they are supposed to give out. I want to modify them more later.


bagder (commented):
Sorry, but I don't think these changes add anything valuable to this document. They just add some words that don't need to be there.

PR #8064, opened 2021-11-26
Docs: Minor punctuation tweaks

GiovanH:
Minor adjustments of punctuation in help output


bagder (commented):
I don't think this changes are necessary or improve the phrases much.

GiovanH:
Entirely reasonable. This is really just something I notice when I run the tool and I figured I might as well make the offer.

PR #8065, opened 2021-11-27
openldap: implement STARTTLS

monnerat:
As this introduces use of `CURLOPT_USE_SSL` option for LDAP, also check this option in ldap.c as it is not supported by this backend.

### DIFF
--- a/docs/cmdline-opts/ssl-reqd.d
+++ b/docs/cmdline-opts/ssl-reqd.d
@@ -1,6 +1,6 @@
 Long: ssl-reqd
 Help: Require SSL/TLS
-Protocols: FTP IMAP POP3 SMTP
+Protocols: FTP IMAP POP3 SMTP LDAP
 Added: 7.20.0
 Category: tls
 Example: --ssl-reqd ftp://example.com
@@ -8,5 +8,9 @@ See-also: ssl insecure
 ---
 Require SSL/TLS for the connection.  Terminates the connection if the server
 does not support SSL/TLS.
+
+This option is handled in LDAP since version 7.81.0. It is fully supported
+by the openldap backend and rejected by the generic ldap backend if explicit
+TLS is required.
 
 This option was formerly known as --ftp-ssl-reqd.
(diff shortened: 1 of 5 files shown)


docs/cmdline-opts/ssl-reqd.d
```diff
@@ -1,6 +1,6 @@
 Long: ssl-reqd
 Help: Require SSL/TLS
-Protocols: FTP IMAP POP3 SMTP
+Protocols: FTP IMAP POP3 SMTP LDAP (openldap only)
```
bagder, line 3: I think the OpenLDAP limitation is better elaborated in the description, not in the protocols line. Maybe also then add that this is supported for LDAP starting in 7.81.0.

docs/cmdline-opts/ssl.d
```diff
@@ -1,6 +1,6 @@
 Long: ssl
 Help: Try SSL/TLS
-Protocols: FTP IMAP POP3 SMTP
+Protocols: FTP IMAP POP3 SMTP LDAP (openldap only)
```
bagder, line 3: Same here.

lib/ldap.c
```diff
@@ -464,6 +464,11 @@ static CURLcode ldap_do(struct Curl_easy *data, bool *done)
 #endif
 #endif /* CURL_LDAP_USE_SSL */
   }
+  else if(data->set.use_ssl > CURLUSESSL_TRY) {
+    failf(data, "LDAP local: explicit TLS not supported");
+    result = CURLE_USE_SSL_FAILED;
```
bagder, line 469: Maybe `CURLE_NOT_BUILT_IN` ? Because this happens when the user selects an option that *could* be supported and work if libcurl was just built differently.

monnerat: Makes sense.
My primary idea was to emphasize on an SSL problem and avoid the same error as if LDAP was not compiled in.

monnerat:
@bagder : Latest push takes all your remarks into account.

bagder (approved):

bagder:
Thanks!

monnerat:
Thanks for pulling.

PR #8066, opened 2021-11-28
auth: compute user:realm:pass digest w/o userhash

gstrauss:
auth: compute user:realm:pass digest w/o userhash

https://datatracker.ietf.org/doc/html/rfc7616#section-3.4.4
  ... the client MUST calculate a hash of the username after
      any other hash calculation ...

Signed-off-by: Glenn Strauss <gstrauss@gluelogic.com>

---

I am implementing HTTP Digest Auth userhash support in lighttpd and I read RFC 7616 Section 3.4.4 differently from how it is currently implemented in cURL from 2b5b37cb9109e7c2e6bfa5ebf54016aff8a1fb48.  The patch in this PR works with my development version of lighttpd, and I am looking for confirmation on whether or not this is an issue in curl, or if I have misunderstood RFC 7616 Section 3.4.4.  Thanks.

### DIFF
--- a/lib/vauth/digest.c
+++ b/lib/vauth/digest.c
@@ -722,8 +722,7 @@ static CURLcode auth_create_digest_http_message(
            unq(nonce-value) ":" unq(cnonce-value)
   */
 
-  hashthis = aprintf("%s:%s:%s", digest->userhash ? userh : userp,
-                                 digest->realm, passwdp);
+  hashthis = aprintf("%s:%s:%s", userp, digest->realm, passwdp);
   if(!hashthis)
     return CURLE_OUT_OF_MEMORY;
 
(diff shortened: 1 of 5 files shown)


gstrauss:
I am looking for feedback before I update the curl digest calculations in the curl HTTP Digest auth tests with `userhash=true`

gstrauss:
I updated tests to make this PR complete.  The hashes in the previous code matched `sha256sum` output, not `shasum -a 512256`, so I updated the algorithm from SHA-512-256 to instead be SHA-256.

gstrauss:
To be clear, the existing code calculates HA1 as H( userhash ":" realm ":" password ) and I believe that HA1 should be calculated as H( username ":" realm ":" password ), the same as with `userhash=false` (the default if the `userhash` param is omitted).

gstrauss:
Manually calculating the `response` in the example in https://datatracker.ietf.org/doc/html/rfc7616#section-3.9.2 suggests that my intepretation of RFC 7616 `userhash` is correct H( username ":" realm ":" password ) -- **not userhash** which curl incorrectly uses and which this PR fixes.

bagder (approved):
Looks correct to me!

bagder:
Thanks!

PR #8076, opened 2021-11-30
HTTP3: update quiche build instructions

ghedo:
The repo repo was re-organized a bit, so the build instructions need to
be updated.


bagder:
Thanks!

PR #8078, opened 2021-12-01
check the return value of BIO_new()

x2018:
Check the return value of BIO_new() in lib/vtls/openssl.c:3836:14 & openssl.c:3565:9 to prevent further memory errors.


bagder (approved):

bagder:
Thanks!

Issue #8079, opened 2021-12-01
CURLOPT_CAINFO_BLOB on mbtedls fails test 678


bagder:
Test 678 seems to fail in mebdtls builds: https://curl.zuul.vexxhost.dev/build/a9732e2c883a4672aabd1c89dc4e40bd

The exact error output is:
~~~
 * Error importing ca cert blob  - mbedTLS: (-0x2180) X509 - The CRT/CRL/CSR format is invalid, e.g. different type expected
~~~
_Originally posted by @bagder in https://github.com/curl/curl/issues/8071#issuecomment-983334607_

/cc @Floni

bagder:
It didn't show up in the CI builds because of the #7522, a bug in the zuul CI which hides most of its runs from github! :cry: 

Your proposed fix sounds perfectly fine!

PR #8071, opened 2021-11-29
mbedTLS: add support for CURLOPT_CAINFO_BLOB

Floni:

PR #8081, opened 2021-12-01
mbedTLS: include NULL byte in blob data length for CURLOPT_CAINFO_BLOB

Floni:
Fix for https://github.com/curl/curl/issues/8079

https://github.com/curl/curl/pull/8071 added support for CURLOPT_CAINFO_BLOB but broke test 678.
The issue is caused by mbedTLS requiring the terminating NULL byte to be included in the length of the data. Since a PEM string is NULL terminated, we can simply increment the length passed to `mbedtls_x509_crt_parse` to meet this requirement.

I this change locally, and now test 678 does succeed.


bagder:
Thanks!

PR #8088, opened 2021-12-03
some changes to comments

coderJeroen:
I made some changes to some comments and to a conditional statement in the file c-hyper.c so the conditional statement isn't evaluated twice.

### DIFF
--- a/lib/base64.c
+++ b/lib/base64.c
@@ -45,7 +45,7 @@
 static const char base64[]=
   "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
-/* The Base 64 encoding with an URL and filename safe alphabet, RFC 4648
+/* The Base 64 encoding with a URL and filename safe alphabet, RFC 4648
    section 5 */
 static const char base64url[]=
   "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
@@ -93,12 +93,12 @@ static size_t decodeQuantum(unsigned char *dest, const char *src)
 /*
  * Curl_base64_decode()
  *
- * Given a base64 NUL-terminated string at src, decode it and return a
- * pointer in *outptr to a newly allocated memory area holding decoded
- * data. Size of decoded data is returned in variable pointed by outlen.
+ * Given a base64 0-terminated string at src, decode it and return a
+ * pointer in *outptr to a newly allocated memory area holding decoded data.
+ * Size of the decoded data is returned in a variable pointed to by *outlen.
  *
- * Returns CURLE_OK on success, otherwise specific error code. Function
- * output shall not be considered valid unless CURLE_OK is returned.
+ * Returns CURLE_OK on success, otherwise a specific error code. Function
+ * output should not be considered valid unless CURLE_OK is returned.
  *
  * When decoded data length is 0, returns NULL in *outptr.
  *
@@ -120,7 +120,7 @@ CURLcode Curl_base64_decode(const char *src,
   *outlen = 0;
   srclen = strlen(src);
 
-  /* Check the length of the input string is valid */
+  /* Check if the length of the input string is valid */
   if(!srclen || srclen % 4)
     return CURLE_BAD_CONTENT_ENCODING;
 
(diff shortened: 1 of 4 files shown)


lib/c-hyper.c
```diff
@@ -357,15 +357,17 @@ CURLcode Curl_hyper_stream(struct Curl_easy *data,
...
+      if (CURL_CSELECT_IN) {
+          if (h->read_waker)
+              hyper_waker_wake(h->read_waker);
+          h->read_waker = NULL;
+      }
+      if (CURL_CSELECT_OUT) {
+          if (h->write_waker)
+              hyper_waker_wake(h->write_waker);
+          h->write_waker = NULL;
+      }
```
bagder, line 370: 1. `make checksrc` will tell you how this breaks the code style
2. You removed the ands from the conditions so this code is now broken!

lib/base64.c
```diff
@@ -283,13 +283,13 @@ static CURLcode base64_encode(const char *table64,
  *
  * Given a pointer to an input buffer and an input size, encode it and
  * return a pointer in *outptr to a newly allocated memory area holding
- * encoded data. Size of encoded data is returned in variable pointed by
- * outlen.
+ * encoded data. Size of the encoded data is returned in a variable pointed
+ * to by *outlen.
  *
- * Input length of 0 indicates input buffer holds a NUL-terminated string.
+ * Input length of 0 indicates input buffer holds a 0-terminated string.
```
bagder, line 289: We call them "NUL-terminated" (or null) through-out the project since 032e838b73578e. I don't think changing the name of them in a few comments is a good idea.

bagder (changes requested):

bagder:
Mixing code changes in one file with comment language fixes in another file in the same PR is also less than ideal since the changes really are completely unrelated.

coderJeroen:
Hi @Badger,

I didn’t know that, since consistency is more important here you can revert my change there. This pull request was more as a practice for myself, since I’d like to contribute to the curl project, but I didn’t know how to do it yet.

coderJeroen:
Hi,
I reverted the changes I previously made that weren't correct, and I saw some things in other code files that I thought could be done a little better, so could you please review it?
By the way, what do you think is the best way to contribute, clone the entire curl base library and have one pull request where you continuously push changes to code files, or create a pull request for every code file you changed? I'm new to GitHub, so I don't really know what's the best way to go about this yet, many thanks in advance!

dfandrich:
You may want to read
https://github.com/curl/curl/blob/master/docs/CONTRIBUTE.md before you start.
If you're looking on something to help with, you could start by looking at the
issues at https://github.com/curl/curl/issues or
https://github.com/curl/curl/blob/master/docs/KNOWN_BUGS  Most of those are
going to be pretty technical and not suited to a curl newbie, but there should
be some that could be tackled without a lot of prior knowledge.

coderJeroen:
Thank you, I'll have a look at the documentation first.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

lib/hash.c
```diff
@@ -204,9 +202,7 @@ Curl_hash_apply(Curl_hash *h, void *user,
 void
 Curl_hash_destroy(struct Curl_hash *h)
 {
-  int i;
-
-  for(i = 0; i < h->slots; ++i) {
+  for(int i = 0; i < h->slots; ++i) {
```
bagder, line 205: This is not C89 compliant code.

lib/base64.c
```diff
@@ -93,7 +93,7 @@ static size_t decodeQuantum(unsigned char *dest, const char *src)
 /*
  * Curl_base64_decode()
  *
- * Given a base64 NUL-terminated string at src, decode it and return a
+ * Given a base64 NULL-terminated string at src, decode it and return a
```
bagder, line 96: I don't think is correct. We can call them "null-terminated" and get away with it, but NULL is a pointer. Strings are not NULL terminated. NUL is the character, the name of the first character in the ASCII character set.

bagder (changes requested):
Keep separate changes in separate PRs. For example, fixing up comments is separate from changing code logic (in another file). Having them separate allows us to merge one of them while still discussing, rejecting or refining the other(s).

bagder:
Also, make sure the title and the description of the PR reflect what the PR wants done. This title currently says "some changes to comments" which is no longer entirely true.

coderJeroen:
Thank you for your comments, I will close this pull request now as it's name isn't correct anymore as you say. Also I've read the documentation and I decided to tackle one of the issues on the known issues list.
________________________________

bagder:
Okay, then I close this and look forward to your next one!

PR #8094, opened 2021-12-03
Wording of -X in docs to refer primarily to 'method' rather than 'command'

themainframe:
Having a poor memory, combined with the not-particularly-descriptive mnemonic (`X`) has led me to fruitlessly scan the manpage/`-h` output numerous times for `method` before falling back to Google.

**Command** feels like awkward parlance in a world where the 95% use-case is HTTP. I could be wrong, but I feel _most people_ looking for this option are looking for how to set the HTTP **method**.

### DIFF
--- a/docs/cmdline-opts/request.d
+++ b/docs/cmdline-opts/request.d
@@ -1,7 +1,7 @@
 Long: request
 Short: X
-Arg: <command>
-Help: Specify request command to use
+Arg: <method>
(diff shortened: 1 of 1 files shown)


bagder (approved):
I think this is fine, even if I'm not sure the world needs more users using -X wrongly...

bagder:
Thanks!

PR #8101, opened 2021-12-06
openldap: process search query response messages one by one

monnerat:
Upon receiving large result sets, this reduces memory consumption and allows starting to output results while the transfer is still in progress.


bagder (approved):

bagder:
Thanks!

monnerat:
Thanks for pulling.

PR #3803, opened 2019-04-23
CURLOPT_FOLLOWLOCATION: add a CURLFOLLOW_NO_CUSTOMREQUEST bit

bagder:
With this change, the argument passed to the CURLOPT_FOLLOWLOCATION
option is treated as a bitmask instead of just a long. If the new
CURLFOLLOW_IGNORE_CUSTOM bit is set in the bitmask, it means that
libcurl will NOT allow a custom method override the HTTP request method
after a redirect is followed. As is otherwise the default behavior (that
surprises many users).

This change is forward compatible because CURLOPT_FOLLOWLOCATION has
been documented to accept the exact value of '1' to enable redirect
following and therefore the other bits were left unused and
undefined. We now add value to another bit. Starting in 7.66.0, the
value 1 and the first bit still enables plain redirect following but the
second bit adds more meaning.

This change is backward compatible in the following way: setting the
CURLFOLLOW_IGNORE_CUSTOM bit in a program that still uses an older
libcurl installation at run-tim will have no effect. This is because
older libcurl code checked if the value was non-zero and then enabled
redirect following. Of course older libcurl will always let the set
CURLOPT_CUSTOMREQUEST string override the method, disregarding what the
HTTP response code suggests.

Test 1563 added to verify the functionality.

PR #8130, opened 2021-12-10
CURLOPT_FOLLOWLOCATION: add a CURLFOLLOW_NO_CUSTOMMETHOD bit

bagder:
(This is a re-run. This change was already proposed for inclusion in 7.66.0 in #3803 but due to lack of interest I retracted it back then. Since this change would allow us to make a `-X` from curl that is "less bad", which also might be interesting when considering [a coming QUERY method](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html), I think/hope it might get some more interest this time around!?)

With this change, the argument passed to the CURLOPT_FOLLOWLOCATION
option is treated as a bitmask instead of just a long. If the new
CURLFOLLOW_NO_CUSTOMREQUEST bit is set in the bitmask, it means that
libcurl will NOT allow a custom method override the HTTP request method
after a redirect is followed. As is otherwise the default behavior (that
surprises many users).

This change is forward compatible because CURLOPT_FOLLOWLOCATION has
been documented to accept the exact value of '1' to enable redirect
following and therefore the other bits were left unused and
undefined. We now add value to another bit. Starting in 7.66.0, the
value 1 and the first bit still enables plain redirect following but the
second bit adds more meaning.

This change is backward compatible in the following way: setting the
CURLFOLLOW_NO_CUSTOMREQUEST bit in a program that still uses an older
libcurl installation at run-tim will have no effect. This is because
older libcurl code checked if the value was non-zero and then enabled
redirect following. Of course older libcurl will always let the set
CURLOPT_CUSTOMREQUEST string override the method, disregarding what the
HTTP response code suggests.

Test 1589 added to verify the functionality.


bagder:
`CURLFOLLOW_NO_CUSTOMMETHOD` is probably a more suitable name. (from @icing)

jay:
> which also might be interesting when considering [a coming QUERY method](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html)

I read that and it looks like cacheable database search results. So then the resource (database) changes and it's invalidated? I wonder how this will work in practice.

docs/libcurl/opts/CURLOPT_FOLLOWLOCATION.3
```diff
@@ -30,9 +30,15 @@ CURLOPT_FOLLOWLOCATION \- follow HTTP 3xx redirects
...
-Location: header can specify a relative or an absolute URL to follow.
+The long parameter \fIenable\fP set to 1 tells the library to follow any
+Location: header that the server sends as part of an HTTP header in a 3xx
+response. The Location: header can specify a relative or an absolute URL to
+follow.
+
+\fIenable\fP is a bitmask. If you set the \fICURLFOLLOW_NO_CUSTOMMETHOD\fP
+bit, it will tell libcurl that the method set with
+\fICURLOPT_CUSTOMREQUEST(3)\fP will not be used after a redirect if the HTTP
+response says so.
```
jay, line 41: Rather than have the user set `curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, CURLFOLLOW_NO_CUSTOMMETHOD | CURLFOLLOW_ENABLE);` , you could instead define CURLFOLLOW_NO_CUSTOMMETHOD as 3 and get rid of CURLFOLLOW_ENABLE. If you feel the bitfield really is the way to go then you should add CURLFOLLOW_DISABLE as well, and document them both.

PR #8133, opened 2021-12-10
check the return value of EVP_MD_CTX_create(EVP_MD_CTX_new)

x2018:
check the return value of EVP_MD_CTX_create(i.e. EVP_MD_CTX_new) in lib/sha256.c:87.
If some memory allocation errors occur, EVP_ MD_ CTX_ New() will return null, then EVP_DigestInit_ex(), EVP_DigestUpdate() and EVP_DigestFinal_ex() will make wrong memory access. So it is better to check the return value of it.

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -288,6 +288,10 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
     post_data_len = (size_t)data->set.postfieldsize;
   Curl_sha256it(sha_hash,
                 (const unsigned char *) post_data, post_data_len);
+  if(!(char *) sha_hash) {
+    goto fail;
+  }
+
   sha256_to_hex(sha_hex, sha_hash, sizeof(sha_hex));
 
   Curl_http_method(data, conn, &method, &httpreq);
@@ -322,6 +326,10 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
 
   Curl_sha256it(sha_hash, (unsigned char *) canonical_request,
                 strlen(canonical_request));
+  if(!(char *) sha_hash) {
+    goto fail;
+  }
+
   sha256_to_hex(sha_hex, sha_hash, sizeof(sha_hex));
 
   /*
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -288,7 +288,7 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
     post_data_len = (size_t)data->set.postfieldsize;
   Curl_sha256it(sha_hash,
                 (const unsigned char *) post_data, post_data_len);
-  if(!(char *) sha_hash) {
+  if(!sha_hash[0]) {
     goto fail;
   }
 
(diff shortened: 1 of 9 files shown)


bagder:
Shouldn't the init function rather return an error to then have `Curl_sha256it()` also return an error for this case so that curl can bail out all the way up?

x2018:
Yes, it is best for this case. However, I hesitated to do that yesterday as there are other implementations under different `macro conditions` and I am not familiar with them. So if we suppose to do that, some definitions and statements need to be also updated.

x2018:
I filtered the locations which call `Curl_sha256it()`, and check the validity of its first parameter after calling it. If you still think it is necessary to change the `return type` of `Curl_sha256it()`, I may consider taking more time to optimize `Curl_sha256it()` and its subsequent functions under different implementations.

x2018:
Sorry. I made mistakes that commit to  this branch. later I will rollback it.

lib/http_aws_sigv4.c
```diff
@@ -288,6 +288,10 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
     post_data_len = (size_t)data->set.postfieldsize;
   Curl_sha256it(sha_hash,
                 (const unsigned char *) post_data, post_data_len);
+  if(!(char *) sha_hash) {
+    goto fail;
+  }
```
bagder, line 293: please make `Curl_sha256it` return error/success instead and check that instead of relying on the output data. Yes, that's a bigger change but is much cleaner and nicer code.

MarcelRaad:
Thanks! That was an oversight in https://github.com/curl/curl/commit/94696e1b51bc05caed7ece734162edfcd3bcd619.

I agree with @bagder , the error should be handled by the caller. That would mean changing the signature of `HMAC_hinit_func`. Which would be good anyway as other implementations and the corresponding MD5 functions can also fail. 

> However, I hesitated to do that yesterday as there are other implementations under different macro conditions and I am not familiar with them. So if we suppose to do that, some definitions and statements need to be also updated.

The other implementations could just always return `CURLE_OK` for now. The CI system should tell us if anything was missed.

x2018:
I changed the return value of `Curl_sha256it()` and `Curl_md5it()`. To be cautious, I still just add a check for the return value of `EVP_MD_CTX_create()`, but the other functions in `*_init`, `*_update` and `*_final` may also fail in some way. As I do not familiar with them and I'm not sure what status they should return when they have an error, so I just retain the original shape(or directly return `CURLE_OK`).

lib/sha256.c
```diff
@@ -491,15 +502,22 @@ static int my_sha256_final(unsigned char *out,
...
 
-  my_sha256_init(&ctx);
+  ret = my_sha256_init(&ctx);
+  if(ret != CURLE_OK)
+    return ret;
+
   my_sha256_update(&ctx, input, curlx_uztoui(length));
   my_sha256_final(output, &ctx);
+  return ret;
 }
```
MarcelRaad, line 523: I can't comment on line 527, but the `HMAC_hinit_func` signature (and the MD5 init function) would have to be updated too to return `CURLcode`. Unfortunately the `CURLX_FUNCTION_CAST` meant for the parameter hides that. Looks good to me otherwise!

x2018: Sorry, I forgot to update it though you reminded me. I will do it after work...

MarcelRaad (approved):
Thanks, looks good to me now! Just waiting if @bagder wants to have another look too.

bagder:
Thanks!

PR #8138, opened 2021-12-13
tests/data/test302[12]: fix MSYS2 path conversion of hostpubsha256

### LINKED ISSUE #8084: Disable msys2 path interpretation in Cirrus CI
### I did this

[This test failure](https://cirrus-ci.com/task/5748643199188992?logs=test#L2926) from Cirrus CI shows test3021 failing:

` curl: (60) Denied establishing ssh session: mismatch sha256 fingerprint. Remote �z�i/XEa"��P�R��h�]��wl�q��) is not equal to C:/msys64/Ad6smkvWEVhIs/nUM1S2fOjaL9djxb6d2zkcbQUgik`

The error message is supposed to show two sha256 fingerprints in base64 format. The gibberish came from a printing mistake and I landed 3467e89b to fix it. The reason the match failed was not because of that but because one of the keys is `C:/msys64/Ad6smkvWEVhIs/nUM1S2fOjaL9djxb6d2zkcbQUgik`, note the erroneous `C:/msys64/`. There should be just a base64 string there.

Tracing it back I don't see anything unusual.
https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/lib/vssh/libssh2.c#L623
https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/tests/data/test3021#L27
https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/tests/runtests.pl#L2291-L2294
https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/tests/sshserver.pl#L402-L404

### I expected the following

My guess is this has something to do with msys2 path interpretation. Maybe the base64 string had a leading / and so msys2 converted that to `C:/msys64/`. There is an msys2 environment variable MSYS2_ARG_CONV_EXCL that can be set to stop msys2 from tampering with arguments. That is already done in appveyor configs, for example:

https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/appveyor.yml#L145

I think that should be done for Cirrus CI as well.

mback2k:
Ref: https://www.msys2.org/wiki/Porting/#filesystem-namespaces

Fixes #8084

### DIFF
--- a/tests/data/test3021
+++ b/tests/data/test3021
@@ -30,6 +30,11 @@ SFTP correct sha256 host key
  <command>
 --hostpubsha256 %SSHSRVSHA256 --key curl_client_key --pubkey curl_client_key.pub -u %USER: sftp://%HOSTIP:%SSHPORT%SSH_PWD/log/file%TESTNUMBER.txt
 </command>
+<setenv>
+# Needed for MSYS2 to not treat the argument as a POSIX path list
+# that has to be converted to Windows paths
+MSYS2_ARG_CONV_EXCL=/
+</setenv>
 <file name="log/file%TESTNUMBER.txt">
 test
 </file>
--- a/tests/data/test3022
+++ b/tests/data/test3022
@@ -30,6 +30,11 @@ SCP correct sha256 host key
  <command>
 --hostpubsha256 %SSHSRVSHA256 --key curl_client_key --pubkey curl_client_key.pub -u %USER: scp://%HOSTIP:%SSHPORT%SSH_PWD/log/file%TESTNUMBER.txt
 </command>
+<setenv>
+# Needed for MSYS2 to not treat the argument as a POSIX path list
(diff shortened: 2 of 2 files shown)


MarcelRaad (approved):
👍

mback2k:
As far as I can tell `/*` was used to escape comments starting with exactly `/*` like shown by @MarcelRaad in https://github.com/curl/curl/issues/8084#issuecomment-985864284.

jay:
> As far as I can tell `/*` was used to escape comments starting with exactly `/*` like shown by @MarcelRaad in [#8084 (comment)](https://github.com/curl/curl/issues/8084#issuecomment-985864284).

Ok that makes sense. I had assumed it was a wildcard.

PR #8140, opened 2021-12-13
openldap: some tiny adjustments

monnerat:
- Early check proper LDAP URL syntax. Reject URLs with a userinfo part.
- Use dynamic memory for `ldap_init_fd`() URL rather than a
  stack-allocated buffer.
- Never chase referrals: supporting it would require additional parallel
  connections and alternate authentication credentials.
- Do not wait 1 microsecond while polling/reading query response data.
- Store last received server code for retrieval with `CURLINFO_RESPONSE_CODE`.

Maybe you will postpone pull of this as `CURLINFO_RESPONSE_CODE` for LDAP might be considered as a feature.

### DIFF
--- a/docs/libcurl/opts/CURLINFO_RESPONSE_CODE.3
+++ b/docs/libcurl/opts/CURLINFO_RESPONSE_CODE.3
@@ -30,15 +30,14 @@ CURLINFO_RESPONSE_CODE \- get the last response code
 CURLcode curl_easy_getinfo(CURL *handle, CURLINFO_RESPONSE_CODE, long *codep);
 .fi
 .SH DESCRIPTION
-Pass a pointer to a long to receive the last received HTTP, FTP or SMTP
-response code. This option was previously known as CURLINFO_HTTP_CODE in
-libcurl 7.10.7 and earlier. The stored value will be zero if no server
-response code has been received. Note that a proxy's CONNECT response should
+Pass a pointer to a long to receive the last received HTTP, FTP, SMTP or
+LDAP (openldap only) response code. This option was previously known as
+CURLINFO_HTTP_CODE in libcurl 7.10.7 and earlier.
+The stored value will be zero if no server response code has been received.
+Note that a proxy's CONNECT response should
 be read with \fICURLINFO_HTTP_CONNECTCODE(3)\fP and not this.
-
-Support for SMTP responses added in 7.25.0.
 .SH PROTOCOLS
-HTTP, FTP and SMTP
+HTTP, FTP, SMTP and LDAP
 .SH EXAMPLE
 .nf
 CURL *curl = curl_easy_init();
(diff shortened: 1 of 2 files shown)


docs/libcurl/opts/CURLINFO_RESPONSE_CODE.3
```diff
@@ -30,15 +30,16 @@ CURLINFO_RESPONSE_CODE \- get the last response code
...
-response code has been received. Note that a proxy's CONNECT response should
+Pass a pointer to a long to receive the last received HTTP, FTP, SMTP or
+LDAP (openldap only) response code. This option was previously known as
+CURLINFO_HTTP_CODE in libcurl 7.10.7 and earlier.
+The stored value will be zero if no server response code has been received.
+Note that a proxy's CONNECT response should
 be read with \fICURLINFO_HTTP_CONNECTCODE(3)\fP and not this.
 
-Support for SMTP responses added in 7.25.0.
+Support for SMTP responses added in 7.25.0, for OpenLDAP in 7.81.0.
```
bagder, line 40: How about moving that line to the `AVAILABILITY` header?

lib/openldap.c
```diff
@@ -231,40 +217,65 @@ static CURLcode oldap_map_error(int rc, CURLcode result)
...
+    rc = ldap_url_parse(data->state.url, ludp);
   if(rc != LDAP_URL_SUCCESS) {
     const char *msg = "url parsing problem";
-    status = CURLE_URL_MALFORMAT;
-    if(rc > LDAP_URL_SUCCESS && rc <= LDAP_URL_ERR_BADEXTS) {
-      if(rc == LDAP_URL_ERR_MEM)
-        status = CURLE_OUT_OF_MEMORY;
+
+    result = rc == LDAP_URL_ERR_MEM? CURLE_OUT_OF_MEMORY: CURLE_URL_MALFORMAT;
+    if(rc > LDAP_URL_SUCCESS && rc <= LDAP_URL_ERR_BADEXTS)
```
bagder, line 246: I know this is better than before, but shouldn't this condition rather use the size of the string array instead so that we can be really **sure** that the following lookup never reads out of boundary by its own merit instead? That way we can verify that the code is fine without having to lookup what the contents of  `LDAP_URL_ERR_BADEXTS` is and has been been.

docs/libcurl/opts/CURLINFO_RESPONSE_CODE.3
monnerat: Right! Let's take this oportunity to do this.

lib/openldap.c
monnerat: Done. This is far better I think. Thanks for the suggestion.

bagder:
Thanks!

monnerat:
Thanks for pulling!

PR #8157, opened 2021-12-16
nss: Remove unused flag in context initialization

danielgustafsson:
When calling `NSS_InitContext()`, NSS explicitly sets `NSS_INIT_NOROOTINIT` without checking for it's presence among the passed flags (ref https://github.com/nss-dev/nss/blob/master/lib/nss/nssinit.c#L915). While it's not incorrect per se to pass it in, remove and document it for clarity to avoid implying that removing it and recompiling curl would alter its behavior.

### DIFF
--- a/lib/vtls/nss.c
+++ b/lib/vtls/nss.c
@@ -1368,10 +1368,14 @@ static CURLcode nss_init_core(struct Curl_easy *data, const char *cert_dir)
     infof(data, "Unable to initialize NSS database: %d (%s)", err, err_name);
   }
 
+  /*
+   * NSS sets NSS_INIT_NOROOTINIT when creating the context regardless of
+   * flags passed in.
+   */
   infof(data, "Initializing NSS with certpath: none");
   nss_context = NSS_InitContext("", "", "", "", &initparams, NSS_INIT_READONLY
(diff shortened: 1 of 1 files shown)


bagder (approved):

jay:
> When calling `NSS_InitContext()`, NSS explicitly sets `NSS_INIT_NOROOTINIT`

Ok but is it documented anywhere? Maybe they change it. What's the harm in keeping it? Also if removing it I find the comment unnecessary, you can just remove it.

danielgustafsson:
> Ok but is it documented anywhere?

The NSS context machinery is pretty much not documented at all, as is equally true for most parts of NSS updated in the past decade.

> Maybe they change it. What's the harm in keeping it? Also if removing it I find the comment unnecessary, you can just remove it.

There is no harm for libcurl, but it can be misleading. Because NSS is so poorly documented, implementations like ours in libcurl are often used as actual API documentation by people implementing NSS support (myself being an example). I personally read this, changed the flag, recompiled and started debugging why it didn't change as expected and only when reading the NSS source code did I realize it was a red herring.

jay:
Ok. It's strange such an NSS function isn't properly documented.

danielgustafsson:
> Ok. It's strange such an NSS function isn't properly documented.

Sadly that's true for a lot of recent changes to NSS.

danielgustafsson:
Indeed we don't.

PR #8161, opened 2021-12-17
nss:set_cipher don't clobber the cipher list

### LINKED ISSUE #8160: Connections not reused when cipher list specified
### I did this
Since 7.52.0, with libcurl built with nss, if a cipher list is specified with multiple entries on subsequent requests such as:
`
curl_easy_setopt(curl, CURLOPT_SSL_CIPHER_LIST, "ecdhe_ecdsa_aes_128_gcm_sha_256,ecdhe_rsa_aes_128_gcm_sha_256,dhe_rsa_aes_128_gcm_sha_256,rsa_aes_128_gcm_sha_25");
`

we see a new https connection created on each request as opposed to reusing the existing connection. This broke with commit
cb4e2be7c proxy: Support HTTPS proxy and SOCKS+HTTP(s)

The issue is that the method set_ciphers (in lib/vtls/nss.c) clobbers the cipher_list entry kept on each connection by replacing any commas with null characters. After this method is called the cipher_list on a connection will now only contain the first entry (in the above example: ecdhe_ecdsa_aes_128_gcm_sha_256) and it is not necessarily the cipher negotiated for the connection.

The above is an issue in ConnectionExists and Curl_ssl_config_matches as it expects the cipher_list of a new request to exactly match the cipher_list found on an existing connection in order for that connection to be reused.

A relatively simple fix is to call strdup on the cipher_list before passing to set_ciphers (I can push this for review if desired). But looking at the original commit that broke this it appears that we may be inadvertently sharing cipher_list pointers -- perhaps further investigation is needed.

### I expected the following
I expected a subsequent request with the same cipher_list (and other properties) to reuse the existing connection.

### curl/libcurl version

libcurl 7.80.0

### operating system

Linux xx.xx.xx.xx 4.19.84-33.70.amzn2.x86_64 #1 SMP Sun Nov 17 03:35:39 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

bagder:
The string is set by the user and needs to remain intact for proper
connection reuse etc.

Reported-by: Eric Musser
Fixes #8160

### DIFF
--- a/lib/vtls/nss.c
+++ b/lib/vtls/nss.c
@@ -343,7 +343,11 @@ static SECStatus set_ciphers(struct Curl_easy *data, PRFileDesc *model,
     else
       len = strlen(cipher);
 
-    if(len && (len < MAX_CIPHER_LENGTH)) {
+    if(len > MAX_CIPHER_LENGTH) {
+      failf(data, "Bad cipher list");
+      return SECFailure;
+    }
+    else if(len) {
       memcpy(name, cipher, len);
       name[len] = 0;
 
--- a/lib/vtls/nss.c
+++ b/lib/vtls/nss.c
@@ -333,6 +333,7 @@ static SECStatus set_ciphers(struct Curl_easy *data, PRFileDesc *model,
     const char *end;
     char name[MAX_CIPHER_LENGTH + 1];
     size_t len;
+    bool found = FALSE;
     while((*cipher) && (ISSPACE(*cipher)))
       ++cipher;
 
@@ -343,7 +344,6 @@ static SECStatus set_ciphers(struct Curl_easy *data, PRFileDesc *model,
       len = strlen(cipher);
 
     if(len && (len < MAX_CIPHER_LENGTH)) {
-      bool found = FALSE;
       memcpy(name, cipher, len);
       name[len] = 0;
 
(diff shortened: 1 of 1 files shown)


lib/vtls/nss.c
```diff
@@ -326,51 +327,48 @@ static SECStatus set_ciphers(struct Curl_easy *data, PRFileDesc * model,
...
-  /* Finally actually enable the selected ciphers */
-  for(i = 0; i<NUM_OF_CIPHERS; i++) {
-    if(!cipher_state[i])
-      continue;
-
-    if(SSL_CipherPrefSet(model, cipherlist[i].num, PR_TRUE) != SECSuccess) {
-      failf(data, "cipher-suite not supported by NSS: %s", cipherlist[i].name);
-      return SECFailure;
+      if(!found) {
+        failf(data, "Unknown cipher in list: %s", name);
```
sfc-gh-emusser, line 364: This failure won't be triggered for cipher names that exceed the max cipher length above.

sfc-gh-emusser (commented):
Thanks for the quick turnaround! One minor issue with the change.

lib/vtls/nss.c
bagder: oh yes...

bagder:
@danielgustafsson I figured you've dabbled with NSS a bit so you might have an opinion on this?

lib/vtls/nss.c
```diff
@@ -359,11 +359,11 @@ static SECStatus set_ciphers(struct Curl_easy *data, PRFileDesc *model,
...
         }
       }
+    }
 
-      if(!found) {
-        failf(data, "Unknown cipher in list: %s", name);
-        return SECFailure;
-      }
+    if(!found && len) {
+      failf(data, "Unknown cipher: %s", name);
```
sfc-gh-emusser, line 365: name won't be populated in the too long case.

jay (approved):

PR #8163, opened 2021-12-18
tftp: set timeout based on the connect-timeout parameter

bagder:
From https://curl.se/mail/archive-2021-12/0015.html

### DIFF
--- a/lib/tftp.c
+++ b/lib/tftp.c
@@ -136,6 +136,7 @@ struct tftp_state_data {
   int             rbytes;
   int             sbytes;
   int             blksize;
+  int             option_timeout;
   int             requested_blksize;
   unsigned short  block;
   struct tftp_packet rpacket;
@@ -221,6 +222,9 @@ static CURLcode tftp_set_timeouts(struct tftp_state_data *state)
 
   /* Set per-block timeout to total */
   timeout = maxtime;
+
+  if(start)
+    state->option_timeout = (timeout > 255) ? 255 : (int)timeout;
 
   /* Average reposting an ACK after 5 seconds */
   state->retry_max = (int)timeout/5;
@@ -507,7 +511,7 @@ static CURLcode tftp_send_first(struct tftp_state_data *state,
                                  (char *)state->spacket.data + sbytes, buf);
 
       /* add timeout option */
-      msnprintf(buf, sizeof(buf), "%d", state->retry_time);
+      msnprintf(buf, sizeof(buf), "%d", state->option_timeout);
       if(result == CURLE_OK)
         result = tftp_option_add(state, &sbytes,
                                  (char *)state->spacket.data + sbytes,


bagder:
This change makes the tftp timeout get set to 255 instead of 6 by default, like in test 271. It doesn't seem right...

divinity76:
what's the point of limiting the option to max 255 seconds?  also maybe a warning like

> `if(t>255){print("warning, timeout capped to 255") ...`

would be in order, if i tell curl to timeout after 300 seconds, and it keeps timing out at 255 seconds, it might be nice to find out why by simply turning on -v

bagder:
libcurl has no warning system. It could output a verbose message, but I doubt that's very useful for this.

tlhackque:
> what's the point of limiting the option to max 255 seconds? also maybe a warning like
> 
> > `if(t>255){print("warning, timeout capped to 255") ...`
> 
> would be in order, if i tell curl to timeout after 300 seconds, and it keeps timing out at 255 seconds, it might be nice to find out why by simply turning on -v

[RFC1784](https://datatracker.ietf.org/doc/html/rfc1784) "TFTP Timeout Interval and Transfer Size Options"

timeout
         The Timeout Interval option, "timeout" (case insensitive).
         This is a NULL-terminated field.

\#secs
         The number of seconds to wait before retransmitting, specified
         in ASCII.  **Valid values range between "1" and "255" octets,**
         inclusive.  This is a NULL-terminated field.

255 is not the value - it's the maximum string length for the value, which is in ASCII.  E.g.  "100" occupies 3 octets.

However, 255 secs is not an unreasonable limit for a timeout, even over a slow dial-up link.  255 secs is 4 1/4 minutes....

6 secs is a much more sensible **default** value for modern networks.

Note that this is a per-packet timeout used to detect lost packets and broken connections.  So anything approaching this value becomes large quite fast.  E.g. a 100 packet (50KB) file would be 425 minutes, or 7 hours if every packet was retried once.  (Which, of course, is unlikely.)    But packets do get dropped - remember that TFTP is a UDP protocol, meaning that delivery is not guaranteed.  When sending router firmware (about that size), I often see a few (<10) packets dropped on a Gbe network.  Dumps can be MBs.

It's not clear why you would couple this to the connect timeout.  Connect happens once (and might involve a few seconds of activating a process); the TFTP timeout happens every packet.  It's more akin to TCP/HTTP keep-alive than connect.

As a side note: TFTP provides no security and is easily disrupted.  It is handy in a protected environment - although even there, I strongly recommend verifying files transfered  with a robust signature.

danielbelz:
> [RFC1784](https://datatracker.ietf.org/doc/html/rfc1784) "TFTP Timeout Interval and Transfer Size Options"

RFC1784 is obsolete. The actual one is [RFC2349](https://datatracker.ietf.org/doc/html/rfc2349)

>       #secs
>         The number of seconds to wait before retransmitting, specified
>         in ASCII.  Valid values range between "1" and "255" seconds,
>         inclusive.  This is a NULL-terminated field.

As mentioned, there isn't a default value defined, and I agree that 255 might be too much as a default value, especially when connect timeout isn't specified. I agree that a default of 6 seconds would be a much better default value. 

> It's not clear why you would couple this to the connect timeout.

As it is, it's already coupled to the connect timeout, just in a different way.
One alternative would be to change the upper and lower bounds for the retry_max. But that would impact the calculation of all internal tftp timeouts.

bagder:
I think that we change the logic somewhat. For example:

The default tftp timeout (called `retry_time`) could be 5 seconds,
I think `retry_max` could be capped lower than 50, maybe at 10. This would actually make a tftp connect fail already at name resolve + 50 seconds by default, but I would be okay with that.

If we then use that as the default timeout for tftp, we can set the `retry_time` as a function of that (`connection-timeout / 10`)  rather than changing  `retry_max`. A connect timeout of 100 would make retry_time 10. Setting connection timeout to 300 would give retry_time 30.

If the connection timeout is set *lower* than 50 seconds, then we could start by reducing the retry_max down to maybe a minimum of 3 before we lower the retry_time. That would fit the range 15 to 50 seconds.

If the connection timeout is shorter than 15 seconds, then we leave retry_max at three and instead shorten retry_time to `timeout / max` ... That might work down to 3 seconds, or at least 6. Setting shorter timeouts like that for TFTP seems very fragile...

bagder:
> Rather than a complex function, you might consider picking 2 or perhaps 3 sets of values for the timing parameters; e.g. "lightly loaded (fast) network"; "typical"; and "slow".

I want to avoid adding another option.

tlhackque:
> > Rather than a complex function, you might consider picking 2 or perhaps 3 sets of values for the timing parameters; e.g. "lightly loaded (fast) network"; "typical"; and "slow".
> 
> I want to avoid adding another option.

Agree that adding a dedicated option isn't desirable; that's why I suggested something generic that other protocols can also use (and add keywords to).

A heuristic function is likely to be wrong in some cases, and certainly the intent will be less clear than "values chosen for ..."  The link between connect timeout and retries is tenuous at best.

If you won't add even a generic option, then you can set  some thresholds in the connect timeout that bundle the retry paramenters - e.g. connect <= 5 seconds == fast, <= 30 seconds == typical, >30 == slow.  (The actual threshold values may need some thought.)

However, documenting the effect of overloading the connect timeout seems even less desirable than adding an option.   You'll have to explain how you've linked connect to the TFTP paramenters, and since it's indirect/for engineering convenience, it's likely to be long, confusing, or both.

However, I'm not doing the work, so my vote doesn't count for much.

danielbelz:
In the mailing list, it was mentioned that the --retry parameter might not being used for TFTP: 
> --retry is for retrying a *failed transfer* when that transfer fails with a
> "transient error", which I'm not sure there is any when it comes to TFTP?

Would it be possible to repurpose this parameter for TFTP connections and calculate the connect timeout as connect-timeout*retry, and make connect-timeout value between 1 and 255? Or would it create too much confusion regarding its actual purpose?

danielbelz:
> I want to avoid adding another option.

And would be acceptable to replace an option? For example, remove --tftp-blksize and replace it with --tftp-option <blksize=x,timeout=y>

bagder:
> Would it be possible to repurpose this parameter for TFTP connections and calculate the connect timeout as connect-timeout*retry

No. That's totally out of the question. The `--retry` option is not even sent to libcurl so you couldn't use it for that anyway.

> And would be acceptable to replace an option? For example, remove --tftp-blksize and replace it with --tftp-option <blksize=x,timeout=y>

No. We cannot replace options since the old ones need to keep working. We can only add.

bagder:
I'll ditch this effort, it needs another direction.

PR #8166, opened 2021-12-18
Change for PEEK in IMAP FETCH command

MonkeybreadSoftware:
We like to fetch email without marking them as read.


MonkeybreadSoftware:
I made a new implementation to better handle UID FETCH custom commands:
https://github.com/curl/curl/pull/8659

PR #8167, opened 2021-12-18
Added peek for IMAP email FETCH command

MonkeybreadSoftware:
When curl fetches an email, it is marked as read.
With the new peek option, we can keep the read status unchanged when loading the email content.

Just add a ";PEEK=1" to the URL to load email.

Ref: https://curl.se/mail/lib-2021-12/0045.html

### DIFF
--- a/lib/imap.c
+++ b/lib/imap.c
@@ -721,22 +721,30 @@ static CURLcode imap_perform_fetch(struct Curl_easy *data)
 
     /* Send the FETCH command */
     if(imap->partial)
-      result = imap_sendf(data, "UID FETCH %s BODY[%s]<%s>",
-                          imap->uid, imap->section ? imap->section : "",
+      result = imap_sendf(data, "UID FETCH %s BODY%s[%s]<%s>",
+                          imap->uid,
+                          imap->peek ? ".PEEK" : "",
+                          imap->section ? imap->section : "",
                           imap->partial);
     else
-      result = imap_sendf(data, "UID FETCH %s BODY[%s]",
-                          imap->uid, imap->section ? imap->section : "");
+      result = imap_sendf(data, "UID FETCH %s BODY%s[%s]",
+                          imap->uid,
+                          imap->peek ? ".PEEK" : "",
+                          imap->section ? imap->section : "");
   }
   else if(imap->mindex) {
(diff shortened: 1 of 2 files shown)


jay:
From what I'm reading in IMAP URL RFC 5092 it looks like PEEK is supposed to be implied without a specific section? I think changing the default at this point could break some expectations though.

[6.  A Specific Message or Message Part](https://datatracker.ietf.org/doc/html/rfc5092#section-6)

~~~
   An IMAP URL referring to a specific message or message part has the
   following form:

      imap://<iserver>/<enc-mailbox>[<uidvalidity>]<iuid>
      [<isection>][<ipartial>][<iurlauth>]
~~~

...

~~~
   The <isection> field is optional.  If not present, the URL refers to
   the entire Internet message as returned by the IMAP command "UID
   FETCH <uid> BODY.PEEK[]".  If present, the URL refers to the object
   returned by a "UID FETCH <uid> BODY.PEEK[<section>]" command.
~~~

bagder:
I struggle with this change since I deeply dislike how we (myself included) have invented a lot of private URL magic for imap. URLs are meant to be agnostic and work cross-applications for doing transfers (actually for identifying remote resources). Inventing our own URL bits goes contrary to that. I also realize that we already did this for IMAP so maybe this boat has already sailed..

Apart from that:
1. This PR lacks documentation for this new functionality
2. There's no test cases that verifies that the feature does what it says

MonkeybreadSoftware:
The general problem is that I'd like to get into the FETCH state, where CURL provides the mail content as output of the transfer. If we use a custom request, the email content goes into the debug log.
If you prefer a different way, feel free to invent one.

otherwise I can look into modification of an Imap test case to test for this and write a few lines of documentation.

bagder:
> If we use a custom request, the email content goes into the debug log.

Isn't this rather the problem we should address then?

MonkeybreadSoftware:
I made a new implementation to better handle UID FETCH custom commands:
https://github.com/curl/curl/pull/8659

bagder:
Does #8659 replace this PR?

MonkeybreadSoftware:
Yes. Since you struggled above, I came up with a new idea to handle this better.

PR #8168, opened 2021-12-18
MINOR: fix format of #error

xquery:
fix spurious spaces in #error preprocessor directive


bagder:
We can just fix those nits later when we go there to fix something else.

PR #8169, opened 2021-12-18
Add build-option to disable environment variables

ida64:


bagder (commented):
1. This isn't documented anywhere
2. This also breaks the public API `curl_getenv()` if used, which seems over-reaching?

ida64:
I apologize, I should have searched around more before making this PR.

It looks like `getenv` is still being used a lot in the code, is this correct? Should it be using with `curl_getenv` instead? Should there be a separate API for fetching CURL only environment variables?

bagder:
> It looks like `getenv` is still being used a lot in the code, is this correct? Should it be using `curl_getenv` instead? Should there be a separate API for fetching CURL only environment variables?

Why? I suppose it depends on what you want to achieve with your change but you haven't explained what the purpose of this is!

bagder:
Needs more work and clarification. Feel free to reopen or create a new one when the details are provided.

PR #8170, opened 2021-12-20
ldap: return CURLE_URL_MALFORMAT for bad URL

bagder:
For consistency, use the same return code for URL malformats,
independently of what scheme that is used. Previously this would return
CURLE_LDAP_INVALID_URL, but starting now that error cannot be returned.


docs/libcurl/libcurl-errors.3
```diff
@@ -184,8 +184,6 @@ This error code has been unified with CURLE_SSL_CACERT since 7.62.0. Its
 previous value was 51.
 .IP "CURLE_BAD_CONTENT_ENCODING (61)"
 Unrecognized transfer encoding.
-.IP "CURLE_LDAP_INVALID_URL (62)"
-Invalid LDAP URL.
```
jay, line 188: I think this is worth leaving in for a while since the web documentation is generated from it. I can imagine for some time there could be searches for CURLE_LDAP_INVALID_URL. Perhaps label it obsolete.

bagder: Fair. I'll amend.

PR #8171, opened 2021-12-20
http: Fix CURLOPT_HTTP200ALIASES

CyberShadow:
The `httpcode < 100` check was also triggered when none of the fields were parsed, thus making the `if(!nc)` block unreachable.

Regression introduced in c67a32fc560913caff7875bb4fe98abb9991798c.


bagder (approved):

bagder:
Seems likely that c67a32fc560913 broke this.

bagder:
Extra bonus points for the excellent added test! :100:

bagder:
Thanks!

Issue #8173, opened 2021-12-21
curl+quiche with HTTP/3 silently ignores certificate problems


bagder:
### I did this

1. I start a test h3 server locally. In my case I run nghttpx with h3 abilities, using a cert + key from curl's test directory.
2. This cert is **not** signed by any CA in my system CA store
3. I run curl against this server to download a resource: `./curl --http3 https://localhost:9443/8GB -o /dev/null`
4. This works fine!

### I expected the following

It should have failed and required `-k` or a suitable `--cacert` line.

### curl/libcurl version

curl from current git. It works similarly using either/both HTTP/3 backends.

### operating system

Tested on Linux but the code is platform independent.

bagder:
@tatsuhiro-t can you point out which ngtcp2 functions I should look closer on to fix this?

@ghedo can you point out which quiche functions I should look closer on to fix this?

tatsuhiro-t:
Usual OpenSSL SSL_CTX or SSL setup is required here for ngtcp2 build.
I think something like https://github.com/curl/curl/blob/2c1dbc1af0a774ae5dce08150131a01e862c13b5/lib/vtls/openssl.c#L3181-L3182 will fix this issue.

bagder:
Thanks @tatsuhiro-t, that seems to fix the verification of the signatures for the OpenSSL branch. But I then also need to verify the CN and subjectAltName fields like we do with `servercert` for TCP+TLS at https://github.com/curl/curl/blob/2c1dbc1af0a774ae5dce08150131a01e862c13b5/lib/vtls/openssl.c#L3863

I figure that needs to be done after the handshake is completed so that we can get the peer cert using the API.

bagder:
Half the check (missing the name check) is done for ngtcp2+quictls in #8178

bagder:
This problem is now fixed in master (8fbd6feddfa587) for the ngtcp2 backend.

PR #8178, opened 2021-12-23
ngtcp2: verify the server cert on connect

bagder:
This makes ngtcp2+quictls acknowledge `CURLOPT_SSL_VERIFYPEER` and `CURLOPT_SSL_VERIFYHOST`

- [X] ngtcp2 + quictls
- [X] ngtcp2 + gnutls
- [ ] quiche

Ref: #8173


bagder:
Since quiche uses boringssl, it can probably be possible to use more or less the same code there, and possibly we can write the generic openssl certificate check code to be usable in all three places, but as a first shot the functionality should get there and only later I'll work on combining code.

PR #8180, opened 2021-12-25
checksrc: detect more kinds of NULL comparisons we avoid

bagder:


lib/checksrc.pl
```diff
@@ -502,12 +502,19 @@ sub scanfile {
...
         # the left of it is a function call
-        if($nostr =~ /^(.*)(if|while)(\(.*[^)]) == NULL/) {
+        if($nostr =~ /^(.*)(if|while)(\(.*[^)]) [!=]= NULL/) {
             checkwarn("EQUALSNULL", $line,
                       length($1) + length($2) + length($3),
                       $file, $l, "we prefer !variable instead of \"== NULL\" comparisons");
         }
 
+        # check for 'NULL ==' in if/while conditions
+        if($nostr =~ /^(.*)(if|while)\(NULL [!=]=/) {
```
jay, line 512: I think you could combine these, and also it doesn't cover cases like (x) && y != NULL. I experimented in regex 101 and [this](https://regex101.com/r/xxEGWg/1) seems to work for all:
~~~perl
/^(.*)(if|while)(\(.*?)([!=]= NULL|NULL [!=]=)/
~~~

bagder: Nice, thanks!

jay (approved):

PR #8185, opened 2021-12-25
set a label of the link to the support document

signalwerk:
All other links in the README.md are set accordingly to the [HTML Living Standard](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) or [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a). The content of an a-element (anchor) should indicate the link's destination. This fixes the only link that is not done like this.


bagder (approved):

bagder:
Thanks!

PR #8191, opened 2021-12-26
mesalink: remove support

### LINKED ISSUE #8188: mesalink: no longer developed
The TLS library mesalink is no longer being developed. It seems the continuation is now called [TabbySSL](https://github.com/ymjing/tabbyssl).

Mesalink has not seen a source code change in over a year.

I propose we drop support for Mesalink because of this. Someone interested could try tabbyssl with the same backend code, or if the OpenSSL API really is good enough, the regular OpenSSL backend code could be used with tabbyssl!

bagder:
Mesalink has stopped development. We can no longer encourage use of it.
It seems to be continued under the name TabbySSL, but no attempts have
(yet) been to make curl support it.

Fixes #8188


jay:
@ymjing

ymjing:
Thanks Daniel. Sorry the MesaLink project does not last long. 
I'd be happy to see the MesaLink TLS backend being deprecated in favor of the new [rustls backend](https://www.abetterinternet.org/post/memory-safe-curl/) sponsored by ISRG.

Issue #8193, opened 2021-12-27
Socket use after close when connecting to remote HTTPS via HTTPS proxy


intxgo:
### I did this

1. Connect to non-existent HTTPS server via HTTPS proxy tunnel. 
2. Observe the process crash

We use  [ProcessStrictHandleCheckPolicy](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-process_mitigation_strict_handle_check_policy) which reveals socket use after close. 

### I expected the following

Curl returns "Received HTTP code 502 from proxy after CONNECT" and gracefully close the HTTPS tunnel.

### curl/libcurl version

7.79.1

[curl -V output]
curl 7.81.0-DEV (x86_64-pc-win32) libcurl/7.81.0-DEV OpenSSL/1.1.1i zlib/1.2.11 WinIDN
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: AsynchDNS HSTS HTTPS-proxy IDN IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI UnixSockets alt-svc libz

### operating system

Windows 10 20H2

bagder:
You didn't show the curl -V output though, so we're clueless about your TLS config.

intxgo:
Thanks, I see, but I'm only linking with curl lib in my project. Perhaps I could provide the CMake definitions used to build the lib if my pool request is not good or clear.

bagder:
I want to A) reproduce the problem and then B) verify that the fix is fine.

bagder:
It does not reproduce for me on Linux, with or without valgrind:

`./src/curl -x https://localhost:8843 https://localhost:7999/ --proxy-insecure`

Where is the socket use after close happening? Can you get a stack trace from your crash?

intxgo:
The stack trace:
```
EXCEPTION_RECORD:  (.exr -1)
ExceptionAddress: 00007ffbc7670d2a (ntdll!KiRaiseUserExceptionDispatcher+0x000000000000003a)
   ExceptionCode: c0000008 (Invalid handle)
  ExceptionFlags: 00000000
NumberParameters: 0
Thread tried to close a handle that was invalid or illegal to close

FAULTING_THREAD:  00002f00

PROCESS_NAME:  curl.exe

[...]

PR #8194, opened 2021-12-27
don't close prematurely socket of HTTPS proxy tunel

intxgo:
In case the remote host is unreachable, the SSL handshake between the proxy and the remote server won't happen. Curl should return "Received HTTP code 502 from proxy after CONNECT" and gracefully close the tunnel, but currently it prematurely closes the socket. This is against the design
https://github.com/curl/curl/blob/d4492b6d125d31ef5c74e4deb6786896606b70cc/lib/multi.c#L2548-L2549

In consequence the prematurely closed socket is used by the SSL clean up code
https://github.com/curl/curl/blob/d4492b6d125d31ef5c74e4deb6786896606b70cc/lib/multi.c#L2568-L2569

Ref:https://github.com/curl/curl/issues/8193

### DIFF
--- a/lib/http_proxy.c
+++ b/lib/http_proxy.c
@@ -677,7 +677,7 @@ static CURLcode CONNECT(struct Curl_easy *data,
       /* When connecting to a remote HTTPS server via established HTTPS proxy
        * tunnel don't close the socket here, SSL have to send shutdown alert
        * before closing it */
-      if (conn->http_proxy.proxytype != CURLPROXY_HTTPS) {
+      if(conn->http_proxy.proxytype != CURLPROXY_HTTPS) {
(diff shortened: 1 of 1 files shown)


bagder:
> In consequence the prematurely closed socket is used by the SSL clean up code

How is that used there? Is `dead_connection` FALSE in that case?

intxgo:
> How is that used there? Is `dead_connection` FALSE in that case?

Yes, since `result` is `CURLE_RECV_ERROR`

bagder:
Thanks! I made a slightly edited take on this in #8195 as it seems that A) if we can skip the close in the HTTPS case we can just as well always skip it and B) it seems suitable to add an assert that catches this case.

intxgo:
Closing as there's a better one to fix the issue ^

PR #8202, opened 2021-12-30
aws_sigv4: Fix "host:" canonical header

matwey:
Content of the "host:" canonical header must coincide with actual "Host:"
header sent to the server. Otherwise, the signature cannot be verified
correctly.

The actual "Host:" header and conn->host.name are different when:

 1. non-standard port is used, i.e. `Host: 127.0.0.1:9000`

 2. IPv6 address is used, i.e. `Host: [::1]:9000`

 3. user may provide his own custom "Host:" header.


outscale-mgo:
Hello,
If I'm not mistaken, this PR resolve the issue too: https://github.com/curl/curl/pull/7966 ?

matwey:
@outscale-mgo indeed, you are right. Do you have an estimate when the #7966 will be accepted?

matwey:
@outscale-mgo I've tested #7966 and unfortunately it doesn't resolve the issue currently. It would be great if you could handle my commends in #7966.

cosmo-ray:
As this issue has been mention recently #7966 so this this issue now :)

hansonchar:
Is this PR ready to be merged or do the 4 failing checks have to pass as a precondition?

bagder:
If #7966 fixes the same problem this PR wants to fix, then shouldn't this just be closed and 7966 should be polished to get merged? (I have not reviewed the PR because I thought so)

bagder:
This issue is already being worked on in #7966

PR #8205, opened 2022-01-01
CURLOPT_HAPROXYPROTOCOL: add support for the HAProxy PROXY protocol v2

remilapeyre:
Add --haproxy-protocol-v2 for the command line tool


bagder:
How about instead extending `CURLOPT_HAPROXYPROTOCOL` to accept either a bitmask or multiple values? The current (old) behavior accepts a '1' for protocol version 1, so it is perfectly possible to use '2' for version 2!

docs/libcurl/symbols-in-versions
```diff
@@ -451,6 +451,7 @@ CURLOPT_FTP_USE_PRET            7.20.0
 CURLOPT_GSSAPI_DELEGATION       7.22.0
 CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS 7.59.0
 CURLOPT_HAPROXYPROTOCOL         7.60.0
+CURLOPT_HAPROXYPROTOCOL         7.81.0
```
bagder, line 454: Typo (missing v2) and plan for 7.82.0!

remilapeyre:
> How about instead extending `CURLOPT_HAPROXYPROTOCOL` to accept either a bitmask or multiple values? The current (old) behavior accepts a '1' for protocol version 1, so it is perfectly possible to use '2' for version 2!

Just to be sure: we are only talking about the way the implementation, not the cli interface? If thought about this but wasn't sure whether we could change CURLOPT_HAPROXYPROTOCOL from a  bit to a bitmask.

In any case, we would still have multiple flag on the command line to choose the PROXY protocol version `--haproxy-protocol` and `--haproxy-protocol-v2` right? I initially wanted to make it possible to use `--haproxy-protocol`, `--haproxy-protocol=v1` and `--haproxy-protocol=v2` but it seems to me that this is not possible.

If you confirm that my understanding is correct I will update the patch to do this.

bagder:
Yes. You can add this without adding a new libcurl option. You cannot avoid a new command line option-

remilapeyre:
Thanks for the feedback @bagder, here's a new version that only uses `CURLOPT_HAPROXYPROTOCOL`.

lib/setopt.c
```diff
@@ -1817,9 +1817,9 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
 #ifndef CURL_DISABLE_PROXY
   case CURLOPT_HAPROXYPROTOCOL:
     /*
-     * Set to send the HAProxy Proxy Protocol header
+     * Set to send the HAProxy Proxy Protocol v1 header
```
bagder, line 1820: It's not just "v1" now...

lib/setopt.c
```diff
@@ -1817,9 +1817,9 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
...
      */
-    data->set.haproxyprotocol = (0 != va_arg(param, long)) ? TRUE : FALSE;
+    data->set.haproxyprotocol = va_arg(param, long);
```
bagder, line 1822: Please make this return an error if an unsupported value is set.

src/tool_getparam.c
```diff
@@ -869,7 +870,10 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
           return PARAM_LIBCURL_DOESNT_SUPPORT;
         break;
       case 'X': /* --haproxy-protocol */
-        config->haproxy_protocol = toggle;
+        config->haproxy_protocol = 1L;
+        break;
+      case 'Y': /* --haproxy-protocol-v2 */
+        config->haproxy_protocol = 2L;
```
bagder, line 876: Maybe `warnf()` if one protocol version option overwrites a previously set protocol option?

lib/http.c
```diff
@@ -1543,25 +1543,45 @@ static CURLcode add_haproxy_protocol_header(struct Curl_easy *data)
...
+      (char []) {
+        /* signature */
+        0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A,
+        /* command is LOCAL and address family is AF_UNSPEC */
+        0x20,
+        /* protocol is UNSPEC */
+        0x00,
+        /* there is no additionnal data */
+        0x00, 0x00
+        },
```
bagder, line 1581: How about making this chunk a `const static` array instead? Will also make it slightly more readable I think.

lib/http.c
```diff
@@ -1543,25 +1543,45 @@ static CURLcode add_haproxy_protocol_header(struct Curl_easy *data)
...
+      16
```
bagder, line 1582: and then you could use `sizeof()` here instead of a magic constant

lib/urldata.h
```diff
@@ -1797,6 +1797,7 @@ struct UserDefined {
   BIT(crlf);            /* convert crlf on ftp upload(?) */
   BIT(strip_path_slash); /* strip off initial slash from path */
   BIT(ssh_compression);            /* enable SSH compression */
+  long haproxyprotocol; /* whether to send HAProxy PROXY protocol header */
```
bagder, line 1800: 1. You could save a few bytes by not storing this as a `long`.
2. Move this field declaration to above the `BIT()` fields to better use struct space.

docs/cmdline-opts/haproxy-protocol-v2.d
```diff
@@ -0,0 +1,15 @@
+Long: haproxy-protocol-v2
+Help: Send HAProxy PROXY protocol v2 header
+Protocols: HTTP
+Added: 7.81.0
```
bagder, line 4: 7.82.0

bagder:
> `--haproxy-protocol-v2 is in the option directory but not in ./../docs/options-in-versions!`

Meaning that you need to amend `docs/options-in-versions` with info about this new option.

remilapeyre:
Hi @bagder, here's a new version that should address all your comments.

bagder:
~~~
http.c: In function ‘add_haproxy_protocol_header’:
http.c:1545:3: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 1545 |   static const char binaryheader[] = {
      |   ^~~~~~

bagder (changes requested):

remilapeyre:
Hi @bagder, I tried to fix all issues. I'm a bit confused why some test fail though.

bagder:
Did you actually get those two new tests to work without having to modify the test server?

bagder:
Also, I would advice you use the `%hex[` construct to add binary data into the tests to compare with. Similar to this:
~~~
<protocol>
%hex[ 0d 0a 0d 0a 00 0d 0a 51 55 49 54 0a 20 00 00 00 ]hex%GET /%TESTNUMBER HTTP/1.1
Host: %HOSTIP:%HTTPPORT
User-Agent: curl/%VERSION
Accept: */*
Testno: %TESTNUMBER
~~~

bagder:
Gah, sorry I messed up that commit. I'll force-push a cleanup

bagder:
Sorry for that. I just wanted to fix the copyright year ranges for you. :disappointed:

bagder:
I think I fixed the tests and test server to do right so that you can verify that curl sends the right data.

Doing this and checking the proxy protocol spec, I have questions though:

The code says `/* protocol is UNSPEC */` (and this is verified by the tests) - but why? The spec says that should be used "for an unknown, unspecified or unsupported protocol". This is neither unknown nor unsupported. We know that it is IPv4, IPv6 or a unix domain socket. So why doesn't this pass on the addresses, just like does for v1?

remilapeyre:
> I think I fixed the tests and test server to do right so that you can verify that curl sends the right data.
> 

Thanks for taking care of this.

> Doing this and checking the proxy protocol spec, I have questions though:
> 
> The code says `/* protocol is UNSPEC */` (and this is verified by the tests) - but why? The spec says that should be used "for an unknown, unspecified or unsupported protocol". This is neither unknown nor unsupported. We know that it is IPv4, IPv6 or a unix domain socket. So why doesn't this pass on the addresses, just like it does for v1?

Yes this is different than what it does for v1. I'm not well versed in what must be done here and I think it depends on the context, for example if cURL is being used to test a connection, or used to send health-checks or used to connect on behalf of a client.

In my case I would like to be able to use `curl --haproxy-protocol-v2 ...` to test a connection, other users might need more knobs. Here's my understanding:

 - for v1 https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt section 2.1. says:

>  For "UNKNOWN", the rest of the line before the
>    CRLF may be omitted by the sender, and the receiver must ignore anything
>   presented before the CRLF is found. Note that an earlier version of this
>   specification suggested to use this when sending health checks, but this
>    causes issues with servers that reject the "UNKNOWN" keyword. Thus is it
>    now recommended not to send "UNKNOWN" when the connection is expected to
>    be accepted, but only when it is not possible to correctly fill the PROXY
>    line.

so the current behavior for v1 seems correct to me.

- for v2 section 2.2 says:

> The lowest four bits represents the command :
>   - \x0 : LOCAL : the connection was established on purpose by the proxy
>     without being relayed. The connection endpoints are the sender and the
>     receiver. Such connections exist when the proxy sends health-checks to the
>     server. The receiver must accept this connection as valid and must use the
>     real connection endpoints and discard the protocol block including the
>     family which is ignored.

which I think is the case that matters here. The rest of the header is therefore set:

> The address family maps to the original socket family without necessarily
> matching the values internally used by the system. It may be one of :
>  
>   - 0x0 : AF_UNSPEC : the connection is forwarded for an unknown, unspecified
>     or unsupported protocol. The sender should use this family when sending
>     LOCAL commands or when dealing with unsupported protocol families. The
>     receiver is free to accept the connection anyway and use the real endpoint
>     addresses or to reject it. The receiver should ignore address information.



and 

> The transport protocol is specified in the lowest 4 bits of the 14th byte :
> 
>   - 0x0 : UNSPEC : the connection is forwarded for an unknown, unspecified
>     or unsupported protocol. The sender should use this family when sending
>     LOCAL commands or when dealing with unsupported protocol families. The
>     receiver is free to accept the connection anyway and use the real endpoint
>     addresses or to reject it. The receiver should ignore address information.
 
so in my case I think everything needs to be set as `UNSPEC`.

bagder:
> Yes this is different than what it does for v1.

That's exactly why I have a really hard time to understand and accept it. Why would the use case be different and much more limited just because you change to v2? To me, the protocol seems to be almost pointless if you don't actually provide any information in there. The intended purpose of the protocol is one and the same for both protocol versions. They're just a different ways to send the data (readable vs binary).

Put another way: is there a good reason to *not* provide the IP information?

bagder:
I cannot accept this PR as it is only half-baked and nobody works on it.

PR #8206, opened 2022-01-01
urlapi: if possible, shorten given numerical IPv6 addresses

bagder:
Extended test 1560 to verify

### DIFF
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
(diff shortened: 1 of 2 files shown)


lib/urlapi.c
```diff
@@ -671,10 +669,22 @@ static CURLUcode hostname_check(struct Curl_URL *u, char *hostname)
       /* hostname is fine */
     }
 #ifdef ENABLE_IPV6
-    hostname[hlen] = 0; /* end the address there */
-    if(1 != Curl_inet_pton(AF_INET6, hostname, dest))
-      return CURLUE_BAD_IPV6;
-    hostname[hlen] = ']'; /* restore ending bracket */
+    {
+      char dest[16]; /* fits a binary IPv6 address */
+      char norm[128];
```
jay, line 674: [MAX_IPADR_LEN](https://github.com/curl/curl/blob/curl-7_80_0/lib/urldata.h#L76-L77)

lib/urlapi.c
```diff
@@ -671,10 +669,22 @@ static CURLUcode hostname_check(struct Curl_URL *u, char *hostname)
...
+      hostname[hlen] = 0; /* end the address there */
+      if(1 != Curl_inet_pton(AF_INET6, hostname, dest))
+        return CURLUE_BAD_IPV6;
+
+      /* check if it can be done shorter */
+      Curl_inet_ntop(AF_INET6, dest, norm, sizeof(norm));
```
jay, line 680: check return

bagder: The only way this function can fail here is if we give it it a too small buffer to create the address in, and the define should make sure that doesn't happen, but yes you're right...

PR #8207, opened 2022-01-01
mbedtls: fix find_package to prevent dev warning due to mismatched arg

trackpadpro:
-- curl version=[7.81.0-DEV]
CMake Warning (dev) at /usr/share/cmake-3.22.1/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (MBEDTLS)
  does not match the name of the calling package (MbedTLS).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  deps/curl/CMake/FindMbedTLS.cmake:31 (find_package_handle_standard_args)
  deps/curl/CMakeLists.txt:473 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.


trackpadpro:
The case difference does cause differences in library link errors if CMake directories are not specified properly, but if the find_package is possible, then the only difference in changing this line of code (in my testing) is that fixing the case prevents dev warnings.

bagder:
Thanks!

PR #8214, opened 2022-01-02
mbedtls: fix v3.1.0 build private access

trackpadpro:
"As a last resort, you can access the field foo of a structure bar by writing bar.MBEDTLS_PRIVATE(foo). Note that you do so at your own risk, since such code is likely to break in a future minor version of Mbed TLS." - https://github.com/ARMmbed/mbedtls/blob/f2d1199edc5834df4297f247f213e614f7782d1d/docs/3.0-migration-guide.md

That future minor version is v3.1.0. I set the >= to == for the version checks relevant to the private members because v3.1.0 is an MbedTLS release, and I am not sure when the private designation was reverted after v3.0.0. The change made to support v3.0.0 causes builds to break for v3.1.0.

Build errors are included in the extended description of the latest commit. I do not believe this error would be exclusive to CMake, but that is what I am using for my project.


bagder:
Seems fair. I'll bump the mbedtls CI job in #8215 to use 3.1.0 once this is merged.

bagder:
Can you rebase/force-push this to remove the irrelevant commits already merged from #8207?

bagder:
Never mind, I'll do that.

bagder:
Thanks!

trackpadpro:
I have actually been wondering what the proper pull request format is. Will read into rebase and force push a bit to see if I can fix it in the future.

PR #8219, opened 2022-01-03
Allow binary zeros in tests

fabiankeil:
This patch set allows to use binary zeros in tests.

I mainly needed binary zeros for Privoxy tests but it doesn't hurt
to explicitly confirm that curl can deal with them as well.


bagder (approved):

bagder:
Thanks, merged as of 87427298272679

PR #8217, opened 2022-01-03
fix the socks5h ATYP field and address passed on

bagder:

PR #8223, opened 2022-01-04
test719: require ipv6 support to run

bagder:
Follow-up to effd2bd7ba2a5fd244
Reported-by: Marc Hörsken
Bug: https://github.com/curl/curl/pull/8217#issuecomment-1004681145

### DIFF
--- a/tests/data/test719
+++ b/tests/data/test719
@@ -30,6 +30,9 @@ Funny-head: yesyes
 #
 # Client-side
 <client>
+<features>
+ipv6
+</features>
 <server>
 http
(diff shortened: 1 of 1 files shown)


mback2k:
Thanks! Something still seems to be wrong with this AppVeyor build failing test 719 through 721:
https://ci.appveyor.com/project/curlorg/curl/builds/42084440/job/qem30th6bj7d0h46?fullLog=true#L5793

bagder:
Ah, because the test also fails to check if "proxy" is enabled...

PR #8233, opened 2022-01-06
check the return value of BIO_new_mem_buf()

x2018:
check the return value of BIO_new_mem_buf() to catch memory errors in time.


bagder (approved):

bagder:
Thanks! BTW, if you like to have your name "properly" attributed to your commits in git (seeing this looks like your 4th one) then I would be happy to make sure it happens. I means instead of x2018 or xkernel. Your choice, you don't have to.

x2018:
Oh. my full name is Xiaoke Wang. Thanks.

bagder:
Now present correctly in git too since 6e7ce9bbfb2

PR #8236, opened 2022-01-06
examples/multi-app.c: call curl_multi_remove_handle as well

### LINKED ISSUE #8234: multi-app.c example - missing call + question
You are missing a `curl_multi_remove_handle()` call in the [multi-app.c](https://curl.se/libcurl/c/multi-app.html) example. Right?

You could also clean-up the easy handles inside the curl_multi_info_read() method, correct?

Unrelated question: [Sometimes I see](https://curl.se/libcurl/c/crawler.html) the `while(..curl_multi_info_read())` inside the main `while(still_running)`. But this [multi-app.c](https://curl.se/libcurl/c/multi-app.html) example the two while loops are after each other.

What is the best way to execute `curl_multi_info_read()`? **Inside** the while loop of the other while (like crawler.c or 10-at-a-time.c examples), or **after each other** like the multi-app.c example?

Thanks!

Regards,
Melroy van den Berg

bagder:
Fixes #8234
Reported-by: Melroy van den Berg

### DIFF
--- a/docs/examples/multi-app.c
+++ b/docs/examples/multi-app.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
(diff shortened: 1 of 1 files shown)


danger89:
LGTM.

PR #8238, opened 2022-01-07
mbedtls: Fix ssl_init error with mbedTLS 3.1.0+

Koromix:
Since mbedTLS 3.1.0, mbedtls_ssl_setup() fails if the provided
config struct is not valid.

mbedtls_ssl_config_defaults() needs to be called before the config
struct is passed to mbedtls_ssl_setup().

### DIFF
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -469,18 +469,18 @@ mbed_connect_step1(struct Curl_easy *data, struct connectdata *conn,
   infof(data, "mbedTLS: Connecting to %s:%ld", hostname, port);
 
   mbedtls_ssl_config_init(&backend->config);
-
-  mbedtls_ssl_init(&backend->ssl);
-  if(mbedtls_ssl_setup(&backend->ssl, &backend->config)) {
-    failf(data, "mbedTLS: ssl_init failed");
-    return CURLE_SSL_CONNECT_ERROR;
-  }
   ret = mbedtls_ssl_config_defaults(&backend->config,
                                     MBEDTLS_SSL_IS_CLIENT,
                                     MBEDTLS_SSL_TRANSPORT_STREAM,
                                     MBEDTLS_SSL_PRESET_DEFAULT);
   if(ret) {
     failf(data, "mbedTLS: ssl_config failed");
+    return CURLE_SSL_CONNECT_ERROR;
+  }
+
+  mbedtls_ssl_init(&backend->ssl);
+  if(mbedtls_ssl_setup(&backend->ssl, &backend->config)) {
+    failf(data, "mbedTLS: ssl_init failed");
     return CURLE_SSL_CONNECT_ERROR;
   }
 


bagder:
@trackpadpro as you've been editing this code recently, does this change look good to you as well?

trackpadpro (commented):
This change looks good to me, but I cannot test it right now on one of my computers. I would like to see the error log you ran into that required this change, but in general, this does not seem like it would cause any problems.

Koromix:
I encountered this problem after updating to mbedTLS 3.1.0 when trying to connect to an SMTP:

```
* Connected to mail.gandi.net (217.70.178.9) port 465 (#0)
* mbedTLS: Connecting to mail.gandi.net:465
* mbedTLS: ssl_init failed
* Closing connection 0
::ffff:127.0.0.1: Error: Failed to perform mail call: SSL connect error
```

As far as I can tell, the config check was introduced here: https://github.com/ARMmbed/mbedtls/pull/4853/files#diff-76b73bf2a8157ee95fdd3b688384812654c63878be4ecce6ee56402eb52b1f4fR3205-R3206

mbedtls_ssl_setup() calls ssl_conf_check(), which fails if the config is not valid.

bagder:
Also, the mbedtls CI job runs a lot of tests just fine with this PR applied so it seems rather safe.

trackpadpro:
I agree that it's rather safe, but to be absolutely certain of backwards compatibility you could always add some version conditionals ("Mbed TLS 2.28 is a long-time support branch. It will be supported with bug-fixes and security fixes until end of 2024."). 

v3.1.0 is MBEDTLS_VERSION_NUMBER == 0x03010000 if you want to add the (possibly overkill) checks

bagder:
It would be enough if someone just built with that version and ran through a few tests

Koromix:
I'll test with mbedTLS 3.0, mbedTLS 2.28 and 2.16 later today, and report back to you.

Koromix:
First, _without_ the patch:

```
# mbedtls-3.1.0
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
curl: (35) mbedTLS: ssl_init failed

# mbedtls-3.0.0
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success

# mbedtls-2.28.0
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success

# mbedtls-2.16.12
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success
```

And now _with the fix_ applied:

```
# mbedtls-3.1.0
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success

# mbedtls-3.0.0
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success

# mbedtls-2.28.0
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success

# mbedtls-2.16.12
$ ./curl --no-progress-meter -o/dev/null https://www.google.fr/ && echo Success
Success
```

bagder:
Awesome, great work. Merging now...

bagder:
Thanks!

PR #8243, opened 2022-01-07
url: given a user in the URL, find pwd for that user in netrc

### LINKED ISSUE #8241: multiple users per host in ~/.netrc not working
Curl does not use different credentials on this:
```
  curl -n ftp://foouser@example.com
  curl -n ftp://baruser@example.com
```

Almost all clients (wget, lftp, ftp, ...) pick the matching entry out of ~/.netrc,
but curl just takes the first ~/.netrc entry for "example.com".

Or, as a standalone example, using http:
```
cat > curl.netrc <<EOF
machine example.com
login foouser
password foouser

machine example.com
login baruser
password baruser
EOF

#> curl -qsS --netrc-file curl.netrc --trace-ascii - http://baruser@example.com | grep "user"
== Info: Server auth using Basic with user 'foouser'
#> curl -qsS --netrc-file curl.netrc --trace-ascii - http://foouser@example.com | grep "user"
== Info: Server auth using Basic with user 'foouser'
```

### curl/libcurl version

curl 7.79.1

bagder:
Add test 380 to verify

Reported-by: Manfred Schwarb
Fixes #8241

### DIFF
--- a/docs/libcurl/opts/CURLOPT_NETRC.3
+++ b/docs/libcurl/opts/CURLOPT_NETRC.3
@@ -5,7 +5,7 @@
 .\" *                            | (__| |_| |  _ <| |___
 .\" *                             \___|\___/|_| \_\_____|
 .\" *
-.\" * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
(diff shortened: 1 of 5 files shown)


bagder:
I need to double-check that this actually follows the `CURLOPT_NETRC.3` documentation...

PR #8255, opened 2022-01-09
rustls: CURLOPT_CAINFO_BLOB support

sagebind:
Add support for `CURLOPT_CAINFO_BLOB` and `CURLOPT_PROXY_CAINFO_BLOB` to the rustls TLS backend. Multiple certificates in a single PEM string are supported just like OpenSSL does with this option.

This is compatible at least with rustls-ffi 0.8+ which is our new minimum version anyway.

I was able to build and run this on Windows, pulling trusted certs from the system and then add them to rustls by setting `CURLOPT_CAINFO_BLOB`. Handy!


bagder (approved):

bagder:
Thanks!

PR #8258, opened 2022-01-11
CURLOPT_RESOLVE.3: change example port to 443

cwill747:
83cc96670811cd0e6cccbf78d2b21bbeb2e4ea45 changed documentation from using
http to https. However, CURLOPT_RESOLVE being set to port 80 in the
documentation means that it isn't valid for the new URL. Update to 443.


jay:
Thanks

PR #8260, opened 2022-01-11
mbedtls: fix CURLOPT_SSLCERT_BLOB (again)

jay:
- Increase the buffer length passed to mbedtls_x509_crt_parse to account
  for the null byte appended to the temporary blob.

Follow-up to 867ad1c which uses a null terminated copy of the
certificate blob, because mbedtls_x509_crt_parse requires PEM data
to be null terminated.

Ref: https://github.com/curl/curl/commit/867ad1c#r63439893
Ref: https://github.com/curl/curl/pull/8146

Closes #xxxx

---

Note the [documentation](https://tls.mbed.org/api/group__x509__module.html#ga033567483649030f7f859db4f4cb7e14) says null termination applies specifically to PEM data (and therefore not to DER?):

_The size of buf, including the terminating NULL byte in case of PEM encoded data._

/cc @Koromix

### DIFF
--- a/docs/libcurl/opts/CURLOPT_SSLCERT_BLOB.3
+++ b/docs/libcurl/opts/CURLOPT_SSLCERT_BLOB.3
@@ -33,8 +33,9 @@ CURLcode curl_easy_setopt(CURL *handle, CURLOPT_SSLCERT_BLOB,
 .SH DESCRIPTION
 Pass a pointer to a curl_blob structure, which contains (pointer and size) a
 client certificate. The format must be "P12" on Secure Transport or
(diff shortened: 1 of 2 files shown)


Koromix:
Thanks!

I've been backing on Mbed TLS yesterday and I made a pull request to not require NUL-termination on PEM strings:
https://github.com/ARMmbed/mbedtls/pull/5411

Hopefully it goes through eventually, if and when it does I will make a PR here to remove the extra allocations and copy on afftected Mbed TLS versions.

bagder (approved):

PR #8262, opened 2022-01-11
curl_multi_socket.3: remove callback and typical usage descriptions

bagder:
1. The callback is better described in the option for setting it. Having
it in a single place reduces the risk that one of them is wrong.

2. The "typical usage" is wrong since the functions described in this
man page are both deprecated so they cannot be used in any "typical" way
anymore.

### DIFF
--- a/docs/libcurl/curl_multi_socket.3
+++ b/docs/libcurl/curl_multi_socket.3
@@ -5,7 +5,7 @@
 .\" *                            | (__| |_| |  _ <| |___
 .\" *                             \___|\___/|_| \_\_____|
 .\" *
-.\" * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+.\" * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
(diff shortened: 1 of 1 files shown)


docs/libcurl/curl_multi_socket.3
```diff
@@ -66,68 +66,6 @@ equivalent to \fIcurl_multi_socket_action(3)\fP with \fBev_bitmask\fP set to
...
-particular socket. Note that a single handle may work with several sockets
-simultaneously.
-
-The \fIs\fP argument is the actual socket value as you use it within your
-system.
-
-The \fIaction\fP argument to the callback has one of five values:
-.RS
-.IP "CURL_POLL_NONE (0)"
-register, not interested in readiness (yet)
```
jay, line 90: note CURL_POLL_NONE is not mentioned in CURLMOPT_SOCKETFUNCTION

bagder: I guess that proves my point... :grin:

Looking at the code, I don't think `CURL_POLL_NONE` can actually be passed to the callback so I think it is correct to not document it as such.

PR #8263, opened 2022-01-11
HTTP/2 stream window options

bagder:
This is #7848 with some minor edits (year/version fixes mostly) plus my added commit to only provide the option if HTTP/2 is actually enabled in the build.


lib/multihandle.h
```diff
@@ -139,7 +139,9 @@ struct Curl_multi {
   struct curltime timer_lastcall; /* the fixed time for the timeout for the
                                     previous callback */
   unsigned int max_concurrent_streams;
-
+#if defined(USE_NGHTTP2) || defined(USE_HYPER)
+  unsigned int stream_window_size;
+#endif
```
jay, line 144: the other uses of stream_window_size are guarded by USE_HTTP2, would that be more appropriate here?

lib/easy.c
```diff
@@ -681,6 +681,12 @@ static CURLcode easy_perform(struct Curl_easy *data, bool events)
   /* Copy the MAXCONNECTS option to the multi handle */
   curl_multi_setopt(multi, CURLMOPT_MAXCONNECTS, data->set.maxconnects);
 
+#ifdef USE_HTTP2
+  /* Copy the STREAM_WINDOW_SIZE option to the multi handle */
+  curl_multi_setopt(multi, CURLMOPT_STREAM_WINDOW_SIZE,
+                    data->set.stream_window_size);
+#endif
```
jay, line 688: The documentation says an individual easy handle's stream size setting is ignored by the multi, not inherited. The user is expected to set CURLMOPT_STREAM_WINDOW_SIZE.

lib/http2.c
```diff
@@ -2378,7 +2376,7 @@ CURLcode Curl_http2_stream_pause(struct Curl_easy *data, bool pause)
   else {
     struct HTTP *stream = data->req.p.http;
     struct http_conn *httpc = &data->conn->proto.httpc;
-    uint32_t window = !pause * HTTP2_HUGE_WINDOW_SIZE;
+    uint32_t window = !pause * data->multi->stream_window_size;
```
jay, line 2379: I'm unclear about how this works. It looks like the connection stream size is always set to `multi->stream_window_size` but an individual stream's window size is only set to `multi->stream_window_size` when unpaused?

What if we continued to set connection stream size to HTTP2_HUGE_WINDOW_SIZE, and got rid of `multi->stream_window_size` (CURLMOPT_STREAM_WINDOW_SIZE), and then set each individual stream size using 
`data->set.stream_window_size` (CURLOPT_STREAM_WINDOW_SIZE)? Would that work, is it better/worse?

lib/multihandle.h
bagder: Yes!

lib/easy.c
bagder: This particular setopt is called within the `curl_easy_perform()` function for the "internal" multi handle so using the easy handle's value for this seems correct to follow the docs.

lib/http2.c
bagder: Streams get a window size set when they are created - the initial window size. That's not something this change modifies. This change only changes the specific **size** used in the window size manipulations.

I can think of two reasons for having it in the multi handle, none of them very strong:

1. For push streams that are created by libcurl itself. Without the multi interface having the default, the callback would need to set the window size in with `curl_easy_setopt()` and we would have to add logic to deal with that. (including calling libcurl back from within the callback).
2. A convenience thing. Setting it in the multi handle magically affects all transfers done using that handle and you then don't have to modify every single transfer you add.

jay: > Streams get a window size set when they are created - the initial window size. That's not something this change modifies. This change only changes the specific **size** used in the window size manipulations.

I still don't understand what you are talking about I need an ELI5 on this. Wouldn't we want to change the initial window size of individual streams with these options? CURLOPT_STREAM_WINDOW_SIZE implies it's going to change the window size and both options are documented as  _"Pass a long indicating the sender's HTTP/2 initial stream window **size** in bytes. This setting affects the initial window size of all streams."_

So does all this mean the options are intended to change the window size of all streams in the handle or all streams combined (ie connection window size), but also if a stream is paused will also change that individual stream's window size?

bagder: # Window size

Each stream has a window size that is maintained during the stream's life time (similar to TCP's segment numbers). It tells the other side how much data we accept without upping the window. As we receive data, the window needs to be updated (so that the window doesn't fill up as then the peer will not send anymore) and nghttp2 does it automatically for us. There's also a window size for data in the other direction that is then maintained by the peer, but we can ignore that for now.

Each stream window size gets their initial values when they are created. That's the value the options are for.

# Pausing streams

When a stream is explicitly paused using the libcurl API, we set the local window size explicitly to zero to say that we cannot accept anymore data, but at that time there might already be up to the previous window_size amount of data in transit to us so we need to be able to buffer that amount of data.

When a stream is again unpaused, we set the window size back to the initial value to tell the peer that we can again accept this amount of data in our end.

jay: > Each stream window size gets their initial values when they are created. That's the value the options are for.

Makes sense. I didn't think that was happening:

> I'm unclear about how this works. It looks like the connection stream size is always set to multi->stream_window_size but an individual stream's window size is only set to multi->stream_window_size when unpaused?

I had thought that NGHTTP2_SETTINGS_INITIAL_WINDOW_SIZE was used to set the default initial connection window size but actually it is used to set the default initial stream window size (via SETTINGS[0] frame).

> As we receive data, the window needs to be updated (so that the window doesn't fill up as then the peer will not send anymore) and nghttp2 does it automatically for us.

Can it automatically update the window size of a paused stream? In other words can it change the paused stream window size from 0 to some other value?

bagder:
I'm dropping this

Issue #8248, opened 2022-01-08
Appveyor CI doesn't have autoreconf anymore


bagder:
It seems the autotools builds on appveyor now permafail because `./buildconf` fails when it runs `autoreconf`:
~~~
*** Do not use buildconf. Instead, just use: autoreconf -fi
./buildconf: line 4: exec: autoreconf: not found
Command exited with code 127
~~~
[example log](https://ci.appveyor.com/project/curlorg/curl/builds/42129185/job/pvgioxqlky4m0gw7)

MarcelRaad:
IIRC, MSYS2 not having some required packages in newer AppVeyor images was exactly the reason why I used the Visual Studio 2015 image in https://github.com/curl/curl/commit/0f234a5cdebd455f324cdf64b604b4a1340dbbfe. Strange that it seems to have worked then by https://github.com/curl/curl/commit/b667bbe2caf7e7d285b225794f6ae041bdf36093 and broke again now.

I wanted to prepare a PR to update to the VS 2022 images anyway, so I'd just use a consistent working image for all autotools builds (2015 or 2017). Or does anyone have a strong opinion on using a newer image? I'd like to avoid installing packages if possible, which sometimes fails in other CI systems.

PR #8265, opened 2022-01-11
AppVeyor: update build images

MarcelRaad:
- Update current VS 2019 images to VS 2022
- Use VS 2022 also for the classic MinGW job, which previously used VS 2015 because the VS 2017 image doesn't have it installed
- Use VS 2017 for the autotools builds as the VS 2019 and 2022 images don't have autotools

Fixes https://github.com/curl/curl/issues/8248


bagder:
That green check-mark is pure beauty! :heavy_check_mark:

PR #8268, opened 2022-01-12
check the return value of SSL_get_ex_data()

x2018:
check the return value of `SSL_get_ex_data()` to prevent potential NULL dereference.
`SSL_set_ex_data()` called by `ossl_associate_connection()` may fail, so it is better to check the return value from `SSL_get_ex_data()`, especially `data`  which will be dereferenced in the following code. And though `sockindex`  seems meaningless in `Curl_ssl_getsessionid()` or `Curl_ssl_addsessionid()`?

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -244,7 +244,7 @@ struct ssl_backend_data {
 #endif
 };
 
-static void ossl_associate_connection(struct Curl_easy *data,
+static bool ossl_associate_connection(struct Curl_easy *data,
                                       struct connectdata *conn,
                                       int sockindex);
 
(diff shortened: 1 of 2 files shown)


lib/vtls/openssl.c
```diff
@@ -2491,13 +2491,12 @@ static int ossl_new_session_cb(SSL *ssl, SSL_SESSION *ssl_sessionid)
...
-  if(!conn)
-    return 0;
-
   data = (struct Curl_easy *) SSL_get_ex_data(ssl, data_idx);
-
   /* The sockindex has been stored as a pointer to an array element */
   sockindex_ptr = (curl_socket_t*) SSL_get_ex_data(ssl, sockindex_idx);
+  if(!conn || !data || !sockindex_ptr)
+    return 0;
+
```
bagder, line 2499: If no error has been returned before, can this really happen?

And if this can happen, doesn't it signify a major error somewhere to call for everything to get closed down?

x2018:
Yes, it looks like this can happen as https://github.com/curl/curl/blob/84853d94f2b8ac1cfd2ed3e3849e2af0f464d22c/lib/vtls/openssl.c#L4549 is `void` type and inside it does not check the return value of `SSL_set_ex_data`, while `SSL_set_ex_data` may fail because of some internal errors.
To raise the error in time, I will change the signature of `ossl_associate_connection` and add several checks these days.

x2018:
I updated the signature of ossl_associate_connection and added several checks. And I retain the usage of Curl_ssl->associate_connection() in vtls.c as before. I am not sure whether these changes make sense. Thank you for taking the time to review this PR.

bagder (approved):

bagder:
Thanks!

PR #8272, opened 2022-01-13
formdata: avoid size_t => long typecast overflows

bagder:
Typically a problem for platforms with 32 bit long and 64 bit size_t

Reported-by: Fabian Yamaguchi
Bug: https://hackerone.com/reports/1444539


dfandrich:
This class of problem could merit a mention in libcurl-security.3, especially since many people likely think long and size_t are the same size everywhere.

PR #8273, opened 2022-01-13
quiche: after connect, verify the host name in the server cert

bagder:
This currently causes valgrind errors in quiche.


bagder:
#8275 is @ghedo's better take on this

PR #8274, opened 2022-01-13
mqtt: free any leftover when done

bagder:
Oss-fuzz found an issue when the "sendleftovers" pointer could leak memory.
Fix this by always freeing it (if still assigned) in the done function.

Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43515

### DIFF
--- a/lib/mqtt.c
+++ b/lib/mqtt.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 2020 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 2020 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  * Copyright (C) 2019, Björn Stenberg, <bjorn@haxx.se>
  *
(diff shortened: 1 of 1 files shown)


lib/mqtt.c
```diff
@@ -692,6 +694,16 @@ static CURLcode mqtt_do(struct Curl_easy *data, bool *done)
...
+static CURLcode mqtt_done(struct Curl_easy *data,
+                          CURLcode status, bool premature)
+{
+  struct MQTT *mq = data->req.p.mqtt;
+  (void)status;
+  (void)premature;
+  Curl_safefree(mq->sendleftovers);
+  return CURLE_OK;
+}
+
```
danielgustafsson, line 706: While I haven't done any testing, from reading this seems perfectly reasonable. Shouldn't however `mqtt_doing` set `mq->sendleftovers` to `NULL` as well as a belts-and-suspenders type operation?

danielgustafsson (approved):

lib/mqtt.c
danielgustafsson: Re-reading it again I concur.

bagder: thanks for the check!

PR #8275, opened 2022-01-13
quiche: verify the server cert on connect

ghedo:
Similarly to c148f0f551f9bea0e3d08f5747b3fe58b811a011, make quiche
correctly acknowledge `CURLOPT_SSL_VERIFYPEER` and
`CURLOPT_SSL_VERIFYHOST`.

Ref: #8173

### DIFF
--- a/lib/vquic/quiche.c
+++ b/lib/vquic/quiche.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
@@ -185,7 +185,8 @@ static SSL_CTX *quic_ssl_ctx(struct Curl_easy *data)
 {
   SSL_CTX *ssl_ctx = SSL_CTX_new(TLS_method());
 
-  SSL_CTX_set_alpn_protos(ssl_ctx, QUICHE_H3_APPLICATION_PROTOCOL,
+  SSL_CTX_set_alpn_protos(ssl_ctx,
+                          (const uint8_t *)QUICHE_H3_APPLICATION_PROTOCOL,
                           sizeof(QUICHE_H3_APPLICATION_PROTOCOL) - 1);
 
   SSL_CTX_set_default_verify_paths(ssl_ctx);
@@ -362,7 +363,6 @@ static CURLcode quiche_has_connected(struct Curl_easy *data,
 
   if(conn->ssl_config.verifyhost) {
     X509 *server_cert;
-    CURLcode result;
     server_cert = SSL_get_peer_certificate(qs->ssl);
     if(!server_cert) {
       return CURLE_PEER_FAILED_VERIFICATION;
--- a/lib/vquic/quiche.c
+++ b/lib/vquic/quiche.c
@@ -181,6 +181,17 @@ static void keylog_callback(const SSL *ssl, const char *line)
   Curl_tls_keylog_write_line(line);
 }
 
+static int verify_callback(int ok, X509_STORE_CTX *store_ctx)
+{
+  SSL *ssl = X509_STORE_CTX_get_ex_data(store_ctx,
+                                        SSL_get_ex_data_X509_STORE_CTX_idx());
+
+  struct quicsocket *qs = SSL_get_app_data(ssl);
+
+  qs->peer_verify_failed = !ok;
+  return ok;
+}
+
 static SSL_CTX *quic_ssl_ctx(struct Curl_easy *data)
 {
   SSL_CTX *ssl_ctx = SSL_CTX_new(TLS_method());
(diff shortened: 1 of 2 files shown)


bagder:
This is great. Just one thing though:

When I try this on a HTTP/3 server using a selfsigned cert, I want to get a return code back like for this case using plain HTTPS. The challenge is really that `quiche_conn_recv` end up returning -10 (`QUICHE_ERR_TLS_FAIL`) for this, which feels very generic. Is there any way I can enhance this to realize when this fails strictly because the cert verification failed?

~~~
$ ./src/curl --http3 https://localhost:7443/ -v
....

*   Trying 127.0.0.1:7443...
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
* Connect socket 5 over QUIC to 127.0.0.1:7443
* Sent QUIC client Initial, ALPN: h3,h3-29,h3-28,h3-27
* STATE: CONNECT => CONNECTING handle 0x5642d6e67618; line 1896 (connection #0)
* quiche_conn_recv() == -10
* connect to 127.0.0.1 port 7443 failed: Failure when receiving data from the peer
*   Trying ::1:7443...
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
* Connect socket 6 over QUIC to ::1:7443
* Sent QUIC client Initial, ALPN: h3,h3-29,h3-28,h3-27
* quiche_conn_recv() == -10
* connect to ::1 port 7443 failed: Failure when receiving data from the peer
* Failed to connect to localhost port 7443 after 14 ms: Failure when receiving data from the peer
* multi_done: status: 56 prem: 1 done: 0
* Closing connection 0
* Expire cleared (transfer 0x5642d6e67618)
curl: (56) quiche_conn_recv() == -10

ghedo:
If `QUICHE_ERR_TLS_FAIL` (`-10`) is returned, then I _think_ you can just call the `ERR_peek_error()` function to get the internal BoringSSL error, which would provide more context https://github.com/google/boringssl/blob/405c7888a7070c7106a94e7dfb96c15d267c0aeb/include/openssl/err.h#L202

bagder:
Unfortunately not. I inserted:

~~~c
      if(QUICHE_ERR_TLS_FAIL == recvd) {
        uint32_t error = ERR_peek_error();
        infof(data, "peek error == %u", error);
      }
~~~

Outputs:
~~~
*   Trying ::1:7443...
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
* Connect socket 6 over QUIC to ::1:7443
* Sent QUIC client Initial, ALPN: h3,h3-29,h3-28,h3-27
* peek error == 0
* quiche_conn_recv() == -10
* connect to ::1 port 7443 failed: Failure when receiving data from the peer
~~~

So, it returns a 0 error.

ghedo:
Hmm, it looks like the error gets cleared internally because quiche calls additional BoringSSL functions after the failure itself.

An alternative could be to look at the actual QUIC error generated by the client, which includes TLS alerts, something like:
```c
    if(QUICHE_ERR_TLS_FAIL == recvd) {
      bool is_app;
      uint64_t error_code;
      const uint8_t *reason;
      size_t reason_len;

      quiche_conn_local_error(qs->conn, &is_app, &error_code,
                              &reason, &reason_len);

      if(error_code > 0x100)
        infof(data, "peek error == %s",
              SSL_alert_desc_string_long(error_code - 0x100));
    }
```
which for me generates this log:
```
* peek error == unknown CA
```
though there are a few of these alerts related to certificate validation (e.g. `SSL_AD_CERTIFICATE_EXPIRED`, 
`SSL_AD_CERTIFICATE_REVOKED`, `SSL_AD_UNKNOWN_CA`, `SSL_AD_BAD_CERTIFICATE`,
`SSL_AD_CERTIFICATE_UNKNOWN`), so checking for each one in code might be annoying.

bagder:
Excellent, thanks! That seems to do it. I pushed a commit based on your lead.

ghedo:
Note that not all error codes > 0x100 are related to peer validation, those are just [QUIC error codes that "embed" TLS alerts](https://datatracker.ietf.org/doc/html/rfc9001#section-4.8) (that is, if you remove 0x100, you get the wire codepoint for the corresponding [TLS alert](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-6)). Of those, only the alerts I listed above (and maybe more) are actually certificate-related.

So you might want to do something like:
```c
switch(error_code - 0x0100) {
case SSL_AD_CERTIFICATE_EXPIRED:
case SSL_AD_CERTIFICATE_REVOKED:
case SSL_AD_UNKNOWN_CA:
case SSL_AD_BAD_CERTIFICATE:
case SSL_AD_CERTIFICATE_UNKNOWN:
  return CURLE_PEER_FAILED_VERIFICATION;

default:
  /* fallthrough */
}
```
though we could probably expose this information better on the quiche API somehow (e.g. have an error code specific to peer verification or something like that).

bagder:
> we could probably expose this information better on the quiche API somehow (e.g. have an error code specific to peer verification or something like that).

... it would certainly make it easier for us, and be a little less magic since we're now assuming a lot of TLS internals will come through without this being "promised" in any API.

ghedo:
In the meantime I think I figured out a better way to do this in curl. Since we already call `SSL_CTX_set_verify()`, we can just implement the callback and store the verification result for later use. Added a commit to do that (I remove the TLS alert logging, though it can be re-added if needed).

ghedo:
Oh, wait, I'm dumb. I just realized we can just call `SSL_get_verify_result()` like the normal OpenSSL backend does and just check that... added yet another commit.

bagder:
Excellent!

bagder:
Thanks!

PR #8295, opened 2022-01-19
HTTP/SASL/SOCKS: Implement CURLOPT_SAFE_AUTH

monnerat:
These options disable clear text password authentication mechanisms over
a non-encrypted connection. In particular:
- HTTP Basic authentication is disabled
- PLAIN and LOGIN mechanisms are disabled for all SASL protocols
- IMAP LOGIN is disabled
- POP3 USER/PASS authentication is disabled
- FTP authentication is disabled unless the user name is "anonymous"
- LDAP simple bind is disabled if not anonymous.
- SOCKS5 proxy basic authentication is disabled.

TLS-encrypted connections (either implicit or explicit) are not affected
as they wont transport clear text passwords.

These options are triggered by the new curl tool options --safe-auth and
--proxy-safe-auth respectively.

New tests 730-746 check these options in various contexts.


bagder:
Test 452 needs to be disabled for rustls for now:

    12:24:46.379717 == Info: rustls_connect: unimplemented

(there's a list of disabled tests in `tests/data/DISABLED` already, just add 452 to that)

monnerat:
Thanks for the hint.
I was scratching my head about this failing test that is very similar to test400 and I did not see the latter was already disabled.

docs/libcurl/opts/CURLOPT_SAFE_AUTH.3
```diff
@@ -0,0 +1,59 @@
...
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+
+CURLcode curl_easy_setopt(CURL *handle, CURLOPT_SAFE_AUTH, long onoff);
+.fi
+.SH DESCRIPTION
+Pass a long as parameter, 0 or 1 where 1 is for enable and 0 for disable.
+When set, This option rejects authentication mechanisms that would
+transfer clear passwords on a non-encrypted connection.
```
bagder, line 35: It feels reasonable to expect a user to mostly want to set this for both host+proxy at the same time, or not at all. Therefore I think it could make sense to be able to set both using the same libcurl option using a bitmask instead of adding two separate options.

lib/http.c
```diff
@@ -90,6 +90,11 @@
 #include "curl_memory.h"
 #include "memdebug.h"
 
+/* Authentication mechanisms that do not transfer clear password. */
+#define SAFE_AUTH_MECHS (CURLAUTH_NEGOTIATE | CURLAUTH_BEARER |         \
+                         CURLAUTH_DIGEST | CURLAUTH_NTLM |              \
+                         CURLAUTH_NTLM_WB | CURLAUTH_AWS_SIGV4)
+
```
bagder, line 97: Shouldn't this be the same bitmask as `CURLAUTH_ANYSAFE` ?

docs/libcurl/opts/CURLOPT_SAFE_AUTH.3
bagder: What happens if this option is set true and `CURLOPT_HTTPAUTH` sets Basic? It would probably help if this spells that out.

monnerat: Thanks for the review.
Sure, these are contradictory options.
In case it occurs, the CURLOPT_SAFE_AUTH takes the precedence. This is also true for SASL mechanisms. I'll add a doc paragraph for that.

lib/http.c
monnerat: Yes, I missed it. Will change.

docs/libcurl/opts/CURLOPT_SAFE_AUTH.3
```diff
@@ -0,0 +1,59 @@
...
+
+For authentication with a proxy, see \fICURLOPT_PROXY_SAFE_AUTH(3)\fP.
+.SH DEFAULT
+0, disabled
+.SH PROTOCOLS
+FTP, HTTP, IMAP, LDAP, POP3 and SMTP
```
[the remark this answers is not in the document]
monnerat: The ssh protocol always login on a secure channel. See https://www.ssh.com/academy/ssh/protocol#how-does-the-ssh-protocol-work.
Thus there's no need to handle these options there.

docs/libcurl/opts/CURLOPT_SAFE_AUTH.3
monnerat: > ... set both using the same libcurl option using a bitmask instead of adding two separate options.

I did it that way because we don't have a `curl_easy_getopt()` (BTW: possible future feature?). But I'll change it if you prefer.
What about flag names? I suggest CURLSAFE_PWD and CURLSAFE_PROXY_PWD.

monnerat:
Ping ?

tlhackque:
I noticed that https://github.com/curl/curl/pull/8986 should be coordinated with this, to resolve the discussion about falling back to insecure cleartext auth.  E.g. if SAFE_AUTH is set, fallback should requuire TLS.

monnerat:
Sure! As ever, there can be races between PRs even if no conflict occurs.
In this case, the latest commited should be adapted.
Thanks for pulling the attention on it.
For the time being, the current PR is expecting your vote: https://curl.se/mail/lib-2022-05/0067.html

tlhackque:
I agree that there should be a mechanism that disables cleartext authentication.  It should be policy statement for the handle.  Protocols that are implicitly secure should ignore the flag.  Protocols that are cleartext only should fail.

I don't think CURLE_LOGIN_DENIED is appropriate - it's documented as the **server** rejected the login.  Find a code that says the client policy rejected the authentication - or add a new error code.

If we were starting from scratch, I'd say this option should be the default - e.g. you have to specify "CURL_UNSAFE_{,PROXY)_AUTH".

Perhaps a CURL_GLOBAL_INIT flag could make it the default for all handles?

Perhaps queue changing the default for the next time (if ever) there's a SONAME bump...

Perhaps the tool doc should suggest putting the cmdline flags in curlrc...

Although backward compatibility is very important, in today's world introducing something the defaults to insecure feels wrong.

Modulo the above comments, I support this work.

monnerat:
> So, what is the voting process. I read something about a thumbs up. Does that mean on any comment?

Alas, the voting instructions are poorly advertised, I'm afraid.

In document https://github.com/curl/curl/blob/master/docs/CONTRIBUTE.md, you can find the paragraph:
```
A pull request sent to the project might get labeled `needs-votes` by a
project maintainer. This label means that in addition to meeting all other
checks and qualifications this pull request must also receive more "votes" of
user support. More signs that people want this to happen. It could be in the
form of messages saying so, or thumbs-up reactions on GitHub.
```
Therefore I presume your comment above is a valid vote :-) Thanks for it !

vszakats:
@monnerat: I missed your PR, and you have my vote for this! Thanks for pointing to it.

bagder:
1. When performed over an authenticated channel, why consider them "unsafe"?
2. When `--safe-auth` is used, should it then not rather return an error if an unsafe auth method (over an unauthenticated protocol) is attempted?

monnerat:
> When performed over an authenticated channel, why consider them "unsafe"?

What are you calling an *authenticated channel"?
The connection we are dealing with is obviously not yet authenticated.
An encrypted (wether authenticated or not) proxy carrier connection may be considered as safe up to the proxy itself, but the final clear connection between the proxy and the server may be sniffed.

>When --safe-auth is used, should it then not rather return an error if an unsafe auth method (over an unauthenticated protocol) is attempted?

This is what will finally happen if no safe auth method can be selected.

bagder:
An *authenticated channel* would be using TLS, SSH or QUIC (without `--insecure`).

Your test 750 does a Basic auth for `http://`. It seems to strip off the auth, not fail the operation. I don't think a user would like that. I think it should rather fail as early as possible because it cannot perform the operation "safely".

bagder:
As written, this is not  a change we will merge in the near term.

Issue #8292, opened 2022-01-17
Cannot build curl 7.81.0 with --with-wolfssl


harrysarson:
### I did this

```
CURL_VERSION=curl-7.81.0

curl https://curl.se/download/$CURL_VERSION.tar.gz | tar -xz
curl -L https://github.com/wolfSSL/wolfssl/archive/refs/tags/v4.8.1-stable.tar.gz | tar -xz

mkdir wolfssl-install
mkdir wolfssl-dir

ls

pushd wolfssl-4.8.1-stable
    
autoreconf -fi
./configure --prefix=$(realpath ../wolfssl-install)  --enable-opensslextra
make -j$(nproc) 
make test 
make install

popd
pushd $CURL_VERSION

autoreconf -fi
./configure --with-wolfssl=$(realpath ../wolfssl-install) 
make -j$(nproc)

popd
```

### I expected the following

Curl can build with wolfssl (all the commands run successfully), as it does with `CURL_VERSION=curl-7.79.1`.  Instead it fails to build with error:

```
md5.c:85:10: fatal error: openssl/md5.h: No such file or directory
   85 | #include <openssl/md5.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
In file included from http_aws_sigv4.c:33:
curl_sha256.h:35:10: fatal error: openssl/sha.h: No such file or directory
   35 | #include <openssl/sha.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:2474: libcurl_la-md5.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from sha256.c:29:
curl_sha256.h:35:10: fatal error: openssl/sha.h: No such file or directory
   35 | #include <openssl/sha.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:2404: libcurl_la-http_aws_sigv4.lo] Error 1
make[2]: *** [Makefile:2621: libcurl_la-sha256.lo] Error 1
In file included from vauth/digest.c:38:
../lib/curl_sha256.h:35:10: fatal error: openssl/sha.h: No such file or directory
   35 | #include <openssl/sha.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

(Note `CURL_VERSION=curl-7.80.0` also fails).

### curl/libcurl version

7.81.0

### operating system

[...]

PR #8298, opened 2022-01-19
projects: Fix Visual Studio wolfSSL configurations

jay:
- Change build-wolfssl.bat to disable SSLv3, enable TLSv1.3, enable
  wolfSSL_DES_ecb_encrypt (needed by NTLM) and enable alt cert chains.

- Disable warning C4214 'bit field types other than int'.

- Add include directory wolfssl\wolfssl.

wolfSSL offers OpenSSL API compatibility that libcurl uses, and some
recent change in libcurl included an include file for wolfSSL like
openssl/foo.h, which has a path like wolfssl\wolfssl\openssl\foo.h.

The include directory issue was reported in #8292 but it's currently
unclear whether this type of change is needed for other build systems.

Bug: https://github.com/curl/curl/issues/8292
Reported-by: Harry Sarson

Closes #xxxx

---

This seems to be working with wolfSSL master but now I get these "identity not guaranteed" warnings:
~~~
..\..\..\..\lib\md5.c(533): warning C4232: nonstandard extension used : 'hmac_hinit' : address of dllimport 'wolfSSL_MD5_Init' is not static, identity not guaranteed
          X:\j\curl\wolfssl\wolfssl\openssl/md5.h(52) : see declaration of 'wolfSSL_MD5_Init'
..\..\..\..\lib\md5.c(535): warning C4232: nonstandard extension used : 'hmac_hupdate' : address of dllimport 'wolfSSL_MD5_Update' is not static, identity not guaranteed
          X:\j\curl\wolfssl\wolfssl\openssl/md5.h(53) : see declaration of 'wolfSSL_MD5_Update'
..\..\..\..\lib\md5.c(537): warning C4232: nonstandard extension used : 'hmac_hfinal' : address of dllimport 'wolfSSL_MD5_Final' is not static, identity not guaranteed
          X:\j\curl\wolfssl\wolfssl\openssl/md5.h(54) : see declaration of 'wolfSSL_MD5_Final'
..\..\..\..\lib\md5.c(550): warning C4232: nonstandard extension used : 'md5_init_func' : address of dllimport 'wolfSSL_MD5_Init' is not static, identity not guaranteed
          X:\j\curl\wolfssl\wolfssl\openssl/md5.h(52) : see declaration of 'wolfSSL_MD5_Init'
..\..\..\..\lib\md5.c(552): warning C4232: nonstandard extension used : 'md5_update_func' : address of dllimport 'wolfSSL_MD5_Update' is not static, identity not guaranteed
          X:\j\curl\wolfssl\wolfssl\openssl/md5.h(53) : see declaration of 'wolfSSL_MD5_Update'
..\..\..\..\lib\md5.c(554): warning C4232: nonstandard extension used : 'md5_final_func' : address of dllimport 'wolfSSL_MD5_Final' is not static, identity not guaranteed
          X:\j\curl\wolfssl\wolfssl\openssl/md5.h(54) : see declaration of 'wolfSSL_MD5_Final'
~~~

### DIFF
--- a/projects/Windows/VC10/lib/libcurl.tmpl
+++ b/projects/Windows/VC10/lib/libcurl.tmpl
@@ -895,13 +895,14 @@
     </Midl>
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>..\..\..\..\include;..\..\..\..\lib;..\..\..\..\..\wolfssl;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\..\..\..\include;..\..\..\..\lib;..\..\..\..\..\wolfssl;..\..\..\..\..\wolfssl\wolfssl;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>_DEBUG;BUILDING_LIBCURL;DEBUGBUILD;USE_WOLFSSL;USE_IPV6;WOLFSSL_DLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <MinimalRebuild>true</MinimalRebuild>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
       <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
       <WarningLevel>Level4</WarningLevel>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
+      <DisableSpecificWarnings>4214</DisableSpecificWarnings>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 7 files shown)


bagder:
It isn't actually what the #8292 bug is about though.

jay:
> It isn't actually what the #8292 bug is about though.

then what is it about? OP doesn't have wolfssl/openssl include directory included. When I build wolfSSL using Visual Studio the includes are in wolfssl-source-directory/wolfssl and openssl compatibility in wolfssl-source-directory/wolfssl/openssl which means if wolfssl-source-directory is the only directory curl is including then we'd have to use `#include <wolfssl/openssl/ssl.h>` (which actually is what we had done in wolfssl.c). The build systems would have to cover this to allow `#include <openssl/ssl.h>` unless it is already included by pkg-config for some build systems, but not others. I will follow up in the issue.

bagder:
> OP doesn't have wolfssl/openssl include directory included

I think you saw it already, but that's probably because that wolfSSL lacks `wolfSSL_DES_ecb_encrypt` and then the configure didn't add that include path.

jay:
> address of dllimport 'wolfSSL_MD5_Init' is not static, identity not guaranteed

I just landed this PR but I'll address the identity not guaranteed issue separately. It will probably involve wrapping calls to these DLL functions instead of taking a static address, similar to what is done in sha256

PR #8299, opened 2022-01-19
cmdline-opts/gen.pl: fix option matching to improve references

bagder:
Previously it could mistakenly match partial names when there are
options that start with the same prefix, leading to the wrong references
used.


bagder:
I ran into this bug with #8300

PR #8301, opened 2022-01-19
cmdline-opts/gen.pl: inhibit "added in" before 7.50

bagder:
Bump up the oldest version to consider for these texts from 7.30.0 to
7.50.0. 7.50.0 was released on Jul 21 2016.

This removes 35 mentions from the generated output.

### DIFF
--- a/docs/cmdline-opts/gen.pl
+++ b/docs/cmdline-opts/gen.pl
@@ -159,8 +159,8 @@ sub too_old {
     elsif($version =~ /^(\d+)\.(\d+)/) {
         $a = $1 * 1000 + $2 * 10;
     }
-    if($a < 7300) {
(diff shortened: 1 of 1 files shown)


dfandrich:
I suggest leaving N years of versions where N is the oldest version with
significant usage. I imagine that will end up being RHEL/RockyLinux with their
10 years of support. Some people may still be (legitimately) concerned with
compatibility with such an old version and would like to have that information
easily available.

bagder:
I was more thinking 5-6 years .The existing limit is at 7.30.0, released on `Apr 12 2013` so less than 9 years...

But you have a fair logic to the reasoning. I'll drop this now and revisit this idea in a few years! :grinning:

PR #8303, opened 2022-01-19
url: make Curl_disconnect return void

bagder:
1. The function would only ever return CURLE_OK anyway
 2. Only one caller actually used the return code
 3. Most callers did (void)Curl_disconnect()

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -830,8 +830,8 @@ static void conn_free(struct connectdata *conn)
  *
  */
 
-CURLcode Curl_disconnect(struct Curl_easy *data,
-                         struct connectdata *conn, bool dead_connection)
+void Curl_disconnect(struct Curl_easy *data,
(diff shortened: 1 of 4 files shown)


lib/url.c
```diff
@@ -851,7 +851,6 @@ CURLcode Curl_disconnect(struct Curl_easy *data,
    */
   if(CONN_INUSE(conn) && !dead_connection) {
     DEBUGF(infof(data, "Curl_disconnect when inuse: %zu", CONN_INUSE(conn)));
-    return CURLE_OK;
   }
```
jay, line 854: "Applying suggestions on deleted lines is not supported", ok github.

```suggestion
    return;
  }
```

jay (changes requested):
otherwise ok

lib/url.c
bagder: oops, good catch, yes indeed!

PR #8307, opened 2022-01-20
install impacket on wolfssl CI

pheiduck:
needed Python Package for wolfssl

### DIFF
--- a/.github/workflows/wolfssl.yml
+++ b/.github/workflows/wolfssl.yml
@@ -25,8 +25,10 @@ jobs:
           configure: --with-wolfssl=$HOME/wssl --enable-debug --enable-werror
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/wolfssl.yml
```diff
@@ -42,7 +42,13 @@ jobs:
 
     - run: ./buildconf && LDFLAGS="-Wl,-rpath,$HOME/wssl/lib" ./configure --enable-warnings --enable-werror ${{ matrix.build.configure }} && make V=1
       name: 'configure and build'
-
+      
+    - run: |
+       sudo apt-get install python-pip
```
bagder, line 47: shouldn't python-pip rather get added to get apt-get install already done above?

pheiduck: I see so we can skip this step

bagder (approved):

PR #8308, opened 2022-01-20
workflows/macos: install impacket

pheiduck:
needed Python Package for SMB tests (on macos ci as well)

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -122,6 +122,9 @@ jobs:
 
     - run: brew update && brew bundle install --no-lock --file /tmp/Brewfile
       name: 'brew install'
+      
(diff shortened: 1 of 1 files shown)


pheiduck:
I think we sholud add this at least to macos CI's otherwise the SMB tests are failing

bagder:
Where do they fail? The CI builds on macos are green...

pheiduck:
`== Contents of files in the log/ dir after test 1451
RUN: failed to start the SMB server
test 1451 SKIPPED: failed starting SMB server
 `
but could maybe a error inside the test 1451?

pheiduck:
the installation makes no difference to her, so anyway it still reports the package is not installed

PR #8313, opened 2022-01-21
workflows/mbedtls: install impacket

pheiduck:
needed Python Package for SMB tests


bagder:
Do we really need to add this to all CI builds? They don't seem to exercise *more* code, only the same code in different builds.

pheiduck:
> Do we really need to add this to all CI builds? They don't seem to exercise _more_ code, only the same code in different builds.

Yes, in fact it is relevant only in the wolfssl ci

PR #8314, opened 2022-01-21
curl --json

bagder:
`--json <data>` is a shortcut for `--data-binary <data> -H "Content-Type: application/json" -H "Accept: application/json"`

While the discussion continues on the curl-users list and in #8312, here's something to check and play with.

### DIFF
--- a/docs/cmdline-opts/Makefile.inc
+++ b/docs/cmdline-opts/Makefile.inc
@@ -5,7 +5,7 @@
 #                            | (__| |_| |  _ <| |___
 #                             \___|\___/|_| \_\_____|
 #
-# Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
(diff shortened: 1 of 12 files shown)


jay:
I'd call it --data-json for consistency with all the other --data-foo

bagder:
[A question](https://github.com/curl/curl/discussions/8312#discussioncomment-2032464) is if this should also set `Accept: application/json, */*`

bmarwell:
Can I still overwrite the accept header? Some restful endpoints might return a document, an image etc instead.

jzakrzewski:
Seems so. Just look at : https://github.com/curl/curl/pull/8314/files#diff-76a23840b36e3f9c5485019056c5ea0e3d195d6c55c9c1f5b2d3a16844674c0b around line 578

bagder:
Yes, exactly like we're used to. The test cases I added verify that both headers this option sets can be overridden.

bagder:
Ok, I'll merge this within 24 hours.

crazysouthern:
Chido

PR #8315, opened 2022-01-21
Fix #8292: allow building curl when wolfssl configured with --enable-opensslextra

### LINKED ISSUE #8292: Cannot build curl 7.81.0 with --with-wolfssl

harrysarson:
This MR puts all the `#include`s of openssl files behind wolfssl `#ifdef`s so that we can use the wolfssl/ prefixed include paths. 

Without these curl only builds when wolfssl is built with enable-all. Running CI against the first commit should fail showing why the fix is useful.

fixes #8292

### DIFF
--- a/.github/workflows/wolfssl.yml
+++ b/.github/workflows/wolfssl.yml
@@ -20,9 +20,14 @@ jobs:
       fail-fast: false
       matrix:
         build:
-        - name: wolfssl
+        - name: wolfssl (configured with --enable-all)
           install:
-          configure: --with-wolfssl=$HOME/wssl --enable-debug --enable-werror
+          wolfssl-configure: --enable-all
+          curl-configure: --with-wolfssl=$HOME/wssl --enable-debug --enable-werror
+        - name: wolfssl (configured with --enable-opensslextra)
+          install:
+          wolfssl-configure: --enable-opensslextra
+          curl-configure: --with-wolfssl=$HOME/wssl --enable-debug --enable-werror
 
     steps:
     - run: |
@@ -35,14 +40,14 @@ jobs:
        tar -xzf v5.0.0-stable.tar.gz
        cd wolfssl-5.0.0-stable
        ./autogen.sh
-       ./configure --enable-tls13 --enable-opensslextra --enable-harden --prefix=$HOME/wssl
+       ./configure --enable-tls13 ${{ matrix.build.wolfssl-configure }} --enable-harden --prefix=$HOME/wssl
        make && make install
 
       name: 'install wolfssl'
 
     - uses: actions/checkout@v2
 
-    - run: ./buildconf && LDFLAGS="-Wl,-rpath,$HOME/wssl/lib" ./configure --enable-warnings --enable-werror ${{ matrix.build.configure }} && make V=1
+    - run: ./buildconf && LDFLAGS="-Wl,-rpath,$HOME/wssl/lib" ./configure --enable-warnings --enable-werror ${{ matrix.build.curl-configure }} && make V=1
       name: 'configure and build'
 
     - run: make V=1 test-ci
--- a/.github/workflows/wolfssl.yml
+++ b/.github/workflows/wolfssl.yml
@@ -35,7 +35,7 @@ jobs:
        tar -xzf v5.0.0-stable.tar.gz
        cd wolfssl-5.0.0-stable
        ./autogen.sh
-       ./configure --enable-tls13 --enable-all --enable-harden --prefix=$HOME/wssl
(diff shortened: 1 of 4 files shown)


bagder:
> Without these curl only builds when wolfssl is built with enable-all

I don't think this is exactly true (you just needed more options than what you used and you didn't test other combos than `--enable-all`), but I also don't think it matters a lot.

.github/workflows/wolfssl.yml
```diff
@@ -35,7 +35,7 @@ jobs:
        tar -xzf v5.0.0-stable.tar.gz
        cd wolfssl-5.0.0-stable
        ./autogen.sh
-       ./configure --enable-tls13 --enable-all --enable-harden --prefix=$HOME/wssl
+       ./configure --enable-tls13 --enable-opensslextra --enable-harden --prefix=$HOME/wssl
```
bagder, line 38: I don't think you should remove the old `--enable-all` build but I think adding an `--enable-opensslextra` is a good ideas. Note that the second build is a *lesser* wolfSSL build and for example NTLM will not be enabled in your curl build using that.

bagder (changes requested):

.github/workflows/wolfssl.yml
harrysarson: I have had a go at splitting the wolfssl CI job in two to test the two configurations.

harrysarson:
> I don't think this is exactly true (you just needed more options than what you used and you didn't test other combos than --enable-all), but I also don't think it matters a lot.

Yep, good point. I guess the correct version of the statement is: this PR reduces the number of things you have to enable when configuring wolfssl in order to to build curl.

bagder (approved):

bagder:
Thanks!

harrysarson:
Thank you too!

PR #8316, opened 2022-01-21
Change qlog file extension to `.sqlog`

LPardue:
quiche has just switched it's qlog serialization format to JSON-SEQ by default . The spec says this SHOULD use `.sqlog` extension.

I believe ngtcp2 also supports JSON-SEQ by default as of https://github.com/ngtcp2/ngtcp2/commit/9baf06fc3f352a1d062b6953ae1de22cae30639d

Let's update curl so that tools know what format we are using!


bagder (approved):

bagder:
Thanks!

PR #8317, opened 2022-01-21
docs: update IETF links to use datatracker

LPardue:
The tools.ietf.org domain has been deprecated a while now, with the
links being redirected to datatracker.ietf.org.

Rather than make people eat that redirect time, this change switches the
URL to a more canonical source.


bagder:
Thanks!

PR #8320, opened 2022-01-22
url: revert the removal of trailing dot from host name

### LINKED ISSUE #8290: Absolute domain names get trailing dot stripped from host request header
curl doesn't honour the domain name part of the redirected URL if it is an absolute name; curl strips the trailing dot.  This causes problems loading some pages where the servers are configured to redirect non-absolute requests to the absolute ones:
```
$ curl -v -IL --max-redirs 1 http://pyropus.ca./
> HEAD / HTTP/1.1
> Host: pyropus.ca
> User-Agent: curl/7.74.0
> Accept: */*
> 
< Location: https://pyropus.ca./
< 
* Connection #0 to host pyropus.ca left intact
* Issue another request to this URL: 'https://pyropus.ca./'
*   Trying 96.126.125.117:443...
*  SSL certificate verify ok.
> HEAD / HTTP/1.1
> Host: pyropus.ca
> User-Agent: curl/7.74.0
> Accept: */*
> 
< Location: https://pyropus.ca./
```

... so that loops until max-redirects is hit.

The RFCs seem to say that the location value should be used as presented by the server.  It should only strip the dot from SNI, not from the Host: request header.

This has come up a couple of times before:
https://github.com/curl/curl/issues/716
https://github.com/curl/curl/issues/3022

... but those seem to be about requests to servers that accidentally (?) include the trailing dot, i.e. the servers are not expecting it.  The server itself isn't issuing a redirect to the absolute version of its domain name.  So accidental breakage in the cases listed shouldn't be a major issue, I think.  While the current behaviour does break sites that are using this feature deliberately.

As for other user agents:
* all GUI browers, to my knowledge, preserve the absolute domain name in the Host: header - I've tested Firefox, Chromium, Vivaldi, Konqueror, and Safari
* wget also preserves the absolute domain name on redirect URLs
* lynx, links, and elinks handle it like curl, erroring out on a redirect loop

I didn't find any other open/closed issues related to this, and the known bugs page doesn't have anything relevant.

### I did this
```
curl -v -IL --max-redirs 1 https://pyropus.ca/
```
### I expected the following

curl to resend the request with the Host: header value set to `pyropus.ca.` .

### curl/libcurl version
```
curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3
Release-Date: 2020-12-09
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli GSS-API HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets
```

### operating system

Linux <hostname> 5.14.16 #1 SMP Thu Nov 4 11:18:07 CST 2021 x86_64 GNU/Linux

Debian 11 / Bullseye

bagder:
Reverts 5de8d84098db1bd24e (May 2014, shipped in 7.37.0) and the
follow-up changes done afterward.

Keep the dot in names for everything except the SNI to make curl behave
more similar to current browsers. This means 'name' and 'name.' send the
same SNI for different 'Host:' headers.

Updated test 1322 accordingly

Fixes #8290
Reported-by: Charles Cazabon
Closes #

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -1574,20 +1574,6 @@ bool Curl_is_ASCII_name(const char *hostname)
 }
 
 /*
- * Strip single trailing dot in the hostname,
- * primarily for SNI and http host header.
- */
-static void strip_trailing_dot(struct hostname *host)
-{
-  size_t len;
-  if(!host || !host->name)
-    return;
-  len = strlen(host->name);
-  if(len && (host->name[len-1] == '.'))
-    host->name[len-1] = 0;
-}
-
-/*
  * Perform any necessary IDN conversion of hostname
  */
 CURLcode Curl_idnconvert_hostname(struct Curl_easy *data,
@@ -4052,17 +4038,6 @@ static CURLcode create_conn(struct Curl_easy *data,
    * Resolve the address of the server or proxy
    *************************************************************/
   result = resolve_server(data, conn, async);
-
-  /* Strip trailing dots. resolve_server copied the name. */
-  strip_trailing_dot(&conn->host);
-#ifndef CURL_DISABLE_PROXY
-  if(conn->bits.httpproxy)
-    strip_trailing_dot(&conn->http_proxy.host);
-  if(conn->bits.socksproxy)
-    strip_trailing_dot(&conn->socks_proxy.host);
-#endif
-  if(conn->bits.conn_to_host)
-    strip_trailing_dot(&conn->conn_to_host);
 
 out:
   return result;
--- a/lib/vtls/bearssl.c
+++ b/lib/vtls/bearssl.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 2019 - 2021, Michael Forney, <mforney@mforney.org>
+ * Copyright (C) 2019 - 2022, Michael Forney, <mforney@mforney.org>
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
@@ -464,6 +464,14 @@ static CURLcode bearssl_connect_step1(struct Curl_easy *data,
       return CURLE_PEER_FAILED_VERIFICATION;
     }
     hostname = NULL;
+  }
+  else {
+    char *snihost = Curl_ssl_snihost(data, hostname, NULL);
+    if(!snihost) {
+      failf(data, "Failed to set SNI");
+      return CURLE_SSL_CONNECT_ERROR;
+    }
+    hostname = snihost;
   }
 
   if(!br_ssl_client_reset(&backend->ctx, hostname, 0))
--- a/lib/vtls/gskit.c
+++ b/lib/vtls/gskit.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
@@ -830,8 +830,13 @@ static CURLcode gskit_connect_step1(struct Curl_easy *data,
 
   /* Process SNI. Ignore if not supported (on OS400 < V7R1). */
   if(sni) {
+    char *snihost = Curl_ssl_snihost(data, sni, NULL);
+    if(!snihost) {
+      failf(data, "Failed to set SNI");
+      return CURLE_SSL_CONNECT_ERROR;
+    }
     result = set_buffer(data, BACKEND->handle,
-                        GSK_SSL_EXTN_SERVERNAME_REQUEST, sni, TRUE);
+                        GSK_SSL_EXTN_SERVERNAME_REQUEST, snihost, TRUE);
     if(result == CURLE_UNSUPPORTED_PROTOCOL)
       result = CURLE_OK;
   }
--- a/lib/vtls/gtls.c
+++ b/lib/vtls/gtls.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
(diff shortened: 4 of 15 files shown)


jay:
> Keep the dot in names for everything except the SNI

I don't see that this PR does anything to strip the trailing dot from SNI. When I run `curld https://httpbin.org./` I get an assert with openssl backend and error 60 with others because the hostname doesn't match:

```
> curld https://httpbin.org./
Assertion failed: data->state.buffer[nlen-1] != '.', file ..\..\..\..\lib\vtls\openssl.c, line 3258

curl 7.82.0-DEV (i386-pc-win32) libcurl/7.82.0-DEV OpenSSL/1.1.1m nghttp2/1.46.0
Release-Date: [unreleased]
Features: AsynchDNS Debug HSTS HTTP2 HTTPS-proxy IPv6 Largefile NTLM SSL Unicode UnixSockets alt-svc

```
```
> curld https://httpbin.org./
curl: (60) schannel: SNI or certificate check failed: SEC_E_WRONG_PRINCIPAL (0x80090322) - The target principal name is incorrect.

curl 7.82.0-DEV (i386-pc-win32) libcurl/7.82.0-DEV Schannel WinIDN
Release-Date: [unreleased]
Features: AsynchDNS Debug HSTS IDN IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI Unicode UnixSockets alt-svc

```
```
> curld https://httpbin.org./
curl: (60)  subject alt name(s) or common name do not match "httpbin.org."
More details here: https://curl.se/docs/sslcerts.html

curl 7.82.0-DEV (x86_64-pc-win32) libcurl/7.82.0-DEV wolfSSL/5.1.1
Release-Date: [unreleased]
Features: AsynchDNS Debug HSTS IPv6 Largefile NTLM SSL Unicode UnixSockets alt-svc
```

Capture

bagder:
Another thing that struck me is that while we agree the SNI host is without the trailing dot, but how about the name we use to verify the server cert? At least one TLS library won't let us use different names for those two different uses. Does it mean we should use the SNI name there or the resolve host name?

[RFC 6125](https://datatracker.ietf.org/doc/html/rfc6125#section-6.4.1) does seem to say

bagder:
The SNI requirement leaves no room for other interpretations:

> "HostName" contains the fully qualified DNS hostname of the server,
> as understood by the client.  The hostname is represented as a byte
> string using ASCII encoding without a trailing dot. 

[RFC 6066 section 3](https://datatracker.ietf.org/doc/html/rfc6066#section-3)

jay:
Yes that part is clear. Are there any server certificates that have the trailing dot in the altname? I don't know of any.

bagder:
Do you know any HTTPS site that cares about the trailing dot in a Host: header? I don't.

jay:
I don't either but the reported issue looks legit.

bagder:
But yes, browsers seem to use the SNI name too (no dot) for checking the server cert.

krono:
@bagder (not nearly having as much experience, but maybe this is helpful:) 

At one point, I had to hand-implement the hostname-matching part for a programming language VM's SSL plugin and tired to make it as similar behaving as other clients I had at hand. So I toyed around: https://github.com/krono/x509hostmatch 

And to make that work, had to dot-drop both the cert-sAN and the incoming hostname…
https://github.com/krono/x509hostmatch/blob/master/x509hostmatch.c#L69

HTH

jay:
I think more changes will be needed for hostname verification. For some backends we call Curl_cert_hostcheck (which strips the trailing dot) and that should be fine. For others though we call a backend specific function which may or may not handle the trailing dot. Tests of the backends I build running `curld https://httpbin.org./`:

OpenSSL: Works. Done by Curl_cert_hostcheck.
Schannel in manual verify mode (ie ca bundle specified): Works. Done by Curl_cert_hostcheck.
Schannel in auto verify mode (ie no ca bundle, use os certs): Works. Done by Schannel's InitializeSecurityContext.
wolfSSL: Doesn't work. Done by wolfSSL's [wolfSSL_check_domain_name](https://www.wolfssl.com/doxygen/group__Setup.html#ga72e50573073c923af27afbdac38ed003).

https://github.com/curl/curl/blob/801bd5138ce31aa0d906fa4e2eabfc599d74e793/lib/vtls/wolfssl.c#L602-L607

---

Another thing, the commit `squashme: more sni conversions for schannel` adds two more SNI conversions for other InitializeSecurityContext calls. The problem is those calls are not necessarily done at the beginning, and currently Curl_ssl_snihost uses data->state.buffer to store the SNI temporarily, however it relies on data->state.buffer not actually being used for something else during this time. I don't know if a separate buffer is needed to avoid overwriting data->state.buffer.

https://github.com/curl/curl/blob/801bd5138ce31aa0d906fa4e2eabfc599d74e793/lib/urldata.h#L1326

bagder:
> The problem is those calls are not necessarily done at the beginning

Then I think the name needs to be copied from when it *is* called in the beginning. I mean, we don't necessarily have to make the function do that generically for all backends since SNI sort of implies it is done before the transfer starts (and the buffer is used for download data).

jay:
> Then I think the name needs to be copied from when it _is_ called in the beginning.

Ok redid the schannel sni code so that now Curl_ssl_snihost is only called once, right after schannel_acquire_credential_handle in step 1, and it copies the SNI hostname needed by the credential.

I still don't like how Curl_ssl_snihost uses the download buffer as the temporary space, I think that could lead to a bug one day. For example I have the sni hostname retrieved separately from schannel_acquire_credential_handle but it really is needed for any Schannel credential handle, and what if a credential handle needs to be created at some point other than step 1? Who is going to remember that Curl_ssl_snihost can only be called from step 1?

I suggest adding a dedicated buffer to an ssl struct, which shouldn't be that large for example as fqdn 255 max so `char snihost[255+1]`.

bagder:
> I suggest adding a dedicated buffer to an ssl struct, which shouldn't be that large for example as fqdn 255 max so char snihost[255+1].

I checked the spec. The SNI field can be 64K. I don't think it is always limited to the max DNS name size. Name resolving does not always mean DNS and there's no 255 byte limit in host names in URLs.

It feels wrong to me to "waste" such a large data field in a long-living struct for something that is only needed when setting up the connection and then isn't needed anymore. The download buffer has been used for this purpose for well over seven years now without problems, which could be an indicator that it might be okay.

bagder:
One additional detail struck me with the kept trailing dot. A connection `example.com.` (with a trailing dot) will not be reused/shared for `example.com` (without a trailing dot). But I also don't think we should change that.

jay:
> I checked the spec. The SNI field can be 64K. I don't think it is always limited to the max DNS name size. Name resolving does not always mean DNS and there's no 255 byte limit in host names in URLs.

Yes but currently and practically we are only sending the DNS hostname, which is all that is allowed.

> One additional detail struck me with the kept trailing dot. A connection to `example.com.` (with a trailing dot) will not be reused/shared for `example.com` (without a trailing dot). But I also don't think we should change that.

agree

bagder:
> Yes but currently and practically we are only sending the DNS hostname, which is all that is allowed.

curl doesn't have any such limit and can certainly use a longer name than what DNS supports. If you use something else than DNS to resolve (like `/etc/hosts` for example), the name can be longer for TLS. *Practically* that might of course be extremely rare. But you can bet that someone will run into that limit if we would impose it! :grinning:

bagder:
This causes a problem for two appveyor builds:
~~~
schannel.c(1035,16): warning C4189: 'hostname': local variable is initialized but not referenced 
~~~

bagder:
I'm getting ready to merge. How do you feel about this @jay ?

jay:
> 'hostname': local variable is initialized but not referenced

Sorry did not see this until after you merged. The debug statement should be left in, it is part of a series of debug messages schannel_connect_step{1,2,3} and so it is problematic to take out just step 2. I will open a PR to readd it, with a (void)hostname to rid of the warning for release builds.

bagder:
I figured they're "just debug messages" and as such they tend to linger around for longer than necessary and they're very easy to put back in there when you actually need them for debugging... but if you think it's better to keep it there, I'm fine with it.

PR #8326, opened 2022-01-25
[fix] minor typo ;)

pheiduck:
githuh should be github

### DIFF
--- a/RELEASE-NOTES
+++ b/RELEASE-NOTES
@@ -86,7 +86,7 @@ advice from friends like these:
   Fabian Yamaguchi, Filip Lundgren, Harry Sarson, Ikko Ashimine, Jan Ehrhardt,
   jhoyla on github, jonny112 on github, Lucas Pardue, Manfred Schwarb,
   Marcel Raad, Melroy van den Berg, Michał Antoniak, Niels Martignène,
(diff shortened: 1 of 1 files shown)


danielgustafsson:
Thats being pulled in from the Git log IIRC by the release notes scripts, and commit f558534e222f0bb6b8889acf3a441b97e4c16885 is credited like that:

```
commit f558534e222f0bb6b8889acf3a441b97e4c16885
Author: pheiduck on githuh <47042125+pheiduck@users.noreply.github.com>
Date:   Thu Jan 20 09:29:12 2022 +0100

    workflows/wolfssl: install impacket

    needed Python Package for SMB tests

    Closes #8307
```
When the release is prepared we can fix this after the last invocation of the release notes generation scripts.

bagder:
Yeah, and since it is now misspelled in the git history, we need to add the correction to `docs/THANKS-filter` to avoid it coming back to haunt us...

bagder:
I fixed it with e35637f193c259d2e4

danielgustafsson:
> On 25 Jan 2022, at 12:06, Daniel Stenberg ***@***.***> wrote:
> 
> ﻿
> I fixed it with e35637f
> 
Thats clever, nice one!

PR #8332, opened 2022-01-26
rtsp: fix memory leak on OOM

bagder:


bagder:
Wrong, uses saferealoc!

PR #8337, opened 2022-01-27
add missing shebang

pheiduck:
to some shellscripts identified by lift


bagder:
I prefer to let the VMS ones be for someone actually using VMS to take care of them. The iconv script is invoked by a CI job that doesn't need the shebang so why do we need to add it?

pheiduck:
> I prefer to let the VMS ones be for someone actually using VMS to take care of them. The iconv script is invoked by a CI job that doesn't need the shebang so why do we need to add it?

my mistake, I have not looked into it in depth, so not a high priority to add it

PR #8340, opened 2022-01-27
configure: set CURL_LIBRARY_PATH for nghttp2

bwncp:
To execute the test program, we might need the library path so that the lib is found at runtime.


bagder (changes requested):
Setting the variable seems like the right change indeed, but I think it should be done slightly further down, in the code block where `USE_NGHTTP2` is set. Because otherwise you might have it set before it is confirmed to be the right version.

bwncp:
> Setting the variable seems like the right change indeed, but I think it should be done slightly further down, in the code block where `USE_NGHTTP2` is set. Because otherwise you might have it set before it is confirmed to be the right version.

Thanks for your feedback. 

Fixed now.

bagder (approved):

bagder:
Thanks!

PR #8343, opened 2022-01-28
mbedtls: remove #include <mbedtls/certs.h>

MAntoniak:
mbedtls/certs.h file contains only certificates example (all definitions is beginning by mbedtls_test_*). None of them is used so we can avoid include the file.


bagder (approved):

bagder:
Thanks!

PR #8344, opened 2022-01-28
Possibility to use a mbedtls library without CRL support

MAntoniak:
Reduce the size of the ssl_backend_data structure.


bagder (approved):

bagder:
Thanks!

PR #8346, opened 2022-01-29
remove the unnecessary use of ternary operator

phyBrackets:
It's look like it return a bool value and it's obvious if the `p` is not null it is treated as true value and if null then treated as false. So, I don't think we need a ternary operator here.

### DIFF
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -199,7 +199,7 @@ static bool conncache_add_bundle(struct conncache *connc,
 {
   void *p = Curl_hash_add(&connc->hash, key, strlen(key), bundle);
 
-  return p?TRUE:FALSE;
+  return p;
 }
 
 static void conncache_remove_bundle(struct conncache *connc,


lib/conncache.c
```diff
@@ -199,7 +199,7 @@ static bool conncache_add_bundle(struct conncache *connc,
 {
   void *p = Curl_hash_add(&connc->hash, key, strlen(key), bundle);
 
-  return p?TRUE:FALSE;
+  return p;
```
jay, line 202: The function returns a bool value so it's not unnecessary.

phyBrackets: yeah right but still , we need that ?

jay: Yes, because p is a pointer.

bagder: shown [in this build](https://ci.appveyor.com/project/curlorg/curl/builds/42379683/job/2v795lcec49kf9g0)
~~~
2>.\conncache.c(202) : error C2440: 'return' : cannot convert from 'void *' to 'bool'

bagder:
I think maybe we can do like #8349 instead. It removes the ternary operator slightly differently.

jay:
Closing in favor of #8349.

Issue #8302, opened 2022-01-19
Curl Progress-Bar Issue


agamsol:
### I did this
```curl -#fkLo "%temp%\Source.txt" "https://pastebin.com/raw/HaAPSSR2"```

### I expected the following
it used to show the progress bar like it always did
`###################################################### 100/100%`

### Actual result
```##O=#    #```  (Gif Below)
![GIF](https://cdn.agamsol.xyz:90/media/cmd_dZzU1K0wBB.gif)

### curl/libcurl version

curl -V output:
- **Windows 11:**
    ```cmd
    C:\Users\Agam>curl -V
    curl 7.79.1 (Windows) libcurl/7.79.1 Schannel
    Release-Date: 2021-09-22
    Protocols: dict file ftp ftps http https imap imaps pop3 pop3s smtp smtps telnet tftp
    Features: AsynchDNS HSTS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI UnixSockets
    ```
- **WIndows 10:**
    ```bat
    C:\Users\Almo>curl -V
    curl 7.79.1 (Windows) libcurl/7.79.1 Schannel
    Release-Date: 2021-09-22
    Protocols: dict file ftp ftps http https imap imaps pop3 pop3s smtp smtps telnet tftp
    Features: AsynchDNS HSTS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI UnixSockets
    ```
### operating system
Command was tested on windows 10 and 11

bagder:
The progress bar was before curl 7.58.0 quite confused for contents without a known size, like the one you get in the example above. It could in fact not know when it reached a 100% before the actual end happened. It then used to output a character per data size, which could lead to excessive amounts of output on screen.

I changed that progress bar back in 2018, as can be seen in my [blog post](https://daniel.haxx.se/blog/2018/01/22/a-flying-curl-progress-bar/) about it.

agamsol:
I gave the command to 4 people and they attempted it

all the 4 had the exact same problem and it is not usual
can you attempt to run the command in your system and see?

bagder:
"The usual" has not been "the usual" since 2018

agamsol:
it is not how it acted 2 days ago

2 days ago it worked just fine on the same machine and it would print the expected output

bagder:
[...]

PR #8347, opened 2022-01-29
tool_cb_prg: show 100% progress bar for chunked and http2

jay:
- Update the total download size of chunked and unknown size http/2
  transfers when they complete successfully.

When the total download size is not known initially then a spaceship
type progress is used. Sometimes it may happen that the download size is
not known until the end, such as with chunked encoding or some HTTP/2
transfers. In that case it is useful to update the total download size
so that at the end the progress switches from the spaceship to the
normal progress bar to show 100% and signal the download was completed
successfully.

Before (condensed for example purpose):
 -=O=-          #     #    #     #

After (condensed for example purpose):
############################ 100.0%

Fixes https://github.com/curl/curl/issues/8302
Closes #xxxx

---

Not totally sure about the logic in http done

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -1654,6 +1654,19 @@ CURLcode Curl_http_done(struct Curl_easy *data,
     return CURLE_GOT_NOTHING;
   }
 
+  /* Set the total download size for when it is not known until end of transfer
+     (eg successful finish of chunked encoding or HTTP/2 stream). */
+  if(status == CURLE_OK && data->req.size == -1 &&
+     !(data->progress.flags & PGRS_DL_SIZE_KNOWN) &&
+     !premature && !conn->bits.retry && !data->set.connect_only &&
+     (
+#ifdef USE_NGHTTP2
+      (conn->httpversion == 20 && http->close_handled) ||
+#endif
+      data->req.chunk)) {
+    Curl_pgrsSetDownloadSize(data, data->req.bytecount);
+  }
+
   return CURLE_OK;
 }
 
--- a/docs/libcurl/opts/CURLINFO_CONTENT_LENGTH_DOWNLOAD.3
+++ b/docs/libcurl/opts/CURLINFO_CONTENT_LENGTH_DOWNLOAD.3
@@ -35,6 +35,11 @@ Pass a pointer to a double to receive the content-length of the download. This
 is the value read from the Content-Length: field. Since 7.19.4, this returns
 -1 if the size is not known.
 
+Some transfers do not have a Content-Length: and instead the server signals
+when the download is complete by using a termination marker (eg HTTP/2
+end-of-stream). In such a case the received content size is the content-length.
+(Added in 7.82.0)
+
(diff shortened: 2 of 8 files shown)


lib/http.c
```diff
@@ -1654,6 +1654,15 @@ CURLcode Curl_http_done(struct Curl_easy *data,
     return CURLE_GOT_NOTHING;
   }
 
+  if(status == CURLE_OK && data->req.size == -1 && !data->req.ignore_cl &&
+     (
+#ifdef USE_NGHTTP2
+      (conn->httpversion == 20 && http->close_handled) ||
+#endif
```
bagder, line 1661: This *probably* needs more attention to also work with hyper (which supports h2 without nghttp2) and HTTP/3, but I can't tell exactly how off the top of my head.

jay: I've somewhat improved on the logic and added to doc for CURLINFO_CONTENT_LENGTH_DOWNLOAD{,_T}. I don't know how it affects hyper. Basically I think total download size should only be set when the length can actually be determined. Notably I tried to exclude the read-until-close where there is no termination point other than close. One could argue that total download size should be set in all cases of successful transfer but having close as the only termination point seems iffy to me, for all we know it could be truncated.

bagder: Your test 1467 works with hyper too, so it looks good!

jay: > Your test 1467 works with hyper too, so it looks good!

That test is only for chunked encoding, so it does not cover the case of HTTP/2 and Hyper. I have some logic for nghttp2 if curl is handling http/2 but not hyper. I just checked and unfortunately the issue is still present when hyper the http backend:

~~~
owner@ubuntu1604-x64-vm:~/curl/tests$ ../src/curl -o /dev/null -# https://pastebin.com/raw/HaAPSSR2
##O=#  #                                                                       
owner@ubuntu1604-x64-vm:~/curl/tests$ ../src/curl -V
curl 7.82.0-DEV (x86_64-pc-linux-gnu) libcurl/7.82.0-DEV OpenSSL/1.1.1i zlib/1.2.8 nghttp2/1.43.0 librtmp/2.3 Hyper/0.14.13 OpenLDAP/2.4.42
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS Debug HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP TrackMemory UnixSockets
~~~

There needs to be some way to determine that the HTTP/2 transfer from hyper was fully downloaded (end of stream reached). I made it check http->close_handled for nghttp2 so perhaps hyper needs something similar.

bagder:
Can I propose we just close this as maybe not worth pursuing?

jay:
Sure. I think it's a good idea but maybe I went about it the wrong way. If there are any more reports of that issue an easier way would probably be to set the upload/download content length on any successful transfer.

PR #8351, opened 2022-01-29
vtls: remove support for NSS

bagder:


jay:
Ref: https://curl.se/mail/lib-2022-01/0120.html

/cc @kdudka

bagder:
Mentioned on [Twitter](https://twitter.com/bagder/status/1486964475261603841) and the [curl-library mailing list](https://curl.se/mail/lib-2022-01/0120.html) on Jan 28 2022. So far not one person have spoken up in favor of *keeping* the support.

kdudka:
I find the argument for proactive removal rather weak but will not fight myself for keeping the NSS support in curl upstream.  Red Hat links libcurl with NSS only in RHEL-7 now and we do not plan any big changes there.  On the other hand, curl user surveys do not show that other upstream consumers went off NSS already.

danielgustafsson:
We carry quite a lot of code that we don't know the status of, and which is likely to be even more obscure than this (Novell Netware support comes to mind), so I'm not sure why NSS support meets a higher bar. That being said, I agree with @kdudka with not caring enough to put up a fight to keep it in given the (probably) small userbase.

bagder:
Most of the unknown-status code in curl is stuff that people actually have to struggle a bit to get going, because of the natural bitrot over time for things we don't repeatedly build and test. And usually one-of-a-kind things.

I think TLS backends are different. If they are to remain "blessed" by us, we can afford to put requirements and demands on those - partly because there are many good supported alternatives around. Each TLS backend is putting on a lot of more of "a maintenance burden" on the greater curl development community that a few random #ifdefs for rare platforms do (in my personal and subjective opinion).

I don't think there's any alarming or dangerous situation for NSS right now so I won't insist on merging this PR of is there's a notable resistance. I do however see the upstream project making it even harder for us to use it while at the same time there is virtually nobody these days that actually build curl with NSS (who couldn't switch to another backend). It is a question of balancing the work vs the benefits.

danielgustafsson:
> Most of the unknown-status code in curl is stuff that people actually have to struggle a bit to get going,

Having spent quality time with the NSS toolchain lately, I'm not sure that the same doesn't apply here =)
 
>  It is a question of balancing the work vs the benefits.

Despite liking NSS, I wholeheartedly agree.

dfandrich:
Rather than suddenly removing support without warning, how about renaming
--with-nss in the next release to
--with-nss-but-this-is-the-final-release-unless-you-speak-up?  That will
greatly increase the chance that we'll hear about NSS users while only mildly
inconveniencing them if there are any. And if there aren't any, remove it for
the release after that.

bagder:
> how about renaming

Cool idea, I'll try that

danielgustafsson:
Maybe we could add a `--deprecated` option to autoconf which would be required to configure support we are thinking about dropping?

sfc-gh-emusser:
Hi @bagder, sorry we just became aware of this. We are actually using the NSS support heavily. We would, at least, like some time to evaluate other options if NSS support isn't going to be permanent and could help maintain it going further. Very much appreciate any consideration here.

danielgustafsson:
Are you using NSS because of an NSS specific feature that you want/need/like?; because of licensing?; because of something wildly different? Any information you can share helps guide this decision.

sfc-gh-emusser:
Hi @danielgustafsson , sorry for the delay in responding. We don't know the specifics of why NSS was originally chosen but fips support and performance (both Arm and x86) are both critical to our application. Even before this PR was raised it was pretty evident to us that there must not be many others using the curl / nss combination (at least in manner we are) so we had already (just recently) been considering moving off of NSS. We are now evaluating other choices including openssl 3 (though we'd need to wait until it is fips certified). We'd really appreciate the curl team not completely removing nss support -- perhaps only deprecating it -- for at least six months.

Thanks,
Eric

bagder:
The removal will not happen before the release of April 27 2022 no matter what.

bagder:
Okay, let's officially push this a little more. We will remove NSS **at the earliest after August 2022**. Therefore, I will close this PR and make a new one later. I will make a separate PR that will change configure to require an additional option to build with NSS just to make users more aware of these plans.

PR #8352, opened 2022-01-30
docs/TheArtOfHttpScripting: fix example URL

neutric:
The GET form post example URL is given as `http://www.example.com/when/junk.cgi`. The analogous POST form example uses URL `http://www.example.com/when.cgi` where it should use `http://www.example.com/when/junk.cgi` as before.


jay:
Thanks

PR #8358, opened 2022-01-31
netware: remove support

bagder:
There are no current users and no Netware related changes have been done
for over 13 years is a clear sign.


danielgustafsson (approved):
Not that I know Netware (apart from admining 3.x servers about 1000 years ago), it seems complete and non-netware builds pass.

PR #8362, opened 2022-01-31
vxworks: drop support

bagder:
No changes or fixes in vxworks related code since 2009 leads me to
believe that this doesn't work anymore.


bagder:
I also asked [on the mailing list](https://curl.se/mail/lib-2022-01/0133.html) for increased visibility

PR #8365, opened 2022-01-31
curl/docs/libcurl/*: fix some formatting of man pages

samueloph:
Details:

 From "mandoc -Tlint":

mandoc: curl_getdate.3:64:2: WARNING: skipping paragraph macro: PP empty
mandoc: curl_global_init_mem.3:56:2: ERROR: skipping end of block that is not open: RE
mandoc: curl_url_cleanup.3:29:2: STYLE: fill mode already enabled, skipping: fi
mandoc: curl_url_dup.3:29:2: STYLE: fill mode already enabled, skipping: fi
mandoc: curl_url_set.3:32:2: STYLE: fill mode already enabled, skipping: fi

  From "test-groff -b -mandoc -T utf8 -rF0 -t -w w -z":

  [ "test-groff" is a developmental version of "groff" ]

troff: <curl_getdate.3>:108: warning: trailing space
troff: <curl_getdate.3>:109: warning: trailing space

Patch originally submitted to Debian at:
https://bugs.debian.org/963559

### DIFF
--- a/docs/libcurl/curl_getdate.3
+++ b/docs/libcurl/curl_getdate.3
@@ -62,7 +62,6 @@ first three letters. This is usually not info that adds anything.
 If a decimal number of the form YYYYMMDD appears, then YYYY is read as the
 year, MM as the month number and DD as the day of the month, for the specified
 calendar date.
-.PP
 .SH EXAMPLE
 .nf
  time_t t;
@@ -110,5 +109,5 @@ On systems with 64 bit time_t: if the year is less than 1583, this function
 will return -1. (The Gregorian calendar was first introduced 1582 so no "real"
 dates in this way of doing dates existed before then.)
 .SH "SEE ALSO"
-.BR curl_easy_escape "(3), " curl_easy_unescape "(3), "
-.BR CURLOPT_TIMECONDITION "(3), " CURLOPT_TIMEVALUE "(3) "
+.BR curl_easy_escape "(3), " curl_easy_unescape "(3),"
+.BR CURLOPT_TIMECONDITION "(3), " CURLOPT_TIMEVALUE "(3)"
--- a/docs/libcurl/curl_url_cleanup.3
+++ b/docs/libcurl/curl_url_cleanup.3
@@ -27,7 +27,7 @@ curl_url_cleanup - free a CURLU handle
 #include <curl/curl.h>
 
 void curl_url_cleanup(CURLU *handle);
-.fi
+.
 .SH DESCRIPTION
 Frees all the resources associated with the given CURLU handle!
 .SH EXAMPLE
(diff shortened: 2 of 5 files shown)


docs/libcurl/curl_getdate.3
```diff
@@ -110,5 +109,5 @@ On systems with 64 bit time_t: if the year is less than 1583, this function
 will return -1. (The Gregorian calendar was first introduced 1582 so no "real"
 dates in this way of doing dates existed before then.)
 .SH "SEE ALSO"
-.BR curl_easy_escape "(3), " curl_easy_unescape "(3), "
-.BR CURLOPT_TIMECONDITION "(3), " CURLOPT_TIMEVALUE "(3) "
+.BR curl_easy_escape "(3), " curl_easy_unescape "(3),"
+.BR CURLOPT_TIMECONDITION "(3), " CURLOPT_TIMEVALUE "(3)"
```
bagder, line 113: This one puzzles me.
1. `mandoc -Tlint curl_getdate.3` on my debian unstable does not warn on this for me
2. I count a total of **79** similar lines, only counting `.3` files in the same directory. Shouldn't then all of them be fixed at the same time?

docs/libcurl/curl_url_cleanup.3
```diff
@@ -27,7 +27,7 @@ curl_url_cleanup - free a CURLU handle
 #include <curl/curl.h>
 
 void curl_url_cleanup(CURLU *handle);
-.fi
+.
```
bagder, line 30: 1. My mandoc doesn't complain on this
2. This pattern is used in all libcurl function man pages, why are these three singled out?
3. Why leave a dot?

bagder (commented):
Thanks, but this left me really curious...

docs/libcurl/curl_getdate.3
samueloph: This one is not from mandoc, but from test-groff, which is not something I have installed but it makes sense I guess, the finding (from the submitter of the patch) was:
```
From "test-groff -b -mandoc -T utf8 -rF0 -t -w w -z":

[ "test-groff" is a developmental version of "groff" ]

troff: <curl_getdate.3>:108: warning: trailing space
troff: <curl_getdate.3>:109: warning: trailing space
```

There's a good chance it can be ignored if you think it's fine (it might be too cosmetic to spend the time to fix across all the files).

bagder: It feels very arbitrary to only fix it in a single file. If we consider this a problem then we should fix it everywhere. If it isn't a problem, then we don't need to fix any file. Why is it a problem? I don't spot any difference at all in the generated result I get with `man -l curl_getdate.3`.

docs/libcurl/curl_url_cleanup.3
samueloph: Hmm, I also don´t get these anymore, apparently mandoc stopped complaining about it sometime ago.

This hunk can be dropped.

I don't know why the original submitter decided to go with a dot, I wasn't the original reviewer of the patch either so I just kept refreshing it. I figured sending it upstream was a good opportunity to drop changes like this.

docs/libcurl/curl_getdate.3
samueloph: Let's drop it then.

I believe this means there's only a single legitimate change to the whole CR, which is for:
`mandoc: curl_getdate.3:64:2: WARNING: skipping paragraph macro: PP empty`

bagder:
Thanks!

PR #8366, opened 2022-01-31
configure: rename --with-nss to --with-nss-maybe-last-release

bagder:
... and output a warning in the configure output with a reference to #8351


bagder:
Let's postpone this one.

PR #8370, opened 2022-02-01
Update urlapi.c

HenrikHolst:
Changed strlen+memcpy into a single strcpy call since in urlapi.c since that is what the code did in practice.


bagder (commented):
That space before the open parentheses breaks the code style, but otherwise this is fine!

bagder:
`make checksrc` in your build tree will also show this.

HenrikHolst:
Yeah sorry about this, did the edit directly on GitHub due to inexperience with git

HenrikHolst:
So the checks are failing left and right, which I suppose is due to the code style? Being a noob here, should I abandon this PR and file a new one or is it possible to somehow edit this one?

bagder:
The smoothest way is to either:
1. fix it, commit again and push that to the same branch (ending up with two commits that we squash on merge) or:
2. amend the existing commit, and force-push to the same branch (sticking to a single commit)

jay:
Thanks

PR #8374, opened 2022-02-02
Threaded resolver fix in configure for AIX xlC compiler

dcassioli:
Appending the correct flag to CFLAGS to enable threaded resolver when building for AIX with IBM xlC compiler.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3472,14 +3472,23 @@ if test "$want_pthreads" != "no"; then
          dnl it doesn't actually work without -lpthread
          USE_THREADS_POSIX=""
          ;;
+      *-ibm-aix*)
+         dnl Check if compiler is xlC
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

asedeno:
This did not preserve behavior for non-AIX xIC compilers. :-(

dcassioli:
> This did not preserve behavior for non-AIX xIC compilers. :-(

We could remove the AIX OS case and put the compiler check in the default case.
@bagder would this be acceptable?

asedeno:
@dcassioli can you test #8542 for your use case? I think that'll work for everyone, but I have no system to test against.

asedeno:
@dcassioli Great, let us know how things go on #8542. Thanks.

PR #8375, opened 2022-02-02
configure: support specification of a nghttp2 library path

bwncp:
This enables using `--with-nghttp2=<dir>` on systems without `pkg-config`.


bagder (approved):

bagder:
Thanks!

PR #8377, opened 2022-02-03
Update setopt.c

HenrikHolst:
Curl_setstropt() allocated memory for the string before checking if the string was within bounds. IMHO the bounds check should be done first.


bagder (approved):

bagder:
Thanks!

PR #8378, opened 2022-02-03
TPF: drop support

bagder:
There has been no TPF related changes done since September 2010 (commit
7e1a45e224e57) and since this is a platform that is relatively different
than many others (== needs attention), I draw the conclusion that this
build is broken since a long time.


monnerat:
If you do that, this is also an open door for all the `CURL_DOES_CONVERSIONS` stuff removal, as it seems this is the only platform using it.
Just my 2 cents!

bagder:
Oh wow, thanks for highlighting this fact. I think that's actually also another strong reason to suspect that the TPF build doesn't work anymore as the conversion stuff is fairly complicated and easy to mess up.

Also, a fairly big gain to be able to rip that out as this define complicates the logic in many places.

monnerat:
Big slashing, isn't it ! :-)

bagder:
Yes, that is a significant chunk. There might even be some more to fix after this when some flows have been simplified.

lib/setopt.c
```diff
@@ -1644,22 +1644,9 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
...
-     * "Convert to network encoding" callback
-     */
-    data->set.convtonetwork = va_arg(param, curl_conv_callback);
-    break;
   case CURLOPT_CONV_FROM_UTF8_FUNCTION:
-    /*
-     * "Convert from UTF-8 encoding" callback
-     */
-    data->set.convfromutf8 = va_arg(param, curl_conv_callback);
+    /* removed */
```
jay, line 1649: I think it would be better to remove them entirely (CURLE_UNKNOWN_OPTION), but if not you'll need /* FALLTHROUGH */

bagder: yeah, might just as well remove them completely

PR #8379, opened 2022-02-03
runtests.pl: tolerate test directories without Makefile.inc

fabiankeil:
Silences the following warnings when using a Makefile.inc-free
TESTDIR using the "-o" argument:

   readline() on closed filehandle D at ./runtests.pl line 592.
   Use of uninitialized value $disttests in pattern match (m//) at ./runtests.pl line 3602.


bagder (approved):

jay:
Thanks

PR #8380, opened 2022-02-03
vtls: Add a new option to set the SSL session cache size

mkauf:
Add an option to set the SSL session cache size for easy handles and for share
handles.

The SSL session cache size can only be set before the cache is created.
For easy handles, it can be changed before the first transfer is started.
For share handles, it can be changed before the SSL session cache is enabled
on the share handle.


bagder:
I'm not against this idea, but this feels like an option that a typical user will have a really hard time to know when to use and if then, what to set the value to.

What made you want to change this value? How do you determine the new value to use? Could just changing the internal defaults be an option?

mkauf:
We use libcurl in a product (reverse proxy) that loads resources from back-end servers. The number of these back-end servers is known, but our customers have different setups. Some have only 1-3 back-end servers, most have 10-30, and probably there are also customers with > 200 back-end servers. With the new libcurl setting, it's possible to tune the size of libcurl's SSL session cache for optimal performance, so that there is one cache entry per back-end server.

If there are too few cache entries (less than the number of back-ends), the performance depends on the sequence in which the back-ends are accessed. In the worst case, the back-ends are accessed in a round-robin way, and a full SSL handshake is necessary for every request.

I agree that most libcurl users probably won't need to change this value. That's fine - if the defaults work, just use the defaults.

bagder:
Can we repurpose `CURLOPT_SSL_SESSIONID_CACHE` for this?

This option is currently used for and documented as setting the cache on/off by setting the long to zero or one. Internally, it treats anything non-zero as one.

If we would treat any value larger than 1 to mean both enable and the new size, we could maybe avoid adding a new option? It would have to keep treating the value `1` as (enable default-size), which is a little weird but maybe a smaller price than an entirely new option?

mkauf:
> Can we repurpose ``CURLOPT_SSL_SESSIONID_CACHE`` for this?

This would work for easy handles, e.g. ``curl_easy_setopt(curl, CURLOPT_SSL_SESSIONID_CACHE, 50L);``. For share handles, this is no solution, because the SSL session cache is enabled in a different way: ``curl_share_setopt(share, CURLSHOPT_SHARE, CURL_LOCK_DATA_SSL_SESSION);``.

I prefer new options for new features, that's more backwards compatible, easier to use and it works with share handles too. But if you prefer the "repurpose" proposal, I can change the pull request accordingly.

jay:
I don't prefer the repurpose, I think it is confusing that 1 would not mean set to size 1

mkauf:
How do we proceed here? Are there open questions? It's probably a "niche feature", but it may be worth implementing.

I can rebase the commits if that helps for a review.

danielgustafsson:
> I don't prefer the repurpose, I think it is confusing that 1 would not mean set to size 1

I agree with @jay, since the current option takes a numeric value I think it would be unfortunate to repurpose it with 1 being a magic value.

docs/libcurl/opts/CURLSHOPT_SHARE.3
```diff
@@ -48,8 +48,9 @@ option.
 SSL session IDs will be shared across the easy handles using this shared
 object. This will reduce the time spent in the SSL handshake when reconnecting
 to the same server. Note SSL session IDs are reused within the same easy
-handle by default. Note this symbol was added in 7.10.3 but was not
-implemented until 7.23.0.
+handle by default. The size of the cache can be changed with
+\fICURLSHOPT_SSL_SESSIONID_CACHE_SIZE(3)\fP. Note this symbol was added in
+7.10.3 but was not implemented until 7.23.0.
```
danielgustafsson, line 53: This change makes it seem like `CURLSHOPT_SSL_SESSIONID_CACHE_SIZE` was added in 7.10.3. I think we need to reword this.

docs/libcurl/opts/CURLSHOPT_SHARE.3
```diff
@@ -80,7 +81,7 @@ multi handle will share PSL cache by default without using this option.
 All
 .SH EXAMPLE
 .nf
-  CURLSHcode sh
+  CURLSHcode sh;
```
danielgustafsson, line 84: This is an unrelated fix which, albeit tiny, should go in separately.

lib/vtls/vtls.c
```diff
@@ -737,9 +737,9 @@ CURLcode Curl_ssl_initsessions(struct Curl_easy *data, size_t amount)
     return CURLE_OUT_OF_MEMORY;
 
   /* store the info in the SSL section */
-  data->set.general_ssl.max_ssl_sessions = amount;
+  data->state.session_size = amount;
   data->state.session = session;
-  data->state.sessionage = 1; /* this is brand new */
+  data->state.sessionage = 0; /* this is brand new */
```
danielgustafsson, line 742: I don't follow, why this change?

lib/share.c
```diff
@@ -39,6 +39,7 @@ curl_share_init(void)
   if(share) {
     share->magic = CURL_GOOD_SHARE;
     share->specifier |= (1<<CURL_LOCK_DATA_SHARE);
+    share->max_ssl_sessions = 8;
```
danielgustafsson, line 42: It's not the fault of this patch, but `8` should be defined with a name or at the very least be given a comment explaining the magic number.

docs/libcurl/opts/CURLSHOPT_SSL_SESSIONID_CACHE_SIZE.3
```diff
@@ -0,0 +1,61 @@
...
+SSL session-ID cache of a share handle is enabled by calling
+curl_share_setopt(share, CURLSHOPT_SHARE, CURL_LOCK_DATA_SSL_SESSION).
+.SH DEFAULT
+8
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+  CURLSHcode sh;
+  struct secrets private_stuff;
```
danielgustafsson, line 44: This line can be removed.

docs/libcurl/opts/CURLSHOPT_SHARE.3
danielgustafsson: Addressed in #8697

mkauf: Ok, thanks!

lib/vtls/vtls.c
mkauf: In ``curl_share_setopt()`` (share.c), the initialization goes like this:
```
share->sessionage = 0;
```
I wanted to use the same value in vtls.c and in share.c. Alternatively share.c could be changed:
```
share->sessionage = 1;
```

docs/libcurl/opts/CURLSHOPT_SHARE.3
mkauf: Thanks, I have swapped the sentences.

lib/share.c
mkauf: I have added a comment that this is the default cache size.

docs/libcurl/opts/CURLSHOPT_SSL_SESSIONID_CACHE_SIZE.3
mkauf: Done.

mkauf:
I have updated the pull request.

docs/libcurl/opts/CURLSHOPT_SSL_SESSIONID_CACHE_SIZE.3
```diff
@@ -0,0 +1,60 @@
...
+CURLSHcode curl_share_setopt(CURLSH *share, CURLSHOPT_SSL_SESSIONID_CACHE_SIZE,
+                             long size);
+
+.fi
+.SH DESCRIPTION
+Set the size of the SSL session-ID cache. The size can only be set before the
+SSL session-ID cache of a share handle is enabled by calling
+curl_share_setopt(share, CURLSHOPT_SHARE, CURL_LOCK_DATA_SSL_SESSION).
+.SH DEFAULT
+8
```
bagder, line 38: I believe this is 8 for shared handles, but 5 by default for an easy handle.

bagder:
Sorry, but I missed to merge this during this feature window. Please feel free to remind me in the next one.

Question: when you use a larger ssl session cache, do you also increase `CURLOPT_MAXCONNECTS` ? It seems like maybe in a common case a user would like them both at basically the same size?

docs/libcurl/opts/CURLSHOPT_SSL_SESSIONID_CACHE_SIZE.3
mkauf: That's correct. The default value 5 for an easy handle is on the corresponding manual page `CURLOPT_SSL_SESSIONID_CACHE_SIZE.3`.

mkauf:
> Question: when you use a larger ssl session cache, do you also increase `CURLOPT_MAXCONNECTS` ? It seems like maybe in a common case a user would like them both at basically the same size?

We increase it to 32 (not the same size as the SSL session cache). In our use case, libcurl works on the local network and creating a new connection is quite fast. In this scenario, reusing SSL sessions is much more important than reusing connections.

Maybe the manual pages for the new libcurl options should mention `CURLOPT_MAXCONNECTS` and `CURLMOPT_MAXCONNECTS`.

bagder:
Still, it seems to me that we could think of a way to fix this (growing the session id cache) without adding a new option. That would then automatically benefit everyone.

PR #8383, opened 2022-02-04
smb: passing a socket for writing and reading

MAntoniak:
I am not familiar with the SMB protocol, but the second parameter for the curl_read and curl_write functions is socket, so give them the correct socket.


bagder (approved):

bagder:
Thanks!

PR #8389, opened 2022-02-04
h2/h3: provide and refer to psuedo headers as defines

bagder:
... and do strlen() on the defines to allow the compiler a chance to
optimize away the strlens.

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -64,6 +64,11 @@
 #define H2BUGF(x) do { } while(0)
 #endif
 
+#define H2_PSEUDO_METHOD ":method"
+#define H2_PSEUDO_SCHEME ":scheme"
+#define H2_PSEUDO_AUTHORITY ":authority"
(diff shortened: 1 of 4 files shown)


bagder:
I should probably change them to sizeof() - 1 to make sure all compilers use a constant there...

lib/http2.c
```diff
@@ -64,6 +64,11 @@
 #define H2BUGF(x) do { } while(0)
 #endif
 
+#define H2_PSUEDO_METHOD ":method"
+#define H2_PSUEDO_SCHEME ":scheme"
+#define H2_PSUEDO_AUTHORITY ":authority"
+#define H2_PSUEDO_PATH ":path"
+#define H2_PSUEDO_STATUS ":status"
```
jay, line 71: s/psuedo/pseudo

bagder: oops!

PR #8394, opened 2022-02-06
curl-openssl: fix SRP check for OpenSSL 3.0

MarcelRaad:
When OpenSSL 3.0 is built with `--api=3.0` and `no-deprecated`, the SRP
functions exist in the library, but are disabled for user code. Check
if they are actually usable instead of only if they exist. Also check
for the functions actually required for TLS-SRP.

TLS-SRP support is still enabled if OpenSSL is configured with just
`--api=3.0` or with `--api=1.1.1 no-deprecated`.

(I'm not very familiar with autoconf and m4 - please let me know if there's a better way to do that.)

### DIFF
--- a/m4/curl-openssl.m4
+++ b/m4/curl-openssl.m4
@@ -384,11 +384,21 @@ dnl ---
 dnl We require OpenSSL with SRP support.
 dnl ---
 if test "$OPENSSL_ENABLED" = "1"; then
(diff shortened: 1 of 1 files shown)


bagder (approved):
Looks fine to me!

MarcelRaad:
Thanks for the review!

PR #8395, opened 2022-02-07
deprecate NSS support

bagder:
Add deprecation plans to docs/DEPRECATE.md

### DIFF
--- a/.github/workflows/nss.yml
+++ b/.github/workflows/nss.yml
@@ -22,7 +22,7 @@ jobs:
         build:
         - name: NSS
           install:
-          configure: --with-nss --enable-debug --enable-werror
+          configure: --with-nss --enable-debug --enable-werror --with-nss-deprecated
 
     steps:
     - run: |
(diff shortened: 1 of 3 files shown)


danielgustafsson:
Another option would be to add a `--enable-deprecated` flag and then require that along with `--with-nss` and other deprecated options. An opt-in to explicitly have to say "I know what I'm getting myself in to" and then we can use that same mechanism for gatekeeping other features on the way out?

bagder:
That's not a bad idea, but I have to say that I like that the way I did it here needs the user to acknowledge *this* exact feature being deprecated and not accepting deprecation *in general*.

jay:
> _this_ exact feature being deprecated and not accepting deprecation _in general_.

Well I'm with you on that but --with-nss-aware doesn't say much to an admin glancing through their company's shell scripts. --DEPRECATED-with-nss or something

bagder:
The flag name needs to start with `--with` for the configure option thing to work. `--with-deprecated-nss` perhaps?

danielgustafsson:
> The flag name needs to start with --with for the configure option thing to work. --with-deprecated-nss perhaps?

I vote for --with-deprecated-nss or --with-nss-deprecated.

jay:
I agree, something with deprecated in the name.

bagder:
Uses `--with-nss-deprecated` now!

PR #8396, opened 2022-02-07
CI: move the NSS job from zuul to GHA

bagder:


bagder:
Clearly unit tests did not build and run properly with NSS before...

PR #8401, opened 2022-02-08
h2h3: pass correct argument types to infof()

bagder:
Detected by Coverity. CID 1497993


bagder:
I'm not sure we should keep this sent without debug, but this is worth fixing anyway!

PR #8402, opened 2022-02-08
maketgz: return error if 'make dist' fails

bagder:
To better detect this problem in CI jobs


MarcelRaad (approved):
failing as intended 👍

bagder:
Yeps, and now rebase it on top of the fix and see that it goes green!

PR #8403, opened 2022-02-08
lib/Makefile: remove config-tpf.h from the dist

bagder:
Follow-up from da15443dddea2bfb. Missed before because the 'distcheck'
CI job was not working as intended.

Ref: https://curl.se/mail/lib-2022-02/0070.html


MarcelRaad (approved):

MarcelRaad:
Thanks, that was fast!

PR #8404, opened 2022-02-08
openldap: pass string length arguments to client_write()

monnerat:
This is an `strlen()` optimization in the same way as commit 4028892f14669937cb47d8a165e059503107812a does.


bagder (approved):

bagder:
Thanks!

monnerat:
Thanks for pulling.

PR #8408, opened 2022-02-08
scripts/cijobs.pl: output data about all currect CI jobs

bagder:
This script parses the config files for all the CI services currently in
use and output the information in a uniform way. The idea is that the
output from this script should be possible to massage into informational
tables or graphs to help us visualize what they are all testing and NOT
testing.


bagder:
[current output from the script](https://github.com/curl/curl/files/8027926/ci.txt)

PR #7801, opened 2021-09-30
lib/mk-ca-bundle.pl: skip certs passed Not Valid After date

bagder:
With this change applied, the now expired 'DST Root CA X3' cert will no
longer be included in the output.

Details: https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/

PR #8411, opened 2022-02-09
mk-ca-bundle.pl: Use stricter logic to process the certificates

jay:
.. and bump version to 1.29.

This change makes the script properly ignore unknown blocks and
otherwise fail when Mozilla changes the certdata format in ways we
don't expect. Though this is less flexible behavior it makes it far less
likely that an invalid certificate can slip through.

Prior to this change the state machine did not always properly reset,
and it was possible that a certificate marked as invalid could then
later be marked as valid when there was conflicting trust info or
an unknown block was erroneously processed as part of the certificate.

Ref: https://github.com/curl/curl/pull/7801#pullrequestreview-768384569

Closes #xxxx

### DIFF
--- a/lib/mk-ca-bundle.pl
+++ b/lib/mk-ca-bundle.pl
@@ -6,7 +6,7 @@
 # *                            | (__| |_| |  _ <| |___
 # *                             \___|\___/|_| \_\_____|
 # *
-# * Copyright (C) 1998 - 2021, Daniel Stenberg, <daniel@haxx.se>, et al.
+# * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
 # *
 # * This software is licensed as described in the file COPYING, which
 # * you should have received as part of this distribution. The terms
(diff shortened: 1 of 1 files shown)


ydroneaud:
This removes EC-ACC '88:49:7f:01:60:2f:31:54:24:6a:e2:8c:4d:5a:ef:10:f1:d8:7e:bb:76:62:6f:4a:e0:b7:f9:5b:a7:96:87:99' certificate from the list of certificates exported by mk-ca-bundle.pl. What's the reason ?

ydroneaud:
OK, certdata.txt contains:

> # For Server Distrust After: Sat Dec 28 00:00:00 2019
> CKA_NSS_SERVER_DISTRUST_AFTER MULTILINE_OCTAL
> \061\071\061\062\062\070\060\060\060\060\060\060\132
> END

(It's not obvious it's distrusted from within Firefox CA certificate tool)

ydroneaud:
It's going to be difficult to process:
EC-ACC is to be trusted until Tue Jan 07 22:59:59 2031, but certificates issued from this CA after Sat Dec 28 00:00:00 2019 shouldn't be trusted.
https://bugzilla.mozilla.org/show_bug.cgi?id=1621159
https://hg.mozilla.org/projects/nss/rev/4d1b7bbeebfe12cb16b2af74cfec4183637014cc

PR #8412, opened 2022-02-09
mk-ca-bundle.vbs: delete this script in favor of mk-ca-bundle.pl

jay:
mk-ca-bundle.vbs is a Windows-specific script for Mozilla certificate
extraction, similar to mk-ca-bundle.pl which runs on any platform. The
vbs version has not been maintained while the perl version has been
maintained with improvements and security fixes. I don't think it's
worth the work to maintain both versions. Windows users should be able
to use mk-ca-bundle.pl without any problems, as long as they have perl.

Closes #xxxx

---

/cc @gknauf


bagder (approved):
Seems totally sensible!

PR #8418, opened 2022-02-09
hostcheck: pass in pattern length too, to avoid a strlen call

bagder:
Removes one strlen() call per SAN name in a cert-check.


bagder:
In a test run using `-L https://google.com`, which makes a redirect, this does 110 strlens less!

PR #8421, opened 2022-02-09
quiche: fix upload for bigger content-length

jpmenil:
When uploading files, we can hit the underlying stream capacity, which in this case will return QUICHE_H3_ERR_DONE (-1), and result in a failure to send data.

Let fix this by handling this return code.

Signed-off-by: Jean-Philippe Menil <jpmenil@gmail.com>

### DIFF
--- a/lib/vquic/quiche.c
+++ b/lib/vquic/quiche.c
@@ -682,7 +682,7 @@ static ssize_t h3_stream_send(struct Curl_easy *data,
     sent = quiche_h3_send_body(qs->h3c, qs->conn, stream->stream3_id,
                                (uint8_t *)mem, len, FALSE);
     if(sent == QUICHE_H3_ERR_DONE) {
-      sent = len;
+      sent = 0;
     }
     if(sent < 0) {
       *curlcode = CURLE_SEND_ERROR;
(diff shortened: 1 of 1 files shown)


lib/vquic/quiche.c
```diff
@@ -681,6 +681,9 @@ static ssize_t h3_stream_send(struct Curl_easy *data,
     H3BUGF(infof(data, "Pass on %zd body bytes to quiche", len));
     sent = quiche_h3_send_body(qs->h3c, qs->conn, stream->stream3_id,
                                (uint8_t *)mem, len, FALSE);
+    if(sent == QUICHE_H3_ERR_DONE) {
+      sent = len;
```
ghedo, line 685: This should be `sent = 0` or something like that, since the `QUICHE_H3_ERR_DONE` error code means that no data was sent, and the write should be retried later.

jpmenil: yeah, seem you're right here, thx !

jpmenil: seem that i got strange result when setting it to 0

jpmenil: so, in fact, i really wonder if quiche must not return BufferTooShort instead of Done here:
https://github.com/cloudflare/quiche/blob/master/quiche/src/h3/mod.rs#L1037

lib/vquic/quiche.c
```diff
@@ -682,7 +682,7 @@ static ssize_t h3_stream_send(struct Curl_easy *data,
     sent = quiche_h3_send_body(qs->h3c, qs->conn, stream->stream3_id,
                                (uint8_t *)mem, len, FALSE);
     if(sent == QUICHE_H3_ERR_DONE) {
-      sent = len;
+      sent = 0;
     }
     if(sent < 0) {
```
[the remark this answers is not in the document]
jpmenil: done

its0x08:
Looks good!
I was waiting for this 😁
Thank you @jpmenil

bagder:
Thanks!

PR #8426, opened 2022-02-10
fix configure error when using '--enable-code-coverage'

gaoxingwang:
fix a spelling error that caused the compilation to fail


bagder (approved):
Nice catch!

bagder:
Thanks!

PR #8428, opened 2022-02-10
Removes strlen() calls per SAN name in a cert-check for chained certificates

HenrikHolst:
Further to #8418 there is for chained certificates a strlen done on the hostname for each iteration of the loop through the chain. For e.g https://curl.se this leads to zero difference since the certificate used there is "sane" but for the https://google.com case which perhaps represents a worst case scenario we can reduce a further 108 calls to strlen.


bagder (approved):

bagder:
Thanks!

PR #8430, opened 2022-02-10
Remove unused data when IPv6 is not supported

MAntoniak:


bagder:
I think #8439 offers a nicer way (== ends up in more readable and maintainable code) to accomplish the same thing for two of the functions: provide "compatibility macros" so that we don't have to do embedded #ifdefs as much.

MAntoniak:
That is look reasonable. The goal of the PR is removed scope_id from connectdata and UserDefined structures . I am going to rebase my PR to use your changes.

bagder:
ok, cool! I'll merge #8439 and then it'll be easier

bagder:
Thanks!

PR #8431, opened 2022-02-10
wolfssl: when SSL_read() returns zero, check the error

bagder:
Returning zero typically indicates end of connection, so if there's no
data read but the connection is alive, it needs to return -1 with
CURLE_AGAIN.

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -923,7 +923,10 @@ static ssize_t wolfssl_recv(struct Curl_easy *data,
     switch(err) {
     case SSL_ERROR_ZERO_RETURN: /* no more data */
       break;
+    case SSL_ERROR_NONE:
+      /* FALLTHROUGH */
     case SSL_ERROR_WANT_READ:
+      /* FALLTHROUGH */
     case SSL_ERROR_WANT_WRITE:
       /* there's data pending, re-invoke SSL_read() */
       *curlcode = CURLE_AGAIN;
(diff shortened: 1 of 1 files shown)


dgarske (commented):
Fix looks good. Thanks Daniel.

lib/vtls/wolfssl.c
```diff
@@ -917,7 +917,7 @@ static ssize_t wolfssl_recv(struct Curl_easy *data,
 
   nread = SSL_read(backend->handle, buf, buffsize);
 
-  if(nread < 0) {
+  if(nread <= 0) {
     int err = SSL_get_error(backend->handle, nread);
 
     switch(err) {
```
jay, line 923:
```suggestion
    switch(err) {
    case SSL_ERROR_NONE: /* this is not an error */
      break;
```
Is an error code guaranteed on 0 read return? I notice in OpenSSL we allow for no stored error (see above suggestion)

regardless fwiw I think the [wolfSSL_read doc](https://www.wolfssl.com/doxygen/group__IO.html#ga33732bde756a527d61a32212b4b9a017) is unclear on 0 read return, because it says return of SSL_FATAL_ERROR (-1) may be due to want read/write but it doesn't say return of 0 (connection closed) can do the same and it seems you're making this change since that can actually happen?

bagder: I don't think that's needed. The docs for [SSL_get_error](https://www.wolfssl.com/doxygen/group__Debug.html#gae30b3ae133f07c6b9d2b567367489b02) says:

> SSL_ERROR_NONE will be returned if ret > 0.

... and we're not calling the function with such an argument.

bagder: I suspect maybe the docs is not correct there.

bagder:
The user who ran into the problem that made me submit this PR to begin with, has confirmed that this fixes their problem.

PR #8438, opened 2022-02-11
projects: add support for Visual Studio 17 (2022)

snir:


MarcelRaad:
Thanks! Looks good to me in general. It only looks a little strange that there's 2017 and 2022 now, but no 2019.

Is there any specific reason you went with the project files?

We currently have 3 build systems for Visual Studio, CMake and nmake in addition to the project files, and could maybe clean that up. The fact that no one has complained about missing Visual Studio 2019 projects so far gave me the impression that it might not be used much anymore.

projects/Windows/VC17/lib/libcurl.tmpl
```diff
@@ -0,0 +1,2381 @@
...
+CURL_LIB_VSSH_H_FILES
+CURL_LIB_VTLS_H_FILES
+  </ItemGroup>
+  <ItemGroup>
+CURL_LIB_RC_FILES
+  </ItemGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <ImportGroup Label="ExtensionTargets">
+  </ImportGroup>
+</Project>
```
jay: remove newline here.

projects/Windows/VC17/lib/libcurl.vcxproj.filters
```diff
@@ -0,0 +1,17 @@
...
+    <Filter Include="Header Files">
+      <UniqueIdentifier>{93995380-89BD-4b04-88EB-625FBE52EBFB}</UniqueIdentifier>
+      <Extensions>h;hh;hpp;hxx;hm;inl;inc;xsd</Extensions>
+    </Filter>
+    <Filter Include="Resource Files">
+      <UniqueIdentifier>{67DA6AB6-F800-4c08-8B7A-83BB121AAD01}</UniqueIdentifier>
+      <Extensions>rc;ico;cur;bmp;dlg;rc2;rct;bin;rgs;gif;jpg;jpeg;jpe;resx;tiff;tif;png;wav;mfcribbon-ms</Extensions>
+    </Filter>
+  </ItemGroup>
+</Project>
```
jay, line 17: remove newline here

projects/build-openssl.bat
```diff
@@ -44,7 +44,11 @@ rem ***************************************************************************
     set OS_PLATFORM=x86
   )
   if defined PROGRAMFILES(x86) (
-    set "PF=%PROGRAMFILES(x86)%"
+    if /i "%~1" == "vc14.3" (
+      set "PF=%PROGRAMFILES%"
+    ) else (
+      set "PF=%PROGRAMFILES(x86)%"
+    )
```
jay, line 51: why is this necessary, and same question for same change in build-wolfssl

jay (changes requested):

jay:
> The fact that no one has complained about missing Visual Studio 2019 projects so far gave me the impression that it might not be used much anymore.

I still use the pregenerated files, but I don't use those versions of Visual Studio. I don't understand the directory scheme because vc15 directory actually should be vc14.1 if I understand correctly, and this vc17 should actually be vc14.3 right?

https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B

Also I think the commit area should be changed from winbuild (which is a different build system) to projects.

MarcelRaad:
@jay 
> I don't understand the directory scheme because vc15 directory actually should be vc14.1 if I understand correctly, and this vc17 should actually be vc14.3 right?

The numbers in the directories reflect the Visual Studio version. So it should probably be `vs17` instead of `vc17`. Or yes, `vc14.3`.

projects/build-openssl.bat
snir: Visual Studio used to be x86 only up to VS17

snir:
@MarcelRaad 
> Is there any specific reason you went with the project files?
> 
> We currently have 3 build systems for Visual Studio, CMake and nmake in addition to the project files, and could maybe clean that up. The fact that no one has complained about missing Visual Studio 2019 projects so far gave me the impression that it might not be used much anymore.

I found the projects files to be the easiest way to get into curl dev on Windows with VS.
Not sure why no one implemented 2019 so far. I guess the base of Windows devs which didn't have 2017 installed is pretty small, as it's probably the most supported VS version for C/C++ projects.


> @jay
> 
> > I don't understand the directory scheme because vc15 directory actually should be vc14.1 if I understand correctly, and this vc17 should actually be vc14.3 right?
> 
> The numbers in the directories reflect the Visual Studio version. So it should probably be `VS17` instead of `VC17`. Or yes, `VC14.3` for the MSVC toolset version.

You're right, it should be VS17, just kept the current convention. It can be addressed in a different PR.

> Also I think the commit area should be changed from winbuild (which is a different build system) to projects.

Yes, should I create a new PR with the commit message fix or force push?

MarcelRaad:
> Yes, should I create a new PR with the commit message fix or force push?

Force-push please :-)

projects/build-openssl.bat
```diff
@@ -44,7 +44,11 @@ rem ***************************************************************************
     set OS_PLATFORM=x86
   )
   if defined PROGRAMFILES(x86) (
-    set "PF=%PROGRAMFILES(x86)%"
+    if /i "%~1" == "vc14.3" (
```
jay, line 47: >Visual Studio used to be x86 only up to VS17

Ok then I suggest add a comment in both those files. Otherwise this PR looks fine.
```suggestion
    rem Visual Studio was x86 only prior to 14.3
    if /i "%~1" == "vc14.3" (
```

jay:
Thanks

PR #8440, opened 2022-02-11
quiche: handle stream reset

### LINKED ISSUE #8437: Support HTTP/3 stream reset detection in quiche backend
### I did this
Looked at the HTTP/3 polling code in quiche backend - https://github.com/curl/curl/blob/96f85a0fef694d764b1a0d981192ac6d3880b1ae/lib/vquic/quiche.c#L619

### I expected the following

A case statement for `QUICHE_H3_EVENT_RESET` https://github.com/cloudflare/quiche/blob/6437b3c2db0dd3c1d6c76cb71d784c874b185d01/quiche/include/quiche.h#L670

### curl/libcurl version

curl @ 96f85a0fef694d764b1a0d981192ac6d3880b1ae

### operating system

n/a

### Rational

The server can reset response streams at any time, These are similar to Finished events except that body data doesn't have to be read to completion before Reset is fired. Not checking for Resets could mean that the client mistakenly sees the stream alive, while the server sees it as terminated.

Furthermore, if curl implements support for cancelling the Request stream (see https://github.com/cloudflare/quiche/blob/6437b3c2db0dd3c1d6c76cb71d784c874b185d01/quiche/include/quiche.h#L329) that will cause a STOP_SENDING to be sent, which will trigger the server to Reset it's side of the stream.

bagder:
A stream reset now causes a CURLE_PARTIAL_FILE error. I'm not convinced
this is the right action nor the right error code.

Reported-by: Lucas Pardue
Fixes #8437

### DIFF
--- a/lib/vquic/quiche.c
+++ b/lib/vquic/quiche.c
@@ -648,6 +648,11 @@ static ssize_t h3_stream_recv(struct Curl_easy *data,
       recvd += rcode;
       break;
 
+    case QUICHE_H3_EVENT_RESET:
+      streamclose(conn, "Stream reset");
+      *curlcode = CURLE_PARTIAL_FILE;
+      return -1;
+
     case QUICHE_H3_EVENT_FINISHED:
       streamclose(conn, "End of stream");
       recvd = 0; /* end of stream */


bagder:
@LPardue any objections to this as a first shot or do you have another idea on how to deal with the situation?

LPardue:
I think returning an error code is ok.

However, there's another reset case that can occur when using quiche_h3_recv_body(). Today this would return a `QUICHE_H3_TRANSPORT_ERR`, which isn't granular enough to know if there was a reset or not. Luckily I have a PR in-flight that exposes more codes; in future you should see `QUICHE_H3_TRANSPORT_ERR_STREAM_RESET`. PR is https://github.com/cloudflare/quiche/pull/1156/files#diff-8677002d221c3e20ec05af5a64f7125a501aade052ab378600fcc3b0a3148c91R677

PR #8441, opened 2022-02-11
mime: use a define instead of the magic number 24

bagder:
MIME_BOUNDARY_DASHES is now the number of leading dashes in the
generated boundary string.

@monnerat approve?


bagder:
Oh, I merged the another one already so I will fix this.

PR #8442, opened 2022-02-11
packages: remove build support for MSVC before VC12 (Visual Studio 2013)

bagder:
With #8438 coming, I think its time to also prune some of the oldest ones.


MarcelRaad:
Hmm, with greatly improved CMake support in the last months, maybe even start another call to remove the VS project files?

Personally, I've been able to use CMake instead of the project files everywhere since at least version 7.79.0 (https://github.com/curl/curl/commit/62dbfa30ba0342ad7586c9c83b432afac49b9d5b) and have since switched to it.

MarcelRaad:
If we want to keep them, I'd maybe be a little more conservative and only remove everything older than 2010 / 10.0. The older ones have more maintenance effort as they use a different format and later Visual Studio versions cannot open them without upgrading them to the current format, so in my opinion that would be the logical cut.

MarcelRaad (approved):

jay:
I'm still using the Visual Studio 2010 project files.

bagder:
> I'm still using the Visual Studio 2010 project files.

Aren't the newer compiler versions also free to download and use? What's the reason to stick with a >decade old version?

jay:
> Aren't the newer compiler versions also free to download and use? What's the reason to stick with a >decade old version?

I have paid tiers of a few older versions of Visual Studio. When I was working I would often have to use a specific version, so I'm sure there are users of older versions. I'm currently doing most of my curl dev from a Windows 7 Enterprise computer (ESU) where VS2010 is more responsive than other versions.

PR #8444, opened 2022-02-12
circleci: add a job using libssh

bagder:


bagder:
I spotted that we have no CI build using libssh

bagder:
This makes test 1459 fail, using `libssh/0.9.3/openssl/zlib`.

On my local machine I have `libssh/0.9.6/openssl/zlib` and I cannot reproduce the failure.

PR #8446, opened 2022-02-12
scripts/ciconfig: show CI job config info

bagder:
Output info about all current CI jobs. configure options only to start with.


bagder:
This script helped me realize that we didn't do any CI builds for libssh and wolfssh...

PR #8447, opened 2022-02-13
projects: Update VC version names for VS2017, VS2022

jay:
- Rename VC15 -> VC14.10, VC17 -> VC14.30.

The projects directory that holds the pre-generated Visual Studio
project files uses VC<ver> to indicate the MSVC version. At some point
support for Visual Studio 2017 (Visual Studio version 15 which uses MSVC
14.10) was added as VC15. Visual Studio 2022 (Visual Studio version 17
which uses MSVC 14.30) project files were recently added and followed
that same format using VC17.

There is no such MSVC version (yet) as VC15 or VC17.

For VS 2017 for example, the name we use is correct as either VS17,
VS2017, VC14.10. I opted for the latter since we use VC for earlier
versions (eg VC10, VC12, etc).

Ref: https://github.com/curl/curl/pull/8438#issuecomment-1037070192

Closes #xxxx

---

Here is one way to do it.

/cc @MarcelRaad  @snir

_edit: Renamed 14.3 to 14.30, renamed 14.1 to 14.10_


MarcelRaad:
Why 14.1 and 14.30? Shouldn't it be 14.1 and 14.3 or 14.10 and 14.30?

jay:
> Why 14.1 and 14.30? Shouldn't it be 14.1 and 14.3 or 14.10 and 14.30?

MS went from 14.1 to 14.11..14.30. I mistakenly thought VS 2022 was 14.3 but it's 14.30. The actual PlatformToolset is v143 though.

jay:
> MS went from 14.1 to 14.11..14.30.

I think you might be right, it is 14.10 and not 14.1. I will install Visual Studio 2017 to get the definitive answer.

jay:
The VS 2017 Visual C++ RTM string is 14.10.25008.0 so I've renamed 14.1 => 14.10. Also, I forgot asked about this 5 years ago in an MS forum and a CSG told me I could use 14.1...

jay:
Last call for feedback before I land this.

MarcelRaad (approved):
Not absolutely sure about YX.Z vs. XY.Z0, but looks good to me!

PR #8450, opened 2022-02-14
lib/urlapi.c:

icing:
- avoid one malloc when setting a new url via curl_url_set()
    and CURLUPART_URL.
  - extract common pattern into a new static function.


bagder (approved):

bagder:
Thanks!

PR #8451, opened 2022-02-14
urldata: remove conn->bits.user_passwd

### LINKED ISSUE #8449: Basic auth credentials passed solely through the url are no longer being reused
### I did this

We use libcurl through [mamba](https://github.com/mamba-org/mamba). After mamba upgraded past libcurl 7.75.0, we started seeing issues with basic auth passed through the URL.

We give curl through `curl_easy_setopt(handle, CURLOPT_URL, url.c_str());` a `url` that has basic auth as part of it.

An example is:

`https://:test@localhost:8005/`

However, we started noticing that reused connections would get a 401 back from the service due to the lack of an auth header. 

### I expected the following

I expected that reused connections would continue to have the basic auth credentials all throughout.

I opened https://github.com/mamba-org/mamba/pull/1141/ where I started iterating to getting a minimal repro and I was able to get that, but from there debugged my way into a potential workaround to this issue. 

If I directly set `curl_easy_setopt(handle, CURLOPT_USERPWD, ":test");`, the issue would no longer come up. So I believe the issue might have something to do with how curl is handling the credentials from the url and they're not being set correctly such that future reuse of the same connection would benefit from the same credentials.

I think this might have something to do with https://github.com/curl/curl/pull/6545 which is very relevant to the release where this got introduced.

### curl/libcurl version

Any libcurl higher than 7.75.0 (I tested against latest as well)

bagder:
While connections for HTTP don't actually have credentials (unless using
NTLM horrors), we maintain that info in there for now and just make sure
to not reset it when a connection is reused.

Fixes #8449
Reported-by: John H. Ayad

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -1738,7 +1738,6 @@ static struct connectdata *allocate_conn(struct Curl_easy *data)
   conn->bits.tunnel_proxy = data->set.tunnel_thru_httpproxy;
 #endif /* CURL_DISABLE_PROXY */
 
-  conn->bits.user_passwd = (data->state.aptr.user) ? TRUE : FALSE;
 #ifndef CURL_DISABLE_FTP
   conn->bits.ftp_use_epsv = data->set.ftp_use_epsv;
   conn->bits.ftp_use_eprt = data->set.ftp_use_eprt;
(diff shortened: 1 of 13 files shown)


bagder:
Clearly not good enough...

lib/url.c
```diff
@@ -2019,45 +2018,47 @@ static CURLcode parseurlandfillconn(struct Curl_easy *data,
...
-        return result;
+      conn->user = decoded;
+      result = Curl_setstropt(&data->state.aptr.user, decoded);
     }
-    else if(uc != CURLUE_NO_PASSWORD)
+    else if(uc != CURLUE_NO_USER)
       return Curl_uc_to_curlcode(uc);
+    else if(data->state.aptr.passwd) {
+      /* no user was set but a password, set a blank user */
+      result = Curl_setstropt(&data->state.aptr.user, "");
```
jay, line 2058: does conn->user need to be set or cleared here

lib/url.c
```diff
@@ -3485,8 +3486,7 @@ static void reuse_conn(struct Curl_easy *data,
 
   /* get the user+password information from the old_conn struct since it may
    * be new for this request even when we re-use an existing connection */
-  conn->bits.user_passwd = old_conn->bits.user_passwd;
-  if(conn->bits.user_passwd) {
+  if(old_conn->user) {
```
jay, line 3489: the other checks that replaced user_passwd are now using the allocated per data aptr.user !=NULL but this one is using conn->user (and became old_conn->user) so now if conn->user and aptr.user are out of sync wouldn't this be as well, in other words you're copying the user/passwd even if aptr.user was set to NULL, unless they are both kept in sync (see earlier comment)

lib/url.c
bagder: Does it really? I can't come up with a scenario or test case when it does!

PR #8452, opened 2022-02-14
avoid using SECONDARYSOCKET when FTP is not support

MAntoniak:
In additional it reduce number of allocation for ssl backend.

I'm not sure about this solution. This will make it more difficult next socket channel. Maybe it doesn't matter? I'm curious to hear your opinions.

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -789,11 +789,17 @@ static CURLcode connect_SOCKS(struct Curl_easy *data, int sockindex,
       conn->http_proxy.host.name :
       conn->bits.conn_to_host ?
       conn->conn_to_host.name :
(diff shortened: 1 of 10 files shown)


bagder:
The likeliness of there ever being another protocol using >1 connection I would say is very slim so we don't need to think about that.

I think the primary problem with this patch is the extreme #ifdef-maze it introduces. It makes the code much harder to read. I think it needs to be done with more "finesse".

MAntoniak:
I originally wanted to reduce the amount of memory allocated. I extracted this work in #8471

bagder:
With #8471 merged, I presume we can close this.

MAntoniak:
Yes, thanks for pull #8471.

PR #8455, opened 2022-02-15
IBM i build improvements

kadler:
Various enhancements to the OS400 build script and docs.

This allows a better default build setup which will stop on failure, which is needed for automated builds to detect status.


bagder:
Thanks!

Issue #8289, opened 2022-01-17
Configure option --with-gssapi=path does not work anymore without pkg-config


mkauf:
### I did this

I have used the "configure" option ``--with-gssapi=path`` to specify the MIT Kerberos to use for GSS API.

This does not work anymore, because pull request #7916 has introduced a bug: When the option ``--with-gssapi=path`` is used, the configure script tries to use pkg-config anyway. If pkg-config is not available, or if pkg-config does not find the necessary information, it is assumed that the GSS library's name is "gssapi". But the correct name (in my case) is "gssapi_krb5".

The configure script prints a (seemingly) unrelated error:
```
checking OpenSSL linking with -ldl... no
checking OpenSSL linking with -ldl and -lpthread... no
configure: OPT_OPENSSL: yes
configure: OPENSSL_ENABLED: 
configure: error: --with-openssl was given but OpenSSL could not be detected
```

From config.log:
```
configure:25314: checking OpenSSL linking with -ldl and -lpthread
[...]
ld: cannot find -lgssapi
```

I think that ``pkg-config`` should only be used if ``--with-gssapi`` is specified without a path, but it should not be used if a path is present. Probably the tool ``krb5-config`` should be used in this case to detect the GSS library's name.

### I expected the following

``configure`` should not abort with an error

### curl/libcurl version

```
curl 7.81.0 (x86_64-pc-linux-gnu) libcurl/7.81.0 OpenSSL/1.1.1m zlib/1.2.11
Release-Date: 2022-01-05
Protocols: ftp ftps http https 
Features: alt-svc GSS-API HSTS HTTPS-proxy IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL TLS-SRP UnixSockets
```

### operating system

CentOS Linux release 7.9.2009 (Core)

bagder:
> think that pkg-config should only be used if --with-gssapi is specified without a path, but it should not be used if a path is present. Probably the tool krb5-config should be used in this case to detect the GSS library's name

I think we should always try pkg-config first, both with and without path given. Simply because pkg-config is the superior way for this kind of data. But if there is no pkg-config available or if gssapi is not supported by it, we should continue to try krb5-config next and then go best-effort without.

dfandrich:
IMHO, if the user is supplying a path, it's there for a reason and should be
used. My expectation as a user would be, if no path is given, for it to first
use pkg-config then fall back to some default location, but if a path is given
[...]

PR #8456, opened 2022-02-15
configure.ac: better use of pkg-config for gssapi, with krb5-config as a fallback

asedeno:
This reverts commit 6fe4e7d3bfc1520da032ea067b4712cc1357f016.

As per #8289, this breaks builds for folks with custom MIT Kerberos installation paths.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1851,6 +1851,18 @@ else
   CPPFLAGS="$save_CPPFLAGS"
 fi
 
+if test x"$want_gss" = xyes; then
+  AC_MSG_CHECKING([if we can link against GSS-API library])
+  AC_LINK_IFELSE([
+    AC_LANG_FUNC_LINK_TRY([gss_init_sec_context])
+  ],[
+    AC_MSG_RESULT([yes])
+  ],[
+    AC_MSG_RESULT([no])
+    AC_MSG_ERROR([--with-gssapi was specified, but a GSS-API library was not found.])
+  ])
+fi
+
 build_libstubgss=no
 if test x"$want_gss" = "xyes"; then
   build_libstubgss=yes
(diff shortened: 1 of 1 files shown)


bagder:
I don't think a straight out revert is the correct fix. I believe the intention in 6fe4e7d3bfc1520da0 was to use the pkg-config file if it is indeed installed in the custom path, and I still think that seems reasonable. If that isn't present there, then a fallback to the `krb5-config` tool seems totally reasonable.

asedeno:
I agree that perhaps being more clever is a good idea, but that can be done separately from fixing the builds that were broken by the change while that fix is figured out.

bagder:
We originally had bug A that was fixed with a commit that introduced bug B. You seem to be suggesting we should immediately reintroduce bug A because of B.

Instead of doing a back-and-forth dance, I think fixing this *proper* is a better idea and I think we are better off not doing a plain revert because if we do, everyone who now suffers from B will just go away and happily leave bug A because they don't suffer from it. I've seen that happen many times.

asedeno:
We had a bug that inconvenienced one person's build system, and fixed that with a bug that broke other people's build systems. There's a severity to be considered.

That said, I have tried to generate a `configure` file from `configure.ac` using autoconf (versions 2.69 and 2.71, both via debian), and I can't get what is in the release bundle. If you can get me explicit instructions, including dependencies, on how to generate the configure that I find in, say, curl 7.81.0, I will try to make a patch that works for everyone.

asedeno:
I believe this will work for everyone involved, and as a bonus, eliminates the debugging that went into figuring out and filing #8289 by trying to link gssapi early.

bagder:
Thanks!

PR #8457, opened 2022-02-15
No longer call poll() in Curl_readwrite

HenrikHolst:
Curl_readwrite() which is only called by multi_runsingle() calls Curl_socket_check() with a zero timeout which in turn makes a call to select/poll to I guess check if the socket is still connected? Anyway since the user is supposed to use their own event system for the multi API there should IMHO be no calls to poll() inside the multi API at all, not to mention that it's an extra syscall.

However with this patch the unit tests refuses to work so either I'm missing something or something else is wrong here.

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -1133,38 +1133,13 @@ CURLcode Curl_readwrite(struct connectdata *conn,
   struct SingleRequest *k = &data->req;
   CURLcode result;
   int didwhat = 0;
-
-  curl_socket_t fd_read;
-  curl_socket_t fd_write;
   int select_res = conn->cselect_bits;
 
   conn->cselect_bits = 0;
 
-  /* only use the proper socket if the *_HOLD bit is not set simultaneously as
-     then we are in rate limiting state in that transfer direction */
-
-  if((k->keepon & KEEP_RECVBITS) == KEEP_RECV)
-    fd_read = conn->sockfd;
-  else
-    fd_read = CURL_SOCKET_BAD;
-
-  if((k->keepon & KEEP_SENDBITS) == KEEP_SEND)
-    fd_write = conn->writesockfd;
-  else
-    fd_write = CURL_SOCKET_BAD;
-
   if(data->state.drain) {
     select_res |= CURL_CSELECT_IN;
     DEBUGF(infof(data, "Curl_readwrite: forcibly told to drain data"));
-  }
-
-  if(!select_res) /* Call for select()/poll() only, if read/write/error
-                     status is not known. */
-    select_res = Curl_socket_check(fd_read, CURL_SOCKET_BAD, fd_write, 0);
-
-  if(select_res == CURL_CSELECT_ERR) {
-    failf(data, "select/poll returned error");
-    return CURLE_SEND_ERROR;
   }
 
 #ifdef USE_HYPER


bagder:
This reminded me why the poll() was there in the first place. The multi API  doesn't pass on the "activity", so `Curl_readwrite()` doesn't know if the socket should be read from or written to. The removal of the poll will then possibly risk doing an extra call to one or both of recv and send. Of course, risking those is probably still better than unconditionally doing the poll call there!

I made things run smoother with this tiny change:
~~~diff
diff --git a/lib/transfer.c b/lib/transfer.c
index 10703f219..f6a0acb57 100644
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -1131,12 +1131,12 @@ CURLcode Curl_readwrite(struct connectdata *conn,
                         bool *comeback)
 {
   struct SingleRequest *k = &data->req;
   CURLcode result;
   int didwhat = 0;
-  int select_res = conn->cselect_bits;
-
+  int select_res = conn->cselect_bits ?
+    conn->cselect_bits : (CURL_CSELECT_IN | CURL_CSELECT_OUT);
   conn->cselect_bits = 0;
 
   if(data->state.drain) {
     select_res |= CURL_CSELECT_IN;
     DEBUGF(infof(data, "Curl_readwrite: forcibly told to drain data"));

HenrikHolst:
Ah, you mean when curl_multi_socket_action is called with a cflags of 0 when the caller is unsure what the flags might be? And incidentally on a timeout one calls curl_multi_socket_action with cflags=0 as well but internally this function does not know that it was called due to a timeout (at which it shouldn't call poll or read or write) or if it was called with cflags=0 due to the caller not knowing which action to take.

HenrikHolst:
Or rather isn't the main issue the loop inside multi_socket, aka this one:

`  do {
    /* the first loop lap 'data' can be NULL */
    if(data) {
      SIGPIPE_VARIABLE(pipe_st);

      sigpipe_ignore(data, &pipe_st);
      result = multi_runsingle(multi, &now, data);
      sigpipe_restore(&pipe_st);

      if(CURLM_OK >= result) {
        /* get the socket(s) and check if the state has been changed since
           last */
        result = singlesocket(multi, data);
        if(result)
          return result;
      }
    }
`

that calls multi:_runsingle over and over until some condition is met and that condition is perhaps not properly set by sendto. Aka where I see this is that curl does a successful sendto with the whole request but then it gets back into the loop at which time the current code calls poll() and the patched code calls recvfrom(), both fails so the loop exits.

But if a sendto is successful and we have no more data to send then the loop should be exited since there are no reason to do anything else until we get another event from whatever event handler the caller is using. I can see that we in a recvfrom situation have to read until EAGAIN due to using nonblocking sockets but for writes the loop feels unnecessary.

HenrikHolst:
My mistake, it's the loop inside multi_runsingle() that does it, the loop in multi_socket() exits after one run in this case.

HenrikHolst:
So the loop in multi_runsingle() runs like this:

data->mstate = MSTATE_DO;
sendto(8, "GET /content/since/1645038037000"..., 225, MSG_NOSIGNAL, NULL, 0) = 225

data->mstate = MSTATE_DID;
<- no syscall, and the loop iterates again ->

data->mstate = MSTATE_PERFORMING;
recvfrom(8, 0x55bd4dcd5288, 16384, 0, NULL, NULL) = -1 EAGAIN

So the loop goes over the socket three times. Sorry for the spamming and I'll be quite for the rest of the day now!

bagder:
> Ah, you mean when curl_multi_socket_action is called with a cflags of 0 when the caller is unsure what the flags might be?

No, I mean when `curl_multi_perform()` is called as a result of action on one of the sockets libcurl uses.

HenrikHolst:
but the caller fills in the cflags with the actions then, or does this also happens inside the api and the flags are no longer set?

bagder:
Nope. [curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html) is called for any change (or even no change) or one or more of the sockets, and it does say which change nor on which socket. It is only `curl_multi_socket_action()` that tells libcurl which socket and which action that occurred.

HenrikHolst:
ah of course, I was mixing the functions in my head... But should not curl_multi_perform simply set the cflags to read+write and then basically call curl_multi_socket_action behind the scene?

bagder:
Pretty much, yes, and that is roughly what my patch provided above makes happen.

HenrikHolst:
but much lower in the stack so that it does repeated calls for not only curl_multi_perform but also curl_multi_socket_action

bagder:
Sure. I'm not denying that things can be improved. The only question is probably if it makes a measurable improvement.

HenrikHolst:
Depends on what counts as measurable I guess. Whenever I call curl_multi_socket_action for timeout there is a 18µs penalty due to the extra syscall:

`poll([{fd=8, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}], 1, 0) = 0 (Timeout) <0.000020>
sendto(8, "GET /content/since/1645134808000"..., 225, MSG_NOSIGNAL, NULL, 0) = 225 <0.000043>
poll([{fd=8, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}], 1, 0) = 0 (Timeout) <0.000018>
`

The first is probably impossible to get rid of since it is used by curl to test if the socket is still a valid one but if we count that one in then what "should" take 43µs to do a single send is now taking 81µs which is close to twice as long. Removing the extra poll at the end there would reduce the overhead by 22% but yes we are talking microseconds here.

bagder:
> Whenever I call curl_multi_socket_action for timeout

On a timeout, it does not normally send. A typical send would happen on socket readability. Calling curl_multi_socket_action for timeout should be relatively unusual.

> The first is probably impossible to get rid of since it is used by curl to test if the socket is still a valid one

It should only check if sockets are still alive when it sets up a new connection, which in most use cases is only once in quite a lot of microseconds. For a transfer in progress it doesn't need poll() to check for "aliveness".

> Removing the extra poll at the end there would reduce the overhead by 22% but yes we are talking only microseconds here.

22% of that single operation in that comparison, yes. Can you measure that curl built with that patch transfers data faster?

HenrikHolst:
Yes curl with the original patch but without your fix doesn't do the second poll so if I measure by counting cycles spent in syscalls with a trace the transfer is 18µs quicker.

Yes on timeout it does not normally send, it's just me that initites the just added easy handle by calling multi_socket_action with the timeout socket.

HenrikHolst:
Looking at a trace of curl_easy_perform on a cached connection to the same server:

`rt_sigaction(SIGPIPE, NULL, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, 8) = 0 <0.000019>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000018>
poll([{fd=4, events=POLLIN}], 1, 0)     = 0 (Timeout) <0.000021>
rt_sigaction(SIGPIPE, NULL, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, 8) = 0 <0.000017>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000019>
poll([{fd=6, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}], 1, 0) = 0 (Timeout) <0.000019>
write(6, "\27\3\3\1\0167\327b8\253\241n.H\305\5Y.u\213\245\374\317\36D\306\33\10\247\25`\32"..., 275) = 275 <0.000054>
poll([{fd=6, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}], 1, 0) = 0 (Timeout) <0.000019>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000019>
poll([{fd=6, events=POLLIN}, {fd=4, events=POLLIN}], 2, 1000) = 0 (Timeout) <1.001142>
rt_sigaction(SIGPIPE, NULL, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, 8) = 0 <0.000019>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000018>
poll([{fd=6, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}], 1, 0) = 0 (Timeout) <0.000019>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000015>
poll([{fd=6, events=POLLIN}, {fd=4, events=POLLIN}], 2, 1000) = 1 ([{fd=6, revents=POLLIN}]) <0.241725>
rt_sigaction(SIGPIPE, NULL, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, 8) = 0 <0.000021>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000019>
poll([{fd=6, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}], 1, 0) = 1 ([{fd=6, revents=POLLIN|POLLRDNORM}]) <0.000019>
read(6, "\27\3\3\2\215", 5)             = 5 <0.000021>
read(6, "\223\30\243)\254+\224\316\356l\4\304\340_\360\312l\361B4\220A!\374\244\31\347R\16\n\216\366"..., 653) = 653 <0.000019>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000019>
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f57680eb210}, NULL, 8) = 0 <0.000017>
`

Not only are there lots of calls to poll with a zero timeout value but there is also quite a lot of rt_sigaction being called. Wouldn't it be better to call send() with the MSG_NOSIGNAL flag than setting and resetting the SIGPIPE signal?

bagder:
> Wouldn't it be better to call send() with the MSG_NOSIGNAL flag

It already does

> than setting and resetting the SIGPIPE signal?

That fiddling is necessary due to the TLS library you use that *doesn't* use the above mentioned method!

HenrikHolst:
Ah ok, so the impossible task is then to get OpenSSL to accept a setting to use MSG_NOSIGNAL. Well one can only try :-)

bagder:
Right, having such an option would remove our need to do that extra signal switwaroo all the time.

HenrikHolst:
Well we have to see what they say but the patch have been sent: https://github.com/openssl/openssl/pull/17734

bagder:
Are you still working on this?

HenrikHolst:
I've sent the patched upstream to OpenSSL but then they wanted a test case for this "new functionality" and besides other work having been in the way I'm not that familiar with the inner workings of OpenSSL yet that I can write that unit test for them so it will have to wait some time more I'm afraid. 

If we are talking only about the poll() then my conclusion is that the first one is probably unavoidable since curl want's to check if the socket is still alive so a syscall have to be done either way and I don't think that poll here is slower than any other one. Your patch removed the second call so that was already a win, so this was not totally in vain :)

bagder:
Closing this for now since there is no current activity. Feel free to reopen/recreate later if there's something to do.

ratacolita:
Thank you for submitting the pull request. I have reviewed it and I must say, your patch has proven to be really useful for my use case. I was using libcurl to create a lightweight load injector, and I noticed an unexpectedly high count of poll syscalls during some tests. After applying your patch, the CPU usage for a single threaded scenario went down from 100% to about 10%, even with a thousand requests being processed. This represents a significant improvement in the performance of the library.

I appreciate the effort you put into addressing the excessive count of rt_sigaction mentioned in the discussions. I have personally resolved this by using gnutls instead of openssl, and it works perfectly for my use case.

I would be happy to assist in any way possible to ensure that this patch is included in the project. Please let me know if there is anything else I can do to help. Thank you again for your contribution.

PR #8458, opened 2022-02-16
docs/DEPRECATE: remove NPN support in August 2022

bagder:
NPN is simply not used these days...


bagder:
No objections heard [on the mailing list](https://curl.se/mail/lib-2022-02/0110.html) either. Proceed to merge.

PR #8459, opened 2022-02-16
des: fix compile break for OpenSSL without DES

MarcelRaad:
When `USE_OPENSSL` was defined but OpenSSL had no DES support and a
different crypto library was used for that, `Curl_des_set_odd_parity`
was called but not defined. This could for example happen on Windows
and macOS when using OpenSSL v3 with deprecated features disabled.

Use the same condition for the function definition as used at the
caller side, but leaving out the OpenSSL part to avoid including
OpenSSL headers.

### DIFF
--- a/lib/curl_des.c
+++ b/lib/curl_des.c
@@ -22,7 +22,12 @@
 
 #include "curl_setup.h"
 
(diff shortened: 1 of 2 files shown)


MarcelRaad:
All callers are here: https://github.com/curl/curl/blob/master/lib/curl_ntlm_core.c

PR #8464, opened 2022-02-16
EVP_sha256() does not appear in OpenSSL until 0.9.7h, not available by default until 0.9.8

asedeno:
We can't use OpenSSL's SHA256 implementation in versions that predate it, and the `EVP_sha256()` function does not appear until OpenSSL 0.9.7h. It is possible it was not enabled by default until later, and someone else can propose moving this threshold up further, but this is the minimum lower bound for this to work.

### DIFF
--- a/lib/sha256.c
+++ b/lib/sha256.c
@@ -40,7 +40,7 @@
 
 #include <openssl/opensslv.h>
 
-#if (OPENSSL_VERSION_NUMBER >= 0x0090700fL)
+#if (OPENSSL_VERSION_NUMBER >= 0x0090800fL)
 #define USE_OPENSSL_SHA256
 #endif
 


bagder:
Does it really appear in 0.9.7h? [this email](https://marc.info/?l=openssl-users&m=135355590501495) seems to suggest that it might not. I presume you're using a version older than 0.9.7h yourself? (I also presume you're fully aware that you really should upgrade to a more modern version)

asedeno:
My determination was made using a git checkout of the openssl repo, checking out the various tags, and grepping for `EVP_sha256`. There were zero results before 0.9.7h, and consistent results from 0.9.7h to 0.9.7m, which is why I say 0.9.7h is a lower bound below which it will definitely not work.

I saw that thread, but it was not clear to me in the changelogs from 0.9.7h to 0.9.7m when something relevant may have changed.

The ancient platform I'm building git (and for git, curl) for is an old Solaris installation with a bunch of packages installed at `/usr/athena/` and OpenSSL 0.9.7.d. It's not mine to upgrade.

asedeno:
You know, building these versions won't take that long. I'll try to pin it down better.

asedeno:
Okay, I tested with a tiny program,

```C
#include <stdio.h>
#include <openssl/evp.h>

int main(char* argv, int argc) {
     printf("%p\n", EVP_sha256);
     return 0;
}
```
and compiled and it, statically linking against libssl.a and libcrypto.a from builds of OpenSSL 0.9.7h-0.9.7m and 0.9.8 with default config (no-asm for 0.9.8, because I needed it to build). Only 0.9.8 worked.

```
$ cc -I./openssl-0.9.7m/include sha256.c ./openssl-0.9.7m/libcrypto.a ./openssl-0.9.7m/libssl.a && ldd ./a.out && ./a.out 
sha256.c: In function ‘main’:
sha256.c:5:21: error: ‘EVP_sha256’ undeclared (first use in this function); did you mean ‘EVP_sha1’?
      printf("%p\n", EVP_sha256);
                     ^~~~~~~~~~
                     EVP_sha1
sha256.c:5:21: note: each undeclared identifier is reported only once for each function it appears in
```

```
 $ cc -I./openssl-0.9.8/include sha256.c ./openssl-0.9.8/libcrypto.a ./openssl-0.9.8/libssl.a && ldd ./a.out && ./a.out 
        linux-vdso.so.1 (0x00007fff165a2000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f123f311000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f123f55f000)
0x556722efc4a0
```
So I think we want to push that back up to 0.9.8 and will update the patch accordingly.

jay:
Thanks

PR #8465, opened 2022-02-16
http: fix "unused parameter ‘conn’" warning

bagder:
Follow-up from 7d600ad1c395

Spotted on appveyor


jay:
Ref: https://github.com/curl/curl/commit/7d600ad1c3959bbf78dc14498335b8a73cfc790c#r66823159

PR #8468, opened 2022-02-17
[no merge] IPFS support for Curl

jchris:
The branch state is currently just an empty shell of a build. Next step is to write the features. I've written to the mailing list asking for feedback about my proposed design. Here is the message I sent:

I am working on a patch to add IPFS support to curl. The goal is not to bring in full P2P node behavior, or provide data into the IPFS network. Rather it is to allow simple retrieval by content identifier (CID) so that shell script authors and other Curl users can use IPFS urls.

Example usage:

```
curl ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/wiki/Vincent_van_Gogh.html
```

This would output the response you’d get from an HTTP request to a public IPFS gateway, eg what you’d get for this:

```
curl https://dweb.link/ipfs/bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/wiki/Vincent_van_Gogh.html
```

(edited to concur with @lidel's comment below)

You can [read more about the interplay between web addresses and IPFS addressees here.](https://docs.ipfs.io/how-to/address-ipfs-on-web/)

Initially I think GET support is all that is needed to be useful. There are a number of public gateways maintained by the community ([ipfs.dweb.link](http://ipfs.dweb.link/) in the above example), so the implementation I am planning is to use a default gateway and make libcurl HTTPS requests internally. The gateway could be overridden with an environment variable, like this:

```
CURL_IPFS_GATEWAY=http://127.0.0.1:8080 curl ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/wiki/Vincent_van_Gogh.html
```

My thinking is this will be a small patch, [smaller than for instance the MQTT support](https://github.com/curl/curl/commit/2522903b792ac5a802f780df60dc4647c58e2477#diff-ed8b659ce184aa0404c6a2642a6bfab86d61f31282502867deb3663204468290).

Are there any example patches adding support for other protocols which are really just wrappers around HTTP? That would help me stick with the coding style. Maybe something in an example curl usage repo? 

Are there technical concerns or other reasons why Curl would be unlikely to accept this patch? I’m happy to include feedback to improve the quality. I’d like to imagine we can start with a small amount of functionality (GET only) and add more as the community adopts it. (For instance, validating the returned content matches the CID requested would be easy to add, but not part of MVP.)


lib/ipfs.c
```diff
@@ -0,0 +1,74 @@
...
+static CURLcode ipfs_do(struct Curl_easy *data, bool *done);
+
+/*
+ * MQTT protocol handler.
+ */
+
+const struct Curl_handler Curl_handler_ipfs = {
+  "IPFS",                             /* scheme */
+  ZERO_NULL,                    /* setup_connection */
+  ipfs_do,                            /* do_it */
```
jchris, line 47: The main feedback I need is: If my feature is essentially just rewriting an `ipfs://<CID>/<PATH>` url to an `https://<CID>.ipfs.dweb.link/<PATH>` url, should I have an ipfs_do function that looks like a normal program that happens to use curl https public APIs -- or should I implement a full Curl_handler that passes through to the `http_do` and similar functions for everything after the URL rewrite.

bagder:
Is there a spec somewhere for this `ipfs://` URL format? I googled for it, but [this page](https://docs.ipfs.io/how-to/address-ipfs-on-web/) came up first, which uses `http://` or `https://` for all accesses...

lidel:
@bagder  the format is the same as HTTP URL, but with Location-based authority component (host+port) replaced with content-addressed one in the form of a unique content identifier ([CID](https://docs.ipfs.io/concepts/content-addressing/)), which itself is an opaque string (`bafy...`, `Qm..`).

Our north star here is to reuse existing standards to maximize interop with existing user agents like browsers and CLI tools, so if something is not clear, HTTP URL rules apply. 

On the formal side of things, we hope to consolidate specs this year, but for now:
- Native addresses:  [docs.ipfs.io/how-to/address-ipfs-on-web/#native-urls](https://docs.ipfs.io/how-to/address-ipfs-on-web/#native-urls)
  - You may find this terse  [IPFS Addressing in Web Browsers Memo](https://github.com/ipfs/in-web-browsers/blob/master/ADDRESSING.md#ipfs-addressing-in-web-browsers) more useful – [Brave](https://brave.com/brave-integrates-ipfs/) and  [Opera](https://blog.ipfs.io/2021-02-08-opera-ios-and-ipfs/) based their support of  `ipfs://` (backed by a public or local gateways) on that memo 
- Real life use and our [collab with Igalia](https://blog.ipfs.io/2021-01-15-ipfs-and-igalia-collaborate-on-dweb-in-browsers/) resulted in `ipfs://` and `ipns://` being registered at IANA: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
- Historical timeline and spec consolidation is tracked  in [ipfs/specs#248](https://github.com/ipfs/specs/issues/248)

bagder:
> I've written to the mailing list asking for feedback

I haven't seen any such email on the list. I'm guessing you sent that from an email address that isn't subscribed, as those are discarded.

jchris:
Thanks for letting me know. I mistakenly sent from the wrong alias. I've sent it again.

On the implementation question -- are there any good examples of protocols which essentially wrap HTTP in Curl? I've looked through the code base but don't see any clear prior art to answer the question I posed here: https://github.com/curl/curl/pull/8468#discussion_r809283552

jchris:
Closing this while we work on an implementation in the tool rather than the library. This pull request might still serve useful for someone wanting an illustration of the build etc changes required to introduce a new protocol.

jchris:
Also, for folks wanting to push this idea forward, my best guess on where to begin a tool implementation is https://github.com/curl/curl/blob/f974bee6235ea049667e9c8f8edc2c84cc5e0f91/src/tool_operate.c#L2398 Also, I'm happy to advise any engineers who are picking this up on what I learned about build layout etc.

bagder:
Look at this point for example:

https://github.com/curl/curl/blob/b0325628b4e78c510691deec32f62b211d2d0bb0/src/tool_operate.c#L1208

... it looks like a place where it could check for `ipfs://` as a scheme and then transform the given URL into a `https://` one to use when continuing...

markg85:
Hi,

Little heads up on my end. I've been made aware of this issue and asked to implement this.
I've done the exact same on the ffmpeg side (still pending approval though) [1] and I've made a proposed spec change for IPFS to expose the gateway it starts [2].

On the ffmpeg side i made the gateway auto detectable in a "happy day" scenario. By that i mean that if you install IPFS as a normal user (so you get a ~/.ipfs folder) that it would automatically find the gateway. This does depend on [2] to be implemented on the IPFS side though. That seems to be OK just some time is needed there from a Go capable dev to make it happen. A user can still define a gateway explicitly by setting the `IPFS_GATEWAY` environment variable. And/or the user can help the auto detection by explicitly setting the `IPFS_PATH` environment variable to where IPFS data is stored. $IPFS_PATH/gateway should then contain a gateway if it's started (again dependent on [2]).

The only question i currently have is if it would be allowed to - in the context of doing a `curl ipfs://<cid>` call - to enable follow redirects if it wasn't explicitly set on the command line?

I'll make a new PR once i have this implemented.

[1] https://ffmpeg.org/pipermail/ffmpeg-devel/2022-March/294251.html
[2] https://github.com/ipfs/specs/issues/261

jchris:
@markg85 we had some discussion about the redirect question, and the initial view was yes, you can follow redirects. Since IPFS is a content addressed network, they don't have the same implications for correctness as a location based redirect would.

PR #8469, opened 2022-02-17
schannel: Move the algIds array out of schannel.h

JDepooter:
This array is only used by the `SCHANNEL_CRED` struct in the `schannel_acquire_credential_handle` function. It can therefore be kept as a local variable. This is a minor update to bbb71507b7bab52002f9b1e0880bed6a32834511.

This change also updates the `NUM_CIPHERS` value to accurately count the number of ciphers options listed in schannel.c, which is 47 instead of 45. It is unlikely that anyone tries to set all 47 values, but if they had tried, the last two would not have been set.

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -325,13 +325,15 @@ get_alg_id_by_name(char *name)
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


bagder (approved):

bagder:
Thanks!

PR #8471, opened 2022-02-17
ssl: reduce allocate space for ssl backend when FTP is not supported.

MAntoniak:
Additionally, adding a backend check for NULL everywhere

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -136,15 +136,6 @@ bool curl_win32_idn_to_ascii(const char *in, char **out);
 #include "curl_memory.h"
 #include "memdebug.h"
 
-/* Count of the backend ssl objects to allocate */
-#ifdef USE_SSL
-#  ifndef CURL_DISABLE_PROXY
-#    define SSL_BACKEND_CNT 4
-#  else
-#    define SSL_BACKEND_CNT 2
-#  endif
-#endif
-
 static void conn_free(struct connectdata *conn);
 
 /* Some parts of the code (e.g. chunked encoding) assume this buffer has at
@@ -752,7 +743,9 @@ static void conn_shutdown(struct Curl_easy *data, struct connectdata *conn)
   /* close the SSL stuff before we close any sockets since they will/may
      write to the sockets */
   Curl_ssl_close(data, conn, FIRSTSOCKET);
+#ifndef CURL_DISABLE_FTP
   Curl_ssl_close(data, conn, SECONDARYSOCKET);
+#endif
 
   /* close possibly still open sockets */
   if(CURL_SOCKET_BAD != conn->sock[SECONDARYSOCKET])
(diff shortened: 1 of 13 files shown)


lib/url.c
```diff
@@ -1668,18 +1675,28 @@ static struct connectdata *allocate_conn(struct Curl_easy *data)
...
+    allocate_result &= conn->ssl[SECONDARYSOCKET].backend != NULL;
+#endif
+#ifndef CURL_DISABLE_PROXY
+    conn->proxy_ssl[FIRSTSOCKET].backend = calloc(1, sslsize);
+    allocate_result &= conn->proxy_ssl[FIRSTSOCKET].backend != NULL;
+#ifndef CURL_DISABLE_FTP
+    conn->proxy_ssl[SECONDARYSOCKET].backend = calloc(1, sslsize);
+    allocate_result &= conn->proxy_ssl[SECONDARYSOCKET].backend != NULL;
+#endif
+#endif
```
bagder, line 1693: I'm not a total fan of how this turns one malloc into four for a regular build. How about instead making the function similar to this? (untested)
~~~c
static struct connectdata *allocate_conn(struct Curl_easy *data)
{
  struct connectdata *conn = calloc(1, sizeof(struct connectdata));
  if(!conn)
    return NULL;

#ifdef USE_SSL
  /* The SSL backend-specific data (ssl_backend_data) objects are allocated as
     a separate array to ensure suitable alignment.
     Note that these backend pointers can be swapped by vtls (eg ssl backend
     data becomes proxy backend data). */
  {
    size_t onesize = Curl_ssl->sizeof_ssl_backend_data;
    size_t totalsize = onesize;
    void *ptr;
    bool allocate_result = true;

#ifndef CURL_DISABLE_FTP
    totalsize *= 2;
#endif
#ifndef CURL_DISABLE_PROXY
    topalsize *= 2;
#endif

    conn->ssl[FIRSTSOCKET].backend = ptr = calloc(1, totalsize);
    if(!ptr)
      return NULL;
#ifndef CURL_DISABLE_FTP
    conn->ssl[SECONDARYSOCKET].backend = ptr += onesize;
#endif
#ifndef CURL_DISABLE_PROXY
    conn->proxy_ssl[FIRSTSOCKET].backend = ptr += onesize;
#ifndef CURL_DISABLE_FTP
    conn->proxy_ssl[SECONDARYSOCKET].backend = ptr += onesize;
#endif
#endif
  }
#endif

lib/vtls/bearssl.c
```diff
@@ -339,6 +339,8 @@ static CURLcode bearssl_connect_step1(struct Curl_easy *data,
   struct in_addr addr;
 #endif
 
+  DEBUGASSERT(backend != NULL);
```
bagder, line 342: I think you should remove `!= NULL` from these checks, as we normally check pointers for being non-NULL like that everywhere else.

lib/vtls/openssl.c
```diff
@@ -1862,6 +1867,8 @@ static CURLcode verifystatus(struct Curl_easy *data,
   ASN1_GENERALIZEDTIME *rev, *thisupd, *nextupd;
   int ret;
 
+  DEBUGASSERT(backend != NULL);
```
bagder, line 1870: This is the first of 4 compiler warnings because you add the assert before the last variable declaration:
~~~
vtls/openssl.c:1872:3: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]

lib/url.c
MAntoniak: I have modified the memory allocation as you wrote.

bagder (approved):

bagder:
Thanks!

PR #8473, opened 2022-02-18
vtls: remove erroneous assert

### LINKED ISSUE #8472: Assertion incorrectly triggering in Curl_ssl_getsessionid()
### I did this

Built a `Debug` version of curl using `gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0`.

Passed the `-D ENABLE_DEBUG=1` flag to CMake to enable assertions in libcurl.

Modified `docs/examples/https.c` to set `CURLOPT_SSL_SESSIONID_CACHE` to zero, as follows:

```
int main(void)
{
  CURL *curl;
  CURLcode res;

  curl_global_init(CURL_GLOBAL_DEFAULT);

  curl = curl_easy_init();
  if(curl) {

    curl_easy_setopt(curl, CURLOPT_URL, "https://example.com/");

    // Added this line
    curl_easy_setopt(curl, CURLOPT_SSL_SESSIONID_CACHE, 0L);

```
Built `https.c` with this command:
```
gcc https.c -I ../../include/ -L ../../build/Debug/lib -l:libcurl-d.so
```

When this code ran, I received this error:
```
a.out: /home/jim/repos/curl/lib/vtls/vtls.c:424: Curl_ssl_getsessionid: Assertion `((CURLPROXY_HTTPS == conn->http_proxy.proxytype && ssl_connection_complete != conn->proxy_ssl[conn->sock[1] == -1 ? 0 : 1].state) ? data->set.proxy_ssl.primary.sessionid : data->set.ssl.primary.sessionid)' failed.
Aborted
```

Reviewing `vtls.c` line 424, it is asserting on the value of sessionid, which is correctly set to zero because of the call to `curl_easy_setopt(curl, CURLOPT_SSL_SESSIONID_CACHE, 0L);`. The zero is interpreted as a boolean false and the assertion fails.

It appears that Line 424 should be removed entirely because it's handled properly on line 426.

### I expected the following

I expected the sample to run correctly and print the content from example.com.

### curl/libcurl version

Using `commit 161cbc502ee0bdbda052d6da17d24fa7835f83e7`, but this problem goes back to at least curl-7.79.1 .

### operating system

Linux mynode 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

bagder:
As assert that required an non-zero session id cache just before the
run-time check that handles it correctly is an incorrect assert.

Reported-by: Jim Beveridge
Fixes #8472

### DIFF
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -421,8 +421,6 @@ bool Curl_ssl_getsessionid(struct Curl_easy *data,
     return TRUE;
 #endif
 
-  DEBUGASSERT(SSL_SET_OPTION(primary.sessionid));
-
   if(!SSL_SET_OPTION(primary.sessionid) || !data->state.session)
     /* session ID re-use is disabled or the session cache has not been
        setup */


jay:
I don't see how it is erroneous, it's doing exactly what I intended, all session id code is supposed to check SSL_SET_OPTION(primary.sessionid)

~~~diff
index fe9a686..01fcac7 100644
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -3225,21 +3225,23 @@ static CURLcode ossl_connect_step1(struct Curl_easy *data,
     return CURLE_SSL_CONNECT_ERROR;
   }
 
-  Curl_ssl_sessionid_lock(data);
-  if(!Curl_ssl_getsessionid(data, conn, SSL_IS_PROXY() ? TRUE : FALSE,
-                            &ssl_sessionid, NULL, sockindex)) {
-    /* we got a session id, use it! */
-    if(!SSL_set_session(backend->handle, ssl_sessionid)) {
-      Curl_ssl_sessionid_unlock(data);
-      failf(data, "SSL: SSL_set_session failed: %s",
-            ossl_strerror(ERR_get_error(), error_buffer,
-                          sizeof(error_buffer)));
-      return CURLE_SSL_CONNECT_ERROR;
+  if(SSL_SET_OPTION(primary.sessionid)) {
+    Curl_ssl_sessionid_lock(data);
+    if(!Curl_ssl_getsessionid(data, conn, SSL_IS_PROXY() ? TRUE : FALSE,
+                              &ssl_sessionid, NULL, sockindex)) {
+      /* we got a session id, use it! */
+      if(!SSL_set_session(backend->handle, ssl_sessionid)) {
+        Curl_ssl_sessionid_unlock(data);
+        failf(data, "SSL: SSL_set_session failed: %s",
+              ossl_strerror(ERR_get_error(), error_buffer,
+                            sizeof(error_buffer)));
+        return CURLE_SSL_CONNECT_ERROR;
+      }
+      /* Informational message */
+      infof(data, "SSL re-using session ID");
     }
-    /* Informational message */
-    infof(data, "SSL re-using session ID");
+    Curl_ssl_sessionid_unlock(data);
   }
-  Curl_ssl_sessionid_unlock(data);
 
 #ifndef CURL_DISABLE_PROXY
   if(conn->proxy_ssl[sockindex].use) {
~~~

jay:
suggest closing in favor of #8484

PR #8474, opened 2022-02-19
Fixed BearSSL bug: TLS Session Resumption (session id) not working

jan2000:

In `bearssl_connect_step1` the function `br_ssl_client_reset()` is called with `resume_session` set to 0. This will make BearSSL clear the specified session id. Setting the `resume_session` paramater to 1 fixes this.

### DIFF
--- a/lib/vtls/bearssl.c
+++ b/lib/vtls/bearssl.c
@@ -474,7 +474,7 @@ static CURLcode bearssl_connect_step1(struct Curl_easy *data,
     hostname = snihost;
   }
 
-  if(!br_ssl_client_reset(&backend->ctx, hostname, 0))
+  if(!br_ssl_client_reset(&backend->ctx, hostname, 1))
     return CURLE_FAILED_INIT;
   backend->active = TRUE;
 


lib/vtls/bearssl.c
```diff
@@ -474,7 +474,7 @@ static CURLcode bearssl_connect_step1(struct Curl_easy *data,
     hostname = snihost;
   }
 
-  if(!br_ssl_client_reset(&backend->ctx, hostname, 0))
+  if(!br_ssl_client_reset(&backend->ctx, hostname, 1))
```
jay, line 477: curl has its own session management. How does this affect the call to `br_ssl_engine_set_session_parameters` above?

michaelforney: Passing 0 here causes BearSSL to disable session resumption, so it won't resume a session with parameters set earlier. Passing 1 instead will cause BearSSL to actually use those parameters.

This was just an oversight on my part; I don't think I actually tested session resumption, so I forgot to enable it.

jay:
Thanks

PR #8475, opened 2022-02-19
Fixed BearSSL bug: error on CURLOPT_SSL_VERIFYPEER 0 and expired cert

jan2000:

The x509_minimal engine of BearSSL is used to parse the certificate chain. When verifypeer is set to 0 the result of the end_chain method (`xm_end_chain`) is overridden to return `BR_ERR_OK` so that the chain is deemed valid even if it is not. The engine is also used to extract the public key from the cert. However, because the x509_minimal engine stops parsing if any validity check fails the parsing can stop before the public key is extracted. Checking the BearSSL code in x509_minimal.t0:`decode-certificate` we can see that this can happen for example for `ERR_X509_EXPIRED`. This will mean that `x509_get_pkey` will return no key if the cert is expired and verifypeer is set to 0, what will result in `bearssl_run_until` returning `CURLE_SSL_CONNECT_ERROR`.

This is fixed by using the x509_decode engine instead of the x509_minimal engine to parse and extract the public key from the first cert of the chain.

### DIFF
--- a/lib/vtls/bearssl.c
+++ b/lib/vtls/bearssl.c
@@ -39,8 +39,10 @@
 struct x509_context {
   const br_x509_class *vtable;
   br_x509_minimal_context minimal;
+  br_x509_decoder_context decoder;
   bool verifyhost;
   bool verifypeer;
+  int cert_num;
 };
 
 struct ssl_backend_data {
@@ -260,6 +262,11 @@ static void x509_start_chain(const br_x509_class **ctx,
 {
   struct x509_context *x509 = (struct x509_context *)ctx;
 
+  if(!x509->verifypeer) {
+    x509->cert_num = 0;
+    return;
+  }
+
   if(!x509->verifyhost)
     server_name = NULL;
   x509->minimal.vtable->start_chain(&x509->minimal.vtable, server_name);
@@ -269,6 +276,13 @@ static void x509_start_cert(const br_x509_class **ctx, uint32_t length)
 {
   struct x509_context *x509 = (struct x509_context *)ctx;
 
+  if(!x509->verifypeer) {
+    /* Only decode the first cert in the chain to obtain the public key */
+    if(x509->cert_num == 0)
+      br_x509_decoder_init(&x509->decoder, NULL, NULL);
+    return;
+  }
+
   x509->minimal.vtable->start_cert(&x509->minimal.vtable, length);
 }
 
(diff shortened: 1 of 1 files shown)


lib/vtls/bearssl.c
```diff
@@ -277,35 +291,52 @@ static void x509_append(const br_x509_class **ctx, const unsigned char *buf,
...
 {
   struct x509_context *x509 = (struct x509_context *)ctx;
-  unsigned err;
 
-  err = x509->minimal.vtable->end_chain(&x509->minimal.vtable);
-  if(err && !x509->verifypeer) {
-    /* ignore any X.509 errors */
-    err = BR_ERR_OK;
+  if(!x509->verifypeer) {
+    return br_x509_decoder_last_error(&x509->decoder);
```
jay, line 320: Why did you change this to return the error in the decoder when !verifypeer?

michaelforney: BearSSL has two APIs that deal with X.509 certificates. One is `br_x509_minimal`, which decodes a chain of certificates, verifying that they are valid, each one certifies the preceding one, and that the root certificate is in the trust anchor list. At the end of the chain, if there is no other error, it returns 0 if trusted, and `BR_ERR_X509_NOT_TRUSTED` if not. In either case, you can access the public key of the end-entity using the `get_pkey` function.

However, if there was some other error (for example, a DN mismatch, where a certificate doesn't certify the preceding one, or if one of the certificates is expired), then `br_x509_minimal` just stops processing the chain, and returns an error from `end_chain`. In this case, `get_pkey` will return NULL, even if it has decoded a key.

So, if we wish to ignore errors with expired or out-of-order certificates, then `br_x509_minimal` is not useful. Instead, we can use the second API, `br_x509_decoder`, which decodes a single certificate. This commit switches the curl x509 engine to using `br_x509_decoder` if !verifypeer, and `br_x509_minimal` otherwise.

The overall effect of this change is the following:
If verifypeer is false, curl will now *only* consider the first certificate in the chain sent by the server. If it can decode the first certificate and extract a key (regardless of the contents of subsequent certificates), then it accepts the chain. Previously, curl would accept the chain *only* if it was completely valid apart from none of the certificates being trusted.

jay: Ok thanks for the thorough explanation

jay:
Thanks

PR #8476, opened 2022-02-19
Fixed BearSSL bug: EXC_BAD_ACCESS on incomplete CA cert

jan2000:

When having a incomplete CA cert (for example missing a newline at the end) in the function `load_cafile` the code after `case BR_PEM_END_OBJ:` will not be executed. This will result in an extra `ta` (`ca.anchors`) that does not have its data set. Because `ta->dn.len` is not set (thus a random value) this will result in EXC_BAD_ACCESS in BearSSL because it expects correct values. The fix is to only add the new `ta` after `case BR_PEM_END_OBJ:`.


michaelforney:
Looks good to me.

jay:
Thanks

PR #8477, opened 2022-02-19
Implemented BearSSL: CURLOPT_SSL_CIPHER_LIST

jan2000:

Implemented the CURLOPT_SSL_CIPHER_LIST option for BearSSL based on the code of SecureTransport.


jay:
You'll have to update the documentation as well. There's CIPHERS.md, CURLOPT_SSL_CIPHER_LIST.3, CURLOPT_PROXY_SSL_CIPHER_LIST.3. Also there's TLS13_cipher docs that need to be updated if you added support for TLS 1.3 ciphers

bagder:
BearSSL doesn't support TLS 1.3 at all. Which makes me wonder why people are even using this library at it seems rather limited. It also seems to have been abandoned, as there's been no commits to their git repo in over a year.

bagder (approved):

bagder:
Thanks!

Issue #8480, opened 2022-02-20
340bb19 breaks intended usage of curl_multi_assign


m6w6:
### I did this
Watched CI break with curl 7.81.

Seriously, 340bb1905f88b13655e89760ca6f4f2d02dc011a breaks the intended usage of `curl_multi_assign`.
> **Quoting [docs](https://curl.se/libcurl/c/curl_multi_socket_action.html):**  
> 
> In a typical application you allocate a struct or at least use some kind of semi-dynamic data for each socket that we must wait for action on when using the [curl_multi_socket_action](https://curl.se/libcurl/c/curl_multi_socket_action.html) approach.
>
> When our socket-callback gets called by libcurl and we get to know about yet another socket to wait for, we can use [curl_multi_assign](https://curl.se/libcurl/c/curl_multi_assign.html) to point out the particular data so that when we get updates about this same socket again, we do not have to find the struct associated with this socket by ourselves.

### I expected the following
`curl_multi_assign` to work from within callbacks.

Coincidentally, `docs/examples/hiperfifo.c` and `docs/examples/multi-event.c` are broken, too.

### curl/libcurl version
7.81/master

### operating system
POSIX

jay:
try #8483

jay:
Thanks

PR #8483, opened 2022-02-20
multi: allow user callbacks to call curl_multi_assign

jay:
Several years ago a change was made to block user callbacks from calling
back into the API when not supported (recursive calls). One of the calls
blocked was curl_multi_assign. Recently the blocking was extended to the
multi interface API, however curl_multi_assign may need to be called
from within those user callbacks (eg CURLMOPT_SOCKETFUNCTION).

I can't think of any callback where it would be unsafe to call
curl_multi_assign so I removed the restriction entirely.

Reported-by: Michael Wallner

Ref: https://github.com/curl/curl/commit/b46cfbc
Ref: https://github.com/curl/curl/commit/340bb19

Fixes https://github.com/curl/curl/issues/8480
Closes #xxxx

---

/cc @zagor

### DIFF
--- a/docs/libcurl/curl_multi_assign.3
+++ b/docs/libcurl/curl_multi_assign.3
@@ -48,6 +48,8 @@ The idea here being that this association (socket to private pointer) is
 something that just about every application that uses this API will need and
 then libcurl can just as well do it since it already has an internal hash
 table lookup for this.
+
+It is acceptable to call this function from your multi callback functions.
 .SH EXAMPLE
 .nf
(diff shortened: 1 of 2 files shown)


bagder (approved):
I think it could also be worth mentioning in the documentation as it is an exception.

jay:
> I think it could also be worth mentioning in the documentation as it is an exception.

Ok what do you think of this line in curl_multi_assign.3: `It is acceptable to call this function from your multi callback functions.`

bagder:
:+1:

PR #8486, opened 2022-02-20
docs: Converted release notes document to markdown.

vijay-jaisankar:
- Should make for better readability on Github. 
- Also, many other documentation files in this repo are in markdown, so this PR makes the whole structure more uniform.


bagder:
Thanks, I appreciate the thoughts and effort but for this particular file we also have a number of support scripts that need to be updated in sync of a format update. [scripts/release-notes.pl](https://github.com/curl/curl/blob/master/scripts/release-notes.pl) and curl-www's [dev/mk-release-notes.pl](https://github.com/curl/curl-www/blob/master/dev/mk-release-notes.pl) being the two primary ones I believe. But also scripts/delta, scripts/contributors.sh and scripts/contrithanks.sh need adjustments and Makefile.am and docs/RELEASE-PROCEDURE.md do as well.

bagder:
Also, if this should use markdown, then:
1. The changes should be written as `[the longer description][num]` where `[num]` is the number for the link mentioned in the bottom. So that linking works correctly.
2. The numerical links in the bottom need to use the format `[num]: URL`

bagder (changes requested):
Several other files need to be updated in sync

vijay-jaisankar:
Thanks for the review. I'll add commits for this soon and tag you once done.

PR #8490, opened 2022-02-21
gha: move the event-based test over from Zuul

bagder:


bagder:
I don't understand why the SSH based tests fail when moved over, but obviously I cannot complete the transition without fixing that. :cry:

pheiduck:
as you mentioned here https://github.com/curl/curl/pull/8596#pullrequestreview-909899946 this could also be applied here.

PR #8493, opened 2022-02-21
vtls: fix checking the second socket

MAntoniak:
Fix checking the second socket within associate and disassociate connection


bagder (approved):

bagder:
Thanks!

PR #8503, opened 2022-02-24
curl: add --remove-on-error

bagder:
If a transfer returns an error, using this option makes curl remove the
leftover downloded (partial) local file before exiting.

Added test 375 to verify


bagder:
This option is actually especially useful when you download N files in parallel with `-Z`, as then it is much harder to script the removal of partial leftovers.

icing:
Nit: curl manual already uses `partial` for range requests:

```
-r, --range
 ...
 Retrieve a byte range (i.e. a partial document) from an HTTP/1.1,
```

Suggestion: either change that or use a new term like `incomplete`?

bagder:
> Suggestion: either change that or use a new term like incomplete?

Ah yes, that's a fine observation...

bagder:
`--remove-leftover` could also work

icing:
Also `partial content` is used in the http spec for response code 206, https://datatracker.ietf.org/doc/html/rfc7233#section-4.1

But that is just one of many protocols, of course.

A download is only `leftover` when it is not removed, or? Seems a Schroedinger kind of problem...

bagder:
I'll wait to see what others think about names. Alternatives to ponder:

    --remove-fragment
    --remove-halfway
    --remove-incomplete
    --remove-leftover
    --remove-on-error
    --remove-partial
    --remove-unfinished

Earnestly:
`--remove-on-error` is nice and has some precdence with GNU make's [`.DELETE_ON_ERROR`](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#:~:text=.DELETE_ON_ERROR)

It also encapsulates nicely the intended consequence for its purpose, which is removal upon an error.

dfandrich:
--remove-on-error and --remove-incomplete are my two favourites.

BenBE:
My favorites are `--remove-on-error`, `--remove-incomplete`, and `--remove-partial`. I think the `--remove-on-error` encapsulates its intent best. The `--remove-incomplete` kinda works too, but is less clear about what means "incomplete" for a download (i.e. could mean if only a part of the file was requested or sth like that). Finally with `--remove-partial`, while this would have been my first go-to suggestion you could cause some confusion if you perform a partial download with `--range` just to `--remove-partial` afterwards. :wink:

That aside: With `rsync` the flag for keeping partially transferred files is `--partial`, thus `--no-partial` could be considered a viable option too. :stuck_out_tongue_winking_eye:

jay:
> If a transfer returns an error,

--remove-on-error

bch:
On Thu, Feb 24, 2022 at 10:53 Jay Satiro ***@***.***> wrote:

> If a transfer returns an error,
>
> --remove-on-error
>

i concur

-bch

bagder:
`--remove-on-error` it is now!

docs/cmdline-opts/remove-on-error.d
```diff
@@ -0,0 +1,12 @@
+Long: remove-on-error
+Help: Remove output file on errors
+See-also: fail
+Category: curl
+Example: --remove-on-error -o output $URL
+Added: 7.83.0
+---
+When curl returns an error when told to save output in a local file, this
+option removes that saved file before exiting. This prevevents curl from
```
colinleroy, line 9: Prevents ;)

bagder: yikes, thanks!

Issue #7351, opened 2021-07-06
Downloading with HTTP/3 produces broken files


vl409:
Downloading ~10Mb file at 10Gbit link produces broken download (garbage within content, see attachment:[bug-data.tar.gz](https://github.com/curl/curl/files/6768587/bug-data.tar.gz))
Downloading same file from same server  using same curl binary over 100Mbit link is fine. 
Downloading same file from same server using ngtcp2 client is fine.
The issue is reproducible. 

The attachment contains:

- digits.good  - original file, plaintext with digits pattern
- digits.broken  - result of downloading with curl: 
``
export SSLKEYLOGFILE=ssl_key_file
./bin/curl -k --http3 -o digits.broken  https://ip:port/digits
``
- ssl_key_file  - ssl keys produces by curl
- capture.pcapng  - captured packets on client side.
- stream-from-wireshark - head of decoded quic stream from wireshark; demonstrates that  captured stream is not broken (compare to digits.broken)

Versions:

curl from git, 0a0bc4a0772826262bcae4ab2627fc28b3345660
``
./bin/curl -V
curl 7.78.0-DEV (x86_64-pc-linux-gnu) libcurl/7.78.0-DEV OpenSSL/3.0.0 zlib/1.2.11 brotli/1.0.9 zstd/1.4.9 libidn2/2.3.1 libpsl/0.21.0 (+libidn2/2.3.0) nghttp2/1.41.0 ngtcp2/0.1.0-DEV nghttp3/0.1.0-DEV libgsasl/1.10.0 OpenLDAP/2.4.57
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB PSL SSL TLS-SRP UnixSockets zstd
``

ngtctp2 from git,  f60c3fb447e0b4605a99fa8cd9226330aae9014a
quic-tls from git, 50a5c6fec45aed8b49e28fd23aeaf590d39becc4
nghttp3 from git, bbc426ecdbc119fd7f12a17850f64c440844af01

curl is configured using the command:
``
    cd curl
    autoreconf -fi
[...]

PR #8504, opened 2022-02-24
ngtcp2: Reset dynbuf when it is fully drained

tatsuhiro-t:
See https://github.com/curl/curl/issues/7351


bagder:
What a lovely green :heavy_check_mark:

bagder:
Thanks!

PR #8505, opened 2022-02-24
Avoid unused parameter error in Curl_resolv_check

sgebbie:
When built without DNS-over-HTTP and without asynchronous resolvers,
neither the dns nor the data parameters are used.

That is Curl_resolv_check appears to call
Curl_resolver_is_resolved(data, dns). But,
with CURL_DISABLE_DOH without CURLRES_ASYNCH, the call is actually
elided via a macro definition.

This fix resolves the resultant: "unused parameter 'data'" error.


bagder (approved):

bagder:
Thanks!

sgebbie:
> Thanks!

@bagder - Thanks for the quick review and fast turn around. Much appreciated.

PR #8511, opened 2022-02-25
libssh: #ifdef include files accordingly

bagder:
Reported-by: 梦终无痕
Bug: https://curl.se/mail/lib-2022-02/0131.html

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -219,6 +219,13 @@ stages:
           prepare: pacman -S --needed --noconfirm --noprogressbar libssh2-devel mingw-w64-x86_64-libssh2
           configure: --host=x86_64-w64-mingw32 --build=x86_64-w64-mingw32 --prefix=/mingw64 --enable-debug --enable-werror --with-libssh2 --with-openssl
           tests: "~571"
+        msys2_mingw64_debug_libssh:
+          name: 64-bit OpenSSL/libssh
+          container_img: ghcr.io/mback2k/curl-docker-winbuildenv/msys2-mingw64:ltsc2019
+          container_cmd: C:\msys64\usr\bin\sh
+          prepare: pacman -S --needed --noconfirm --noprogressbar libssh-devel mingw-w64-x86_64-libssh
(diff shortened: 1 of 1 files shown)


illusory-dream:
S_IFLNK and S_IFMT will be undefined variable without including unistd.h. Could it work as expecting if we define them manually.

bagder:
> Could it work as expecting if we define them manually.

We can define them manually, sure. But isn't there a header file you can include that has them defined? Like maybe `windows.h` for example?

bagder:
This fix seems to build on Windows now, at least with mingw

lib/vssh/libssh.c
```diff
@@ -81,18 +81,33 @@
...
+#ifdef HAVE_FCNTL_H
 #include <fcntl.h>
+#endif
+
+#ifndef S_IFMT
+#define S_IFMT  00170000
+#endif
+#ifndef S_IFLNK
+#define S_IFLNK  0120000
+#endif
```
jay, line 105: I don't think these should be defined, they may have different values on different platforms. For example my CRT uses 0xF000 for IFMT mask and there is no IFLNK file type. Also the code these are used for doesn't make sense to me:

~~~c
          if((sshc->readdir_attrs->flags & SSH_FILEXFER_ATTR_PERMISSIONS) &&
             ((sshc->readdir_attrs->permissions & S_IFMT) ==
              S_IFLNK)) {
~~~
So these are remote permissions? What if the client has IFMT, IFLNK values different from the server?

/cc @nmav

bagder: Ah, I found these in liibssh/sftp.h:
~~~
#define SSH_S_IFMT   00170000
#define SSH_S_IFLNK  0120000

gvanem:
But I noted that `WSAStartup()` is now called twice:
  * In libcurl's `Curl_win32_init()`.
  * And libssh's `ssh_socket_init()`.

Any way to avoid that?

bagder:
It seems tricky because libssh has no apparent way to inhibit that. `ssh_socket_init()` is called from within `ssh_init()` but there's no way for us to tell it not to.

jay:
> But I noted that `WSAStartup()` is now called twice:

that's fine as long as there's a corresponding cleanup call. multiple startup calls increment an internal counter like curl_global_init.

PR #6064, opened 2020-10-12
win32: disable output buffering since line-buffering is broken

mback2k:
Related to the missing/incomplete output issues being troubleshooted with #6049 and #5904.

PR #8516, opened 2022-02-26
tool and tests: force flush of all buffers at end of program

mback2k:
On Windows data can be lost in buffers in case of abnormal program
termination, especially in process chains as seen due to flaky tests.
Therefore flushing all buffers manually should avoid this data loss.

In the curl tool we play the safe game by only flushing write buffers,
but in the testsuite where we manage all buffers, we flush everything.

This should drastically reduce the CI and testsuite flakiness.

Supersedes https://github.com/curl/curl/pull/7833 and https://github.com/curl/curl/pull/6064

### DIFF
--- a/src/tool_main.c
+++ b/src/tool_main.c
@@ -277,6 +277,11 @@ int main(int argc, char *argv[])
     main_free(&global);
   }
 
+#ifdef WIN32
+  /* Flush buffers of all streams opened in write or update mode */
+  fflush(NULL);
+#endif
+
 #ifdef __NOVELL_LIBC__
   if(!getenv("_IN_NETWARE_BASH_"))
     tool_pressanykey();
--- a/tests/libtest/first.c
+++ b/tests/libtest/first.c
@@ -179,5 +179,10 @@ int main(int argc, char **argv)
     PR_Cleanup();
 #endif
 
+#ifdef WIN32
+  /* flush buffers of all streams regardless of mode */
+  _flushall();
+#endif
+
   return result;
 }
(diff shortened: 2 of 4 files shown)


mback2k:
No flakiness observed after 6 Azure CI runs. @bagder @jay

bagder (approved):

jay:
I don't understand why this works. If the program is aborting then these flush calls shouldn't matter. Is it possible it is a coincidence that it appears to work? If a program exits normally (return from main) then the files should be flushed and closed without these additional calls.

mback2k:
> I don't understand why this works. If the program is aborting then these flush calls shouldn't matter. Is it possible it is a coincidence that it appears to work? If a program exits normally (return from main) then the files should be flushed and closed without these additional calls.

Me neither. I think it most likely works by avoiding data getting lost in process / output forwarding chains. Remember, we are intermixing a couple of different more Unix-based technologies in the testsuite even if it is running under msys. So we have at least the native Win32 binaries, msys-based Perl and Shell, with a couple of scripts in each of those interpreters being intertwined and nested. My theory is that sometimes somewhere in those process chains, the output data get's lost in a buffer.

mback2k:
@jay I would like to land this for now and observe the long term statistics on Azure Pipelines after ~30 days, what do you think?

jay:
ok

mback2k:
The intermediate picture looks quite good so far, not a single test failure on Azure CI since merging this PR:
image
Source: https://dev.azure.com/daniel0244/curl/_test/analytics?definitionId=1&contextType=build filtered to **master** branch.

Issue #8361, opened 2022-01-31
[Windows] Long-filename problem


gvanem:
There is an issue with access to filenames with >= 260 characters in the Windows version of curl.
I used this `<curl_root>\upload-long-file.bat` file to verify:
```
@echo off
setlocal
::
:: File to upload: 261 bytes long (446 bytes) [1]
::
set LONG_FILE=f:\MingW32\src\GUI\Harlinn\2022.01.11\Tests\3rdParty\gdal\data\gdrivers\sentinel2\fake_l2a_MSIL2Ap\S2A_MSIL2A_20170823T094031_N0205_R036_T34VFJ_20170823T094252.SAFE\GRANULE\L2A_T34VFJ_A011330_20170823T094252\IMG_DATA\R60m\L2A_T34VFJ_20170823T094031_AOT_60m.jp2

::
:: A copy of the above, but 260 bytes long
::
set SHORT_FILE=f:\MingW32\src\GUI\Harlinn\2022.01.11\Tests\3rdParty\gdal\data\gdrivers\sentinel2\fake_l2a_MSIL2Ap\S2A_MSIL2A_20170823T094031_N0205_R036_T34VFJ_20170823T094252.SAFE\GRANULE\L2A_T34VFJ_A011330_20170823T094252\IMG_DATA\R60m\L2A_T34VFJ_20170823T094031_AOT_60m.jp

set URL=ftp://ftp.xxx.yyy.com

if %1. == long. (
  echo on
  %~dp0\src\curl.exe -vT %LONG_FILE% %URL%
) else (
  echo on
  %~dp0\src\curl.exe -vT %SHORT_FILE% %URL%
)
```

With a `upload-long-file.bat long`, curl fails to open the `%LONG_FILE`:
```
curl: Can't open 'f:\MingW32\src\GUI\Harlinn\2022.01.11\Tests\3rdParty\gdal\data\gdrivers\sentinel2\fake_l2a_MSIL2Ap\
S2A_MSIL2A_20170823T094031_N0205_R036_T34VFJ_20170823T094252.SAFE\GRANULE\
L2A_T34VFJ_A011330_20170823T094252\IMG_DATA\R60m\L2A_T34VFJ_20170823T094031_AOT_60m.jp2'!
curl: try 'curl --help' or 'curl --manual' for more information
curl: (26) Failed to open/read local data from file/application
```

[...]

PR #8518, opened 2022-02-27
tool_doswin: Prepare for long path support in Windows 10

jay:
- Remove PATH_MAX (MAX_PATH, 260) limitations where appropriate.

Some PATH_MAX usage remains for MS-DOS specific code.

Windows 10 1607 and later supports long paths (ie paths longer than
MAX_PATH, up to 32k) if the user opts in via LongPathsEnabled registry
key and the application supports it (ie app code not limited by MAX_PATH
and app manifest contains attribute ws2:longPathAware).

This change addresses all the use of PATH_MAX in the curl tool that
would prevent adding ws2:longPathAware to the manifest. However, it does
not actually add the attribute. That would have to be a separate change
for each build system to create/amend a manifest with that attribute.

Prior to this change long paths could be accessed only via the literal
path prefix `\\?\` (eg `\\?\C:\reallylongpath`) which bypassed some MAX_PATH
checks in both the curl tool and the Windows API. That can still be used
as a workaround with the curl tool in all versions of Windows, though
the behavior is not guaranteed.

It's my opinion that, in general, long path support is not guaranteed
to work properly 100% of the time even if we make all the appropriate
changes. Refer to the bug discussion for more information.

Ref: https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

Bug: https://github.com/curl/curl/issues/8361
Reported-by: Gisle Vanem

Closes #xxxx


bagder:
What's the status for this?

jay:
> What's the status for this?

I was kind of on the fence about this and so it went stale. I will take another look.

bagder:
Should we just close this? It's not really a problem that affects users much.

jay:
Closing in favor of #9288 which adds the issue to known bugs.

PR #8524, opened 2022-02-28
make Curl_getconnectinfo work with connection cache from share handle

meierfra-ergon:
I experienced that curl_easy_getinfo() with CURLINFO_ACTIVESOCKET does not return a socket, if I use a easy_handle with a share_handle set (curl_easy_setopt() with CURLOPT_SHARE).
The reason is that in the curl function Curl_getconnectinfo() does not check if the connection cache from the share_handle was used.
I propose the following pull request to address this issue.

BTW: A long time ago [1], I also proposed a patch in the same function to not only retrieve the last used connection of an easy_handle but also the currently active connection. If the above PR is acceptable I could also create a PR with this extended behavior. We use this for years now ;-).

[1] https://curl.se/mail/lib-2015-08/0116.html

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -74,6 +74,7 @@
 #include "warnless.h"
 #include "conncache.h"
 #include "multihandle.h"
+#include "share.h"
 #include "version_win32.h"
 #include "quic.h"
 #include "socks.h"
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

Cmo1976:
@

Cmo1976:
Thank you for the help

PR #8525, opened 2022-02-28
rtsp: don't let CSeq error override earlier errors

jay:
- When done, if an error has already occurred then don't check the
  sequence numbers for mismatch.

A sequence number may not have been received if an error occurred.

Prior to this change a sequence mismatch error would override earlier
errors. For example, a server that returns nothing would cause error
CURLE_GOT_NOTHING in Curl_http_done which was then overridden by
CURLE_RTSP_CSEQ_ERROR in rtsp_done.

Closes #xxxx

---

Before:
~~~
* multi_done: status: 0 prem: 0 done: 0
* Empty reply from server
* The CSseq of this request 1 did not match the response 0
* Closing connection 0
curl: (85) Empty reply from server
~~~

After:
~~~
* multi_done: status: 0 prem: 0 done: 0
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
~~~

### DIFF
--- a/lib/rtsp.c
+++ b/lib/rtsp.c
@@ -219,7 +219,7 @@ static CURLcode rtsp_done(struct Curl_easy *data,
 
   httpStatus = Curl_http_done(data, status, premature);
 
-  if(rtsp) {
+  if(rtsp && !status && !httpStatus) {
     /* Check the sequence numbers */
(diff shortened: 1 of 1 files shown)


jay:
I can't think of anything going wrong with this but also I don't use the RTSP feature so one more call for comments before I land it.

Issue #8112, opened 2021-12-07
Windows: '--local-port range' is unable to detect ports in TIME_WAIT, and so does not reliably try further ports in the range


gclinch:
### I did this
On a Windows client:

The first run succeeds:
```
> curl -v --local-port 600-605 http://github.com/
*   Trying 140.82.121.4:80...
* Local port: 600
* Connected to github.com (140.82.121.4) port 80 (#0)
[...]
```

The same command a second time fails:
```
> curl -v --local-port 600-605 http://github.com/
*   Trying 140.82.121.4:80...
* Local port: 600
* connect to 140.82.121.4 port 80 failed: Address already in use
* Failed to connect to github.com port 80 after 13 ms: Address already in use
* Closing connection 0
curl: (7) Failed to connect to github.com port 80 after 13 ms: Address already in use
[...]
```

### I expected the following
On a Linux client, both runs succeed:

First, port 600 is used:
```
# curl -v --local-port 600-605 http://github.com/
*   Trying 140.82.121.4...
* TCP_NODELAY set
* Local port: 600
* Connected to github.com (140.82.121.4) port 80 (#0)
[...]
```

And on the second execution, port 601 is used:
```
# curl -v --local-port 600-605 http://github.com/
*   Trying 140.82.121.4...
* TCP_NODELAY set
* Bind to local port 600 failed, trying next
* Local port: 601
* Connected to github.com (140.82.121.4) port 80 (#0)
[...]
```

It appears that given an already in-use port number (as seen through netstat / ss), Windows bind will succeed (then failing later on the connect call), whilst Linux bind will not succeed.  Curl's localport "try next" logic considers only whether the bind call succeeds.

### curl/libcurl version

```
curl 7.80.0 (x86_64-pc-win32) libcurl/7.80.0 OpenSSL/3.0.0 (Schannel) zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.2 libssh2/1.10.0 nghttp2/1.46.0 libgsasl/1.10.0
Release-Date: 2021-11-10
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
[...]

PR #8528, opened 2022-03-01
connect: improve binding to local ports on Windows

inPhraZ:
There is a configuration parameter in windows registry
called TcpTimedWaitDelay that determines the time that a
connection stays in the TIME_WAIT state when it is closing.
It's minimum and default value is 30 seconds.
As long as a connection is in the TIME_WAIT state, the socket pair
cannot be re-used, BUT bind() still returns 0 in this case.
curl now can detect the state of a given port using
GetTcpTable2() for IPv4 connections and avoids to bind that port
if the state is TIME_WAIT

ref: https://docs.microsoft.com/en-us/troubleshoot/windows-client/networking/tcpip-and-nbt-configuration-parameters#optional-tcpip-parameters-that-you-can-configure-by-using-registry-editor

Fixes: https://github.com/curl/curl/issues/8112
Reported-by: gclinch on github

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -926,6 +926,7 @@ if(NOT UNIX)
   check_include_file_concat("ws2tcpip.h"     HAVE_WS2TCPIP_H)
   check_include_file_concat("winsock2.h"     HAVE_WINSOCK2_H)
   check_include_file_concat("wincrypt.h"     HAVE_WINCRYPT_H)
+  check_include_file_concat("iphlpapi.h"     HAVE_IPHLPAPI_H)
 endif()
 
 check_include_file_concat("stdio.h"          HAVE_STDIO_H)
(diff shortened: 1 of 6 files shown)


jay:
This is an interesting solution and I am very interested in what others will think of it. My initial thoughts are:

I think it would probably have some cost to enumerate. The bind address would have to be checked as well as the port. IPv4 / IPv6 (af == AF_INET6) differences, the latter needs IPv6 enumeration check. Also how reliable is it if the port is in some other state and the check passes because it's not timewait and then on connect same problem?

bagder:
I need more guidance and feedback on this proposal from people using and knowing Windows.

bagder:
This improvement proposal has not received much interest and I cannot deem the necessity myself. As such, I cannot merge this. Sorry.

Issue #7606, opened 2021-08-21
Configure fails to identify OpenSSL 3.0.0 headers


rsbeckerca:
When configuring with OpenSSL 3.0.0 beta2, configure is unable to recognize the OpenSSL version. The library version is also incorrectly identified. 1.1.1 is not in the library path nor in /usr/local/lib. 3.0.0 is the version reported by the library in /usr/local-ssl3.0/lib.

### I did this
```
./configure with OpenSSL 3.0.0
... snip ...
configure: Added /usr/local-ssl3.0/lib to CURL_LIBRARY_PATH
checking for OpenSSL headers version... unknown - ( (3<<28) |(0<<20) |(0<<4) |0x0L )
checking for OpenSSL library version... 1.1.1
checking for OpenSSL headers and library versions matching... fail
configure: WARNING: Can not compare OpenSSL headers and library versions.
```
### I expected the following
```
configure: Added /usr/local-ssl3.0/lib to CURL_LIBRARY_PATH
checking for OpenSSL headers version... 3.0.0
checking for OpenSSL headers and library versions matching... matched
```
### curl/libcurl version
7.77.0

### operating system
NONSTOP_KERNEL myhost L20 10 NSX-D NSX NSX-D NonStop Kernel

rsbeckerca:
configure seems to be identifying version 3.0, but does not look like it is acting correctly:
```
configure:24521: checking for OpenSSL >= v3
configure:24543: cc -c -c99 -D_XOPEN_SOURCE_EXTENDED=1 -Wlp64 -WIEEE_float -I/us
r/local-ssl3.0/openssl/include -I/usr/local-ssl3.0/include -I/usr/local-ssl3.0/i
nclude/openssl conftest.c >&5
   return 0;
   ^
"/home/ituglib/randall/jenkins/.jenkins/workspace/Curl_3.0_Pipeline/conftest.c",
 line 61: warning(203):
          statement is unreachable

configure:24543: $? = 0
configure:24545: result: yes
```

bagder:
Why do you provide a lot of custom flags to configure? What happens if you just use plain configure as intended?

Also, configure  from 7.77.0 will not build correctly with current OpenSSL built from their git due to how they've changed. Fixed in 1e5e93d899e. I don't know if beta2 has that change or if that was done later.

We build with OpenSSL3 (from their git master) in the CI, it works. As I told on the mailing list, I build frequently and I did it from source just today. On Linux x86-64.

[...]

PR #7503, opened 2021-07-26
curl-openssl.m4: check lib64 for the pkg-config file

bagder:
OpenSSL recently started putting the libs in $prefix/lib64 on 'make
install', so we check that directory for pkg-config data if the
'lib' check fails.

PR #8536, opened 2022-03-04
curl-openssl.m4: prefer lib64 over lib

jay:
Prior to this change if the user had two versions of OpenSSL installed
in the same folder (eg /usr/local/ssl) then the possibly older OpenSSL
in lib was preferred over OpenSSL 3 in lib64.

Ref: https://github.com/curl/curl/issues/7606
Ref: https://github.com/curl/curl/pull/7503

Closes #xxxx

---

I'm not sure about this one. I don't understand why on some systems OpenSSL 3 is using lib64 and others it's using just lib. I have OpenSSL 1.1.1 installed in /usr/local/ssl and the libraries are in lib. When I installed OpenSSL 3.0.1 in /usr/local/ssl using the same commands it put the libraries in lib64. Since only the most recent OpenSSL version may use lib64 by default I would think prefer that over lib?  This is in Ubuntu where there isn't usually (or ever?) lib64 and $libsuff is empty.

### DIFF
--- a/m4/curl-openssl.m4
+++ b/m4/curl-openssl.m4
@@ -71,16 +71,17 @@ if test "x$OPT_OPENSSL" != xno; then
 
     dnl Try pkg-config even when cross-compiling.  Since we
     dnl specify PKG_CONFIG_LIBDIR we're only looking where
-    dnl the user told us to look
-    OPENSSL_PCDIR="$OPT_OPENSSL/lib/pkgconfig"
+    dnl the user told us to look.
+
+    dnl OpenSSL 3 may use lib64 instead of lib like older versions
+    OPENSSL_PCDIR="$OPT_OPENSSL/lib64/pkgconfig"
     if test -f "$OPENSSL_PCDIR/openssl.pc"; then
       AC_MSG_NOTICE([PKG_CONFIG_LIBDIR will be set to "$OPENSSL_PCDIR"])
       PKGTEST="yes"
     fi
 
     if test "$PKGTEST" != "yes"; then
-      # try lib64 instead
-      OPENSSL_PCDIR="$OPT_OPENSSL/lib64/pkgconfig"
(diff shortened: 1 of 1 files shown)


bagder (approved):

jay:
It looks like they [changed](https://github.com/openssl/openssl/issues/16244) OpenSSL 3.0.0 to create a multilib directory even if it doesn't already exist, so the lib files could be in lib64, lib32, lib-x64, lib/x64, lib/64 etc now where as before it would be just lib. They have a --libdir=lib that can be used for the old behavior.

Probably the change I made is not technically correct, but $libsuff is empty for me in Ubuntu x64. Either way I think we are wrong now. We can't use pkgconfig anymore on cross compile (at least I don't think) because we could end up with the wrong OpenSSL. To truly be correct here could require copying the logic OpenSSL uses to generate the suffix, but that seems like a lot of work.

bagder:
> because we could end up with the wrong OpenSSL

I think it works now too because most users only have a single OpenSSL installed. In fact, if you have two OpenSSLs in the same prefix, it's not even clear what curl *should* do.

jay:
OpenSSL 3.0.0 overwrote the 1.1.1 installation except that it placed the libs in /usr/local/ssl/lib64 and so the 1.1.1 libs are still in /usr/local/ssl/lib. I think that is going to surprise a lot of people. I'll use --libdir=lib from now on.

I'm going to think of a better way to handle this here, maybe a dirty check host string for 64 then try lib64, if not try lib32. That will cover some not all. You can `git grep multilib` from ossl master to see all the possibilities.

bagder:
What's the status on this now?

jay:
> What's the status on this now?

Dead for now. Preferring lib64 over lib may have unintended consequences and I never got around to seeing if I could emulate the OpenSSL 3 code.

AFAICT this is only an issue for users that have artifacts of old OpenSSL installations in their OpenSSL path.

Issue #8538, opened 2022-03-04
[Win32] crash in unit1610.exe


gvanem:
Compiling and running `unit1610.exe www.curl.se`, caused a crash inside OpenSSL;
calling a func-ptr `c_set_error_mark` which is NULL:
```
0x0
unit1610!ERR_set_mark(void)+0x8
unit1610!CONF_modules_load_file_ex(struct ossl_lib_ctx_st * libctx = 0x00000000, char * filename = 0x00000000 "", char * appname = 0x00000000 "", unsigned long flags = 0x32)+0xb6
unit1610!CONF_modules_load_file(char * filename = 0x00000000 "", char * appname = 0x00000000 "", unsigned long flags = 0x32)+0x13
unit1610!ossl_config_int(struct ossl_init_settings_st * settings = 0x00000000)+0x32
unit1610!ossl_init_config_ossl_(void)+0x7
unit1610!CRYPTO_THREAD_run_once(long * once = 0x00ea90d8, <function> * init = 0x009fbc40)+0x36
unit1610!OPENSSL_init_crypto(unsigned int64 opts = 0x40, struct ossl_init_settings_st * settings = 0x00000000)+0x314
unit1610!ossl_engine_table_select(struct st_engine_table ** table = 0x00ea911c, int nid = 0x2a0, char * f = 0x00d61d60 "crypto/engine/tb_digest.c", int l = 0x45)+0x17
unit1610!ENGINE_get_digest_engine(int nid = 0x2a0)+0x15
unit1610!evp_md_init_internal(struct evp_md_ctx_st * ctx = 0x047a0738, struct evp_md_st * type = 0x00d427d8, struct ossl_param_st * params = 0x00000000, struct engine_st * impl = 0x00000000)+0x157
unit1610!EVP_DigestInit_ex(struct evp_md_ctx_st * ctx = 0x047a0738, struct evp_md_st * type = 0x00d427d8, struct engine_st * impl = 0x00000000)+0x13
unit1610!Curl_sha256it(unsigned char * output = 0x0072f8f0 "???", unsigned char * input = 0x0072f8dc "1", unsigned int length = 1)+0x1f
unit1610!test(char * arg = 0x0c7b19d5 "www.curl.se")+0x5a
...
```

Also some context is not set correctly and the `filename == NULL`. But AFAICS, OpenSSL would then use some 
default `openssl.conf` file (?)

[...]

PR #8540, opened 2022-03-04
unit1610: init SSL library before calling SHA256 functions

jay:
The SSL library must be initialized (via global initialization) because
libcurl's SHA256 functions may call SHA256 functions in the SSL library.

Reported-by: Gisle Vanem

Fixes https://github.com/curl/curl/issues/8538
Closes #xxxx

### DIFF
--- a/tests/unit/unit1610.c
+++ b/tests/unit/unit1610.c
@@ -25,12 +25,14 @@
 
 static CURLcode unit_setup(void)
 {
-  return CURLE_OK;
+  int res = CURLE_OK;
+  global_init(CURL_GLOBAL_ALL);
+  return res;
 }
 
 static void unit_stop(void)
 {
-
+  curl_global_cleanup();
 }
 
 UNITTEST_START


bagder (approved):

bagder:
(The hyper builds seem to permafail since a while back, unrelated to this change of course)

gvanem:
There's a similar issue in `unit1620.exe` giving `Error: resolver_init failed` on Windows.
So why not like this in both `unit16*.c` files:
```diff
--- a/unit1620.c 2022-03-04 06:19:29
+++ b/unit1620.c 2022-03-05 15:02:50
@@ -28,7 +28,7 @@

 static CURLcode unit_setup(void)
 {
-  return CURLE_OK;
+  return curl_global_init(CURL_GLOBAL_ALL);
 }
```

jay:
ok but unit_setup uses the global_init macro which requires an int res variable like in this PR. I will check on why global init is necessary here but I'd guess it's winsock needs initialization for the resolver to work. which resolver are you using?

gvanem:
> which resolver are you using?

C-Ares.

Issue #8276, opened 2022-01-13
-pthread appended to CFLAGS even though the compiler is IBM XLC


dcassioli:
https://github.com/curl/curl/blob/master/configure.ac#:~:text=AC_CHECK_FUNC(pthread_create%2C%20%5BUSE_THREADS_POSIX,PTHREAD%3D%22%20%2Dpthread%22

In configure.ac (and the generated configure script) the -pthread flag gets appended to CFLAGS even in the case of IBM XLC compiler. The issue was experienced in AIX 7.1 with XLC version 11.1.

The -pthread flag has no meaning in XLC, and it's probably getting interpreted as -p (enable profiling) -t (list of toolchain programs to prepend the prefix specified by the -B flag to, i.e., 'a' for 'assembler', 'l' for 'linker', etc.), and then 'hread' is parsed as argument for -t (which is garbage, so the flag is simply discarded with a warning).

In discussion: https://curl.se/mail/lib-2019-03/0049.html a possible fix is proposed.

bagder:
I think it would make sense if you would write up a PR for the change you think makes sense for you on that platform. We don't have many AIX users or developers around.

dcassioli:
> The patch back then seems to suggest to check for `*-ibm-aix*` and treat it like hpux is already treated in there: it simply skips adding a command line option but then also disables threading (and the threaded resolver). It's a very crude "fix".

You are right, of course, I'm aware it's not a real fix but it allows a user to pass the -qthreaded flag to C/CPPFLAGS manually (not much of an improvement, I admit) and, most of all, it doesn't backfire and result in having profiling enabled.

> 1.Isn't there a chance that gcc/clang can run on AIX and have such an option?
I've found some literature about gcc working on AIX, but it's not supported by IBM and it feels like a "hack". I'd rather have XLC fully supported (apart from this little issue, curl is building fine).

> 2. Don't you rather want the corresponding xlc take used so that you can use the threaded resolver?
I'd love to. The main reason why I opened this issue was to (indirectly) ask whether some patch was already being implemented; if you don't mind, I'm going to test a possible solution and submit it as a PR.

Thank you for your patience and your quick reply.

bagder:
> I've found some literature about gcc working on AIX, but it's not supported by IBM and it feels like a "hack". I'd rather have XLC fully supported (apart from this little issue, curl is building fine).

[...]

PR #8542, opened 2022-03-05
configure.ac: move -pthread CFLAGS setting back where it used to be

asedeno:
The fix for #8276 proposed in #8374 set `CFLAGS="$CFLAGS -pthead"` earlier than it used to be set, applying it in cases where it should not have been applied.

This moves the AIX XLC check to a new `case $host in` block inside of the `if test "$USE_THREADS_POSIX" != "1"` block, where `CFLAGS="$CFLAGS -pthead"` used to happen.

I'd like to see this tested on AIX before it gets merged.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3522,23 +3522,27 @@ if test "$want_pthreads" != "no"; then
          dnl it doesn't actually work without -lpthread
          USE_THREADS_POSIX=""
          ;;
-      *-ibm-aix*)
-         dnl Check if compiler is xlC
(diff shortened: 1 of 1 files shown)


asedeno:
x-ref: 8541

asedeno:
@badger, if you think this looks good, I think we should merge this now. I'm willing to do any follow-up work needed if this doesn't work as expected for AIX XLC.

bagder:
Thanks!

dcassioli:
@asedeno I've just verified that it's working on AIX. Sorry for taking so long.

PR #4576, opened 2019-11-09
include: fixed build by MCST lcc compiler

r-a-sattarov:
Fixed build by MCST lcc compiler on MCST Elbrus 2000 architecture and do some code cleanup.

e2k (Elbrus 2000) - this is VLIW/EPIC architecture, like Intel Itanium architecture.

About this architecture:
- https://en.wikipedia.org/wiki/Elbrus_2000

PR #8546, opened 2022-03-05
include: updated definition macro for MCST-LCC compiler

r-a-sattarov:
in mcst-lcc compiler => 1.25 added a new macro definition to determine compiler

enhancement of the previous patch - https://github.com/curl/curl/pull/4576

### DIFF
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -121,7 +121,7 @@
 #  define CURL_TYPEOF_CURL_SOCKLEN_T int
 
 #elif defined(__LCC__)
-#  if defined(__e2k__) /* MCST eLbrus C Compiler */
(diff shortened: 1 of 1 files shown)


bagder:
> in mcst-lcc compiler => 1.25 added a new macro definition to determine compiler

And we can assume most lcc users to have that version and nobody still uses the old?

r-a-sattarov:
> > in mcst-lcc compiler => 1.25 added a new macro definition to determine compiler
> 
> And we can assume most lcc users to have that version and nobody still uses the old?

Yes, the current version is 1.25 (version 1.27 is in development), and version 1.23 is no longer supported (and almost not used).

bagder (approved):

bagder:
Thanks!

PR #8548, opened 2022-03-06
runtests: make 'oldlibssh' be before 0.9.4

pghmcfc:
The 'oldlibssh' feature indicates that the error code returned by libssh for a broken known_hosts file should be 67 rather than 60 (test1459). This feature was added as part of #8444 with 'oldlibssh' mapping to libssh versions prior to 0.9.6, and then refined as part of #8511 to map to versions prior to 0.9.5.

In Red Hat Enterprise Linux 8.5 there is a patched version of libssh version 0.9.4 (https://git.centos.org/rpms/libssh/blob/c8/f/SOURCES) in which test1459 fails because it returns the "new" value rather than the "old" one. It's plausible that one of the patches is responsible for this rather than the underlying code but I don't think so.

This change therefore drops the 'oldlibssh' version check to map to libssh versions older than 0.9.4, which fixes builds on RHEL-8.

### DIFF
--- a/tests/FILEFORMAT.md
+++ b/tests/FILEFORMAT.md
@@ -395,7 +395,7 @@ Features testable here are:
 - `ld_preload`
 - `libssh2`
 - `libssh`
-- `oldlibssh` (versions before 0.9.6)
+- `oldlibssh` (versions before 0.9.4)
 - `libz`
 - `manual`
 - `Mime`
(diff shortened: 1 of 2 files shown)


bagder (approved):
I thought I had a libssh 0.9.4 install somewhere show up as "old" but now when I look for it I can't find it so I'll go with your move here until we find a reason to think otherwise.

bagder:
Thanks!

PR #8556, opened 2022-03-07
tls: enable CURLOPT_SSL_EC_CURVES with BoringSSL

### LINKED ISSUE #8553: CURLOPT_SSL_EC_CURVES and '--curves' ignored with BoringSSL
### I did this

* Compiled curl with BoringSSL as per [this guide](https://everything.curl.dev/source/build/tls/boringssl)
* Ran curl with `curl --curves X25519:P-256:P-384:P-521 ...`

### I expected the following

Expected the correct curves to be set in the TLS client hello message. However, the curves are not set. This is from Wireshark:
```
Extension: supported_groups (len=8)
    Type: supported_groups (10)
    Length: 8
    Supported Groups List Length: 6
    Supported Groups (3 groups)
        Supported Group: x25519 (0x001d)
        Supported Group: secp256r1 (0x0017)
        Supported Group: secp384r1 (0x0018)
```
(Notice that `secp521r1` is missing).

The reason is this code in `lib/vtls/openssl.c`:
```c
#if ((OPENSSL_VERSION_NUMBER >= 0x10101000L) && \
     !defined(LIBRESSL_VERSION_NUMBER) &&       \
     !defined(OPENSSL_IS_BORINGSSL))
#define HAVE_SSL_CTX_SET_CIPHERSUITES
#define HAVE_SSL_CTX_SET_POST_HANDSHAKE_AUTH
/* SET_EC_CURVES is available under the same preconditions: see
 * https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set1_groups.html
 */
#define HAVE_SSL_CTX_SET_EC_CURVES
#endif
```

However, BoringSSL does supply the `SSL_CTX_set1_curves_list()` function, here's the [documentation](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set1_curves_list). It was added in [this commit](https://boringssl.googlesource.com/boringssl/+/5fd1807d95f75895ae99e336ac21b422f3cc6bd3).

The fix is simple and I have a patch already. I was wondering whether there is a specific reason it was disabled with BoringSSL.

### curl/libcurl version

```
curl 7.82.0 (x86_64-pc-linux-gnu) libcurl/7.82.0 BoringSSL zlib/1.2.11 brotli/1.0.7 nghttp2/1.40.0
Release-Date: 2022-03-05
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL UnixSockets
```

### operating system

Linux ThinkPad 5.10.0-1055-oem #58-Ubuntu SMP Thu Jan 6 20:44:40 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

lwthiker:
The CURLOPT_SSL_EC_CURVES option (used by the '--curves' flag)
in libcurl was ignored when compiling with BoringSSL because
HAVE_SSL_CTX_SET_EC_CURVES was explicitly disabled if BoringSSL was
detected.  However, this feature is supported in BoringSSL since
5fd1807d. This commit enables it, and also reduces the required minimal
OpenSSL version to 1.0.2 as per OpenSSL's official documentation.

Fixes #8553.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -208,9 +208,18 @@
      !defined(OPENSSL_IS_BORINGSSL))
 #define HAVE_SSL_CTX_SET_CIPHERSUITES
 #define HAVE_SSL_CTX_SET_POST_HANDSHAKE_AUTH
-/* SET_EC_CURVES is available under the same preconditions: see
- * https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set1_groups.html
+#endif
+
+/*
+ * Whether SSL_CTX_set1_curves_list is available.
+ * OpenSSL: supported since 1.0.2, see
+ *   https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set1_groups.html
+ * BoringSSL: supported since 5fd1807d95f7 (committed 2016-09-30)
+ * LibreSSL: not tested.
  */
+#if ((OPENSSL_VERSION_NUMBER >= 0x10002000L) && \
+     !defined(LIBRESSL_VERSION_NUMBER)) || \
+    defined(OPENSSL_IS_BORINGSSL)
 #define HAVE_SSL_CTX_SET_EC_CURVES
 #endif
 


bagder:
Thanks!

PR #8562, opened 2022-03-08
Add back setsockopt path to curlx_nonblock

donny-dont:
The implementation using setsockopt was removed when BeOS support was purged. However this functionality wasn't BeOS specific and `HAVE_SETSOCKOPT_SO_NONBLOCK` is still present in the codebase.

### DIFF
--- a/lib/nonblock.c
+++ b/lib/nonblock.c
@@ -73,6 +73,12 @@ int curlx_nonblock(curl_socket_t sockfd,    /* operate on this */
   long flags = nonblock ? 1L : 0L;
   return IoctlSocket(sockfd, FIONBIO, (char *)&flags);
 
+#elif defined(HAVE_SETSOCKOPT_SO_NONBLOCK)
+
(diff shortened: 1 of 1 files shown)


bagder:
What is it needed for?

bagder:
Why do you avoid to write "Orbis OS", presumably the PlayStation 4 system software? It would help if you'd mention that maybe as a comment in the code, to avoid a future mistake in the similar vein.
 
Does curl build fine otherwise out of the box for Orbis if this fix is brought back?

donny-dont:
Yea so PlayStation 4/5 use cURL and this change makes it so it'll build out of the box.

donny-dont:
Noted the bit of code is for `game consoles`.

bagder:
I would still like that comment to mention more exact what OS it is for, but I can add that.

bagder:
Thanks!

donny-dont:
Thank you!

PR #8569, opened 2022-03-09
ngtcp2: disconnect the QUIC connection proper

### LINKED ISSUE #8534: http/3: CURL does not send connection close message
### I did this
I sent an HTTP3 request with CURL (NGTCP2 version) to my http3 server by:
curl  -k --http3 https://1.1.1.1:443

### I expected the following
After curl received the file, i expect it to send a CONNECTION_CLOSE message.
However no CONNECTION_CLOSE is sent.
Further health-check packets from server gets "ICMP port unreachable" response.
This suggests that the client disconnected and close the connection, 
but without sending close message, as it should.

### curl/libcurl version
This happens only on latest curl based on NGTCP2.
issue is NOT seen with QUICHE version.

[curl -V output]
curl 7.82.0-DEV (x86_64-pc-linux-gnu) libcurl/7.82.0-DEV OpenSSL/3.0.0 zlib/1.2.11 ngtcp2/0.3.0-DEV nghttp3/0.3.0-DEV
Release-Date: [unreleased]
Features: alt-svc AsynchDNS HSTS HTTP3 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets

This looks like a bug in CURL code, in the inegration code with NGTCP2.

Comparing files under curl/lib/vquic
we can see that qs_disconnect function is implemented differently for ngtcp2.c and quiche.c
as the quiche.c implementation calls conn_close while ngtcp.c does not.

### operating system
Ubuntu

Linux 23 5.13.0-30-generic #33~20.04.1-Ubuntu SMP Mon Feb 7 14:25:10 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

bagder:
Fixes #8534
Reported-by: mehatzri on github

### DIFF
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -868,6 +868,7 @@ static void qs_disconnect(struct quicsocket *qs)
                                           NULL, /* pkt_info */
                                           (uint8_t *)buffer, sizeof(buffer),
                                           0, /* error_code */
+                                          NULL, 0, /* reason + len */
                                           ts);
   if(rc > 0) {
     while((send(qs->conn->sock[FIRSTSOCKET], buffer, rc, 0) == -1) &&
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -858,7 +858,7 @@ static int ng_getsock(struct Curl_easy *data, struct connectdata *conn,
 
 static void qs_disconnect(struct quicsocket *qs)
 {
-  char buffer[256];
+  char buffer[NGTCP2_MAX_UDP_PAYLOAD_SIZE];
   ngtcp2_tstamp ts;
   ngtcp2_ssize rc;
   if(!qs->conn) /* already closed */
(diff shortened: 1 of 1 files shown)


bagder:
Maybe a bit crude, but should work I hope?

lib/vquic/ngtcp2.c
```diff
@@ -858,8 +858,22 @@ static int ng_getsock(struct Curl_easy *data, struct connectdata *conn,
 
 static void qs_disconnect(struct quicsocket *qs)
 {
+  char buffer[256];
```
tatsuhiro-t, line 861: It would be nice to extend this to at least NGTCP2_MAX_UDP_PAYLOAD_SIZE (1200).
RFC 9000 demands that Client Initial packet must be padded to 1200 bytes.  It looks like the current ngtcp2 code does not enforce this.

bagder: fixed!

tatsuhiro-t (approved):

PR #8579, opened 2022-03-10
openssl: check SSL_get_peer_cert_chain return value

jpmenil:
Ensure to check SSL_get_peer_cert_chain return value, as it can be NULL

Signed-off-by: Jean-Philippe Menil <jpmenil@gmail.com>


bagder (approved):

bagder:
Thanks!

PR #8587, opened 2022-03-12
docs/opts: Mention Schannel client cert type is P12

jay:
Schannel backend code behaves same as Secure Transport, it expects a P12
certificate file or the name of a certificate already in the user's OS
key store. Also, both backends ignore CURLOPT_SSLKEY (tool: --key)
because they expect the private key to already be available from the
keystore or P12 certificate.

Ref: https://github.com/curl/curl/discussions/8581#discussioncomment-2337260

Closes #xxxx

### DIFF
--- a/src/tool_listhelp.c
+++ b/src/tool_listhelp.c
@@ -62,7 +62,7 @@ const struct helptxt helptext[] = {
    "Verify the status of the server cert via OCSP-staple",
    CURLHELP_TLS},
   {"    --cert-type <type>",
-   "Certificate type (DER/PEM/ENG)",
+   "Certificate type (DER/PEM/ENG/P12)",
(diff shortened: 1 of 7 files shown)


src/tool_listhelp.c
```diff
@@ -62,7 +62,7 @@ const struct helptxt helptext[] = {
    "Verify the status of the server cert via OCSP-staple",
    CURLHELP_TLS},
   {"    --cert-type <type>",
-   "Certificate type (DER/PEM/ENG)",
+   "Certificate type (DER/PEM/ENG/P12)",
```
bagder, line 65: This file is generated, so this change should also be updated in `docs/cmdline/cert-type.d`.

PR #8590, opened 2022-03-13
add short cmdline-opt for no-clobber like in wget

pheiduck:


bagder (changes requested):
`-nc`  cannot be made to work as a short option in curl as it is the equivalent to `-n -c` and that already has a meaning you cannot change like that. See also https://curl.se/mail/archive-2022-03/0033.html

PR #8593, opened 2022-03-14
header api: add curl_easy_header and curl_easy_nextheader

bagder:
Early docs: https://github.com/curl/curl/wiki/get-headers-v2

- [X] code for the two functions
- [X] man pages for the functions
- [X] example showing the API functions in use
- [X] add test case for `curl_easy_header`
- [X] figure out how to deal with CONNECT response headers
- [X] figure out how to deal with 1xx response headers
- [X] add test case for `curl_easy_header` with CONNECT
- [X] documented as EXPERIMENTAL
- [x] add test case for `curl_easy_header` with 1xx
- [X] add test case for `curl_easy_header` with trailers
- [X] add test case for `curl_easy_header` in a redirect case
- [X] test case for `curl_easy_nextheader`
- [X] make it build fine when HTTP support is disabled 
- [X] use a single malloc per stored header
- [X] make the function calls able to access headers from multiple responses from the transfer
- [X] allow `curl -w` to output header contents with `%header{name}`
- [X] supports `-w %{header_json}` - ouputs all headers as JSON
- [X] make `-w %{header_json}` handle duplicate headers
- [X] add test case for `-w %header{name}`
- [X] add test case for `-w %{header_json}`
- [X] support/handle psuedo headers
- [X] skip making a test for psuedo headers (due to lacking test infra)
- [X] make the API (and -w options) build opt-in while EXPERIMENTAL
- [X] make the tests run only if the headers API is enabled in the build
- [X] make sure the docs appear nicely on the website (pending PR)

----

Remove the wiki page (to keep the docs in a single place) after initial merge


jay:
does this cover 100 status responses, for example 

HTTP/1.1 101 Switching Protocols
Upgrade: websocket
Connection: Upgrade

bagder:
Well, it should cover what we want it to cover. 1xx responses are part of HTTP and quite common so we need to support it somehow. The question is probably how to do it the best way.

bagder:
The more complicated setup for 100 might be something like this:
~~~
HTTP/1.1 100 Continue
Date: Tue, 08 Nov 2010 14:49:00 GMT

HTTP/1.1 200 OK
Date: Tue, 09 Nov 2010 14:49:00 GMT
Server: test-server/fake
~~~

If the the two responses have duplicate header names...

bagder:
I think its best if we store each received header according to exactly where it comes from and then provide them separately:
~~~c
/* for the 'type' argument below */
#define CURLH_NORMAL    0 /* plain headers */
#define CURLH_TRAILER   1 /* trailers */
#define CURLH_CONNECT   2 /* CONNECT headers */
#define CURLH_1XX       3 /* 1xx headers */

CURL_EXTERN CURLHcode curl_easy_header(CURL *easy,
                                       const char *name,
                                       size_t index,
                                       unsigned int type,
                                       struct curl_header **hout);

bagder:
Test 262 verifies that HTTP with embedded binary zeroes work fine. I think the time has come to drop support for that protocol violation and rather call that "illegal HTTP" and bail out. I'll work on that in a separate PR so that I can make this one not have to bother about that.

lib/headers.c
```diff
@@ -0,0 +1,314 @@
...
+
+  /* Find the end of the header name */
+  while(*header && (*header != ':'))
+    ++header;
+
+  if(*header)
+    /* Skip over colon, null it */
+    *header++ = 0;
+  else
+    return CURLE_BAD_FUNCTION_ARGUMENT;
```
jay, line 200: Something like BADHEADER (ie no colon) will cause this error. I think at worst we should ignore it. I realize all headers require colons but there is probably some server somewhere doing that.

Also, the only reason this works is because hs was calloc'd and so the last byte of the memory pointed to by hs is null, eg header points to BADHEADER\r\0. I'm not sure if that was intended and there is no bounds checking for end until the end of the function. In other words if header pointed to some string of length n > hlen then it looks like this function may in some cases read past hlen.

bagder: > Something like BADHEADER (ie no colon) will cause this error. I think at worst we should ignore it. I realize all headers require colons but there is probably some server somewhere doing that.

I think we should fail on such responses. Being lenient about such protocol violations I think is not going to help anyone but rather risks that such a server can trigger badness somewhere somehow. Then of course we should probably detect and trigger on that error before this function gets called!

I just added a better, explicit, zero termination but as you said: it makes a calloc so it should be good.

bagder:
I'm ready to merge this within the next few days. Any comments to take into account before that happens?

nanonyme:
Hmm, was this supposed to hide the new experimental ABI by default if --enable-headers-api is not given? Because it doesn't seem it does.

bagder:
Can you clarify what you're asking?

nanonyme:
[There was ](https://daniel.haxx.se/blog/2022/03/22/a-headers-api-for-libcurl/) clearly states that these new functions should not be exposed as they are experimental. Yet it looks based on our ABI checks that they are indeed exposed by default even without the flag to expose them https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/2386983527 Is this intended behaviour or should I file a bug?

bagder:
The functions are *present* even when disabled, but they don't do anything. They have no functionality.

nanonyme:
I see. But just exposing them means people may call them. As it was said, the ABI is subject to change up to function signatures so isn't it completely unsafe to keep them there even as no-op functions?

bagder:
Why would someone call them if they don't do anything? And yes, providing experimental features risks that users don't read our warnings and go ahead and use them anyway. I don't see an easy way to prevent that.

nanonyme:
Couldn't you just use USE_HEADERS_API macro to hide them as private symbols?

dfandrich:
On Thu, Apr 28, 2022 at 02:34:37PM -0700, Daniel Stenberg wrote:
> Why would someone call them if they don't do anything? 

In an application in the year 2024 uses the functions then someone downgrades
the library or runs the binary on CentOS with an old libcurl, it would silently
fail instead of raising a run-time dynamic link error.

PR #8594, opened 2022-03-14
tests/server/util.h: align WIN32 condition with util.c

mback2k:
-   tests/server/util.h: align WIN32 condition with util.c

    There is no need to test for both _WIN32 and WIN32 as curl_setup.h
    automatically defines the later if the first one is defined.

    Also tests/server/util.c is only checking for WIN32 arouund the
    implementation of win32_perror, so just defining _WIN32
    would not be sufficient for a successful compilation.

-   lib/warnless.[ch]: only check for WIN32 and ignore _WIN32

    curl_setup.h automatically defines WIN32 if just _WIN32 is defined.

### DIFF
--- a/lib/warnless.c
+++ b/lib/warnless.c
@@ -360,7 +360,7 @@ curl_socket_t curlx_sitosk(int i)
 
 #endif /* USE_WINSOCK */
 
-#if defined(WIN32) || defined(_WIN32)
+#if defined(WIN32)
 
 ssize_t curlx_read(int fd, void *buf, size_t count)
 {
@@ -372,7 +372,7 @@ ssize_t curlx_write(int fd, const void *buf, size_t count)
   return (ssize_t)write(fd, buf, curlx_uztoui(count));
 }
 
-#endif /* WIN32 || _WIN32 */
+#endif /* WIN32 */
 
 #if defined(__INTEL_COMPILER) && defined(__unix__)
 
--- a/lib/warnless.h
+++ b/lib/warnless.h
@@ -22,6 +22,8 @@
  *
  ***************************************************************************/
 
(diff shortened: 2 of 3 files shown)


jay:
This looks fine except for warnless.h which does not start with a `#include "curl_setup.h"` (therefore WIN32 possibly not yet defined). If I had to guess we omitted it because warnless.h is included by curlx.h. However, other header files that are included by curlx.h use `#include "curl_setup.h"` so I'm not sure what to make of it...

mback2k:
Thanks @jay, I though the same. Completely green CI seems to prove that the changed `warnless.[ch]` still works, but if we are not sure about the original intent, I may also drop the 2nd commit on merging this PR. @bagder do you maybe have insights?

jay:
I think at this point nobody outside the project is using curlx separately or curl_setup.h includes fine for them? I don't understand how that could be though. I think it's fine to add `#include "curl_setup.h"` to the top of warnless.h because the other curlx headers already include it and it's been like that for quite a while, years.

bagder (approved):

bagder:
I would presume that all current users of `warnless.h` right now include`curl_setup.h` themselves which makes this problem not appear.

mback2k:
Thanks, @bagder is this a yes or no regarding @jay suggestions? In case of a no, I would merge the PR as it is now.

> I think it's fine to add `#include "curl_setup.h"` to the top of warnless.h because the other curlx headers already include it and it's been like that for quite a while, years.

mback2k:
I added `#include "curl_setup.h"` to `warnless.h`. Do you guys think this is ready to be merged now? @bagder @jay

jay (approved):

PR #5888, opened 2020-08-28
[WIP] timeval: add curlx helper functions for timeval conversions

mback2k:
This is an early work-in-progress I just wanted to save here after talking about this with @bagder on IRC.

This PR will be updated to include more functions and update more existing places to use these functions.

PR #8595, opened 2022-03-15
timediff.[ch]: add curlx helper functions for timeval conversions

mback2k:
Also move timediff_t definitions from timeval.h to timediff.h and
then make timeval.h include the new standalone-capable timediff.h.

---

New version of #5888.

### DIFF
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -205,6 +205,7 @@ LIB_CFILES =         \
   system_win32.c     \
   telnet.c           \
   tftp.c             \
+  timediff.c         \
   timeval.c          \
   transfer.c         \
   url.c              \
@@ -325,6 +326,7 @@ LIB_HFILES =         \
   system_win32.h     \
   telnet.h           \
   tftp.h             \
+  timediff.h         \
   timeval.h          \
   transfer.h         \
   url.h              \
--- a/lib/vtls/gskit.c
+++ b/lib/vtls/gskit.c
@@ -74,6 +74,7 @@
 #include "connect.h" /* for the connect timeout */
 #include "select.h"
 #include "strcase.h"
+#include "timediff.h"
 #include "x509asn1.h"
 #include "curl_printf.h"
 
@@ -975,11 +976,12 @@ static CURLcode gskit_connect_step2(struct Curl_easy *data,
 
   for(;;) {
     timediff_t timeout_ms = nonblocking? 0: Curl_timeleft(data, NULL, TRUE);
+    stmv.tv_sec = 0;
+    stmv.tv_usec = 0;
     if(timeout_ms < 0)
       timeout_ms = 0;
-    stmv.tv_sec = timeout_ms / 1000;
-    stmv.tv_usec = (timeout_ms - stmv.tv_sec * 1000) * 1000;
-    switch(QsoWaitForIOCompletion(BACKEND->iocport, &cstat, &stmv)) {
+    switch(QsoWaitForIOCompletion(BACKEND->iocport, &cstat,
+                                  curlx_mstotv(&stmv, timeout_ms))) {
     case 1:             /* Operation complete. */
       break;
     case -1:            /* An error occurred: handshake still in progress. */
--- a/lib/asyn-ares.c
+++ b/lib/asyn-ares.c
@@ -65,6 +65,7 @@
 #include "connect.h"
 #include "select.h"
 #include "progress.h"
+#include "timediff.h"
 
 #  if defined(CURL_STATICLIB) && !defined(CARES_STATICLIB) &&   \
   defined(WIN32)
@@ -290,7 +291,7 @@ int Curl_resolver_getsock(struct Curl_easy *data,
 
   timeout = ares_timeout((ares_channel)data->state.async.resolver, &maxtime,
                          &timebuf);
-  milli = (timeout->tv_sec * 1000) + (timeout->tv_usec/1000);
+  milli = (long)curlx_tvtoms(timeout);
   if(milli == 0)
     milli += 10;
   Curl_expire(data, milli, EXPIRE_ASYNC_NAME);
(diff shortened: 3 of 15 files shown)


lib/Makefile.inc
```diff
@@ -205,6 +205,7 @@ LIB_CFILES =         \
   telnet.c           \
   tftp.c             \
   timeval.c          \
+  timeutil.C         \
```
jay, line 208: s/C/c/

jay:
Should be added to generate.bat curlx section:
~~~diff
diff --git a/projects/generate.bat b/projects/generate.bat
index 0bbe3fb..efded08 100644
--- a/projects/generate.bat
+++ b/projects/generate.bat
@@ -215,6 +215,7 @@ rem
       call :element %1 lib "curl_multibyte.c" %3
       call :element %1 lib "version_win32.c" %3
       call :element %1 lib "dynbuf.c" %3
+      call :element %1 lib "timeutil.c" %3
     ) else if "!var!" == "CURL_SRC_X_H_FILES" (
       call :element %1 lib "config-win32.h" %3
       call :element %1 lib "curl_setup.h" %3
@@ -225,6 +226,7 @@ rem
       call :element %1 lib "curl_multibyte.h" %3
       call :element %1 lib "version_win32.h" %3
       call :element %1 lib "dynbuf.h" %3
+      call :element %1 lib "timeutil.h" %3
     ) else if "!var!" == "CURL_LIB_C_FILES" (
       for /f "delims=" %%c in ('dir /b ..\lib\*.c') do call :element %1 lib "%%c" %3
     ) else if "!var!" == "CURL_LIB_H_FILES" (
~~~
and also CURLX_ sections in src/Makefile.inc and tests/server/Makefile.inc

lib/Makefile.inc
mback2k: Thanks, also noticed this due to the many CI failures.

bagder:
Maybe one day we should name all these curlx_* header/source files in special way, or put them in a subdir or something, to make it more apparent that they are magic and are used outside of the regular source dir.

mback2k:
Before we think of additional changes, do you think this PR is ready from a code/content change perspective? @bagder @jay

mback2k:
Before we think of additional changes, do you think this PR is ready from a code/content change perspective? @bagder @jay

lib/vtls/gskit.c
```diff
@@ -978,9 +979,8 @@ static CURLcode gskit_connect_step2(struct Curl_easy *data,
     timediff_t timeout_ms = nonblocking? 0: Curl_timeleft(data, NULL, TRUE);
     if(timeout_ms < 0)
       timeout_ms = 0;
-    stmv.tv_sec = timeout_ms / 1000;
-    stmv.tv_usec = (timeout_ms - stmv.tv_sec * 1000) * 1000;
-    switch(QsoWaitForIOCompletion(BACKEND->iocport, &cstat, &stmv)) {
+    switch(QsoWaitForIOCompletion(BACKEND->iocport, &cstat,
+                                  curlx_mstotv(timeout_ms))) {
```
jay, line 983: this one doesn't look right

jay: maybe

~~~c
    curlx_mstotv(&stmv, timeout_ms);
    switch(QsoWaitForIOCompletion(BACKEND->iocport, &cstat, &smtv)) {
~~~

?

jay:
Any curlx refactor would be separate from this. Aside from the gskit comment I just left this pr looks fine to me.

lib/vtls/gskit.c
mback2k: hm, to me it looks like using `select`-semantics for this parameter is just fine according to IBM documentation:

> **struct timeval * timeToWait**
> (Input) A pointer to a timeval structure that contains the time in seconds and microseconds for which the QsoWaitForIOCompletion() call should block if there is no completion status to receive.
> _If this parameter is null, QsoWaitForIOCompletion() waits indefinitely. If this value is specified, and 0 seconds 0 microseconds are specified, QsoWaitForIOCompletion() returns immediately._

curlx_mstotv aligns exactly do that 2nd sentence by returning NULL if stmv is also NULL (or timeout_ms is < 0) and setting the timeval members to 0 if timeout_ms is 0.

jay: curlx_mstotv takes two parameters not one. also I don't know why we'd want to wait indefinitely that doesn't make sense

mback2k: You are right, sorry, I totally missed the 2nd parameter thing. If timeout_ms is actually less then 0 and curlx_mstotv returns NULL, then smtv would be uninitialized and therefore the behaviour would still be undefined anyway.

mback2k: Actually timeout_ms cannot be less then 0 because of this preceding code:
```
    if(timeout_ms < 0)
      timeout_ms = 0;
```
So we won't ever block as curlx_mstotv will never return NULL in this case.

jay, line 983: Yes. IMO it's clearer to init smtv to 0 , call curlx_mstotv and pass &smtv as the param because that is always !=NULL but this would work as well:
```suggestion
                                  curlx_mstotv(&stmv, timeout_ms))) {
```
this one though there's an assumption that curlx_mstotv will always return != NULL. though that may be the case now code could change.

mback2k: yes, but with your first suggestion you assume that curlx_mstotv always works on the timeval struct provided. Right now the idea is to use the return value which at the time of this writing is based on the first argument, but this could also change at some point.

mback2k:
@jay what do you think of the solution in the most recent version of this PRs changes to `gskit.c` now?

jay:
> @jay what do you think of the solution in the most recent version of this PRs changes to `gskit.c` now?

LGTM. I've marked the review conversations as resolved.

mback2k:
CI is ✅, can this go in during feature freeze @bagder?

bagder (approved):
:+1:

PR #8596, opened 2022-03-15
Ci: use autoreconf -fi instead buildconf

pheiduck:
It should be enough to use autoreconf -fi instead buildconf

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -169,8 +169,8 @@ stages:
     - script: sudo apt-get update && sudo apt-get install -y clang-tools-10 clang-9 libssl-dev libssh2-1-dev libpsl-dev libbrotli-dev libzstd-dev
       displayName: 'apt install'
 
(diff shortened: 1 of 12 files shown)


bagder (changes requested):
Agreed, but when doing this change wouldn't it make sense to change at **all** uses of buildconf in github workflow files? (perhaps even in all CI config files)
~~~
$ git grep -l buildconf
workflows/linux-hyper.yml
workflows/macos.yml
workflows/mbedtls.yml
workflows/nss.yml
workflows/rustls.yml
workflows/wolfssl.yml

pheiduck:
> Agreed, but when doing this change wouldn't it make sense to change at **all** uses of buildconf in github workflow files? (perhaps even in all CI config files)

Saw that I missed some of it, I'm on it ;)

bagder (approved):

bagder:
Thanks!

PR #8601, opened 2022-03-16
http: reject header contents with nul bytes

bagder:
They are not allowed by the protocol and allowing them risk that curl
misbehaves somewhere where c functions are used but won't work on the
full contents. Further, they are not supported by hyper and they cause
problems for the new comign headers API work.

Updated test 262 to verify.


bagder:
I made it return `CURLE_WEIRD_SERVER_REPLY`, as it seemed the best available error code.

PR #8607, opened 2022-03-18
tool_urlglob: return error when the output file name is blank

bagder:
Reported-by: iblanes on github
Fixes #8606


bagder:
Nah. To be discussed more in #8606

PR #8630, opened 2022-03-24
dynbuf: make dyn_ptr consistently return NULL when length is zero

bagder:
Previously it would return a pointer if the buffer previously had
contents. This makes the code follow the documentation.

Added test case 1614 to verify this behavior.

Spotted-by: iblanes
Ref: #8606

### DIFF
--- a/lib/dynbuf.c
+++ b/lib/dynbuf.c
@@ -5,7 +5,7 @@
  *                            | (__| |_| |  _ <| |___
  *                             \___|\___/|_| \_\_____|
  *
- * Copyright (C) 2020, Daniel Stenberg, <daniel@haxx.se>, et al.
+ * Copyright (C) 2020 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
  *
  * This software is licensed as described in the file COPYING, which
  * you should have received as part of this distribution. The terms
@@ -222,25 +222,25 @@ CURLcode Curl_dyn_addf(struct dynbuf *s, const char *fmt, ...)
 }
 
 /*
- * Returns a pointer to the buffer.
+ * Returns a pointer to the buffer if it has a length.
  */
 char *Curl_dyn_ptr(const struct dynbuf *s)
 {
   DEBUGASSERT(s);
   DEBUGASSERT(s->init == DYNINIT);
   DEBUGASSERT(!s->leng || s->bufr);
-  return s->bufr;
+  return s->leng ? s->bufr : NULL;
 }
 
 /*
- * Returns an unsigned pointer to the buffer.
+ * Returns an unsigned pointer to the buffer if it has a length.
  */
 unsigned char *Curl_dyn_uptr(const struct dynbuf *s)
 {
   DEBUGASSERT(s);
   DEBUGASSERT(s->init == DYNINIT);
   DEBUGASSERT(!s->leng || s->bufr);
-  return (unsigned char *)s->bufr;
+  return s->leng ? (unsigned char *)s->bufr : NULL;
 }
 
 /*
--- a/lib/http.c
+++ b/lib/http.c
@@ -3349,6 +3349,8 @@ static statusline
 checkprotoprefix(struct Curl_easy *data, struct connectdata *conn,
                  const char *s, size_t len)
 {
+  if(!len)
+    return STATUS_UNKNOWN;
 #ifndef CURL_DISABLE_RTSP
   if(conn->handler->protocol & CURLPROTO_RTSP)
     return checkrtspprefix(data, s, len);
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -200,7 +200,7 @@ test1566 test1567 test1568 test1569 test1570 \
 test1590 test1591 test1592 test1593 test1594 test1595 test1596 \
 \
 test1600 test1601 test1602 test1603 test1604 test1605 test1606 test1607 \
-test1608 test1609 test1610 test1611 test1612 test1613 \
+test1608 test1609 test1610 test1611 test1612 test1613 test1614 \
 \
 test1620 test1621 \
 \
--- a/tests/data/test1614
+++ b/tests/data/test1614
@@ -1,0 +1,22 @@
+<testcase>
+<info>
+<keywords>
+unittest
+dynbuf
+</keywords>
+</info>
+
+#
+# Client-side
+<client>
+<server>
+none
+</server>
+<features>
+unittest
+</features>
+ <name>
+dynbuf
+ </name>
+</client>
+</testcase>
--- a/tests/unit/Makefile.inc
+++ b/tests/unit/Makefile.inc
@@ -32,7 +32,7 @@ UNITPROGS = unit1300 unit1301 unit1302 unit1303 unit1304 unit1305 unit1307 \
  unit1330 unit1394 unit1395 unit1396 unit1397 unit1398 \
  unit1399 \
  unit1600 unit1601 unit1602 unit1603 unit1604 unit1605 unit1606 unit1607 \
- unit1608 unit1609 unit1610 unit1611 unit1612 \
+ unit1608 unit1609 unit1610 unit1611 unit1612 unit1614 \
  unit1620 unit1621 \
  unit1650 unit1651 unit1652 unit1653 unit1654 unit1655 \
  unit1660 unit1661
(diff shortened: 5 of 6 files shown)


bagder:
Circle CI seems to have removed the arm hosts from the free tier, now saying:

*"job execution unauthorized: This job has been blocked because resource-class:arm.medium is not available on your plan. Please upgrade to continue building"*

jay:
IMO 0 is a valid length and the buffer should still be returned.

~~~
    Curl_dyn_add(&s,"");
    p = Curl_dyn_ptr(&s);  // should be != NULL
    x = Curl_dyn_len(&s);  // should be 0
~~~

monnerat:
I fear it would introduce some problems where we do want a zero-length string, like in https://github.com/curl/curl/blob/f718a915477aed881808244b9f2ad82cee3812f9/lib/mime.c#L317-L327

jay:
Yes. I think the documentation is actually wrong in this case. If a valid pointer must always be returned by dyn_ptr/uptr then a placeholder could be used instead of malloc(1) which some memory managers internally may put at some low number like 256 and there might be a heavier cost on realloc to a large number.

~~~diff
diff --git a/docs/DYNBUF.md b/docs/DYNBUF.md
index 16aca8a..e92f379 100644
--- a/docs/DYNBUF.md
+++ b/docs/DYNBUF.md
@@ -84,9 +84,9 @@ larger than the buffer length.
 char *Curl_dyn_ptr(const struct dynbuf *s);
 ```
 
-Returns a `char *` to the buffer if it has a length, otherwise a NULL. Since
-the buffer may be reallocated, this pointer should not be trusted or used
-anymore after the next buffer manipulation call.
+Returns a `char *` to the buffer. Since the buffer may be reallocated, this
+pointer should not be trusted or used anymore after the next buffer
+manipulation call.
 
 ## uptr
 
@@ -94,9 +94,9 @@ anymore after the next buffer manipulation call.
 unsigned char *Curl_dyn_uptr(const struct dynbuf *s);
 ```
 
-Returns an `unsigned char *` to the buffer if it has a length, otherwise a
-NULL. Since the buffer may be reallocated, this pointer should not be trusted
-or used anymore after the next buffer manipulation call.
+Returns an `unsigned char *` to the buffer. Since the buffer may be
+reallocated, this pointer should not be trusted or used anymore after the next
+buffer manipulation call.
 
 ## len
 
diff --git a/lib/dynbuf.c b/lib/dynbuf.c
index ada7e0c..8483a1b 100644
--- a/lib/dynbuf.c
+++ b/lib/dynbuf.c
@@ -39,7 +39,8 @@ void Curl_dyn_init(struct dynbuf *s, size_t toobig)
 {
   DEBUGASSERT(s);
   DEBUGASSERT(toobig);
-  s->bufr = NULL;
+  s->bufr = s->placeholder;
+  s->placeholder[0] = 0;
   s->leng = 0;
   s->allc = 0;
   s->toobig = toobig;
@@ -55,7 +56,10 @@ void Curl_dyn_init(struct dynbuf *s, size_t toobig)
 void Curl_dyn_free(struct dynbuf *s)
 {
   DEBUGASSERT(s);
-  Curl_safefree(s->bufr);
+  if(s->bufr != s->placeholder) {
+    free(s->bufr);
+    s->bufr = s->placeholder;
+  }
   s->leng = s->allc = 0;
 }
 
@@ -95,10 +99,9 @@ static CURLcode dyn_nappend(struct dynbuf *s,
   if(a != s->allc) {
     /* this logic is not using Curl_saferealloc() to make the tool not have to
        include that as well when it uses this code */
-    void *p = realloc(s->bufr, a);
+    void *p = realloc(s->bufr == s->placeholder ? NULL : s->bufr, a);
     if(!p) {
-      Curl_safefree(s->bufr);
-      s->leng = s->allc = 0;
+      Curl_dyn_free(s);
       return CURLE_OUT_OF_MEMORY;
     }
     s->bufr = p;
diff --git a/lib/dynbuf.h b/lib/dynbuf.h
index 252411f..b9b7d03 100644
--- a/lib/dynbuf.h
+++ b/lib/dynbuf.h
@@ -47,6 +47,7 @@ struct dynbuf {
 #ifdef DEBUGBUILD
   int init;     /* detect API usage mistakes */
 #endif
+  char placeholder[1]; /* bufr points to this before allocation */
 };
 
 void Curl_dyn_init(struct dynbuf *s, size_t toobig);
~~~

monnerat:
> Why? It's just what the dynbuf_ptr function returns that changes, not how the buffer is managed. The length is still zero.

In the curl/lib/mime.c sequence above, function `escape_string()` terminates by "converting" the local struct dynbuf into an allocated char * without freeing the dynbuf. It uses `Curl_dyn_ptr()` for that purpose as dynbuf struct fields should not be accessed directly (bold-emphasized in https://github.com/curl/curl/blob/master/docs/DYNBUF.md). `escape_string()` is not supposed to return NULL unless an error occurs.

The change in this PR breaks that. If you apply, how can I perform this "conversion" without having to process the zero-length case separately?

Please note the for loop initialization clause: its role is to force allocation even if no characters are stored later.

Similar sequences occur elsewhere in the current code.

bagder:
> The change in this PR breaks that. If you apply, how can I perform this "conversion" without having to process the zero-length case separately?

I don't think I can. But this code currently assumes that there will be a string (pointer) to return even when nothing has been added to the buffer, and right now that's not what we've guaranteed the function to do so it is already a "grey zone" behavior already I think...

monnerat:
> I  don't think I can. But this code currently assumes that there will be a string (pointer) to return even when nothing has been added to the buffer, and right now that's not what we've guaranteed the function to do so it is a "grey zone" behavior already I think...

Something **has** been added by the for loop initialization clause: a zero-length string !

Thus I would be in favor of slightly updating/precising the documentation rather than changing the current behavior, in order to light this grey zone. Motivated primarily to avoid buffer overflows, the goal of dynbuf is to ease and simplify dynamic (re)allocation and having to check for particular cases is not a simplification.

In addition, the existence of this particular case is purely artificial and necessitates (IMO) useless additional code.

Jay's placeholder proposal is not satisfactory either since it is not "freeable" outside dynbuf, but as he says, a zero-length string is valid.

The current functionality can be precised as such:
- `Curl_dyn_init` only makes the dynbuf usable. No allocation. If nothing more is performed on the dynbuf, `Curl_dyn_ptr` returns NULL.
- ` Curl_dyn_add*` forces allocation even with a zero-length string. The first call to one of these can be considered as the real "initialization", in terms of allocation.
- A failure deallocates and clears the pointer., as `Curl_dyn_free` will do.
- `Curl_dyn_*ptr` return the pointer, whatever its value is.
- If the "parentship" of the buffer is transferred to the caller, do not use or free the dynbuf anymore or `Curl_dyn_init` it again.

An alternative would be to implement new function `Curl_dyn_adopt` that returns the pointer inconditionally and calls `Curl_dyn_init`.

jay:
If there is code that needs to take ownership of ptr how about a function for that

char *foo = Curl_dynbuf_release_ptr(&dyn);
if(foo) // released

release could reset dynbuf and return s->bufr unless s->bufr == s->placeholder in which case return malloc(1)

bagder:
@monnerat:
> the goal of dynbuf is to ease and simplify dynamic (re)allocation and having to check for particular cases is not a simplification.

And yet that was *exactly* what made me do this PR in the first place. As right now you *do* have to check because the result is not necessarily consistent... I'm not saying I'm married to doing it this way, just pointing out that it is also a matter of viewpoint.

@jay:
> release could reset dynbuf and return s->bufr unless s->bufr == s->placeholder in which case return malloc(1)

The new malloc for returning nothing from a never added-to string that doesn't actually need a malloc at all makes that solution hurt my soul little. I think I would prefer sticking to the current code and rather document it better.

jay:
alright how about this then

~~~c
char *Curl_dyn_release_ptr(struct dynbuf *s)
{
  char *ptr = s->bufr != s->placeholder ? s->bufr : NULL;
  s->bufr = NULL;
  s->leng = s->allc = 0;
  return ptr;
}
~~~

bagder:
> alright how about this then

What do you think the benefit of doing this would be?

bagder:
I think we can at least conclude that we shouldn't merge this PR as it currently stands, so I'll close it and rethink.

jay:
> What do you think the benefit of doing this would be?

I think it's bad practice for a ptr returned by Curl_dyn_ptr to be freed outside of its dynbuf so I suggested a function to "release" the dynbuf buffer (ie take ownership of ptr).

bagder:
In the current cases where the ptr is returned like that, the struct is stored on the stack so whatever is done within there is gone in the next instruction...

PR #8631, opened 2022-03-24
New ARM CI because Circle CI seems to have removed the arm hosts

pheiduck:
reference: https://github.com/curl/curl/pull/8630#issuecomment-1077675400
@bagder I need your feedback about it, and its still in progress. ;)


pheiduck:
And yes, I need to touch up here and there, I still need to figure that out. I just wanted to put that as an idea

bagder:
Seems like a fine idea!

pheiduck:
maybe we should use fosshost for this @bagder?

pheiduck:
not sure why this does not execute the commands :/

pheiduck:
Best solution for this is fosshost + selfhosted arm runner

pheiduck:
How much space does the Hyper Build need?

pheiduck:
I think we should go with fosshost and selfhosted arm runner on it.

pheiduck:
Or Azure CLI can also be used for this.

pheiduck:
Otherwise we use builds as documented

pheiduck:
I think we need another solution for this :( I don’t get this to work.

PR #8644, opened 2022-03-28
mqtt: better handling of TCP disconnect mid-message

bagder:
Reported-by: Jenny Heino
Bug: https://hackerone.com/reports/1521610


bagder:
The CI fuzz job has started to permail fail, I've filed an issue: https://github.com/google/oss-fuzz/issues/7468

PR #8645, opened 2022-03-29
setopt: make curl_easy_vsetopt public

bagder:
This moves the internal Curl_vsetopt function into the public API so
that users can set options with a va_list.

Signed-off-by: Drew DeVault <sir@cmpwn.com>

This is https://curl.se/mail/lib-2022-03/0089.html with 3 copyright year ranges fixed.

### DIFF
--- a/docs/libcurl/Makefile.inc
+++ b/docs/libcurl/Makefile.inc
@@ -42,6 +42,7 @@ man_MANS = \
  curl_easy_strerror.3 \
  curl_easy_unescape.3 \
  curl_easy_upkeep.3 \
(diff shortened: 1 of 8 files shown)


bagder:
[quote from](https://curl.se/mail/lib-2022-03/0104.html) Drew DeVault:

> I'm going to abandon my patch, which was only to facilitate a temporary hack on my end anyway. This was not a good way for this to end.

keynmol:
Howdy! 
If I (or someone) were to attempt to bring this back into Curl's public API (I promise to check the Github Actions output!), would it be accepted?

PR #8651, opened 2022-03-30
vtls: provide a unified APLN-disagree string for all backends

### LINKED ISSUE #8643: Proposed ALPN failure clarification: append ", continuing with post-TLS negotiation"
Following up on the previous reports of the technically accurate but confusing to non-experts ALPN notification, a clarification that would have helped me would be:

    ALPN: server did not agree on a protocol, continuing with post-TLS negotiation

I would have submitted a PR, but this message appears in multiple places, so I wasn't sure if the preferred approach would be to just change them all or to factor them out as a shared constant (assuming that the proposed change is even acceptable in the first place).

(I would have commented on https://github.com/curl/curl/issues/2749 instead of filing a new ticket but it's locked)

bagder:
Also rephrase to make it sound less dangerous:

 "ALPN: server did not agree on a protocol. Uses default."

Reported-by: Nick Coghlan
Fixes #8643

### DIFF
--- a/lib/vtls/bearssl.c
+++ b/lib/vtls/bearssl.c
@@ -884,7 +884,7 @@ static CURLcode bearssl_connect_step3(struct Curl_easy *data,
                           BUNDLE_MULTIPLEX : BUNDLE_NO_MULTIUSE);
     }
     else
-      infof(data, "ALPN, server did not agree to a protocol");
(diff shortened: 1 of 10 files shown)


bagder:
Maybe this will also mark the beginning of us making more TLS-specific strings generic between the backends to make them look and appear more similar to users...

Issue #1150, opened 2016-11-29
IMAP custom FETCH request not working correctly


do-you-even-curl:
### I did this
I make a custom IMAP request `UID FETCH 34 BODY[]`. When I call `curl_easy_perform` on it, my custom write function only receives this:
> \* 2 FETCH (UID 34 BODY[] {571}
> 

verbose output:
> \> A004 UID FETCH 34 BODY[]
> < * 2 FETCH (UID 34 BODY[] {571}
> < Return-Path: ...
> < Received: from ...
> < X-Sieve: ...
> < To: ...
> < From: ...
> < Subject: test
> < Message-ID: ...
> < Date: Tue, 29 Nov 2016 17:42:09 +0100
> < MIME-Version: 1.0
> < Content-Type: text/plain; charset=utf-8; format=flowed
> < Content-Transfer-Encoding: 7bit
> <
> < this is a test
> < )
> < A004 OK Completed (0.000 sec)

### I expected the following
I expected my custom write function to receive the entire header and body of the mail instead of just the first line of the response. I expected it to behave the same as when calling `curl_easy_perform` without a custom request, but with the URL set to imaps://servername/INBOX;UID=1.

This is the verbose output when calling imaps://servername/INBOX;UID=1:
> \> A004 FETCH 1 BODY[]
> < * 2 FETCH (BODY[] {571}
> \* Found 571 bytes to download
> \* Written 571 bytes, 0 bytes are left for transfer
> < )
> < A004 OK Completed (0.000 sec)

In this case my custom write function gets called with the entire header and body of the mail.

### curl/libcurl version
curl-7.52.0-20161129.tar.bz2

### operating system
ubuntu 14.04

P.S.: The problem doesn't seem to be connected to the fact that one does a `UID FETCH` and the other a normal `FETCH` because the exact same problem appears when I execute `FETCH 1 BODY[]` as a custom request instead of executing it implicitly by calling the URL imaps://servername/INBOX;UID=1.

jay:
Dupe, see #536. Unfortunately it's a low priority nobody has had the time to work on it. You can parse the data from the header callback for now.

Issue #536, opened 2015-11-20
IMAP custom requests always use the LIST response handler


eXeC64:
Using a custom request such as `FETCH 9 BODY[HEADER.FIELDS (DATE FROM TO SUBJECT)]` yields only the following response data. `* 9 FETCH (BODY[HEADER.FIELDS (DATE FROM TO SUBJECT)] {147}`

The 147 bytes of actual data are not retrieved/handled. This stops any kind of custom `FETCH` request from being useful, as in many cases the actual response data is simply inaccessible through libcurl.

jay:
Parsing isn't done for custom IMAP requests iirc. Should it be? That's the question I guess. To your report though the data is received but remains part of the headers. You can either issue effectively the same command using an IMAP URL or parse the headers.

[This has come up before](http://curl.haxx.se/mail/lib-2014-05/0047.html). Some commands there may not be any alternative other than go custom and parse the header data. Also note there is a known bug with IMAP where long lines are truncated ([knownbugs #90](http://curl.haxx.se/docs/knownbugs.html)). I recall I did take a look at both of those issues at different times but they were too complicated to fix in the time I had. cc to Steve (@captain-caveman2k) in case he's still interested.

So for your example you should be able to do the same thing as Steve showed to me in that thread:
```c
  curl_easy_setopt(curl, CURLOPT_URL, "imaps://imap.gmail.com:993/INBOX/;UID=9/;SECTION=HEADER.FIELDS%20(DATE%20FROM%20TO%20SUBJECT)");
  curl_easy_perform(curl);
```
and read from the body. You will need a recent version of curl (I forget how recent -- just use the latest version).

Or if you have to...
```c
  curl_easy_setopt(curl, CURLOPT_URL, "imaps://imap.gmail.com:993/INBOX");
  curl_easy_setopt(curl, CURLOPT_CUSTOMREQUEST, "FETCH 9 BODY[HEADER.FIELDS (DATE FROM TO SUBJECT)]");
  curl_easy_perform(curl);
```
and read from the header which would be like
```c
* 9 FETCH (BODY[HEADER.FIELDS (DATE FROM TO SUBJECT)] {90}
Date: Thu, 19 Nov 2015 14:42:05 -0800 (PST)
From: <foo@bar.com>(foobar)
Subject: qux

)
A004 OK Success
```
then parse that

ancow:
[Related discussion on the mailing list.](https://curl.haxx.se/mail/lib-2014-04/0069.html)

[...]

PR #8659, opened 2022-03-31
Added detection of custom FETCH and UID FETCH commands for IMAP

MonkeybreadSoftware:
Added detection of custom FETCH and UID FETCH commands, so the output is correctly (and not redirected to debug log).


lib/imap.c
```diff
@@ -660,9 +660,27 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+        state(data, IMAP_FETCH);
+      }
+
+      /* e.g. "UID FETCH 107 BODY.PEEK[]" */
+
+      if(imap->custom_params)
+      if(strncasecompare(imap->custom, "UID", 3) == 1)
+      if(strncasecompare(imap->custom_params, " FETCH", 6) == 1) {
+        state(data, IMAP_FETCH);
+      }
```
bagder, line 680: 1.This code doesn't use the proper indenting, two spaces for new levels.
2.I don't think it should change state if it immediately again changes to something else. Change only once.

MonkeybreadSoftware: I added spaces for intending.
I think setting IMAP_FETCH twice should not be a problem.
it is much easier code to do it this way than to create one huge if to handle both cases together.

But feel free to adjust if needed.

bagder: The problem is primarily that it makes it harder to understand what's going on when you look at state transitions in a verbose log but also that it isn't actually changing the state if you immediately change it again so it's sort of a lie.

If we keep state changes to *actual* state changes, we keep the ability to at some point add state init logic there, the way we've done in some of our other state machines.

MonkeybreadSoftware: We could of course add "return result; " after changing state to fetch.
And then do the IMAP_LIST if the compares didn't set it to fetch.

MonkeybreadSoftware:
I added a test, which seems to work here.
I hope the ID I picked is still available.

lib/imap.c
MonkeybreadSoftware: I committed a new suggestion for the implementation with an isFetch flag, which only changes state once now.
Comments?

MonkeybreadSoftware:
What is still needed for this to proceed and get added to one of the next releases?

sawoods:
For reference #536, #1150 and [Issue with IMAP UID FETCH](https://curl.se/mail/lib-2014-04/0069.html)

MonkeybreadSoftware:
Could we get this added for the current feature window?
I'd really like to remove all the work arounds, where we take the email content from debug messages.

Anthropos888:
I also hope that this feature gets added. The UID FETCH response isn't complete so I have to get the complete response from the debug message.

bagder:
I restarted the appveyor builds to see if the errors remain.

bagder:
@MonkeybreadSoftware are you still looking to get this change merged? It merge conflicts now.

(We clearly lack someone with IMAP knowledge and will to push and take responsibility for IMAP changes in the project, which often makes them fall through the cracks and get forgotten.)

MonkeybreadSoftware:
Yes, I'd love to have this merged as then I don't need to patch it into myself every release.

MonkeybreadSoftware:
The conflict was only the test case number, so I moved 1662 to 1665 to get a free one.

PR #8668, opened 2022-04-01
Support socks5h proxy using unix sockets

balki:
**Usage:**

    curl -x "socks5h://unix/run/tor/socks" "https://example.com"

Ref: https://curl.se/mail/archive-2021-03/0012.html

This is a proof of concept to make sure it is working. Please take over from here or guide next steps - tests, documentation etc.,

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -2411,7 +2411,7 @@ static CURLcode parse_proxy(struct Curl_easy *data,
   int port = -1;
   char *proxyuser = NULL;
   char *proxypasswd = NULL;
-  char *host;
+  char *host = NULL;
   bool sockstype;
   CURLUcode uc;
   struct proxy_info *proxyinfo;
@@ -2548,14 +2548,14 @@ static CURLcode parse_proxy(struct Curl_easy *data,
     goto error;
   }
 #ifdef USE_UNIX_SOCKETS
-  if(sockstype && strcmp(UNIX_SOCKET_PREFIX, host) == 0) {
+  if(sockstype && strcasecompare(UNIX_SOCKET_PREFIX, host)) {
     uc = curl_url_get(uhp, CURLUPART_PATH, &path, CURLU_URLDECODE);
     if(uc) {
       result = CURLE_OUT_OF_MEMORY;
       goto error;
     }
     /* path will be "/", if no path was was found */
-    if(!(path[0]=='/' && path[1]=='\0')) {
+    if(strcmp("/", path)) {
       is_unix_proxy = TRUE;
       free(host);
       host = aprintf(UNIX_SOCKET_PREFIX"%s", path);
@@ -2566,21 +2566,23 @@ static CURLcode parse_proxy(struct Curl_easy *data,
       Curl_safefree(proxyinfo->host.rawalloc);
       proxyinfo->host.rawalloc = host;
       proxyinfo->host.name = host;
+      host = NULL;
     }
   }
 
   if(!is_unix_proxy) {
 #endif
-  Curl_safefree(proxyinfo->host.rawalloc);
-  proxyinfo->host.rawalloc = host;
-  if(host[0] == '[') {
-    /* this is a numerical IPv6, strip off the brackets */
-    size_t len = strlen(host);
-    host[len-1] = 0; /* clear the trailing bracket */
-    host++;
-    zonefrom_url(uhp, data, conn);
-  }
-  proxyinfo->host.name = host;
+    Curl_safefree(proxyinfo->host.rawalloc);
+    proxyinfo->host.rawalloc = host;
+    if(host[0] == '[') {
+      /* this is a numerical IPv6, strip off the brackets */
+      size_t len = strlen(host);
+      host[len-1] = 0; /* clear the trailing bracket */
+      host++;
+      zonefrom_url(uhp, data, conn);
+    }
+    proxyinfo->host.name = host;
+    host = NULL;
 #ifdef USE_UNIX_SOCKETS
   }
 #endif
@@ -2588,6 +2590,7 @@ static CURLcode parse_proxy(struct Curl_easy *data,
   error:
   free(proxyuser);
   free(proxypasswd);
+  free(host);
   free(scheme);
 #ifdef USE_UNIX_SOCKETS
   free(path);
@@ -3429,8 +3432,8 @@ static CURLcode resolve_server(struct Curl_easy *data,
       unix_path = conn->unix_domain_socket;
 #ifndef CURL_DISABLE_PROXY
     else if(conn->socks_proxy.host.name
-        && strncmp(UNIX_SOCKET_PREFIX"/",
-          conn->socks_proxy.host.name, sizeof(UNIX_SOCKET_PREFIX)) == 0)
+        && !strncmp(UNIX_SOCKET_PREFIX"/",
+          conn->socks_proxy.host.name, sizeof(UNIX_SOCKET_PREFIX)))
       unix_path = conn->socks_proxy.host.name + sizeof(UNIX_SOCKET_PREFIX) - 1;
 #endif
 
--- a/lib/url.c
+++ b/lib/url.c
@@ -147,6 +147,10 @@ static void conn_free(struct connectdata *conn);
 # error READBUFFER_SIZE is too small
 #endif
 
+#ifdef USE_UNIX_SOCKETS
+#define UNIX_SOCKET_PREFIX "localhost"
+#endif
+
 /*
 * get_protocol_family()
 *
@@ -2414,6 +2418,11 @@ static CURLcode parse_proxy(struct Curl_easy *data,
   CURLU *uhp = curl_url();
   CURLcode result = CURLE_OK;
   char *scheme = NULL;
+#ifdef USE_UNIX_SOCKETS
+  char *path = NULL;
+  bool is_unix_proxy = FALSE;
+#endif
+
 
   if(!uhp) {
     result = CURLE_OUT_OF_MEMORY;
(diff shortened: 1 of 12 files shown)


lib/url.c
```diff
@@ -2519,6 +2519,20 @@ static CURLcode parse_proxy(struct Curl_easy *data,
   }
 
   /* now, clone the proxy host name */
+#ifdef USE_UNIX_SOCKETS
+  if(proxytype == CURLPROXY_SOCKS5_HOSTNAME
```
bagder, line 2523: Is there any reason to limit this to `SOCKS5_HOSTNAME` and not allow SOCKS4a too?

lib/url.c
```diff
@@ -2519,6 +2519,20 @@ static CURLcode parse_proxy(struct Curl_easy *data,
...
+      && strncmp("unix/", proxy + 10, 5) == 0) { /* len("socks5h://") = 10 */
```
bagder, line 2524: The host name *might* start with that scheme, the type could also be set with a separate option and then the proxy name starts with the host name. This cannot assume the first 10 bytes are known.

I also find it a little unfortunate to reserve the `unix` host name like this, as there might actually be a proxy in use that is named like that. Maybe we should instead make the `--unix-socket` option accept a `socks...` prefix?

bagder (commented):
For tests, I think test case 1435 can serve as good inspiration as it already uses `--unix-socket`

lib/url.c
balki: I thought it may not work as dns resolution is done locally in SOCKS4a using the same function. But after quick test, looks like it should work. I will update for all socks proxies.

lib/url.c
balki: > The host name might start with that scheme, the type could also be set with a separate option and then the proxy name starts with the host name. This cannot assume the first 10 bytes are known.

I will update to handle that case.

> there might actually be a proxy in use that is named like that.

In that case, `unix/path` cannot be a valid URI for a socks proxy as it can only have `host:port` syntax. Since I am checking for `unix/`, I think it should not affect anyone having a server named `unix` with a socks proxy running in it.

> Maybe we should instead make the --unix-socket option accept a socks... prefix

I wish to have the ability to set the proxy via environment variable like `ALL_PROXY=socks5h://unix/run/tor/socks`, so that programs that internally use curl does not have to change the command line.

dfandrich:
How about a syntax like this:

ALL_PROXY=socks5h:///unix/run/tor/socks

i.e. an extra slash, making the host part of the URL empty. This is similar to
how a file: URL would refer to the same file.

bagder:
> i.e. an extra slash, making the host part of the URL empty. This is similar to how a file: URL would refer to the same file.

Unfortunately, the URL parser accepts one, two or three slashes where there should be two so three slashes will be accepted as if they were two:

~~~
$ curl -x socks5h:///moo/hello localhost
curl: (7) Failed to connect to moo port 1080 after 0 ms: Connection refused
~~~

Another option is to use a short but otherwise illegal host name. For example just two dots (`..`):
~~~
$ curl -x socks5h://../run/tor/socks localhost
curl: (6) Could not resolve host: ..

lib/url.c
balki: Updated to support all socks proxies and not depend on the scheme prefix in proxy string. Please confirm if implementation/syntax is ok. I will then add a test.

dfandrich:
On Sat, Apr 02, 2022 at 02:55:00PM -0700, Daniel Stenberg wrote:
> Unfortunately, the URL parser accepts one, two or three slashes where there
> should be two so three slashes will be accepted as if they were two:

That sounds wrong—it goes against RFC 1738 and RFC 8089, anyway. Is this a
WHATWG thing?

bagder:
> That sounds wrong—it goes against RFC 1738 and RFC 8089, anyway. Is this a WHATWG thing?

Yes. :cry:  The lax nature of WHATWG's URL spec makes such URLs occasionally get used in the wild, so this is our adjustment to accept such URLs better.The WHATWG spec actually allows any amount of slashes (even zero), and both backslashes and slashes (mixed) but curl only allows regular forward slashes and only one, two or three of them.

balki:
> For tests, I think test case 1435 can serve as good inspiration as it already uses `--unix-socket`

I looked at the test setup. I think the following changes are required along with writing a testcase.

1. Update this server to support taking unixsocket as argument. https://github.com/curl/curl/blob/master/tests/server/socksd.c
2. Pass the option from here: https://github.com/curl/curl/blob/master/tests/runtests.pl

Q: Where should the temporary socket be created and how/when should it be cleaned up?

With last update, If the hostname is `unix` and path is not just '/', the path is considered to to be a unix socket.

Please correct if I am missing anything.

balki:
@bagder Please take a look again. 
* Changed syntax to `socks5h://localhost/path/to/sock`
* Added tests

bagder:
The torture test fail of 1467 means that there is a memory/resource leak somewhere in the exit path.

bagder (changes requested):
There is no added documentation?

balki:
> The torture test fail of 1467 means that there is a memory/resource leak somewhere in the exit path.

Reviewed again. Could not find the leak.. Can you please share instructions to reproduce failure locally? I am on a ubuntu desktop. `make tests-full` pass fine.

Added documentation for the command line flags.

lib/url.c
```diff
@@ -2542,6 +2551,30 @@ static CURLcode parse_proxy(struct Curl_easy *data,
...
   }
+#ifdef USE_UNIX_SOCKETS
+  if(sockstype && strcmp(UNIX_SOCKET_PREFIX, host) == 0) {
+    uc = curl_url_get(uhp, CURLUPART_PATH, &path, CURLU_URLDECODE);
+    if(uc) {
+      result = CURLE_OUT_OF_MEMORY;
+      goto error;
+    }
+    /* path will be "/", if no path was was found */
+    if(!(path[0]=='/' && path[1]=='\0')) {
```
bagder, line 2562: Wouldn't it be easier to just do:
~~~c
if(strcmp("/", path)) {

lib/url.c
```diff
@@ -2542,6 +2551,30 @@ static CURLcode parse_proxy(struct Curl_easy *data,
...
+      }
+      Curl_safefree(proxyinfo->host.rawalloc);
+      proxyinfo->host.rawalloc = host;
+      proxyinfo->host.name = host;
+    }
+  }
+
+  if(!is_unix_proxy) {
+#endif
   Curl_safefree(proxyinfo->host.rawalloc);
```
bagder, line 2578: This line should be indented to the right here, right?

lib/url.c
```diff
@@ -2542,6 +2551,30 @@ static CURLcode parse_proxy(struct Curl_easy *data,
     result = CURLE_OUT_OF_MEMORY;
     goto error;
   }
+#ifdef USE_UNIX_SOCKETS
+  if(sockstype && strcmp(UNIX_SOCKET_PREFIX, host) == 0) {
```
bagder, line 2555: We usually avoid comparisons with zero like this and instead use the not operator. But the host name check should probably also be case insensitive. Try using `strcasecompare()`but note the return code.
~~~c
if(sockstype && strcasecompare(UNIX_SOCKET_PREFIX, host) ) {

lib/url.c
```diff
@@ -3388,25 +3427,35 @@ static CURLcode resolve_server(struct Curl_easy *data,
...
 #ifdef USE_UNIX_SOCKETS
-    if(conn->unix_domain_socket) {
+    char *unix_path = NULL;
+
+    if(conn->unix_domain_socket)
+      unix_path = conn->unix_domain_socket;
+#ifndef CURL_DISABLE_PROXY
+    else if(conn->socks_proxy.host.name
+        && strncmp(UNIX_SOCKET_PREFIX"/",
+          conn->socks_proxy.host.name, sizeof(UNIX_SOCKET_PREFIX)) == 0)
```
bagder, line 3437: another case sensitive check?

bagder:
The clue is **torture tests**. It's a special way to run the tests, which verifies the exit paths heavily.

> Reviewed again. Could not find the leak.. Can you please share instructions to reproduce failure locally? I am on a ubuntu desktop. make tests-full pass fine.

~~~
$ ./configure --enable-debug (+ c-ares or non-threaded resolver)
$ make
$ cd tests
$ make
$ ./runtests -n -t 1467
~~~

This then makes all allocations in the test fail, one by one and it verifies that it never leaks memory in any case.

The 1467 failure you got looked like this:

~~~
test 1467...[HTTP GET via SOCKS5 proxy via unix sockets]
 144 functions found, but only fail 25 (17.36%)
** MEMORY FAILURE
Leak detected: memory still allocated: 10 bytes
At 7f7f53e04bc8, there's 10 bytes.
 allocated by escape.c:144
LIMIT urlapi.c:1415 strdup reached memlimit
 Failed on function number 93 in test.
 invoke with "-t93" to repeat this single case.
 ~~~

It shows that escape.c:144 allocated a line that was never freed. It was when fallible function number 93 was made to fail. If you wanted reproduce that exact fail number, you can do this:

~~~
./runtests -t94 1467
~~~

I usually first try with -t to run all loops torture to see that it reproduces, it might not be the exact same number in a local build. Also, use -n to switch off valgrind since running *with* valgrind makes every loop take several times longer, which can make this test really slow. Using valgrind in the last step when you re-run a single number is very useful because valgrind can help you pinpoint the entire leak call stack much better.

bagder:
if you still have problems rerunning the test, I can have a go at it soon.

lib/url.c
balki: Fixed

lib/url.c
balki: Fixed. Was trying to save a function call. It should anyway be in-lined I hope.

lib/url.c
balki: Fixed

lib/url.c
balki: `host.name` here is set in line 2565, so no need insensitive compare.

balki:
> if you still have problems rerunning the test, I can have a go at it soon.

Thank you. Was able to reproduce and fix it. Waiting for CI run.

balki:
Fixed torture test and added documentation. Please review. curl/check failure does not show any logs or errors except for message "Build Failed". Please help understand the error.

balki:
> This looks very good and I'm ready to merge. Just one thing left that I can think of:
> 
> You added docs for the command line tool but not for the libcurl option that actually passes the string to libcurl. `CURLOPT_PROXY`. It needs to be similarly extended so that users can learn how to specify a socks proxy over unix domain socket with it.

Thank you. Updated CURLOPT_PROXY documentation.

bagder (approved):

bagder:
Thanks!

PR #8669, opened 2022-04-02
Updated all Copyright end Dates

Klay4:
Changed all the copyright years end dates to 2022.


bagder:
Thanks, but we maintain the copyright ranges to end on the last year the file was actually modified. We don't bulk update them unconditionally.

We do this as part of our release procedure with the help of the `./scripts/copyright.pl` script.

PR #8670, opened 2022-04-04
opts: deprecate RANDOM_FILE and EGDSOCKET

bagder:
These two options were only ever used for the OpenSSL backend for
versions before 1.1.0. They were never used for other backends and they
are not used with recent OpenSSL versions. They were never used much by
applications.

The defines RANDOM_FILE and EGD_SOCKET can still be set at build-time
for ancient EOL OpenSSL versions.

### DIFF
--- a/docs/libcurl/opts/CURLOPT_EGDSOCKET.3
+++ b/docs/libcurl/opts/CURLOPT_EGDSOCKET.3
@@ -5,7 +5,7 @@
 .\" *                            | (__| |_| |  _ <| |___
 .\" *                             \___|\___/|_| \_\_____|
 .\" *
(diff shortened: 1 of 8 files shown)


jay:
> Deprecated option. Serves no purpose anymore.

IMO this is unclear. deprecated means still supported but going to be removed eventually. while that may be true of the option's name it's not true of the option's function which has been removed. "Deprecated option. This option has no effect since 7.83.0." Also explain it can still be set at build time or document that somewhere else, eg "It can still be set at build time using define EGD_SOCKET."

jay:
> The **AVAILABILITY** section in the bottom has the exact version information.

I was referring to the curl tool options

bagder:
aha

PR #5017, opened 2020-03-02
global_init: atomic initialize counter

bagder:
With the use a Windows specific function and _Atomic for the rest.
    
`_Atomic` is a C11 feature that when available makes the init reference counter atomic and thread-safe.
    
init: call the init functions independent of counter
    
cleanup: decrements the counter and will only do the cleanups when the counter reaches zero

Part of the thread-safe `curl_global_init` effort.

PR #4064, opened 2019-06-22
global_init: Make it thread-safe when possible

jay:
- Use compare-and-swap with a full memory barrier to synchronize
  calls to curl_global_init, curl_global_init_mem, curl_global_cleanup
  and curl_global_sslset.

This is a locking method that does not require libcurl to be built with
a threading library. If compare-and-swap with a full memory barrier is
supported by the compiler (gcc, clang) or the OS (Windows) then libcurl
init/cleanup will be locked.

As discused in #3990 curl_global_ functions are not thread-safe which is
a problem for some users of libcurl that are themselves libraries. The
curl_global_ function requirements may be impossible to meet, eg:

"You must not call it when any other thread in the program (i.e. a
thread sharing the same memory) is running. This doesn't just mean no
other thread that is using libcurl. Because curl_global_init calls
functions of other libraries that are similarly thread unsafe, it could
conflict with any other thread that uses these other libraries."

Even if the user builds libcurl to depend only on libraries that offer
documented thread-safe init/cleanup, there is still libcurl's own
init/cleanup where we have a chicken-and-egg problem of how to
initialize and clean up our lock on many platforms. To solve that I've
implemented a lock that has only its state using compare-and-swap and
does not require any initialization or clean up.

We still cannot guarantee curl_global_ functions are thread-safe, but
this change should make it safer for the user.

Ref: https://github.com/curl/curl/issues/3990
Ref: https://curl.haxx.se/libcurl/c/curl_global_init.html
Ref: ??? probably this has come up before elsewhere

Closes #xxxx

---

/cc @nicowilliams @kdudka

PR #8680, opened 2022-04-05
curl_global_init(): make it thread-safe

tguillem:
Making curl_global_init() thread-safe can be very useful for libraries that can't control what other libraries (or the main application) are doing with Curl. That way, curl_global_init() doesn't have to be called at the very beginning of the process and can be called anytime (from another library init function, that is run later, for example).

PS: I didn't know that number to choose for the new test (3026).


lib/easy.c
```diff
@@ -89,6 +89,55 @@
 static unsigned int  initialized;
 static long          init_flags;
 
+#if defined(USE_THREADS_POSIX) || (defined(USE_THREADS_WIN32) && _WIN32_WINNT >= 0x0600)
```
bagder, line 92: Note that these `USE_THREADS_` defines are for using the threaded name resolvers. Meaning that builds that for example use c-ares for DNS don't have those defines set.

tguillem: Thanks for the review.

Would it be possible to enable pthread and win32 threads by default or via another configure option (like --init-thread-safe)?

By the way, this PR introduces a leak of the CRITICAL_SECTION on Windows (we can't clean it from curl_global_cleanup since this would prevent a future curl_global_init to work).

A potential alternative would be to implement a mutex using the "Futuxes Are Tricky" paper from Drepper (and use WIN32 WaitOnAddress() as futuxes) but this also requires adding atomics...

jay:
Also see #4064

bagder:
> Would it be possible to enable pthread and win32 threads by default or via another configure option (like --init-thread-safe)?

Sure it would be possible. It would imagine that check should be enabled by default.

But I would also like to see the #4064 and #5017 approaches considered properly, as they might be able to make us reach this goal without using a thread library.

lib/easy.c
```diff
@@ -133,6 +182,9 @@ static char *leakpointer;
  */
 static CURLcode global_init(long flags, bool memoryfuncs)
 {
+  if ((flags & CURL_GLOBAL_NEED_THREADSAFE) && !GLOBAL_INIT_IS_THREADSAFE)
+    return CURLE_NOT_THREADSAFE;
```
bagder, line 186: For the documentation it should be noted that passing in `CURL_GLOBAL_NEED_THREADSAFE` when the application is using an older libcurl (from before this support is added) may still have libcurl use an unsafe global_init.

tguillem:
> > Would it be possible to enable pthread and win32 threads by default or via another configure option (like --init-thread-safe)?
> 
> Sure it would be possible. It would imagine that check should be enabled by default.
> 
> But I would also like to see the #4064 and #5017 approaches considered properly, as they might be able to make us reach this goal without using a thread library.

On Windows, we could do implement a lock via Interlocked (Atomic like) and WaitOnAddress (futex like)  : 
There won't be protection against lock convoy or spinlock optimization, but I don't think we care if we use this lock only for the global init (it's very unlikely that init/cleanup are called a lot).

On Linux, we could do the same with atomics and futexes, but futuxes are not available on all POSIX systems (we could fall back to a wait loop but I'm not very fan). That's why I would prefer using pthread.

lib/easy.c
tguillem: I have added a new flag in curl_version_info_data.features instead.

lib/easy.c
```diff
@@ -84,11 +84,25 @@
 #include "curl_printf.h"
 #include "curl_memory.h"
 #include "memdebug.h"
+#include "easy_lock.h"
 
 /* true globals -- for curl_global_init() and curl_global_cleanup() */
 static unsigned int  initialized;
```
dfandrich, line 90: I think this is going to have to be marked volatile to ensure that the right value is used in all threads.

tguillem: The pthread_mutex and the SRWLock use a memory barrier, so volatile here is not needed.

tguillem:
> > Would it be possible to enable pthread and win32 threads by default or via another configure option (like --init-thread-safe)?
> 
> Sure it would be possible. It would imagine that check should be enabled by default.
> 
> But I would also like to see the #4064 and #5017 approaches considered properly, as they might be able to make us reach this goal without using a thread library.

If a program need a thread-safe init for curl, then it is likely already linking with the pthread library (on posix). I prefer the mutex approach to keep it simple since atomics or spinlocks are too often wrongly used (lock convoy, fairness, active wait).

For Windows, the SRWLock is a good a light solution too.

tguillem:
Hello,

I just proposed an alternative that doesn't need pthread: https://github.com/curl/curl/commit/7417fd4d1018342f4ebe76e06a8b14f30b5e4946

We still use SRWLOCK for Windows since it is simpler that way (don't need to use Windows Atomic API, and it is very likely to be a spinlock internally).

bagder (approved):
:+1: this is good stuff. Let's do this.

lib/easy_lock.h
```diff
@@ -0,0 +1,67 @@
...
+
+static inline void curl_simple_lock_lock(curl_simple_lock *lock)
+{
+  for(;;) {
+    if(!atomic_exchange_explicit(lock, true, memory_order_acquire))
+      break;
+    /* Reduce cache coherency traffic */
+    while(atomic_load_explicit(lock, memory_order_relaxed)) {
+      /* Reduce load (not mandatory) */
+#if defined(__i386__) || defined(__x86_64__)
```
piru, line 49: How about using `sched_yield()` here when available?

tguillem: I think we can, like the following, after all #ifdefs ?

```
#elif defined(_POSIX_PRIORITY_SCHEDULING)
sched_yield()
#endif
```

bagder: Or like this:

~~~diff
diff --git a/configure.ac b/configure.ac
index fdc4523e5..ce66b5592 100644
--- a/configure.ac
+++ b/configure.ac
@@ -3443,10 +3443,11 @@ AC_CHECK_FUNCS([fnmatch \
   getrlimit \
   gettimeofday \
   if_nametoindex \
   mach_absolute_time \
   pipe \
+  sched_yield \
   setlocale \
   setmode \
   setrlimit \
   usleep \
   utime \
diff --git a/lib/easy_lock.h b/lib/easy_lock.h
index 9e01793ac..41627c4d4 100644
--- a/lib/easy_lock.h
+++ b/lib/easy_lock.h
@@ -48,10 +48,12 @@ static inline void curl_simple_lock_lock(curl_simple_lock *lock)
       /* Reduce load (not mandatory) */
 #if defined(__i386__) || defined(__x86_64__)
       __builtin_ia32_pause();
 #elif defined(__aarch64__)
       asm volatile("yield" ::: "memory");
+#elif defined(HAVE_SCHED_YIELD)
+      sched_yield();
 #endif
     }
   }
 }

tguillem: Indeed :+1:

bagder:
Thanks!

mback2k:
I don't think this PR was sufficiently tested as the later merge of #8976 with 26 failed CI runs unfortunately shows.

bagder:
We saw **zero** interest from anyone to do more tests and we had no indication that this was not good enough on Windows. I think we merged it fine. We keep suffering on Windows because we have too few people around who know about and care for Windows (before things are merged), and the Windows CI jobs are flaky.

The test seems to have problems in itself, but does it show that this code isn't good enough and doesn't do what it's set out to do?

mback2k:
The Windows CI is far less flaky nowadays. Most failures are now caused by the underlying infrastructure or memory issues, not the tests themselves anymore. So, if you see more than 1 CI failing it is definitely worth checking.

I am sorry, but I personally can't keep up with this all the time although I am trying very hard. I tried to make the CI as stable as possible and still so, and then people just ignoring it again makes me question the whole purpose.

I don't know if this code without the test is broken, I just focused on the test last night. I will try to replicate a Windows specific test with CreateThread.

mback2k:
At least I would determine that a test pass rate of 99.91 % on Azure CI can be considered quite stable:
image

jay:
IMO this is not any of our fault. IIRC any CI we've used has always been flaky on Windows, weird input/output issues that can't be reproduced, and a bunch of hacks we have to create to get things working. They seem to be way more sensitive to timing and available resources. And credit to Marc for his work on this, they are definitely more stable than they used to be.

PR #8698, opened 2022-04-12
[WIP] Curl_quic_idle

tatsuhiro-t:
Curl_quic_idle to be called when no HTTP level read or write is
performed.  It is a good place to handle timer expiry for QUIC
transport (.e.g, retransmission).

This is kinda straw man, to add a machinery to vquic to handle general timer expiry stuff when curl has no HTTP data to send or receive.

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -1222,6 +1222,13 @@ CURLcode Curl_readwrite(struct connectdata *conn,
         infof(data, "Done waiting for 100-continue");
       }
     }
+
+#ifdef ENABLE_QUIC
+    if(conn->transport == TRNSPRT_QUIC)
+      result = Curl_quic_idle(data);
+    if(result)
+      return result;
+#endif
   }
 
   if(Curl_pgrsUpdate(data))
--- a/lib/quic.h
+++ b/lib/quic.h
@@ -54,6 +54,7 @@ void Curl_quic_done(struct Curl_easy *data, bool premature);
 bool Curl_quic_data_pending(const struct Curl_easy *data);
 void Curl_quic_disconnect(struct Curl_easy *data,
                           struct connectdata *conn, int tempindex);
+CURLcode Curl_quic_idle(struct Curl_easy *data);
 
 #else /* ENABLE_QUIC */
 #define Curl_quic_done_sending(x)
--- a/lib/vquic/msh3.c
+++ b/lib/vquic/msh3.c
@@ -495,4 +495,11 @@ bool Curl_quic_data_pending(const struct Curl_easy *data)
   return stream->recv_header_len || stream->recv_data_len;
 }
 
+CURLcode Curl_quic_idle(struct Curl_easy *data)
+{
+  (void)data;
+  H3BUGF(infof(data, "Curl_quic_idle"));
+  return CURLE_OK;
+}
+
 #endif /* USE_MSH3 */
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -1883,4 +1883,26 @@ bool Curl_quic_data_pending(const struct Curl_easy *data)
   return Curl_dyn_len(&stream->overflow) > 0;
 }
 
+/*
+ * Called from transfer.c:Curl_readwrite when neither HTTP level read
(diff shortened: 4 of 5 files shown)


lib/transfer.c
```diff
@@ -1222,6 +1222,13 @@ CURLcode Curl_readwrite(struct connectdata *conn,
         infof(data, "Done waiting for 100-continue");
       }
     }
+
+#ifdef ENABLE_QUIC
+    if(conn->transport == TRNSPRT_QUIC)
+      result = Curl_quic_idle(data);
+    if(result)
+      return result;
```
danielgustafsson, line 1230: The contract of this API isn't immediately clear to me, do we want library housekeeping functions be able to abort the transfer here? Also, other subsystem housekeeping, like removing expired cookies etc, is pushed to the respective subsystem to perform at leisure, if we introduce something like this maybe we should design something which can offload more generically?

bagder:
Thoughts on this, @ghedo or @nibanks  ?

lib/transfer.c
danielgustafsson: The thing that sort of bothers me is that it feels like a layering violation of `Curl_handler`, in that before adding this we should perhaps see if this is a problem generically faced (but not currently addressed) which we can solve in a bigger picture context.

This might be completely wrong but I think the discussion is worth having before special casing.

nibanks:
If you're trying to drive a QUIC stack you need to be able to run at any time IMO, not just if the app has stuff to send. This is why MsQuic drives this stuff itself. It's no trivial matter to drive the code.

lib/transfer.c
tatsuhiro-t: If we do not have this kind of problem before QUIC is introduced, we do not need to make this generic.
QUIC itself is quite special in a sense that it is a proper transport protocol, the same layer of TCP.
Either way, without this kind of facility, curl is unable to handle QUIC properly.

tatsuhiro-t:
If you have something to send, you can do that along with sending application data.
Other than that, it is basically timer-based.

nibanks:
> If you have something to send, you can do that along with sending application data. Other than that, it is basically timer-based.

I agree. But they why not have a mechanism to indicate the next timeout up to curl instead of just polling in certain states? What if tomorrow, something changes and you need curl to poll in some other state, when there isn't application data? For instance, MsQuic supports the ability to configure keep alives sent periodically at the transport layer. In that scenario, there is no app data and no receives, but just a timer. What about idle timeout? So a good general approach is either to always poll or just give the next timeout to curl and it is responsible for calling you back once it elapses.

tatsuhiro-t:
Curl has Curl_expire function to let QUIC stack set a timer.  When it is expired Curl_readwrite is called.  Then it checks sockets or something and calling in to more protocol specific handler.  But if there is no data to send or write, nothing happens.
The proposed function call is hit in this situation.  If I observed correctly, Curl_readwrite is also called periodically with or without timer.  This is why ngtcp2 Curl_quic_idle checks timer expiry explicitly.  So the current approach is somehow the hybrid of timer based and periodic poll based, but the polling interval is not derived from QUIC stack, it is generally not useful.  It might be better to call the function only when timer is expired.

tatsuhiro-t:
Thank you for the feedback.  It seems like we agreed that some kind method is necessary to send QUIC packet when there is no HTTP data to send.  The proposed function is called when timer is expired or periodically, and it serves the purpose.

The argument about whether this kind of method should be generic or not, I'd like to make it QUIC specific in order to minimize the impact to the code base.  The absence of this kind of functionality suggests that there is no real requirement for non-QUIC use at the moment.

bagder:
@tatsuhiro-t  wrote:

> The absence of this kind of functionality suggests that there is no real requirement for non-QUIC use at the moment.

I agree with this. If we at a later point in time realize we need this feature somewhere else, we can move it then.

tatsuhiro-t:
Fixed commit message, and it is ready to merge.

bagder:
Cool, thanks. I will just wait and ship 7.83.1 first this Wednesday, just to not rock the boat unnecessarily this close to release.

tatsuhiro-t:
Sure, there is no rush.  I hope this would resolve some known issues that cause stalled transfer.

PR #8699, opened 2022-04-12
cookies: Improve errorhandling for reading cookiefile

danielgustafsson:
The existing programming had some issues with errorhandling for reading
the cookie file. If the file failed to open, we would silently ignore it
and continue as if there was no file (or stdin) passed. In this case, we
would also call fclose() on the NULL FILE pointer, which is undefined
behavior. Fix by ensuring that the FILE pointer is set before calling
fclose on it, and error out in case the cookiefile cannot be opened.

Closes: #xxxx

### DIFF
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -1188,12 +1188,15 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
     fp = stdin;
     fromfile = FALSE;
   }
-  else if(file && !*file) {
-    /* points to a "" string */
+  else if(!file || !*file) {
+    /* points to an empty string or NULL */
     fp = NULL;
   }
-  else
-    fp = file?fopen(file, FOPEN_READTEXT):NULL;
+  else {
+    fp = fopen(file, FOPEN_READTEXT);
+    if(!fp)
+      infof(data, "WARNING: failed to open cookie file \"%s\"", file);
+  }
 
   c->newsession = newsession; /* new session? */
 
@@ -1227,7 +1230,7 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
      */
     remove_expired(c);
 
-    if(fromfile)
+    if(fromfile && fp)
       fclose(fp);
   }
 


danielgustafsson:
Erroring out on nonexisting cookiefile is too invasive to backwards compat, so let's not go there.

bagder (approved):

lib/cookie.c
```diff
@@ -1188,12 +1188,17 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
     fp = stdin;
     fromfile = FALSE;
   }
-  else if(file && !*file) {
-    /* points to a "" string */
+  else if((file && !*file) || (!file)) {
```
jay, line 1191:
```suggestion
  else if(!file || !*file) {
```

lib/cookie.c
```diff
@@ -1188,12 +1188,17 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
...
+    /* points to an empty string or NULL */
     fp = NULL;
   }
+  else if(file) {
```
jay, line 1195: this is always going to be true because of block above

lib/cookie.c
```diff
@@ -1188,12 +1188,17 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
...
+    fp = fopen(file, FOPEN_READTEXT);
+    if(!fp)
+      infof(data, "WARNING: failed to open cookie file \"%s\"", file);
+  }
   else
```
jay, line 1200: this is unreachable because of block above

jay (changes requested):

danielgustafsson:
@jay thanks, fixed up with your comments.

lib/cookie.c
```diff
@@ -1227,7 +1230,7 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
      */
     remove_expired(c);
 
-    if(fromfile)
+    if(fromfile && fp)
```
jay, line 1233: this statement is in block if(fp) so you can skip the check here

PR #8714, opened 2022-04-15
fail.d: tweak the description

bagder:
Suggested-by: Robert Charles Muir
Ref: https://twitter.com/rcmuir/status/1514915401574010887

### DIFF
--- a/docs/cmdline-opts/fail.d
+++ b/docs/cmdline-opts/fail.d
@@ -1,15 +1,15 @@
 Long: fail
 Short: f
 Protocols: HTTP
-Help: Fail silently (no output at all) on HTTP errors
+Help: Fail fast with no output on HTTP errors
 See-also: fail-with-body
 Category: important http
 Example: --fail $URL
 Added: 4.0
 ---
-Fail silently (no output at all) on server errors. This is mostly done to
-enable scripts etc to better deal with failed attempts. In normal cases
-when an HTTP server fails to deliver a document, it returns an HTML document
+Fail fast with no output at all on server errors. This is mostly done to
+enable scripts etc to better deal with failed attempts. In normal cases when
+an HTTP server fails to deliver a document, it returns an HTML document
 stating so (which often also describes why and more). This flag will prevent
 curl from outputting that and return error 22.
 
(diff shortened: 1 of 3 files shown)


rmuir:
Thank you for opening the issue!

I didn't intend to attack the documentation, but IMO this PR improves the commandline help (and manpage) a lot. "fail silently" has bad connotations, in some ways is the opposite of what this flag will do for you. It is great to avoid the phrase especially for the small summary.

docs/cmdline-opts/fail.d
```diff
@@ -1,15 +1,15 @@
...
+Help: Fail fast with no output on HTTP errors
 See-also: fail-with-body
 Category: important http
 Example: --fail $URL
 Added: 4.0
 ---
-Fail silently (no output at all) on server errors. This is mostly done to
-enable scripts etc to better deal with failed attempts. In normal cases
-when an HTTP server fails to deliver a document, it returns an HTML document
+Fail fast with no output at all on server errors. This is mostly done to
```
danielgustafsson, line 10: How about s/done/useful/ here?

docs/cmdline-opts/fail.d
```diff
@@ -1,15 +1,15 @@
 Long: fail
 Short: f
 Protocols: HTTP
-Help: Fail silently (no output at all) on HTTP errors
+Help: Fail fast with no output on HTTP errors
```
danielgustafsson, line 4: Is "Fail fast" actually true, it doesn't really fail faster than any other way (apart from not outputting) or than the server allows it to. I don't have a better idea off the cuff, but fast seem so imply something it doesn't do.

bagder: It does fail "fast" in the way that it stops reading the response after the headers if the response code tells it to.

docs/cmdline-opts/fail.d
bagder: Totally reasonable!

bagder: I changed some more words too. Check the update!

docs/cmdline-opts/fail.d
danielgustafsson: Fair.

danielgustafsson (approved):
LGTM

Issue #8705, opened 2022-04-13
Build curl lib with Mingw compiler


medbsl:
I want to use the curl lib in my project (c++ QT creator on windows), I download the zip lib file from the website https://curl.se/download.html (Version 7.82.0) and I wanted to build it using the Makefile with this commend mingw32-make mingw32 as described in the documentation https://curl.se/docs/install.html

but unfortunately, I am getting an error

mingw32-make mingw32
[....]
gcc   -s -m32 -static -o curl.exe curl.res slist_wc.o tool_binmode.o tool_bname.o 
tool_cb_dbg.o tool_cb_hdr.o tool_cb_prg.o tool_cb_rea.o tool_cb_see.o tool_cb_wrt.o 
tool_cfgable.o tool_dirhie.o tool_doswin.o tool_easysrc.o tool_filetime.o tool_findfile.o 
tool_formparse.o tool_getparam.o tool_getpass.o tool_help.o tool_helpers.o tool_hugehelp.o 
tool_libinfo.o tool_listhelp.o tool_main.o tool_msgs.o tool_operate.o tool_operhlp.o 
tool_panykey.o tool_paramhlp.o tool_parsecfg.o tool_progress.o tool_strdup.o tool_setopt.o 
tool_sleep.o tool_urlglob.o tool_util.o tool_vms.o tool_writeout.o tool_writeout_json.o 
tool_xattr.o strtoofft.o nonblock.o warnless.o curl_ctype.o curl_multibyte.o version_win32.o 
dynbuf.o -L../lib -lcurl -lwldap32 -lws2_32
../lib\libcurl.a: error adding symbols: Archive has no index; run ranlib to add one
collect2.exe: error: ld returned 1 exit status
make[1]: *** [curl.exe] Fehler 1
make[1]: Leaving directory `D:/Dev/API/curl-7.82.0/src'
make: *** [mingw32] Fehler 2

I tried to run ranlib lib\libcurl.a but the issue is still there,

does anyone have an idea about it?

One more question: is it possible also to build it with mingw64?

bagder:
Both mingw32 and mignw64 should work. but I suspect that makefile target is not working.

If you want to build with mingw, then you should probably go all msys and do configure + make the autotools style: https://everything.curl.dev/build/autotools

Thoughts @jay or @vszakats ?

vszakats:
[...]

PR #8717, opened 2022-04-16
docs/INSTALL: refer mingw users to invoke configure

bagder:
Remove Makefile.dist from the repository and the dist tarball.

Ref: #8705


jay:
I'm fine with this but I'll defer to Viktor since he's using it in curl-for-win.

mnaser:
recheck

vszakats:
This will make it more difficult to launch `Makefile.m32`-based builds, and curl-for-win scripts will have to be updated accordingly (looks to be a simple update). It seems this build method becomes even more niche after this point. I still prefer it over autotools/cmake, but it'd seem even more urgent to switch to one of these mainstream ones. Any patch or help would be more than welcome here. (I had one try and fail with cmake in the past.)

bagder:
Nah sorry, I misunderstood and thought this wasn't used. Since it is actually in use, I don't feel a need to rush this now.

Issue #8708, opened 2022-04-14
fd gets closed twice when using libssh for sftp


JayDommaschk:
### I did this

Hello,
I made the following simple test program to demonstrate the issue:

```cpp
#include <curl/curl.h>
#include <sys/stat.h>
#include <unistd.h>

int main()
{
    FILE *file = fopen("/tmp/test.c", "rb");
    struct stat file_info;
    stat("/tmp/test.c", &file_info);

    char SERVER_URL[512];
    CURLcode res = CURLE_OK;
    CURL *curl = curl_easy_init();

    curl_version_info_data *ver = curl_version_info(CURLVERSION_NOW);
    printf("curl version: %u.%u.%u\n", (ver->version_num >> 16) & 0xff,
                                       (ver->version_num >> 8) & 0xff,
                                       ver->version_num & 0xff);

    curl_easy_setopt(curl, CURLOPT_USERNAME, "user");
    curl_easy_setopt(curl, CURLOPT_PASSWORD, "password");

    snprintf(SERVER_URL, sizeof(SERVER_URL), "sftp://%s:%s/~/%s", "192.168.2.11", "22", "curltest.txt");
    curl_easy_setopt(curl, CURLOPT_URL, SERVER_URL);

    curl_easy_setopt(curl, CURLOPT_INFILESIZE_LARGE, (curl_off_t)(file_info.st_size));
    curl_easy_setopt(curl, CURLOPT_FTP_CREATE_MISSING_DIRS, CURLFTP_CREATE_DIR_RETRY);

    curl_easy_setopt(curl, CURLOPT_READDATA, file);
    curl_easy_setopt(curl, CURLOPT_UPLOAD, 1L);

    curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);

    res = curl_easy_perform(curl);

    printf("curl result: %s (%d)\n", curl_easy_strerror(res), res);
}
```
Now if I compile it and run "strace -e trace=open,close" on it, I get the following:

```
strace -e trace=open,close ./curltest 
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
[...]

PR #8718, opened 2022-04-18
fix libssh double socket close

JayDommaschk:
This is meant so solve issue #8708 

In myssh_connect in lib/vssh/libssh.c, the fd of conn->sock[FIRSTSOCKET] is assigned to ssh->ssh_session (line 2219). That gets closed on ssh_disconnect (line 1972). By setting conn->sock[FIRSTSOCKET] to CURL_SOCKET_BAD afterwards, we can avoid closing the fd a second time on cleanup of conn.

### DIFF
--- a/lib/vssh/libssh.c
+++ b/lib/vssh/libssh.c
@@ -1970,6 +1970,10 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
       }
 
       ssh_disconnect(sshc->ssh_session);
+      /* conn->sock[FIRSTSOCKET] is closed by ssh_disconnect behind our back,
+         explicitly mark it as closed with the memdebug macro: */
+      fake_sclose(conn->sock[FIRSTSOCKET]);
+      conn->sock[FIRSTSOCKET] = CURL_SOCKET_BAD;
 
       SSH_STRING_FREE_CHAR(sshc->homedir);
       data->state.most_recent_ftp_entrypath = NULL;


bagder:
Thank for your contribution, but this doesn't seem like an entirely good fix. As this build shows:

https://github.com/curl/curl/runs/6059033071?check_suite_focus=true#step:5:3560

... it cases build failures because the memdebug system only sees the socket getting creating but never closed. Are you *sure* libssh itself (always) closes the socket in this case? It seems like something it shouldn't do. Also what libssh versions are you testing with?

The [libssh docs](https://api.libssh.org/master/group__libssh__session.html#ga0f048a4c0dbe02cfb7e9c5b6d0db0f27) is wonderfully brief so it doesn't help much. Maybe we need to ask the team over there?

bagder:
See https://gitlab.com/libssh/libssh-mirror/-/issues/122

JayDommaschk:
Sorry for the bad fix!

As you already said, libssh seems to close the socket unconditionally. I tested with version if 0.9.6 of libssh. (but also with some slightly older version, 0.9.2, on a different machine)

So I will happily look into why the build failes. But let's see first what the folks as libssh say about the issue, maybe they will already bring some enlightening.

bagder:
There's a now a pending fix for libssh to stop closing our socket here: https://gitlab.com/libssh/libssh-mirror/-/merge_requests/240

This puts us in the difficult spot that we still need to handle all the libssh versions before this fix. So we need add some kind of hack that tells the memdebug system that libssh closed the socket for us... Let me give that a try.

bagder:
Sorry for force-pushing, but I messed it up and wanted to clean it up a little.

@JayDommaschk would you like me to rather make the commit using the full name "Jay Dommaschk" to give you proper credit?

JayDommaschk:
While my ego might like receiving credit, I wouldn't mind if I didn't. The most important thing is that this problem is fixed. If anything, that would motivate me to fix more problems to get credit in the future ;) 

Anyway, thanks a lot for everything!

bagder:
You will get the credit all right, I was mostly asking how you want your name written in the commit. I'll use your full name as mentioned above. Merging!

bagder:
Thanks!

PR #8719, opened 2022-04-18
unit1620: call global_init before calling Curl_open

jay:
Curl_open calls the resolver init and on Windows if the resolver backend
is c-ares then the Windows sockets library (winsock) must already have
been initialized (via global init).

Ref: https://github.com/curl/curl/pull/8540#issuecomment-1059771800

Closes #xxxx

--

/cc @gvanem


bagder (approved):
Seems reasonable!

PR #8727, opened 2022-04-20
cookies: Statement is in block if(fp) skip check in Line:1233

pheiduck:
https://github.com/curl/curl/pull/8699#discussion_r851583907
Suggested-by: Jay Satiro <raysatiro@yahoo.com>

### DIFF
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -1230,7 +1230,7 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
      */
     remove_expired(c);
 
-    if(fromfile && fp)
+    if(fromfile)
       fclose(fp);
   }
 


lib/cookie.c
```diff
@@ -1230,7 +1230,7 @@ struct CookieInfo *Curl_cookie_init(struct Curl_easy *data,
      */
     remove_expired(c);
 
-    if(fromfile && fp)
+    if(fromfile)
```
danielgustafsson, line 1233: If `fopen()` fails on line 1197, `fromfile` is set but `fp` isn't and shouldn't be `fclose`d. Thus, we need this check.

pheiduck: I see, then this is not important.

jay: > Thus, we need this check.

no we don't, it's in the if(fp) block

danielgustafsson:
> no we don't, it's in the if(fp) block

Sorry, I was a bit thick, *now* I see what you mean. Will fix.

PR #8730, opened 2022-04-21
docs: updates spellings with full words

kushaldas:
For clearing meanings.


bagder (approved):

bagder:
Thanks! (I also removed a few contractions in those files while at it)

PR #8738, opened 2022-04-24
ngtcp2: Deal with sub-millisecond timeout

tatsuhiro-t:


lib/vquic/ngtcp2.c
```diff
@@ -1830,6 +1830,9 @@ static CURLcode ng_flush_egress(struct Curl_easy *data,
     }
     else {
       timeout = expiry - ts;
+      if(timeout / NGTCP2_MILLISECONDS * NGTCP2_MILLISECONDS < timeout) {
+        timeout += NGTCP2_MILLISECONDS;
+      }
```
bagder, line 1835: This logic is a bit funny-looking and I think it could be rephrased to become easier to understand, but if not it could be explained by a comment at least.

Couldn't it instead be written like this? (or maybe I  misunderstand it)
~~~
if(timeout % NGTCP2_MILLISECONDS)
  timeout += NGTCP2_MILLISECONDS;

tatsuhiro-t: Fix committed via 0593e0259ac27c2351fd1195ff075a6a848d9105

bagder (approved):

PR #8740, opened 2022-04-24
mprintf: Ignore non-literal format string

gjasny:
Using the non-literal format string here is intentional and the warning leads to noise in the best case or an error in the worst.

### DIFF
--- a/lib/mprintf.c
+++ b/lib/mprintf.c
@@ -956,9 +956,16 @@ static int dprintf_formatf(
 
         *fptr = 0; /* and a final zero termination */
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-nonliteral"
+#endif
         /* NOTE NOTE NOTE!! Not all sprintf implementations return number of
            output characters */
         (sprintf)(work, formatbuf, p->data.dnum);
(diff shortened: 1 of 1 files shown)


bagder:
How do you to get this warning?

gjasny:
We have a custom CMake toolchain that sets `-Wformat=2` in the `CMAKE_C_FLAGS`.

danielgustafsson:
What if someone has a toolchain or process which relies on trapping that?

gjasny:
AFAIU the compiler warning was added to not pass a user-controlled format string to printf like in this example:

```c
void log(const char* msg) {
  printf(msg); // use printf("%s", msg) instead
}
```

As heuristic for non-user controlled strings the compiler looks for a string literal (which is always compile-time defined). But here in `dprintf_formatf` the format string is assembled on-the-fly but still not user-controlled. This renders the compiler warning into a false-positive because the format string is not user-controlled.

To suppress the warning we save all the enabled warnings, disable the `format-nonliteral` warning and after the function call restore all the warnings.

I cannot think of a case where one would like to see the warning anyways.

bagder (approved):
Seems small and innocuous enough as long as we don't have to do this dance on many places.

bagder:
Thanks!

PR #8745, opened 2022-04-25
libssh2: make the md5 comparison fail if wrong length

bagder:
Making it just skip the check unless exactly 32 is too brittle. Even if
the docs says it needs to be exactly 32, it is be safer to make the
comparison fail here instead.

Reported-by: Harry Sintonen
Bug: https://hackerone.com/reports/1549461


piru (commented):
`if(pubkey_md5) {` is unneccesary here. It is of course harmless but it is not really needed.

PR #8757, opened 2022-04-27
Fix do_pubkey not able to handle EC public keys

sergio-nsk:


bagder:
How do you test/experience this problem and fix?

sergio-nsk:
Verified on https://www.cloudflare.com/ by using `CURLINFO_CERTINFO`, that returned `num_of_certs` is 2, but `certinfo` points to `NULL`.

lib/vtls/x509asn1.c
```diff
@@ -945,6 +945,29 @@ static int do_pubkey(struct Curl_easy *data, int certnum,
...
+     * ECC public key is all the data, a value of type BIT STRING mapped to
+     * OCTET STRING. The first octet of the OCTET STRING indicates whether the
+     * key is compressed or uncompressed.
+     */
+    const unsigned long len =
+      (unsigned long)((pubkey->end - pubkey->beg - 2) * 4);
+    if(!certnum)
+      infof(data, "   ECC Public Key (%lu bits)", len);
+    if(data->set.ssl.certinfo) {
+      const char *q = curl_maprintf("%lu", len);
```
bagder, line 959: I think this would be better as a buffer on the stack so that it doesn't risk failing a malloc here (this function also doesn't handle a malloc failure properly here)

bagder (changes requested):

lib/vtls/x509asn1.c
sergio-nsk: What do you mean? `q = ...; if (q) { ...; free(q); }` - how can it improperly handle a malloc failure? This code is borrowed from the bellow lines https://github.com/curl/curl/pull/8757/files#diff-7a2f5f9d345bd490dc6054928fc0785ced111e5687914a4e6c9979e70e0e4c9aR997

sergio-nsk: If you want a stack buffer here, then other parts of the function should be modified for consistency. Do we this?

bagder: I meant that for a malloc failure, it should bail out everything completely without leaking memory and return error. Avoiding the malloc uses less resources, is faster and does not risk a malloc failure!

bagder (approved):

bagder:
Thanks!

PR #8762, opened 2022-04-27
Pass remote_port to MsH3ConnectionOpen

unasuke:
## What

MsH3 supported additional "Port" parameter to connect not hosted on 443 port QUIC website.

* https://github.com/nibanks/msh3/releases/tag/v0.3.0
* https://github.com/nibanks/msh3/pull/37

image

### DIFF
--- a/lib/vquic/msh3.c
+++ b/lib/vquic/msh3.c
@@ -95,7 +95,9 @@ static const MSH3_REQUEST_IF msh3_request_if = {
 
 void Curl_quic_ver(char *p, size_t len)
 {
-  (void)msnprintf(p, len, "msh3/%s", "0.3.0");
+  uint32_t v[4];
+  MsH3Version(v);
+  (void)msnprintf(p, len, "msh3/%d.%d.%d.%d", v[0], v[1], v[2], v[3]);
 }
 
 CURLcode Curl_quic_connect(struct Curl_easy *data,
--- a/lib/vquic/msh3.c
+++ b/lib/vquic/msh3.c
@@ -95,7 +95,7 @@ static const MSH3_REQUEST_IF msh3_request_if = {
 
 void Curl_quic_ver(char *p, size_t len)
 {
-  (void)msnprintf(p, len, "msh3/%s", "0.0.1");
+  (void)msnprintf(p, len, "msh3/%s", "0.3.0");
 }
 
 CURLcode Curl_quic_connect(struct Curl_easy *data,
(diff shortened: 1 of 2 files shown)


lib/vquic/msh3.c
```diff
@@ -95,7 +95,7 @@ static const MSH3_REQUEST_IF msh3_request_if = {
 
 void Curl_quic_ver(char *p, size_t len)
 {
-  (void)msnprintf(p, len, "msh3/%s", "0.0.1");
+  (void)msnprintf(p, len, "msh3/%s", "0.3.0");
```
danielgustafsson, line 98: This seems brittle, can we persuade the good msh3 folks to add version information to `msh3.h` which we can consume?

lib/vquic/msh3.c
```diff
@@ -121,7 +121,10 @@ CURLcode Curl_quic_connect(struct Curl_easy *data,
     return CURLE_FAILED_INIT;
   }
 
-  qs->conn = MsH3ConnectionOpen(qs->api, conn->host.name, unsecure);
+  qs->conn = MsH3ConnectionOpen(qs->api,
+                                conn->host.name,
+                                (uint16_t)conn->remote_port,
+                                unsecure);
```
danielgustafsson, line 127: This will put a hard requirement on a very recent version of the library, that should go into the `--with-msh3` autoconf check to test for this signature no?

lib/vquic/msh3.c
nibanks: I agree! That's why I created https://github.com/nibanks/msh3/issues/23. Feel free to take a crack at it.

nibanks (commented):
LGTM 👍

lib/vquic/msh3.c
unasuke: RIght. it's useful if version string API in mash3. @nibanks What do you think about?

lib/vquic/msh3.c
unasuke: I think it's very safe if there is a signature checking, but too early to introduce it probably.

bagder: I think for now it is fine to require a very recent version of msh3 to build. It is early days and I assume it might move a bit going forward. We've worked similar to that with the other h3 libraries. They're all "beta" and this is functionality still labeled "experimental" in curl.

unasuke:
It is unlikely that I will be able to get this to pass all the tests...

lib/vquic/msh3.c
bagder: As https://github.com/nibanks/msh3/pull/38 has been merged now, this can now get the runtime version!

bagder:
> It is unlikely that I will be able to get this to pass all the tests...

I'm afraid you see a lot of unrelated flaky CI build failures right now. They're not happening because of anything in this PR.

lib/vquic/msh3.c
unasuke: Ok! I'll try to it.

unasuke: @bagder I addressed it at deb1bc3

image

nibanks: LGTM!

bagder:
Thanks!

PR #8763, opened 2022-04-27
msh3: print boolean value as text representation

danielgustafsson:
Print the boolean value as its string representation instead of with `%hhu` which isn't a format we typically use. I don't have strong feelings wrt this in case others disagree, but `%hhu` struck me as odd when skimming the code.

### DIFF
--- a/lib/vquic/msh3.c
+++ b/lib/vquic/msh3.c
@@ -357,7 +357,7 @@ static void MSH3_CALL msh3_complete(MSH3_REQUEST *Request, void *IfContext,
   struct HTTP *stream = IfContext;
   (void)Request;
   (void)AbortError;
-  H3BUGF(printf("* msh3_complete, aborted=%hhu\n", Aborted));
+  H3BUGF(printf("* msh3_complete, aborted=%s\n", Aborted?"true":"false"));
   msh3_lock_acquire(&stream->recv_lock);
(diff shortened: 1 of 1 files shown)


nibanks (commented):
LGTM 👍Though I'd add extra spaces between the `?` and `:`, but that's just me!

danielgustafsson:
> LGTM 👍Though I'd add extra spaces between the `?` and `:`, but that's just me!

So embarrassingly enough I was about to say that `checksrc.pl` mandates the lack of spaces, but it actually doesn't. I feel like I should've known that.

So, let's turn to science to see what to do:
```sh
 $ git grep "?\"[^:]\+:\"" | wc -l
      60
 $ git grep "? \"[^:]\+ : \"" | wc -l
      65
```
It's pretty close but adding spaces clearly wins, so I will go do that before merging.

That got me curious though, what other permutations turn up?
```sh
 $ git grep "? \"[^:]\+[^ ]: \"" | wc -l
      35
 $ git grep "? \"[^:]\+ :\"" | wc -l
       1
```
Maybe we should have a rule about this after all? =)

bagder:
> Maybe we should have a rule about this after all? =)

Hehe, I was also convinced checksrc already checked for this, but apparently not...

PR #8769, opened 2022-04-29
CI/macos: do a brew upgrade to keep packages up-to-date

pheiduck:


.github/workflows/macos.yml
```diff
@@ -81,7 +81,7 @@ jobs:
     # Run this command with retries because of spurious failures seen
     # while running the tests, for example
     # https://github.com/curl/curl/runs/4095721123?check_suite_focus=true
-    - run: brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done
+    - run: brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done && brew upgrade
```
danielgustafsson, line 84: Since we run a `brew update` first, and `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK` isn't set then we should have all the latest versions already no?

pheiduck: the current CI exit with this "error":
```
You have 13 outdated formulae and 2 outdated casks installed.
You can upgrade them with brew upgrade
or list them with brew outdated.
```

pheiduck: another one is this:
```
fatal: unable to access 'https://github.com/mongodb/homebrew-brew/': The requested URL returned error: [4](https://github.com/curl/curl/runs/6224510054?check_suite_focus=true#step:3:4)29
Error: Fetching /usr/local/Homebrew/Library/Taps/mongodb/homebrew-brew failed!
```

pheiduck: brew update updates the downloaded git repository with the latest code from GitHub.
brew upgrade updates the actual packages to match the versions in the updated local git repository.

maybe we should replace update with upgrade

pheiduck: Those packages installed by the CI image, so its not affect the brew bundle used by us.

danielgustafsson: According to my understanding, `brew update` will ensure that packages in our bundle are up to date, so we should definitely keep doing that. `brew upgrade` will update whatever packages were installed in the base image which also aren't dependencies of any the packages in the bundle. The latter should be a safe operation.

PR #8775, opened 2022-04-29
http: move Curl_allow_auth_to_host()

### LINKED ISSUE #8772: building with --disable-http-auth fails with "undefined reference to `Curl_allow_auth_to_host'"
### I did this
configure with `--disable-http-auth`

### I expected the following
building curl shoud succeed

### curl/libcurl version
7.83.0

The problem is here: https://github.com/curl/curl/blob/685170b729a1b704233036381cc66b1275732af4/lib/http.c#L1920

This call to `Curl_allow_auth_to_host()` is not protected by a `CURL_DISABLE_HTTP_AUTH` check but the implementation of the function is.

bagder:
It was mistakenly put within the CURL_DISABLE_HTTP_AUTH #ifdef

Reported-by: Michael Olbrich
Fixes #8772

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -651,6 +651,21 @@ CURLcode Curl_http_auth_act(struct Curl_easy *data)
   return result;
 }
 
+/*
+ * Curl_allow_auth_to_host() tells if authentication, cookies or other
+ * "sensitive data" can (still) be sent to this host.
+ */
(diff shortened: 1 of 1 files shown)


bagder:
Closed by commit d7b970e46ba29a7e558e21d19f485977ffed6266

PR #8777, opened 2022-04-29
md files and scripts: use autoreconf -fi instead buildconf

pheiduck:
Signed-off-by: Philip H <47042125+pheiduck@users.noreply.github.com>

### DIFF
--- a/scripts/coverage.sh
+++ b/scripts/coverage.sh
@@ -21,7 +21,7 @@
 #
 ###########################################################################
 
-/usr/bin/autoreconf -fi
+autoreconf -fi
 mkdir -p cvr
 cd cvr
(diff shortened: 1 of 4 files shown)


scripts/coverage.sh
```diff
@@ -21,7 +21,7 @@
 #
 ###########################################################################
 
-autoreconf -fi
+/usr/bin/autoreconf -fi
```
bagder, line 24: Why should this one use an absolute path when no others do?

pheiduck: revet it here: [revert: absolute path and whitespace](https://github.com/curl/curl/pull/8777/commits/ff663e5a652f0b28c9292f8642c01cbcaa2204f4)
thought it would be useful in a bash script

bagder (approved):

bagder:
Thanks!

PR #8780, opened 2022-04-30
[ci] use clang-14 on CIs which needs clang for dependencies

pheiduck:
Signed-off-by: Philip H <47042125+pheiduck@users.noreply.github.com>


pheiduck:
brainstormed about this again: maybe better to use the pkgs as they come from the CI image

PR #8782, opened 2022-04-30
gskit: remove unused function set_callback

danielgustafsson:
This function has been unused since the initial commit of the GSKit backend in 0eba02fd41263218a0eaedbe65af739a9bb2da63. @monnerat, am I missing something?

### DIFF
--- a/lib/vtls/gskit.c
+++ b/lib/vtls/gskit.c
@@ -293,27 +293,6 @@ static CURLcode set_numeric(struct Curl_easy *data,
 }
 
 
-static CURLcode set_callback(struct Curl_easy *data,
-                             gsk_handle h, GSK_CALLBACK_ID id, void *info)
-{
-  char buffer[STRERROR_LEN];
-  int rc = gsk_attribute_set_callback(h, id, info);
-
-  switch(rc) {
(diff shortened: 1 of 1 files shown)


monnerat:
You're right, it's unused.

My motivation for itt was getting the whole certificate chain: the only place where the latter is available is as a callback parameter and I planned to get the certs from there.

Unfortunately this is not possible to pass a user pointer to this callback, effectively precluding the possibility to associate the cert chain with a data/conn structure.

There's a comment about it at https://github.com/curl/curl/blob/a7b29122cb35051b65780f37fe2dcaed4f31386b/lib/vtls/gskit.c#L1070-L1074.

See also https://www.ibm.com/docs/api/v1/content/ssw_ibm_i_71/apis/gsk_attribute_set_callback.htm and search for `pgsk_cert_validation_callback` on the page.

I kept the function around in hope IBM would add such a parameter but they never did. 9 years later, I'm convinced they'll never do thus we'll never have the whole chain in CAINFO and I approve this removal.

Thanks for looking at it.

danielgustafsson:
Thanks for clarifying!

PR #8791, opened 2022-05-04
test{898,974,976}: add 'HTTP proxy' keywords

fabiankeil:
... so the tests can be automatically skipped when
testing external HTTP proxies like Privoxy.


bagder (approved):

bagder:
Thanks!

PR #8793, opened 2022-05-04
gskit_connect_step1: fixed bogus setsockopt calls

piru:
setsockopt takes a reference to value, not value. With the current code this just leads to -1 return value with errno EFAULT.


bagder (approved):
It is curious that it has been this wrong all since cb4e2be7c6 when these lines were introduced...

bagder:
Thanks!

monnerat:
> It is curious that it has been this wrong all since https://github.com/curl/curl/commit/cb4e2be7c6d42ca0780f8e0a747cecf9ba45f151 when these lines were introduced...

2 possible explanations:

1. The system default values are satisfactory
2. No one ever securely connected via an https proxy from OS/400 libcurl (I don't remember if I tested it) !

PR #8796, opened 2022-05-04
mbed_connect_step1: bail out if rng init fails

piru:
There was a failf() call but no actual error return.

It is likely that the anything using the rng later would fail due errors anyway, but better bail out right away, I think.


bagder (approved):

bagder:
Thanks!

PR #8797, opened 2022-05-05
Relax UndefinedBehavior sanitizer

sergio-nsk:
`while(i--)` causes runtime error: unsigned integer overflow: 0 - 1 cannot be
represented in type 'size_t' (aka 'unsigned long')


sergio-nsk (commented):
[Here](https://github.com/curl/curl/pull/8797/files#diff-72e7d4123a3c69e8c31616bfea3a66f88afb86864cbaaeab76e455c0d4c19a72R254)

bagder (approved):
A benign use, but I can't see any harm in this change if it silences a warning here.

bagder:
Thanks!

PR #8798, opened 2022-05-05
sectransp_connect_step1: bail out if SSLSetPeerDomainName fails

piru:
Before the code would only warn about SSLSetPeerDomainName() errors.

Caller requested host verification and if we can't do it we must fail to connect.

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -2045,8 +2045,9 @@ static CURLcode sectransp_connect_step1(struct Curl_easy *data,
     err = SSLSetPeerDomainName(backend->ssl_ctx, snihost, snilen);
 
     if(err != noErr) {
-      infof(data, "WARNING: SSL: SSLSetPeerDomainName() failed: OSStatus %d",
+      failf(data, "SSL: SSLSetPeerDomainName() failed: OSStatus %d",
             err);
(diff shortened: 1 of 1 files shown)


jay:
Was the intention to continue even if SSLSetPeerDomainName fails? There is a "WARNING" there which makes me think so. @nickzman

nickzman:
I'm not sure if this is even necessary. If you look at [the function's internal implementation](https://github.com/apple-oss-distributions/Security/blob/2ac04335f51a001d0495c0119af8d352100fa770/OSX/libsecurity_ssl/lib/sslContext.c#L897), you'll see that it fails only under two circumstances: the context is not set, and the session is already active. Neither of which will ever be hit unless the code in this function gets scrambled, so I will be very surprised if this function fails for anyone anywhere.

@piru , what are you trying to accomplish?

piru:
@nickzman Defensive programming. The API call has an error code. If the error code is set, the operation can be assumed to have failed. Continuing when this operation has failed will (likely) result in security check bypass (hostname will not be validated).

Now, if the function can practically never fail, what's the harm in checking the result code and bailing out if error occurred?

Also:

What is this code trying to accomplish by continuing even if error occurred?

bagder (approved):

bagder:
Thanks!

PR #8800, opened 2022-05-05
fixup: redundancy on contributors

pheiduck:
pheiduck on github <47042125+pheiduck@users.noreply.github.com> thats me
Philip H <47042125+pheiduck@users.noreply.github.com> thats as well

we should use only Philip H <47042125+pheiduck@users.noreply.github.com>


bagder:
Thanks!

PR #8801, opened 2022-05-05
forgot to add Philip H

pheiduck:
follow up https://github.com/curl/curl/pull/8800


bagder:
There's no need to manually add people to `docs/THANKS`. I update that set with `scripts/contrithanks.sh` in association with every release.

pheiduck:
@bagder I was not aware of that, then I close this PR.

PR #8803, opened 2022-05-05
libcurl: add option to disallow cookies from HTTP 100-199

danielgustafsson:
RFC 6265 section 3 specifies that clients MAY ignore Set-Cookie headers in 100-level (informational) responses. This is further specified in draft-ietf-httpbis-rfc6265bis-09 section 5.3 which maintains the wording that clients "MAY ignore".

This adds an option to libcurl to ignore Set-Cookie headers in 100-level responses. The default value is to allow cookies set in 100-level responses in order to preserve backwards compat.


dfandrich:
Is this actually useful? Are there servers out there that present Set-Cookie
header in 100-level responses that are not desired for some reason? Are there
real cases where lacking this option breaks something?

danielgustafsson:
> Is this actually useful?

I honestly don't really know, I implemented it mainly since I want libcurl to closely match the state mechanism RFC (and it seemed like a fun hack, which was something I needed). It was added in [draft-22](https://datatracker.ietf.org/doc/html/draft-ietf-httpstate-cookie-22) in the lead up to RFC 6265 but I'm not sure if the reasoning/discussions can be found archived somewhere.

> Are there servers out there that present Set-Cookie header in 100-level responses that are not desired for some reason? 

At this point I expect there to be a malicious use of every benign feature on the Internet.

> Are there real cases where lacking this option breaks something?

Not that I can think of, it's yet another way to protect the against potential malicious acts by rogue servers.

dfandrich:
If a rouge server can set a cookie in a 100 response, it could set it in a 200
response, so I don't see how this protects against anything.

bagder:
Yeah, I'm with @dfandrich here and I'm not convinced there's a need or desire for this option.

PR #8804, opened 2022-05-05
Fix "make ca-firefox"

ryandesign:
With curl 7.83.0, `make ca-firefox` fails with:

```
make: *** No rule to make target `lib/firefox-db2pem.sh', needed by `ca-firefox'.  Stop.
```

firefox-db2pem.sh was [moved from lib to scritps](https://github.com/curl/curl/commit/8e22fc68e7dda43e9f0b6857b1057d0e9131a4b2).

After fixing the paths in the Makefile it fails with:

```
generate a fresh ca-bundle.crt
scripts/firefox-db2pem.sh lib/ca-bundle.crt
make: scripts/firefox-db2pem.sh: Permission denied
make: *** [ca-firefox] Error 1
```

That's fixed with a further `chmod a+x`.


bagder (approved):
Would it perhaps be better to run the script with sh to not rely on `+x` ?

bagder:
Ah right, so let's keep it for consistency

bagder:
Thanks!

PR #8808, opened 2022-05-06
http2: always debug print stream id in decimal

steini2000:
Sometimes the stream id is printed with %x and sometimes with %d or %u. Confusing.

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -643,7 +643,7 @@ static int push_promise(struct Curl_easy *data,
                                               frame->promised_stream_id,
                                               newhandle);
     if(rv) {
-      infof(data, "failed to set user_data for stream %d",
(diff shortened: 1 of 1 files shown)


bagder (commented):
How about using %u instead? It seems a lot of places already use that, so it would make it even more consistent.

steini2000:
Hmm. Sometimes %d is used and sometimes %u. Data type is int32_t, so %d seemed to be more appropriate for me. Maybe I should change all stream id printf to either %u or %d. Just let me know what you prefer.

bagder:
Streams are identified with an unsigned 31-bit integer, so it doesn't matter which type we use. They are still unsigned. I think using %u makes that clearer in the code.

jay:
Thanks

PR #8811, opened 2022-05-07
Fix 357 For Hyper

gsquire:
This patch fixes the hyper API such that PUT requests that receive a 417 response can retry without the Expect header.

I followed the original implementation in #4964 but I'm not sure if there is anything else we need to validate in this API.


bagder (approved):

bagder:
Thanks!

PR #8821, opened 2022-05-09
fix handling of trailing host name dot for HSTS

bagder:


lib/hsts.c
```diff
@@ -238,10 +247,21 @@ struct stsentry *Curl_hsts(struct hsts *h, const char *hostname,
...
+    char buffer[MAX_HSTS_HOSTLEN + 1];
     time_t now = time(NULL);
     size_t hlen = strlen(hostname);
     struct Curl_llist_element *e;
     struct Curl_llist_element *n;
+
+    if((hlen > MAX_HSTS_HOSTLEN) || !hlen)
+      return NULL;
+    memcpy(buffer, hostname, hlen);
+    if(hostname[hlen-1] == '.')
```
jay, line 259:
```suggestion
    if(hlen && hostname[hlen-1] == '.')
```
It looks like this function purges expired entries, so should it run even when !hlen?

bagder: It is only ever called with a (valid) host host name, so I don't see how it can be zero length without that being an error somewhere earlier.

Issue #8818, opened 2022-05-09
(🐞) Double quotes are removed from `-d`/`--data`/`--data-raw` on Windows


KotlinIsland:
### I did this
`curl -v -d '"a"'`

### I expected the following
`> Content-Length: 3`

### Instead I actually got
`> Content-Length: 1`

### curl/libcurl version

```
> curl -V
curl 7.79.1 (Windows) libcurl/7.79.1 Schannel
Release-Date: 2021-09-22
Protocols: dict file ftp ftps http https imap imaps pop3 pop3s smtp smtps telnet tftp
Features: AsynchDNS HSTS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI UnixSockets
```
### operating system
Windows 10 21H2

### FYI
On `curl 7.81.0 (x86_64-pc-linux-gnu)` this works correctly and doesn't remove the double quotes.

### Additional info
```powershell
> curl -v -d '"' "https://example.com"
curl: no URL specified!
```
Hmm, I think there is a url specified.

jzakrzewski:
AFAIK Windows CMD does not consider single quotes as special, so you have to quote the parameters using double quotes and escape any double quotes that are a part of the parameters.
This is a shell thing, not a curl issue.

jay:
The Windows command prompt does not use literal quotes like Unix shells do. You will have to double quote the argument and then escape inner double quotes with a backslash `\`. I made a program called [getargs](https://gist.github.com/jay/19aba48653bd591cf4b90eb9249a302c) that can show the arguments passed to it:

~~~
getargs -v -d '"' "https://example.com"

argv[0]: getargs
argv[1]: -v
argv[2]: -d
argv[3]: '' https://example.com
~~~

The command shell interpreter is removing the double quotes and keeping the single quotes, which is likely not what you intended.

~~~
getargs  -v -d '"a"'

argv[0]: getargs
argv[1]: -v
argv[2]: -d
argv[3]: 'a'
~~~

~~~
getargs -v -d "\"a\""

argv[0]: getargs
argv[1]: -v
argv[2]: -d
argv[3]: "a"
~~~

KotlinIsland:
@jzakrzewski @jay This isn't using CMD, this is using PowerShell I didn't even bother using CMD.

CMD:
```cmd
> echo '"' a   b  '"'
'"' a   b  '"'

PS:
```powershell
> echo '"' a   b '"'
"
a
b
"
```

[...]

PR #8823, opened 2022-05-09
FAQ: Clarify Windows double quote usage

jay:
- Windows command prompt doesn't use literal quoting via single quotes.

- Windows command prompt inner double quotes are escaped with a
  backslash.

- Windows powershell does use single quotes but curl is not a powershell
  script so the arguments may not be passed on correctly.

- Windows powershell inner double quotes seems can be passed to curl if
  the outer quotes are double quotes and an escape of backslash-backtick
  is used.

Command prompt example:

~~~
getargs -v -d "\"a\""

argv[0]: getargs
argv[1]: -v
argv[2]: -d
argv[3]: "a"
~~~

Ref: https://github.com/curl/curl/issues/8818

Reported-by: KotlinIsland@users.noreply.github.com

Closes #xxxx

### DIFF
--- a/docs/FAQ
+++ b/docs/FAQ
@@ -588,7 +588,13 @@ FAQ
   Exactly what kind of quotes and how to do this is entirely up to the shell
   or command line interpreter that you are using. For most unix shells, you
   can more or less pick either single (') or double (") quotes. For
-  Windows/DOS prompts I believe you are forced to use double (") quotes.
+  Windows/DOS command prompts you must use double (") quotes, and if you have
+  inner double quotes you can escape them with a backslash.
+
+  For Windows powershell the arguments are not always passed on as expected
+  because curl is not a powershell script. You may or may not be able to use
+  single quotes. To escape inner double quotes seems to require a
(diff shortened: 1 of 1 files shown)


docs/FAQ
```diff
@@ -588,7 +588,13 @@ FAQ
...
   or command line interpreter that you are using. For most unix shells, you
   can more or less pick either single (') or double (") quotes. For
-  Windows/DOS prompts I believe you are forced to use double (") quotes.
+  Windows/DOS command prompts you must use double (") quotes, and if you have
+  inner double quotes you can escape them with a backslash.
+
+  For Windows powershell the arguments are not always passed on as expected
+  because curl is not a powershell script. You may or may not be able to use
+  single quotes. To escape inner double quotes seems to require a
+  backslash-backtick escape sequence and the outer quotes as double quotes.
```
danielgustafsson, line 597: Does this warrant a small example, or is that overkill?

jay: I think it's overkill. It's already a stretch to explain that inner double quotes need escaping.

PR #8827, opened 2022-05-09
Change the HTTP writedelay unit to milliseconds

fabiankeil:
This allows to use write delays for large responses without
resulting in the test taking an unreasonable amount of time.
    
In many cases delaying writes by a whole second or more isn't
necessary for the desired effect.


bagder (approved):

bagder:
Thanks!

PR #8831, opened 2022-05-10
x509asn1: mark msnprintf return as unchecked

danielgustafsson:
We have lots of unchecked msnprintf calls, and this particular msnprintf
call isn't more interesting than the others, but this one yields a Coverity
warning to implicitly silence it.


bagder (approved):
Making the ignore explicit is generally good I think.

Issue #8816, opened 2022-05-08
Segmentation fault when using --proxy-negotiate --proxy-user : 


cacheguard:
### I did this

Hi,
The context: password for the user foo on my kdc is foo. I'm using squid as a remote proxy.

On the client:

echo foo | kinit foo
klist # The output is OK

curl --proxy-negotiate --proxy-user : --proxy 192.168.1.254:3128 --url http://example.com

### I expected the following
The content of http://example.com as the output

BUT -- Got a Segmentation fault

### curl/libcurl version

[curl -V output] curl 7.81.0 (i686-pc-linux-gnu) libcurl/7.81.0 OpenSSL/1.1.1m zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libssh2/1.10.0 OpenLDAP/2.6.1
Release-Date: 2022-01-05
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTPS-proxy IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL TLS-SRP UnixSockets zstd

### operating system

Linux built from scratched (based on LFS 11.1-rc1)

Linux kernel 4.19.0-20-amd64

bagder:
I don't have any such proxy setup to reproduce with. Any chance you can try with the latest curl and see if it still happens? And if it does, can you post a stack trace from the crash here? (ideally of course with curl built with debug enabled so that we get more symbols shown)

jay:
> #0 0x00007ffff790fdf8 in gss_display_status () from /usr/lib/libgss.so.3
> #1 0x00007ffff7ed7740 in display_gss_error (status=0, type=2, buf=0x7fffffffde60 "An unsupported mechanism was requested. ", len=40) at curl_gssapi.c:95

It looks like it crashes in the second iteration of the loop below. When there are multiple errors in the queue gss_display_status requires the parameters remain the same to retrieve the additional errors. AFAICS that is done correctly so I don't know why the crash.

https://github.com/curl/curl/blob/1669b17d3a1a1fd824308544ca0ec02a2a4f50ea/lib/curl_gssapi.c#L86-L110

danielgustafsson:
[...]

PR #8832, opened 2022-05-11
gssapi: improve handling of errors from gss_display_status

danielgustafsson:
In case gss_display_status() returns an error, avoid trying to add
it to the buffer as the message may well be a NULL pointer.  While
there, also initialize the status_string variable correctly.

Originally this fix comes from a discussion in issue #8816.

Closes: #xxxx
Reviewed-by: xxxx

### DIFF
--- a/lib/curl_gssapi.c
+++ b/lib/curl_gssapi.c
@@ -89,7 +89,7 @@ static size_t display_gss_error(OM_uint32 status, int type,
   OM_uint32 maj_stat;
   OM_uint32 min_stat;
   OM_uint32 msg_ctx = 0;
-  gss_buffer_desc status_string;
+  gss_buffer_desc status_string = GSS_C_EMPTY_BUFFER;
 
   do {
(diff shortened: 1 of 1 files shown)


jay (approved):
status_string is also missing an initialization in socks_gssapi, you could amend for that or if not i will add it in a separate commit

~~~diff
diff --git a/lib/socks_gssapi.c b/lib/socks_gssapi.c
index 8ef2f8f..0895d94 100644
--- a/lib/socks_gssapi.c
+++ b/lib/socks_gssapi.c
@@ -51,7 +51,7 @@ static int check_gss_err(struct Curl_easy *data,
   if(GSS_ERROR(major_status)) {
     OM_uint32 maj_stat, min_stat;
     OM_uint32 msg_ctx = 0;
-    gss_buffer_desc status_string;
+    gss_buffer_desc status_string = GSS_C_EMPTY_BUFFER;
     char buf[1024];
     size_t len;
 
~~~

PR #8840, opened 2022-05-13
fix secure cookie domain overlay

bagder:
@piru: I assume you don't mind me using your patch here


piru:
> @piru: I assume you don't mind me using your patch here

I don't mind at all. Fix the typo though: s/exsiting/existing/g

tests/data/test1561
```diff
@@ -35,12 +35,12 @@ nomnom
...
-Set-Cookie: super=secret; domain=example.com; path=/%TESTNUMBER/; httponly;
-Set-Cookie: super=secret; domain=example.com; path=/15; httponly;
+Set-Cookie: super=plain; domain=example.com; path=/%TESTNUMBER; httponly;
+Set-Cookie: super=plain; domain=example.com; path=/%TESTNUMBER/; httponly;
+Set-Cookie: super=plain; domain=example.com; path=/15; httponly;
 Set-Cookie: public=yes; domain=example.com; path=/foo;
-Set-Cookie: supersuper=secret; domain=example.com; path=/%TESTNUMBER/login/en;
-Set-Cookie: supersuper=secret; domain=example.com; path=/%TESTNUMBER/login;
+Set-Cookie: supersuper=plain; domain=example.com; path=/%TESTNUMBER/login/en;
+Set-Cookie: supersuper=plain; domain=example.com; path=/%TESTNUMBER/login;
```
danielgustafsson, line 43: Why this change?

lib/cookie.c
```diff
@@ -1059,11 +1060,39 @@ Curl_cookie_add(struct Curl_easy *data,
...
+        }
+      }
+    }
+    lastc = clist;
+    clist = clist->next;
+  }
+
+  clist = c->cookies[myhash];
+  replace_old = FALSE;
+  while(clist) {
```
danielgustafsson, line 1078: Traversing clist twice adds a computational cost, might be unavoidable but we should probably at the least add a comment on why this is done.

tests/data/test1561
bagder: For making the test easier to read. I decided to change the contents *set* in the two responses (HTTPS vs HTTP) just to make it more obvious in the final output which of the same name cookies that actually survive.

lib/cookie.c
bagder: I'll see if I can work it down to a single loop

bagder: Just needed a little massaging. Thanks for highlighting this.

PR #8847, opened 2022-05-13
make sure --fail-with-body works with --retry

bagder:


jakub-bochenski:
Thanks a lot! Are the some preview build available so that I can check this out or do I have to wait for next release?

PR #8850, opened 2022-05-14
Curl_parsenetrc: don't access local pwbuf outside of scope

piru:
Accessing local variables outside of the scope is forbidden and depending on the compiler can result in the value being overwritten. Fixed by moving the `pwbuf` to be in scope.


bagder (approved):

bagder:
Thanks!

PR #8854, opened 2022-05-15
sectransp: check for a function defined when __BLOCKS__ is undefined

### LINKED ISSUE #8846: build error on macos+gcc
There is a long history of this issue.
https://github.com/curl/curl/issues?q=is%3Aissue+SecTrustEvaluateAsync+is%3Aclosed

curl 7.83.1
macos sdk12.3, gcc11 from homebrew

Error:
```
[org.sw.demo.badger.curl.libcurl-7.83.1]/lib/vtls/sectransp.c
/Users/egor/.sw/storage/pkg/3e/f5/47f6/src/sdir/lib/vtls/sectransp.c: In function 'collect_server_cert':
/Users/egor/.sw/storage/pkg/3e/f5/47f6/src/sdir/lib/vtls/sectransp.c:2967:6: error: 'SecTrustEvaluateAsync' undeclared (first use in this function); did you mean 'SecTrustEvaluate'?
 2967 |   if(SecTrustEvaluateAsync) {
      |      ^~~~~~~~~~~~~~~~~~~~~
      |      SecTrustEvaluate
```

Changing it to [SecTrustEvaluateWithError](https://developer.apple.com/documentation/security/2980705-sectrustevaluatewitherror) (not deprecated) fixes the issue.
You can compare OS version support with [SecTrustEvaluateAsync](https://developer.apple.com/documentation/security/1400632-sectrustevaluateasync).
[SecTrustEvaluate](https://developer.apple.com/documentation/security/1394363-sectrustevaluate) works as well, but it's also deprecated.

nickzman:
SecTrustEvaluateAsync() is defined in the macOS 10.7 SDK, but it
requires Grand Central Dispatch to be supported by the compiler, and
some third-party macOS compilers do not support Grand Central Dispatch.
SecTrustCopyPublicKey() is not present in macOS 10.6, so this shouldn't
adversely affect anything.

Fixes #8846
Reported-by: Egor Pugin

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -2964,7 +2964,7 @@ collect_server_cert(struct Curl_easy *data,
      private API and doesn't work as expected. So we have to look for
      a different symbol to make sure this code is only executed under
      Lion or later. */
-  if(SecTrustEvaluateAsync) {
+  if(SecTrustCopyPublicKey) {
 #pragma unused(server_certs)
     err = SSLCopyPeerTrust(backend->ssl_ctx, &trust);
     /* For some reason, SSLCopyPeerTrust() can return noErr and yet return


bagder:
@egorpugin, does this fix it for you then?

egorpugin:
Yes, it works.
But note that this function is deprecated as well.
https://developer.apple.com/documentation/security/1396135-sectrustcopypublickey

In some future SDK updates it may be removed.

PR #8858, opened 2022-05-16
feat: add support for yaml file extension

IbrahimFathy19:
add new content type for yaml files instead of default `HTTPPOST_CONTENTTYPE_DEFAULT=application/octet-stream`

credits goes to [How does `curl` on the command line determine the MIME type of a file being uploaded?](https://unix.stackexchange.com/questions/396385/how-does-curl-on-the-command-line-determine-the-mime-type-of-a-file-being-uplo)

### DIFF
--- a/lib/mime.c
+++ b/lib/mime.c
@@ -1729,7 +1729,9 @@ const char *Curl_mime_contenttype(const char *filename)
     {".htm",  "text/html"},
     {".html", "text/html"},
     {".pdf",  "application/pdf"},
-    {".xml",  "application/xml"}
+    {".xml",  "application/xml"},
+    {".yaml",  "application/x-yaml"},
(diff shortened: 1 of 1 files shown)


danielgustafsson (commented):
Seems reasonable

IbrahimFathy19:
I think the debate will be using `application/x-yaml` or `text/yaml`

bagder:
The content-type is possible to set by the user, I'm not convinced it is worth opening the Pandora's box here to bike-shed about the types and have a never-ending flow of more proposal for extensions to feature by default.

bagder:
If there is no strong reason provided for why this should be added to the default list, I think we should rather go with "it can be set by the user" and void adding/changing any of the detected extensions.

bagder:
it can be set by the user

PR #8862, opened 2022-05-16
bindlocal: Don't use a random port if port number would wrap

piru:
Earlier if `CURLOPT_LOCALPORT` + `CURLOPT_LOCALPORTRANGE` would go past port 65535 the code would fall back to random port rather than giving up.

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -470,8 +470,10 @@ static CURLcode bindlocal(struct Curl_easy *data,
     }
 
     if(--portnum > 0) {
-      infof(data, "Bind to local port %hu failed, trying next", port);
       port++; /* try next port */
+      if(port == 0)
+        break;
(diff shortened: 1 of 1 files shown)


bagder (approved):

lib/connect.c
```diff
@@ -472,6 +472,8 @@ static CURLcode bindlocal(struct Curl_easy *data,
     if(--portnum > 0) {
       infof(data, "Bind to local port %hu failed, trying next", port);
       port++; /* try next port */
+      if(port == 0)
+        break;
```
bagder, line 476: Maybe it would be nicer to have the infof() done after the check? "trying next" might turn out wrong otherwise.

piru: Good point there, fixed that.

bagder:
Thanks!

PR #8864, opened 2022-05-17
wolfSSL: explicitly use compatibility layer

elms:
This change removes adding an include `$prefix/wolfssl` or similar to allow for openssl include aliasing. Include paths of
`wolfssl/openssl/` are used to explicitly use wolfSSL includes. This fixes cmake builds(see https://github.com/microsoft/vcpkg/pull/24348) as well as avoiding potentially using openSSL headers since include path order is not guaranteed.


bagder (approved):
Nice cleanup!

bagder:
Thanks!

PR #8873, opened 2022-05-17
GHA: align all install, configure and build steps again

mback2k:
This will be the first attempt to streamline the steps between workflows. Next up afterwards: merging workflow files again.

### DIFF
--- a/.github/workflows/bearssl.yml
+++ b/.github/workflows/bearssl.yml
@@ -21,17 +21,18 @@ jobs:
       matrix:
         build:
         - name: bearssl
-          install:
-          configure: --with-bearssl=$HOME/bear --enable-debug --enable-werror
-        - name: clang-bearssl
-          install:
-          configure: CC=clang --with-bearssl=$HOME/bear --enable-debug --enable-werror
+          install: libpsl-dev libbrotli-dev libzstd-dev zlib1g-dev
+          configure: LDFLAGS="-Wl,-rpath,$HOME/bear/lib" --with-bearssl=$HOME/bear --enable-debug
+        - name: bearssl-clang
+          install: libpsl-dev libbrotli-dev libzstd-dev zlib1g-dev clang
+          configure: CC=clang LDFLAGS="-Wl,-rpath,$HOME/bear/lib" --with-bearssl=$HOME/bear --enable-debug
 
     steps:
     - run: |
-        sudo apt-get install libtool autoconf automake pkg-config clang stunnel4 libpsl-dev libbrotli-dev libzstd-dev zlib1g-dev
+        sudo apt-get update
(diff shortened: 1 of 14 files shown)


bagder:
> streamline the steps between workflows

What does that mean?

pheiduck:
I think we should update the workflow names of the Linux CI‘s as well, like shown on PRs:
For example:
`Linux / bearssl`
This will make the GH - Actions more organized.

mback2k:
> > streamline the steps between workflows
> 
> 
> 
> What does that mean?

The previous repetition of workflow files created during migration away from Zuul is actually not needed. You opted to modify the actual install step instead of modifying the install variable in the build matrix for example.

My plan is to 1) make workflows more similar again, e.g. by splitting autoreconf, configure and make in the same way over all workflows and having the same set of base arguments in steps while using variables to differentiate builds again.

Then 2) merge those workflow files again which don't need custom build steps, e.g. for a 3rd party TLS library. 

Would this be okay for you?

mback2k:
> I think we should update the workflow names of the Linux CI‘s as well, like shown on PRs:
> 
> For example:
> 
> `Linux / bearssl`
> 
> This will make the GH - Actions more organized. 

If you mean renaming the workflow files: yes, I plan to do that during step 2) mentioned above.

bagder:
I suppose I just don't see why we need them so similar, but sure if you want to work on that go ahead.

I think a big (the biggest maybe outside the flakiness) space for improvement is still the CI job names that appear here on GitHub for pull requests. I think maybe prefixing 30 jobs with `Linux / ` is not too useful. I suppose the azure builds are the worst in that aspect as they have very long, repetitive names with lots of useless cruft in them. Example: **curl.curl (windows windows msys1_mingw64_debug_schannel)**

mback2k:
> I suppose I just don't see why we need them so similar, but sure if you want to work on that go ahead.

Simplification and don’t repeat yourself (DRY). Especially to ease future maintenance of the workflows.

> I think a big (the biggest maybe outside the flakiness) space for improvement is still the CI job names that appear here on GitHub for pull requests. I think maybe prefixing 30 jobs with `Linux / ` is not too useful.

I will try to optimize the naming as part of this work then.

> I suppose the azure builds are the worst in that aspect as they have very long, repetitive names with lots of useless cruft in them. Example: **curl.curl (windows windows msys1_mingw64_debug_schannel)**

I would also like to get rid of the `curl.curl (windows windows` prefix, but unfortunately Azure CI is not that flexible.

pheiduck:
make and make example: is given this error on some CIs: collect2: `error: ld returned 1 exit status`
openssl3-clang: clang: `error: linker command failed with exit code 1 (use -v to see invocation)`

mback2k:
I know, this PR is still work in progress...

PR #8875, opened 2022-05-18
strcase: some optimisations.

gevaerts:
Lookup tables for toupper() and tolower() make Curl_strcasecompare()
about 1.5 times faster. Reorganising Curl_strcasecompare() to fully
exit early then also allows simplifying the check at the end, for
another 15%. In total, the changes make Curl_strcasecompare() around
1.6 to 1.7 times faster.

Note that these optimisation assume ASCII. The original Curl_raw_toupper()
and raw_tolower() look like they already made that assumption, but confirmation would be good.

For short strings (I tested 4 to 8 characters) this is competitive with glibc, with this code being slightly faster on some CPUs and some lengths, and glibc being slightly faster on others. For longer strings glibc starts getting significantly faster (but the locale issue of course remains)


bagder:
Confirmed: we only (need to) support ASCII with these.

bagder (approved):

bagder:
Thanks!

Issue #8877, opened 2022-05-19
curl exceeds max-time when retries enabled


Nuru:
### I did this

```
$ time curl --max-time 10 --retry 5 --retry-connrefused https://BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out
Warning: Problem : timeout. Will retry in 1 seconds. 5 retries left.
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out
Warning: Problem : timeout. Will retry in 2 seconds. 4 retries left.
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out
Warning: Problem : timeout. Will retry in 4 seconds. 3 retries left.
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out
Warning: Problem : timeout. Will retry in 8 seconds. 2 retries left.
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out
Warning: Problem : timeout. Will retry in 16 seconds. 1 retries left.
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out

real	1m16.354s
user	0m0.012s
sys	0m0.010s
```

### I expected the following

I expected `curl` to quit after 10 seconds. Note that not only did it not quit after 10 seconds, it did not even quit after the retry delay backed off to something greater than 10 seconds. 

### curl/libcurl version

Current version for Debian 11.3:

curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3
Release-Date: 2020-12-09
[...]

Issue #8876, opened 2022-05-19
Connection times out at 75% of --connnection-timeout


Nuru:
### I did this

```
time curl -v --connect-timeout 20  https://BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com
*   Trying 10.10.78.234:443...
* After 9997ms connect time, move on!
* connect to 10.10.78.234 port 443 failed: Connection timed out
*   Trying 10.10.81.137:443...
* After 4997ms connect time, move on!
* connect to 10.10.81.137 port 443 failed: Connection timed out
* Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out
* Closing connection 0
curl: (28) Failed to connect to BD71190E2FE991E56354DF0200B85B61.gr7.us-east-1.eks.amazonaws.com port 443: Connection timed out

real	0m14.998s
user	0m0.005s
sys	0m0.008s
```

### I expected the following

I expected `curl` to wait `--connection-timeout` seconds for the connection to succeed. Instead it waits half that time, then retries the connection waiting 1/4 of the `connection_timeout`, then quits.

### curl/libcurl version

Current version for Debian 11.3:

curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3
Release-Date: 2020-12-09
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli GSS-API HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets

### operating system

Debian 11.3
Linux 8024f36bbeba 5.10.104-linuxkit #1 SMP Thu Mar 17 17:08:06 UTC 2022 x86_64 GNU/Linux

bagder:
This is done on purpose, so that curl doesn't "waste" the entire timeout time on the first address when it has multiple ones to try within the given time-out. Then it has an algorithm to rather continue to the next to try that as well and so on.

[...]

PR #8879, opened 2022-05-19
max-time.d: clarify max-time sets max transfer time

jay:
Prior to this change the doc said --max-time set the maximum time of the
'whole operation' which is not accurate. The option maps to
CURLOPT_TIMEOUT_MS which sets maximum transfer time.

For example, the maximum time on a transfer is reset if the transfer is
retried (--retry).

Reported-by: Nuru@users.noreply.github.com

Fixes https://github.com/curl/curl/issues/8877
Closes #xxxx

### DIFF
--- a/docs/cmdline-opts/max-time.d
+++ b/docs/cmdline-opts/max-time.d
@@ -2,16 +2,20 @@ Long: max-time
 Short: m
 Arg: <fractional seconds>
 Help: Maximum time allowed for transfer
-See-also: connect-timeout
+See-also: connect-timeout retry-max-time
 Category: connection
 Example: --max-time 10 $URL
 Example: --max-time 2.92 $URL
 Added: 4.0
 ---
-Maximum time in seconds that you allow the whole operation to take.  This is
+Maximum time in seconds that you allow each transfer to take.  This is
 useful for preventing your batch jobs from hanging for hours due to slow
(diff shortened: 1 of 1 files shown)


docs/cmdline-opts/max-time.d
```diff
@@ -8,10 +8,14 @@ Example: --max-time 10 $URL
...
 ---
-Maximum time in seconds that you allow the whole operation to take.  This is
+Maximum time in seconds that you allow each transfer to take.  This is
 useful for preventing your batch jobs from hanging for hours due to slow
 networks or links going down.  Since 7.32.0, this option accepts decimal
 values, but the actual timeout will decrease in accuracy as the specified
 timeout increases in decimal precision.
 
+If you enable retrying the transfer (--retry) then the maximum time counter is
+reset each time the transfer is retried. You can use --retry-max-time to limit
```
danielgustafsson, line 18: How about adding `--retry-max-time` as a `See-also` parameter?

jay: Good idea, done

danielgustafsson (approved):
:+1:

Nuru:
Please also note that `--max-time` also sets `--connect-timeout`, which triggers some other odd behavior when the target host has multiple addresses. (See #8876)

bagder:
Thanks!

PR #8888, opened 2022-05-20
CURLINFO_CAPATH/CAINFO: get the default CA paths from libcurl

bagder:


docs/libcurl/opts/CURLINFO_CAINFO.3
```diff
@@ -0,0 +1,60 @@
...
+
+The \fBpath\fP pointer will be NULL if there is no default path.
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+CURL *curl = curl_easy_init();
+if(curl) {
+  CURLcode res;
+  curl_easy_getinfo(curl, CURLINFO_CAINFO, &cainfo);
```
PerMalmberg, line 47: Maybe declare `cainfo` for a complete example?

docs/libcurl/opts/CURLINFO_CAPATH.3
```diff
@@ -0,0 +1,60 @@
...
+
+The \fBpath\fP pointer will be NULL if there is no default path.
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+CURL *curl = curl_easy_init();
+if(curl) {
+  CURLcode res;
+  curl_easy_getinfo(curl, CURLINFO_CAPATH, &cainfo);
```
PerMalmberg, line 47: Same as above; declare `cainfo` for a complete example?

bagder: Good catches. Will amend!

dfandrich:
Out of curiousity, how does this interact with the dynamic TLS library feature? It looks like it doesn't and just hard codes one path, but, presumably, the path returned by this feature isn't necessarily the one that's being used by the TLS library being selected at run-time. In fact, isn't this path going to be completely ignored on some TLS libraries? These should at least be documented if they can't be fixed.

bagder:
Good point. I'll convert your comment to an issue to make sure we don't forget it!

PR #8900, opened 2022-05-24
headers api: remove EXPERIMENTAL tag

bagder:


bagder:
A complete lack of feedback, but I feel fairly confident this is good enough.

xquery:
FWIW I have played with it a bit - at least the 'happy paths' seem robust enough

mback2k:
Somewhat related note: test 1940 is consistently broken on the Cirrus CI Windows builds due to invalid line endings. Maybe something is written with the wrong mode there?

Generally speaking: Windows CI stability is now at a point there it makes sense to always review failures instead of ignoring them without looking at all and thinking the flakiness is the cause. 😉

bagder:
I've seen it, but the test runs fine on appveyor builds. The line endings *should* be fine as the section is marked as text. I don't know why it happens.

PR #8906, opened 2022-05-24
CI/azure: shorten names of Windows CI jobs

mback2k:
Removes redundant information and strips "_debug" from job names.

Suggested-by: Daniel Stenberg

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -193,7 +193,7 @@ stages:
   variables:
     agent.preferPowerShellOnContainers: true
   jobs:
-  - job: windows
+  - job: msys
     # define defaults to make sure variables are always expanded/replaced
     variables:
       container_img: ''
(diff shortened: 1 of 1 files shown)


bagder (approved):
:+1: I wish there was a way to improve that "curl.curl" part...

mback2k:
hm, it seems like this is the Azure CI pipeline name with the `.` having been replaced with a `_`. I don't know how rename affects currently running jobs, but I could give it a try and shorten it to just `curl`:
image

bagder:
> I could give it a try and shorten it to just `curl`:

Or just "c", we know the project is curl... Or maybe "azure" ?

mback2k:
> > I could give it a try and shorten it to just `curl`:
> 
> Or just "c", we know the project is curl... Or maybe "azure" ?

hm, what if you ever run more than the curl pipeline in Azure? e.g. if we ever move over my docker build env pipeline?

bagder:
> hm, what if you ever run more than the curl pipeline in Azure? e.g. if we ever move over my docker build env pipeline?

I don't understand the question. A) We could change the name then and B) surely such new jobs can have other names?

mback2k:
> > hm, what if you ever run more than the curl pipeline in Azure? e.g. if we ever move over my docker build env pipeline?
> 
> I don't understand the question. A) We could change the name then and B) surely such new jobs can have other names?

Ok, that is right. So which one would you prefer? I will try to change it then during a time there no Azure job is running.

bagder:
Maybe start with just plain "curl".

PR #8910, opened 2022-05-24
docs/CONTRIBUTE.md: document the 'needs-votes' concept

bagder:
A pull request sent to the project might get labeled `needs-votes` by a
project maintainer. This label means that in addition to meeting all
other checks and qualifications this pull request must also receive at
least **5** thumbs-up reactions on GitHub to be considered for merging.

Discussed: https://curl.se/mail/lib-2022-05/0067.html

### DIFF
--- a/docs/CONTRIBUTE.md
+++ b/docs/CONTRIBUTE.md
@@ -169,6 +169,11 @@ actual content changed also allows you to retrigger the tests for that commit.
 When you adjust your pull requests after review, consider squashing the
 commits so that we can review the full updated version more easily.
 
+A pull request sent to the project might get labeled `needs-votes` by a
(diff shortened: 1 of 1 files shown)


jay:
I'd rather we not box ourselves in and I think the informal process we have now works well. 5 users could think anything is a good idea. I would leave any voting to collaborators, and only when it's really needed. If you really want to tag them I think needs-more-feedback or something, which we already say. I will follow up in the list thread.

bagder:
Tagging doesn't work as I/we don't know who to tag. I think 4 or 5 of the pending PRs we have now are such pull requests. Virtually no one but the author and me have interacted in them and I don't know if these are changes we actually want.

PR #8912, opened 2022-05-25
Digest auth fixes

Karlson2k:
RFC 7616 (and 2617) requires values to be "unquoted" before used for digest calculations. The same mentioned in the code comments.
`realm="DOMAIN\\host"` and `realm=DOMAN\\host` are different realms, but they are processed as identical values at the moment, which is not compliant with RFC. See [RFC7616](https://datatracker.ietf.org/doc/html/rfc7616#page-8).
The only place where unquoting can be done correctly is header parsing function. 
This PR adds unquoting (de-escaping) of all values during header parsing and quoting (escaping) of the values during header forming. This approach should be most straightforward and easy to read/maintain as all values are processed in the same way as required by RFC.

The second commit adds detection of more erroneous situation during header parsing.
The third commit adds fix for NULL dereference discovered by Fuzzer. Server header may not define `realm`, so code must tolerate NULL value for `digest->realm`.

Can be squashed into single commit.

### DIFF
--- a/lib/vauth/digest.c
+++ b/lib/vauth/digest.c
@@ -695,7 +695,7 @@ static CURLcode auth_create_digest_http_message(
   }
 
   if(digest->userhash) {
-    hashthis = aprintf("%s:%s", userp, digest->realm);
+    hashthis = aprintf("%s:%s", userp, digest->realm ? digest->realm : "");
     if(!hashthis)
       return CURLE_OUT_OF_MEMORY;
 
@@ -715,7 +715,8 @@ static CURLcode auth_create_digest_http_message(
            unq(nonce-value) ":" unq(cnonce-value)
   */
 
-  hashthis = aprintf("%s:%s:%s", userp, digest->realm, passwdp);
+  hashthis = aprintf("%s:%s:%s", userp, digest->realm ? digest->realm : "",
+                     passwdp);
   if(!hashthis)
     return CURLE_OUT_OF_MEMORY;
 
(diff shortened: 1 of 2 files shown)


Karlson2k:
MQTT leak detected by Fuzzer seems to be unrelated: https://github.com/curl/curl/runs/6595703522?check_suite_focus=true
Test 1014 `Compare curl --version with curl-config --features` failure seems to be unrelated: https://object-storage.public.mtl1.vexxhost.net/swift/v1/bfd521072e894ebb99e66f72619daa8a/zuul-logs-curl_63e/8912/9d2ce6be57149d73f6ec0a95e83acea7c418c525/check/curl-cmake-ngtcp2/63eb2f4/job-output.txt
Similar failures in test 1940 on Cirrus CI builds seems to be unrelated (some CRLF -> LF conversions): https://cirrus-ci.com/task/6002205147791360?logs=test#L2822
Failed 1554 `HTTP with shared connection cache` seems to be unrelated as well (two additional Mutex lock/unlock): https://ci.appveyor.com/project/curlorg/curl/builds/43659712/job/d2aiea2f68i1dona

The PR looks to be ready.

Karlson2k:
The last patch also fixed wrongly hashed `(nil)` string when `realm` is not present in header from server.

Karlson2k:
Rebased

Karlson2k:
Rebased again.

bagder (approved):

bagder:
Thanks!

Karlson2k:
Thank you!

PR #8914, opened 2022-05-25
c-hyper: mark status line as status for Curl_client_write()

### LINKED ISSUE #8894: USE_HYPER and CURLE_BAD_FUNCTION_ARGUMENT
Building libcurl with `-DUSE_HYPER` always causes the annoying error:
`curl: (43) A libcurl function was given a bad argument` on any address.
I'm building with the most up-to date version of `Hyper.dll.lib` on Win-10.

Now, a simple `curl.exe -vL4 www.curl.se`, just gives this:
```
...
* STATE: DO => DID handle 0x10011228; line 2145 (connection #0)
* STATE: DID => PERFORMING handle 0x10011228; line 2264 (connection #0)
* HTTP 1.1 or later with persistent connection
* multi_done: status: 43 prem: 1 done: 0
* Curl_quic_done
* Closing connection 0
* Expire cleared (transfer 0x10011228)
curl: (43) A libcurl function was given a bad argument
```

So it would be nice if in verbose-mode (`-v`) libcurl could state at which point that is happening.
Perhaps only with `-DCURLDEBUG`. I was thinking of adding something like this to `memdebug.h`:
```c
static int inline curl_dbg_bad_function_argument (int verbose, const char *file, unsigned line)
{
  if (verbose)
     fprintf(stderr, "%s(%u): Throwing a 'CURLE_BAD_FUNCTION_ARGUMENT'\n", file, line);
  return (CURLE_BAD_FUNCTION_ARGUMENT);
}

#define curl_dbg_BAD_FUNCTION_ARGUMENT(verb)   curl_dbg_bad_function_argument (verb, __FILE__, __LINE__)
...
#ifndef curl_dbg_BAD_FUNCTION_ARGUMENT
#define curl_dbg_BAD_FUNCTION_ARGUMENT(verb) CURLE_BAD_FUNCTION_ARGUMENT
#endif
```

And using this every place where a `data->set.verbose` is present.

bagder:
To make sure the headers API can filter it out as not a regular header.

Reported-by: Gisle Vanem
Fixes #8894

### DIFF
--- a/lib/c-hyper.c
+++ b/lib/c-hyper.c
@@ -288,7 +288,7 @@ static CURLcode status_line(struct Curl_easy *data,
              len);
 
   if(!data->state.hconnect || !data->set.suppress_connect_headers) {
-    writetype = CLIENTWRITE_HEADER;
+    writetype = CLIENTWRITE_HEADER|CLIENTWRITE_STATUS;
     if(data->set.include_header)
       writetype |= CLIENTWRITE_BODY;
     result = Curl_client_write(data, writetype,
--- a/tests/data/DISABLED
+++ b/tests/data/DISABLED
@@ -49,10 +49,16 @@
 671
 672
 673
+1117
+1274
 1417
 1533
 1540
 1591
+1940
+1941
+1942
+1943
 %endif
 2043
 # Tests that are disabled here for rustls are SUPPOSED to work


bagder:
As a result of this fix, the HTTP server is properly working again in CI builds and that's the reason why the Hyper CI job now suddenly shows unrelated test fails. They were simply not detected before because the HTTP test server wouldn't run because curl couldn't verify it because HTTP with Hyper was broken with the headers API enabled.

PR #8918, opened 2022-05-25
cmake: enable curl.rc for all Windows targets

vszakats:
Before this patch, it was only enabled for MSVC. This syncs this
configuration with libcurl.rc, which was already included for all
Windows targets.


vszakats:
@Karlson2k This enables `curl.rc` for `curl.exe`, either when it is linked against a static libcurl or `libcurl.dll`.

vszakats:
However, `libcurl.rc` is at the moment added to both `libcurl.dll` and static libcurl library, of which the latter is indeed wrong. [PR](https://github.com/curl/curl/pull/8923) opened.

PR #8923, opened 2022-05-26
cmake: do not add `libcurl.rc` to the static libcurl library

vszakats:
Ref: https://github.com/curl/curl/pull/8918#issuecomment-1138263855


Karlson2k:
Thanks! Looks fine.

Issue #8920, opened 2022-05-25
test 1940 is consistently broken on the Cirrus CI Windows builds


bagder:
Somewhat related note: test 1940 is consistently broken on the Cirrus CI Windows builds due to invalid line endings. Maybe something is written with the wrong mode there?

Generally speaking: IMHO Windows CI stability is now at a point there it makes sense to always review failures instead of ignoring them without looking at all and thinking the flakiness is the cause. 😉

_Originally posted by @mback2k in https://github.com/curl/curl/pull/8900#issuecomment-1137770369_

The test runs fine on appveyor builds. The line endings should be fine as the section is marked as text. I don't know why it happens.

mback2k:
Thanks for the analysis so far @jay , but maybe a stupid question: why are we expecting headers to terminate with LF anyway? Aren't headers normally terminated with CRLF and this is how we do it in other tests usually? Maybe this is the reason the problem only surfaces this time.

bagder:
> Does it?

No I was wrong. I must have hallucinated. The question is then why it doesn't have CRLF there since it is marked as text...

mback2k:
> > Does it?
> 
> 
> 
> No I was wrong. I must have hallucinated. The question is then why it doesn't have CRLF there since it is marked as text...

It is not marked as text as shown in @jay's screenshot.

An another information: non of the AppVeyor builds using msys are actual native Win32 cross-builds. All of them compile to the msys-specific Linux emulation layer. Only the VS builds there and the Azure Windows builds are Win32 native by cross-compiling from msys using mingw.

bagder:
> It is not marked as text as shown in @jay's screenshot.

With **marked** I mean using that property in the test case. Like this on line 50:

https://github.com/curl/curl/blob/a77d14d22ae93aa15670522e453f78744649b8cc/tests/data/test1940#L50

I think the screenshot shows the stdout data isn't treated the way that the test case tells it to.

bagder:
> And another information: non of the AppVeyor builds using msys are actual native Win32 cross-builds

But why does that matter? I seem to recall that they need the text mode set nonetheless?

mback2k:
> > It is not marked as text as shown in @jay's screenshot.
> 
> 
> 
> With **marked** I mean using that property in the test case. Like this on line 50:
> 
> 
> 
> https://github.com/curl/curl/blob/a77d14d22ae93aa15670522e453f78744649b8cc/tests/data/test1940#L50
> 
> 
> 
[...]

PR #8925, opened 2022-05-26
DISABLE test 1940 on the Cirrus CI Windows builds

pheiduck:
temp fix for: https://github.com/curl/curl/issues/8920


tests/data/DISABLED
```diff
@@ -55,7 +55,10 @@
 1533
 1540
 1591
+# Test that are disabled here for Cirrus CI Windows builds
+%if Windows
```
bagder, line 59: This disables the test on **all** windows builds, not just the Cirrus ones!

pheiduck: Any idea to do this only on Cirrus ones?

bagder: As discussed in #8920, it does seem to work on appveyor for example. I prefer we first try to actually fix the test case.

PR #8927, opened 2022-05-27
merged commit from Add SChannel TLS 1.3 support #8419 of wyattoday

gvollant:
This is a rebase of @wyattoday schannel PR
https://github.com/curl/curl/pull/7784
https://github.com/curl/curl/pull/8419

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -83,7 +83,12 @@
 #endif
 #endif
 
-#if defined(CryptStringToBinary) && defined(CRYPT_STRING_HEX)
+#ifndef BCRYPT_CHACHA20_POLY1305_ALGORITHM
+#define BCRYPT_CHACHA20_POLY1305_ALGORITHM L"CHACHA20_POLY1305"
+#endif
+
+#if defined(CryptStringToBinary) && defined(CRYPT_STRING_HEX) \
+    && !defined(DISABLE_SCHANNEL_CLIENT_CERT)
 #define HAS_CLIENT_CERT_PATH
 #endif
 
@@ -117,6 +122,10 @@
 
 #ifndef SP_PROT_TLS1_2_CLIENT
 #define SP_PROT_TLS1_2_CLIENT           0x00000800
+#endif
+
+#ifndef SP_PROT_TLS1_3_CLIENT
+#define SP_PROT_TLS1_3_CLIENT           0x00002000
 #endif
 
 #ifndef SCH_USE_STRONG_CRYPTO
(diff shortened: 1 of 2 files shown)


gvollant:
I just created this PR to test a rebased version of #8419

@wyattoday @jay I can close it if this is a bad idea

MarcelRaad:
My only concern is that I'm not sure if all those explicit checks for Windows >= 10.0.20348 are really needed, and they complicate the code quite a bit:
- If a user really wants to use TLS 1.3 on an older Windows version, why not let it fail (or succeed!) "naturally" rather than failing hard?
- Do we really have to manually disable all those ciphers? That's quite a lot of code.

Also, the `HAS_CLIENT_CERT_PATH` changes look unrelated, and there's a build error with Visual Studio 2008 visible on AppVeyor:
```
2>.\vtls\schannel.c(779) : error C2059: syntax error : '{'
2>.\vtls\schannel.c(790) : error C2059: syntax error : '{'
2>.\vtls\schannel.c(863) : error C2059: syntax error : '{'
2>.\vtls\schannel.c(873) : error C2059: syntax error : '{'
2>.\vtls\schannel.c(884) : error C2059: syntax error : '{'
2>.\vtls\schannel.c(903) : error C2059: syntax error : '{'
2>.\vtls\schannel.c(999) : error C2061: syntax error : identifier 'schannel_connect_step1'
2>.\vtls\schannel.c(999) : error C2059: syntax error : ';'
2>.\vtls\schannel.c(999) : error C2059: syntax error : 'type'
2>.\vtls\schannel.c(1782) : warning C4013: 'schannel_connect_step1' undefined; assuming extern returning int
```
and on classic MinGW visible on Azure Pipelines:
```
vtls/schannel.h:35:10: fatal error: winternl.h: No such file or directory
   35 | #include <winternl.h>
```

gvollant:
Commit https://github.com/curl/curl/pull/8927/commits/0dce3c9750380669fe662a7a7b79c09fa8b75dda try fix the VS2008 syntax problem.

It seem winternl.h is used to define UNICODE_STRING ,also defined in subauth.h

jay:
IMO this didn't need another PR. We have been discussing it in #8419, though it seems to have gone stale. I am unsure of the change or I would have added it already. We have discussed PRs like this, the reporter obviously put in a lot of work, we don't want to cast it aside but we're not definite on it.

gvollant:
> IMO this didn't need another PR. We have been discussing it in #8419, though it seems to have gone stale. I am unsure of the change or I would have added it already. We have discussed PRs like this, the reporter obviously put in a lot of work, we don't want to cast it aside but we're not definite on it.

I just made a rebase, resolve conflict and resolve syntax compatibility problem vith VS 2008, and replace one header (for another compatibility problem)

bagder:
It sounds as if @wyattoday intends to continue his work in #8419, does this mean that we should close #8927 again?

PR #8930, opened 2022-05-28
curl: clamp --parallel-max to MAX_PARALLEL instead of resetting to default value

JustAnotherArchivist:
Previously, `--parallel-max 300` would use 300 concurrent transfers, but `--parallel-max 301` would unexpectedly use only 50. This clamps higher values to the maximum.


jay:
Thanks

PR #8931, opened 2022-05-28
Re-enable --no-remote-name

verhovsky:
The docs for `--remote-name-all` say

> you must use "-o -" or --no-remote-name.

but with curl 7.83 on macOS I get

```
$ curl --no-remote-name
curl: option --no-remote-name: used '--no-' for option that isn't a boolean
curl: try 'curl --help' or 'curl --manual' for more information
```

So either the option should be put back (it was incorrectly marked as `ARG_NONE` instead of `ARG_BOOL`) or the docs should be updated.

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -312,7 +312,7 @@ static const struct LongShort aliases[]= {
   {"N",  "buffer",                   ARG_BOOL},
          /* 'buffer' listed as --no-buffer in the help */
   {"o",  "output",                   ARG_FILENAME},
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
I would be great to also write up a test case that uses this feature so that we can make sure we don't break it again in the future...

bagder:
Ok, I have written a test locally that verifies. I'll make a separate PR after I've merged this.

bagder:
Thanks!

Issue #8810, opened 2022-05-06
--aws-sigv4 is missing the amz-content-sha256 header?


vlovich:
R2 from Cloudflare requires x-amz-content-sha256 to be present on every request. This is part of the SigV4 docs: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

> Note: The x-amz-content-sha256 header is required for all AWS Signature Version 4 requests. It provides a hash of the request payload. If there is no payload, you must provide the hash of an empty string.

### I did this
```
curl https://<account>.r2.cloudflarestorage.com/mybucket/myfile --user '<id>:<secret>' --aws-sigv4 "aws:amz:auto:s3"
```

When I use `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` as the content-sha256 then it seems to works (i.e. `x-amz-content-sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`).

However, this is incorrect for POST requests which fail with:
```
<Error><Code>XAmzContentSHA256Mismatch</Code><Message>The provided &apos;x-amz-content-sha256&apos; header does not match what was computed.</Message></Error>
```
since `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` is not the payload.

### I expected the following
A successful get and post.

### curl/libcurl version

```
curl 7.82.0 (x86_64-pc-linux-gnu) libcurl/7.82.0 OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.0) libssh2/1.10.0 nghttp2/1.47.0
Release-Date: 2022-03-05
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets zstd
```

### operating system

```
Linux vlovich 5.16.16-arch1-1 #1 SMP PREEMPT Mon, 21 Mar 2022 22:59:40 +0000 x86_64 GNU/Linux
```

jamshid:
[...]

PR #8935, opened 2022-05-29
aws_sigv4: Include and sign X-Amz-Content-Sha256.

AndrewJDR:
This commit adds the X-Amz-Content-Sha256 header to all requests
as defined by the AWS documentation:

https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

Note, some SigV4 server implementations appear to strictly enforce this requirement while
others don't, which is why curl worked prior to this commit for some implementations
and not for others.

See Github Issue #8810


AndrewJDR:
I tried my best with the test fixes. Looking at the other PRs, it seems some of the tests just don't work properly in general, but I'm not entirely sure.

lib/http_aws_sigv4.c
```diff
@@ -250,6 +251,17 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
   memcpy(date, timestamp, sizeof(date));
   date[sizeof(date) - 1] = 0;
 
+  if(data->set.postfieldsize < 0)
+    post_data_len = strlen(post_data);
+  else
+    post_data_len = (size_t)data->set.postfieldsize;
```
bagder, line 257: This typecast is wrong, it is a `curl_off_t`. I see now that you didn't add this but only moved the line, but the type should still be fixed.

outscale-mgo:
> This commit adds the X-Amz-Content-Sha256 header to all requests as defined by the AWS documentation:
> 
> https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
> 
> Note, some SigV4 server implementations appear to strictly enforce this requirement while others don't, which is why curl worked prior to this commit for some implementations and not for others.
> 
> See Github Issue #8810

That completely true, AWS S3 V4 signature require `X-Amz-Content-Sha256`, but most API(like EC2, ELB...) doesn't require it, see https://docs.aws.amazon.com/general/latest/gr/sigv4_elements.html.

So by forcing `X-Amz-Content-Sha256` you might broke every API that are not S3-storage like. (as I guess EC2 and other doesn't expect an hashed payload)

Note that `X-Amz-Content-Sha256` not been computed here, is a real problem, partially fix by this PR (https://github.com/curl/curl/pull/7966) as the PR add support for other header, but even so,
you still need to do the hash manually.

If `X-Amz-Content-Sha256` and payload hashing is push into this algorithm (which would be a good thing), it must be done in an optional way, that can be configure by the user.

lib/http_aws_sigv4.c
AndrewJDR: @bagder Should the def on line 95, `size_t post_data_len` also be changed to `curl_off_t post_data_len`?

AndrewJDR:
> So by forcing X-Amz-Content-Sha256 you might broke every API that are not S3-storage like. (as I guess EC2 and other doesn't expect an hashed payload)

Ah okay. I've only been working with S3-like services -- all of which require this header -- so far. So if the goal is to maintain support those non-S3 services, it's probably best if someone else pick up this PR that's in a better position to test against both S3-like and non-S3-like services.

> Note that X-Amz-Content-Sha256 not been computed here, is a real problem, partially fix by this PR (#7966) as the PR add support for other header, but even so, you still need to do the hash manually.

My understanding is that `X-Amz-Content-Sha256` is a payload hash (i.e. the request's post data, or a hash of an empty string if there's no request body), and that seems to be computed here on line 258 of `lib/http_aws_sigv4.c`:
```
258   if(Curl_sha256it(sha_hash, (const unsigned char *) post_data ...
```
This worked in my testing, but I only tested with the hash-of-an-empty-string case -- a GET request with no request body at all.

There's also the idea of supporting the `X-Amz-Content-Sha256: UNSIGNED-PAYLOAD` capability, which I didn't attempt in this PR.

cbodley:
apologies, i duplicated some of this work in https://github.com/curl/curl/pull/9588 before i saw this one

> There's also the idea of supporting the X-Amz-Content-Sha256: UNSIGNED-PAYLOAD capability, which I didn't attempt in this PR.

i did use `UNSIGNED-PAYLOAD` there, specifically for PUT requests with service=s3. that approach avoids the need for a real payload hashing implementation - maybe that can help unblock progress here?

> So by forcing X-Amz-Content-Sha256 you might broke every API that are not S3-storage like. (as I guess EC2 and other doesn't expect an hashed payload)

we could make `x-amz-content-sha256` conditional on service=s3, if the reviewers think that's worth doing. to me, it seems unlikely to break other services if we send it anyway - so long as it's actually a payload hash, and not one of s3's special values like 'UNSIGNED-PAYLOAD' and 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD'

cbodley:
@AndrewJDR @outscale-mgo kindly requesting your thoughts on the comment above

outscale-mgo:
> we could make x-amz-content-sha256 conditional on service=s3, if the reviewers think that's worth doing. to me, it seems unlikely to break other services if we send it anyway - so long as it's actually a payload hash, and not one of s3's special values like 'UNSIGNED-PAYLOAD' and 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD'

I mostly like #9588 implementation (though it need to be rebase now that #7966 have been merged), but I'm not a big fan of checking `s3` in the URL, because there are s3-like services that need X-Amz-Content-Sha256. (pretty sure Google have one for instance, we have one at outscale too). 
The best solution would be to let the user choose if per want to compute `x-amz-content-sha256`.
So my guess is that we should add an option like `--aws-content-sha256`, or something similar.
Maybe there is other way, like checking if `x-amz-content-sha256` is present but doesn't contain a sha256 nor s3's special values then we compute it for the user, but this solution seems ratter unconventional.

cbodley:
hi @outscale-mgo, thanks for taking a look!

> Maybe there is other way, like checking if x-amz-content-sha256 is present but doesn't contain a sha256 nor s3's special values then we compute it for the user, but this solution seems ratter unconventional.

taking this idea in a slightly different direction..

from my perspective, the two main requirements are:
* s3 requests provide a `x-amz-content-sha256` header, even if it's value is UNSIGNED-PAYLOAD
* the value of that header matches the hash we write in our `CanonicalRequest` when generating the signature

given the hard requirement for s3, i would prefer that we generate valid requests without requiring additional user input - so if a request with service=s3 doesn't have that header already, then we add `x-amz-content-sha256: UNSIGNED-PAYLOAD` and sign the request accordingly

if the request *does* already have this header, then we should use its value in `CanonicalRequest` and include its name in `SignedHeaders`. this approach allows users to provide real hashes when they can, including the google/outscale cases that require them

> The best solution would be to let the user choose if per want to compute x-amz-content-sha256.

in my opinion, trying to generate real payload hashes here isn't going to work in general. we won't necessarily have the entire payload in memory at this point, and trying to read in the payload twice to generate this hash would be problematic for libcurl users that stream in the request body

i was hoping to sidestep this issue entirely for s3 through the use of `UNSIGNED-PAYLOAD`. however, the approach above could still be used with `x-amz-content-sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD`, provided the user includes chunk signatures as part of the request body. that's how s3, at least, would handle the streaming payloads

what do you think?

cbodley:
to summarize the proposed algorithm:

1. check for x-amz-content-sha256 header
2. if service=s3 and no header existed, add x-amz-content-sha256: UNSIGNED-PAYLOAD
3. for CanonicalRequest, use value from header or calculate the hash from postfields
4. for SignedHeaders, add x-amz-content-sha256 if present

outscale-mgo:
> hi @outscale-mgo, thanks for taking a look!
> 
> > Maybe there is other way, like checking if x-amz-content-sha256 is present but doesn't contain a sha256 nor s3's special values then we compute it for the user, but this solution seems ratter unconventional.
> 
> taking this idea in a slightly different direction..
> 
> from my perspective, the two main requirements are:
> 
>     * s3 requests provide a `x-amz-content-sha256` header, even if it's value is UNSIGNED-PAYLOAD
> 
>     * the value of that header matches the hash we write in our `CanonicalRequest` when generating the signature
> 
> 
> given the hard requirement for s3, i would prefer that we generate valid requests without requiring additional user input - so if a request with service=s3 doesn't have that header already, then we add `x-amz-content-sha256: UNSIGNED-PAYLOAD` and sign the request accordingly
> 

Using the same logic of "s3 URL require x-amz-content-sha256", We could have done a "every aws URL require sigv4 signature", and not having a parameter for that.

But it would not have work with every other provider that would require the sigv4.
And worst than that, if someday amazon change the algorithm require for s3, all curl version having hardcoded `s3` detection in they parameter will fail.

> if the request _does_ already have this header, then we should use its value in `CanonicalRequest` and include its name in `SignedHeaders`. this approach allows users to provide real hashes when they can, including the google/outscale cases that require them
> 

This is already done by #7966 merge, but the hash still need to be done manually, and it's as you said "the value of that header matches the hash we write in our `CanonicalRequest` when generating the signature", so let's the user have the ability to use that, and not only for s3.

> > The best solution would be to let the user choose if per want to compute x-amz-content-sha256.
> 
> in my opinion, trying to generate real payload hashes here isn't going to work in general. we won't necessarily have the entire payload in memory at this point, and trying to read in the payload twice to generate this hash would be problematic for libcurl users that stream in the request body
> 
> i was hoping to sidestep this issue entirely for s3 through the use of `UNSIGNED-PAYLOAD`. however, the approach above could still be used with `x-amz-content-sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD`, provided the user includes chunk signatures as part of the request body. that's how s3, at least, would handle the streaming payloads
> 
> what do you think?

By generating "real payload hash", I mean use what we've already have, like you've done in your PR, and push the header when requested by user.
The part I disagree with you, are
- to force it for s3. (even if s3 require it, maybe some user generate s3 hash themself, and having curl recompute it, could be annoying for the user)
- to not have the ability to push x-PROVIDER-content-sha256 for non s3 service:  it can be useful for google, yandex, outscale...  just type "s3 compatible API" on your search engine, there's a lot of answer,  each of those API might not work with the hard check on `service == "s3"` because `service` is also use to generate credential_scope, and you could have an s3-compatible API, with a credential scope that need something that is not "s3".

cbodley:
> By generating "real payload hash", I mean use what we've already have, like you've done in your PR, and push the header when requested by user.

got it, thanks

> The part I disagree with you, are
> 
> * to force it for s3. (even if s3 require it, maybe some user generate s3 hash themself, and having curl recompute it, could be annoying for the user)

the proposal was `if service=s3 and no header existed, add x-amz-content-sha256: UNSIGNED-PAYLOAD` - so if the user already provided something, we won't annoy them by modifying it

> * to not have the ability to push x-PROVIDER-content-sha256 for non s3 service:  it can be useful for google, yandex, outscale...  just type "s3 compatible API" on your search engine, there's a lot of answer,  each of those API might not work with the hard check on `service == "s3"` because `service` is also use to generate credential_scope, and you could have an s3-compatible API, with a credential scope that need something that is not "s3".

hmm this stretches my definition of a 's3-compatible' service, which is that you can point unmodified s3 clients at it

for clients and servers to interoperate, they need to agree on a well-defined protocol, and the credential scope is part of how clients specify which protocol they're using

because the special values like 'UNSIGNED-PAYLOAD' are only specified for amazon's s3, i didn't propose any changes for non-s3 services here. are you requesting that i find a way to add 'UNSIGNED-PAYLOAD' to s3-compatible services that don't identify as such?

cbodley:
> And worst than that, if someday amazon change the algorithm require for s3, all curl version having hardcoded `s3` detection in they parameter will fail.

such API changes would break all s3 clients, not just curl. luckily, amazon doesn't have a habit of breaking their APIs. recall for example that they added sigv4 instead of trying to change sigv2

bagder:
> At the end, the signature V4 is just a signature protocol, and could be implement by any Service providers. so curl code should be usable by any service implementing signature V4, that independently of the service name to which you communicate at the end.

I completely agree with this! :+1:

outscale-mgo:
I've just think at a generic solution that could work as generic and non generic, and shouldn't add too much work on your PR:
we could add an optional 5rd argument to `--aws-sigv4 <provider1[:provider2[:region[:service-name[:service-type]]]]>`
Then if `service-type` is not present you match the service type on `service-name`, otherwise you match it on `service-type`.

> if only this were true. :( there are cases where v4 signatures work differently for s3. for example, s3 allows slashes in its object names, so the process to generate the CanonicalURI (which goes into CanonicalRequest) is different than other services. from https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html:

Yes I know, V4 is weird, and different (yet very similar) if we use it for s3 or any other service.


On this PR https://github.com/curl/curl/pull/7966 (which try to do URI encode for V4 signature, and is still WIP), I've encounter the same problem of having 2 different kind of encoding, to solve this problem, the solution I went for was to add `--aws-sigv4-nb-encoding`, which could be 0,1,2. 0 for letting the user encode it's string manually, 1 for non S3 service, and 2 for s3. But I don't know is it's a good solution.

I think there is 2 different approach to having a V4 generic: 
- 1rst one is to follow this https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html by default, and have an option for every differences you could find (like the one in s3).
- 2nd one is to have a way to match which v4 we're trying to have (like with the `service-type` parameter).

Both have advantage and disadvantage:
1rst solution have the advantage of been more generic, and if someday there's a new v4-like signature that use half the stuff done by `s3`, and half the stuff done by `default v4`, then no change on curl would be require to support this new signature. The disadvantage is that we could end up with a lot of new options just for the v4.
The 2nd solution is obviously easier to use from an user perspective, but could require more change into curl code, every-time a new V4-like appear. (which hopefully doesn't happen that much)
Another advantage of 1rst solution, is that if we go with it, it would by pretty easy to have a curl wrapper that implement 2nd solution, and add the curl options for the user.

outscale-mgo:
> > 1. check for x-amz-content-sha256 header
> > 2. if service=s3 and no header existed, add x-amz-content-sha256: UNSIGNED-PAYLOAD
> > 3. for CanonicalRequest, use value from header or calculate the hash from postfields
> > 4. for SignedHeaders, add x-amz-content-sha256 if present
> 
> in the meantime, i've opened #9804 to address points 1 and 3, without the special case for service=s3 in point 2
> 
> point 4 was indeed addressed in #7966, thanks @outscale-mgo

For the point 2: once we have decide on the condition to enable it, it seems to me it would be more useful for the user, if we add x-amz-content-sha256 that contain the hash calculated from postfields, because adding x-amz-content-sha256: UNSIGNED-PAYLOAD is easier than calculating the hash.

cbodley:
> For the point 2: once we have decide on the condition to enable it, it seems to me it would be more useful for the user, if we add x-amz-content-sha256 that contain the hash calculated from postfields, because adding x-amz-content-sha256: UNSIGNED-PAYLOAD is easier than calculating the hash.

@outscale-mgo there are some request methods, like GET and POST, that the existing payload-hashing logic is handling well. but i do think we'll need to rely on UNSIGNED-PAYLOAD for requests like PUT

cbodley:
> I think there is 2 different approach to having a V4 generic:
> - 1rst one is to follow this https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html by default, and have an option for every differences you could find (like the one in s3).
> - 2nd one is to have a way to match which v4 we're trying to have (like with the service-type parameter).

thanks @outscale-mgo, i like how you've framed this

my preference would be for this second approach, because curl could automatically do the right thing when users point it at AWS. in essence, i view your `service-name[:service-type]` idea as the special-case for service=s3 i proposed, plus a config abstraction to sign *as if* the service was s3

i don't have a strong opinion on how that configuration works, but tend to agree that it's better to extend `--aws-sigv4` than to add a separate option

cbodley:
now that https://github.com/curl/curl/pull/9804 has merged, i've raised the followup https://github.com/curl/curl/pull/9995 to use UNSIGNED-PAYLOAD for aws:s3 requests. that includes a `bool sign_as_s3` flag that can be adjusted for other services, but i left that extra abstraction/configuration out for future work

bagder:
Closing due to inactivity

PR #8950, opened 2022-06-02
examples/crawler.c: use the curl license

bagder:
With permission from Jeroen Ooms

URL: https://github.com/curl/curl/pull/8869#issuecomment-1144742731


danielgustafsson (approved):
Thanks @jeroen !

mxmehl:
Oh, I also have done this in #8869 just now before seeing this PR. I wouldn't mind including this PR but would then like to ask you to add `SPDX-License-Identifier: curl` nearby so I don't have to re-add that in the REUSE branch.

PR #8961, opened 2022-06-05
select: return error from "lethal" poll/select errors

### LINKED ISSUE #8921: multi_wait() ignores Curl_poll() errors
`multi_wait()` ignores the potential error from `Curl_poll()` at https://github.com/curl/curl/blob/d67f41acd3c7c5f6f0ce46ae9a1c1db1457b90d3/lib/multi.c#L1315

If `Curl_poll()` returns immediately the function will then return `CURLM_OK` right away without performing any waiting. This will lead `curl_multi_poll()`/`curl_multi_wait()` to return immediately as well. Effectively this will lead to  a busyloop as there is no status returned that would lead to terminating a loop of a typical `curl_multi_perform()`  + `curl_multi_poll()` app, at least if the app doesn't bail out once the connections are completed (`still_running == 0` doesn't lead to loop termination).

### I did this

Had condition where `poll()` (or `select()` if using `poll()` emulation) return a permanent error without waiting.

### I expected the following

`curl_multi_poll()` / `curl_multi_wait()` to return an error if `Curl_poll()` fails.

The question is what error to return though as none of the `CURLM_` errors seem to fit the bill.

### curl/libcurl version

7.83.1

### operating system

This is a generic issue, but it is likely more likely to get triggered on specific platforms.

bagder:
Reported-by: Harry Sintonen
Fixes #8921


bagder:
@piru, does this fix the problem you could repro?

piru:
@bagder, It's looking good now.

PR #8964, opened 2022-06-06
Document security considerations: "CRLF in headers"

Haxatron:
Ref: https://hackerone.com/reports/1589877

### DIFF
--- a/docs/libcurl/libcurl-security.3
+++ b/docs/libcurl/libcurl-security.3
@@ -130,6 +130,16 @@ as necessary. Alternately, an app could leave \fICURLOPT_FOLLOWLOCATION(3)\fP
 enabled but set \fICURLOPT_REDIR_PROTOCOLS(3)\fP and install a
 \fICURLOPT_OPENSOCKETFUNCTION(3)\fP or \fICURLOPT_PREREQFUNCTION(3)\fP callback
 function in which addresses are sanitized before use.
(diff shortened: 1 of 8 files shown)


jay:
I don't think this warning is really needed but if it is does it really need two paragraphs?

Haxatron:
Hi, the rationale (from the hacker1 report) was that untrusted user input that contain newlines could be used to perform all sorts of bad stuff. This has appeared before, such as in https://medium.com/@tomnomnom/crlf-injection-into-phps-curl-options-e2e0d7cfe545. 

This update to the documentation aims to address that by informing users that libcurl will not sanitize headers, and that user have to sanitize headers by themselves. However, after thinking about it for a while, I think only libcurl-security.3 should have this warning, to reduce clutter in the options documentation. wdyt?

jay:
> This update to the documentation aims to address that by informing users that libcurl will not sanitize headers, and that user have to sanitize headers by themselves. However, after thinking about it for a while, I think only libcurl-security.3 should have this warning, to reduce clutter in the options documentation. wdyt?

I agree. I just landed that and I changed the warning sentence slightly so that it is broader than just new headers injection: "someone malicious may be able to modify the request in a way you didn't intend such as injecting new headers." Someone malicious could do \r\n\r\n and inject into the body, and there are probably other things I'm not thinking of.

PR #8972, opened 2022-06-07
curl_global_init.3: this is now (usually) thread-safe

bagder:
Follow-up to 23af112f5556

### DIFF
--- a/docs/libcurl/curl_global_init.3
+++ b/docs/libcurl/curl_global_init.3
@@ -44,16 +44,18 @@ In normal operation, you must specify CURL_GLOBAL_ALL. Do not use any other
 value unless you are familiar with it and mean to control internal operations
 of libcurl.
 
-\fBThis function is not thread safe.\fP You must not call it when any other
+This function is thread-safe since libcurl 7.84.0 if
+\fIcurl_version_info(3)\fP has CURL_VERSION_THREADSAFE_INIT feature bit set
+(most platforms).
+
+If this is not thread-safe, you must not call this function when any other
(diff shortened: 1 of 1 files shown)


docs/libcurl/curl_global_init.3
```diff
@@ -44,16 +44,18 @@ In normal operation, you must specify CURL_GLOBAL_ALL. Do not use any other
...
 thread in the program (i.e. a thread sharing the same memory) is running.
 This does not just mean no other thread that is using libcurl. Because
 \fIcurl_global_init(3)\fP calls functions of other libraries that are
 similarly thread unsafe, it could conflict with any other thread that uses
-these other libraries.
-
-If you are initializing libcurl from a Windows DLL you should not initialize it
-from DllMain or a static initializer because Windows holds the loader lock
-during that time and it could cause a deadlock.
+these other libraries. If you are initializing libcurl from a Windows DLL you
```
jay, line 56: The Windows loader lock warning paragraph should remain separate because even if initialization is thread-safe the loader can still deadlock if libcurl is initialized during DLL initialization. There's no way for us to prevent that, it is inherent in the way Windows works. DLLs can be loaded for any number of reasons it's not something that's limited to explicit calls in the library code.

jay (changes requested):

docs/libcurl/curl_global_init.3
jay: Ah I was 10 min too late. I fixed it in ae8a329.

PR #8975, opened 2022-06-07
maintain --path-as-is after redirects

bagder:


llamasoft:
I just tried out this branch from source and it works for me! 👍

PR #8976, opened 2022-06-08
test3026: add missing control file

bagder:
Follow-up from 2ed101256414ea5

Makes the test run, makes 'make dist' work

This single test takes 24-25 seconds on my machine. I think it's too
slow to have it run unconditionally among the test... For this reason I
tag it with a "slow" keyword.

### DIFF
--- a/tests/data/test3026
+++ b/tests/data/test3026
@@ -1,0 +1,44 @@
+<testcase>
+<info>
+<keywords>
+curl_global_init
+thread-safe
+slow
+</keywords>
+</info>
+
+#
+# Server-side
+<reply>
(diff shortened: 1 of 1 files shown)


bagder:
The test seems to run ok in most CIs. Let's add it like this (for now)

mback2k:
This PR had 26 failed CI runs and was still merged. :-(

mback2k:
Intermediate results of my test analysis: 32-bit Windows builds just fail the test and 64-bit Windows builds randomly timeout/fail, visible mostly on MinGW-based Windows builds in the CI.

Issue #8973, opened 2022-06-07
cmake doesn't check for _Atomic


bagder:
Starting with #8680, libcurl has a thread-safe `curl_global_init` on most platforms **iff using `_Atomic`** works and gets used. Currently, the cmake build does not check for this symbol, which makes is unfortunate and therefore cannot build the init function thread-safe on several platforms!

bagder:
This also makes test 1014 fail, at least on Appveyor: curl-config doesn't know about `threadsafe-init`

PR #8982, opened 2022-06-08
cmake: Check for C11 atomics and set config

cblichmann:
Context #8973

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1055,6 +1055,31 @@ if(HAVE_SIZEOF_LONG_LONG)
   set(HAVE_LL 1)
 endif()
 
+# Check for atomics support
+check_c_source_compiles(
+  "
+  #include <stdatomic.h>
+
+  int main(void)
+  {
+    atomic_int i = 0;
+    return atomic_load(&i);
+  }" HAVE_ATOMIC
+)
+
+# Check for SRW lock support
+if(WIN32)
+  check_c_source_compiles(
+    "
+    #include <windows.h>
+
+    #if (WINVER < 0x600) && (_WIN32_WINNT < 0x600)
+    #error
+    #endif
+    " HAVE_SRW_LOCK
+  )
+endif()
+
 find_file(RANDOM_FILE urandom /dev)
 mark_as_advanced(RANDOM_FILE)
 
@@ -1464,6 +1489,7 @@ _add_if("HTTP3"         USE_NGTCP2 OR USE_QUICHE)
 _add_if("MultiSSL"      CURL_WITH_MULTI_SSL)
 _add_if("HTTPS-proxy"   SSL_ENABLED AND (USE_OPENSSL OR USE_GNUTLS OR USE_NSS))
 _add_if("unicode"       ENABLE_UNICODE)
+_add_if("threadsafe"    (NOT WIN32 AND HAVE_ATOMIC) OR (HAVE_SRW_LOCK))
 string(REPLACE ";" " " SUPPORT_FEATURES "${_items}")
 message(STATUS "Enabled features: ${SUPPORT_FEATURES}")
 
--- a/lib/curl_config.h.cmake
+++ b/lib/curl_config.h.cmake
@@ -421,6 +421,9 @@
 /* Define to 1 if the compiler supports the 'long long' data type. */
 #cmakedefine HAVE_LONGLONG 1
 
+/* Define to 1 if the compiler supports C11 atomics. */
+#cmakedefine HAVE_ATOMIC 1
+
 /* Define to 1 if you have the malloc.h header file. */
 #cmakedefine HAVE_MALLOC_H 1
 
--- a/m4/curl-functions.m4
+++ b/m4/curl-functions.m4
@@ -6577,9 +6577,11 @@ AC_DEFUN([CURL_ATOMIC],[
   AC_MSG_CHECKING([if _Atomic is available])
   AC_COMPILE_IFELSE([
     AC_LANG_PROGRAM([[
+      #include <stdatomic.h>
       $curl_includes_unistd
     ]],[[
-      _Atomic int i = 0;
+      atomic_int i = 0;
+      (void) atomic_load(&i);
     ]])
   ],[
     AC_MSG_RESULT([yes])


cblichmann:
Since thread-safe init is a significant feature (IMO), I think it does.
I'll update the PR.

CMakeLists.txt
```diff
@@ -1053,6 +1053,16 @@ if(HAVE_SIZEOF_LONG_LONG)
   set(HAVE_LL 1)
 endif()
 
+# Check for atomics support
+check_c_source_compiles(
+  "
+  int main(int, char**)
+  {
+    _Atomic int i = 0;
+    return 0;
```
friendlyanon, line 1062:
```suggestion
  #include <stdatomic.h>
  
  int main(void)
  {
    atomic_int i = 0;
    return atomic_load(&i);
```

I would personally change the test to this, so there is no chance of warnings being raised here. This is enough to not make Clang emit anything with `-Weverything`.

cblichmann: Sure. I'll change the check in the Autotools code as well then.

bagder: is `atomic_int` really C11? I can only find references to it in C++ docs?

cblichmann:
Any tips for how I should be dealing with the test failures? Some seem unrelated to this change and others I don't quite understand yet.

Unrelated (missing `psl` feature):
```
2022-06-08 16:31:48.687221 | ubuntu-bionic | test 1014...[Compare curl --version with curl-config --features]
2022-06-08 16:31:49.525915 | ubuntu-bionic | Mismatch in features lists:
2022-06-08 16:31:49.525971 | ubuntu-bionic | curl:        alt-svc asynchdns brotli hsts http3 https-proxy idn ipv6 largefile ntlm psl ssl threadsafe-init unixsockets zstd
2022-06-08 16:31:49.525984 | ubuntu-bionic | curl-config: alt-svc asynchdns brotli hsts http3 https-proxy idn ipv6 largefile ntlm ssl threadsafe-init unixsockets zstd
2022-06-08 16:31:49.526489 | ubuntu-bionic |  postcheck FAILED
```

Failing on Windows using msys v2:
```
2022-06-08T18:00:16.9622957Z test 3026...[curl_global_init thread-safety]
2022-06-08T18:00:17.6232259Z 
2022-06-08T18:00:17.6233251Z lib3026.exe returned 127, when expecting 0
2022-06-08T18:00:17.6233869Z  exit FAILED
2022-06-08T18:00:17.6238697Z == Contents of files in the log/ dir after test 3026
2022-06-08T18:00:17.6245538Z === Start of file commands.log
2022-06-08T18:00:17.6249392Z  ./libtest/lib3026.exe none >log/stdout3026 2>log/stderr3026
2022-06-08T18:00:17.6250062Z === End of file commands.log
2022-06-08T18:00:17.6255678Z === Start of file ftpserver.cmd
2022-06-08T18:00:17.6259999Z  Testnum 3026
2022-06-08T18:00:17.6260525Z === End of file ftpserver.cmd
2022-06-08T18:00:17.6266788Z === Start of file stderr3026
2022-06-08T18:00:17.6271480Z  URL: none
2022-06-08T18:00:17.6283756Z  lib3026.c:65 Couldn't create thread, errno 11
2022-06-08T18:00:17.6284691Z === End of file stderr3026
```

Shouldn't the above also fail when we're preparing the build files with autotools?

bagder:
The issue here is that the `threadsafe-init` feature needs to be enabled for curl-config for Windows as well, and not only for `_Atomic` present. Like is done in this configure check:

https://github.com/curl/curl/blob/5121d98c22dc077d72c68e7b9436a04f55bd7249/configure.ac#L4261-L4276

Keitagit-kun:
So, if target _WIN32_WINNT < 0x600 threadsafe-init feature should be disabled (will be not selectable)?

bagder:
Exactly. On Windows we rely on `SWRLOCK` instead of atomic and then presumably that is the lowest Windows version that supports it. See [easy_lock.h](https://github.com/curl/curl/blob/master/lib/easy_lock.h)

bagder:
This [fails on macos](https://github.com/curl/curl/runs/6834074919?check_suite_focus=true#step:6:9):
~~~
Run cmake -H. -Bbuild -DCURL_WERROR=ON -DPICKY_COMPILER=ON -DOPENSSL_ROOT_DIR=/usr/local/opt/libressl -DCURL_DISABLE_LDAP=ON -DCURL_DISABLE_LDAPS=ON
CMake Error at CMakeLists.txt:1080:
  Parse error.  Expected "(", got newline with text "
  ".
-- Configuring incomplete, errors occurred!

cblichmann:
Sorry about that, I missed the parenthesis afer `endif`. Apparently some CMake version simply accept this (as did my local one).

cblichmann:
According to #8996, we can ignore test 3026 failing.

cblichmann:
Let's do the fix for `-H` in a different PR :)

CMakeLists.txt
```diff
@@ -1464,6 +1489,7 @@ _add_if("HTTP3"         USE_NGTCP2 OR USE_QUICHE)
 _add_if("MultiSSL"      CURL_WITH_MULTI_SSL)
 _add_if("HTTPS-proxy"   SSL_ENABLED AND (USE_OPENSSL OR USE_GNUTLS OR USE_NSS))
 _add_if("unicode"       ENABLE_UNICODE)
+_add_if("threadsafe-init" (NOT WIN32 AND HAVE_ATOMIC) OR (HAVE_SRW_LOCK))
```
bagder, line 1492: This feature has been renamed to `threadsafe` while this PR has been in the works!

bagder:
Can you figure out why the appveyor windows builds using cmake don't get the `threadsafe` feature listed in `curl-config` output thus making test 1014 fail?

[example build](https://ci.appveyor.com/project/curlorg/curl/builds/43856658/job/fal7pjkuelama6di#L3616)

cblichmann:
Looking into this. I first rebased on `master` though.

vszakats (approved):

bagder:
The cmake build still [fails test 1014 on appveyor](https://ci.appveyor.com/project/curlorg/curl/builds/43865783/job/hmw0924i9qu7so2l#L3609). Which seems to be because `curl-config --feature` doesn't get the `threadsafe` feature correctly.

bagder:
This needs a rebase and force-push. I think the curl-functions edit can be removed and the cmake check should make sure to also check for the presence of the `stdatomic.h`  header file before it defines `HAVE_ATOMIC`, to match how we improved the autotools check.

Issue #8984, opened 2022-06-09
imap_perform_authentication() does not seem to fallback to cleartext authentication as expected.


kop316:
Hello! For a bit of background, I developed and maintain an application called vvmd (a Visual Voicemail application for mobile phones).

Visual Voicemail works primarily over IMAP, so I use libcurl as my backend.

### I did this

Recently, we discovered that IMAP authentication was not working for a subset of T-Mobile USAs IMAP servers:
https://gitlab.com/kop316/vvmd/-/issues/12

Mobile carriers are notorious for misconfigured IMAP servers, so in `vvmd`, I do a variety of login tests to determine the correct authentication method for an IMAP Voicemail server.

In T-mobile's case, the IMAP server can advertise
```
< * OK [CAPABILITY AUTH=PLAIN AUTH=DIGEST-MD5 IMAP4rev1 QUOTA UIDPLUS] mStore IMAP4rev1 Server is ready.
> A001 CAPABILITY
< * CAPABILITY AUTH=PLAIN AUTH=DIGEST-MD5 IMAP4rev1 QUOTA UIDPLUS
```

But attempts to authenticate with `AUTH=DIGEST-MD5` do not work with T-Mobile's IMAP server, only `AUTH=PLAIN`:

```
> A001 CAPABILITY
< * CAPABILITY AUTH=DIGEST-MD5 IMAP4rev1 QUOTA UIDPLUS
< A001 OK CAPABILITY completed.
> A002 AUTHENTICATE DIGEST-MD5
< + cmVhbG09InJlYWxtQG1hdmVuaXIuY29tIixub25jZT0iTURRNVptWmhOek10TVRZek5DMDBNVEppTFRrMVlXRXRNelpoWldKak16ZGlaR1ZqIixxb3A9ImF1dGgiLGNoYXJzZXQ9dXRmLTgsYWxnb3JpdGhtPW1kNS1zZXNz
[...]

Issue #8979, opened 2022-06-08
IMAP doesn't do LOGIN as needed


MonkeybreadSoftware:
### I did this

We tried to login on IMAP server for Outlook 365.

### I expected the following

Usually for IMAP it works, but this server doesn't show AUTH=LOGIN, but that is needed to login.

### curl/libcurl version

CURL 7.83.1 with openssl

### operating system

macOS 10.12.

So this is our CURL log:

```
  Trying 52.98.175.2:993...
Connected to outlook.office365.com (52.98.175.2) port 993 (#0)
TLSv1.3 (OUT), TLS handshake, Client hello (1):
TLSv1.3 (IN), TLS handshake, Server hello (2):
TLSv1.2 (IN), TLS handshake, Certificate (11):
TLSv1.2 (IN), TLS handshake, Server key exchange (12):
TLSv1.2 (IN), TLS handshake, Server finished (14):
TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
TLSv1.2 (OUT), TLS handshake, Finished (20):
TLSv1.2 (IN), TLS handshake, Finished (20):
SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
Server certificate:
 subject: C=US; ST=Washington; L=Redmond; O=Microsoft Corporation; CN=outlook.com
 start date: Oct 24 00:00:00 2021 GMT
 expire date: Oct 23 23:59:59 2022 GMT
 issuer: C=US; O=DigiCert Inc; CN=DigiCert Cloud Services CA-1
 SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* OK The Microsoft Exchange IMAP4 service is ready. [...]
A001 CAPABILITY
* CAPABILITY IMAP4 IMAP4rev1 AUTH=PLAIN AUTH=XOAUTH2 SASL-IR UIDPLUS ID UNSELECT CHILDREN IDLE NAMESPACE LITERAL+
A001 OK CAPABILITY completed.
No known authentication mechanisms supported
Closing connection 0
TLSv1.2 (OUT), TLS alert, close notify (256):
```

As you see server can do AUTH=PLAIN AUTH=XOAUTH2.
Now we do the same in Apple Mail:

```
INITIATING CONNECTION Jun 08 14:34:04.800 host:outlook.office365.com -- port:143 -- socket:0x0 -- thread:0x60000249d100

[...]

PR #8986, opened 2022-06-10
imap: Fallback to the LOGIN command when the AUTHENTICATE command fails

captain-caveman2k:
As per section 6.2.3 of RFC-9051 allow the fallback to a clear text
LOGIN when a SASL based AUTHENTICATE fails.

### DIFF
--- a/lib/imap.c
+++ b/lib/imap.c
@@ -21,7 +21,7 @@
  * RFC2195 CRAM-MD5 authentication
  * RFC2595 Using TLS with IMAP, POP3 and ACAP
  * RFC2831 DIGEST-MD5 authentication
- * RFC3501 IMAPv4 protocol
+ * RFC3501 Internet Message Access Protocol - Version 4rev1
  * RFC4422 Simple Authentication and Security Layer (SASL)
  * RFC4616 PLAIN authentication
  * RFC4752 The Kerberos V5 ("GSSAPI") SASL Mechanism
@@ -29,6 +29,7 @@
  * RFC5092 IMAP URL Scheme
  * RFC6749 OAuth 2.0 Authorization Framework
  * RFC8314 Use of TLS for Email Submission and Access
+ * RFC9051 Internet Message Access Protocol (IMAP) - Version 4rev2
  * Draft   LOGIN SASL Mechanism <draft-murchison-sasl-login-00.txt>
  *
  ***************************************************************************/
@@ -1025,6 +1026,9 @@ static CURLcode imap_state_auth_resp(struct Curl_easy *data,
     default:
       break;
     }
+  else if((!imapc->login_disabled) && (imapc->preftype & IMAP_TYPE_CLEARTEXT))
+    /* Perform clear text authentication on failure as per RFC-9051 */
+    result = imap_perform_login(data, conn);
 
   return result;
 }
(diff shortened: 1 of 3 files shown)


lib/imap.c
```diff
@@ -1025,6 +1026,9 @@ static CURLcode imap_state_auth_resp(struct Curl_easy *data,
     default:
       break;
     }
+  else if((!imapc->login_disabled) && (imapc->preftype & IMAP_TYPE_CLEARTEXT))
```
jay, line 1029: Ref: https://datatracker.ietf.org/doc/html/rfc9051#section-6.2.3

Though the burden is on the server to provide logindisabled should this fallback be allowed without ssl, in other words why not this:

```suggestion
  else if(!imapc->login_disabled && imapc->ssldone &&
          (imapc->preftype & IMAP_TYPE_CLEARTEXT))
```

jay:
Ref: https://github.com/curl/curl/issues/8979
Ref: https://github.com/curl/curl/issues/8984

lib/imap.c
emilengler: While I support the prohibition of plain-text passwords without TLS, I'm a bit afraid that this would break scripts.

captain-caveman2k: I think the login disabled flag is sufficient as we do for the other use of cleartext in IMAP.

Not only that but introducing the dependency on TLS would prevent curl or libcurl being used in systems where other mechanisms are in place to secure the network - such as a closed wall system that has no outside access.

captain-caveman2k: I note that this change has broken test849 - whilst I have fixed it up locally to add the fallback I am still deliberating whether cleartext fallback should be used or not when an authzid has been provided.

Does anyone have any strong views either way?

captain-caveman2k:
I also note that this change has broken test849 - whilst I have fixed it up locally to add the fallback I am still deliberating whether cleartext fallback should be used or not when an authzid has been provided.

Does anyone have any strong views either way?

kop316:
If I may ask, is there anything I can do to help this get merged? This fixes a serious issue for me, and I don't want to tell folks the solution to a problem is to add an unmerged branch of curl.

jay:
> If I may ask, is there anything I can do to help this get merged? This fixes a serious issue for me, and I don't want to tell folks the solution to a problem is to add an unmerged branch of curl.

It needs an update from the OP to proceed. I'm sort of ok with it. I don't like providing fallbacks on unencrypted protocols. OTOH the user has the option of explicitly using SSL/TLS to prevent MITM.

monnerat:
> As per section 6.2.3 of RFC-9051 allow the fallback to a clear text LOGIN when a SASL based AUTHENTICATE fails.

This also is an implicit recognition  that 2 auth mechanism on the same server may have different user/credential pairs: the mechanism is then a significant parameter of the authentication and should be checked for cache re-use too!

bagder:
> This also is an implicit recognition that 2 auth mechanisms on the same server may have different user/credential pairs

Yikes, that would make a major impact. But the way I read the spec, it could also be a way to just deal with AUTHENTICATE not working for whatever reason, which would be less impactful.

kop316:
Hello,

While I understand it needs an update from the OP, is there anything I could do to update the patch so that this could get merged? Based on the discussion, it seems that it may be ok to merge as is, or a slight tweak is required. This is required to get a program I have in a few linux distributions to work, and (understandably), none are willing to carry an unmerged patch.

Thank you!

bagder:
This work has stalled since a long time so I'm closing this to not make the impression this is still actually in progress. Feel free to create a new PR when/if work resumes.

PR #8988, opened 2022-06-10
rtmp: deprecate "wrong" RTMP protocol defines

bagder:
The 32 bit protocol bitmask is getting full so adding new protocols
become complicated when they no longer fit.

There are 4 defines that were wrongly added to the set back in 2010 that
we now change and move out to make room for a few more protocols until
we hit the ceiling for real. I say wrongly, because they are not
separate enough protocols to warrant their own protocol bits like
this. They are:

 CURLPROTO_RTMPT
 CURLPROTO_RTMPE
 CURLPROTO_RTMPTE
 CURLPROTO_RTMPTS

Leaving CURLPROTO_RTMP and CURLPROTO_RTMPS as-is.

RTMP is extremely rarely used by curl users (1.5% use RTMP, 1.7% RTMPS),
and there's likely much fewer users who ever used one of these modified
defines.

Modifying these defines is less than ideal and involves a certain risk
to cause issues for users who are actually using them individually.

### DIFF
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -1028,17 +1028,23 @@ typedef CURLSTScode (*curl_hstswrite_callback)(CURL *easy,
 #define CURLPROTO_SMTPS  (1<<17)
 #define CURLPROTO_RTSP   (1<<18)
 #define CURLPROTO_RTMP   (1<<19)
-#define CURLPROTO_RTMPT  (1<<20)
-#define CURLPROTO_RTMPE  (1<<21)
-#define CURLPROTO_RTMPTE (1<<22)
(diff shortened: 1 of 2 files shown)


hyc:
The question is what curl needs to know about these to keep them straight. If curl doesn't care itself, and the details are all left to another library (like librtmp) then perhaps you only need CURLPROTO_RTMP.

The others certainly are distinct on the wire, e.g. RTMPT is tunneled in HTTP, RTMPE is encrypted by Adobe's proprietary scheme, etc...

bagder:
i'm gonna go ahead in a different way

PR #8989, opened 2022-06-10
version: rename threadsafe-init to threadsafe

vszakats:
Referring to Daniel's article [^1], making the init function thread-safe
was the last bit to make libcurl thread-safe as a whole. So the name of
the feature may as well be the more concise 'threadsafe', also
telling the story that libcurl is now fully thread-safe, not just its
init function. Chances are high that libcurl wants to remain so in the
future, so there is little likelyhood of ever needing any other distinct
`threadsafe-<name>` feature flags.

[^1]: https://daniel.haxx.se/blog/2022/06/08/making-libcurl-init-more-thread-safe/

---
1. Comments are welcome whether this sounds something reasonable
and desired.
1. If so, it raises the question if `CURL_VERSION_THREADSAFE_INIT`
may be renamed to `CURL_VERSION_THREADSAFE` also?


jay:
Well we have a document of [thread safety guidelines](https://curl.se/libcurl/c/threadsafe.html) (which, reminds me, needs to be updated for curl_global_init change) so to have a broader flag indicating that the library is thread-safe I think we should refer to that document in the flag description. Basically we can say libcurl is thread-safe but we can't guarantee that of the libraries it's using, though in many cases as @bagder mentioned (somewhere) that is a non issue these days.

bagder (approved):
While this particular change modifies the init, I wouldn't mind a shorter name and since the rest was already thread-safe it is also not wrong. I approve.

vszakats:
@bagder: What do you think about shortening the constant name accordingly?: `CURL_VERSION_THREADSAFE_INIT` → `CURL_VERSION_THREADSAFE`

bagder:
I'm **for** that as well.

PR #8991, opened 2022-06-11
projects: Fix the third-party SSL library build paths for Visual Studio

captain-caveman2k:
The paths used by the build batch files were inconsistent with those in
the Visual Studio project files - which means the OpenSSL and wolfSSL
builds would fail trying to link.


captain-caveman2k:
Note: This is not intended to fix the inconsistency in the Visual Studio version arguments between build-openssl.bat, build-wolfssl.bat and generate.bat.

PR #8992, opened 2022-06-11
setopt: add CURLOPT_PROTOCOLS_STR and CURLOPT_REDIR_PROTOCOLS_STR

bagder:
... as replacements for deprecated CURLOPT_PROTOCOLS and
CURLOPT_REDIR_PROTOCOLS as these new ones do not risk running into the
32 bit limit the old ones are facing.

CURLINFO_PROTOL is now deprecated.

The curl tool is updated to use the new options.


dfandrich:
Wouldn't a curl_slist of strings instead of a single string be more useful? It would be easier to parse and probably faster, too.

lib/setopt.c
```diff
@@ -146,6 +146,86 @@ static CURLcode setstropt_userpwd(char *option, char **userp, char **passwdp)
...
+  bool found_comma = FALSE;
+  static struct scheme {
+    const char *name;
+    long bit;
+  } const protos[] = {
+    { "dict", CURLPROTO_DICT },
+    { "file", CURLPROTO_FILE },
+    { "ftp", CURLPROTO_FTP },
+    { "ftps", CURLPROTO_FTPS },
+    { "gopher", CURLPROTO_GOPHER },
```
emilengler, line 160: I think you forgot `gophers`?

bagder:
> Wouldn't a curl_slist of strings instead of a single string be more useful? It would be easier to parse and probably faster, too.

Faster to parse, but perhaps a bit more inconvenient to create?

lib/setopt.c
bagder: I certainly did!

lib/setopt.c
```diff
@@ -146,6 +146,85 @@ static CURLcode setstropt_userpwd(char *option, char **userp, char **passwdp)
...
+    { "imaps", CURLPROTO_IMAPS },
+    { "ldap", CURLPROTO_LDAP },
+    { "ldaps", CURLPROTO_LDAPS },
+    { "mqtt", CURLPROTO_MQTT },
+    { "pop3", CURLPROTO_POP3 },
+    { "pop3s", CURLPROTO_POP3S },
+    { "rtmp", CURLPROTO_RTMP },
+    { "rtmpe", CURLPROTO_RTMPE },
+    { "rtmps", CURLPROTO_RTMPS },
+    { "rtmpT", CURLPROTO_RTMPT },
```
CommanderRoot, line 174: I don't think the uppercase "T" is wanted here?

bagder: Agreed, that looks bad. Fixing!

docs/libcurl/opts/CURLOPT_PROTOCOLS_STR.3
```diff
@@ -0,0 +1,75 @@
...
+If trying to set a non-existing protocol or if no matching protocol at all is
+set, it returns error.
+
+These are the available protocols:
+
+DICT, FILE, FTP, FTPS, GOPHER, GOPHERS, HTTP, HTTPS, IMAP, IMAPS, LDAP, LDAPS,
+POP3, POP3S, RTMP, RTMPE, RTMPS, RTMPT, RTMPTE, RTMPTS, RTSP, SCP, SFTP, SMB,
+SMBS, SMTP, SMTPS, TELNET, TFTP
+
+As a short-cut, you can set "ALL" as a short-cut to enable all protocols. Note
```
rybak, line 49: "as a short-cut" is repeated twice in this sentence. Same thing in `CURLOPT_REDIR_PROTOCOLS_STR.3`.

bagder: Thanks, fixed now!

include/curl/curl.h
```diff
@@ -2141,6 +2142,15 @@ typedef enum {
   /* set the SSH host key callback custom pointer */
   CURLOPT(CURLOPT_SSH_HOSTKEYDATA, CURLOPTTYPE_CBPOINT, 317),
 
+  /* specify which protocols that are allowed to be used for the transfer,
+     which thus helps the app which takes URLs from users or other external
+     inputs and want to restrict what protocol(s) to deal with. Defaults to
+     all built-in prootocols. */
```
williamdes, line 2148:
```suggestion
     all built-in protocols. */
```

bagder: Thanks!

docs/libcurl/opts/CURLOPT_PROTOCOLS_STR.3
```diff
@@ -0,0 +1,77 @@
...
+CURLOPT_PROTOCOLS_STR \- allowed protocols
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+
+CURLcode curl_easy_setopt(CURL *handle, CURLOPT_PROTOCOLS_STR, char *spec);
+.fi
+.SH DESCRIPTION
+Pass a pointer to a string that holds a comma-separated list of case
+instensitive protocol names (URL schemes) to allow in the transfer. This
```
maxdymond, line 36: sp:insensitive

bagder: Thanks!

dfandrich:
On Mon, Jun 13, 2022 at 02:14:13AM -0700, Daniel Stenberg wrote:
> +instensitive protocol names (URL schemes) to allow in the transfer. This

Typo here, too: insensitive

bagder:
> Typo here, too: insensitive

Thanks! I fixed both of them actually.

dfandrich (commented):
Typo in commit message: CURLINFO_PROTOL

dfandrich:
Do the protocols accepted by CURLOPT_PROTOCOLS_STR exactly match the ones supplied by curl_version_info()? There should probably be a reference to that function in this man page as being the way to retrieve all allowed protocols.

bagder:
> Typo in commit message: CURLINFO_PROTOL

Fixed locally now.

> Do the protocols accepted by CURLOPT_PROTOCOLS_STR exactly match the ones supplied by curl_version_info()? 

Yes - all protocols `curl_version_info` shows are supported by this option and spelled the exact same way. `CURLOPT_PROTOCOLS_STR` supports the extra RTMP* ones from the `CURLPROTO_*` defines as well, and they are not listed by `curl_version_info`.

> There should probably be a reference to that function in this man page as being the way to retrieve all allowed protocols.

Good idea. I'll amend!

PR #8995, opened 2022-06-12
Websockets

bagder:
**Warning:** regular rebase + force-pushes are still performed on this branch.

[Initial design thoughts](https://github.com/curl/curl/wiki/WebSockets)

- [X] setopt documentation
- [X] configure
- [X] decide to not rely on any external websocket library
- [X] Send HTTP request
- [X] Send HTTP request when issued with hyper
- [X] websocket URL parser (+ tests for it)
- [X] curl_version_info() protocols[] fixed
- [X] HTTP Upgrade response check
- [X] extended CURLOPT_CONNECT_ONLY to offer **upgrade-only**
- [X] curl_ws_send documentation
- [X] curl_ws_recv documentation
- [X] curl_ws_send implementation
- [X] curl_ws_recv implementation
- [X] decide how to do the test server: extend sws
- [X] do proper non-blocking buffer management when reading slow/a lot
- [X] test server support for websockets (partial)
- [X] test cases running (ws test infrastructure working)
- [X] document curl_ws_meta();
- [X] implement curl_ws_meta();
- [X] add test using curl_ws_meta();
- [X] fail gracefully if not getting the expected101 response code + test it
- [X] allow overriding request headers in the ws initial request

## Left for follow-up work

-  make websockets work with hyper
- do the sha1 response verification
- consider a `curl_ws_poll()`
- make nc-like functionality for the curl tool
- make sure ws code paths are fuzzed
-  disable PING-PONG automation
- `CURLWS_COMPRESS`


calvin2021y:
Hi  bagder,  thanks for the great work.

I plan to use libcurl with event loop to replace libwebsockets.   The new implement will support non-blocked frame recv ?

I guess call curl_ws_recv will blocked, so some thing like CURLOPT_WS_RECVFUNCTION, CURLOPT_WS_RECVDATA should exists to handle the callback inside a event loop.

bagder:
I think maybe it might be better to take this to the [discussions](https://github.com/curl/curl/discussions) rather than to have this "side conversation" here in the PR.

> I plan to use libcurl with event loop to replace libwebsockets. The new implement will support non-blocked frame recv ?

I'm not sure what that means. The proposed API is documented in the wiki and in somewhat more detail in the docs section in the files in this PR.

> I guess call curl_ws_recv will blocked

It will return a full frame is there is one, I think it might not return anything if there isn't a complete frame to return.

> so some thing like CURLOPT_WS_RECVFUNCTION, CURLOPT_WS_RECVDATA should exists to handle the callback inside a event loop.

This has been discussed at length and I am currently providing the websockets data using the "regular" write callback `CURLOPT_WRITEFUNCTION`. Additional details about the delivered frame will be available from within the callback using the `curl_ws_meta()` function.

bagder:
@seanmonstar I'm now looking at making my WebSockets PR work with Hyper, but already when I create the HTTP request hyper says "header map at capacity" and dies. What does it mean?

~~~
thread '<unnamed>' panicked at 'header map at capacity', /home/daniel/.cargo/registry/src/github.com-1ecc6299db9ec823/http-0.2.8/src/header/map.rs:1453:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
panic unwind caught, aborting
~~~

seanmonstar:
> header map at capacity

Huh, I had not seen that before. Checking the source, it looks like there's a max of 32768 individual header entries in the map. I don't remember the rationale, perhaps to use a smaller integer type as indices...

Are you calling `hyper_headers_add` over 32,000 times on a single map? That sounds unlikely...

bagder:
Certainly not intentionally. Maybe I messed up. I will go back and double-check.

seanmonstar:
I doubt you are. I wonder what this could be, I've never seen anyone hit that error in equivalent Rust code. I'm now suspicious of the C API code I wrote -.-

bagder:
I figured it out. I passed in the *wrong* `hyper_headers *` in the first argument - I accidentally used the return value from `hyper_request_new();` instead of the variable holding the return value from `hyper_request_headers()`!

Sorry for the noise.

bagder:
I intend to merge this soon.

PR #8999, opened 2022-06-13
add missing SPDX-License-Identifier

bagder:
@mxmehl can you figure out why the REUSE CI job didn't catch these?

### DIFF
--- a/docs/libcurl/opts/CURLINFO_CAINFO.3
+++ b/docs/libcurl/opts/CURLINFO_CAINFO.3
@@ -18,6 +18,8 @@
 .\" * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
 .\" * KIND, either express or implied.
 .\" *
+.\" * SPDX-License-Identifier: curl
+.\" *
 .\" **************************************************************************
 .\"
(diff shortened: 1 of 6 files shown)


bagder:
The REUSE CI check said :heavy_check_mark: in the master branch for the commit immediately before I merged this PR, so clearly the CI job did not detect that the `SPDX-License-Identifier` was missing on 5 files. The copyright.pl script did though when I ran it manually.

bagder:
Aha, so that ignores files *recursively* ? That's not very good. We have 923 files under docs/ and 624 of them have copyright lines.

The ones we are careful are the markdown ones in the docs/ dir only.

mxmehl:
Yes, the wildcard does that for every file in this directory. But **ignore** is the wrong word: it does recursively add this copyright and license info to these files, even if quite far away from the actual files (which is why the dep5 file should be the last resort in REUSE).[^1]

We could do the following: change line 12 to `Files: docs/*.md` and make sure all files except the markdown files have valid copyright/licensing information. The same could apply to the other wildcards in this file.

Unfortunately I can only work on this from late July on as I am currently actually on parental leave. You could create an issue with this and other quirks you may find and assign me to it unless you'd prefer to tackle this on your own (which shouldn't be hard).

[^1]: We are working on a better, more flexible solution that can be put closer to the described files.

PR #8996, opened 2022-06-13
test3026: disable on win32

bagder:
... as it's not likely to have working pthreads

PR #9012, opened 2022-06-14
test3026: add support for Windows using native Win32 threads

mback2k:
Also make sure the result is initialized to an invalid value (`CURL_LAST`) before the corresponding thread is launched.

Follow up to #8996

### DIFF
--- a/tests/data/test3026
+++ b/tests/data/test3026
@@ -20,7 +20,6 @@ slow
 <features>
 threadsafe
 threaded-resolver
-!win32
 </features>
 <server>
 none
--- a/tests/libtest/lib3026.c
+++ b/tests/libtest/lib3026.c
@@ -26,11 +26,69 @@
 #include "testutil.h"
 #include "warnless.h"
 
-#ifdef HAVE_PTHREAD_H
+#define NUM_THREADS 100
+
+#ifdef WIN32
+static DWORD WINAPI run_thread(LPVOID ptr)
+{
+  CURLcode *result = ptr;
+
+  *result = curl_global_init(CURL_GLOBAL_ALL);
+  if(*result == CURLE_OK)
+    curl_global_cleanup();
+
+  return 0;
+}
+
+int test(char *URL)
+{
+  CURLcode results[NUM_THREADS];
+  HANDLE ths[NUM_THREADS];
+  unsigned tid_count = NUM_THREADS, i;
+  int test_failure = 0;
+  curl_version_info_data *ver;
+  (void) URL;
+
+  ver = curl_version_info(CURLVERSION_NOW);
+  if((ver->features & CURL_VERSION_THREADSAFE) == 0) {
+    fprintf(stderr, "%s:%d On Windows but the "
+            "CURL_VERSION_THREADSAFE feature flag is not set\n",
+            __FILE__, __LINE__);
+    return -1;
+  }
+
+  for(i = 0; i < tid_count; i++) {
+    HANDLE th;
+    results[i] = CURL_LAST; /* initialize with invalid value */
+    th = CreateThread(NULL, 0, run_thread, &results[i], 0, NULL);
+    if(!th) {
+      fprintf(stderr, "%s:%d Couldn't create thread, errno %d\n",
+              __FILE__, __LINE__, GetLastError());
+      tid_count = i;
+      test_failure = -1;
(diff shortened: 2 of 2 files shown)


tests/libtest/lib3026.c
```diff
@@ -61,7 +120,8 @@ int test(char *URL)
   }
 
   for(i = 0; i < tid_count; i++) {
-    int res = pthread_create(&tids[i], NULL, run_thread, &results[i]);
+    results[i] = CURL_LAST;
+    res = pthread_create(&tids[i], NULL, run_thread, &results[i]);
```
vszakats, line 124: #nit: Any reason for moving `int res;` a level up?

mback2k: To avoid the error about mixing declarations and code.

vszakats: This should compile fine:
```c
  for(i = 0; i < tid_count; i++) {
    int res;
    results[i] = CURL_LAST;
    res = pthread_create(&tids[i], NULL, run_thread, &results[i]);
```

mback2k:
It seems like the threadsafe unit is broken on 32-Bit Windows.

jay:
> It seems like the threadsafe init is broken on 32-Bit Windows.

All the CI failures I see are errno 8 (ERROR_NOT_ENOUGH_MEMORY). Creating 1000 test threads is probably too many. Which CI are you referring to? You could try changing the stack size? Also I think on error it should say GetLastError instead of errno since it's actually an error code from the former.

tests/libtest/lib3026.c
```diff
@@ -26,12 +26,70 @@
...
+  unsigned tid_count = NUM_THREADS, i;
+  int test_failure = 0;
+  curl_version_info_data *ver;
+  HANDLE th;
+  (void) URL;
+
+  ver = curl_version_info(CURLVERSION_NOW);
+  if((ver->features & CURL_VERSION_THREADSAFE) == 0) {
+    fprintf(stderr, "%s:%d On Windows but the "
+            "CURL_VERSION_THREADSAFE feature flag is not set\n",
```
jay, line 56: Is this really an error, isn't it expected when target is < 0x600?

bagder:
> Creating 1000 test threads is probably too many

If that is a problem, we can probably consider it tested with 100 too...

tests/libtest/lib3026.c
mback2k: Yes, but I think the test should not be executed at all in that case.

mback2k:
Yeah, I will try to reduce the number of threads in case a 32-but build is done.

tests/libtest/lib3026.c
bagder: I don't think we have a very convenient way to disable a test for that specific condition though. It can be made as a `<precheck>` I suppose.

mback2k: Hm, can't I detect the feature not being present via the test data file?

mback2k: Thanks @jay

tests/libtest/lib3026.c
mback2k: Thanks @vszakats

tests/data/test3026
```diff
@@ -15,11 +15,8 @@ slow
 #
 # Client-side
 <client>
-# require the threaded resolver only because it means pthreads might
-# be used for it
 <features>
-threaded-resolver
```
bagder, line 21: It still needs the threaded-resolver feature present as otherwise we don't know it has thread support present.

mback2k: That may be true for Linux, but I don't think this is correct for Windows builds, right?

bagder: Right, it's primarily for non-Windows reasons but then again there aren't many windows builds done without `threaded-resolver`.

jay: I see. The test will fail for non-Windows builds of libcurl that don't link pthreads.

mback2k: Okay, then I will keep `threaded-resolver`.

mback2k:
CI failures are unrelated. I think this is ready for merge. What do you think @bagder @jay @vszakats ?

vszakats (approved):

PR #9014, opened 2022-06-14
ci: avoid `cmake -Hdir`

### LINKED ISSUE #9008: Drop misuse of CMake's `-H` flag
Reference: https://github.com/curl/curl/pull/8982#issuecomment-1155436824

For justification, see this commit https://github.com/microsoft/vscode-cmake-tools/commit/2c181deca469b37be944921c533eef74e0d7de8a and the issue it closes.

emanuele6:
This is an undocumented option, but behaves similarly to the `-S <path>` introduced in cmake 3.10.

Since curl's current CMakeLists.txt file supports cmake 3.2 or newer, we are replacing `-Hsdir -Bbdir` with the more portable `-Bbdir sdir` instead of `-Sbdir -Bbdir`.

Fixes #9008

Also see https://github.com/curl/curl/issues/9008#issuecomment-1155795538


friendlyanon (commented):
Please see the response in the issue. Either make sure that CMake 3.13 or newer is used in CI, or use the old style of invoking.

CMake >=3.13:

```
cmake -Bbuild # -S can be omitted if -S. is used
```

CMake <3.13

```
mkdir build
cd build
cmake .. # cwd is binary dir, path argument is source dir
```

scripts/zuul/script.sh
```diff
@@ -90,7 +90,7 @@ if [ "$T" = "tidy" ]; then
 fi
 
 if [ "$T" = "cmake" ]; then
-  cmake -H. -Bbuild -DCURL_WERROR=ON $C
+  cmake -S. -Bbuild -DCURL_WERROR=ON $C
```
friendlyanon, line 93: The Zuul CI scripts are run on Ubuntu Bionic:

https://github.com/curl/curl/blob/f0f587d3471c8efe8687186e5749a4e88498d53f/zuul.d/jobs.yaml#L36

Bionic only packages CMake 3.10.2: https://packages.ubuntu.com/bionic/cmake

I looked at the list of checks, but I don't see where Zuul CI is. Does it have GitHub Actions integration?

emanuele6: Interesting.

If that is true, would not that mean that this didn't work even before this patch (since `-B` and `-S` were both added in cmake v3.13)?

---

> I looked at the list of checks, but I don't see where Zuul CI is. 

They are here: <https://curl.zuul.vexxhost.dev/builds>

> Does it have GitHub Actions integration?

Afaik, it doesn't show up most of the times.

> file
> https://curl.zuul.vexxhost.dev/status

It looks like the job for this PR is still in queue.

friendlyanon:
Relying on undocumented features is not a good thing, no matter how you look at it.

> So if you don't try to change minimal version to 3.13, I see no reason to introduce this change, IMO.

To quote myself from https://github.com/curl/curl/issues/9008#issuecomment-1156641393:

> This is not a change that is concerned with the project code at all. This is purely about how `cmake` is invoked on the command line or in scripts.

emanuele6:
I used the manual `mkdir`/`cd`/`cmake` sequence for the ubuntu jobs.

@bagder I drafted this PR mainly because I was waiting for the Zuul CI to start working again to make sure I didn't break something.

snikulov (approved):
LGTM.

emanuele6:
"intruced" -> "introduced"

bagder:
Thanks!

PR #9039, opened 2022-06-22
Avoid supplying 0 length `msg_control` to sendmsg()

icing:
* testing on macOS 12.4, sendmsg() often*) fails with EINVAL when
  a msg_control buffer is provided in sendmsg(), even though
  msg_controllen was set to 0.
  Supplying the buffer only when GSO is actually used, resolves the issue.

*) Note: this all is a bit fishy. Failures seem to happen on local ipv4 connections
regularly, while external requests over ipv6 seem to work fine. 

Then again, the code change should do no harm, in my understanding.

### DIFF
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -1653,16 +1653,12 @@ static CURLcode do_sendmsg(size_t *psent, struct Curl_easy *data, int sockfd,
   struct iovec msg_iov = {(void *)pkt, pktlen};
   struct msghdr msg = {0};
   uint8_t msg_ctrl[32];
-  size_t ctrllen = 0;
   ssize_t sent;
 #if defined(__linux__) && defined(UDP_SEGMENT)
   struct cmsghdr *cm;
 #endif
 
   *psent = 0;
-
-  assert(sizeof(msg_ctrl) >= CMSG_SPACE(sizeof(uint16_t)));
-
   msg.msg_iov = &msg_iov;
   msg.msg_iovlen = 1;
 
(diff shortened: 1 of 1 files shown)


tatsuhiro-t (approved):

lib/vquic/ngtcp2.c
```diff
@@ -1666,21 +1666,23 @@ static CURLcode do_sendmsg(size_t *psent, struct Curl_easy *data, int sockfd,
...
-  msg.msg_controllen = sizeof(msg_ctrl);
-
 #if defined(__linux__) && defined(UDP_SEGMENT)
   if(pktlen > gsolen) {
+    /* Only set this, when we need it. macOS, for example,
+     * does not seem to like a msg_control of length 0. */
+    msg.msg_control = msg_ctrl;
+    msg.msg_controllen = sizeof(msg_ctrl);
+
     ctrllen += CMSG_SPACE(sizeof(uint16_t));
```
jay, line 1676: I'm not familiar with how these macros work but I wonder is separate `ctrllen` still needed now, for example would this be correct `msg.msg_controllen = sizeof(msg_ctrl) + CMSG_SPACE(sizeof(uint16_t));` ?

jay: hm if not, how about this:
msg.msg_controllen = sizeof(msg_ctrl);
msg.msg_controllen += CMSG_SPACE(sizeof(uint16_t));

tatsuhiro-t: Because we use UDP_SEGMENT only, just assigning `CMSG_SPACE(sizeof(uint16_t))` to `msg.msg_controllen` should work:

```c
    msg.msg_control = msg_ctrl;
    msg.msg_controllen = CMSG_SPACE(sizeof(uint16_t));
```

icing: Just added that change and cleaned up the local var and assertion.

Issue #8737, opened 2022-04-22
Issues with ftps through CONNECT


MarvinWinks:
Using curl executable, directory listings against several ftps sites stopped working as of version 7.77 of curl. I tried compiling and testing against each of the versions from 7.76 to 7.82.

On versions 7.77 and after, the TLS handshake completes successfully. But the directory listing does not return.

I also tried writing a simple main.cpp using libcurl. I am finding that starting with v7.77, CURLOPT_WRITEFUNCTION does not get called.

### I did this

I ran the curl executable to do a directory listing of an ftps site. Command-line options used:

  -vv
  -x <proxy>
  -u <username:password>
  --ftp-ssl
  --proxytunnel
  <ftp site with scheme as ftp://>

### I expected the following

The directory listing as output.

### curl/libcurl version

I tried several versions:
*Newest version that works: 7.76.1*
curl 7.76.1 (x86_64-unknown-linux-gnu) libcurl/7.76.1 OpenSSL/1.0.2zd zlib/1.2.11 libssh2/1.9.0
Release-Date: 2021-04-14
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS GSS-API HTTPS-proxy IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL TLS-SRP UnixSockets

*Newest version tried (that does not work): 7.82.0*
curl 7.82.0 (x86_64-unknown-linux-gnu) libcurl/7.82.0 OpenSSL/1.0.2zd zlib/1.2.11 libssh2/1.9.0
Release-Date: 2022-03-05
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS GSS-API HSTS HTTPS-proxy IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL TLS-SRP UnixSockets

### operating system

Linux <hostname redacted> 3.10.0-957.35.2.el7.x86_64 #1 SMP Wed Sep 18 05:51:28 EDT 2019 x86_64 x86_64 x86_64 GNU/Linux

jay:
Is there a public server that we can use to reproduce?

MarvinWinks:
Hi Jay,

[...]

PR #9043, opened 2022-06-23
ftp: restore protocol state after http proxy connect

tblah:
`connect_init()` (`lib/http_proxy.c`) swaps out the protocol state while working on the proxy connection, this is then restored by Curl_connect_done() after the connection completes.

`ftp_do_more()` extracted the protocol state pointer to a local variable at the start of the function then calls `Curl_proxy_connect()`. If the proxy connection completes, `Curl_proxy_connect()` will call `Curl_connect_done()` (via `Curl_proxyCONNECT()`), which restores `data->req.p` to point to the ftp protocol state instead of the http proxy protocol state, but the local variable in `ftp_do_more()` still pointed to the old value.

Ultimately this meant that the state worked on by `ftp_do_more()` was the http proxy state not the ftp state initialized by `ftp_connect()`, but subsequent calls to any ftp_ function would use the original state.

For my use-case, the visible consequence was that `ftp->downloadsize` was never set and so downloaded data was never returned to the application.

This commit updates the ftp protocol state pointer in `ftp_do_more()` after `Curl_proxy_connect()` returns, ensuring that the correct state pointer is used.

Fixes https://github.com/curl/curl/issues/8737

I was not able to reproduce #8737 in a curl unit test, but it can be reproduced by trying to list files on an ftp server accessed via an http proxy. I can provide dockerfiles for an ftp server and an http proxy if anyone is interested.

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -3565,8 +3565,10 @@ static CURLcode ftp_do_more(struct Curl_easy *data, int *completep)
   bool connected = FALSE;
   bool complete = FALSE;
 
-  /* the ftp struct is inited in ftp_connect() */
-  struct FTP *ftp = data->req.p.ftp;
+  /* the ftp struct is inited in ftp_connect(). If we are connecting to an HTTP
+   * proxy then the state will not be valid until after that connection is
+   * complete */
+  struct FTP *ftp = NULL;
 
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #9044, opened 2022-06-24
tidy-up: delete unused build configuration settings

vszakats:
Delete unused macros (most of them feature guards):

- `CURL_INCLUDES_SYS_UIO` [1]
- `HAVE_ALLOCA_H` [2]
- `HAVE_CRYPTO_CLEANUP_ALL_EX_DATA` (unused since de71e68000c8624ea13f90b136f8734dd0fb1bdc)
- `HAVE_DLFCN_H`
- `HAVE_DLOPEN`
- `HAVE_DOPRNT`
- `HAVE_FCNTL`
- `HAVE_GETHOSTBYNAME` [3]
- `HAVE_GETOPT_H`
- `HAVE_GETPASS`
- `HAVE_GETPROTOBYNAME`
- `HAVE_GETSERVBYNAME`
- `HAVE_IDN_FREE*`
- `HAVE_INET_ADDR`
- `HAVE_IOCTL`
- `HAVE_KRB4`
- `HAVE_KRB_GET_OUR_IP_FOR_REALM`
- `HAVE_KRB_H`
- `HAVE_LDAPSSL_H`
- `HAVE_LDAP_INIT_FD`
- `HAVE_LIBDL`
- `HAVE_LIBNSL`
- `HAVE_LIBRESOLV*`
- `HAVE_LIBUCB`
- `HAVE_LL`
- `HAVE_LOCALTIME_R`
- `HAVE_MALLOC_H`
- `HAVE_MEMCPY`
- `HAVE_MEMORY_H`
- `HAVE_NETINET_IF_ETHER_H`
- `HAVE_NI_WITHSCOPEID`
- `HAVE_OPENSSL_CRYPTO_H`
- `HAVE_OPENSSL_ERR_H`
- `HAVE_OPENSSL_PEM_H`
- `HAVE_OPENSSL_PKCS12_H`
- `HAVE_OPENSSL_RAND_H`
- `HAVE_OPENSSL_RSA_H`
- `HAVE_OPENSSL_SSL_H`
- `HAVE_OPENSSL_X509_H`
- `HAVE_PEM_H`
- `HAVE_POLL`
- `HAVE_RAND_SCREEN`
- `HAVE_RAND_STATUS`
- `HAVE_RECVFROM`
- `HAVE_SETSOCKOPT`
- `HAVE_SETVBUF`
- `HAVE_SIZEOF_LONG_DOUBLE`
- `HAVE_SOCKIO_H`
- `HAVE_SOCK_OPTS`
- `HAVE_STDIO_H`
- `HAVE_STRCASESTR`
- `HAVE_STRFTIME`
- `HAVE_STRLCAT`
- `HAVE_STRNCMPI`
- `HAVE_STRNICMP`
- `HAVE_STRSTR`
- `HAVE_STRUCT_IN6_ADDR`
- `HAVE_TLD_H`
- `HAVE_TLD_STRERROR`
- `HAVE_UNAME`
- `HAVE_USLEEP`
- `HAVE_WINBER_H`
- `HAVE_WRITEV`
- `HAVE_X509_H`
- `LT_OBJDIR`
- `NEED_BASENAME_PROTO`
- `NOT_NEED_LIBNSL`
- `OPENSSL_NO_KRB5`
- `RECVFROM_TYPE*`
- `SIZEOF_LONG_DOUBLE`
- `STRERROR_R_TYPE_ARG3`
- `USE_YASSLEMUL`
- `_USRDLL` (from CMake) [4]

[1] Maybe some related parts in `m4/curl-functions.m4` and
    `configure.ac` could also be deleted?

[2] Maybe the related comment can also be deleted in
    `packages/vms/generate_config_vms_h_curl.com`

[3] There are more instances of this in autotools, but I did not dare to
    touch those. Looked like it is used to detect socket support.

[4] This is necessary for MFC (Microsoft Foundation Class) DLLs to
    force linking MFC components statically to the DLL. `libcurl.dll`
    does not use MFC, so this define can be deleted.
    Ref: https://docs.microsoft.com/cpp/build/regular-dlls-statically-linked-to-mfc

Closes #xxxx


bagder:
How did you find these? Can we create a script or test or something for this?

vszakats:
@bagder It was a byproduct of an effort in syncing autotools-cmake-Makefile.m32 builds to produce the exact same binaries [^1]. I dumped command-line `-D` + `curl-config.h` / `config-win32.h` options for all three into a sorted list and started comparing. Then I `git grep` for those that were only used by _some_ of the build tools.

Some unused settings are still being generated by `autotools`, e.g. most `HAVE_OPENSSL_*_H` settings, as the strings seem generated and not visible as-is, by `grep`.

It's probably possible to automate this to some extent: The source for most of these settings is `lib/curl_config.h.cmake`, `lib/curl_config.h.in` and `lib/config-*.h`. So parsing them and then looking the results up in the source tree can work.

[^1]: This was a [success](https://github.com/curl/curl-for-win/commit/99f89b27a9cc440e6373ad2a9cb71217e91f20f7) for the generated `.o` files which are now all identical. Besides the `OS` string, which is different for each build tool, so I synced those with patches for these tests.

vszakats:
Here's a rough attempt to automate it:
```shell
#!/bin/sh

autoheader configure.ac  # generate lib/curl_config.h.in

{
  grep -o -E -h '#define +[A-Z][A-Z0-9_]{3,}' lib/config-*.h | sed -E 's|#define +||g'
  grep -o -E '#cmakedefine +[A-Z][A-Z0-9_]{3,}' lib/curl_config.h.cmake | sed -E 's|#cmakedefine +||g'
  grep -o -E '#undef +[A-Z][A-Z0-9_]{3,}' lib/curl_config.h.in | sed -E 's|#undef +||g'
} | sort -u | while read -r def; do
  c="$(git grep "${def}" | grep -v -E -c '(/libcurl.tmpl|^lib/config-|^lib/curl_config.h.cmake|^CMakeLists.txt|^CMake/Platforms/WindowsCache.cmake)')"
  if [ "${c}" = '0' ]; then
    echo "unused: ${def}"
  fi
done
```

jay:
I think the HAVE_ ones at least should be ok. OPENSSL_NO_KRB5 is only defined by config-dos but I'm not sure why as I'd expect it to be defined by opensslconf.h.

bagder:
> OPENSSL_NO_KRB5 is only defined by config-dos but I'm not sure why

Yeah, that feels like an overstep. We should leave openssl to define those things.

vszakats:
Rebased onto master.

vszakats:
@bagder: At this point this is committable. The few FIXMEs are rather questions, that's beyond my insight to touch. Do you think I shall continue with this?

bagder (approved):

PR #9045, opened 2022-06-24
Extra libs in auxiliary path are ignored

yurashek:
Hello Dan,
It is a nit only but it is common the path to auxiliary libraries should prepend the standard paths.
Could it be possible to fix it?
Thanks,
Yura

### DIFF
--- a/m4/curl-openssl.m4
+++ b/m4/curl-openssl.m4
@@ -138,8 +138,8 @@ if test "x$OPT_OPENSSL" != xno; then
   fi
 
   dnl finally, set flags to use SSL
-  CPPFLAGS="$CPPFLAGS $SSL_CPPFLAGS"
-  LDFLAGS="$LDFLAGS $SSL_LDFLAGS"
+  CPPFLAGS="$SSL_CPPFLAGS $CPPFLAGS"
+  LDFLAGS="$SSL_LDFLAGS $LDFLAGS"
 
   AC_CHECK_LIB(crypto, HMAC_Update,[
      HAVECRYPTO="yes"


jay:
How is this common? What are you trying to solve?

yurashek:
> How is this common? What are you trying to solve?

It is because linker will search -L paths in the order how the paths are listed so if you need to introduce new added API (i.e. openssl3) which is placed on the other than standard path (/usr/lib:/lib) then you need to list the newly added API first than the original API (ssl(1) in this case placed on standard path) will be found by the linker. Otherwise the API on standard path is used.
No problem. No one knows all details about linking  ...except of Ali Bahrami and his mentors :-)

jay:
The order of paths specified by -L is not guaranteed so if you have OpenSSL in multiple paths then you would need to resolve that. Normally system default paths would not be specified by -L for example you normally would not see -L/usr/lib

yurashek:
> ~The order of paths specified by -L is not guaranteed~ (or, rather, the order that we add library paths is not guaranteed) so if you have OpenSSL in multiple paths then you would need to resolve that.

Order of searching the -L paths *is* specified by the order of -L options of linker. What is uncertain is how these options are assembled by autoconf (my fix do not concern of cmake build) so this is the reason why I am fixing m4 macro directly.

> Normally system default paths would not be specified by -L for example you normally would not see -L/usr/lib

Sure but this is because the linker is searching /usr/lib:/lib paths internally to ensure these paths will be searched as the last instance. Unfortunately some FOSS components emits these default paths on pkg-config --libs what is wrong but happens. This is exactly the reason to put an "extra" options to linker before the "standard" options.

FYI the SSL_LDFLAGS here are fed by pkg-config:

 $ PKG_CONFIG_PATH=/usr/openssl/3/lib/64/pkgconfig  pkg-config --libs libssl
-L/usr/openssl/3/lib/sparcv9 -lssl

jay:
> Unfortunately some FOSS components emits these default paths on pkg-config --libs what is wrong but happens. This is exactly the reason to put an "extra" options to linker before the "standard" options.

That is not a bug in curl though. I don't see a reason to change the order just to fix your case. Also, it just happens to work. The order is just not guaranteed so again my position is if you have multiple paths with OpenSSL included by -L then you need to address that.

yurashek:
> That is not a bug in curl though. I don't see a reason to change the order just to fix your case. Also, it just happens to work. The order is just not guaranteed so again my position is if you have multiple paths with OpenSSL included by -L then you need to address that.

This is logical bug in position of flags in m4 macros for curl. As I described on the same beginning auxiliary options should ever prepend the standard options.
If the additional info looks to be too complicated you can ignore it. It can not change this fact.

Standard way to address selected library is to use selected pkg-config file. This bug in curl breaks the standard way.

If there are linkers which can not assure the parameters (of the same category) will be interpreted in the order as they are listed it can not be affected by the change proposed by me but behavior of the deterministic linkers will be correct after this change. You argument looks to be irrational to me but even though you might be right. Can you explain your rational arguments?
Thank you,
Yura

jay:
Closing since this is not a curl bug. You may be able to workaround the issue by supplying CFLAGS and LDFLAGS when invoking configure, for example if it's installed to prefix /usr/local/ssl and the libdir is named lib:
~~~
CFLAGS="-I/usr/local/ssl/include" \
LDFLAGS="-L/usr/local/ssl/lib -Wl,-rpath,/usr/local/ssl/lib" \
./configure --with-ssl=/usr/local/ssl
~~~

PR #9046, opened 2022-06-24
cmake: do not force bump build target to Vista

vszakats:
CMake has a setting `ENABLE_INET_PTON` defaulting to `ON`, which
does nothing else than fixing the Windows build target to Vista.
This was done even when the toolchain did not have Vista support
(e.g. original MinGW), breaking such builds.

On other systems it did not make a user-facing difference, because
libcurl has its own pton() implementation, so it works equally well
with or without Vista's inet_pton().

This patch drops the setting `ENABLE_INET_PTON`. inet_pton() is now
used whenever building for Vista or newer, either when requested
manually via
  `CURL_TARGET_WINDOWS_VERSION=0x600`,
or by default with modern toolchains (e.g. mingw-w64). Older envs
will fall back to curl's pton().

Ref: https://github.com/curl/curl/pull/9027#issuecomment-1164157604
Ref: https://github.com/curl/curl/pull/8997#issuecomment-1164344155

Closes #xxxx


vszakats:
This PR deprecates `CURL_TARGET_WINDOWS_VERSION`. It was introduced in January 2021. It's a CMake and curl-specific setting for something that seems relatively straightforward to specify in a universal way by passing `-D_WIN32_WINNT=`. As a CMake-noob, it is possible I'm missing certain merits/aspects of the dedicated-setting approach. It's also not a required change to reach the goals of the PR (in its current form.)

I'm happy to discuss what to do with this option. Keep it as-is, leave it deprecated, delete it?

MarcelRaad:
I'm not very familiar with this aspect of CMake, but several other projects I use provide a way to add compiler or preprocessor options without overriding the defaults via `-DCMAKE_C(XX)_FLAGS`. IIUC, that's required for respecting what's set via environment variables and toolchain files.

vszakats:
@MarcelRaad Thanks for your feedback. If this brings things closer to the CMake-way, we can certainly keep it.

I haven't tried it, but CMake says to also support [`CFLAGS` envvar](https://cmake.org/cmake/help/v3.24/envvar/CFLAGS.html), so `CFLAGS=-D_WIN32_WINNT=...` might also work.

jzakrzewski:
Projects tend to provide dedicated options for aspects that they deem "frequently used" or "overly complicated to setup otherwise". So the question would be if this setting falls into one of those categories, I guess. I for sure don't know ;)

vszakats:
Thanks for the inputs. I deleted the deprecation part. It's nice to make passing this setting a little bit friendlier.

MarcelRaad (approved):

jay (approved):
I would save this for the next feature window which is in [several days](https://calendar.google.com/calendar/u/0/embed?src=c9u5d64odop9js55oltfarjk6g@group.calendar.google.com&ctz=Europe/Stockholm) barring any issues worthy of a patch release

PR #9048, opened 2022-06-25
Apply cookie limits

bagder:


bagder:
Hm, hyper has its own (very) strict limits on the size of the received HTTP response headers so test 444 must be made much smaller to work there as well...

PR #9056, opened 2022-06-27
Fix build

vvb2060:
```
curl/src/main/native/curl/lib/easy_lock.h:54:7: error: use of undeclared identifier 'asm'
        asm volatile("yield" ::: "memory");
        ^
  1 error generated.
```
when build with `-std` options

---

```
curl/src/main/native/curl/lib/vquic/ngtcp2.c:1081:3: error: incompatible function pointer types initializing 'nghttp3_acked_stream_data' (aka 'int (*)(struct nghttp3_conn *, long long, unsigned long long, void *, void *)') with an expression of type 'int (nghttp3_conn *, int64_t, size_t, void *, void *)' (aka 'int (struct nghttp3_conn *, long long, unsigned int, void *, void *)') [-Werror,-Wincompatible-function-pointer-types]
C/C++:   cb_h3_acked_stream_data, /* acked_stream_data */
C/C++:   ^~~~~~~~~~~~~~~~~~~~~~~
C/C++: 1 error generated.
```
when build with `-Werror` options on armeabi-v7a
https://github.com/ngtcp2/nghttp3/commit/68fbf0576dfc705cfda3e89007713a8141b5f0fc

### DIFF
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -113,7 +113,7 @@ static CURLcode ng_process_ingress(struct Curl_easy *data,
 static CURLcode ng_flush_egress(struct Curl_easy *data, int sockfd,
                                 struct quicsocket *qs);
 static int cb_h3_acked_stream_data(nghttp3_conn *conn, int64_t stream_id,
-                                   size_t datalen, void *user_data,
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #9057, opened 2022-06-27
libcurl-security.3: Fix typo on macro "SH_"

samueloph:
During the packaging of the latest curl release for Debian, Lintian warned me about a typo which causes the section name "Secrets in memory" to not be rendered in the manpage due to "SH_" not being recognized as a header.

Thanks,


bagder:
Thanks!

PR #9061, opened 2022-06-28
easy_lock: switch to using atomic_int instead of bool

### LINKED ISSUE #9055: Code using atomics should link with -latomic
Building curl-7.84.0 for 64-bit RISC-V Linux using GCC 12.1.0 fails for me:

```
  CCLD     curl
ld: ../lib/.libs/libcurl.so: undefined reference to `__atomic_exchange_1'
```

This is because GCC 12 doesn't inline sub-word-sized atomic operations ([although future versions may do](https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg283119.html)), and the new code in `easy_lock.h` does an atomic exchange on a single-byte `atomic_bool`.

The proper fix would be to link with GCC's `libatomic` - maybe `CURL_ATOMIC` should check for this? Changing the type of `curl_simple_lock` to `atomic_int` also works as a workaround.

bagder:
To work with more compilers without requiring separate libs to
link. Like with gcc-12 for RISC-V on Linux.

Reported-by: Adam Sampson
Fixes #9055

### DIFF
--- a/lib/easy_lock.h
+++ b/lib/easy_lock.h
@@ -40,8 +40,8 @@
 #include <sched.h>
 #endif
 
-#define curl_simple_lock atomic_bool
-#define CURL_SIMPLE_LOCK_INIT false
+#define curl_simple_lock atomic_int
+#define CURL_SIMPLE_LOCK_INIT 0
 
 static inline void curl_simple_lock_lock(curl_simple_lock *lock)
 {


bagder:
@atsampson like this?

atsampson:
Yep, that's the same as my local workaround, which seems to work - I didn't change the initialisation value but that shouldn't make any difference.

bagder:
Thanks @atsampson!

jay:
Yes it's ok to cherry pick it. There have been only build fixes. AFAICS that upstream commit has one test failing in appveyor cmake builds, test 1014 which is because curl-config generated by cmake is missing the threadsafe feature. That is being worked on in #8982.

AlexGhiti:
Thanks for your answer, it fixed the issue for me so it should land soon in Ubuntu :)

Thanks again,

Alex

AlexGhiti:
Hmmm actually, it keeps failing on armhf on the test 574. You can see the build log here https://launchpadlibrarian.net/614228570/buildlog_ubuntu-kinetic-armhf.curl_7.84.0-2ubuntu1~ppa0_BUILDING.txt.gz. I took a look at `tests/libtest/lib574.c` which does not seem to do much to me. Would you have any idea what could be wrong?

jay:
It "keeps failing" as in it is 100% reproducible? From that log it looks like it failed because it hangs in the CI but I don't know why that is. If you change `make test` to run just that test and print the log output, does it show anything? `make TFLAGS='-v -p 574' test`

AlexGhiti:
Sorry for the late answer, I did not manage to reproduce it locally, and it finally passed after 2 fails in a row: either the infrastructure was flaky or the issue was fixed in the meantime by another package...I don't know.

So no problem actually, thanks again for your quick answer :)

PR #9065, opened 2022-06-28
cmake: support ngtcp2 boringssl backend

donny-dont:
Update the ngtcp2 find module to detect the boringssl backend. Determine if the underlying OpenSSL implementation is BoringSSL and if so use that as the ngtcp2 backend.


donny-dont:
Makes it so https://github.com/curl/curl/pull/8789 will build in CMake. I'm guessing maybe the autotools one already works with the ngtcp2 + BoringSSL backend. If not then something like this would also need to be done.

donny-dont:
Hey @bagder anything in particular you want addressed here? I was able to build a version of curl with HTTP/3 support and run it in on the curl network backend in WebKit.

bagder:
I was hoping someone who knows cmake would chime in...

jzakrzewski:
I don't have time to build & test, but otherwise it looks clean to me.

bagder:
Thanks!

donny-dont:
Thanks @jzakrzewski and @bagder !

Issue #8653, opened 2022-03-30
breaking change in 7.82.0: mandatory login token in .netrc


standsed:
### I did this
Create/append ~/.netrc file with record without login token, e.g.,
`machine curl.com password S3cr3t`

run
`curl https://host.com --netrc --verbose`
### I expected the following
Header to be sent
Authorization: Basic Onp4Y3Y=
### curl/libcurl version
Does not work: 
docker image: curlimages/curl:7.82.0
curl 7.82.0-DEV (x86_64-pc-linux-musl) libcurl/7.82.0-DEV OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.9 libssh2/1.10.0 nghttp2/1.46.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets

Works:
docker image: curlimages/curl:7.81.0
curl 7.81.0-DEV (x86_64-pc-linux-musl) libcurl/7.81.0-DEV OpenSSL/1.1.1l zlib/1.2.11 brotli/1.0.9 libssh2/1.10.0 nghttp2/1.46.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets

curlimages/curl:7.77.0 also ok

jay:
Bisected to 7d600ad (#8449) which removed the user_passwd flag in favor of checking user pointer. AFAICT that means now that a username must always be set if a password is set. There is code in parseurlandfillconn which sets an empty username if no username was found in the URL but a password was found:

https://github.com/curl/curl/blob/64db5c575d9c5536bd273a890f50777ad1ca7c13/lib/url.c#L2066-L2069

netrc user/pass comes in later in override_login, but it doesn't have the same logic to create a blank username if there's a password but no username. I'm not sure the reasons for that but here is some similar logic that works:

~~~diff
diff --git a/lib/url.c b/lib/url.c
[...]

PR #9066, opened 2022-06-29
url: treat missing usernames in netrc as empty

jay:
- If, after parsing netrc, there is a password with no username then
  set a blank username.

This used to be the case prior to 7d600ad (precedes 7.82). Note
parseurlandfillconn already does the same thing for URLs.

Reported-by: Raivis <standsed@users.noreply.github.com>
Testing-by: Domen Kožar

Fixes https://github.com/curl/curl/issues/8653
Closes #xxxx

---

Edit: This could use a test so it doesn't happen again, so consider it a WIP for now.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -3041,6 +3041,12 @@ static CURLcode override_login(struct Curl_easy *data,
       /* don't update the user name below */
       userp = NULL;
     }
+    /* no user was set but a password, set a blank user */
+    if(userp && !*userp && *passwdp) {
+      *userp = strdup("");
+      if(!*userp)
+        return CURLE_OUT_OF_MEMORY;
+    }
   }
 #endif
 


bagder (approved):

orgads:
I've noticed another regression. I *do* have username in netrc, but if I provide a username in the URL, none of them is used, and no Authorization header is sent (even if it is the same user).

Example:
.netrc:
```
machine server login user password 5up3r53cr31
```

```bash
$ curl -n http://user@server/
Unauthorized
```

Is this covered by this PR too?

emilengler:
> Is this covered by this PR too?

No.

orgads:
This works when user is not provided, but if you provide a username, it is used without the password. Not sure what's the expected behavior on this case though.

.netrc
```
machine curl.com password 123456
```
```bash
$ curl -vkn http://curl.com/ 2>&1 | grep Auth
> Authorization: Basic OjEyMzQ1Ng==
# This is :123456

$ curl -vkn http://aaa@curl.com/ 2>&1 | grep Auth
> Authorization: Basic YWFhOg==
# This is aaa:
```

orgads:
Fixed the case I described here in #9248, and added tests for all the netrc regression fixes in #9256.

bagder:
Made a draft until we get a test case for this. Does #9256 verify this PR?

orgads:
> Made a draft until we get a test case for this. Does #9256 verify this PR?

It does

orgads:
@badger Do you want me to create a new PR that includes all the netrc fixes with the additional tests?

PR #977, opened 2016-08-23
Add libproxy support

dwmw2:
This adds optional support for using libproxy to determine which proxy to use for a given URL.

The original libproxy has evolved into a C++ monstrosity which potentially loads a JavaScript interpreter into the address space of anything which uses it... but the API is clean and simple:
https://github.com/libproxy/libproxy/blob/master/libproxy/proxy.h

The PacRunner dæmon provides a service where you can simply ask it, over D-Bus, which proxy to use for a given URL. Not only does it have a plugin for the original libproxy, but it also provides its own replacement libproxy implementation which does _nothing_ but field the call to PacRunner.

I am mentoring a Google Summer of Code project this year which is fixing NetworkManager to pick up proxy information from DHCP, VPN servers and other sources, and to allow it to be an explicit part of per-network configuration — and to poke it into PacRunner appropriately.

Thus the expectation will be that asking libproxy should _consistently_ give correct answers. Once that's true, I hope to change Linux distribution packaging guidelines to suggest that packaged applications _should_ be using libproxy's results (or at least PacRunner's results) by default.

This implements libproxy support for curl because it seems like the more generic option, rather than talking to PacRunner directly over D-Bus. This is just the basic functionality; we need to look at how we make it accessible from the curl command line tool, and/or enable it by default.

PR #9068, opened 2022-06-29
Draft: Add libproxy support

janbrummer:
Updated MR initially made by dwmw2 (#977)

As our company uses a pac proxy with dozens of entries such a support is necessary within curl. Let's revive this old (but updated) PR and get back to talking.

@dwmw2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,6 +226,13 @@ mark_as_advanced(CURL_DISABLE_TFTP)
 option(CURL_DISABLE_VERBOSE_STRINGS "disables verbose strings" OFF)
 mark_as_advanced(CURL_DISABLE_VERBOSE_STRINGS)
 
+option(ENABLE_LIBPROXY "Define if you want to enable libproxy support" ON)
+if(ENABLE_LIBPROXY)
+  find_package(libproxy REQUIRED)
+  set(ENABLE_LIBPROXY ON)
+  list(APPEND CURL_LIBS ${LIBPROXY_LIBRARIES})
+endif()
+
 # Corresponds to HTTP_ONLY in lib/curl_setup.h
 option(HTTP_ONLY "disables all protocols except HTTP (This overrides all CURL_DISABLE_* options)" OFF)
 mark_as_advanced(HTTP_ONLY)
@@ -1467,6 +1474,7 @@ _add_if("HTTP3"         USE_NGTCP2 OR USE_QUICHE)
 _add_if("MultiSSL"      CURL_WITH_MULTI_SSL)
 _add_if("HTTPS-proxy"   SSL_ENABLED AND (USE_OPENSSL OR USE_GNUTLS OR USE_NSS))
 _add_if("unicode"       ENABLE_UNICODE)
+_add_if("libproxy"      ENABLE_LIBPROXY)
 string(REPLACE ";" " " SUPPORT_FEATURES "${_items}")
 message(STATUS "Enabled features: ${SUPPORT_FEATURES}")
 
--- a/configure.ac
+++ b/configure.ac
@@ -23,10 +23,10 @@
 #***************************************************************************
 dnl Process this file with autoconf to produce a configure script.
 
-AC_PREREQ(2.59)
+AC_PREREQ([2.71])
 
 dnl We don't know the version number "statically" so we use a dash here
-AC_INIT([curl], [-], [a suitable curl mailing list: https://curl.se/mail/])
+AC_INIT([curl],[-],[a suitable curl mailing list: https://curl.se/mail/])
 
 XC_OVR_ZZ50
 XC_OVR_ZZ60
@@ -1369,6 +1369,90 @@ if test X"$OPT_BROTLI" != Xno; then
     fi
   else
     dnl no brotli, revert back to clean variables
+    LDFLAGS=$CLEANLDFLAGS
+    CPPFLAGS=$CLEANCPPFLAGS
+    LIBS=$CLEANLIBS
+  fi
+fi
+
+dnl **********************************************************************
+dnl Check for the presence of LIBPROXY libraries and headers
+dnl **********************************************************************
+
+dnl Default to compiler & linker defaults for LIBPROXY files & libraries.
+OPT_LIBPROXY=off
+AC_ARG_WITH(libproxy,dnl
+AS_HELP_STRING([--with-libproxy=PATH],[Where to look for libproxy, PATH points to the LIBPROXY installation; when possible, set the PKG_CONFIG_PATH environment variable instead of using this option])
+AS_HELP_STRING([--without-libproxy],[disable LIBPROXY]),
+  OPT_LIBPROXY=$withval)
+
+if test X"$OPT_LIBPROXY" != Xno; then
+  dnl backup the pre-libproxy variables
+  CLEANLDFLAGS="$LDFLAGS"
+  CLEANCPPFLAGS="$CPPFLAGS"
+  CLEANLIBS="$LIBS"
+
+  case "$OPT_LIBPROXY" in
+  yes)
+    dnl --with-libproxy (without path) used
+    CURL_CHECK_PKGCONFIG(libproxy)
+
+    if test "$PKGCONFIG" != "no" ; then
+      LIB_PROXY=`$PKGCONFIG --libs-only-l libproxy`
+      LD_PROXY=`$PKGCONFIG --libs-only-L libproxy`
+      CPP_PROXY=`$PKGCONFIG --cflags-only-I libproxy`
+      version=`$PKGCONFIG --modversion libproxy`
+      DIR_PROXY=`echo $LD_PROXY | $SED -e 's/-L//'`
+    fi
+
+    ;;
+  off)
+    dnl no --with-libproxy option given, just check default places
+    ;;
+  *)
+    dnl use the given --with-libproxy spot
+    PREFIX_PROXY=$OPT_LIBPROXY
+    ;;
+  esac
+
+  dnl if given with a prefix, we set -L and -I based on that
+  if test -n "$PREFIX_PROXY"; then
+    LIB_PROXY="-lproxy"
+    LD_PROXY=-L${PREFIX_PROXY}/lib$libsuff
+    CPP_PROXY=-I${PREFIX_PROXY}/include
+    DIR_PROXY=${PREFIX_PROXY}/lib$libsuff
+  fi
+
+  LDFLAGS="$LDFLAGS $LD_PROXY"
+  CPPFLAGS="$CPPFLAGS $CPP_PROXY"
+  LIBS="$LIB_PROXY $LIBS"
+
+  AC_CHECK_LIB(proxy, px_proxy_factory_new)
+
+  AC_CHECK_HEADERS(proxy.h,
+    curl_libproxy_msg="enabled"
+    LIBPROXY_ENABLED=1
+  )
+
+  if test X"$OPT_LIBPROXY" != Xoff &&
+     test "$LIBPROXY_ENABLED" != "1"; then
+    AC_MSG_ERROR([libproxy libs and/or directories were not found where specified!])
+  fi
+
+  if test "$LIBPROXY_ENABLED" = "1"; then
+    if test -n "$DIR_LIBPROXY"; then
+       dnl when the libproxy shared libs were found in a path that the run-time
+       dnl linker doesn't search through, we need to add it to LD_LIBRARY_PATH
+       dnl to prevent further configure tests to fail due to this
+
+       if test "x$cross_compiling" != "xyes"; then
+         LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$DIR_LIBPROXY"
+         export LD_LIBRARY_PATH
+         AC_MSG_NOTICE([Added $DIR_LIBPROXY to LD_LIBRARY_PATH])
+       fi
+    fi
+  else
+    dnl no libproxy; revert back to clean variables
     LDFLAGS=$CLEANLDFLAGS
     CPPFLAGS=$CLEANCPPFLAGS
     LIBS=$CLEANLIBS
@@ -4479,6 +4563,7 @@ AC_MSG_NOTICE([Configured to build curl/libcurl:
   HTTP2:            ${curl_h2_msg}
   HTTP3:            ${curl_h3_msg}
   ECH:              ${curl_ech_msg}
+  libproxy support: ${curl_libproxy_msg}
   Protocols:        ${SUPPORT_PROTOCOLS}
   Features:         ${SUPPORT_FEATURES}
 ])
--- a/docs/libcurl/curl_easy_setopt.3
+++ b/docs/libcurl/curl_easy_setopt.3
@@ -182,6 +182,9 @@ Proxy port to use. See \fICURLOPT_PROXYPORT(3)\fP
 Proxy type. See \fICURLOPT_PROXYTYPE(3)\fP
 .IP CURLOPT_NOPROXY
 Filter out hosts from proxy use. \fICURLOPT_NOPROXY(3)\fP
+.IP CURLOPT_LIBPROXY
+Use the libproxy API to determine which proxy, if any, to
+use. \fICURLOPT_LIBPROXY(3)\fP
 .IP CURLOPT_HTTPPROXYTUNNEL
 Tunnel through the HTTP proxy. \fICURLOPT_HTTPPROXYTUNNEL(3)\fP
 .IP CURLOPT_CONNECT_TO
--- a/lib/curl_libproxy.c
+++ b/lib/curl_libproxy.c
@@ -1,0 +1,75 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 2011 - 2016, Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at http://curl.haxx.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ ***************************************************************************/
+
+#include "curl_setup.h"
+
+#ifdef ENABLE_LIBPROXY
+
+#include "urldata.h"
+
+#include <proxy.h>
+/* The last 3 #include files should be in this order */
+#include "curl_printf.h"
+#include "curl_memory.h"
+#include "memdebug.h"
+
+static pxProxyFactory *factory;
+
+CURLcode Curl_libproxy_global_init(void)
+{
+  factory = px_proxy_factory_new();
+  if(!factory)
+    return CURLE_OUT_OF_MEMORY;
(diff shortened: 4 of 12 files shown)


MarcelRaad:
What I don't understand is why libcurl has to do that itself. If this is actually for using libcurl from another application, the application could just pass the proxy found by libproxy to libcurl.

If the use case is for the curl tool, a wrapper script passing the proxy to curl might be another solution, but that might of course be quite some effort if that's a common use case.

CMakeLists.txt
```diff
@@ -226,6 +226,13 @@ mark_as_advanced(CURL_DISABLE_TFTP)
 option(CURL_DISABLE_VERBOSE_STRINGS "disables verbose strings" OFF)
 mark_as_advanced(CURL_DISABLE_VERBOSE_STRINGS)
 
+option(ENABLE_LIBPROXY "Define if you want to enable libproxy support" ON)
+if(ENABLE_LIBPROXY)
+  find_package(libproxy REQUIRED)
+  set(ENABLE_LIBPROXY ON)
+  list(APPEND CURL_LIBS ${LIBPROXY_LIBRARIES})
+endif()
+
```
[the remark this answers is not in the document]
janbrummer: Thanks for your feedback, i will updated it.

janbrummer:
> Speaking of Windows, proxy detection can be done with reasonably small amount of code and with minor, if any, build-level work. An example [here](https://github.com/vszakats/hb/blob/main/contrib/hbwin/win_proxyc.c). [ This is "works for me" code and hasn't been vetted with a wide spread of users and systems, though the idea/code is similar to what Chromium did at the time. ]

I'm totally open to change the implementation and use something else. We just have the need to have a working solution in curl.

CMakeLists.txt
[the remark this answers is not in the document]
janbrummer: Actually the default value has been set on purpose. I want  to ensure that newer builds of curl are using it by default. I could change the naming then to  disable libproxy. Not sure how to deal with it here.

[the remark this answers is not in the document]
jzakrzewski: By default if available - this is what all other options are doing (unless they're OFF by default).
I don't want my scripts suddenly fail because of a new optional component.

configure.ac
```diff
@@ -23,10 +23,10 @@
 #***************************************************************************
 dnl Process this file with autoconf to produce a configure script.
 
-AC_PREREQ(2.59)
+AC_PREREQ([2.71])
 
 dnl We don't know the version number "statically" so we use a dash here
-AC_INIT([curl], [-], [a suitable curl mailing list: https://curl.se/mail/])
+AC_INIT([curl],[-],[a suitable curl mailing list: https://curl.se/mail/])
```
bagder, line 29: These who changes seem unrelated. Surely you don't *need* 2.71?

docs/libcurl/curl_easy_setopt.3
```diff
@@ -182,6 +182,9 @@ Proxy port to use. See \fICURLOPT_PROXYPORT(3)\fP
 Proxy type. See \fICURLOPT_PROXYTYPE(3)\fP
 .IP CURLOPT_NOPROXY
 Filter out hosts from proxy use. \fICURLOPT_NOPROXY(3)\fP
+.IP CURLOPT_LIBPROXY
```
bagder, line 185: The option name should describe what it does, what it's for. Not what library it uses. It might end up not using that library at some point.

Also, since this is about setting a proxy, wouldn't it be better if this used the existing proxy options somehow to also avoid the problem with more mutexed options?

lib/curl_libproxy.c
```diff
@@ -0,0 +1,75 @@
...
+
+#include "urldata.h"
+
+#include <proxy.h>
+/* The last 3 #include files should be in this order */
+#include "curl_printf.h"
+#include "curl_memory.h"
+#include "memdebug.h"
+
+static pxProxyFactory *factory;
```
bagder, line 35: I don't see why this needs to use a global variable. I think it shouldn't.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,75 @@
...
+
+      /* We only cope with one; can't fall back on failure */
+      result = libproxy_results[0];
+      for(i=1; libproxy_results[i]; i++)
+        free(libproxy_results[i]);
+      free(libproxy_results);
+    }
+  }
+
+  return result;
```
bagder, line 72: 1. When is this pointer free'd again?
2. libproxy doesn't return the proxy URL using the same schemes as curl supports does it? I mean for SOCKS4/4a/5/5h does it?
3. Can a libproxy proxy be used over unix domain sockets?

docs/libcurl/curl_easy_setopt.3
```diff
@@ -182,6 +182,9 @@ Proxy port to use. See \fICURLOPT_PROXYPORT(3)\fP
...
+Use the libproxy API to determine which proxy, if any, to
+use. \fICURLOPT_LIBPROXY(3)\fP
```
bagder, line 187: This man page is missing.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,75 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 2011 - 2016, Daniel Stenberg, <daniel@haxx.se>, et al.
```
bagder, line 8: We want the copyright ranges to end the year of the most recent change: 2022.

lib/easy.c
```diff
@@ -194,6 +194,10 @@ static CURLcode global_init(long flags, bool memoryfuncs)
   }
 #endif
 
+#ifdef ENABLE_LIBPROXY
+  Curl_libproxy_global_init();
+#endif
```
bagder, line 199: Provide the functions as empty macros if the feature is not built-in, then you don't need `#ifdefs` in the code like this.

lib/easy.c
```diff
@@ -310,6 +314,10 @@ void curl_global_cleanup(void)
   free(leakpointer);
 #endif
 
+#ifdef ENABLE_LIBPROXY
+  Curl_libproxy_global_cleanup();
+#endif
```
bagder, line 319: another macro opportunity.

bagder (commented):
I miss documentation for the new option, and no tests verifying it.

I miss [documentation for libproxy](https://github.com/libproxy/libproxy/issues/52). There also seems to be no tests in the libproxy project nor does it run any CI jobs. It does not give an impression of something to lean on in production.

Any thoughts on how user of curl the command line tool would like to use it?

(I've also set this as 'needs-votes' as I'm not personally convinced this is actually a PR we want merged - I welcome more feedback and opinions.)

CMakeLists.txt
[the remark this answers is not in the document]
mcatanzaro: > By default if available - this is what all other options are doing (unless they're OFF by default).
> I don't want my scripts suddenly fail because of a new optional component.

If you do "by default if available" then it's going to be enabled at random depending on what build dependencies are specified and how eagle-eyed the human packager is to check all the various build options. [Here's the standard "don't do that"](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies) wiki page. Humans are just not smart enough to avoid traps like this. Please pick a default and force people building curl to deviate intentionally.

libproxy should definitely be enabled by default on Linux platforms. Not sure about others.

P.S. If you use meson instead of CMake, it is possible to safely use auto dependencies. But doesn't matter, because you use CMake.

mcatanzaro:
> I miss [documentation for libproxy](https://github.com/libproxy/libproxy/issues/52). There also seems to be no tests in the libproxy project nor does it run any CI jobs. It does not give an impression of something to lean on in production.

libproxy is barely maintained and objectively pretty terrible -- no docs or tests or CI are _far_ from the worst problems here -- but it's also standard on Linux. If your software does not use it, it's probably broken. Otherwise, I'm not sure how you'd know to check GNOME proxy settings, KDE proxy settings, desktop xyz proxy settings, proxy environment variables, how to read and execute web proxy autoconfig JS, etc. Even if you don't do HTTP at all, you still need to check desktop proxy settings because a SOCKS proxy could be configured.

Currently some software has been migrating from libsoup to libcurl, and I've noticed incorrect proxy handling is the most common issue with such migrations. libsoup handles proxies transparently, but libcurl does not. Applications using libcurl currently need to manually proxy proxy settings (bad pun intended) from libproxy to curl proxy options, which is easy to get wrong and annoying. This may be by design, but it's questionable.

There is one better option: depend on GLib and use GProxyResolver, which is smart enough to use libproxy only if it really needs to. (I maintain this code and am slowly reducing its use of libproxy.) But that does you no good unless you're willing to depend on GLib.

CMakeLists.txt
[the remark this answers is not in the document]
mcatanzaro: Ah, I'll note I disagree with the wiki page here:

> In the case of binary-based distributions, like RPM or DEB based ones, automagic dependencies does not change anything: if the user has something installed and is building by hand, it's usually what he wants to enable, while if it's the maintainer, he'll just have to add a dependency over the packages required to run the binaries he has created.

Basically the wiki page suggests this is only bad for Gentoo and not normal Linux distributions. It's not true. If you use automagic dependencies, Fedora/Debian/Ubuntu/yocto/buildroot/whatever packagers probably won't notice, and whether the feature gets enabled or disabled is likely to depend on what BuildRequires are specified. In short, not a good idea. Failing the build unless you pass `-DENABLE_FOO=OFF` is a cheap way to avoid broken downstream packages and hardly difficult to update your build scripts to handle.

lib/curl_libproxy.c
mcatanzaro: I'll let jbrummer answer 1.

Regarding 2, libproxy may return `socks://` (which means: try `socks5://`, then `socks4a://`, then `socks4://`, in that order), or it may return `direct://` (which means "connect directly"). These need to be translated into curl proxy syntax. Then I don't know about `socks5h://` because GLib doesn't know about `socks5h://`... but I guess that should that be attempted first, before `socks5://`?

Regarding 3. libproxy just returns URIs. It would not generally make sense to connect to an HTTP proxy via anything other than TCP, but it's theoretically possible, if you are running a proxy server locally and have taught it to respond via UNIX domain sockets rather than TCP. Same for the other common proxy types, HTTPS and FTP and SOCKS (at least I think SOCKS goes over TCP?). But in theory, if you pass a different URI scheme, it could return any sort of proxy URI.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,75 @@
...
+    char **libproxy_results = px_proxy_factory_get_proxies(factory, url);
+
+    if(libproxy_results) {
+      int i;
+
+      /* We only cope with one; can't fall back on failure */
+      result = libproxy_results[0];
+      for(i=1; libproxy_results[i]; i++)
+        free(libproxy_results[i]);
+      free(libproxy_results);
```
mcatanzaro, line 68: This will crash on Windows, see [this bug](https://github.com/libproxy/libproxy/issues/43). [Example solution](https://gitlab.gnome.org/GNOME/glib-networking/-/commit/8bfbdf3a8b949f0a2264a829c2a6f953482fba2a). Just make sure to require 0.4.16.

lib/url.c
```diff
@@ -2404,8 +2404,17 @@ static char *detect_proxy(struct Curl_easy *data,
       proxy = curl_getenv(envp);
     }
   }
+
   if(proxy)
     infof(data, "Uses proxy env variable %s == '%s'", envp, proxy);
+#ifdef ENABLE_LIBPROXY
+  if(/*data->set.libproxy && */!proxy) {
+    char *libproxy_proxy = Curl_libproxy_detect_proxy(data->state.url);
```
mcatanzaro, line 2412: Looks like `libproxy_proxy` is leaked, right?

Also, stray debug comment above.

lib/curl_libproxy.c
mcatanzaro: > libproxy may return socks:// (which means: try socks5://, then socks4a://, then socks4://, in that order)

BTW this does not mean libproxy might not return more specific URI schemes too. libproxy has basically no documentation, so better assume it will return whatever it wants. ;) But specifically `socks://` and `direct://` are going to require special handling here. In particular, if you do not see `direct://` then you must not connect directly.

CMakeLists.txt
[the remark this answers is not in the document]
jzakrzewski: I'm generally in the "OFF by default" camp for everything, but this is an established pattern in curl and the only thing worse than this pattern, IMHO, is inconsistency.

bagder:
To add to the problems mentioned already, libproxy also contains and uses its own

1. (bad) URL parser
2. (bad) **blocking** HTTP/1.1 retriever function (no TLS support) (including a blocking name resolve with its own DNS cache)

I also find it curious that there's no expire time set for a proxy. With PAC files etc supported, proxies can very well change depending on things like time of day or date but with this API the application doesn't get any clues on when to recheck -  a browser rechecks with the PAC for *every* URL.

CMakeLists.txt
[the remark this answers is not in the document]
bagder: I think the general curl pattern is that we go with features OFF to start with, then over time if we deem the "thing" almost a defacto thing that most users want and use, we can *possibly* switch the default to ON if present and working.

janbrummer:
@bagder What about @mcatanzaro proposal to make use of glib? This is an established library and works very well. Any thoughts? Maybe this is the better way.

dwmw2:
> What I don't understand is why libcurl has to do that itself. If this is actually for using libcurl from another application, the application could just pass the proxy found by libproxy to libcurl.

Has to do *what* for itself? Work out which proxy to use for a given URL?

The point here is that libcurl *doesn't* have to do that for itself. It just asks libproxy. In the ideal case, libproxy ends up just doing a D-Bus call to a standalone tool like pacrunner, which has already started up a JS interpreter with the PAC file _(that in turn was provided from DHCP or VPN config)_ and can serve a response. 

Should the application do that for itself? I don't really think so; no more than the application should have to do DNS for itself, or TLS for itself. Unless it *wants* to override such things, we shouldn't make it.

dwmw2:
> libproxy is barely maintained and objectively pretty terrible

That's all true. Libproxy itself is a vile C++ monstrosity which has long since jumped the shark.

The libproxy *API*, on the other hand, is sane and simple enough and other implementations exist. There's one bundled with pacrunner, which basically just D-Bus calls as I mentioned before. And pacrunner gets its configuration prodded into it by NetworkManager/ConnMan etc., from the underlying DHCP/VPN services or manual network configuration etc.

I'm all for using something other than the original libproxy; this PR is about the *API*. I am kind of ambivalent about doing a GLib/GProxyResolver thing that reloads the PAC file into a JS interpreter in *every* process instead of the pacrunner approach, but if you wrap that into an implementation of the libproxy API, that's just fine. It's just another implementation.

janbrummer:
@dwmw2 Hi, it's been a while :)

With the pacrunner approach you mean something like https://docs.01.org/clearlinux/latest/guides/clear/autoproxy.html ?

dwmw2:
> @dwmw2 Hi, it's been a while :)

Indeed :)

> With the pacrunner approach you mean something like https://docs.01.org/clearlinux/latest/guides/clear/autoproxy.html ?

Yes. PacRunner is a separate dæmon which loads a PAC file into a JS interpreter like duktape (it supports others including mozjs, v8,etc.). It can have its configuration poked into it by systemd, NetworkManager, ConnMan etc. according to the underlying network that we're connected to. Then clients just query it — instead of *all* of the clients having to obtain the information about the current network from whatever happens to be managing it, and having to fetch the PAC file for themselves and run it in their own JS interpreter.

dwmw2:
cf.
 - https://wiki.gnome.org/Projects/NetworkManager/Proxies
 - https://news.softpedia.com/news/networkmanager-1-6-to-support-discovery-and-configuration-of-web-proxy-settings-511118.shtml
 - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/4dce38c37f5fb116b50a6545b483ef628d6fceb6/src/nm-pacrunner-manager.c

bagder:
> What about @mcatanzaro proposal to make use of glib

glib seems like such a huge beast to "suck in" for such a limited functionality, but I don't know anything about it so I can't even tell if that would make it easier to make use of curls own URL parser and HTTP transfer abilities.

janbrummer:
@bagder FTR: I'm currently in contact with all participant of this bug to prepare the best solution for curl. Furthermore I've contacted libproxy maintainer and it looks like I (and my company) are going to support the development and fix the missing issues.

How many thumbs up do you need to remove the needs-votes label? Looks like 19 is already the highest count in your ticket system.

CMakeLists.txt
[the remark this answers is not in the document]
mcatanzaro: You partially inspired me to write a blog post. [See rule three](https://blogs.gnome.org/mcatanzaro/2022/07/15/best-practices-for-build-options/).

bagder:
I [blogged](https://daniel.haxx.se/blog/2022/08/12/the-dream-of-auto-detecting-proxies/) about this case a few days ago.

I need to see an improved libproxy before I will accept its use by libcurl. I outlined 5 different areas in my blogpost in which I would like to see improvements, and while I don't demand that they all are top notch, I also don't think you should expect me to accept this PR in the near term since there is work to do first.

So, I'm glad there area thumbs-ups and that people want to use this. Let's make sure it becomes good enough for us to use it.

juchem:
Forgive me for jumping in on the discussion.

After looking at the proposed changes, it seems like the PR basically wants to do 2 things: 
- enable/disable use of libproxy (the setopt change);
- when enabled, detect the appropriate proxy for a given URL through libproxy.

The rest is build system and other necessary evil stuff.

I propose there's a way for this and similar extensions to be achieved without necessitating changes to libcurl.

For this specific case one could, instead of calling `curl_easy_perform` direcly, implement a simple helper function that:
1. queries the easy handle for the URL;
2. calls libproxy for the appropriate proxy;
3. sets the proxy, if any, using `curl_easy_setopt` with option `CURLOPT_PROXY` & friends;
4. potentially set additional metadata using `curl_easy_setopt` with `CURLOPT_PRIVATE`;
5. call `curl_easy_perform`

Similarly, instead of calling `curl_easy_cleanup` / `curl_easy_reset`, one would implement a similar wrapper that would query the easy handle for a proxy and, if found, properly deallocate the memory for the proxy string.

Looking at the documentation for `CURLOPT_URL`, there doesn't seem to be a way to retrieve the URL once set (perhaps `curl_easy_getinfo` with `CURLINFO_EFFECTIVE_URL`?). Likewise for `CURLOPT_PROXY` & friends.

If all of the above holds true, then instead of adding libproxy support one would add `curl_easy_getinfo` support for the missing information, which would prove to be a much more powerful and general purpose feature. Ideally I would like to see `curl_easy_getopt` implemented, but that may be a much bigger project.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,75 @@
...
+
+  factory = NULL;
+}
+
+char *Curl_libproxy_detect_proxy(const char *url)
+{
+  char *result = NULL;
+
+  if(factory) {
+    char **libproxy_results = px_proxy_factory_get_proxies(factory, url);
```
mcatanzaro, line 59: Just so we don't forget: this should use an asynchronous API. That means changing libproxy to expose an async API, and also changing curl to use it in an asynchronous manner. Network requests should block, but the application thread should not block.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,75 @@
...
+
+    if(libproxy_results) {
+      int i;
+
+      /* We only cope with one; can't fall back on failure */
```
mcatanzaro, line 64: This is unfortunate too. Ideally if curl cannot connect to the first result, it should try the other results sequentially instead of ignoring them. i.e. curl should understand that proxy configuration is a list of possible proxy addresses, not just a single address.

(But certainly trying just the first result is better than not.)

PR #9070, opened 2022-06-30
hyper: use wakers for curl pause/resume

seanmonstar:
Fixes the tests 670-673 (possibly others, if they relate to pausing...)

More details in #8600


bagder (approved):
Nice, four down!

bagder:
Thanks!

PR #9073, opened 2022-06-30
Add ISC to license expression

jlovejoy:
text of the ISC license is in this file, so the SPDX license expression should be updated

@karsten-klein pointed this out on the mailing list and forwarded to me (to give credit where credit it due)


bagder (approved):

bagder:
Thanks!

PR #9074, opened 2022-06-30
Fixes related to digest authorisation

Karlson2k:
This PR fixes a few things related to "session" digest HTTP authorisation:
* remember generated `H(A1)` value and reuse it for the next requests
* always generate new `cnonce` when `cnonce` is used in calculation (for both session and non-session)

---
Some commits were moved to separate PRs:
* #9077
* #9090
* #9076
* #9094
---
I've made some investigation. 
See https://github.com/curl/curl/pull/9074#issuecomment-1177963896 and https://github.com/curl/curl/pull/9074#issuecomment-1178021367

PR #9077, opened 2022-07-01
digest: reject broken header with session protocol and without qop

Karlson2k:
Session algorithms require use of `cnonce`.
`cnonce` could be used only when `qop=auth` or `qop=auth-int`. `cnonce` is not used when `qop` is empty or missing (missing `qop` triggers [RFC2069](https://datatracker.ietf.org/doc/html/rfc2069) compatibility mode, which does not use `cnonce`).
When `qop` is empty, `cnonce` is not sent by curl so there is no way for the server to check calculations where `cnonce` is involved, like `H(A1)`.

Cherry-picked from PR #9074

### DIFF
--- a/lib/vauth/digest.c
+++ b/lib/vauth/digest.c
@@ -637,6 +637,10 @@ CURLcode Curl_auth_decode_digest_http_message(const char *chlg,
   if(!digest->nonce)
     return CURLE_BAD_CONTENT_ENCODING;
 
+  /* "<algo>-sess" protocol versions require "auth" or "auth-int" qop */
+  if(!digest->qop && (digest->algo & SESSION_ALGO))
+    return CURLE_BAD_CONTENT_ENCODING;
+
   return CURLE_OK;
 }
 


bagder (commented):
Have you tried to add a test case that verifies this behavior?

bagder:
Seeing this code made me file #9079 since it'll make it possible to just `&` instead of check for three different values.

Karlson2k:
> Have you tried to add a test case that verifies this behavior?

Not yet. I'm working on #9074 currently. May return to this one later, but for me it's fine if simple test case would be added by someone else. :)

> Seeing this code made me file #9079 since it'll make it possible to just `&` instead of comparing three different values.

I like the idea.

bagder:
> for me it's fine if simple test case would be added by someone else.

It seems we don't even have a single test case using a Digest `-sess` algorithm right now so I can't copy/clone one and I don't know enough of this to make up a test case.

Can you record a working set of HTTP requests and response that we can use to create a test from?

Karlson2k:
> It seems we don't even have a single test case using a Digest `-sess` algorithm right now so I can't copy/clone one and I don't know enough of this to make up a test case.

Yes, curl doesn't have any single test for session digest. However, as no predefined pseudorandom sequence can be enforced for current libcurl (even for testing), it makes a little sense to write test for sessions, as all cnonce and response values would need to be filtered, like it is implemented for a few tests with qop=auth. Actually, any digest test with non-empty qop could be used for sessions digest as the testsuite doesn't check generated values, only presence of them.

Maybe it would make sense to bring back option for predefined pseudorandom sequence, at least for debug builds? Or make it internal only so only tests could use it, while it would not be available in public API (libcurl already have such function, like session reset for easy handle).

> Can you record a working set of HTTP requests and response that we can use to create a test from?
This is the problem. Neither Apache nor Nginx supports sessions (they don't support SHA256 as well).
Maybe IIS has some support, need to check SSPI documentation.

bagder:
> as no predefined pseudorandom sequence can be enforced for current libcurl (even for testing),

Where did you get that from? Debug builds of curl even has a *fixed* random seed when it runs tests, so random is not an issue there.

Karlson2k:
The PR itself is correct, I think. 
Rebased.

Karlson2k:
Rebased

Karlson2k:
Rebased again.

bagder (approved):

bagder:
thanks!

PR #9079, opened 2022-07-01
digest: provide a special bit for "sess" algos

bagder:
Also shortened the names and moved them to the .c file since they are
private for this source file only. Also made them #defines instead of
enum.


Karlson2k:
I like it.
I'd go further and replace some `switch()`es with `if(... &... )`

PR #9085, opened 2022-07-02
Deduplicate exported link libraries for static build

dg0yt:
Another backport from vcpkg:
For static builds, the link link libraries used to be listed both in `Libs` and in `Libs.private` of `libcurl.pc`, leading to duplicate output when using `pkg-config --static` (or when merging the fields in the build system, as vcpkg does, proably for similar reasons as curl has for extending `Libs`).
Fixing this for `libcurl.pc` is easy but implies a similar change to `curl-config` (new, not backported from vcpkg).


dg0yt:
I have to add that I couldn't verify `curl-config`, or impact on other build systems.

bagder:
curl-config is different than pkg-config though. When you invoke curl-config, you use *either* `--libs` or `--static-libs`, not both.

Can you elaborate on what the curl-config edit does?

bagder:
conflicting and no response for months

PR #9087, opened 2022-07-03
docs: sync authentication guidance in username docs

jay:
- Use the auth username guidance from --user, which seems to be the most
  complete, in --proxy-user, CURLOPT_USERNAME, CURLOPT_PROXYUSERNAME,
  CURLOPT_USERPWD, CURLOPT_PROXYUSERPWD.

Prior to this change the documentation was out of sync as to the
recommended ways the user can set the username for NTLM, Kerberos, etc.

Bug: https://github.com/jeroen/curl/issues/237
Reported-by: Oscar Lane

Closes #xxxx

---

@captain-caveman2k note this PR was prompted by jeroen/curl#237 where it is unclear that the username must be blank for SSPI to use the user's Windows credentials.

IMO in the proxy counterparts rather than duplicate all the info we could refer back to the non-proxy version, eg CURLOPT_PROXYUSERPWD could refer to CURLOPT_USERPWD. However they are not exactly the same. Also, there would still be copies in CURLOPT_USERNAME and CURLOPT_USERPWD. So another way to handle this could be make a libcurl-username.3 and refer to that?

---

/cc @jeroen @OscarLane

### DIFF
--- a/docs/cmdline-opts/proxy-user.d
+++ b/docs/cmdline-opts/proxy-user.d
@@ -11,9 +11,23 @@ See-also: proxy-pass
 ---
 Specify the user name and password to use for proxy authentication.
 
-If you use a Windows SSPI-enabled curl binary and do either Negotiate or NTLM
-authentication then you can tell curl to select the user name and password
-from your environment by specifying a single colon with this option: "-U :".
+When using Kerberos V5 with a Windows based server you should include the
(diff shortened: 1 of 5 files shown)


jay:
Dead. I think this will need a different approach rather than duplicate the auth guidance in a bunch of places.

PR #9090, opened 2022-07-03
digest: fix missing increment of 'nc' value for auth-int

Karlson2k:
`nc` must be incremented each time when `nc` is used.
See [RFC7616](https://datatracker.ietf.org/doc/html/rfc7616#section-3.4) (and [RFC2617](https://datatracker.ietf.org/doc/html/rfc2617#section-3.2.2)).

### DIFF
--- a/lib/vauth/digest.c
+++ b/lib/vauth/digest.c
@@ -850,10 +850,8 @@ static CURLcode auth_create_digest_http_message(
                        digest->qop,
                        request_digest);
 
-    if(strcasecompare(digest->qop, "auth"))
-      digest->nc++; /* The nc (from RFC) has to be a 8 hex digit number 0
-                       padded which tells to the server how many times you are
-                       using the same nonce in the qop=auth mode */
+    /* Increment nonce-count to use another nc value for the next request */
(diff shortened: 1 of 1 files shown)


jay:
> `nc` must be incremented each time when `nc` is used.

nc isn't used if qop isn't used so the existing code appears to be correct. If you are referring the latest RFC requiring qop you can see that libcurl has some legacy compatibility and can operate without qop (and therefore nc). Sorry but I just don't see this as an improvement, IMO it's clearer as is.

jay:
Oops I missed your title... you are right the increment is missing for auth-int. I think in that case it could still be handled in the if block for qop.

Karlson2k:
> Oops I missed your title... you are right the increment is missing for auth-int. I think in that case it could still be handled in the if block for qop.

The additional idea was to use `nc` as an indication of `nonce` used at least one time, but this could be handled in another commit.

_I know that digest HTTP auth is defined by RFC 7616, RFC 2617, and (old, deprecated, but still supported) RFC 2069._

jay:
Thanks

PR #9091, opened 2022-07-03
select: do not return fatal error on EINTR from poll()

t-8ch:
The same was done for select() in 5912da253b64de3cb2a1a229558077219b2c8a35
but poll() was missed.

Downstream report: https://bugs.archlinux.org/task/75201
See #8921 and #8961


jay:
Thanks

pitrou:
Seems like this PR missed one code path: https://github.com/curl/curl/issues/11135

PR #9093, opened 2022-07-03
aws_sigv4: use actual Host header, trim values

obataku:
improve compliance with AWS SigV4 spec:
* use actual Host header value in CanonicalHeaders;
* properly trim whitespace in header values

TODO: add a test

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -66,6 +66,38 @@ static void sha256_to_hex(char *dst, unsigned char *sha, size_t dst_l)
   }
 }
 
+static char *aws_canonical_value(const char *hdr)
+{
+  /*
+   * per AWS:
+   *   CanonicalHeadersEntry =
+   *     Lowercase(HeaderName) + ':' + Trimall(HeaderValue)  + '\n'
+   *   The Trimall function removes excess white space before and
+   *   after values, and converts sequential spaces to a single space.
+   * Curl_copy_header_value() handles the first part; we do the rest
+   */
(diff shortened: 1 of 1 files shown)


bagder:
Also the fuzzer found this:
~~~
AddressSanitizer:DEADLYSIGNAL
=================================================================
==33==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7ff91a9f30fc bp 0x7ffd1756a270 sp 0x7ffd17569a28 T0)
==33==The signal is caused by a READ memory access.
==33==Hint: address points to the zero page.
SCARINESS: 10 (null-deref)
    #0 0x7ff91a9f30fc  (/lib/x86_64-linux-gnu/libc.so.6+0x1880fc)
    #1 0x4bc268 in strchr /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc
    #2 0x671f33 in Curl_output_aws_sigv4 /src/curl/lib/http_aws_sigv4.c:239:12
    #3 0x65ad10 in output_auth_headers /src/curl/lib/http.c:695:14

bagder:
Also see #8202 which sounds like a dupe

cosmo-ray:
sorry to bump like this, but https://github.com/curl/curl/pull/7966 try to do the same. but also sort header in alphabetic order, and add user header to the signature computing, which is require for some APIs. (mostly s3)

bagder:
No response and this seems to be addressed in other PRs.

PR #9094, opened 2022-07-04
Fixes for changed authorisation doing second request reusing handle

Karlson2k:
This PR should fix situation not tested by testsuite: changed password or (and) username for easy handle which has been authorized before.
This is applicable only when `libcurl` is used directly, `curl` binary resets easy handle after each request.

If new username/password (or authorisation type) is used then for the next transfer authorisation status must be reset otherwise new username/password will not be used.
As for digest authorisation servers may associate unique nonce values with clients (for session types protocols servers *must* remember nonce-client association together with first `cnonce` (or `H(A1)`) value as the same value is used for the rest of the session), so if `nonce` value was used to send the response it is better to get a fresh `nonce` to avoid any problems with authentication.

As centralized reset of digest data is implemented, a workaround in SSPI code to detect new username/password is removed.

To reduce the number of requests, added tracking whether server 'nonce' has been used at least one time. For example if Digest header was decoded when doing NTLM or Basic authorisation, but `nonce` hasn't been used, the code will use already received `nonce`.

This PR also resets `state.authhost.avail` and `state.authhost.avail` as otherwise digest headers are rejected as duplication by https://github.com/curl/curl/blob/e6f8445edef8e7996d1cfb141d6df184efef972c/lib/http.c#L1031-L1032

Small notification is printed as debug string, when authorisation status and/or digest are reset so user may understand better the ongoing processes.

With accurate tracking of used `nonce` number of required requests are even lower than needed before this patch. Removed not needed now comment about duplicated 401 responses from tests 2027 and 2030.

Added new type of tests with seven additional tests. Added report for connection number in tests to deal with NTLM which requires new/existing connection switching when one of the connection is authorised by  NTLM.

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -72,6 +72,7 @@
 #include "url.h"
 #include "getinfo.h"
 #include "vtls/vtls.h"
+#include "vauth/vauth.h"
 #include "select.h"
 #include "multiif.h"
 #include "connect.h"
@@ -1487,12 +1488,6 @@ CURLcode Curl_pretransfer(struct Curl_easy *data)
     Curl_pgrsResetTransferSizes(data);
     Curl_pgrsStartNow(data);
 
-    /* In case the handle is re-used and an authentication method was picked
-       in the session we need to make sure we only use the one(s) we now
-       consider to be fine */
-    data->state.authhost.picked &= data->state.authhost.want;
-    data->state.authproxy.picked &= data->state.authproxy.want;
-
 #ifndef CURL_DISABLE_FTP
     data->state.wildcardmatch = data->set.wildcard_enabled;
     if(data->state.wildcardmatch) {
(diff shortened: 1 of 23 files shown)


Karlson2k:
Added more tests and implemented a special tool.
_Switch from NTLM to other types should be tested and may need additional fixes._
However current version of PR should not break anything.

Karlson2k:
Rebased, added new `sws` features to deal with NTLM correctly.

Karlson2k:
> This really needs a test case or two to prove itself working and to make sure we don't break it later.

@bagder Tests were added, ready for review.

Karlson2k:
Rebased

Karlson2k:
Rebased again

bagder:
This is a massive change that we never manged to merge.

We need to split this up and do it more step by step in order to better succeed. For example: let's start with a reproducible test case that we can use to prove the existing bug to begin with. Then we can discuss how that bug should be fixed best.

Issue #5669, opened 2020-07-09
Curl static compilation problem (X509_NAME) on Windows with BoringSSL 1.1.0


samuel-tranchet:
I have a recent problem when I try to compile Curl static lib on windows with BoringSSL. The above commands where working fine 6 months ago, but with latest versions of Curl and BoringSSL I came into an issue in the nmake command.

### I did this

```
call buildconf.bat

cd "winbuild"

set RTLIBCFG=static
set _CL_=-MTd -Ob0 -Od -Zi -FS

call nmake /NOLOGO /X - /F MakeFile.vc VC=16 mode=static MACHINE=x64 DEBUG=yes WITH_PREFIX="C:/cpplibs/Curl" WITH_SSL=static SSL_PATH="C:/cpplibs/BoringSSL" ENABLE_SSPI=yes ENABLE_IPV6=yes ENABLE_IDN=yes ENABLE_WINSSL=no
```

### I expected the following

I expected a successful build but I get the following in the console :

```
Generating prerequisite files
* C:\cpplibs\Sources\Curl\latest\Makefile
* C:\cpplibs\Sources\Curl\latest\src\tool_hugehelp.c
configuration name: libcurl-vc16-x64-release-static-ssl-static-ipv6-sspi

cl /O2 /DNDEBUG /MT /DCURL_STATICLIB /I. /I ../lib /I../include /nologo /W4 /EHsc /DWIN32 /FD /c /DBUILDING_LIBCURL /DUSE_OPENSSL /I"C:/cpplibs/BoringSSL"  /DUSE_WIN32_IDN /DWANT_IDN_PROTOTYPES  /DUSE_IPV6  /DUSE_WINDOWS_SSPI /Fo"..\builds\libcurl-vc16-x64-release-static-ssl-static-ipv6-sspi-obj-lib/altsvc.obj"  ..\lib\altsvc.c
altsvc.c

... 100+ successfull calls

[...]

PR #9110, opened 2022-07-06
openssl: fix BoringSSL symbol conflicts with LDAP and Schannel

vszakats:
Same issue as here [1], but this time when building curl with BoringSSL
for Windows with LDAP(S) or Schannel support enabled.

Apply the same fix [2] for these source files as well.

This can also be fixed by moving `#include "urldata.h"` _before_
including `winldap.h` and `schnlsp.h` respectively. This seems like
a cleaner fix, though I'm not sure why it works and if it has any
downside. (see [below](https://github.com/curl/curl/pull/9110#issuecomment-1175624864))

[1] https://github.com/curl/curl/issues/5669
[2] https://github.com/curl/curl/commit/fbe07c6829ba8c5793c84c2856526e19e9029ab9

### DIFF
--- a/lib/ldap.c
+++ b/lib/ldap.c
@@ -45,6 +45,13 @@
 # else
 #  include <winber.h>
 # endif
+  /* Undefine wincrypt conflicting symbols for BoringSSL. */
+# undef X509_NAME
+# undef X509_EXTENSIONS
+# undef PKCS7_ISSUER_AND_SERIAL
+# undef PKCS7_SIGNER_INFO
+# undef OCSP_REQUEST
+# undef OCSP_RESPONSE
 #else
 # define LDAP_DEPRECATED 1      /* Be sure ldap_init() is defined. */
 # ifdef HAVE_LBER_H
--- a/lib/ldap.c
+++ b/lib/ldap.c
@@ -37,6 +37,18 @@
  * OpenLDAP library versions, USE_OPENLDAP shall not be defined.
  */
 
+/* Wincrypt must be included before anything that could include OpenSSL. */
+#if defined(WIN32)
+#include <wincrypt.h>
+/* Undefine wincrypt conflicting symbols for BoringSSL. */
+#undef X509_NAME
+#undef X509_EXTENSIONS
+#undef PKCS7_ISSUER_AND_SERIAL
+#undef PKCS7_SIGNER_INFO
+#undef OCSP_REQUEST
+#undef OCSP_RESPONSE
+#endif
+
 #ifdef USE_WIN32_LDAP           /* Use Windows LDAP implementation. */
 # include <winldap.h>
 # ifndef LDAP_VENDOR_NAME
@@ -45,13 +57,6 @@
 # else
 #  include <winber.h>
 # endif
-  /* Undefine wincrypt conflicting symbols for BoringSSL. */
-# undef X509_NAME
-# undef X509_EXTENSIONS
-# undef PKCS7_ISSUER_AND_SERIAL
-# undef PKCS7_SIGNER_INFO
-# undef OCSP_REQUEST
-# undef OCSP_RESPONSE
 #else
 # define LDAP_DEPRECATED 1      /* Be sure ldap_init() is defined. */
 # ifdef HAVE_LBER_H
--- a/lib/ldap.c
+++ b/lib/ldap.c
@@ -38,7 +38,7 @@
  */
 
 /* Wincrypt must be included before anything that could include OpenSSL. */
-#if defined(WIN32)
+#if defined(USE_WIN32_CRYPTO)
 #include <wincrypt.h>
 /* Undefine wincrypt conflicting symbols for BoringSSL. */
 #undef X509_NAME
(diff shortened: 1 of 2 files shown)


vszakats:
Alternative fix:
```diff
diff --git a/lib/ldap.c b/lib/ldap.c
index 51a32dc96..13b684104 100644
--- a/lib/ldap.c
+++ b/lib/ldap.c
@@ -37,6 +37,8 @@
  * OpenLDAP library versions, USE_OPENLDAP shall not be defined.
  */
 
+#include "urldata.h"
+
 #ifdef USE_WIN32_LDAP           /* Use Windows LDAP implementation. */
 # include <winldap.h>
 # ifndef LDAP_VENDOR_NAME
@@ -56,7 +58,6 @@
 # endif /* HAVE_LDAP_SSL && HAVE_LDAP_SSL_H */
 #endif
 
-#include "urldata.h"
 #include <curl/curl.h>
 #include "sendf.h"
 #include "escape.h"
diff --git a/lib/vtls/schannel.h b/lib/vtls/schannel.h
index 0b4c4d934..45ecb029b 100644
--- a/lib/vtls/schannel.h
+++ b/lib/vtls/schannel.h
@@ -28,12 +28,12 @@
 
 #ifdef USE_SCHANNEL
 
+#include "urldata.h"
+
 #include <schnlsp.h>
 #include <schannel.h>
 #include "curl_sspi.h"
 
-#include "urldata.h"
-
 /* <wincrypt.h> has been included via the above <schnlsp.h>.
  * Or in case of ldap.c, it was included via <winldap.h>.
  * And since <wincrypt.h> has this:
```

jay:
Interesting. Could boringssl have fixed it to undefine the symbols? For all versions we have the way that works, I would just copy that, note it must come after curl_setup.h which always comes first.

https://github.com/curl/curl/blob/45ac4d019475df03562fe0ac54eb67e1d1de0ca7/lib/vtls/openssl.c#L36-L46

vszakats:
Probably could be fixed in BoringSSL, but I see little chance getting it fixed there. Likely Google don't mix it with `wincrypt.h` in their own source code to justify a workaround. A local note about include order should be fine I think.

vszakats:
Found this in `openssl.c`:
```
/* Wincrypt must be included before anything that could include OpenSSL. */
```

This was a similar bug where apparently the exact reverse action solved the problem for OpenSSL.

Ref: https://github.com/curl/curl/pull/5606

jay:
I would explicitly include wincrypt, followed by the undefines, as early as possible so that something that includes something that includes BoringSSL can't come before it.

jay:
I mean this, it's not really an alternative I just added the include

~~~diff
diff --git a/lib/ldap.c b/lib/ldap.c
index 51a32dc..292684b 100644
--- a/lib/ldap.c
+++ b/lib/ldap.c
@@ -26,6 +26,18 @@
 
 #if !defined(CURL_DISABLE_LDAP) && !defined(USE_OPENLDAP)
 
+/* Wincrypt must be included before anything that could include OpenSSL. */
+#if defined(USE_WIN32_CRYPTO)
+#include <wincrypt.h>
+/* Undefine wincrypt conflicting symbols for BoringSSL. */
+#undef X509_NAME
+#undef X509_EXTENSIONS
+#undef PKCS7_ISSUER_AND_SERIAL
+#undef PKCS7_SIGNER_INFO
+#undef OCSP_REQUEST
+#undef OCSP_RESPONSE
+#endif
+
 /*
  * Notice that USE_OPENLDAP is only a source code selection switch. When
  * libcurl is built with USE_OPENLDAP defined the libcurl source code that
diff --git a/lib/vtls/schannel.h b/lib/vtls/schannel.h
index 0b4c4d9..05323c7 100644
--- a/lib/vtls/schannel.h
+++ b/lib/vtls/schannel.h
@@ -28,6 +28,18 @@
 
 #ifdef USE_SCHANNEL
 
+/* Wincrypt must be included before anything that could include OpenSSL. */
+#if defined(USE_WIN32_CRYPTO)
+#include <wincrypt.h>
+/* Undefine wincrypt conflicting symbols for BoringSSL. */
+#undef X509_NAME
+#undef X509_EXTENSIONS
+#undef PKCS7_ISSUER_AND_SERIAL
+#undef PKCS7_SIGNER_INFO
+#undef OCSP_REQUEST
+#undef OCSP_RESPONSE
+#endif
+
 #include <schnlsp.h>
 #include <schannel.h>
 #include "curl_sspi.h"
~~~

either that or maybe it could just be included once in whatever curl header is actually causing this problem by including boringssl, rather than putting it in each unit

vszakats:
@jay: Thanks, got it. But, speaking of MinGW, both `schannel.h` and `winldap.h` include `wincrypt.h` unconditionally anyway, so this reduces the fix to cases when `USE_WIN32_CRYPTO` is defined in the build. Do we want that?

vszakats:
I re-set it to use `USE_WIN32_CRYPTO`. This macro should always be set for MinGW, and I assume on systems where it isn't set, `schannel.h` and `winldap.h` would not attempt to pull in `wincrypt.h`. In UWP apps maybe.

PR #9114, opened 2022-07-06
tool_operate.c: fix several problems about the usage of curl_easy_init()

x2018:
On the one hand, curl_easy_init() returns NULL if some errors happen.
So it is better to check the return of it to propagate the proper status.
Without the check, if curl_easy_init() fails, transfer_per_config() will
return CURLE_UNKNOWN_OPTION since curl_easy_getinfo() returns that status
when the first argument is NULL.

On the other hand, we should call curl_easy_cleanup() to prevent potential
memory leaks on other error handling paths.

### DIFF
--- a/lib/getinfo.c
+++ b/lib/getinfo.c
@@ -560,7 +560,7 @@ CURLcode Curl_getinfo(struct Curl_easy *data, CURLINFO info, ...)
   CURLcode result = CURLE_UNKNOWN_OPTION;
 
   if(!data)
-    return result;
+    return CURLE_BAD_FUNCTION_ARGUMENT;
 
   va_start(arg, info);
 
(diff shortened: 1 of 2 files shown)


jay:
There may some places in the code where we call curl_easy_getinfo without first checking the curl_easy_init return. You could change curl_easy_getinfo to return CURLE_BAD_FUNCTION_ARGUMENT for NULL handle. Regardless though, I'm not opposed to cleaning them up.

x2018:
Thank you for your suggestion. I removed the check for curl_easy_init() before curl_easy_getinfo(), and made Curl_getinfo() return CURLE_BAD_FUNCTION_ARGUMENT for handling NULL of the first argument.

bagder (approved):

bagder:
Thanks!

PR #9116, opened 2022-07-07
fix [Known Bugs]1.11 CURLOPT_SEEKFUNCTION not called with CURLFORM_STREAM

nickychen123456:
When using libcurl to POST form data using a FILE* with the CURLFORM_STREAM option of curl_formadd(). I notice that if the connection drops at just the right time, the POST is reattempted without the data from the file. It seems like the file stream position is not getting reset to the beginning of the file. I found the CURLOPT_SEEKFUNCTION option and set that with a function that performs an fseek() on the FILE*. However, setting that did not seem to fix the issue or even get called.

I have fixed this know issue locally, please help to review it, thanks

Fixes https://github.com/curl/curl/issues/9120

### DIFF
--- a/lib/formdata.c
+++ b/lib/formdata.c
@@ -716,7 +716,7 @@ int curl_formget(struct curl_httppost *form, void *arg,
   curl_mimepart toppart;
 
   Curl_mime_initpart(&toppart, NULL); /* default form is empty */
-  result = Curl_getformdata(NULL, &toppart, form, NULL);
+  result = Curl_getformdata(NULL, &toppart, form, NULL, NULL);
   if(!result)
     result = Curl_mime_prepare_headers(&toppart, "multipart/form-data",
                                        NULL, MIMESTRATEGY_FORM);
@@ -802,7 +802,8 @@ static CURLcode setname(curl_mimepart *part, const char *name, size_t len)
 CURLcode Curl_getformdata(struct Curl_easy *data,
                           curl_mimepart *finalform,
                           struct curl_httppost *post,
-                          curl_read_callback fread_func)
+                          curl_read_callback fread_func,
+                          curl_seek_callback seek_func)
 {
   CURLcode result = CURLE_OK;
   curl_mime *form = NULL;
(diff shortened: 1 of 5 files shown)


jay:
@monnerat

monnerat:
The form API is deprecated and the mime API should be used instead. That said, even in the pre-mime times, the form stream was not rewound unless `CURLOPT_SEEKDATA` was set to the same handle as the `CURLFORM_STREAM` argument (this strategy should still work nowadays: we should have a test for it !). The [CURLFORM_STREAM](https://curl.se/libcurl/c/curl_formadd.html#CURLFORMSTREAM) documentation does not mention a use of the seek function.

In other words, you are adding a new feature to a deprecated API. One of the main reason for replacement by mime was those kind of caveats and I think it is a perfect case where form-->mime migration is recommended.

Regarding your code: I don't think it's necessary to copy the `seek_func` value from `set` to `state` as it does not change during transfer, you can use it directly from `set`. The rest seems OK and very clear. Maybe update the `CURLFORM_STREAM` documentation to explicit this?

bagder:
@nickychen123456 thanks, but can you first elaborate on why you are using this deprecated API? It seems wrong to do anything about it other than leaving it alone and pushing users over to the replacement we have provided since curl 7.56.0.

nickychen123456:
@bagder @monnerat 
Our product has used libcurl for a long time, and has always used CURLFORM_STREAM to send files. We don't want to replace it with mime, because the code changes may cause risks, so we hope to fix this problem based on CURLFORM_STREAM.

We hope that this change could be merged, it is a great help for our product. Many thanks.

bagder:
> We don't want to replace it with mime, because the code changes may cause risks

I read that as you rather have us take the risk instead of you. The formadd API has been deprecated since October 2017. Also, clearly you need to change a lot of code now anyway to add the use of a seek callback and you need to switch to a libcurl version that supports this callback for the formadd API.

bagder (changes requested):
This looks simple enough, but we need a test case or two as well to verify that it works and to prevent it from breaking as we go forward.

Also, should this use of the callback also perhaps be mentioned in the docs for `CURLFORM_STREAM` ?

bagder:
No response.

PR #9117, opened 2022-07-07
build: improve OS string in CMake and `config-win32.h`

vszakats:
This patch makes CMake fill the "OS string" with the value 
of `CMAKE_C_COMPILER_TARGET`, if passed. This typically
contains a triplet, the same that is passed to `./configure` via
`--host=`.

For non-CMake, non-autotools, Windows builds, this patch 
adds the ability to override the default `OS` value in
`lib/config-win32.h`.

With these its possible to get the same OS string across the
three build systems.

This patch supersedes the earlier, partial, CMake-only solution:
435f395f3f8c11eebfcc243ca55ebcc11a19b8b8, thus retiring the `CURL_OS_SUFFIX` CMake option.

Closes #xxxx


jay:
Do you see anyone besides yourself manually setting the OS string? It seems to me a better solution would be improve cmake's detected OS string.

vszakats:
Anybody who wants to create reproducible builds across multiple build system will face these, but yes, it might be only me doing that. Not sure how to improve this in CMake, but even if doable, I suspect any autodetection would be inherently different from `autotools` results and will definitely remain different from what `config-win32.h` does. This is also a better fix for CMake than the earlier one already committed.

dfandrich:
I would also rather have the CMake build choose the right system triplet on its
own. This is entirely a automatible process that doesn't need to be have a
manual override. It wouldn't surprise me if there were a CMake macro available
to do this already, bu worst-case, you could just include config.guess in the
tarball and call that to get the same triplet as autoconf.

vszakats:
Actually autotools isn't even guessing the triplet in my (or any cross-build) case, I'm sending it to `./configure` via `--host=`. This value then find its way into the version string.

The same could be done with CMake, where the triplet is passed (for cross-builds at least) via `CMAKE_C_COMPILER_TARGET`. That can be used to fill `OS`, if not empty:
```cmake
set(OPERATING_SYSTEM "${CMAKE_SYSTEM_NAME}")
if(CMAKE_C_COMPILER_TARGET)
  set(OS "\"${CMAKE_C_COMPILER_TARGET}\"")
else()
  set(OS "\"${CMAKE_SYSTEM_NAME}\"")
endif()
```

This would change the default behaviour for CMake builds where this value is passed (unless we put it behind an extra switch).

jay:
I don't understand how this makes it reproducible in Windows pregenerated config. If the triplet is the same it's already the same string. Similar for cmake, if anything "Windows" is always "Windows"? Not that it should be that, but you can see what I mean. IMO CURL_OS_SUFFIX is a workaround for the actual problem which is lack of cmake detailed target recognition

vszakats:
@jay: In case of cross-builds there no automatic target recognition (there cannot be), it's an input provided manually to the build systems by the builder. Currently this input is only used to form the OS string with autotools. Makefile.m32 uses some hard-coded value, and CMake just puts 'Windows' there. It means that even though I do build for the exact same target with the same triplet and with precisely the same build options, the three build systems generate objects that differ only in the object compiled from `version.c`. (At the moment I need to resort to patching curl to force them use the same OS string.)

In general, CMake's solution to put 'Windows' or other crude OS string there seems wrong in general, especially when someone clearly passed a triplet to it via `CMAKE_C_COMPILER_TARGET`.

The (updated) proposed solution makes CMake use the above value, if present. It also lets the user override `OS` in `config-win32.h` (used by Makefile.m32 and perhaps some MSVC builds.) The latter could be tweaked to produce the desired results automatically (e.g. by recognizing MinGW env). Though with more complexitly, fragility and making it less future-proof.

jay:
Ok for most recent cmake change. Are you saying you need to override OS in config-win32? What is wrong with the OS logic already there, is it not accurate enough?

vszakats:
@jay: That's correct, I'd like to see the same OS string for CMake, Makefile.m32 and autotools builds.

jay (approved):

PR #9124, opened 2022-07-08
CMake: remove APPEND in export(TARGETS)

rouault:
When running cmake several times, new content was appended to already
existing generated files, which is not appropriate


jzakrzewski (approved):
LGTM.
There is a single target in each file anyway, so appending did not make sense from the beginning I guess.

PR #9125, opened 2022-07-08
CMake: link curl to its dependencies with PRIVATE

rouault:
The current PUBLIC visibility causes issues for downstream users.
Cf https://github.com/OSGeo/PROJ/pull/3172#issuecomment-1157942986

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -95,7 +95,7 @@ if(NOT BUILD_SHARED_LIBS)
     set_target_properties(${LIB_NAME} PROPERTIES INTERFACE_COMPILE_DEFINITIONS CURL_STATICLIB)
 endif()
 
-target_link_libraries(${LIB_NAME} ${CURL_LIBS})
+target_link_libraries(${LIB_NAME} PRIVATE ${CURL_LIBS})
 
 set_target_properties(${LIB_NAME} PROPERTIES
   COMPILE_DEFINITIONS BUILDING_LIBCURL


lib/CMakeLists.txt
```diff
@@ -95,7 +95,14 @@ if(NOT BUILD_SHARED_LIBS)
...
 
-target_link_libraries(${LIB_NAME} ${CURL_LIBS})
+if(BUILD_SHARED_LIBS)
+    target_link_libraries(${LIB_NAME} PRIVATE ${CURL_LIBS})
+else()
+    # It is arguable that the default PUBLIC visibility is correct for
+    # static builds, but it might be a convenient workaround for now,
+    # as correct dependency tracking for static build requires more effort.
+    target_link_libraries(${LIB_NAME} ${CURL_LIBS})
+endif()
```
dg0yt, line 105: `PRIVATE` should be correct for both shared and static linkage. CMake takes care of this in exported CMake config by adding a `$<LINK_ONLY:target>` generator expression in the `INTERFACE_LINK_LIBRARIES` for private static linkage.

rouault: indeed. pull request amended to use PRIVATE unconditionally

dfandrich:
Is this the equivalent of the autoconf option --enable-symbol-hiding or is it
something else?

dg0yt:
> Is this the equivalent of the autoconf option --enable-symbol-hiding ...

No. It is not on this level. It is more like `pkg-config`'s `Libs.private`.

jzakrzewski (approved):
Seems correct.
Before the change we actually got unnecessary overlinking when using with `FetchContent`. Since libcurl does not use any symbols from 3-party libraries in its interface this is the way to go.

Note this may break some projects that assumed that libcurl will pull those dependencies in, but they should have never assume it in the first place and link, what they use.

Issue #8733, opened 2022-04-22
trying to compile and install curl with -http3 from the source code 


ShawnBian:
I just strictly follow the below links to install curl with http3 from my centos8, unfortunately, it is failed and really drive me crazy :
https://github.com/curl/curl/blob/master/docs/HTTP3.md

[root@centos8 curl]# LDFLAGS="-Wl,-rpath,/usr/local/openssl/lib" ./configure --with-openssl=/usr/local/openssl --with-nghttp3=/usr/local/nghttp3 --with-ngtcp2=/usr/local/ngtcp2 --enable-alt-svc
it throw below error when I tried the last step:
checking for libidn2 options with pkg-config... no
configure: IDN_LIBS: "-lidn2"
configure: IDN_LDFLAGS: ""
configure: IDN_CPPFLAGS: ""
configure: IDN_DIR: ""
checking if idn2_lookup_ul can be linked... no
checking idn2.h usability... no
checking idn2.h presence... no
checking for idn2.h... no
configure: WARNING: Cannot find libraries for IDN support: IDN disabled
checking for pkg-config... (cached) /usr/bin/pkg-config
checking for libnghttp2 options with pkg-config... no
checking for nghttp2_session_set_local_window_size in -lnghttp2... no
checking for pkg-config... (cached) /usr/bin/pkg-config
checking for libngtcp2 options with pkg-config... found
configure: -l is -lngtcp2 
configure: -I is -I/usr/local/ngtcp2/include 
configure: -L is -L/usr/local/ngtcp2/lib 
checking for ngtcp2_conn_client_new_versioned in -lngtcp2... yes
checking ngtcp2/ngtcp2.h usability... yes
checking ngtcp2/ngtcp2.h presence... yes
checking for ngtcp2/ngtcp2.h... yes
configure: Added /usr/local/ngtcp2/lib to CURL_LIBRARY_PATH
checking for pkg-config... (cached) /usr/bin/pkg-config
checking for libngtcp2_crypto_openssl options with pkg-config... no
configure: error: --with-ngtcp2 was specified but could not find ngtcp2_crypto_openssl pkg-config file.

really appreciated it if someone know the reason for the error above. thanks

bagder:
>  configure: error: --with-ngtcp2 was specified but could not find ngtcp2_crypto_openssl pkg-config file.

[...]

PR #9126, opened 2022-07-09
Fix HTTP3.md?

ruokeqx:
I meet the same problem with https://github.com/curl/curl/issues/8733 .  And i realize the real path is `<somewhere1>/lib64`. Should we change the Doc?


ruokeqx:
I mean if we follow the Doc step by step, the autual path is `<somewhere1>/lib64/pkgconfig`

PR #9139, opened 2022-07-11
base64: base64url encoding has no padding

monnerat:
See RFC4648 section 5 and RFC7540 section 3.2.1.

Suppress generation of '=' padding of base64url encoding. This is
accomplished by considering the string beginning at offset 64 in the
character table as the padding: this is "=" for base64, "" for base64url.

Also use strchr() to replace character search loops where possible.

Suppress erroneous comments about empty encoding results.

Adjust unit test 1302 to unpadded base64url encoding and add tests for
empty results.

### DIFF
--- a/lib/base64.c
+++ b/lib/base64.c
@@ -43,8 +43,9 @@
 #include "memdebug.h"
 
 /* ---- Base64 Encoding/Decoding Table --- */
+/* Padding character string starts at offset 64. */
 static const char base64[]=
-  "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+  "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
 
 /* The Base 64 encoding with an URL and filename safe alphabet, RFC 4648
    section 5 */
@@ -54,25 +55,18 @@ static const char base64url[]=
 static size_t decodeQuantum(unsigned char *dest, const char *src)
 {
   size_t padding = 0;
-  const char *s, *p;
+  const char *s;
   unsigned long i, x = 0;
 
   for(i = 0, s = src; i < 4; i++, s++) {
     if(*s == '=') {
-      x = (x << 6);
+      x <<= 6;
       padding++;
     }
     else {
-      unsigned long v = 0;
-      p = base64;
-
-      while(*p && (*p != *s)) {
-        v++;
-        p++;
-      }
-
-      if(*p == *s)
-        x = (x << 6) + v;
+      const char *p = strchr(base64, *s);
+      if(p)
+        x = (x << 6) + curlx_uztoul(p - base64);
       else
         return 0;
     }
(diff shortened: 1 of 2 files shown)


lib/base64.c
```diff
@@ -126,19 +120,17 @@ CURLcode Curl_base64_decode(const char *src,
...
+    /* A maximum of two = padding characters is allowed */
+    if(padptr[1] == '=')
       padding++;
-  }
 
-  /* Check the = padding characters weren't part way through the input */
-  if(length + padding != srclen)
-    return CURLE_BAD_CONTENT_ENCODING;
+    /* Check the = padding characters weren't part way through the input */
+    if(padptr + padding != src + srclen)
```
jay, line 131:
```suggestion
    if(padptr && (padptr + padding != src + srclen))
```

lib/base64.c
```diff
@@ -224,27 +217,30 @@ static CURLcode base64_encode(const char *table64,
...
     switch(inputparts) {
     case 1: /* only one byte read */
-      msnprintf(output, 5, "%c%c==",
-                table64[obuf[0]],
-                table64[obuf[1]]);
+      i = msnprintf(output, 5, "%c%c%s%s",
+                    table64[obuf[0]],
+                    table64[obuf[1]],
+                    padstr,
+                    padstr);
```
jay, line 224: I don't see how using `table64 + 64` is better, if anything it's harder to understand.

jay (changes requested):

lib/base64.c
monnerat: This is ot needed as it is the if condition for the containing conditional block (line 124).

lib/base64.c
monnerat: In the base64url case, we have a padding string of length 0, in base 64, this is "=" of length 1.
We thus cannot assume padding is a single character: we must use a string. Having the latter appended to the char table itself is the simplest way to achieve it. It is much more readable to have `padstr` rather than `table64 + 64` in the calls, and probably optimized by the compiler.

jay: Ok I think I understand now. There's two base64 encodings, the first regular, 
"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
the second for url
"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_"
and in the latter case padstr (table+64) is going to be null?

lib/base64.c
```diff
@@ -187,6 +179,7 @@ static CURLcode base64_encode(const char *table64,
   char *output;
   char *base64data;
   const char *indata = inputbuff;
+  const char *padstr = table64 + 64;
```
bagder, line 182: maybe `const char *padstr = &table64[64]` instead? I think it makes the code clearer what it wants to point to.

monnerat: Done.

bagder (approved):

bagder:
Thanks!

PR #9140, opened 2022-07-11
runtests: increase valgrind max thread count limit

monnerat:
By default, valgrind limits the number of threads in the target program
to 500 while test 3026 creates at least 1000.

Increase the valgrind limit to avoid a systematic test 3026 failure.

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -727,6 +727,7 @@ sub torture {
                 $valgrindcmd .= "--suppressions=$srcdir/valgrind.supp ";
                 # $valgrindcmd .= "--gen-suppressions=all ";
                 $valgrindcmd .= "--num-callers=16 ";
+                $valgrindcmd .= "--max-threads=1030 ";
                 $valgrindcmd .= "${valgrind_logfile}=$LOGDIR/valgrind$testnum";
                 $cmd = "$valgrindcmd $testcmd";
             }
@@ -4197,6 +4198,7 @@ sub singletest {
             $valgrindcmd .= "--suppressions=$srcdir/valgrind.supp ";
            # $valgrindcmd .= "--gen-suppressions=all ";
             $valgrindcmd .= "--num-callers=16 ";
+            $valgrindcmd .= "--max-threads=1030 ";
             $valgrindcmd .= "${valgrind_logfile}=$LOGDIR/valgrind$testnum";
             $CMDLINE = "$valgrindcmd $CMDLINE";
         }


monnerat:
> Maybe it is easier if the test is limited to 100 for all platforms and iterate 10x.

This can be an alternate solution if this satisfies the test requirements.

SukanyaHanumanthu:
Hi Jay, 

Is this a fix for the below ?

```
TESTDONE: 1520 tests were considered during 500 seconds.
TESTDONE: 1080 tests out of 1081 reported OK: 99%

TESTFAIL: These test cases failed: 3026 

[91mmake[1]: *** [quiet-test] Error 1
[0mmake[1]: Leaving directory `/tmp/curl-7.84.0/tests'
[91mmake: *** [test] Error 2
[0mThe command '/bin/sh -c cd curl-${CURL_VERSION} &&     ./configure         --prefix ${CURL_BUILD_INSTALL_PREFIX}         --disable-shared         --with-pic         --without-ssl         --without-zlib &&     make &&     make test &&     make install' returned a non-zero code: 2
```

jay:
> Is this a fix for the below ?

I don't know, run that specific test and check the output: `./runtests.pl -p 3026`. #9012 is supposed to be a Windows fix but we may expand it.

SukanyaHanumanthu:
I think expanding it to the other platforms is the right thing to do. The mentioned failure happened in Linux.

bagder:
I too think it would rather be better to reduce the number of threads in the test.

mback2k:
I will improve my PR regarding this tonight.

SukanyaHanumanthu:
A bit more details into the error

```
test 3020...OK (1498 out of 1504, remaining: 00:02, took 0.004s, duration: 08:20)
--
[91mUse of uninitialized value $USER in substitution (s///) at ./runtests.pl line 3453.
[0m[91mUse of uninitialized value $USER in substitution (s///) at ./runtests.pl line 3453.
[0mtest 3025...OK (1503 out of 1504, remaining: 00:00, took 0.004s, duration: 08:20)
test 3026... exit FAILED
```

monnerat:
@SukanyaHanumanthu: No, this PR does not fix your error.
The problem fixed here is `log/valgrind3026` containing:
```
Use --max-threads=INT to specify a larger number of threads
and rerun valgrind

valgrind: the 'impossible' happened:
   Max number of threads is too low
```
> Looks like [this](https://github.com/curl/curl/commit/2ed101256414ea5274e85e6f3d4c011b46442ac4) commit introduced the error, is there a way to skip this test while doing make test?

You can locally add 3026 in `tests/data/DISABLED`.

monnerat:
This PR is no longer needed: test 3026 thread count has been reduced to 100 by commit 0484127.
Closing.

PR #9147, opened 2022-07-12
Allow memory leak instead of waiting for name resolver threads

### LINKED ISSUE #2975: waiting for name resolver threads before quitting can cause delays even with --max-time
### I did this
1. Edit `/etc/resolve.conf` to use a bad DNS server.
2. `time curl https://github.com/ --max-time 1 -v`
And `curl` exits after 5 seconds. The output is 
```
time curl https://github.com/ --max-time 1 -v
* Resolving timed out after 1000 milliseconds
* Could not resolve host: github.com
* stopped the pause stream!
* Closing connection 0
curl: (28) Resolving timed out after 1000 milliseconds
curl https://github.com/ --max-time 1 -v  0.02s user 0.02s system 0% cpu 5.042 total
```

### I expected the following
`curl` exits after 1 second.
Document at https://ec.haxx.se/usingcurl-timeouts.html says "When the set time has elapsed, curl will exit no matter what is going on at that moment", so I think this is a bug.

### curl/libcurl version
curl 7.61.0 (x86_64-pc-linux-gnu) libcurl/7.61.0 OpenSSL/1.1.0i zlib/1.2.11 libidn2/2.0.5 libpsl/0.20.2 (+libidn2/2.0.4) nghttp2/1.32.0
Release-Date: 2018-07-11
Protocols: dict file ftp ftps gopher http https imap imaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: AsynchDNS IDN IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz TLS-SRP HTTP2 UnixSockets HTTPS-proxy PSL

### operating system
Arch Linux

ferrieux:
This PR solves the responsibility issue mentioned in #2975 between `libcurl` and `curl` by exposing an option, `CURLOPT_MAY_LEAK`, that curl then sets to 1, in order to tell `libcurl` to skip waiting for resolver thread termination.
In that case, `libcurl` indeed allows itself to *leak* threads until `curl` 's exit.

Note: I would have *loved* to add a test case; however, doing this implies emulating an unresponsive DNS server, and while I can do it on Linux, intrusively as root with `iptables`, or with a mount namespace and a tweaked `/etc/resolv.conf`, I know of no simple userland, non-root, portable way of saying `curl -m 3 http://host-that-will-hang-DNS`. Welcoming any suggestions !

Note2: this is my first PR, apologies if I missed  important things in contribution etiquette...


ferrieux:
Note what looks like a permafail of Cirrus, FreeBSD 12.3 and 13.1 both stumbling on a missing Python package...
I assume you detected this already, as it happens on master too.

lib/asyn-thread.c
```diff
@@ -536,7 +536,7 @@ void Curl_resolver_kill(struct Curl_easy *data)
   /* If we're still resolving, we must wait for the threads to fully clean up,
      unfortunately.  Otherwise, we can simply cancel to clean up any resolver
      data. */
-  if(td && td->thread_hnd != curl_thread_t_null)
+  if(td && td->thread_hnd != curl_thread_t_null && !data->set.may_leak)
```
bagder, line 539: This makes the else block run instead, which isn't what you want is it?

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
```diff
@@ -0,0 +1,56 @@
...
+.SH NAME
+CURLOPT_MAY_LEAK \- allow to exit quickly
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+
+CURLcode curl_easy_setopt(CURL *handle, CURLOPT_MAY_LEAK,
+                          boolean value);
+.SH DESCRIPTION
+Pass a boolean as a parameter, true meaning libcurl should skip lengthy cleanups
```
bagder, line 35: It should pass a `long` as a parameter to make it C89 compliant where there is no native boolean type, and the value to trigger should probably be set to `1` (to allow for future differentiators).

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
```diff
@@ -0,0 +1,56 @@
...
+.SH DEFAULT
+0
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+CURL *curl = curl_easy_init();
+if(curl) {
+  CURLcode ret;
+  curl_easy_setopt(curl, CURLOPT_MAY_LEAK, 1);
```
bagder, line 47: Pass the value as `1L` to work (reliably) on windows.

docs/libcurl/curl_easy_setopt.3
```diff
@@ -684,6 +684,8 @@ Share object to use. See \fICURLOPT_SHARE(3)\fP
 Mode for creating new remote files. See \fICURLOPT_NEW_FILE_PERMS(3)\fP
 .IP CURLOPT_NEW_DIRECTORY_PERMS
 Mode for creating new remote directories. See \fICURLOPT_NEW_DIRECTORY_PERMS(3)\fP
+.IP CURLOPT_MAY_LEAK
```
bagder, line 687: I think "may leak" is a weird name of the option since that's just an effect of what you're asking for. You're actually asking for libcurl to *not* wait around (for threads) when closing down a transfer. Maybe "fast exit", "quick exit" or "fast shutdown" or something is a better name?

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
```diff
@@ -0,0 +1,56 @@
...
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+
+CURLcode curl_easy_setopt(CURL *handle, CURLOPT_MAY_LEAK,
+                          boolean value);
+.SH DESCRIPTION
+Pass a boolean as a parameter, true meaning libcurl should skip lengthy cleanups
+that are intended to avoid all kinds of leaks (threads etc.), as the caller program
+is about to call exit() anyway. This allows for a swift termination after a DNS timeout for example.
```
bagder, line 37: I think you should be specific about what exactly this might leak at the time of this implementation: a name resolver thread.

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
```diff
@@ -0,0 +1,56 @@
...
+  curl_easy_setopt(curl, CURLOPT_MAY_LEAK, 1);
+  ret = curl_easy_perform(curl);
+}
+.fi
+.SH AVAILABILITY
+Added in 7.85.0
+.SH RETURN VALUE
+Returns CURLE_OK if the option is supported, and CURLE_UNKNOWN_OPTION if not.
+.SH "SEE ALSO"
+.BR Nothing
```
bagder, line 56: As the leak this might trigger is related to name resolving, I think you can refer to `CURLOPT_IPRESOLVE(3)` here instead of "nothing". Possibly even `CURLOPT_RESOLVE(3)` since that is about providing IP addresses for names ahead of time.

src/tool_operate.c
```diff
@@ -1230,6 +1230,14 @@ static CURLcode single_transfer(struct GlobalConfig *global,
         }
 #endif
 
+        /* We'll clean everything up via exit(), so don't bother with
+           slow cleanups.
+           Note: avoid having this setopt added to the --libcurl source
+           output. */
+        result = curl_easy_setopt(curl, CURLOPT_MAY_LEAK, 1L);
```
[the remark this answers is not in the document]
ferrieux: Sure, make it a conditional default, then. On the operating systems where we know `_exit(2)` works, let's do it.

lib/asyn-thread.c
ferrieux: This is what I want: see the full code block below.

```c
 if(td && td->thread_hnd != curl_thread_t_null && !data->set.may_leak)
     (void)thread_wait_resolv(data, NULL, FALSE);
   else
     Curl_resolver_cancel(data);
```

As it turns out, the `else` block does call `Curl_resolver_cancel`, which (1) doesn't block and (2) is useful as it avoids a race condition if the thread happens to finish at the same time:

```c
 /* within destroy_async_data() in async_thread.c */

    Curl_mutex_acquire(td->tsd.mtx);
    done = td->tsd.done;
    td->tsd.done = 1;
    Curl_mutex_release(td->tsd.mtx);

    if(!done) {
      Curl_thread_destroy(td->thread_hnd);
    }
*/

File: lib/url.c [hunk 8]
```diff
@@ -638,6 +638,7 @@ CURLcode Curl_init_userdefined(struct Curl_easy *data)
 #endif
     ;
   Curl_http2_init_userset(set);
+  set->may_leak = 0;
```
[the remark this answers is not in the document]
ferrieux: Turned to `0L` instead, as @bagder  suggests to use a `long` instead of a `boolean`.

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
ferrieux: DONE

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
ferrieux: DONE

docs/libcurl/curl_easy_setopt.3
ferrieux: DONE: "quick exit"

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
ferrieux: DONE

docs/libcurl/opts/CURLOPT_MAY_LEAK.3
ferrieux: DONE

docs/libcurl/curl_easy_setopt.3
dawidmedycki: g

src/tool_operate.c
```diff
@@ -1230,6 +1230,17 @@ static CURLcode single_transfer(struct GlobalConfig *global,
         }
 #endif
 
+#if defined(__unix) || defined (WIN32)
```
bagder, line 1233: I think this is an unnecessary precaution. If there's a platform with a problem with this, we should add a reversed check for that specific one.

docs/libcurl/opts/CURLOPT_QUICK_EXIT.3
```diff
@@ -0,0 +1,60 @@
...
+.nf
+CURL *curl = curl_easy_init();
+if(curl) {
+  CURLcode ret;
+  curl_easy_setopt(curl, CURLOPT_QUICK_EXIT, 1L);
+  ret = curl_easy_perform(curl);
+}
+.fi
+.SH AVAILABILITY
+Added in 7.85.0
```
bagder, line 56: Unfortunately we missed this feature window as well, so 7.87.0 is the next version to target with this change.

docs/libcurl/symbols-in-versions
```diff
@@ -677,6 +677,7 @@ CURLOPT_MAIL_AUTH               7.25.0
 CURLOPT_MAIL_FROM               7.20.0
 CURLOPT_MAIL_RCPT               7.20.0
 CURLOPT_MAIL_RCPT_ALLLOWFAILS   7.69.0
+CURLOPT_QUICK_EXIT                7.85.0
```
bagder, line 680: here as well

bagder (commented):
This PR also needs a rebase + force-push to address merge conflicts.

src/tool_operate.c
ferrieux: Ok, 2b813a9

ferrieux:
> This PR also needs a rebase + force-push to address merge conflicts.

DONE

docs/libcurl/opts/CURLOPT_QUICK_EXIT.3
ferrieux: DONE

docs/libcurl/symbols-in-versions
ferrieux: DONE

ferrieux:
Appveyor tests 714 and 715 fail only in the CI, in Windows. They are OK in local "make test" on Linux. Dunno whether it is specific, or master fails too .  Any help appreciated .

ferrieux:
Just spent a couple of hours chasing CI failures that ended up crippling master too :(
Is there ongoing work on a test suite cleanup ?
Note the inspection of the master's CI failures is not enough, as there are some Heisenbugs.
For example, 2302 (Websockets) clearly has a race condition, that triggers an assert typically only on first run after a full make (in local "TFLAGS=2302 make test-full" on master). You might want to have a closer look ;)

src/tool_operate.c
```diff
@@ -1229,6 +1229,17 @@ static CURLcode single_transfer(struct GlobalConfig *global,
 
         use_proto = url_proto(per->this_url);
 
+#ifndef EXIT_MAY_BE_LEAKY
```
bagder, line 1232: Isn't this name the other way around? If this is defined, then it will not leak. But perhaps more importantly: since nothing sets this define maybe we should just not use it and let someone who actually needs it add it later?

bagder (changes requested):
Can you please rebase and fix the merge conflict? I figure we need a test case for this as well, but I can volunteer to write one.

src/tool_operate.c
ferrieux: You're obviously right, sorry. Removed the ifdef as suggested.

ferrieux:
Resolved. And yes, I'd appreciate help on the tests: not only the new one testing the fix itself, but also the failing ones, some of which may be shared with master (e.g. the websocket one mentioned above).

.github/scripts/spellcheck.words
```diff
@@ -71,6 +71,7 @@ buildbot
 buildconf
 Caddy
 calloc
+cancelling
```
bagder, line 74: This is not how it is spelled!

bagder (approved):
I'll fix the typo as I merge.

bagder:
Thanks!

ferrieux:
Thanks a lot Daniel.

.github/scripts/spellcheck.words
cthart: Cancelling, travelling, modelling, ... that's how I spell them. Yet another difference between British and American English. https://english.stackexchange.com/questions/338/when-is-l-doubled

PR #9152, opened 2022-07-13
curl: writeout: fix repeated header outputs

### LINKED ISSUE #9150: `-w %header{name}` prints broken string on the second request
### I did this

```
$ curl https://github.com/curl/curl -s -o /dev/null -w "%header{date}\n"
Wed, 13 Jul 2022 07:03:27 GMT
```
but,
```
$ curl https://github.com/curl/curl#[1-2] -s -o /dev/null -w "%header{date}\n"
Wed, 13 Jul 2022 07:03:44 GMT
%header{date
```

### I expected the following

```
$ curl https://github.com/curl/curl#[1-2] -s -o /dev/null -w "%header{date}\n"
Wed, 13 Jul 2022 07:03:44 GMT
Wed, 13 Jul 2022 07:03:44 GMT
```

### curl/libcurl version

curl 7.84.0 (x86_64-pc-cygwin) libcurl/7.84.0 OpenSSL/1.1.1q zlib/1.2.12 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.2) libssh2/1.10.0 nghttp2/1.48.0
Release-Date: 2022-06-27

### operating system

MSYS_NT-10.0-19044 hostname 3.3.5-341.x86_64 2022-07-04 21:33 UTC x86_64 Msys
on
Windows 10 Pro 21H2 19044.1766

bagder:
The function stores a terminating zero in the buffer for convenience, but
when on repeated calls that would cause problems. Starting now, the original
content is put back again and repeated calls work fine.

Fixes #9150
Reported-by: highmtworks on github

### DIFF
--- a/src/tool_writeout.c
+++ b/src/tool_writeout.c
@@ -338,8 +338,8 @@ void ourWriteOut(const char *writeinfo, struct per_transfer *per,
       else {
         /* this is meant as a variable to output */
         char *end;
+        size_t vlen;
         if('{' == ptr[1]) {
-          char keepit;
           int i;
           bool match = FALSE;
           end = strchr(ptr, '}');
@@ -348,10 +348,10 @@ void ourWriteOut(const char *writeinfo, struct per_transfer *per,
             fputs("%{", stream);
             continue;
           }
-          keepit = *end;
-          *end = 0; /* null-terminate */
+          vlen = end - ptr;
           for(i = 0; variables[i].name; i++) {
-            if(curl_strequal(ptr, variables[i].name)) {
+            if((strlen(variables[i].name) == vlen) &&
+               curl_strnequal(ptr, variables[i].name, vlen)) {
               match = TRUE;
               switch(variables[i].id) {
               case VAR_ONERROR:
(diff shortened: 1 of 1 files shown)


jay:
Though we already do this save and restore elsewhere and it works fine I really don't like it, it is not correct because ptr is derived from `const char *` argument and we are stripping the const. It just happens to work the way we use it now since we aren't passing const. I think the easiest way to remedy this would be update the prototype and I've added a commit for that.

bagder:
I don't think we should remove the const but rather consider making the code not write to it.

bagder:
I now made the function not touch the passed in buffer.

src/tool_writeout.c
```diff
@@ -380,21 +380,26 @@ void ourWriteOut(const char *writeinfo, struct per_transfer *per,
             }
           }
           if(!match) {
-            fprintf(stderr, "curl: unknown --write-out variable: '%s'\n", ptr);
+            fprintf(stderr, "curl: unknown --write-out variable: '%.*s'\n",
+                    (int)vlen, ptr);
           }
           ptr = end + 1; /* pass the end */
```
jay, line 386: typo here as well, suggest remove comment, otherwise lgtm

jay (approved):

src/tool_writeout.c
bagder: But the comment is unrelated to this change, I prefer not to modify it (now)

Issue #9128, opened 2022-07-09
The command line arguments '-u user:password' are not hidden


bagder:
### Discussed in https://github.com/curl/curl/discussions/9127

<div type='discussions-op-text'>

<sup>Originally posted by **LitterWhite** July  9, 2022</sup>
I seem to have found a bug.

When starting curl, I pass username and password via command line, like this:
```shell
curl -u user:password ...
```
Then check the cmdline of curl by ps and expect user name and password are hidden,like this:
```shell
curl -u              ...
```
But they are not.

I check the code and found the arg of 'user:password' will be clean at getparameter()

src/tool_getparam.c: getparameter()
```c
case 'u':
      /* user:password  */
      GetStr(&config->userpwd, nextarg);
      cleanarg(nextarg);
      break;
```

src/tool_paramhlp.c: cleanarg()
```c
void cleanarg(char *str)
{
#ifdef HAVE_WRITABLE_ARGV
  /* now that GetStr has copied the contents of nextarg, wipe the next
   * argument out so that the username:password isn't displayed in the
   * system process list */
  if(str) {
    size_t len = strlen(str);
    memset(str, ' ', len);
  }
#else
  (void)str;
#endif
}
```

However, the pointer 'nextarg' pass to cleanarg is not the argv[] from main(argc, argv), it is copyed from argv[] at parse_args()

src/tool_getparam.c: parse_args()
```c
        char *nextarg = (i < (argc - 1))
          ? curlx_convert_tchar_to_UTF8(argv[i + 1])
          : NULL;

        result = getparameter(orig_opt, nextarg, &passarg, global, config);

        curlx_unicodefree(nextarg);
```

lib/curl_multibyte.h
```c
#if defined(UNICODE) && defined(WIN32)

#define curlx_convert_UTF8_to_tchar(ptr) curlx_convert_UTF8_to_wchar((ptr))
#define curlx_convert_tchar_to_UTF8(ptr) curlx_convert_wchar_to_UTF8((ptr))

typedef union {
  unsigned short       *tchar_ptr;
  const unsigned short *const_tchar_ptr;
  unsigned short       *tbyte_ptr;
  const unsigned short *const_tbyte_ptr;
} xcharp_u;

#else

#define curlx_convert_UTF8_to_tchar(ptr) (strdup)(ptr)
#define curlx_convert_tchar_to_UTF8(ptr) (strdup)(ptr)
```

Thus, the user name and password in argv are not cleaned up.

The last version is this:

[...]

PR #9154, opened 2022-07-14
tool_getparam: fix cleanarg for wide characters in Windows

jay:
Follow-up to bf7e887 which fixed cleaning sensitive arguments from the
command line in Windows. That fix (if it works) only works for Windows
multibyte builds (char) and not Windows Unicode builds (wchar_t).

Ref: https://github.com/curl/curl/issues/9128

Closes #xxxx

---

This doesn't work the way I expect. HAVE_WRITABLE_ARGV isn't defined in my Visual Studio Unicode build. If I override that and the sensitive arguments are cleaned, in Process Explorer I can see the command line doesn't change. Is this supposed to work at all on Windows? @MarcelRaad

### DIFF
--- a/lib/curl_setup_once.h
+++ b/lib/curl_setup_once.h
@@ -456,10 +456,16 @@ typedef unsigned int bit;
 
 #ifdef __VMS
 #define argv_item_t  __char_ptr32
+#define argv_strlen_func  strlen
+#define argv_fill_char  ' '
 #elif defined(_UNICODE)
-#define argv_item_t wchar_t *
+#define argv_item_t  wchar_t *
+#define argv_strlen_func  wcslen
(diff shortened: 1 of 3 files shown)


bagder:
I've tried to google this, and I can't find any source that claims this trick actually works on Windows. On the contrary, most places I find agree on that this isn't possible to achieve on Windows.

jay:
As discussed in #9195, Windows does not use cleanarg. It has writable argv but it doesn't matter because either way (char or wchar_t) they're not pointers into the command line, they're copies parsed from it so blanking them out does nothing.

Closing in favor of #9196 which addresses the compiler warnings with cleanarg and wchar_t.

PR #9165, opened 2022-07-16
doh: use https protocol by default

### LINKED ISSUE #9163: [Suggestion] `--doh-url` should not assert `http://` when omitted protocol
&#xA0;&#xA0;&#xA0;&#xA0;DNS over HTTPS hence only HTTPS would be plausible.
&#xA0;&#xA0;&#xA0;&#xA0;It's value handling should be special-cased.

&#xA0;&#xA0;&#xA0;&#xA0;Also, would it be proper to add an option "--doh" simply as the alias for "--doh-url"?
&#xA0;&#xA0;&#xA0;&#xA0;Or perhaps making "--doh" a boolean switch controlling whether to use DoH or not?

vszakats:
The only allowed protocol is https, so it makes sense to use that
by default if not passed explicitly by the user.

Fixes #9163
Closes #xxxx

### DIFF
--- a/lib/doh.c
+++ b/lib/doh.c
@@ -243,6 +243,7 @@ static CURLcode dohprobe(struct Curl_easy *data,
        the gcc typecheck helpers */
     struct dynbuf *resp = &p->serverdoh;
     ERROR_CHECK_SETOPT(CURLOPT_URL, url);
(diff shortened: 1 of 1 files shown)


jay (approved):
add a `Reported-by:` line. since this user does not have a name you'd use their handle and that it's github like "xxx@users.noreply.github.com" or "xxx on github"

PR #9170, opened 2022-07-17
h2h3: Fix bug overriding the 'TE: Trailers' header

### LINKED ISSUE #9171: curl fails to replace a 'TE: Trailers' header with 'te: trailers' when using HTTP/2
### I did this

Attempt to supply the `TE: Trailers` header when accessing an HTTP/2 server. For example:
```
curl -v --http2 -o /dev/null -H 'TE: Trailers' https://www.wikipedia.org
```

### I expected the following

Expected the connection to work. Instead received:
```
curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
```

The server closed the connection because it got the value 'Trailers' with an uppercase instead of 'trailers' with a lowercase.

### curl/libcurl version

`curl 7.85.0-DEV (x86_64-pc-linux-gnu) libcurl/7.85.0-DEV OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 nghttp2/1.40.0`

### operating system

`Linux ThinkPad-T14s-Gen-2i 5.14.0-1036-oem #40-Ubuntu SMP Mon May 9 09:15:08 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

lwthiker:
A 'TE: Trailers' header is explicitly replaced by 'te: trailers'
(lowercase) in Curl_pseudo_headers() when building the list of HTTP/2
or HTTP/3 headers. However, this is then replaced again by the original
value due to a bug, resulting in the uppercased version being sent. Some
HTTP/2 servers reject the whole HTTP/2 stream when this is the case.

Fixes #9171

### DIFF
--- a/lib/h2h3.c
+++ b/lib/h2h3.c
@@ -258,9 +258,6 @@ CURLcode Curl_pseudo_headers(struct Curl_easy *data,
       nva[i].valuelen = (end - hdbuf);
     }
 
-    nva[i].value = hdbuf;
-    nva[i].valuelen = (end - hdbuf);
-
     ++i;
   }
 


bagder (approved):

bagder:
Thanks!

PR #9173, opened 2022-07-17
lib3026: reduce the number of threads to 100

bagder:
Down from 1000, to make it run and work in more systems.


ericonr:
Confirmed to work for me!

MarcelRaad (approved):

PR #9187, opened 2022-07-20
curl-wolfssl.m4: add options header

neheb:
Needed for certain configurations of wolfSSL. Otherwise, missing header error may occur.

Tested with OpenWrt.


bagder (approved):

bagder:
Thanks!

PR #9196, opened 2022-07-23
tool_getparam: fix cleanarg() for unicode builds

### LINKED ISSUE #9195: build warnings with Visual Studio unicode builds since bf7e887b24
[build log](https://ci.appveyor.com/project/curlorg/curl/builds/44255779/job/3drakckomc27qqtt)

~~~
C:/projects/curl/src/tool_getparam.c: In function 'parse_args':
C:/projects/curl/src/tool_getparam.c:2450:17: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
           clear = argv[i + 1];
                 ^
[ 87%] Building C object src/CMakeFiles/curl.dir/tool_help.c.obj
make[2]: *** [src/CMakeFiles/curl.dir/build.make:294: src/CMakeFiles/curl.dir/tool_getparam.c.obj] Error 1

bagder:
Use the correct type, and make cleanarg an empty macro if the cleaning
ability is absent.

Fixes #9195

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -541,24 +541,26 @@ static ParameterError GetSizeParameter(struct GlobalConfig *global,
   return PARAM_OK;
 }
 
-static void cleanarg(char *str)
-{
 #ifdef HAVE_WRITABLE_ARGV
+static void cleanarg(argv_item_t str)
+{
   /* now that GetStr has copied the contents of nextarg, wipe the next
    * argument out so that the username:password isn't displayed in the
    * system process list */
   if(str) {
-    size_t len = strlen(str);
+    size_t len = strlen((char *)str);
     memset(str, ' ', len);
   }
-#else
-  (void)str;
-#endif
 }
+#else
+#define cleanarg(x)
+#endif
 
 ParameterError getparameter(const char *flag, /* f or -long-flag */
                             char *nextarg,    /* NULL if unset */
-                            char *clearthis,
+#ifdef HAVE_WRITABLE_ARGV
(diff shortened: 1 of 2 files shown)


src/tool_getparam.c
```diff
@@ -541,24 +541,24 @@ static ParameterError GetSizeParameter(struct GlobalConfig *global,
...
-{
 #ifdef HAVE_WRITABLE_ARGV
+static void cleanarg(argv_item_t str)
+{
   /* now that GetStr has copied the contents of nextarg, wipe the next
    * argument out so that the username:password isn't displayed in the
    * system process list */
   if(str) {
-    size_t len = strlen(str);
+    size_t len = strlen((char *)str);
```
MarcelRaad, line 551: Doesn't this only work coincidentally because `HAVE_WRITABLE_ARGV` is never set when `wchar_t` is used and otherwise the terminating null character would be found too early for ASCII characters?
Maybe just leave out the cast and change the signature back to `char *` if we can safely assume that this function is never used for `wchar_t`, or otherwise use `wcslen` in a `wchar_t` overload.

bagder: I see your point.

But with that fix, we would instead get a warning because we pass in an `argv_item_t` to a `char *` parameter, so it would still need a typecast somewhere.

jay: Do any of the builds have strong type checking? The only other type at issue here is VMS __char_ptr32. Removing the cast could help us out with a warning if something like this happens again.

jay (approved):

Issue #9086, opened 2022-07-03
HTTP/3: Windows build sends lots of empty UDP datagrams


tatsuhiro-t:
### I did this

Compile curl windows build (mingw64 works) and contact HTTP/3 server with --http3 option.
You need to either check server log or capture the packets to see the empty UDP datagrams.
The number of UDP datagrams varies, but I observed ~500.

It looks like the empty UDP datagram is sent by this call: https://github.com/curl/curl/blob/0defae2fe524230f8b818d406d19e56f360bcc54/lib/multi.c#L1238

which is introduced via https://github.com/curl/curl/commit/b36442b24305f3cda7c13cc64b46838995a4985b

### I expected the following

curl should not send empty UDP datagrams on HTTP/3 connection because such UDP datagram is invalid on QUIC specification.

### curl/libcurl version

master branch
[curl -V output]

jay:
@mback2k

bagder:
That's an awful lot of send()s...

bagder:
They were added for performance reasons (but not tested for h3 I bet). What exactly happens if we don't send() there?

mback2k:
If we don't send, then we won't get another write-ready event from select if we don't send in between do calls, effectively blocking till the next readiness event. I guess we could try to just use this workaround with TCP, as UDP should always be ready for writing/sending.

bagder:
> If we don't send, then we won't get another write-ready event from select

From **WSAEventSelect** then I presume? select() and poll() are not edge-trigged, they *should* return writable for as long as it is writable.

> we could try to just use this workaround with TCP, as UDP should always be ready for writing/sending,

Yes we should, because send()ing zero bytes might be an almost no-op with TCP but as @tatsuhiro-t figured out: it clearly is much more than that for UDP.

bagder:
It doesn't look very easy to add that check though... Can we back-peddle the use of WSAEventSelect() there ?

mback2k:
> > If we don't send, then we won't get another write-ready event from select
> 
> From **WSAEventSelect** then I presume? select() and poll() are not edge-trigged, they _should_ return writable for as long as it is writable.

Yes, sorry, I meant **WSAEventSelect** in this context of course, but since I was responding via the mobile app I didn't have time and context to fully type it out.

mback2k:
> It doesn't look very easy to add that check though... Can we back-peddle the use of WSAEventSelect() there ?

[...]

PR #9203, opened 2022-07-25
Fix for HTTP/3: Windows build sends lots of empty UDP datagrams

mback2k:
Attempt to #9086 by: fix sending empty UDP packets to reset FD_WRITE

On Windows the internal state of FD_WRITE as returned from
WSAEnumNetworkEvents is only reset after successful send().
Since multi_wait can be called multiple times without such
a send() happening in between calls, we need to call send()
ourselves to make sure we still get a fresh FD_WRITE signal.

For TCP connections an empty send() does not actually send
something on the wire, but for UDP an empty message packet
is actually send on the wire, leading to hundreds of such.

This commit makes the send()-workaround apply only to TCP.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1176,6 +1176,9 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
   bool ufds_malloc = FALSE;
 #ifdef USE_WINSOCK
   WSANETWORKEVENTS wsa_events;
+  DWORD stype;
+  int ltype = sizeof(stype);
+  char *ptype = (char *)&stype;
   DEBUGASSERT(multi->wsa_event != WSA_INVALID_EVENT);
 #endif
 #ifndef ENABLE_WAKEUP
@@ -1274,7 +1277,10 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
           s = sockbunch[i];
 #ifdef USE_WINSOCK
           mask |= FD_WRITE|FD_CONNECT|FD_CLOSE;
-          send(s, NULL, 0, 0); /* reset FD_WRITE */
+          if(getsockopt(s, SOL_SOCKET, SO_TYPE, ptype, &ltype) == SOCKET_ERROR)
+            stype = SOCK_STREAM; /* assume TCP */
+          if(stype == SOCK_STREAM)
+            send(s, NULL, 0, 0); /* reset FD_WRITE */
 #endif
           ufds[nfds].fd = s;
           ufds[nfds].events = POLLOUT;
@@ -1300,6 +1306,7 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
 
   /* Add external file descriptions from poll-like struct curl_waitfd */
   for(i = 0; i < extra_nfds; i++) {
+    curl_socket_t s = extra_fds[i].fd;
 #ifdef USE_WINSOCK
     long mask = 0;
     if(extra_fds[i].events & CURL_WAIT_POLLIN)
@@ -1308,15 +1315,18 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
       mask |= FD_OOB;
     if(extra_fds[i].events & CURL_WAIT_POLLOUT) {
       mask |= FD_WRITE|FD_CONNECT|FD_CLOSE;
-      send(extra_fds[i].fd, NULL, 0, 0); /* reset FD_WRITE */
+      if(getsockopt(s, SOL_SOCKET, SO_TYPE, ptype, &ltype) == SOCKET_ERROR)
+        stype = SOCK_STREAM; /* assume TCP */
+      if(stype == SOCK_STREAM)
+        send(s, NULL, 0, 0); /* reset FD_WRITE */
     }
-    if(WSAEventSelect(extra_fds[i].fd, multi->wsa_event, mask) != 0) {
+    if(WSAEventSelect(s, multi->wsa_event, mask) != 0) {
       if(ufds_malloc)
         free(ufds);
       return CURLM_INTERNAL_ERROR;
     }
 #endif
-    ufds[nfds].fd = extra_fds[i].fd;
+    ufds[nfds].fd = s;
     ufds[nfds].events = 0;
     if(extra_fds[i].events & CURL_WAIT_POLLIN)
       ufds[nfds].events |= POLLIN;
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1166,6 +1166,8 @@ static DWORD multi_getsocktype(struct Curl_multi *multi, curl_socket_t s)
   int res;
 
   struct Curl_sh_entry *entry = sh_getentry(&multi->sockhash, s);
+  if(!entry)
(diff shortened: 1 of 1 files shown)


mback2k:
@tatsuhiro-t hey, can you please test drive this workaround for me and compare the performance before and after this PR?

tatsuhiro-t:
I saw almost no difference in between.  I also saw no difference when I tested HTTP/1.1 upload with 0 bytes send() commented out, so I might not have a good environment to test this thing.

mback2k:
The test upload should be at least something around 100 MB to notice a difference.

tatsuhiro-t:
I actually used 100MiB file.

bagder:
Does this calling of getsockopt() on all the sockets risk slowing things down?

It struck me that as we already have a "socket hash table" with all the sockets in use within the multi handle, it could easily be extended to have the socket type in there as well. It could then make it a hash table lookup instead, if this getsockopt() approach turns out unfavorable.

mback2k:
> Does this calling of getsockopt() on all the sockets risk slowing things down?

I would think so, yes. But unfortunately, it is currently quite hard to test with the state of uploads using HTTP/3.

> It struck me that as we already have a "socket hash table" with all the sockets in use within the multi handle, it could easily be extended to have the socket type in there as well. It could then make it a hash table lookup instead, if this getsockopt() approach turns out unfavorable.

Good idea! Would you mind taking a look into this or pointing me into the right direction? Where would the hash table value be initialized and stored?

bagder:
The function `multi.c:sh_getentry()` returns a `struct Curl_sh_entry` from the socket hash given a socket that is used by libcurl. That struct could probably be extended to also contain sock type, maybe even lazily initialized so you could do the getsockopt() in case it doesn't know and then store the result in case you need to know again for this socket later.

mback2k:
@bagder I attempted to implement the caching, but with the code committed in 7ef77d2833c5aad99dc7511f4f55cbf29594788e including debug fprintf statements, I always get `entry=0`, so the checked socket is not already in the hashmap. Should I add it in that case?

Also I observed that `multi_wait` is only used to check a socket for write readiness with the HTTP/3 protocol up until the request headers are sent. After that, I no longer get any `entry=0` line, but just `* ngh3_stream_recv returns 0 bytes and EAGAIN` lines.

For TFTP and pre-HTTP/3 uploads, I get `entry=0` lines throughout the transfer as somewhat expected. @tatsuhiro-t

mback2k:
@bagder @tatsuhiro-t I could really use your help and insight on this topic. Thanks in advance!

mback2k:
Force-pushed with a new commit adding the socket to the hash table if it wasn't there yet:
https://github.com/curl/curl/blob/edf8ef992188d2d0ce673583b1f72daf8897b556/lib/multi.c#L1170-L1171

Let's see what CI will tell us if this is the right approach...

mback2k:
CI failures are unrelated to the socket hash table. But @bagder , I still have doubts about adding sockets from extra_fds to that hash table belonging to a multi handle. Since we are not managing the extra_fds within libcurl, but the calling application does, those socket handles/numbers can change over time and eventually get re-used by a connection of a different type. So we have no way to invalidate our cache for these sockets and may never notice the changed socket type. This means we probably should only cache the socket's connection type for _our_ sockets, right?

mback2k:
@tatsuhiro-t although this still contains debug code, can you review the impact of this PR again please?

tatsuhiro-t:
It works for me.  No empty packets were observed.

mback2k:
@tatsuhiro-t thanks a lot! @bagder does the logic in this PR make sense to you now?

mback2k:
I am still not sure if using `sh_addentry` this way is correct, but I am now requesting review on this PR in general.

mback2k:
> The regular fd part if I recall the sockets should already have entries, and if they don't I can't readily tell if it's ok to create one without setting the rest of the entry as would be done in singlesocket. So, perhaps getsockopt should be added in singlesocket [here](https://github.com/curl/curl/blob/curl-7_85_0/lib/multi.c#L2897) and then multi_getsocktype return the type but not set it lazily

In all my tests the entry wasn't existing for the socket in the regular fd part at the time multi_getsocktype was called. So I guess then I just probably just stay with always calling getsockopt without a caching layer.

jay:
Interesting. That's not what I expected. In that case I'd wait for @bagder's input on this.

bagder:
The socket gets added to the sockhash on:

https://github.com/curl/curl/blob/aa970c4c08775afcd0c2853be89b0a6f02582d50/lib/multi.c#L2879

I have not analyzed why the sockets you check are not present in that hash. It seems curious to me.

mback2k:
Thanks to both of you. I tried to incorporate your feedback, @jay.

lib/multi.c
```diff
@@ -1155,6 +1158,26 @@ CURLMcode curl_multi_fdset(struct Curl_multi *multi,
...
+    return SOCK_STREAM;
+  if(entry->socketype)
+    return entry->socketype;
+  if(getsockopt(s, SOL_SOCKET, SO_TYPE, (char *)&st, &sl) == SOCKET_ERROR)
+    return SOCK_STREAM;
+
+  entry->socketype = st;
+  return st;
+}
+#endif
```
jay, line 1179: I think the function could use a comment that it is only to be called on libcurl sockets and not the extra sockets, and also explain why it defaults to SOCK_STREAM when  no entry is available or the socket type cannot be obtained. Conversely the second check sets the type to 0 when it cannot be obtained

mback2k: I think this function *can* be called for extra sockets as they may overlap with libcurl sockets. We cannot assume that they are fully distinct or can we?

jay: I encourage you to test your caching method against getsockopt on 1000 sockets so we can see if it has any performance gain and is worth pursuing. For example
~~~c
  if(getsockopt(s, SOL_SOCKET, SO_TYPE, (char *)&st, &sl) != SOCKET_ERROR &&
     st == SOCK_STREAM) {
    send(s, NULL, 0, 0); /* reset FD_WRITE */
  }
~~~

icing:
I would propose to beef up the `getsock()` interfaces to collect not only the `sockfd` and read/write flags, but also the type. The code which opened the socket does know what type it is. So, there is really no need to ask for that in `multi_wait()`.

Since this is not done in a day, I think we could merge this PR and then do the refactoring afterwards, eliminating `multi_getsocktype()` again.

mback2k:
Unfortunately, curl does not open the sockets in extra_fds. But yes, I hope to get to this again over the holidays period. Currently to busy at work, sorry.

mback2k:
Still too busy with life and work, sorry. Maybe @icing or @jay could take over from here?

My curl time will be very limited till much later this year, sorry.

jay:
See #10430 for my take on this, which is calling getsockopt directly without caching, and then reset FD_WRITE only if TCP. It's unclear to me if extra user fd's that could possibly? be neither TCP nor UDP would need to reset FD_WRITE.

PR #9207, opened 2022-07-26
tool_operate: make --doh-url "" switch it off

bagder:
A possible future addition could be to parse the URL first too to verify
that it is valid before trying to use it.

### DIFF
--- a/docs/cmdline-opts/doh-url.d
+++ b/docs/cmdline-opts/doh-url.d
@@ -16,4 +16,7 @@ name lookups take place over SSL. However, the certificate verification
 settings are not inherited and can be controlled separately via
 --doh-insecure and --doh-cert-status.
 
+This option is unset if an empty string "" is used as the URL. (Added in
+7.85.0)
+
 If this option is used several times, the last one will be used.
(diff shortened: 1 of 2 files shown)


jay:
I'm skeptical about doing this with strings because we may have logic elsewhere that needs to be addressed (as is the case here, see the squashme for env logic) and that can lead to bugs. An alternative would be handle it during parsing,

~~~diff
diff --git a/src/tool_getparam.c b/src/tool_getparam.c
index 9bbd51d..0aec45d 100644
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -700,7 +700,10 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
           return err;
         break;
       case 'C': /* doh-url */
-        GetStr(&config->doh_url, nextarg);
+        if(nextarg && *nextarg)
+          GetStr(&config->doh_url, nextarg);
+        else /* this option is documented as unset by an empty string */
+          Curl_safefree(config->doh_url);
         break;
       case 'd': /* ciphers */
         GetStr(&config->cipher_list, nextarg);
~~~

bagder:
Yeah, setting it to NULL again is clearly a better approach.

jay (approved):

PR #9211, opened 2022-07-26
improve/freebsd-ci: Bootstrapping pip installer

pheiduck:
This PR improves the maintenance of freebsd-ci.
Since py310-pip does not exist. we have to use the "main" python pkg.
So we can use python 3.10.x right now.
The idea is to reduce the maintenance effort and keep the CI stable as long as possible. ;)
@bagder @jay @danielgustafsson


pheiduck:
Works with `python -m ensurepip --default-pip`
Ready for Review.

PR #6341, opened 2020-12-17
connect: On linux, enable reporting of all ICMP errors on UDP sockets

crrodriguez:
The linux kernel does not report all ICMP errors back to userspace
due to historical reasons.
IP*_RECVERR sockopt must be turned on to have the correct behaviour
which is to pass all ICMP errors to userspace.

See
https://bugzilla.kernel.org/show_bug.cgi?id=202355<

PS: You almost certainly want this option to be set in your DNS resolver too (ares, or whatever, glibc was already fixed)

PR #9215, opened 2022-07-26
connect: revert the use of IP*_RECVERR

### LINKED ISSUE #9209: HTTP/3: Busy loop due to IP_RECVERR
### I did this

Upload a large file with HTTP/3.
Lots of POLLIN event occurs but recvfrom read nothing.

It is caused by IP_RECVERR socket option [1] set for UDP socket.  ngtcp2 does UDP Path MTU Discovery and it might send larger packet than the path can sustain.  In this case ICMP message is received, it is resulted in POLLERR, which is translated into POLLIN by curl code.  Because IP_RECVERR, the normal recvfrom call does not clear error queue, thus poll keeps returning POLLERR.
It is only cleared by recvmsg with MSG_ERRQUEUE.  But it looks like there is no code in curl to do this.

[1] https://github.com/curl/curl/blob/e28edb6678cd02aea7fc73340c223f26dba8f5b7/lib/connect.c#L1670-L1682

### curl/libcurl version

curl master branch with ngtcp2 QUIC backend
[curl -V output]

### operating system

Linux

bagder:
The options were added in #6341 and d13179d, but cause problems: Lots of
POLLIN event occurs but recvfrom read nothing.

Reported-by: Tatsuhiro Tsujikawa
Fixes #9209

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -1667,20 +1667,6 @@ CURLcode Curl_socket(struct Curl_easy *data,
   }
 #endif
 
-#if defined(__linux__) && defined(IP_RECVERR)
-  if(addr->socktype == SOCK_DGRAM) {
-    int one = 1;
-    switch(addr->family) {
-    case AF_INET:
-      (void)setsockopt(*sockfd, SOL_IP, IP_RECVERR, &one, sizeof(one));
-      break;
-    case AF_INET6:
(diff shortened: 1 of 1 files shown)


jay:
@crrodriguez

bagder:
@crrodriguez already commented on this idea [here](https://github.com/curl/curl/issues/9209#issuecomment-1196029978)

bagder:
Also the original patch only fixed some theoretical problems, nothing that was ever actually reported to hurt curl.

PR #9216, opened 2022-07-27
hyper: enable obs-folded multiline headers

seanmonstar:
This enables the option in hyper to allow obs-folded multiline headers, which makes test 1940 pass.

---

I was hopeful this would fix test 1274, but it seems _that_ test is expecting to receive the bytes _before_ the value is unfolded, and hyper unfolds as part of the parsing step. The header value is still received, just without the newlines. So a curl user depending on a folded header could still work, if they don't blindly unfold.

We can either edit that test directly to allow the expected output to be unfolded if hyper is enabled, or disable the test directly.


bagder (approved):

bagder:
Yeah, we probably need to edit the 1274 test a little to make it run with hyper.

bagder:
Thanks!

PR #9217, opened 2022-07-28
hyper: customize test1274 to how hyper unfolds headers

seanmonstar:
I tried following the way I see some other test data files do this, but after I made the changes locally and re-run `./runtests.pl 1274`, it still sees the old output. Perhaps there's a step in `make test` that I'm skipping locally?

### DIFF
--- a/tests/data/DISABLED
+++ b/tests/data/DISABLED
@@ -72,7 +72,6 @@
 # 1021 re-added here due to flakiness
 1021
 1117
-1274
 1417
 1533
 1540
--- a/tests/data/test1274
+++ b/tests/data/test1274
@@ -13,14 +13,22 @@ header line folding
 <data>
 HTTP/1.1 200 OK
 Date: Tue, 09 Nov 2010 14:49:00 GMT
+%if hyper
+Server: test-server/ fake folded
+%else
 Server: test-server/
  fake
  folded
+%endif
 Last-Modified: Tue, 13 Jun 2000 12:10:00 GMT
 ETag: "21025-dc7-39462498"
 Content-Length: 6
+%if hyper
+Connection: close
+%else
 Connection:                                              
    close
+%endif
 
 -foo-
 </data>


bagder:
Hm, curious. I agree that it looks weird as the change seems fine. I need to look into it.

bagder:
The issue here is quite simply that the built-in code does no unfolding of the headers that it stores with `-D`, the only unfolding it does is when storing the headers for the headers API.

bagder:
I propose we make a conditional in the test case, like this:
~~~diff
index e789a7232..785f74768 100644
--- a/tests/data/test1274
+++ b/tests/data/test1274
@@ -59,17 +59,25 @@ Accept: */*
 
 </protocol>
 <file name="log/out%TESTNUMBER">
 HTTP/1.1 200 OK
 Date: Tue, 09 Nov 2010 14:49:00 GMT
+%if hyper
+Server: test-server/ fake folded^M
+%else
 Server: test-server/
  fake
  folded
+%endif
 Last-Modified: Tue, 13 Jun 2000 12:10:00 GMT
 ETag: "21025-dc7-39462498"
 Content-Length: 6
+%if hyper
+Connection: close^M
+%else
 Connection:                                              
    close
+%endif
 
 </file>
 </verify>
 </testcase>

seanmonstar:
I _think_ that's what I did in this PR, but maybe I'm missing an important detail.

bagder:
Ah, yes now I see it, you edited the `<data>` part with ifdefs which is part that the server uses to serve data with - although it also doubles and is also used to check that was received (stored really) was exactly what was sent. In this case, with hyper, what is sent is not exactly what is stored really.

New edit/proposal attached. Gzipped to make the line endings stick. This version makes the test work for me.

[test1274.gz](https://github.com/curl/curl/files/9300561/test1274.gz)

bagder:
If you have no objections I can merge it like that.

seanmonstar:
Ohhh, I see now. I had done them in the wrong section. Thanks for clarifying! Sure, no objections here :)

bagder:
Thanks!

PR #9228, opened 2022-07-30
Add details to error: "unable to set client certificate"

opensignature:
error "unable to set client certificate" is, imho, too generic.
I would add the message from openssl.
In my case:
from: "curl: (58) unable to set client certificate"
to: curl: (58) unable to set client certificate [error:0A00018F:SSL routines::ee key too small]

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -872,8 +872,9 @@ int cert_stuff(struct Curl_easy *data,
           }
 
           if(SSL_CTX_use_certificate(ctx, params.cert) != 1) {
-            failf(data, "unable to set client certificate");
-            X509_free(params.cert);
+            failf(data, "unable to set client certificate [%s]",
+                  ossl_strerror(ERR_get_error(), error_buffer,
+                                sizeof(error_buffer)));
             return 0;
           }
           X509_free(params.cert); /* we don't need the handle any more... */


dfandrich:
Isn't the OpenSSL detailed error displayed as a debug log?

emilengler:
At that point I think it would be better to write a wrapper function for `failf` to always include the OpenSSL error instead of this patch at just one place in the code.

opensignature:
> Isn't the OpenSSL detailed error displayed as a debug log?

"debug log" is for developers. I think it is easier for a user to try to interpret, for example, the message "ee key too small" that enable debug log.
Consider that with the introduction of openssl version 1.1.1 several packages have become buggy.
"dh key too small", "ee key too small", "ca md too weak" caused by the SECLEVEL 2 setting the security level to 112 bit.
RSA and DHE keys need to be at least 2048 bit long but still today many keys (for example those in the smartcards used to authenticate) are 1024 bit long.

bagder:
>  to: curl: (58) unable to set client certificate [error:0A00018F:SSL routines::ee key too small]

That's a horribly user hostile way to phrase the error, but I can't really think of how it can be improved easily... :frowning_face:

dfandrich:
On Sun, Jul 31, 2022 at 02:30:33AM -0700, opensignature wrote:
> "debug log" is for developers.

Debug log is for anyone who needs more information about an error than is
displayed in the error messages. There are plenty of cases where this is the
case.

opensignature:
> Debug log is for anyone who needs more information about an error than is displayed in the error messages

ok, sorry, maybe I don't know curl very well but which debug log are you referring?

"This curl uses a libcurl built with Debug. This enables more error-tracking
and memory debugging etc. For curl-developers only!" [1]

[1] https://github.com/curl/curl/blob/df45fd1794b1ff3052a96a39c1fccfc8bc5ec6a9/docs/cmdline-opts/version.d#L31

dfandrich:
On Mon, Aug 01, 2022 at 01:21:18AM -0700, opensignature wrote:
> ok, sorry, maybe I don't know curl very well but which debug log are you
> referring?

It's also called the verbose log. It's the one you get with the -v option on
the command-line or the CURLOPT_DEBUGFUNCTION in libcurl.

> "This curl uses a libcurl built with Debug. This enables more error-tracking
> and memory debugging etc. For curl-developers only!" [1]

That's something completely different. That refers to enabling a debug build
with the --enable-debug configure option.

opensignature:
> It's also called the verbose log. It's the one you get with the -v option on the command-line or the CURLOPT_DEBUGFUNCTION in libcurl.

With the -v option curl shows "unable to set client certificate" only.
For this reason I proposed PR.
in https://github.com/curl/curl/blob/9153ba708be87ed6e7c25e1b4864f86fadeb95ad/lib/vtls/openssl.c
there are 106 calls to failf, of these only about twenty integrate the error with the message of the openssl library

bagder:
thanks!

PR #9229, opened 2022-07-30
docs: add dns category to --resolve

emilengler:
This commit adds the dns category to the --resolve command line option,
because it can be interpreted as both: a low-level connection option and
an option related to the resolving of a hostname.

It is also not common for dns options to belong to the connection
category and vice versa.  --ipv4 and --ipv6 are both good examples.


bagder (approved):

bagder:
Thanks!

PR #9231, opened 2022-07-30
cmdline-opts/gen.pl: improve speed

emanuele6:
The character capure and the (`\W`) capture substituted with themselves is unnecessary. I replaced them with a lookahead assertion and a `\b` assertion.

I also quoted the injection of $k in the regex since the unquoted injection was not intended and could potentially cause problems.

This improved the speed of

 ./gen.pl mainpage *.d > /dev/null

from 12.3 seconds to 0.6 seconds on my PC!

Reported-by: Emil Engler <me@emilengler.com>
Fixes https://github.com/curl/curl/issues/9230
Closes https://github.com/curl/curl/pull/9231

Issue #8231, opened 2022-01-06
Wrong coding


7-904-001-67-28:
.\curl.exe -I "https://rr3---sn-5hneknee.googlevideo.com/videoplayback?expire=1641445341&ei=fSPWYcq6L8i4mLAPldaKoAY&ip=78.110.173.172&id=o-AD59IMs80-W1AInJlj538lj22yr6_UQJGYicDi7m0T9W&itag=22&source=youtube&requiressl=yes&vprv=1&mime=video%2Fmp4&ns=xuag_3aEIBW3xH2evJ6re2oG&cnr=14&ratebypass=yes&dur=3567.304&lmt=1641408177845900&fexp=23886206,24001373,24007246&beids=23886206&c=WEB&txp=5532434&n=13nYxvgU4PaM1w&sparams=expire%2Cei%2Cip%2Cid%2Citag%2Csource%2Crequiressl%2Cvprv%2Cmime%2Cns%2Ccnr%2Cratebypass%2Cdur%2Clmt&sig=AOq0QJ8wRgIhALtdKWZ0UzoGEGSfVA066Z6YN5o86wxVbDxjwfaNk_UrAiEAjXVzdMcvdiH0FZIfgMeSUpO9Mu2otWkV2rIRz-ksedk%3D&title=%D0%93%D0%BB%D0%BE%D0%B1%D0%B0%D0%BB%D1%8C%D0%BD%D0%BE%D0%B5%20%D0%BD%D0%B0%D1%81%D1%82%D1%83%D0%BF%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5%20%D0%B0%D0%B2%D1%82%D0%BE%D1%80%D0%B8%D1%82%D0%B0%D1%80%D0%B8%D0%B7%D0%BC%D0%B0%3F&redirect_counter=1&rm=sn-aigez67s&req_id=32ab84cd98cba3ee&cms_redirect=yes&ipbypass=yes&mh=pL&mip=190.2.132.214&mm=31&mn=sn-5hneknee&ms=au&mt=1641429420&mv=m&mvi=3&pl=22&lsparams=ipbypass,mh,mip,mm,mn,ms,mv,mvi,pl&lsig=AG3C_xAwRQIhAIqs7eddtpCQSd5fW8uQgQ5g0TmNl8WMgca7k4ojOPK4AiAd5qtVhFrxRIKXXxYn4rmAXAQhv_xvUJo7ZaHmHULi6A%3D%3D"
[...]

PR #9232, opened 2022-07-30
cmdline-opts/gen.pl: improve performance

### LINKED ISSUE #9230: `./gen.pl` is horribly slow
### I did this

```sh
engler@lennon:~/src/curl/docs/cmdline-opts$ time ./gen.pl mainpage *.d > /dev/null 
    1m49.23s real     0m54.19s user     0m53.80s system
```

### I expected the following

A faster generation of the man-page.

### curl/libcurl version

git master

### operating system

`OpenBSD lennon 7.1 GENERIC.MP#3 amd64`

emilengler:
On some systems, the gen.pl script takes nearly two minutes for the
generation of the main-page, which is a completely unacceptable time.

The commit author thinks, that the terrible performance comes from some
low-level locale checks regarding the use of "\W" in a regular
expression.

This commit replaces the "\W" with [^a-zA-Z0-9_], which is, according to
regex101.com, functionally equivalent to the previous operation, except
that it is obviously limited to ASCII, which is fine, as the curl
project is English-only anyway.

See #8299
Fixes #9230
Closes #XXXX

<hr>

cc @bagder

Does this still fixes the issue you've encountered six months ago? I wasn't exactly sure what your patch was meant to fix.

### DIFF
--- a/docs/cmdline-opts/gen.pl
+++ b/docs/cmdline-opts/gen.pl
@@ -103,10 +103,10 @@ sub printdesc {
             print ".fi\n"; # fill-in
         }
         # skip lines starting with space (examples)
-        if($d =~ /^[^ ]/) {
+        if($d =~ /^[^ ]/ && $d =~ /--/) {
             for my $k (keys %optlong) {
                 my $l = manpageify($k);
-                $d =~ s/--$k([^a-z0-9_-])(\W)/$l$1$2/;
+                $d =~ s/--\Q$k\E([^a-z0-9_-])([^a-zA-Z0-9_])/$l$1$2/;
             }
         }
         # quote "bare" minuses in the output


emanuele6:
@emilengler oh, i was working on a fix, I think we can avoid the capture entirely, see #9231

emilengler:
@emanuele6 Wow, this was pretty quick... I've replaced with a functionally equivalent thing but with a restriction to ASCII.

emilengler:
Here is a small benchmark comparing the two pull requests:

#9231
```sh
engler@lennon:~/src/curl/docs/cmdline-opts$ time ./gen.pl mainpage *.d > /dev/null
    0m32.87s real     0m26.59s user     0m04.31s system
```

#9232 (this one)
```sh
engler@lennon:~/src/curl/docs/cmdline-opts$ time ./gen.pl mainpage *.d > /dev/null 
    0m35.86s real     0m29.44s user     0m05.45s system
```

emilengler:
The output of this `./gen.pl mainpage *.d` has the same checksum as the slow one found in git master.

emanuele6:
Let's ignore #8231 and just use this PR.

Since you are at it, you could also quote the injection of `$k` as I sugested in the other PR.

I also think you can use `/e` to only run `manpageify when necessary. That reduce execution time from 10.3seconds (current patch on this PR) to 8.6seconds on my PC:

```diff
diff --git a/docs/cmdline-opts/gen.pl b/docs/cmdline-opts/gen.pl
index 3c37fcb4a..5bf427aa9 100755
--- a/docs/cmdline-opts/gen.pl
+++ b/docs/cmdline-opts/gen.pl
@@ -105,8 +105,7 @@ sub printdesc {
         # skip lines starting with space (examples)
         if($d =~ /^[^ ]/) {
             for my $k (keys %optlong) {
-                my $l = manpageify($k);
-                $d =~ s/--$k([^a-z0-9_-])([^a-zA-Z0-9_])/$l$1$2/;
+                $d =~ s/--\Q$k\E([^a-z0-9_-][^a-zA-Z0-9_])/manpageify($k) . $1/e;
             }
         }
         # quote "bare" minuses in the output
```

emilengler:
I think this is ready for merge.

dfandrich:
This discussion made me think of other optimizations that could be done to speed up the process. My take (which is orthogonal to this one) is in #9233.

bagder:
Maybe the logic should check if there is a single `--` occurrence in the line first before it loops around all the long options to try to find them? It feels like it could speed up the script much more than this...

A secondary fix would be to avoid the loop somehow.

emilengler:
> Maybe the logic should check if there is a single `--` occurrence in the line first before it loops around all the long options to try to find them? It feels like it could speed up the script much more than this...
> 
> A secondary fix would be to avoid the loop somehow.

The first idea is brilliant. It reduces the run-time even further to just 3 seconds on my system. I just fixed it.

The second is also pretty nice, but I just tried implementing it and it became somewhat weird when a line contained more than one opt (because of the quoting). I'll be leaving this up to someone else to implement.

bagder (approved):

PR #9233, opened 2022-07-31
gen.pl: Generate the option man page text multithreaded

dfandrich:
Doing it in parallel reduces the time take by 60% on an octacore system.
The number of threads is hard-coded to 8 which is a reasonable
compromise between memory and speed on current systems. Ideally, this
could be tuned to the number of CPU cores available.

### DIFF
--- a/docs/cmdline-opts/gen.pl
+++ b/docs/cmdline-opts/gen.pl
@@ -39,6 +39,20 @@ Unfortunately it seems some perls like msysgit can't handle a global input-only
 =end comment
 =cut
 
+# Load the Proc::ParallelLoop module only if it's available
+my $haveparallel = eval {
+  require Proc::ParallelLoop;
+  Proc::ParallelLoop->import();
+  1;
+};
+if(!$haveparallel) {
+  # Stub to avoid a syntax error when the module isn't loaded
+  sub pareach {
+    die "Not supported"
+  };
+}
+my $parallel = 8;  # number of parallel tasks to use
+
 my %optshort;
 my %optlong;
 my %helplong;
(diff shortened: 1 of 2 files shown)


emilengler:
```
Can't locate Proc/ParallelLoop.pm in @INC (you may need to install the Proc::ParallelLoop module) (@INC contains: /usr/local/libdata/perl5/site_perl/amd64-openbsd /usr/local/libdata/perl5/site_perl /usr/libdata/perl5/amd64-openbsd /usr/libdata/perl5) at ./gen.pl line 42.
```

At that point I do not think it is worth introducing yet another dependency.

emilengler:
> ```
> 
> Can't locate Proc/ParallelLoop.pm in @INC (you may need to install the Proc::ParallelLoop module) (@INC contains: /usr/local/libdata/perl5/site_perl/amd64-openbsd /usr/local/libdata/perl5/site_perl /usr/libdata/perl5/amd64-openbsd /usr/libdata/perl5) at ./gen.pl line 42.
> 
> ```
> 
> 
> 
> At that point I do not think it is worth introducing yet another dependency.

Many CI tests are also failing witj that error.

bagder:
Is it possible to make it a "soft" dependency and only use it if available?

dfandrich:
If perl allows that sort of thing, I don't see why not; it wouldn't add much
code at all. I'll try to figure out how to do that.

dfandrich:
It's ugly doing it in perl, but I've pushed a change that only optionally uses that module. The trick now is discoverability of this speedup.

docs/cmdline-opts/gen.pl
```diff
@@ -39,6 +39,20 @@
 =end comment
 =cut
 
+# Load the Proc::ParallelLoop module only if it's available
```
emilengler, line 42: I would also explain why we make use of that module (main-page generation)

bagder:
@dfandrich do you think this still makes it worth it after the other performance improvements we already landed?

I couldn't even figure out which debian package that brings this perl module so I haven't yet tried it out myself.

dfandrich:
I was thinking the same thing. I haven't tried the other patch yet, but if it
really drops the run time to 2 seconds, this patch isn't really worth the
bother any longer.

bagder:
I haven't measured how long it now takes in my build, but it is certainly in the 2 seconds ballpark.

bagder:
I think we can drop this as it is also fairly complicated and the current version generates its output much faster than before.

PR #9234, opened 2022-07-31
tool: comment and refactor getparameter()

emilengler:
This commit adds more comments to the getparameter() functions and its
associated structures. Beside this, it also refactors certain parts of
the code that were unclear to the commit author while auditing them.

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -61,7 +61,7 @@
 } while(0)
 
 struct LongShort {
-  const char *letter; /* short name option */
+  const char *letter; /* "short" name option (see comment in aliases) */
   const char *lname;  /* long name option */
   enum {
     ARG_NONE,   /* stand-alone but not a boolean */
@@ -556,7 +556,8 @@ static void cleanarg(argv_item_t str)
 #define cleanarg(x)
 #endif
 
-ParameterError getparameter(const char *flag, /* f or -long-flag */
+ParameterError getparameter(const char *flag, /* "-a", "-ab", "--long" or "long"
+                                               */
                             char *nextarg,    /* NULL if unset */
                             argv_item_t clearthis,
                             bool *usedarg,    /* set to TRUE if the arg
@@ -581,8 +582,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
 
   if(('-' != flag[0]) || ('-' == flag[1])) {
     /* this should be a long name */
-    const char *word = ('-' == flag[0]) ? flag + 2 : flag;
-    size_t fnam = strlen(word);
+    const char *word = flag + 2;
     int numhits = 0;
     bool noflagged = FALSE;
 
@@ -593,8 +593,10 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
       noflagged = TRUE;
     }
 
+    /* search for the accompanying internal short letter */
     for(j = 0; j < sizeof(aliases)/sizeof(aliases[0]); j++) {
-      if(curl_strnequal(aliases[j].lname, word, fnam)) {
+      /* treat word like a prefix in aliases and be happy with a unique hit */
+      if(curl_strnequal(aliases[j].lname, word, strlen(word))) {
         longopt = TRUE;
         numhits++;
         if(curl_strequal(aliases[j].lname, word)) {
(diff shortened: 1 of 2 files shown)


src/tool_getparam.c
```diff
@@ -61,7 +61,7 @@
 } while(0)
 
 struct LongShort {
-  const char *letter; /* short name option */
+  const char *letter; /* internal short name option (see comment in aliases) */
```
bagder, line 64: Not entirely true, as it can be the normal plain public short option **or** an internal one.

src/tool_getparam.c
```diff
@@ -556,7 +556,7 @@ static void cleanarg(argv_item_t str)
 #define cleanarg(x)
 #endif
 
-ParameterError getparameter(const char *flag, /* f or -long-flag */
+ParameterError getparameter(const char *flag, /* the raw argument from argv */
```
bagder, line 559: I don't think this helps and it's not true. For example when the function is used to handle arguments set in .curlrc.

src/tool_getparam.c
```diff
@@ -565,7 +565,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
                             struct OperationConfig *config)
 {
   char letter;
-  char subletter = '\0'; /* subletters can only occur on long options */
+  char subletter; /* subletters can only occur on long options */
```
bagder, line 568: I don't think we should remove "safety initializations" like this.

src/tool_getparam.c
```diff
@@ -579,10 +579,9 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
   (void)clearthis; /* for !HAVE_WRITABLE_ARGV builds */
   *usedarg = FALSE; /* default is that we don't use the arg */
 
-  if(('-' != flag[0]) || ('-' == flag[1])) {
+  if(flag[0] == '-' && flag[1] == '-') {
```
bagder, line 582: This is not the same logic. The previous version also caught long option names without leading dash-dash!

src/tool_getparam.c
```diff
@@ -593,8 +592,10 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
       noflagged = TRUE;
     }
 
+    /* search for the accompanying internal short letter */
     for(j = 0; j < sizeof(aliases)/sizeof(aliases[0]); j++) {
-      if(curl_strnequal(aliases[j].lname, word, fnam)) {
+      /* only compare the prefix in order to guess the remainder */
```
bagder, line 597: It doesn't "guess" the remainder, it is happy with a prefix as long as it is a unique hit. I think "guess" gives the reader the wrong idea.

src/tool_getparam.c
emilengler: But aren't they treated nearly the same?

src/tool_getparam.c
emilengler: Okay, fair. However, the current comment isn't true either, as it has been always `-f` or `--long-flag` for me, but I may have overseen an edge-case.

src/tool_getparam.c
emilengler: ACK

src/tool_getparam.c
emilengler: > The previous version also caught long option names without leading dash-dash!

I also thought about this initially, but I was unable to trigger that case (invoked like `curl -output foo example.com`; `curl output foo example.com`; ...).

src/tool_getparam.c
emilengler: ACK

src/tool_getparam.c
emilengler: ACK

src/tool_getparam.c
bagder: I don't think the fact that it was wrong before is a good reason to change it to another wrong...

How about listing the 4(?) different formats the arguments can be provided?
~~~c
/* "-a", "-ab", "--long" or "long" */

src/tool_getparam.c
bagder: The previous logic matched "--verbose" and "verbose" but not "-v", I believe that was the point.

Now you only match "--verbose"

src/tool_getparam.c
emilengler: sounds good

src/tool_getparam.c
```diff
@@ -581,8 +582,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
 
   if(('-' != flag[0]) || ('-' == flag[1])) {
     /* this should be a long name */
-    const char *word = ('-' == flag[0]) ? flag + 2 : flag;
-    size_t fnam = strlen(word);
+    const char *word = flag + 2;
```
bagder, line 585: What if `flag` doesn't start with two dashes? Just removing that condition surely will change the logic here?

emilengler:
I'll be closing that for now. I currently lack the time working on this refactor, which seemed simpler to me at first. Feel free to cherry-pick some things from it.

PR #9236, opened 2022-07-31
curl_easy_header: Add CURLH_PSEUDO to sanity check

### LINKED ISSUE #9235: curl_easy_header fails when requesting CURLH_PSEUDO headers
### I did this
Call `curl_easy_header` with the `CURLH_PSEUDO` origin, either alone or in a bitmask of other origins.

### I expected the following
To receive a pointer to the header struct. Instead libcurl returns `CURLHE_BAD_ARGUMENT`. It looks like pseudo headers were omitted from a sanity check: https://github.com/curl/curl/blob/master/lib/headers.c#L77

### curl/libcurl version
libcurl/7.84.0 OpenSSL/3.0.4 (Schannel) zlib/1.2.12 brotli/1.0.9 libidn2/2.3.2 libssh2/1.10.0 nghttp2/1.48.0 ngtcp2/0.6.0 nghttp3/0.5.0 libgsasl/1.10.0

### operating system
Windows 10 x64

charonn0:
Fixes #9235

### DIFF
--- a/lib/headers.c
+++ b/lib/headers.c
@@ -74,7 +74,8 @@ CURLHcode curl_easy_header(CURL *easy,
   struct Curl_header_store *hs = NULL;
   struct Curl_header_store *pick = NULL;
   if(!name || !hout || !data ||
-     (type > (CURLH_HEADER|CURLH_TRAILER|CURLH_CONNECT|CURLH_1XX|CURLH_PSEUDO)) ||
+     (type > (CURLH_HEADER|CURLH_TRAILER|CURLH_CONNECT|CURLH_1XX|\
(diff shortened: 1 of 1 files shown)


bagder (changes requested):
This would've been just fine if it wasn't too long! Our source code lines need to be within 79 columns.

bagder (approved):

charonn0:
I'm trying to make sense of the three failed tests, but I'm a little out of my depth.

bagder:
@charonn0 they're unfortunately flaky tests that happen without them being your fault

bagder:
Thanks!

PR #9245, opened 2022-08-04
Create CITATION.cff

oliver-peoples:
I've written a couple reports and internal research papers which frequently reference curl. I have my own bibtex file I use, but I thought it would be useful to add a standard CITATION.cff one.


bagder:
I presume this is the format: https://citation-file-format.github.io/

I think it's a bad idea to put this file in the source tree root, for a single reason: it needs to be updated at least for every release if not more often - depending how the list of authors is done. It would be much better to have it automatically generated and made available on curl.se. The format page does however seem to indicate that the file should rather be in the source code repo which saddens me.

emilengler:
> I think it's a bad idea to put this file in the source tree root, for a single reason: it needs to be updated at least for every release if not more often - depending how the list of authors is done. It would be much better to have it automatically generated and made available on curl.se. The format page does however seem to indicate that the file should rather be in the source code repo which saddens me.

I think we should only go with @bagder as the author and an `et al.`, as it is the case with the man-page and the COPYING file. This is because there is no fair way to determine when someones name should appear within that `CITATION.tff` file. By the number of commits? Unfair, because a single commit that changes hundreds of lines may have a deeper impact on the copyright than 100 commits where each of them just fixes a single typo. By the maintainer status? Unfair, because former maintainers and contributors who made more contributions than other maintainers are excluded then.

bagder:
Anyone who wants to reference curl should reference [curl.se](https://curl.se/), not the GitHub repository URL.

bagder (changes requested):
As commented

bagder:
No response, closing

PR #9247, opened 2022-08-04
url: really use the user provided in the url when netrc entry exists

### LINKED ISSUE #9243: REG [7.81->7.82] Authorization doesn't work for user in URL + netrc
### I did this
I have username in netrc, but if I provide a username in the URL, none of them is used, and no Authorization header is sent (even if it is the same user).

Example:
.netrc:
```
machine server login user password 5up3r53cr31
```
```bash
$ curl -n http://user@server/
Unauthorized
```

The issue is caused by (and fixed by reverting) d1237ac906ae7e3cd7a22c3a2d3a135a97edfbf5.

### I expected the following
I expect it to send `Authorization` header.

### curl/libcurl version
```
curl 7.84.0 (x86_64-w64-mingw32) libcurl/7.84.0 OpenSSL/1.1.1q (Schannel) zlib/1.2.12 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.2 libssh2/1.10.0 nghttp2/1.48.0
Release-Date: 2022-06-27
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe TLS-SRP zstd
```

### operating system
Windows 10

orgads:
If the user is specified as part of the URL, and the same user exists
in .netrc, Authorization header was not sent at all.

The user and password fields were assigned in conn->user and password
but the user was not assigned to data->state.aptr, which is the field
that is used in output_auth_headers and friends.

Fix by assigning the user also to aptr.

Amends commit d1237ac906ae7e3cd7a22c3a2d3a135a97edfbf5.

- Fixes #9243

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -3039,29 +3039,25 @@ static CURLcode override_login(struct Curl_easy *data,
       conn->user = strdup(*userp);
       if(!conn->user)
         return CURLE_OUT_OF_MEMORY;
-      /* don't update the user name below */
-      userp = NULL;
     }
   }
 #endif
 
   /* for updated strings, we update them in the URL */
-  if(userp) {
-    if(*userp) {
-      CURLcode result = Curl_setstropt(&data->state.aptr.user, *userp);
-      if(result)
-        return result;
-    }
-    if(data->state.aptr.user) {
-      uc = curl_url_set(data->state.uh, CURLUPART_USER, data->state.aptr.user,
-                        CURLU_URLENCODE);
-      if(uc)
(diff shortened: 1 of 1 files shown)


orgads:
I'm not sure how to add tests for this.

emilengler (commented):
LGTM

bagder:
Fixed in c40ec3178f8ce10927c1

PR #9248, opened 2022-08-04
netrc: Use the password from lines without login

orgads:
If netrc entry has password with empty login, use it for any username.

Example:
.netrc:
machine example.com password 123456

curl -vn http://user@example.com/

Fix it by initializing state_our_login to TRUE, and reset it only when
finding an entry with the same host and different login.


orgads:
I'm not sure how to add tests for this.

emilengler:
I'm personally a bit in doubt of this feature, because it implicates that a default user password is a usable approach, whereas in reality, this is a security nightmare.

PR #9250, opened 2022-08-05
test44[2-4]: add '--resolve' to the keywords

fabiankeil:
... so the tests can be automatically skipped when
using an external proxy like Privoxy.


bagder (approved):

bagder:
Thanks!

PR #9251, opened 2022-08-05
cmake: set TESTCASES as sources

orgads:
This includes them in the project tree for IDEs.

### DIFF
--- a/tests/data/CMakeLists.txt
+++ b/tests/data/CMakeLists.txt
@@ -27,4 +27,4 @@ include("${CMAKE_CURRENT_BINARY_DIR}/Makefile.inc.cmake")
 
 # Prints all available test cases. Do not quote TESTCASES, it must be displayed
 # as a space-separated string rather than comma-separated (a list in CMake).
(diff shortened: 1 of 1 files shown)


bagder:
Please elaborate on why this change is necessary. This changes the 'show' target for the makefile which AFAIK isn't used anymore.

orgads:
For IDEs that use CMake file api (like Qt Creator), this change lists the testcase files in the project tree.

bagder:
I don't think you are correct.

This change proposed changing the 'show' target in the makefile in there, I really do not see how an IDE would magically use that. I have now removed that target completely since it isn't in use anymore.

This PR therefore has conflicts.

orgads:
Only files that are listed as SOURCES (for any target) are shown in the project tree on Qt Creator.

So I just picked the first target I found :)

Maybe @cristianadam can elaborate, and suggest an alternative.

bagder:
Ok. Then I think you need to create a target for that and properly document what it's for.

I would perhaps argue that test cases are not source code, but I don't know why you want this nor how your IDE works so I will not.

PR #9252, opened 2022-08-05
cmake: move libcurl to src directory on Windows

orgads:
It has to be in the same directory for the executable to use it.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -111,6 +111,10 @@ endif()
 set_target_properties(${LIB_NAME} PROPERTIES PREFIX "")
 set_target_properties(${LIB_NAME} PROPERTIES IMPORT_PREFIX "")
 
+if(WIN32 OR CYGWIN)
+  set_target_properties(${LIB_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "../src")
+endif()
+
 if(CURL_HAS_LTO)
   set_target_properties(${LIB_NAME} PROPERTIES
     INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE


bagder (changes requested):
Did you push the wrong commit? The description doesn't match the commit.

orgads:
Oops, sorry

orgads:
Fixed, ready for review.

bagder:
What problem does this fix? People and CI build with cmake on Windows fine already, don't they?

orgads:
It build good, but you cannot execute curl directly from src. This change moves licurl.dll to src, which enables execution of curl from there.

vszakats:
If adding `src` to the PATH can also resolve this, I'd prefer it. It would be a more universal solution and without introducing a platform- (and build-method) specific build exception. BTW, isn't this also an issue for non-Windows CMake builds?

jzakrzewski:
On platforms supporting RPATH/RUNPATH CMake uses them by default AFAIR, so the binaries in the build tree can find everything. This is again removed during the install phase.

And for Windows I typically either use a common output directory for everything (using CMAKE_XXX_OUTPUT_DIRECTORY and/or target properties) or set PATH in my dev env, as @vszakats says.

orgads:
CMake build on linux sets rpath, so it's possible to run from `src`.

Why do you care where the dll is located? This is only in the build directory, it doesn't affect install. On install the executable and the library are both installed in `bin` anyway.

vszakats:
@orgads 
- installing is optional (esp on Windows), so there is fair chance some scripts/people do their own thing and expect files to be in certain locations. These would need to be updated (just checked and curl-for-win is also affected. Even though it does do `install`, it needs to know the DLL location to pick up its `.map` file.) For mutltiplatform/multi-build scripts this can be more cumbersome, as they would need to follow the conditions of this exception.
- non-Windows is not only Linux (or systems with rpath)
- a CMake exception will not resolve this for the other build methods (Makefile.m32, autotools, MSVC?). Those may (or may not) have different workarounds for this.

What is wrong with using a `PATH` here?

orgads:
> * installing is optional (esp on Windows), so there is fair chance some scripts/people do their own thing and expect files to be in certain locations. These would need to be updated (just checked and curl-for-win is also affected. Even though it does do `install`, it needs to know the DLL location to pick up its `.map` file.) For mutltiplatform/multi-build scripts this can be more cumbersome, as they would need to follow the conditions of this exception.

Ok, so how about creating a symbolic link or copying the dll?

> * non-Windows is not only Linux (or systems with rpath)

I don't have other systems to test on.

> * a CMake exception will not resolve this for the other build methods (Makefile.m32, autotools, MSVC?). Those may (or may not) have different workarounds for this.

The MSVC build creates the exe and the dll in the same directory (e.g. `build/Win32/VC14.30/DLL Debug - DLL Windows SSPI - DLL WinIDN`)

Autotools creates `src/.libs/curl.exe` which is the real curl, and a "fake" curl exeutable in `src/curl.exe`, which is created using `libtool`, and loads the correct dll from `lib/.libs/libcurl-4.dll` (looks like some kind of rpath alternative). CMake doesn't support that as far as I can tell.

> What is wrong with using `PATH` here?

I'll have to remember setting it every time I build curl, and if I have different build directories, I'll need to set it differently for each one of them.

vszakats:
> I'll have to remember setting it every time I build curl, and if I have different build directories, I'll need to set it differently for each one of them.

The entry point of tests is `tests/runtests.pl`. Is there anything against extending PATH there?

orgads:
> > I'll have to remember setting it every time I build curl, and if I have different build directories, I'll need to set it differently for each one of them.
> 
> The entry point of tests is `tests/runtests.pl`. Is there anything against extending PATH there?

That's an option for running auto tests, but what if I want to run curl manually?

vszakats:
`make install` is one solution for these cases. Not sure about the exact goal/use-case here, but if libcurl.dll happens to depend on further DLLs, copying it next to curl.exe is not enough.

orgads:
> `make install` is one solution for these cases. Not sure about the exact goal/use-case here, but if libcurl.dll happens to depend on further DLLs, copying/moving it next to curl.exe is not enough.

My workflow when I try to debug an issue is build -> run `src/curl` with the params I want to test. Adding another step of install can be long and disturbing. Do you install every time you build?

jay:
The cmake build behaves like the autotools build in that it will compile the lib files to a lib dir (ie libcurl output files pdb dll lib etc) and the src files to a src dir (ie curl tool output files). Unlike autotools, cmake doesn't output a wrapper on Windows to accommodate running the exe without install.

The change in this PR does not seem correct to put the dll in the src dir, it seems contrary to the way cmake does it or the way we are using cmake. Is it typical in other projects for cmake output files to all be in one directory, are there any disadvantages?

orgads:
Qt Creator (which has a sandal executable and many libraries), and most projects I know, can be executed directly from the build directory.

bagder:
I don't sense that people think this is a good idea, and it is an idea that anyone can easily just have their own local build script do.

PR #9254, opened 2022-08-05
runtests: convert protocol to CRLF on Windows

orgads:
The protocol field is saved as binary (the rest of the file is LF but
this part is CRLF). It is overwritten by both VS Code and Qt Creator,
that replace it with LF to match the rest, and this breaks the tests.

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -4480,6 +4480,11 @@ sub singletest {
         # check if there's any attributes on the verify/protocol section
         my %hash = getpartattr("verify", "protocol");
 
+        if($has_textaware) {
+            # text mode when running on windows: fix line endings
+            map s/\r\n/\n/g, @protstrip;
+            map s/\n/\r\n/g, @protstrip;
+        }
         if($hash{'nonewline'}) {
             # Yes, we must cut off the final newline from the final line
             # of the protocol data


orgads:
Actually this should not be Windows-specific, as it will break on other platforms if tests file are rewritten with LF.

How about adding something like `eol="lf"` attribute to `<protocol>` (default to `crlf`), and translate according to this attribute?

emilengler:
> It is overwritten by both VS Code and Qt Creator, that replace it with LF to match the rest, and this breaks the tests.

This is not our fault.

bagder (changes requested):
This is a big no-no. The protocol needs to be checked *binary* and your system clearly broke the tests if it modified the newlines. Stop it from doing that.

orgads:
How about my other suggestion in the comment? https://github.com/curl/curl/pull/9254#issuecomment-1206290093

How do you edit the test files?

emilengler:
> How about my other suggestion in the comment? https://github.com/curl/curl/pull/9254#issuecomment-1206290093
> 
> 
> 
> How do you edit the test files?

I personally use a combination of vi and emacs.

orgads:
Having different newlines in a single file is *very* odd. I've never seen anything like it, especially when the content is XML, so it's text anyway. Why not use a markup for the line endings, and handle it in the runner script?

bagder:
> Why not use a markup for the line endings, and handle it in the runner script?

Mostly because

1. it has not been a problem
2. it would require changing some 1500+ files

orgads:
1. How many developers contribute tests? Am I the first to hit this issue?
2. It is very easy to write a script that will adapt all the files if you agree to this change.

bagder:
> How many developers contribute tests? Am I the first to hit this issue?

~~~
git shortlog -s -- tests | wc -l
220
~~~

> It is very easy to write a script that will adapt all the files if you agree to this change.

I would be open to the idea, but it depends on the exact change.

PR #9256, opened 2022-08-05
tests: add tests for netrc login/password combinations

orgads:
Covers the following PRs:
- #9066
- #9247
- #9248


orgads:
Replaced the inline base64 values with `%b64[user:pass]b64%`

PR #9257, opened 2022-08-06
netrc: remove the static 4096 buffer

emilengler:
This commit replaces the static 4096-byte buffer for the .netrc file
with a dynamically allocated string based on the length of the .netrc
file-stream.

Closes #XXXX

### DIFF
--- a/lib/netrc.c
+++ b/lib/netrc.c
@@ -53,6 +53,22 @@ enum host_lookup_state {
 #define NETRC_SUCCESS 0
 
 /*
+ * Return the file size as an integer (max. ~2GB) or a -1 in case of a failure.
+ */
+static int get_file_sz(FILE *fp)
+{
+  long sz;
+
+  fseek(fp, 0L, SEEK_END); /* Go to the end */
+  sz = ftell(fp); /* Get the position of the end which is the file size */
+  fseek(fp, 0L, SEEK_SET); /* Go back to the position from where we started */
+
+  if(sz > INT_MAX - 1) /* -1 because we add +1 later for '\0' */
+    return -1;
+  return (int)sz;
+}
+
+/*
  * Returns zero on success.
  */
 static int parsenetrc(const char *host,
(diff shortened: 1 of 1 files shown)


emilengler:
Wait! I forgot to free the memory resulting in a memory leak...

emilengler:
> Wait! I forgot to free the memory resulting in a memory leak...

fixed

lib/netrc.c
```diff
@@ -81,8 +97,20 @@ static int parsenetrc(const char *host,
...
-    char netrcbuffer[4096];
-    int  netrcbuffsize = (int)sizeof(netrcbuffer);
+    char *netrcbuffer = NULL;
+    int  netrcbuffsize;
+
+    netrcbuffsize = get_file_sz(file);
+    if(netrcbuffsize++ == -1) {
+      retcode = NETRC_FAILED;
+      goto out;
+    }
```
bagder, line 107: Without this change, a *super huge* .netrc file was a nuisance and slow. With this change, you introduce a super huge alloc as well. I fear this can have some undesired consequences. **If** we should allocate memory for the entire file, then I think it needs to have a capped maximum size, like maybe 8MB

bagder (commented):
Is this addressing a problem somewhere for someone with lines longer than 4K?

lib/netrc.c
```diff
@@ -81,8 +97,20 @@ static int parsenetrc(const char *host,
   file = fopen(netrcfile, FOPEN_READTEXT);
   if(file) {
     bool done = FALSE;
-    char netrcbuffer[4096];
-    int  netrcbuffsize = (int)sizeof(netrcbuffer);
+    char *netrcbuffer = NULL;
+    int  netrcbuffsize;
+
+    netrcbuffsize = get_file_sz(file);
+    if(netrcbuffsize++ == -1) {
```
bagder, line 104: This will make having netrc as a (named) pipe fail.

emilengler: Fair point, as @dfandrich pointed out as well. This method is only applicable if `FILE *` is a seekable stream, which isn't necessarily always the case.

lib/netrc.c
emilengler: ACK. I will close this PR then, because at that point it'd be only a somewhat endless race.

emilengler:
See comments.

PR #9258, opened 2022-08-06
tool: reintroduce set file comment code for AmigaOS

Futaura:
Amiga specific code which put the URL in the file comment was perhaps
accidentally removed in b88940850002a3f1c25bc6488b95ad30eb80d696 having
originally been added in 5c215bdbdfde8b2350cdcbac82aae0c914da5314.
Reworked to fit the code changes and added it back in.

Reported-by: Michael Trebilcock
Originally-added-by: Chris Young

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -327,6 +327,22 @@ static CURLcode pre_transfer(struct GlobalConfig *global,
   return result;
 }
 
+#ifdef __AMIGA__
+static void AmigaSetComment(struct per_transfer *per,
+                            CURLcode result)
+{
+  struct OutStruct *outs = &per->outs;
+  if(!result && outs->s_isreg && outs->filename) {
+    /* Set the url (up to 80 chars) as comment for the file */
+    if(strlen(per->this_url) > 78)
+      per->this_url[79] = '\0';
+    SetComment(outs->filename, per->this_url);
+  }
+}
+#else
+#define AmigaSetComment(x,y) Curl_nop_stmt
+#endif
+
 /*
  * Call this after a transfer has completed.
  */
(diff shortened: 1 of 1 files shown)


bagder (commented):
How about making it an Amiga specific function that uses a blank macro for non-Amiga? In an attempt to maybe reduce the ifdefs within that already very complicated function. Maybe like this:
~~~diff
index 839efb1a1..853e2fbb2 100644
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -325,10 +325,25 @@ static CURLcode pre_transfer(struct GlobalConfig *global,
     per->input.fd = per->infd;
   }
   return result;
 }
 
+#ifdef __AMIGA__
+static void AmigaSetComment(CURLcode result,
+                            struct OutStruct *outs)
+{
+  if(!result && outs->s_isreg && outs->filename) {
+    /* Set the url (up to 80 chars) as comment for the file */
+    if(strlen(per->this_url) > 78)
+      per->this_url[79] = '\0';
+    SetComment(outs->filename, per->this_url);
+  }
+}
+#else
+#define AmigaSetComment(x,y)
+#endif
+
 /*
  * Call this after a transfer has completed.
  */
 static CURLcode post_per_transfer(struct GlobalConfig *global,
                                   struct per_transfer *per,
@@ -601,10 +616,12 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
       notef(global, "Removing output file: %s\n", outs->filename);
       unlink(outs->filename);
     }
   }
 
+  AmigaSetComment(result, outs);
+
   /* File time can only be set _after_ the file has been closed */
   if(!result && config->remote_time && outs->s_isreg && outs->filename) {
     /* Ask libcurl if we got a remote file time */
     curl_off_t filetime = -1;
     curl_easy_getinfo(curl, CURLINFO_FILETIME_T, &filetime);

Futaura:
Sure - whatever you prefer. I just added it back more or less in the same way as it had been originally committed.

Futaura:
As suggested, I've moved the code into a function.

bagder (approved):

bagder:
Thanks!

PR #9259, opened 2022-08-06
file: add handling of native AmigaOS paths

Futaura:
On AmigaOS 4.x, handle native absolute paths, whilst blocking relative
paths. Also allow unix style paths if feature enabled at link time.

Inspiration-from: Michael Trebilcock

### DIFF
--- a/lib/file.c
+++ b/lib/file.c
@@ -71,6 +71,8 @@
 
 #if defined(WIN32) || defined(MSDOS) || defined(__EMX__)
 #define DOS_FILESYSTEM 1
+#elif defined(__amigaos4__)
+#define AMIGA_FILESYSTEM 1
 #endif
 
 #ifdef OPEN_NEEDS_ARG3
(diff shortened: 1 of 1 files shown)


lib/file.c
```diff
@@ -196,8 +198,33 @@ static CURLcode file_connect(struct Curl_easy *data, bool *done)
     return CURLE_URL_MALFORMAT;
   }
 
+  #if defined(AMIGA_FILESYSTEM)
```
bagder, line 201: We usually go with just `ifdef` when checking a single define.

lib/file.c
```diff
@@ -196,8 +198,33 @@ static CURLcode file_connect(struct Curl_easy *data, bool *done)
...
+  /*
+   * A leading slash in an AmigaDOS path denotes the parent
+   * directory, and hence we block this as it is relative.
+   * Absolute paths start with 'volumename:', so we check for
+   * this first. Failing that, we treat the path as a real unix
```
bagder, line 206: Can the volume name be any length? Or maybe it doesn't matter as the open will just fail anyway if tried with a too long name?

Futaura: > Can the volume name be any length? Or maybe it doesn't matter as the open will just fail anyway if tried with a too long name?

I'm not sure on the length limit (probably 107 characters, at a guess), but it may also differ depending which filesystem is used. Normally, volume names are not that long though, but equally always longer than the drive letter than Windows uses.

lib/file.c
Futaura: Sorry - I'll fix that (this was originally part an `#elif`), but I moved it.

lib/file.c
```diff
@@ -196,8 +198,33 @@ static CURLcode file_connect(struct Curl_easy *data, bool *done)
...
+   * path, but only if the application was compiled with -lunix.
+   */
+  fd = -1;
+  file->path = real_path;
+
+  if(real_path[0] == '/') {
+    extern int __unix_path_semantics;
+    if(strchr(real_path, ':')) {
```
emilengler, line 214: Consider doing the `real_path + 1` also here.

bagder (approved):

bagder:
thanks!

PR #9260, opened 2022-08-06
AmiSSL v5 support and AmigaOS 4.x configure + resolver

Futaura:
These commits bump the minimum AmiSSL requirement to 5.1 (OpenSSL 3.0), which also allows some of the old workarounds like function stubs to be removed from all the sources (as libamisslstubs.a now covers those from the AmiSSL SDK). The configure script now works better for AmigaOS 4.x and has AmiSSL v5 detection. Finally, the threaded resolver has been fixed to work properly on AmigaOS 4.x, as it was not actually thread safe before.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -813,6 +813,7 @@ dnl return value in RECV_TYPE_RETV.
 
 AC_DEFUN([CURL_CHECK_FUNC_RECV], [
   AC_REQUIRE([CURL_CHECK_HEADER_WINSOCK2])dnl
+  AC_REQUIRE([CURL_INCLUDES_BSDSOCKET])dnl
   AC_CHECK_HEADERS(sys/types.h sys/socket.h)
   #
   AC_MSG_CHECKING([for recv])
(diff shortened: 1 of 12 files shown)


bagder (commented):
This is curious. As you have AmiSSL as a library for Amiga, then why doesn't that just define `OPENSSL_NO_MD4` and `OPENSSL_NO_MD5` ?

bagder:
This PR contains very different changes, which makes it very hard to work with. I think you should keep separate changes in separate PRs so that they can be discussed, approved and merged independently.

Futaura:
> This is curious. As you have AmiSSL as a library for Amiga, then why doesn't that just define `OPENSSL_NO_MD4` and `OPENSSL_NO_MD5` ?

As far as curl's usage is concerned, AmiSSL is the same as OpenSSL. AmiSSL is currently using OpenSSL 3.0.5 which enables MD4 and MD5 by default. The functions are both now deprecated in OpenSSL, but AmiSSL still supports all the deprecated functions and has the legacy provider. Curl should use the OpenSSL functions when available.

On this topic, I don't quite understand why the `#ifdef` handling whether MD4 and MD5 are disabled are different. The `md5.c` code looks more logical, but not sure why `md4.c` doesn't use a similar check, but instead explicitly checks `OPENSSL_VERSION_MAJOR`.

bagder:
> On this topic, I don't quite understand why the #ifdef handling whether MD4 and MD5 are disabled are different.

Just by accident.

Futaura:
Ok, I'll try to split this.

PR #9261, opened 2022-08-06
New feature for curl --in64

lucky-sideburn:
Hi! 

Sometimes I need to send log curl strings to customers. Some of them has windows so they have not base64 command. For this reasom I have written some code to do this:

[rocky@dev01 curl]$ ./src/curl  --in64 aHR0cHM6Ly9kZXZvcHN0cmliZS5pdCAtSSAtawo
Decoding Base64 args...

HTTP/1.1 200 OK
Date: Sat, 06 Aug 2022 19:30:17 GMT
Server: Apache/2.4.53 (Debian)
X-Powered-By: PHP/7.4.29
Link: <https://devopstribe.it/wp-json/>; rel="https://api.w.org/"
Content-Type: text/html; charset=UTF-8

### DIFF
--- a/src/tool_base64.h
+++ b/src/tool_base64.h
@@ -1,0 +1,148 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.se/docs/copyright.html.
+ *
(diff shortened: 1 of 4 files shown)


dfandrich:
Why not just supply your customers with the base64 binary, or give them URLs or
command-lines they can use that are already properly quoted for their
environment?  I don't see this as a feature that adds any real value.

lucky-sideburn:
Hi @dfandrich 

I thought this embedded curl feature might be useful.. If you want I delete the pull request :)

dfandrich:
Sure, some people would find it useful, but there are costs to every feature
added. curl follows the Unix philosophy, where small programs work together to
accomplish particular tasks. In this case, the base64 program is the small
program that can already accomplish the task that this feature adds.
It is unfortunate that Windows does not follow the Unix philosophy.

bagder:
I'm seconding @dfandrich's concerns. This seems like a very niche feature.

lucky-sideburn:
Hi @bagder thank you. I totaly agree with @dfandrich about the Unix philosophy and yes, it is a niche feature! Can I close the pr?

bagder:
I think you can, yes

PR #9265, opened 2022-08-07
amigaos: fix threaded resolver on AmigaOS 4.x

Futaura:
Replace ip4 resolution function on AmigaOS 4.x, as it requires
runtime feature detection and extra code to make it thread safe.

### DIFF
--- a/lib/amigaos.c
+++ b/lib/amigaos.c
@@ -25,8 +25,14 @@
 #include "curl_setup.h"
 
 #ifdef __AMIGA__
-#  include "amigaos.h"
-#  if defined(HAVE_PROTO_BSDSOCKET_H) && !defined(USE_AMISSL)
+
+#include "hostip.h"
+#include "amigaos.h"
+
+#ifdef HAVE_PROTO_BSDSOCKET_H
+#  if defined(__amigaos4__)
+#    include <bsdsocket/socketbasetags.h>
+#  elif !defined(USE_AMISSL)
 #    include <amitcp/socketbasetags.h>
 #  endif
 #  ifdef __libnix__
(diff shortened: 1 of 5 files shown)


lib/amigaos.c
```diff
@@ -38,8 +44,142 @@
...
+ *
+ * With the threaded resolver enabled, try to use gethostbyname_r() where
+ * available, otherwise (re)open bsdsocket.library and fallback to
+ * gethostbyname().
+ */
+
+#include <proto/bsdsocket.h>
+
+static struct SocketIFace *__CurlISocket = NULL;
+static uint32 SocketFeatures = 0;
```
bagder, line 71: statics really? Are you sure these are thread-safe?

Futaura: I'm with you on the statics, but this is the only way I could think of, without forcing unnecessary extra overhead on every resolve. Normally, I'd rely on the global ISocket variable which the application will have (if using AmiSSL), but since I need to enable the `SBTC_CAN_SHARE_LIBRARY_BASES` on it to make it thread safe, I didn't want to interfere with that, or indeed allow the application to interfere with libcurl by switching that feature off!

`Curl_amiga_init()` isn't thread safe by itself, when it comes to initialising `__CurlISocket`, but as long as `GLOBAL_INIT_IS_THREADSAFE` is defined it will be thread safe, as it will be on AmigaOS 4.x builds - that should be enough, right? Once initialised, the functionality is completely thread safe (the old code wasn't as it relied on `gethostbyname()`.

This code is going to get reused for enabling `getaddrinfo()` support to in the future, but that function appears to be broken in the AmigaOS 4.x tcp/ip stack at the moment!

bagder:
This PR appears to have merge conflicts now. Can you please rebase and force-push it?

bagder:
It still says "This branch cannot be rebased due to conflicts". I really did mean rebase and force-push...

Futaura:
Sorry - was assuming the buttons on the GitHub website would help me out, but seemed to created a bigger mess! Having booted up linux, hopefully my first attempt at a rebase is correct, but please let me know if not. Thanks.

bagder:
Thanks!

gvanem:
Real retro stuff: [curl on Amiga](https://watt-32.net/misc/Amiga-curl-version.html).  (just a test of HTML and web-fonts).

PR #9266, opened 2022-08-07
configure: -pthread not available on AmigaOS 4.x

Futaura:
The most recent GCC builds for AmigaOS 4.x do not allow -pthread and
exit with an error. Instead, need to explictly specify -lpthread.


bagder (approved):

bagder:
thanks!

PR #9267, opened 2022-08-07
amissl: make AmiSSL v5 a minimum requirement

Futaura:
AmiSSL v5 is the latest version, featuring a port of OpenSSL 3.0.
Support for previous OpenSSL 1.1.x versions has been dropped, so
makes sense to enforce v5 as the minimum requirement. This also
allows all the AmiSSL stub workarounds to be removed as they are
now provided in a link library in the AmiSSL SDK.


bagder (approved):
We have no Amiga CI and from what I can tell very few developers using it, so as long as all other platforms are still fine and the developer of this PR says it works fine on Amiga I figure we need to accept their word for it.

bagder:
thanks!

PR #9268, opened 2022-08-07
configure: fixup bsdsocket detection code for AmigaOS 4.x

Futaura:
The code that detects bsdsocket.library for AmigaOS did not work
for AmigaOS 4.x. This has been fixed and also cleaned up a little
to reduce duplication. Wasn't technically necessary before, but is
required when building with AmiSSL instead of OpenSSL.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -871,7 +871,7 @@ $curl_includes_bsdsocket
 #endif
 #define RECVCALLCONV PASCAL
 #else
-                      $curl_includes_bsdsocket
+$curl_includes_bsdsocket
 #ifdef HAVE_SYS_TYPES_H
 #include <sys/types.h>
 #endif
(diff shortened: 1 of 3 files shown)


acinclude.m4
```diff
@@ -873,10 +871,7 @@ struct Library *SocketBase = NULL;
 #endif
 #define RECVCALLCONV PASCAL
 #else
-#ifdef HAVE_PROTO_BSDSOCKET_H
-#include <proto/bsdsocket.h>
-struct Library *SocketBase = NULL;
-#endif
+                      $curl_includes_bsdsocket
```
bagder, line 874: Is this huge leading white space a mistake?

Futaura: Perhaps 😕. It is indented to match the level of the code below, but I didn't do that anywhere else, so I'll remove the white space on this.

bagder:
thanks!

PR #9269, opened 2022-08-07
amissl: allow AmiSSL to be used with AmigaOS 4.x builds

Futaura:
Enable AmiSSL to be used instead of static OpenSSL link libraries.
for AmigaOS 4.x, as it already is in the AmigaOS 3.x build.


bagder:
Thanks!

Issue #9274, opened 2022-08-08
DNS cache bug about ai_socktype and ai_protocol


Cering:
### I did this
- Send an http3 request first and set `CURLOPT_FRESH_CONNECT` and `CURLOPT_FORBID_REUSE` to 1, for avoid reusing http3 connection
- After that, send an http(tcp) request. This will not reuse the http3 connection, but will use the dns cache
- However, the second request return fail because of error 7(Couldn't connect to server)
- The demo code is [demo.cpp.txt](https://github.com/curl/curl/files/9279417/demo.cpp.txt), and major output info: 
```
New Request[0]: https://www.google.com
curl_debug[0]: STATE: INIT => CONNECT handle 0x13c2a68; line 1833 (connection #-5000)
curl_debug[0]: Added connection 0. The cache now contains 1 members
curl_debug[0]: STATE: CONNECT => RESOLVING handle 0x13c2a68; line 1879 (connection #0)
curl_debug[0]: family0 == v4, family1 == v6
curl_debug[0]:   Trying 142.251.42.228:443...
curl_debug[0]:  CAfile: /etc/pki/tls/certs/ca-bundle.crt
curl_debug[0]:  CApath: none
curl_debug[0]: Connect socket 5 over QUIC to 142.251.42.228:443
curl_debug[0]: Sent QUIC client Initial, ALPN: h3,h3-29,h3-28,h3-27
...
curl_debug[0]: multi_done: status: 0 prem: 0 done: 0
curl_debug[0]: The cache now contains 0 members
curl_debug[0]: Closing connection 2
curl_debug[0]: Expire cleared (transfer 0x13c2a68)
mycurl_process: https://www.google.com.hk/, curl_code: 0(No error), http_version: 30, total_time: 0.760068, connect_time: 0.618590, download_bytes: 49727, download_speed: 65424


New Request[1]: https://www.google.com
curl_debug[0]: STATE: INIT => CONNECT handle 0x13c2a68; line 1833 (connection #-5000)
curl_debug[0]: Added connection 3. The cache now contains 1 members
curl_debug[0]: Hostname www.google.com was found in DNS cache
curl_debug[0]: family0 == v4, family1 == v6
curl_debug[0]: The cache now contains 0 members
curl_debug[0]: Closing connection 3
curl_debug[0]: Expire cleared (transfer 0x13c2a68)
[...]

PR #9275, opened 2022-08-08
Curl_socket(): fix socktype and protocol

Cering:
for fix https://github.com/curl/curl/issues/9274

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -1605,8 +1605,8 @@ CURLcode Curl_socket(struct Curl_easy *data,
 
   addr->family = ai->ai_family;
   addr->socktype = (conn->transport == TRNSPRT_TCP) ? SOCK_STREAM : SOCK_DGRAM;
-  addr->protocol = conn->transport != TRNSPRT_TCP ? IPPROTO_UDP :
-    ai->ai_protocol;
+  addr->protocol = (conn->transport == TRNSPRT_TCP) ? IPPROTO_TCP :
+    IPPROTO_UDP;
   addr->addrlen = ai->ai_addrlen;
 
   if(addr->addrlen > sizeof(struct Curl_sockaddr_storage))


lib/connect.c
```diff
@@ -1607,6 +1607,12 @@ CURLcode Curl_socket(struct Curl_easy *data,
   addr->socktype = (conn->transport == TRNSPRT_TCP) ? SOCK_STREAM : SOCK_DGRAM;
   addr->protocol = conn->transport != TRNSPRT_TCP ? IPPROTO_UDP :
     ai->ai_protocol;
+  if(addr->socktype == SOCK_STREAM && addr->protocol == IPPROTO_UDP)
+   /* If ai from an dns cache whose ai_protocol is IPPROTO_UDP, and want to
+    * create TRNSPRT_TCP socket here, change addr->protocol to IPPROTO_TCP.
+    * Otherwise socket() will return a 'Protocol not supported' error.
+    */
+    addr->protocol = IPPROTO_TCP;
```
bagder, line 1615: Can't we simplify this and instead just do this:
~~~diff
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -1603,12 +1603,11 @@ CURLcode Curl_socket(struct Curl_easy *data,
    * if this has been set, before that, it is initialized from parameters.
    */
 
   addr->family = ai->ai_family;
   addr->socktype = (conn->transport == TRNSPRT_TCP) ? SOCK_STREAM : SOCK_DGRAM;
-  addr->protocol = conn->transport != TRNSPRT_TCP ? IPPROTO_UDP :
-    ai->ai_protocol;
+  addr->protocol = conn->transport == TRNSPRT_TCP ? IPPROTO_TCP : IPPROTO_UDP;
   addr->addrlen = ai->ai_addrlen;
 
   if(addr->addrlen > sizeof(struct Curl_sockaddr_storage))
      addr->addrlen = sizeof(struct Curl_sockaddr_storage);
   memcpy(&addr->sa_addr, ai->ai_addr, addr->addrlen);

Cering: Yes! This is graceful

bagder (dismissed):

bagder:
Argh. I forgot unix domain sockets, and when I tried to adapt to those I spotted they they were wrongly marked TCP so it wasn't that easy, and when setting them to use `TRNSPRT_UNIX` they broke `Curl_socket()` ...

IOW: I did #9276, which actually also is a fix for #9274 but different than this. If you're fine with that, we can close this PR again.

Cering:
will fixed in https://github.com/curl/curl/pull/9276

PR #9281, opened 2022-08-08
ldap: adapt to conn->port now being an 'int'

bagder:
Remove typecasts. Fix printf() formats.

Follow-up from 764c6bd3bf.
Pointed out by Coverity CID 1507858.


emilengler (commented):
Isn't `%d` more natural for an `int`?

bagder:
since the output is always unsigned I figured %u makes it more obvious

emilengler:
> since the output is always unsigned I figured %u makes it more obvious 

But, at that point, wouldn't it mane more sense to change conn->port to an unsigned int anyway? Although not the scope of this PR, obviously.

bagder:
I don't think it does, as that will just cause some more typecasts or require conversions elsewhere. The printf format string is for how it outputs the number, %d and %u both work on the same int input. It's not an error.

PR #9283, opened 2022-08-08
refactor: split resolve_server() into functions

emilengler:
This commit splits the branch-heavy resolve_new() function into various
sub-functions, in order to reduce the amount of nested
if/else-statements.

Beside this, it also removes many else-sequences, by returning in the
previous if-statement.

Closes #XXXX

<hr>

This is still in draft-mode, because I would like to let the CI jobs run over it, as I am not 100% sure about not having made a mistake.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -3414,6 +3414,146 @@ static CURLcode parse_connect_to_slist(struct Curl_easy *data,
   return result;
 }
 
+#ifdef USE_UNIX_SOCKETS
+static CURLcode resolve_unix(struct Curl_easy *data,
+                             struct connectdata *conn,
+                             char *unix_path)
+{
+  struct Curl_dns_entry *hostaddr = NULL;
+  bool longpath = FALSE;
+
+  DEBUGASSERT(unix_path);
+  DEBUGASSERT(conn->dns_entry == NULL);
+
+  /* Unix domain sockets are local. The host gets ignored, just use the
+   * specified domain socket address. Do not cache "DNS entries". There is
+   * no DNS involved and we already have the filesystem path available. */
+  hostaddr = calloc(1, sizeof(struct Curl_dns_entry));
+  if(!hostaddr)
+    return CURLE_OUT_OF_MEMORY;
+
+  hostaddr->addr = Curl_unix2addr(unix_path, &longpath,
+                                  conn->bits.abstract_unix_socket);
+  if(!hostaddr->addr) {
+    if(longpath)
+      /* Long paths are not supported for now */
+      failf(data, "Unix socket path too long: '%s'", unix_path);
+    free(hostaddr);
+    return longpath ? CURLE_COULDNT_RESOLVE_HOST : CURLE_OUT_OF_MEMORY;
+  }
+
+  hostaddr->inuse++;
+  conn->dns_entry = hostaddr;
+  return CURLE_OK;
+}
+#endif
+
+#ifndef CURL_DISABLE_PROXY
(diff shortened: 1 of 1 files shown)


bagder:
I think you meant split **resolve_server()** into functions, right? That "new" string is something you inserted in several places but I don't think you need it...

emilengler:
> I think you meant split **resolve_server()** into functions, right? That "new" string is something you inserted in several places but I don't think you need it...

Oh yes! Well I've used the `new` string so many times to distinguish it from the cache resolving, but I think it's not worth it.

emilengler:
Marking this ready for review. Instead of looking at the diff, I suggest comparing the two `lib/url.c` files side by side.

lib/url.c
```diff
@@ -3414,142 +3414,166 @@ static CURLcode parse_connect_to_slist(struct Curl_easy *data,
...
+}
 
-      else if(!hostaddr) {
-        failf(data, "Couldn't resolve proxy '%s'", host->dispname);
-        result = CURLE_COULDNT_RESOLVE_PROXY;
-        /* don't return yet, we need to clean up the timeout first */
-      }
-    }
+/* Perform a fresh resolve */
+static CURLcode resolve_new(struct Curl_easy *data,
```
bagder, line 3529: So not resolve **host** or **name** ? When I read "new" it sounds like this is the new version of the resolve function and it makes my head itch.

lib/url.c
```diff
@@ -3414,142 +3414,166 @@ static CURLcode parse_connect_to_slist(struct Curl_easy *data,
...
+
+#ifndef CURL_DISABLE_PROXY
+  if(!unix_path && conn->socks_proxy.host.name &&
+     !strncmp(UNIX_SOCKET_PREFIX"/",
+              conn->socks_proxy.host.name, sizeof(UNIX_SOCKET_PREFIX)))
+    unix_path = conn->socks_proxy.host.name + sizeof(UNIX_SOCKET_PREFIX) - 1;
+#endif
+
+  if(unix_path) {
+    conn->transport = TRNSPRT_UNIX;
```
bagder, line 3544: shouldn't this assignment rather be moved into `resolve_unix` ?

lib/url.c
```diff
@@ -3414,142 +3414,166 @@ static CURLcode parse_connect_to_slist(struct Curl_easy *data,
...
+static CURLcode resolve_new(struct Curl_easy *data,
+                            struct connectdata *conn,
+                            bool *async)
+{
+#ifdef USE_UNIX_SOCKETS
+  char *unix_path = conn->unix_domain_socket;
+
+#ifndef CURL_DISABLE_PROXY
+  if(!unix_path && conn->socks_proxy.host.name &&
+     !strncmp(UNIX_SOCKET_PREFIX"/",
```
bagder, line 3538: Note for another PR: we should change this to a case insensitive string check.

lib/url.c
emilengler: Well, I named it *new* because it makes a *fresh* look-up. Maybe rename it to fresh instead?

lib/url.c
emilengler: I think it is better suited there, because the function itself doesn't "decide" on it anymore.

lib/url.c
bagder: "fresh" is better in my eyes, yes

emilengler: Done.

bagder (approved):

bagder:
Thanks!

PR #9289, opened 2022-08-10
imap: use ISALNUM() for alphanumeric checks

emilengler:
This commit replaces a self-made character check for alphanumeric
characters within imap_is_bchar() with the ISALNUM() macro, as it is
reduces the size of the code and makes the performance better, due to
ASCII arithmetic.

Closes #XXXX


bagder (approved):

bagder:
Thanks!

PR #9290, opened 2022-08-10
Adding quic support via wolfSSL

icing:
- based on  wolfSSL (commit fa979230050dde47f6ac084c6f3b232a5e729eb5)
- based on ngtcp2 PR https://github.com/ngtcp2/ngtcp2/pull/505
- configure adapted to build against ngtcp2 wolfssl crypto lib
- quic code added for creation of WOLFSSL* instances

### DIFF
--- a/lib/vquic/ngtcp2.c
+++ b/lib/vquic/ngtcp2.c
@@ -38,6 +38,9 @@
 #elif defined(USE_GNUTLS)
 #include <ngtcp2/ngtcp2_crypto_gnutls.h>
 #include "vtls/gtls.h"
+#elif defined(USE_WOLFSSL)
+#include <ngtcp2/ngtcp2_crypto_wolfssl.h>
+#include "vtls/wolfssl.h"
 #endif
 #include "urldata.h"
 #include "sendf.h"
@@ -101,6 +104,11 @@ struct h3out {
   "+CHACHA20-POLY1305:+AES-128-CCM:-GROUP-ALL:+GROUP-SECP256R1:" \
   "+GROUP-X25519:+GROUP-SECP384R1:+GROUP-SECP521R1:" \
   "%DISABLE_TLS13_COMPAT_MODE"
+#elif defined(USE_WOLFSSL)
+#define QUIC_CIPHERS                                                          \
+  "TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_"               \
+  "POLY1305_SHA256:TLS_AES_128_CCM_SHA256"
+#define QUIC_GROUPS "P-256:P-384:P-521"
 #endif
 
 /* ngtcp2 default congestion controller does not perform pacing. Limit
@@ -201,6 +209,12 @@ static int keylog_callback(gnutls_session_t session, const char *label,
 
   Curl_tls_keylog_write(label, crandom.data, secret->data, secret->size);
   return 0;
+}
+#elif defined(USE_WOLFSSL)
+static void keylog_callback(const WOLFSSL *ssl, const char *line)
+{
+  (void)ssl;
+  Curl_tls_keylog_write_line(line);
 }
 #endif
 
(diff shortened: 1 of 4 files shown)


lib/vquic/ngtcp2.c
```diff
@@ -395,7 +409,117 @@ static int quic_init_ssl(struct quicsocket *qs)
   gnutls_server_name_set(qs->ssl, GNUTLS_NAME_DNS, hostname, strlen(hostname));
   return 0;
 }
+#elif defined(USE_WOLFSSL)
+
+static void wolfssl_logging(const int logLevel, const char *const logMessage)
+{
+    fprintf(stderr, " ** wolfssl[%d]: %s\n", logLevel, logMessage);
```
emilengler, line 416: Is there a specific reason to use libc `fprintf` instead of a curl-internal printf-pedant, such as `infof`?

lib/vquic/ngtcp2.c
```diff
@@ -395,7 +409,117 @@ static int quic_init_ssl(struct quicsocket *qs)
...
+}
+
+static WOLFSSL_CTX *quic_ssl_ctx(struct Curl_easy *data)
+{
+  struct connectdata *conn = data->conn;
+  WOLFSSL_CTX *ssl_ctx = wolfSSL_CTX_new(wolfTLSv1_3_client_method());
+
+#if 0
```
emilengler, line 424: Is there any reason for leaving this disabled sequence in the code?

lib/vquic/ngtcp2.c
```diff
@@ -395,7 +409,117 @@ static int quic_init_ssl(struct quicsocket *qs)
...
+
+  if(ngtcp2_crypto_wolfssl_configure_client_context(ssl_ctx) != 0) {
+    failf(data, "ngtcp2_crypto_wolfssl_configure_client_context failed");
+    return NULL;
+  }
+
+  wolfSSL_CTX_set_default_verify_paths(ssl_ctx);
+
+  if(wolfSSL_CTX_set_cipher_list(ssl_ctx, QUIC_CIPHERS) != 1) {
+    char error_buffer[256];
```
emilengler, line 439: `= { '\0' }`?

lib/vquic/ngtcp2.c
```diff
@@ -1569,8 +1701,14 @@ static CURLcode ng_has_connected(struct Curl_easy *data,
     if(result)
       return result;
     infof(data, "Verified certificate just fine");
-#else
+#elif defined(USE_GNUTLS)
     result = Curl_gtls_verifyserver(data, conn, conn->quic->ssl, FIRSTSOCKET);
+#elif defined(USE_WOLFSSL)
+    char *snihost = Curl_ssl_snihost(data, SSL_HOST_NAME(), NULL);
+    if(!snihost ||
```
emilengler, line 1708: Should this really be an *OR*? The latter expression makes less sense then.

lib/vquic/ngtcp2.c
bagder: That logging callback doesn't seem to have a custom pointer coming in so it will be hard to use `infof()` there.

But: the logging is `#if 0`'ed below so it seems this function should be as well? Or maybe even turn ifdef them on DEBUG_WOLFSSL or something to make it easier to switch on the debugging in a future build?

lib/vquic/ngtcp2.c
icing: I took this from `lib/vtls/wolfssl.c_616` as I thought QUIC and TLS should behave the same in this case. Have I mistunderstood?

lib/vquic/ngtcp2.c
icing: See comment above.

lib/vquic/ngtcp2.c
icing: I today regard this as an experiment. And am in favour of removing this entirely. If ever wolfSSL gets a "real" logging callback with a customer pointer, *then* I think this should be connected here.

But as things are, it is better removed.

lib/vquic/ngtcp2.c
icing: `ERR_error_string_n()` really makes no assumptions about the content of the buffer passed. We could 0 it, if your prefer this, though.

icing:
I have a question about the handling of a client cert. I could  not find it in the vtls case for wolfSSL and the vquic code in `ngtcp2.c:815` is missing its wolfSSL equivalent. Any ideas?

lib/vquic/ngtcp2.c
bagder: No, I believe it is correct!

lib/vquic/ngtcp2.c
bagder: Yeah, we can just as well remove that logging function. I'll do it when I merge.

bagder:
Thanks!

vszakats:
Bumped into the error below while testing a [wolfSSL build with H3](https://github.com/curl/curl-for-win/commit/5a0efe80eb44a8800607a0e891ad43dde43da0cd):

vquic/ngtcp2.c:418:9: error: use of undeclared identifier 'wolfssl_logging'
  (void)wolfssl_logging;
        ^
1 error generated.
Deleting the line above fixes it. Then remains this new warning:

vquic/ngtcp2.c:214:13: warning: unused function 'keylog_callback' [-Wunused-function]
static void keylog_callback(const WOLFSSL *ssl, const char *line)
            ^
1 warning generated.

bagder:
My fault.

PR #9291, opened 2022-08-10
configure: introduce CURL_SIZEOF

bagder:
This is a rewrite of the previously used GPL+exception licensed file.


m4/curl-functions.m4
```diff
@@ -5939,3 +5939,62 @@ AC_DEFUN([CURL_ATOMIC],[
     ])
   ])
 ])
+
+# Rewrite inspired by the functionality once provided by
+# AX_COMPILE_CHECK_SIZEOF. Uses the select() "trick" to find the size of the
```
pythontech, line 5944: Did you mean: switch() "trick"

bagder: oops, yes!

MarcelRaad (approved):

Issue #8915, opened 2022-05-25
Socket-leak in 'vquic/msh3.c'


gvanem:
Building with most options enabled and doing:
```
set CURL_MEMDEBUG=c:\temp\memdebug.curl
curl --http3 https://www.curl.se
perl tests\memanalyze.pl c:\temp\memdebug.curl
```

gives: `Open file descriptor created at connect.c:1632`

bagder:
@nibanks when I just now reproduced this issue with valgrind on my curl+msh3 build, it certainly shows leaking memory.

But the more primary problem I fell over is that I get a lot of "Conditional jump or move depends on uninitialised value(s)" errors from valgrind that overshadow the leak and that make it hard to work on the leak.

Here is my valgrind output from a curl build on current git master: [error.txt](https://github.com/curl/curl/files/9307795/error.txt) curl's output is here: 
[curl-out.txt](https://github.com/curl/curl/files/9307810/curl-out.txt)

I used this command line:

`valgrind --log-file=error ./src/curl --http3 https://curl.se -v`

nibanks:
I'll take a look!

nibanks:
I _think_ the following would fix the `EncodeHeaders` issue:
```
--- a/lib/msh3.hpp
+++ b/lib/msh3.hpp
@@ -52,6 +52,7 @@ enum H3SettingsType {
 // Contiguous buffer for (non-null-terminated) header name and value strings.
 struct H3HeadingPair : public lsxpack_header_t {
     char Buffer[512] = {0};
+    H3HeadingPair() { memset(this, sizeof(lsxpack_header_t), 0); }
     bool Set(const MSH3_HEADER* Header) {
         if (Header->NameLength + Header->ValueLength > sizeof(Buffer)) return false;
         buf = Buffer;
diff --git a/msquic b/msquic
index 030273f..3a46222 160000
--- a/msquic
+++ b/msquic
```
If you're able to add the new line to msh3.hpp to try while I look at the other error, I'd appreciate it.

nibanks:
For this one, I'm not sure why it claims there is uninitialized memory access.
```
==1929954== Syscall param sendmsg(mmsg[0].msg_hdr.msg_iov[0]) points to uninitialised byte(s)
==1929954==    at 0x4E40936: sendmmsg (sendmmsg.c:31)
==1929954==    by 0x4FAD649: CxPlatSocketSendInternal (in /home/daniel/build-msh3/lib/libmsquic.so.2.0.0)
[...]

PR #9297, opened 2022-08-11
Update Docs to msh3 v0.4.0

nibanks:
Updates msh3 usage to [latest release](https://github.com/nibanks/msh3/releases/tag/v0.4.0), which includes:

- Uses latest release of msquic ([v2.1](https://github.com/microsoft/msquic/releases/tag/v2.1.0)).
- A few fixes, including a partial one for #8915.

FYI, it'd be great to hook up some curl automation to verify things, but I'm not sure how to do that.

Also, I'm not sure it's worth mentioning here, but by moving to libmsquic.2.1.0, we have [official, signed packages published](https://packages.microsoft.com/ubuntu/20.04/prod/pool/main/libm/libmsquic/) that could be referenced instead of building it yourself.

### DIFF
--- a/docs/HTTP3.md
+++ b/docs/HTTP3.md
@@ -140,7 +140,7 @@ Build curl:
 
 Build msh3:
 
-     % git clone --depth 1 --recursive https://github.com/nibanks/msh3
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #9298, opened 2022-08-12
docs/cmdline-opts: fix example and categories for --form-escape

emanuele6:
The example was missing a "--form" argument
I also replaced "--form" with "-F" to shorten the line a bit since it
was already very long.

And I also moved --form-escape from the "post" category to the "upload"
category (this is what I originally wanted to fix, before also noticing
the mistake in the example).


bagder (approved):

bagder:
Thanks!

PR #9299, opened 2022-08-12
krb5: change SPDX-License-Identifier to LicenseRef-BSD-3-KTH

### LINKED ISSUE #9220: Licensing: some license identifiers potentially wrong/unprecise
On the [REUSE mailing list](https://lists.fsfe.org/pipermail/reuse/2022q2/000149.html), @karsten-klein pointed out some potential unprecise or wrong licensing information they found with their scanners. They also provided some extra comments on single points in [this email](https://lists.fsfe.org/pipermail/reuse/2022q2/000150.html). I think it's best to address this here. 

My comments, where due, are in the brackets behind the individual points):

1. `m4/ax_compile_check_sizeof.m4` should be tagged as "GPL-3.0-or-later WITH Autoconf-exception-2.0" instead of just be "GPL-3.0-or-later"
2. ~~`lib/curl_path.c` includes an ISC snippet, but this is not marked with an SPDX-License-Identifier~~ (fixed in #9073 already by @jlovejoy)
3. `lib/sha256.c` includes a quote that parts of the code are based on public domain. This is not captured. (Will be possible to do with snippets, upcoming feature of SPDX 2.3 and the next REUSE specification)
4. `lib/md4.c` includes contains code under Public Domain or simple redistribution terms (“heavily cut down BSD license”). This is not captured. (Will be possible to do with snippets, upcoming feature of SPDX 2.3 and the next REUSE specification)
5. `lib/md5.c` includes contains code under Public Domain or simple redistribution terms (“heavily cut down BSD license”). This is not captured. (Will be possible to do with snippets, upcoming feature of SPDX 2.3 and the next REUSE specification)
6. `lib/krb5.c` contains a variant of the BSD-3-Clause. The SPDX-License-Identifier says BSD-3-Clause. In `LICENSES/BSD-3-Clause.txt` is however the standard/default text not matching the original license text. (We can do that with [custom licenses](https://reuse.software/faq/#custom-license))

Being extremely picky:

7. The files

```
tests/data/test222
tests/data/test230
tests/data/test232
tests/data/test314
tests/data/test396
tests/data/test1123
```

contain references to X11 License and suggest that curl is under such license. I would propose to rework those test cases to not cause any ambiguity.

Depending on review by the maintainer(s) I should be able to prepare a pull request to fix these issue.

bagder:
The license header in the file has replaced the words "COPYRIGHT
HOLDER(S)" with "INSTITUTE" in three places compared to the original
BSD-3-Clause license which makes it a custom BSD-3 license.

That edit was presumably done by the original authors of the file in the
mid 90s.

Fixes #9220


bagder:
Googling the phrase `THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS` (one of the lines in the header) shows that this version of the BSD-3-clause has been used by KTH and the "Swedish Institute of Computer Science" in numerous places for decades. Both of them being Swedish higher academic organizations.

Perhaps most widely used by [Heimdal](https://github.com/heimdal/heimdal/blob/master/LICENSE)

bagder:
To understand if those three replaced words constitutes a "changed license" by SPDX standards, I read their [guideline](https://spdx.dev/license-list/matching-guidelines/). Here's the quote I'm using mostly:

> 2.1.3 Guideline:  Replaceable Text  Some licenses include text that refers to the specific copyright holder or author, yet the rest of the license is exactly the same as a generic version. The intent here is to avoid the inclusion of a specific name in one part of the license resulting in a non-match where the license is otherwise an exact match (e.g., the third clause and disclaimer in the BSD licenses, or the third, fourth, and fifth clauses of Apache 1.1). In these cases, there should be a positive license match.

I conclude that this is exactly what we have here. So, this in spite of this edit this: **"should be a positive license match"**

In other words, this PR should not be accepted as the license can still be considered to be `BSD-3-Clause` as per the SPDX definition.

PR #9303, opened 2022-08-12
connect: close the happy eyeballs loser connection when using QUIC

bagder:


bagder:
This fixes the larger part of #8915

lib/connect.c
```diff
@@ -910,6 +910,8 @@ CURLcode Curl_is_connected(struct Curl_easy *data,
         conn->tempsock[i] = CURL_SOCKET_BAD;
         post_SOCKS(data, conn, sockindex, connected);
         connkeep(conn, "HTTP/3 default");
+        if(conn->tempsock[other] != CURL_SOCKET_BAD)
+          Curl_quic_disconnect(data, conn, other);
```
nibanks, line 914: At least for msh3, I think we need to update `msh3_cleanup` to free the request (if non-null) too, to full fix the problem. Right?

bagder: Yeah, I have a follow-up separate PR for msh3 coming up that fixes the cleanup etc to make sure it behaves correctly and doesn't leak. See #9304 for a more complete take.

nibanks: Ok. Well, I can't approve the PR, but LGTM. 👍🏻

bagder: thanks!

PR #9307, opened 2022-08-13
NPN: remove support for and use of

bagder:
Next Protocol Negotiation is a TLS extension that was created and used for
agreeing to use the SPDY protocol (the precursor to HTTP/2) for HTTPS. In the
early days of HTTP/2, before the spec was finalized and shipped, the protocol
could be enabled using this extension with some servers.

curl supports the NPN extension with some TLS backends since then, with a
command line option `--npn` and in libcurl with `CURLOPT_SSL_ENABLE_NPN`.

HTTP/2 proper is made to use the ALPN (Application-Layer Protocol Negotiation)
extension and the NPN extension has no purposes anymore. The HTTP/2 spec was
published in May 2015.

Today, use of NPN in the wild should be extremely rare and most likely totally
extinct. Chrome removed NPN support in Chrome 51, shipped in
June 2016. Removed in Firefox 53, April 2017.


src/tool_getparam.c
```diff
@@ -736,7 +736,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
         global->tracetype = TRACE_BIN;
         break;
       case 'G': /* --npn */
-        config->nonpn = (!toggle)?TRUE:FALSE;
+        warnf(global, "--npn is no longer supported\n");
```
jay, line 739: this option is most commonly used as --no-npn, so the user may find it confusing to read --npn in that case
```suggestion
        warnf(global, "%s is no longer supported\n",
              (toggle ? "--npn" : "--no-npn"));
```

jay (approved):

MarcelRaad (approved):

PR #9311, opened 2022-08-14
memory debug tracking, various overflow check proposal.

devnexen:
cheap checks as possible at least for compilers providing those.

### DIFF
--- a/lib/setup-win32.h
+++ b/lib/setup-win32.h
@@ -51,6 +51,7 @@
 #    endif
 #  endif
 #  include <tchar.h>
+#  include <intsafe.h>
 #  ifdef UNICODE
      typedef wchar_t *(*curl_wcsdup_callback)(const wchar_t *str);
 #  endif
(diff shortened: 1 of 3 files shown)


bagder:
When you you expect these to actually trigger?

lib/setup-win32.h
```diff
@@ -51,6 +51,7 @@
 #    endif
 #  endif
 #  include <tchar.h>
+#  include <intsafe.h>
```
bagder, line 54: [log](https://dev.azure.com/daniel0244/curl/_build/results?buildId=10781&view=logs&j=94094f6a-a458-5f8b-6d31-7c5891e4b222&t=fb0e0277-bde5-5166-fa85-ad3bd70f67da&l=24)
~~~
setup-win32.h:54:12: fatal error: intsafe.h: No such file or directory
   54 | #  include <intsafe.h>

devnexen:
> When you you expect these to actually trigger?

you might be right, it was more an extra precaution than actual use.

PR #9312, opened 2022-08-14
cmake: improve detection of threadsafe feature

mback2k:
Avoids failing test 1014 by replicating configure checks
for HAVE_ATOMIC and _WIN32_WINNT with custom CMake tests.

Follow up to #8680 and #8997

### DIFF
--- a/CMake/CurlTests.c
+++ b/CMake/CurlTests.c
@@ -516,3 +516,50 @@ main() {
   return 0;
 }
 #endif
+#ifdef HAVE_ATOMIC
+/* includes start */
+#ifdef HAVE_SYS_TYPES_H
+#  include <sys/types.h>
+#endif
+#ifdef HAVE_UNISTD_H
+#  include <unistd.h>
+#endif
+#ifdef HAVE_STDATOMIC_H
+#  include <stdatomic.h>
+#endif
+/* includes end */
+
+int
+main() {
+  _Atomic int i = 1;
+  i = 0;  // Force an atomic-write operation.
+  return i;
+}
+#endif
+#ifdef HAVE_WIN32_WINNT
+/* includes start */
+#ifdef WIN32
+/*
+ * Don't include unneeded stuff in Windows headers to avoid compiler
+ * warnings and macro clashes.
+ * Make sure to define this macro before including any Windows headers.
+ */
+#  ifndef WIN32_LEAN_AND_MEAN
+#    define WIN32_LEAN_AND_MEAN
+#  endif
+#  ifndef NOGDI
+#    define NOGDI
+#  endif
+#  include "../lib/setup-win32.h"
+#endif
+/* includes end */
+
+#define enquote(x) #x
+#define expand(x) enquote(x)
+#pragma message("_WIN32_WINNT=" expand(_WIN32_WINNT))
+
+int
+main() {
+  return 0;
+}
+#endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -982,6 +982,7 @@ check_include_file_concat("pwd.h"            HAVE_PWD_H)
 check_include_file_concat("setjmp.h"         HAVE_SETJMP_H)
 check_include_file_concat("signal.h"         HAVE_SIGNAL_H)
 check_include_file_concat("ssl.h"            HAVE_SSL_H)
+check_include_file_concat("stdatomic.h"      HAVE_STDATOMIC_H)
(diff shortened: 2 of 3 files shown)


MarcelRaad (approved):
👍

mback2k:
@MarcelRaad with the current state of this PR I almost get all AppVeyor builds to work again regarding test 1014, but one is still failing and quite weird. My new CMake `_WIN32_WINNT` [detection test returns](https://ci.appveyor.com/project/curlorg/curl/builds/44533140/job/3owg7l4bx7a2hj5h?fullLog=true#L182) `0x0501` for it, but [later during the compilation](https://ci.appveyor.com/project/curlorg/curl/builds/44533140/job/3owg7l4bx7a2hj5h?fullLog=true#L264) of `easy_lock.h` the actual value is `0x0601` and I am not sure where this is coming from. Do you have any idea?

MarcelRaad:
Hmm, I don't see it on first glance either. This is Visual Studio 2010 with the Windows 7 SDK, which defaults to 0x0601, so that should be correct. I'll take another look later.

mback2k:
@MarcelRaad I was missing the include of `sdkddkver.h` in `CurlTests.c`. Now the Windows part is working correctly.

MarcelRaad:
But you had the include of windows.h, which should include sdkddkver.h itself, shouldn't it? And even if it didn't, nothing should define `_WIN32_WINNT` to `0x0501`. 😕 Whatever, glad that it works now and thanks for letting me know!

Still/again looks good to me. Only are the includes of curl_setup.h and sdkddkver.h both needed? I thought that curl_setup.h was needed because the include of windows.h, and ultimately sdkddkver.h, was expected to happen through that.

mback2k:
`sdkddkver.h` seems not to be automatically included by `windows.h`, so I am manually including `sdkddkver.h` first now to set the defaults based on the SDK environment. Then I include `curl_setup.h` to have the same behavior as an actual build would hopefully have, with all the includes that you mentioned.

MarcelRaad:
Strange. I checked several Windows SDK versions I have installed and one of the first things all of them do is unconditionally include <sdkddkver.h>, typically right after including <winapifamily.h> for recent versions. 😕

MarcelRaad:
Wow. config-win32.h is for non-CMake builds, but `HAVE_CONFIG_H` was not set yet for that CMake check, right? That absolutely makes sense then as setup-win32.h, which includes windows.h & friends, is only included later.

MarcelRaad:
Maybe we should include setup-win32.h directly then to avoid pulling in config-win32.h? Then we should have the same behavior as libcurl itself in regards to Windows SDK includes and definitions and don't need any additional includes.

mback2k:
Ok, so include `setup-win32.h` instead of `windows.h` or `sdkddkver.h` and `curl_setup.h`, right?

MarcelRaad:
yes 👍

mback2k:
AppVeyor looks good now, thanks @MarcelRaad. Unfortunately I had to include some boilerplate in the test before including setup-win32.h to avoid compiler warnings and errors. Also I am not sure regarding the relevance of the other CI failures yet.

MarcelRaad:
Ah, right, that stuff is still in curl_setup.h for the winapifamily.h include. I actually wanted to move that whole block to setup-win32.h, but that's not possible because of a circular dependency: config-win32.h needs the definition of `CURL_WINDOWS_APP`. That check needs `_WIN32_WINNT`, which is only reliably defined after the windows.h include in setup-win32.h. Which needs curl_config.h/config-win32.h again for the header availability macros. So probably the order is best as it is right now, as the user should define `_WIN32_WINNT` anyway when building for Windows.

But I think we could move the definition of `WIN32_LEAN_AND_MEAN` and `NOGDI` to setup-win32.h. AFAICS, the winapifamily.h needed for `CURL_WINDOWS_APP` is independent of those two macros.

MarcelRaad:
> Also I am not sure regarding the relevance of the other CI failures yet.

The macOS HTTP2 test failures show up in other PRs too.

mback2k:
> But I think we could move the definition of `WIN32_LEAN_AND_MEAN` and `NOGDI` to setup-win32.h. AFAICS, the winapifamily.h needed for `CURL_WINDOWS_APP` is independent of those two macros.

Okay, but let us save that for after fixing AppVeyor for now to see if that would break other things. Then the test can be updated accordingly.

MarcelRaad (approved):

PR #9317, opened 2022-08-15
url: reject URLs with hostnames longer than 65535 bytes

bagder:
It *probably* causes other problems too since DNS can't resolve such
long names, but the SNI field in TLS is limited to 16 bits length.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -2025,6 +2025,10 @@ static CURLcode parseurlandfillconn(struct Curl_easy *data,
     if(!strcasecompare("file", data->state.up.scheme))
       return CURLE_OUT_OF_MEMORY;
   }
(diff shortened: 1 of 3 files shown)


MarcelRaad:
Unfortunately, the new test fails on all Windows CI builds with
```
curl returned 126, when expecting 3

[...]
sh: ../src/curl.exe: Argument list too long
```

Windows is probably not the only platform with that problem, but I'm not exactly sure how to fix this in a generic way. My MinGW-w64 defines `ARG_MAX` to 14500 and `_POSIX_ARG_MAX` to 4096. I cannot find it at all in MSVC headers, but https://docs.microsoft.com/en-US/troubleshoot/windows-client/shell-experience/command-line-string-limitation says it's 8191 characters for the Windows command prompt.

PR #9324, opened 2022-08-16
libssh: setting atime or mtime > 32bit is now just skipped

bagder:
The libssh API used caps the time to an unsigned 32bit variable. Avoid
nasty surprises by instead not setting such time.

Spotted by Coverity.


bagder:
I posted a question/issue to the libssh devs on how this should be handled: https://gitlab.com/libssh/libssh-mirror/-/issues/138

PR #9325, opened 2022-08-16
libssh2: setting atime or mtime >32bit on 4-bytes-long systems

bagder:
Since the libssh2 API uses 'long' to store the timestamp, it cannot
transfer >32bit times on Windows and 32bit architecture builds.

Avoid nasty surprises by instead not setting such time.

Spotted by Coverity

### DIFF
--- a/lib/vssh/libssh2.c
+++ b/lib/vssh/libssh2.c
@@ -1766,8 +1766,15 @@ static CURLcode ssh_statemach_act(struct Curl_easy *data, bool *block)
           sshc->actualcode = CURLE_QUOTE_ERROR;
           break;
         }
-        sshp->quote_attrs.atime = (unsigned long)date;
-        sshp->quote_attrs.flags = LIBSSH2_SFTP_ATTR_ACMODTIME;
+#if SIZEOF_TIME_T > SIZEOF_LONG
+        if(date > 0xffffffff)
+          ;
+        else
+#endif
+        {
+          sshp->quote_attrs.atime = (unsigned long)date;
+          sshp->quote_attrs.flags = LIBSSH2_SFTP_ATTR_ACMODTIME;
+        }
       }
       else if(strncasecompare(cmd, "mtime", 5)) {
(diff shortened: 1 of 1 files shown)


bagder:
I filed an issue about this to the libssh2 project: https://github.com/libssh2/libssh2/issues/735

lib/vssh/libssh2.c
```diff
@@ -1766,8 +1766,15 @@ static CURLcode ssh_statemach_act(struct Curl_easy *data, bool *block)
           sshc->actualcode = CURLE_QUOTE_ERROR;
           break;
         }
-        sshp->quote_attrs.atime = (unsigned long)date;
-        sshp->quote_attrs.flags = LIBSSH2_SFTP_ATTR_ACMODTIME;
+#if SIZEOF_TIME_T > SIZEOF_LONG
+        if(date > 0xffffffff)
```
emilengler, line 1770: I personally would use `UINT_MAX` or at least add a comment explaining that this case is for >32 bit.

emilengler (commented):
> Avoid nasty surprises by instead not setting such time.

Shouldn't there be at least a warning?

lib/vssh/libssh2.c
bagder: It felt wrong to use an int related define when none of the variables involved are ints and I want to make the check exactly 32 bit.

I figure we could do infof() there, but probably it should even rather return error?

emilengler: > It felt wrong to use an int related define when none of the variables involved are ints and I want to make the check exactly 32 bit.

I totally follow you here, makes sense.

> I figure we could do infof() there, but probably it should even rather return error?

I think an error would probably be even better, because some software heavily relies on the mtime of files. For example, I've seen many website generators (no matter if dynamic or static) which make use of that value in order to sort the entries. A wrong mtime would be a disaster in such cases.

bagder: I made #9328

PR #9327, opened 2022-08-17
url: output the maximum when rejecting a url

emilengler:
This commit changes the failf message to output the maximum length,
when curl refuses to process a URL because it is too long.

See: #9317
Closes: #XXXX

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -154,6 +154,9 @@ static void conn_free(struct connectdata *conn);
 #define UNIX_SOCKET_PREFIX "localhost"
 #endif
 
+/* Reject URLs exceeding this length */
+#define MAX_URL_LEN 0xffff
+
 /*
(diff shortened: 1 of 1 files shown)


lib/url.c
```diff
@@ -2027,7 +2027,7 @@ static CURLcode parseurlandfillconn(struct Curl_easy *data,
       return CURLE_OUT_OF_MEMORY;
   }
   else if(strlen(data->state.up.hostname) > 0xffff) {
-    failf(data, "Too long host name");
+    failf(data, "Too long host name (maximum is 65535)");
```
bagder, line 2030: maybe make a `#define` for it and use it in both places?

emilengler: sounds good.

bagder (approved):

bagder:
Thanks!

PR #9329, opened 2022-08-17
curl.h: Include <sys/select.h> on SunOS

tnn2:
It is needed for fd_set to be visible to downstream consumers that
use <curl/multi.h>. Header is known to exist at least as far back
as Solaris 2.6.


bagder (approved):
:+1: this `#if` line might be the coolest one in the project :grin:

bagder:
Thanks!

PR #9337, opened 2022-08-19
Allow to configure DNS cache hash map size

uranix:
Following this thread https://curl.se/mail/lib-2021-03/0036.html I propose a simple patch that allows to configure DNS cache hash map size via a preprocessor variable CURL_DNSCACHE_HASH_TABLE_SIZE. This cache becomes a CPU bottleneck when there are thousands of entries.


bagder:
If you are right, and I have no reason to question that, then won't applications like the ability to actually set this instead of relying on a custom build?

uranix:
Is `curl_share_setopt` / `curl_multi_setopt` a proper place to do this?

PR #9338, opened 2022-08-20
Implement `curl_easy_init_ex` that returns an error code.

juchem:
In a production grade software project it is often useful to check, log
and/or collect error codes for better error handling, diagnostics and
bug fixing.

The majority of libcurl functions provide error codes as part of their
interface. `curl_easy_init` is one of the most important functions of
the library, yet it lacks proper means of retrieving an error code when
it fails, making the bug fixing process more involved on certain
situations.

Looking closer at the implementation of `curl_easy_init`, it's apparent
that the error codes are readily available. All that's missing is an
interface to extract them to the call site.

This commit aims to provide an alternative interface capable of
surfacing the error code, while otherwise being identical in behavior to
`curl_easy_init`. By not directly modifying `curl_easy_init`'s
interface, the library remains backwards compatible.

`curl_easy_init_ex`'s implementation is almost identical to
`curl_easy_init`, with slight changes to accomodate for the new
interface.

Having an identical interface to `Curl_open`, it allows `curl_easy_init`
to be implemented as a thin wrapper on top of it, adding virtually zero
overhead (likely zero through trivial compiler optimizations). This has
two main benefits:
- avoid code duplication;
- leverage the vast existing test suite already in place for
  `curl_easy_init` without requiring immediate effort towards quality
  assurance.

Documentation for `curl_easy_init` has also been updated to accommodate
the new interface.

### DIFF
--- a/docs/libcurl/curl_easy_init.3
+++ b/docs/libcurl/curl_easy_init.3
@@ -58,8 +58,16 @@ if(curl) {
 .SH AVAILABILITY
 Always
 .SH RETURN VALUE
-If this function returns NULL, something went wrong and you cannot use the
-other curl functions.
+Returns a valid non-null handle on success.
+
+On failure, returns NULL. If \fIcurl_global_init(3)\fP was called prior to
+\fIcurl_easy_init(3)\fP, then it means the process has ran out of memory. If
+\fIcurl_global_init(3)\fP was not called beforehand, it could also be due to a
+global initialization failure.
+
+In order to better isolate failure cases, the user is strongly advised to
+manually call \fIcurl_global_init(3)\fP instead of relying on libcurl doing it
+implicitly.
 .SH "SEE ALSO"
 .BR curl_easy_cleanup "(3), " curl_global_init "(3), " curl_easy_reset "(3), "
 .BR curl_easy_perform "(3) "


juchem:
FWIW, I've looked into making a similar contribution to `curl_multi_init`, but the change there seems a little bit less trivial.

I decided to first gauge interest in this addition, as well as keeping the changes small.

jay:
curl_easy_init failures are highly unlikely and pretty much fatal (read: out of memory). However, if the user forgot to call curl_global_init then curl_easy_init will do it implicitly, and yes it would be good to know the error code for that but IMO the user should just be calling curl_global_init.

bagder:
I am not convinced this adds enough new value to be worth the trouble of adding a new function. This function fails very rarely and I'm not sure that the existing error codes will and provide enough granularity to actually help a lot - since this doesn't have a way to return an error message.

juchem:
Understood. Those are reasonable rationales for not adding a new API.

@bagder given the well understood behavior of easy-init, would it then be outrageous to propose instead a change in its documentation, to give more strong guarantees about the fail case?

Right now the docs say "something went wrong", which is not very helpful without an error code. What about saying that either the process is out of memory, or initialization failed; possibly giving emphasis to the fact that, if global-init has been manually called before, then the only mode of failure is out-of-memory?

emilengler:
> Right now the docs say "something went wrong", which is not very helpful without an error code. What about saying that either the process is out of memory, or initialization failed; possibly giving emphasis to the fact that, if global-init has been manually called before, then the only mode of failure is out-of-memory?

It is very likely, that the value of `errno` will give further clues about what caused the failure in detail. I'm not quite sure if it's worth documenting though, because it is the responsibility of libc and not libcurl.

bagder:
@juchem so maybe updating and clarifying the current docs will be more productive then, than adding a new API just to be able to return `CURLE_OUT_OF_MEMORY` or `CURLE_FAILED_INIT`, neither which will help an application author much more than the current API...

juchem:
Closing this due to community consensus.

PR #9340, opened 2022-08-21
Introduce options to control dns cache hash table slots

uranix:
I accidentally closed my PR #9337, so here it is again

Following this thread https://curl.se/mail/lib-2021-03/0036.html I propose a simple patch that allows to configure DNS cache hash map size via a preprocessor variable CURL_DNS_CACHE_HASH_TABLE_SIZE. This cache becomes a CPU bottleneck when there are thousands of entries.

I also added CURLMOPT_DNS_CACHE_HASH_TABLE_SIZE / CURLSLOPT_DNS_CACHE_HASH_TABLE_SIZE options so the parameter could be tuned via curl_multi_setopt / curl_share_setopt


dfandrich:
I haven't looked at this patch, but isn't this something that should be set
automatically? Can't the hash table size be increased automatically as more
entries are added to keep lookups optimum? This seems like the kind of
housekeeping operation that the application needn't be burdened about.

uranix:
That would be rather complex and dangerous change to the hash table internals, I'm not confident I can do that without breaking things.

Personally I'd be satisfied with compile-time variable, but @bagder suggested there should be an API to adjust the value.

bagder:
How big do you set it with what kind of DNS cache size? Could we make this scale better without changing this value?

I didn't say it should be an API, I actually *asked* about it... It is still a value that is really hard for a user to know what to set and when.

bagder:
Maybe extending `Curl_multi_handle()` with an argument for the DNS hash table size would be enough? It could remain `7` for the easy interface and maybe get set to a sensible higher value for the multi interface (`71` maybe?). Could that perhaps be a good enough middle ground?

uranix:
On the second though I agree that user should not care about such details of implementation and the options are overkill. More appropriate constants are used in #9376 and I'm perfectly ok with them

PR #9345, opened 2022-08-22
documenting stronger guarantees for `curl_easy_init` failures

juchem:
In a production grade software project it is often useful to check, log
and/or collect error codes for better error handling, diagnostics and
bug fixing.

The majority of libcurl functions provide error codes as part of their
interface. `curl_easy_init` is one of the most important functions of
the library, yet it lacks proper means of retrieving an error code when
it fails, making the bug fixing process more involved on certain
situations.

The failures modes for `curl_easy_init` are well understood so after some
discussion in #9338, the community decided towards not adding a new interface
for surfacing the error codes of a failed easy handle creation.

This proposal is a follow up that aims to document and strengthen the API
contract by making the documentation clearer on the topic.

### DIFF
--- a/docs/libcurl/curl_easy_init.3
+++ b/docs/libcurl/curl_easy_init.3
@@ -58,8 +58,12 @@ if(curl) {
 .SH AVAILABILITY
 Always
 .SH RETURN VALUE
-If this function returns NULL, something went wrong and you cannot use the
-other curl functions.
+It returns a pointer to a CURL easy handle on success.
+
+On failure, it returns NULL. If \fIcurl_global_init(3)\fP was called prior to
+\fIcurl_easy_init(3)\fP, then it means the process has ran out of memory.
+Future revisions may add other failure modes. If \fIcurl_global_init(3)\fP was
+not called beforehand, it could also be due to a global initialization failure.
 .SH "SEE ALSO"
 .BR curl_easy_cleanup "(3), " curl_global_init "(3), " curl_easy_reset "(3), "
 .BR curl_easy_perform "(3) "


juchem:
@bagder this is a follow up from the [`curl_easy_init_ex` proposal](https://github.com/curl/curl/pull/9338) that touches only documentation for `curl_easy_init`.

Let me know if there's any interest in such a documentation change, as well as necessary changes which I'll gladly apply.

docs/libcurl/curl_easy_init.3
```diff
@@ -58,8 +58,16 @@ if(curl) {
 .SH AVAILABILITY
 Always
 .SH RETURN VALUE
-If this function returns NULL, something went wrong and you cannot use the
-other curl functions.
+Returns a valid non-null handle on success.
```
bagder, line 61: Maybe "it returns a pointer to a CURL easy handle on success." instead?

Talking about "valid non-null" might imply that there can be a invalid non-null somehow.

docs/libcurl/curl_easy_init.3
```diff
@@ -58,8 +58,16 @@ if(curl) {
...
+
+On failure, returns NULL. If \fIcurl_global_init(3)\fP was called prior to
```
bagder, line 63: On failure, **it** returns NULL.

docs/libcurl/curl_easy_init.3
```diff
@@ -58,8 +58,16 @@ if(curl) {
...
+\fIcurl_easy_init(3)\fP, then it means the process has ran out of memory. If
+\fIcurl_global_init(3)\fP was not called beforehand, it could also be due to a
+global initialization failure.
```
bagder, line 66: I don't think it should be this specific. We don't want to exclude the possibility that this function can fail differently in a future.

docs/libcurl/curl_easy_init.3
```diff
@@ -58,8 +58,16 @@ if(curl) {
...
+
+In order to better isolate failure cases, the user is strongly advised to
+manually call \fIcurl_global_init(3)\fP instead of relying on libcurl doing so
+implicitly.
```
bagder, line 70: This is already stated above, no need to repeat.

docs/libcurl/curl_easy_init.3
bagder: How is the user supposed to know when that was written and if their current version is such a future version? They probably are already using a future version when they read that and then having the "future version" text is superfluous and this docs is mostly phrased in a peculiar way.

juchem: By the way, throughout this discussion there's an underlying assumption that out-of-memory errors are rare. Although that has been true in recent history with memory getting bigger and cheaper, that's only true when the software has full reigns of the machine.

In a containerized world, it's increasingly common to enforce soft and hard memory limits, specially on 3rd-party services and agents. I've seen first hand cases where debugging time lowers dramatically by having an out-of-memory error code vs an uninformed crash. That being said, there are other ways to detect out-of-memory, yet detecting by proxy remains less powerful than detecting _in loco_.

That said, it seems that the docs are already in the intended state and there's no interested in the proposed change, so I'll just drop this PR.

Thank you all for your time reviewing and sharing your experience as maintainers.

bagder: curl is used in a *wide* variety of machines, environments and circumstances. Some of them with tiny system resources. I don't think out-of-memory is rare in all those instances. I don't see how this is very relevant to this PR though.

Issue #9346, opened 2022-08-22
Curl --cert/-E can't load cert with ':' in filename


jacobtolar:
### I did this

```
curl --key /path/to/key --cert /path/to/some-name:not-a-password.cert https://.../
```

### I expected the following
Successful use of client cert

### Observed behavior

```
curl: (58) could not load PEM client certificate, OpenSSL error error:02001002:system library:fopen:No such file or directory, (no key found, wrong pass phrase, or wrong file format?)
```

### curl/libcurl version

Built from master branch just now: 

```
./configure --disable-shared --enable-debug --enable-maintainer-mode LDFLAGS="-L/opt/homebrew/Cellar/openssl@1.1/1.1.1q/lib" CFLAGS="-I/opt/homebrew/Cellar/openssl@1.1/1.1.1q/include" --with-openssl --prefix=/tmp/curlbuild
make -j8 install

/tmp/curlbuild/bin/curl -V
curl 7.85.0-DEV (aarch64-apple-darwin21.5.0) libcurl/7.85.0-DEV OpenSSL/1.1.1q zlib/1.2.11 libidn2/2.3.3 nghttp2/1.48.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS Debug HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP TrackMemory UnixSockets
```

### operating system

```
uname -a
Darwin V30123H0G5 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64
```

jacobtolar:
The cert parameter is, off course, documented as:

```
       -E, --cert <certificate[:password]>
```

```
              If curl is built against the NSS SSL library then this option can tell curl the nickname of the certificate to use within the NSS database defined by the
              environment variable SSL_DIR (or by default /etc/pki/nssdb). If the NSS PEM PKCS#11 module (libnsspem.so) is available then PEM files may be loaded. If you want
              to use a file from the current directory, please precede it with "./" prefix, in order to avoid confusion with a nickname.  If the nickname contains ":", it
[...]

PR #9349, opened 2022-08-22
Document --cert escape handling

jacobtolar:
See: https://github.com/curl/curl/issues/9346

Includes two commits:

1) Clarify in the documentation for `--cert` that escape sequences for `:` and `/` apply to file paths as well as NSS nicknames

2) In lib/vtls/openssl.c, prints the file path (if available) of the certificate that was used if an error occurs when loading it.

### DIFF
--- a/docs/cmdline-opts/cert.d
+++ b/docs/cmdline-opts/cert.d
@@ -32,11 +32,6 @@ loaded.
 If you provide a path relative to the current directory, you must prefix the
 path with "./" in order to avoid confusion with an NSS database nickname.
 
-If the NSS nickname or certificate filename contains the character ":", it
-must be prefixed by "\\" so that it is not recognized as the password
-delimiter. Similarly, if the nickname or filename contains "\\", it must be
-escaped as "\\\\" so that it is not recognized as an escape character.
-
 If curl is built against OpenSSL library, and the engine pkcs11 is available,
 then a PKCS#11 URI (RFC 7512) can be used to specify a certificate located in
 a PKCS#11 device. A string beginning with "pkcs11:" will be interpreted as a
--- a/docs/cmdline-opts/cert.d
+++ b/docs/cmdline-opts/cert.d
@@ -14,9 +14,14 @@ Tells curl to use the specified client certificate file when getting a file
 with HTTPS, FTPS or another SSL-based protocol. The certificate must be in
 PKCS#12 format if using Secure Transport, or PEM format if using any other
 engine. If the optional password is not specified, it will be queried for on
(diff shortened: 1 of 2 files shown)


docs/cmdline-opts/cert.d
```diff
@@ -14,19 +14,24 @@ Tells curl to use the specified client certificate file when getting a file
 with HTTPS, FTPS or another SSL-based protocol. The certificate must be in
 PKCS#12 format if using Secure Transport, or PEM format if using any other
 engine. If the optional password is not specified, it will be queried for on
-the terminal. Note that this option assumes a \&"certificate" file that is the
-private key and the client certificate concatenated! See --cert and --key to
-specify them independently.
+the terminal if required. Note that this option assumes a \&"certificate"
```
bagder, line 17: It feels a bit superfluous to mention both "optional" and "required" in the same sentence. Maybe remove "optional" here?

docs/cmdline-opts/cert.d
```diff
@@ -14,19 +14,24 @@ Tells curl to use the specified client certificate file when getting a file
...
 by the environment variable SSL_DIR (or by default /etc/pki/nssdb). If the
 NSS PEM PKCS#11 module (libnsspem.so) is available then PEM files may be
-loaded. If you want to use a file from the current directory, please precede
-it with "./" prefix, in order to avoid confusion with a nickname. If the
-nickname contains ":", it needs to be preceded by "\\" so that it is not
-recognized as password delimiter. If the nickname contains "\\", it needs to
-be escaped as "\\\\" so that it is not recognized as an escape character.
+loaded.
+
+If you intend to use a certificate file and provide a path relative to the
```
bagder, line 27: We can make it shorter:

    If you provide a file path relative to the

docs/cmdline-opts/cert.d
```diff
@@ -14,19 +14,24 @@ Tells curl to use the specified client certificate file when getting a file
...
+current directory, you must prefix the path with "./" in order to avoid
+confusion with an NSS database nickname.
+
+If the NSS nickname or certificate filename contains the character ":", it
+must be prefixed by "\\" so that it is not recognized as the password
+delimiter. Similarly, if the nickname or filename contains "\\", it must be
+escaped as "\\\\" so that it is not recognized as an escape character.
```
bagder, line 34: You add the mention of the escaping for file names in the section for NSS details which not reads as if this is only for NSS somehow, when it really isn't. Right?

bagder: It sounds sensible.

bagder:
Thanks (slightly edited before merge)!

PR #9360, opened 2022-08-24
CI/runtests.pl: add param for dedicated curl to talk to APIs

mback2k:
This should make it possible to also report test failures
if our freshly build curl binary is not fully functional.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -95,7 +95,7 @@ stages:
       displayName: 'test'
       env:
         AZURE_ACCESS_TOKEN: "$(System.AccessToken)"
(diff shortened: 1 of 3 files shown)


Developer64688 (commented):
Hecho

mback2k:
@bagder requesting your review primarily for the `runtests.pl` changes.

bagder (approved):
I can't quite see why `-vc` couldn't just be used, but I'm fine with this and the change looks fine.

mback2k:
A few months ago you said not to set -vc on CI, so I removed it from the test-ci target. Also the system curl may not have all features we are testing on CI and cannot validate all relevant test servers.

PR #9361, opened 2022-08-24
multi_wait: fix skipping to populate revents for extra_fds

mback2k:
Splitting from #9203 to merge before the pending release:

On Windows revents was not populated for extra_fds if
multi_wait had to wait due to the Curl_poll pre-check
not signalling any readyness. This commit fixes that.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1366,8 +1366,9 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
         unsigned r = ufds[curlfds + i].revents;
         unsigned short mask = 0;
 #ifdef USE_WINSOCK
+        curl_socket_t s = extra_fds[i].fd;
         wsa_events.lNetworkEvents = 0;
-        if(WSAEnumNetworkEvents(extra_fds[i].fd, NULL, &wsa_events) == 0) {
+        if(WSAEnumNetworkEvents(s, NULL, &wsa_events) == 0) {
           if(wsa_events.lNetworkEvents & (FD_READ|FD_ACCEPT|FD_CLOSE))
             mask |= CURL_WAIT_POLLIN;
           if(wsa_events.lNetworkEvents & (FD_WRITE|FD_CONNECT|FD_CLOSE))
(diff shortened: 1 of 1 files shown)


mback2k:
Waiting for #9364 to be merged to re-run tests.

lib/multi.c
```diff
@@ -1361,9 +1362,11 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
           if(ret && pollrc <= 0 && wsa_events.lNetworkEvents)
             retcode++;
         }
-        WSAEventSelect(extra_fds[i].fd, multi->wsa_event, 0);
-        if(pollrc <= 0)
+        WSAEventSelect(s, multi->wsa_event, 0);
+        if(pollrc <= 0) {
+          extra_fds[i].revents = mask;
```
MarcelRaad, line 1367: Maybe get rid of the `mask` variable and always use `extra_fds[i].revents` instead to avoid this mistake completely?

MarcelRaad (approved):

lib/multi.c
mback2k: I think the mask variable is there to avoid redundant array access in this performance relevant code path.

bagder: Should this still ignore Curl_poll() returning -1 ? We fixed it elsewhere previously in 5912da253

jay: > I think the mask variable is there to avoid redundant array access in this performance relevant code path.

I don't see anything wrong with it but I doubt we are saving many cycles here.

> Should this still ignore Curl_poll() returning -1 ? We fixed it elsewhere previously in [5912da2](https://github.com/curl/curl/commit/5912da253b64de3cb2a1a229558077219b2c8a35)

It seems that fix covered this code as well, it's just not visible in the review unless you use the up arrow to expand more lines. Though further up there is a place where it could happen,

https://github.com/curl/curl/blob/45ac4d019475df03562fe0ac54eb67e1d1de0ca7/lib/multi.c#L1307-L1314

jay (approved):

lib/multi.c
mback2k: I will look into this in a separate follow-up PR.

mback2k:
Just checked: failed CIs are irrelevant for this change. So going to merge this.

lib/multi.c
mback2k: > Should this still ignore Curl_poll() returning -1 ? We fixed it elsewhere previously in [5912da2](https://github.com/curl/curl/commit/5912da253b64de3cb2a1a229558077219b2c8a35)

@bagder I believe https://github.com/curl/curl/pull/9372 could be a good attempt at improving this situation here.

Issue #9364, opened 2022-08-25
macOS CI jobs fail HTTP/2 tests 1700 + 1701


bagder:
Since a few days back, most (all?) of them fail. [example log](https://github.com/curl/curl/runs/8006644474?check_suite_focus=true)

jay:
~~~
2022-08-25T00:48:19.104Z 2176 2177 3ae25fb5 INFO (shrpx_http_downstream_connection.cc:1479) [DCONN:0x010e5acf00] HTTP parser failure: (HPE_CR_EXPECTED) Missing expected CR after header value
 2022-08-25T00:48:19.105Z 2176 2177 3ae25fb5 INFO (shrpx_http2_upstream.cc:1288) [DCONN:0x010e5acf00] HTTP parser failure
~~~

but it doesn't say what header

@tatsuhiro-t

tatsuhiro-t:
llhttp which nghttpx uses for HTTP/1.x tightens validation on header field separator and it now only accepts CR + LF.
I think the backend returned only LF without CR.

jay:
Ok. #9365 is the attempted fix for those two tests

PR #9365, opened 2022-08-25
tests: fix http2 tests to use CRLF headers

jay:
Prior to this change test1700 and test1701 did not use CRLF headers
everywhere. Recent changes in nghttp2 (??? ref here) nghttpx proxy
requires curl's HTTP/1.1 test server to use CRLF headers.

Fixes https://github.com/curl/curl/issues/9364
Closes #xxxx


jay:
I've expanded this to include tests 358 and 359 as a preventative measure. I can't tell if they are affected.

MarcelRaad:
Yes, 358 was failing too: https://github.com/curl/curl/runs/8006644539?check_suite_focus=true#step:10:7334

bagder (approved):

mback2k (approved):

kdudka (approved):
Thanks!  This fixes the failure for me on Fedora rawhide with `nghttp2-1.49.0`.  As for reference, `git bisect` points to the following commit: https://github.com/nghttp2/nghttp2/commit/9d389e874484d56bdf84f696b58381d5e4c72014

PR #9367, opened 2022-08-25
configure: if asked to use TLS, fail if no TLS lib was detected

bagder:
Previously the configure script would just warn about this fact and continue with TLS disabled build which is not always helpful. TLS should be explicitly disabled if that is what the user wants.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1969,6 +1969,9 @@ x1)
   SSL_ENABLED="1"
   AC_MSG_NOTICE([built with one SSL backend])
   ;;
+xD)
+  # explicitly built without TLS
+  ;;
 *)
   # more than one SSL backend is enabled
   AC_SUBST(SSL_ENABLED)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1955,11 +1957,11 @@ if test "x$curl_cv_native_windows" = "xyes" &&
   LIBS="-lbcrypt $LIBS"
 fi
 
-case "x$OPENSSL_ENABLED$GNUTLS_ENABLED$NSS_ENABLED$MBEDTLS_ENABLED$WOLFSSL_ENABLED$SCHANNEL_ENABLED$SECURETRANSPORT_ENABLED$BEARSSL_ENABLED$RUSTLS_ENABLED"
+case "x$OPENSSL_ENABLED$GNUTLS_ENABLED$NSS_ENABLED$MBEDTLS_ENABLED$WOLFSSL_ENABLED$SCHANNEL_ENABLED$SECURETRANSPORT_ENABLED$BEARSSL_ENABLED$RUSTLS_ENABLED$SSL_DISABLED"
```
MarcelRaad, line 1960: Won't we run into the multi-SSL case then with SSL disabled?

bagder: Oops, yes we do. I need to special-case that!

PR #9368, opened 2022-08-25
multi: use a pipe instead of a socketpair on apple platforms

### LINKED ISSUE #6132: curl_multi_poll always returns immediately if the wakeup socket has been disconnected
`curl_multi_poll` delegates to `Curl_multi_wait`, which checks `multi->wakeup_pair[0]` for POLLIN activity then attempts to read from it.

Unfortunately, waking an iPad breaks the wakeup socket pair, triggering a `POLLIN` event and resulting in `SOCKERRNO` being set to `ENOTCONN` in `Curl_multi_wait`'s final `#ifdef ENABLE_WAKEUP` block.

This condition, and other possible error conditions on the wakeup socket, are not handled, so the condition remains on the FD and `curl_multi_poll` will never block again.

Even if the iOS wake scenario is to be considered an edge case, it seems to me we'd be best off looking for and handling any wakeup socket errors anyway.

rcombs:
Sockets may be shut down by the kernel when the app is moved to the background,
but pipes are not.

Removed from KNOWN_BUGS

Fixes #6132

### DIFF
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -120,7 +120,6 @@ problems may have been fixed or changed somewhat since this was written.
  11.10 Blocking socket operations in non-blocking API
  11.11 A shared connection cache is not thread-safe
  11.12 'no_proxy' string-matches IPv6 numerical addresses
- 11.13 wakeup socket disconnect causes havoc
  11.14 Multi perform hangs waiting for threaded resolver
  11.15 CURLOPT_OPENSOCKETPAIRFUNCTION is missing
  11.16 libcurl uses renames instead of locking for atomic operations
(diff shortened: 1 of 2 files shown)


bagder:
Apple platforms you say, but is this also true for macOS or is it actually iOS etc it affects?

rcombs:
The issue only affects iOS/iPadOS/tvOS/watchOS; I gated the #ifdef on all Apple platforms since it's a simpler check and it's safe to use pipes on macOS as well (plus it makes it easier to test that everything works properly).

bagder:
Right thanks, I think I agree that it makes sense to use the same approach on macOS.

bagder:
I can imagine that with this groundwork in place, a few other platforms might prefer the pipe as well going forward...

rcombs:
Yeah, TBH I would expect it's _probably_ safe to use on all non-Windows (and Windows doesn't use the socketpair code anymore anyway); I just gated it to apple since I can't test all the more obscure targets that might have weird semantics I don't know about.

bagder:
> I just gated it to apple since I can't test all the more obscure targets

I think that is good, it makes it a more controlled and gradual transition.

bagder (approved):

bagder:
Thanks!

mback2k:
Just a best-practice note: this should probably have waited and be rebased onto #9365 first, to ensure the macOS tests still work.

bagder:
That PR fixed HTTP/2 tests (that wasn't even macOS specific), while this change has nothing at all with h2 to do - and I noticed lots of CI builds building and using this code on macOS and other systems. It was clear to me that this change would not break with or without 9365.

PR #9369, opened 2022-08-25
BugFix: SymLink name is not correctly read for SFTP in LibSSH2

skundu07:
When reading symbolic link name for a file, we need to add the file name to base path.


bagder (approved):
Nice catch!

bagder:
Just this little nit:
~~~
vssh/libssh2.c:2341:103: warning: Longer than 79 columns (LONGLINE)
               result = Curl_dyn_addf(&sshp->readdir_link, "%s%s", sshp->path, sshp->readdir_filename);

PR #9371, opened 2022-08-25
CI/azure: align torture shallowness with GHA

mback2k:
There 25 is used with FTP tests skipped, and 20 for FTP tests.
This should make torture tests stay within the 60min timeout.


bagder (approved):
Yeah, making sure it gets to run through is better than having a higher value and timeout.

PR #9375, opened 2022-08-26
lib: consolidate Win32 setup definitions in setup-win32.h

mback2k:
Suggested-by: Marcel Raad

Follow up to #9312

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -50,17 +50,6 @@
 #endif
 
 #ifdef WIN32
-/*
- * Don't include unneeded stuff in Windows headers to avoid compiler
- * warnings and macro clashes.
- * Make sure to define this macro before including any Windows headers.
- */
-#  ifndef WIN32_LEAN_AND_MEAN
-#    define WIN32_LEAN_AND_MEAN
-#  endif
-#  ifndef NOGDI
-#    define NOGDI
-#  endif
 /* Detect Windows App environment which has a restricted access
  * to the Win32 APIs. */
 # if (defined(_WIN32_WINNT) && (_WIN32_WINNT >= 0x0602)) || \
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -50,6 +50,17 @@
 #endif
 
 #ifdef WIN32
+/*
+ * Don't include unneeded stuff in Windows headers to avoid compiler
+ * warnings and macro clashes.
+ * Make sure to define this macro before including any Windows headers.
+ */
+#  ifndef WIN32_LEAN_AND_MEAN
+#    define WIN32_LEAN_AND_MEAN
+#  endif
+#  ifndef NOGDI
+#    define NOGDI
+#  endif
 /* Detect Windows App environment which has a restricted access
  * to the Win32 APIs. */
 # if (defined(_WIN32_WINNT) && (_WIN32_WINNT >= 0x0602)) || \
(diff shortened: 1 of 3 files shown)


MarcelRaad (approved):
Thanks!

lib/curl_setup.h
```diff
@@ -50,17 +50,6 @@
...
- * Don't include unneeded stuff in Windows headers to avoid compiler
- * warnings and macro clashes.
- * Make sure to define this macro before including any Windows headers.
- */
-#  ifndef WIN32_LEAN_AND_MEAN
-#    define WIN32_LEAN_AND_MEAN
-#  endif
-#  ifndef NOGDI
-#    define NOGDI
-#  endif
```
jay, line 63: I don't think it is a good idea to remove it here, we want it defined as early as possible in case anything that comes after includes something that includes windows.h

MarcelRaad: According to this comment, it should be safe:
https://github.com/curl/curl/blob/c3b4090d835ac1fe945020a15ca71bddac30fbe0/lib/curl_setup.h#L210
But right, probably all these platform setup includes should be before the curl.h include. Any idea why they're not? Especially the `_UNICODE` define makes no sense after including any C standard library headers.

jay: I don't know why curl.h is included before the system includes, that's a good question, it wasn't always like that. 9506d01 added it @bagder

> Especially the `_UNICODE` define makes no sense

I assume you are referring to this in setup-win32, which comes after curl.h is included (which includes windows.h):

https://github.com/curl/curl/blob/45ac4d019475df03562fe0ac54eb67e1d1de0ca7/lib/setup-win32.h#L38-L46

I think we didn't notice because all build systems that can make a Windows Unicode build define both UNICODE and _UNICODE together. It could be changed to a sanity check

~~~diff
diff --git a/lib/setup-win32.h b/lib/setup-win32.h
index c16928d..6e897bc 100644
--- a/lib/setup-win32.h
+++ b/lib/setup-win32.h
@@ -37,10 +37,10 @@
 
 #ifdef HAVE_WINDOWS_H
 #  if defined(UNICODE) && !defined(_UNICODE)
-#    define _UNICODE
+#    error "UNICODE is defined but _UNICODE is not defined"
 #  endif
 #  if defined(_UNICODE) && !defined(UNICODE)
-#    define UNICODE
+#    error "_UNICODE is defined but UNICODE is not defined"
 #  endif
 #  include <winerror.h>
 #  include <windows.h>
~~~

bagder: As I recall it, my change in https://github.com/curl/curl/commit/9506d01ee50d5908138ebad0fd9fbd39b66bd64d was primarily to stop including the other headers we don't have anymore, and it seemed to work so I didn't think twice about it...

mback2k: @MarcelRaad @jay I am a little bit lost now, can you take it from here for me?

jay: I've pushed a commit with my suggestion which is don't remove it from curl-setup. Separately from that I will propose a PR to change the UNICODE logic as mentioned earlier but commit it when the feature window opens. As to curl.h include I think it is probably? ok to move it to the end. If @bagder is ok with it we can try moving it to the end after the feature window opens. Include order is like jenga sometimes.

MarcelRaad: Makes sense to me! Also, including system.h instead of curl.h should probably be sufficient now that curlbuild.h and curlrules.h are gone.

mback2k: Should I merge this PR as it is right now or will there further changes? @jay @MarcelRaad

MarcelRaad: I think this is good as it is! We can work on the curl.h include when this is merged.

PR #9376, opened 2022-08-27
multi: use larger dns hash table for multi interface

bagder:
Have curl_multi_init() use a much larger DNS hash table than used for the easy interface to scale and perform better when used with _many_ host names.

curl_share_init() sets an in-between size.

Inspired-by: Ivan Tsybulin
See #9340


bagder:
@uranix, what do you think about this approach?

uranix:
It is great! 71 is probably fine for our environment and if not could be tuned in compile-time.

PR #9381, opened 2022-08-29
cookie: reject cookies with "control bytes"

bagder:


bagder:
The libssh errors are taken to #9382

PR #9387, opened 2022-08-29
ftp: ignore a 550 response to MDTM

### LINKED ISSUE #9357: FTP download fails with "No such file or directory" when CURLOPT_FILETIME is enabled
### I did this

I want to download a file from an FTP server and enabled CURLOPT_FILETIME by default in my curl context.
However, some FTP servers are configured more strictly so that the MDTM command is rejected with 550 error "Permission denied".
curl translates the 550 to "File not found", however, the file is actually present and when I don't set CURLOPT_FILETIME, then the download works fine. See the following trace:

```
*   Trying 1.1.1.1...
* TCP_NODELAY set
* Connected to ftp.example.com (1.1.1.1) port 21 (#0)
< 220 ProFTPD Server (Ftp Server) [1.1.1.1]
 [1.1.1.1]
> AUTH SSL
< 500 AUTH not understood
> AUTH TLS
< 500 AUTH not understood
> USER myuser
< 331 Password required for myuser
> PASS foobarpassword
< 230 User myuser logged in
> PWD
< 550 PWD: Permission denied
> MDTM 1_0_7.image
* ftp_perform ends with SECONDARY: 0
< 550 1_0_7.image: Operation not permitted
* Given file does not exist
* Remembering we are in dir ""
* Connection #0 to host ftp.example.com left intact
```

### I expected the following

I would expect that the download still works with CURLOPT_FILETIME set - personally, I consider it more as optional thing - if it works then it's fine - if not then it's still ok. In function ftp_state_mdtm_resp for example an unsupported reply is also handled in such a way, that no error is thrown.
However, I know that the FTP server responses are some kind of implementation specific: for some servers 550 is really "File not found", some servers (as above) use it for permission things.
In my eyes, there are three options to handle it:
1. don't fail on every 550 anymore in ftp_state_mdtm_resp, just ignore it -> this would defer the error in case the file really does not exists
2. introduce something like CURLOPT_FILETIME_OPTIONAL which deals with this special situation and allows the library user to fine-control the behavior -> it would only fail hard if the user does not give this flag (keep current behavior) but ignores the 550 in case this flag is set
3. not only look at the numeric error code but use a heuristic on the plain text description, e.g. look for "found" or "permitted" in the assumption that the FTP server always respon in English and decide upon the found string whether the 550 can be ignored or not
4. any other idea :-)

I want to discuss here, whether you also consider it as issue and which solution should be worked on.

### curl/libcurl version

curl 7.58.0 (x86_64-pc-linux-gnu) libcurl/7.58.0 OpenSSL/1.1.1 zlib/1.2.11 libidn2/2.0.4 libpsl/0.19.1 (+libidn2/2.0.4) nghttp2/1.30.0 librtmp/2.3
Release-Date: 2018-01-24
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz TLS-SRP HTTP2 UnixSockets HTTPS-proxy PSL

But this is still present in latest Github branch.

### operating system

I'm still on Ubuntu 18.04 on my dev host, but I also tested with newer libcurl on an embedded Linux system (Yocto based).

mhei:
The 550 is overused as a return code for multiple error case, e.g.
file not found and/or insufficient permissions to access the file.

So we cannot fail hard in this case.

Adjust test 511 since we now fail later.
Add new test 3027 which check that when MDTM failed, but the file could
actually be retrieved, that in this case no filetime is provided.

Reported-by: Michael Heimpold
Fixes #9357

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -2131,9 +2131,11 @@ static CURLcode ftp_state_mdtm_resp(struct Curl_easy *data,
   default:
     infof(data, "unsupported MDTM reply format");
     break;
-  case 550: /* "No such file or directory" */
-    failf(data, "Given file does not exist");
-    result = CURLE_REMOTE_FILE_NOT_FOUND;
+  case 550: /* 550 is used for several different problems, e.g.
+               "No such file or directory" or "Permission denied".
+               It does not mean that the file does not exist at all. */
+    infof(data, "MDTM failed: file does not exist or permission problem,"
+          " continuing");
     break;
   }
 
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -245,4 +245,4 @@ test2200 test2201 test2202 test2203 test2204 test2205 \
 test3000 test3001 test3002 test3003 test3004 test3005 test3006 test3007 \
 test3008 test3009 test3010 test3011 test3012 test3013 test3014 test3015 \
 test3016 test3017 test3018 test3019 test3020 test3021 test3022 test3023 \
-test3024 test3025 test3026
+test3024 test3025 test3026 test3027
--- a/tests/data/test3027
+++ b/tests/data/test3027
@@ -2,16 +2,21 @@
 <info>
 <keywords>
 FTP
+CURLOPT_FILETIME
+MDTM fail
 </keywords>
 </info>
 
-#
 # Server-side
 <reply>
 <servercmd>
-REPLY MDTM 550 bluah you f00l!
-REPLY SIZE 550 bluah you f00l!
+REPLY MDTM 550 Permission denied
 </servercmd>
+<data>
+data blobb
+</data>
+
+# data is sent to stdout
 </reply>
 
 # Client-side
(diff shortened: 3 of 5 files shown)


bagder:
try `make checksrc` to verify the code style...

lib/ftp.c
```diff
@@ -2131,9 +2131,11 @@ static CURLcode ftp_state_mdtm_resp(struct Curl_easy *data,
...
     infof(data, "unsupported MDTM reply format");
     break;
-  case 550: /* "No such file or directory" */
-    failf(data, "Given file does not exist");
-    result = CURLE_REMOTE_FILE_NOT_FOUND;
+  case 550: /* 550 is used for several different problems, e.g.
+               "No such file or directory" or "Permission denied".
+               It does not mean that the file does not exist at all. */
+    infof(data, "MDTM failed: file does not exist or permission problem,"
+          " trying to continue");
```
bagder, line 2138: It actually continues, it does't *try* to.

bagder (commented):
I think the log message could be improved but otherwise this is :+1:

lib/ftp.c
mhei: Good catch :smile: I've updated the wording, maybe it fits better now. Thanks for your support and patience :+1:

bagder (approved):

bagder:
Thanks!

PR #9395, opened 2022-08-30
appveyor: pass target windows version to MSVS 2022 + cmake + msys build

bagder:
Fixes #9214

Suggested-by: Marcel Raad

PR #9393, opened 2022-08-30
appveyor: remove permafailing CI build

bagder:
The Visual Studio 2022 / cmake / msys build has failed to build for months. Removing it.

Fixes #9214

PR #4662, opened 2019-11-30
Support IPv6 adresses without getaddrinfo

MarcelRaad:
This makes the autotools build behave like the CMake build in this regard: support IPv6 addresses (`ENABLE_IPV6`) by default as long as `AF_INET6` and `sockaddr_in6` are available. `getaddrinfo` is only needed for DNS resolution (`CURLRES_IPV6`).

This also fixes the CMake build with `ENABLE_IPV6` set to the default `ON`, `ENABLE_THREADED_RESOLVER` set to `OFF`, and no `getaddrinfo` available. The theaded resolver code guards the calls to `getaddrinfo` with `HAVE_GETADDRINFO`, but the synchronous resolver code assumes that `CURLRES_IPV6` implies `HAVE_GETADDRINFO`, which it did only implicitly when using the autotools build.

Tested with classic MinGW, which disables `getaddrinfo` by default by targeting Windows 2000, and both the threaded and synchronous resolvers.

PR #9396, opened 2022-08-30
cmake: fix original MinGW builds

### LINKED ISSUE #9214: curl_addrinfo.c error: 'getaddrinfo' undeclared in old-MinGW CI builds
The appveyor CI builds with cmake using Visual Studio 2022 are permafailing since a while back. See [log](https://ci.appveyor.com/project/curlorg/curl/builds/44285128/job/78pv85j280cnnghh)

The compiler errors are:
~~~
C:/projects/curl/lib/curl_addrinfo.c: In function 'Curl_getaddrinfo_ex':
C:/projects/curl/lib/curl_addrinfo.c:192:5: error: implicit declaration of function 'freeaddrinfo'; did you mean 'getaddrinfo'? [-Werror=implicit-function-declaration]
  192 |     freeaddrinfo(aihead);
      |     ^~~~~~~~~~~~
      |     getaddrinfo
C:/projects/curl/lib/curl_addrinfo.c:192:5: error: nested extern declaration of 'freeaddrinfo' [-Werror=nested-externs]
C:/projects/curl/lib/curl_addrinfo.c: In function 'curl_dbg_getaddrinfo':
C:/projects/curl/lib/curl_addrinfo.c:556:14: error: 'getaddrinfo' undeclared (first use in this function); did you mean 'addrinfo'?
  556 |   int res = (getaddrinfo)(hostname, service, hints, result);
      |              ^~~~~~~~~~~
      |              addrinfo
C:/projects/curl/lib/curl_addrinfo.c:556:14: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [lib/CMakeFiles/libcurl.dir/build.make:242: lib/CMakeFiles/libcurl.dir/curl_addrinfo.c.obj] Error 1

vszakats:
1. Re-enable `HAVE_GETADDRINFO` detection on Windows

   Commit d08ee3c83d6bd416aef62ff844c98e47c4682429 (in 2013) added logic
   that automatically assumed `getaddrinfo()` to be present for builds with
   IPv6 enabled. As it turns out, certain toolchains (e.g. original MinGW)
   by default target older Windows versions, and thus do not support
   `getaddrinfo()` out of the box. The issue was masked for a while by CMake
   builds forcing a newer Windows version, but that logic got deleted in
   commit 8ba22ffb2030ed91312fc8634e29516cdf0a9761. Since then, some CI
   builds started failing due to IPv6 enabled, `HAVE_GETADDRINFO` set, but
   `getaddrinfo()` in fact missing.

   It also turns out that IPv6 works without `getaddrinfo()` since commit
   67a08dca27a6a07b36c7f97252e284ca957ff1a5 (from 2019, via #4662). So, to
   resolve all this, we can now revert the initial commit, thus restoring
   `getaddrinfo()` detection and support IPv6 regardless of its outcome.

   Reported-by: Daniel Stenberg

2. Omit `bcrypt` with original MinGW

   Original (aka legacy/old) MinGW versions do not support `bcrypt`
   (introduced with Vista). We already have logic to handle that in
   `lib/rand.c` and autotools builds, where we do not call the unsupported
   API and do not link `bcrypt`, respectively, when using original MinGW.

   This patch ports that logic to CMake, fixing the link error:
   `c:/mingw/bin/../lib/gcc/mingw32/9.2.0/../../../../mingw32/bin/ld.exe: cannot find -lbcrypt`

   Ref: https://ci.appveyor.com/project/curlorg/curl/builds/44624888/job/40vle84cn4vle7s0#L508
   Regression since 76172511e7adcf720f4c77bd91f49278300ec97e

Fixes #9214
Fixes #9393
Fixes #9395
Closes #9396


MarcelRaad (approved):

vszakats:
The `getaddrinfo()` issue is successfully fixed: https://ci.appveyor.com/project/curlorg/curl/build/job/40vle84cn4vle7s0

There is another issue with `bcrypt` missing in legacy MinGW, so the build still fails, but at the linking phase now. That should be a different fix.

[ IMO we should delete these legacy MinGW builds from the CI. ]

Due to the above, I've deleted #9393 and #9395 from the "Fixes" list.

vszakats:
The solution for the `bcrypt` issue is probably to create a legacy-MinGW detector and guard the line `list(APPEND CURL_LIBS "bcrypt")` in `./CMakeLists.txt` with this condition. This is already done in autotools builds. If, we aim to keep support for legacy-MinGW that is.

mback2k:
Since when does curl require bcrypt? I can't remember that change or requirement regarding WinCNG. I only know it from libssh2.

vszakats:
@mback2k It has been merged with this PR #9027. We discussed it 2-3 months ago there.

mback2k:
> [ IMO we should delete these legacy MinGW builds from the CI. ]

@vszakats Thanks. Okay, I would still vote for supporting classic MinGW builds and therefore pre-Vista without bcrypt as long as it is feasible. I think curl is one of those projects trying to stay backwards compatible to legacy OS as long as possible, right?

vszakats:
@mback2k: Given that mingw-w64 surpassed it in all aspects long ago, and mingw-w64 is actually maintained and a free upgrade, while legacy-mingw hasn't been maintained for many years, I think old mingw has not much real use over being an interesting cuiriosity or challange. But, this requires a huge amount of effort each time touching Windows-related things, holding back development and functioning as a time sink. I'm not happy/ready investing time in such things, but if someone else is, it's fine of course.

mback2k:
@vszakats I am okay with dropping the classic MinGW CMake build from AppVeyor while those using autotools stay on Azure.

MarcelRaad:
From my experience with other open source projects, classic MinGW is still in wide use (by people who don't even know about MinGW-w64 and just get told "install MinGW") and not supporting or not testing it just creates avoidable noise on the mailing lists or issue trackers.

But this PR makes sense anyway.

vszakats:
Added CMake logic to fix the bcrypt issue and converted this PR into a legacy-MinGW patch with two disctinct fixes.

mback2k (approved):
LGTM with green CI 👍

vszakats:
I can merge this ASAP (= before reopening the feature window) to make upcoming CI runs greener. Any objection?

bagder (approved):

PR #9400, opened 2022-08-30
FIxing compilation when HAVE_WRITABLE_ARGV not def

jeffque:
Pull request #9399 changed the variable `clearthis` from being a function arg to be a local variable that exists only when the compilation flag `HAVE_WRITABLE_ARGV` is defined.

The not defined case implied that the variable won't exists, but still there were calls to it in other places. This PR puts all mentions of this specific variable under the `HAVE_WRITABLE_ARGV` definition

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -694,7 +694,9 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
         break;
       case 'B': /* OAuth 2.0 bearer token */
         GetStr(&config->oauth_bearer, nextarg);
+#ifdef HAVE_WRITABLE_ARGV
         cleanarg(clearthis);
+#endif
         config->authtype |= CURLAUTH_BEARER;
         break;
       case 'c': /* connect-timeout */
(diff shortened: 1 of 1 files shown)


bagder:
See line 556. If `HAVE_WRITABLE_ARGV` is not defined, the `cleanarg()` is an empty macro:

https://github.com/curl/curl/blob/5b059ba8954c6dcf305ff4def77fe394c7b57401/src/tool_getparam.c#L556

bagder:
on Windows the define is not set and I see many CI jobs on Windows build warning-free, I don't think this change is needed

jay:
> I don't think this change is needed

I don't either. @jeffque how does this fix your compilation, what message were you seeing?

jeffque:
@jay @bagder , in my hubris I misread a snippet in issue #9128 as been the working code in master branch. You see, I called it a function where it clearly is a macro when `HAVE_WRITABLE_ARGV` isn't defined as pointed out in the first comment in this PR.

PR #9401, opened 2022-08-31
http: fix fail code returned if 3xx or 4xx on range request

sergio-nsk:
Data range can be requested and 401 Unauthorized can be responded with a body
unrelated to the requested content, and without Content-Range header, with
unrelated Content-Length value. `curl_easy_perform()` fails with the error
`CURLE_RANGE_ERROR`. It should not fail.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -2911,6 +2911,11 @@ CURLcode Curl_http_firstwrite(struct Curl_easy *data,
      (data->state.httpreq == HTTPREQ_GET) &&
      !k->ignorebody) {
 
+    if(!(200 <= data->info.httpcode && data->info.httpcode < 300)) {
+      /* We wanted to resume a download, but the server doesn't respond with a
+       * successful code, we don't know if the server supports byte ranges */
+      return CURLE_OK;
+    }
     if(k->size == data->state.resume_from) {
       /* The resume point is at the end of file, consider this fine even if it
          doesn't allow resume from here. */
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -251,7 +251,7 @@ test2500 test2501 test2502 \
 test3000 test3001 test3002 test3003 test3004 test3005 test3006 test3007 \
 test3008 test3009 test3010 test3011 test3012 test3013 test3014 test3015 \
 test3016 test3017 test3018 test3019 test3020 test3021 test3022 test3023 \
-test3024 test3025 test3026 test3027 test3028 test3029 test3030 \
+test3024 test3025 test3026 test3027 test3028 test3029 test3030 test3031 \
 \
 test3100 test3101 \
 test3200
--- a/tests/data/test3031
+++ b/tests/data/test3031
@@ -5,14 +5,15 @@ HTTP
 HTTP GET
 Resume
 Largefile
-FAILURE
+Not found
+SUCCESS
 </keywords>
 </info>
 #
 # Server-side
 <reply>
 <data>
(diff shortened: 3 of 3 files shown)


sergio-nsk:
Also fixes test99: `0123456789123\n` data, `0123456789123.` in the dump are 14 bytes, that exceed `Content-Lenght: 13` in the response.

sergio-nsk:
The title in the commit message is changed to be the same as PR's title.

sergio-nsk:
What is a usual time for accepting PRs? 7.86.0 is released 5 days ago. Any chance the change will be merged before 7.87.0?

bagder:
Why do we want curl to return "OK" for this situation? You ask for a range transfer of a specific URL and the server cannot deliver that. Why is that fine when you specify a non-existing URL?

sergio-nsk:
> Why do we want curl to return "OK" for this situation? You ask for a range transfer of a specific URL and the server cannot deliver that. Why is that fine when you specify a non-existing path?

If a server returns a status 4xx, it's not a curl or transfer error. As I wrote above, a server can require authentication, it doesn't mean "the server cannot deliver that", a client app should get and handle 401, not curl error.

bagder:
> 404 does not mean the server is unable to do a ranged transfer.

True, but the transfer you tried to perform does not want to know if the server can do a ranged request. It wants the specified transfer of a specific resource to happen and this clearly was not a successful resume of that.

I fear that changing this now after so many years is going to introduce changed behaviors that will make existing applications misbehave.

sergio-nsk:
libcurl is useless if it returns `CURLE_RANGE_ERROR` in case of other than 100-299 codes.
1. `CURLOPT_USERNAME` and `CURLOPT_PASSWORD` become useless in case of ranged requests. The app must check `CURLINFO_RESPONSE_CODE`, reset `CURLOPT_USERNAME` and `CURLOPT_PASSWORD` and use `CURLOPT_HTTPHEADER` to add a list with `WWW-Authenticate: <type> realm=<realm>`. This will only work for Basic authentication. It will be very hard to implement a workaround for the NTLM authentication.
2. `CURLOPT_FOLLOWLOCATION` becomes not working if it is set. The app must check `CURLINFO_RESPONSE_CODE` and use `curl_easy_header` to retrieve `Location`, then change the URL and perform an updated ranged request.

> I fear that changing this now after so many years is going to introduce changed behaviors that will make existing applications misbehave.

If you merge my change, the developer you so much care of, will get rid of at least one headache, and I am among them. `*_perform()`  call will just work as supposed to work with ranged or regular requests, without custom handlers.

nmoinvaz:
1. Why doesn't cURL check for 200 or [206](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206#:~:text=The%20HTTP%20206%20Partial%20Content,Range%20header%20of%20the%20request) http status code when resuming a content request?

Without this PR, the code would return `CURLE_RANGE_ERROR`:
```c
    /* we wanted to resume a download, although the server doesn't seem to
     * support this and we did this with a GET (if it wasn't a GET we did a
     * POST or PUT resume) */
    failf(data, "HTTP server doesn't seem to support "
          "byte ranges. Cannot resume.");
    return CURLE_RANGE_ERROR;
```
But perhaps returning `CURLE_RANGE_ERROR` is not accurate because technically we do not know if the server supports ranged requests or not. In the case of a 401/407/301 the server will not even attempt to access the content or verify the validity of the content range.

bagder:
I don't think this has been motivated enough.

PR #9403, opened 2022-08-31
misc: spelling fixes

vszakats:
Found using codespell 2.2.1.

Also delete redundant protocol designator from an archive.org URL.


tests/data/test368
```diff
@@ -27,7 +27,7 @@ Connection: close
 http
 </server>
  <name>
-Appened dash if -r range specified without one
+Appended dash if -r range specified without one
```
bagder, line 30: I think this should be "Append"

bagder (approved):
Lovely, just found one little nit...

tests/data/test368
vszakats: Indeed it should, thanks for reviewing!

PR #9412, opened 2022-09-01
test3026: try to reduce runtime within legacy mingw build environments

mback2k:
Right now the test seems to take between 15 and 20 minutes
on our Windows CI environments. _Which is quite extreme for
just 100 threads, but anyway 42 looks like a better number._

### DIFF
--- a/tests/libtest/lib3026.c
+++ b/tests/libtest/lib3026.c
@@ -62,10 +62,8 @@ int test(char *URL)
      libcurl calls LoadLibrary/FreeLibrary it only increases/decreases the
      library's refcount rather than actually loading/unloading the library,
      which would affect the test runtime. */
-#ifdef WIN32
   (void)win32_load_system_library(TEXT("secur32.dll"));
   (void)win32_load_system_library(TEXT("iphlpapi.dll"));
-#endif
 
   for(i = 0; i < tid_count; i++) {
     HANDLE th;
--- a/tests/libtest/lib3026.c
+++ b/tests/libtest/lib3026.c
@@ -57,6 +57,16 @@ int test(char *URL)
     return -1;
   }
 
+  /* On Windows libcurl global init/cleanup calls LoadLibrary/FreeLibrary for
+     secur32.dll and iphlpapi.dll. Here we load them beforehand so that when
+     libcurl calls LoadLibrary/FreeLibrary it only increases/decreases the
+     library's refcount rather than actually loading/unloading the library,
+     which would affect the test runtime. */
+#ifdef WIN32
+  (void)win32_load_system_library(TEXT("secur32.dll"));
+  (void)win32_load_system_library(TEXT("iphlpapi.dll"));
+#endif
+
   for(i = 0; i < tid_count; i++) {
     HANDLE th;
     results[i] = CURL_LAST; /* initialize with invalid value */
--- a/tests/libtest/testutil.c
+++ b/tests/libtest/testutil.c
@@ -131,7 +131,7 @@ double tutil_tvdiff_secs(struct timeval newer, struct timeval older)
 }
 
 #ifdef WIN32
-HMODULE win32_load_system_library(TCHAR *filename)
+HMODULE win32_load_system_library(const TCHAR *filename)
 {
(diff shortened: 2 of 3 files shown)


jay:
> Right now the test seems to take between 15 and 20 minutes
> on our Windows CI environments.

The test should not even take 1 minute, why does it take 15?

bagder:
If a 100 threads take 20 minutes, won't 42 minutes then still take a rather excessive 8.4 minutes?

Maybe the test should rather be changed to have a max time for which it can execute?

mback2k:
As I guessed it does not seem to be a linear issue.

Before / without this PR: https://ci.appveyor.com/project/curlorg/curl/builds/44661247/job/fq5hqnkj4bjr90p5#L4987
```
test 3026...[curl_global_init thread-safety]
-------e--- OK (1510 out of 1510, remaining: 00:00, took 683.379s, duration: 31:14)
```
After / with this PR: https://ci.appveyor.com/project/curlorg/curl/builds/44653968/job/hk9etjq6tuwo0ca7#L4988
```
test 3026...[curl_global_init thread-safety]
-------e--- OK (1510 out of 1510, remaining: 00:00, took 87.706s, duration: 19:43
```

Overall this seems to cut quite some time from the AppVeyor CI runs.

Before vs. After cmake builds (top 4 MSVC, bottom 4 MSYS/MinGW):
image image

Before vs. After autotools builds (3 MSYS/MinGW and 1 CygWin):
image image

Seems like this is primarily affecting the 4 cmake-based MSYS/MinGW in the middle.

mback2k:
To sum up, only builds using the following compilers are affected by the slowness after all:

- C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin
- C:\mingw-w64\x86_64-7.2.0-posix-seh-rt_v5-rev1\mingw64\bin
- C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin
- C:\MinGW\bin

So I guess it is their threading model which is causing the trouble. @MarcelRaad any suggestions?

bagder:
This is still set a draft but wants a review. Isn't that contradictory?

mback2k:
I will need to reproduce this locally and figure out a better way to shorten the runtime for legacy MinGW builds.

mback2k:
Update: I was able to reproduce this locally with classic/legacy MinGW. Next up is identifying the root cause / finding workaround.

mback2k:
Root cause: the slow part is loading of `iphlpapi.dll` (or any other library) as part of `Curl_win32_init` during `curl_global_init`.

bagder:
Should we just close this?

jay:
> Root cause: the slow part is loading of `iphlpapi.dll` (or any other library) as part of `Curl_win32_init` during `curl_global_init`.

I pushed a change to your branch that calls loadlibrary on secur32 and iphlpapi in the main thread. This way each time curl_global_init is called from the worker threads and it calls loadlibrary on those two libraries it should just increase their refcount. This may or may not work depending on if initialization of the library is taking a long time, and not the actual loading.

mback2k:
@jay to resolve the merge conflict which is blocking Azure Pipelines and AppVeyor you may delete my remaining commit.

jay:
job `windows msys v1_mingw32` now shows:

~~~
2022-11-16T07:40:24.7492812Z test 3026...[curl_global_init thread-safety]
2022-11-16T07:40:25.0123541Z -------e--- OK (1526 out of 1530, remaining: 00:02, took 0.268s, duration: 17:07)
~~~

Does that mean this is fixed or should I be checking a different job?

mback2k:
Thanks a lot @jay. Yeah, all of the msys v1 jobs experienced the slowness issue, so if one is fixed, all of them should be.

Will you take it completely from here?

jay:
> Will you take it completely from here?

Yup np. I'm going to wait on the CI for this last fixup to complete then I will land it.

mback2k:
Sounds good, thanks again.

mback2k:
CI looks good to me. I think this is ready for merge.

PR #9413, opened 2022-09-01
GHA: prepare workflow merge by aligning structure again

mback2k:
My long-term plan is still to merge at least all the Linux builds into a single workflow-file for easier management and reduced duplication (like the macOS one, although bigger). Staying aligned with the matrix build structure will make that merge easier.

### DIFF
--- a/.github/workflows/ngtcp2-gnutls.yml
+++ b/.github/workflows/ngtcp2-gnutls.yml
@@ -25,14 +25,14 @@ jobs:
       matrix:
         build:
         - name: gnutls
-          install:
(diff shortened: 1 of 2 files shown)


bagder:
So what's the idea behind the separation of the install packages in the separate places?

mback2k:
The idea is to have common packages which are always required for all Linux CIs on the apt step line, but those that are variant specific in the matrix build config. Of course there can still be a small duplication then there for similar variants, but with YAML anchors there is also a solution for that.

bagder:
Oh right, you mean if you would merge those matrix builds into a unified file?

mback2k:
Yes, exactly.

mback2k:
Any objections against merging this @bagder ?

bagder:
No objections, but I don't see the point in this change unless you also merge the files.

mback2k:
Thanks. That will be my next endeavor.

PR #9414, opened 2022-09-01
configure: fail if '--without-ssl' has been set together with an explicit parameter for an ssl lib

samueloph:
As part of packaging 7.85.0 on Debian, I noticed a build issue due to our packaging using the wrong configure flags, it would set both `--without-ssl` and `--with-nss` or `--with-gnutls`, this would result in the build being marked as a MultiSSL one.

I have addressed the issue by removing the wrong parameter (`--without-ssl`), which I believe was added with the intention of behaving like `--without-openssl` (though this parameter is not available), but it looks like it's a good idea to make the configure script stricter to avoid this from happening altogether, so I figure you might be interested in this change.

Commit description below:
 A side effect of a previous change to configure (576e507c78bdd2ec88da442a5354f9dd661b9a8a)
 exposed a non-critical issue that can happen if configure is called
 with both '--without-ssl' and some parameter setting the use of a
 ssl library (e.g. --with-gnutls). The configure script would end
 up assuming this is a MultiSSL build, due to the way the case
 statement is written.

 I have changed the order of the variables in the string concatenation
 for the case statement and also tweaked the options so that
 --without-ssl never turns the build into a MultiSSL one and
 also clearly stating that there are conflicting parameters
 if the user sets it like described above.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1957,7 +1957,7 @@ if test "x$curl_cv_native_windows" = "xyes" &&
   LIBS="-lbcrypt $LIBS"
 fi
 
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #9408, opened 2022-09-01
urlapi: remove two fields from the URL handle struct

bagder:
'scratch' and 'temppath' are better off as plain local variables.

PR #9429, opened 2022-09-05
curl_ctype: convert to macros-only

bagder:
This no longer provide functions, only macros. Runs slightly faster and produces smaller output.


bagder:
I tested with and without this PR in my performance runs in #9408 and there is a fractional gain, both in speed and in output size.

PR #9447, opened 2022-09-07
headers: reset the requests counter at transfer start

### LINKED ISSUE #9424: curl_easy_nextheader returns null on all subsequent requests after having followed a redirect
### I did this
I'm calling `curl_easy_nextheader` in a loop to count the number of requests in the previous transfer. This works until a transfer involves HTTP redirection. After the redirected request completes, `curl_easy_nextheader` will work as expected. However, on all subsequent requests `curl_easy_nextheader` will return null if the `request` parameter is greater than -1.

### I expected the following
That `curl_easy_nextheader` would not be affected by previous transfers.

### curl/libcurl version
libcurl/7.85.0 OpenSSL/3.0.5 (Schannel) zlib/1.2.12 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.3 libssh2/1.10.0 nghttp2/1.49.0 ngtcp2/0.8.0 nghttp3/0.7.0 libgsasl/1.10.0

### operating system
Windows 10 build 19044.1889

bagder:
If not, reusing an easy handle to do a subsequent transfer would continue the counter from the previous invoke, which then would make use of the header API difficult/impossible as the request counter mismatched.

Add libtest 1947 to verify.

Reported-by: Andrew Lambert
Fixes #9424

### DIFF
--- a/lib/headers.c
+++ b/lib/headers.c
@@ -335,6 +335,7 @@ CURLcode Curl_headers_push(struct Curl_easy *data, const char *header,
 static void headers_init(struct Curl_easy *data)
 {
   Curl_llist_init(&data->state.httphdrs, NULL);
+  data->state.requests = 0;
 }
 
 /*
--- a/lib/headers.c
+++ b/lib/headers.c
@@ -335,7 +335,6 @@ CURLcode Curl_headers_push(struct Curl_easy *data, const char *header,
 static void headers_init(struct Curl_easy *data)
 {
   Curl_llist_init(&data->state.httphdrs, NULL);
-  data->state.requests = 0;
 }
 
 /*
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -1439,6 +1439,7 @@ CURLcode Curl_pretransfer(struct Curl_easy *data)
   if(result)
     return result;
 
+  data->state.requests = 0;
   data->state.followlocation = 0; /* reset the location-follow counter */
   data->state.this_is_a_follow = FALSE; /* reset this */
   data->state.errorbuf = FALSE; /* no error has occurred */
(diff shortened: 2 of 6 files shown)


lib/headers.c
```diff
@@ -335,6 +335,7 @@ CURLcode Curl_headers_push(struct Curl_easy *data, const char *header,
 static void headers_init(struct Curl_easy *data)
 {
   Curl_llist_init(&data->state.httphdrs, NULL);
+  data->state.requests = 0;
```
jay, line 338: is there some reason this can't be put in pretransfer?

bagder: No. I couldn't quite make up my mind where it belongs. But I take it you think `Curl_pretransfer` is a better place? Curl_headers_cleanup() is called from pretransfer and this headers_init() is called from Curl_headers_cleanup().

jay: Yes but I see what you are saying. That's sort of hard to figure out. I suggest don't call init implicitly from cleanup, I would actually do it reverse:

~~~diff
 /*
  * Curl_headers_init(). Init the headers subsystem.
  */
-static void headers_init(struct Curl_easy *data)
+CURLcode Curl_headers_init(struct Curl_easy *data)
 {
+  CURLcode rc = Curl_headers_cleanup(data);
+  if(rc)
+    return rc;
   Curl_llist_init(&data->state.httphdrs, NULL);
+  return CURLE_OK;
 }
~~~
 then in pretransfer

result = Curl_headers_init(
if (result)
error

data->state.requests = 0;
data->state.followlocation = 0; /* reset the location-follow counter */
data->state.this_is_a_follow = FALSE; /* reset this */

etc

Though requests is only used by headers functions right now it could cause problems not to have it in pretransfer if something code is added that evaluates it elsewhere

bagder: Maybe, I just think that's a little much to "just" fix this particular bug. I rather do the smaller change now and move the init to Curl_pretransfer. See my follow-up commit for this.

PR #9448, opened 2022-09-07
http2: make nghttp2 less picky about field whitespace

bagder:
In nghttp2 1.49.0 it returns error on leading and trailing whitespace in header fields according to language in the recently shipped RFC 9113.

nghttp2 1.50.0 introduces an option to switch off this strict check and this change enables this by default which should make curl behave more similar to how it did with nghttp2 1.48.0 and earlier.

We might want to consider making this an option in the future.

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1277,6 +1277,27 @@ void Curl_http2_done(struct Curl_easy *data, bool premature)
   }
 }
 
+static int client_new(struct connectdata *conn,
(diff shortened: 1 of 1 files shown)


jay (approved):

bagder:
I verified that this PR built fine with current nghttp2 1.50.0-DEV version in a local build, and that it didn't cause any obvious issues, but I did not try it against any white-space-field sites.

PR #4750, opened 2019-12-23
Added CURLPROTO_RTSP to redir_protocols

janssen70:
area: RTSP

The RTSP protocol got broken in 7.66.0 by changing the way of initialising redir_protocols in url.c. It would still connect to unauthenticated streams, but not to authenticating ones. This pull request fixes that.

PR #9449, opened 2022-09-07
New libcurl test for RTSP Basic authentication

lminiero:
As requested in #4750, this is an attempt to provide a libcurl test to address the currently broken RTSP authentication support. More precisely, this test will check Basic authentication, but in case the test is fine, a different one can be provided for Digest as well. As it is now, as expected it will break on the protocol check, since the second DESCRIBE should contain an `Authorization: Basic` header, which libcurl is not sending (the main problem that should be solved).

This is the first time I work on a curl/libcurl test, so there are good chances I did something wrong or failed to comply with some assumptions or requirements: in case, I apologize in advance, and I'll gladly accept feedback on how to fix what's wrong. To make an example, I arbitrarily chose `4000` as a number for the test, since I didn't know if there was a rule for the numbering.

With respect to the libcurl check itself, notice that the code currently also includes the ugly workaround that until 7.78.0 still got RTSP authentication working (even though it was apparently broken already), that is:

    /* Ugly workaround needed after curl 7.66 */
    test_setopt(curl, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_RTSP);
    test_setopt(curl, CURLOPT_HTTP09_ALLOWED, 1L);

Not sure if this should be removed in the finalized test, in case this is considered ok to merge eventually. With the snippet above as part of the code, trying this test on curl 7.78.0 succeeds:

```
test 4000...[RTSP Authentication check]
--pd---e-v- OK (1   out of 1  , remaining: 00:00, took 2.757s, duration: 00:02)
TESTDONE: 1 tests were considered during 3 seconds.
TESTDONE: 1 tests out of 1 reported OK: 100%
```

The same test launched on master will fail instead:

```
test 4000...[RTSP Authentication check]

 4000: protocol FAILED:
--- log/check-expected	2022-09-07 16:13:51.391543891 +0200
+++ log/check-generated	2022-09-07 16:13:51.391543891 +0200
@@ -5,5 +5,4 @@
 DESCRIBE rtsp://127.0.0.1:39189/4000 RTSP/1.0[CR][LF]
 CSeq: 2[CR][LF]
 Accept: application/sdp[CR][LF]
-Authorization: Basic dXNlcjpwYXNzd29yZA==[CR][LF]
 [CR][LF]

 - abort tests
TESTDONE: 1 tests were considered during 2 seconds.

TESTFAIL: These test cases failed: 4000 
```

Removing the workaround causes both older and newer versions of libcurl to fail the test sooner than that, with `log/stderr4000` containing

    * Protocol "rtsp" not supported or disabled in libcurl

which seems in line with the discussion made in #4750.

Please let me know if this test is acceptable, or if there's any changes that should be made. Thanks!


bagder:
This looks like a fine start!

> I arbitrarily chose 4000 as a number for the test

That's fine. The test numbers don't mean anything, they just make each test has its own unique id.

> With respect to the libcurl check itself, notice that the code currently also includes the ugly workaround that until 7.78.0 still got RTSP authentication working. Not sure if this should be removed in the finalized test

I don't see why we would want "an ugly workaround" in the test? Especially since the test doesn't run OK with it either in current master.

In my mind, the point with this test case was to use it as a tool to verify that whatever fix is necessary to get RTSP + auth to work actually works. Without workarounds.

lminiero:
> I don't see why we would want "an ugly workaround" in the test? Especially since the test doesn't run OK with it either in current master.
> 
> In my mind, the point with this test case was to use it as a tool to verify that whatever fix is necessary to get RTSP + auth to work actually works. Without workarounds.

Makes sense! I only added it to highlight what caused it to break from previous versions (where the workaround helped), and how without the workaround both still failed because of a deeper issue. I'll get rid of it shortly.

bagder:
There's been a month with nobody trying to actually fix these problems that this test highlights.

I think we basically have two options on how to proceed in spite of no one fixing RTSP:

1. Close this PR and just leave it for later until someone tries to get things working again
2. Merge it in a disabled status, to enable it at a later point once someone gets going with RTSP

Thoughts?

lminiero:
I don't have a strong opinion on either, if not that merging it (even with a disabled state) might help people be aware of pending tests that need fixing. Of course there are other places where the existence of this problem is reported, so it's not really needed I guess. Sorry if I haven't worked on a fix myself in the meanwhile, but unfortunately I'm not familiar with the curl internals, at least not at the depth fixing this problem would probably need. Should that change I'll definitely take a look.

janssen70:
Hi,

Using @lminiero's test I have this now:

```
test 4000...[RTSP Authentication check]
--pd---e-v- OK (1   out of 1  , remaining: 00:00, took 1.968s, duration: 00:01)
TESTDONE: 1 tests were considered during 2 seconds.
TESTDONE: 1 tests out of 1 reported OK: 100%

```

 but I have not fixed the CURLOPT_REDIR_PROTOCOLS problem. I did:

1. Add back to the test: test_setopt(curl, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_RTSP);
2. Patched rtsp_do() to set the first_host, first_remote_port and first_remote_protocol fields

I added the diff for 2) below. I don't see how I can propose a change to a PR. What do you think of it?

@bagder, would it be acceptable to have rtsp authentication working again, with enabled test, but accept the  CURLOPT_REDIR_PROTOCOLS workaround? To me it seems of great value
Btw, I see no need for setting CURLOPT_HTTP09_ALLOWED

Below the diff for rtsp.c.  @lminiero, maybe you can add it to the PR?

```
diff --git a/lib/rtsp.c b/lib/rtsp.c
index 6d3bf97e6..30505cd7c 100644
--- a/lib/rtsp.c
+++ b/lib/rtsp.c
@@ -267,6 +267,18 @@ static CURLcode rtsp_do(struct Curl_easy *data, bool *done)
   rtsp->CSeq_sent = data->state.rtsp_next_client_CSeq;
   rtsp->CSeq_recv = 0;
 
+  if(!data->state.first_host) {
+    /* Set first_remote_port like Curl_http_host(). Needed for correct
+     * evaluation by Curl_auth_allowed_to_host() in vauth/vauth.c
+     */
+    data->state.first_host = strdup(conn->host.name);
+    if(!data->state.first_host)
+      return CURLE_OUT_OF_MEMORY;
+
+    data->state.first_remote_port = conn->remote_port;
+    data->state.first_remote_protocol = conn->handler->protocol;
+  }
+
   /* Setup the 'p_request' pointer to the proper p_request string
    * Since all RTSP requests are included here, there is no need to
    * support custom requests like HTTP.
```

bagder:
> @bagder, would it be acceptable to have rtsp authentication working again, with enabled test, but accept the CURLOPT_REDIR_PROTOCOLS workaround? To me it seems of great value

Great value for whom? If we accept that hack to get RTSP auth working, then applications using other schemes suddenly *unwillingly* also start accepting redirects to RTSP URLs. That seems totally wrong.

Not to mention that if we accept this hack, I fear that every RTSP user will be happy and satisfied and no one will work on doing the *proper* fix.

So no, I don't think it is in the project's long-term interest to accept this work-around.

janssen70:
> So no, I don't think it is in the project's long-term interest to accept this work-around.

I understand your reasoning. I'd appreciate if the test in this PR gets merged then in disabled status. It helps with trying to fix the redirect-thing

> Great value for whom?

I mentioned that :)

bagder:
I just created #9870 which is my take on fixing this. It includes this test case, renumbered to 3100 and @janssen70's clue about the first host fields.

PR #9450, opened 2022-09-07
content_encoding: handle alignment explicitly

monnerat:
The variable-sized encoding-specific storage of a struct contenc_writer currently relies on void * alignment that may be insufficient with regards to the specific storage fields, although having not caused any problems yet.

In addition, gcc 11.3 issues a warning on access to fields of partially allocated structures that can occur when the specific storage size is 0:

```
  content_encoding.c: In function ‘Curl_build_unencoding_stack’:
  content_encoding.c:980:21: warning: array subscript ‘struct contenc_writer[0]’ is partly outside array bounds of ‘unsigned char[16]’ [-Warray-bounds]
    980 |     writer->handler = handler;
        |     ~~~~~~~~~~~~~~~~^~~~~~~~~
  In file included from content_encoding.c:49:
  memdebug.h:115:29: note: referencing an object of size 16 allocated by ‘curl_dbg_calloc’
    115 | #define calloc(nbelem,size) curl_dbg_calloc(nbelem, size, __LINE__, __FILE__)
        |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  content_encoding.c:977:60: note: in expansion of macro ‘calloc’
    977 |   struct contenc_writer *writer = (struct contenc_writer *)calloc(1, sz);
```

To resolve both problems, this commit:
- adds an align.h header file to define alignment macros,
- removes the "params" field of structure contenc_writer.
- computes the allocation size and effective offset of specific storage from the struct contenc_writer address in an alignment-aware way. This makes sure the structure is fully allocated and specific storage properly aligned (warning disappears).

New align.h file is generic and may be re-used in another module if ever needed.

### DIFF
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -219,6 +219,7 @@ LIB_CFILES =         \
   wildcard.c
 
 LIB_HFILES =         \
+  align.h            \
   altsvc.h           \
   amigaos.h          \
   arpa_telnet.h      \
--- a/lib/align.h
+++ b/lib/align.h
@@ -1,0 +1,68 @@
+#ifndef HEADER_CURL_ALIGN_H
+#define HEADER_CURL_ALIGN_H
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+
+/* Universal type alignment macros. */
+
+#include "curl_setup.h"
+#include <stddef.h>
+
+/* Alignment for type.*/
+#define ALIGN(type)     offsetof(struct { char c; type t; }, t)
+
+/* Union of all scalar types for universal alignment computation. */
+union all_scalar_types {
+  char          c;
+  short         s;
+  int           i;
+  long          l;
+#ifdef HAVE_LONGLONG
+  long long     ll;
+#endif
+  float         f;
+  double        d;
+#ifdef HAVE_LONGDOUBLE
+  long double   ld;
+#endif
+  void *        p;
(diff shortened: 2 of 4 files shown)


bagder:
How do you get those warnings to show?

bagder:
This warning is about out of bounds access for an array, it says nothing about alignment. I don't 

~~~
content_encoding.c:980:21: warning: array subscript ‘struct contenc_writer[0]’ is partly outside array bounds of ‘unsigned char[16]’ [-Warray-bounds]

bagder:
(very sorry, I pressed the wrong button my mistake)

bagder:
I'm asking, because I would us to look for a simpler solution. This was not ideal before but this align-macro thing makes it worse IMHO.

monnerat:
> How do you get those warnings to show?

With gcc 11.3.1 on Fedora 35.
```
export CFLAGS="-g -O3"
./configure --prefix=/usr						\
	--with-gssapi							\
	--with-libssh							\
	--with-ssl							\
	--enable-debug
```
The warnings do not occur with optimizaion level < 2.

(hidden comment):
> This warning is about out of bounds access for an array, it says nothing about alignment.

Right, but they have the same cause and the warnings took my attention on it.
That said, there are no problems with the current partial allocation of structure because when it occurs, unallocated storage is not used.
I would say also the warning message is not very clear...

> I'm asking, because I would us to look for a simpler solution.

That's what I was thinking of! IMO we may resolve this case by emulating C++-like subclasses for specific parameters, letting the compiler manage alignment.
Please hold-on with the current PR (I'll withdraw it when case resolved). I'll submit another one for the _subclass_ approach.

> This was not ideal before but this align-macro thing makes it worse IMHO.

Less readable, I agree and really prefer the current code, but not worse in terms of alignment and compiler's analysis!

bagder:
> Less readable, I agree and really prefer the current code, but not worse in terms of alignment and compiler's analysis!

Yes, sorry. I meant the readability.

bagder:
A simpler way to hush up the compiler, at the cost of a few extra bytes of heap allocation:
~~~
--- a/lib/content_encoding.c
+++ b/lib/content_encoding.c
@@ -971,11 +971,11 @@ static const struct content_encoding error_encoding = {
 static struct contenc_writer *
 new_unencoding_writer(struct Curl_easy *data,
                       const struct content_encoding *handler,
                       struct contenc_writer *downstream)
 {
-  size_t sz = offsetof(struct contenc_writer, params) + handler->paramsize;
+  size_t sz = sizeof(struct contenc_writer) + handler->paramsize;
   struct contenc_writer *writer = (struct contenc_writer *)calloc(1, sz);
 
   if(writer) {
     writer->handler = handler;
     writer->downstream = downstream;

monnerat:
> A simpler way to hush up the compiler, at the cost of a few extra bytes of heap allocation:

Sure, this was my first evidence !

But it does not resolve alignment potential problem and waste some unused memory.

monnerat:
> I don't believe any existing CPU architectures actually have an alignment problem with this.

I don't believe either but this is purely fortuitous.
Suppose we have 32-bit pointers with alignment 4, 64-bit double with alignment 8.
If a parameter structure contains a double (maybe in a substructure), it requires alignment 8, while only 4 is guaranteed by `void * params`.
Fortunately `params` is at offset 8... only a chance !

Anyway, I would suggest to obsolete this PR and replace by #9455

monnerat:
PR is obsolete: closing.

PR #9363, opened 2022-08-25
schannel: when importing PFX, disable key persistence

DHowett:
By default, the PFXImportCertStore API persists the key in the user's
key store (as though the certificate was being imported for permanent,
ongoing use.)

The documentation specifies that keys that are not to be persisted
should be imported with the flag `PKCS12_NO_PERSIST_KEY`.

NOTE: This flag is only supported on versions of Windows newer than XP
and Server 2003. On lower versions, we'll gracefully fall back.

Closes #9300

PR #9460, opened 2022-09-09
schannel: when importing PFX, disable key persistence

### LINKED ISSUE #9300: RSA key file leak in C:\ProgramData\Microsoft\Crypto\RSA\S-1-5-18
I use this code to perform tls mutual authentication. It works. However, anytime curl_easy_perform returns, there is a file generated in  C:\ProgramData\Microsoft\Crypto\RSA\S-1-5-18. And curl_easy_cleanup doesn't delete the file.  Is this a bug or there is a attribute that I can set to auto delete the file ?
~~~c
    CURLcode res;
    int result = 0;
    char errBuf[CURL_ERROR_SIZE]{0};
    char *version = curl_version();

    struct curl_slist *http_headers = NULL;
    ret                             = "";
    http_headers = curl_slist_append(http_headers, "Accept: application/json");
    http_headers = curl_slist_append(http_headers, "Content-Type: application/json");
    http_headers = curl_slist_append(http_headers, "charsets: utf-8");
    curl_easy_setopt(curl, CURLOPT_HTTPHEADER, http_headers);

    curl_easy_setopt(curl, CURLOPT_URL, apiPath);
    curl_easy_setopt(curl, CURLOPT_POSTFIELDS, str);
    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, WriteData);
    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &ret);
    curl_easy_setopt(curl, CURLOPT_POST, 1);
    curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1);
    curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, errBuf);

    curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, 60);
    curl_easy_setopt(curl, CURLOPT_TIMEOUT, 60);
    curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, true);
    curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, true);
    curl_easy_setopt(curl, CURLOPT_SSLCERTTYPE, "P12");
    curl_easy_setopt(curl, CURLOPT_SSLCERT, personalCertificateFile.c_str());
    curl_easy_setopt(curl, CURLOPT_KEYPASSWD, passwd.c_str());
    curl_slist_free_all(http_headers);
    curl_easy_cleanup(curl);
    return result;
~~~
### curl/libcurl version
7.85.0-DEV

### operating system
Windows 10

DHowett:
By default, the PFXImportCertStore API persists the key in the user's
key store (as though the certificate was being imported for permanent,
ongoing use.)

The documentation specifies that keys that are not to be persisted
should be imported with the flag `PKCS12_NO_PERSIST_KEY`.
NOTE: this flag is only supported on versions of Windows newer than XP
and Server 2003.

When you import a P12 into memory alone, you must keep the handle to the
certificate store alive until you're done with the credential; minor
changes were made to the schannel credential context to support closing
the client certificate store handle at the end of an SSL session.

Supersedes #9363
Closes #9300

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -186,6 +186,10 @@
 #define ALG_CLASS_DHASH ALG_CLASS_HASH
 #endif
 
+#ifndef PKCS12_NO_PERSIST_KEY
+#define PKCS12_NO_PERSIST_KEY 0x00008000
+#endif
+
 static Curl_recv schannel_recv;
 static Curl_send schannel_send;
 
@@ -676,7 +680,13 @@ schannel_acquire_credential_handle(struct Curl_easy *data,
         else
           pszPassword[0] = 0;
 
-        cert_store = PFXImportCertStore(&datablob, pszPassword, 0);
+        if(curlx_verify_windows_version(6, 0, 0, PLATFORM_WINNT,
+                                        VERSION_GREATER_THAN_EQUAL))
+          cert_store = PFXImportCertStore(&datablob, pszPassword,
+                                          PKCS12_NO_PERSIST_KEY);
+        else
+          cert_store = PFXImportCertStore(&datablob, pszPassword, 0);
+
         free(pszPassword);
       }
       if(!blob)
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -576,6 +576,8 @@ schannel_acquire_credential_handle(struct Curl_easy *data,
   }
 
 #ifdef HAS_CLIENT_CERT_PATH
+  HCERTSTORE client_cert_store = NULL;
+
   /* client certificate */
   if(data->set.ssl.primary.clientcert || data->set.ssl.primary.cert_blob) {
     DWORD cert_store_name = 0;
@@ -758,7 +760,7 @@ schannel_acquire_credential_handle(struct Curl_easy *data,
         return CURLE_SSL_CERTPROBLEM;
       }
     }
-    CertCloseStore(cert_store, 0);
+    client_cert_store = cert_store;
   }
 #else
   if(data->set.ssl.primary.clientcert || data->set.ssl.primary.cert_blob) {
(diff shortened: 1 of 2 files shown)


DHowett:
This was tested for use in `libcurl` by the original reporter, who verified it as of https://github.com/curl/curl/issues/9300#issuecomment-1241613007.

DHowett:
The tests pass locally at the top of my branch (04c11a4), and it seems like these specific build runs (mingw64, WinIDN) are failing in a few other pull requests. I marked the pull request as draft until I could validate the tests, so I'm going to mark it as ready for review now.

lib/vtls/schannel.c
```diff
@@ -2447,6 +2467,10 @@ static void schannel_session_free(void *ptr)
     if(cred->refcount == 0) {
       s_pSecFn->FreeCredentialsHandle(&cred->cred_handle);
       curlx_unicodefree(cred->sni_hostname);
+      if(cred->client_cert_store) {
+        CertCloseStore(cred->client_cert_store, 0);
+        cred->client_cert_store = NULL;
+      }
```
jay, line 2473:
```suggestion
#ifdef HAS_CLIENT_CERT_PATH
      if(cred->client_cert_store) {
        CertCloseStore(cred->client_cert_store, 0);
        cred->client_cert_store = NULL;
      }
#endif
```
according to 1592ea9 HAS_CLIENT_CERT_PATH checks for CryptStringToBinary because some builds like mingw which target versions of windows earlier than xp it might not be available, and I assume neither will CertCloseStore.  

also, client_cert_store declaration could be guarded the same way, however that would require moving the HAS_CLIENT_CERT_PATH logic into the EXPOSE_SCHANNEL_INTERNAL_STRUCTS section. i can do that or you can do it, you can see HAS_MANUAL_VERIFY_API for an example. basically we have to adjust for original mingw (if we are still supporting that even)

jay (changes requested):

lib/vtls/schannel.c
DHowett: Ah! That was my original intent, but I realized that the structure was declared before `HAS_CLIENT_CERT_PATH` would be available. I can restructure it a bit to follow what you did for `HAS_MANUAL_VERIFY_API`.

In the meantime, I'll also try to grab a build environment that will let me verify that it works properly.

DHowett:
Pushed an update. I've taken your updated commit message into the PR body for tracking purposes. Thanks for the review!

jay:
I've pushed a change to your branch to always include wincrypt, since now there is a test to see if CryptStringToBinary is defined, which is defined by wincrypt if it is defined. Assuming the CI has no problems I will review again but otherwise this looks ready.

DHowett:
Thanks! That makes sense.

Surprising number of OpenSSL and WolfSSL failures in CI. Are those already known/baselined?

jay:
The 3 CI failures I see are unrelated. The BoringSSL failure might be an actual problem, though unrelated, and the other two look like random failures that occasionally happen due to resource constraints.

DHowett:
Is there anything I need to do to get this over the finish line for the coming release? :smile:

Thanks!

bagder:
@jay any remaining nits or can we merge?

jay:
Thanks!

PR #9463, opened 2022-09-09
schannel: ban server ALPN selection during recv renegotiation

jay:
By the time schannel_recv is renegotiating the connection, libcurl has already decided on a protocol and it is too late for the server to select a protocol via ALPN.

Ref: https://github.com/curl/curl/issues/9451

Closes #xxxx

---

I'm not entirely sure if this is correct, I don't really understand the way ALPN is expected to work on renegotiation. For example, if the connection is immediately renegotiated before any application data is received, then maybe it would work to change protocols based on server ALPN selection?


jay:
I've modified this slightly to allow the server to select an ALPN protocol on renegotiation as long as it is the same as what was originally selected, since it's not documented how schannel handles this.

PR #9469, opened 2022-09-10
CI: debug if we randomly are killing ourselves on Win32

mback2k:
Not to be merged.

Previous example failure: https://github.com/curl/curl/runs/8285644244

### DIFF
--- a/tests/ftp.pm
+++ b/tests/ftp.pm
@@ -133,6 +133,8 @@ sub pidterm {
                 my $filter = "PID eq $pid";
                 my $result = `tasklist -fi \"$filter\" 2>nul`;
                 if(index($result, "$pid") != -1) {
+                    print("RUN: Win32 pid $pid signalled to die\n");
+                    print("RUN: Win32 pid $pid = $result\n");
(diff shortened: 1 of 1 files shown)


mback2k:
Reproduced failure: https://dev.azure.com/mback2k/curl/_build/results?buildId=1075&view=logs&j=e7c72124-b275-5783-0311-a91da4bbcc8b&t=73dc882c-50a4-5b1c-38aa-d437ef5d9e59&l=2170

mback2k:
It seems like some processes are killed by `runtests.pl` after test 677 although that and the following test do not ask for that. Weird!

@bagder any idea why this might be happening?

bagder:
Is this the `killpid()` function getting called? If so, from where?

mback2k:
> Is this the `killpid()` function getting called? If so, from where?

That is also what I was wondering. I don't see a reason for it to be called. Is there some kind of stacktrace in Perl to figure this out?

mback2k:
Closing, will come back to this at a later point. Right now I don't think just the Windows CI is a main source of flakiness anymore. Actually it seems much more stable recently.

PR #9472, opened 2022-09-11
setopt: use the handler table for protocol name to number conversions

monnerat:
This also returns error `CURLE_UNSUPPORTED_PROTOCOL` rather than `CURLE_BAD_FUNCTION_ARGUMENT` when a listed protocol name is not found.

A new `schemelen` parameter is added to `Curl_builtin_scheme()` to support this extended use.

Note that disabled protocols are not recognized anymore.

Test programs 1597 and 1911 adapted accordingly.

### DIFF
--- a/lib/setopt.c
+++ b/lib/setopt.c
@@ -148,85 +148,36 @@ static CURLcode setstropt_userpwd(char *option, char **userp, char **passwdp)
 #define C_SSLVERSION_VALUE(x) (x & 0xffff)
 #define C_SSLVERSION_MAX_VALUE(x) (x & 0xffff0000)
 
-static CURLcode protocol2num(char *str, curl_prot_t *val)
+static CURLcode protocol2num(const char *str, curl_prot_t *val)
 {
-  bool found_comma = FALSE;
-  static struct scheme {
-    const char *name;
-    curl_prot_t bit;
-  } const protos[] = {
-    { "dict", CURLPROTO_DICT },
-    { "file", CURLPROTO_FILE },
-    { "ftp", CURLPROTO_FTP },
-    { "ftps", CURLPROTO_FTPS },
-    { "gopher", CURLPROTO_GOPHER },
-    { "gophers", CURLPROTO_GOPHERS },
-    { "http", CURLPROTO_HTTP },
-    { "https", CURLPROTO_HTTPS },
-    { "imap", CURLPROTO_IMAP },
-    { "imaps", CURLPROTO_IMAPS },
-    { "ldap", CURLPROTO_LDAP },
-    { "ldaps", CURLPROTO_LDAPS },
-    { "mqtt", CURLPROTO_MQTT },
-    { "pop3", CURLPROTO_POP3 },
-    { "pop3s", CURLPROTO_POP3S },
-    { "rtmp", CURLPROTO_RTMP },
-    { "rtmpe", CURLPROTO_RTMPE },
-    { "rtmps", CURLPROTO_RTMPS },
-    { "rtmpt", CURLPROTO_RTMPT },
-    { "rtmpte", CURLPROTO_RTMPTE },
-    { "rtmpts", CURLPROTO_RTMPTS },
-    { "rtsp", CURLPROTO_RTSP },
-    { "scp", CURLPROTO_SCP },
-    { "sftp", CURLPROTO_SFTP },
-    { "smb", CURLPROTO_SMB },
-    { "smbs", CURLPROTO_SMBS },
-    { "smtp", CURLPROTO_SMTP },
-    { "smtps", CURLPROTO_SMTPS },
-    { "telnet", CURLPROTO_TELNET },
-    { "tftp", CURLPROTO_TFTP },
-#ifdef USE_WEBSOCKETS
-    { "ws", CURLPROTO_WS },
-    { "wss", CURLPROTO_WSS },
-#endif
-    { NULL, 0 }
-  };
-
   if(!str)
     return CURLE_BAD_FUNCTION_ARGUMENT;
-  else if(curl_strequal(str, "all")) {
-    *val = (curl_prot_t)~0;
+
+  if(curl_strequal(str, "all")) {
+    *val = ~(curl_prot_t) 0;
     return CURLE_OK;
   }
 
   *val = 0;
 
   do {
+    const char *token = str;
     size_t tlen;
-    struct scheme const *pp;
-    char *token;
-    token = strchr(str, ',');
-    found_comma = token ? TRUE : FALSE;
-    if(!token)
-      token = strchr(str, '\0');
-    tlen = token - str;
+
+    str = strchr(str, ',');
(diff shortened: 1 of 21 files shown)


monnerat:
@bagder: it seems not recognizing disabled protocols is a big problem for the tool and the tests.
What do we do ?

- Accept this restriction and adapt tests & tools to it (your websocket fix to protocol2num goes in this direction).
- Ignore any unrecognized word: real errors not detected anymore.
- Forget about this change

lib/url.c
```diff
@@ -1853,15 +1853,18 @@ static struct connectdata *allocate_conn(struct Curl_easy *data)
...
   const struct Curl_handler * const *pp;
   const struct Curl_handler *p;
   /* Scan protocol handler table and match against 'scheme'. The handler may
      be changed later when the protocol specific setup function is called. */
+  if(schemelen == CURL_ZERO_TERMINATED)
+    schemelen = strlen(scheme);
   for(pp = protocols; (p = *pp) != NULL; pp++)
-    if(strcasecompare(p->scheme, scheme))
-      /* Protocol found in table. Check if allowed */
+    if(strncasecompare(p->scheme, scheme, schemelen) && !p->scheme[schemelen])
```
bagder, line 1866: won't this make looking for `http` also match a handler having `https` as scheme?

monnerat: No, because it also tests for `!p->scheme[schemelen]`

bagder: oh of course!

bagder:
Hm, I think maybe the first alternative is the best:

**Accept this restriction and adapt tests & cli tool to it**

I believe I intended for `CURLOPT_PROTOCOLS_STR` to work that way, as I read the description in the man page that way. The downside with that approach is for applications that would think "we can support these protocols XXXX" and then support them for when the libcurl actually supports it and not support them if they don't...

But if we don't work that way, we also can't detect typos...

monnerat:
> Hm, I think maybe the first alternative is the best:

OK, I will change that way. It will probably require several checks and take some time. Turning this PR into a draft now.

monnerat:
@bagder: with this PR, the cli tool relies on the protocols listed by `curl_version_info`(). It would therefore require to list also `RTMPE`, `RTMPTE`, `RTMPS` and `RTMPTS` if available (not yet done in this PR). Any objection ?

bagder:
> require to list also RTMPE, RTMPTE, RTMPS and RTMPTS if available

I'm not super happy about that, as they will then end up in the `curl -V` output etc but at the same time I figure that might be the better solution overall. I mean, it makes sense to treat all schemes the same way consistently.

We could *possibly* add something special for the curl tool that just doesn't show all of them.

monnerat:
> I'm not super happy about that, as they will then end up in the curl -V output 

I understand your dilemma. In all cases, the tool should adapt. I'll look at it. Thanks for your reply.

monnerat:
I've added "warts" code snippets to the cli tool to support the library whether it returns RTMP?* protocols or not. Those protocols are NOT returned yet.

tests/data/test1401
```diff
@@ -91,7 +91,7 @@ int main(int argc, char *argv[])
   curl_easy_setopt(hnd, CURLOPT_FTP_SKIP_PASV_IP, 1L);
 %endif
   curl_easy_setopt(hnd, CURLOPT_TCP_KEEPALIVE, 1L);
-  curl_easy_setopt(hnd, CURLOPT_PROTOCOLS_STR, "http,ftp,file");
+  curl_easy_setopt(hnd, CURLOPT_PROTOCOLS_STR, "http");
```
bagder, line 94: This is the only test case verifying `CURLOPT_PROTOCOLS_STR` with `---libcurl`, shouldn't it use more than one protocol ?

We can just make the test require support for those protocols to be present.

bagder (commented):
I think this looks very good. Just some minor comments/questions.

lib/setopt.c
```diff
@@ -148,85 +148,36 @@ static CURLcode setstropt_userpwd(char *option, char **userp, char **passwdp)
...
-    struct scheme const *pp;
-    char *token;
-    token = strchr(str, ',');
-    found_comma = token ? TRUE : FALSE;
-    if(!token)
-      token = strchr(str, '\0');
-    tlen = token - str;
+
+    str = strchr(str, ',');
+    tlen = str? (size_t) (str - token): strlen(token);
```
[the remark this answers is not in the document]
monnerat: No. If I drop it, I get:
```
setopt.c:168:18: warning: operand of ‘?:’ changes signedness from ‘long int’ to ‘size_t’ {aka ‘long unsigned int’} due to unsignedness of other operand [-Wsign-compare]
  168 |     tlen = str?  (str - token): strlen(token);
      |                  ^~~~~~~~~~~~~
```
I presume pointer subtraction leads to some type equivalent to `ssize_t` or `ptrdiff_t`.

lib/setopt.c
```diff
@@ -148,85 +148,36 @@ static CURLcode setstropt_userpwd(char *option, char **userp, char **passwdp)
...
+      *val |= h->protocol;
     }
-    if(found_comma)
-      str = token + 1;
-  } while(found_comma);
+  } while(str++);
+
   if(!*val)
-    /* no matching protocol */
+    /* no protocol listed */
```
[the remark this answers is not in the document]
monnerat: No. unmatching protocols are rejected above with `CURLE_UNSUPPORTED_PROTOCOL`.

lib/setopt.c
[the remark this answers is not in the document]
bagder: Curios, since there was a `tlen = token - str;` in the old version without a typecast!

tests/data/test1401
monnerat: Yes, but how to check for 3 protocols in the same test? I didn't find a way to make the program "auto-adaptive" and still match the `<verify>` data.

bagder: In the `<client>` section you list all features that are required:
~~~
<features>
http
ftp
file
</features>

lib/setopt.c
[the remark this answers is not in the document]
bagder: :+1:

lib/setopt.c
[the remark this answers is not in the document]
monnerat: Maybe because the context was assignment and not the `?:` operator.
Do I rephrase or keep cast?

tests/data/test1401
monnerat: Thanks for the hint. I didn't know protocol names were features !

lib/setopt.c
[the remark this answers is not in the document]
bagder: Keep it

tests/data/test1401
monnerat: Fixed now!

bagder (approved):

bagder:
Thanks!

monnerat:
Thanks for your help.

PR #9474, opened 2022-09-11
Correct missing uppercases in Markdown files

Benjamin-Loison:
To detect these typos I used:

```
grep -rn '\. [a-z]' . | \
uniq | \
grep -v '\. lib' | \
grep -v '[0-9]\. [a-z]' | \
grep -v '\.\. [a-z]' | \
grep -v '\. curl' | \
grep -v 'e.g. [a-z]' | \
grep -v 'eg. [a-z]' | \
grep -v '\etc. [a-z]' | \
grep -v 'i.e\. [a-z]' | \
grep --color=always '\. [a-z]' | \
grep '\.md'
```

### DIFF
--- a/docs/BUGS.md
+++ b/docs/BUGS.md
@@ -245,7 +245,7 @@
  tracker.
 
  The `TODO` document is full of ideas and suggestions of what we can add or
- fix one day. you are always encouraged and free to grab one of those items and
+ fix one day. You are always encouraged and free to grab one of those items and
  take up a discussion with the curl development team on how that could be
  implemented or provided in the project so that you can work on ticking it odd
  that document.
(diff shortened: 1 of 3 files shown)


bagder (approved):

bagder:
It might be fun to convert such a check into a CI job/tool. I notice that the command line above still points out:

    docs/CHECKSRC.md:143:is sometimes a need to ignore specific warnings. checksrc allows a few

... and maybe `checksrc` should be allowed to have a lowercase c just like curl is allowed in the line?

bagder:
Thanks!

dfandrich:
How about turning this into a test to prevent future errors?

PR #9482, opened 2022-09-12
tool_operate: Check USE_MANUAL before call hugehelp

xiaoxiang781216:
since PARAM_MANUAL_REQUESTED isn't returned when USE_MANUAL isn't defined

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -2679,8 +2679,10 @@ CURLcode operate(struct GlobalConfig *global, int argc, argv_item_t argv[])
       if(res == PARAM_HELP_REQUESTED)
         tool_help(global->help_category);
       /* Check if we were asked for the manual */
+#ifdef USE_MANUAL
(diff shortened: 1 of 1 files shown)


bagder:
I suggest we do #9485 instead.

xiaoxiang781216:
> Ah right, it doesn't return `PARAM_MANUAL_REQUESTED` then, but is the call hurting? I think an added #ifdef there is a cost I rather avoid.

The problem is that we run curl on RTOS which use a special build system and then hard to generate hugehelp from the script dynamically. It's also useless to generate dummy function, that why I prefer to remove hugehelp from call site.

bagder:
I think that should rather be fixed by making `hugehelp()` a blank macro when it doesn't do anything. I extended #9485 with that.

xiaoxiang781216:
> I think that should rather be fixed by making `hugehelp()` a blank macro when it doesn't do anything. I extended #9485 with that.

Yes, the change looks good, let me close this PR.

PR #9483, opened 2022-09-12
dynbuf: map curlx_dyn_xxx to Curl_dyn_ when BUILDING_LIBCURL is defined

xiaoxiang781216:
so the tool and library could be bundled in the same binary

### DIFF
--- a/lib/dynbuf.h
+++ b/lib/dynbuf.h
@@ -41,8 +41,21 @@
 #define Curl_dyn_reset(a) curlx_dyn_reset(a)
 #define Curl_dyn_tail(a,b) curlx_dyn_tail(a,b)
 #define Curl_dyn_setlen(a,b) curlx_dyn_setlen(a,b)
-#define curlx_dynbuf dynbuf /* for the struct name */
+#else
+#define curlx_dyn_init(a,b) Curl_dyn_init(a,b)
+#define curlx_dyn_add(a,b) Curl_dyn_add(a,b)
+#define curlx_dyn_addn(a,b,c) Curl_dyn_addn(a,b,c)
+#define curlx_dyn_addf Curl_dyn_addf
+#define curlx_dyn_vaddf Curl_dyn_vaddf
+#define curlx_dyn_free(a) Curl_dyn_free(a)
+#define curlx_dyn_ptr(a) Curl_dyn_ptr(a)
+#define curlx_dyn_uptr(a) Curl_dyn_uptr(a)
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
> I don't understand why you need this change. Can you please elaborate?

we run curl on an RTOS(https://github.com/apache/incubator-nuttx) which don't use dynamically normally, which mean both library and tool compile and link into one image. Without this change, tool will complain that curlx_dyn can't found.

bagder:
When you build the command line tool, `BUILDING_LIBCURL` is not supposed to be defined. The command line tool is not libcurl. It sounds like you set that define and that is the reason for your problem. The existing top `#ifndef` in dynbuf.h is already present there exactly to cater for when the command line code uses the dynbuf.c code. This is code that is reused by the command line tool, it is not part of the API.

xiaoxiang781216:
> When you build the command line tool, `BUILDING_LIBCURL` is not supposed to be defined. The command line tool is not libcurl. It sounds like you set that define and that is the reason for your problem.

Yes, it's hard to get the different macro definition, since both library and tool is built together in the same phase for bare metal or RTOS environment.

BTW, since BUILDING_LIBCURL is defined in curl_config.h, could you explain how tools get the different definition from library if both reference the same file.

bagder:
> since BUILDING_LIBCURL is defined in curl_config.h

That seems to be a bug in the cmake build. See #9498

bagder:
When we move out `BUILDING_LIBCURL` from `lib/curl_config.h`, is there then any reason left for this PR?

xiaoxiang781216:
> When we move out `BUILDING_LIBCURL` from `lib/curl_config.h`, is there then any reason left for this PR?

#9498 make the thing better, but it can't resolve my(many bare metal or RTOS ) issue since it's hard to give the same macro different value for different source file if they use IDE or handwriting Makefile.

bagder:
Sure, but curl is a tool and libcurl is a library. They are not meant to get built in the same go with the same options. `BUILDING_LIBCURL` should *only* be defined when the sources are built for libcurl.

xiaoxiang781216:
I understand the requirement, but it's hard to apply the different options in the bare metal build environment.:(

bagder:
It's really not that hard. You just have to make sure you build the files in lib/ with one set of flags and the ones in src/ with another. You're asking us to add a weirdo work-around to save you from a little bit of work and I don't think that's the right way to go about this.

PR #9491, opened 2022-09-12
scripts: use `grep -E` instead of `egrep`

a1346054:
latest GNU grep emits a deprecation message when `egrep` is invoked. All platforms that can be expected to run the three changed scripts already have `grep -E` as a replacement.

### DIFF
--- a/MacOSX-Framework
+++ b/MacOSX-Framework
@@ -83,12 +83,12 @@ if test $PPC64_NEEDED -gt 0
 then
   SDK64=10.5
   ARCHES64='-arch x86_64 -arch ppc64'
-  SDK64=`ls  $SDK_PATH|grep 10.5|head -1`
+  SDK64=`ls  $SDK_PATH | grep "10\.5" | head -1`
 else
  ARCHES64='-arch x86_64'
  #We "know" that 10.4 and earlier do not support 64bit
- OLD_SDK64=`ls  $SDK_PATH|egrep -v "10.[0-4]"|head -1`
- NEW_SDK64=`ls -r $SDK_PATH|egrep -v "10.[0-4][^0-9]" | head -1`
+ OLD_SDK64=`ls  $SDK_PATH | grep -v "10\.[0-4]" | head -1`
+ NEW_SDK64=`ls -r $SDK_PATH | grep -v "10\.[0-4][^0-9]" | head -1`
  if test $USE_OLD -gt 0
   then
    SDK64=$OLD_SDK64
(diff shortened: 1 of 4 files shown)


emanuele6:
There is no reason to use `-E` with these `grep` invocations since you are using `grep` and not `egrep`.

MacOSX-Framework
```diff
@@ -87,8 +87,8 @@ then
 else
  ARCHES64='-arch x86_64'
  #We "know" that 10.4 and earlier do not support 64bit
- OLD_SDK64=`ls  $SDK_PATH|egrep -v "10.[0-4]"|head -1`
- NEW_SDK64=`ls -r $SDK_PATH|egrep -v "10.[0-4][^0-9]" | head -1`
+ OLD_SDK64=`ls  $SDK_PATH|grep -Ev "10.[0-4]"|head -1`
+ NEW_SDK64=`ls -r $SDK_PATH|grep -Ev "10.[0-4][^0-9]" | head -1`
```
emanuele6, line 91: you don't need to use `-E` here

a1346054: I didn't want to change the script more than I had to.

vszakats: Strictly speaking, `.` should be escaped here as well. This, and dropping `-E` are fine within this PR IMO.

We have more `egrep`s in this file:
```
.github/workflows/hacktoberfest-accepted.yml:            egrep -i "^Close[sd]? " | sort | uniq | tee log
.github/workflows/hacktoberfest-accepted.yml:          egrep -o "#([0-9]+)" log | cut -d# -f2 | sort | uniq | xargs -t -n1 -I{} \
.github/workflows/hacktoberfest-accepted.yml:            egrep -o '^([0-9]+):[0-9]{4}-(09-30T|10-|11-01T)' | cut -d: -f1 | sort | uniq | xargs -t -n1 -I {} \
.github/workflows/hacktoberfest-accepted.yml:          egrep -o "github.com/(.+)/(.+)/pull/([0-9]+)" log | sort | uniq | xargs -t -n1 -I{} \
.github/workflows/hacktoberfest-accepted.yml:            egrep -o '^([0-9]+):[0-9]{4}-(09-30T|10-|11-01T)' | cut -d: -f1 | sort | uniq | xargs -t -n1 -I {} \
```

a1346054:
Fixed up the commit by removing `-E` where not needed for the grep pattern, and additionally escaping the `.` in the pattern.

Also replaced `egrep` in hacktoberfest-accepted.yml with `grep -E`

MacOSX-Framework
a1346054: Addressed your comments in the fixed up commit.

vszakats (approved):

bagder:
Thanks!

PR #9492, opened 2022-09-13
wolfSSL: Fix session management bug.

haydenroche5:
Prior to this commit, non-persistent pointers were being used to store sessions. When a WOLFSSL object was then freed, that freed the session it owned, and thus invalidated the pointer held in curl's cache. This commit makes it so we get a persistent (deep copied) session pointer that we then add to the cache. Accordingly, wolfssl_session_free, which was previously a no-op, now needs to actually call SSL_SESSION_free.

This bug was discovered by a wolfSSL customer.


bagder (approved):

bagder:
Thanks!

PR #9497, opened 2022-09-13
CI: retire zuul

pheiduck:
Signed-off-by: Philip Heiduck


bagder:
I rather we keep them alive until they have been reproduced on other services.

bagder:
Let's remove entries from the zuul scripts when we move them somewhere else. Then the last moved stuff could remove all traces of Zuul.

PR #9501, opened 2022-09-13
CI/GHA: merge bearssl and hyper into initial linux workflow

mback2k:
Begin work on merging all Linux workflows into one file.

[skip azp]


mback2k:
I will have to rethink this as GitHub unfortunately does not seem to support YAML anchors at the moment:
> The workflow is not valid. .github/workflows/linux.yml: Anchors are not currently supported. Remove the anchor 'bearssl_install_script'
image

My goal is still to reduce YAML duplication and also get rid of this redundancy and convolution in the Checks tab of PRs:
image

pheiduck:
I would appreciate it if we remove / hide workflows that have not been running for a while. What only the owner of the repo can do.

mback2k:
> I would appreciate it if we remove / hide workflows that have not been running for a while. What only the owner of the repo can do.

Good idea, I will try to do that.

mback2k:
bearssl and hyper builds looking good merged together:
image image 

@bagder @pheiduck what do you think about this approach (including install_steps)? Should I continue with more workflows?

bagder:
It looks good! shouldn't this PR also remove `.github/workflows/linux-hyper.yml`  ?

mback2k:
It does, but GitHub renders it as a rename instead.

PR #9511, opened 2022-09-15
Fix a problem with doing PUT then POST using the same handle

bagder:


bagder:
Suddenly, the libressl job on macOS failed test 1564 *wakeup before poll with no easy handles* here, but I deem that totally unrelated. But something to keep an eye on.

PR #9512, opened 2022-09-15
Use NMake batch-rules for compilation

chemodax:
Currently NMake based build invokes cl.exe for every file. This pull request enable NMake batch-mode rules for cl.exe. In this case cl.exe invoked for multiple .c files.

This is significantly improves compilation time. For example in my environment: 40 s → 20 s.


jay:
> This is significantly improves compilation time. For example in my environment: 40 s → 20 s.

I see the same. Thanks

PR #9520, opened 2022-09-16
gha: add a CI job that runs proselint on all markdown files

bagder:


.github/workflows/proselint.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: install prereqs
+      run: sudo apt-get install python3-proselint
+
+    # config file help: https://github.com/amperser/proselint/
+    - name: store proselint config
+      run: |
+        echo '{' > $HOME/.proselintrc
```
cmeister2, line 28: This seems a weird way of doing this, why not store it under the CI folder then copy it into place?

cmeister2 (commented):
Think the proselintrc file handling needs a rethink.

.github/workflows/proselint.yml
```diff
@@ -0,0 +1,37 @@
+name: proselint
+
+on:
+  # Trigger the workflow on push or pull requests, but only for the
+  # master branch
```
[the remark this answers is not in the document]
bagder: nope.

.github/workflows/proselint.yml
bagder: sounds like a better approach indeed, I'll try that!

bagder:
ok, seems to work now, rebased and squashed into something readable

.github/workflows/proselint.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: check prose
+      run: a=`git ls-files '*.md' | grep -v docs/CHECKSRC.md` && proselint $a README
+
+    - name: create second proselint config
+      run: grep '^## C2:' .github/workflows/proselint.yml | sed 's/^## C2://' > $HOME/.proselintrc
+
+    - name: check special prose
+      run: a=docs/CHECKSRC.md && proselint $a
+
+## C: {
```
cmeister2, line 40: This just feels like a kludge. I'd put this in something like .github/proselint/special.json, then have a 
```
cp .github/proselint/special.json ~/.config/proselint/config.json
```
(this is the new path for proselint as per https://github.com/amperser/proselint#checks )

.github/workflows/proselint.yml
```diff
@@ -0,0 +1,56 @@
...
+## C:    "checks": {
+## C:      "typography.diacritical_marks": false,
+## C:      "typography.symbols": false,
+## C:      "annotations.misc": false
+## C:    }
+## C: }
+
+# This is for CHECKSRC and files with agressive exclamation mark needs
```
cmeister2, line 48: nit: sp: aggressive

cmeister2 (commented):
I have a minor quibble with how the proselint json files are mastered, but this is only a recommendation. (I feel like introducing grep and sed into the process is a mistake)

.github/workflows/proselint.yml
bagder: I opted to not create a stand-alone json file ingit because they don't have comments and then I can't easily provide copyright info and would need to have yet another exception for that in `.reuse/dep5` and I rather not.

cmeister2: Ok - could something like this work instead?

```
max@LAPTOP-BL9ORR19:~$ cat special.yml
# I have comments because I am a YAML file
{
  "checks": {
    "typography.diacritical_marks": false,
    "typography.symbols": false,
    "annotations.misc": false,
  }
}
max@LAPTOP-BL9ORR19:~$ grep -v "#" special.yml
{
  "checks": {
    "typography.diacritical_marks": false,
    "typography.symbols": false,
    "annotations.misc": false,
  }
}
max@LAPTOP-BL9ORR19:~$ yq . special.yml
{
  "checks": {
    "typography.diacritical_marks": false,
    "typography.symbols": false,
    "annotations.misc": false
  }
}
```

YAML is a superset of JSON - you can write the necessary copyright comments with lines starting #, then either grep them out or use `yq` (would need installing along with `jq`) to parse and output the JSON content.

bagder: But if we have to run grep anyway, does it really matter?

bagder: Since we have to copy the file to a different place anyway, we can just call i ".json" and still do the grep and allow `#`-comments...

bagder: That doesn't really solve that we need two different though, so we'd need to do two different files I guess. Or do something crazier.

cmeister2: Ideally the benefits here comes from having editor tooling which understands how to edit JSON and YAML. If more options need adding in future then having a JSON/YAML aware editor means that people will make fewer mistakes; it's also more obvious to people where changes need to be added and how to add them. 

IMO we should be generally aiming (in all aspects) for "least surprise" to developers - if I were a new dev coming into this project and wanted to add a setting to proselint for some reason, I'd expect that to be mastered in its own file as opposed to mixed in with the CI definition. 

There are reasons for wanting to put significant chunks of code in CI definitions (e.g. they're being shared between projects) but I don't think this is one of them.

Anyway, I'm not going to argue this further! It's only my recommendation.

bagder: If I make two files called `proselint.json` and `proselint2.json` for this, it still requires some good insight for a developer to know how to update them and to what, and probably they need to read the `proselint.yml` file to learn that, and then they have the whole picture anyway...

bagder: I'll update to that and see how that feels in comparison.

mback2k: You could also just include the config in each file creation step with a multi-line (using | in YAML) shell command to use echo or tee.

bagder:
Thanks @mback2k, I've squashed this all into two comments now.

PR #9521, opened 2022-09-16
windows: add .rc support to autotools builds

vszakats:
After this update autotools builds will compile and link `.rc` resources
to Windows executables. Bringing this feature on par with CMake
and Makefile.m32 builds. And also making it unnecessary to improvise
these steps manually, while monkey patching build files, e.g. [0].

You can customize the resource compiler via the `RC` envvar, and its
options via `RCFLAGS`.

This harmless warning may appear throughout the build, even though the
autotools manual documents [1] `RC` as a valid tag:
`libtool:   error: ignoring unknown tag RC`

[0] https://github.com/curl/curl-for-win/blob/535f19060d4b708f72e75dd849409ce50baa1b84/curl-autotools.sh#L376-L382
[1] https://www.gnu.org/software/libtool/manual/html_node/Tags.html

Closes #9521

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,8 @@ CURL_CHECK_OPTION_ECH
 
 XC_CHECK_PATH_SEPARATOR
 
+LT_LANG([Windows Resource])
+
 #
 # save the configure arguments
 #
@@ -568,6 +570,7 @@ CURL_DARWIN_CFLAGS
 CURL_DARWIN_SYSTEMCONFIGURATION
 CURL_SUPPORTS_BUILTIN_AVAILABLE
 
+AM_CONDITIONAL([OS_WINDOWS], [test "$curl_cv_native_windows" = "yes"])
 
 dnl ************************************************************
 dnl switch off particular protocols
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -75,6 +75,12 @@ AM_CPPFLAGS += -DBUILDING_LIBCURL
 AM_LDFLAGS =
 AM_CFLAGS =
 
+# Makefile.inc provides the CSOURCES and HHEADERS defines
+include Makefile.inc
+
+libcurl_la_SOURCES = $(CSOURCES) $(HHEADERS)
+libcurlu_la_SOURCES = $(CSOURCES) $(HHEADERS)
+
 libcurl_la_CPPFLAGS_EXTRA =
 libcurl_la_LDFLAGS_EXTRA =
 libcurl_la_CFLAGS_EXTRA =
@@ -102,6 +108,11 @@ endif
 
 if USE_CPPFLAG_CURL_STATICLIB
 libcurl_la_CPPFLAGS_EXTRA += -DCURL_STATICLIB
+else
+if OS_WINDOWS
+libcurl_la_SOURCES += $(LIB_RCFILES)
+$(LIB_RCFILES): $(top_srcdir)/include/curl/curlver.h
+endif
 endif
 
 if DOING_CURL_SYMBOL_HIDING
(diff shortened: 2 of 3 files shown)


vszakats:
Pending issues:
- handling non-windres environments (= MSVC). If this is a valid use-case with curl autotools builds.
- what happens if `RC` isn't set?

vszakats:
CI seems fine now, with two unrelated (flaky?) single-test failures @ AppVeyor, and `curl/check` (@ Zuul) failing, which does fail with mostly all commits these days.

vszakats:
FWIW, tried removing `--tag=RC` to avoid the nonsense libtool warning `error: ignoring unknown tag RC`, which resulted in this:
```
libtool: compile: unable to infer tagged configuration
libtool:   error: specify a tag with '--tag'
```

¯\_(ツ)_/¯

monnerat:
A side effect of this commit is new warning messages from `autoreconf -fi`:
```
configure.ac:125: warning: AC_PROG_CC was called before XC_CHECK_PROG_CC
m4/xc-cc-check.m4:86: XC_CHECK_PROG_CC is expanded from...
configure.ac:125: the top level
configure.ac:352: warning: LT_LANG was called before LT_INIT
/usr/share/aclocal/libtool.m4:61: LT_INIT is expanded from...
m4/xc-lt-iface.m4:442: XC_LIBTOOL is expanded from...
configure.ac:352: the top level
configure.ac:125: warning: AC_PROG_CC was called before XC_CHECK_PROG_CC
m4/xc-cc-check.m4:86: XC_CHECK_PROG_CC is expanded from...
configure.ac:125: the top level
configure.ac:352: warning: LT_LANG was called before LT_INIT
/usr/share/aclocal/libtool.m4:61: LT_INIT is expanded from...
m4/xc-lt-iface.m4:442: XC_LIBTOOL is expanded from...
configure.ac:352: the top level
configure.ac:125: warning: AC_PROG_CC was called before XC_CHECK_PROG_CC
m4/xc-cc-check.m4:86: XC_CHECK_PROG_CC is expanded from...
configure.ac:125: the top level
configure.ac:352: warning: LT_LANG was called before LT_INIT
aclocal.m4:83: LT_INIT is expanded from...
m4/xc-lt-iface.m4:442: XC_LIBTOOL is expanded from...
configure.ac:352: the top level
```
I tried locally to move the `LT_LANG([Windows Resource])` after the `XC_LIBTOOL` and it does suppress the warnings. However I am not sure it doesn't break your commit. Can you investigate? Thanks.

vszakats:
@monnerat: Great find! This also fixes the odd warning noted here earlier, and in my tests the .rc files continue to compile and link fine.

Do you want to commit the fix?

monnerat:
> Do you want to commit the fix?

No, I let you care of your baby !
> Strangely, I'm not getting those autoreconf -fi warnings on my system.

libtool or autotools version dependent ? OS dependent ? who knows !

vszakats:
Sure, I've opened a [PR](https://github.com/curl/curl/pull/9582).

> libtool or autotools version dependent ? OS dependent ? who knows !

The perfect motto for autotools :) (to be fair: for build systems in general)

monnerat:
Thanks for fix.

lazka:
The latest release fails to build, could this be related? (missing make dependency?)

```
  /bin/sh ../libtool --tag=RC --mode=compile windres -I../../curl-7.86.0/include -DCURL_EMBED_MANIFEST  -i ../../curl-7.86.0/src/curl.rc -o curl.o
    CC       tool_hugehelp.o
    CC       ../lib/strtoofft.o
    CC       ../lib/timediff.o
  libtool: compile:  windres -I../../curl-7.86.0/include -DCURL_EMBED_MANIFEST -i ../../curl-7.86.0/src/curl.rc  -o .libs/curl.o
    CC       ../lib/nonblock.o
    CC       ../lib/warnless.o
    CC       ../lib/curl_multibyte.o
    CC       ../lib/version_win32.o
    CC       ../lib/dynbuf.o
    CCLD     curl.exe
  clang: error: no such file or directory: 'curl.o'
  make[2]: *** [Makefile:1011: curl.exe] Error 1
```

lazka:
I've reduced it to `./configure --enable-shared --disable-static --with-schannel` on current master, if that helps

I can also create a new issue if wanted.

bagder:
> I can also create a new issue if wanted.

Please do, I think we track it better that way

vszakats:
@lazka: Thanks! A separate Issue would be helpful indeed. `--enable-shared --disable-static` is the exact combination where in curl-for-win I explicitly _skip_ building `curl.exe` (due to fatal `ltwrapper` issues post-link, and we also happen to not need this binary). This explains why this escaped my local tests. We also miss it from CI tests apparently. I could reproduce it now.

Do you have any suggestion for a fix?

lazka:
I've filed an issue at https://github.com/curl/curl/issues/9803

PR #9523, opened 2022-09-17
GHA: spellcheck markdown files and man pages

bagder:
This spellchecker checks markdown files. For this reason this job converts all man pages in the repository to markdown with pandoc before the check runs. I am unfortunately not aware of any tool that can run spellcheck on the "raw" nroff file.

This is somewhat of an experiment. I am not sure this will make total sense as maybe the amount of unknown words will be too large and not manageable. But I do know that I write a lot of typos and by checking our documentation with tools we might improve it.


bagder:
This one puzzles me:
~~~
pandoc docs/libcurl/opts/CURLOPT_PROXY_TLS_CIPHERS.3.3 -o docs/libcurl/opts/CURLOPT_PROXY_TLS_CIPHERS.3.md 
pandoc: docs/libcurl/opts/CURLOPT_PROXY_TLS_CIPHERS.3.3: openBinaryFile: does not exist (No such file or directory)
~~~

It finds a file name there that doesn't seem to exist... ?

bagder:
If the approach works, using [pyLanguagetool](https://github.com/Findus23/pyLanguagetool) is an option to try.

bagder:
Next step is to auto-generate words that need to be accepted, like `CURLOPT_[word]`.

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,680 @@
...
+    - name: setup the custom wordlist
+      run: grep '^#W' .github/workflows/spellcheck.yml | sed 's/#W://' >  wordlist.txt
+
+    - name: accept all option names
+      run: find docs/libcurl/opts/ -name "*.3" | cut -d_ -f2 | sed 's/\.3//' | sort -u >> wordlist.txt
+
+    - name: Check Spelling
+      uses: UnicornGlobal/spellcheck-github-actions@master
+
+#W:ACTIVESOCKET
```
cmeister2, line 42: Similar to #9520, I think these sorts of files should be mastered entirely separately from the CI definitions.

bagder: And I have the same reasoning here: I don't think I can do comments in that file so copyright info can't be provided.

bagder:
I need to ponder if I can come up with a way to not trigger too many errors for the synopsis and example parts of the man pages.

bagder:
I also think I need to skip `BINDINGS.md` completely...

.github/workflows/spellcheck.yml
dfandrich: Based on some of the words in this list, spaces are allowed. So, why not make the first word be

    #W:Copyright 2022 Daniel Stenberg, all rights reserved, etc.

.github/workflows/spellcheck.yml
```diff
@@ -143,12 +146,19 @@ jobs:
...
 #W:CURLINFO
+#W:CURLKHTYPE
+#W:CURLM
 #W:CURLMOPT
 #W:CURLOPT
+#W:CURLPROXY
 #W:curlrc
+#W:CURLSHE
+#W:CURLSHOPT
+#W:CURLSHOPT
```
dfandrich, line 161: This is a dupe.

.github/workflows/spellcheck.yml
```diff
@@ -178,6 +188,7 @@ jobs:
 #W:Dudka
 #W:Dymond
 #W:EBCDIC
+#W:ECDHE
 #W:EFnet
 #W:eg
```
dfandrich, line 193: This isn't a word: it should probably be "e.g."

.github/workflows/spellcheck.yml
```diff
@@ -222,11 +233,13 @@ jobs:
 #W:fuzzer
```
dfandrich, line 233: (Github isn't letting place this comment on the correct line). "fiddlings" isn't actually a real word.

.github/workflows/spellcheck.yml
```diff
@@ -222,11 +233,13 @@ jobs:
...
+#W:GCM
 #W:gdb
 #W:Gergely
 #W:getaddrinfo
 #W:gethostname
 #W:Getinfo
+#W:getinfo
 #W:GETing
 #W:getpwuid
 #W:ggcov
```
dfandrich, line 245: Similarly, "gitter" (below) isn't a word.

.github/workflows/spellcheck.yml
```diff
@@ -277,6 +290,8 @@ jobs:
 #W:HTTPS
```
dfandrich, line 290: "HackerOne" is spelled three different ways (above).

.github/workflows/spellcheck.yml
```diff
@@ -277,6 +290,8 @@ jobs:
...
 #W:https
 #W:Huawei
+#W:IANA
+#W:ICONV
 #W:IDN
 #W:IDNA
 #W:IETF
```
dfandrich, line 297: "ipod", if it's the music device, should be "iPod". I actually can't find some of these words in the repo so I can't see the context.

.github/workflows/spellcheck.yml
```diff
@@ -416,6 +431,7 @@ jobs:
 #W:Ok
```
dfandrich, line 431: "Kubernetes" (above) should be capitalized.

.github/workflows/spellcheck.yml
```diff
@@ -416,6 +431,7 @@ jobs:
...
 #W:OpenBSD
```
dfandrich, line 432: "NMetware" above doesn't look right to me.

.github/workflows/spellcheck.yml
```diff
@@ -416,6 +431,7 @@ jobs:
...
 #W:openldap
+#W:OpenLDAP
 #W:Opensocket
```
dfandrich, line 435: The capital letter here doesn't seem right.

.github/workflows/spellcheck.yml
```diff
@@ -613,6 +633,8 @@ jobs:
 #W:urlget
```
dfandrich, line 633: "Stenberg"? That's an odd one. Is it Danish? ☺

.github/workflows/spellcheck.yml
dfandrich: I would also like to use this word list to spell check Everything Curl. I've created my own list there but it would be nice to reuse this one. To make that easier, it would be nice if this word list were 1) in a separate file, and 2) a list of bare words, without the #W: prefix. That could be added in the workflow.

dfandrich (commented):
I couldn't find a bunch of these words in the repo so I couldn't figure out the context.

.github/workflows/spellcheck.yml
bagder: Which one? You mean `OpenLDAP` is not right?

bagder:
> I would also like to use this word list to spell check Everything Curl

Funny, because this list comes from the CI job that I added recently to spellcheck everything curl :grimacing:

bagder:
> I need to ponder if I can come up with a way to not trigger too many errors for the synopsis and example parts of the man pages.

It struck me that I will solve this by making a script that simply strips out those parts of the man pages before they are converted to markdown.

.github/workflows/spellcheck.yml
dfandrich: No; I complained about GitHub not letting me put my comments on the right lines. In most cases, my comment refers to the line immediately above (Opensocket in this case), but if Github didn't let me do that, then I mention the actual word in my comment. Maybe there's a better way to do this kind of review.

.github/workflows/cleanspell.pl
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/perl
```
williamdes, line 1: I am not sure why this is a perl script, would not bash be best?

bagder: Maybe. I have a feeling this filter is going to grow and get more complicated very soon so a shell script will get quirky very soon.

bagder:
It just took 97 commits and now the spell job ran green! :balloon:

bagder:
We are at 861 words in the custom dictionary now. I am sure we can trim that and remove words going forward but I think this is a decent enough first shot.

.github/workflows/spellcheck.yml
bagder: nothing uses Opensocket, removed

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2022 Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+
+name: spell
+on:
+  push:
+    branches:
+    - master
+    paths:
```
mback2k, line 10: I guess if you want to filter on paths for this workflow, you would need to include "**.3", "**.33" and "**.1", right?

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,44 @@
...
+    - '**.md'
+    - '!docs/BINDINGS.md'
+  pull_request:
+    branches:
+    - master
```
mback2k, line 15: I guess the same filter should then also apply to pull_requests and both filters also need to include the relevant files under `.github/workflows`

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,44 @@
...
+# SPDX-License-Identifier: curl
+
+name: spell
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - '**.md'
+    - '!docs/BINDINGS.md'
```
mback2k, line 12: I think paths are excluded/ignored by specifying them in another key called `paths-ignore` instead of `paths`.

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  check:
+
```
mback2k, line 19: maybe this empty line can go away, please? ;)

.github/workflows/spellcheck.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright (C) 2022 Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+#
+# Docs: https://github.com/UnicornGlobal/spellcheck-github-actions
+matrix:
```
mback2k, line 6: The workflows folder is reserved for actual workflow definitions. This YAML files is interpreted by GitHub and leads to invalid workflow runs, see: https://github.com/curl/curl/actions/workflows/spellcheck.yaml

They recommend to use `.github/scripts/` for workflow-related scripts, see https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#adding-scripts-to-your-workflow

.github/workflows/spellcheck.yml
bagder: `**.3`  and `**.1` at least, since the `.33` files are just the trick this job makes as an intermediate step

bagder:
@mback2k I believe I've addressed your initial comments now. Anything else you think needs polish?

bagder:
I'll merge this tomorrow unless someone objects

.github/scripts/spellcheck.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright (C) 2022 Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+#
+# Docs: https://github.com/UnicornGlobal/spellcheck-github-actions
+matrix:
+- name: Markdown
+  expect_match: false
+  apsell:
```
cmeister2, line 9: `apsell`? Not `aspell`?

cmeister2 (commented):
One minor thing in the GHA commit.

mback2k:
I will look later today.

.github/scripts/spellcheck.yaml
bagder: I've seen that as well. I decided not to touch it since it works. The `apsell` comes from the example config over here: https://github.com/marketplace/actions/check-spelling-js-vue-html-markdown-text

... where you can also see that both `aspell` and `apsell` are used. I suspect apsell is a typo but that English is the default anyway.

mback2k (approved):
LGTM

Issue #9524, opened 2022-09-17
lib1597.c nonstandard extension used : 'exp' : cannot be initialized using address of automatic variable


bagder:
commit 9d5132904 landed a change to lib1597.c that isn't C89 compliant and some [builds on appveyor](https://ci.appveyor.com/project/curlorg/curl/builds/44805354/job/xnlccnpi5d7b5mty#L983) caught it.

The same change caused another warning but that is handled in #9522 

/cc @monnerat

monnerat:
Sorry for that!
I think putting this table and CURLcode vars in global scope will fix the problem.
I'll soon submit a PR for it.

PR #9525, opened 2022-09-17
lib1597: make it C89-compliant again

monnerat:
Automatic variable addresses cannot be used in an initialisation aggregate.

Follow-up to 9d51329

Reported-by: Daniel Stenberg
Fixes: #9524

Let's see what AppVeyor says!

### DIFF
--- a/tests/libtest/lib1597.c
+++ b/tests/libtest/lib1597.c
@@ -38,18 +38,18 @@ int test(char *URL)
   CURL *curl = NULL;
   int res = 0;
   CURLcode result = CURLE_OK;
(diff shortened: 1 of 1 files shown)


bagder:
I believe this shows that appveyor says: :heavy_check_mark: - even if it failed on one of those flaky ones again.

bagder:
Thanks!

monnerat:
> --- even if it failed on one of those flaky ones again.

This is the problem with AppVeyor: as it almost always fails, real problems are buried in false positives :-(

Thanks for commit.

PR #9528, opened 2022-09-17
docs: fix whitespace

jbampton:


bagder:
As I'm fixing spelling problems in those two files in #9523, I'll merge this little fix into that huge take.

bagder:
Thanks for the report, these flaws were fixed in master now via other PRs!

PR #9531, opened 2022-09-17
CI/linkcheck: only run if a Markdown file is changed

mback2k:
This saves CI resources and therefore a little energy.


mback2k:
Could also be used for #9520

cmeister2 (approved):
LGTM

mback2k:
Of course this also needs to run if the workflow itself was changed.

cmeister2 (approved):

PR #9532, opened 2022-09-17
CI: move distcheck job from Azure Pipelines to GitHub Actions

mback2k:
This is in preparation of adding trigger excludes to Azure Pipelines.

### DIFF
--- a/.github/workflows/distcheck.yml
+++ b/.github/workflows/distcheck.yml
@@ -1,0 +1,101 @@
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+
+name: dist
+
+on:
+  push:
+    branches:
+    - master
+    - '*/ci'
+  pull_request:
+    branches:
+    - master
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  maketgz-and-verify-in-tree:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v4
+
+    - run: sudo apt-get purge -y curl libcurl4 libcurl4-doc
+      name: 'remove preinstalled curl libcurl4{-doc}'
+
+    - run: autoreconf -fi
+      name: 'autoreconf'
+
+    - run: ./configure --without-ssl
+      name: 'configure'
+
+    - run: make V=1 && make V=1 clean
+      name: 'make and clean'
+
+    - run: ./maketgz 99.98.97
+      name: 'maketgz'
+
+    - uses: actions/upload-artifact@v3
+      with:
+        name: 'release-tgz'
+        path: 'curl-99.98.97.tar.gz'
+
+    - run: |
+        echo "::stop-commands::$(uuidgen)"
+        tar xvf curl-99.98.97.tar.gz
+        pushd curl-99.98.97
+        ./configure --prefix=$HOME/temp --without-ssl
+        make
+        make TFLAGS=1 test
+        make install
+        popd
+        # basic check of the installed files
+        bash scripts/installcheck.sh $HOME/temp
+        rm -rf curl-99.98.97
+      name: 'verify in-tree configure build including install'
+
+  verify-out-of-tree-docs:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs: maketgz-and-verify-in-tree
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: 'release-tgz'
+
+    - run: |
+        echo "::stop-commands::$(uuidgen)"
+        tar xvf curl-99.98.97.tar.gz
+        touch curl-99.98.97/docs/{cmdline-opts,libcurl}/Makefile.inc
+        mkdir build
+        pushd build
+        ../curl-99.98.97/configure --without-ssl
+        make
+        make TFLAGS='-p 1 1139' test
+        popd
+        rm -rf build
+        rm -rf curl-99.98.97
+      name: 'verify out-of-tree configure build including docs'
+
+  verify-out-of-tree-cmake:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs: maketgz-and-verify-in-tree
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: 'release-tgz'
+
+    - run: |
+        echo "::stop-commands::$(uuidgen)"
+        tar xvf curl-99.98.97.tar.gz
+        pushd curl-99.98.97
+        mkdir build
+        pushd build
+        cmake ..
+        make
+      name: 'verify out-of-tree cmake build'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -116,59 +116,6 @@ stages:
         AZURE_ACCESS_TOKEN: "$(System.AccessToken)"
         TFLAGS: "-ac /usr/bin/curl -r $(tests)"
 
-- stage: distcheck
-  dependsOn: []
-  jobs:
-  - job: ubuntu
-    timeoutInMinutes: 30
-    pool:
-      vmImage: 'ubuntu-latest'
-    steps:
-    - script: autoreconf -fi && ./configure --without-ssl
-      displayName: 'configure'
-
-    - script: make && ./maketgz 99.98.97
-      displayName: 'make tarball'
-
-    - script: |
-        tar xf curl-99.98.97.tar.gz
-        cd curl-99.98.97
-        ./configure --prefix=$HOME/temp --without-ssl
-        make
-        make TFLAGS=1 test
-        make install
-        # basic check of the installed files
-        cd ..
-        bash scripts/installcheck.sh $HOME/temp
-        rm -rf curl-99.98.97
-
-      displayName: 'verify in-tree configure build'
-
-    - script: |
-        # verify out-of-tree build
-        tar xf curl-99.98.97.tar.gz
-        touch curl-99.98.97/docs/{cmdline-opts,libcurl}/Makefile.inc
-        mkdir build
-        cd build
-        ../curl-99.98.97/configure --without-ssl
-        make
-        make TFLAGS='-p 1 1139' test
-        # verify cmake build
-        cd ..
-        rm -rf curl-99.98.97
-
-      displayName: 'verify out-of-tree configure build'
-
-    - script: |
-        tar xf curl-99.98.97.tar.gz
-        cd curl-99.98.97
-        mkdir build
-        cd build
-        cmake ..
-        make
-
-      displayName: 'verify out-of-tree cmake build'
-
 - stage: scanbuild
   dependsOn: []
   jobs:


.github/workflows/distcheck.yml
```diff
@@ -0,0 +1,68 @@
...
+    branches:
+    - master
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v3
+
```
pheiduck, line 22: we need to prepare some dependencies for **maketgz**:
```
run: |
sudo apt-get update
apt-get install libgsasl-dev libssl-dev
name: 'install prereqs and impacket'
```

mback2k: I don't think these are required. @bagder I could see you install any of those in Azure CI either, right?

bagder: The more dependencies configure finds, the more "complete" build it makes but I don't think that matters too much when we just verify the dist.

bagder:
So why does this job break like this? :confused:

mback2k:
It fails with `2022-10-07T20:11:20.3892506Z /usr/bin/sed: couldn't write 8192 items to stdout: No space left on device`

bagder:
> It fails with  No space left on device

It says that, but then after that error message it manages to generate four separate release tarballs at almost 20MB in total. So not entirely out of space clearly.

From what I read, GitHub actions provide [14GB disk space](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and I thought we do much larger build-from-source jobs than this.

We could add a `make clean` after that first make to clean off temp files before `maketgz` runs.

mback2k:
I guess we are not running into a file size, but file count limit, probably due to storage-based anti-mining protection or something.

mback2k:
I opened a support ticket with GitHub to have them help us investigate the root cause here.

mback2k:
GitHub Support came back to me with this additional information:




> We are seeing an exit code 2 which means that your step verify in-tree configure build including install sh commands might be the issue. For example, your [script](https://github.com/curl/curl/blob/6a7c776ee6cf21fab0e60c9d95ef1cb9a9f637bc/scripts/installcheck.sh#L40) but this is hard to tell because we do not have the full logs. Can you do some more troubleshooting on your end and let us know your results.

> 
> The error message that may assist you in troubleshooting:
> ```
> [2022-10-17 19:54:11Z ERR  JobRunner] System.IO.IOException: No space left on device : '/home/runner/runners/2.298.2/_diag/Worker_20221017-194616-utc.log'
>    at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
>    at System.IO.Strategies.BufferedFileStreamStrategy.FlushWrite()
>    at System.IO.Strategies.BufferedFileStreamStrategy.WriteSpan(ReadOnlySpan`1 source, ArraySegment`1 arraySegment)
>    at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
>    at System.Diagnostics.TextWriterTraceListener.Flush()
>    at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
>    at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
>    at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
>    at GitHub.Runner.Common.JobServerQueue.ProcessWebConsoleLinesQueueAsync(Boolean runOnce)
>    at GitHub.Runner.Common.JobServerQueue.ShutdownAsync()
>    at GitHub.Runner.Worker.JobRunner.ShutdownQueue(Boolean throwOnFailure)
>    at GitHub.Runner.Worker.JobRunner.CompleteJobAsync(IJobServer jobServer, IExecutionContext jobContext, AgentJobRequestMessage message, Nullable`1 taskResult)
> ```
> Hope this helps!

Sounds like we have two issues here: 1) writing too much to stdout/stderr causing a space issue with their worker logs and 2) the check eventually failing with exit code 2, but that cannot be verified due to 1).

I will try to split the job into a job-sequence to reduce the amount of log output per job.

bagder:
:+1:

jay:
> writing too much to stdout/stderr

uh what? That sounds like a bug on their end. We have to watch our printfs now? How much is "too much"?

bagder:
I don't understand how to debug this. How do we figure out why a certain step is "skipped" ?

mback2k:
I am still in contact with GitHub Support. They are again investigating this for us.

mback2k:
So GitHub confirmed we are running out of diskspace during `maketgz`. Further investigation on my side lead to:

`./config.status --recheck` is causing a `tests/libtest/Makefile.tmp3` to completely fill the disk (14 GB).

bagder:
Sounds like #9843 in the works.

mback2k:
That seems to be the case as the file has an endless distclean section with an endless repetitive stream of rm entries.

bagder:
Right now, I think the only way to accomplish this would be to also install another automake version than 1.16.5 as a pre-req.

mback2k:
Dowgrading to Ubuntu 18.04 (deprecated and soon to be removed from GitHub Actions) with automake 1.5 fixes the issue.

But this is not a feasible solution due to the deprecation... :/

andy5995:
> Dowgrading to Ubuntu 18.04 (deprecated and soon to be removed from GitHub Actions) with automake 1.5 fixes the issue.
> 
> But this is not a feasible solution due to the deprecation... :/

You can do it easily with docker.

https://github.com/0ad-matters/0ad-appimage/blob/061344d03a46d25698161f7fc9b6e7ac22448aed/.github/workflows/appimage.yml#L28

In that example, I pull an Ubuntu 18.04 docker image created using the Dockerfile in the repo root.

Although what you might prefer is to make a docker image using a more recent version of Ubuntu, or some other distro. As part of the Dockerfile, install an older version of automake. With Arch I know, you can install several version of [automake from the AUR](https://aur.archlinux.org/packages?O=0&SeB=nd&K=automake).

Also, it looks like automake 1.15 is available with Ubuntu focal (20.04) https://packages.ubuntu.com/focal/automake-1.15

pheiduck:
Another solution could be install automake like this `sudo apt install automake=1.5`

pheiduck:
solution (2) is build **automake 1.15.1** from source tarball, as it is a small package it will not very hard encrease the CI time

mback2k:
@bagder after https://github.com/curl/curl/commit/73e9e6d767296c5759972d07f2441c8993ccbb9d this is finally working!

But now the following error appeared: https://github.com/curl/curl/actions/runs/4478314300/jobs/7870978453?pr=9532#step:4:5404
```
bash: scripts/installcheck.sh: No such file or directory
Error: Process completed with exit code 127.
```

Is this expected? Is `scripts/installcheck.sh` supposed to be part of the release archive or not?

pheiduck:
> Is this expected? Is scripts/installcheck.sh supposed to be part of the release archive or not?

It's not included on the 8.0.1 tarball, so maybe not?

mback2k:
> > Is this expected? Is scripts/installcheck.sh supposed to be part of the release archive or not?
> 
> It's not included on the 8.0.1 tarball, so maybe not?

Which could be an oversight that is now detected by this CI. That is why I was asking in the first place.

bagder:
> Is this expected? Is scripts/installcheck.sh supposed to be part of the release archive or not?

I don't think it needs to be shipped. The script is only meant for checking that an install is actually installing what is present in the source tree.

mback2k:
> > Is this expected? Is scripts/installcheck.sh supposed to be part of the release archive or not?
> 
> I don't think it needs to be shipped. The script is only meant for checking that an install is actually installing what is present in the source tree.

Okay, then I will re-work this PR to work without it being present in the release archive. Thanks for the feedback!

.github/workflows/distcheck.yml
```diff
@@ -0,0 +1,123 @@
...
+        echo "::stop-commands::$(uuidgen)"
+        tar xvf curl-99.98.97.tar.gz
+        pushd curl-99.98.97
+        ./configure --prefix=$HOME/temp --without-ssl
+        make
+        make TFLAGS=1 test
+        make install
+        popd
+        # basic check of the installed files
+        bash scripts/installcheck.sh $HOME/temp
```
bagder, line 75: This should run the script from the git tree since the `installcheck.sh` itself is not included in the tarball.

Maybe just:

`bash ../scripts/installcheck.sh $HOME/temp`

bagder:
What is the reason this PR is still a draft?

mback2k:
Not enough time to rework it.

mback2k:
@bagder I got `scripts/installcheck.sh` going, but now it yields:
```
--- /dev/fd/63	2023-04-19 20:07:18.019125178 +0000
+++ /dev/fd/62	2023-04-19 20:07:18.023125203 +0000
@@ -277,7 +277,6 @@
 CURLOPT_PROXY_TLSAUTH_USERNAME.3
 CURLOPT_PROXY_TRANSFER_MODE.3
 CURLOPT_PUT.3
-CURLOPT_QUICK_EXIT.3
 CURLOPT_QUOTE.3
 CURLOPT_RANDOM_FILE.3
 CURLOPT_RANGE.3
@@ -312,8 +311,6 @@
 CURLOPT_SOCKS5_GSSAPI_SERVICE.3
 CURLOPT_SSH_AUTH_TYPES.3
 CURLOPT_SSH_COMPRESSION.3
-CURLOPT_SSH_HOSTKEYDATA.3
-CURLOPT_SSH_HOSTKEYFUNCTION.3
 CURLOPT_SSH_HOST_PUBLIC_KEY_MD5.3
 CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256.3
 CURLOPT_SSH_KEYDATA.3
ERROR: installed libcurl docs mismatch
```
Is this maybe a general problem currently on master or a problem of this PR?

pheiduck:
@mback2k I think this happens because curl is already installed on the Ubuntu CI.
We could uninstall it:
```
apt purge -y curl
```

pheiduck:
Okay, the installed lib/curl does not seem to be the culprit.

pheiduck:
It must be removed before the CI starts the intended process.
```
ERROR: installed libcurl docs mismatch
```

.github/workflows/distcheck.yml
```diff
@@ -0,0 +1,99 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  maketgz-and-verify-in-tree:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v3
+
```
pheiduck, line 26:
```suggestion

  - run: sudo apt-get purge -y curl libcurl4 libcurl4-doc
    name: 'remove preinstalled curl libcurl4{-doc}'
```

mback2k: Thanks, but even this does not solve the problem.

pheiduck:
```
$ find /usr -name "libcurl*"
$ libconfig -v | less
```

bagder:
Clearly this is not moving much. Close this now, maybe return in a future.

mback2k:
I am returning to this topic today, so re-opening to trigger another test-run.

mback2k:
It is finally working! https://github.com/curl/curl/actions/runs/6239889256

I found some docs to be missing from Makefile.inc, see: https://github.com/curl/curl/pull/9532/commits/ede7f41c9133d2386f8eeb9a280cbf039376d016

PR #9533, opened 2022-09-17
CI/GHA: cancel outdated CI runs on new PR changes

mback2k:
Avoid letting outdated CI runs continue if a PR receives new changes. Outside a PR we let them continue running
by tying the concurrency to the commit hash instead.

Also only let one CodeQL or Hacktoberfest job run at a time.

Other CI platforms we use have this build in, but GitHub unfortunately neither by default nor with a simple option.

This saves CI resources and therefore a little energy.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -5,8 +5,6 @@
 name: CodeQL
 
 on:
-  # Trigger the workflow on push or pull requests, but only for the
-  # master branch
   push:
     branches:
     - master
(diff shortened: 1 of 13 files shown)


mback2k:
References:
- https://docs.github.com/en/actions/using-jobs/using-concurrency
- https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
- https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

bagder (approved):
I can't comment on the yaml content, but I approve of what you want to do wit this.

cmeister2 (approved):
I did read the docs and it seems like this will work, so happy to approve 😅

PR #9536, opened 2022-09-19
winbuild/MakefileBuild.vc: handle spaces in libssh(2) include paths

bagder:
Patched-by: Mark Itzcovitz
Bug: https://curl.se/mail/lib-2022-09/0038.html


MarcelRaad (approved):

dfandrich:
What happens if the directories contain double quotes?

bagder:
> What happens if the directories contain double quotes?

I bet it all goes crazy. But presumably that's rarer than space. Also, this quoting of the path variables is done in many other places already within this makefile so this change basically just makes these variables treated the same ways as the other.

MarcelRaad:
File and folder names for NTFS (or on Windows?) cannot contain double quotes. I cannot find a real reference, but neither Windows Exporer nor mkdir allow it and the Windows API function references at least link to these "best practices" that say that they're reserved characters:
https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

dfandrich:
I created a directory with double quotes on a NTFS filesystem on Linux without
issue (using ntfs-3g). Maybe that's a bug in that driver.

jay:
> I created a directory with double quotes on a NTFS filesystem on Linux without issue (using ntfs-3g).

A filesystem may allow it but Windows doesn't. It is possible to trick Windows into writing banned characters (refer to naming conventions) but that is not a case we should cover.

bagder:
I think we can assume that double-quotes in Windows paths are highly unusual. Spaces are not.

PR #9540, opened 2022-09-19
Forced IPv4 resolve mode to use only A queries.

dkarpov1970:
These changes force using IPv4 name resolution queries for the "IP4 only" resolve mode in IPv6-enabled dual-stack libcurl.
They protect "IPv4 only" transfers from undesired bad IPv6-related side effects and make IPv4 transfers in dual-stack libcurl behave the same way as in IPv4 single-stack libcurl.

The e-mail thread describing the problem and some ideas to solve it can be found at:
https://curl.se/mail/lib-2022-05/0004.html


bagder (approved):

bagder:
Thanks!

PR #9542, opened 2022-09-20
curl-compilers.m4: for gcc + want warnings, set gnu89 standard

bagder:
To better verify that the code is C89

### DIFF
--- a/lib/vquic/msh3.c
+++ b/lib/vquic/msh3.c
@@ -381,9 +381,6 @@ static void MSH3_CALL msh3_shutdown(MSH3_REQUEST *Request, void *IfContext)
   (void)stream;
 }
 
-static_assert(sizeof(MSH3_HEADER) == sizeof(struct h2h3pseudo),
-              "Sizes must match for cast below to work");
-
 static ssize_t msh3_stream_send(struct Curl_easy *data,
                                 int sockindex,
                                 const void *mem,
@@ -396,6 +393,9 @@ static ssize_t msh3_stream_send(struct Curl_easy *data,
   struct h2h3req *hreq;
 
   (void)sockindex;
+  /* Sizes must match for cast below to work" */
+  DEBUGASSERT(sizeof(MSH3_HEADER) == sizeof(struct h2h3pseudo));
+
   H3BUGF(infof(data, "msh3_stream_send %zu", len));
 
   if(!stream->req) {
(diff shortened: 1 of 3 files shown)


MarcelRaad:
This will be a problem with third-party libraries using features from later standards, like C++ comments, in their public headers. Maybe add it to some CI builds instead?

bagder:
yeah, but more important this identifies numerous places within libcurl itself that is not proper C89. That was my primary thought with this, to try it out. It might not make sense to land in the short term.

The question also comes down to: are those nits worth fixing if nobody has suffered from any problems with them while we are slowly considering bumping to C99 in a future...

bagder:
I decided they are probably worth fixing because:

If (and this has not been decided yet) we change the requirement to C99 at some point, then whatever is in git before that switch will be the last C89 compliant curl code and it could be worth having that in a fine shape. Compliance wise.

monnerat:
FYI:
I did the same kind of test locally with:
```
CFLAGS="... -std=C89 ..."
CPPFLAGS="... -D_XOPEN_SOURCE=700 ..."
```
This exhibited problems fixed (sucessfully) in 6fbf7d3, 9eec754 and c3e634d: no compiler message for the current master.
Of course, not everything was enabled in my local test.

I had to define `_XOPEN_SOURCE` to enable definition of some library functions. There are configure tests for most of the latter (i.e.: `HAVE_FCHMOD`), but it seems `AC_CHECK_FUNCS` does not use our predefined `CFLAGS` while checking.
A partular case is function `kill()`: it is used unconditionally by curl but its definition is disabled by `-std=C89`. There are probably a lot of other external definitions hidden by `-std=C89` but re-enabled by -D_XOPEN_SOURCE=700.

As a consequence we might question ourselves if curl's C89 requirement only targets the language or also the libraries.

dfandrich:
C89 doesn't provide nearly enough of an API to build curl (e.g. sockets) so
it's safe to assume that we're talking about language features only. Various
APIs outside C89 are also required (e.g. POSIX). I used to run an autobuild that
tried to build with a minimum set of such APIs (by setting things like
-D_XOPEN_SOURCE=600) to try to make sure that it wouldn't depend on too much
more. It might be interesting to turn that into a CI build.

monnerat:
`-D_XOPEN_SOURCE=600` was not enough for me: `strndup` requires 700.

monnerat:
> C89 doesn't provide nearly enough of an API to build curl (e.g. sockets)

Agreed: see the comment about `kill()` above. There might be others.
But the sockets declarations are not conditioned on POSIX/XOPEN, although not in the standard C library. In fact, they were already present in 4.2BSD (1983).

bagder:
> it's safe to assume that we're talking about language features only.

Yes exactly: *language features only*. Everything else we can detect and adapt to.

PR #9544, opened 2022-09-20
test1119: scan all public headers

bagder:
Previously this test only scanned a subset of the headers, which made us accidentally miss symbols that were provided in the others. Now, the script iterates over all headers present in include/curl.


bagder:
This updated scan also correctly finds the #9552 issue, a wrongly spelled `CURLH_*` define.

bagder:
#9523 fixes that issue, and I will rerun this build once that has land to confirm.

PR #9546, opened 2022-09-20
tool: remove protocol count limitation

monnerat:
Replace bit mask protocol sets by null-terminated arrays of protocol tokens. These are the addresses of the protocol names returned by curl_version_info().

Protocol names are sorted case-insensitively before output to satisfy CI tests matches consistency.

The protocol list returned by curl_version_info() is augmented with all RTMP protocol variants.

Test 1401 adjusted for new alpha order output.

There is still however an output string length limitation (currently 704 characters for dynbuf) left.
This implements what has been discussed in #9534.

### DIFF
--- a/lib/version.c
+++ b/lib/version.c
@@ -338,6 +338,11 @@ static const char * const protocols[] = {
 #endif
 #ifdef USE_LIBRTMP
   "rtmp",
+  "rtmpe",
+  "rtmps",
+  "rtmpt",
+  "rtmpte",
+  "rtmpts",
 #endif
 #ifndef CURL_DISABLE_RTSP
   "rtsp",
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -75,8 +75,7 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
   const char *str = ptr;
   const size_t cb = size * nmemb;
   const char *end = (char *)ptr + cb;
-  char *scheme;
-  proto_t protocol = proto_last;
+  const char *scheme = NULL;
 
   /*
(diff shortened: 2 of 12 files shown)


lib/version.c
```diff
@@ -339,6 +339,21 @@ static const char * const protocols[] = {
...
+#endif
+#ifdef USE_LIBRTMP
+  "rtmpt",
+#endif
+#ifdef USE_LIBRTMP
+  "rtmpte",
+#endif
+#ifdef USE_LIBRTMP
+  "rtmpts",
+#endif
```
bagder, line 356: I think you could put them all within a single `#ifdef` clause, it would look less weird to me.

bagder:
> There is still however an output string length limitation (currently 704 characters for dynbuf) left.

How did you pick 704 as the maximum string length?

lib/version.c
monnerat: Agreed, but I wanted to respect the comment above.
Will change.

monnerat:
> How did you pick 704 as the maximum string length?

`#define MAX_PROTOSTRING (64*11) /* Enough room for 64 10-chars proto names. */`

Both 64 an 10 should be large enough for some years :-)

bagder:
Ah sorry, I missed that.

> Both 64 an 10 should be large enough for some years :-)

Yeah I totally agree. Although the maximum length for a dynbuf is there to mostly detect/prevent havoc and abuse, so it would not really be bad to also just set the max to 7040 or any other arbitrary value that isn't in the megabytes range.

But I'm also fine with 704. With mostly 5-letter scheme names and a comma, it allows 117 of them.

lib/version.c
monnerat: Change pushed.

bagder (approved):

PR #9551, opened 2022-09-21
unit test 1655: make it C89-compliant

monnerat:
Initializations performed in unit test 1655 use automatic variables in aggregates and thus can only be computed at run-time. Using gcc in C89 dialect mode produces warning messages like:

```
unit1655.c:96:7: warning: initializer element is not computable at load time [-Wpedantic]
   96 |     { toolong, DOH_DNS_NAME_TOO_LONG },  /* expect early failure */
      |       ^~~~~~~
```

Fix the problem by converting these automatic pointer variables to static arrays.

### DIFF
--- a/tests/unit/unit1655.c
+++ b/tests/unit/unit1655.c
@@ -47,7 +47,7 @@ UNITTEST_START
  * Prove detection of other invalid input.
  */
 do {
-  const char *max =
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #9553, opened 2022-09-21
hostip: lazily wait to figure out if IPv6 works until needed

bagder:
The check may take many milliseconds, so now that is done once the value is first needed. Also, this change makes sure that the value is not used if the resolve is set to be IPv4-only.


dkarpov1970:
Looks good to me!

PR #9562, opened 2022-09-21
chkspeed: improve portability

monnerat:
The example program chkspeed uses `strncasecmp`() which is not portable across systems. Replace calls to this function by tests on characters.

The initial motivation for this change is the use of a wrong header file for the function definition, causing compilation messages in C89 mode.


bagder (approved):

bagder:
Thanks!

PR #9569, opened 2022-09-22
Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")

pheiduck:
ensure compatibility with macOS 13

### DIFF
--- a/lib/mprintf.c
+++ b/lib/mprintf.c
@@ -964,7 +964,7 @@ static int dprintf_formatf(
 #endif
         /* NOTE NOTE NOTE!! Not all sprintf implementations return number of
            output characters */
-        (sprintf)(work, formatbuf, p->data.dnum);
+        (snprintf)(work, formatbuf, p->data.dnum);
 #ifdef __clang__
 #pragma clang diagnostic pop
 #endif


lib/mprintf.c
```diff
@@ -964,7 +964,7 @@ static int dprintf_formatf(
 #endif
         /* NOTE NOTE NOTE!! Not all sprintf implementations return number of
            output characters */
-        (sprintf)(work, formatbuf, p->data.dnum);
+        (snprintf)(work, formatbuf, p->data.dnum);
```
bagder, line 967: 1. This is not how `snprintf`  works
2. `snprintf` is not in C89 and thus not always available
3. This use of `sprintf()` is the only one in libcurl and has been reviewed and vetted

pheiduck: okay, then we have to use `CFLAGS: "-Wno-deprecated-declarations"` for the m1 arm CI to build it.
Otherwise this error will be shown.

bagder: Or we fix it by checking in configure if snprintf exists and make sure to use it conditionally and correctly in `mprintf.c`

bagder: Let me give it a shot

bagder: #9570

rsbeckerca: IIRC, does not `snprintf` take the `work` size in characters as the second argument?

bagder: Yes, hence my comment above:

> This is not how snprintf works

But I've already merged #9570 which had that corrected.

PR #9570, opened 2022-09-22
mprintf: use snprintf if available

bagder:
This is the single place in libcurl code where it uses the "native" s(n)printf() function. Used for writing floats. The use has been reviewed and vetted and uses a HUGE target buffer, but switching to snprintf() still makes this safer and removes build-time warnings.

Reported-by: Philip Heiduck

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3535,6 +3535,7 @@ AC_CHECK_FUNCS([fnmatch \
   setlocale \
   setmode \
   setrlimit \
+  snprintf \
   utime \
   utimes
 ],[
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1079,6 +1079,7 @@ check_symbol_exists(getrlimit      "${CURL_INCLUDES}" HAVE_GETRLIMIT)
 check_symbol_exists(setlocale      "${CURL_INCLUDES}" HAVE_SETLOCALE)
 check_symbol_exists(setmode        "${CURL_INCLUDES}" HAVE_SETMODE)
 check_symbol_exists(setrlimit      "${CURL_INCLUDES}" HAVE_SETRLIMIT)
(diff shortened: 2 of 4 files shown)


bagder:
@jay @MarcelRaad does windows have snprintf widely enough you think so that we can add `#define HAVE_SNPRINTF` to `config-win32.h` ?

MarcelRaad:
It should be safe to enable it for Visual Studio 2015 and later (`_MSC_VER >= 1900`). IIRC, VS 2013 and earlier had a non-conforming implementation (or maybe no implementation at all and only `_snprintf`, which would probably a smaller problem).

bagder:
I think it is fine to just ignore it for older version. This is almost like a no-op anyway and is mostly to satisfy humans and tools.

configure.ac
```diff
@@ -3535,6 +3535,7 @@ AC_CHECK_FUNCS([fnmatch \
   setlocale \
   setmode \
   setrlimit \
+  snprintf \
```
MarcelRaad, line 3538: Is it sufficient to have the function here or do we want it to have conforming behavior? Early MSVC and other implementations have issues related to null-termination and return values.

bagder: Hm, complicated. Can the autotools build even be used to compile with MSVC though?

MarcelRaad: No, but CMake can. And it's not only MSVC, unfortunately. At least also HP, according to https://demin.ws/blog/english/2013/01/28/use-snprintf-on-different-platforms/.

bagder: The HPUX behavior shown there actually seems fine for our use. Of that listing, the worst one is Windows 7, Visual Studio 12 since it doesn't zero terminate when capped. All the others do.

So no, with MSVC it seems just detecting it is not enough...

PR #9572, opened 2022-09-22
lib517: fix C89 constant signedness

monnerat:
In C89, positive integer literals that overflow an int but not an unsigned int may be understood as a negative int.

```
lib517.c:129:3: warning: this decimal constant is unsigned only in ISO C90
   {"Sun, 06 Nov 2044 08:49:37 GMT", 2362034977 },
   ^
```


bagder (approved):

bagder:
Thanks!

PR #9575, opened 2022-09-22
Disable use of FLOSS for 64-bit NonStop builds.

rsbeckerca:
Older 32-bit builds currently need FLOSS. This dependency may be removed in future OS releases.

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -322,8 +322,10 @@
 #include <assert.h>
 #endif
 
-#ifdef __TANDEM /* for nsr-tandem-nsk systems */
-#include <floss.h>
+#ifdef __TANDEM /* for ns*-tandem-nsk systems */
+# if ! defined __LP64
+#  include <floss.h> /* FLOSS is only used for 32-bit builds. */
+# endif
 #endif
 
 #ifndef STDC_HEADERS /* no standard C headers! */


jay:
I can't find anything about this, do you have a reference?

rsbeckerca:
@jay The original port was included in 2004 - before I was involved. In those days, there was no 64-bit NonStop OS. As time went by, NonStop moved to ia64 and x86. Parts of the utilities, including FLOSS, which is a bit like glibc, stayed in 32-bit land. The APIs in the NonStop POSIX side have evolved and cover far more than they did in 2004. As a result, for 64-bit builds, FLOSS is rarely needed, and not at all in curl. There are pending build issues we are trying to address for 32-bit code that avoid the need for FLOSS but I am not there yet with the HPE Dev team. As part of moving curl to 64-bit builds, I found that FLOSS not only is not needed, but causes link issues if included, so it has to be removed from all parts of the build. At present, FLOSS only exists in 32-bit form on NonStop, so creating a 64-bit curl with FLOSS is not possible. There is still a need in my community for 32-bit libcurl.a, so that has to be retained and needs FLOSS.

Therefore, the solution, as I see it, is disabling the use of `floss.h` when a 64-bit build is being done, with the compiler generated knob being `__LP64`.

bagder (approved):
Since @rsbeckerca is working on and with this platform, I figure he is the most suitable person to have a say on how this build should be adjusted.

bagder:
Thanks!

PR #9576, opened 2022-09-23
tool: adjust handling of non-existent url scheme

monnerat:
The test for a non-existent protocol scheme in an URL is currently always false and the warning message can never occur. Fortunately this has no impact as this test is now unconditionally excluded.
This commit fixes it however, in case this warning is reenabled.

A debug assertion is also added to verify protocols included/excluded in a set are always tokenized.

Follow-up to commit 677266c.

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -1249,14 +1249,7 @@ static CURLcode single_transfer(struct GlobalConfig *global,
         if(result)
           break;
 
-        /* here */
         use_proto = url_proto(per->this_url);
-#if 0
-        if(!use_proto) {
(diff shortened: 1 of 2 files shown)


bagder (commented):
I don't think this improves the code. I think we should rather remove the `#if 0` section that checks `use_proto` and leave the "???" handling in the `url_proto` function as it is.

monnerat:
> I think we should rather remove the #if 0 section that checks use_proto

I agree. There is very little chance to have the warning reenabled anyway since it's redundant with the later error returned by `curl_easy_perform`().

I'll update this PR for that.

bagder (approved):

bagder:
Thanks!

PR #9583, opened 2022-09-24
version: add a feature names array to curl_version_info_data

monnerat:
New field `feature_names` contains a pointer  to a null-terminated sorted array of feature names. Bitmask field `features` is deprecated.

This will allow us to support more than 32 features.

Whether the `strcasecmp4sort`() procedure should be moved to strcase.c is questionable.


lib/version.c
```diff
@@ -496,11 +502,69 @@ static curl_version_info_data version_info = {
...
+{
+#if defined(HAVE_STRCASECMP)
+  return strcasecmp(*(char * const *) p1, *(char * const *) p2);
+#elif defined(HAVE_STRCMPI)
+  return strcmpi(*(char * const *) p1, *(char * const *) p2);
+#elif defined(HAVE_STRICMP)
+  return stricmp(*(char * const *) p1, *(char * const *) p2);
+#else
+  return strcmp(*(char * const *) p1, *(char * const *) p2);
+#endif
```
bagder, line 559: Not a fan of using these functions in libcurl. We use our own functions all over for a reason: they are stable and independent of the locales. This set is not.

I would rather that we provide a version of our internal one with a return code that is fine to sue for qsort.

bagder: But otoh: why not just have the list in a sorted order to begin with to avoid the qsort?

monnerat: >  I would rather that we provide a version of our internal one with a return code that is fine to sue for qsort.

I was hesitating. I'll create a function in strcase.c for that.
> But otoh: why not just have the list in a sorted order to begin with to avoid the qsort?

Because next features wont have a `CURL_VERSION_*` bitmask and thus would be entered in array by another loop.
I plan to make features for `MIME` and `form-API` in a very near future. Maybe `headers-API` could be one too.

bagder: > Because next features wont have a CURL_VERSION_* bitmask and thus would be entered in array by another loop.

Ok, I don't think I miss something because I don't see the point in doing additional ones in a second loop.

>  I plan to make features for MIME and form-API in a very near future. Maybe headers-API could be one too.

Why add those? I can't recall anyone ever having asked for for missed them being present there.

bagder: The #ifdefs could also be used *in* the list, so that disabled features also removes the strings from memory.

monnerat: > Ok, I think I miss something because I don't see the point in doing additional ones in a second loop.

Because they don't have a `CURL_VERSION_*` bitmask.

> Why add those? I can't recall anyone ever having asked for for missed them being present there.

I ran "make check" on a build with `--disable-mime` and several tests fail.
I plan also to really deprecate the form api using a `CURL_DISABLE_FORM_API` conditional and tests for it should check the feature.

> The #ifdefs could also be used in the list, so that disabled features also remove the strings from memory.

The loop introduced in this PR checks the `CURL_VERSION_*` bits in `features` field because some of them are added at run-time. That cannot be done for features without a bitmask.

Alternative: if the bitmask in the table is 0, always include the feature in the array. That could be feasible.

Anyway suppressing the sort would disallow determining the presence of a new feature at run-time and still have the array in alpha order.

bagder: > The loop introduced in this PR checks the CURL_VERSION_* bits in features field because some of them are added at run-time. That cannot be done for features without a bitmask.

Features that are disabled at build-time don't need their strings as features so they could be removed with #ifdef use. In the spirit of removing things from occupying space when the are disabled.

monnerat: > Features that are disabled at build-time don't need their strings as features so they could be removed with #ifdef use. 

Sure! But that will duplicate all conditional stuff of `version_info.features` initialisation for current features. The idea behind having 2 loops and 2 tables was to avoid this duplication by testing the bitmask for current features in the first loop and to have a compile-time conditioned table for the new ones. I think they may be combined, but I would retain unconditioned entries with a bitmask to avoid conditions duplication.

monnerat: I just pushed a version that:

- has the `feature_names_table` sorted
- has all strings for existing features
- supports (conditioned) new features in single table.
- has no qsort.
- introduces `MIME` feature (as an example).

bagder: We have several users who are extremely memory and foot print conscious. They .build their own libcurl with only a small set of features and protocols enabled.

For those users, it is important that every feature "carries its own weight" so that when the feature is disabled, as much as possible of that feature is gone and not present in the build. For example: it does not make sense to include the string "http3" in a build that does not have the feature enabled. When we can easily avoid it.

monnerat: > For those users, it is important that every feature "carries its own weight" so that when the feature is disabled, as much as possible of that feature is gone and not present in the build.

I'm aware of that and, as you might have understood, avoiding duplication of conditionals is only motivated by not introducing and error-prone context.
I have an idea for combining both wishes at the expense of a extra function pointer in each compiled-in table entry... let me the time to code it!

monnerat: The last version pushed should satisfy everything at this point.

docs/libcurl/curl_version_info.3
```diff
@@ -216,6 +221,14 @@ backends). (Added in 7.21.4)
 libcurl was built with support for Unix domain sockets.
 (Added in 7.40.0)
 .RE
+
+The use of this field is deprecated as not all features may be represented in
+the bitmask. Use \fIfeature_names\fP instead.
+
+\fIfeature_names\fP is a pointer to an array of string pointers, containing the
+names of the features that libcurl supports. The array is terminated by a NULL
+entry.
```
bagder, line 230: This mentioned that `features` is deprecated, but the man page still only documents the bits in that bitmask. It says nothing about the potential names in `feature_names` that users of this function probably are very interested in. The list above should probably list both?

monnerat: Yes, I can do it. Note this means the man page will still have to be updated when a new feature is introduced.

bagder: That was always the case. How is a user supposed to know what the feature means if it isn't documented?

monnerat: As `features` is deprecated, do you expect me to remove the bit definitions in the doc ?

bagder: I think it is way too early for that. The documentation also needs to work as a resource for people reading existing code, and there will be code using this flags for the foreseeable future and then the documentation needs to describe them.

monnerat: > I think it is way too early for that.

Then the doc should be ready now !

docs/libcurl/curl_version_info.3
```diff
@@ -216,6 +223,89 @@ backends). (Added in 7.21.4)
...
+libcurl was built with memory tracking debug capabilities. This is mainly of
+interest for libcurl hackers.
+.IP Unicode
+libcurl was built with Unicode support on Windows. This makes non-ASCII
+characters work in filenames and options passed to libcurl.
+.IP UnixSockets
+libcurl was built with support for Unix domain sockets.
+.IP zstd
+supports HTTP zstd content encoding using zstd library.
+.RE
```
bagder, line 307: This is much better, but now you've duplicated the documentation for each feature, and they're not duplicated identically. I suggest we instead list each feature with both name and bitmask define name, and have *one* explanation for it.

monnerat: Can I drop descriptions of `CURL_VERSION_CONV` and `CURL_VERSION_KERBEROS4`? They're obsolete and have no associated name. Reason: I want to label descriptions by names and these are not named! Or do I associate a dummy name that can never occur :-( ?

monnerat: > each feature with both name and bitmask define name, and have one explanation for it.

Done.
(I wish man had a built-in `.TABLE` macro!)

docs/libcurl/curl_version_info.3
```diff
@@ -106,6 +110,9 @@ typedef struct {
   /* when 'age' is CURLVERSION_TENTH or higher (>= 7.77.0), the members
      below exist */
   const char *gsasl_version; /* human readable string. */
+  /* when 'age' is CURLVERSION_ELEVENTH or higher (>= 7.86.0), the members
```
bagder, line 113: 7.87.0 now!!

docs/libcurl/symbols-in-versions
```diff
@@ -1097,6 +1097,7 @@ CURLUSESSL_CONTROL              7.17.0
 CURLUSESSL_NONE                 7.17.0
 CURLUSESSL_TRY                  7.17.0
 CURLVERSION_EIGHTH              7.72.0
+CURLVERSION_ELEVENTH            7.86.0
```
bagder, line 1100: 7.87.0

src/tool_libinfo.c
```diff
@@ -68,23 +68,81 @@ static struct proto_name_tokenp {
...
+  {"HSTS",           &feature_hsts,       CURL_VERSION_HSTS},
+  {"HTTP2",          &feature_http2,      CURL_VERSION_HTTP2},
+  {"HTTP3",          &feature_http3,      CURL_VERSION_HTTP3},
+  {"HTTPS-proxy",    NULL,                CURL_VERSION_HTTPS_PROXY},
+  {"IDN",            NULL,                CURL_VERSION_IDN},
+  {"IPv6",           NULL,                CURL_VERSION_IPV6},
+  {"Kerberos",       NULL,                CURL_VERSION_KERBEROS5},
+  {"Largefile",      NULL,                CURL_VERSION_LARGEFILE},
+  {"libz",           &feature_libz,       CURL_VERSION_LIBZ},
+  {"MIME",           NULL,                0},
```
bagder, line 106: This is not an actual feature is it?

bagder (commented):
If libcurl returns more/other features in its name list than the list in tool_libinfo.c knows about, I assume they will still get output fine?

docs/libcurl/curl_version_info.3
monnerat: Time past soooo fast !

docs/libcurl/symbols-in-versions
monnerat: Idem

src/tool_libinfo.c
monnerat: No it isn't. This is a leftover from local tests. Do not harm, but as useless, removed now.

monnerat:
All your remarks taken into account now. Thanks for pointing on them.

> If libcurl returns more/other features in its name list than the list in tool_libinfo.c knows about, I assume they will still get output fine?

Yes. the feature list is taken directly from the name array:

https://github.com/curl/curl/blob/f8b8dc4113d4a3115e56933980b9ab0f76de32a8/src/tool_help.c#L177-L182

In case an old library does not provide the feature array, the later is built by the tool from the flags:

https://github.com/curl/curl/blob/f8b8dc4113d4a3115e56933980b9ab0f76de32a8/src/tool_libinfo.c#L160-L170

bagder:
Thanks!

PR #9585, opened 2022-09-24
[MSDOS] name-clash in url.c

gvanem:
In commit https://github.com/curl/curl/commit/764c958c52edb427f39530f53623e85d34409904, the new function `resolve_ip()`
clashes with an [internal function](https://github.com/gvanem/Watt-32/blob/master/inc/tcp.h#L376) in Watt-32. Hence undefine it in libcurl.

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -519,9 +519,6 @@
 #    ifdef byte
 #      undef byte
 #    endif
-#    ifdef resolve_ip
-#      undef resolve_ip
-#    endif
 
 #  endif /* MSDOS */
 
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -519,6 +519,9 @@
 #    ifdef byte
 #      undef byte
 #    endif
+#    ifdef resolve_ip
+#      undef resolve_ip
+#    endif
 
 #  endif /* MSDOS */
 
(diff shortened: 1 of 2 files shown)


bagder:
Since `resolve_ip` is a static function in url.c, wouldn't it be a cleaner fix to just add `#undef resolve_ip /* avoid a Watt-32 duplicate symbol */` just before the function declaration in that file?

dfandrich:
Or just rename the function?

bagder:
> Or just rename the function?

oh yes, that's actually even better

gvanem:
> oh yes, that's actually even better

Ok. Fine by me.

gvanem:
Perhaps I misunderstood something. Should I rename `resolve_ip()` in `url.c` to avoid this name-clash?

MarcelRaad (approved):

lib/curl_setup.h
```diff
@@ -519,6 +519,9 @@
 #    ifdef byte
 #      undef byte
 #    endif
+#    ifdef resolve_ip
```
MarcelRaad, line 522: Now we don't need this anymore, do we?

gvanem: Ops. No. Removing.

bagder (approved):

bagder:
Thanks!

PR #9586, opened 2022-09-25
autotools: allow --enable-symbol-hiding with windows

vszakats:
This local autotools logic was put in place in 9e24b9c7afbcb81120af4cf3f6cdee49a06d8224 (in 2012) which disabled it for Windows unconditionally. Testing reveals that it actually works with tested toolchains (mingw-w64 and CI ones), so let's allow this build feature on that platform. Bringing this in sync with CMake, which already supported this.

Closes #9586

### DIFF
--- a/m4/curl-confopts.m4
+++ b/m4/curl-confopts.m4
@@ -458,8 +458,7 @@ AC_DEFUN([CURL_CONFIGURE_SYMBOL_HIDING], [
   AC_MSG_CHECKING([whether hiding of library internal symbols will actually happen])
   CFLAG_CURL_SYMBOL_HIDING=""
   doing_symbol_hiding="no"
-  if test x"$curl_cv_native_windows" != "xyes" &&
-    test "$want_symbol_hiding" = "yes" &&
+  if test "$want_symbol_hiding" = "yes" &&
     test "$supports_symbol_hiding" = "yes"; then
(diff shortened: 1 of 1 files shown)


vszakats:
Tested fine with mingw-w64. Let's test with the legacy mingw and the others.

jay:
But does it actually do anything on windows?

https://github.com/curl/curl/blob/93d092867f0f2c78571983040ef75e078ee1a4c4/include/curl/curl.h#L112-L126

vszakats:
Sure, unless I'm fatally misunderstanding something. After the patch `libcurl.dll` is longer exports every existing public function linked to it, only the curl API. This let me drop the manual solution using a pre-rolled `.def` file (as used in `Makefile.m32` builds). Ref this curl-for-win patch: https://github.com/curl/curl-for-win/commit/aa0da2b0f28dee1407c97b37c6c4861772e359cd

The compiler macro is a noop on Windows, but there is extra logic run around the linking phase, and it may also trigger some linker options. I didn't investigate the internals further.

vszakats:
Example build here: https://ci.appveyor.com/project/curlorg/curl-for-win/builds/44876993

jay (approved):
> there is extra logic run around the linking phase

ah ok

PR #9588, opened 2022-09-25
http: Curl_output_aws_sigv4 compatibility fixes

cbodley:
some fixes after interop testing against the s3-compatible [Ceph Object Gateway](https://docs.ceph.com/en/latest/radosgw/):

* when the server runs on a non-standard port, the `Host: localhost:8000` header included the port number but the `host:localhost` in `canonical_headers` did not. this led to `SignatureDoesNotMatch` errors

* when a request omits the `x-amz-content-sha256` header, the server uses `UNSIGNED-PAYLOAD` as the payload hash when generating its `canonical_request`. the curl client is using the hash of an empty string, so this also led to `SignatureDoesNotMatch` errors. if the client includes its empty hash in a `x-amz-content-sha256` header, it works

* even with the `x-amz-content-sha256` header added, s3 PUT requests failed with `XAmzContentSHA256Mismatch` because the client was sending its empty hash. using `UNSIGNED-PAYLOAD` for s3 PUT requests avoids having to calculate a real payload hash

see https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html for more about `x-amz-content-sha256` in s3


jay:
Ref: https://github.com/curl/curl/pull/8935

bagder (changes requested):
This PR fails many CI tests and it has merge conflicts.

cbodley:
> when the server runs on a non-standard port, the Host: localhost:8000 header included the port number but the host:localhost in canonical_headers did not. this led to SignatureDoesNotMatch errors

this one has since been fixed in https://github.com/curl/curl/pull/7966

the other issues with x-amz-content-sha256 with are still being discussed in https://github.com/curl/curl/pull/8935 and https://github.com/curl/curl/pull/9804

PR #9589, opened 2022-09-25
vtls: use a thread lock for multissl init

jay:
.. because multissl_setup() and multissl_version() can be called by functions that do not hold the global init lock.

Ref: https://curl.se/mail/lib-2022-09/0104.html

Closes #xxxx

---

This PR is to further discussion on the mailing list about why multissl_setup and multissl_version would need to be protected by a lock. But why would they not be called during libcurl initialization? Yet there's a bunch of functions calling multissl_setup without a lock, why? IMO it seems like compensating for user error not initializing. I think we probably don't need this PR but just in case I'm wrong here's how it can be done.

### DIFF
--- a/lib/easy.c
+++ b/lib/easy.c
@@ -326,7 +326,7 @@ CURLsslset curl_global_sslset(curl_sslbackend id, const char *name,
 
   global_init_lock();
 
-  rc = Curl_init_sslset_nolock(id, name, avail);
+  rc = Curl_init_sslset(id, name, avail);
 
   global_init_unlock();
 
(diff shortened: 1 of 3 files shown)


jay:
Closing in favor of #9583 which will clarify the documentation for curl_version_info that it is not thread safe before initialization occurs.

> Yet there's a bunch of functions calling multissl_setup without a lock, why?

I still don't know the answer to this, however it seems those functions are stub functions used by Curl_ssl_multi which is the Curl_ssl struct used before an SSL library is set. In other words if no SSL library was set then when libcurl calls connect_nonblocking it maps to multissl_connect_nonblocking which initializes multissl, then calls the actual multissl_connect_nonblocking. I just don't see how it's possible that no SSL library would be set by that point.

PR #9591, opened 2022-09-25
autotools: reduce brute-force when detecting recv/send arg list

vszakats:
autotools uses brute-force to detect `recv`/`send`/`select` argument lists, by interating through _all_ argument type combinations on each `./configure` run. This logic exists since 01fa02d0b545e1433dced2430561f8c0c72b74a9 (from 2006) and was a bit later extended with Windows support.

This results in a worst-case number of compile + link cycles as below:
- `recv`: 96
- `send`: 192
- `select`: 60
Total: 348 (the number of curl C source files is 195, for comparison)

Notice that e.g. curl-for-win autotools builds require two `./configure` invocations, doubling these numbers.

`recv` on Windows was especially unlucky because `SOCKET` (the correct choice there) was listed _last_ in one of the outer trial loops. This resulted in lengthy waits while autotools was trying all invalid combinations first, wasting cycles, disk writes and slowing down iteration.

This patch reduces the amount of idle work by reordering the tests in a way to succeed first on a well-known platform such as Windows, and also on non-Windows by testing for POSIX prototypes first, on the assumption that these are the most likely candidates these days. (We do not touch `select`, where the order was already optimal for these platforms.)

For non-Windows, this means to try a return value of `ssize_t` first, then `int`, reordering the buffer argument type to try `void *` first, then `byte *`, and prefer the `const` flavor with `send`. If we are here, also stop testing for `SOCKET` type in non-Windows builds.

After the patch, detection on Windows is instantaneous. It should also be faster on popular platforms such as Linux and BSD-based ones.

If there are known-good variations for other platforms, they can also be fast-tracked like above, given a way to check for that platform inside the autotools logic.

Closes #9591

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -851,11 +851,21 @@ $curl_includes_bsdsocket
   if test "$curl_cv_recv" = "yes"; then
     AC_CACHE_CHECK([types of args and return type for recv],
       [curl_cv_func_recv_args], [
+      # Win32: int     (SOCKET, char *, int   , int)
+      # POSIX: ssize_t (int   , void *, size_t, int)
+      if test "$curl_cv_native_windows" = "yes"; then
+        retv_test1='int'    ; arg1_test1='SOCKET' ; arg2_test1='char *'; arg3_test1='int'
+        retv_test2='ssize_t'; arg1_test2='int'    ; arg2_test2='void *'; arg3_test2='size_t'
+      else
+        retv_test1='ssize_t'; arg1_test1='int'    ; arg2_test1='void *'; arg3_test1='size_t'
+        retv_test2='int'    ; arg1_test2='ssize_t'; arg2_test2='char *'; arg3_test2='int'
+      fi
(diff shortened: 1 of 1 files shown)


bagder (approved):
Nice work!

I have thought about ripping this out completely from cmake/autoconf and instead replace it with a basic `#ifdef` sequence but it hasn't become more than thoughts in my head so  I think this is a step forward!

vszakats:
Thanks! Ripping it out would definitely be the best. Perf is fine now, but there is a significant complexity in the code and build systems around this, most of it probably overkill.

Data is in: The patch shaved off almost 5 minutes from curl-for-win builds. A 10% improvement.

with patch: https://ci.appveyor.com/project/curlorg/curl-for-win/builds/44877232
without: https://ci.appveyor.com/project/curlorg/curl-for-win/builds/44876993

bagder:
I took a first step in #9592. I imagine it might need a while and polish before we can switch.

PR #9598, opened 2022-09-26
windows: adjust name/visibility of two internal public functions

vszakats:
According to `docs/INTERNALS.md`, internal function names spanning source files start with uppercase `Curl_`. Bring these two functions in alignment with this.

This also stops exporting them from `libcurl.dll`.

Closes #xxxx


bagder (approved):
:+1:

PR #9600, opened 2022-09-26
http, vauth: always provide Curl_allow_auth_to_host() functionality

monnerat:
This function is currently located in the lib/http.c module and is therefore disabled by the `CURL_DISABLE_HTTP` conditional token.

As it may be called by TLS backends, disabling HTTP results in an undefined reference error at link time.

Move this function to vauth/vauth.c to always provide it and rename it as `Curl_auth_allowed_to_host`() to respect the vauth module naming convention.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -721,21 +721,6 @@ CURLcode Curl_http_auth_act(struct Curl_easy *data)
   return result;
 }
 
-/*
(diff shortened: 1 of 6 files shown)


bagder (approved):

bagder:
Thanks!

PR #9605, opened 2022-09-27
configure: deprecate builds with small curl_off_t

bagder:
If curl_off_t turns out to be smaller than 8 bytes, --with-n64-deprecated needs to be used to allow the build to continue. This is to highlight the fact that support for such builds is going away next year.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3388,6 +3388,18 @@ AC_CHECK_TYPE(long long,
 )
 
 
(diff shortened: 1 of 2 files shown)


bagder:
I tested this by locally bumping the check to check if it is less than `9` bytes instead of `8` - and made sure the command line switch did its job.

jay (approved):

PR #9610, opened 2022-09-27
lib: sanitize conditional exclusion around MIME

monnerat:
The introduction of `CURL_DISABLE_MIME` came with some additional bugs:
- Disabled MIME is compiled-in anyway if SMTP and/or IMAP is enabled.
- `CURLOPT_MIMEPOST`, `CURLOPT_MIME_OPTIONS` and `CURLOPT_HTTPHEADER` are conditioned on HTTP, although also needed for SMTP and IMAP MIME mail uploads.

In addition, the `CURLOPT_HTTPHEADER` and `--header` documentation does not mention their use for MIME mail.

This commit fixes the problems above.

### DIFF
--- a/docs/libcurl/opts/CURLOPT_HTTPHEADER.3
+++ b/docs/libcurl/opts/CURLOPT_HTTPHEADER.3
@@ -37,6 +37,11 @@ Pass a pointer to a linked list of HTTP headers to pass to the server and/or
 proxy in your HTTP request. The same list can be used for both host and proxy
 requests!
 
+When used within an IMAP or SMTP request to upload a MIME mail, the given
+header list establishes the document-level MIME headers to prepend to the
+uploaded document described by \fICURLOPT_MIMEPOST(3)\fP. This does not affect
+raw mail uploads.
+
 The linked list should be a fully valid list of \fBstruct curl_slist\fP
 structs properly filled in. Use \fIcurl_slist_append(3)\fP to create the list
 and \fIcurl_slist_free_all(3)\fP to clean up an entire list. If you add a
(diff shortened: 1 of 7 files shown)


docs/libcurl/opts/CURLOPT_HTTPHEADER.3
```diff
@@ -98,6 +98,9 @@ implies that if you tell libcurl to follow redirects
 in the subsequent request. Redirects can of course go to other hosts and thus
 those servers will get all the contents of your custom headers too.
 
+This option must also be used to set the top-level headers of a MIME mail when
+uploading the latter via SMTP or IMAP.
+
```
jay, line 103: how is this a security concern

bagder:
I was not completely aware of this. I think the man page needs more massaging to better convey this fact. It is even unfortunate that it is named `HTTPHEADER` when in fact then isn't only for HTTP headers.

Should we consider a new alias for this called `REQUESTHEADERS` or something?

docs/libcurl/opts/CURLOPT_HTTPHEADER.3
monnerat: > how is this a security concern

Thanks for pointing this out: I placed the paragraph in a wrong section.

monnerat:
> Should we consider a new alias for this called REQUESTHEADERS or something?

I think we should. We already did this kind of thing (i.e.: `CURLOPT_FTP_SSL`) but at the time of mime coding, I did not want to make a revolution on existing public symbols. Then it came out of my mind.

IMAP and SMTP mime mails use this option because the mime api we implemented supports setting headers on parts only. Mime mail data need to support document-level headers.

We currently already have the (undocumented) `CURLOPT_RTSPHEADERS` alias. As the use in the mime case is not related to the request and is multi-protocol, I think a good alias would be `CURLOPT_MAILHEADERS`. I tried to imagine a more generic name but discarded the use of simply `CURLOPT_HEADERS` because it might be confused with the existing `CURLOPT_HEADER` option. Maybe you have a better idea?

bagder:
(It is of course a bit off topic for this PR)

I think i would prefer a new name to use for the option globally rather than to continue to invent special protocol aliases.

`CURLOPT_SENDHEADERS` ?
`CURLOPT_MYHEADERS` ?

An obvious downside with a name change (even when the old name is still supported) is that this is a very commonly used option so there exist a bazillion examples of its use on stackoverflow and other places online, and they will look worse if we change the name... That could be an argument for just adding an alias like `CURLOPT_MAILHEADERS`.

monnerat:
Documentation updated.  `CURLOPT_HTTPHEADER` **not** renamed.

bagder (approved):

bagder:
Thanks!

PR #9616, opened 2022-09-28
Makefile.m32: various cleanups [ci skip]

vszakats:
- Add `-lcrypt32` just once, and add it always for simplicity.
- Delete broken link and reference to the pre-Vista WinIDN add-on. MS no longer distribute it.
- Delete related `WINIDN_PATH` option. IDN is a system lib since Vista.
- Sync `LIBCARES_PATH` default with the rest of dependencies.
- Delete version numbers from dependency path defaults.
- `libgsasl` package is now called `gsasl`.
- Delete `libexpat` and `libxml2` references. No longer used by curl.
- Delete `Edit the path below...` comments. We advise to predefine those envvars instead.
- `libcares.a` is not an internal dependency anymore. Stop using it as such.
- `windres` `--include-dir` → `-I`, `-F` → `--target=` for readability.
- Delete `STRIP`, `CURL_STRIP`, `AR` references from `src/Makefile.m32`. They were defined, but never used.
- Stop to `clean` some objects twice in `src/Makefile.m32`.
- Delete cvs-specific leftovers.
- Finish resource support in examples make file.
- Delete `-I<root>/lib` from examples make file.
- Fix copyright start year in examples make file.
- Delete duplicate `ftpuploadresume` input in examples make file.
- Sync OpenSSL lib order, `SYNC` support, `PROOT` use, dependency path defaults, variables names and other internal bits between the three make files.
- `lib/Makefile.m32` accepted custom options via `DLL_LIBS` envvar. This was lib-specific and possibly accidental. Use `CURL_LDFLAG_EXTRAS_DLL` envvar for the same effect.
- Fix linking `curl.exe` and examples to wrong static libs with auto-detected OpenSSL 1.0.2 or earlier.
- Add `-lgdi32` for OpenSSL 1.0.2 and earlier only.
- Add link to Novell LDAP SDK and use a relative default path. Latest version is from 2016, linked to an outdated OpenSSL 1.0.1.
- Whitespace and comment cleanups.

Closes #9616

---
TODO:
- Delete built-in detection/logic for OpenSSL 1.0.2 and earlier, the Novell LDAP SDK and the other LDAP SDK (which is _not_ OpenLDAP). Write up the necessary custom envvars to configure them.


jay:
What happens if we always include -lcrypt32, is there some reason not to do that?

vszakats:
@jay: Assuming a crypt32 implib exists, nothing would happen. If it doesn't, link would fail for all builds. With the current patch, it would only fail with Schannel or OpenSSL enabled (which kinda is also _always_). Can't tell off-hand when crypt32 may be missing, but both cmake and autotools make sure to check for it. Maybe to support pre-XP envs?

Anyway, it's a good idea to just add it always.

jay:
Ok. I tried an old version of mingw and -lcrypt32 works there. I'm not sure what other build environments m32 makefiles are meant to support.

vszakats:
If we're a this, `gdi32` is no longer necessary with OpenSSL 1.1.x and up. For older versions it can be added via a custom option. I tend to think we should remove it by default. What do you think?

PR #9619, opened 2022-09-29
GHA: build tests in a separate step from the running of them

bagder:
... to make the output smaller for when you want to look at test failures.

### DIFF
--- a/.github/workflows/msh3.yml
+++ b/.github/workflows/msh3.yml
@@ -64,10 +64,10 @@ jobs:
     - run: make V=1
       name: 'make'
 
-    - run: make V=1 examples
-      name: 'make examples'
+    - run: make V=1 -C tests
+      name: 'make tests'
 
     - run: make V=1 test-ci
-      name: 'test'
+      name: 'run tests'
       env:
         TFLAGS: "${{ matrix.build.tflags }}"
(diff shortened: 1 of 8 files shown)


.github/workflows/msh3.yml
```diff
@@ -64,10 +64,10 @@ jobs:
     - run: make V=1
       name: 'make'
 
-    - run: make V=1 examples
-      name: 'make examples'
+    - run: make V=1 -C tests
```
mback2k, line 67: Why skip building examples here?

.github/workflows/torture.yml
```diff
@@ -54,7 +54,10 @@ jobs:
     - run: make V=1
       name: 'make'
 
-    - run: make V=1 test-torture
-      name: 'test'
+    - run: make V=1 -C tests
+      name: 'make tests'
+
+    - run: make V=1 test-ci
+      name: 'run tests'
```
mback2k, line 61: Isn't the purpose of this workflow to just run torture tests?

.github/workflows/msh3.yml
bagder: It was not intentional to do that change, but then again we could probably just as well skip them in this build as they are built in many others.

bagder: maybe we should consider only building the example in a single specific example job and remove the building of them from all others?

mback2k: Good idea, but we should at least cover all platforms and architectures as we had issues on some in the past with the examples.

PR #9623, opened 2022-09-29
docs: Update Caddy example in HTTP3.md

mholt:
Caddy 2.6 enables HTTP/3 by default and is no longer experimental. This config yields a more useful test endpoint related to protocols and also retains the example of reverse proxying if desired.

Since this example uses `caddy start` do you think we should explain to use `caddy stop` to stop the process? Or maybe an example `curl` request to see the result?

### DIFF
--- a/docs/HTTP3.md
+++ b/docs/HTTP3.md
@@ -291,26 +291,21 @@ that exists in curl's test dir.
 
 ### Caddy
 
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #9626, opened 2022-09-30
feat.h: defines what features to enable

bagder:
All C files now check for the *presence* of features with `FEAT` instead of the former often used `CURL_DISABLE` that checked for disabled features. This simplifies #ifs and #ifdefs in numerous places.

`CURL_DISABLE` defines can still be set by config-*.h files or by configure/cmake.

feat.h defines what fatures to build, respecting the will of any `CURL_DISABLE` defines set.

**DRAFT** to get a sense for how it works and feels.

### DIFF
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -267,6 +267,7 @@ LIB_HFILES =         \
   easyif.h           \
   easyoptions.h      \
   escape.h           \
+  feat.h             \
   file.h             \
   fileinfo.h         \
   fopen.h            \
--- a/lib/altsvc.c
+++ b/lib/altsvc.c
@@ -27,7 +27,7 @@
  */
 #include "curl_setup.h"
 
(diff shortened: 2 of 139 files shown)


jay:
This doesn't look like a simplification unless I missed it, just another set of defines based on the existing ones. Also why didn't you use the CURL namespace? like CURL_FEAT_? Is this that you don't like the negation of CURL_DISABLE_?

bagder:
Well, it's not done yet. But...

- Why would we need a curl namespace? They're not public symbols. Shorter makes easier to read.
- The idea is to reverse the defines - to stick to defining features and protocols we want instead of the opposite -  so that `#ifdefs` in the code in general become easier to read and write.
- Ideally we get rid of the `CURL_DISABLE_` ones completely, but I haven't yet figured out the nicest way to do that.

bagder:
Not quite happy with this PR as it is right now and I think I don't have enough bandwidth to take it further right now. Closing and regrouping until later.

PR #9565, opened 2022-09-21
CI: macOS w/ m1 for curl build and testing

pheiduck:

PR #9627, opened 2022-09-30
CI: add some more jobs from macOS x86_64 for macOS arm64

pheiduck:
continue: #9565

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -135,9 +135,19 @@ windows_task:
     %container_cmd% -l -c "cd $(echo '%cd%') && %make_cmd% TFLAGS='!IDN !SCP ~612 ~1056 %tests%' test-ci"
 
 macos_task:
-  name: macOS arm64 normal
+  name: macOS arm64
   macos_instance:
     image: ghcr.io/cirruslabs/macos-ventura-xcode:latest
+
+  matrix:
+    - name: macOS arm64 normal
+      env:
+        install_packages: nghttp2
+        configure: --without-ssl
+    - name: macOS arm64 debug
+      env:
+        install_packages: nghttp2
+        configure: --without-ssl --enable-debug
 
   env:
     CIRRUS_CLONE_DEPTH: 10
@@ -145,36 +155,13 @@ macos_task:
     CFLAGS: "-Wno-vla -mmacosx-version-min=10.9"
 
   pkginstall_script:
-    - brew update
-    - echo libtool autoconf automake pkg-config | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
+    - echo libtool autoconf automake pkg-config ${install_packages} | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
     - brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done
+    - sudo -H python3 -m pip install --upgrade pip
+    - sudo -H python3 -m pip install impacket
   configure_script:
     - autoreconf -fi
-    - ./configure --enable-warnings --enable-werror --without-ssl --enable-websockets
-  compile_script:
-    - make V=1 && make V=1 examples && cd tests && make V=1
-  test_script:
-    - sudo make V=1 test-ci
-  install_script:
-    - sudo make V=1 install
-
-macos_task:
-  name: macOS arm64 debug
-  macos_instance:
-    image: ghcr.io/cirruslabs/macos-ventura-xcode:latest
-
-  env:
-    CIRRUS_CLONE_DEPTH: 10
-    MAKE_FLAGS: -j 4
-    CFLAGS: "-Wno-vla -mmacosx-version-min=10.9"
-
-  pkginstall_script:
-    - brew update
-    - echo libtool autoconf automake pkg-config nghttp2 | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
-    - brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done
-  configure_script:
-    - autoreconf -fi
-    - ./configure --enable-warnings --enable-werror --enable-debug --without-ssl --enable-websockets
+    - ./configure --enable-warnings --enable-werror --enable-websockets ${configure}
   compile_script:
     - make V=1 && make V=1 examples && cd tests && make V=1
   test_script:


pheiduck:
the cmake gcc builds are failing because of an error which was fixed here: https://github.com/curl/curl/commit/935b1bd4544a23a91d68ffb9f86983e92747e9a5

pheiduck:
@bagder @mback2k can be reviewed I skip the cmake gcc builds here because the failed to often.

mback2k:
Please rework this into using a matrix build to avoid the huge amount of repetition.

pheiduck:
> Please rework this into using a matrix build to avoid the huge amount of repetition.

I hope it still builds :crossed_fingers:

pheiduck:
@mback2k The builds are failing what is wrong?

mback2k:
This is YAML, so if you want to put a multi-line string somewhere you have to use a modifier after the colon `:`, e.g. `: |`, see:
https://yaml-multiline.info/

pheiduck:
> This is YAML, so if you want to put a multi-line string somewhere you have to use a modifier after the colon `:`, e.g. `: |`, see: https://yaml-multiline.info/

hm, anyway still failing

pheiduck:
@fkorotkov could you help out here?

.cirrus.yml
```diff
@@ -133,49 +133,97 @@ windows_task:
...
 
 macos_task:
-  name: macOS arm64 normal
+  name: macOS arm64
   macos_instance:
-    image: ghcr.io/cirruslabs/macos-ventura-xcode:latest
-
+    image: ${container_img}
+    pkgs: ${prepare}
+    config: ${configure}
```
fkorotkov, line 140: These two lines seems will be ignored.

.cirrus.yml
```diff
@@ -133,49 +133,97 @@ windows_task:
...
+    - name: macOS arm64 torture-ftp
+      env:
+        container_img: ghcr.io/cirruslabs/macos-ventura-xcode:latest
+        prepare: echo libtool autoconf automake pkg-config nghttp2 openssl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
+        configure: --enable-warnings --enable-werror --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=$(brew --prefix openssl) --enable-websockets
+
+  prepare_script: |
+    ${pkgs} && brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done && sudo -H python3 -m pip install --upgrade pip && sudo -H python3 -m pip install impacket
+  configure_script: |
+    autoreconf -fi && ./configure ${config}
```
fkorotkov, line 223: I guess you want something like this:

```suggestion
  prepare_script: |
    ${prepare} && brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done && sudo -H python3 -m pip install --upgrade pip && sudo -H python3 -m pip install impacket
  configure_script: |
    autoreconf -fi && ./configure ${configure}
```

pheiduck:
@fkorotkov @mback2k the Problem is that the Brewfile is empty and on the build config variable is "the same problem"

pheiduck:
`Error while parsing tasks: INVALID_ARGUMENT: yaml: line 154: mapping values are not allowed in this context!`

Not sure why this is showing

dfandrich:
There's a no-break space between prepare: and | on line 152.

I find forcing CI YAML files through a YAML pretty-printer before submission useful to catch syntax issues. A one-liner like this works great for me:

    python3 -c 'import yaml,sys,pprint; print(pprint.PrettyPrinter(indent=4).pprint(yaml.load(sys.stdin,Loader=yaml.SafeLoader)))' < .cirrus.yml

(I posted this one-liner yesterday by e-mail, but it seems Github ate it).

pheiduck:
> There's a no-break space between prepare: and | on line 152.
> 
> I find forcing CI YAML files through a YAML pretty-printer before submission useful to catch syntax issues. A one-liner like this works great for me:
> 
> ```
> python3 -c 'import yaml,sys,pprint; print(pprint.PrettyPrinter(indent=4).pprint(yaml.load(sys.stdin,Loader=yaml.SafeLoader)))' < .cirrus.yml
> ```
> 
> (I posted this one-liner yesterday by e-mail, but it seems Github ate it).

Doesn't solve it anyway, maybe we need to use it without matrixbuilds

pheiduck:
I revert it back to the first commit have to revise it again later.

mback2k:
Suggestion: I will make a matrix build out of the two jobs already there for macOS and then you can add the additional ones on top of that.

mback2k:
I can take a look later, I guess you are escaping things wrong.

.cirrus.yml
```diff
@@ -133,46 +133,81 @@ windows_task:
...
     MAKE_FLAGS: -j 4
-    CFLAGS: "-Wno-vla -mmacosx-version-min=10.9"
+    CFLAGS: "-Wno-vla -mmacosx-version-min=10.8"
 
   pkginstall_script:
-    - brew update
-    - echo libtool autoconf automake pkg-config nghttp2 | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
+    - echo libtool autoconf automake pkg-config ${install_packages} | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
     - brew update && for i in 1 2 3; do brew bundle install --no-lock --file /tmp/Brewfile && break || sleep 1; done
+    - export PKG_CONFIG_PATH=${prefix}
```
mback2k, line 205: Try `$(eval echo $prefix)` instead of just `${prefix}`

pheiduck: :/ doesn't work

mback2k:
I guess the variable export does not survive till the next step.

pheiduck:
> I guess the variable export does not survive till the next step.

Another solution is to use `brew link pkg --force` it will create symlinks in `/usr/local`

mback2k:
Your latest commit has mixed/invalid indentation.

mback2k:
I would recommend to add an install step like this:

for package in ${link_packages}; do brew link ${package}; done

pheiduck:
> I would recommend to add an install step like this:
> 
> for package in linkpackages;dobrewlink{package} --force; done
> 
> This will do nothing if link_packages is empty or undefined.

My last 2 commit have solved the problem 🎉

mback2k:
Generally LGTM, but please rebase onto master instead of merging from it and then squash your changes into one commit.

pheiduck:
Wasn't that what I wound reopen https://github.com/curl/curl/pull/9676

pheiduck:
Had to reopened it https://github.com/curl/curl/pull/9677 because it closed due a bad rebasing

dfandrich:
I find forcing CI YAML files through a YAML pretty-printer before submission useful to catch syntax issues. A one-liner like this works great for me:

    python3 -c 'import yaml,sys,pprint; print(pprint.PrettyPrinter(indent=4).pprint(yaml.load(sys.stdin, Loader=yaml.SafeLoader)))' < appveyor.yml

PR #9630, opened 2022-10-01
docs: update sourceforge project links [ci skip]

vszakats:
PROJECT.sourceforge.net addresses support HTTPS now, seemingly by default. [^1]

Until recently [^3][^4], equivalent .io addresses were offering HTTPS, with a redirect to the HTTP .net site for projects that did not opt-in to HTTPS and staying on .io for those that did. PROJECT.sourceforge.net did not offer HTTPS at all.

Now, .io addresses became a project-selectable option [^2], so _some_ of them perm-redirect to the .net URL, but with _HTTP_, even though all those got HTTPS enabled. Then HTTP gets perm-redirected back to HTTPS in a next step. Possible deployment glitch? Other project stay on the .io address. Needs to be detected on a per-project basis.

FWIW, both .io and .net are served through Cloudflare.

The step back from .io to .net for user content is a little bit puzzling, as well as the insecure redirect between them and the ambiguity in deciding which flavor is the canonical one (= without a redirect).

This patch converts all .io addresses (back) to .net, so now they are all accessible via HTTPS, without a redirect.

[^1]: https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/#changes-as-of-september-2022
[^2]: https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/diff?v1=57&v2=56
[^3]: https://sourceforge.net/p/forge/documentation/Convert%20your%20website%20to%20HTTPS/
[^4]: https://sourceforge.net/blog/introducing-https-for-project-websites/

### DIFF
--- a/docs/BINDINGS.md
+++ b/docs/BINDINGS.md
@@ -20,7 +20,7 @@ C++: [curlpp](https://github.com/jpbarrette/curlpp/) Written by Jean-Philippe Ba
 [curlcpp](https://github.com/JosephP91/curlcpp) by Giuseppe Persico and [C++
 Requests](https://github.com/libcpr/cpr) by Huu Nguyen
 
-[Ch](https://chcurl.sourceforge.io/) Written by Stephen Nestinger and Jonathan Rogado
+[Ch](https://chcurl.sourceforge.net/) Written by Stephen Nestinger and Jonathan Rogado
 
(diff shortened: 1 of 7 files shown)


vszakats:
UPDATE: .io → .net redirection may only happen for projects that did not opt-in to HTTPS before. E.g. this .io URL is not being redirected at the time of writing this: https://libpng.sourceforge.io/

PR #9631, opened 2022-10-01
Makefile.m32: delete legacy component bits [ci skip]

vszakats:
- Drop auto-detection of OpenSSL 1.0.2 and earlier. Now defaulting to OpenSSL 1.1.0 and later, LibreSSL and BoringSSL.

- Drop `Invalid path to OpenSSL package` detection. OpenSSL has been using a standard file layout since 1.1.0, so this seems unnecessary now.

- Drop special logic to enable Novell LDAP SDK support.

- Drop special logic to enable OpenLDAP LDAP SDK support. This seems to be distinct from native OpenLDAP, with support implemented inside `lib/ldap.c` (vs. `lib/openldap.c`) back when the latter did not exist yet in curl.

- Add `-lwldap32` only if there is no other LDAP library (either native OpenLDAP, or SDKs above) present.

- Update `doc/INSTALL.md` accordingly.

After this patch, it's necessary to make configration changes when using OpenSSL 1.0.2 or earlier, or the two LDAP SDKs.

OpenSSL 1.0.2 and earlier:
```
export OPENSSL_INCLUDE = <path-to-openssl>/outinc
export OPENSSL_LIBPATH = <path-to-openssl>/out
export OPENSSL_LIBS = -lssl32 -leay32 -lgdi32
```

Novell LDAP SDK, previously enabled via `USE_LDAP_NOVELL=1`:
```
export CURL_CFLAG_EXTRAS = -I<path-to-sdk>/inc -DCURL_HAS_NOVELL_LDAPSDK
export CURL_LDFLAG_EXTRAS = -L<path-to-sdk>/lib/mscvc -lldapsdk -lldapssl -lldapx
```

OpenLDAP LDAP SDK, previously enabled via `USE_LDAP_OPENLDAP=1`:
```
export CURL_CFLAG_EXTRAS = -I<path-to-sdk>/include -DCURL_HAS_OPENLDAP_LDAPSDK
export CURL_LDFLAG_EXTRAS = -L<path-to-sdk>/lib -lldap -llber
```

I haven't tested these scenarios, and in general we recommend using a recent OpenSSL release. Also, WinLDAP (the Windows default) and OpenLDAP (via `-DUSE_OPENLDAP`) are the LDAP options actively worked on in curl.

Closes #9631

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -172,9 +172,9 @@ adjust as necessary. It is also possible to override these paths with
 environment variables, for example:
 
 ```cmd
-set ZLIB_PATH=c:\zlib-1.2.8
-set OPENSSL_PATH=c:\openssl-1.0.2c
-set LIBSSH2_PATH=c:\libssh2-1.6.0
+set ZLIB_PATH=c:\zlib-1.2.12
+set OPENSSL_PATH=c:\openssl-3.0.5
+set LIBSSH2_PATH=c:\libssh2-1.10.0
 ```
 
 It is also possible to build with other LDAP installations than MS LDAP;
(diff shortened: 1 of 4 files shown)


vszakats:
It seems these old LDAP options were not touched since 2007. They were used in NetWare and some other deleted makefiles. I wonder if these are still being in active use these days?

Issue #7475, opened 2021-07-22
7.78.0 breaks static linking


moparisthebest:
[static-curl](https://github.com/moparisthebest/static-curl) started spitting out dynamically linked instead of static executables starting with 7.78.0, the script has not changed, and I verified 7.77.0 in the same alpine container produces a static executable as expected.  I ran a git bisect using bisect.sh which is pasted at the end of this issue, with instructions, which pointed to 265b14d6b37c4298bd5556fabcbc37d36f911693 as the first bad commit.  I further determined that `git checkout 9cf516adc67b67388e22bb8c33cb3fabd68f2ac6 -- configure.ac 'src/Makefile.am'` makes static linking work at both master and 265b14d6b37c4298bd5556fabcbc37d36f911693 , I believe the exact culprit is [this change to src/Makefile.am](https://github.com/curl/curl/commit/265b14d6b37c4298bd5556fabcbc37d36f911693#diff-4cb884d03ebb901069e4ee5de5d02538c40dd9b39919c615d8eaa9d364bbbd77L69) but I'm not versed enough in autotools to figure out exactly why.

bisect.sh:
```sh
#!/bin/sh

# run these in order, in a directory with this script:
# docker run --rm -v $(pwd):/tmp -w /tmp -it alpine sh
# apk add git autoconf automake libtool build-base clang
# git clone https://github.com/curl/curl
# cd curl
# git bisect start curl-7_78_0 curl-7_77_0 --
# git bisect run /tmp/bisect.sh
# last good: 9cf516adc67b67388e22bb8c33cb3fabd68f2ac6
# first bad: 265b14d6b37c4298bd5556fabcbc37d36f911693

set -exu

export CC=clang

git clean -dxf
autoreconf -fi

# set up any required curl options here
LDFLAGS="-static" PKG_CONFIG="pkg-config --static" ./configure --disable-shared --enable-static --disable-libcurl-option --without-brotli --disable-manual --disable-unix-sockets --disable-dict --disable-file --disable-gopher --disable-imap --disable-smtp --disable-rtsp --disable-telnet --disable-tftp --disable-pop3 --without-zlib --disable-threaded-resolver --disable-ipv6 --disable-smb --disable-ntlm-wb --disable-tls-srp --disable-crypto-auth --without-ngtcp2 --without-nghttp2 --disable-ftp --disable-mqtt --disable-alt-svc --disable-largefile  --without-ssl

make -j4 V=1 curl_LDFLAGS=-all-static

[...]

PR #9632, opened 2022-10-02
Makefile.m32: major rework [ci skip]

vszakats:
This patch overhauls `Makefile.m32` scripts, fixing a list of quirks, making its behaviour and customization envvars align better with other build systems, aiming for less code, that is easier to read, use and maintain.

Details:
- Rename customization envvars:
  `CURL_CC` -> `CC`
  `CURL_RC` -> `RC`
  `CURL_AR` -> `AR`
  `CURL_LDFLAG_EXTRAS_DLL` -> `CURL_LDFLAGS_LIB`
  `CURL_LDFLAG_EXTRAS_EXE` -> `CURL_LDFLAGS_BIN`
- Drop `CURL_STRIP` and `CURL_RANLIB`. These tools are no longer used.
- Accept `CFLAGS`, `CPPFLAGS`, `RCFLAGS`, `LDFLAGS` and `LIBS` envvars.
- Drop `CURL_CFLAG_EXTRAS`, `CURL_LDFLAG_EXTRAS`, `CURL_RCFLAG_EXTRAS` in
  favor of the above.
- Do not automatically enable `zlib` with `libssh2`. `zlib` is optional
  with `libssh2`.
- Omit unnecessary `CPPFLAGS` options when building `curl.exe` and
  examples.
- Drop support for deprecated `-winssl` `CFG` option. Use `-schannel`
  instead.
- Avoid late evaluation where not necessary (`=` -> `:=`).
- Drop support for `CURL_DLL_A_SUFFIX` to override the implib suffix.
  Instead, use the standard naming scheme by default: `libcurl.dll.a`.
  The toolchain recognizes the name, and selects it automatically when
  asking for a `-shared` vs. `-static` build.
- Stop applying `strip` to `libcurl.a`. Follow-up from
  16a58e9f93c7e89e1f87720199388bcfcfa148a4. There was no debug info to
  strip since then.
- Stop setting `-O3`, `-W`, `-Wall` options. You can add these to
  `CFLAGS` as desired.
- Always enable `-DCURL_DISABLE_OPENSSL_AUTO_LOAD_CONFIG` with OpenSSL,
  to avoid that vulnerability on Windows.
- Add `-lbrotlicommon` to `LIBS` when using `brotli`.
- Do not enable `-nghttp3` without `-ngtcp2`.
- `-ssh2` and `-rtmp` options no longer try to auto-select a TLS-backend.
  You need to set the backend explicitly. This scales better and avoids
  issues with certain combinations (e.g. `libssh2` + `wolfssl` with no
  `schannel`).
- Default to OpenSSL TLS-backend with `ngtcp2`. Possible to override via
  `NGTCP2_LIBS`.
- Old, alternate method of enabling components (e.g. `SSH2=1`) no longer
  supported.
- Delete `SPNEGO` references. They were no-ops.
- Drop support for Win9x environments.
- Allow setting `OPENSSL_LIBS` independently from `OPENSSL_LIBPATH`.
- Support autotools/CMake `libssh2` builds by default.
- Respect `CURL_DLL_SUFFIX` in `-dyn` mode when building `curl.exe` and
  examples.
- Assume standard directory layout with `LIBCARES_PATH`. (Instead of the
  long gone embedded one.)
- Stop static linking with c-ares by default. Add
  `CPPFLAGS=-DCARES_STATICLIB` to enable it.
- Reorganize internal layout to avoid redundancy and emit clean diffs
  between src/lib and example make files.
- Delete unused variables.
- Code cleanups/rework.
- Comment and indentation fixes.

Closes #9632


jay:
Do you know of the reason curl_LDADD was used instead of LDFLAGS? It seems like that was an intentional choice but I don't know what advantage it would have.

vszakats:
I don't know the history of it, but the name seems arbitrary. It has no special meaning AFAIK and I don't recognize it as a naming convention either. It was internal-only, so any name could have worked.

For the `lib` makefile `DLL_LIBS` was used for the same purpose. That one was not initialized (accidentally or not), so DLL-only LDFLAGS could be passed via such envvar.

jay:
Somewhat or maybe not related, while looking into this I see autotools had curl_LDFLAGS which was removed in favor of passing LDFLAGS to make in order to not confuse configure. #7475 has some example of how it was used. #9633 finishes it off unless I did it wrong, I can't tell if it's supposed to be -static and -all-static or just the latter

PR #9636, opened 2022-10-03
websockets: remodeled API to support 63 bit frame sizes

bagder:
curl_ws_recv() now receives data to fill up the provided buffer, but can return a partial fragment. The function now also get a pointer to a curl_ws_frame struct with metadata that also mentions the offset and total size of the fragment (of which you might be receiving a smaller piece). This way, large incoming fragments will be "streamed" to the application. When the curl_ws_frame struct field 'bytesleft' is 0, the final fragment piece has been delivered.

curl_ws_recv() was also adjusted to work with a buffer size smaller than the fragment size. (Possibly needless to say as the fragment size can now be 63 bit large).

curl_ws_send() now supports sending a piece of a fragment, in a streaming manner, in addition to sending the entire fragment in a single call if it is small enough. To send a huge fragment, curl_ws_send() can be used to send it in many small calls by first telling libcurl about the total expected fragment size, and then send the payload in N number of separate invokes and libcurl will stream those over the wire.

The struct curl_ws_meta() returns is now called 'curl_ws_frame' and it has been extended with two new fields: *offset* and *bytesleft*. To help describe the passed on data chunk when a fragment is delivered in many smaller pieces.

The documentation has been updated accordingly.

### DIFF
--- a/docs/libcurl/curl_ws_meta.3
+++ b/docs/libcurl/curl_ws_meta.3
@@ -60,11 +60,11 @@ This field specify the age of this struct. It is always zero for now.
 This is a bitmask with individual bits set that describes the WebSocket
 data. See the list below.
 .IP offset
-When this frame is a continuation of fragment data aleady delivered, this is
+When this frame is a continuation of fragment data already delivered, this is
 the offset into the final fragment where this piece belongs.
 .IP bytesleft
 If this is not a complete fragment, the \fIbytesleft\fP field informs about
-how many additonal bytes are expected to arrive before this fragment is
+how many additional bytes are expected to arrive before this fragment is
 complete.
 .SH FLAGS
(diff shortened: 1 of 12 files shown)


bagder:
Any objections to me merging this soon?

jay:
> Any objections to me merging this soon?

I haven't been following this but if you are changing the API I would wait until the next feature window.

bagder:
> I haven't been following this but if you are changing the API I would wait until the next feature window.

Really? That just delays things more, as then people will get the old API in the next release when that is already history.

We do have it experimental to be able to change things (including the API), and I've been informed in clear terms that the limitations of the current API makes it hard to use.

jay:
Ok. Those are all good points.

bagder:
Thanks for keeping things in check!

PR #9637, opened 2022-10-03
curl_formadd.3: bolden the deprecation notice

jay:
Ref: https://github.com/curl/curl/pull/9621

Closes #xxxx

---

/cc @monnerat

### DIFF
--- a/docs/libcurl/opts/CURLOPT_HTTPPOST.3
+++ b/docs/libcurl/opts/CURLOPT_HTTPPOST.3
@@ -32,6 +32,8 @@ CURLOPT_HTTPPOST \- multipart formpost content
 CURLcode curl_easy_setopt(CURL *handle, CURLOPT_HTTPPOST,
                           struct curl_httppost *formpost);
 .SH DESCRIPTION
+\fBThis function is deprecated.\fP Use \fICURLOPT_MIMEPOST(3)\fP instead.
+
 Tells libcurl you want a \fBmultipart/formdata\fP HTTP POST to be made and you
 instruct what data to pass on to the server in the \fIformpost\fP argument.
 Pass a pointer to a linked list of \fIcurl_httppost\fP structs as parameter.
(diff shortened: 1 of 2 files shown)


monnerat:
Thanks, this goes in the good direction. Maybe do the same boldening in `CURLOPT_HTTPPOST.3` too ?

bagder (approved):

bagder:
And move the note to the top of the description I think.

jay:
Done, please give it another look.

IMO[ CURLOPT_MIMEPOST](https://curl.se/libcurl/c/CURLOPT_MIMEPOST.html) should have an inline HTTP example instead of just referring to the SMTP MIME example code. The example could be copied from libcurl-tutorial,

https://github.com/curl/curl/blob/93d092867f0f2c78571983040ef75e078ee1a4c4/docs/libcurl/libcurl-tutorial.3#L496-L518

bagder (approved):

docs/libcurl/opts/CURLOPT_HTTPPOST.3
```diff
@@ -32,6 +32,8 @@ CURLOPT_HTTPPOST \- multipart formpost content
 CURLcode curl_easy_setopt(CURL *handle, CURLOPT_HTTPPOST,
                           struct curl_httppost *formpost);
 .SH DESCRIPTION
+\fBThis function is deprecated.\fP Use \fICURLOPT_MIMEPOST(3)\fP instead.
```
monnerat, line 35: Isn't it an option rather than a function ?

bagder: Agreed!

jay: > Isn't it an option rather than a function ?

Oops was a c&p, fixed now

~~~diff
-\fBThis function is deprecated.\fP Use \fICURLOPT_MIMEPOST(3)\fP instead.
+\fBThis option is deprecated.\fP Use \fICURLOPT_MIMEPOST(3)\fP instead.
~~~

PR #9638, opened 2022-10-04
Improve usability of CMake build as a sub-project

jbms:
- Renames `uninstall` -> `curl_uninstall`
- Ensures all export rules are guarded by CURL_ENABLE_EXPORT_TARGET


CMakeLists.txt
```diff
@@ -440,8 +440,12 @@ if(CURL_USE_OPENSSL)
   endif()
 
   set(CMAKE_REQUIRED_INCLUDES ${OPENSSL_INCLUDE_DIR})
-  check_symbol_exists(RAND_egd "${CURL_INCLUDES}" HAVE_RAND_EGD)
-  check_symbol_exists(OPENSSL_IS_BORINGSSL "openssl/base.h" HAVE_BORINGSSL)
+  if(NOT (HAVE_RAND_EGD))
```
emanuele6, line 443: Shouldn't this be `if(NOT (DEFINED HAVE_RAND_EGD))`?
```suggestion
  if(NOT (DEFINED HAVE_RAND_EGD))
```

CMakeLists.txt
```diff
@@ -440,8 +440,12 @@ if(CURL_USE_OPENSSL)
...
+    check_symbol_exists(RAND_egd "${CURL_INCLUDES}" HAVE_RAND_EGD)
+  endif()
+  if(NOT (DEFINED HAVE_BORINGSSL))
```
emanuele6, line 446: All the other cmake checks for HAVE_\* options use `if()`/`if(NOT)` without `DEFINED`. I think we should not introduce this inconsistency.

CMakeLists.txt
jbms: Thanks for catching this, fixed.

CMakeLists.txt
jbms: I added an explanatory comment at the top.  I am checking for DEFINED rather than for the boolean value, because someone could define it to OFF, and in that case we still want to skip the symbol detection.

bagder:
Thanks!

PR #9643, opened 2022-10-04
tool: avoid generating ambiguous escaped characters in --libcurl

monnerat:
C string hexadecimal-escaped characters may have more than 2 digits. This results in a wrong C compiler interpretation of a 2-digit escaped character when followed by an hex digit character.

The solution retained here is to represent such characters as 3-digit octal escapes.

Adjust and extend test 1465 for this case.


bagder (approved):

bagder:
Thanks!

PR #9644, opened 2022-10-04
Fix spelling

jbampton:


lib/urlapi.c
```diff
@@ -781,7 +781,7 @@ static CURLUcode decode_host(struct dynbuf *host)
  * The function handles a query part ('?' + stuff) appended but it expects
  * that fragments ('#' + stuff) have already been cut off.
  *
- * Note that this funciton *writes* a byte into the source buffer during its
+ * Note that this function *writes* a byte into the source buffer during its
```
bagder, line 784: Actually, that sentence is wrong since f703cf971ccd so an even better fix it to remove it completely!

bagder (approved):

bagder:
Thanks!

PR #9646, opened 2022-10-04
CI/GHA: merge msh3 and openssl3 builds into linux workflow

mback2k:
Continue work on merging all Linux workflows into one file.

Follow up to #9501


mback2k:
Relevant workflow succeeded: https://github.com/curl/curl/actions/runs/3185259091

PR #9653, opened 2022-10-05
tool: reorganize function c_escape around a dynbuf

monnerat:
This is a bit shorter and a lot safer.

Substrings of unescaped characters are added by a single call to reduce overhead.

Extend test 1465 to handle more kind of escapes.


bagder (approved):
Me like!

PR #9658, opened 2022-10-06
strcase: add and use Curl_timestrcmp

bagder:
This is a strcmp() alternative function for comparing "secrets", designed to take the same time no matter the content to not leak match/non-match info to observers based on how fast it is.

The time this function takes is only a function of the shortest input string.

Reported-by: Trail of Bits

### DIFF
--- a/lib/netrc.c
+++ b/lib/netrc.c
@@ -198,9 +198,9 @@ static int parsenetrc(const char *host,
           /* we are now parsing sub-keywords concerning "our" host */
           if(state_login) {
             if(specific_login) {
-              state_our_login = !strcmp(login, tok);
+              state_our_login = !Curl_timestrcmp(login, tok);
             }
-            else if(!login || strcmp(login, tok)) {
+            else if(!login || Curl_timestrcmp(login, tok)) {
               if(login_alloc) {
                 free(login);
                 login_alloc = FALSE;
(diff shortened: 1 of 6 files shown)


bagder:
If anyone wants to bikeshed on the function name, please tell me a better one.

dfandrich:
FWIW, I looked into the names for similar functions used by other projects, but didn't get a chance to write this in time to really bikeshed it. But here goes anyway! Of the 24 functions/class names I found, 46% had some variation of "constant time", "const" "ct" in the name, and 50% had "time" or "timing" in the name (there is some overlap between the two). Of the rest, 12% had "crypto", "secure" or "slow" in the name (one sample for each), and 25% had completely generic names.

So, `Curl_timestrcmp` would fall into one of the more popular categories, but `Curl_consttimestrcmp` would fit into the even more popular category of names. But, the current name shouldn't be very surprising to someone new.

For completeness and posterity, here are the names I found:

```areEqual
ConstantTimeCompare
constant_time_compare
constant_time_eq
constant-time-equal
constantTimeEquals
constantTimeMemcmp
constEq
crypto-equality
ct_compare
fixed_time_eq
isEqual
mem_neq
osmo_constant_time_cmp
secure_compare
slowEquals
Str_compareConstantTime
strctcmp
streql
str_equals_timing_almost_safe
str_iseq
TimeConstantComparer
timingSafeEqual
VerifyBufsEqual
```

PR #9659, opened 2022-10-06
cookie: reject cookie names or content with TAB characters

bagder:
TABs in name and content seem allowed by RFC 6265: "the algorithm strips leading and trailing whitespace from the cookie name and value (but maintains internal whitespace)"
    
Cookies with TABs in the names are rejected by Firefox and Chrome.
    
TABs in content are stripped out by Firefox, while Chrome discards the whole cookie.
    
TABs in cookies also cause issues in saved netscape cookie files.

### DIFF
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -538,7 +538,7 @@ Curl_cookie_add(struct Curl_easy *data,
     do {
       /* we have a <what>=<this> pair or a stand-alone word here */
       name[0] = what[0] = 0; /* init the buffers */
-      if(1 <= sscanf(ptr, "%" MAX_NAME_TXT "[^;\r\n=] =%"
+      if(1 <= sscanf(ptr, "%" MAX_NAME_TXT "[^;\t\r\n=] =%"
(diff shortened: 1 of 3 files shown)


danielgustafsson (approved):

bagder:
According to https://github.com/httpwg/http-extensions/issues/2262 the browsers are now adapting to new language in 6265bis (that was not present in 6265) which more than previously encourages implementers to allow TABs in there.

I'm not a fan, but it seems they have made up their minds. In the name of (future) interop, we probably need to keep the support and instead fix the file saving end.

bagder:
The opposite take, that instead encodes the TABs in name and values then saving, is in #9662

bagder:
Decision: we merge this PR: we start rejecting cookies with tab in name or value since they are not accepted widely and thus do not interoperate between clients and they break the cookie file when saved so they do not even work correctly with curl.

If the unlikely event happens and this behavior changes drastically in the future, we can consider going back and then take the #9662 route instead.

PR #9661, opened 2022-10-06
doc: avoid deprecated options in examples where possible

monnerat:
Examples targeting a deprecated feature/option are commented with a warning about it.
Other examples are adapted to not use deprecated features/options.


bagder (approved):

bagder:
Thanks!

PR #9662, opened 2022-10-07
cookie: encode TABs in names and value when saving cookiejar

bagder:
... and decode such cookies on load.

TABs are accepted within names and contents according to RFC 6265 (leading and trailing ones are stripped) but have not been accepted widely by browsers in the past. curl has supported them always.

New wording in the upcoming 6265bis draft has made the browsers decide to (suddenly) accept them going forward. This is a reason for not just discard cookies like this on sight.

The Netscape cookie file format that curl uses to save cookies to, uses a very basic TAB-separated format. Previously curl created a broken file (or downright wrong cookies) when it saved the TABs in names and values raw and as-is in that file.

Starting now, all TABs in names or content will be encoded as '=' when the cookies are saved. When such encoding is done, curl also adds a new column in the file with the keyword 'enc' (for encoded).

The equals character cannot syntactically be a part of the name nor value so old curl version cannot make them appear there. When an encoded cookie jar line is read by an old curl implementation, it will discard encoded cookie lines entirely because there is one field too many on the line.

This encoding is an extension to the original file format, meaning that when exchanging the file with other tools using this format, these cookies will not be understood.

With this change, the cookie load function will of course decode such content and store them "correctly" formatted in memory.

Updated test 46 and 172 to verify.


bagder:
The encoding of the name and values is simple but effective. If anyone has ideas on how to do it better, do let us know.

dfandrich:
From the commit comment, I assumed = encoding meant MIME quoted printable escaping (i.e., tab would be encoded as =09) but on reading the code, it's just a single literal =. Was this chosen because = is not allowed in a cookie name? What is the chance that we will need to encode other characters in the future, meaning using a more extensible form of escaping would be more prudent?

bagder:
> Was this chosen because = is not allowed in a cookie name?

Yes. And because it is a single-letter so I can do the encode/decode without growing/reallocing the string.

> What is the chance that we will need to encode other characters in the future, meaning using a more extensible form of escaping would be more prudent?

I think that is a **low** risk (but of course not non-existing), as this is the first encoding we've added in over twenty years.

I also considered using URL-encoding, but we don't have any convenient function for that so it would mean more allocs and probably more code. I dismissed base64 since the current format keeps the cookies readable, I wanted a format that maintains that to a high degree.

The new format with a new field in the file do allows us to change to other/more/different encodings or other file format changes in a future if we deem it necessary.

But I'm willing to reconsider!

bagder:
We decided to reject such cookies instead, #9659 has been merged. This encoding version can be used if we at a later point decide to revert this decision.

PR #9665, opened 2022-10-07
Minor fixes found using experimental web sockets without CONNECT_ONLY

pseligman:
- Fixed an issue where `is_in_callback` was getting cleared when using web sockets with debug logging enabled
- Ensure the handle is `is_in_callback` when calling out to `fwrite_func`
- Don't send along a masking key for 0-length frames (empty ping)
- Change the `write` vs. `send_data` decision to whether or not the handle is in `CONNECT_ONLY` mode.
- Account for `buflen` not including the header length in `curl_ws_send`

### DIFF
--- a/lib/sendf.c
+++ b/lib/sendf.c
@@ -735,9 +735,10 @@ void Curl_debug(struct Curl_easy *data, curl_infotype type,
     static const char s_infotype[CURLINFO_END][3] = {
       "* ", "< ", "> ", "{ ", "} ", "{ ", "} " };
     if(data->set.fdebug) {
+      bool inCallback = Curl_is_in_callback(data);
       Curl_set_in_callback(data, true);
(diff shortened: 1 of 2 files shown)


bagder:
It looks good. The CI test fail should be fixed with #9666 I think.

pseligman:
> It looks good. The CI test fail should be fixed with #9666 I think.

In retrospect, the previous behavior of sending along 4 mask bytes and 0 payload bytes seems more correct when the client wants to send an empty ping payload.  I've updated the patch to match previous behavior (and unbreak test 2302)

bagder:
Can you please fix the merge conflict and force-push? Sorry for that.

bagder (approved):

bagder:
Thanks!

PR #9667, opened 2022-10-07
lib: feature deprecation warnings in gcc >= 4.3

monnerat:
Add a deprecated attribute to functions and enum values that should not be used anymore.
This uses a gcc 4.3 dialect, thus is only available for this version of gcc and newer. Note that the `_Pragma`() keyword is introduced by C99, but is available as part of the gcc dialect even when compiling in C89 mode.
    
It is still possible to disable deprecation at a calling module compile time by defining `CURL_DISABLE_DEPRECATION`.
    
Gcc type checking macros are made aware of possible deprecations.
    
Some testing support Perl programs are adapted to the extended declaration syntax.
    
Several test and unit test C programs intentionally use deprecated functions/options and are annotated to not generate a warning.

New test 1222 checks the deprecation status in doc and header files.


bagder (approved):

docs/libcurl/symbols-in-versions
```diff
@@ -886,6 +888,7 @@ CURLOPT_WS_OPTIONS              7.86.0
 CURLOPT_XFERINFODATA            7.32.0
 CURLOPT_XFERINFOFUNCTION        7.32.0
 CURLOPT_XOAUTH2_BEARER          7.33.0
+CURLOPTDEPRECATED               7.89.0
```
bagder, line 891: 7.87.0 and merge asap?

monnerat: Yes, done and rebased.
Maybe wait for the CI tests to complete before merge?

bagder: :+1:

monnerat: CI tests complete now. Failed tests seems not related.

bagder:
Thanks!

PR #9672, opened 2022-10-07
CI/cirrus: merge existing macOS jobs into a job matrix

mback2k:
Ref: #9627


mback2k:
What do you think about this @pheiduck ?

pheiduck:
@mback2k LGTM it works, we could continue with it

PR #9676, opened 2022-10-09
CI: add some more jobs from macOS x86_64 for macOS arm64 refactor yml structur for mac_task

pheiduck:
continue: https://github.com/curl/curl/pull/9565


jay:
Please stick with #9627 instead of opening another PR

PR #9677, opened 2022-10-09
CI: add some more jobs from macOS x86_64 for macOS arm64

pheiduck:
Ref/Rebase for: https://github.com/curl/curl/pull/9627
continue: https://github.com/curl/curl/pull/9565

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -184,12 +184,12 @@ macos_task:
       env:
         install_packages: nghttp2 openssl
         configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
-        tflags: -n -t --shallow=25 !FTP
+        TFLAGS: -n -t --shallow=25 !FTP
     - name: macOS arm64 torture-ftp
       env:
         install_packages: nghttp2 openssl
         configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
-        tflags: -n -t --shallow=20 FTP
+        TFLAGS: -n -t --shallow=20 FTP
 
   env:
     CIRRUS_CLONE_DEPTH: 10
@@ -207,6 +207,6 @@ macos_task:
   compile_script:
     - make V=1 && make V=1 examples && cd tests && make V=1
   test_script:
-    - sudo make V=1 test-ci ${tflags}
+    - sudo make V=1 test-ci
   install_script:
     - sudo make V=1 install
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -184,10 +184,12 @@ macos_task:
       env:
         install_packages: nghttp2 openssl
         configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
+        tflags: -n -t --shallow=25 !FTP
     - name: macOS arm64 torture-ftp
       env:
         install_packages: nghttp2 openssl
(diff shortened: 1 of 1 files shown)


pheiduck:
@mback2k not sure where the Windows CI changes came from

pheiduck:
@bagder This can be merged as it works now as expected. work was done here: https://github.com/curl/curl/pull/9627

.cirrus.yml
```diff
@@ -148,11 +148,51 @@ macos_task:
...
+        install_packages: nghttp2 libressl
+        configure: --enable-debug --with-openssl=/opt/homebrew/opt/libressl
+    - name: macOS arm64 torture
+      env:
+        install_packages: nghttp2 openssl
+        configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
+    - name: macOS arm64 torture-ftp
+      env:
+        install_packages: nghttp2 openssl
+        configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
```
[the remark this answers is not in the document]
pheiduck: On x86_64 `torture` and `torture-ftp` also identical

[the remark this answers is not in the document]
mback2k: No, they are not. They use different TFLAGS, see:
https://github.com/curl/curl/blob/master/.github/workflows/macos.yml#L72
and
https://github.com/curl/curl/blob/master/.github/workflows/macos.yml#L77

.cirrus.yml
```diff
@@ -205,6 +207,6 @@ macos_task:
   compile_script:
     - make V=1 && make V=1 examples && cd tests && make V=1
   test_script:
-    - sudo make V=1 test-ci
+    - sudo make V=1 test-ci ${tflags}
```
mback2k, line 210: The `make` command does not take these options as parameter. Instead you need to pass them as environment variables. Since it is an environment variable anyway in the `env:`, it should be enough to uppercase it there and not change the `test_script`.

.cirrus.yml
```diff
@@ -184,10 +184,12 @@ macos_task:
       env:
         install_packages: nghttp2 openssl
         configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
+        tflags: -n -t --shallow=25 !FTP
```
mback2k, line 187: Use uppercase `TFLAGS:`, reasoning see below.

.cirrus.yml
```diff
@@ -184,10 +184,12 @@ macos_task:
...
     - name: macOS arm64 torture-ftp
       env:
         install_packages: nghttp2 openssl
         configure: --enable-debug --disable-shared --disable-threaded-resolver --with-openssl=/opt/homebrew/opt/openssl
+        tflags: -n -t --shallow=20 FTP
```
mback2k, line 192: Use uppercase `TFLAGS:`, reasoning see below.

mback2k (changes requested):

.cirrus.yml
mback2k: Please try to add the parameter `-E` to `sudo` to make it keep environment variables. Right not it seems TFLAGS is lost.

mback2k: This seems to work now.

pheiduck:
@mback2k I think it is ready for the mainline. We could also optimize the `CFLAGS` later.
I am not sure if we need `sudo -E` to compile then.

mback2k (approved):

mback2k:
Merged via 1de60b18de9db237a5e1f0efca6103b6cb6165cc, but unfortunately I put the wrong PR number into that commit. I am sorry about that.

PR #9687, opened 2022-10-10
cmake: enable more detection on Windows

vszakats:
Enable `HAVE_UNISTD_H`, `HAVE_STRTOK_R` and `HAVE_STRCASECMP` detection on Windows, instead of having predefined values.

With these features detected correctly, CMake Windows builds get closer to the autotools and `config-win32.h` ones.

This also fixes detecting `HAVE_FTRUNCATE` correctly, which required `unistd.h`.

Fixing `ftruncate()` in turn causes a build warning/error with legacy MinGW/MSYS1 due to an offset type size mismatch. This env misses to detect `HAVE_FILE_OFFSET_BITS`, which may be a reason. This patch force-disables `HAVE_FTRUNCATE` for this platform.

Reviewed-by: Daniel Stenberg

Closes #9687


bagder (approved):

vszakats:
The good news is that this fix seems to also somehow fix to correctly detect
`HAVE_FTRUNCATE` with modern MinGW.

The bad new is that this in turn breaks legacy MinGW/MSYS1 builds because 
of the enabled `ftruncate()`:
```
[ 86%] Building C object src/CMakeFiles/curl.dir/tool_operhlp.c.obj
C:/projects/curl/src/tool_operate.c: In function 'post_per_transfer':
C:/projects/curl/src/tool_operate.c:576:48: error: conversion from 'curl_off_t' {aka 'long long int'} to 'off_t' {aka 'long int'} may change value [-Werror=conversion]
  576 |         if(ftruncate(fileno(outs->stream), outs->init)) {
      |                                            ~~~~^~~~~~
[ 86%] Building C object src/CMakeFiles/curl.dir/tool_panykey.c.obj
make[2]: *** [src/CMakeFiles/curl.dir/build.make:456: src/CMakeFiles/curl.dir/tool_operate.c.obj] Error 1
[ 86%] Building C object src/CMakeFiles/curl.dir/tool_paramhlp.c.obj
[ 86%] Building C object src/CMakeFiles/curl.dir/tool_parsecfg.c.obj
```

Is there someone with access to this old env with an idea how to fix it?

vszakats:
Updated to keep `ftruncate()` disabled for old MinGW.

PR #9688, opened 2022-10-10
autotools: allow unix sockets on Windows

vszakats:
Fixes: https://github.com/curl/curl-for-win/blob/73a070d96fd906fdee929e2f1f00a9149fb39239/curl-autotools.sh#L44-L47

Closes #9688

Cleaner patch w/o whitespace changes: https://github.com/curl/curl/pull/9688/files?w=1

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3905,17 +3905,24 @@ AS_HELP_STRING([--disable-unix-sockets],[Disable Unix domain sockets]),
        ]
 )
 if test "x$want_unix_sockets" != "xno"; then
-  AC_CHECK_MEMBER([struct sockaddr_un.sun_path], [
+# if test "x$curl_cv_native_windows" = "xyes" &&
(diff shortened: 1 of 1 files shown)


bagder:
> Also: Surely this can be done more elegantly for someone better versed in autotools.

Why does it need this Windows adjustment at all? Is that not merely a problem with an include file? The struct field it checks for is used for Windows as well in `lib/connect.c`.

vszakats:
On native Windows there is never a `sys/un.h` header, so the check fails consistently.

vszakats:
There is a complex detection logic inside `lib/curl_setup.h` already. That could be duplicated inside `configure.am`, but that seems (to me) less elegant and certainly more fragile/maintenance-heavy.

vszakats:
After fixing the `configure` `Features:` line, tests are green. (except some unrelated-looking issues)

PR #9691, opened 2022-10-10
--url-query for curl

bagder:
This option adds a piece of data, usually a name + value pair, to the end of the URL query part. The syntax is identical to that used for `--data-urlencode` with one extension:

If the argument starts with a `+` (plus), the rest of the string is provided as-is unencoded.
    
This allows users to "build" query parts with options and URL encoding even when not doing GET requests, which the already provided option `-G` (`--get`) is limited to.


bagder:
**Note**

This format does not URL encode the left side of the `name=value` pairs. It assumes that side is already encoded. This was something I decided when I made the `--data-urlencode` option because if it isn't encoded, it might contain a `=` and then we cannot know which is data and which is separator.

The `=content` syntax will URL encode the entire thing.

PR #9697, opened 2022-10-11
test1275: verify upercase after period in markdown

bagder:
Script based on the #9474 pull-request logic, but implemented in perl.

Updated docs/URL-SYNTAX.md accordingly.

Suggested-by: Dan Fandrich


MarcelRaad:
The new script seems to be missing in the release archive:
https://curl.se/dev/log.cgi?id=20221013041732-2358088#prob1

```
+Can't open perl script "/curl-7.86.0-20221013/tests/markdown-uppercase.pl": No such file or directory[LF]
```

PR #9699, opened 2022-10-11
urlapi.c: nitpick grammar in function comment

12932:
because the 'u' in URL is actually a consonant *sound* it is only correct to write "a URL"

sorry this is a bit nitpicky :P

https://english.stackexchange.com/questions/152/when-should-i-use-a-vs-an 
https://www.techtarget.com/whatis/feature/Which-is-correct-a-URL-or-an-URL

### DIFF
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -116,7 +116,7 @@ static const char *find_host_sep(const char *url)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


12932:
done 👍

bagder (approved):

bagder:
Thanks!

PR #9700, opened 2022-10-11
tool_paramhelp: str2udouble 'max' changed to double

Ricardicus:
It is Hacktoberfest. I found myself here. Nitpicking in this glorious repository.

My LSP told me:

"Implicit conversion from 'long' to 'double' may lose precision"

I therefore thought that I perhaps could make a little PR.

Actually, with this PR; one may now be able to wait even longer for a connection to fail!
And that might also be a good thing.

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -808,7 +808,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
         break;
       case 'c': /* connect-timeout */
         err = str2udouble(&config->connecttimeout, nextarg,
-                          LONG_MAX/1000);
+                          (double)LONG_MAX/1000);
         if(err)
           return err;
         break;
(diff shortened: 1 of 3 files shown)


bagder:
The maximum accepted timeout value for connect timeout is set to be `LONG_MAX/1000` for a reason:

1. [CURLOPT_CONNECTTIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html) accepts a long
2. The option takes a millisecond counter so we multiply the given value by 1000
3. We therefore need the value to be be less than `LONG_MAX/1000` to not overflow/wrap in the math

Ricardicus:
Ok, I see. I will change it back to LONG_MAX/1000

bagder:
> Maybe it should be long all the way?

It parses a double to allow fractions of seconds: `1.234`, but it could certainly store that as milliseconds in a long.

Ricardicus:
I see.. I will think of a way to perhaps not compare a long to a double; that would be the only value in a PR like this hehe.

Ricardicus:
I think this is as far as this PR can go!

bagder (approved):

bagder:
"This branch cannot be rebased due to conflicts"

Can you please rebase on master, fix the issues and force-push?

Ricardicus:
There it is!

bagder:
Thanks!

PR #9255, opened 2022-08-05
fix Cygwin/MSYS compilation

orgads:
_getpid is Windows API. On Cygwin variants it should remain getpid.

Fixes #8220.

Issue #8220, opened 2022-01-03
cmake: curl subdirectory failed build due to getpid within smb.c


trackpadpro:
### I did this

Added cURL as a subdirectory within a CMake build. I am using MbedTLS, but that seems unrelated to this error. When I have more time I will likely try other CMake-compatible libraries. I only started to get this error once I started calling the cURL library within my main.cpp. 

### I expected the following

Everything to build. If I did not alter anything, I got the following:
```
[  0%] Building C object deps/curl/lib/CMakeFiles/libcurl.dir/smb.c.o
/c/Users/Valentin/Downloads/stateSpaceSolver/deps/curl/lib/smb.c: In function 'smb_format_message':
/c/Users/Valentin/Downloads/stateSpaceSolver/deps/curl/lib/smb.c:36:16: warning: implicit declaration of function '_getpid'; did you mean 'getpid'? [-Wimplicit-function-declaration]
   36 | #define getpid _getpid
      |                ^~~~~~~
/c/Users/Valentin/Downloads/stateSpaceSolver/deps/curl/lib/smb.c:371:9: note: in expansion of macro 'getpid'
  371 |   pid = getpid();
      |         ^~~~~~
/c/Users/Valentin/Downloads/stateSpaceSolver/deps/curl/lib/smb.c:36:16: warning: nested extern declaration of '_getpid' [-Wnested-externs]
   36 | #define getpid _getpid
      |                ^~~~~~~
/c/Users/Valentin/Downloads/stateSpaceSolver/deps/curl/lib/smb.c:371:9: note: in expansion of macro 'getpid'
  371 |   pid = getpid();
      |         ^~~~~~
[  0%] Linking C shared library libcurl.dll
/usr/lib/gcc/x86_64-pc-msys/11.2.0/../../../../x86_64-pc-msys/bin/ld: CMakeFiles/libcurl.dir/smb.c.o:smb.c:(.text+0x5aa): undefined reference to `_getpid'
collect2: error: ld returned 1 exit status
```

If I simply commented out the definition of getpid as _getpid, then I got no errors, but my libcurl.dll was not found when running the executable. I am fairly certain that my library inclusion is correct within my project's CMakeLists.txt and main.cpp, so I am assuming that there was a build problem caused by this line change. 

### curl/libcurl version

7.81.0-DEV

### operating system

Windows 10 (CMake through MSYS2 [msys64])

bagder:
[...]

PR #9701, opened 2022-10-11
smb: replace CURL_WIN32 with WIN32

vszakats:
PR #9255 aimed to fix a Cygwin/MSYS issue. It used the `CURL_WIN32` macro, but that one is not defined here, while compiling curl itself. This patch changes this to `WIN32`, assuming this was the original intent.

Regression from 1c52e8a3795ccdf8ec9c308f4f8f19cf10ea1f1a.

Ref: #8220

Closes #9701

/cc @orgads, @MarcelRaad

### DIFF
--- a/lib/smb.c
+++ b/lib/smb.c
@@ -34,7 +34,7 @@
 #include <process.h>
 #ifdef CURL_WINDOWS_APP
 #define getpid GetCurrentProcessId
-#elif defined(CURL_WIN32)
+#elif defined(WIN32)
 #define getpid _getpid
 #endif
 #endif


MarcelRaad:
The PR 9244 in the commit message should be 9255, right?

MarcelRaad (approved):

vszakats:
Maybe this condition could be simplified to never depend on `HAVE_PROCESS_H` on Windows (which is defined by CMake/`config_win32.h`, but not by autotools), and always use the lowest-level pubic Win32 API instead: `GetCurrentProcessId()`. Untested:

```c
#ifdef WIN32
#define getpid GetCurrentProcessId
#elif defined(HAVE_PROCESS_H)
#include <process.h>
#endif
```

Opinions?

vszakats:
@MarcelRaad: You're right, I mis-copy-pasted that one.

PR #9702, opened 2022-10-11
smb: tidy-up getpid on Windows

vszakats:
Always use the `GetCurrentProcessId()` on Windows.

It was always used for UWP apps, leaving other Windows builds with `getpid()` or `_getpid()` dependening on build configuration.

Ref: https://github.com/curl/curl/pull/9701#issuecomment-1274928405

### DIFF
--- a/lib/smb.c
+++ b/lib/smb.c
@@ -30,7 +30,7 @@
 
 #define BUILDING_CURL_SMB_C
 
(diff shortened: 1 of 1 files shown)


vszakats:
Closing in favor of https://github.com/curl/curl/pull/9703

PR #9703, opened 2022-10-11
tidy-up: process.h detection and use

vszakats:
This patch aims to cleanup the use of `process.h` header and the macro `HAVE_PROCESS_H` associated with it.

- `process.h` is always available on Windows. In curl, it is required only for `_beginthreadex()` in `lib/curl_threads.c`.

- `process.h` is also available in MS-DOS. In curl, its only use was in `lib/smb.c` for `getpid()`. But `getpid()` is in fact declared by `unistd.h`, which is always enabled via `lib/config-dos.h`. So the header is not necessary.

- `HAVE_PROCESS_H` was detected by CMake, forced to 1 on Windows and left to real detection for other platforms. It was also set to always-on in `lib/config-win32.h` and `lib/config-dos.h`. In autotools builds, there was no detection and the macro was never set.

Based on these observations, in this patch we:

- Rework Windows `getpid` logic in `lib/smb.c` to always use the equivalent direct Win32 API function `GetCurrentProcessId()`, as we already did for Windows UWP apps. This makes `process.h` unnecessary here on Windows.

- Stop #including `process.h` into files where it was not necessary. This is everywhere, except `lib/curl_threads.c`.

  > Strangely enough, `lib/curl_threads.c` compiled fine with autotools because `process.h` is also indirecty included via `unistd.h`. This might have been broken in autotools MSVC builds, where the latter header is missing.

- Delete all remaining `HAVE_PROCESS_H` feature guards, for they were unnecessary.

- Delete `HAVE_PROCESS_H` detection from CMake and predefined values from `lib/config-*.h` headers.

Closes #xxxx


jay (approved):

vszakats:
Testing an MS-DOS build with this patch, would be helpful. To be on the safe side.

Is there someone in the position to do it?

jay:
> Testing an MS-DOS build with this patch, would be helpful.

@gvanem

vszakats:
Managed to test with MS-DOS, and unless I'm overlooking something, `smb.c` is _excluded_ from MS-DOS builds since 24b30d259caf2317bc38873b2f6ee20dd9682cc7, due to `SIZEOF_CURL_OFF_T == 4` there and SMB requiring 64-bit integers. In that case, the `MSDOS` guard can just be deleted. Does that sound right?

jay:
Yes. FYI I think at one point @bagder or someone mentioned dropping support for 32-bit curl_off_t in [libcurl 8](https://github.com/curl/curl/wiki/libcurl-8) but I can't find the thread.

vszakats:
@jay: I saw the libcurl8 plans. I'm wondering if 64-bit integers are universally missing from MS-DOS compilers, or is it just not implemented in `lib/config-dos.h`? Such integers are supported for file offsets in later MS-DOS editions.

Anyway, as a test, I deleted the `SIZEOF_CURL_OFF_T` condition from `smb.c` and it compiled without warnings with DJGPP. So I think the PR is ready to merge as it is.

bagder:
Yeah, I can't think of any reason why compilers on MS-DOS shouldn't support 64 bit types. I think that was just someone's "mis-thought" when the smb code was written and nobody has bothered about since.

The planned deprecation of builds for systems without 64 bit types is mentioned here: https://curl.se/dev/deprecate.html

gvanem:
@jay 

> Testing an MS-DOS build with this patch, would be helpful.

I have no longer a way to run DOS/djgpp programs. I scrapped my old Win-XP box.

vszakats:
I used these DJGPP builds that run on Linux/macOS/modern-Windows and FreeBSD:
https://github.com/andrewwutw/build-djgpp

vszakats:
More MS-DOS testing revealed that `getpid()` is not declared in `process.h` but in `unistd.h`, which is always included due to `HAVE_UNISTD_H == 1` in `lib/config-dos.h`. It means the MS-DOS-specific lines in `smb.c` are not necessary after all.

vszakats:
I flagged this next-feature-window, but it isn't a feature really. Also well-tested by now.

Any objections to merging?

jay:
> I flagged this next-feature-window, but it isn't a feature really

next-feature-window it doesn't have to be a feature, for example it could be something that has a possibility of breaking builds that's why you're waiting until after the release for the next feature window. there's usually a several day period (like we're in now)  where we can see if a patch release is necessary before the window opens. at this point i'd just wait the few days.

vszakats:
You're right, let's do that!

PR #9705, opened 2022-10-11
tests: avoid CreateThread if _beginthreadex is available

mback2k:
CreateThread is not threadsafe if mixed with CRT calls.
_beginthreadex on the other hand can be mixed with CRT.


MarcelRaad (approved):
Might be good if we could define `curl_win_thread_handle_t` in some common place instead of multiple times separately, but I don't see a common include. Looks good to me!

PR #9710, opened 2022-10-12
tool_formparse: unroll the NULL_CHECK and CONST_FREE macros

bagder:
To make the code read more obvious

### DIFF
--- a/src/tool_formparse.c
+++ b/src/tool_formparse.c
@@ -38,9 +38,6 @@
 
 #include "memdebug.h" /* keep this as LAST include */
 
-/* Macros to free const pointers. */
-#define CONST_SAFEFREE(x)       Curl_safefree(*((void **) &(x)))
-
 /* tool_mime functions. */
 static struct tool_mime *tool_mime_new(struct tool_mime *parent,
                                        toolmimekind kind)
@@ -64,7 +61,7 @@ static struct tool_mime *tool_mime_new_parts(struct tool_mime *parent)
 }
 
 static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
-                                            const char *data)
+                                            char *data)
 {
   struct tool_mime *m = NULL;
 
@@ -72,7 +69,7 @@ static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
   if(data) {
     m = tool_mime_new(parent, TOOLMIME_DATA);
     if(!m)
-      free((char *)data);
+      free(data);
     else
       m->data = data;
   }
@@ -80,7 +77,7 @@ static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
 }
 
 static struct tool_mime *tool_mime_new_filedata(struct tool_mime *parent,
-                                                const char *filename,
+                                                char *filename,
                                                 bool isremotefile,
                                                 CURLcode *errcode)
 {
@@ -94,7 +91,7 @@ static struct tool_mime *tool_mime_new_filedata(struct tool_mime *parent,
     if(filename) {
       m = tool_mime_new(parent, TOOLMIME_FILE);
       if(!m)
-        free((char *)filename);
+        free(filename);
       else {
         m->data = filename;
         if(!isremotefile)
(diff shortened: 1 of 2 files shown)


src/tool_formparse.c
```diff
@@ -73,7 +72,7 @@ static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
   if(data) {
     m = tool_mime_new(parent, TOOLMIME_DATA);
     if(!m)
-      CONST_FREE(data);
+      free((char *)data);
```
jay, line 75: Instead of freeing pointers to const I would rather we change the function prototypes:

~~~diff
diff --git a/src/tool_formparse.c b/src/tool_formparse.c
index 14189d4..c54f9ef 100644
--- a/src/tool_formparse.c
+++ b/src/tool_formparse.c
@@ -64,7 +64,7 @@ static struct tool_mime *tool_mime_new_parts(struct tool_mime *parent)
 }
 
 static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
-                                            const char *data)
+                                            char *data)
 {
   struct tool_mime *m = NULL;
 
@@ -72,7 +72,7 @@ static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
   if(data) {
     m = tool_mime_new(parent, TOOLMIME_DATA);
     if(!m)
-      free((char *)data);
+      free(data);
     else
       m->data = data;
   }
@@ -80,7 +80,7 @@ static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
 }
 
 static struct tool_mime *tool_mime_new_filedata(struct tool_mime *parent,
-                                                const char *filename,
+                                                char *filename,
                                                 bool isremotefile,
                                                 CURLcode *errcode)
 {
@@ -94,7 +94,7 @@ static struct tool_mime *tool_mime_new_filedata(struct tool_mime *parent,
     if(filename) {
       m = tool_mime_new(parent, TOOLMIME_FILE);
       if(!m)
-        free((char *)filename);
+        free(filename);
       else {
         m->data = filename;
         if(!isremotefile)
~~~

/cc @monnerat

bagder: > Instead of freeing pointers to const I would rather we change the function prototypes:

Agreed

monnerat:
Some explanations about the original code:

- initial design was made around libcurl 's private structs that have const fields, requiring then casted `free`() calls. This was the reason for `CONST_FREE`() and subsisted even after converting to tool's own struct usage.
- Different error return values help debuging and/or problem tracking.
- I'm always reluctant about using a goto, even in the case of error return!

bagder:
> Different error return values help debuging and/or problem tracking.

Agreed. But since those different values are not used in the code today, you need to edit the code to debug the code with them. And when you edit code, you can also do other things. This change of mine could also do different return codes, but as nothing used those I figured this way would be simpler.

> I'm always reluctant about using a goto, even in the case of error return!

I think using goto and a single cleanup path is better than trying to cleanup properly on N different points within the same function.

But primarily I do not like using macros that makes reading and following the code hard. And I think these do, hence me proposing we get rid of them. A balance of course.

monnerat:
They were just explanation comments, not criticisms!
> But since those different values are not used in the code today, you need to edit the code to debug the code with them. 

A not yet known problem may occur even if you don't edit the code; in this case, gdb is your friend and checking the returned code may light your mind about what has caused the error. But I agree this was more useful during development than now.
> But primarily I do not like using macros that makes reading and following the code hard.

I can understand that, although the macro names were meaningful.
I wish we had exceptions in C ;-)

bagder:
> I can understand that, although the macro names were meaningful.

The names were fine, but they still hide what is actually going on. Which probably is less of a concern when you are the author of the code, more of a problem when reading someone else's code using such macros.

It's not a critique of how it was done. I just think removing them is a way to make the code ever so slightly more readable. (After I spent some time reading code in that file.)

There's still a `SET_TOOL_MIME_PTR` macro in there that I also don't like, but I'm saving that for another time.

monnerat:
> The names were fine, but they still hide what is actually going on. Which probably is less of a concern when you are the author of the code, more of a problem when reading someone else's code using such macros.

So is it with actual non-macro functions!

bagder:
> > The names were fine, but they still hide what is actually going on. Which probably is less of a concern when you are the author of the code, more of a problem when reading someone else's code using such macros.
> 
> So is it with actual non-macro functions!

The point is that the macros make it *harder* to read and follow. Whoever wrote it.

PR #9712, opened 2022-10-12
lib: set more flags in config-win32.h

vszakats:
The goal is to add any flag that affect the created binary, to get in sync with the ones built with CMake and autotools.

I took these flags from curl-for-win [0], where they've been tested with mingw-w64 and proven to work well.

This patch brings them to curl as follows:

- Enable unconditionally those force-enabled via `CMake/WindowsCache.cmake`:

  - `HAVE_SETJMP_H`
  - `HAVE_STRING_H`
  - `HAVE_SIGNAL` (CMake equivalent is `HAVE_SIGNAL_FUNC`)

- Expand existing guards with mingw-w64:

  - `HAVE_STDBOOL_H`
  - `HAVE_BOOL_T`

- Enable Win32 API functions for Windows Vista and later:

  - `HAVE_INET_NTOP`
  - `HAVE_INET_PTON`

- Set sizes, if not already set:
    
  - `SIZEOF_OFF_T = 8`
  - `_FILE_OFFSET_BITS = 64` when `USE_WIN32_LARGE_FILES` is set,
    and using mingw-w64.

- Add the remaining for mingw-w64 only. Feel free to expand as desired:

  - `HAVE_LIBGEN_H`
  - `HAVE_FTRUNCATE`
  - `HAVE_BASENAME`
  - `HAVE_STRTOK_R`

Future TODOs:

- `HAVE_SIGNAL` has a different meaning in CMake. It's enabled when both the `signal()` function and the `SIGALRM` macro are found. In autotools and this header, it means the function only. For the function alone, CMake uses `HAVE_SIGNAL_FUNC`.

[0] https://github.com/curl/curl-for-win/blob/c9b9a5f273c94c73d2b565ee892c4dff0ca97a8c/curl-m32.sh#L53-L58

Closes #xxxx

### DIFF
--- a/lib/config-win32.h
+++ b/lib/config-win32.h
@@ -79,7 +79,7 @@
 
 /* Define to 1 if you have the <stdbool.h> header file. */
 #if (defined(_MSC_VER) && (_MSC_VER >= 1800)) || \
-    (defined(__MINGW32__) && defined(__MINGW64_VERSION_MAJOR))
+    defined(__MINGW64_VERSION_MAJOR)
 #define HAVE_STDBOOL_H 1
 #endif
 
@@ -146,7 +146,7 @@
 #define HAVE_STRING_H 1
 
 /* Define to 1 if you have the <libgen.h> header file. */
-#if defined(__MINGW32__) && defined(__MINGW64_VERSION_MAJOR)
+#if defined(__MINGW64_VERSION_MAJOR)
 #define HAVE_LIBGEN_H 1
 #endif
 
@@ -162,7 +162,7 @@
 
 /* Define to 1 if bool is an available type. */
 #if (defined(_MSC_VER) && (_MSC_VER >= 1800)) || \
-    (defined(__MINGW32__) && defined(__MINGW64_VERSION_MAJOR))
+    defined(__MINGW64_VERSION_MAJOR)
 #define HAVE_BOOL_T 1
 #endif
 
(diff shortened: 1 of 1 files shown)


vszakats:
The AppVeyor CI job failure is unrelated:
```
TESTFAIL: These test cases failed: 1275
test 1275...[Verify capital letters after period in markdown files]
```

https://ci.appveyor.com/project/curlorg/curl/builds/45050447/job/odv664nav6sytial#L7280

bagder:
> The AppVeyor CI job failure is unrelated:

Wow, that was a spectacularly weird error!

bagder (approved):

peterpiekarski:
@bagder, @vszakats  , building libcurl for windows (64bit with VS2022), it appears sizeof(off_t) is equal to 4 and not 8. Looking at the CRT's types.h file (Windows 10 SDK 10.0.19041.0), off_t is defined as
    typedef long _off_t; // file offset value
where "long" is 32 bit in Windows builds.

After updating from 7.84.0 to 7.86.0, a static_assert in our code was triggered, which we added among others to make sure libcurl was not misconfigured for the target:
	static_assert(SIZEOF_OFF_T == sizeof(off_t), "off_t type size not defined correctly");

I was also surprised to find that off_t and SIZEOF_OFF_T appear to be not used anywhere inside libcurl, and I was wondering why it was added in the first place if not used.

vszakats:
@peterpiekarski Thanks for your report. Before this patch this macro settled on 4 by default, indeed. I didn't recognize this value is CRT-dependent. So maybe this method would be better by limiting this line to MinGW:

```c
#if defined(__MINGW32__)
#  define SIZEOF_OFF_T 8
#endif
```

As a temporary workaround you can use `-DSIZEOF_OFF_T=4`.

PR #9713, opened 2022-10-12
First version of Multipath TCP support in curl

obonaventure:
Initial support for Multipath TCP for curl. Multipath TCP is a TCP extension defined in RFC8684 that is supported by recent Linux kernels (5.13+). Multipath TCP allows a TCP connection to use different paths (e.g. Wi-Fi and cellular on smartphones, IPv4 and IPv6, ...). Multipath TCP is enabled by default on recent Linux distributions (Debian, Ubuntu, Redhat,...)

This patch provides the --multipath command line option that enables Multipath TCP when set. If running on a kernel that does not support Multipath TCP, curl remembers the fact that Multipath TCP does not work and does not attempt anymore to use Multipath TCP when creating a TCP connection. Another possibility would be to always enable Multipath TCP by default on Linux.

With the patch, curl reports the activation of Multipath TCP when interacting with a Multipath TCP server.

Compile with -DMPTCP to enable Multipath TCP support
```
* Multipath TCP is enabled.
*   Trying [2001:41d0:a:ffcf::1]:80...
* Failed to set TCP_KEEPIDLE on fd 5
* Failed to set TCP_KEEPINTVL on fd 5
* Connected to test.multipath-tcp.org (2001:41d0:a:ffcf::1) port 80 (#0)
* Multipath TCP was accepted by the remote server.
> GET / HTTP/1.1
> Host: test.multipath-tcp.org
> User-Agent: curl/7.86.0-DEV
> Accept: */*
>
* Mark bundle as not supporting multiuse < HTTP/1.1 200 OK
< Content-Type: text/html
< Accept-Ranges: bytes
< ETag: "4215149735"
< Last-Modified: Tue, 05 Jul 2022 16:11:47 GMT
< Content-Length: 389
< Date: Wed, 12 Oct 2022 16:32:51 GMT
< Server: lighttpd/1.4.59
<
<!DOCTYPE html>
<html>
<head>
<title>Welcome to test.multipath-tcp.org!</title>
<style>
    body {
        width: 35em;
        margin: 0 auto;
        font-family: Tahoma, Verdana, Arial, sans-serif;
    }
</style>
</head>
<body>
<h1>Welcome to test.multipath-tcp.org !</h1>
<p>This web server runs Multipath TCP v1</p>

<p><em>Thank you for using Multipath TCP.</em></p> </body>
</html>
* Connection #0 to host test.multipath-tcp.org left intact
```
First version of Multipath TCP support in curl, compile with ./configure CPPFLAGS="-DMPTCP"


include/curl/curl.h
```diff
@@ -2157,6 +2157,11 @@ typedef enum {
   /* websockets options */
   CURLOPT(CURLOPT_WS_OPTIONS, CURLOPTTYPE_LONG, 320),
 
+#ifdef MPTCP
+  /* Set Multipath TCP */
+  CURLOPT(CURLOPT_TCP_MULTIPATH, CURLOPTTYPE_LONG, 316),
+#endif
+
```
bagder, line 2164: This is a public header, it cannot use `#ifdefs` like this. The symbols need to be provided no matter what.

lib/connect.c
```diff
@@ -1550,6 +1550,24 @@ bool Curl_connalive(struct connectdata *conn)
...
+      infof(data,"Multipath TCP used %d subflows on this connection",inf.mptcpi_subflows);
+    }
+    else {
+      infof(data,"Multipath TCP was DISABLED on this connection");
+    }
+  }
+#endif /* MPTCP */
+
+
+
```
bagder, line 1570: This block looks like it's there for debugging? Maybe use `#ifdef DEBUGBUILD` as well?

lib/connect.c
```diff
@@ -1641,7 +1663,7 @@ CURLcode Curl_socket(struct Curl_easy *data,
                                     (struct curl_sockaddr *)addr);
     Curl_set_in_callback(data, false);
   }
-  else
+  else
```
bagder, line 1666: Please don't modify whitespace.

lib/connect.c
```diff
@@ -1680,6 +1702,7 @@ CURLcode Curl_socket(struct Curl_easy *data,
 #endif
 
   return CURLE_OK;
+
```
bagder, line 1705: and here

lib/connect.c
```diff
@@ -1709,6 +1732,7 @@ void Curl_conncontrol(struct connectdata *conn,
     conn->bits.close = closeit; /* the only place in the source code that
                                    should assign this bit */
   }
+
```
bagder, line 1735: and here

lib/socketpair.c
```diff
@@ -42,6 +42,8 @@
 #ifdef HAVE_NETINET_IN_H
 #include <netinet/in.h> /* IPPROTO_TCP */
 #endif
+
+
```
bagder, line 46: Just no

lib/socketpair.c
```diff
@@ -73,8 +75,8 @@ int Curl_socketpair(int domain, int type, int protocol,
   (void)domain;
   (void)type;
   (void)protocol;
-
-  listener = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+  printf("mptcp\n");
```
bagder, line 78: debug leftover?

lib/socketpair.c
```diff
@@ -73,8 +75,8 @@ int Curl_socketpair(int domain, int type, int protocol,
...
+  listener = socket(AF_INET, SOCK_STREAM, IPPROTO_MPTCP);
```
bagder, line 79: why?

lib/socketpair.h
```diff
@@ -34,4 +34,11 @@ int Curl_socketpair(int domain, int type, int protocol,
 #define Curl_socketpair(a,b,c,d) socketpair(a,b,c,d)
 #endif
 
+#ifdef MPTCP
+#ifndef IPPROTO_MPTCP
+#define IPPROTO_MPTCP 262
+#endif
+#endif
+
+
```
bagder, line 43: Why duplicate this snippet?

src/tool_getparam.c
```diff
@@ -250,6 +250,9 @@ static const struct LongShort aliases[]= {
   {"Ek", "tlsuser",                  ARG_STRING},
   {"El", "tlspassword",              ARG_STRING},
   {"Em", "tlsauthtype",              ARG_STRING},
+#ifdef MPTCP
+  {"EM", "multipath",                ARG_BOOL},
+#endif
```
bagder, line 255: The cmdline options cannot be ifdef'ed like this, as the tool might not know what features the lib was built with.

bagder (commented):
Thanks for your contribution. I think this is a future that might get appreciated. I think the PR needs a little more polish though. There is also no docs. I imagine creating tests for this is hard.

lib/connect.c
obonaventure: Yes, good suggestion

lib/socketpair.h
obonaventure: This is mistake, apologies. Windows does not support MPTCP and nothing needs to be changed in this file

obonaventure:
Thanks for your comments, I'm looking at them and will propose an update. I'll propose some documentation. For the tests, I can propose a modification to the http server to enable multipath tcp and then add a few tests. These tests would only work on recent linux kernels.

bagder:
Additionally: macOS supports multipath TCP as well so it would make sense to provide this feature there was well.

Are you still up for working on this PR?

bagder:
Abandoned.

gvanem:
Looks like someone else has added MTCP to curl (on Windows at least): 
https://github.com/ngi-mptcp/curl/wiki/Multipath-TCP-on-Windows-10

With the help of wintap and [LKL](https://github.com/ngi-mptcp/lkl-next/); a *Linux Kernel Library*. Fascinating stuff.

Issue #9706, opened 2022-10-11
SChannel certificate chain order reversed in Windows 11 22H2


nmoinvaz:
### I did this

```c
curl_easy_getinfo(curl_handle, CURLINFO_CERTINFO, &cert_info)
```

### I expected the following

The ordering of the certificate chain should be consistent across different versions of Windows and different platforms.

### Problem details

This problem report follows in the foot steps of:

* https://github.com/curl/curl/pull/4519
* https://github.com/curl/curl/issues/4518
* Commit: https://github.com/curl/curl/pull/4519/commits/55a143c8d3800b6634ff0dc0afa284ebf736ec60
* Mailing list mention: https://curl.se/mail/lib-2018-10/0099.html

Older versions of Chromium that used SChannel for TLS/SSL implementation reported:

```c
  // In testing, enumerating the store returned from SChannel appears to
  // enumerate certificates in reverse of the order they were added, meaning
  // that issuer certificates appear before the subject certificates. This is
  // likely because the default Windows memory store is implemented as a
  // linked-list. Reverse the list, so that intermediates are ordered from
  // subject to issuer.
  // Note that the store also includes the end-entity (server) certificate,
  // so exclude this certificate from the set of |intermediates|.
```

https://github.com/youtube/h5vcc/blob/7da685fe633dac18fd8db35624f7bade8619c977/external/chromium/net/socket/ssl_client_socket_win.cc#L1559

Chromium now uses BoringSSL, but you can see there was a problem with the way that SChannel returned certificate chains.

In Windows 11 22H2 it appears that the incorrect ordering of certificates has been corrected and that subject certificates are now correctly appearing before issuer certificates. So reversing the ordering of certificates in the chain in curl is no longer necessary according to how newer versions of Windows are behaving.

I have also filed the issue on Microsoft Feedback Hub to get clarification, but I am not hopeful. I want to bring awareness about the issue and hopefully we can get a solution for `CURLINFO_CERTINFO` when using SChannel that works consistently in all versions of Windows.

### curl/libcurl version

7.84

### operating system

Windows 11 22H2 OS Build 22621.674

jay:
> I have also filed the issue on Microsoft Feedback Hub to get clarification

Is there a public link to this issue?

bagder:
Do we need to add a run-time version-check (for Windows 11 22H2?) to figure out what order to use?

nmoinvaz:
> Is there a public link to this issue?

[...]

PR #9721, opened 2022-10-13
lib3028: check CURLOPT_CERTINFO certificate chain contains leaf first.

nmoinvaz:
This lib test checks that the certificate chain returned from `curl_easy_getinfo(curl, CURLINFO_CERTINFO, &cert_info)` is in the correct order: leaf first, root last. 

This was inspired by the fact that on recent versions of Windows 11 22H2, the certificates returned from SChannel are in reverse order.  See #9706.

I have named it `lib3028` but I can rename to whatever you want, because I didn't see anywhere what each range of tests targeted.

### DIFF
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -248,4 +248,4 @@ test2300 test2301 test2302 test2303 \
 test3000 test3001 test3002 test3003 test3004 test3005 test3006 test3007 \
 test3008 test3009 test3010 test3011 test3012 test3013 test3014 test3015 \
 test3016 test3017 test3018 test3019 test3020 test3021 test3022 test3023 \
-test3024 test3025 test3026 test3027
+test3024 test3025 test3026 test3027 test3028
--- a/tests/data/test3028
+++ b/tests/data/test3028
@@ -1,0 +1,53 @@
+<testcase>
+<info>
+<keywords>
+HTTPS
+HTTP GET
+</keywords>
+</info>
+
+#
+# Server-side
+<reply>
+<data>
+HTTP/1.1 200 OK
+Date: Tue, 09 Nov 2010 14:49:00 GMT
+Server: test-server/fake
+Content-Length: 7
+
+MooMoo
+</data>
+</reply>
+
+#
+# Client-side
+<client>
+<features>
+SSL
+</features>
+<server>
+https
+</server>
+ <name>
+Verify certificate chain order with simple HTTPS GET
+ </name>
+<tool>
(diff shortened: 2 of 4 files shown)


nmoinvaz:
On my second commit I realized that we can just compare the entire distinguished name since curl is the one constructing both in the exact same way with ASN1 class. I can squash if needed.

nmoinvaz:
Output looks like this on failure:
```
cert 0
  subject: C=US, O=DigiCert Inc, CN=DigiCert TLS Hybrid ECC SHA384 2020 CA1
  issuer: C=US, O=DigiCert Inc, OU=www.digicert.com, CN=DigiCert Global Root CA
cert 1
  subject: C=US, ST=California, L=San Francisco, O=GitHub, Inc., CN=github.com
  issuer: C=US, O=DigiCert Inc, CN=DigiCert TLS Hybrid ECC SHA384 2020 CA1
cert 0 issuer does not match cert 1 subject
certificate chain is not in order
```
Output looks like this on success:
```
cert 0
  subject: C=US, ST=California, L=San Francisco, O=GitHub, Inc., CN=github.com
  issuer: C=US, O=DigiCert Inc, CN=DigiCert TLS Hybrid ECC SHA384 2020 CA1
cert 1
  subject: C=US, O=DigiCert Inc, CN=DigiCert TLS Hybrid ECC SHA384 2020 CA1
  issuer: C=US, O=DigiCert Inc, OU=www.digicert.com, CN=DigiCert Global Root CA
certificate chain is in order
```

dfandrich:
> I have named it lib3028 but I can rename to whatever you want, because I didn't
> see anywhere what each range of tests targeted.

There used to be a specified ranges for tests but that was removed in commit
a8ac1be7 because people were ignoring it anyway.

nmoinvaz:
Squashed commits.

bagder:
It looks good.

This is doesn't actually *run* the test though, you need a `tests/data/test3028` file for that.

nmoinvaz:
I'm not exactly sure what I should do for a `test3028`, so I just copied `test300` which is a simple HTTPS GET. When I tested `lib3028` I just would pass it https://example.com/ or https://github.com/.

nmoinvaz:
Looks like that didn't work. I'm not exactly sure what I'm doing wrong.

nmoinvaz:
@bagder do you have any hints about what I might be doing wrong with this test data? 🙏🏻

nmoinvaz:
Nevermind. I figured it out.

PR #9726, opened 2022-10-13
cmake: set HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID on Windows

vszakats:
This configuration option is enabled unconditionally in `lib/config-win32.h`. Make it apply to CMake builds as well.

While here, delete a broken check for `HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID` from `CMakeLists.txt`. This came with the initial commit [1], but did not include the actual verification code inside `CMake/CurlTests.c`, so it always failed. A later commit [2] added a second test, for non-Windows platforms.

Enabling this flag causes test 1056 to fail with CMake builds, just like they do with autotools builds. Let's apply the same solution and ignore the results here as well.

[1] 4c5307b45655ba75ab066564afdc0c111a8b9291
[2] aec7c5a87c8482b6ddffa352d7d220698652262e
 
Closes #xxxx


bagder (approved):

vszakats:
AppVeyor CI, CMake **debug** builds (both MSVC and MinGW) are consistently failing test 1056:
https://ci.appveyor.com/project/curlorg/curl/builds/45065933/job/8hidqd39hbaot4fu#L4047

Seems related. I rarely do debug builds, so any help is appreciated.

MarcelRaad:
It doesn't happen in release builds only because they're not running tests at all.

Test 1056 is also failing for the autotools builds on AppVeyor, it's just disabled there. So it probably can be safely disabled for CMake too. I don't know the reason why it fails though.

vszakats:
It might be because it should not be enabled without IPv6. Committed a fix.

mback2k:
I wonder if maybe test1056 is not correct if it should not run without full IPv6 support, but it has the corresponding keyword and feature required?
https://github.com/curl/curl/blob/86f2d8f67f0e4255128e448e531d91aec4cfd6be/tests/data/test1056#L7
https://github.com/curl/curl/blob/86f2d8f67f0e4255128e448e531d91aec4cfd6be/tests/data/test1056#L46

vszakats:
@mback2k: Strangely, today these (8) failures [reappeared](https://ci.appveyor.com/project/curlorg/curl/builds/45074196) under a [different PR](https://github.com/curl/curl/pull/9734), that [did not have](https://github.com/curl/curl/commit/fa62a26022308b7d1512cf6cbaec99612c9f361b) this patch. Rebasing to this one "fixed it".

mback2k:
So, do I understand it correctly that https://github.com/curl/curl/commit/fa62a26022308b7d1512cf6cbaec99612c9f361b makes test 1056 work again?

vszakats:
Nope, sorry, I had to rebase it to _this_ patch, which disabled those tests. What I didn't understand is why did they (re)appear there in the first place.

mback2k:
Maybe because your PR enabled the IPv6 feature for those builds?

vszakats:
That's the interesting bit, that PR is not supposed to change things around IPv6. It adds an _additional_ safeguard to `lib/hostip4.c` to avoid a potential misconfiguration with GETADDRNAME vs GETADDRNAME_THREADSAFE. Such mis-configuration should not happen with curl build systems (and if it does, it would fail at link time), so at this point it should not change anything at all, but certainly not test results.

PR #9727, opened 2022-10-13
cmake: set HAVE_GETADDRINFO_THREADSAFE on Windows

vszakats:
autotools enables this configuration option unconditionally for Windows [^1]. Do the same in CMake.

The above will make this work for all reasonably recent environments. The logic present in `lib/config-win32.h` [^2] has the following exceptions which we did not cover in this CMake update:

- Builds targeting Windows 2000 and earlier
- MS Visual C++ 5.0 (1997) and earlier

Defining `HAVE_GETADDRINFO_THREADSAFE` without
`HAVE_GETADDRINFO` results in a broken build:
https://ci.appveyor.com/project/curlorg/curl/builds/45062422/job/tjyv75wxfdhaledj#L357
Add condition to ensure this doesn't happen. (It would probably be nicer to check for this in C).
[PENDING]

[^1]: https://github.com/curl/curl/blob/68fa9bf3f5d7b4fcbb57619f70cb4aabb79a51f6/m4/curl-functions.m4#L2067-L2070 

[^2]: https://github.com/curl/curl/blob/68fa9bf3f5d7b4fcbb57619f70cb4aabb79a51f6/lib/config-win32.h#L511-L528

Closes #xxxx


vszakats:
I did a patch that adds a C-level solution for this, do want it?
```patch
Date:   2022-10-14 14:36:21 +0000

    sync conditions for Curl_getaddrinfo_ex() def and use
    
    `Curl_getaddrinfo_ex()` gets _defined_ with `HAVE_GETADDRINFO` set.
    But, `hostip4.c` _used_ it with `HAVE_GETADDRINFO_THREADSAFE` set alone.
    It meant a build with the latter, but without the former flag could
    result in calling this function but not defining it, and failing to
    link.
    
    Patch this by adding an extra check for `HAVE_GETATTRINFO` around
    the call.
    
    Before this patch, build systems prevented this condition. Now they
    don't need to.

diff --git a/lib/hostip4.c b/lib/hostip4.c
index 1dd54e879..df4f2c89b 100644
--- a/lib/hostip4.c
+++ b/lib/hostip4.c
@@ -125,14 +125,15 @@ struct Curl_addrinfo *Curl_getaddrinfo(struct Curl_easy *data,
 struct Curl_addrinfo *Curl_ipv4_resolve_r(const char *hostname,
                                           int port)
 {
-#if !defined(HAVE_GETADDRINFO_THREADSAFE) && defined(HAVE_GETHOSTBYNAME_R_3)
+#if !(defined(HAVE_GETADDRINFO) && defined(HAVE_GETADDRINFO_THREADSAFE)) && \
+   defined(HAVE_GETHOSTBYNAME_R_3)
   int res;
 #endif
   struct Curl_addrinfo *ai = NULL;
   struct hostent *h = NULL;
   struct hostent *buf = NULL;
 
-#if defined(HAVE_GETADDRINFO_THREADSAFE)
+#if defined(HAVE_GETADDRINFO) && defined(HAVE_GETADDRINFO_THREADSAFE)
   struct addrinfo hints;
   char sbuf[12];
   char *sbufptr = NULL;
@@ -280,14 +281,16 @@ struct Curl_addrinfo *Curl_ipv4_resolve_r(const char *hostname,
     h = NULL; /* set return code to NULL */
     free(buf);
   }
-#else /* HAVE_GETADDRINFO_THREADSAFE || HAVE_GETHOSTBYNAME_R */
+#else /* (HAVE_GETADDRINFO && HAVE_GETADDRINFO_THREADSAFE) ||
+          HAVE_GETHOSTBYNAME_R */
   /*
    * Here is code for platforms that don't have a thread safe
    * getaddrinfo() nor gethostbyname_r() function or for which
    * gethostbyname() is the preferred one.
    */
   h = gethostbyname((void *)hostname);
-#endif /* HAVE_GETADDRINFO_THREADSAFE || HAVE_GETHOSTBYNAME_R */
+#endif /* (HAVE_GETADDRINFO && HAVE_GETADDRINFO_THREADSAFE) ||
+           HAVE_GETHOSTBYNAME_R */
 
   if(h) {
     ai = Curl_he2ai(h, port);
```

PR #9730, opened 2022-10-13
schannel: don't reverse cert chain for CURLOPT_CERTINFO on Win 11 22H2+.

nmoinvaz:
On newer versions of Windows 11 (22H2 or higher) `QueryContextAttributes` and `CertEnumCertificatesInStore` return the certificate chain in the expected order, leaf first so there is no need to reverse the insertion order of the certificates when adding to the chain. See #9706.

Co-authored-by: Sergey Markelov

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -1665,7 +1665,12 @@ add_cert_to_certinfo(const CERT_CONTEXT *ccert_context, void *raw_arg)
   if(valid_cert_encoding(ccert_context)) {
     const char *beg = (const char *) ccert_context->pbCertEncoded;
     const char *end = beg + ccert_context->cbCertEncoded;
-    int insert_index = (args->certs_count - 1) - args->idx;
+    int insert_index = args->idx;
+    /* Windows prior to 11 22H2 returned certificates in reverse order. */
+    if(curlx_verify_windows_version(10, 0, 22621, PLATFORM_WINNT,
+                                    VERSION_LESS_THAN)) {
+      insert_index += args->certs_count - 1;
+    }
     args->result = Curl_extract_certinfo(args->data, insert_index,
                                          beg, end);
     args->idx++;


lib/vtls/schannel.c
```diff
@@ -1665,7 +1665,12 @@ add_cert_to_certinfo(const CERT_CONTEXT *ccert_context, void *raw_arg)
   if(valid_cert_encoding(ccert_context)) {
     const char *beg = (const char *) ccert_context->pbCertEncoded;
     const char *end = beg + ccert_context->cbCertEncoded;
-    int insert_index = (args->certs_count - 1) - args->idx;
+    int insert_index = args->idx;
+    /* Windows prior to 11 22H2 returned certificates in reverse order. */
+    if(curlx_verify_windows_version(10, 0, 22621, PLATFORM_WINNT,
+                                    VERSION_LESS_THAN)) {
+      insert_index += args->certs_count - 1;
+    }
```
vszakats, line 1673: `curlx_verify_windows_version()` looks to be a quite heavy function, making many API calls, even an indirect one, and it can even recurse. Haven't made measurements though. [^1]

Maybe worth caching the version result in a `raw_arg` variable?

(Having said that, this version check call is now made from a lot of low-level places in `schannel.c`.)

[^1]: https://github.com/curl/curl/blob/57e2bb52aaa7d16adef6d3d1c4b75c242148a85e/lib/version_win32.c#L196-L314

nmoinvaz: Yeah I noticed that too. It seemed to be used all over the place. The results probably should be memoized if it takes too long.

bagder: It seems reasonable to make such an optimization (if feasible) in a separate PR though.

jay:
AFAICT MS has not acknowledged whether or not they intended to change the order. It may be a bug. Do you know if this affects earlier versions of Windows 11 such as 21H2?

nmoinvaz:
I only experienced this after upgrading to 22H2 and I was previously on 21H2.

PR #9734, opened 2022-10-14
lib: sync guard for Curl_getaddrinfo_ex() definition and use

vszakats:
`Curl_getaddrinfo_ex()` gets _defined_ with `HAVE_GETADDRINFO` set. But, `hostip4.c` _used_ it with `HAVE_GETADDRINFO_THREADSAFE` set alone. It meant a build with the latter, but without the former flag could result in calling this function but not defining it, and failing to link.

Patch this by adding an extra check for `HAVE_GETATTRINFO` around the call.

Before this patch, build systems prevented this condition. Now they don't need to, so delete the build-system-specific logic from CMake.

Follow-up to 67d88626d44ec04b9e11dca4cfbf62cd29fe9781

Closes #xxxx


CMakeLists.txt
```diff
@@ -1079,9 +1079,6 @@ check_symbol_exists(_strtoi64      "${CURL_INCLUDES}" HAVE__STRTOI64)
 check_symbol_exists(strerror_r     "${CURL_INCLUDES}" HAVE_STRERROR_R)
 check_symbol_exists(siginterrupt   "${CURL_INCLUDES}" HAVE_SIGINTERRUPT)
 check_symbol_exists(getaddrinfo    "${CURL_INCLUDES}" HAVE_GETADDRINFO)
-if(NOT HAVE_GETADDRINFO)
-  set(HAVE_GETADDRINFO_THREADSAFE OFF)
-endif()
```
jay, line 1084: wouldn't it make sense to leave this

vszakats: It wouldn't hurt, but why guard against this twice?

Also without it, our CI wouldn't be really verifying this PR.

jay: I don't see where else it is checked. What I am saying is it wouldn't make sense that HAVE_GETADDRINFO_THREADSAFE is defined when not HAVE_GETADDRINFO.

jay: Ok. IMO it wouldn't hurt to leave it there.

jay (approved):

CMakeLists.txt
vszakats: Fine with me, I'll put it back there.

vszakats: @jay: Rephrasing the Windows logic as this:
```cmake
if(WIN32)
   set(HAVE_GETADDRINFO_THREADSAFE ${HAVE_GETADDRINFO})
 endif()
```
would IMO express our intent much cleaner and self contained in these three lines, with identical results. It would also allow dropping the related line from `WindowsCache.cmake`.

Meaning: When we have `getaddrinfo()`, it is always threadsafe (on Windows).

What do you think?

jay: Works for me

PR #9738, opened 2022-10-15
CI/cirrus: remove superfluous double-quotes and sudo

mback2k:
Follow up to #9565 and #9677

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -142,63 +142,64 @@ macos_task:
   matrix:
     - name: macOS arm64 normal
       env:
-        CFLAGS: "-Wno-vla -mmacosx-version-min=10.9"
+        CFLAGS: -Wno-vla -mmacosx-version-min=10.9
         install_packages: nghttp2
         configure: --without-ssl
     - name: macOS arm64 debug
       env:
-        CFLAGS: "-Wno-vla -mmacosx-version-min=10.9"
+        CFLAGS: -Wno-vla -mmacosx-version-min=10.9
         install_packages: nghttp2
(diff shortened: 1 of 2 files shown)


mback2k:
@pheiduck I don't think sudo is required for the build and test steps after all. I will only be needed for the install step.

pheiduck:
@mback2k As far as I know we used it to parse the `TFLAGS` to test step, but I think it is unnecessary since it is a variable.
We can remove it also on build step.

mback2k:
There is no sudo command to escalate privileges required for build or test steps, just for the install step it is needed.

You had sudo there before, we added just the -E to it with: https://github.com/curl/curl/commit/4896a8d59cfd792cafa790ec16effe7e97fe8a26

Now I am removing it completely and therefore the environment variables are present anyway.

mback2k:
@pheiduck do you have an idea why "macOS arm64 libssh2" is failing here?

pheiduck:
The output say this:
`Warning: Couldn't find a known_hosts file!`

pheiduck:
@mback2k maybe openssh from homebrew will help not sure that the latest version is shipped on macOS 13.

mback2k:
@pheiduck tried the openssh server from Homebrew, but unfortunately that does not seem to help.

I would guess that libssh2 is not modern enough, so the problem is the other way around.

pheiduck:
@mback2k than we can drop this task until libssh2 works w/ M1 Chip.

mback2k:
I think I found a solution, let's see what the CI will say.

PR #9739, opened 2022-10-16
sigv4: set the max string lengths with strlen(), not sizeof()

bagder:
... as otherwise we are off-by-one. Also, change some msnprintf() calls to use sizeof(target buffer) instead of a define to make sure the right target size is used.

Second take at fixing OSS-fuzz issue 52349


bagder:
Never mind

PR #9745, opened 2022-10-16
file: refactor file_do()

emilengler:
This commit refactors the file_do() function, by placing much of the logic into a separate function called fetch_file(), which in turn contains a partially refactored version of the previous code found for fetching files.

The driving force behind that commit is the possibility to add a later feature for a directory listing, such as the one found in FTP.

See #9746

PR #9746, opened 2022-10-16
file: implement support for directory listing 

emilengler:
~Based on #9745~

This commit implements support to the file:// protocol for directory
listings, similar to FTP.

Directory entries are printed as follows:
`MODE    SIZE    MTIME    NAME`

An example directory listing:
```sh
[engler@lennon curl]$ src/curl file:///home/engler
drwxr-xr-x	0	Oct 12 14:09	Desktop
drwxr-xr-x	362	Oct 14 08:31	Downloads
drwxr-xr-x	0	Oct 12 14:09	Templates
drwxr-xr-x	0	Oct 12 14:09	Public
drwxr-xr-x	0	Oct 12 14:09	Documents
drwxr-xr-x	0	Oct 12 14:09	Music
drwxr-xr-x	108	Oct 14 08:34	Pictures
drwxr-xr-x	0	Oct 12 14:09	Videos
drwxr-xr-x	96	Oct 15 17:58	projects
-rw-r--r--	2278067	Oct 13 18:51	wallpaper.jpg
drwxr-xr-x	12	Oct 14 11:21	go
```

### DIFF
--- a/lib/file.c
+++ b/lib/file.c
@@ -50,6 +50,10 @@
 #include <fcntl.h>
 #endif
 
+#ifdef HAVE_DIRENT_H
+#include <dirent.h>
+#endif
+
 #include "strtoofft.h"
 #include "urldata.h"
 #include <curl/curl.h>
@@ -80,6 +84,9 @@
 #else
 #  define open_readonly(p,f) open((p),(f))
 #endif
+
+#define MODE_LEN 11
+#define MTIME_LEN 64
 
 /*
  * Forward declarations.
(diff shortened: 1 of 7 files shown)


lib/file.c
```diff
@@ -380,126 +387,222 @@ static CURLcode file_upload(struct Curl_easy *data)
...
+  result = Curl_gmtime(time(NULL), &now);
+  if(result)
+    return result;
 
-  if(data->set.upload)
-    return file_upload(data);
+  if(tm.tm_year == now.tm_year)
+    strftime(str, MTIME_LEN, "%b %d %H:%M", &tm);
+  else
+    strftime(str, MTIME_LEN, "%b %d %Y", &tm);
```
vszakats, line 438: Would it easier to parse with the year always included? It would also allow parsing without knowing in which year the request was returned.

Unless there is a convention or standard to follow there, parsing may be even easier if using an ISO timestamp here, e.g.: `%Y-%m-%dT%H:%M:%SZ`. With the bonus of having seconds included.

vszakats:
curl FTP file listing (works)[https://everything.curl.dev/libcurl/examples/ftplist] such that it requires an ending slash to trigger it. This ensures that the result is either a file listing or an error (vs. either, depending on the state of the filesystem).

lib/file.c
bagder: I too would prefer a stable syntax, independent of time passing.

> Unless there is a convention or standard to follow

I guess @emilengler tried to make this look like an FTP listing so the convention would then be the format common FTP servers (ls implementations?) show.

emilengler: Done e308e3d8dfccf3ddf5eaa46c5c34dd3be0539dbc

emilengler:
> curl FTP file listing [works](https://everything.curl.dev/libcurl/examples/ftplist) such that it requires an ending slash to trigger it. This ensures that the result is either a file listing or an error (vs. either, depending on the state of the filesystem). It'd be useful to replicate that here as well IMO.

This is a good suggestion. I'll fix that.

gvanem:
What about the targets (like MSVC) that does not have `<dirent.h>` ?

emilengler:
> What about the targets (like MSVC) that does not have `<dirent.h>` ?

Nothing will happen in these cases. There are preprocessor conditions that check for the existence of `fdopendir` and `<dirent.h>`. However, I am unsure if the current way of ignoring is good practice. Maybe `fetch_file` should be called in that case, just like it happened earlier?

bagder:
The fuzzer found problems. This now has merge conflicts.

bagder:
This PR has conflicts and still has test failures.

dfandrich:
Does this patch support --list-only as well?  That would simplify one use case I have.

emilengler:
I haven't forgot the PR yet, but I'm currently very busy. It may take until christmas holidays, until I find some free time again :/

emilengler:
Rebased to 99f72c01c214124ee2b18f8c7184215cfc2609c3

emilengler:
> Does this patch support --list-only as well? That would simplify one use case I have.

Not yet, but it seems like a good suggestion. I will first try to investigate the CI and fuzzer fails though.

PR #9750, opened 2022-10-17
CURLOPT_AUTOREFERER: default to origin-only

bagder:
Automatic referers now default to only setting the "origin" (scheme + host + port). For privacy, especially when redirecting to non-secure schemes.

CURLOPT_AUTOREFERER now accepts the value '2' to force it to also include path and query in referer header fields.

A future improvement might be to support the Referrer-Policy header and rather act according to the server's will.

- [x] use origin-only by default
- [ ] add test case for referer-full
- [ ] provide command line option for referer-full

### DIFF
--- a/docs/TODO
+++ b/docs/TODO
@@ -77,6 +77,7 @@
  5.5 auth= in URLs
  5.6 alt-svc should fallback if alt-svc does not work
  5.7 Require HTTP version X or higher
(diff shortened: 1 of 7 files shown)


bagder:
After feedback on the mailing list, this is not the way forward.

PR #9755, opened 2022-10-17
Check for HAVE_STDATOMIC_H in easy_lock

donny-dont:
The check for `HAVE_STDATOMIC_H` looks to see if the `stdatomic.h` header is present.

### DIFF
--- a/lib/easy_lock.h
+++ b/lib/easy_lock.h
@@ -47,7 +47,7 @@ typedef PVOID SRWLOCK, *PSRWLOCK;
 #define curl_simple_lock_lock(m) AcquireSRWLockExclusive(m)
 #define curl_simple_lock_unlock(m) ReleaseSRWLockExclusive(m)
 
-#elif defined (HAVE_ATOMIC)
+#elif defined(HAVE_ATOMIC) && defined(HAVE_STDATOMIC_H)
 #include <stdatomic.h>
 #if defined(HAVE_SCHED_YIELD)
 #include <sched.h>


mback2k:
Please don't remove the check for HAVE_ATOMIC as that is required for the feature to work in general and implies HAVE_STDATOMIC_H via autotools/CMake.

bagder:
Does this fix a problem or why this change? As @mback2k says, I don't think the current approach is wrong.

donny-dont:
If its the case that `HAVE_ATOMIC` means `HAVE_STDATOMIC_H` that's simply not true. On PlayStation 4 I'm getting `HAVE_ATOMIC` as `true` from the CMake build, but `HAVE_STDATOMIC_H` is false so easy_lock does not compile.

All the curl functions in [easy_lock.h](https://github.com/curl/curl/blob/master/lib/easy_lock.h#L77-L99) rely on atomic functions defined in [stdatomic.h](https://en.cppreference.com/w/c/atomic) so `HAVE_STDATOMIC_H` looks to be the correct guard.

The only other case of `HAVE_ATOMIC` in the codebase is in [CurlTests.c](https://github.com/curl/curl/blob/master/CMake/CurlTests.c#L497-L515) which guards `HAVE_STDATOMIC_H`.

mback2k:
> If its the case that `HAVE_ATOMIC` means `HAVE_STDATOMIC_H` that's simply not true. On PlayStation 4 I'm getting `HAVE_ATOMIC` as `true` from the CMake build, but `HAVE_STDATOMIC_H` is false so easy_lock does not compile.
> 
> All the curl functions in [easy_lock.h](https://github.com/curl/curl/blob/master/lib/easy_lock.h#L77-L99) rely on atomic functions defined in [stdatomic.h](https://en.cppreference.com/w/c/atomic) so `HAVE_STDATOMIC_H` looks to be the correct guard.

That is true and I guess the correct approach would be to check for **both** `HAVE_ATOMIC` **and** `HAVE_STDATOMIC_H` then as the test (see below) can also work without `HAVE_STDATOMIC_H` as you experienced.
> 
> The only other case of `HAVE_ATOMIC` in the codebase is in [CurlTests.c](https://github.com/curl/curl/blob/master/CMake/CurlTests.c#L497-L515) which guards `HAVE_STDATOMIC_H`.

That is the place there the functional test for atomic support is done. `HAVE_ATOMIC` is defined during the test and if it compiles it will be defined for the curl build.

donny-dont:
Updated to have both checks.

bagder (approved):

bagder:
Thanks!

PR #9756, opened 2022-10-18
schannel: Don't reset recv/send function pointers on renegotiation

### LINKED ISSUE #9451: Windows nghttp2 + schannel + tls 1.3 error: Received HTTP/0.9 when not allowed
### I did this
Compiling curl with nghttp2 1.49 and schannel support with MSVC is causing TLS 1.3 connections to fail with the error `Unsupported protocol` (CURLE_UNSUPPORTED_PROTOCOL) with the details `Received HTTP/0.9 when not allowed`. This doesn't happen 100% of the time, but it almost always happens for me.

This only happens on Windows 11 or Windows Server 2022. Windows 10 does not seem affected.

Bisecting, this started with #8419.

Roughly I did:
1. Built nghttp2 with cmake
2. Built curl with cmake with options:
   * `-DCURL_USE_SCHANNEL=ON`
   * `-DUSE_NGHTTP2=ON`
   * `-DNGHTTP2_INCLUDE_DIR=...`
   * `-DNGHTTP2_LIBRARY=...`
3. Run `curl.exe -O https://github.com/curl/curl/releases/download/curl-7_85_0/curl-7.85.0.zip` (may need to run a few times)

The exact URL doesn't seem too important, as long as the server supports 1.3. It seems to fail more readily with larger files (very small files like https://example.com/ don't fail quite as often).

```
C:\Temp\curl\lib\Debug>..\..\src\Debug\curl.exe -O  https://github.com/curl/curl/releases/download/curl-7_85_0/curl-7.85.0.zip
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (1) Received HTTP/0.9 when not allowed
```

It also happens with really trivial programs like `simple.c` from the examples directory.  Lowering the max TLS version below 1.3 makes the problem go away.

### I expected the following
Should not fail.

### curl/libcurl version
Latest tested version is 5390c9d77a5df2f2bc3d3fd8f8c97dead890b950. Seems to be happening with any version since #8419.

```
curl 7.86.0-DEV (Windows) libcurl/7.86.0-DEV Schannel nghttp2/1.49.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS HSTS HTTP2 IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI threadsafe UnixSockets
```

### operating system

Windows 11 22000.856

JDepooter:
Fixes #9451 

These function pointers will have been set when the initial TLS handshake was completed. If they are unchanged, there is no need to set them again. If they have been changed, as is the case with HTTP/2, we don't want to override that change. That would result in the http2_recv/send functions being completely bypassed.

It should be investigated whether the same change should be made to the other TLS backends. They don't seem to have the same amount of renegotiation logic that SChannel does. It appears that OpenSSL/BoringSSL may outright forbid TLS renegotiation.

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -1935,8 +1935,7 @@ schannel_connect_common(struct Curl_easy *data, struct connectdata *conn,
 
   if(ssl_connect_done == connssl->connecting_state) {
     connssl->state = ssl_connection_complete;
-    if(!connssl->backend->recv_renegotiating)
-    {
+    if(!connssl->backend->recv_renegotiating) {
       /* On renegotiation, we don't want to reset the existing recv/send
        * function pointers. They will have been set after the initial TLS
        * handshake was completed. If they were subsequently modified, as
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -1935,8 +1935,16 @@ schannel_connect_common(struct Curl_easy *data, struct connectdata *conn,
 
   if(ssl_connect_done == connssl->connecting_state) {
     connssl->state = ssl_connection_complete;
-    conn->recv[sockindex] = schannel_recv;
-    conn->send[sockindex] = schannel_send;
+    if(!connssl->backend->recv_renegotiating)
+    {
+      /* On renegotiation, we don't want to reset the existing recv/send
+       * function pointers. They will have been set after the initial TLS
+       * handshake was completed. If they were subsequently modified, as
+       * is the case with HTTP/2, we don't want to override that change.
+       */
+      conn->recv[sockindex] = schannel_recv;
+      conn->send[sockindex] = schannel_send;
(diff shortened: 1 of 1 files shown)


bagder:
~~~
./vtls/schannel.c:1939:5: warning: badly placed open brace (BRACEPOS)
     {
     ^

JDepooter:
> ```
> ./vtls/schannel.c:1939:5: warning: badly placed open brace (BRACEPOS)
>      {
>      ^
> ```

Sorry, forgot to check formatting. I've pushed a new commit fixing this.

jay:
Thanks!

PR #9757, opened 2022-10-18
CURLOPT_AUTOREFERER.3: highlight the privacy leak risk

bagder:


vszakats:
I happen to use this option to retain a meaningful source URL when downloading certain files. GitHub ones are maybe the most trivial examples:

```
$ curl --disable -LO --xattr https://github.com/curl/curl/releases/download/curl-7_85_0/curl-7.85.0.tar.xz
$ xattr -l curl-7.85.0.tar.xz
user.xdg.origin.url: https://objects.githubusercontent.com/github-production-release-asset-2e65be/
569041/ce852[...]a60?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AK[...]%2F20221018
%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20221018T[...]Z&X-Amz-Expires=300
&X-Amz-Signature=[...]&X-Amz-SignedHeaders=host&actor_id=0&key_id=0&repo_id=569041
&response-content-disposition=attachment%3B%20filename%3Dcurl-7.85.0.tar.xz
&response-content-type=application%2Foctet-stream
```

That is a temporary URL, so unusable for re-download, and non-trivial even to find out which repo it came from.

Workaround:
```
$ curl --disable --referer ';auto' -LO --xattr https://github.com/curl/curl/releases/download/curl-7_85_0/curl-7.85.0.tar.xz
$ xattr -l curl-7.85.0.tar.xz
user.xdg.origin.url: [...as above...]
user.xdg.referrer.url: https://github.com/curl/curl/releases/download/curl-7_85_0/curl-7.85.0.tar.xz
```

The referrer now points to the original URL.

This is a rather bad solution with unwanted side-effects, loss-of-privacy being just one of them.

This use-case might be fixed by storing the original URL (after unglobbing) to `user.xdg.origin.url`. By default, or with an extra option?

bagder:
> This use-case might be fixed by storing the original URL

Maybe that is actually the more correct action, as that is after all the URL you used in the command line...

Slightly unrelated to this PR though. I will proceed and merge this. Feel free to create an issue about your remark and we can work on it.

Issue #1455, opened 2017-04-28
CURLOPT_FTP_SKIP_PASV_IP unusable


Zenju:
Using CURLOPT_FTP_SKIP_PASV_IP requires knowledge if the IP as returned in a server's response to PASV is routable or not. Often this information is not available before actually trying to make a connection: If the IP is not routable (e.g. 192.168.1.1) then libcurl will hang. Therefore simply falling back to adding CURLOPT_FTP_SKIP_PASV_IP only in the failure case is not a good solution. Conversely always using CURLOPT_FTP_SKIP_PASV_IP is also premature.
Other FTP clients solve this by doing a simple check if the IP in PASV's response is routable or not and only if it is not, fall back to the IP of the control connection.
Is there a reason why libcurl does not support this or has just nobody bothered to code it yet? For IPv4 it's almost trivial. Heck, I'd do it myself even, because I need it! :)
Option could be named: CURLOPT_FTP_SKIP_PASV_IP_IF_NOT_ROUTABLE

bagder:
> Using CURLOPT_FTP_SKIP_PASV_IP requires knowledge if the IP as returned in a server's response to PASV is routable or not

No, that's what what it requires. It basically assumes that the returned IP is useless information and that the client can connect back again to the same IP it is already connected. That's a rather bold assumption, but many FTP servers work this way still and this is how the option is documented to work so I see no fundamental problem with it.

> If the IP is not routable (e.g. 192.168.1.1) then libcurl will hang

1. If you set `CURLOPT_FTP_SKIP_PASV_IP`, the returned IP would be *skipped* and not cause a hang.
2. Trying an IP like that will in many cases not cause a hang but a rather quick failed connection

But sure, it can be a bit annoying. Broken servers are annoying.

> Other FTP clients solve this by doing a simple check if the IP in PASV's response is routable or not and only if it is not, fall back to the IP of the control connection.

How does a client determine that? And secondly, private IP addresses like that are often used in LANs so they *are* routable and fine, just plain wrong to connect back to if the FTP server is remote and responds with such an address.

> Is there a reason why libcurl does not support this or has just nobody bothered to code it yet?

The latter. But as I mention above, being "routable" doesn't mean it works. It would probably work even better if there was a fail-safe mode that would go back and try the original IP if the one used from the response doesn't work...

Zenju:
> No, that's what what it requires

[...]

PR #1470, opened 2017-05-05
Enhance CURLOPT_FTP_SKIP_PASV_IP

Zenju:
In reference to: 
https://github.com/curl/curl/issues/1455

PR #9758, opened 2022-10-18
Enhance CURLOPT_FTP_SKIP_PASV_IP

tenzap:
This is a refresh of https://github.com/curl/curl/pull/1470
I took the original authors PR, squashed it, and tried to refresh it towards the latest version of curl

URL_FTP_SKIP_PASV_IP_ALWAYS instructs libcurl to ignore the IP address the server suggests in its 227-response to libcurl's PASV command when libcurl connects the data connection. Instead libcurl will re-use the same IP address it already uses for the control connection. But it will use the port number from the 227-response.

The allowed values are:

- CURL_FTP_SKIP_PASV_IP_NEVER Always use the IP returned by the server.

- CURL_FTP_SKIP_PASV_IP_ALWAYS Always ignore the IP returned by the server.

- CURL_FTP_SKIP_PASV_IP_IF_NOT_ROUTABLE Only ignore the returned IP if it is within the private address space (see RFC1918), which is a common server configuration error.

Closes: #1455


tenzap:
@bagder maybe ?

src/tool_getparam.c
```diff
@@ -1198,7 +1198,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
         config->ignorecl = toggle;
         break;
       case 'q': /* --ftp-skip-pasv-ip */
-        config->ftp_skip_ip = toggle;
+        config->ftp_pasvp_ip_rule = CURL_FTP_SKIP_PASV_IP_ALWAYS;
```
bagder, line 1201: this breaks behavior for users who do `--no-ftp-skip-pasv-ip` to switch off the option

src/tool_cfgable.h
```diff
@@ -177,7 +177,8 @@ struct OperationConfig {
   bool doh_verifystatus;
   bool create_dirs;
   bool ftp_create_dirs;
-  bool ftp_skip_ip;
+  long ftp_pasvp_ip_rule;   /* how to handle the IP address the FTP server
+                               passes on to us */
```
bagder, line 181: I think you should keep treating this as a boolean cmdline option

docs/libcurl/symbols-in-versions
```diff
@@ -38,6 +38,9 @@ CURL_FORMADD_NULL               7.9.8         7.56.0
 CURL_FORMADD_OK                 7.9.8         7.56.0
 CURL_FORMADD_OPTION_TWICE       7.9.8         7.56.0
 CURL_FORMADD_UNKNOWN_OPTION     7.9.8         7.56.0
+CURL_FTP_SKIP_PASV_IP_ALWAYS          7.86.0
+CURL_FTP_SKIP_PASV_IP_IF_NOT_ROUTABLE 7.86.0
+CURL_FTP_SKIP_PASV_IP_NEVER           7.86.0
```
bagder, line 43: This cannot be merged until 7.87.0

docs/libcurl/opts/CURLOPT_FTP_SKIP_PASV_IP.3
```diff
@@ -45,7 +52,7 @@ abuse by malicious servers.
 
 This option has no effect if PORT, EPRT or EPSV is used instead of PASV.
 .SH DEFAULT
-1 since 7.74.0, was 0 before then.
+CURL_FTP_SKIP_PASV_IP_ALWAYS
```
bagder, line 55: This removes the information about the historic change and I think it should not do that.

docs/libcurl/opts/CURLOPT_FTP_SKIP_PASV_IP.3
```diff
@@ -29,14 +29,21 @@ CURLOPT_FTP_SKIP_PASV_IP \- ignore the IP address in the PASV response
...
+the data connection. Instead libcurl will re-use the same IP address it already uses
+for the control connection. But it will use the port number from the 227-response.
+The allowed values are:
+.IP CURL_FTP_SKIP_PASV_IP_NEVER
+Always use the IP returned by the server.
+.IP CURL_FTP_SKIP_PASV_IP_ALWAYS
+Always ignore the IP returned by the server.
+.IP CURL_FTP_SKIP_PASV_IP_IF_NOT_ROUTABLE
+Only ignore the returned IP if it is within the private address space (see RFC1918),
+which is a common server configuration error.
```
bagder, line 46: I think this should also mention from which version this option works.

include/curl/curl.h
```diff
@@ -2202,6 +2202,12 @@ typedef enum {
   /* Convenient "aliases" */
 #define CURLOPT_RTSPHEADER CURLOPT_HTTPHEADER
 
+  /*parameters for the CURLOPT_FTP_SKIP_PASV_IP option */
```
bagder, line 2205: A space after `/*` perhaps?

bagder (commented):
I think we need a test case for this as well, verifying that it works

tenzap:
@Zenju now that I squashed and refreshed your commits, would you please act on the comments?

src/tool_cfgable.h
tenzap: original author permits 3 values, so I don't think bool would fit here.

docs/libcurl/symbols-in-versions
tenzap: changed to 7.87

docs/libcurl/opts/CURLOPT_FTP_SKIP_PASV_IP.3
tenzap: updated

docs/libcurl/opts/CURLOPT_FTP_SKIP_PASV_IP.3
tenzap: updated, but I guess @zenju has better knowledge of what he wanted to do with this 
PR

include/curl/curl.h
tenzap: done

bagder (changes requested):
this breaks existing behavior

bagder:
This can't be merged and no response in weeks. I would also like a stronger motivation why we want/need this feature. Letting the server decide IP address is a security risk.

PR #9760, opened 2022-10-18
hostip: guard PF_INET6 use

vszakats:
Some platforms (e.g. Amiga OS) do not have `PF_INET6`. Adjust the code for these.

```
hostip.c: In function 'fetch_addr':
hostip.c:308:12: error: 'PF_INET6' undeclared (first use in this function)
       pf = PF_INET6;
            ^~~~~~~~
```

Regression from 1902e8fc511078fb5e26fc2b907b4cce77e1240d

Closes #xxxx

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -303,8 +303,10 @@ static struct Curl_dns_entry *fetch_addr(struct Curl_easy *data,
     bool found = false;
     struct Curl_addrinfo *addr = dns->addr;
 
+#ifdef PF_INET6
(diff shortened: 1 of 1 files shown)


vszakats:
The initial patch fixes the compiler error, but the initial commit is most likely not the correct fix. Other variations may be using `ENABLE_IPV6` instead, or excluding a larger block of code in this case. This came while cross-compiling for Amiga OS.

bagder (approved):

PR #9768, opened 2022-10-19
curl: make --xattr save the original URL

bagder:


vszakats:
True. Well, auto-referrer might trigger to save the effective URL to the origin xattr, but this seems way more trouble than it's worth.

Perhaps the best solution is to just not use automatic referrers.

[ Off-topic: A generic solution could be a `--write-out`-like option where the user could customize what information to write to which xattr, e.g. `--xattr-out 'user.xdg.origin.url=%{url_effective}'`. New variables, like download timestamp or hash can be added to that. `--xattr` would trigger the default set of these. ]

bagder:
@mback2k the Appveyor builds now all appear as failed, but looking at the details there is no link to figure out more!

It appears as if it still shows the previous rounds build failure while in fact it has restarted a new build after the PR was updated. Can this be it?

bagder:
Right, the new test does not work on Windows...

vszakats:
We should just disable these tests for Windows. On Windows there are no xattrs. There is a similar feature called [file streams](https://learn.microsoft.com/en-us/windows/win32/fileio/file-streams), but there is no `fsetxattr()` API for it.

bagder:
I figure there are more platforms than Windows so I tried to make it a proper feature for the test suite to require.

mback2k:
I will look at the AppVeyor issue tonight.

bagder:
Now, how do I add that extra include path for `disabled` when built with cmake.... ?

PR #9772, opened 2022-10-19
ftp: fix for FTP growing file support

fractal-access:
When using the option CURLOPT_IGNORE_CONTENT_LENGTH (set.ignorecl in code) to support growing files in FTP, the code should ignore the initial size it gets from the server as this will not be the final size of the file. This is done in ftp_state_quote() to prevent a size request being issued in the initial sequence. However, in a later call to ftp_state_get_resp() the code attempts to get the size of the content again if it doesn't already have it, by parsing the response from the RETR request. This fix prevents this parsing of the response to get the size when the set.ignorecl option is set. This should maintain the size value as -1, unknown, in this situation.

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -2465,6 +2465,7 @@ static CURLcode ftp_state_get_resp(struct Curl_easy *data,
 
     if((instate != FTP_LIST) &&
        !data->state.prefer_ascii &&
+       !data->set.ignorecl &&
        (ftp->downloadsize < 1)) {
       /*
        * It seems directory listings either don't show the size or very
--- a/tests/FILEFORMAT.md
+++ b/tests/FILEFORMAT.md
@@ -307,6 +307,7 @@ about to issue.
    appear at once when a file is transferred
 - `RETRNOSIZE` - Make sure the RETR response doesn't contain the size of the
   file
+- `RETRSIZE [size]` - Force RETR response to contain the specified size
 - `NOSAVE` - Don't actually save what is received
 - `SLOWDOWN` - Send FTP responses with 0.01 sec delay between each byte
 - `PASVBADIP` - makes PASV send back an illegal IP in its 227 response
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -67,7 +67,7 @@ test370 test371 test372 test373 test374 test375 test376 test378 test379 \
 test380 test381 test383 test384 test385 test386 test387 test388 test389 \
 test390 test391 test392 test393 test394 test395 test396 test397 test398 \
 test399 test400 test401 test402 test403 test404 test405 test406 test407 \
-test408 test409 test410 test411 test412 test413 test414 test415 \
+test408 test409 test410 test411 test412 test413 test414 test415 test416 \
 \
 test430 test431 test432 test433 test434 test435 test436 \
 \
--- a/tests/data/test416
+++ b/tests/data/test416
@@ -1,0 +1,51 @@
+<testcase>
+<info>
+<keywords>
+FTP
+EPSV
+RETR
+Range
+</keywords>
+</info>
+# Server-side
+<reply>
+<data>
+0123456789abcdef0123456789abcdef0123456789abcdef
+</data>
+<datacheck>
+0123456789abcdef0123456789abcdef0123456789abcdef
+</datacheck>
+<servercmd>
+RETRSIZE 7
+</servercmd>
+<size>
+8
+</size>
+</reply>
+
+# Client-side
+<client>
+<server>
+ftp
+</server>
+ <name>
+FTP growing file support
+ </name>
+ <command>
+--ignore-content-length ftp://%HOSTIP:%FTPPORT/%TESTNUMBER
+</command>
+</client>
+
+# Verify data after the test has been "shot"
+<verify>
+<protocol>
+USER anonymous
+PASS ftp@example.com
+PWD
+EPSV
+TYPE I
+RETR %TESTNUMBER
+QUIT
+</protocol>
+</verify>
+</testcase>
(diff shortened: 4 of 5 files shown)


fractal-access:
Testing
=====
To find this issue and test I have done the following. I created some code to slow copy a file at 5 KB/s. I then use a standard curl_easy_perform() to get this file from an FTP server while it grows. This will cause libcurl to retrieve the amount of file based on the size it gets at the start of the interaction and raise and error due to getting a, 450 Transfer aborted, error from the server. I then set the option:

curl_easy_setopt(m_curlHandle, CURLOPT_IGNORE_CONTENT_LENGTH, 1L)

I would expect this to now get the complete file, but it fails in the same way as before. After applying the fix, libcurl continues reading the data connection until the server indicates the end. I then get the full file after it has finished being slow copied.

For my tests I used the ProFTPD server with a tailmode filter addition to support growing files. The main function of this is to not automatically finish the transfer if the server reaches eof, but to also make sure the file hasn't been updated within a timeout period. I think this could be tested against a standard FTP server if the bandwidth option is also used in libcurl and set to a value smaller than the growing file rate.

I not sure how easy it would be to get this type of test into the libcurl test suite, but if anyone has any thoughts on this please let me know.

dfandrich:
> I not sure how easy it would be to get this type of test into the libcurl test
> suite, but if anyone has any thoughts on this please let me know.

How about returning a small size in response to the SIZE command, but actually return a much longer file, as though it were added to during the transfer? test135 is probably a good one to use as a template for this.

fractal-access:
> I not sure how easy it would be to get this type of test into the libcurl test suite, but if anyone has any thoughts on this please let me know.
> How about returning a small size in response to the SIZE command, but actually return a much longer file, as though it were added to during the transfer? test135 is probably a good one to use as a template for this.

That's a good idea, thanks for the suggestion.

bagder (changes requested):
Needs a test.

fractal-access:
I have attempted to make a test which hopefully is acceptable, but please check.

I added a new test at the first available number, 416, and based it off test 135.

After looking into this I found to test the real situation I needed to get the test FTP server to reply with a smaller size than the final size of the data in the RETR command.
I wasn't sure of the best way to do this. I've gone for adding another value in the <servercmd> section. I have added RETRSIZE [size], that when set will explicitly set the value of the size returned by ftpserver.pl. In ftpserver.pl, I have tried to follow the scheme for other value settings.

bagder:
The `REPLY` command already exists to provide a custom response for a given FTP command. No need to invent another command for the same thing!

fractal-access:
Thanks for the suggestion. As far as I can tell REPLY will fake the response but doesn't actually perform the request. The test needs the fake size and also the data to be sent. Is there a configuration using the current parameters that allows for the fake response as well as actually getting the request performed?

bagder:
I figured you just want to provide a regular request/transfer and send X bytes a response but have the SIZE response say the data is Y bytes, where Y is smaller than X. The way a transfer of a growing file would appear.

fractal-access:
The implementation of --ignore-content-length/CURLOPT_IGNORE_CONTENT_LENGTH is that it skips the SIZE command in the standard libcurl FTP interaction. Because of this it ends up getting the size from the parse of the return from the RETR command. Therefore, the test needs to send a smaller length value back to the RETR request to mimic the interaction.

bagder:
Yes, which is why you make the RETR response say size X while the response returns size Y.

fractal-access:
Yes, that's correct. Can you think of a way to do this without having to introduce a new parameter?

bagder:
Looking into this made me question the need for this PR. The functionality already exists and seems to work!

Test 1137 already tests FTP with `--ignore-content-length`. How is that not enough?

fractal-access:
Growing files don't currently work on the latest libcurl version. This can been seen by looking at the code, where it can be seen that the size will still be set by parsing the return from the RETR command. It can also be shown by running the new test against the current version, which will fail. 

Test 1137 uses --ignore-content-length, but does not test growing files.

bagder:
I disagree.

The current code already works with growing files if you use `--ignore-content-length`.

It does this by avoiding the `SIZE` command all together and thus it does not know how large the file is and it only uses the disconnect as a signal to when the file ends.

Test 1137 verifies this. You can see that it does not send a SIZE, and when it doesn't know the size upfront, it also does not know if the transferred files grows or shrinks or whatever. curl will just accepts what it gets and consider the file complete when connection is closed.

If you look at what curl does when your test 416 is run, you will see that your extra work for the ftp server is not even used.

fractal-access:
There has never been any doubt that --ignore-content-length turns off the SIZE request. The issue is that curl will also establish the size by parsing the length out of the RETR request return. Do you think this is not the case? If so, can you explain what the code does from line 2457 of ftp.c? 

This can also be seen with a simple test with verbose turned on. The size curl has is printed at 2515:

 infof(data, "Getting file with size: %" CURL_FORMAT_CURL_OFF_T,
             size);

If --ignore-content-length worked this should print out -1, agreed? 

Try something like:

curl ftp://hlm:hlm@maudv03//highres/nd/destination.mxf --ignore-content-length -v

which gives:

Getting file with size: 1024000
Not
Getting file with size: -1

bagder:
Ah yes, I forgot about that weirdness. I would say that the parsing of that response should also be skipped if this option is set. Maybe just like this?
~~~diff
diff --git a/lib/ftp.c b/lib/ftp.c
index f72d803bb..e626bc829 100644
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -2444,10 +2444,11 @@ static CURLcode ftp_state_get_resp(struct Curl_easy *data,
      *
      * Debugging help from Salvatore Sorrentino on February 26, 2003.
      */
 
     if((instate != FTP_LIST) &&
+       !data->set.ignorecl &&
        !data->state.prefer_ascii &&
        (ftp->downloadsize < 1)) {
       /*
        * It seems directory listings either don't show the size or very
        * often uses size 0 anyway. ASCII transfers may very well turn out

bagder (approved):
After I've slowly recalled how this works. I think this is the way to do it.

bagder:
Thanks!

PR #9778, opened 2022-10-21
noproxy: fix builds without AF_INET6

vszakats:
Regression from 1e9a538e05c0107c54ef81d9de7cd0b27cd13309

Closes #xxxx

---

This fixes the build error, but it's not necessarily the best fallback.


bagder (approved):
Thanks, I forgot that adjustment...

PR #9780, opened 2022-10-21
docs: make sure libcurl opts examples pass in long arguments

### LINKED ISSUE #9779: Bugs in all curl_set_opt calls with enum arguments as values.
### I did this

Let's consider the example with [CURLOPT_SSLVERSION](https://curl.se/libcurl/c/CURLOPT_SSLVERSION.html).

#### The function synopsis
```c
CURLcode curl_easy_setopt(CURL *curl, CURLoption option, ...);
```
#### The call semantics
```c
CURLcode curl_easy_setopt(CURL *handle, CURLOPT_SSLVERSION, long version);
```
#### [Example](https://curl.se/libcurl/c/CURLOPT_SSLVERSION.html)
```c
curl_easy_setopt(curl, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1);
```
#### Reading the value (shortened)
```c
  case CURLOPT_SSLVERSION:
      arg = va_arg(param, long);
```
The issue on all 64 bit OSes except Windows: the enum value `CURL_SSLVERSION_TLSv1` promoted to `int` (32 bit) is used for the call to the variable arg function, the value is read as `long` (64 bit), a possible buffer over-read vulnerability.

### I expected the following

enum values are passed as long values.

### curl/libcurl version

libcurl 7.84.0

### operating system

Ubuntu 22.04.1, macOS 12.4

bagder:
Reported-by: Sergey
Fixes #9779

### DIFF
--- a/docs/libcurl/opts/CURLINFO_CONDITION_UNMET.3
+++ b/docs/libcurl/opts/CURLINFO_CONDITION_UNMET.3
@@ -52,7 +52,8 @@ if(curl) {
   curl_easy_setopt(curl, CURLOPT_TIMEVALUE, 1577833200L);
 
   /* If-Modified-Since the above time stamp */
-  curl_easy_setopt(curl, CURLOPT_TIMECONDITION, CURL_TIMECOND_IFMODSINCE);
+  curl_easy_setopt(curl, CURLOPT_TIMECONDITION,
+                   (long)CURL_TIMECOND_IFMODSINCE);
 
   /* Perform the request */
   res = curl_easy_perform(curl);
--- a/docs/libcurl/opts/CURLOPT_ALTSVC_CTRL.3
+++ b/docs/libcurl/opts/CURLOPT_ALTSVC_CTRL.3
@@ -75,7 +75,7 @@ HTTPS
 .nf
 CURL *curl = curl_easy_init();
 if(curl) {
-  curl_easy_setopt(curl, CURLOPT_ALTSVC_CTRL, CURLALTSVC_H1);
+  curl_easy_setopt(curl, CURLOPT_ALTSVC_CTRL, (long)CURLALTSVC_H1);
   curl_easy_setopt(curl, CURLOPT_ALTSVC, "altsvc-cache.txt");
   curl_easy_perform(curl);
 }
--- a/docs/libcurl/opts/CURLOPT_FTPSSLAUTH.3
+++ b/docs/libcurl/opts/CURLOPT_FTPSSLAUTH.3
@@ -54,7 +54,7 @@ if(curl) {
   curl_easy_setopt(curl, CURLOPT_URL, "ftp://example.com/file.txt");
   curl_easy_setopt(curl, CURLOPT_USE_SSL, CURLUSESSL_TRY);
   /* funny server, ask for SSL before TLS */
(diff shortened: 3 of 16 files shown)


jay:
I think we discussed this at one point and determined that because of the way the 32-bit arguments are passed it was technically not necessary. The only way I see this possibly being a problem is if the arch type is 32-bits and the long type is 64-bits. In that case, rather than typecast (long)CURLVAL we could ditch the enums for macros, for example

~~~diff
diff --git a/include/curl/curl.h b/include/curl/curl.h
index e28dd0b..19910a6 100644
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -2202,21 +2202,22 @@ typedef enum {
   /* Convenient "aliases" */
 #define CURLOPT_RTSPHEADER CURLOPT_HTTPHEADER
 
-  /* These enums are for use with the CURLOPT_HTTP_VERSION option. */
-enum {
-  CURL_HTTP_VERSION_NONE, /* setting this means we don't care, and that we'd
-                             like the library to choose the best possible
-                             for us! */
-  CURL_HTTP_VERSION_1_0,  /* please use HTTP 1.0 in the request */
-  CURL_HTTP_VERSION_1_1,  /* please use HTTP 1.1 in the request */
-  CURL_HTTP_VERSION_2_0,  /* please use HTTP 2 in the request */
-  CURL_HTTP_VERSION_2TLS, /* use version 2 for HTTPS, version 1.1 for HTTP */
-  CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE,  /* please use HTTP 2 without HTTP/1.1
-                                           Upgrade */
-  CURL_HTTP_VERSION_3 = 30, /* Makes use of explicit HTTP/3 without fallback.
-                               Use CURLOPT_ALTSVC to enable HTTP/3 upgrade */
-  CURL_HTTP_VERSION_LAST /* *ILLEGAL* http version */
-};
+  /* macros for use with the CURLOPT_HTTP_VERSION option. */
+#define CURL_HTTP_VERSION_NONE 0L  /* setting this means we don't care, and
+                                      that we'd like the library to choose the
+                                      best possible for us! */
+#define CURL_HTTP_VERSION_1_0 1L   /* please use HTTP 1.0 in the request */
+#define CURL_HTTP_VERSION_1_1 2L   /* please use HTTP 1.1 in the request */
+#define CURL_HTTP_VERSION_2_0 3L   /* please use HTTP 2 in the request */
+#define CURL_HTTP_VERSION_2TLS 4L  /* use version 2 for HTTPS, version 1.1 for
+                                      HTTP */
+#define CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE 5L  /* please use HTTP 2 without
+                                                   HTTP/1.1 Upgrade */
+#define CURL_HTTP_VERSION_3 30L    /* Makes use of explicit HTTP/3 without
+                                      fallback. Use CURLOPT_ALTSVC to enable
+                                      HTTP/3 upgrade */
+#define CURL_HTTP_VERSION_LAST 31L  /* *ILLEGAL* http version */
+
 
 /* Convenience definition simple because the name of the version is HTTP/2 and
    not 2.0. The 2_0 version of the enum name was set while the version was
~~~

bagder:
We can probably do this, but since people use the docs even for older versions of curl it seems reasonable to keep the typecasts in there for a while.

I cannot think of any ABI breakage caused by switching from enums to defines like proposed above.

PR #9781, opened 2022-10-22
windows: fail early with a missing windres in autotools

vszakats:
`windres` is not always auto-detected by autotools when building for Windows. When this happened, the build failed with a confusing error due to the empty `RC` command:

```
/bin/bash ../libtool --tag=RC --mode=compile  -I../include -DCURL_EMBED_MANIFEST  -i curl.rc -o curl.o
[...]
Usage: /sandbox/curl/libtool [OPTION]... [MODE-ARG]...
Try 'libtool --help' for more information.
libtool:   error: unrecognised option: '-I../include'
```

Improve this by verifying if `RC` is set, and fail with a clear error otherwise.

Follow-up to 6de7322c03d5b4d91576a7d9fc893e03cc9d1057

Ref: https://curl.se/mail/lib-2022-10/0049.html
Reported-by: Thomas Glanzmann
Closes #9781

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -570,7 +570,13 @@ CURL_DARWIN_CFLAGS
 CURL_DARWIN_SYSTEMCONFIGURATION
 CURL_SUPPORTS_BUILTIN_AVAILABLE
 
-AM_CONDITIONAL([OS_WINDOWS], [test "$curl_cv_native_windows" = "yes"])
+AM_CONDITIONAL([HAVE_WINDRES],
+  [test "$curl_cv_native_windows" = "yes" && test -n "${RC}"])
+
+if test "$curl_cv_native_windows" = "yes"; then
+  AM_COND_IF([HAVE_WINDRES],,
+    [AC_MSG_ERROR([windres not found in PATH. Windows builds require windres. Cannot continue.])])
+fi
 
 dnl ************************************************************
 dnl switch off particular protocols
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -109,7 +109,7 @@ endif
 if USE_CPPFLAG_CURL_STATICLIB
 libcurl_la_CPPFLAGS_EXTRA += -DCURL_STATICLIB
 else
-if OS_WINDOWS
+if HAVE_WINDRES
 libcurl_la_SOURCES += $(LIB_RCFILES)
 $(LIB_RCFILES): $(top_srcdir)/include/curl/curlver.h
 endif
@@ -154,7 +154,7 @@ tidy:
 optiontable:
 	perl optiontable.pl < $(top_srcdir)/include/curl/curl.h > easyoptions.c
 
-if OS_WINDOWS
+if HAVE_WINDRES
 .rc.lo:
 	$(LIBTOOL) --tag=RC --mode=compile $(RC) -I$(top_srcdir)/include $(RCFLAGS) -i $< -o $@
 endif
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -55,7 +55,7 @@ include Makefile.inc
 
 # CURL_FILES comes from Makefile.inc
 curl_SOURCES = $(CURL_FILES)
-if OS_WINDOWS
+if HAVE_WINDRES
 curl_SOURCES += $(CURL_RCFILES)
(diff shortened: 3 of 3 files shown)


vszakats:
After iterating through a few options, I'm siding that this is a local configuration issue. Unless there is a valid environment that misses a `windres` tool (or a valid case/bug where autotools is not detecting it), it seems better to hard-fail in this case, than to fail silently by not compiling the resources.

jay:
@sithglan's error from the mailing list is below for reference. I think that if I saw this error it would not be immediately apparent to me what is happening. Is it worthwhile to test if rc tool exists in the configuration and fail with "rc required" or something, or do you think this is just specific to him? I haven't heard anything else about this.

~~~
make[2]: Entering directory '/srv/kvm/scratch/vlconnect/curl/src'
  CC tool_libinfo.o
/bin/bash ../libtool --tag=RC --mode=compile -I../include -DCURL_EMBED_MANIFEST -i curl.rc -o curl.o
libtool: error: ignoring unknown tag RC
Usage: /srv/kvm/scratch/vlconnect/curl/libtool [OPTION]... [MODE-ARG]...
Try 'libtool --help' for more information.
libtool: error: unrecognised option: '-I../include'
make[2]: *** [Makefile:20990: curl.o] Error 1
make[2]: Leaving directory '/srv/kvm/scratch/vlconnect/curl/src'
make[1]: *** [Makefile:1488: all-recursive] Error 1
make[1]: Leaving directory '/srv/kvm/scratch/vlconnect/curl/src'
make: *** [Makefile:1238: all-recursive] Error 1
~~~

vszakats:
The way it fails is definitely confusing. I'm not entirely sure what's happening in that env, but the cross-compiler prefix seems wrong (`mingw32-`). This may be a reason why several autotools auto-detection fail (e.g. `checking for mingw32-ar... no`). `i686-w64-mingw32-` is the correct value there, which is also the expected one for x86.

From the log it seems a similar issue happened in the past with OpenSSL, hence the explicit `RC=` line in the make invocation for it: `make -j ${HECS} RC=/usr/bin/i686-w64-mingw32-windres`. This is missing from the curl `make` command, and adding it would fix it.

`./configure` output already has this:
```
checking for mingw32-windres... no
checking for windres... no
```

How could we improve that, to be actually noticable?

jay:
If it's required then AC_MSG_ERROR but now I'm not so sure because what you describe sounds more like user error. @sithglan

sithglan:
Hello Viktor,

> The way it fails is definitely confusing. I'm not entirely sure what's
> happening in that env, but the cross-compiler prefix seems wrong
> (`mingw32-`). This may be a reason why several autotools
> auto-detection fail (e.g. `checking for mingw32-ar... no`).
> `i686-w64-mingw32-` is the correct value there, which is also the
> expected one for x86.

thank you, that indeed fixed my issue:

(infra) [/srv/kvm/scratch/vlconnect] git diff compile_dependencies.sh
diff --git a/compile_dependencies.sh b/compile_dependencies.sh
index cdd0d27..cd8da42 100644
--- a/compile_dependencies.sh
+++ b/compile_dependencies.sh
@@ -85,7 +85,7 @@ build_curl_openssl()
         autoreconf -fi
         CC=i686-w64-mingw32-gcc ./configure CFLAGS="${CFLAGS}" \
                 --with-ssl=`pwd`/../local/win32/openssl \
-                --host=mingw32 \
+                --host=i686-w64-mingw32 \
                 --prefix=`pwd`/../local/win32/openssl \
                 $LIBCURL_COMMON_PARAMS
         make -j ${HECS}

> From the log it seems a similar issue happened in the past with
> OpenSSL, hence the explicit `RC=` line in the make invocation for it:
> `make -j ${HECS} RC=/usr/bin/i686-w64-mingw32-windres`. This is
> missing from the curl `make` command, and adding it would fix it.

Specifying the RC= line does not hurt but also was not necessary after
specifying the right 'host' value.

Cheers,
        Thomas

vszakats:
Thank you Thomas for your logs and report, I'm happy it worked out fine.

I think this patch would improve error handling, so it'd be nice to merge. We're close to the release, and this isn't a showstopper, so maybe after the release would be the best.

PR #9785, opened 2022-10-23
ci: bump to gcc-11 for macos

vszakats:
Ref: https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/
Ref: https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

Closes #xxxx


cmeister2 (approved):
LGTM

PR #9787, opened 2022-10-23
noproxy: include netinet/in.h for htonl()

vszakats:
Solve the Amiga build warning by including `netinet/in.h`.

`krb5.c` and `socketpair.c` are using `htonl()` too. This header is already included in those sources.

Closes #9787

### DIFF
--- a/lib/noproxy.c
+++ b/lib/noproxy.c
@@ -30,6 +30,10 @@
 #include "strcase.h"
 #include "noproxy.h"
 
+#ifdef HAVE_NETINET_IN_H
+#include <netinet/in.h>
+#endif
+
 /*
  * Curl_cidr4_match() returns TRUE if the given IPv4 address is within the
  * specified CIDR address range.


bagder:
We also use `htonl()` in krb5.c and socketpair.c...

jzakrzewski:
Also: shouldn't this check for big-endian architecture?

vszakats:
@jzakrzewski: It definitely should. Can you create a patch for that?

@bagder: Indeed, it'd be best to have this as an internal public function, similar to e.g. `Curl_inet_pton()`.

This leads a little farther than anticipated :) 

[ Also reminds me that this function in [nghttp2](https://github.com/nghttp2/nghttp2/blob/cb11cfcd2c3d434da91538722a7692fe0ff0d8e9/lib/nghttp2_net.h)/[ngtcp2](https://github.com/ngtcp2/ngtcp2/blob/cfec0f9c6a9fb8cd7c587769dd7409f86573b4ec/lib/ngtcp2_net.h) might need a check to see if they work as intended in ARM64 Windows builds. /cc @tatsuhiro-t  ]

vszakats:
Well, sidestepping all the complications with an own-rolled function, I've found this simpler fix, at least for Amiga:
```diff
diff --git a/lib/noproxy.c b/lib/noproxy.c
index 701a09e04..2c1f8f954 100644
--- a/lib/noproxy.c
+++ b/lib/noproxy.c
@@ -30,6 +30,10 @@
 #include "strcase.h"
 #include "noproxy.h"
 
+#ifdef HAVE_NETINET_IN_H
+#include <netinet/in.h>
+#endif
+
 /*
  * Curl_cidr4_match() returns TRUE if the given IPv4 address is within the
  * specified CIDR address range.
```

We should find a better place for this to cover the other functions, in case they don't have it already.

Or maybe include the 'whole set' as used in many places in curl already:
```
#ifdef HAVE_NETINET_IN_H
#include <netinet/in.h>
#endif
#ifdef HAVE_NETINET_IN6_H
#include <netinet/in6.h>
#endif
#ifdef HAVE_NETDB_H
#include <netdb.h>
#endif
#ifdef HAVE_ARPA_INET_H
#include <arpa/inet.h>
#endif
#ifdef __VMS
#include <in.h>
#include <inet.h>
#endif
```

bagder:
network byte order *is* big endian, so on big endian systems this is a no-op.

See the [glibc code](https://codebrowser.dev/glibc/glibc/inet/htonl.c.html).

But yeah, if this isn't needed even for Amiga then we should probably just drop it.

jzakrzewski:
> network byte order _is_ big endian, so on big endian systems this is a no-op.

My point exactly

vszakats:
[ Also on Windows, ARM/ARM64 runs in little-endian mode, so ngtcp2/nghttp2 are fine. ]

vszakats:
Updated to pull in the header. `krb5.c` and `socketpair.c` do include it already, so they should be fine when those features are enabled.

bagder (approved):

PR #9789, opened 2022-10-24
netrc: replace fgets with Curl_get_line

bagder:
Make the parser only accept complete lines and avoid problems with overly long lines.

Reported-by: Hiroki Kurosawa

### DIFF
--- a/lib/curl_get_line.c
+++ b/lib/curl_get_line.c
@@ -25,7 +25,7 @@
 #include "curl_setup.h"
 
 #if !defined(CURL_DISABLE_COOKIES) || !defined(CURL_DISABLE_ALTSVC) ||  \
-  !defined(CURL_DISABLE_HSTS)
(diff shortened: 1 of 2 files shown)


fredericgermain:
For info, this PR was included in curl 7.86.0 shipped with MacOS Ventura 13.2 (MacOS Ventura 13.1 had curl 7.85.0).

But https://github.com/curl/curl/pull/9973 was not included on curl 7.86.0, so there is a bug about .netrc files not ending with newlines not being handled properly on mac venture 13.2.

bagder:
That sounds like an Apple problem.

fredericgermain:
When I compile 7.86.0 on linux, there is this problem. it's a curl 7.86.0 problem.

bagder:
Sure, which is why we fixed the problem and included that in an update.

PR #9793, opened 2022-10-24
winidn: drop WANT_IDN_PROTOTYPES

vszakats:
`WANT_IDN_PROTOTYPES` was necessary to avoid using a header that
came via an optional package. MS stopped distributing this package some
years ago and the winidn definitions are part of standard headers (via 
`windows.h`) since Vista.

Auto-detect Vista inside `lib/idn_win32.c` and enable the manual
definitions if building for an older Windows.

This allows to delete this manual knob from all build-systems.

Also drop the `_SAL_VERSION` sub-case:

The definitions are now only enabled with old systems. We assume that
code analysis is not run on such systems, allowing us to delete the
SAL-friendly flavour of these.


jay (approved):
The two commits should be combined. The first commit message's first sentence says:

> This was necessary to avoid using a header that came via an optional package.

It would be clearer to say

> WANT_IDN_PROTOTYPES was necessary to avoid using a header that came via an optional package.

because otherwise it could be read as "This (commit) was necessary to avoid using a header that came via an optional package."

vszakats:
Thanks @jay, update the text, squashed and rebased on master.

PR #9796, opened 2022-10-24
avoid duplicated include files

bagder:


dfandrich:
Looks fine to me, but maybe it's better to unconditionally include memdebug.h et. al. in md4/md5/sha256 so it's not forgotten in new backends or if existing backends are changed. Also, the regex would be better with "\s" instead of " " to catch tabs as well as spaces.

bagder:
> Looks fine to me, but maybe it's better to unconditionally include memdebug.h et. al. in md4/md5/sha256 so it's not forgotten in new backends or if existing backends are changed.

A good idea. But to do that properly we need to rearrange the code a bit so that all include files are included at the top and not distributed all over like now. Which also would make the files a little cleaner I think.

> Also, the regex would be better with "\s" instead of " " to catch tabs as well as spaces.

Good point. Fixing.

bagder:
I will wait with merging this until after the 7.86.0 release

PR #9801, opened 2022-10-26
sendf: Remove unnecessary null buffer check

JDepooter:
At this point, the `psnd->buffer` will always exist. We have already allocated a new buffer if one did not previously exist, and returned from the function if the allocation failed.


bagder (approved):

bagder:
Thanks!

PR #9802, opened 2022-10-26
Various typo and grammar fixes in non-code files

Ayesh:
- Replace `Github` with `GitHub`.
- Replace `windows` with `Windows`
- Replace `advice` with `advise` where a verb is used.
- A few fixes on removing repeated words.
- Replace `a HTTP` with `an HTTP`

### DIFF
--- a/docs/CONTRIBUTE.md
+++ b/docs/CONTRIBUTE.md
@@ -148,7 +148,7 @@ changes.
 
 ### About pull requests
 
-With github it is easy to send a [pull
+With GitHub it is easy to send a [pull
 request](https://github.com/curl/curl/pulls) to the curl project to have
 changes merged.
 
--- a/docs/EXPERIMENTAL.md
+++ b/docs/EXPERIMENTAL.md
@@ -10,7 +10,7 @@ Experimental support in curl means:
    "carved in stone".
 2. You must enable the feature when invoking configure as otherwise curl will
    not be built with the feature present.
-3. We strongly advice against using this feature in production.
+3. We strongly advise against using this feature in production.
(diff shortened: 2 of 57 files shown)


bagder:
`github` in lowercase should probably be removed from `.github/scripts/spellcheck.words` so that the CI jobs better can detect this mistake going forward.

The use of **a HTTP**... is used in numerous more places if you want to fix it even more. A fair share of them in source comments:
~~~
$ git grep -i ' a http' | wc -l
86

Ayesh:
Awesome, I will get to it and push to the same PR. Thank you.

Ayesh:
I force-pushed ([diff](https://github.com/curl/curl/compare/5a91b1b29783e09a7e55ca36cd4eaa0bc454c895..c0b08a0e813f038367dcddb19e705514621ee432))with the rest of the `s/a http/an http/` patterns, and manually double checked every chunk. I also capitalized a couple `http` words in the comments to `HTTP` where appropriate. 

This seems to be all of the typos with it, at least upto cd95ee9f771361.

Apparently the spellchecker[ picks up](https://github.com/curl/curl/actions/runs/3327118439/jobs/5501550359) on correct GitHub URLs as a typo; not sure what to do.

bagder:
But it made two lines too long:
~~~
./url.c:1384:81: warning: Longer than 79 columns (LONGLINE)
         /* The requested connection does not use an HTTP proxy or it uses SSL or
./urldata.h:923:81: warning: Longer than 79 columns (LONGLINE)
      struct only because we can do just about any protocol through an HTTP proxy

bagder (approved):

bagder:
Thanks!

PR #9805, opened 2022-10-26
windows: fix linking .rc to shared curl with autotools

### LINKED ISSUE #9803: Build fails with `./configure --enable-shared --disable-static` on Windows
This issue is a continuation from https://github.com/curl/curl/pull/9521#issuecomment-1291864501

### I did this

```bash
# using gcc/clang + mingw-w64
autoreconf -vfi
./configure --enable-shared --disable-static --with-schannel
make -j
```

```bash
libtool: compile:  windres -I../include -DCURL_EMBED_MANIFEST -i curl.rc  -DDLL_EXPORT -DPIC -o .libs/curl.o
  CCLD     curl.exe
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find curl.o: No such file or directory
collect2.exe: error: ld returned 1 exit status
```

git bisect -> 6de7322c03d5b4d915

```bash
$ git bisect good
6de7322c03d5b4d91576a7d9fc893e03cc9d1057 is the first bad commit
commit 6de7322c03d5b4d91576a7d9fc893e03cc9d1057
Author: Viktor Szakats
Date:   Sun Sep 18 12:05:35 2022 +0000
```

### I expected the following

it to build

### curl/libcurl version

7.86.0

### operating system

Windows 11

vszakats:
`./configure --enable-shared --disable-static` fails when trying to link a shared `curl.exe`, due to `libtool` magically changing the output filename of `windres` to one that it doesn't find when linking:

```
/bin/sh ../libtool --tag=RC --mode=compile windres -I../../curl-7.86.0/include -DCURL_EMBED_MANIFEST  -i ../../curl-7.86.0/src/curl.rc -o curl.o
libtool: compile:  windres -I../../curl-7.86.0/include -DCURL_EMBED_MANIFEST -i ../../curl-7.86.0/src/curl.rc  -o .libs/curl.o
[...]
CCLD     curl.exe
clang: error: no such file or directory: 'curl.o'
```

Let's resolve this by skipping `libtool` and calling `windres` directly when building `src` (aka `curl.exe`). Leave `lib` unchanged, as it does need the `libtool` magic. This solution is compatible with building a static `curl.exe`.

This build scenario is not CI-tested.

While here, delete an obsolete comment about a permanent `libtool` warning that we've resolved earlier.

Regression from 6de7322c03d5b4d91576a7d9fc893e03cc9d1057

Reported-by: Christoph Reiter
Fixes #9803
Closes #xxxx

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -155,7 +155,6 @@ optiontable:
 	perl optiontable.pl < $(top_srcdir)/include/curl/curl.h > easyoptions.c
 
 if OS_WINDOWS
-# Warning is "normal": libtool:   error: ignoring unknown tag RC
 .rc.lo:
 	$(LIBTOOL) --tag=RC --mode=compile $(RC) -I$(top_srcdir)/include $(RCFLAGS) -i $< -o $@
 endif
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -157,7 +157,6 @@ listhelp:
 	(cd $(top_srcdir)/docs/cmdline-opts && ./gen.pl listhelp *.d) > tool_listhelp.c
 
 if OS_WINDOWS
-# Warning is "normal": libtool:   error: ignoring unknown tag RC
 .rc.o:
-	$(LIBTOOL) --tag=RC --mode=compile $(RC) -I$(top_srcdir)/include -DCURL_EMBED_MANIFEST $(RCFLAGS) -i $< -o $@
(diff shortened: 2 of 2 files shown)


vszakats:
This build-case is not tested in our CI, nor is it in curl-for-win builds.

Please test and leave a comment here in case there is any issue with ths committed solution.

lazka:
I can confirm that this fixed things: https://github.com/msys2/MINGW-packages/pull/13788

thanks!

Issue #9431, opened 2022-09-05
CURLE_RECV_ERROR schannel: SEC_E_DECRYPT_FAILURE


egorpugin:
### I did this

libcurl
GET https://raw.githubusercontent.com/SoftwareNetwork/database/master//db.version

### I expected the following

Everything worked ok for a long time.

### curl/libcurl version

7.85.0

### operating system

win11, native tls (schannel)

Output
```
*   Trying 185.199.110.133:443...
* Connected to raw.githubusercontent.com (185.199.110.133) port 443 (#0)
* schannel: disabled automatic use of client certificate
* ALPN: offers h2
* ALPN: offers http/1.1
* ALPN: server accepted h2
* Using HTTP2, server supports multiplexing
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* h2h3 [:method: GET]
* h2h3 [:path: /SoftwareNetwork/database/master//db.version]
* h2h3 [:scheme: https]
* h2h3 [:authority: raw.githubusercontent.com]
* h2h3 [accept: */*]
* Using Stream ID: 1 (easy handle 0x2913cc49bd0)
> GET /SoftwareNetwork/database/master//db.version HTTP/2
Host: raw.githubusercontent.com
accept: */*

* schannel: failed to decrypt data, need more data
* schannel: failed to read data from server: SEC_E_DECRYPT_FAILURE (0x80090330) - The specified data could not be decrypted.
* Failed receiving HTTP2 data
* Connection #0 to host raw.githubusercontent.com left intact
```

Same for http1.1.
`curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);`
```
*   Trying 185.199.110.133:443...
* Connected to raw.githubusercontent.com (185.199.110.133) port 443 (#0)
* schannel: disabled automatic use of client certificate
* ALPN: offers http/1.1
* ALPN: server accepted http/1.1
> GET /SoftwareNetwork/database/master//db.version HTTP/1.1
Host: raw.githubusercontent.com
Accept: */*

* schannel: failed to read data from server: SEC_E_DECRYPT_FAILURE (0x80090330) - The specified data could not be decrypted.
* Closing connection 0
* schannel: shutting down SSL/TLS connection with raw.githubusercontent.com port 443
```

jay:
[...]

PR #9807, opened 2022-10-26
sendf: Add debug logging for postponed buffer usage

JDepooter:
The existence of the "postponed" buffer came as a surprise when debugging #9431. I thought I would add some debug logging around the usage of this buffer. I am open to suggestions on improved wording of the messages.

### DIFF
--- a/lib/sendf.c
+++ b/lib/sendf.c
@@ -145,9 +145,9 @@ bool Curl_recv_has_postponed_data(struct connectdata *conn, int sockindex)
          psnd->recv_size > psnd->recv_processed;
 }
 
-static CURLcode pre_receive_plain(struct Curl_easy *data,
-                                  struct connectdata *conn, int num)
+static CURLcode pre_receive_plain(struct Curl_easy *data, int num)
 {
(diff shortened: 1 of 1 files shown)


jay:
Couldn't this create a lot of noise if it shows repeatedly? Is there value in debugging this?

JDepooter:
As I understand it, this buffer is only used on Windows when doing a write on a socket which has unread data. The fact that doing a write caused a read to happen was surprising to me, so I thought it may be worthwhile noting it in the debug log.

I'm not sure how often it happens that curl will write to a socket when there is unread data on the socket. In my scenario, it produced two extra log messages (one read, one write) during a TLS handshake. The SChannel code already produces quite a number of log messages about encrypted/decrypted buffer sizes, so this didn't actually make a noticeable difference in the amount of logging. This buffer was never actually used during the HTTP transfer, because all the data is read before anything further is written. It's possible this could produce a lot of messages in some other protocol which interleaves reads and writes.

I'm not bothered if this doesn't get merged.

Issue #9811, opened 2022-10-27
OS400 fails to build 7.86.0, curl_socklen_t referenced before its defined


jonrumsey:
### I did this
Attempted to build curl 7.86.0 on IBM i V7R4M0

```
CZM1003:  /curl/src/curl-7.86.0/lib/setup-os400.h, 54.31: CZM0046(30) Syntax error.
CZS1601:  Program CHKSTRINGS is not created because statement errors occurred.
CZM1613:  The compilation failed.
```
The order of the includes in curl_setup.h was changed by 
https://github.com/curl/curl/pull/9453
so that curl.h is now included after platform specific headers (i.e. setup-os400.h) and the typedef of curl_socklen_t is now after being referenced in a prototype like;

```
   53       |extern int Curl_getnameinfo_a(const struct sockaddr *sa,                 
   54       |                              curl_socklen_t salen,                      
===========> ..............................a..........................................
*=SEVERE==========> a - CZM0046  Syntax error.                                        
   55       |                              char *nodename, curl_socklen_t nodenamelen,
   56       |                              char *servname, curl_socklen_t servnamelen,
   57       |                              int flags);                                
```

I think the right fix here is to keep the order of the includes but change the prototype of the platform specific ASCII wrapper to use the platform socklen_t, to match the prototype for getnameinfo, and not use the curl_socklen_t.

### I expected the following
Clean build

### curl/libcurl version
curl 7.86.0

### operating system
IBM i V7R4M0

PR #9812, opened 2022-10-27
Use platform socklen_t in os400 Curl_getnameinfo_a

jonrumsey:
Curl_getnameinfo_a() is prototyped before including curl.h as an ASCII'fied wrapper for the platform getnameinfo(), which itself is prototyped with socklen_t arguments, so the wrapper prototype should use the platform socklen_t and not curl_socklen_t.

This should resolve https://github.com/curl/curl/issues/9811


bagder (approved):
Since Jon runs the tests and this is limited to OS400, I have no reason to doubt this!

bagder:
Thanks!

PR #9815, opened 2022-10-27
curl: timeout in the read callback

bagder:
The read callback can timeout if there's nothing to read within the given maximum period. Example use case is when doing "curl -m 3 telnet://example.com" or anything else that expects input on stdin or similar that otherwise would "hang" until something happens and then not respect the timeout.

This fixes KNOWN_BUG 8.1, first filed in July 2009.


bagder:
This fix cleverly does not work on Windows, because Windows has a defective select() function that cannot wait on stdin. Making it work on Windows will be left as a future exercise.

PR #9816, opened 2022-10-27
Fix build on NonStop requiring htonl that is defined in arpa/inet.h.

rsbeckerca:
The location of htonl() is not in the typical place on NonStop.

### DIFF
--- a/lib/noproxy.c
+++ b/lib/noproxy.c
@@ -34,6 +34,10 @@
 #include <netinet/in.h>
 #endif
 
+#ifdef HAVE_ARPA_INET_H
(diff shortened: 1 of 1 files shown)


jay:
That include is also missing from lib/krb5.c which calls htonl, so is a similar change needed there as well?

rsbeckerca:
> That include is also missing from lib/krb5.c which calls htonl, so is a similar change needed there as well?

I did not see an error from the build of that file.

dfandrich:
Unless Kerberos has been enabled, it won't be compiled.

jay:
I amended the commit to include in krb5 as well. Thanks

PR #9817, opened 2022-10-27
Correctly define curl_off_t on NonStop ia64 and x86 for 32-bit builds.

rsbeckerca:
This change is not required for 64-bit builds on NonStop where curl_off_t would be 64-bit automatically.


jay:
Thanks

PR #9820, opened 2022-10-28
docs/EARLY-RELEASE.md: how to determine an early release

bagder:
URL: https://curl.se/mail/lib-2022-10/0079.html

### DIFF
--- a/docs/EARLY-RELEASE.md
+++ b/docs/EARLY-RELEASE.md
@@ -1,0 +1,64 @@
+# How to determine if an early patch release is warranted
+
+In the curl project we do releases every 8 weeks. Unless we break the cycle
+and to an early patch release.
+
+We do frequent releases partly to always have the next release "not too far
+away".
+
+## Bugfix
+
+During the release cycle, and especially in the beginning of a new cycle,
+there are times when a bug is reported and after it has been subsequently
+fixed correctly, the discussion might be brought up: is this bug and
+associated fix important enough for an early patch release.
+
+The question can only be properly asked once a fix has been created and landed
+in the git master branch.
+
+## Early release
+
+An early patch release means that we ship a new, complete and full release
+called `major.minor.patch` where the `patch` part is increased by one since
+the previous release. A curl release is a curl release. There is no small or
+big and we never release just a patch. There is only "release".
+
+## Questions to ask
+
+ - Is there a security advisory rated high or critical?
+ - Is there a data corruption bug?
+ - Did the bug cause an API/ABI breakage?
+
+If the answer is yes to one or more of the above, an early release might
+indeed be warranted.
+
+More questions to ask ourselves when doing the assessment if the answers to
+the three ones above are all 'no'.
+
+ - Does the bug cause curl to prematurely terminate?
+ - How common is the affected buggy option/feature/protocol/platform to get
+   used?
+ - How large is the estimated impacted user base?
+ - Does the bug block something crucial for applications or other adoption of
+   curl "out there" ?
+ - Does the bug cause problems for curl developers or others on "the curl
+   team" ?
+ - Is the bug limited to the curl tool only? That might have a smaller impact
+   than a bug also present in libcurl.
+ - Is there a (decent) workaround?
+ - Is it a regression? Is the bug introduced in this release?
+ - Can the bug be fixed "easily" by applying a patch?
+ - Does the bug break the build? Most users don't build curl themselves.
+ - How long is it left until the already scheduled next release?
+ - Can affected users safely rather revert to a former release until the next
+   scheduled release?
+
+## If an early release is deemed necessary
+
+Unless done for security or similarly very important reasons, an early release
+should never be done within two weeks of the previous version was shipped.
+
+This, to enable us to collect and bundle more fixes into the same release to
+make the release more worthwhile for everyone and to allow more time for fixes
+to settle and things to get tested. Getting a release in shape and done in
+style is work that should not be rushed.
--- a/docs/EARLY-RELEASE.md
+++ b/docs/EARLY-RELEASE.md
@@ -55,7 +55,7 @@ the three ones above are all 'no'.
 
 ## If an early release is deemed necessary
 
-Unless done for security or similarly very important reasons, an early release
+Unless done for security or similarly important reasons, an early release
 should never be done within two weeks of the previous version was shipped.
 
 This, to enable us to collect and bundle more fixes into the same release to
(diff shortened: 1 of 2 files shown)


rodwiddowson:
I missed the mail conversation until now but I think there may be another question - which I apply when weighing up these issues.
_Is rolling back to the previous release an acceptable workaound for affected users?_

Obviously there are many cases when it isn't.  like "The affected release carried a fix for a Security Advisory" or "This bug is a security issue and may go unnoticed".

In the specific case I would have thought that waiting until Tuesday before cranking a release might be prudent....

docs/EARLY-RELEASE.md
```diff
@@ -0,0 +1,64 @@
+# How to determine if an early patch release is warranted
+
+In the curl project we do releases every 8 weeks. Unless we break the cycle
+and to an early patch release.
```
danielgustafsson, line 4: s/to/do/

danielgustafsson:
> .. or "This bug is a security issue and may go unnoticed".

For bugs that are found to be security issues I think the [security process](https://github.com/curl/curl/blob/master/docs/SECURITY-PROCESS.md) already has wording around that. If we feel that's not giving enough direction I think we should address that there instead.

rodwiddowson:
> For bugs that are found to be security issues I think the security process already has wording ..

Agreed 110%

docs/EARLY-RELEASE.md
bagder: :heavy_check_mark:

bagder:
> I missed the mail conversation until now but I think there may be another question - which I apply when weighing up these issues. _Is rolling back to the previous release an acceptable workaound for affected users?_

I phrased it like this in the current draft:

 - Can affected users safely rather revert to a former release until the next scheduled release?

> Obviously there are many cases when it isn't. like "The affected release carried a fix for a Security Advisory" or "This bug is a security issue and may go unnoticed".

Right. Like our current situation: we shipped 7.86.0 with fixes for **four** security related problems. Asking users to stick to the former release due to the existence of a problem in the new release might then implicitly ask them to live with those four security problems for another release cycle ... Not necessarily the most responsible way.

> In the specific case I would have thought that waiting until Tuesday before cranking a release might be prudent....

I went into scheduling a little bit in the bottom for the cases we actually deem an early release to be the right choice. It is also important to not rush out releases to make sure things are always done in an orderly way and that all procedures are followed.

docs/EARLY-RELEASE.md
```diff
@@ -0,0 +1,64 @@
...
+## Early release
+
+An early patch release means that we ship a new, complete and full release
+called `major.minor.patch` where the `patch` part is increased by one since
+the previous release. A curl release is a curl release. There is no small or
+big and we never release just a patch. There is only "release".
+
+## Questions to ask
+
+ - Is there a security advisory rated high or critical?
```
danielgustafsson, line 28: Come to think of it, if there was, wouldn't that warrant a new release which bumps the minor rev and not just the patch rev? This only applies to *high* and *critical* of course.

docs/EARLY-RELEASE.md
```diff
@@ -0,0 +1,64 @@
...
+ - Can the bug be fixed "easily" by applying a patch?
+ - Does the bug break the build? Most users don't build curl themselves.
+ - How long is it left until the already scheduled next release?
+ - Can affected users safely rather revert to a former release until the next
+   scheduled release?
+
+## If an early release is deemed necessary
+
+Unless done for security or similarly important reasons, an early release
+should never be done within two weeks of the previous version was shipped.
```
rybak, line 59: Grammar: "within two weeks of" should be followed by a noun-phrase, without a verb. You might want to consider something like:
```suggestion
should never be done within two weeks of the release of the previous version.
```

docs/EARLY-RELEASE.md
```diff
@@ -0,0 +1,64 @@
...
+   team" ?
+ - Is the bug limited to the curl tool only? That might have a smaller impact
+   than a bug also present in libcurl.
+ - Is there a (decent) workaround?
+ - Is it a regression? Is the bug introduced in this release?
+ - Can the bug be fixed "easily" by applying a patch?
+ - Does the bug break the build? Most users don't build curl themselves.
+ - How long is it left until the already scheduled next release?
+ - Can affected users safely rather revert to a former release until the next
+   scheduled release?
```
danielgustafsson, line 54: - Is it a performance regression which has no functionality side-effects? If so it has to be substantial.

henning-schild:
Great to see how you guys deal with all of that and thanks so much!

The breakage can be seen as severe and security relevant on its own. Any machine in a network with proxies will potentially/partially loose network connectivity after installing that affected release. Which is clearly related to the A in CIA.

Maybe people could even get stuck on a broken release. Because their package manager is based on curl and their repo mirror based on `no_proxy`. Would require some package sideloading tricks to get back to live.

bagder:
> Maybe someone could ping the maintainers of alpine and arch linux.

There are bugs in *every* curl release. I presume distro maintainers have ways to deal with that.

docs/EARLY-RELEASE.md
bagder: Would it really? If we just fix an internal flaw somewhere that is still just a patch release, even if that fixes a really serious security flaw - I think.

docs/EARLY-RELEASE.md
bagder: :heavy_check_mark:

docs/EARLY-RELEASE.md
bagder: :heavy_check_mark:

docs/EARLY-RELEASE.md
rsbeckerca: If you are on a monthly cycle, and this fix comes out 2 days into the cycle and you want to release it as a patch, but will be more thoroughly packaging the fix by the end of the cycle, perhaps it is a patch, not a minor release.

danielgustafsson: If we are past the opening of the feature window then all releases are by definition minor rev bumps since new features will have gone in. (or well, *may* have gone in..)

danielgustafsson: > Would it really? If we just fix an internal flaw somewhere that is still just a patch release, even if that fixes a really serious security flaw - I think.

For signalling purposes I think it can be argued for minor rev bump, but technically it should be a patch following the current process.

bagder:
Surely this should *not* have the `hacktoberfest-accepted` tag set?

PR #9822, opened 2022-10-28
noproxy: fix tail-matching

### LINKED ISSUE #9821: noproxy tailmatching issue
```
curl 7.86.1-DEV (x86_64-pc-linux-gnu) libcurl/7.86.1-DEV OpenSSL/1.1.1n libidn2/2.3.0
Release-Date: [unreleased]
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IDN IPv6 Largefile NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets
```

commit d4fed2a13a81d23e73f1fb491c335a1b1d91e3fb

```
export https_proxy=http://127.0.0.5:3128
export http_proxy=http://127.0.0.5:3128
export no_proxy=localhost,.google.com,.google.de
curl -vv -o /dev/null http://www.google.com/testfile
```

And again it tries to go via that proxy where it should not.

_Originally posted by @henning-schild in https://github.com/curl/curl/issues/9813#issuecomment-1294657686_

bagder:
Also ignore trailing dots in both host name and comparison pattern.

Regression in 7.86.0 (from 1e9a538e05c0)

Extended test 1614 to verify better.

Reported-by: Henning Schild
Fixes #9821


tests/unit/unit1614.c
```diff
@@ -77,6 +77,12 @@ UNITTEST_START
     { NULL, NULL, 0, FALSE} /* end marker */
   };
   struct noproxy list[]= {
+    { "www.example.com", "localhost,.example.com,.example.de", TRUE},
+    { "www.example.com.", "localhost,.example.com,.example.de", TRUE},
+    { "example.com", "localhost,.example.com,.example.de", TRUE},
+    { "example.com.", "localhost,.example.com,.example.de", TRUE},
+    { "www.example.com", "localhost,.example.com.,.example.de", TRUE},
+    { "www.example.com", "localhost,www.example.com.,.example.de", TRUE},
```
henning-schild, line 85: Maybe also test for `example.com` and not just `.example.com`, some tools out there can not do the leading dot so people might not have it in their env.

Not too related to the issue at hand, but for sure a good improvement of the test-suite to prevent potential problems in the future.

bagder: A good idea. I added three more checks. That what you had in mind?

henning-schild: Yes, looking good!

PR #9823, opened 2022-10-28
Fix pointer mismatchings and other warnings

AtariDreams:

Note that none of the public facing functions had their signature changed.

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -398,18 +398,23 @@ static CURLcode bindlocal(struct Curl_easy *data,
 #ifdef HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID
         char *scope_ptr = strchr(myhost, '%');
         if(scope_ptr)
-          *(scope_ptr++) = 0;
+          *(scope_ptr++) = '\0';
 #endif
         if(Curl_inet_pton(AF_INET6, myhost, &si6->sin6_addr) > 0) {
           si6->sin6_family = AF_INET6;
           si6->sin6_port = htons(port);
 #ifdef HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID
-          if(scope_ptr)
+          if(scope_ptr) {
             /* The "myhost" string either comes from Curl_if2ip or from
                Curl_printable_address. The latter returns only numeric scope
                IDs and the former returns none at all.  So the scope ID, if
                present, is known to be numeric */
-            si6->sin6_scope_id = atoi(scope_ptr);
+            unsigned long scope_id = strtoul(scope_ptr, NULL, 10);
+            if(scope_id > UINT_MAX)
+              return CURLE_UNSUPPORTED_PROTOCOL;
+
+            si6->sin6_scope_id = (unsigned int)scope_id;
+          }
 #endif
         }
         sizeof_sa = sizeof(struct sockaddr_in6);
@@ -866,7 +871,7 @@ CURLcode Curl_is_connected(struct Curl_easy *data,
   int error = 0;
   struct curltime now;
   int rc = 0;
-  unsigned int i;
+  int i;
 
   DEBUGASSERT(sockindex >= FIRSTSOCKET && sockindex <= SECONDARYSOCKET);
 
@@ -1207,7 +1212,7 @@ static CURLcode singleipconnect(struct Curl_easy *data,
     return result;
 
   /* store remote address and port used in this connection attempt */
-  if(!Curl_addr2string((struct sockaddr*)&addr.sa_addr, addr.addrlen,
+  if(!Curl_addr2string(&addr.sa_addr, addr.addrlen,
                        ipaddress, &port)) {
     /* malformed address or bug in inet_ntop, try next address */
     failf(data, "sa_addr inet_ntop() failed with errno %d: %s",
@@ -1262,7 +1267,7 @@ static CURLcode singleipconnect(struct Curl_easy *data,
 #endif
     ) {
     result = bindlocal(data, sockfd, addr.family,
-                       Curl_ipv6_scope((struct sockaddr*)&addr.sa_addr));
+                       Curl_ipv6_scope(&addr.sa_addr));
     if(result) {
       Curl_closesocket(data, conn, sockfd); /* close socket and bail out */
       if(result == CURLE_UNSUPPORTED_PROTOCOL) {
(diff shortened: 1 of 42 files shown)


lib/base64.c
```diff
@@ -55,10 +55,10 @@ static const char base64url[]=
 static size_t decodeQuantum(unsigned char *dest, const char *src)
 {
   size_t padding = 0;
-  const char *s;
+  const char *s = src;
   unsigned long i, x = 0;
 
-  for(i = 0, s = src; i < 4; i++, s++) {
+  for(i = 0; i < 4; i++, s++) {
```
bagder, line 61: This is not a pointer mismatch change.

lib/connect.c
```diff
@@ -409,7 +409,7 @@ static CURLcode bindlocal(struct Curl_easy *data,
                Curl_printable_address. The latter returns only numeric scope
                IDs and the former returns none at all.  So the scope ID, if
                present, is known to be numeric */
-            si6->sin6_scope_id = atoi(scope_ptr);
+            si6->sin6_scope_id = (unsigned int)strtoul(scope_ptr, NULL, 10);
```
bagder, line 412: if we change this, shouldn't we also make a proper overflow check as well? After all, `strtoul()` might return a value that cannot be stored as a scope id...

lib/connect.c
```diff
@@ -866,7 +866,7 @@ CURLcode Curl_is_connected(struct Curl_easy *data,
   int error = 0;
   struct curltime now;
   int rc = 0;
-  unsigned int i;
+  int i;
```
bagder, line 869: why?

lib/connect.c
```diff
@@ -1056,8 +1056,8 @@ CURLcode Curl_is_connected(struct Curl_easy *data,
       hostname = conn->http_proxy.host.name;
     else
 #endif
-      if(conn->bits.conn_to_host)
-        hostname = conn->conn_to_host.name;
+    if(conn->bits.conn_to_host)
+      hostname = conn->conn_to_host.name;
```
bagder, line 1060: why?

lib/curl_fnmatch.c
```diff
@@ -71,18 +71,19 @@ typedef enum {
...
   parsekey_state state = CURLFNM_PKW_INIT;
 #define KEYLEN 10
   char keyword[KEYLEN] = { 0 };
-  int found = FALSE;
-  int i;
-  unsigned char *p = *pattern;
+  bool found = FALSE;
+  unsigned int i;
+  char *p = *pattern;
+  char c;
```
bagder, line 82: Why the changed scope for `c` here?

lib/curl_fnmatch.c
```diff
@@ -140,33 +141,33 @@ static char_class charclass(unsigned char c)
...
-    unsigned char endrange = *p++;
+    char endrange = *p++;
 
     if(endrange == '\\')
       endrange = *p++;
     if(endrange >= c && charclass(endrange) == cc) {
       while(c++ != endrange)
         if(charclass(c) == cc)  /* Chars in class may be not consecutive. */
-          charset[c] = 1;
+          charset[(unsigned int)c] = 1;
```
bagder, line 159: This change **introduces** two typecasts. Why is that better?

lib/curl_fnmatch.c
```diff
@@ -245,7 +246,7 @@ static int setcharset(unsigned char **p, unsigned char *charset)
       if(c == ']')
         return SETCHARSET_OK;
       state  = CURLFNM_SCHS_DEFAULT;
-      charset[c] = 1;
+      charset[(unsigned int)c] = 1;
```
bagder, line 249: Here too, you add **seven** typecasts. Why is this better?

bagder (commented):
I review a part of this PR and it made me question what you are doing so I paused.

lib/curl_fnmatch.c
AtariDreams: In case c is negative

lib/curl_fnmatch.c
AtariDreams: In case c is negative

lib/connect.c
AtariDreams: const int other = i ^ 1;

lib/curl_fnmatch.c
AtariDreams: c is defined elsewhere, causing shadowing

lib/connect.c
AtariDreams: Because the formatting here is broken

lib/connect.c
bagder: and why does that get better if `i` is not unsigned?

lib/connect.c
bagder: in your opinion then perhaps, but not in mine. The `if` is intended based on the `else` on the line before.

lib/connect.c
AtariDreams: Because I only goes up to 1 anyway. It silences warnings, assuming you intended to use the variable like a Boolean.

bagder: What warning? There's nothing wrong there, having a warning for that seems overzealous.

lib/curl_fnmatch.c
bagder, line 159: I still don't see the point in this change. It still *adds* two typecasts, that's contrary to the main motivation for this PR.

AtariDreams: It has to be because char is signed

bagder:
[quite bad](https://github.com/curl/curl/actions/runs/3352731574/jobs/5555067542)

AtariDreams:
Not anymore

tests/server/util.c
```diff
@@ -378,18 +378,16 @@ static char raw_toupper(char in)
...
     }
     max--;
     first++;
     second++;
   }
-  if(0 == max)
-    return 1; /* they are equal this far */
 
-  return raw_toupper(*first) == raw_toupper(*second);
+  return 1; /* they are equal this far */
```
bagder, line 390: this change is completely unrelated to the topic and purpose of this PR

bagder:
This PR starts to smell like a million tiny changes you think should be done, but most of them are now not about "pointer mismatches and other warnings". It's a busload of various unrelated changes that should not be in the same PR. It makes this PR very hard to review and next to impossible to ultimately merge.

jay:
You must have very strict compiler settings for the warnings you are seeing. Many of these changes are not needed.

We allow counter variables to go negative or wrap around. For example this is acceptable:

int foo = 5;
while(foo--) {
}
return

We allow subscript values to be a signed integer type. For example this is acceptable:

int foo = 5;
bar[foo]

We allow char * function parameters.

I'm closing this PR in favor of #9835 which is what is left after my review.

PR #9829, opened 2022-10-30
Fix pointers, return types, and casting

AtariDreams:


bagder:
Is this PR replacing #9823?

jay:
> Is this PR replacing #9823?

I'm going through that one now and though I think a lot of it is unnecessary there are some good ideas. I'm trying to pick out what is good and roll it up separately.

jay:
Closing. This appears to be a duplicate of #9823 which was closed in favor of #9835.

PR #9830, opened 2022-10-30
Make strncasecompare more readable

AtariDreams:
It is less convoluted this way.


bagder:
Looking at this, it strikes me we should probably rather re-use code from the lib rather than having a separate implementation (and is not even identical). Like #9831

AtariDreams:
Okay then

PR #9834, opened 2022-10-31
gen.pl: do not generate CURLHELP bitmask lines > 79 characters

monnerat:
If a command line option is in many help categories, there is a risk that `CURLHELP` bitmask source lines generated for `listhelp` are longer than 79 characters.
This commit takes care of folding such long lines.

The case does not occur yet as no command line option lists too many categories.
I discovered this potential problem while keeping the ManageSieve PR up-to-date.


bagder:
Thanks!

bagder:
closed in 3b9af11c77d6325

PR #9835, opened 2022-10-31
lib: fix some type mismatches and remove unneeded typecasts

jay:

Ref: https://github.com/curl/curl/pull/9823

Closes #xxxx

---

This is a subset of the changes submitted by @AtariDreams in #9823.

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -4167,7 +4167,7 @@ CURLcode ftp_parse_url_path(struct Curl_easy *data)
         /* get path before last slash, except for / */
         size_t dirlen = slashPos - rawPath;
         if(dirlen == 0)
-            dirlen = 1;
+          dirlen = 1;
 
         ftpc->dirs = calloc(1, sizeof(ftpc->dirs[0]));
         if(!ftpc->dirs) {
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -1174,7 +1174,7 @@ static CURLcode ftp_state_use_port(struct Curl_easy *data,
   /* get the name again after the bind() so that we can extract the
      port number it uses now */
   sslen = sizeof(ss);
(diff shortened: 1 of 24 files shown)


lib/ftp.c
```diff
@@ -4167,7 +4167,7 @@ CURLcode ftp_parse_url_path(struct Curl_easy *data)
         /* get path before last slash, except for / */
         size_t dirlen = slashPos - rawPath;
         if(dirlen == 0)
-            dirlen++;
+            dirlen = 1;
```
bagder, line 4170: maybe fix the indent level while at it!

lib/noproxy.c
```diff
@@ -222,7 +222,12 @@ bool Curl_check_noproxy(const char *name, const char *no_proxy)
           slash = strchr(check, '/');
           /* if the slash is part of this token, use it */
           if(slash) {
-            bits = atoi(slash + 1);
+            unsigned long longbits = strtoul(slash + 1, NULL, 10);
+            if(longbits > UINT_MAX)
+              /* this cannot match */
+              break;
```
bagder, line 228: The cidr match functions already return error if the bits number is too big.

lib/http.c
```diff
@@ -650,7 +650,7 @@ CURLcode Curl_http_auth_act(struct Curl_easy *data)
   if(!data->set.str[STRING_BEARER])
     authmask &= (unsigned long)~CURLAUTH_BEARER;
 
-  if(100 <= data->req.httpcode && 199 >= data->req.httpcode)
+  if(100 <= data->req.httpcode && data->req.httpcode <= 199)
```
MarcelRaad, line 653: Not sure what this is supposed to fix. The rest looks good to me!

MarcelRaad (approved):

lib/http.c
jay: This one isn't supposed to fix anything it's just a style change.

PR #9839, opened 2022-11-01
Add link for building libcurl with Bazel

cpsauer:
Hi again, @bagder!

Following #9732 and seeing people struggling to use libcurl from Bazel, I released our files for easily and correctly doing so. They work across a few platforms in this initial release (Android + all the Apple platforms). More to come.

I'm hoping you might be willing to link to the new project from your docs to help Bazelers find it, since I think it'd solve an acute pain point for them. (Bazel's CMake/autotools wrappers aren't up to the task, sadly.)

On me to keep it in sync with your build. We're actively using it, so I'd better! And if you aren't satisfied long term, obviously feel free to revert out this link. Heard you loud and clear on not wanting to support more build systems yourself. But we needed this one and I figured it'd be good to share it with the world.

Thanks for all you do,
Chris

P.S. Fun fact I just discovered reading the docs: Curl and I share a birthday!

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -21,6 +21,10 @@ The curl port in vcpkg is kept up to date by Microsoft team members and
 community contributors. If the version is out of date, please [create an issue
 or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
 
+## Building libcurl using Bazel
+
+Please refer to [hedronvision/bazel-make-cc-https-easy](https://github.com/hedronvision/bazel-make-cc-https-easy).
+
 ## Building from git
 
 If you get your code off a git repository instead of a release tarball, see


bagder:
I'm generally positive to this idea, but I think the page you link to is not the page the link makes it out to be. I can't even find a description there on how you build curl with Bazel.

cpsauer:
Oh no! Well, there was certainly no deceit intended, I promise. 

I'll stand by it indeed being instructions/code to get people building and using libcurl via Bazel. But maybe Bazel is sufficiently weird/different that it doesn't come across that way? I do earnestly think it's in the form Bazel users would expect and most want to use. [And there are quite a few Bazel users these days, given how much force Google has put behind it, how clean its graph-based dependency model is, and how many other companies have converted. Though _certainly_ not as many as there are curl users!]

cpsauer:
As evidence that that project is explicitly oriented around getting libcurl to build for Bazel users: 
Almost all the code there is dedicated to building libcurl, hence curl featuring in the about string and tags, instructions culminating in adding `"@curl"` to their `deps`, and directing people to a libcurl example and your docs. When people do that, Bazel plus the extensions in that repo will automatically pull down libcurl and its dependencies and handle platform configuration and building and linking into whatever larger project is being built. 

Google's Bazel is weird in that it's explicitly dependency-graph based, but that's also it's magic; the goal of bazel library rules isn't usually to produce a libcurl binary (though one could with `bazel build @curl` after loading these extensions), but rather to be able to easily reference the library in your dependency graph and have everything just work, even as you reconfigure the larger graph around it. Maybe where things got twisted is the recommendation of the libCPR wrapper for curl for C++, but I think that's probably right. Most Bazel folks will be using C++ and want object oriented bindings around libcurl, but we'll certainly always keep libcurl available to them directly. They just add `"@curl"` to their `deps`--and if they add `"@cpr"`, well, that implicitly brings in libcurl, since that's what's doing the work under the hood. Building libcurl with bazel is definitely the core of what's being provided here

jzakrzewski:
Sorry to be a bit negative, but the link goes to your project that doesn't seem to be about "building curl" (but has indeed the build file for that and includes a line telling people how to use it as as a dependency). It has no instructions to "build curl" and, IIUC, the build cannot be parameterized in any way, except for editing the build file (but maybe it is the Bazel way?).
It may be an example how to write such a file, but I am a bit sceptical about endorsing it in its current current shape, but maybe I am expecting too much.

bagder:
On a page about building curl with Bazel, I expect to find (clear) step by step instructions on how to build curl. If that is not on the page, or if the page contains a lot of other things that aren't strictly related to building curl with Bazel, then I don't think it is a page we should direct people to.

cpsauer:
I appreciate your directness. Normally I'd close this (your site, your call), but it just really seems like we're talking past each other somehow, and that's a shame. I'm just not sure why.

To try to backtrace the problem: We're all on the same page that this is about **lib**curl, right? (The library, not the command line tool. Not much point in adding another build system for the tool, but valuable to help more people compile the library into their projects.)

cpsauer:
If we're on the same page, then seriously guys, this is about as laser focused on being able to build libcurl (from source) into a Bazel project as can possibly be done. 

Listing it as a dependency *is* the step that causes libcurl to be built by bazel.

cpsauer:
Hopefully it goes without saying that I have enormous respect for you guys and I really appreciate your time. I've got to sleep for now (very different timezone!), but this seems likely to me a miscommunication somewhere, and I'd love to earn your trust if I still can.

bagder:
> this is about as laser focused on being able to build libcurl (from source) into a Bazel project as can possibly be done.

Seriously, you cannot mean that. If it was "laser focused" on building libcurl, they why is the document you link to titled **Hedron's Bazel Rules for C++ HTTPS Requests — User Interface** and not **building libcurl with Bazel** ?

cpsauer:
I do. It's phrased that way because a good interface starts with the user need it solves. Most people who arrive directly at the repo (e.g., from the Bazel docs) will be trying to solve the need of making network requests and looking for "Bazel rules" to do so. It then guides them all into building libcurl.

Folks who arrive from your link will often, naturally, have already decided that they want to use libcurl to satisfy that need. That's great! I can make them a separate landing page for them if you want, but it'll be the exact same steps: Configure their bazel workspace to be able to build libcurl, and then start depending on `"@curl"`. Workspace setup followed by `"@curl"` is what Bazel users will arrive looking for; they won't be confused.

Earnestly, guys, I'm trying to help as many Bazel folks solve their network request needs with libcurl as I can. That's my only goal with this, and I think it's a pretty solid effort we could be aligned on. Is there any way I can reset this discussion back to the presumption of honesty and good intent? That's what this is.

bagder:
> I'm trying to help as many Bazel folks solve their network request needs with libcurl as I can

I don't doubt your intentions. I just don't think the page is suitable for a "this how you build libcurl with Bazel" link from the curl docs.

cpsauer:
Bummer. But ok!
I'm hopeful people will still find it; we've gotten enthusiastic merges from all the other stakeholder projects.

Please do let me know if there's ever a good place in the curl docs for helping bazel users browsing, looking to build libcurl into their projects. (As opposed to using bazel to build curl/libcurl in isolation, which I suspect neither you nor they are much interested in?) 

Otherwise, I'll keep looking for ways to help if/when I see them! Like with fixing up the Android build instructions in #9732.

Cheers, and again, thanks for all the good you do for the world!
Chris

PR #9845, opened 2022-11-02
curl: don't ignore query part when using --remote-name

### LINKED ISSUE #9826: New behaviour of -O option (stripping query) breaks existing user scripts
### I did this

`curl -O 'http://mirrors.mageia.org/api/mageia.9.i586.list?reason=upgrade'`

### I expected the following

The transferred data to be stored in a file named `mageia.9.i586.list?reason=upgrade`.

### curl/libcurl version

curl 7.86.0 (i586-mageia-linux-gnu) libcurl/7.86.0 OpenSSL/3.0.5 zlib/1.2.13 brotli/1.0.9 libidn2/2.3.4 libpsl/0.21.1 (+libidn2/2.3.2) libssh/0.10.4/gnutls/zlib nghttp2/1.50.0
Release-Date: 2022-10-26
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets

### operating system

Linux localhost 6.0.6-desktop-1.mga9 #1 SMP PREEMPT_DYNAMIC Sat Oct 29 10:40:13 UTC 2022 i686 GNU/Linux

martinwhitaker:
PR #9684 changed the behaviour of --remote-name to ignore the possible query part of the URL when generating the output file name. This breaks existing code that use curl and expect the old behaviour. This update restores the original behaviour whilst still using the URL parser.

Fixes #9826

### DIFF
--- a/src/tool_operhlp.c
+++ b/src/tool_operhlp.c
@@ -161,6 +161,9 @@ CURLcode get_url_file_name(char **filename, const char *url)
 
   if(!curl_url_set(uh, CURLUPART_URL, url, CURLU_GUESS_SCHEME) &&
      !curl_url_get(uh, CURLUPART_PATH, &path, 0)) {
+    char *query = NULL;
+    curl_url_get(uh, CURLUPART_QUERY, &query, 0);
+
     curl_url_cleanup(uh);
 
     pc = strrchr(path, '/');
@@ -175,7 +178,12 @@ CURLcode get_url_file_name(char **filename, const char *url)
       /* no slash => empty string */
       pc = "";
 
-    *filename = strdup(pc);
+    if(query) {
+      *filename = aprintf("%s?%s", pc, query);
+      curl_free(query);
+    }
+    else
+      *filename = strdup(pc);
     curl_free(path);
     if(!*filename)
       return CURLE_OUT_OF_MEMORY;
--- a/tests/data/test1210
+++ b/tests/data/test1210
@@ -53,7 +53,11 @@ User-Agent: curl/%VERSION
 Accept: */*
 
 </protocol>
-<file name="log/%TESTNUMBER">
+%if win32
+<file name="log/%TESTNUMBER_junk">
+%else
+<file name="log/%TESTNUMBER?junk">
+%endif
 12345
 </file>
 


bagder (approved):
I don't like it but I understand why we need to do this.

bagder:
Did we convert the question mark on Windows before to make it work there?

bagder (changes requested):
test fails on Windows

martinwhitaker:
I've fixed the test to work on Windows. The call to `sanitize_file_name` was there before PR #9684, so the question mark would have been converted.

There are still a few failing tests, but they don't appear to be be related to this PR.

vszakats:
I think curl stripping query params from the filename is the _expected_ (least surprise) behaviour here (saying this after having to rename unexpectedly mangled filenames way too many times). I also understand it may break compatibility if someone is actually building on this unexpected behaviour. Can we reconsider reverting this or implementing some other solution? Or maybe wait a bit to see if this causes widespread breakage?

martinwhitaker:
Well, this has broken the Mageia installler and package manager (urpmi) and is blocking the alpha release of Mageia 9. Is that widespread enough?

And I don't consider stripping the query is the least surprising behaviour. A different query will normally result in different data being returned, so it makes sense to me to store it in a different file.

wget will similarly include the query part in the saved file name.

vszakats:
Can't speak for `urpmi` specificially, but perhaps there is a reasonable way to update the affected script to not rely on this specific behaviour (and making it work with both 7.86.0 and earlier versions). Also, 7.86.0 is out already, so scripts encountering this version will need to adapt for it, unless they have a way to force a specific earlier (or later?) curl version.

I'm for collecting a bit more data points before reverting.

Smackd0wn:
The "original behaviour" prior to 7.86.0 also included the fragment part in the filename. For example,

`curl -O "https://example.com/index.html#[1-10]"`  (adapted from [ref.](https://daniel.haxx.se/blog/2022/12/23/the-curl-fragment-trick/))

would produce output `index.html#1` to `index.html#10`. But that behaviour is not restored by this patch. According to Hyrum's law, that must also break some people's workflow. Should that be reverted as well?


While we're at it, I'd say URL-decoded filename would be the "least surprising behaviour" (wget also does it). But that's another story, and I think "do not break existing scripts and applications" is one of the core principles of curl ([ref.](https://daniel.haxx.se/blog/2022/09/08/http-http-http-http-http-http-http/))

Either way, it would be nice to update the man page on `--remote-name` accordingly.

vszakats:
@Smackd0wn 👍 for URL-decoding and updating the man page.

The fragment trick example can be modified like these to keep copies of each download:
`curl -O 'https://example.com/index.html#[1-10]' --no-clobber` (needs 7.83.0)
`curl 'https://example.com/index.html#[1-10]' --output 'index.html##1'`

bagder:
Every bugfix is a behavior change, but this is one we want to stick with and it has not caused a lot of blow back.

PR #9848, opened 2022-11-03
docs: add missing parameters for retry flag

starbops:
Option `--retry` expects a proper numerical parameter but there's no one in some examples.


bagder (approved):

bagder:
Thanks!

PR #9850, opened 2022-11-03
websockets: install websockets.h only if enabled

xypiie:
install and include websockets.h only if it's enabled by --enable-websockets

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4238,6 +4238,7 @@ AS_HELP_STRING([--disable-websockets],[Disable WebSockets support]),
   esac ],
      AC_MSG_RESULT(no)
 )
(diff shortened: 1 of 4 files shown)


bagder:
I disagree. The headers are made to be able to be shared by multiple curl installations no matter their build configs. There is also no harm in having this extra header even if the support is not built-in.

jay:
I disagree with this. We want the public API functions to still be available, though I think the return code in this case is wrong. When not built in the websocket functions should return CURLE_NOT_BUILT_IN instead of CURLE_OK. See #9851

xypiie:
Thank you for your feedback and good description. @jay Yes, the return code of the stubs were confusing me as well. Thanks for taking care.

PR #9851, opened 2022-11-03
ws: return CURLE_NOT_BUILT_IN when websockets not built in

jay:
- Change curl_ws_recv & curl_ws_send to return CURLE_NOT_BUILT_IN when websockets support is not built in.

Prior to this change they returned CURLE_OK.

Closes #xxxx


bagder (approved):

bagder:
Thanks!

PR #9855, opened 2022-11-05
Connection Filters v0.1

icing:
Connection filters (cfilter) addition to curl:

        - general construct/destroy in connectdata
        - default implementations of callback functions
        - connect: cfilters for connect and accept
        - socks: cfilter for socks proxying
        - http_proxy: cfilter for http proxy tunneling
        - vtls: cfilters for primary and proxy ssl
        - change in general handling of data/conn
          - Curl_cfilter_setup() sets up filter chain based on data settings, if none are installed by the protocol handler setup
          - Curl_cfilter_connect() boot straps filters into `connected` status, used by handlers and multi to reach further stages
          - Curl_cfilter_is_connected() to check if a conn is connected, e.g. all filters have done their work
          - Curl_cfilter_get_select_socks() gets the sockets and READ/WRITE indicators for multi select to work
          - Curl_cfilter_data_pending() asks filters if the have incoming data pending for recv
          - Curl_cfilter_recv()/Curl_cfilter_send are the general callbacks installed in conn->recv/conn->send for io handling
          - Curl_cfilter_attach_data()/Curl_cfilter_detach_data() inform filters and addition/removal of a `data` from their connection
          - adding vtl functions to prevent use of Curl_ssl globals directly in other parts of the code.

    Fixing bugs found by building with cares:

    - report successul connects only once
    - use accept cfilter also in ftp port commands
    - have accept cfilter take a new socket after accept() happened
    - fix FILE handler to not leak memory when handler->connect_it() is called twice

### DIFF
--- a/docs/CONNECTION-FILTERS.md
+++ b/docs/CONNECTION-FILTERS.md
@@ -1,0 +1,127 @@
+# curl connection filters
+
+Connection filters is a design in the internals of curl, not visible in its public API. They were added 
+in curl v7.xx.x. This document describes the concepts, its high level implementation and the motivations.
+
+## Filters
+
+A "connection filter" is a piece of code that is responsible for handling a range of operations
+of curl's connections: reading, writing, waiting on external events, connecting and closing down - to name the most important ones.
+
+The most important feat of connection filters is that they can be stacked on top of each other (or "chained" if you prefer that metaphor). In the common scenario that you want to retrieve a `https:` url with curl, you need 2 basic things to send the request and get the response: a TCP connection, represented by a `socket` and a SSL instance en- and decrypt over that socket. You write your request to the SSL instance, which encrypts and writes that data to the socket, which then sends the bytes over the network.
+
+With connection filters, curl's internal setup will look something like this (cf for connection filter):
+
+```
+Curl_easy *data         connectdata *conn        cf-ssl        cf-socket
++----------------+      +-----------------+      +-------+     +--------+
+|https://curl.se/|----> | properties      |----> | keys  |---> | socket |--> OS --> network
++----------------+      +-----------------+      +-------+     +--------+
+
+ Curl_write(data, buffer)
+  --> Curl_cfilter_write(data, data->conn, buffer)
+       ---> conn->filter->write(conn->filter, data, buffer)
+```
+
+While connection filters all do different things, they look the same from the "outside". The code in `data` and `conn` does not really know **which** filters are installed. `conn` just writes into the first filter, whatever that is.
+
+Same is true for filters. Each filter has a pointer to the `next` filter. When SSL has encrypted the data, it does not write to a socket, it writes to the next filter. If that is indeed a socket, or a file, or a HTTP/2 connection is of no concern to the SSL filter.
+
+And this allows the stacking, as in:
+
+```
+Direct:
+  http://localhost/      conn -> cf-socket
+  https://curl.se/       conn -> cf-ssl -> cf-socket
+Via http proxy tunnel:
+  http://localhost/      conn -> cf-http-proxy -> cf-socket 
+  https://curl.se/       conn -> cf-ssl -> cf-http-proxy -> cf-socket
+Via https proxy tunnel:
+  http://localhost/      conn -> cf-http-proxy -> cf-ssl -> cf-socket 
+  https://curl.se/       conn -> cf-ssl -> cf-http-proxy -> cf-ssl -> cf-socket
+Via http proxy tunnel via SOCKS proxy:
+  http://localhost/      conn -> cf-http-proxy -> cf-socks -> cf-socket 
+```
+
+### Connecting/Closing
+
+Before `Curl_easy` can send the request, the connection needs to be established. This means that all connection filters have done, whatever they need to do: waiting for the socket to be connected, doing the TLS handshake, performing the HTTP tunnel request, etc. This has to be done in reverse order: the last filter has to do its connect first, then the one above can start, etc.
+
+Each filter does in principle the following:
+
+```
+static CURLcode 
+myfilter_cf_connect(struct Curl_cfilter *cf,
+                    struct Curl_easy *data,
+                    bool *done)
+{
+  CURLcode result;
+
+  if(cf->connected) {            /* we and all below are done */
+    *done = TRUE;
+    return CURLE_OK;
+  }
+                                 /* Let the filters below connect */
+  result = cf->next->cft->connect(cf->next, data, blocking, done);
+  if(result || !*done)    
+    return result;               /* below errored/not finished yet */
+
+  /* MYFILTER CONNECT THINGS */  /* below connected, do out thing */
+  *done = cf->connected = TRUE;  /* done, remember, return */
+  return CURLE_OK;
+}
+```
+
+Closing a connection then works similar. The `conn` tells the first filter to close. Contrary to connecting,
+the filter does its own things first, before telling the next filter to close.
+
+### Efficiency
+
+There are two things curl is concerned about: efficient memory use and fast transfers.
+
+The memory footprint of a filter is relatively small:
+
+```
+struct Curl_cfilter {
+  const struct Curl_cftype *cft; /* the type providing implementation */
+  struct Curl_cfilter *next;     /* next filter in chain */
+  void *ctx;                     /* filter type specific settings */
+  struct connectdata *conn;      /* the connection this filter belongs to */
+  int sockindex;                 /* TODO: like to get rid off this */
+  BIT(connected);                /* != 0 iff this filter is connected */
+};
+```
+The filter type `cft` is a singleton, one static struct for each type of filter. The `ctx` is where a filter will hold its specific data. That varies by filter type. A http-proxy filter will keep the ongoing state of the CONNECT here, but free it after its has been established. The SSL filter will keep the `SSL*` (if OpenSSL is used) here until the connection is closed. So, this varies.
+
+`conn` is a reference to the connection this filter belongs to, so nothing extra besides the pointer itself.
+
+Several things, that before were kept in `struct connectdata`, will now go into the `filter->ctx` *when needed*. So, the memory footprint for connections that do *not* use a http proxy, or socks, or https will be lower.
+
+As to transfer efficiency, writing and reading through a filter comes at near zero cost *if the filter does not transform the data*. A http proxy or socks filter, once it is connected, will just pass the calls through. Those filters implementations will look like this:
+
+```
+ssize_t  Curl_cf_def_send(struct Curl_cfilter *cf, struct Curl_easy *data,
+                          const void *buf, size_t len, CURLcode *err)
+{
+  return cf->next->cft->do_send(cf->next, data, buf, len, err);
+}
+```
+The `recv` implementation is equivalent.
+
+## Filter Types
+
+The (currently) existing filter types are: SOCKET, SOCKET-ACCEPT, SSL, HTTP-PROXY and SOCKS-PROXY. Vital to establishing and read/writing a connection. But filters are also a good way to implement tasks for *managing* a connection:
+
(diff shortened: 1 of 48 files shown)


.gitignore
```diff
@@ -21,6 +21,7 @@
 .libs
 .project
 .settings
+.idea
```
xquery, line 24: nitpick - I've found its fine to do this with my local users gitignore instead of inject here ... otherwise we could add a quite a few more

docs/CONNECTION-FILTERS.md
```diff
@@ -0,0 +1,127 @@
...
+
+Connection filters is a design in the internals of curl, not visible in its public API. They were added 
+in curl v7.xx.x. This document describes the concepts, its high level implementation and the motivations.
+
+## Filters
+
+A "connection filter" is a piece of code that is responsible for handling a range of operations
+in regard to curl's connections: reading, writing, waiting on external events, connecting and closing down - to name the most important ones.
+
+The most important feat of connection filters is that they can bet stacked on top of each other (or "chained" if you prefer that metaphor). In the very common scenario that you want to retrieve a `https:` url with curl, you need 2 basic things to send the request and get the response: a TCP connection, represented by a `socket` and a SSL instance en- and decrypt over that socket. You write your request to the SSL instance, which encrypts and writes that data to the socket, which then sends the bytes over the network.
```
xquery, line 11: type 'bet' = 'be'

xquery: Generally I have an uneasy feeling about such a broad design concept being introduced ... for example saying 'stacked' or 'chained' does not provide enough concrete detail (for me) ... it is hinted at eg. it is called out the difference between transforming and non transforming pipelines (a set of filters implies a pipeline right ?)  .... I am not looking for academic rigor but even a scant mention of Jackson Structured Programming or 'anything' might help constrain the scope of what is being attempted here.

.gitignore
icing: will remove. found some .vs stuff there and thought...

docs/CONNECTION-FILTERS.md
```diff
@@ -0,0 +1,127 @@
...
+  BIT(connected);                /* != 0 iff this filter is connected */
+};
+```
+The filter type `cft` is a singleton, one static struct for each type of filter. The `ctx` is where a filter will hold its specific data. That varies by filter type. A http-proxy filter will keep the ongoing state of the CONNECT here, but free it after its has been established. The SSL filter will keep the `SSL*` (if OpenSSL is used) here until the connection is closed. So, this varies.
+
+`conn` is a reference to the connection this filter belongs to, so nothing extra besides the pointer itself.
+
+Several things, that before were kept in `struct connectdata`, will now go into the `filter->ctx` *when needed*. So, the memory footprint for connections that do *not* use a http proxy, or socks, or https will be lower.
+
+As to transfer efficiency, writing and reading through a filter comes at near zero cost *if the filter does not transform the data*. A http proxy or socks filter, once it is connected, will just pass the calls through. Those filters implementations will look like this:
```
xquery, line 100: more generically, efficiency in a pipeline of filters is related to 'polarity' ... understanding when 'pull' versus 'push' is better performing is key eg. there maybe things that are already performing well today that will suffer with this abstraction (or maybe thats just maxwell's demon whispering in my ear) ...  perhaps I have to read this entire PR to grok its scope (is this applicable to all protocols or just http)

docs/CONNECTION-FILTERS.md
icing: thanks for noticing the typo.

As to the "uneasy" comment, I am not sure what you are looking for. Another concept than the one implemented or a better description of it?

docs/CONNECTION-FILTERS.md
```diff
@@ -0,0 +1,127 @@
...
+                          const void *buf, size_t len, CURLcode *err)
+{
+  return cf->next->cft->do_send(cf->next, data, buf, len, err);
+}
+```
+The `recv` implementation is equivalent.
+
+## Filter Types
+
+The (currently) existing filter types are: SOCKET, SOCKET-ACCEPT, SSL, HTTP-PROXY and SOCKS-PROXY. Vital to establishing and read/writing a connection. But filters are also a good way to implement tasks for *managing* a connection:
```
xquery, line 113: is it a goal of this work to (eventually) refactor everything into filters ?

docs/CONNECTION-FILTERS.md
```diff
@@ -0,0 +1,127 @@
...
++----------------+      +-----------------+      +-------+     +--------+
+
+ Curl_write(data, buffer)
+  --> Curl_cfilter_write(data, data->conn, buffer)
+       ---> conn->filter->write(conn->filter, data, buffer)
+```
+
+While connection filters all do different things, they look the same from the "outside". The code in `data` and `conn` does not really know **which** filters are installed. `conn` just writes into the first filter, whatever that is.
+
+Same is true for filters. Each filter has a pointer to the `next` filter. When SSL has encrypted the data, it does not write to a socket, it writes to the next filter. If that is indeed a socket, or a file, or a HTTP/2 connection is of no concern to the SSL filter.
```
xquery, line 28: each 'filter' itself maybe a lego block though have you done analysis on impact on streamability eg. will we introduce logical filters that branch ?

docs/CONNECTION-FILTERS.md
bagder: The work puts the foundation. Where it makes sense and helps we can then convert more code over as we go along. IMHO.

docs/CONNECTION-FILTERS.md
xquery: no not another concept, just a more concrete description of the concept you are presenting ... pipelines of filters can be very attractive from a code organisation/syntax/composability pov ... just trying to grasp the operational impact ... and more importantly constraint on scope (esp on this initial commit).

docs/CONNECTION-FILTERS.md
bagder: I don't think we need to speculate too much about what we *might* do or not do in the future. This work is a refactor and remodel of how we deal with protocols and "layers" today. If we need to make it more fancy in the future, then let's do that in the future.

docs/CONNECTION-FILTERS.md
icing: I wanted to highlight the possibilities. If it makes send, for a concrete function, to do that is outside the scope of this. In general, my own opinion is that when a function should work for all protocol handlers the same, adding them as a filter maybe the easiest way in terms of interdependencies and needed code changes.

docs/CONNECTION-FILTERS.md
bagder: It is applicable for all protocols

icing: To explain a bit more: curl continues to work as before where it comes to IO, read/write/select handling, multi states, etc. This PR does not change that.

What this PR changes is how the functionality of, for example http proxy tunneling, is realized. Not how it behaves. Handlers do no longer need to have special code branches to check if proxy tunneling is in place or not. They call connect() until it is done or errors. And then send/receive as before.

docs/CONNECTION-FILTERS.md
icing: I am not able to prohibit or demand future changes to curl. I can (and will) not write a rule book here.

With the goal of Proxying over HTTP/2, the current implementation needs to change. The current way to "stacking" `SSL*`either on a socket or on another `SSL*` to not work for this. curl needs to be able to become more flexible in how connections are setup and handled. 

If we do a `https:` transfer over a HTTP/2 proxy tunnel, we need a SSL, talking to a HTTP/2 session, talking to an SSL instance, talking to a socket. This is a way to model such cases, so that we can combine these thing in the manner we need.

docs/CONNECTION-FILTERS.md
```diff
@@ -0,0 +1,127 @@
...
+
+And this allows the stacking, as in:
+
+```
+Direct:
```
xquery, line 33: question (mostly to aid my understanding) - will these filter pipelines be static or dynamically composable ?

icing: Right now, there is only the possibility to add filters. This is needed for STARTTLS. This is dynamic in the sense that, for example IMAP sets up its plain connection, connects() and starts talking. If it decides to do STARTTLS, it adds an SSL filter and calls connect() again until `done`. Then it can continue the conversation.

docs/CONNECTION-FILTERS.md
```diff
@@ -0,0 +1,127 @@
...
+the filter does its own things first, before telling the next filter to close.
+
+### Efficiency
+
+There are two things curl is concerned about: efficient memory use and fast transfers.
+
+The memory footprint of a filter is relatively small:
+
+```
+struct Curl_cfilter {
```
xquery, line 85: wondering out loud - would we ever surface up any generic filter data eg. does each filter have a concept of timeout or would we want to hook in a debug callback ?

docs/CONNECTION-FILTERS.md
xquery: somewhat related - would filters be 'pausable' ?

bagder:
Windows [build failures](https://ci.appveyor.com/project/curlorg/curl/builds/45324814/job/dha5ecwu517nq5q7#L318):
~~~
In file included from C:/projects/curl/lib/curl_setup.h:705,
                 from C:/projects/curl/lib/cfilters.c:25:
C:/projects/curl/lib/cfilters.c: In function 'Curl_cfilter_setup':
C:/projects/curl/lib/cfilters.c:291:12: error: 'index' undeclared (first use in this function)
  291 |            index, conn->bits.tunnel_proxy, conn->bits.httpproxy,
      |            ^~~~~
C:/projects/curl/lib/curl_setup_once.h:279:19: note: in definition of macro 'DEBUGF'
  279 | #define DEBUGF(x) x
      |                   ^
C:/projects/curl/lib/cfilters.c:291:12: note: each undeclared identifier is reported only once for each function it appears in
  291 |            index, conn->bits.tunnel_proxy, conn->bits.httpproxy,
      |            ^~~~~
C:/projects/curl/lib/curl_setup_once.h:279:19: note: in definition of macro 'DEBUGF'
  279 | #define DEBUGF(x) x
      |                   ^
make[2]: *** [lib/CMakeFiles/libcurl.dir/build.make:182: lib/CMakeFiles/libcurl.dir/cfilters.c.obj] Error 1

docs/CONNECTION-FILTERS.md
bagder: Filters as implemented here don't introduce any new features. Things in curl remain "pausable" after this PR to the same extent they are today: a little.

lib/cfilters.h
```diff
@@ -0,0 +1,206 @@
...
+                          const void *buf, size_t len, CURLcode *err);
+ssize_t  Curl_cf_def_recv(struct Curl_cfilter *cf, struct Curl_easy *data,
+                          char *buf, size_t len, CURLcode *err);
+void     Curl_cf_def_attach_data(struct Curl_cfilter *cf,
+                                 struct Curl_easy *data);
+void     Curl_cf_def_detach_data(struct Curl_cfilter *cf,
+                                 struct Curl_easy *data);
+
+
+CURLcode Curl_cfilter_create(struct Curl_cfilter **pcf,
```
bagder, line 139: Is there a pattern to when you use `_cf_` vs `_cfilter_` in symbol names? Would it not make sense to stick to `_cf_`  for consistency?

icing: I tried to differentiate between calls on `data, conn, index` and call on a cfilter itself.

vszakats:
I'd throw in the idea to use `cfilt` instead of `cfilter` (e.g. `Curl_cfilt_setup()`). Easier to type, shorter, still easy to grep, and IMO still unambigiously meaning "filter". Another alternative: Not shorter, still easier to type and also clarifying `c`, could be `connfilt` (e.g. `Curl_connfilt_setup()`). [ My 2cents, and feel free to disregard. ]

bagder (approved):

PR #9858, opened 2022-11-05
noproxy: tailmatch like in 7.85.0 and earlier

### LINKED ISSUE #9842: noproxy tail matching still has an issue
Contrary to https://github.com/curl/curl/blob/master/tests/unit/unit1614.c#L88, traditional behaviour for no_proxy is for example.com to also match host.example.com. 

### I did this

```
$ curl -V
curl 7.86.0 (x86_64-unknown-openbsd7.2) libcurl/7.86.0 LibreSSL/3.6.0 zlib/1.2.13 nghttp2/1.50.0
Release-Date: 2022-10-26
Features: alt-svc AsynchDNS HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe UnixSockets

$ no_proxy=google.com https_proxy=http://127.0.0.1:9999/ curl -sI https://www.google.com/ | head -5
curl: (7) Failed to connect to 127.0.0.1 port 9999 after 0 ms: Couldn't connect to server
```

### I expected the following

```
$ curl -V
curl 7.85.0 (x86_64-unknown-openbsd7.2) libcurl/7.85.0 LibreSSL/3.6.0 zlib/1.2.12 nghttp2/1.49.0
Release-Date: 2022-08-31
Features: alt-svc AsynchDNS HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe UnixSockets

$ no_proxy=google.com https_proxy=http://127.0.0.1:9999/ curl -sI https://www.google.com/ | head -5
HTTP/2 200 
content-type: text/html; charset=ISO-8859-1
p3p: CP="This is not a P3P policy! See g.co/p3phelp for more info."
date: Wed, 02 Nov 2022 03:13:12 GMT
server: gws
```

As also seen in at least lynx, w3m, OpenBSD ftp(1).

### curl/libcurl version

[see above]

### operating system

`OpenBSD $somehostname 7.2 GENERIC.MP#819 amd64`

bagder:
A regfression in 7.86.0 (via 1e9a538e05c010) made the tailmatch work differently than before. This restores the logic to how it used to work:

All names listed in NO_PROXY are tailmatched against the used domain name, if the lengths are identical it needs a full match.

Update the docs, update test 1614.

Reported-by: Stuart Henderson
Fixes #9842

### DIFF
--- a/docs/libcurl/opts/CURLOPT_NOPROXY.3
+++ b/docs/libcurl/opts/CURLOPT_NOPROXY.3
@@ -40,10 +40,6 @@ list is matched as either a domain which contains the hostname, or the
 hostname itself. For example, "ample.com" would match ample.com, ample.com:80,
 and www.ample.com, but not www.example.com or ample.com.org.
 
-If the name in the \fInoproxy\fP list has a leading period, it is a domain
-match against the provided host name. This way ".example.com" will switch off
-proxy use for both "www.example.com" as well as for "foo.example.com".
-
 Setting the \fInoproxy\fP string to "" (an empty string) will explicitly
 enable the proxy for all host names, even if there is an environment variable
 set for it.
--- a/lib/noproxy.c
+++ b/lib/noproxy.c
@@ -187,22 +187,24 @@ bool Curl_check_noproxy(const char *name, const char *no_proxy)
             tokenlen--;
 
           if(tokenlen && (*token == '.')) {
-            /* A: example.com matches '.example.com'
-               B: www.example.com matches '.example.com'
-               C: nonexample.com DOES NOT match '.example.com'
-            */
-            if((tokenlen - 1) == namelen)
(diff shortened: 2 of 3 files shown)


sthen:
Many thanks, this fixes the things I use, and matches a table of behaviour of older versions that I found in https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/ other than the intentional change to use CIDR.

For the diff to the test, would it be worth keeping that test line but changing to TRUE? i.e.

`{ "www.example.com", "localhost,example.com,.example.de", TRUE},`

bagder:
> For the diff to the test, would it be worth keeping that test line but changing to TRUE?

Thanks, I added that as well just now.

Issue #9840, opened 2022-11-01
hyper: These test cases failed: 209 213 1008 


bagder:
As of recently, these three tests seem to fail in every CI job that builds with hyper enabled.

[Example log](https://github.com/curl/curl/actions/runs/3367179038/jobs/5584421315)

PR #9860, opened 2022-11-07
hyper, forget handled task ref

icing:
Hyper: fix handling of hyper_task's when they reuse the same memory address.


bagder (approved):

bagder:
Fixes #9840

PR #9861, opened 2022-11-07
websockets: fix handling of partial frames

icing:
buffer used and send length calculations are fixed when a partial websocket frame has been received.


bagder (approved):

bagder:
Thanks!

PR #9863, opened 2022-11-07
cookie: compare cookie prefixes case insensitively

bagder:
Adapted to language in rfc6265bis draft-11.


danielgustafsson (approved):
LGTM

PR #9870, opened 2022-11-08
Fix RTSP basic auth

bagder:
known bug 6.8

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -696,6 +696,15 @@ CURLcode Curl_http_auth_act(struct Curl_easy *data)
     data->req.newurl = strdup(data->state.url); /* clone URL */
     if(!data->req.newurl)
       return CURLE_OUT_OF_MEMORY;
+#ifndef CURL_DISABLE_RTSP
+    /*
+     * Authentication is treated as a redirect in Curl_follow(), so if this is
+     * done using RTSP we make it allow these "redirects" to RTSP (only). A
+     * safe assumption as no other redirects should happen from RTSP.
+     */
+    if(conn->handler->protocol & CURLPROTO_RTSP)
+      data->set.redir_protocols = CURLPROTO_RTSP;
+#endif
   }
   else if((data->req.httpcode < 300) &&
           (!data->state.authhost.done) &&
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -73,7 +73,6 @@ problems may have been fixed or changed somewhat since this was written.
  6.5 NTLM does not support password with § character
  6.6 libcurl can fail to try alternatives with --proxy-any
  6.7 Do not clear digest for single realm
- 6.8 RTSP authentication breaks without redirect support
  6.9 SHA-256 digest not supported in Windows SSPI builds
  6.10 curl never completes Negotiate over HTTP
  6.11 Negotiate on Windows fails
@@ -617,15 +616,6 @@ problems may have been fixed or changed somewhat since this was written.
 6.7 Do not clear digest for single realm
 
  https://github.com/curl/curl/issues/3267
-
-6.8 RTSP authentication breaks without redirect support
-
- RTSP authentication broke in 7.66.0. A work-around is to enable RTSP in
- CURLOPT_REDIR_PROTOCOLS. Authentication should however not be considered an
- actual redirect so a "proper" fix needs to be different and not require users
- to allow redirects to RTSP to work.
-
- See https://github.com/curl/curl/pull/4750
 
 6.9 SHA-256 digest not supported in Windows SSPI builds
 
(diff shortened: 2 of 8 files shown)


janssen70:
Once more a comment on closed PR... Seems to make sense though
Given that there has been discussion on strdup-usage before:
The unconditional free/strdup implies that on an rtsp-interleaved link (i.e. one that repeats CURL_RTSPREQ_RECEIVE) memory overhead happens a few tens to a few thousand times per second depending on payload size and presence (absence) of server-side interleaving optimizations. From that point of view it's better to assign first_host only when it is not set.

bagder:
Agreed. It's even strange that it can update the first host...

lib/http.c
```diff
@@ -696,6 +696,15 @@ CURLcode Curl_http_auth_act(struct Curl_easy *data)
...
     if(!data->req.newurl)
       return CURLE_OUT_OF_MEMORY;
+#ifndef CURL_DISABLE_RTSP
+    /*
+     * Authentication is treated as a redirect in Curl_follow(), so if this is
+     * done using RTSP we make it allow these "redirects" to RTSP (only). A
+     * safe assumption as no other redirects should happen from RTSP.
+     */
+    if(conn->handler->protocol & CURLPROTO_RTSP)
+      data->set.redir_protocols = CURLPROTO_RTSP;
```
monnerat, line 706: This probably breaks reuse of handle with another protocol.
If handle is used for RTSP then for HTTP, redirection won't work on HTTP unless reset externally between both uses.
I would rather use a copy of `set.redir_protocols` into a new `state.redir_protocol` field in the whole library.

bagder: Yikes, that is indeed a good remark. I need to fix this...

bagder: The RTSP code also wrongly sets `data->set.opt_no_body` in several places...

monnerat: I did not see ithem before but you're right. I noticed this particular one while rebasing the sieve PR, as it also features redirection.

I've just done a quick check for more and there may be some (although not all listed):

```
$ find lib -name '*.c' | xargs grep -e '->set\.[A-Za-z0-9_.]* *= ' | grep -Fv 'setopt.c
> url.c'
lib/easy.c:    dst->set.postfields = dst->set.str[i];
lib/easy.c:  outcurl->set.buffer_size = data->set.buffer_size;
lib/http.c:      data->set.redir_protocols = CURLPROTO_RTSP;
lib/ftp.c:    data->set.ftp_filemethod = FTPFILE_MULTICWD;
lib/ftp.c:  data->set.fwrite_func = Curl_ftp_parselist;
lib/ftp.c:  data->set.out = data;
lib/ftp.c:      data->set.fwrite_func = ftpwc->backup.write_function;
lib/ftp.c:      data->set.out = ftpwc->backup.file_descriptor;
lib/doh.c:    doh->set.fmultidone = doh_done;
lib/doh.c:    doh->set.dohfor = data; /* identify for which transfer this is done */
lib/rtsp.c:  data->set.opt_no_body = TRUE; /* most requests don't contain a body */
lib/rtsp.c:    data->set.opt_no_body = FALSE;
lib/rtsp.c:    data->set.opt_no_body = FALSE;
lib/rtsp.c:    data->set.opt_no_body = FALSE;
lib/rtsp.c:      data->set.opt_no_body = TRUE;
lib/transfer.c:      data->set.trailer_data = NULL;
lib/transfer.c:      data->set.trailer_callback = NULL;
lib/transfer.c:      data->set.upload = false;
lib/http2.c:        node->data->set.stream_depends_on = child;
lib/http2.c:      parent->set.stream_dependents = 0;
lib/http2.c:      (*tail)->data->set.stream_depends_e = FALSE;
lib/http2.c:  child->set.stream_depends_on = parent;
lib/http2.c:  child->set.stream_depends_e = exclusive;
lib/http2.c:      parent->set.stream_dependents = data->next;
lib/http2.c:  child->set.stream_depends_on = 0;
lib/http2.c:  child->set.stream_depends_e = FALSE;
lib/multi.c:  data->state.conn_cache->closure_handle->set.timeout = data->set.timeout;
lib/conncache.c:  connc->closure_handle->set.buffer_size = READBUFFER_MIN;
lib/vtls/vtls.c:  data->set.general_ssl.max_ssl_sessions = amount;
$
```
Maybe we can plan to have only setopt values in `set` and r/w runtime values in `state`? Just an idea...

bagder: > Maybe we can plan to have only setopt values in set and r/w runtime values in state?

I think we should, yes. And then possibly invent and set up some some kind of framework/markup or something that lets us scan for mistakes like the present.

bagder: #9888 addresses parts of this

monnerat: > And then possibly invent and set up some some kind of framework/markup

Suggestion:

- declare `set` as `const`
- use a cast with a comment in the few function allowed to alter it.

Not very elegant but will cause a compile error if such a mistake occurs again. The C language does not offer many alternatives to achieve that. And it does not require any external test or tool.

PR #9871, opened 2022-11-08
KNOWN_BUGS cleanup

bagder:
A range of items mentioned in the document that I think are not actually bugs and thus do not belong therein.

### DIFF
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -51,7 +51,6 @@ problems may have been fixed or changed somewhat since this was written.
  5.4 Build with statically built dependency
  5.5 cannot handle Unicode arguments in non-Unicode builds on Windows
  5.6 make distclean loops forever
- 5.7 Visual Studio project gaps
  5.8 configure finding libs in wrong directory
  5.9 Utilize Requires.private directives in libcurl.pc
  5.10 curl hangs on SMB upload over stdin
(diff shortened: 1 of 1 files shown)


micheljung:
As for 1.2, [it's not true that this has NEVER been reported](https://curl.se/mail/lib-2020-01/0063.html)

bagder:
Ah wow I clearly forgot about that: so wrong of me. It *has* been reported, but it is still such a very rare thing that I do not consider it a bug.

micheljung:
okay, so whether something is a bug or not is not decided by if it works according to [the RFC that even has an explicit example for this](https://www.rfc-editor.org/rfc/rfc9110.html#section-11.6.1-6) but by how many people report it? 😀

It cost us many man hours and forced us to come up with an ugly workaround which is still in production today just to deal with this non-bug. I'd rather have spent a fraction of that sponsoring you or someone else to fix it. Unfortunately, that was not my decision to make :(

bagder:
Sort of, yes. It is now re-labeled as "a known implementation restriction", not a bug.  It does not serve the project well to keep items in the known-bugs list that none of us intends to fix and that none of the project developers actually considers a bug.

We need to stay pragmatic, we can never read or act on RFC specifications by their exact words only. curl does not live in a vacuum.

This does not rule out that a future version of curl could have this auth header handling improved.

jay:
It's tricky because the comma in WWW-Authenticate is used in the scheme as well as to separate schemes, which means you have to parse each scheme in order to know when the next scheme starts. You can see the code in Curl_http_input_auth does loop for multiple schemes but I guess it would need to be improved.

https://github.com/curl/curl/blob/cd95ee9f771361acf241629d2fe5507e308082a2/lib/http.c#L1001-L1139

Take your example
~~~
curl -kv -L --negotiate -u : -b ~/cookie.txt -c ~/cookie.txt http://localhost:8081/negotiate
~~~
I would think in this case curl sends an Authorization for the first request, then if it gets this in reply:

WWW-Authenticate: Negotiate, Basic realm="TM1"

It assumes failure. Interesting it passes on `", Basic realm="TM1"` as the negotiate data, which can't possibly be right according to [rfc 4559](https://datatracker.ietf.org/doc/html/rfc4559#page-3). That causes Curl_input_negotiate to return a bad encoding error instead of a login denied error.

I think he has a point that this should be in known bugs even if we aren't working on it. This is a known bug that libcurl does not handle multiple auth on the same line well, even though it may try (it does loop around, I don't know what good that does in cases with multiple options per scheme)

PR #9872, opened 2022-11-08
config-win32: fix SIZEOF_OFF_T for MSVC and old MinGW

vszakats:
The previously set default value of 8 (64-bit) is only correct for mingw-w64 and only when `_FILE_OFFSET_BITS` is set to 64 (the default when building curl). For MSVC, old MinGW and other Windows compilers, the correct value is 4 (32-bit). Adjust condition accordingly. Also drop the manual override option.

Regression in 7.86.0 (from 68fa9bf3f5d7b4fcbb57619f70cb4aabb79a51f6)

Bug: https://github.com/curl/curl/pull/9712#issuecomment-1307330551
Reported-by: Peter Piekarski
Closes #xxxx

### DIFF
--- a/lib/config-win32.h
+++ b/lib/config-win32.h
@@ -332,8 +332,8 @@
 #define SIZEOF_CURL_OFF_T 8
 
 /* Define to the size of `off_t', as computed by sizeof. */
-#ifndef SIZEOF_OFF_T
-#define SIZEOF_OFF_T 8
+#if defined(__MINGW32__)
+#  define SIZEOF_OFF_T 8
 #endif
 
 /* ---------------------------------------------------------------- */
--- a/lib/config-win32.h
+++ b/lib/config-win32.h
@@ -333,7 +333,7 @@
 
 /* Define to the size of `off_t', as computed by sizeof. */
 #if defined(__MINGW32__)
-#  define SIZEOF_OFF_T 8
+#define SIZEOF_OFF_T 8
 #endif
 
 /* ---------------------------------------------------------------- */
--- a/lib/config-win32.h
+++ b/lib/config-win32.h
@@ -331,11 +331,6 @@
 /* Define to the size of `curl_off_t', as computed by sizeof. */
 #define SIZEOF_CURL_OFF_T 8
 
-/* Define to the size of `off_t', as computed by sizeof. */
-#if defined(__MINGW32__)
-#define SIZEOF_OFF_T 8
-#endif
-
 /* ---------------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


lib/config-win32.h
```diff
@@ -332,7 +332,7 @@
 #define SIZEOF_CURL_OFF_T 8
 
 /* Define to the size of `off_t', as computed by sizeof. */
-#ifndef SIZEOF_OFF_T
+#if defined(__MINGW32__)
```
jay, line 335: Besides MSVC, MinGW and MinGW-w64, what other compilers could use config-win32? If we know the size the compilers will use I think we should be explicit about it. For example

```suggestion
#if defined(_MSC_VER)
#define SIZEOF_OFF_T 4
#elif defined(__MINGW32__)
```

jay: Also note original mingw uses 32-bit off_t, so this would have to be further revised to only account when mingw-w64 is using a 64-bit off_t which may be all? some? of the time

vszakats: @jay: Thanks for the old-mingw info, I keep forgetting about it. As for the rest, do you think it's best to explicitly list all platforms that need 4?

Like so:
```c
#if defined(_MSC_VER)
#define SIZEOF_OFF_T 4
#elif defined(__MINGW64_VERSION_MAJOR)
#define SIZEOF_OFF_T 8
#elif defined(__MINGW32__)
#define SIZEOF_OFF_T 4
#endif
```

Or we could stay short, like:
```c
#if defined(__MINGW64_VERSION_MAJOR)
#define SIZEOF_OFF_T 8
#else
#define SIZEOF_OFF_T 4
#endif
```

jay: If you are sure that MinGW-w64 uses off_t 64 bits. MS data type for off_t is 32 bits.

vszakats: I verified, and it's 64 bits when `_FILE_OFFSET_BITS == 64` (otherwise it's 32). This is so for mingw-w64 1.0 and later releases (= all releases), thanks to this [commit](https://github.com/Alexpux/mingw-w64/commit/4d3b28a9929ea58511e7165cb7eb1bcdd01151ad) (from 2011-02-19).

This should cover everything we know so far:
```c
#if defined(__MINGW64_VERSION_MAJOR) && \
  defined(_FILE_OFFSET_BITS) && (_FILE_OFFSET_BITS == 64)
#define SIZEOF_OFF_T 8
#else
#define SIZEOF_OFF_T 4
#endif
```
and moving the block under the defaulting logic for `_FILE_OFFSET_BITS`.

jay (approved):

Issue #9873, opened 2022-11-08
segmentation fault using a easy handle after CURLE_WRITE_ERROR


Edrusb:
### I did this

the application I maintain and that relies on libcurl sometime needs to stop reading file content from ftp/sftp depending on the data read so far. 

For years it was implemented by CURLOPT_WRITEFUNCTION  + CURLOPT_WRITEDATA  and the pointed to callback returned 0 instead of the size*nmemb value when reading needed to stop (sftp and ftp protocols).

Today, if doing that way still works as expected, (curl_easy_perform() returns the expected CURLE_WRITE_ERROR code) any subsequent further action on the handle triggers a segmentation fault with the following stack (if that can help):

```
(gdb) n
[New Thread 0x7ffff7292700 (LWP 25812)]

Thread 5 "dar" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7292700 (LWP 25812)]
0x00007ffff75bb298 in ?? () from /lib/x86_64-linux-gnu/libnss_files.so.2
(gdb) bt
#0  0x00007ffff75bb298 in ?? () from /lib/x86_64-linux-gnu/libnss_files.so.2
#1  0x00007ffff75bc544 in _nss_files_gethostbyname4_r () from /lib/x86_64-linux-gnu/libnss_files.so.2
#2  0x0000000000c05a76 in gaih_inet.constprop ()
#3  0x0000000000c068f5 in getaddrinfo ()
#4  0x000000000067d3c4 in Curl_getaddrinfo_ex ()
#5  0x0000000000675e83 in getaddrinfo_thread ()
#6  0x000000000067decb in curl_thread_create_thunk ()
#7  0x00000000008d9b97 in start_thread (arg=<optimized out>) at pthread_create.c:477
#8  0x0000000000c0cb5f in clone ()
(gdb)
```

I tried to curl_easy_reset() first and reapplying all option (curl_easy_setopt()) on this handle, this just leads to the same fate...
I tried also after a curl_easy_cleanup() + curl_easy_init() ... same fate...

### I expected the following
if I'm doing something wrong which might still be possible, an error message would be welcome ;)

### curl/libcurl version

```
curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.85.0 OpenSSL/1.1.1n zlib/1.2.11 zstd/1.4.8 libssh2/1.9.0
Release-Date: 2020-12-09
[...]

PR #9874, opened 2022-11-08
lib: add CURL_WRITEFUNC_ERROR to signal write callback error

jay:
Prior to this change if the user wanted to signal an error from their write callbacks they would have to use logic to return a value different from the number of bytes (nmemb) passed to the callback. Also, the inclination of some users has been to just return 0 to signal error, which is incorrect as that may be the number of bytes passed to the callback.

To remedy this the user can now return CURL_WRITEFUNC_ERROR instead.

Ref: https://github.com/curl/curl/issues/9873

Closes #xxxx


bagder (approved):
A good idea!

PR #9882, opened 2022-11-10
rtsp: only store first_host once

bagder:
Suggested-by: Erik Janssen @janssen70
URL: https://github.com/curl/curl/pull/9870#issuecomment-1309499744


janssen70:
Perfect :)
With the latest changes rtsp becomes rather okay!

PR #9890, opened 2022-11-11
Websocket fixes for partial frames and buffer updates.

icing:
- buffers updated correctly when handling partial frames
- callbacks no longer invoked for incomplete payload data of 0 length
- curl_ws_recv no longer returns with 0 length partial payload


bagder (approved):

bagder:
Thanks!

PR #9894, opened 2022-11-11
Test telnet server close

icing:
Slight change in test telnet server code to have it invoke its close() method.


bagder:
This round at least didn't cause any telnet issues. Should we re-run any particular test a few times to be more sure or do you think this is likely to actually be the proper fix?

icing:
In any case, it is a better test code as it calls `close()` on the server object, as an application is expected to. I could imagine the Python internals to abort sockets not properly closed in deconstruction. But I have not checked.

I have locally not seen the telnet test fails on my mac with this change. I'd say, merge it and we revisit such we see another telnet CONNRESET failure.

bagder:
Excellent, thanks!

PR #9895, opened 2022-11-11
vtls: fix build without proxy support

bagder:
Follow-up to dafdb20a26d0c890

/cc @icing


icing:
Ah, right. Nice catch.

PR #9897, opened 2022-11-11
log2changes.pl: wrap long lines at 80 columns

### LINKED ISSUE #9896: OS400 sh makefile.sh > makelog 2>&1 fails.
File CHANGES has line(s) that are longer that 240 bytes. It can not be copied to ${TARGETLIB]/DOCS, makefile.sh ends with these lines in makelog:
CPC7305:  Member CHANGES added to file DOCS in CURLIBMI1.
CPDA037:  Could not write to member CHANGES in file DOCS in library CURLIBMI1.
CPC7309:  Member CHANGES removed from file DOCS in CURLIBMI1.
CPFA0D4:  File system error occurred.  Error number 3003.

Editing the file to remove the long line(s) allows it to work.

bagder:
Also, only use author names in the output.

Fixes #9896
Reported-by: John Sherrill

### DIFF
--- a/scripts/log2changes.pl
+++ b/scripts/log2changes.pl
@@ -43,6 +43,9 @@ sub printmsg {
         print $p.substr($msg, 0, 77, "")."\n";
         $p="  ";
     }
+    if($msg eq "") {
+        $p = "";
+    }
     print "$p$msg\n";
 }
 
--- a/scripts/log2changes.pl
+++ b/scripts/log2changes.pl
@@ -37,6 +37,15 @@ sub nicedate {
     return $date;
 }
 
+sub printmsg {
+    my ($p, $msg)=@_;
+    while(length($msg) > 77) {
+        print $p.substr($msg, 0, 77, "")."\n";
+        $p="  ";
+    }
+    print "$p$msg\n";
+}
+
 print
 '                                  _   _ ____  _
                               ___| | | |  _ \| |
@@ -47,7 +56,6 @@ print
                                   Changelog
 ';
 
-my $line;
 my $tag;
 while(<STDIN>) {
     my $l = $_;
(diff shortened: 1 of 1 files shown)


scripts/log2changes.pl
```diff
@@ -37,6 +37,15 @@ sub nicedate {
     return $date;
 }
 
+sub printmsg {
+    my ($p, $msg)=@_;
+    while(length($msg) > 77) {
+        print $p.substr($msg, 0, 77, "")."\n";
+        $p="  ";
```
vszakats, line 44: To avoid outputting lines with spaces-only (untested):
```perl
$o=substr($msg, 0, 77, "");
if $o ne "" {
   $o=$p.$o
}
print $o."\n";
```

Fixed earlier in fde1cb9247792dfeeb6e44a07c84ed3957cb4707.

bagder: I can't see them actually occurring? I tried this:
~~~diff
diff --git a/scripts/log2changes.pl b/scripts/log2changes.pl
index f4c06b750..fbaf177a9 100755
--- a/scripts/log2changes.pl
+++ b/scripts/log2changes.pl
@@ -41,11 +41,12 @@ sub printmsg {
     my ($p, $msg)=@_;
     while(length($msg) > 77) {
         print $p.substr($msg, 0, 77, "")."\n";
         $p="  ";
     }
-    print "$p$msg\n";
+    print "$p$msg\n" if($msg !~ /^\s+\z/);
+    print "HERE\n" if($msg =~ /^\s+\z/);
 }
 
~~~
... but I don't see any `HERE` output when I try this?

vszakats: I'm not sure why the test doesn't trigger, but the spaces are definitely there in the output:
Screen Shot 2022-11-14 at 11 31 24

Also, my patch was incomplete, here's one that fixes the above (and multiple issues with my patch):
```diff
--- log2changes-a.pl	2022-11-14 10:23:38 +0000
+++ log2changes-b.pl	2022-11-14 10:27:51 +0000
@@ -40,9 +40,16 @@
 sub printmsg {
     my ($p, $msg)=@_;
     while(length($msg) > 77) {
-        print $p.substr($msg, 0, 77, "")."\n";
+        $o=substr($msg, 0, 77, "");
+        if($o eq "") {
+            $p="";
+        }
+        print $o."\n";
         $p="  ";
     }
+    if($msg eq "") {
+        $p="";
+    }
     print "$p$msg\n";
 }
```

bagder: Ah, now I get it. I think it can be simplified to:
~~~difff
index f4c06b750..ee8a753eb 100755
--- a/scripts/log2changes.pl
+++ b/scripts/log2changes.pl
@@ -41,10 +41,13 @@ sub printmsg {
     my ($p, $msg)=@_;
     while(length($msg) > 77) {
         print $p.substr($msg, 0, 77, "")."\n";
         $p="  ";
     }
+    if($msg eq "") {
+        $p = "";
+    }
     print "$p$msg\n";
 }
 
 print
~~~
... since it it isn't likely that lines longer than 77 columns would return a "" (or even spaces-only) for those initial 77 columns.

PR #9898, opened 2022-11-12
ntlm: silence ubsan warning about copying from null target_info pointer.

nmoinvaz:
runtime error: null pointer passed as argument 2, which is declared to never be null

#0 0xe7909c in Curl_ntlm_core_mk_ntlmv2_resp curl/lib/curl_ntlm_core.c:661:3


bagder (approved):

bagder:
Thanks!

PR #9903, opened 2022-11-13
ntlm: fix typo in comment of encrypt_des

rybak:


lib/curl_ntlm_core.c
```diff
@@ -188,7 +188,7 @@ static void setup_des_key(const unsigned char *key_56,
 /*
  * Expands a 56 bit key KEY_56 to 64 bit and encrypts 64 bit of data, using
  * the expanded key.  The caller is responsible for giving 64 bit of valid
- * data is IN and (at least) 64 bit large buffer as OUT.
+ * data as IN and (at least) 64 bit large buffer as OUT.
```
bagder, line 191: It still reads a little quirky I think. How about rewriting that sentence to:

IN points to 64 bit source data, OUT points to a 64 bit output buffer.

PR #9927, opened 2022-11-16
mime: relax easy/mime structures binding

monnerat:
Recent changes to the library have released the strict need for an early binding of mime structures to an easy handle.

This constraint currently forces to create the handle before the mime structure and the latter cannot be attached to another handle once created (see https://curl.se/mail/lib-2022-08/0027.html).

This commit removes the handle pointers from the mime structures allowing more flexibility on their use.

When an easy handle is duplicated, bound mime structures must however still be duplicated too as their components hold send-time dynamic information.

### DIFF
--- a/docs/libcurl/curl_mime_init.3
+++ b/docs/libcurl/curl_mime_init.3
@@ -31,11 +31,14 @@ curl_mime_init - create a mime handle
 curl_mime *curl_mime_init(CURL *easy_handle);
 .fi
 .SH DESCRIPTION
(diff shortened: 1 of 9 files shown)


bagder (approved):

jay:
> "Recent changes to the library have released the strict need for an early binding of mime structures to an easy handle."

Can you please cite those changes as references in the commit, or just briefly explain what they are?

monnerat:
> Can you please cite those changes as references in the commit

Done.

bagder:
Thanks!

PR #9928, opened 2022-11-16
github-actions: clarify workflows permissions, set least possible privilege

diogoteles08:
Set top-level permissions to None on all workflows and then set per-job permissions, giving only the necessary ones. This avoids that new jobs inherit unwanted privileges.

Previously most of the workflows did not have written permissions, so their permissions were depending on the permission set as default on the Github repo settings

Discussion: https://curl.se/mail/lib-2022-11/0028.html

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -13,11 +13,11 @@ on:
     branches:
     - master
 
-permissions: {}
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
+
+permissions: {}
 
 jobs:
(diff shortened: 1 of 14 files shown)


.github/workflows/fuzz.yml
```diff
@@ -13,6 +13,8 @@ on:
     branches:
     - master
 
+permissions: {}
```
mback2k, line 16: Please put in the same order on the job files, between concurrency and jobs sections, thanks.

bagder:
@cmeister2 @mback2k any objections to me merging this ?

mback2k (approved):
LGTM

bagder:
Thanks!

PR #9932, opened 2022-11-17
cookie: expire cookies at once when max-age is negative

### LINKED ISSUE #9930: curl doesn't expire cookies with negative Max-Age
I noticed that when curl receives a `Set-Cookie` header with `Max-Age=-1` (or any negative number) it will not immediately expire the cookie.  Rather, the cookie will stay in the cookie jar when curl exits.

I believe that this is related to c3fdecad which works only for `Max-Age=0`, see Issue #3351.

However, @badger quotes RFC6265 in [this comment](https://github.com/curl/curl/issues/3351#issuecomment-445470627) as stating that: 

> If delta-seconds is less than or equal to zero (0) ....

In my opinion, patch c3fdecad may work only if delta-seconds is equal to zero and not if it is less than.  I have verified this with a server: when I send `Max-Age: 0`, the cookie is immediately removed from the cookie jar, whereas if I set `Max-Age: -1` it is not removed until the next time I run `curl -c ...`

My output with `Max-Age: -1`
```
* Added cookie auth_token="" for domain localhost, path /, expire 0
< Set-Cookie: auth_token=; Path=/; Max-Age=-1; HttpOnly
...
```
next invocation:
```
> Cookie: auth_token=
```
and the cookie jar contains:
```
#HttpOnly_localhost	FALSE	/	FALSE	0	auth_token	
```

By contrast, `Max-Age=0` yields:
```
* Added cookie auth_token="" for domain localhost, path /, expire 1
< Set-Cookie: auth_token=; Path=/; Max-Age=0; HttpOnly
```
and the cookie is gone from the cookie jar.

bagder:
Update test 329 to verify

Reported-by: godmar on github
Fixes #9930

### DIFF
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -779,10 +779,16 @@ Curl_cookie_add(struct Curl_easy *data,
       offt = curlx_strtoofft((*co->maxage == '\"')?
                              &co->maxage[1]:&co->maxage[0], NULL, 10,
                              &co->expires);
-      if(offt == CURL_OFFT_FLOW)
+      switch(offt) {
+      case CURL_OFFT_FLOW:
         /* overflow, used max value */
         co->expires = CURL_OFF_T_MAX;
-      else if(!offt) {
+        break;
+      case CURL_OFFT_INVAL:
+        /* negative or otherwise bad, expire */
+        co->expires = 1;
+        break;
+      case CURL_OFFT_OK:
         if(!co->expires)
           /* already expired */
           co->expires = 1;
@@ -791,6 +797,7 @@ Curl_cookie_add(struct Curl_easy *data,
           co->expires = CURL_OFF_T_MAX;
         else
           co->expires += now;
+        break;
       }
     }
     else if(co->expirestr) {
--- a/tests/data/test329
+++ b/tests/data/test329
@@ -14,6 +14,7 @@ Date: Tue, 09 Nov 2010 14:49:00 GMT
 Server: test-server/fake
 Content-Type: text/html
 Funny-head: yesyes
+Set-Cookie: testn1=yes; path=/we/want/; domain=.host.foo.com; Max-Age=-1;
 Set-Cookie: test=yes; path=/we/want/; domain=.host.foo.com; Max-Age=0;
 Content-Length: 4
 


lib/cookie.c
```diff
@@ -779,10 +779,16 @@ Curl_cookie_add(struct Curl_easy *data,
...
-      if(offt == CURL_OFFT_FLOW)
+      switch(offt) {
+      case CURL_OFFT_FLOW:
         /* overflow, used max value */
         co->expires = CURL_OFF_T_MAX;
-      else if(!offt) {
+        break;
+      case CURL_OFFT_INVAL:
+        /* negative or otherwise bad, expire */
+        co->expires = 1;
```
jay, line 789: It was unexpected to me that curlx_strtoofft does not parse negative numbers, what is the reason for that?

bagder: I can't recall the exact reasoning anymore, but I did look over all current uses of this function and I could not spot any other code path that would expect a negative value to pass through fine.

danielgustafsson (commented):
LGTM, but don't you need this hunk as well?:
```diff
diff --git a/lib/cookie.c b/lib/cookie.c
index a3c699d48..c8aba77a8 100644
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -731,11 +731,11 @@ Curl_cookie_add(struct Curl_easy *data,
            * Optional.  The Max-Age attribute defines the lifetime of the
            * cookie, in seconds.  The delta-seconds value is a decimal non-
            * negative integer.  After delta-seconds seconds elapse, the
-           * client should discard the cookie.  A value of zero means the
-           * cookie should be discarded immediately.
+           * client should discard the cookie.  A value of zero, or a negative
+           * value, means the cookie should be discarded immediately.
            */
           strstore(&co->maxage, whatptr);
-          if(!co->maxage) {
+          if(!co->maxage <= 0) {
             badcookie = TRUE;
             break;
           }

```

bagder:
> LGTM, but don't you need this hunk as well?:

No, because that code just copies the max-age string, it is actually parsed further down. The check you modified just checks if the clone of the data worked.

I think this is done this way so that max-age gets preference over expires.

lib/cookie.c
jay: #9934. Also I notice MS doc for _strtoi64 says it does not set ERANGE on overflow/underflow so I will investigate further.

jay: > I could not spot any other code path that would expect a negative value to pass through fine.

hm appears I misread..

danielgustafsson:
> > LGTM, but don't you need this hunk as well?:
> 
> No, because that code just copies the max-age string, it is actually parsed further down. The check you modified just checks if the clone of the data worked.
> 
> I think this is done this way so that max-age gets preference over expires.

Ah yes, of course, sorry for the noise.

PR #9934, opened 2022-11-17
strtoofft: support parsing negative numbers in curlx_strtoofft

jay:
Prior to this change curlx_strtoofft would return CURL_OFFT_INVAL when the input string was a negative number.

Ref: https://github.com/curl/curl/pull/9932/files#r1024837115

Closes #xxxx


bagder:
but why? This just risks opening up for flaws. Like this now brings back the cookie parser mistake I fix in #9932

jay:
Yeah I misread your comment in the other PR.

PR #9935, opened 2022-11-17
haproxy build fix for configs with proxy but no http

icing:
Fixing haproxy filter to be only available when PROXY and HTTP are enabled in curl configuration. We build this config in an azure CI job `ubuntu disable_http_smtp_imap`.


bagder (approved):

bagder:
Thanks!

PR #9943, opened 2022-11-18
Revert "sendf: change Curl_read_plain to wrap Curl_recv_plain"

bagder:
This reverts commit 12e1def51a75392df62e65490416007d7e68dab9.

It introduced SOCKS proxy fails, like test 700 never ending.

Reopens #9431 

/cc @jay

### DIFF
--- a/lib/krb5.c
+++ b/lib/krb5.c
@@ -454,14 +454,14 @@ static int ftp_send_command(struct Curl_easy *data, const char *message, ...)
 /* Read |len| from the socket |fd| and store it in |to|. Return a CURLcode
    saying whether an error occurred or CURLE_OK if |len| was read. */
 static CURLcode
-socket_read(struct Curl_easy *data, curl_socket_t fd, void *to, size_t len)
+socket_read(curl_socket_t fd, void *to, size_t len)
 {
   char *to_p = to;
   CURLcode result;
   ssize_t nread = 0;
 
   while(len > 0) {
-    result = Curl_read_plain(data, fd, to_p, len, &nread);
+    result = Curl_read_plain(fd, to_p, len, &nread);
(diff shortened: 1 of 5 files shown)


icing:
Several tests failing for me: 250, 251, 700, ...

The fix for 700 is to add:
```
@@ -471,6 +471,7 @@ ssize_t Curl_recv_plain(struct Curl_easy *data, int num, char *buf,
       ) {
       /* this is just a case of EWOULDBLOCK */
       *code = CURLE_AGAIN;
+      nread = 0;
     }
     else {
       char buffer[STRERROR_LEN];
```

but 250 and friends are still failing here.

@jay : as you intended in your PR, the `pre_recved` handling needs to be at the lowest layer. In our new "connection filter" world, this should be the "socket" filter that sits at the end of the filter chain.

Once we have your change working, I'd like `Curl_read/write_plain()` to go through the socket filter installed and have the socket filter "own" the pre_recvd buffer. Then we can eliminate that one at the `conn` itself.

bagder:
Merged in 18383fbf72b31b9ae58f17df8e75158a818c31eb

jay:
There was another inconsistency that I did not catch. I've proposed #9949 with the original commit and a follow up commit to address the issue:

Though Curl_recv_plain could return 0 on CURLE_AGAIN as @icing suggests, IMO this is incorrect. Curl_send_plain already returns 0 on CURLE_AGAIN and IMO that is incorrect as well, because both should imitate a normal recv/send and return -1 on any error. However, callers of Curl_read_plain and Curl_write_plain expect that bytes read/written param receives 0 when CURLE_AGAIN is returned. The full details can be read in the follow-up commit which I left separate to make it easier to review.

PR #9945, opened 2022-11-18
H3 special setup of connection filters

icing:
Restore HTTP/3 to working condition after connection filter introduction.

HTTP/3 needs a special filter chain, since it does the TLS handling itself. This PR adds special setup handling in the HTTP protocol handler that takes are of it.

When a handler, in its setup method, installs filters, the default behaviour for managing the filter chain is overridden.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -247,6 +247,7 @@ static CURLcode h3_setup_conn(struct Curl_easy *data,
 
 #else /* ENABLE_QUIC */
   (void)conn;
+  (void)data;
   DEBUGF(infof(data, "QUIC is not supported in this build"));
   return CURLE_NOT_BUILT_IN;
 #endif /* !ENABLE_QUIC */
--- a/lib/http.c
+++ b/lib/http.c
@@ -246,6 +246,7 @@ static CURLcode h3_setup_conn(struct Curl_easy *data,
   return Curl_cfilter_socket_set(data, conn, FIRSTSOCKET);
 
 #else /* ENABLE_QUIC */
+  (void)conn;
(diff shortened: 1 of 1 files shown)


bagder:
The msh3 build:
~~~
In file included from curl_setup.h:322,
                 from http.c:25:
http.c: In function ‘h3_setup_conn’:
http.c:227:15: error: suggest parentheses around assignment used as truth value [-Werror=parentheses]
  227 |   DEBUGASSERT(conn->transport = TRNSPRT_QUIC);
      |               ^~~~
http.c:227:3: note: in expansion of macro ‘DEBUGASSERT’
  227 |   DEBUGASSERT(conn->transport = TRNSPRT_QUIC);
      |   ^~~~~~~~~~~

bagder:
Thanks!

lib/http.c
```diff
@@ -224,7 +224,7 @@ static CURLcode h3_setup_conn(struct Curl_easy *data,
 #ifdef ENABLE_QUIC
   /* We want HTTP/3 directly, setup the filter chain ourself,
    * overriding the default behaviour. */
-  DEBUGASSERT(conn->transport = TRNSPRT_QUIC);
+  DEBUGASSERT(conn->transport == TRNSPRT_QUIC);
```
catull, line 227: Perhaps this can be transposed to

```c
DEBUGASSERT(TRNSPRT_QUIC == conn->transport);
```
The rationale is, you cannot accidentally assign `conn->transport` to the constant `TRNSPRT_QUIC`.

icing: @catull yes, that is a better way of doing it.

PR #9947, opened 2022-11-18
hyper: classify headers as CONNECT and 1XX

seanmonstar:
I was debugging tests 1941 and 1942, and eventually noticed that `curl_easy_header` was fetching headers from CONNECT and 1XX responses as `CURLH_HEADER`, which was incorrect. I found that the issue was that the `writetype` was not being set, as it is in `http.c`.

I don't know if there's a better or more idiomatic way you prefer this written.


bagder (approved):
nice!

bagder:
Thanks!

PR #9948, opened 2022-11-18
ftp fix in regard to connection filters

icing:
Two fixes for ftp in regard to the new conneciton filters:

* "AUTH TLS" actually adds TLS to the primary connection
* SSL in PASV secondary connection works again.

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -2747,6 +2747,13 @@ static CURLcode ftp_statemachine(struct Curl_easy *data,
       if((ftpcode == 234) || (ftpcode == 334)) {
         /* this was BLOCKING, keep it so for now */
         bool done;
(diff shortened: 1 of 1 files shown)


icing:
Explanation: the PASV connection stalled, because the server was not starting the TLS handshake until the primary connection had send more commands. We tried to reach full connect status in do_more and that never happened unless do_more continue sending stuff on the first.

Afterthoughts: 

* If we really need to have reached the PASV socket on the server, this fix is not enough. If a proxy tunnel is in play, ftp do_more would continue once the proxy is connected.
* The `conn->secondaryhostname` was a surprise to me. Our TLS hostname verification uses always the name from the first socket. Seems to work for ftp...

bagder:
Thanks!

PR #9949, opened 2022-11-18
sendf: change Curl_read_plain to wrap Curl_recv_plain (take2)

jay:
_(This is the same commit from #9904, which landed earlier today but was reverted because SOCKS tests were hanging, and a follow-up commit to fix the reason for the revert. I have left the follow-up commit separate to make it easier to review but I plan to squash it into the parent.)_

---
_**(Original commit)**_

### sendf: change Curl_read_plain to wrap Curl_recv_plain

Prior to this change Curl_read_plain would attempt to read the socket directly. On Windows that's a problem because recv data may be cached by libcurl and that data is only drained using Curl_recv_plain. 

Rather than rewrite Curl_read_plain to handle cached recv data, I changed it to wrap Curl_recv_plain, in much the same way that Curl_write_plain already wraps Curl_send_plain. 

Curl_read_plain -> Curl_recv_plain
Curl_write_plain -> Curl_send_plain

This fixes a bug in the schannel backend where decryption of arbitrary TLS records fails because cached recv data is never drained. We send data (TLS records formed by Schannel) using Curl_write_plain, which calls Curl_send_plain, and that may do a recv-before-send ("pre-receive") to cache received data. The code calls Curl_read_plain to read data (TLS records from the server), which prior to this change did not call Curl_recv_plain and therefore cached recv data wasn't retrieved, resulting in malformed TLS records and decryption failure (SEC_E_DECRYPT_FAILURE). 

The bug has only been observed during Schannel TLS 1.3 handshakes. Refer to the issue and PR for more information. 

---

_**(Notes regarding original commit)**_

The pre-receive behavior for Windows was designed to call recv() before every send(), enough to fill a small buffer, because Winsock destroys unread received data in its own buffers when send() fails. Pre-receive is also designed to only be used with HTTP(S) and when the connection has "plain" sockets, eg TLS and HTTP/2 haven't yet been negotiated.

The bug has only been observed during Schannel TLS 1.3 handshakes where the server may send a record immediately after the server's handshake `Finished` record. In #9431 it is observed that the server sends `New Session Ticket` immediately after `Finished`. The client (us) reply to the server's `Finished` with our own `Finished` using Curl_write_plain, which could then pre-receive some part of the `New Session Ticket` received before our `Finished` was sent.

---

_**(Follow-up commit)**_

### squashme: fix Curl_{read,write}_plain for CURLE_AGAIN

- Change Curl_send_plain to return -1 on result CURLE_AGAIN. Curl_recv_plain already does this.

- Change Curl_read_plain (which wraps Curl_recv_plain) and Curl_write_plain (which wraps Curl_send_plain) to set the bytes read/written to 0 on any error.

This is a follow up to the parent commit which irons out inconsistencies in Curl_{read,recv,write,send}_plain. 

Callers of Curl_read_plain/Curl_write_plain expect 0 to be received as bytes read/written when any error is returned, unlike Curl_recv_plain/Curl_send_plain which are expected to return -1 for bytes read/written on any error, like normal recv/send behavior. 

Without this follow-up, on error:

Curl_recv_plain returns -1 and *result receives error.
Curl_send_plain returns -1 and *result receives error (!= CURLE_AGAIN).
Curl_send_plain returns 0 and *result receives error (== CURLE_AGAIN).
Curl_read_plain returns error and *n receives -1.
Curl_write_plain returns error and *written receives 0.

With this follow-up, on error

Curl_recv_plain returns -1 and *result receives error.
Curl_send_plain returns -1 and *result receives error.
Curl_read_plain returns error and *n receives 0.
Curl_write_plain returns error and *written receives 0.

### DIFF
--- a/lib/krb5.c
+++ b/lib/krb5.c
@@ -454,14 +454,14 @@ static int ftp_send_command(struct Curl_easy *data, const char *message, ...)
 /* Read |len| from the socket |fd| and store it in |to|. Return a CURLcode
    saying whether an error occurred or CURLE_OK if |len| was read. */
 static CURLcode
-socket_read(curl_socket_t fd, void *to, size_t len)
+socket_read(struct Curl_easy *data, curl_socket_t fd, void *to, size_t len)
 {
   char *to_p = to;
   CURLcode result;
   ssize_t nread = 0;
 
   while(len > 0) {
-    result = Curl_read_plain(fd, to_p, len, &nread);
+    result = Curl_read_plain(data, fd, to_p, len, &nread);
     if(!result) {
       len -= nread;
       to_p += nread;
@@ -502,15 +502,15 @@ socket_write(struct Curl_easy *data, curl_socket_t fd, const void *to,
   return CURLE_OK;
 }
 
-static CURLcode read_data(struct connectdata *conn,
-                          curl_socket_t fd,
+static CURLcode read_data(struct Curl_easy *data, curl_socket_t fd,
                           struct krb5buffer *buf)
 {
+  struct connectdata *conn = data->conn;
   int len;
   CURLcode result;
   int nread;
 
-  result = socket_read(fd, &len, sizeof(len));
+  result = socket_read(data, fd, &len, sizeof(len));
   if(result)
     return result;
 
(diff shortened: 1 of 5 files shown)


jay:
It looks like I will need to amend this commit, just based on what I've seen so far. Basically, if Curl_read_plain and Curl_send_plain return any CURLE error then the caller may expect read/written param to receive 0 instead of -1.

PR #9962, opened 2022-11-22
tls backends using connection filter IO

icing:
TLS backends using connection filters for IO, enabling HTTPS-proxy.

        - OpenSSL (and compatible)
        - BearSSL
        - gnutls
        - mbedtls
        - rustls
        - schannel
        - secure-transport
        - wolfSSL (v5.0.0 and newer)

        This leaves the only the following without HTTPS-proxy support:
        - gskit
        - nss
        - wolfSSL (versions earlier than v5.0.0)

### DIFF
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -156,6 +156,46 @@ static void mbed_debug(void *context, int level, const char *f_name,
 #else
 #endif
 
+static int bio_cf_write(void *bio, const unsigned char *buf, size_t blen)
+{
+  struct Curl_cfilter *cf = bio;
+  struct ssl_connect_data *connssl = cf->ctx;
+  struct Curl_easy *data = connssl->call_data;
+  ssize_t nwritten;
+  CURLcode result;
+
+  DEBUGASSERT(data);
+  nwritten = Curl_conn_cf_send(cf->next, data, (char *)buf, blen, &result);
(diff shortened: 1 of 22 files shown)


lib/vtls/mbedtls.c
```diff
@@ -924,13 +961,14 @@ static void mbedtls_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
   char buf[32];
-  (void) data;
 
   DEBUGASSERT(backend);
 
   /* Maybe the server has already sent a close notify alert.
      Read it to avoid an RST on the TCP connection. */
+  backend->io_data = data;
   (void)mbedtls_ssl_read(&backend->ssl, (unsigned char *)buf, sizeof(buf));
+  backend->io_data = NULL;
```
jay, line 971: Though I'm loathe to bring up conn->data (which was eliminated about 2 years ago) is there a better way then the association of data like this at a bunch of seemingly arbitrary points? Yes I'm sure they're not arbitrary but I'm struggling to follow how this works. Is there a more central location where data could be assigned, like vtls for example?

icing: I agree that this is looking for a more elegant solution. The sprinkling of those assignments in the code is not to my liking really, either.

bagder (approved):

bagder:
Thanks!

Issue #9963, opened 2022-11-22
--http2-prior-knowledge still offers http/1.1


kit-ty-kate:
### I did this

```
curl -sv https://exn.st --http2-prior-knowledge -o /dev/null
```

### I expected the following

```
[…]
* Connected to exn.st (XXX...) port 443 (#0)
* ALPN: offers h2
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
[…]
> GET / HTTP/2
[…]
```
but got
```
[…]
* Connected to exn.st (XXX...) port 443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
[…]
> GET / HTTP/1.1
[…]
```

I expected to not see `ALPN: offers http/1.1` given what the documentation says on the website (https://curl.se/docs/http2.html):
```
curl offers the --http2-prior-knowledge command line option to enable use of HTTP/2 without HTTP/1.1 Upgrade.
```

### curl/libcurl version

```
curl 7.86.0 (aarch64-unknown-linux-gnu) libcurl/7.86.0 OpenSSL/3.0.7 zlib/1.2.13 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.4 libpsl/0.21.1 (+libidn2/2.3.4) libssh2/1.10.0 nghttp2/1.51.0
Release-Date: 2022-10-26
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

```
Linux alarm 5.19.0-asahi-5-1-ARCH #1 SMP PREEMPT_DYNAMIC Sat, 20 Aug 2022 09:23:11 +0000 aarch64 GNU/Linux
```

jay:
[--http2-prior-knowledge](https://curl.se/docs/manpage.html#--http2-prior-knowledge) :

[...]

PR #9964, opened 2022-11-22
Do not offer http/1.1 when --http2-prior-knowledge is given

kit-ty-kate:
Fixes https://github.com/curl/curl/issues/9963

Temporary note: at the moment this PR only fixes this for the openssl backend. Since it would be a simple copy/past of the solution to other backend, i'd like to be sure that this solution would be accepted first before committing to changing the rest of the backends.

Also feel free to push into this branch if it's faster for you this way. I might not have time / forget to follow up on this PR.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -3507,10 +3507,12 @@ static CURLcode ossl_connect_step1(struct Curl_cfilter *cf,
     }
 #endif
 
-    protocols[cur++] = ALPN_HTTP_1_1_LENGTH;
-    memcpy(&protocols[cur], ALPN_HTTP_1_1, ALPN_HTTP_1_1_LENGTH);
-    cur += ALPN_HTTP_1_1_LENGTH;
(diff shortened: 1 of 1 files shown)


bagder:
I think this is a fine change as it provides a way for users to say h2-only which wasn't there before, but it also needs to be documented and we should work on getting it done for more backends while at it.

Also, this is probably to be viewed as a "change" and therefore I prefer we wait until the feature window is open until we merge. Which gives us more time to work on this for other TLS backends as well.

jay:
Sure, I agree. I can't think of what this would break.

bagder:
Any chance you can rebase this and fix the merge conflict?

bagder:
Lack of attention.

Sharathji1990:
I understand this is a old post, but I tried this commit, but after this change, curl only send HTTP1.1 in alpn and it doesn't add any H2 in alpn header.

bagder:
This PR was never merged since it had conflicts and was abandoned.

PR #9965, opened 2022-11-22
connect: remove include of <sys/filio.h> on NetWare

ryandesign:
The code that used to use `FIONBIO` from <sys/filio.h> was moved to nonblock.c in d709cb2a and bdbfe1f8.

At least it looks that way to me. I don't have a NetWare system to test on.

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -48,9 +48,6 @@
 #include <arpa/inet.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


danielgustafsson:
Netware isn't supported anymore as of 3b16575ae938dec2a29454631a12aa52b6ab9c67 so this seems like a dead leftover. I've opened #9966 for a removal of all Netware leftovers.

PR #9966, opened 2022-11-22
netware: remove leftover traces

danielgustafsson:
Commit 3b16575ae938dec2a29454631a12aa52b6ab9c67 removed support for building on Novell Netware, but a few leftover traces remained. This removes the last bits.


lib/nonblock.c
```diff
@@ -31,9 +31,6 @@
 #include <fcntl.h>
 #endif
 
-#if (defined(HAVE_IOCTL_FIONBIO) && defined(NETWARE))
```
ryandesign, line 34: Could you keep this `#include <sys/filio.h>` but change the condition to `#ifdef HAVE_SYS_FILIO_H`? I'm trying to revive support for classic Mac OS and it needs this.

danielgustafsson: Couldn't your patch just as well include it if it needs it? I understand that it'll be a merge conflict once this goes in (if it does) but it should be trivially fixed.

bagder (approved):
:rocket:

lib/nonblock.c
ryandesign: OK.

danielgustafsson:
> OK.
> 
Also, as a sidenote, I think it’s great to see a stab at reviving MacOS support! Keep up the good work!

PR #7280, opened 2021-06-19
curl_endian: remove the unused Curl_write64_le function

bagder:
The last usage was removed in cca455a36

PR #9968, opened 2022-11-22
curl_endian: remove Curl_write64_le from header

ryandesign:
The actual function was already removed in 4331c6dc.

See #7280


bagder (approved):

bagder:
Thanks!

Issue #9969, opened 2022-11-23
curl respects the locale's number format setting for parsing fractional seconds, it does not enforce dot


danielfaust:
### I did this

In `bash` and `zsh`: `curl --connect-timeout 3.14 https://example.com`

### I expected the following

The following printed out to stdout:
```
<!doctype html>
<html>
<head>
    <title>Example Domain</title>

...

</html>
```

### The actual result was

`curl: option --connect-timeout: expected a proper numerical parameter`

### curl/libcurl version

```
curl 7.81.0 (x86_64-pc-linux-gnu) libcurl/7.81.0 OpenSSL/3.0.2 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libidn2/2.3.2 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.43.0 librtmp/2.3 OpenLDAP/2.5.13
Release-Date: 2022-01-05
```
as well as
```
curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3
Release-Date: 2020-01-08
```

### operating system

Ubuntu 22.04 LTS
Linux hostname 5.15.0-53-generic #59-Ubuntu SMP Mon Oct 17 18:53:30 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**With `LC_NUMERIC="de_DE.UTF-8"`**

as well as 

Ubuntu 20.04 LTS
Linux hostname 5.15.0-52-generic #58~20.04.1-Ubuntu SMP Thu Oct 13 13:09:46 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**With `LC_NUMERIC="de_DE.UTF-8"`**

### workaround

Add `export LC_NUMERIC="en_US.UTF-8"` to environments requiring fractional seconds in the timeouts.

bagder:
Isn't your command line and claim contradict each other? German uses *comma* right and not a *period* ? (like Swedish btw)

curl uses the dot-version (as it ignoring the locale) fixed and on purpose so that command lines become transferable between systems. This is intentional. The documentation could probably make that clearer.

danielfaust:
[...]

PR #9972, opened 2022-11-23
curl: use dot as decimal separator independent of locale

bagder:
Independently of what the locale says

Reported-by: Daniel Faust
Ref: #9969


bagder:
Unfortunately, this change is also likely to break a few scripts out there... :cry:

MarcelRaad (approved):
> Unfortunately, this change is also likely to break a few scripts out there... 😢

At least they're broken consistently then. Or maybe wait 12 weeks for the last v7 release before breaking stuff? :-D

jay:
You are changing the locale on everything not just decimals.

bagder:
Yes, but is there anywhere we *want* locale dependent behavior?

jay:
System error messages. The default is c so you could delete the statement.

bagder:
The IDN tests seem to fail with this so I guess I have to do it differently.

PR #9973, opened 2022-11-23
make Curl_get_line not require a newline on the last line

robdewit:
Improve backwards compatibility by allowing a last line not ending with newline char.


robdewit:
Lines 51-58 could be left out if calling code is not relying on the newline as terminator.

lib/curl_get_line.c
```diff
@@ -43,12 +43,22 @@ char *Curl_get_line(char *buf, int len, FILE *input)
...
       size_t rlen = strlen(b);
-      if(rlen && (b[rlen-1] == '\n')) {
+      if(rlen && ((b[rlen-1] == '\n') || feof(input))) {
         if(partial) {
           partial = FALSE;
           continue;
         }
-        return b;
+        if(b[rlen-1] != '\n') {
+          /* make sure line is '\n' terminated if EOF */
```
bagder, line 52: This doesn't actually know that it is EOF, it could also be a line longer than 'len'. If you want it to act only on EOF, then the feof() check probably needs to set a separate flag for that purpose.

bagder:
I suppose the title means improved backwards compatibility for older .netrc handling? This function is used for more than that though.

lib/curl_get_line.c
robdewit: I admit it's a bit hacky, but from man fgets, I get that the line will never be more than len chars long.

```
       char *fgets(char *s, int size, FILE *stream);

       fgets()  reads  in at most one less than size characters from stream and stores them into the buffer pointed to by s.  Reading stops after an EOF or a
       newline.  If a newline is read, it is stored into the buffer.  A terminating null byte ('\0') is stored after the last character in the buffer.
```

robdewit:
I already thought it would be used in more places. For netrc only, the insertion of the missing newline would not be necessary. I guess it would be good practice for any file being read, to not throw away the last line if there is no newline. Unless, of course, the newline is part of the config specs.

lib/curl_get_line.c
bagder: > The man page explains "Reading stops after an EOF or a  newline."

If the line is `len` + newline or `len` + EOF we get the same `len` number of bytes without newline in the buffer. You then cannot know if the missing newline is because of too long buffer or if at EOF, just by checking for missing newline. If the line is *shorter* than `len` and is missing a newline, you know it is EOF.

robdewit: You're right, back to the drawing board.

robdewit: I now see it is impossible to know whether the last line is correct without parsing it when it completely fills the buffer, so I chose to opt for the most likely case: assume it is truncated.

As this makes it unreliable I could see why you would not want to allow a last non-newline line at all. 

Ignoring them does potentially break old config files though.

robdewit: @bagder Pushed a new commit, fixing the missing of truncated lines.

lib/curl_get_line.c
```diff
@@ -41,17 +41,39 @@ char *Curl_get_line(char *buf, int len, FILE *input)
...
-      /* read a partial, discard the next piece that ends with newline */
-      partial = TRUE;
+      else if(feof(input)) {
+        if(partial) {
+          partial = FALSE;
+          continue;
+        }
+        if(rlen < (size_t) len) {
+          /* b is EOF terminated, insert missing \n */
+          b[rlen] = '\n';
```
bagder, line 65: Shouldn't this also insert a null byte?

lib/curl_get_line.c
```diff
@@ -41,17 +41,39 @@ char *Curl_get_line(char *buf, int len, FILE *input)
...
+          return b;
+        }
+        else {
+          /* EOF terminated string using full buffer size
+           * This could either be truncated or correct.
+           * Most probably truncated.
+           */
+          partial = TRUE;
```
bagder, line 73: As feof() already returned true here, surely this cannot be a truncated (partial) line?

lib/curl_get_line.c
robdewit: yes.  I will fix it.

lib/curl_get_line.c
robdewit: Indeed, it needs rephrasing. As it cannot be converted to valid by adding a '\n' within the given buffer size, my choice would be to ignore it like a partial line. 

We need to make a choice 
* Changing the fgets to a maximum of (len-1) would make it possible to add the \n, but it would also possibly introduce other backward compatibility issues (assuming someone uses a 4K config line somewhere)
* Accepting the last line without \n because it is not truncated would maybe introduce compatibility issues in other routines that call this function

My choice would be to ignore it (like in the code, but rephrased). It would allow all non-newline last lines under 4K long and ignore that one 4K long last line without a newline. 

```
else
  /* Max buffer reached + EOF - this line is impossible to add a \n to so we'll ignore it */
  break;
```

Agreed?

bagder: I think that's totally fine to ignore in this case. It's just a special edge case anyway. Agreed!

bagder (approved):

lib/curl_get_line.c
```diff
@@ -41,17 +41,37 @@ char *Curl_get_line(char *buf, int len, FILE *input)
...
           continue;
         }
         return b;
       }
-      /* read a partial, discard the next piece that ends with newline */
-      partial = TRUE;
+      else if(feof(input)) {
+        if(partial) {
+          partial = FALSE;
+          continue;
```
bagder, line 61: shouldn't this just break? We're at end of file, no point in looping.

lib/curl_get_line.c
```diff
@@ -41,17 +41,37 @@ char *Curl_get_line(char *buf, int len, FILE *input)
...
+          b[rlen] = '\n';
+          b[rlen + 1] = '\0';
+          return b;
+        }
+        else
+          /* Maximum buffersize reached + EOF
+           * This line is impossible to add a \n to so we'll ignore it
+           */
+          break;
+      }
```
bagder, line 74: Isn't there a missing else and `partial = TRUE` somewhere? Nothing sets `partial` anymore!

bagder (changes requested):

lib/curl_get_line.c
robdewit: It clearly needs unit testing, so I added some of that. Not sure if I covered all edge cases, though.

bagder:
Test 8 fails on Windows, which I suspect is because it has a ^Z in there and that means EOF for text files on Windows somehow. We can probably just remove that. See patch below.

This probably becomes an issue in this PR because cookies are read with Curl_get_line().

~~~diff
diff --git a/tests/data/test8 b/tests/data/test8
index fa44a5a6a..3feff1926 100644
--- a/tests/data/test8
+++ b/tests/data/test8
@@ -67,11 +67,10 @@ Set-Cookie: cookie20=%hex[%14-junk]hex%
 Set-Cookie: cookie21=%hex[%15-junk]hex%
 Set-Cookie: cookie22=%hex[%16-junk]hex%
 Set-Cookie: cookie23=%hex[%17-junk]hex%
 Set-Cookie: cookie24=%hex[%18-junk]hex%
 Set-Cookie: cookie25=%hex[%19-junk]hex%
-Set-Cookie: cookie26=%hex[%1a-junk]hex%
 Set-Cookie: cookie27=%hex[%1b-junk]hex%
 Set-Cookie: cookie28=%hex[%1c-junk]hex%
 Set-Cookie: cookie29=%hex[%1d-junk]hex%
 Set-Cookie: cookie30=%hex[%1e-junk]hex%
 Set-Cookie: cookie31=%hex[%1f-junk]hex%

tests/unit/unit3200.c
```diff
@@ -0,0 +1,152 @@
...
+        break;
+      default:
+        abort_unless(1, "Unknown case");
+        break;
+    }
+    fclose(fp);
+  }
+  return 0;
+}
+UNITTEST_STOP
```
bagder, line 152: I love the unit test!

robdewit: I don't have any Windows environment at hand, but I'm confused. eof() should only trigger on the condition of having reached the end of input, not some control character. So my guess would be that the curl_get_line() function should just return the string including whatever control character is in there. It is then up to the parser that calls curl_get_line to decide what to do with that.

I added a line containing ^Z in the unit test just to check what Windows would make of it. 

Maybe the code needs something like `#ifndef WIN32` to not break compatibility in WIndows - which was the whole purpose of the PR anyway - maximum downwards compatibility.

robdewit: According to the OK result of the changed unit test on Windows, curl_get_line() seems to correctly return a string including the ^Z character. So I guess the fail in test 8 is caused by differences in parsing the cookie on Windows compared to Linux or MacOS.

What to do? Remove the offending line from test 8 won't fix the difference in test output.

robdewit: Hold on, the test was not quite conclusive, retrying, waiting for Windows output.

robdewit: I think the difference in processing is caused by cookie.c:1267 opening the file as FOPEN_READTEXT,  which makes Window interpret (some) control characters. The previous version of curl_get_line() would discard the line as a partial read because of the missing \n. The new version reads the data up until, but excluding the ^Z, so the cookie.c:452 invalid_octets() function does not throw the data away. What's left is a valid `cookie26=` cookie. 

Still no clue how to handle this, I guess there's good reasons to use the FOPEN_READTEXT.

bagder: As I recall things we went with text file simply because cookies are text. there were no considerations made for silly ^Z-handling. I just submitted #10017 to change this, let's see how that goes :smile:

bagder:
if you rebase this PR onto master now and force-push, I think the ^Z issue on Windows should be gone and I think we are ready to merge

bagder (approved):

bagder:
Thanks a lot for your hard work and patience with this!

fredericgermain:
I guess it more that you should not merge a commit and do a release on the same day? (linked to https://github.com/curl/curl/pull/9789 where you close comments).

bagder:
1. Why not?
2. How is that relevant to this bugfix?

bagder:
If you want to help out in the project, please do. We will appreciate it. Coming here months later just because Apple is slow to ship the latest curl version is not helpful.

fredericgermain:
I think integrating a commit and doing a release the same day with it is a very bad practice., but it's your project in the end.

This bugfix helped fix the problem in release 7.86.0, which you pushed in https://github.com/curl/curl/pull/9789, exactly the day of the tagging of 7.86.0. And you closed the comment about discussing the issue, so adding here.

I'm not an Apple fan boy, but I'm very skeptical about your comment "That sounds like an Apple problem". Most distribution makes months to integrate new releases, but you might have changed to have a faster integration when it's linked to security fixing, and it's not the case here.
I'd be actually interested to know the distribution you're using, we'll see if they are actually faster than Apple to ship it.

Please understand that curl is used by many people, and it might need nice to have a longer insight into the commits being integrated.

I would have helped with the project, but it's fixed by this commit, so I don't have anything to add so far.

Also, I'd like to add that closing a ticket for comments is quite childish.

bagder:
I closed that issue because your complaining did not add anything of value, and virtually nobody is going to see or follow a thread of comments added to a long-time closed issue. You mentioned that Apple shipped a curl with a bug we fixed in a later release.

Now I'm going to waste even more time and effort on this non-issue because you insist. Because I am childish I guess.

If you want to argue about or contribute to improve our development process then you are most welcome to do so, but doing this in an old closed issue or pull request is also a fairly bad way at doing it because it will not reach the audience that would be interested. You should rather take that in a *new* discussion here on GitHub or even better on the curl-library mailing list. I would then of course urge you to rather bring your proposals for what to improve rather than to just whine on something that failed previously.

It *is* now "an Apple problem" because Apple builds and ships curl bundled with macOS and they *could* opt to ship 7.87.0 already now. We fixed the problem already. Numerous distributions and operating systems already provide the latest curl release to their users. I personally use Debian unstable, it usually offers the latest curl release within days of our releases.

To the issue of merging a bugfix just days before a release:

- this particular bugfix was a security fix, which we always merge late in the cycle on purpose because it will reveal to observers that there was a security flaw that can be exploited.  Still, it was offered as a pull request, it ran through all the tests fine and not a single person objected to it or remarked on its functionality - until several weeks later. Thus, the closeness in time to release was actually irrelevant.
- a fixed bug is something we want to include in the next release and most users would *expect* us to ship bugfixes that we have already working, in the pending release. It's a balance of course, but just giving something more time is also not a guarantee for us finding problems with code.
- I think our development process should rely on procedures and established ways that do not make us have to just wait for some period of time between a fix and a release. Code style, reviews, tests, verification, code scans, tortures should be enough. In this particular case, what failed us was that we did not have a test case using this edge case.

PR #9974, opened 2022-11-23
vtls: repair build with disabled proxy

bagder:


icing:
So, this removes the PROXY ifdefs in openssl.c, because the called functions handle that case already?

bagder:
> So, this removes the PROXY ifdefs in openssl.c, because the called functions handle that case already?

This *adds* the ifdef in openssl.c because those variables are unused when there is no proxy support built-in.

icing:
> > So, this removes the PROXY ifdefs in openssl.c, because the called functions handle that case already?
> 
> This _adds_ the ifdef in openssl.c because those variables are unused when there is no proxy support built-in.

Clearly, need more coffee. 

Yes, this looks necessary since structures in `connectdata` will not exist in such a build.

bagder:
Thanks!

PR #9976, opened 2022-11-24
CI: freebsd build heimdal from source

pheiduck:
freebsd have dropped heimdal?

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -43,7 +43,7 @@ freebsd_task:
   pkginstall_script:
     - pkg update -f
     - pkg install -y autoconf automake libtool pkgconf brotli openldap24-client libpsl libssh2 openssh-portable libidn2 librtmp libnghttp2 nghttp2 stunnel unzip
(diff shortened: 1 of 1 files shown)


danielgustafsson:
I'm not sure why pkg install fails, but don't think heimdal has been dropped: https://www.freshports.org/security/heimdal/

`pkg install heimdal` works fine on my FreeBSD 14 box, but I don't have a 13 installation to try on. The db5 package which it depends on is however deprecated (yet still available), maybe that holds a clue?

pheiduck:
it is no longer found on 12 and 13:
```
pkg: No packages available to install matching 'heimdal' have been found in the repositories
```

danielgustafsson:
That could just as well be an issue with the mirror though? As it's still in the ports tree, listed as a package and installable on FreeBSD 14 it seems there is more to this. If we do anything, shouldn't we the use the [ports tree entry for heimdal](https://cgit.freebsd.org/ports/tree/security/heimdal) to ensure we are building a version representative of what users might have?

pheiduck:
okay, then we'll wait and see how it develops.
I split the PR so we can merge the pip install fix

pheiduck:
Will be closed its solved on 13.1 maybe 12.4 later as well

PR #9977, opened 2022-11-24
CI: fixup: pip install --use-pep517 ...

pheiduck:
e.g. "DEPRECATION: pycryptodomex is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559"


pheiduck:
isn't needed

Issue #9982, opened 2022-11-25
Crash creating quic connection (nghttp3/ngtcp2 backend)


Fullmetal5:
### I did this
Building using the instructions at https://curl.se/docs/http3.html (plus --enable-debug) for ngtcp2 then attempting to make a http3/quic connection crashes.
This didn't use to crash just a few commits ago but I haven't bisected to find the offending commit.
EDIT: Commit af22c2a546ab862ab577c8d9d3609af0de178974 appears to have introduced the problem.
Built on current master (280cbeee2789749fa9b2267ce7a07813645897de)

```
./curl -v -4 --http3 https://example.com/
* STATE: INIT => CONNECT handle 0x55d6e888f038; line 1909 (connection #-5000)
* HTTP/3 direct conn setup(conn #-1, index=0)
Segmentation fault (core dumped)
```

Backtrace from gdb:
```
(gdb) r -v -4 --http3 https://example.com/
Starting program: /test/pre4/bin/curl -v -4 --http3 https://example.com/
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
* STATE: INIT => CONNECT handle 0x5555555e8038; line 1909 (connection #-5000)
* HTTP/3 direct conn setup(conn #-1, index=0)

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7ec503b in Curl_conn_socket_set (data=0x5555555e8038, sockindex=0) at connect.c:1838
1838      DEBUGASSERT(!data->conn->cfilter[sockindex]);
(gdb) bt
#0  0x00007ffff7ec503b in Curl_conn_socket_set (data=0x5555555e8038, sockindex=0) at connect.c:1838
#1  0x00007ffff7eec98c in h3_setup_conn (data=0x5555555e8038, conn=0x5555555e6d28) at http.c:246
#2  0x00007ffff7eeca6b in http_setup_conn (data=0x5555555e8038, conn=0x5555555e6d28) at http.c:276
#3  0x00007ffff7f48df4 in setup_connection_internals (data=0x5555555e8038, conn=0x5555555e6d28) at url.c:2177
#4  0x00007ffff7f4c4b0 in create_conn (data=0x5555555e8038, in_connect=0x7fffffffdee0, async=0x7fffffffdf28) at url.c:3711
[...]

PR #9983, opened 2022-11-26
conn setup phase clarifications

icing:
In regard to #9982, re-add `struct connectdata *conn` as parameter to cfilter setup methods.

- `Curl_ssl_get_config()` now returns the first config if no SSL proxy filter is active
- socket filter starts connection only on first invocation of its connect method


bagder:
Thanks!

PR #9984, opened 2022-11-27
build: restore old msdos, amigaos makefiles

vszakats:
- Restore original legacy Makefiles for MS-DOS and AmigaOS. Some envs do not recognize double-quoted command-line arguments.
- Status of MS-DOS/AmigaOS platform support inside `Makefile.mk` changes to experimental.

Regression from a8861b6ccdd7ca35b6115588a578e36d765c9e38 (Partial revert)

Reported-by: Gisle Vanem
Fixes https://github.com/curl/curl/pull/9764#issuecomment-1328216388
Closes #xxxx

---

The extra clutter in `Makefile.mk` that belong to these two platorms is quite tiny: 23 lines. Extra 30 lines for `WIN32`-specific stuff, which can be removed if we support Windows-only here.

[ Not saying that these couldn't be made work with `Makefile.mk`, but I don't have the capacity to test/support the needs of these platforms. Also, supporting them seems to be difficult without hurting legibility of mingw32-specific logic. ]

### DIFF
--- a/.github/scripts/spellcheck.words
+++ b/.github/scripts/spellcheck.words
@@ -161,6 +161,7 @@ distro
 distro's
 distros
 DJGPP
+djgpp
 dlist
 DLL
 dll
--- a/.github/scripts/spellcheck.words
+++ b/.github/scripts/spellcheck.words
@@ -160,8 +160,6 @@ dir
 distro
 distro's
 distros
-DJGPP
-djgpp
 dlist
 DLL
 dll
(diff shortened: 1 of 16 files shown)


vszakats:
Pending discussions:
- Why double quotes might not be working on Windows?
- Dropping support for spaces in `*_PATH` variables?
- Use `"-L<path>"` style quoting? (vs. `-L"<path>"`)

vszakats:
Closing with these conclusions/notes: https://github.com/curl/curl/pull/9764#issuecomment-1330650743

PR #9986, opened 2022-11-27
build: assume errno.h is always available

jay:
- Remove errno.h detection from all build configurations.

errno.h is a standard header according to C89.

Closes #xxxx

---

Note HAVE_ERRNO_H was undefined in config-win32 when BSD-style lwIP TCP/IP stack was used. I'm not sure what that's about.
Note HAVE_ERRNO_H was not defined by config-win32ce so possibly some build system that builds for Windows CE is missing it.

/cc @gvanem


bagder (approved):

vszakats:
Yep, `errno.h` is not present on WinCE MSVC + ARM. Likely less of an issue these days.

PR #7600, opened 2021-08-20
aws_sigv4: URI encoding

outscale-mgo:
It seems AWS signature V4 was lacking part of the algorith, AWSv4 should in theory URL Encode, the data before using it in the algorytme.

problem is that depending of whatever V4 signature is use for s3 or another API, the URL encoding is not the same.
https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html
https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

In one case the data is single encoded, and in the other case, it is double encode.

So the patch add a new option to curl, that enable the user to choose the number of encoding you want. (default is 0)

Here is the changes of this PR:
* add `--aws-sigv4-nb-encoding` option, default to 0
* add some static function in https_aws_sigv4 file .
* uri_encore `data->state.up.query`  and `data->state.up.path `, if the option is set.

some notes:

0. I didn't test, nor implement double encoding yet.
1. I couldn't use strlen_url in urlapi.c as the algorytm use for the encoding is slightly different
2. I'm not sure hostname trim is useful, maybe curl already do it when storing it in `hostname` ? 
3. AWS documentation say nothing explicitly about `=` encoding, but say that each value and key of the query need to be encoded, while the `=` use to slipt the key from the value need to be left as it. as `data->state.up.query` contain all queries parameters as a single string, I need to determine which '=' are part of a value and which one a "true" equal

should fix: #7559 #9717

PR #9995, opened 2022-11-28
aws_sigv4: fall back to UNSIGNED-PAYLOAD for sign_as_s3

cbodley:
all s3 requests default to `UNSIGNED-PAYLOAD` and add the required `x-amz-content-sha256` header. this allows `CURLAUTH_AWS_SIGV4` to correctly sign s3 requests to amazon with no additional configuration

adds a `bool sign_as_s3` flag that can be extended with detection/configuration for other s3-compatible services (out of scope)


cbodley:
cc @outscale-mgo

lib/http_aws_sigv4.c
```diff
@@ -408,6 +411,54 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
...
+
+  if(payload_hash) {
+    /* Use payload hash from existing content-sha256 header */
+  }
+  else if(sign_as_s3) {
+    /* Use UNSIGNED-PAYLOAD for the payload hash and content-sha256 header */
+    char *header = NULL;
+    struct curl_slist *head = NULL;
+
+    payload_hash_len = sizeof(S3_UNSIGNED_PAYLOAD) - 1;
```
bagder, line 429: Can we add a DEBUGASSERT() here that verifies that `hash_len` is shorter than `sizeof(sha_hex)` ?

bagder:
@outscale-mgo can we get your comments on this PR?

lib/http_aws_sigv4.c
```diff
@@ -310,6 +312,7 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
   char provider1[MAX_SIGV4_LEN + 1]="";
   char region[MAX_SIGV4_LEN + 1]="";
   char service[MAX_SIGV4_LEN + 1]="";
+  bool sign_as_s3 = false;
```
outscale-mgo, line 315: `s/sign_as_s3/default_unsigned_payload/g`
because sign_as_s3 seems to imply that s3 require unsigned_payload, so `default_unsigned_payload` may be more explicit. (plus if another API would require unsigned_payload by default the code will be  more straightforward to read if the boolean doesn't imply it's use only for 3s)

outscale-mgo (commented):
I like how the bolean is use overall.
as we talk here https://github.com/curl/curl/pull/8935, what I'm not a big fan of is this condition
```
  sign_as_s3 = (strcasecompare(provider0, "aws") &&
                strcasecompare(service, "s3"));
```
first because it bind some code to s3, but TBH if we see this code as a first step for allowing other API to use `sign_as_s3` it's not a big deal either.

what I find a little more annoying, is that s3 might require the hash calculated [here](https://github.com/curl/curl/blob/56390897d84c2fc0c9638af3be297093847e68d8/lib/http_aws_sigv4.c#L447) to be set as `x-%s-content-sha256`. and by using `aws:s3` as the condition to set  `UNSIGNED_PAYLOAD`, we remove the future possibility to use curl to calculate the payload for `x-%s-content-sha256`. 
UNSIGNED_PAYLOAD is the 'best' default value for `s3` because it's the one that will work most of the time.
but, currently if a user want to have an hash in `x-%s-content-sha256`, they need to compute it, it's not ideal, especially as the hash they need to compute is already computed in curl code. so ideally we need to find a way for s3 user be able to tell curl to set [this](https://github.com/curl/curl/blob/56390897d84c2fc0c9638af3be297093847e68d8/lib/http_aws_sigv4.c#L447) as `x-%s-content-sha256`, because even if it's not the header that will always work, it's the one the user might need the most (as it's the one with the most complex algorithm, so the one you want to avoid the most). 
it's not exclusive to keep `aws:s3` as condition for defaulting to UNSIGNED_PAYLOAD and then add another condition to use the calculated hash for the header, even if `aws:s3` is set, but it might not be very intuitive either.

cbodley:
@outscale-mgo thanks for the feedback!

i went with UNSIGNED-PAYLOAD for all s3 requests because i view the existing payload hashing as unreliable. it only works for certain requests, and it's been hard for me to specify exactly which those are. so i worried that getting this detection logic wrong would lead to signing bugs, where UNSIGNED-PAYLOAD would always work

but maybe it's not so hard to specify which requests can't be signed? the only non-empty payload we currently support is from [CURLOPT_POSTFIELDS](https://curl.se/libcurl/c/CURLOPT_POSTFIELDS.html). @bagder are [CURLOPT_READFUNCTION](https://curl.se/libcurl/c/CURLOPT_READFUNCTION.html) and [CURLOPT_READDATA](https://curl.se/libcurl/c/CURLOPT_READDATA.html) the only other ways to provide a request body? if so, then we would only need to use UNSIGNED-PAYLOAD for those 2 cases

cbodley:
happy new year, @outscale-mgo @bagder. a friendly reminder about this one

bagder:
`CURLOPT_MIMEPOST` is also a popular way to send POST data.

cbodley:
> `CURLOPT_MIMEPOST` is also a popular way to send POST data.

ok, thanks. it looks like the existing code will generate the wrong signature (using the checksum of an empty buffer) for these requests. if that mimepost data is all in memory, we *should* be able to calculate a real checksum

for now, i pushed a rebase/update that uses s3's UNSIGNED-PAYLOAD for all 3 cases (CURLOPT_MIMEPOST, CURLOPT_READDATA, CURLOPT_READFUNCTION), and otherwise falls back to the existing code

i still need to update the doc and add test cases with/without each of those CURLOPTs

lib/http_aws_sigv4.c
```diff
@@ -300,6 +300,71 @@ static char *parse_content_sha_hdr(struct Curl_easy *data,
...
+                                     char *sha_hex, size_t sha_hex_size)
+{
+  char *header = NULL;
+  struct curl_slist *head = NULL;
+  CURLcode ret = CURLE_OUT_OF_MEMORY;
+
+  /* use UNSIGNED-PAYLOAD when we can't reliably calculate a checksum */
+  if(data->set.method == HTTPREQ_POST_MIME || /* from CURLOPT_MIMEPOST */
+     data->set.in_set || /* from CURLOPT_READDATA or CURLOPT_READFUNCTION */
+     data->set.fread_func_set) { /* from CURLOPT_READFUNCTION */
```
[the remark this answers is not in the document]
cbodley: @bagder do you have any suggestions on how best to detect whether we expect a request body, outside of CURLOPT_POSTFIELDS/CURLOPT_MIMEPOST? do you think `data->set.method` alone could do this reliably?

if this gets too complicated, i'd be tempted to switch back to UNSIGNED-PAYLOAD by default for s3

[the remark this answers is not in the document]
bagder: data->set.method should be reliable, but of course a body can also be zero-bytes long

cbodley:
rebased and updated to make use only the request method and infilesize

test case 1960 checks that UPLOAD with INFILESIZE=0 calculates the checksum of an empty buffer

other test cases planned:
* non-empty UPLOAD -> UNSIGNED-PAYLOAD
* MIMEPOST -> UNSIGNED-PAYLOAD
* POSTFIELDS -> real checksum
* GET -> empty checksum

lib/http_aws_sigv4.c
```diff
@@ -300,6 +300,76 @@ static char *parse_content_sha_hdr(struct Curl_easy *data,
...
+  /* add the required content-sha256 header */
+  header = curl_maprintf("x-%s-content-sha256: %s", provider1, sha_hex);
+  if(!header)
+    goto fail;
+  head = Curl_slist_append_nodup(data->set.headers, header);
+  if(!head) {
+    free(header);
+    goto fail;
+  }
+  data->set.headers = head;
```
cbodley, line 367: the final test case 1964 doesn't add any headers to the request. in that case only, valgrind complains that this memory is leaked. is there a safe way to do this here?

below, `Curl_output_aws_sigv4()` adds a `x-amz-date` header at the end of `auth_headers`:
```c
  auth_headers = curl_maprintf("Authorization: %s4-HMAC-SHA256 "
                               "Credential=%s/%s, "
                               "SignedHeaders=%s, "
                               "Signature=%s\r\n"
                               "%s\r\n",
                               provider0,
                               user,
                               credential_scope,
                               sha_hex,
                               date_header);
```
is that the correct way to add `x-amz-content-sha256`?

cbodley: i added a commit that does this. the leak is gone and the tests still pass

cbodley:
@outscale-mgo would you please do another round of review?

this successfully interops with ceph's s3 server in its current form

> s/sign_as_s3/default_unsigned_payload/g

i chose to stick with the former, since this flag controls both the use of UNSIGNED-PAYLOAD and the addition of a content-sha256 header. i assume that other s3-like services will want both

> by using aws:s3 as the condition to set UNSIGNED_PAYLOAD, we remove the future possibility to use curl to calculate the payload for x-%s-content-sha256

it now calculates real checksums where it can. are you happy with the way that `calc_s3_payload_hash()` and `calc_payload_hash()` are composed here?

outscale-mgo:
> 

on it, I try to give you feedback before today, or tomorrow :)

outscale-mgo (commented):
overall I'm pretty happy with this PR.
I should be able to reuse `sign_as_s3` variable to handle URI encoding here:
https://github.com/curl/curl/pull/7600.
something that I might have to rethink in my PR , is if I should let the user to have a possibility to speficy the number of encoding perself, or just for the URI encoding.

lib/http_aws_sigv4.c
```diff
@@ -300,6 +308,69 @@ static char *parse_content_sha_hdr(struct Curl_easy *data,
...
+    goto fail;
+
+  sha256_to_hex(sha_hex, sha_hash, sha_hex_size);
+fail:
+  return ret;
+}
+
+#define S3_UNSIGNED_PAYLOAD "UNSIGNED-PAYLOAD"
+
+static CURLcode calc_s3_payload_hash(struct Curl_easy *data,
```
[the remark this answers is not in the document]
cbodley: > I guess the +2 if for '\n\r'

that's right, but i had also forgotten to add another 2 for the ': ' between the header name/value

cbodley:
@outscale-mgo thanks for the great feedback. i believe i've addressed all of your comments

cbodley:
@bagder any more feedback on your end?

cbodley:
oops, i'm looking through the failed checks, and some of them are due to new test cases. i'll look into those

tests/data/test1961
```diff
@@ -0,0 +1,77 @@
...
+PUT /aws_sigv4/testapi/test HTTP/1.1
+Host: exam.ple.com:9000
+Authorization: AWS4-HMAC-SHA256 Credential=xxx/19700101/us-east-1/s3/aws4_request, SignedHeaders=content-type;host;x-amz-content-sha256;x-amz-date, Signature=4a6e9b4af0542ffb83744c6852f8e1bfec14f2a67e6f6f037b39f172f79d62af
+X-Amz-Date: 19700101T000000Z
+x-amz-content-sha256: UNSIGNED-PAYLOAD
+Transfer-Encoding: chunked
+Expect: 100-continue
+
+0
+
```
cbodley, line 74: the failing tests 1961 and 1965 are both missing the final two lines of this 'terminating chunk'. the tests succeed in my local build configured with `--with-openssl --enable-debug`

both libtests rely on a `CURLOPT_READFUNCTION` that does nothing but `return 0`. should they be doing something different?

cbodley: i'm guessing that these test failures didn't see those expected lines of response body because the server `<reply>` didn't include the expected `100 Continue`? i added that to test1961 and test1965 to see if it helps

cbodley: @bagder the combination of a redirect and 100-continue seemed to be the issue there. does that final SQUASHME commit look sensible? if so i'll squash

otherwise i *think* the other failed checks are unrelated to this PR

cbodley:
a conflicting test1960 was merged, so i rebased and bumped the new test cases from the range 1960-1965 to 1970-1975

cbodley:
@bagder a friendly reminder; i believe this is ready

bagder:
Thanks!

PR #10003, opened 2022-11-29
ci: Update the fuzzing job so that the correct dirs are provided to CIFuzz

cmeister2:
I don't think the CIFuzz job is currently doing the right thing because of how the Docker image gets created in oss-fuzz. I've tested this in curl-fuzzer and it appears to pick up the right commits (i.e. I put a script change in with the CI change and that was reflected in the job)

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -21,11 +21,32 @@ jobs:
   fuzzing:
     runs-on: ubuntu-latest
     steps:
+    # These two checkout steps check out:
+    # 
+    # - `curl` at the current ref
+    # - `curl_fuzzer` at HEAD of the default branch
+    #
+    # under the `fuzz` directory.
+    - name: Checkout curl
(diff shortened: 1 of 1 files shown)


cmeister2:
Closing this. It seems by default we don't need to do this because OSS-Fuzz runs this step by default:

```
Running: docker run --rm --privileged --shm-size=2g --platform linux/amd64 -e FUZZING_ENGINE=libfuzzer -e CIFUZZ=True -e SANITIZER=address -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c++ -e OUT=/github/workspace/build-out --volumes-from 26746b5491a9 gcr.io/oss-fuzz/curl /bin/bash -c 'cd / && rm -rf /src/curl/* && cp -r /github/workspace/storage/curl /src && cd - && compile'.
```
which overrides the checked-out /src/curl directory with whatever is in the branch.

PR #10006, opened 2022-11-30
curl.h: include <sys/select.h> on SerenityOS

kennethmyhra:
This is a patch that's needed for the curl port on SerenityOS we would like to upstream.


bagder (approved):

bagder:
Thanks!

PR #10015, opened 2022-12-01
Fix for BoringSSL BIO result interpretation mixups.

icing:
Boring seem to interpret a simple `retry` indicator as a SYSCALL error instead of gently returning WANT_READ, as OpenSSL does.

This PR adds a workaround that preserves the actual CURLcode of the recent IO operation and checks to to catch such conditions.


bagder:
Thanks!

PR #10017, opened 2022-12-02
cookie: open cookie jar as a binary file

bagder:
On Windows there is a difference and for text files, ^Z means end of file which is not desirable.

Ref: #9973


jay:
This change means that CR \r won't be stripped from the end of lines.

bagder:
> This change means that CR \r won't be stripped from the end of lines.

They are already stripped explicitly by cookie.c code though, so it might just work...

Issue #10013, opened 2022-12-01
curl compiled with BoringSSL/quiche for H3 throws "Resource temporarily unavailable" for non-H3 requests


rmarx:
### I did this

I built curl with HTTP/3 support following [the guide](https://github.com/curl/curl/blob/master/docs/HTTP3.md#quiche-version). 
This via a dockerfile at https://github.com/rmarx/curl-http3.

Loading urls over H3 with the `--http3` flag works fine, it's when I **omit** the `--http3` flag (hoping to use HTTP/1.1 or HTTP/2 as "normal" and for example test alt-svc) that I get the "Resource temporarily unavailable" error. 

For example:
```
% docker run -it --rm rmarx/curl-http3 curl -IL https://www.cloudflare.com --verbose
*   Trying 104.16.123.96:443...
* Connected to www.cloudflare.com (104.16.123.96) port 443 (#0)
* ALPN: offers http/1.1
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
* [CONN-0-0][CF-SSL] TLSv1.2 (OUT), TLS handshake, Client hello (1):
* BoringSSL SSL_connect: Resource temporarily unavailable in connection to www.cloudflare.com:443 
* Closing connection 0
curl: (35) BoringSSL SSL_connect: Resource temporarily unavailable in connection to www.cloudflare.com:443
```

Shelling into the docker container and running curl inside gives the same error, so it's not a "docker interface" issue. 

I see that there are [several similar issues](https://github.com/curl/curl/issues?q=is%3Aissue+%22Resource+temporarily+unavailable%22+is%3Aclosed+) reported for other TLS libs, e.g., #6896 and #6864, though they seem to fail when loading HTTP/3, not H1 or H2. Additionally, the fixes there don't immediately seem applicable here, so I'm not sure what to do.  

### I expected the following

That the curl build with BoringSSL+quiche also loads URLs over HTTP/1.1 and HTTP/2, not just HTTP/3. 

### curl/libcurl version

curl 7.87.0-DEV (aarch64-unknown-linux-gnu) libcurl/7.87.0-DEV BoringSSL quiche/0.16.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
[...]

PR #10021, opened 2022-12-02
openssl: return -1 on errors

bagder:
BIO_read and BIO_write return negative numbers on error, including retryable ones. This seems to have been a regression from 55807e6. Both branches should be returning -1.

The APIs are patterned after POSIX read and write which, similarly, return -1 on errors, not zero, with EAGAIN treated as an error.

Reverts commit 5c27f6c452f02507b3bb9506e1079ea2f3943e4e and solves it differently.

Bug: https://github.com/curl/curl/issues/10013#issuecomment-1335308146
Reported-by: David Benjamin @davidben 

/cc @icing

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -713,13 +713,8 @@ static int bio_cf_out_write(BIO *bio, const char *buf, int blen)
   BIO_clear_retry_flags(bio);
   connssl->backend->io_result = result;
   if(nwritten < 0) {
-    if(CURLE_AGAIN == result) {
+    if(CURLE_AGAIN == result)
       BIO_set_retry_write(bio);
-      nwritten = 0;
-    }
-    else {
-      nwritten = -1;
-    }
   }
   return (int)nwritten;
 }
@@ -743,13 +738,8 @@ static int bio_cf_in_read(BIO *bio, char *buf, int blen)
   BIO_clear_retry_flags(bio);
   connssl->backend->io_result = result;
   if(nread < 0) {
-    if(CURLE_AGAIN == result) {
+    if(CURLE_AGAIN == result)
       BIO_set_retry_read(bio);
-      nread = 0;
-    }
-    else {
-      nread = -1;
-    }
   }
   return (int)nread;
 }


bagder:
@icing did I understand you right that you rather keep some of the old logic?

icing:
> @icing did I understand you right that you rather keep some of the old logic?

I'd feel more comfortable that way. I do simply not trust that all OpenSSL-like implementations expose the same quirks in all versions, as the error handling in the API is not well defined.

If you think, the -1 way should work everywhere and we rather pound on non-confirming implementations, then I won't object. ;)

bagder:
I think your assessment of the situation here has merit. But should we perhaps at least still make it always return -1 for BIO in those functions? Like this:
~~~diff
index dd7201ac7..e7a1caabf 100644
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -711,17 +711,12 @@ static int bio_cf_out_write(BIO *bio, const char *buf, int blen)
   /* DEBUGF(infof(data, CFMSG(cf, "bio_cf_out_write(len=%d) -> %d, err=%d"),
          blen, (int)nwritten, result)); */
   BIO_clear_retry_flags(bio);
   connssl->backend->io_result = result;
   if(nwritten < 0) {
-    if(CURLE_AGAIN == result) {
+    if(CURLE_AGAIN == result)
       BIO_set_retry_write(bio);
-      nwritten = 0;
-    }
-    else {
-      nwritten = -1;
-    }
   }
   return (int)nwritten;
 }
 
 static int bio_cf_in_read(BIO *bio, char *buf, int blen)
@@ -741,17 +736,12 @@ static int bio_cf_in_read(BIO *bio, char *buf, int blen)
   /* DEBUGF(infof(data, CFMSG(cf, "bio_cf_in_read(len=%d) -> %d, err=%d"),
          blen, (int)nread, result)); */
   BIO_clear_retry_flags(bio);
   connssl->backend->io_result = result;
   if(nread < 0) {
-    if(CURLE_AGAIN == result) {
+    if(CURLE_AGAIN == result)
       BIO_set_retry_read(bio);
-      nread = 0;
-    }
-    else {
-      nread = -1;
-    }
   }
   return (int)nread;
 }
 
 static BIO_METHOD *bio_cf_method = NULL;

jay (approved):
@icing if wolfSSL's OpenSSL emulation can't handle it properly then wouldn't that be a bug in wolfSSL?

Ref: https://github.com/curl/curl/blob/f8ad07c031adec4a76fbb60bd1e8d413657753f6/lib/vtls/wolfssl.c#L311-L355

icing:
@bagder: looks good to me

@jay: yes, but let's first find a common way for all the other *SSL thingies. Then we verify the wolfSSL behaviour on this.

bagder:
I simplified this PR and only fixed the -1 return codes now, as discussed above.

PR #10031, opened 2022-12-04
Makefile.mk: improve a GNU Make hack [ci skip]

vszakats:
Replace the hack of using `$() ` to represent a single space. The new method avoids the `--warn-undefined-variables` debug warning.

Closes #xxxxx

### DIFF
--- a/lib/Makefile.mk
+++ b/lib/Makefile.mk
@@ -324,7 +324,7 @@ ifdef WIN32
   _LIBS += -lws2_32 -lcrypt32 -lbcrypt
 endif
 
-ifneq ($(findstring 11,$(subst $() ,,$(SSLLIBS))),)
+ifneq ($(findstring 11,$(subst $(subst ,, ),,$(SSLLIBS))),)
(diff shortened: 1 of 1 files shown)


lib/Makefile.mk
```diff
@@ -324,7 +324,7 @@ ifdef WIN32
   _LIBS += -lws2_32 -lcrypt32 -lbcrypt
 endif
 
-ifneq ($(findstring 11,$(subst $() ,,$(SSLLIBS))),)
+ifneq ($(findstring 11,$(subst $(subst ,, ),,$(SSLLIBS))),)
```
jay, line 327: Would something like $(space) work

vszakats: I'm not aware of such built-in function. We can initialize such variable using the same hack, but I'd only do it once we have another place which needs a space at a problematic location.

jay: Ok. It would be easier to understand IMO but that's not critical.

jay (approved):

PR #10032, opened 2022-12-05
base64 decode faster

bagder:
- by using a lookup table instead of strchr()
 - by doing full quantums first, then padding

# Test

The test decodes "short" strings, up to 2948 bytes in length and it loops a thousand times.

[test code](https://gist.github.com/bagder/959b087c933465317407b20bc0fb0b53)

## Numbers

Best out of three runs on my local machine.

Old code: 1m47.593s
New code: 24.098s

Speedup: 4.46 times.


maage:
I see tiny speedup on your test program, like 10s -> 9s with this patch.
https://github.com/maage/curl/commit/492d018641f8824308f15dad93a08442bb6cfb13

bagder:
Yes, I think that is pretty obvious if we can accept the added memory 
requirement. I went with my solution to keep it smaller.

bagder:
maybe I should allow the user to select which way to go with an ifdef...

lucasexe:
Thanks @bagder for this great improvement.
I would like to discuss again about the @maage's sugestion.

In a quick analysis, your original solution uses less memory but there are some caveats:
 - In the original solution we have: 80 bytes in the data/rodata segment + 256 bytes in stack  + operations to copy the data to stack).
 - In the second version we have only 256 bytes in the rodata segment.
 
The original solution has a memory peak greater than the second one during the function execution.
 
Please also note that "operations to copy the data to stack" will also require some space in the code segment. Any compiler with a decent optimizer will replace both 2 memcpy to several other instructions. (Please check this example: https://godbolt.org/z/o8995W7qv):

GCC (with -O2) optimizes the memcpys to other instructions that requires 112 bytes in code segment. 
Clang (with -O2) optimizes the memcpys to other instructions that requires 192 bytes in code segment.

So, compiling with GCC, the original solution spends 80 + 112 = 192 bytes + 256 (if the function is called). Compiling with Clang, spends 80 + 192 = 272 bytes (more than the second solution) + 256 bytes in the stack.

With all those numbers, I would strongly recommend using the second version.
What do you think?

bagder:
> What do you think?

I've moved on. I don't believe in adding ifdefs for this because nobody will care enough to alter those. If someone feels strongly for further changes, propose those in a PR and explain why we want that.

PR #10033, opened 2022-12-05
The BIO-1 way for the wolfSSL vtls backend

icing:
Removing special BIO return code handling, relying solely on retry flag in BIO, similar to OpenSSL vtls implementation.


bagder:
Your commit here includes `unit3200` which is a mistake. Maybe rebase and force-push without it?

jay:
before pushing this upstream the commit message needs to be modified to meet the guidelines and include a reference. example:

~~~
wolfssl: remove special BIO return code handling

Rely solely on retry flag in BIO, similar to OpenSSL vtls implementation.

Ref: https://github.com/curl/curl/pull/10021#issuecomment-1336147053

Closes #xxxx
~~~

icing:
Rebased, squash-merged, without the now ignored file, painted pink with a light vanilla taste.

bagder:
Thanks!

PR #10036, opened 2022-12-05
curl.h: name all public function parameters

ryandesign:
Most public function parameters already have names; this adds those that were missing.

-----

I took the parameter names from where the functions were defined in the C sources. In some cases this does not match the documentation. For example, in lib/slist.c (and, with this PR, in curl.h) the second parameter of `curl_slist_append` is called `data` while [in the documentation](https://curl.se/libcurl/c/curl_slist_append.html) it's called `string`.

-----

My motivation for making these changes arose when I wrote a script to parse the curl headers in part to get the public function parameter names and I noticed a few functions didn't have them. Do you have any objections to adding these?


bagder (approved):

bagder:
Thanks!

PR #10027, opened 2022-12-04
imap: Fallback to the LOGIN command when the AUTHENTICATE command fails

kop316:
Hello!

This is a continuation of https://github.com/curl/curl/pull/8986

As per section 6.2.3 of RFC-9051 allow the fallback to a clear text LOGIN when a SASL based AUTHENTICATE fails.

I see two notes on the original PR:

https://github.com/curl/curl/pull/8986#issuecomment-1179588099

There's an option to prevent this without TLS, but what would break scripts: https://github.com/curl/curl/pull/8986#discussion_r894117162

The other note is: https://github.com/curl/curl/pull/8986#issuecomment-1190134446

The way I read the RFC, it looks like the same authentication should be used, but I may be wrong on that.

I am happy to attempt what I need to do to get this accepted. Thank you!

PR #10041, opened 2022-12-06
imap: Provide method to disable SASL if it is advertised

kop316:
Hello!

There is no method to be able to use plain LOGIN (Plain LOGIN vs. SASL LOGIN) if an IMAP server advertises SASL capabilities. However, this may be desirable for e.g. a misconfigured server.

Per: https://www.ietf.org/rfc/rfc5092.html#section-3.2

";AUTH=<enc-auth-type>" looks to be the correct way to specify what authenication method to use, regardless of SASL or not.

Since https://www.rfc-editor.org/rfc/rfc3501.html#section-6.1.1 provides no way to have no SASL mechanisms, "XNONE" looks to be the best RFC compliant way of doing it.

While this does provide a method for cleartext password being sent, a user must expliciticly call `AUTH=XNONE` to excerise this.

I should note I originally thought of adding this in `Curl_sasl_parse_url_auth_option()` to add a more general way of disabling SASL authentication (it did not break any scripts), but I wasn't sure which would be more appropriate.

### DIFF
--- a/docs/cmdline-opts/login-options.d
+++ b/docs/cmdline-opts/login-options.d
@@ -16,3 +16,9 @@ You can use login options to specify protocol specific options that may be
 used during authentication. At present only IMAP, POP3 and SMTP support
 login options. For more information about login options please see RFC
 2384, RFC 5092 and IETF draft draft-earhart-url-smtp-00.txt
+
+Since 8.2.0, IMAP supports the login option "AUTH=+LOGIN". With this option,
+curl uses the plain (not SASL) LOGIN IMAP command even if the server advertises
+SASL authentication. Care should be taken in using this option, as it will send
+out your password in plain text. This will not work if the IMAP server disables
+the plain LOGIN (e.g. to prevent password snooping).
--- a/docs/libcurl/opts/CURLOPT_LOGIN_OPTIONS.3
+++ b/docs/libcurl/opts/CURLOPT_LOGIN_OPTIONS.3
@@ -43,6 +43,12 @@ options, such as the preferred authentication mechanism via "AUTH=NTLM" or
 "AUTH=*", and should be used in conjunction with the \fICURLOPT_USERNAME(3)\fP
 option.
 
+Since 8.2.0, IMAP supports the login option "AUTH=+LOGIN". With this option,
+curl uses the plain (not SASL) LOGIN IMAP command even if the server advertises
+SASL authentication. Care should be taken in using this option, as it will send
+out your password in plain text. This will not work if the IMAP server disables
+the plain LOGIN (e.g. to prevent password snooping).
+
 The application does not have to keep the string around after setting this
 option.
 .SH DEFAULT
--- a/lib/imap.c
+++ b/lib/imap.c
@@ -1925,6 +1925,7 @@ static CURLcode imap_parse_url_options(struct connectdata *conn)
   CURLcode result = CURLE_OK;
   struct imap_conn *imapc = &conn->proto.imapc;
   const char *ptr = conn->options;
+  bool prefer_login = false;
 
   while(!result && ptr && *ptr) {
     const char *key = ptr;
@@ -1938,26 +1939,39 @@ static CURLcode imap_parse_url_options(struct connectdata *conn)
     while(*ptr && *ptr != ';')
       ptr++;
 
-    if(strncasecompare(key, "AUTH=", 5))
+    if(strncasecompare(key, "AUTH=+LOGIN", 11)) {
+      /* User prefers plaintext LOGIN over any SASL, including SASL LOGIN */
+      prefer_login = true;
+      imapc->sasl.prefmech = SASL_AUTH_NONE;
+    }
+    else if(strncasecompare(key, "AUTH=", 5)) {
+      prefer_login = false;
       result = Curl_sasl_parse_url_auth_option(&imapc->sasl,
                                                value, ptr - value);
-    else
+    }
+    else {
+      prefer_login = false;
       result = CURLE_URL_MALFORMAT;
+    }
 
     if(*ptr == ';')
       ptr++;
   }
 
-  switch(imapc->sasl.prefmech) {
-  case SASL_AUTH_NONE:
-    imapc->preftype = IMAP_TYPE_NONE;
-    break;
-  case SASL_AUTH_DEFAULT:
-    imapc->preftype = IMAP_TYPE_ANY;
-    break;
-  default:
-    imapc->preftype = IMAP_TYPE_SASL;
-    break;
+  if(prefer_login)
+    imapc->preftype = IMAP_TYPE_CLEARTEXT;
+  else {
+    switch(imapc->sasl.prefmech) {
+    case SASL_AUTH_NONE:
+      imapc->preftype = IMAP_TYPE_NONE;
+      break;
+    case SASL_AUTH_DEFAULT:
+      imapc->preftype = IMAP_TYPE_ANY;
+      break;
+    default:
+      imapc->preftype = IMAP_TYPE_SASL;
+      break;
+    }
   }
 
   return result;
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -102,6 +102,7 @@ test700 test701 test702 test703 test704 test705 test706 test707 test708 \
 test709 test710 test711 test712 test713 test714 test715 test716 test717 \
 test718 test719 test720 test721 \
 \
+test799 \
 test800 test801 test802 test803 test804 test805 test806 test807 test808 \
 test809 test810 test811 test812 test813 test814 test815 test816 test817 \
 test818 test819 test820 test821 test822 test823 test824 test825 test826 \
--- a/tests/data/test799
+++ b/tests/data/test799
@@ -1,0 +1,53 @@
+<testcase>
+<info>
+<keywords>
+IMAP
+Clear Text
+SASL AUTH +LOGIN
+</keywords>
+</info>
+
+#
+# Server-side
+<reply>
+<servercmd>
+AUTH PLAIN
+REPLY LOGIN A002 OK LOGIN completed
+</servercmd>
+<data>
+From: me@somewhere
+To: fake@nowhere
+
+body
+
+--
+  yours sincerely
+</data>
+</reply>
+
+#
+# Client-side
+<client>
+<server>
+imap
+</server>
+ <name>
+IMAP with --login-options 'AUTH=+LOGIN'
+ </name>
+ <command>
+'imap://%HOSTIP:%IMAPPORT/%TESTNUMBER/;MAILINDEX=1' -u user:secret --login-options AUTH=+LOGIN
+</command>
+</client>
+
+#
+# Verify data after the test has been "shot"
+<verify>
+<protocol>
+A001 CAPABILITY
+A002 LOGIN user secret
+A003 SELECT %TESTNUMBER
+A004 FETCH 1 BODY[]
+A005 LOGOUT
+</protocol>
+</verify>
+</testcase>


bagder:
I'm totally fine with doing it this way. Two things:

- This needs to be documented.
- Why `XNONE` and not just `NONE`?

kop316:
> I'm totally fine with doing it this way. Two things:

Sounds good!

> 
>     * This needs to be documented.

That is fine! How do I go about documenting this?

> 
>     * Why `XNONE` and not just `NONE`?

In https://www.rfc-editor.org/rfc/rfc3501.html :

```
capability      = ("AUTH=" auth-type) / atom
                    ; New capabilities MUST begin with "X" or be
                    ; registered with IANA as standard or
                    ; standards-track
```

I read that to mean that since `NONE` is not an IANA standard (they don't seem to have any reserved words for this), I needed to put an `X` in front of it. However, if you disagree and prefer `NONE`, that is fine too and I can change it.

kop316:
Hello!

I just wanted to follow up with this. I believe I have the documentation added (though that's pending review). Should I also change `XNONE` to `NONE`? I don't mind that, I just wasn't sure if my reasoning in the previous comment was correct or not.

kop316:
Hello and happy new year!

I just wanted to follow up on this again. I was waiting for feedback on the comments here: https://github.com/curl/curl/pull/10041#issuecomment-1339186739

to do any changes, and I wanted to check in to see what your thoughts were.

kop316:
> I'm totally fine with doing it this way. Two things:
> 
>     * This needs to be documented.
> 
>     * Why `XNONE` and not just `NONE`?

Hello,

Based on this, I added documentation and revert it to `NONE`. Please let me know if there is anything else you would like me to do for it.

jay:
The other PR is held up by a security concern and my lack of SASL knowledge. If AUTH NONE is used I would think then there is no authentication? I don't see 'none' in the rfcs

monnerat:
I really prefer this way of resolving the problem rather than the fallback of #10027: the later causes a very fast double authentication attempt that may appear to the server as an attack and finally locks the user and/or ip.

The problem is real: there is no way to select a non-SASL login method if some SASL are advertised by the server. It exists since introduction of SASL in curl and probably also affects POP3, SMTP and LDAP.

The ideal solution would be to consider ALL authentication methods (SASL + protocol login + APOP, etc) without discrimination and allow giving them in the `AUTH=` parameter, but I currently don't have time for that change, although plenty of ideas.

In the meantime I would prefer something like `AUTH=XLOGIN` and reserve the `NONE`/`XNONE` in case a "no-authentication" feature is needed some day.

Please note that POP3 URL requires a non-SASL mechanism name to begin with a '`+`' (RFC 2384 page 2), but there is no such restriction for IMAP (RFC 2362). `XLOGIN` or `+LOGIN` seem good possible mechanism names for IMAP, as `LOGIN` is already used by SASL The former has the advantage to be POP3-compatible if introduced some day.

kop316:
@jay @monnerat 

This PR will solve the underlying issue I am facing, so i am fine with this instead of the other one. 

Reading though https://github.com/curl/curl/pull/10041#issuecomment-1502520496 , if I change `NONE` to `XLOGIN` or `+LOGIN`, would this PR be good for merging? I am admittedly confused, as it sounded like `+LOGIN` may be better for POP from:

>  POP3 URL requires a non-SASL mechanism name to begin with a '+' (RFC 2384 page 2)

But 

> The former has the advantage to be POP3-compatible if introduced some day.

Implied that `XLOGIN` is preferable.

Please let me know which one you would prefer and I will change as appropriate.

Thank you!

monnerat:
> Please let me know which one you would prefer and I will change as appropriate.

I'm in favor of `+LOGIN`: this will allow a more hortogonal option string between protocols (we already have `+APOP` in POP3). But others' opinion matter, of course.

kop316:
> > Please let me know which one you would prefer and I will change as appropriate.
> 
> I'm in favor of `+LOGIN`: this will allow a more hortogonal option string between protocols (we already have `+APOP` in POP3). But others' opinion matter, of course.

I personally do not have a strong opinion, as they would all work for me equally as well. I went ahead and changed `NONE` to `+LOGIN` in another commit. Perhaps in a few days if no one chimes in I just squash for merging?

jay:
@monnerat does this look correct to you now, if SASL_AUTH_NONE is set to use IMAP_TYPE_CLEARTEXT; because SASL_AUTH_NONE is used when prefs are reset so should it be defaulting to cleartext in that case?

monnerat:
> does this look correct to you now, 

It looks like, yes.

> if SASL_AUTH_NONE is set to use IMAP_TYPE_CLEARTEXT; because SASL_AUTH_NONE is used when prefs are reset so should it be defaulting to cleartext in that case?

This is the center of the initial problem and I'm not the father of this code. Probably some more testing should be done on it, but as stated earlier, I don't have time for it right now. However I'm quite confident but I won't take this responsibility.

kop316:
> This is the center of the initial problem and I'm not the father of this code. Probably some more testing should be done on it, but as stated earlier, I don't have time for it right now. However I'm quite confident but I won't take this responsibility.

Sorry, just so I am clear, are you saying more testing needs to be done on the MR, or more testing needs to be done on the initial problem to completely fix it?

I just want to make sure you are not needing anything else from me.

monnerat:
> Sorry, just so I am clear, are you saying more testing needs to be done on the MR, or more testing needs to be done on the initial problem to completely fix it?

I'm just saying the PR looks good, but as this part of libcurl code is not very clear, I cannot predict possible border effects :-)

kop316:
Ahh, fair enough!  I could never get the code to reach the line here: https://github.com/curl/curl/pull/10041/files#diff-9dd142f0e3406c05d5878d67ca2460c070b4e8f684abd673e6e38f58493cb008L1953 . That is why I manually changed it how I did. However if there is a desire to change it, please let me know and I will.

kop316:
@jay @monnerat Hello! I just wanted to follow up with this. 

It sounded like the PR is good, but I also coded up this method to do the same thing: https://github.com/curl/curl/compare/master...kop316:curl:wip/sasl2 and it passes CI as well in case there was a worry about side effects.

monnerat:
I'm not OK with the other:

> #define SASL_AUTH_DISABLE       0xff00

These are supposed to be a mechanisn bits mask and not a specific token value.
It is equivalent of `SASL_MECH_OAUTHBEARER | SASL_MECH_SCRAM_SHA_1 | SASL_MECH_SCRAM_SHA_256 | some_other_yet_undefined_mechanisms` and will lock our inevitable future use of the latter in an orthogonal way.

kop316:
> I'm not OK with the other:
> 
> > #define SASL_AUTH_DISABLE       0xff00
> 
> These are supposed to be a mechanisn bits mask and not a specific token value. It is equivalent of `SASL_MECH_OAUTHBEARER | SASL_MECH_SCRAM_SHA_1 | SASL_MECH_SCRAM_SHA_256 | some_other_yet_undefined_mechanisms` and will lock our inevitable future use of the latter in an orthogonal way.

That's fair, thanks for explaining! It sounds like I will stick to the original PR then.

jay:
@kop316  I just pushed a commit to your branch that adds a separate 'prefer_login' variable that will determine whether or not to set IMAP_TYPE_CLEARTEXT. @monnerat are there any problems with this approach? If so then Chris can just force push back to where it was.

monnerat:
> are there any problems with this approach?

No, this should work.

kop316:
> @kop316 I just pushed a commit to your branch that adds a separate 'prefer_login' variable that will determine whether or not to set IMAP_TYPE_CLEARTEXT. @monnerat are there any problems with this approach? If so then Chris can just force push back to where it was.

Sounds good to me, if it passes CI it should work for my issue!

kop316:
> The test for this doesn't look like it tests for LOGIN preference over SASL, just LOGIN by itself.

I modeled it after: https://github.com/curl/curl/blob/master/tests/data/test819#L15 , in which the server is replaying that it supports SASL AUTH PLAIN. Without the MR, curl would attempt to authenticate with SASL PLAIN instead of LOGIN. So to the best of my knowledge, it is testing that curl will prefer LOGIN even if presented with a SASL preference.

Is there something I am missing?

jay:
> Is there something I am missing?

No it's probably something I'm missing, as I'm not too familiar with SASL. I will take a look tonight to confirm.

kop316:
> > Is there something I am missing?
> 
> No it's probably something I'm missing, as I'm not too familiar with SASL. I will take a look tonight to confirm.

No worries! if you have any questions I will try my best to answer. I think this comment describes the issue: https://github.com/curl/curl/pull/10041#issuecomment-1502520496 

> there is no way to select a non-SASL login method if some SASL are advertised by the server. It exists since introduction of SASL in curl and probably also affects POP3, SMTP and LDAP.

So my thought process for the test was to present curl with a SASL method (in this case, `SASL PLAIN`) of logging in, and make sure that with the `AUTH=+LOGIN` curl still uses plain (not SASL) LOGIN.

kop316:
> > Is there something I am missing?
> 
> No it's probably something I'm missing, as I'm not too familiar with SASL. I will take a look tonight to confirm.

No worries! if you have any questions I will try my best to answer. I think this comment describes the issue: https://github.com/curl/curl/pull/10041#issuecomment-1502520496 

> there is no way to select a non-SASL login method if some SASL are advertised by the server. It exists since introduction of SASL in curl and probably also affects POP3, SMTP and LDAP.

So my thought process for the test was to present curl with a SASL method (in this case, `SASL PLAIN`) of logging in, and make sure that with the `AUTH=+LOGIN` curl still uses plain (not SASL) LOGIN.

jay:
> So my thought process for the test was to present curl with a SASL method (in this case, `SASL PLAIN`) of logging in, and make sure that with the `AUTH=+LOGIN` curl still uses plain (not SASL) LOGIN.

So on further review:

AFAICT your test expects curl to send the `LOGIN user secret` not `AUTHENTICATE PLAIN` or `AUTHENTICATE LOGIN` for `AUTH PLAIN` or `AUTH LOGIN` server capability, respectively. Therefore there's no reason to handle AUTHENTICATE in servercmd section. OK.

What is confusing to me is the difference between server `AUTH LOGIN` and client `AUTH=+LOGIN`. If `AUTH LOGIN` was set server side that would mean the server sends `CAPABILITY IMAP4 AUTH=LOGIN`. If that was the only mech the client would send base64 encoded login like `[NUL]user[NUL]secret`. Ok I get that's not what you want. If `AUTH=+LOGIN` was set client side the plaintext login would be used like `LOGIN user secret`. Would people familiar with these protocols confuse those two because it confused me. Should the doc be improved to clarify this?

kop316:
> > So my thought process for the test was to present curl with a SASL method (in this case, `SASL PLAIN`) of logging in, and make sure that with the `AUTH=+LOGIN` curl still uses plain (not SASL) LOGIN.
> 
> So on further review:
> 
> AFAICT your test expects curl to send the `LOGIN user secret` not `AUTHENTICATE PLAIN` or `AUTHENTICATE LOGIN` for `AUTH PLAIN` or `AUTH LOGIN` server capability, respectively. Therefore there's no reason to handle AUTHENTICATE in servercmd section. OK.

Yep, that is correct.

> 
> What is confusing to me is the difference between server `AUTH LOGIN` and client `AUTH=+LOGIN`. If `AUTH LOGIN` was set server side that would mean the server sends `CAPABILITY IMAP4 AUTH=LOGIN`. If that was the only mech the client would send base64 encoded login like `[NUL]user[NUL]secret`. Ok I get that's not what you want. If `AUTH=+LOGIN` was set client side the plaintext login would be used like `LOGIN user secret`. Would people familiar with these protocols confuse those two because it confused me. Should the doc be improved to clarify this?

If it confused you, the doc should probably be improved for it. I made an update to the documentation, does that help?

jay:
Looks good, thanks. This PR is an enhancement and not a bug fix so it will have to wait until [the next feature window](https://calendar.google.com/calendar/u/0/embed?src=c9u5d64odop9js55oltfarjk6g@group.calendar.google.com&ctz=Europe/Stockholm) which means it will be in the July release.

vszakats:
Can we consider limiting this fallback to IMAPS connections only?

My concern is apps enabling this under the hood and leaking passwords in cleartext over the network.

bagder:
> IMAPS connections only

I think that might be clever.

Just a caution around that: I imagine that the larger use of TLS for IMAP is done with STARTTLS, so they are actually using an `IMAP://`  URL but with CURLOPT_USE_SSL set. So they start out clear text but upgrade into TLS. It unfortunately makes it a little harder for libcurl to know if it is an "IMAPS" connection at any given point.

monnerat:
> Can we consider limiting this fallback to IMAPS connections only?
> My concern is apps enabling this under the hood and leaking passwords in cleartext over the network.

That's what PR #8295 is intended to prevent, in a more global scope as SASL may also transmit clear passwords.

vszakats:
`STARTTLS` is trivial to downgrade-attack, so maybe it in this context we can consider is as cleartext IMAP.

monnerat:
> IMO this adds a high risk for leaking email account passwords.

Yes and no !

- Yes because `AUTH=+LOGIN` does require a cleartext password authentication.
- No because using an `AUTH=LOGIN` SASL mechanism is not better. In addition it can be selected automatically if it's the only SASL mechanism advertised by the server.
- No again, because if a server does not advertise SASL, IMAP LOGIN is used automatically.
- An still no, because `AUTH=+LOGIN` must be requested explicitly if SASL is available.

IMO, disabling cleartext password and using SASL mechanism or not should not be bound. The above mentioned PR treats this (disjoint) problem globally, It is currently awaiting your vote and/or comment. Maybe it should be implemented the other way round (i.e.: disabling cleartext password by default) at the cost of backwards incompatibility but the topic comes periodically in PR comments and mailing list, so something equivalent should be implemented to definitively  close this chapter. I really don't understand why #8295 has not yet been adopted (or requested for changes) with regards to all these password security gripes.

jay:
I disagree, no more work is needed. This PR adds an option to specifically use plaintext LOGIN instead of server specified SASL. If the user sets it as a login option then that is what we give them.

vszakats:
Related news from a few days ago: https://mailbox.org/en/post/mailbox-org-discovers-unencrypted-password-transmission-in-mymail (on HN: https://news.ycombinator.com/item?id=35845308)

In this case it was the lack of TLS + one of the variations of cleartext password. Adding more ways to achieve this doesn't help.

kop316:
Hello,

Given that we are in the feature window for the next curl release, I wanted to check back in to see if there is anything else needed from me for this.

Thank you!

jay:
Thanks

PR #10042, opened 2022-12-06
config-mac: define HAVE_SYS_IOCTL_H

ryandesign:
~~Define `HAVE_SYS_FILIO_H` in config-mac.h and include <sys/filio.h> in nonblock.c when `HAVE_SYS_FILIO_H` is defined.~~

Define `HAVE_SYS_IOCTL_H` in config-mac.h. This is needed to compile nonblock.c on classic Mac OS with GUSI because nonblock.c uses `FIONBIO` which is defined in <sys/filio.h> which is included by <sys/ioctl.h>.

### DIFF
--- a/lib/config-mac.h
+++ b/lib/config-mac.h
@@ -53,6 +53,7 @@
 #define HAVE_UTIME_H            1
 #define HAVE_SYS_TIME_H         1
 #define HAVE_SYS_UTIME_H        1
+#define HAVE_SYS_IOCTL_H        1
 
 #define TIME_WITH_SYS_TIME      1
 


lib/config-mac.h
```diff
@@ -53,6 +53,7 @@
 #define HAVE_UTIME_H            1
 #define HAVE_SYS_TIME_H         1
 #define HAVE_SYS_UTIME_H        1
+#define HAVE_SYS_FILIO_H        1
```
danielgustafsson, line 56: Daft question since I don't know Mac OS other than as a user (a long time ago..). Is `filio.h` available on all versions of Mac OS on which it would be feasible to build libcurl? In other words, does it go further back than OS9?

Disclaimer: I'm mainly curious, I don't think it's a problem to define this if it doesn't

lib/nonblock.c
```diff
@@ -31,6 +31,9 @@
 #include <fcntl.h>
 #endif
 
+#ifdef HAVE_SYS_FILIO_H
```
danielgustafsson, line 34: I wonder if we should invent a `__CURL_MACOS_CLASSIC__` guard in `lib/config-mac.h` and use that as well here to keep the header from not being included when not needed?

lib/config-mac.h
ryandesign: Classic Mac OS does not have filio.h or other Unix headers. To build curl on classic Mac OS, the [Grand Unified Socket Interface (GUSI)](https://sourceforge.net/projects/gusi/) is required. It provides an early 2000s Unix-like socket and threads interface for classic Mac OS. It is compatible with System 7, Mac OS 8, and Mac OS 9.

lib/nonblock.c
ryandesign: Is your concern that including sys/filio.h when present would be a problem for some OS other than classic Mac OS?

lib/config-mac.h
danielgustafsson: Gotcha, thanks.

lib/nonblock.c
ryandesign: I was modeling my change on tests/server/tftpd.c where it uses the same condition:

```c
#ifdef HAVE_SYS_FILIO_H
/* FIONREAD on Solaris 7 */
#include <sys/filio.h>
#endif
```

But let me double-check if this is needed. Even modern macOS 10.15.7 defines `FIONBIO` in <sys/filio.h> so I was surprised curl had gotten away with not including that header here, but I see now that nonblock.c also includes <sys/ioctl.h> if present which does (on both modern macOS and in GUSI) include <sys/filio.h> already.

danielgustafsson: Aha, even better!

ryandesign: The error without my change is:

```sh
#----------------------------------------------------------
### MWC68K Compiler Error:
#	return ioctl(sockfd, FIONBIO, &flags);
#	                     ^^^^^^^
# undefined identifier 'FIONBIO'
#----------------------------------------------------------
File "nonblock.c"; Line 64
#----------------------------------------------------------
# errors caused tool to abort
make[3]: *** [nonblock.68k.o] Error 1
```

When I first saw that, I grep'd the GUSI headers to see where `FIONBIO` was defined, found it in <sys/filio.h>, found an existing include of <sys/filio.h> in nonblock.c that only took effect on NetWare (which was removed in 2d9fee450427bf20284278022f829f7cf3c05d21), and I figured I should change it to include it more broadly.

However it appears to be sufficient solely to `#define HAVE_SYS_IOCTL_H 1` in config-mac.h. Shall I change this PR to just do that? Or open a new PR to do that?

danielgustafsson: You can update this PR, that's fine.

jay:
Thanks

PR #10043, opened 2022-12-06
KNOWN_BUGS: remove items not considered bugs any more

bagder:
- CURL_GLOBAL_SSL

This option was changed in libcurl 7.57.0 and clearly it has not caused too many issues and a lot of time has passed.

- Schannel disable CURLOPT_SSL_VERIFYPEER and verify hostname

This is a restriction in the Schannel API set, not a libcurl bug.

- Store TLS context per transfer instead of per connection

This is a possible future optimization. One that is much less important and interesting since the added support for CA caching.

- Microsoft telnet server

This bug was filed in May 2007 against curl 7.16.1 and we have not received further reports.

- active FTP over a SOCKS

Actually, proxies in general is not working with active FTP mode. This is now added in proxy documentation.

- DICT responses show the underlying protocol

curl still does this, but since this is now an established behavior since forever we cannot change it easily and adding an option for it seems crazy as this protocol is not so little its not worth it. Let's just live with it.

### DIFF
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -12,25 +12,18 @@ check the changelog of the current development status, as one or more of these
 problems may have been fixed or changed somewhat since this was written.
 
  1. HTTP
- 1.3 STARTTRANSFER time is wrong for HTTP POSTs
- 1.4 multipart formposts file name encoding
  1.5 Expect-100 meets 417
- 1.8 DoH is not used for all name resolves when enabled
- 1.11 CURLOPT_SEEKFUNCTION not called with CURLFORM_STREAM
 
  2. TLS
  2.3 Unable to use PKCS12 certificate with Secure Transport
  2.4 Secure Transport will not import PKCS#12 client certificates without a password
  2.5 Client cert handling with Issuer DN differs between backends
- 2.6 CURL_GLOBAL_SSL
  2.7 Client cert (MTLS) issues with Schannel
  2.8 Schannel disable CURLOPT_SSL_VERIFYPEER and verify hostname
  2.9 TLS session cache does not work with TFO
- 2.10 Store TLS context per transfer instead of per connection
  2.11 Schannel TLS 1.2 handshake bug in old Windows versions
  2.12 FTPS with Schannel times out file list operation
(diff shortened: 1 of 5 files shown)


docs/KNOWN_BUGS
```diff
@@ -116,9 +109,6 @@ problems may have been fixed or changed somewhat since this was written.
  13. TCP/IP
  13.2 Trying local ports fails on Windows
 
- 14. DICT
```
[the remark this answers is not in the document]
bagder: Maybe, but I figured it is rather unlikely that we will add another known bug for DICT in the short term future! :grimacing:  We can just add it again if that unlikely event happens...

docs/KNOWN_BUGS
```diff
@@ -12,25 +12,17 @@ check the changelog of the current development status, as one or more of these
...
- 1.8 DoH is not used for all name resolves when enabled
- 1.11 CURLOPT_SEEKFUNCTION not called with CURLFORM_STREAM
 
  2. TLS
  2.3 Unable to use PKCS12 certificate with Secure Transport
  2.4 Secure Transport will not import PKCS#12 client certificates without a password
  2.5 Client cert handling with Issuer DN differs between backends
- 2.6 CURL_GLOBAL_SSL
  2.7 Client cert (MTLS) issues with Schannel
- 2.8 Schannel disable CURLOPT_SSL_VERIFYPEER and verify hostname
```
jay, line 27: I've filed https://github.com/curl/curl/pull/10056 to address this

bagder: Awesome. I will make sure I merge this PR after yours.

docs/libcurl/opts/CURLOPT_FTPPORT.3
```diff
@@ -59,6 +59,11 @@ Examples with specified ports:
   [::1]:1234-4567
 .fi
 
+We strongly advice against specify the address with a name, as it causes
```
andy5995, line 62:
```suggestion
We strongly advice against specifying the address with a name, as it causes
```

dfandrich:
Also, "advise", not "advice".

PR #10044, opened 2022-12-06
runtests: silence nghttpx errors

bagder:
Also, move the output of the nghttpx_h3 info to the general "Env:" line in the test output header.

Reported-by: Marcel Raad
Ref: https://github.com/curl/curl/commit/ca15b7512e8d1199e55fbaa206ef01e64b8f147d#commitcomment-92015094


bagder:
@icing you ok with this?

icing:
> @icing you ok with this?

Sure!

PR #10046, opened 2022-12-06
tool_formparse: avoid clobbering on function params

danielgustafsson:
While perfectly legal to do, clobbering function parameters and using them as local variables is confusing at best and rarely improves code readability.  Fix by using a local variable instead, no functionality is changed.

### DIFF
--- a/src/tool_formparse.c
+++ b/src/tool_formparse.c
@@ -61,17 +61,18 @@ static struct tool_mime *tool_mime_new_parts(struct tool_mime *parent)
 }
 
 static struct tool_mime *tool_mime_new_data(struct tool_mime *parent,
-                                            char *data)
+                                            char *mime_data)
 {
+  char *mime_data_copy;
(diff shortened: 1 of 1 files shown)


bagder (approved):
Also, we sort of reserve the use of the name `data` all over libcurl code to point to the easy handle struct, and this is used for a different purpose...

Would it also make sense to change the function argument to `const char *data` you think?

danielgustafsson:
> Also, we sort of reserve the use of the name `data` all over libcurl code to point to the easy handle struct, and this is used for a different purpose...
> 
> Would it also make sense to change the function argument to `const char *data` you think?

I think it does, maybe even go further and rename from data to `const char *mime_data` and in turn rename the local variable something else?

bagder:
I would approve!

danielgustafsson:
Pushed a rebase with the discussed changes.

bagder (approved):

PR #10049, opened 2022-12-07
Classic Mac OS: fix socklen_t, curl_off_t, long long

ryandesign:
Change `CURL_TYPEOF_CURL_SOCKLEN_T` from `int` to `unsigned int` when `__MWERKS__` is defined (i.e. Metrowerks compilers i.e. CodeWarrior).

This fixes a build failure when compiling for classic Mac OS with GUSI with CodeWarrior compilers.

-----

The error was:

```sh
#----------------------------------------------------------
### MWC68K Compiler Error:
#	if(getsockname(sockfd, (struct sockaddr *) &add, &size) < 0) {
#	                                                      ^
# cannot convert
# 'int *' to
# 'unsigned int *'
#----------------------------------------------------------
File "connect.c"; Line 462
#----------------------------------------------------------
### MWC68K Compiler Error:
#	if(0 != getsockopt(sockfd, SOL_SOCKET, SO_ERROR, (void *)&err, &errSize))
#	                                                                       ^
# cannot convert
# 'int *' to
# 'unsigned int *'
#----------------------------------------------------------
File "connect.c"; Line 535
#----------------------------------------------------------
```

-----

It's unclear to me what system was envisioned when the `__MWERKS__` block was originally added in 2008 in 3ac692991912e596959aaedb49ff2798537e7af5. Metrowerks CodeWarrior was originally a classic Mac OS compiler in the 90s but, later, versions for BeOS, Windows, and Mac OS X existed, and it could cross-compile code for other systems as well. The Metrowerks company no longer exists today but CodeWarrior still does and is now used for embedded systems. I'm not certain that all of the systems CodeWarrior supports or supported will have the same values for the constants being defined here.

For example, POSIX says `socklen_t` should be unsigned.

* GUSI for classic Mac OS started using `socklen_t` in version 2. It is unsigned in the earliest version still posted on the site (2.1.3 from 2000) and in the last version (2.2.3 from 2002).
* Mac OS X first used `socklen_t` in version 10.3 (2003) and it was signed. By Mac OS X 10.4 (2005) it was unsigned.
* BeOS (1995-2001) apparently used a signed `socklen_t` at least for a time according to a quick Internet search. Don't know what Haiku (the reimplementation of BeOS) does.
* No idea what Windows or embedded systems do.

I'm not sure how these potential differences should be handled here.

### DIFF
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -170,7 +170,7 @@
 #  define CURL_FORMAT_CURL_OFF_TU    "llu"
 #  define CURL_SUFFIX_CURL_OFF_T     LL
 #  define CURL_SUFFIX_CURL_OFF_TU    ULL
-#  define CURL_TYPEOF_CURL_SOCKLEN_T int
+#  define CURL_TYPEOF_CURL_SOCKLEN_T unsigned int
 
 #elif defined(__TANDEM)
 # if ! defined(__LP64)
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -165,11 +165,20 @@
 #  define CURL_TYPEOF_CURL_SOCKLEN_T unsigned int
 
 #elif defined(macintosh)
-#  define CURL_TYPEOF_CURL_OFF_T     long long
-#  define CURL_FORMAT_CURL_OFF_T     "lld"
-#  define CURL_FORMAT_CURL_OFF_TU    "llu"
-#  define CURL_SUFFIX_CURL_OFF_T     LL
-#  define CURL_SUFFIX_CURL_OFF_TU    ULL
+#  include <ConditionalMacros.h>
+#  if TYPE_LONGLONG
+#    define CURL_TYPEOF_CURL_OFF_T     long long
+#    define CURL_FORMAT_CURL_OFF_T     "lld"
+#    define CURL_FORMAT_CURL_OFF_TU    "llu"
+#    define CURL_SUFFIX_CURL_OFF_T     LL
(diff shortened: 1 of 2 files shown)


jay:
> I'm not certain that all of the systems CodeWarrior supports or supported will have the same values for the constants being defined here.

Maybe we should stop using __MWERKS__ or add more qualifiers? I think we can assume that section was probably for old macs.

dfandrich:
FWIW, Haiku defines socklen_t as uint32_t in sys/socket.h

ryandesign:
> Maybe we should stop using `__MWERKS__` or add more qualifiers?

Perhaps so. Several of the other top-level conditions in that file are already about operating systems (e.g. `_WIN32_WCE`, `__VMS`, `__OS400__`), not compilers.

> I think we can assume that section was probably for old macs.

Probably. I've rewritten the PR for this. I've divided it into a few hopefully sensibly distinct changes.

jay:
Is SIZEOF_LONG defined for you? I don't see it in config-mac and there are a few places in the library where it's evaluated

lib/mprintf.c:#if (SIZEOF_CURL_OFF_T > SIZEOF_LONG)
lib/mprintf.c:#if (SIZEOF_SIZE_T > SIZEOF_LONG)
lib/mprintf.c:#if (SIZEOF_CURL_OFF_T > SIZEOF_LONG)

ryandesign:
Good point! I've added a commit to define `SIZEOF_LONG`.

bagder (approved):

bagder:
Thanks!

Issue #3284, opened 2018-11-17
Schannel can't disable only CURLOPT_SSL_VERIFYPEER and still verify the host name


bagder:
### I did this

Martin Galvan [reported on the mailing list](https://curl.haxx.se/mail/lib-2018-10/0113.html):

> I was looking at the WinSSL code which processes
> CURLOPT_SSL_VERIFYPEER and CURLOPT_SSL_VERIFYHOST, and noticed that,
> if CURLOPT_SSL_VERIFYPEER isn't enabled, the host name checking won't
> be performed:

> https://github.com/curl/curl/blob/master/lib/vtls/schannel_verify.c#L539
> https://github.com/curl/curl/blob/master/lib/vtls/schannel.c#L1117

### I expected the following

If the TLS backend has the API for it, the options should be independent.

### curl/libcurl version

current git master

### operating system

Windows

jay:
I'll take a look but I thought it was already independent.
https://github.com/curl/curl/blob/196677150f711a96c38ed123e621f1d4e995b2e5/lib/vtls/schannel.c#L536-L541

jay:
It fails with SEC_E_WRONG_PRINCIPAL when verifypeer is enabled, but succeeds when it's disabled. (I'm using https://cdn0.nflximg.net as test server). When verifypeer is disabled then manual verification is enabled SCH_CRED_MANUAL_CRED_VALIDATION. What I think is happening is the manual verification disables the server name check even though we don't. Seems like they could have warned us of that in the doc.

bagder:
@jay any idea on how to deal with this? Should we just close or can we document this somehow somewhere?

PR #3285, opened 2018-11-17
schannel: allow verifyhost independently of verifypeer

bagder:
As documented and as working in other TLS backends.

Fixes #3284

PR #10056, opened 2022-12-08
schannel: verify hostname independent of verify cert

jay:
Prior to this change when CURLOPT_SSL_VERIFYPEER (verifypeer) was off and CURLOPT_SSL_VERIFYHOST (verifyhost) was on we did not verify the hostname in schannel code.

This fixes KNOWN_BUG 2.8 "Schannel disable CURLOPT_SSL_VERIFYPEER and verify hostname". We discussed a fix several years ago in #3285 but it went stale.

Assisted-by: Daniel Stenberg

Bug: https://curl.haxx.se/mail/lib-2018-10/0113.html
Reported-by: Martin Galvan

Ref: https://github.com/curl/curl/blob/curl-7_86_0/docs/KNOWN_BUGS#L304-L308
Ref: https://github.com/curl/curl/pull/3285

Fixes https://github.com/curl/curl/issues/3284
Closes #xxxx

---

This PR looks like a lot of change but it's really not. Most of the change is I moved two functions out of the HAS_MANUAL_VERIFY_API guard and then made one of them public (Curl_verify_host) so that hostnames could be verified independently of the server certificate.

### DIFF
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -21,7 +21,6 @@ problems may have been fixed or changed somewhat since this was written.
  2.4 Secure Transport will not import PKCS#12 client certificates without a password
  2.5 Client cert handling with Issuer DN differs between backends
  2.7 Client cert (MTLS) issues with Schannel
- 2.8 Schannel disable CURLOPT_SSL_VERIFYPEER and verify hostname
  2.11 Schannel TLS 1.2 handshake bug in old Windows versions
  2.12 FTPS with Schannel times out file list operation
  2.13 CURLOPT_CERTINFO results in CURLE_OUT_OF_MEMORY with Schannel
@@ -162,12 +161,6 @@ problems may have been fixed or changed somewhat since this was written.
 2.7 Client cert (MTLS) issues with Schannel
 
  See https://github.com/curl/curl/issues/3145
-
-2.8 Schannel disable CURLOPT_SSL_VERIFYPEER and verify hostname
-
- This seems to be a limitation in the underlying Schannel API.
-
- https://github.com/curl/curl/issues/3284
 
 2.11 Schannel TLS 1.2 handshake bug in old Windows versions
 
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -810,9 +810,9 @@ schannel_acquire_credential_handle(struct Curl_cfilter *cf,
 
     SCH_CREDENTIALS credentials = { 0 };
     TLS_PARAMETERS tls_parameters = { 0 };
-    CRYPTO_SETTINGS crypto_settings[4] = { 0 };
-    UNICODE_STRING blocked_ccm_modes[1] = { 0 };
-    UNICODE_STRING blocked_gcm_modes[1] = { 0 };
+    CRYPTO_SETTINGS crypto_settings[4] = { { 0 } };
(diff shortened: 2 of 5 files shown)


bagder:
I had totally forgot about those old discussions. Thanks for this.

jay:
Original mingw CI failed so I've added some fixes, also moved the functions back to their original location opting instead to change the position of the HAS_MANUAL_VERIFY_API guards so that it is easier to see the changes in the diff.

bagder:
You probably want a ` /* !checksrc! disable TYPEDEFSTRUCT 4 */` before those typedefs.

jay:
Done. What's odd is that there are other typedef structs in the header, possibly an issue with checksrc parsing?

I will take a look again at this tomorrow so if you don't want to wait for the KNOWN_BUGS PR you could omit this bug.

bagder:
I've removed it from #10043

bagder:
@jay what's the state of this PR now?

jay:
This was held up by #9706. I can't find anything in the doc to support his claim that [SECPKG_ATTR_REMOTE_CERT_CONTEXT](https://learn.microsoft.com/en-us/windows/win32/secauthn/querycontextattributes--schannel#SECPKG_ATTR_REMOTE_CERT_CONTEXT) is not the server certificate although it does not say explicitly that it is the server cert: _"Finds a certificate context that contains the end certificate supplied by the server."_ That's good enough for me but I was hoping to look into #9706 and see what happens in Windows 11 that causes him to think this I just haven't had the time yet.

jay:
I've reviewed https://github.com/curl/curl/issues/9706#issuecomment-1370403119 and I do not see anything in the MS documentation that indicates the entire certificate store (the certificate chain) associated with the server cert context (ie the server certificate) is searched by functions that take the server cert as a parameter. I tested in Windows 11 and our code that uses SECPKG_ATTR_REMOTE_CERT_CONTEXT works the same as it does in earlier versions (which makes sense or we would have heard about this in reports). And finally I checked the ReactOS source code and I don't see where CertGetNameString enumerates the entire chain on the basis of one certificate. Even if it did that would be out of line with MS doc.

Based on above I see no reason to delay this any further.

/cc @nmoinvaz

nmoinvaz:
There appear to be two places in curl code where `SECPKG_ATTR_REMOTE_CERT_CONTEXT` is used. 

- In [one place](https://github.com/curl/curl/blob/a9c1e28160b8b42f17bd7cdd27808c4649410c5c/lib/vtls/schannel.c#L1812) `traverse_cert_store` is called directly after, which calls `CertEnumCertificatesInStore`. This appears to use the `CERT_CONTEXT->hCertStore` parameter. 
- In the [other place](https://github.com/curl/curl/blob/a9c1e28160b8b42f17bd7cdd27808c4649410c5c/lib/vtls/schannel_verify.c#L583), it uses `CERT_CONTEXT->cbCertEncoded`.  If the `CERT_CONTEXT->cbCertEncoded` is always the end-entity (leaf) certificate, then it would make sense why functions like `CertGetNameString` work on it. 

In my previous [comment](https://github.com/curl/curl/issues/9706#issuecomment-1370403119), I assumed that you were using the `hCertStore` parameter.

Regarding #9706, it might be possible to instead infer the ordering of the chain in `hCertStore` by using `cbCertEncoded` as a guide. If it comes first, then no need to reorder, otherwise we need to reorder. I will have to do some research on that.

PR #10059, opened 2022-12-08
Use SecPKCS12Import on temporary keychain if p12 certificate not found

### LINKED ISSUE #10038: PKCS#12 certificate does not work with Secure Transport unless itself or private key is in keychain
### I did this

Built cURL 7.86.0 with the following setup:

```
Host setup:       x86_64-apple-darwin22.1.0
Install prefix:   [redacted path]/curl
Compiler:         gcc
  CFLAGS:          -target x86_64-apple-macos10.13 -Qunused-arguments -Wno-pointer-bool-conversion -g -O0 -pedantic -Wall -Wextra -Wpointer-arith -Wwrite-strings -Wshadow -Winline -Wnested-externs -Wmissing-declarations -Wmissing-prototypes -Wno-long-long -Wfloat-equal -Wno-multichar -Wsign-compare -Wundef -Wno-format-nonliteral -Wendif-labels -Wstrict-prototypes -Wdeclaration-after-statement -Wcast-align -Wno-system-headers -Wshorten-64-to-32 -Wunused -Wvla -Wshift-sign-overflow -Wmissing-variable-declarations -Wdouble-promotion -Wcomma -Wassign-enum -Wextra-semi-stmt -Werror=partial-availability
  CPPFLAGS:        -isystem [redacted path]/zlibstat/include -isystem [redacted path]/mbedtls/include -isystem [redacted path]/libssh2/include -isystem [redacted path]/nghttp2/include
  LDFLAGS:         -framework CoreFoundation -framework SystemConfiguration -L[redacted path]/zlibstat/lib -framework CoreFoundation -framework Security -L[redacted path]/mbedtls/lib -L[redacted path]/libssh2/lib -L[redacted path]/nghttp2/lib
  LIBS:            -lnghttp2 -lssh2 -lssh2 -lmbedtls -lmbedx509 -lmbedcrypto -lldap -lzstd -lbrotlidec -lz

curl version:     7.86.0
SSL:              enabled (Secure Transport, mbedTLS)
SSH:              enabled (libSSH2)
zlib:             enabled
brotli:           enabled (libbrotlidec)
zstd:             enabled (libzstd)
GSS-API:          no      (--with-gssapi)
GSASL:            no      (libgsasl not found)
TLS-SRP:          no      (--enable-tls-srp)
resolver:         POSIX threaded
IPv6:             enabled
Unix sockets:     enabled
IDN:              no      (--with-{libidn2,winidn})
Build libcurl:    Shared=no, Static=yes
Built-in manual:  no      (--enable-manual)
--libcurl option: enabled (--disable-libcurl-option)
Verbose errors:   enabled (--disable-verbose)
Code coverage:    disabled
SSPI:             no      (--enable-sspi)
ca cert bundle:   no
ca cert path:     no
ca fallback:      no
LDAP:             enabled (OpenLDAP)
LDAPS:            enabled
RTSP:             no      (--enable-rtsp)
RTMP:             no      (--with-librtmp)
PSL:              no      (libpsl not found)
Alt-svc:          enabled (--disable-alt-svc)
Headers API:      enabled (--disable-headers-api)
HSTS:             enabled (--disable-hsts)
HTTP1:            enabled (internal)
HTTP2:            enabled (nghttp2)
HTTP3:            no      (--with-ngtcp2, --with-quiche --with-msh3)
ECH:              no      (--enable-ech)
WebSockets:       no      (--enable-websockets)
Protocols:        FILE FTP FTPS HTTP HTTPS IMAP IMAPS LDAP LDAPS MQTT POP3 POP3S SCP SFTP SMTP SMTPS
Features:         AsynchDNS HSTS HTTP2 IPv6 Largefile MultiSSL NTLM NTLM_WB SSL UnixSockets alt-svc brotli libz threadsafe zstd
```

Once built, used `./curl --cert [redacted path]/1571753451.p12:test --verbose https://itv.mit-xperts.com/clientssl/check --cert-type 'P12'`

Please note the 1571753451.p12 in use is expired, but this does not impact the bug as the endpoint returns a 400 when the certificate is bad.

### I expected the following

```
*   Trying 89.58.32.80:443...
* Connected to itv.mit-xperts.com (89.58.32.80) port 443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* (304) (OUT), TLS handshake, Client hello (1):
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Request CERT (13):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Certificate (11):
* (304) (OUT), TLS handshake, CERT verify (15):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-AES256-GCM-SHA384
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=itv.mit-xperts.com
*  start date: Nov 19 16:41:36 2022 GMT
*  expire date: Feb 17 16:41:35 2023 GMT
*  subjectAltName: host "itv.mit-xperts.com" matched cert's "itv.mit-xperts.com"
*  issuer: C=US; O=Let's Encrypt; CN=R3
*  SSL certificate verify ok.
* Using HTTP2, server supports multiplexing
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* h2h3 [:method: GET]
* h2h3 [:path: /clientssl/check]
* h2h3 [:scheme: https]
* h2h3 [:authority: itv.mit-xperts.com]
* h2h3 [user-agent: curl/7.84.0]
* h2h3 [accept: */*]
* Using Stream ID: 1 (easy handle 0x7fad8480c600)
> GET /clientssl/check HTTP/2
> Host: itv.mit-xperts.com
> user-agent: curl/7.84.0
> accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
< HTTP/2 400 
< server: nginx/1.20.1
< date: Mon, 05 Dec 2022 18:47:37 GMT
< content-type: text/html
< content-length: 215
< 
<html>
<head><title>400 The SSL certificate error</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<center>The SSL certificate error</center>
<hr><center>nginx/1.20.1</center>
</body>
</html>
* Connection #0 to host itv.mit-xperts.com left intact
```

### Instead, I got the following

```
* STATE: INIT => CONNECT handle 0x7fb5ea00ec08; line 1879 (connection #-5000)
* STATE: CONNECT => RESOLVING handle 0x7fb5ea00ec08; line 1925 (connection #0)
*   Trying 89.58.32.80:443...
* STATE: RESOLVING => CONNECTING handle 0x7fb5ea00ec08; line 2009 (connection #0)
* Connected to itv.mit-xperts.com (89.58.32.80) port 443 (#0)
* STATE: CONNECTING => PROTOCONNECT handle 0x7fb5ea00ec08; line 2074 (connection #0)
* ALPN: offers h2
* ALPN: offers http/1.1
* SSL: Can't find the certificate "[redacted path]/1571753451.p12" and its private key in the Keychain.
* multi_done: status: 58 prem: 1 done: 0
* Closing connection 0
* Expire cleared (transfer 0x7fb5ea00ec08)
curl: (58) SSL: Can't find the certificate "[redacted path]/1571753451.p12" and its private key in the Keychain.
```

### curl/libcurl version (not working one)

```
curl 7.86.0 (x86_64-apple-darwin22.1.0) libcurl/7.86.0 SecureTransport (mbedTLS/2.28.0) zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libssh2/1.10.0 nghttp2/1.48.0
Release-Date: 2022-10-26
Protocols: file ftp ftps http https imap imaps ldap ldaps mqtt pop3 pop3s scp sftp smtp smtps 
Features: alt-svc AsynchDNS brotli Debug HSTS HTTP2 IPv6 Largefile libz MultiSSL NTLM NTLM_WB SSL threadsafe TrackMemory UnixSockets zstd
```

### curl/libcurl version (working one)

```
curl 7.84.0 (x86_64-apple-darwin22.0) libcurl/7.84.0 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.47.0
Release-Date: 2022-06-27
```

### operating system

`macOS Ventura Version 13.0 (22A380)`

### Why is this happening?

From what I can see, the issue originates in `CopyIdentityFromPKCS12File` in sectransp.c:

For iOS, `SecPKCS12Import` is used which automatically adds the certificate to keychain.

For macOS, we do not want the certificate to be added to keychain automatically, so `SecItemImport` is used instead.

Now, if we don't have the private key in the keychain, when we call `SecIdentityCreateWithCertificate` to get the `SecIdentityRef`, it fails because it cannot find the certificate or its private key in the keychain. If either certificate or private key are in the keychain, this works flawlessly.

The dirty workaround is to default to the iOS behaviour and use `SecPKCS12Import` instead, similar to what was done in #4800.

A proper workaround (although not very elegant) would be to create a temporary keychain and use `SecPKCS12Import` to get the `SecIdentityRef`, followed by a a `SecKeychainDelete` to cleanup.

Others seems to have come across this issue outside of cURL (https://developer.apple.com/forums/thread/31711) and the great Quinn The Eskimo seems to have filed it as a bug (r. 25,140,029).

I'm more than happy to provide the p12 certificate privately if needed, just let me know :)

andreiaugustin:
Using Secure Transport, if we are using a path to a p12 certificate which itself or private key is not in the keychain, it will fail with errSecItemNotFound. With this change, we are trying a last-resort solution of creating a temporary keychain, importing the p12 certificate via SecPKCS12Import and successively delete the temporary keychain.

This is to fix #10038.

Please note that with my current build (latest changes at time of writing, commit 3f06b423c4986a5e6c8ce44559d6859991b1e06e), I always seem to be getting "curl: (35) Internal SSL engine error encountered during the SSL handshake" with or without this change. However, to make sure it works, I instead used the latest released version at time of writing (7.86.0) and I can confirm it works OK on my end.

The downsides of this change is that the APIs to create and delete a keychain (SecKeychainCreate and SecKeychainDelete) have been deprecated at macOS 10.10.

The upside is that you won't need to add a p12 cert or its private key to the default keychain if using Secure Transport.

If the deprecated APIs will ever be removed, a more robust solution I can think of would be to use SecPKCS12Import and try removing the certificate that was just added with SecItemDelete, but would need to find a way to target that specific certificate in order to avoid deleting other certificates/private keys.

Let me know if you'd like anything changed in the PR!

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -3056,10 +3056,10 @@ static CURLcode collect_server_cert(struct Curl_cfilter *cf,
 static CURLcode sectransp_connect_step3(struct Curl_cfilter *cf,
                                         struct Curl_easy *data)
 {
+  struct ssl_connect_data *connssl = cf->ctx;
 #if CURL_BUILD_MAC_10_7
   DeleteTemporaryKeychain();
 #endif
-  struct ssl_connect_data *connssl = cf->ctx;
 
   /* There is no step 3!
    * Well, okay, let's collect server certificates, and if verbose mode is on,
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -140,6 +140,10 @@
 #define ioErr -36
 #define paramErr -50
 
+#if CURL_BUILD_MAC_10_7
+#define tempKeychain "CURL_RESERVED_TEMPORARY_KEYCHAIN"
+#endif
+
 struct ssl_backend_data {
   SSLContextRef ssl_ctx;
   bool ssl_direction; /* true if writing, false if reading */
@@ -1153,6 +1157,51 @@ static OSStatus CopyIdentityWithLabel(char *label,
   return status;
 }
 
+#if CURL_BUILD_MAC_10_7
+static SecKeychainRef GetTemporaryKeychain()
+{
+  SecKeychainRef keychain = NULL;
+  OSStatus status = SecKeychainCreate(tempKeychain,
+                                      0,
+                                      "",
+                                      false,
+                                      NULL,
+                                      &keychain);
+  if(status == errSecDuplicateKeychain)
+    status = SecKeychainOpen(tempKeychain, &keychain);
+
+  if(status == errSecSuccess && keychain)
+    return keychain;
+  return NULL;
+}
+
+static void DeleteTemporaryKeychain()
+{
+  SecKeychainRef keychain = GetTemporaryKeychain();
+  if(keychain) {
+    SecKeychainDelete(keychain);
+    CFRelease(keychain);
+  }
+}
+
+static CFTypeRef ExtractIdentityFromArray(CFArrayRef pArray)
+{
+  CFIndex i, count;
+  count = CFArrayGetCount(pArray);
+
+  for(i = 0; i < count; i++) {
+    CFTypeRef item = (CFTypeRef) CFArrayGetValueAtIndex(pArray, i);
+    CFTypeID  itemID = CFGetTypeID(item);
+
+    if(itemID == CFDictionaryGetTypeID()) {
+      return (CFTypeRef) CFDictionaryGetValue(
+                                              (CFDictionaryRef) item,
+                                              kSecImportItemIdentity);
+    }
+  }
+}
+#endif /* CURL_BUILD_MAC_10_7 */
+
 static OSStatus CopyIdentityFromPKCS12File(const char *cPath,
                                            const struct curl_blob *blob,
                                            const char *cPassword,
@@ -1257,6 +1306,50 @@ static OSStatus CopyIdentityFromPKCS12File(const char *cPath,
 #endif
       }
     }
+
+#if CURL_BUILD_MAC_10_7
+    if(status == errSecItemNotFound) {
+      /* Try importing into temporary Keychain */
+      SecKeychainRef curlKeychain = GetTemporaryKeychain();
+      if(curlKeychain) {
+        const void *cKeys[] = {
+          kSecImportExportKeychain,
+          kSecImportExportPassphrase
+        };
+        const void *cValues[] = {
+          curlKeychain,
+          password
+        };
+        CFDictionaryRef options =
+        CFDictionaryCreate(NULL, cKeys, cValues,
+                           password ? 2L : 1L,
+                           NULL, NULL);
+        CFArrayRef tempItems = NULL;
+        if(options) {
+          status = SecPKCS12Import(pkcs_data, options, &tempItems);
+          CFRelease(options);
+        }
+        if(status == errSecSuccess &&
+           tempItems &&
+           CFArrayGetCount(tempItems)) {
+          CFTypeRef identity = ExtractIdentityFromArray(tempItems);
+          if(identity) {
+            CFRetain(identity);
+            *out_cert_and_key = (SecIdentityRef) identity;
+          }
+          else
+            status = errSecItemNotFound;
+        }
+        else
+          status = errSecItemNotFound;
+        if(tempItems)
+          CFRelease(tempItems);
+      }
+      if(curlKeychain)
+        CFRelease(curlKeychain);
+    }
+#endif /* CURL_BUILD_MAC_10_7 */
+
 
     if(items)
       CFRelease(items);
@@ -2503,6 +2596,9 @@ static CURLcode sectransp_connect_step2(struct Curl_cfilter *cf,
   err = SSLHandshake(backend->ssl_ctx);
 
   if(err != noErr) {
+#if CURL_BUILD_MAC_10_7
+    DeleteTemporaryKeychain();
+#endif
     switch(err) {
       case errSSLWouldBlock:  /* they're not done with us yet */
         connssl->connecting_state = backend->ssl_direction ?
@@ -2959,6 +3055,9 @@ static CURLcode collect_server_cert(struct Curl_cfilter *cf,
 static CURLcode sectransp_connect_step3(struct Curl_cfilter *cf,
                                         struct Curl_easy *data)
 {
+#if CURL_BUILD_MAC_10_7
+  DeleteTemporaryKeychain();
+#endif
   struct ssl_connect_data *connssl = cf->ctx;
 
   /* There is no step 3!
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -1199,6 +1199,7 @@ static CFTypeRef ExtractIdentityFromArray(CFArrayRef pArray)
                                               kSecImportItemIdentity);
     }
   }
+  return NULL;
 }
 #endif /* CURL_BUILD_MAC_10_7 */
 


vszakats:
Without trying, this seems prone to race conditions and may also leave the keychain on disk in case of errors. In any case the keychain is NOT password protected on disk, while it's storing the imported key after de-encrypting it. It's also a question how well creating/deleting a separate keychain perform.

I wonder if the alternative `SecPKCS12Import()` solution may work better, together with a unique auto-generated key ID/name.

andreiaugustin:
You are right, it's going to leave the keychain on disk in case of errors or crashes, although it should delete it on the next successful run. Furthermore, there may be other cases in which `DeleteTemporaryKeychain()` needs to be called if it returns because of a failure (e.g. if new steps get added in the future).

`SecPKCS12Import()` can be used to get the identity out of the p12 file and it adds it to the keychain as a side effect (doesn't happen on iOS). Unless we can find a reliable way of building the query part of `SecItemDelete()` to target the specific certificate that was added, we would have to import into the keychain manually and then remove it with a unique auto-generated identifier like you've suggested.

bagder:
1. Creating a file on disk sounds slow
2. What's the ownership/rights on that file?
3. It uses a fixed name so others can easily look for it, even if it exists for a short while
4. Does it save it in "current dir" hoping it can write there? What if it can't? Shouldn't it use a full path to somewhere "safe" ?
5. The deletion seems to be able to happen while another curl is about to use it. Imagine 100 parallel curl invokes.
6. And the API is deprecated

Maybe the answer is simply: don't use Secure Transport?

andreiaugustin:
Indeed it’s not the most elegant workaround as I’ve previously stated, but is the alternative of simply forgetting about Secure Transport the best solution?

What do you think about an extra option, specifically for Secure Transport, to favour using ` SecPKCS12Import` like we do for iOS and leave the certificate in the user’s default keychain?

That way, we leave it up to the user to decide if adding the certificate or private key to their default keychain is a price worth paying for their request succeeding.

bagder:
> is the alternative of simply forgetting about Secure Transport the best solution?

That is certainly *a* solution.

I don't think we can accept this brittle solution that stores and may leave data on disk using a deprecated API. I also don't know if there is any better way to do this with Secure Transport.

andreiaugustin:
The only other option I can think of is for Apple to allow `SecItemImport` importing the private key too so it doesn't go looking for it in a keychain, or alternatively for them to make `SecPKCS12Import` optionally work like it does on iOS where it does not import into keychain.

I'm leaning towards a solution in which we offer a switch option to allow importing the certificate into the user's default keychain (so using `SecPKCS12Import`). It acknowledges that there's no other cleaner solution that can be implemented today, but it makes the workaround optional. That way, we don't have to ask users to use a different TLS backend. Furthermore, the user themselves would be acknowledging they're happy for their default keychain to be polluted.

I imagine anybody coming across the same issue, if not moving to a different TLS backend, would just be manually adding the certificate to their keychain - this would only be automated it if they wish to pass the switch?

Overall, I agree that it feels like it needs to be fixed lower in the stack, ideally by Apple giving better options so we don't compromise curl's speed and security.

vszakats:
My concern is, even if implemented behind an explicit option, apps will start using this and litter user's Keychains anyway, with them powerless to opt-out. Here by 'user' I mean the end-user, who is often a different person than the user of libcurl (who is the developer making this choice on behalf of their end-users). It's also rarely something the end-user can discover at all, or only after the fact.

Now, if a developer really wants to do this, they can (already) do it before calling libcurl, with the exact same effects.

andreiaugustin:
Good point! I agree - no need to provide the option to make the situation worse directly in curl when they can do it themselves.

It seems the best way to workaround this is by switching over to LibreSSL as a TLS backend or, worse, have the libcurl user (developer) add the certificate to the keychain before calling into libcurl.

Another idea could be converting the p12 to pem, but I imagine that's also something better done outside of libcurl.

I think it might be best to close this PR then, thanks a lot for your time and considerations, I appreciate it! :)

PR #10064, opened 2022-12-09
contributors.sh: actually use $CURLWWW instead of just setting it.

gevaerts:
The script was all set up for flexibility where curl-www is in the filesystem, but then hard-coded ../curl-www anyway...


bagder (approved):

bagder:
Thanks!

PR #10065, opened 2022-12-09
contributors.sh: also include people from the curl-fuzzer repository.

gevaerts:
curl-fuzzer is not tagged in the same way, so we can't do the easy thing that works for curl-www. We try to use a timestamp based approach, which will go wrong if the commit date isn't close to the merge/push date, but it's not immediately clear if there's a way to avoid that.

Ideas welcome!

### DIFF
--- a/scripts/contributors.sh
+++ b/scripts/contributors.sh
@@ -45,6 +45,11 @@ if [ -z "$CURLWWW" ] ; then
     CURLWWW=../curl-www
 fi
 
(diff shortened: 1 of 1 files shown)


bagder (approved):
I think it could work good enough.

gevaerts:
I've adjusted the script to use the tag timestamp if possible, falling back to the commit timestamp if needed.

bagder:
Thanks!

Closed via 99f72c01c21 which accidentally used the wrong `closes`!

PR #10071, opened 2022-12-09
workflows/linux.yml: merge 3 common packages

andy5995:
Normally I wouldn't open and close so many pull requests for one thing, but rather squash commits and force-push on the same branch.

Whenever I try to force-push, git-cola is reporting the branch doesn't exist. So I've had to delete it from origin and recreate it! Ugh!


mback2k (approved):

bagder:
Thanks!

Issue #10053, opened 2022-12-07
Problems with connection SSL status


monnerat:
### I did this
`curl --ssl smtps://...`

### I expected the following
Single SSL negotiation.

### curl/libcurl version
current github snapshot

[curl -V output]
```
curl 7.87.0-DEV (x86_64-pc-linux-gnu) libcurl/7.87.0-DEV OpenSSL/3.0.5 zlib/1.2.12 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.4 libpsl/0.21.1 (+libidn2/2.3.3) libssh/0.10.4/openssl/zlib nghttp2/1.51.0 librtmp/2.3 libgsasl/1.10.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli Debug gsasl GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP TrackMemory UnixSockets zstd
```
### operating system
Fedora 37 64-bit

`Linux patrick.monnerat 6.0.11-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Dec 2 20:47:45 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

### Comment
There are problems in `Curl_ssl_cf_get_ssl()`:
1) The first cfilter is skipped, resulting in missing a leading SSL layer.
2) A clear connection over an SSL proxy will be considered as encrypted with the proxy SSL layer.

We currently do not have test support for implicit SSL imap/pop3/smtp, thus I can't write tests for these.

I did not try the command above, but I'm pretty sure it negotiates  SSL twice. I discovered these problems while maintaining the SAFEAUTH PR and by reading the code.

bagder:
/cc @icing 

Does the `--ssl` option change anything? I mean does the bug reproduce without it as well?

monnerat:
> Does the --ssl option change anything? I mean does the bug reproduce without it as well?

Yes: it requires explicit SSL on an implicit SSL connection. Function `Curl_ssl_use()` returns FALSE after implicit negotiation because the SSL layer is the first in chain. Then the protocol retries to negotiate SSL.
Again, I did not try it.

[...]

PR #10077, opened 2022-12-10
tests: support implicit SSL mail protocols

monnerat:
And as bonus, use a hash table in runtests.pl for server ports and introduce 3 new tests for cases exposed in https://github.com/curl/curl/issues/10053.


bagder (approved):

monnerat:
From the current CI test results:
1) I suppose new tests should be disabled for rustls ?
2) I suspect we have a race problem when starting the protocol server just before stunnel. runtests.pl should probably make sure the server is running before starting stunnel. For now, move imaps/pop3s/smtps tests after first corresponding clear protocol tests.

bagder:
> I suppose new tests should be disabled for rustls ?

Why? rustls is supposed to generally work fine.

> I suspect we have a race problem

Hm, yes it is supposed to first verify that the non-stunnel server works before it starts stunnel. See `startservers` for FTPS for example, it first runs `runpingpongserver` (which also verifies that it works) before it runs `runftpsserver` that starts stunnel.

monnerat:
> Why? rustls is supposed to generally work fine.

It has 12 disabled tests that are supposed to work with. That's why I asked.
> Hm, yes it is supposed to first verify that the non-stunnel server works before it starts stunnel. See startservers for FTPS for example, it first runs runpingpongserver (which also verifies that it works) before it runs runftpsserver that starts stunnel.

I've reused the ftps code for the mail protocols (I even renamed `runftpsserver` as `runsecureserver`) thus I expect new protos behave the same. Investigating.

bagder:
> It has 12 disabled tests that are supposed to work with. That's why I asked.

Right, but we use TLS in many more tests than just those. These are just tests that for some reason fail with rustls. It is still expected to work just like the other backends unless there's some anomaly.

tests/data/test987
```diff
@@ -0,0 +1,51 @@
...
+SMTPS with redundant explicit SSL request
+ </name>
+<stdin>
+From: different
+To: another
+
+body
+</stdin>
+ <command>
+-k --ssl-reqd --mail-rcpt recipient@example.com --mail-from sender@example.com -T - smtps://%HOSTIP:%SMTPSPORT/%TESTNUMBER
```
bagder, line 29: I think these three tests should have `--ssl-reqd` removed. This option should only be necessary for transfers that are upgraded to TLS, while this one should use it from the start already (when using the `smtps://` URL scheme) and will fail if it can't do the TLS handshake.

monnerat: > I think these three tests should have --ssl-reqd removed. This option should only be necessary for transfers that are upgraded to TLS, 

No, this is intentional: see the test names. `--ssl-reqd` (or simply `--ssl`) should simply be ignored for implicit TLS protocols. Issue #10053 caused the TLS upgrade even for implicit TLS protocols and the goal of these tests is to check the upgrade is not attempted.

Regarding the CI test failures: I can't reproduce them locally thus it's hard to see what happens in the cloud. Obviouly it's a bad port reuse problem which seems to come from concurrent jobs. The server checks you mention above deal with pids, not ports. I wonder if checking for the port to be open can clear this problem.
Maybe it should also be good to wait for the ports to be closed when a server process is terminated, to avoid `EADDRINUSE` errors. I noticed  that, among random failures in CI jobs, many of them come from `EADDRINUSE`, even outside this PR: if we can resolve this, all CI tests will benefit of it.
I'm currently quite busy for a few days, so don't expect much activity on this PR soon.

bagder:
There are test failures now with NSS and rustls. Are they permafails or flaky?

monnerat:
> There are test failures now with NSS and rustls. Are they permafails or flaky?

They seem to be non-flaky.
I've succeeded in reproducing the NSS failure locally and it shows the transfer never ends (stuck idle until timeout).
I did not try with rustls as it's way too complicated to compile/install locally.
I'm now 80% sure this does not come from the PR changes themselves, but rather are backend-specific problems wich we did not experience before as we were not able to run TLS mail tests. Help would be appreciated.
In all cases this PR is not ready for commit.

bagder:
>  In all cases this PR is not ready for commit.

Maybe make it a "draft" PR for the time being then to make it more visible to observers?

bagder:
> I've succeeded in reproducing the NSS failure locally and it shows the transfer never ends (stuck idle until timeout).

Since NSS is already deprecated and scheduled for deletion in August, it feels like a boring job to try to work on a bugfix now for an issue that probably has existed for a long time already and obviously has not disturbed any users enough to report it to us. It feels attractive to just mark this single test disabled for NSS builds.

> I did not try with rustls as it's way too complicated to compile/install locally.

rustls also already has a range of tests disabled so it wouldn't be too terrible to add a few more to that list. This backend is stilled marked experimental for this reason.

It generally takes someone with a little insight into rustls and its API to work on these issues.

monnerat:
Thanks for your comment.

In the meantime I've found the problem with NSS: the `data_pending` method is not implemented and  `Curl_none_data_pending` is used for it, always telling there is no data ready for input.

When some TLS data is buffered by the backend, there's therefore no chance to see it and wake the handle. The problem is likely to occur with all backends not specifically implementing `data_pending`.

I've managed to make it work with NSS in two ways:
1. By adding a delay in `ftpserver.pl` between the imap data output and postfetch output: this forces the postfetch output to be transmitted in a separate packet and therefore not buffered. This was just a confirmation test and not a problem solution.
2. By effectively implementing `nss_data_pending`:.

rustls does implement `data_pending` but I suspect it does not work as it should.

I then consider this problem as the first one being detected by this PR and already existing before it. Do you want a separate issue report for it ? I think it should be fixed (for all backends) before we commit the current PR to avoid "parasitic" reports in the CI tests.

bagder:
> Do you want a separate issue report for it ?

I think at least a separate PR would make sense, yes!

monnerat:
> I think at least a separate PR would make sense, yes!.

I can produce one for NSS only. However the problem should probably be handled more generally, either for each backend or by supporting (blocking?) data reads from backends that do not implement `data_pending`.

The rustls backend code is out of my skill.

monnerat:
Should be OK now.
New tests successful with NSS after commits ee0f739 and f22cd67 applied.
Still fail with rustls: timeout in hello handshake --> rustls problem; new tests disabled for it.
Gskit should fail too because `data_pending` method is not implemented.
Other backends do not seem to have problems.

bagder:
Thanks! and sorry for taking so long

monnerat:
> Thanks! and sorry for taking so long

Never mind and thanks for pulling.

PR #10088, opened 2022-12-13
lib: use size_t or int etc instead of longs

bagder:
Since long is not using a consistent data size in curl builds.


danielgustafsson (approved):

bagder:
I found more cases in other files, so this PR becomes a more generic "unsigned long" cleanup thing.

bagder:
The fact that the `CURLAUTH_*` defines have `unsigned long` in them is pretty annoying now =)

PR #10105, opened 2022-12-16
socks: fix username max size is 255 (0xFF)

0x0a0d:
https://github.com/curl/curl/blob/50e781d3ecccedbfef323ce468dd281d483b44d0/lib/socks.c#L681-L687


danielgustafsson (approved):
Agreed, this looks like a correct fix. Seems like commit f4b69012307580119409100f5934d7621ea545c9 got the username length check wrong but the password length check right.

bagder (approved):

bagder:
Thanks!

PR #10114, opened 2022-12-19
more HTTP/2 test cases

icing:
Adding 3 new HTTP/2 test cases, plus https: support for nghttpx
- a simple https get
- a simple https post
- a multi get of 4 requests and check that same connection was used


bagder:
Thanks!

PR #10115, opened 2022-12-19
CI: event-based: often get cancled due timeout encrease it to 2h

pheiduck:
Example:
https://github.com/curl/curl/actions/runs/3730372864/jobs/6327357216#step:20:249


pheiduck:
hm, cancled after 2h as well maybe a test will get stuck there?

bagder:
We should check which the last test is before it times out to see if it repeatedly is the same

PR #10136, opened 2022-12-22
Support for Request Payload with msh3

nibanks:
This updates the msh3 HTTP/3 code to support sending request data. A few option issues that can probably be handled in future PRs:

* How do we make sure curl doesn't free the send data before the send is complete?
* How do we know when to FIN the request?
  * Currently, we only will support a single send call from curl and FIN after that.


bagder:
"This branch cannot be rebased due to conflicts"

Can you rebase this against master and force-push?

nibanks:
I merged `master` but it still is complaining?

nibanks:
Ok. Just force pushed. I think I fixed everything.

bagder:
Thanks!

PR #10145, opened 2022-12-22
libssh2: try sha2 algos for hostkey methods

### LINKED ISSUE #10143: SFTP rsa-sha2-256, rsa-sha2-512 key exchange issue
Compiled and linked curl to libssh2 current github master dev branch in order to get rsa-sha2-256 and rsa-sha2-512 server-key-exchange methods.

First SSH server connect works fine, server key gets saved to ssh local keyfile. Subsequent connects failing always because curl offers only "ssh-rsa", not "rsa-sha2-256" and "rsa-sha2-512" anymore, to server.

Could fix the issue with a little change to /lib/vssh/libssh2.c :
In ssh_force_knownhost_key_type()
`static const char * const hostkey_method_ssh_rsa
    = "rsa-sha2-256,rsa-sha2-512,ssh-rsa";`
instead of
`static const char * const hostkey_method_ssh_rsa
    = "ssh-rsa";`

This is the debug output when SFTP/SSH connect fails:
`* Connected to 10.45.120.110 (10.45.120.110) port 22 (#0)`
`* Found host 10.45.120.110 in my_ssh_known_hosts.txt`
`* Set "ssh-rsa" as SSH hostkey type`
`* Failure establishing ssh session: -5, Unable to exchange encryption keys`
`* Closing connection 0`

I think as soon a new libssh2 release is available (1.11.0), someone needs to address this issue in order to get rsa-sha2-256 and rsa-sha2-512. The old ssh-rsa (SHA1) is disabled within OpenSSH since version 8.8, 2021-09-26.

bagder:
As is supported by recent libssh2, but should just be ignored by older versions.

Reported-by: norbertmm on github
Fixes #10143

### DIFF
--- a/lib/vssh/libssh2.c
+++ b/lib/vssh/libssh2.c
@@ -840,6 +840,8 @@ static CURLcode ssh_force_knownhost_key_type(struct Curl_easy *data)
 #endif
   static const char * const hostkey_method_ssh_rsa
     = "ssh-rsa";
+  static const char * const hostkey_method_ssh_rsa_all
+    = "rsa-sha2-256,rsa-sha2-512,ssh-rsa";
   static const char * const hostkey_method_ssh_dss
     = "ssh-dss";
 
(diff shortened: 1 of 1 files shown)


norbertmm:
One test related to the PR is failing.

bagder:
> One test related to the PR is failing.

It seems to just be something flaky, as there are numerous of SFTP tests working for the same build.

norbertmm:
Nevertheless I will test on ubuntu bionic tonight, just to be sure.

norbertmm:
Found the cause for the failing test: There is a bug in libssh2 version <1.9.0. The removal of unsupported methods does not really work as intended within libssh2_session_method_pref(). See https://github.com/libssh2/libssh2/issues/611. The missing 'continue' ...

So a workaround depending on the used libssh2 lib version is necessary. See my PR https://github.com/curl/curl/pull/10154

With this workaround I sucessfully tested on ubuntu bionic like in the failing test suite. Needed to build libssh2 1.8.0 for that testing by myself as the Ubuntu binary package is linked against libgcrypt.so and therefore does not support .ssh/known_hosts file at all.

Issue #10150, opened 2022-12-23
7.87.0: core dump in curl_easy_cleanup


0-wiz-0:
### I did this
Use [lgogdownloader](https://github.com/Sude-/lgogdownloader/) with curl 7.87.0. 
```
lgogdownloader --download --repair --game $SOME_GAME
```
### I expected the following
After updating curl from 7.86.0 to 7.87.0, lgogdownloader started core dumping on exit.

The backtrace is:
```
(gdb) bt
#0  0x000076a23f2c46a0 in ?? ()
#1  0x000076a234df9c43 in bio_read_intern (b=0x76a23f25a800, data=data@entry=0x76a23ee74b43, dlen=dlen@entry=5,
    readbytes=readbytes@entry=0x7f7ffff118f8)
    at /usr/src/crypto/external/bsd/openssl/dist/crypto/bio/bio_lib.c:270
#2  0x000076a234dfa161 in BIO_read (b=<optimized out>, data=data@entry=0x76a23ee74b43, dlen=dlen@entry=5)
    at /usr/src/crypto/external/bsd/openssl/dist/crypto/bio/bio_lib.c:296
#3  0x000076a235225b21 in ssl3_read_n (s=s@entry=0x76a23ef03000, n=n@entry=5, max=5, extend=extend@entry=0,
    clearold=clearold@entry=1, readbytes=readbytes@entry=0x7f7ffff119c8)
    at /usr/src/crypto/external/bsd/openssl/dist/ssl/record/rec_layer_s3.c:296
#4  0x000076a235222c65 in ssl3_get_record (s=s@entry=0x76a23ef03000)
    at /usr/src/crypto/external/bsd/openssl/dist/ssl/record/ssl3_record.c:207
#5  0x000076a23522740a in ssl3_read_bytes (s=<optimized out>, type=23, recvd_type=0x0, buf=0x7f7ffff11bb0 "", len=32,
    peek=0, readbytes=0x7f7ffff11b98)
    at /usr/src/crypto/external/bsd/openssl/dist/ssl/record/rec_layer_s3.c:1301
#6  0x000076a2352615c9 in ssl3_read_internal (s=0x76a23ef03000, buf=0x7f7ffff11bb0, len=32, peek=0,
    readbytes=0x7f7ffff11b98) at /usr/src/crypto/external/bsd/openssl/dist/ssl/s3_lib.c:4465
[...]

PR #10151, opened 2022-12-23
Removing attached easy handles from OpenSSL SSL instances. Refs #10150.

icing:
- keeping the "current" easy handle registered at SSL* is no longer necessary, since the "calling" data object is already stored in the cfilter's context (and used by other SSL backends from there).
- The "detach" of an easy handle that goes out of scope is then avoided.
- This could be the cause of the crash reported in #10150.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -277,8 +277,6 @@
 #endif /* !LIBRESSL_VERSION_NUMBER */
 
 struct ssl_backend_data {
-  struct Curl_easy *logger; /* transfer handle to pass trace logs to, only
-                               using sockindex 0 */
   /* these ones requires specific SSL-types */
   SSL_CTX* ctx;
   SSL*     handle;
@@ -788,9 +786,6 @@ static void bio_cf_free_methods(void)
 #endif
 
 
-static bool ossl_attach_data(struct Curl_cfilter *cf,
-                             struct Curl_easy *data);
-
 /*
  * Number of bytes to read from the random number seed file. This must be
  * a finite value (because some entropy "files" like /dev/urandom have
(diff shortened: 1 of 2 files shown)


lib/vtls/openssl.c
```diff
@@ -3842,10 +3842,15 @@ static CURLcode ossl_connect_step1(struct Curl_cfilter *cf,
...
 
   BIO_set_data(bio, cf);
+#ifdef HAVE_SSL_SET0_WBIO
   BIO_up_ref(bio);
   SSL_set0_rbio(backend->handle, bio);
   SSL_set0_wbio(backend->handle, bio);
-
+#else
+  /* old style, with some weird backward compat handling,
+   * we prefer the more controlled case above. */
```
bagder, line 3851: "old" here seems to be before 1.1.0 (according to https://www.openssl.org/docs/man1.1.1/man3/SSL_set0_rbio.html) - maybe that can be mentioned here as it might be useful at some point

icing: Yes, also libressl added it in some of its versions.

icing: Rewrote the comment.

bagder (approved):

bagder:
Thanks!

lib/vtls/openssl.c
vszakats: Strangely, latest LibreSSL (3.7.0) has `SSL_set0_rbio`, but not `SSL_set0_wbio`. (it has this instead: `BIO * SSL_get_wbio(const SSL *s);`)

icing: That sounds very cautious of them.

lib/vtls/openssl.c
```diff
@@ -2674,7 +2644,7 @@ static void ossl_trace(int direction, int ssl_ver, int content_type,
   connssl = cf->ctx;
   DEBUGASSERT(connssl);
   DEBUGASSERT(connssl->backend);
```
sjadhavar, line 2646: This can be removed I guess.

PR #10154, opened 2022-12-24
workaround for bug in libssh2 version < 1.9.0

norbertmm:
workaround for bug in libssh2 version < 1.9.0


bagder:
Merged into and taken care of in #10145, thanks!

PR #10155, opened 2022-12-24
cmake: fix to find snprintf

yrashk:
I haven't had the time to check other configurations, but on my macOS Ventura 13.1 with XCode 14.2 cmake does not find `snprintf`.

curl continues to build and work just fine, however since HAVE_SNPRINTF variable is cached, other dependencies (for example [json-c](https://github.com/json-c/json-c)) skip the check due to the way `check_symbol_exists` and assume that `snprintf` is not defined.

Solution: ensure stdio.h is checked for definitions

This way other dependencies have a fair shot at checking whether this function is available without being presented with a false negative.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1007,7 +1007,6 @@ check_include_file_concat("utime.h"          HAVE_UTIME_H)
 
 check_include_file_concat("stddef.h"         HAVE_STDDEF_H)
 check_include_file_concat("stdint.h"        HAVE_STDINT_H)
-check_include_file_concat("stdio.h"         HAVE_STDIO_H)
 check_include_file_concat("sys/utsname.h"   HAVE_SYS_UTSNAME_H)
 
 check_type_size(size_t  SIZEOF_SIZE_T)
@@ -1093,7 +1092,7 @@ check_symbol_exists(setrlimit      "${CURL_INCLUDES}" HAVE_SETRLIMIT)
 
 if(NOT MSVC OR (MSVC_VERSION GREATER_EQUAL 1900))
   # earlier MSVC compilers had faulty snprintf implementations
-  check_symbol_exists(snprintf       "${CURL_INCLUDES}" HAVE_SNPRINTF)
+  check_symbol_exists(snprintf       "stdio.h" HAVE_SNPRINTF)
 endif()
 check_function_exists(mach_absolute_time HAVE_MACH_ABSOLUTE_TIME)
 check_symbol_exists(inet_ntop      "${CURL_INCLUDES}" HAVE_INET_NTOP)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1007,6 +1007,7 @@ check_include_file_concat("utime.h"          HAVE_UTIME_H)
 
 check_include_file_concat("stddef.h"         HAVE_STDDEF_H)
 check_include_file_concat("stdint.h"        HAVE_STDINT_H)
+check_include_file_concat("stdio.h"         HAVE_STDIO_H)
 check_include_file_concat("sys/utsname.h"   HAVE_SYS_UTSNAME_H)
 
 check_type_size(size_t  SIZEOF_SIZE_T)


yrashk:
I am also trying to patch CMake to solve the underlying issue I ran into: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8042

vszakats:
I deleted that check in 4d73854462f30948acab12984b611e9e33ee41e6, because it wasn't used in the curl codebase.

It seems unexpected to check for features we do not use locally (for CMake novice at least).

Could this be solved by fixing our `snprintf()` detection? Or to disable caching this to avoid side-effects in other projects?

If we end up checking for `HAVE_STDIO_H`, IMO a comment would be helpful telling the reason.

yrashk:
Of course, the more significant underlying issue is how CMake's `check_symbol_exists` caches defined variables regardless of which include files were checked. This is what I am trying to address with my merge requested I mentioned above. Even if it makes in some capacity, it'll be a long time before most of users will have that version of CMake installed.

I am not sure what you mean by "check for features we do not use locally"

Here's one use of HAVE_SNPRINTF here: https://github.com/curl/curl/blob/1c567f797bce0befce109bceac6cf751331ae43f/lib/mprintf.c#L972-L975

The surface is indeed tiny, but `HAVE_SNPRINTF` being optional, it would just revert to `sprintf` on platforms where `snprintf.h` can only be found in `stdio.h`.  `snprintf`'s man page does indicate that it should come from `stdio.h` (and so does POSIX standard). Not sure why curl would not check `stdio.h` for `snprintf` based on this.

I wonder if in fact curl never/rarely detects snprintf as being available? I just spun up a bare Ubuntu docker image and tried to build curl on it, and here's what I got:

```
-- Looking for snprintf
-- Looking for snprintf - not found
```

This suggests to me that it curl may be NOT using `snprintf` AT ALL right now?


Not sure about the implications of having HAVE_STDIO_H defined you had in mind. Can you elaborate?

bagder:
1. it should not check for the function *in an include file*, it should check if it is present when linking.
2. I may have reintroduced the snprintf() use again (in 935b1bd4544a23a9) after @vszakats removed that check

bagder:
> This suggests that curl may be not using snprintf at all right now.

Sure it does, if you use configure or another build system that sets that define.

yrashk:
> > This suggests that curl may be not using snprintf at all right now.
> 
> Sure it does, if you use configure or another build system that sets that define.

Fair, I meant in the context of using CMake. Apologies for the lack of clarity on this.

yrashk:
> 1. it should not check for the function _in an include file_, it should check if it is present when linking.

This PR does not change the way the presence of this function is checked, it simply gives `snprintf` a chance to be found while employing the same method (`check_symbol_exists`).

vszakats:
@yrashk what I meant is that the patch restores `HAVE_STDIO_H`. That macro isn't used anywhere in curl. I understand (maybe wrongly) that if we're checking for `snprintf`, we'd define `HAVE_SNPRINTF` (or whatever we use in the code to guard its use).

yrashk:
@vszakats there's a check for snprintf:

https://github.com/curl/curl/blob/master/CMakeLists.txt#L1095

But it is failing because stdio.h is not included

vszakats:
Got it. Still looks odd, and maintaining my opinion that a comment documenting this peculiarity would be useful. This could prevent accidentally deleting this check in the future because of no local macro references. If there is a more self-documenting way to force-include `stdio.h` to make the `snprintf` test pass, it'd probably be a better solution. It also makes me wonder how many other necessary checks may I have deleted that was there for a similar hidden side-effect.

yrashk:
@vszakats got it! I've updated the PR to clarify which check requires `stdio.h`. Do you think this is any better?

vszakats:
@yrashk Thank you for your update, it clears up all concerns from my end.

bagder (approved):

bagder:
Thanks!

Issue #10165, opened 2022-12-27
7.87.0 regression: `--haproxy-protocol` fails with `SSL_ERROR_SYSCALL`


Sergio-IME:
It works with 7.86.0
Happens with dafdb20a26d0c890e83dea61a104b75408481ebd from #9855 (89ee5cfb38b22f9ff68c34aa55ca2c242be90826 right before, is working)
The server is apache 2.4.54

### I did this...
```console
curl -vvv --haproxy-protocol -k https://127.0.0.1/
```

### ...and got this
```console
*   Trying 127.0.0.1:443...
* Connected to 127.0.0.1 (127.0.0.1) port 443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* [CONN-0-0][CF-SSL] TLSv1.0 (OUT), TLS header, Certificate Status (22):
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS handshake, Client hello (1):
* OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to 127.0.0.1:443 
* Closing connection 0
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to 127.0.0.1:443
```

### I expected the following
```console
*   Trying 127.0.0.1:443...
* Connected to 127.0.0.1 (127.0.0.1) port 443 (#0)
> PROXY TCP4 127.0.0.1 127.0.0.1 59730 443
* ALPN: offers h2
* ALPN: offers http/1.1
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Finished (20):
[...]

PR #10167, opened 2022-12-27
Fix HAProxy protocol information to be send before TLS handhshake.

icing:
- refs #10165
- reverse order of haproxy and final ssl cfilter
- make haproxy avaiable on PROXY builds, independent of HTTP support as it can be used with any protocol.


bagder (approved):

bagder:
Thanks!

PR #10178, opened 2022-12-29
tiny tweak on tests/ftpserver.pl data timeout

icing:
Lower the normal DATA connect timeout in the server to speed up torture tests.

- tests/ftpserver.pl blocks when expecting a DATA connection from the client.
- the previous 10 seconds were encountered repeatedly in torture tests and let to long waits.
- 2 seconds should still be sufficient for current hw, but CI will show.


bagder (approved):

bagder:
Thanks!

PR #10184, opened 2022-12-30
vtls: use ALPN HTTP/1.0 when HTTP/1.0 is used

bagder:
Previously libcurl would use the HTTP/1.1 ALPN id even when the application specified HTTP/1.0.

Reported-by: William Tang


bagder:
grrrr, closed by df856cb5c9466

PR #10185, opened 2022-12-31
Add additional condition for including stdint.h in lib/http.h

camblomquist:
stdint.h was only included in http.h when ENABLE_QUIC was defined, but symbols from stdint.h are also used when USE_NGHTTP2 is defined. This causes build errors when USE_NGHTTP2 is defined but ENABLE_QUIC is not.

I don't know if such a configuration *should* be possible, but I had a friend encounter it while building media-autobuild_suite.

Fixes m-ab-s/media-autobuild_suite#2353

### DIFF
--- a/lib/http.h
+++ b/lib/http.h
@@ -42,7 +42,7 @@ typedef enum {
 
 #ifndef CURL_DISABLE_HTTP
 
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #10187, opened 2022-12-31
http: Decode transfer encoding first

jbrobst:
The unencoding stack is added to as `Transfer-Encoding` and `Content-Encoding` fields are encountered with no distinction between the two, meaning the stack will be incorrect if, e.g., the message has both fields and a non-chunked `Transfer-Encoding` comes first. This PR fixes this by ordering the stack with transfer encodings first.

Additionally, something I did not change but I think may be a bug is that `MAX_ENCODE_STACK` doesn't limit `writer_stack`'s length, just the number of encodings in a single field line.

### DIFF
--- a/lib/content_encoding.c
+++ b/lib/content_encoding.c
@@ -977,7 +977,8 @@ static const struct content_encoding error_encoding = {
 static struct contenc_writer *
 new_unencoding_writer(struct Curl_easy *data,
                       const struct content_encoding *handler,
-                      struct contenc_writer *downstream)
+                      struct contenc_writer *downstream,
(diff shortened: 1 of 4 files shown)


bagder:
@monnerat: any comments on this?

monnerat:
Yes, this makes sense. Without it we depend on the ordering of headers within the received stream.
The special case of `chunked` encoding still remains though as it is handled externally (always first). I guess it should be taken as a known limitation, unless we move it to the regular unencoding stack, which is a big work.

bagder:
Thanks!

PR #10192, opened 2023-01-01
Update to v0.6 msh3

nibanks:
> **Note** - This PR cannot be completely validated because msh3 usage was broken with previous refactoring. I am looking into seeing what it will take to fix that.

This PR updates usage of msh3 to the latest version (v0.6), updating the API usage as necessary.


lib/vquic/msh3.c
```diff
@@ -569,15 +612,17 @@ CURLcode Curl_cf_msh3_create(struct Curl_cfilter **pcf,
...
-  (void)ai; /* TODO: msh3 resolves itself? */
+  infof(data, "msh3: initializing");
+
   ctx = calloc(sizeof(*ctx), 1);
   if(!ctx) {
     result = CURLE_OUT_OF_MEMORY;
     goto out;
   }
 
+  ctx->ai = ai;
```
nibanks, line 624: @icing why isn't this passed into the `cf_msh3_connect` function, so I don't have to cache it here?

lib/vquic/msh3.c
```diff
@@ -447,8 +493,13 @@ static CURLcode cf_connect_start(struct Curl_cfilter *cf,
 {
   struct cf_msh3_ctx *ctx = cf->ctx;
   bool insecure = !cf->conn->ssl_config.verifypeer;
+  MSH3_ADDR addr = {0};
+  if(ctx->ai) {
+    memcpy(&addr, ctx->ai->ai_addr, ctx->ai->ai_addrlen);
+  }
+  MSH3_SET_PORT(&addr, (uint16_t)cf->conn->remote_port);
```
nibanks, line 500: Is the address info above always specified? Will it already contain the remote port?

icing:
Please have a look at #10204 where msh3 is unbroken again.

PR #10204, opened 2023-01-02
Making MSH3 HTTP/3 implementation run again in its cfilter.

icing:
- test 2500, single GET works
- test 2501, single POST stalls
- test 2502, multiple, sequential GETs each use a new connection since MsH3ConnectionGetState(qconn) no longer reports CONNECTED after one GET.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -79,7 +79,7 @@
 #include "memdebug.h"
 
 
-#define DEBUG_CF 0
+#define DEBUG_CF 1
 
 #if DEBUG_CF
 #define CF_DEBUGF(x) x
@@ -203,9 +203,9 @@ tcpkeepalive(struct Curl_easy *data,
   }
 }
 
-static void assign_sockaddr_ex(struct Curl_sockaddr_ex *dest,
-                               const struct Curl_addrinfo *ai,
-                               int transport)
+void Curl_sock_assign_addr(struct Curl_sockaddr_ex *dest,
+                           const struct Curl_addrinfo *ai,
+                           int transport)
 {
   /*
    * The Curl_sockaddr_ex structure is basically libcurl's external API
@@ -296,7 +296,7 @@ CURLcode Curl_socket_open(struct Curl_easy *data,
     /* if the caller doesn't want info back, use a local temp copy */
     addr = &dummy;
 
-  assign_sockaddr_ex(addr, ai, transport);
+  Curl_sock_assign_addr(addr, ai, transport);
   return socket_open(data, addr, sockfd);
 }
 
@@ -786,6 +786,8 @@ static void cf_socket_close(struct Curl_cfilter *cf, struct Curl_easy *data)
         CF_DEBUGF(infof(data, CFMSG(cf, "cf_socket_close(%d) no longer at "
                   "conn->sock[%d], discarding"), (int)ctx->sock));
       }
+      if(cf->sockindex == FIRSTSOCKET)
+        cf->conn->remote_addr = NULL;
     }
     else {
       /* this is our local socket, we did never publish it */
@@ -797,8 +799,6 @@ static void cf_socket_close(struct Curl_cfilter *cf, struct Curl_easy *data)
     ctx->active = FALSE;
   }
 
-  if(cf->sockindex == FIRSTSOCKET)
-    cf->conn->remote_addr = NULL;
   cf->connected = FALSE;
 }
 
@@ -1297,7 +1297,7 @@ CURLcode Curl_cf_tcp_create(struct Curl_cfilter **pcf,
     goto out;
   }
   ctx->transport = TRNSPRT_TCP;
-  assign_sockaddr_ex(&ctx->addr, ai, ctx->transport);
+  Curl_sock_assign_addr(&ctx->addr, ai, ctx->transport);
   ctx->sock = CURL_SOCKET_BAD;
 
   result = Curl_cf_create(&cf, &cft_tcp, ctx);
@@ -1376,7 +1376,7 @@ CURLcode Curl_cf_udp_create(struct Curl_cfilter **pcf,
     goto out;
   }
   ctx->transport = TRNSPRT_UDP;
-  assign_sockaddr_ex(&ctx->addr, ai, ctx->transport);
+  Curl_sock_assign_addr(&ctx->addr, ai, ctx->transport);
   ctx->sock = CURL_SOCKET_BAD;
 
   result = Curl_cf_create(&cf, &cft_udp, ctx);
@@ -1426,7 +1426,7 @@ CURLcode Curl_cf_unix_create(struct Curl_cfilter **pcf,
     goto out;
   }
   ctx->transport = TRNSPRT_UNIX;
-  assign_sockaddr_ex(&ctx->addr, ai, ctx->transport);
+  Curl_sock_assign_addr(&ctx->addr, ai, ctx->transport);
   ctx->sock = CURL_SOCKET_BAD;
 
   result = Curl_cf_create(&cf, &cft_unix, ctx);
--- a/lib/cf-socket.h
+++ b/lib/cf-socket.h
@@ -99,6 +99,14 @@ void Curl_sndbufset(curl_socket_t sockfd);
 #endif
 
 /**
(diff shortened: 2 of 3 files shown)


lib/vquic/msh3.c
```diff
@@ -110,19 +119,22 @@ static CURLcode msh3_data_setup(struct Curl_cfilter *cf,
   struct HTTP *stream = data->req.p.http;
   (void)cf;
 
-  H3BUGF(infof(data, "msh3_data_setup"));
-  stream->recv_buf = malloc(MSH3_REQ_INIT_BUF_LEN);
+  DEBUGASSERT(stream);
   if(!stream->recv_buf) {
```
nibanks, line 123: Why would it not already be `NULL`?

lib/vquic/msh3.c
```diff
@@ -131,13 +143,18 @@ static bool msh3request_ensure_room(struct HTTP *stream, size_t len)
 {
   uint8_t *new_recv_buf;
   const size_t cur_recv_len = stream->recv_header_len + stream->recv_data_len;
+
   if(cur_recv_len + len > stream->recv_buf_alloc) {
     size_t new_recv_buf_alloc_len = stream->recv_buf_alloc;
     do {
       new_recv_buf_alloc_len <<= 1; /* TODO - handle overflow */
     } while(cur_recv_len + len > new_recv_buf_alloc_len);
+    CF_DEBUGF(fprintf(stderr, "* enlarging buffer to %zu\n",
```
nibanks, line 152: Why are you sending these all to `stderr`?

lib/cf-socket.c
```diff
@@ -79,7 +79,7 @@
 #include "memdebug.h"
 
 
-#define DEBUG_CF 0
+#define DEBUG_CF 1
```
nibanks, line 82: Did you mean to enable this by default?

lib/vquic/msh3.c
```diff
@@ -189,7 +211,8 @@ static void MSH3_CALL msh3_header_received(MSH3_REQUEST *Request,
               (int)Header->ValueLength, Header->Value);
   }
 
-  stream->recv_header_len += total_len - 1; /* don't include null-terminator */
+  stream->recv_header_len += total_len;
```
nibanks, line 214: Why this change?

lib/vquic/msh3.c
```diff
@@ -244,15 +276,17 @@ static void MSH3_CALL msh3_complete(MSH3_REQUEST *Request, void *IfContext,
 
 static void MSH3_CALL msh3_shutdown(MSH3_REQUEST *Request, void *IfContext)
 {
-  struct HTTP *stream = IfContext;
+  struct Curl_easy *data = IfContext;
+  struct HTTP *stream = data->req.p.http;
   (void)Request;
   (void)stream;
```
nibanks, line 282: I wonder if it's not simply better to do:
```
  (void)Request;
  (void)IfContext;
```

lib/vquic/msh3.c
```diff
@@ -446,9 +525,23 @@ static CURLcode cf_connect_start(struct Curl_cfilter *cf,
                                  struct Curl_easy *data)
 {
   struct cf_msh3_ctx *ctx = cf->ctx;
-  bool insecure = !cf->conn->ssl_config.verifypeer;
+  bool verify = !!cf->conn->ssl_config.verifypeer;
+
+  if(verify && (cf->conn->ssl_config.CAfile || cf->conn->ssl_config.CApath)) {
+    /* TODO: find out if there is a way to provide trust anchors to MSH3 */
```
nibanks, line 531: The latest MsQuic does support this. We can wire it up through msh3 eventually.

lib/vquic/msh3.c
```diff
@@ -512,14 +622,43 @@ static CURLcode cf_msh3_connect(struct Curl_cfilter *cf,
...
+  CF_DEBUGF(infof(data, CFMSG(cf, "clearing ctx")));
   if(ctx) {
     if(ctx->qconn)
       MsH3ConnectionClose(ctx->qconn);
     if(ctx->api)
       MsH3ApiClose(ctx->api);
+
+    if(ctx->active) {
+      /* We share our socket at cf->conn->sock[cf->sockindex] when active.
+       * If it is no longer there, someone has stolen (and hopefully
```
nibanks, line 639: "someone has stolen [it]"?? Could you explain this some?

lib/vquic/msh3.c
```diff
@@ -512,14 +622,43 @@ static CURLcode cf_msh3_connect(struct Curl_cfilter *cf,
...
+       * closed it) and we just forget about it.
```
nibanks, line 640:
```suggestion
       * If it is no longer there, someone has stolen (and hopefully
       * closed) it and we just forget about it.
```

lib/vquic/msh3.c
```diff
@@ -529,7 +668,7 @@ static void cf_msh3_close(struct Curl_cfilter *cf, struct Curl_easy *data)
 
   (void)data;
   if(ctx) {
-    cf_msh3_ctx_clear(ctx);
+    cf_msh3_ctx_clear(cf, data);
```
nibanks, line 671: No need to check `if(ctx)` since `cf_msh3_ctx_clear` already handles that. But really, this function should simply be merged with `cf_msh3_ctx_clear`. No need to have separate ones, IMO.

lib/vquic/msh3.c
icing: There is another path via `DATA_SETUP` event that might already initialize this.

lib/vquic/msh3.c
icing: stderr is easier for debugging than stdout and Curl's own `infof()` cannot be called, since this is on the "wrong" thread.

lib/vquic/msh3.c
```diff
@@ -538,11 +677,44 @@ static void cf_msh3_destroy(struct Curl_cfilter *cf, struct Curl_easy *data)
...
 
+static CURLcode cf_msh3_query(struct Curl_cfilter *cf,
+                              struct Curl_easy *data,
+                              int query, int *pres1, void **pres2)
+{
+  struct cf_msh3_ctx *ctx = cf->ctx;
+
+  switch(query) {
+  case CF_QUERY_MAX_CONCURRENT: {
+    /* TODO: we do not have access to this so far, fake it */
```
nibanks, line 693: You're asking how many parallel requests the peer is currently allowing you to open? Why do you need that? msh3/msquic supports queueing requests until the peer is ready, so the app doesn't actually need to know how many instantaneous requests might be allowed (this value can change at any point in time).

lib/cf-socket.c
icing: Leftover from my debugging. Should change that.

lib/vquic/msh3.c
icing: The lengths were off by one in my testing and the header were not correctly parsed.

lib/vquic/msh3.c
nibanks: Could you explain the state machine a little?

lib/vquic/msh3.c
icing: Ideally, the connection filters that create sockets should also be the ones to close them. However, there are code paths on curl that access the `conn->sock[index]` ones and seem to do that. I need to track those cases down.

lib/vquic/msh3.c
nibanks: Ok.

lib/vquic/msh3.c
nibanks: Ok. I guess we need some tests. 😄

lib/vquic/msh3.c
nibanks: That doesn't sound good. If we do expose the socket from MsQuic up to curl and it closes it, I expect really bad things would happen.

lib/vquic/msh3.c
bagder: If the conditions are right, libcurl adds transfers to an existing connection if it can and multiplexes them. Otherwise it can create a new connection for it. For this to work, it needs to know if the new transfer can be put on the existing connection.

nibanks (commented):
LGTM. Once it's in, I'll merge/rebase my PR.

bagder:
Thanks!

PR #10205, opened 2023-01-02
copyright: update all copyright lines and remove year ranges

bagder:
- they are mostly pointless in all major jurisdictions
- many big corporations and projects already don't use them
- saves us from pointless churn
- git keeps history for us
- the year range is kept in COPYING.

checksrc is updated to allow non-year using copyright statements

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -5,7 +5,7 @@
 #                            | (__| |_| |  _ <| |___
 #                             \___|\___/|_| \_\_____|
 #
-# Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
 #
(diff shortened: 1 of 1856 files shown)


MarcelRaad (approved):
I haven't reviewed everything individually, but the idea looks good to me.

kdudka:
I am not sure about the legal implications but it would make backports easier in long term.

bagder:
Several people pointed out [this blog post](https://hynek.me/til/copyright-years/) and highlighted lots of projects from known companies and organizations that already dropped them.

Some said it is fine as long as it is kept in the main COPYING file, so I did that. Not saying it is necessarily a legal requirement, but keeping it in one places seems fine to me.

cmeister2 (approved):
Haven't reviewed in detail but support the concept.

bagder:
Did another pass and cleaned up more. Especially other names and year ranges not following the common style.

bagder:
Welcome to the new curl :grin:

PR #10206, opened 2023-01-02
runtest.pl: add expected fourth return value

scrtt:
Fixes warning in autobild log: "Use of uninitialized value $HTTP2TLSPORT in substitution iterator at /tests/runtests.pl line 3516"

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -1542,7 +1542,7 @@ sub runhttp2server {
 
     # don't retry if the server doesn't work
     if ($doesntrun{$pidfile}) {
-        return (0, 0, 0);
+        return (0, 0, 0, 0);
     }
 
     my $pid = processexists($pidfile);


bagder (approved):

MarcelRaad (approved):
Thanks!

bagder:
Thanks!

MarcelRaad:
This is a regression from commit https://github.com/curl/curl/commit/c7fb341c0eec9adf0bbb1da27622449bf5db45e8. Is there anything we could do to make this visible in the PRs instead of the autobuilds? I remember similar warnings like this in the past.

bagder:
Yeah, they are however "just" warnings so the tests still work fine. A quick googling showed me a trick on how we can "promote" these warnings to errors, which should make them more noticeable. I'll try it in a PR!

PR #10208, opened 2023-01-02
runtests: consider warnings fatal and error on them

bagder:
To help us detect and fix warnings in this script easier and faster.

Ref: #10206

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -75,7 +75,8 @@ BEGIN {
 }
 
 use strict;
-use warnings;
+# Promote all warnings to fatal
+use warnings FATAL => 'all';
 use Cwd;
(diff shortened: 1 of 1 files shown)


MarcelRaad (approved):
Cool! 👍

tests/runtests.pl
```diff
@@ -124,6 +124,9 @@ BEGIN
 require "azure.pm";
 require "appveyor.pm";
 
+# Promote all warnings to fatal
+$SIG{__WARN__} = sub { die @_; };
```
jahir, line 128: shouldn't `use warnings FATAL => 'all';` do the same?

bagder: [This page](https://www.perlmonks.org/?node_id=931025) says it does the same, yes.

PR #10213, opened 2023-01-03
Remove/Replace Curl_addrinfo references at connectdata and filters

icing:
Keep sockaddr local in the socket filters.

- copy `struct Curl_addrinfo` on filter setup into context
- replace `struct Curl_addrinfo *` with `struct Curl_sockaddr_ex *` in connectdata that is set and NULLed by the socket filter
- this means we have no reference to the resolver info in connectdata or its filters

- trigger the CF_CTRL_CONN_INFO_UPDATE event when the complete filter chain reaches connected status
- update easy handle connection information on CF_CTRL_DATA_SETUP event.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -203,28 +203,10 @@ tcpkeepalive(struct Curl_easy *data,
   }
 }
 
-/*
- * Create a socket based on info from 'conn' and 'ai'.
- *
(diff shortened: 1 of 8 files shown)


bagder (approved):

bagder:
Thanks!

PR #10214, opened 2023-01-03
cmake: set SOVERSION also for macOS

rbrich:
https://github.com/curl/curl/pull/10023 added support for SOVERSION in CMake, which is great. But Mac is missing in the listed platforms. It works fine on Mac and it matches the Autotools output, which also sets the soversion.

The installed libs I get after adding Darwin to the list:
```
libcurl.4.8.0.dylib
libcurl.4.dylib -> libcurl.4.8.0.dylib
libcurl.dylib -> libcurl.4.dylib
```

I think this is correct.


bagder (approved):
Excellent!

bagder:
Thanks!

PR #10224, opened 2023-01-04
curl -D headerfile to use append only operations

icing:
Fix headerfile writing in the curl tool to not rely on the first transfer started to be the first to get a response.

All transfers now write the headefile (-D) in append mode, making sure that the order of transfer responses does not lead to overwrites of previous data.


bagder (approved):

bagder:
Thanks!

PR #10225, opened 2023-01-04
nss: implement data_pending method

monnerat:
NSS currently uses the default `Curl_none_data_pending`() method which always returns false, causing TLS buffered input data to be missed.

The current commit implements the `nss_data_pending`() method that properly monitors the presence of available TLS data.

Other backends should be checked (and fixed) for possible similar problems.


bagder:
Thanks!

PR #10232, opened 2023-01-05
CI: increase '-j' when running make

andy5995:


cmeister2:
I'm not averse to the general principle, though I worry that running tests in parallel will cause interesting problems.

bagder:
I double-checked, `make test` works fine to invoke with `-j`, the tests themselves will not run in parallel.

mback2k:
On native shells it should be quite unproblematic, but just to mention it already: Cygwin and msys shells are known to have trouble with make parallelism.

dfandrich:
IIRC Github actions only get 2 CPUs so going beyond that won't help much.

andy5995:
> IIRC Github actions only get 2 CPUs so going beyond that won't help much.

Looks right

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

Except 3 on macOS VMs

vszakats:
In local build tests (many thousands of them) on macOS with an 4-core HT Intel CPU, `nproc` returns 8 (logical cores), yet `-j4` gives better results then `-j8`. Using `nproc` as-is for `-j` seems overoptimistic to me. Maybe taking half of it, or maxing the value at 4 could help. Though in CI's (in our free tiers at least), there is seldom more than 2 cores.

mback2k:
We also tested -j2 on Cirrus CI (also 2 cores) before and it didn't improve the overall speed in any meaning full way. I guess the 2 cores are busy enough with the various processes spawned during an autotools-based build and context switching between them.

andy5995:
I manually checked the difference on a couple workflows. There seemed to be a worthwhile difference.

I'll do that again soon and make a few notes here.

bagder:
@andy5995 do you have any results/measurements on the differences you can add here to help us make a decision about this?

andy5995:
The right column shows the results with the PR applied

The left column shows results from the workflow https://github.com/curl/curl/pull/10349

image


Here, the left column is from https://github.com/curl/curl/pull/10390

image

There's more contrast here because 'j3' is specified in the macos workflow

image

thesamesam:
Does this affect the total build time at all across all jobs? (Do github throttle or something?)

andy5995:
> Does this affect the total build time at all across all jobs? (Do github throttle or something?)
> 

@thesamesam I don't think so. I don't see anything about it mentioned at

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

or

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

Issue #10227, opened 2023-01-04
Regression in SecureTransport


hjmallon:
### I did this

This is reproduceable with curl and libcurl. I originally saw it in libcurl but I bisected it in curl.

```sh
curl https://google.co.uk/

# Sometimes you will get (with 7.87)
curl: (35) Internal SSL engine error encountered during the SSL handshake

# Sometimes you will get the correct response (with 7.87)
...301 Moved...

# However with curl 7.47 it works every time.
```

### I expected the following

It to work every time like 7.47 does.

### curl/libcurl version

```
% ./src/curl -V
curl 7.87.0-DEV (x86_64-apple-darwin22.2.0) libcurl/7.87.0-DEV SecureTransport zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.4 nghttp2/1.51.0 librtmp/2.3
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe UnixSockets zstd
```

### operating system

`Darwin pixmac570 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:08:47 PST 2022; root:xnu-8792.61.2~4/RELEASE_X86_64 x86_64`

### git bisect

I have done a git bisect with

```
autoreconf -fi
./configure --with-secure-transport --without-openssl
make -j10
./src/curl https://google.co.uk/
```

and it says the first bad commit it

```
commit 55807e6c056f27846d70cec70ee6ac3f0e5b3bbe
Author: Stefan Eissing <stefan@eissing.org>
Date:   Fri Nov 25 14:06:43 2022 +0100

    tls: backends use connection filters for IO, enabling HTTPS-proxy
    
     - OpenSSL (and compatible)
     - BearSSL
     - gnutls
     - mbedtls
     - rustls
     - schannel
     - secure-transport
     - wolfSSL (v5.0.0 and newer)
    
     This leaves only the following without HTTPS-proxy support:
     - gskit
     - nss
     - wolfSSL (versions earlier than v5.0.0)
    
    Closes #9962

 CMakeLists.txt       |   5 +-
 configure.ac         |  13 ++-
[...]

PR #10235, opened 2023-01-05
Fixing secure-transport IO for incomplete read/writes. Refs #10227.

icing:
SecureTransport expects result code errSSLWouldBlock when the requested length could not be sent/recieved in full. The previous code returned noErr, which let SecureTransport to believe that the IO had terminated prematurely.

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -862,7 +862,7 @@ static OSStatus bio_cf_in_read(SSLConnectionRef connection,
     }
     nread = 0;
   }
-  else if(nread < *dataLength) {
+  else if((size_t)nread < *dataLength) {
     rtn = errSSLWouldBlock;
   }
   *dataLength = nread;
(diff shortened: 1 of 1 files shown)


hjmallon:
I have just tested this locally and it appears to fix the `curl: (35) Internal SSL engine error encountered during the SSL handshake` issue! Thanks (this is excelled software support for $0).

bagder (approved):

bagder:
Thanks!

wolfv:
We've been seeing this error on macOS x64 with curl 7.87 (logs: https://github.com/mamba-org/mamba/actions/runs/3847932244/jobs/6555036152)

```
    Download error (56) Failure when receiving data from the peer [https://conda.anaconda.org/conda-forge/noarch/repodata.json]
    SSLRead() return error -50
```

Do you think that could also be fixed with this?

icing:
@wolfv that seems very likely, yes.

PR #10236, opened 2023-01-05
runtests: fix detection of TLS backends

monnerat:
Built-in TLS backends are detected at test time by scanning for their names in the version string line returned by the cli tool: as this line may also list the libssh configuration that mentions its own backend, the curl backend may be wrongly determined.

In example, if the version line contains "libssh/0.10.4/openssl/zlib", OpenSSL is detected as a curl-configured backend even if not.

This fix requires the backend names to appear as full words preceded by spacing in the version line to be recognized as curl TLS backends.

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -3127,45 +3127,45 @@ sub checksystem {
                 $has_win32 = 1;
                 $has_mingw = 1 if ($curl =~ /-pc-mingw32/);
             }
-           if ($libcurl =~ /(winssl|schannel)/i) {
+           if ($libcurl =~ /\s(winssl|schannel)\b/i) {
                $has_schannel=1;
                $has_sslpinning=1;
            }
-           elsif ($libcurl =~ /openssl/i) {
+           elsif ($libcurl =~ /\sopenssl\b/i) {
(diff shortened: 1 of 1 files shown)


bagder:
Ah, nice catch!

tests/runtests.pl
```diff
@@ -3127,45 +3127,45 @@ sub checksystem {
...
+           elsif ($libcurl =~ /\srustls-ffi\b/i) {
                $has_rustls=1;
            }
-           elsif ($libcurl =~ /nss/i) {
+           elsif ($libcurl =~ /\snss\b/i) {
                $has_nss=1;
                $has_sslpinning=1;
            }
-           elsif ($libcurl =~ /wolfssl/i) {
+           elsif ($libcurl =~ /\swolfssl<b/i) {
```
MarcelRaad, line 3149: Should this be `\b` instead of `<b`?

monnerat: Yes, this is a typo. Thanks for pulling my attention.

monnerat: Fixed now.

bagder:
Thanks!

Issue #10163, opened 2022-12-26
Broken code in lib/vtls/x509asn1.c Curl_verifyhost


piru:
### Description

`Curl_verifyhost` has some broken code due to recent code shuffling. The issue is visible here:
https://github.com/curl/curl/blob/curl-7_87_0/lib/vtls/x509asn1.c#L1355

The code doesn't compile and it also leaks some memory for empty ("") strings if parenthesis are added. The issue was introduced by commit https://github.com/curl/curl/commit/6b19247e794cfdf4ec63c5880d8f4f5485f653ab

The easiest solution is to revert the first section of the patch. Note how `utf8asn1str` always sets the pointer to NULL on error.

### curl/libcurl version

curl-7_87_0 tag

danielgustafsson:
> `Curl_verifyhost` has some broken code due to recent code shuffling. The issue is visible here: https://github.com/curl/curl/blob/curl-7_87_0/lib/vtls/x509asn1.c#L1355

Meh, I was twiddling with codeformattting and ended up with the wrong patch =( This codepath only compiles and runs for  GSKit which makes it quite rare. Thanks for spotting it.

> The code doesn't compile and it also leaks some memory for empty ("") strings if parenthesis are added. The issue was introduced by commit [6b19247](https://github.com/curl/curl/commit/6b19247e794cfdf4ec63c5880d8f4f5485f653ab)

Is there valid input which can produce an empty string be returned from `utf8asn1str()` here? We could add a check for zero-length hostnames but I'm not convinced it's a case that will happen.

> The easiest solution is to revert the first section of the patch. 

I think the easiest solution is to fix the issue by apply the following:
```
diff --git a/lib/vtls/x509asn1.c b/lib/vtls/x509asn1.c
index 4c1c9a8b7..2ab07eca7 100644
--- a/lib/vtls/x509asn1.c
+++ b/lib/vtls/x509asn1.c
@@ -1352,7 +1352,7 @@ CURLcode Curl_verifyhost(struct Curl_cfilter *cf,
           len = utf8asn1str(&dnsname, CURL_ASN1_IA5_STRING,
                             name.beg, name.end);
           if(len > 0) {
-            if(size_t)len == strlen(dnsname)
+            if((size_t)len == strlen(dnsname))
               matched = Curl_cert_hostcheck(dnsname, (size_t)len,
                                             connssl->hostname, hostlen);
             free(dnsname);
```
[...]

PR #10238, opened 2023-01-05
Fix compile errors and warning in x509asn1.c when built for GSKit

jonrumsey:
Correct a few small compilation errors and warnings as detailed in https://github.com/curl/curl/issues/10163


bagder (approved):

bagder:
Thanks!

PR #10241, opened 2023-01-05
GHA: add a quiche CI job

bagder:


pheiduck:
> Can't figure out why the quiche build fails... 😢

Because of this?:
```
error: could not find native static library `ssl`, perhaps an -L flag is missing?
```

bagder:
That's the error, not the reason.

I build boringssl first then point out that build to the quiche build to use (with `QUICHE_BSSL_PATH`), but I can't seem to make it work. I build quiche like this locally since years.

bagder:
@LPardue @ghedo can you spot any obvious mistake here?

Issue #10231, opened 2023-01-04
Windows: Don't crash if CURLOPT_READ/WRITEFUNCTION is NULL


DanielGibson:
### I did this

When not setting CURLOPT_WRITEFUNCTION, but setting CURLOPT_WRITEDATA, curl assumes that the write-data is a `FILE *` and that it should use fwrite() as write-function.
As documented on https://curl.se/libcurl/c/CURLOPT_WRITEDATA.html, this will crash on Windows (as far as I understand, this happens when curl.dll and the application that links it are built with different compilers or even just different versions of Visual Studio).

(And it's the same for CURLOPT_READFUNCTION etc)

### I expected the following

I think it would be preferable if instead of crashing, libcurl returned an error (when calling `curl_easy_perform()`, or possible alternatives that invoke the same behavior) if CURLOPT_WRITEDATA is set but CURLOPT_WRITEFUNCTION is not (on Windows).

This way the mistake should (hopefully!) be found when integrating curl (or when first testing the application on Windows), instead of possibly a lot later when updating curl.dll and by chance using one built with a different compiler (version).
This would also give *users* a better chance of replacing curl.dll with a newer one (that fixes know vulnerabilities) when the application they're using isn't updated anymore (or the application developer doesn't update curl.dll).

*(again, same for the read case)*

### curl/libcurl version

Probably all that assume that CURLOPT_WRITEDATA (or CURLOPT_READDATA) sets a *FILE *` if CURLOPT_WRITEFUNCTION (CURLOPT_READFUNCTION) isn't set (or is set to NULL).

### operating system

Windows

maraakate:
Hi, I work with Daniel on some projects and I've been bit by this bug a couple of years ago and forgot about it.  Then had someone come to me for help with getting strange crashes when they changed their implementation from static to dynamic with curl.  Spent a full day on this and then it finally "clicked" that the temp files being written out were 0 sized and noticed writefunction was null.

I agree with Daniel, an assert or something besides letting it crash would be more helpful.  I was getting strange crashes on false from ucrt that led me down false rabbit holes.  Even ran the code through ASAN assuming there was some kind of heap corruption.

jay:
It is technically possible for normal function if just WRITEDATA is set on Windows even though the documentation says otherwise. The problem is doing that will lead to crashes if the calling code's C runtime is not the same as libcurl's C runtime, which sounds like what happened to you. We have really made it as clear as possible to not do it, and I think that's enough.

[...]

PR #10243, opened 2023-01-05
pretransfer: fail on Windows if WRITEDATA set without WRITEFUNCTION

jay:
.. and also do the same for set HEADERDATA without a set HEADERFUNCTION or WRITEFUNCTION; and READDATA without a set READFUNCTION.

Prior to this change if libcurl was running from a DLL and the user set the DATA option without also setting an applicable FUNCTION option then libcurl could crash writing to a FILE stream that was created by a different C runtime. This is because in Windows each DLL that is part of a program may or may not have its own C runtime.

We already warn about this in the WRITEDATA doc but apparently that is not enough.

Reported-by: Daniel Gibson

Fixes https://github.com/curl/curl/issues/10231
Closes #xxxx

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -1314,6 +1314,20 @@ static CURLcode check_for_option_conflicts(struct Curl_easy *data)
 #endif
 
     if(dll) {
+      if(data->set.is_ioctl_func_from_user &&
+         (!data->set.is_in_set_from_user ||
+          !data->set.is_fread_set_from_user)) {
+        failf(data, "Windows DLL users must set CURLOPT_READDATA and "
+                    "CURLOPT_READFUNCTION if CURLOPT_IOCTLFUNCTION is set.");
+        return CURLE_BAD_FUNCTION_ARGUMENT;
+      }
+      if(data->set.is_seek_func_from_user &&
+         (!data->set.is_in_set_from_user ||
+          !data->set.is_fread_set_from_user)) {
+        failf(data, "Windows DLL users must set CURLOPT_READDATA and "
+                    "CURLOPT_READFUNCTION if CURLOPT_SEEKFUNCTION is set.");
+        return CURLE_BAD_FUNCTION_ARGUMENT;
+      }
       if(data->set.is_in_set_from_user &&
          !data->set.is_fread_set_from_user) {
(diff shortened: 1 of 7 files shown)


dfandrich:
Something probably also needs to be done about CURLOPT_IOCTLFUNCTION and fseek.

jay:
> Something probably also needs to be done about CURLOPT_IOCTLFUNCTION and fseek.

ok see latest commit

MarcelRaad:
Wow, that's a lot of code just for users not reading the docs.

lib/urldata.h
```diff
@@ -1805,7 +1805,22 @@ struct UserDefined {
...
+  /* has WRITEFUNCTION been set to non-NULL by user? */
+  BIT(is_fwrite_func_from_user);
+  /* has HEADERFUNCTION been set to non-NULL by user? */
+  BIT(is_fwrite_header_from_user);
+  /* has READDATA been set by user? */
+  BIT(is_in_set_from_user);
+  /* has WRITEDATA been set by user? */
+  BIT(is_out_from_user);
+  /* has HEADERDATA been set by user? */
+  BIT(is_writeheader_from_user);
```
bagder, line 1823: These should be `#ifdef WIN32`, right?

lib/transfer.c
```diff
@@ -1286,6 +1287,75 @@ void Curl_init_CONNECT(struct Curl_easy *data)
...
+         !data->set.is_fwrite_func_from_user &&
+         !data->set.is_fwrite_header_from_user) {
+        failf(data, "Windows DLL users must set CURLOPT_HEADERFUNCTION or "
+                    "CURLOPT_WRITEFUNCTION if CURLOPT_HEADERDATA is set.");
+        return CURLE_BAD_FUNCTION_ARGUMENT;
+      }
+    }
+  }
+
+  return CURLE_OK;
```
bagder, line 1353: I think you could simplify this logic and still cover 99% of the mistakes: If the `*DATA*` argument is not set, but the `*FUNCTION*` is, that's the main problematic case. You can do that without introducing any `is_*_from_user` bits.

bagder:
Can we close this? I don't see a lot of love for it.

bagder:
No love. Leave as is.

PR #10248, opened 2023-01-06
QUIC connect fix, improved error/debug messages

icing:
Improve connect error message, debugging info, fix quic connect reporting false success.

- ECONNECTREFUSED has not its own fail message in quic filters
- Debug logging in connect eyballing improved
- Fix bug in ngtcp2/quiche that could lead to false success reporting.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1555,10 +1555,14 @@ CURLcode Curl_cf_socket_peek(struct Curl_cfilter *cf,
   if(Curl_cf_is_socket(cf) && cf->ctx) {
     struct cf_socket_ctx *ctx = cf->ctx;
 
-    *psock = ctx->sock;
-    *paddr = &ctx->addr;
-    *premote_ip_str = ctx->r_ip;
-    *premote_port = ctx->r_port;
+    if(psock)
+      *psock = ctx->sock;
+    if(paddr)
+      *paddr = &ctx->addr;
+    if(premote_ip_str)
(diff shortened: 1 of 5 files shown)


bagder:
Seems to fail [this quiche build](https://object-storage.public.mtl1.vexxhost.net/swift/v1/bfd521072e894ebb99e66f72619daa8a/zuul-logs-curl_498/10248/ffbc6ff4939913e993cf97103819368713094932/check/curl-cmake-boringssl-quiche/4981179/job-output.txt).

bagder:
Sorry, the quiche problem is in #10260 

This PR fixes the crash for me, but it does show another, slightly less serious issue:

    curl --http3 https://curl.se:5000 -v

Ie connect to a port where nothing is listening and it seems to me that the IPv6 attempts fail fairly fast and then the IPv4 connect attempts look really odd and it takes a long time retrying.

icing:
> Sorry, the quiche problem is in #10260
> 
> This PR fixes the crash for me, but it does show another, slightly less serious issue:
> 
> ```
> curl --http3 https://curl.se:5000 -v
> ```
> 
> Ie connect to a port where nothing is listening and it seems to me that the IPv6 attempts fail fairly fast and then the IPv4 connect attempts look really odd and it takes a long time retrying.

I am seeing this as well. I was debug tracing it and the ipv4 socket never gave an error, while the ipv6 pretty fast gave a ECONNREFUSED. Is this some UDP craziness?

bagder:
Thanks!

PR #10251, opened 2023-01-06
GHA: rename wolfssl and torture workflows

andy5995:


andy5995:
The practical intent of this is to make it easier to view workflow information from the Actions tab. Right now, there are far too many "Linux" workflows:

image

mback2k (approved):
Okay for now, although further CI file merges are still pending my side.

bagder:
I don't quite see the point.

|before|now|
|----|----|
| **Linux/wolfssl** | **wolfssl/wolfssl** |
| **Linux/torture** | **torture/torture** |

This change just removes an information bit and instead add a (pointless) repetition in the name. Also, there is an existing separate **macos/torture** job that goes well with the **Linux/** version as it hints what differences there are.

bagder:
I think this makes the output worse, as explained above.

PR #10252, opened 2023-01-06
GHA: ignore pushes when only other yml files are changed

andy5995:
The first time I tried this... I'm not sure if I got the syntax correct yet. I'll finish it up after a little feedback

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,13 +9,19 @@ on:
     branches:
     - master
     - '*/ci'
-    paths-ignore:
-    - '**/*.md'
+    paths:
+    - '**'
+    - '!**/*.yml'
+    - '!**/*.md'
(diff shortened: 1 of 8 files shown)


andy5995:
The intent of this is to automatically prevent a lot of tests from running. For example, if 'linux.yml' is the only file changed, only that workflow will run. Right now, most all workflows will run.

cmeister2:
It's not obvious to me that having a negating rule in `paths-ignore` is valid syntax; https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore seems to indicate that the `!` character only operates in the `paths` key.

andy5995:
@cmeister2 Thanks for pointing that out!

andy5995:
rebased

andy5995:
Apparently https://github.com/curl/curl/pull/10742 was merged instead, so closing this PR.

PR #10255, opened 2023-01-07
setopt: move the SHA256 opt within #ifdef libssh2

bagder:
Because only the libssh2 backend supports it and thus this should return error if this option is used with
other backends.

Reported-by: Harry Sintonen


bagder:
I've asked @tatsuhiro-t about the new nghttpx build fails. It doesn't seem to build with h3 support atm.

PR #10257, opened 2023-01-08
Designate ngtcp2 and its dependency versions

tatsuhiro-t:
Designate ngtcp2 and its dependency versions so that the CI build does not fail without our control.


bagder (approved):
:+1: will make us less sensitive for changes upstream

mback2k (approved):
Makes sense 👍

bagder:
Thanks!

PR #10264, opened 2023-01-09
add options for http3-only transfer

bagder:
... and make the previous http3 become become the one that allows a fallback to a lower HTTP version if the HTTP/3 connection fails.

**Reminder: HTTP/3 support is experimental. We reserve the right to do changes like this.**


docs/cmdline-opts/http3.d
```diff
@@ -13,10 +13,15 @@ Example: --http3 $URL
 Multi: mutex
 Experimental: yes
 ---
-Tells curl to use HTTP version 3 directly to the host and port number used in
-the URL. A normal HTTP/3 transaction will be done to a host and then get
-redirected via Alt-Svc, but this option allows a user to circumvent that when
-you know that the target speaks HTTP/3 on the given host and port.
+Tells curl to try HTTP/3 to the host used in the URL, but fallback to use
+earlier HTTP versions if the HTTP/3 connection establishement fails. HTTP/3
+can only be used for HTTPS and not for HTTP URLs.
```
icing, line 18: Will it be silently ignored then for `http:`?

docs/libcurl/opts/CURLOPT_HTTP_VERSION.3
```diff
@@ -66,13 +66,12 @@ prior knowledge that the server supports HTTP/2 straight away. HTTPS requests
...
-downgrade to earlier HTTP version if the server does not support HTTP/3.
-
-For more reliably upgrading to HTTP/3, set the preferred version to something
-lower and let the server announce its HTTP/3 support via Alt-Svc:. See
-\fICURLOPT_ALTSVC(3)\fP.
+(Added in 7.66.0) This option makes libcurl attempt to use HTTP/3 to the host
+given in the URL, with fallback to earlier HTTP versions if needed.
+.IP CURL_HTTP_VERSION_3ONLY
+(Added in 7.88.0) Setting this value will make libcurl attempt to use HTTP/3
+directly to server given in the URL but will not gracefully downgrade to
```
icing, line 73: remove "gracefully".

icing (changes requested):

docs/cmdline-opts/http3.d
bagder: That's a good question. I believe it right now returns an error if asked to do 3 with HTTP... I think we can go with that for now.

PR #10265, opened 2023-01-10
autotools: use AC_SEARCH_LIBS instead of AC_CHECK_LIB

andy5995:
Maybe this would help with https://github.com/curl/curl/issues/10127 ?

https://www.gnu.org/software/autoconf/manual/autoconf-2.71/html_node/Libraries.html#index-AC_005fCHECK_005fLIB-1

>AC_CHECK_LIB requires some care in usage, and should be avoided in some
common cases. Many standard functions like gethostbyname appear in the standard C library on some hosts, and in special libraries like nsl on other hosts. On some hosts the special libraries contain variant implementations that you may not want to use. These days it is normally better to use AC_SEARCH_LIBS([gethostbyname], [nsl]) instead of AC_CHECK_LIB([nsl], [gethostbyname]).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1054,7 +1054,7 @@ dnl gethostbyname without lib or in the nsl lib?
 AC_CHECK_FUNC(gethostbyname,
               [HAVE_GETHOSTBYNAME="1"
               ],
-              [ AC_CHECK_LIB(nsl, gethostbyname,
+              [ AC_SEARCH_LIBS([gethostbyname], [nsl],
                              [HAVE_GETHOSTBYNAME="1"
                              LIBS="-lnsl $LIBS"
                              ])


bagder:
The way I read it, is that this macro is primarily good for looking for a function through multiple libs. In this case, it only looks for it in a single lib and then these two macros are doing virtually the same things.

andy5995:
> The way I read it, is that this macro is primarily good for looking for a function through multiple libs. In this case, it only looks for it in a single lib and then these two macros are doing virtually the same things.

A bit more explanation quoted here about AC_SEARCH_LIBS:

>Macro: AC_SEARCH_LIBS (function, search-libs, [action-if-found], [action-if-not-found], [other-libraries])

>Search for a library defining function if it’s not already available. This equates to calling ‘AC_LINK_IFELSE([AC_LANG_CALL([], [function])])’ first with no libraries, then for each library listed in search-libs. 

I see a *little* more difference, though I'm not convinced that would help with #10127 .

bagder:
The issue was closed, and I don't think this change would improve much for this case.

PR #10271, opened 2023-01-10
improved debug logging for cfilters

icing:
New curl_log.[ch] to keep failf/infof and debug logging implementations.

- new functions and macros for cfilter debugging
- set CURL_DEBUG with names of cfilters where debug logging should be enabled
- use GNUC __attribute__ to enable printf format checks during compile

Usage on a debug build:

```
> CURL_DEBUG=happy-eyeballs,ssl curl -v https://curl.se 
* STATE: INIT => CONNECT handle 0x7f9a0c80ae08; line 1908 (connection #-5000)
* STATE: CONNECT => RESOLVING handle 0x7f9a0c80ae08; line 1954 (connection #0)
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* STATE: RESOLVING => CONNECTING handle 0x7f9a0c80ae08; line 2028 (connection #0)
* [CONN-0-HAPPY-EYEBALLS] ipv6 starting
*   Trying [2a04:4e42:a00::347]:443...
* [CONN-0-HAPPY-EYEBALLS] ipv6 connect -> 0, connected=0
* [CONN-0-HAPPY-EYEBALLS] ipv6 connect -> 0, connected=1
...
* [CONN-0-SSL] bio_cf_out_write(len=517) -> 517, err=0
...
```

Tried to make the macros performant and work for all compilers. Let's see what CI says.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -767,20 +767,20 @@ static void cf_socket_close(struct Curl_cfilter *cf, struct Curl_easy *data)
        * closed it) and we just forget about it.
        */
       if(ctx->sock == cf->conn->sock[cf->sockindex]) {
-        LOG_CF_DEBUG(data, cf, "cf_socket_close(%d) active", (int)ctx->sock);
+        DEBUGF(LOG_CF(data, cf, "cf_socket_close(%d) active", (int)ctx->sock));
         socket_close(data, cf->conn, !ctx->accepted, ctx->sock);
         cf->conn->sock[cf->sockindex] = CURL_SOCKET_BAD;
       }
       else {
-        LOG_CF_DEBUG(data, cf, "cf_socket_close(%d) no longer at "
-                     "conn->sock[], discarding", (int)ctx->sock);
+        DEBUGF(LOG_CF(data, cf, "cf_socket_close(%d) no longer at "
+                      "conn->sock[], discarding", (int)ctx->sock));
       }
       if(cf->sockindex == FIRSTSOCKET)
         cf->conn->remote_addr = NULL;
     }
     else {
       /* this is our local socket, we did never publish it */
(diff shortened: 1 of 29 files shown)


lib/cf-socket.c
```diff
@@ -1153,8 +1142,8 @@ static ssize_t cf_socket_recv(struct Curl_cfilter *cf, struct Curl_easy *data,
 
   DEBUGASSERT(data->conn == cf->conn);
   nread = Curl_recv_plain(data, cf->sockindex, buf, len, err);
-  CF_DEBUGF(infof(data, CFMSG(cf, "recv(len=%zu) -> %d, err=%d"),
-            len, (int)nread, *err));
+  DEBUGF(LOG_CF(data, cf, "recv(len=%zu) -> %d, err=%d", len, (int)nread,
+                *err));
```
jay, line 1146: I appreciate why this may be necessary temporarily, and useful, but even for me this is sometimes too verbose. It looks like this PR is going to change CF logging so that it cannot be disabled on DEBUGBUILDs any longer.

icing: Theses statements will not produce output, unless you enable the logging for a certain cfilter type.

```
CURL_DEBUG=tcp curl -v url
```

will enable the logging of the 'TCP' socket cfilter. 

The idea here is to have logging available, should you wish, without recompiling the debug-enabled curl.

jay: Neat. We will need documentation explaining how it works.

lib/http2.c
```diff
@@ -883,8 +882,7 @@ static int on_frame_recv(nghttp2_session *session, const nghttp2_frame *frame,
       Curl_expire(data_s, 0, EXPIRE_RUN_NOW);
```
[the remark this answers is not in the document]
icing: @sonatype-lift ignore

bagder:
Thanks!

PR #10272, opened 2023-01-10
Treat CMAKE_REQUIRED_DEFINITIONS as a list

marc-groundctl:
According to [the documentation](https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html) `CMAKE_REQUIRED_DEFINITIONS` is a list, so add to it using `list(APPEND` instead of string concatenation.


jay:
Thanks

PR #10275, opened 2023-01-11
easyoptions: Fix header printing in generation script

danielgustafsson:
The optiontable.pl script prints the header comment when generating easyoptions.c,  but it wasn't escaping all characters which jumbled the curl ascii logo.  Fix by escaping.


bagder (approved):

bagder:
Closed via 4579626c273fe68 (but with typo)

Issue #10269, opened 2023-01-10
assert failures/crashes from h2_session_send 


tinypancake:
### I did this
Http pings, always crashes on DEBUGASSERT in h2_session_send for stream->stream_id != -1

### Backtrace:

```
(230110_16:38:35.499)AlexaApp: ../../curl-7.82.0/lib/http2.c:1605: h2_session_send: Assertion `stream->stream_id != -1' failed.
(230110_16:38:35.580)SIGABRT receiverd 
(230110_16:38:35.580)=============================backtrace=============================
(230110_16:38:35.580)obtained 18 stack frames:
(230110_16:38:35.580)backtrace:
(230110_16:38:35.580)[0] /usr/bin/AlexaApp() [0x12568] 
(230110_16:38:35.580)[1] /usr/bin/AlexaApp(sighander2+0x38) [0x12624] 
(230110_16:38:35.580)[2] /lib/libc.so.6(__default_sa_restorer+0) [0xb6ae7db0] 
(230110_16:38:35.580)[3] /lib/libc.so.6(+0x7a110) [0xb6b2a110] 
(230110_16:38:35.580)[4] /lib/libc.so.6(raise+0x14) [0xb6ae6e40] 
(230110_16:38:35.580)[5] /lib/libc.so.6(abort+0xf0) [0xb6ad15b8] 
(230110_16:38:35.580)[6] /lib/libc.so.6(+0x302d4) [0xb6ae02d4] 
(230110_16:38:35.580)[7] /lib/libc.so.6(__assert_fail+0x44) [0xb6ae0378] 
(230110_16:38:35.580)[8] /usr/lib/libcurl.so.4(+0x23528) [0xb4703528] 
(230110_16:38:35.580)[9] /usr/lib/libcurl.so.4(+0x23594) [0xb4703594] 
(230110_16:38:35.580)[10] /usr/lib/libcurl.so.4(+0x250d8) [0xb47050d8] 
(230110_16:38:35.580)[11] /usr/lib/libcurl.so.4(+0x20f58) [0xb4700f58] 
(230110_16:38:35.580)[12] /usr/lib/libcurl.so.4(+0x2f2ec) [0xb470f2ec] 
(230110_16:38:35.580)[13] /usr/lib/libcurl.so.4(curl_multi_perform+0x6c) [0xb4710294] 
[...]

PR #10278, opened 2023-01-11
output warning for debug-enabled version

bagder:
A debug-enabled version of curl is not intended for production

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1861,6 +1861,15 @@ static CURLMcode multi_runsingle(struct Curl_multi *multi,
     multistate(data, MSTATE_COMPLETED);
   }
 
+#ifdef DEBUGBUILD
+  if(!multi->warned) {
+    infof(data, "!!! WARNING !!!");
+    infof(data, "This is a debug build of libcurl, "
+          "do not use in production.");
+    multi->warned = true;
+  }
+#endif
+
   do {
     /* A "stream" here is a logical stream if the protocol can handle that
        (HTTP/2), or the full connection for older protocols */
--- a/lib/multihandle.h
+++ b/lib/multihandle.h
@@ -170,6 +170,9 @@ struct Curl_multi {
 #endif
   BIT(dead); /* a callback returned error, everything needs to crash and
                 burn */
+#ifdef DEBUGBUILD
+  BIT(warned);                 /* true after user warned of DEBUGBUILD */
+#endif
 };
 
 #endif /* HEADER_CURL_MULTIHANDLE_H */
--- a/src/tool_help.c
+++ b/src/tool_help.c
@@ -164,8 +164,8 @@ void tool_version_info(void)
 {
   const char *const *builtin;
   if(is_debug())
-    printf("WARNING: this libcurl is Debug-enabled, "
-           "do not use in production\n\n");
+    fprintf(stderr, "WARNING: this libcurl is Debug-enabled, "
+            "do not use in production\n\n");
 
   printf(CURL_ID "%s\n", curl_version());
 #ifdef CURL_PATCHSTAMP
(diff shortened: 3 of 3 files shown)


jay:
Ref: https://curl.se/mail/lib-2023-01/0039.html

I think it's too much but it looks like the list disagrees. Do you have a reference to the issue that was reported with the unintentional use of debugbuild? What about put it in curl_version, all caps, DEBUGBUILD. That way it is also visible from libcurl. Nevermind that DEBUGBUILD can override the version string if the user use CURL_VERSION env...

~~~
curl 7.87.1-DEV (i386-pc-win32) libcurl/7.87.1-DEV DEBUGBUILD OpenSSL/1.1.1s nghttp2/1.50.0
Release-Date: [unreleased]
Features: alt-svc AsynchDNS Debug HSTS HTTP2 HTTPS-proxy IPv6 Largefile NTLM SSL Unicode UnixSockets
~~~

~~~diff
diff --git a/lib/version.c b/lib/version.c
index a69e2ca..dae8335 100644
--- a/lib/version.c
+++ b/lib/version.c
@@ -109,7 +109,7 @@ static void zstd_version(char *buf, size_t bufsz)
  * zeros in the data.
  */
 
-#define VERSION_PARTS 16 /* number of substrings we can concatenate */
+#define VERSION_PARTS 17 /* number of substrings we can concatenate */
 
 char *curl_version(void)
 {
@@ -173,6 +173,9 @@ char *curl_version(void)
 #endif
 
   src[i++] = LIBCURL_NAME "/" LIBCURL_VERSION;
+#ifdef DEBUGBUILD
+  src[i++] = "DEBUGBUILD";
+#endif
 #ifdef USE_SSL
   Curl_ssl_version(ssl_version, sizeof(ssl_version));
   src[i++] = ssl_version;
~~~

bagder:
The debug-by-mistake was in here #10269

I don't think putting **DEBUGBUILD** in there is much different to what we do already do today. It doesn't stress or emphasize how unwise it is to use such build in production.

jay:
It looks as though that reporter is using libcurl but not the curl tool. What about in multi,

~~~
> curld  -v https://curl.se -o NUL
* !!! WARNING !!!
* This is a debug build of libcurl, do not use in production.
* STATE: INIT => CONNECT handle 0x1d37ae8; line 1917 (connection #-5000)
~~~

~~~diff
diff --git a/lib/multi.c b/lib/multi.c
index b6187cd..4032748 100644
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1861,6 +1861,15 @@ static CURLMcode multi_runsingle(struct Curl_multi *multi,
     multistate(data, MSTATE_COMPLETED);
   }
 
+#ifdef DEBUGBUILD
+  if(!multi->warned) {
+    infof(data, "!!! WARNING !!!");
+    infof(data, "This is a debug build of libcurl, "
+          "do not use in production.");
+    multi->warned = true;
+  }
+#endif
+
   do {
     /* A "stream" here is a logical stream if the protocol can handle that
        (HTTP/2), or the full connection for older protocols */
diff --git a/lib/multihandle.h b/lib/multihandle.h
index 7dd6a0a..6cda65d 100644
--- a/lib/multihandle.h
+++ b/lib/multihandle.h
@@ -170,6 +170,9 @@ struct Curl_multi {
 #endif
   BIT(dead); /* a callback returned error, everything needs to crash and
                 burn */
+#ifdef DEBUGBUILD
+  BIT(warned);                 /* true after user warned of DEBUGBUILD */
+#endif
 };
 
 #endif /* HEADER_CURL_MULTIHANDLE_H */
~~~

bagder:
Right, maybe doing both would make sense. (as proposed by @dfandrich on the list)

bagder:
I'll merge this, but let's keep our ears and eyes open to see if this causes too much problems.

Issue #10253, opened 2023-01-07
http/2 arbitrary hangs in Windows since connection filter refactor


jay:
### I did this

@icing I am seeing arbitrary hangs during http/2 transfers in Windows and I've bisected it to 71b7e016 `lib: connect/h2/h3 refactor`. I think it has something to do with the pre-receive behavior that is on by default in Windows.

When I use curl on Windows built with `-DDONT_USE_RECV_BEFORE_SEND_WORKAROUND` I cannot reproduce. When I use curl on Ubuntu built with `-DUSE_RECV_BEFORE_SEND_WORKAROUND` I can reproduce. The hangs seem arbitrary but I've noticed smaller transfers seem more likely to hang. For example https://httpbin.org/bytes/1 (get 1 random byte) it takes about half a dozen attempts for me to see a hang. I have a local nginx server where the odds of curl hanging are better if I make up a non-existent path like https://localhost/asdf. It looks like timing has something to do with it as well.

<details>
  <summary>Click to expand sample output from a hang in Ubuntu</summary>
  
~~~
> GET /bytes/1 HTTP/2
> Host: httpbin.org
> user-agent: curl/7.87.1-DEV
> accept: */*
> 
* STATE: DO => DID handle 0x622000019908; line 2198 (connection #0)
* multi changed, check CONNECT_PEND queue
* STATE: DID => PERFORMING handle 0x622000019908; line 2317 (connection #0)
* http2_recv: easy 0x622000019908 (stream 1) win 33554432/65535
* [CONN-0-0][CF-SSL] TLSv1.2 (IN), TLS header, Supplemental data (23):
* nread=40
* Got SETTINGS
* MAX_CONCURRENT_STREAMS == 128
* ENABLE_PUSH == TRUE
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
* h2_process_pending_input: All data in connection buffer processed
* [CONN-0-0][CF-SSL] TLSv1.2 (OUT), TLS header, Supplemental data (23):
* http2_recv returns AGAIN for stream 1
* multi changed, check CONNECT_PEND queue
~~~
</details>

[...]

PR #10280, opened 2023-01-11
Conn_recv/send_plain refactoring, USE_RECV_BEFORE_SEND_WORKAROUND in cf-socket

icing:
Refactoring of the diverse Curl_send/recv_plain() and Curl_write/read_plain() functions, fix of stalling HTTP/2 processing when USE_RECV_BEFORE_SEND_WORKAROUND is in effect. Refs #10253.

- Curl_write_plain/Curl_read_plain are gone, usage replace with Curl_conn_send/recv() calls.
- Curl_send_plain/Curl_recv_plain have been internalized in `cf-socket.c?
- The "pre-receive" buffering (feature USE_RECV_BEFORE_SEND_WORKAROUND) has also been
   internalised in `cf-socket.c`.
- the HTTP/2 cfilter (nghttp2) now sets `data->state.drain` if, at the end of an operation, the "lower" filter chain reports pending data. This fixes the stalling of h2 processing when this feature is in effect.


jay:
Sorry that I did not have time to review this. I did test it and it seems to be working but I always see a lot more of drain in the verbose output:

~~~
* http2_recv returns AGAIN for stream 1
* Curl_readwrite: forcibly told to drain data
* http2_recv: easy 0xb66030 (stream 1) win 33554432/33554432
* Curl_readwrite: forcibly told to drain data
* http2_recv: easy 0xb66030 (stream 1) win 33554432/33554432
* Curl_readwrite: forcibly told to drain data
* http2_recv: easy 0xb66030 (stream 1) win 33554432/33554432
~~~

Which is probably this so maybe it's expected now:

> the HTTP/2 cfilter (nghttp2) now sets data->state.drain if, at the end of an operation, the "lower" filter chain reports pending data. This fixes the stalling of h2 processing when this feature is in effect.

Another website that can be used to reproduce the issue before this fix is `curl -v -H "Connection: close" https://httpd.apache.org/ -o NUL`

Though this PR fixes the issue and is a welcome improvement I am going to propose in a separate PR disabling by default USE_RECV_BEFORE_SEND_WORKAROUND in Windows builds. It has led to bugs that are hard to track down and that there is no easy way to test for. Like, this looks fixed but I can't determine for certain if it's fixed in all scenarios which is a problem. I'm basically running commands thousands of times and getting good feedback. I think that if buffering of socket data improves we can revisit enabling it by default.

icing:
@jay thanks for the testing. I agree that this windows workaround is hard to verify for correctness. Building on top of this PR with improvements is most welcome.

I believe a general buffering implementation in the socket filters is a Good Thing, if we can get a grip on the "pending" and "drain" handling.

bagder:
Thanks!

PR #10285, opened 2023-01-12
OpenSSL: use local BIO_METHOD instance

icing:
Make the BIO_METHOD instance a local variable of the OpenSSL connection filter.

This avoids UAF issues when `curl_global_cleanup()` is called before all transfers have been completely handled.


bagder (approved):

bagder:
The Windows CI failures are being addressed here: #10286

bagder:
Thanks!

PR #10287, opened 2023-01-12
CI: macOS fixup nghttp2 install

pheiduck:
[Fixup] Warning: The following dependents of upgraded formulae are outdated but will not be upgraded because they are not bottled:
  mongodb-community@5.0

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -140,19 +140,19 @@ jobs:
           CFLAGS: "-mmacosx-version-min=10.15 -Wno-error=undef -Wno-error=conversion"
         build:
         - name: OpenSSL
-          install: mongodb-community@5.0 nghttp2 openssl
+          install: nghttp2 openssl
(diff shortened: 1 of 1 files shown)


bagder:
How come we need it at all in the first place? Can't we just remove it instead?

pheiduck:
> How come we need it at all in the first place? Can't we just remove it instead?
Yes sure we can remove it, I want to fix this CI issue here: https://github.com/curl/curl/actions/runs/3904373833/jobs/6670071643#step:3:590

bagder:
I appreciate that! I just don't understand how mongodb is being involved in our builds in the first place nor how you could identify that being the problem...

Also, how come the circle CI macOS builds that also use homebrew don't see the same problem?

pheiduck:
> I appreciate that! I just don't understand how mongodb is being involved in our builds in the first place nor how you could identify that being the problem...
> 
> Also, how come the circle CI macOS builds that also use homebrew don't see the same problem?

I am not sure. I think we have to open an issue at the homebrew upstream project. Because it fails on all github macOS CIs, or we build it from scratch.

PR #10288, opened 2023-01-12
CI: macOS fixup nghttp2 install

pheiduck:
branch rename of https://github.com/curl/curl/pull/10287
CI issue: https://github.com/curl/curl/actions/runs/3904373833/jobs/6670071643#step:3:590
Goal fixup macOS github CIs to install nghttp2
Solved by uninstall both Python3 versions 3.11 and 3.10
3.9 is used to install pip pkgs.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -87,6 +87,8 @@ jobs:
           configure: --enable-debug --disable-ldap --with-openssl=/usr/local/opt/openssl
           macosx-version-min: 10.15
     steps:
+    - run: brew uninstall azure-cli pipx python@3.10 python@3.11
+      name: 'uninstall Python'
     - run: echo libtool autoconf automake pkg-config ${{ matrix.build.install }} | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
       name: 'brew bundle'
 
@@ -149,6 +151,8 @@ jobs:
           install: nghttp2 openssl libssh2
           generate: -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl -DCURL_USE_LIBSSH2=ON
     steps:
+    - run: brew uninstall azure-cli pipx python@3.10 python@3.11
+      name: 'uninstall Python'
     - run: echo libtool autoconf automake pkg-config ${{ matrix.build.install }} | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
       name: 'brew bundle'
 
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -87,8 +87,9 @@ jobs:
           configure: --enable-debug --disable-ldap --with-openssl=/usr/local/opt/openssl
           macosx-version-min: 10.15
     steps:
+    # We need to uninstall Python 3.10 and 3.11 otherwise nghttp2 fail to install
     - run: brew uninstall azure-cli pipx python@3.10 python@3.11
-      name: 'uninstall Python'
+      name: 'uninstall unused dependencies'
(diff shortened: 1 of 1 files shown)


pheiduck:
So the issue seems to be python 3.11 is installed twice. https://github.com/orgs/Homebrew/discussions/4117#discussioncomment-4690084

jay:
Do you have a reference to the brew issue that caused this problem?

pheiduck:
> ~Do you have a reference to the brew issue that caused this problem?~ Nevermind, I missed your post above. So, can't we just stop it from being installed twice? Where is that happening?

```
==> Pouring python@3.10--3.10.9.monterey.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3
Target /usr/local/bin/2to3
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.10

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.10

Possible conflicting files are:
/usr/local/bin/2to3 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/2to3
/usr/local/bin/idle3 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/idle3
/usr/local/bin/pydoc3 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/pydoc3
/usr/local/bin/python3 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/python3
/usr/local/bin/python3-config -> /Library/Frameworks/Python.framework/Versions/3.11/bin/python3-config
==> /usr/local/Cellar/python@3.10/3.10.9/bin/python3.10 -m ensurepip
==> /usr/local/Cellar/python@3.10/3.10.9/bin/python3.10 -m pip install -v --no-d
==> Caveats
Python has been installed as
  /usr/local/bin/python3

Unversioned symlinks `python`, `python-config`, `pip` etc. pointing to
`python3`, `python3-config`, `pip3` etc., respectively, have been installed into
  /usr/local/opt/python@3.10/libexec/bin

You can install Python packages with
  pip3 install <package>
They will install into the site-package directory
  /usr/local/lib/python3.10/site-packages

tkinter is no longer included with this formula, but it is available separately:
  brew install python-tk@3.10

See: https://docs.brew.sh/Homebrew-and-Python
==> Summary
🍺  /usr/local/Cellar/python@3.10/3.10.9: 3,108 files, 56.4MB
==> Running `brew cleanup python@3.10`...
Removing: /usr/local/Cellar/python@3.10/3.10.8... (3,136 files, 56.9MB)
Removing: /Users/runner/Library/Caches/Homebrew/python@3.10--3.10.8... (14.6MB)
==> Upgrading python@3.11
  3.11.0 -> 3.11.1 

==> Pouring python@3.11--3.11.1.monterey.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.11
Target /usr/local/bin/2to3-3.11
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.11'
```

pheiduck:
@jay we can try:
```
brew link --overwrite python@3.10
brew link --overwrite python@3.11
```

pheiduck:
@jay overwrite symlink does not help same error.
So I think the solution is to get rid of Python 3.10 and 3.11 as long as possible

jay:
What doesn't make sense to me is why we have to do that. It seems like we are coming in after the fact on cleanup when this shouldn't happen in the first place. Why does it happen? Isn't this a brew problem?

pheiduck:
@jay the PR is now no longer necessary, since the CI toolchain was updated yesterday https://github.com/actions/runner-images/commit/fda2b5e0d667a52902356cec0c296872dfd56056

Issue #10291, opened 2023-01-13
mingw on Windows build errors


bagder:
mingw now fails to build current git master on windows, see these recent [build logs](https://ci.appveyor.com/project/curlorg/curl/builds/45926363)

Pretty sure db91dbbf2c0a853 caused this regression.

icing:
Hmm, a gcc that reports __STDC_VERSION__ as c99 and yet refuses %z. well done.

bagder:
I don't think it hurts us too much if we just disable this check for mingw builds.

PR #10293, opened 2023-01-13
Disable log printf format checks on MINGW builds.

icing:
The MINGW compiler reports C99 and yet refuses to recognize %z formats. Disable printf log checks for these builds.


icing:
Refs #10291

PR #10295, opened 2023-01-13
[Windows] Missing 'sleep()' for MSVC

gvanem:
Done similar to how `lib1515.c` and `lib1542.c` does it for Windows.
Although MinGW do have `sleep()`, this patch should not hurt.
But perhaps add this to `test.h` instead?

### DIFF
--- a/tests/libtest/lib1515.c
+++ b/tests/libtest/lib1515.c
@@ -37,10 +37,6 @@
 
 #define DNS_TIMEOUT 1
 
-#if defined(WIN32) || defined(_WIN32)
-#define sleep(sec) Sleep ((sec)*1000)
-#endif
-
 static int debug_callback(CURL *curl, curl_infotype info, char *msg,
                           size_t len, void *ptr)
 {
--- a/tests/libtest/lib1542.c
+++ b/tests/libtest/lib1542.c
@@ -36,10 +36,6 @@
 #include "warnless.h"
 #include "memdebug.h"
 
-#if defined(WIN32) || defined(_WIN32)
-#define sleep(sec) Sleep ((sec)*1000)
-#endif
-
 int test(char *URL)
 {
   CURL *easy = NULL;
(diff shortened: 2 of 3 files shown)


bagder:
> But perhaps add this to `test.h` instead?

That seems clever:
~~~shell
$ git grep -l 'sleep(' *.c
lib1515.c
lib1542.c
lib2301.c
lib2302.c
lib2304.c

jay:
I agree. server\util.h already does something similar:

https://github.com/curl/curl/blob/c12fb3ddaf48e709a7a4deaa55ec485e4df163ee/tests/server/util.h#L45-L49

gvanem:
So what about this PR? Should I or @bagder merge it and change the other files in another PR?

jay:
> So what about this PR?

You could update this PR with your proposed changes then force push to lib2304-sleep, for example

git checkout lib2304-sleep
#make changes
#stage changes using git gui or some other way
git commit --amend
git push -f

curl_setup.h is included by test.h so you don't need to check _WIN32 just WIN32
https://github.com/curl/curl/blob/c12fb3ddaf48e709a7a4deaa55ec485e4df163ee/lib/curl_setup.h#L48-L50

Review [push access guidelines](https://github.com/curl/curl/wiki/push-access-guidelines), specifically commit style, for example `libtest: add sleep() macro for Windows`

gvanem:
> You could update this PR

Sorry, no. That's beyond my git knowledge.

jay:
Ok, I understand that, but it doesn't have to be. Amending a single commit can be done in `git gui`. If you can't or don't want to do that then you could add another commit to this branch with the new changes and I will squash them together (essentially, your existing changes in lib2304 would disappear) before I commit it.

(Also, as described in the guidelines you should work in your own fork of the curl repo. There are some exceptions for this suchas @bagder.)

Here's a screenshot from git gui

Capture

jay:
@gvanem if you want I can finish this

gvanem:
@jay Yes, please do.

MarcelRaad (approved):

jay:
Thanks

Issue #10309, opened 2023-01-16
Use heap after free in libcurl linked with Schannel


sergio-nsk:
From here, in `multi_runsingle()` on the line curl/lib/multi.c:1934, `host.rawalloc` is freed: `Curl_safefree(existing->host.rawalloc);`
```
1: 0x1870da80 is located 0 bytes inside of 26-byte region [0x1870da80,0x1870da9a)
1: freed by thread T0 here:
1:     #0 0x14ba221 in free D:\a01\_work\38\s\src\vctools\crt\asan\llvm\compiler-rt\lib\asan\asan_malloc_win.cpp:109
1:     #1 0x1b2323c in reuse_conn C:\actions-runner\_work\client\client\third-party\curl\lib\url.c:3387
1:     #2 0x1b1bde6 in create_conn C:\actions-runner\_work\client\client\third-party\curl\lib\url.c:3747
1:     #3 0x1b16b0c in Curl_connect C:\actions-runner\_work\client\client\third-party\curl\lib\url.c:3946
1:     #4 0x1b02c71 in multi_runsingle C:\actions-runner\_work\client\client\third-party\curl\lib\multi.c:1934
```
Later in the same function `multi_runsingle()` on the line curl/lib/multi.c:2469, it calls `schannel_shutdown()` on the line curl/lib/vtls/schannel.c:2505 that accesses `connssl->hostname` that definitely points to somewhere inside the freed memory.
```
1: ==476==ERROR: AddressSanitizer: heap-use-after-free on address 0x1870da80 at pc 0x014bf5f9 bp 0x004f779c sp 0x004f7790
1: READ of size 3 at 0x1870da80 thread T0
1:     #0 0x14bf611 in __asan_wrap_strlen D:\a01\_work\38\s\src\vctools\crt\asan\llvm\compiler-rt\lib\sanitizer_common\sanitizer_common_interceptors.inc:375
1:     #1 0x1b6b684 in dprintf_formatf C:\actions-runner\_work\client\client\third-party\curl\lib\mprintf.c:842
1:     #2 0x1b6[864](https://github.com/snxd/client/actions/runs/3934099623/jobs/6728495573#step:8:865)c in curl_mvsnprintf C:\actions-runner\_work\client\client\third-party\curl\lib\mprintf.c:1036
1:     #3 0x1b372c7 in Curl_infof C:\actions-runner\_work\client\client\third-party\curl\lib\sendf.c:251
[...]

Issue #10273, opened 2023-01-10
use-after-free in Curl_ssl_addsessionid()


stefantalpalaru:
### I did this

I built Transmission with `-fsanitize=address` and ran into this use-after-free problem inside Curl code: https://gist.github.com/stefantalpalaru/45bbfea53e051078bedabfd090934eeb

The shared handle is used from a single thread, so I don't think it's a case of https://curl.se/docs/knownbugs.html#A_shared_connection_cache_is_not

Transmission is sharing everything it can for that handle: https://github.com/transmission/transmission/blob/0af121004c0ca5321d1625e465640fa379c25d9a/libtransmission/web.cc#L727-L745

### curl/libcurl version

curl 7.87.0 (x86_64-pc-linux-gnu) libcurl/7.87.0 (NSS/3.86) OpenSSL/1.1.1s zlib/1.2.13 zstd/1.5.2 c-ares/1.18.1 nghttp2/1.51.0
Release-Date: 2022-12-21
Protocols: dict file ftp ftps http https imap imaps mqtt pop3 pop3s rtsp smtp smtps tftp
Features: AsynchDNS HTTP2 HTTPS-proxy IPv6 Largefile libz MultiSSL NTLM SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

Linux amd2 6.0.15-gentoo #1 SMP PREEMPT_DYNAMIC Mon Dec 26 19:39:26 CET 2022 x86_64 AMD Ryzen 9 5950X 16-Core Processor AuthenticAMD GNU/Linux

bagder:
What more can you tell us about what curl did before this situation happened? If we can't reproduce the problem, this is going to be much harder.

The stack trace implicates the shared session id cache.

stefantalpalaru:
I tried to disable sharing by commenting out this line: https://github.com/transmission/transmission/blob/0af121004c0ca5321d1625e465640fa379c25d9a/libtransmission/web.cc#L481

Nine and a half hours later, the same use-after-free appeared, so it's not related to `CURLOPT_SHARE`.
(Maybe `curl_share_setopt(_, CURLSHOPT_SHARE, _)` overrides a missing `curl_easy_setopt(_, CURLOPT_SHARE, _)`?)

> If we can't reproduce the problem, this is going to be much harder.

I know, but it takes a highly variable amount of time, seeding some random torrent with Transmission to see this. It can be minutes, it can be hours.

stefantalpalaru:
[...]

PR #10310, opened 2023-01-17
vtls hostname handling fix

icing:
Fixes the hostname in vtls filters to use a copy instead of referencing the `connectdata` instance since this may get free'ed on connection reuse. Refs #10309, #10273.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -3369,6 +3369,20 @@ static void reuse_conn(struct Curl_easy *data,
   }
 #endif
 
+  /* Finding a connection for reuse in the cache matches, among other
+   * things on the "remote-relevant" hostname. This is not necessarily
+   * the authority of the URL, e.g. conn->host. For example:
+   * - we use a proxy (not tunneling). we want to send all requests
+   *   that use the same proxy on this connection.
+   * - we have a "connect-to" setting that may redirect the hostname of
+   *   a new request to the same remote endpoint of an existing conn.
+   *   We want to reuse an existing conn to the remote endpoint.
+   * Since connection reuse does not match on conn->host necessarily, we
+   * switch `existing` conn to `temp` conn's host settings.
+   * TODO: is this correct in the case of TLS connections that have
+   *       used the original hostname in SNI to negotiate? Do we send
+   *       requests for another host through the different SNI?
+   */
   Curl_free_idnconverted_hostname(&existing->host);
   Curl_free_idnconverted_hostname(&existing->conn_to_host);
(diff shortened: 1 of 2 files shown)


stefantalpalaru:
Patch tested and it works as advertised.

icing:
@jay, if you find the time, could you merge this PR?

jay:
Thanks. We don't use the free((void *) pattern , so I removed const qualifier from hostname and dispname. Regarding the TODO SNI note, I don't understand.

https://github.com/curl/curl/blob/f8da4f2f2d0451dc0a126ae3e5077b4527ccdc86/lib/url.c#L3382-L3384

`hostname` is different from `dispname` when the hostname is a decoded IDN. In that case `hostname` is the punycode ascii hostname and `dispname` is the original international hostname. SNI hostname is from `hostname` and not `dispname`. I do not understand what you mean by different requests have different SNI, if they have the same hostname.

icing:
> Thanks. We don't use the free((void *) pattern , so I removed const qualifier from hostname and dispname. Regarding the TODO SNI note, I don't understand.
> 
> https://github.com/curl/curl/blob/f8da4f2f2d0451dc0a126ae3e5077b4527ccdc86/lib/url.c#L3382-L3384
> 
> `hostname` is different from `dispname` when the hostname is a decoded IDN. In that case `hostname` is the punycode ascii hostname and `dispname` is the original international hostname. SNI hostname is from `hostname` and not `dispname`. I do not understand what you mean by different requests have different SNI, if they have the same hostname.

Reading the comment again today, I should have been more clear by giving an example.

```
--resolve hostA:port:dest1 urlA --resolve hostB:port:dest1 urlB
```
will `urlB` reuse the connection from `urlA`? If they yes and both are `https`, then `urlB` is sent over a TLS connection with SNI `hostA`. Which depends on the server if it allows this.

jay:
> --resolve hostA:port:dest1 urlA --resolve hostB:port:dest1 urlB

Resolve and connect-to do not affect the SNI. In your example the urlB transfer does not reuse the hostA connection unless hostA and hostB hostnames are the same.

icing:
Ah, yes it not only checks the `conn_to` but also `conn->host`. All fine. Thanks!

icing:
Relatedly, the issue was caused by the freeing of `conn->host` on every connection reuse. I believe, this then to be (mostly?) unnecessary. 

Just a side issue, since I feel more comfortable with the vtls filter holding  its copy now.

PR #10312, opened 2023-01-17
HAPPY_EYEBALL test case and fixes

icing:
Connect strategy testing for attempts, timeouts and happy-eyeball handling.

- test2600 is a unit test that triggers various connect conditions and monitors behaviour, available in a debug build only.
- this exposed edge cases in connect.c that have been fixed here.


bagder:
Thanks!

PR #10316, opened 2023-01-17
.cirrus.yml: fixup typo arm64 > aarch64

pheiduck:


dfandrich:
I wouldn't call it a "typo". arm64 is used sometimes and aarch64 others. It's more a matter of taste, IMHO.

jay:
I agree, AFAICT they are both the same and this comes down to preference. I think we are already familiar with arm64 tag so why not just leave it.

PR #10320, opened 2023-01-19
openssl: interop with AWS-LC

jeking3:
This pull request adds support for building against AWS-LC.
AWS-LC is a BoringSSL/OpenSSL derivative.  For more information see https://github.com/awslabs/aws-lc/
This is my first PR into curl - please let me know if I missed anything or changed something I was not supposed to.

### Changes

* CMakeLists.txt changes to detect AWS-LC (very minor)
* Compile-time branches needed to support AWS-LC
* GitHub Actions script to build with autoconf and cmake against AWS-LC

### Testing
* awslc.yaml GHA workflow result (PASS): https://github.com/jeking3/curl/actions/runs/3953692960


jeking3:
I did a force push to fix a LONGLINE formatting error coming from checksrc.pl during the build in CI.

jeking3:
The 6 macOS build failures appear to be pre-existing to this PR and should be resolved by #10288.
The curl/check curl-novalgrind-ngtcp2-with-openssl failure appears to be pre-existing to this PR.

jay:
This looks fine to me but we are in a [feature freeze](https://calendar.google.com/calendar/u/0/embed?src=c9u5d64odop9js55oltfarjk6g@group.calendar.google.com&ctz=Europe/Stockholm) and not adding new symbols, so this will have to wait until the next feature window opens.

jeking3:
What does this mean in terms of timing before it is able to get into a release?  The next one I see on the calendar you provided is in March.  I was hoping to be able to get this into 7.88.0 to simplify some near-term plans.  Given it was submitted on the day of the freeze and is very low risk, would you consider an exception?

jeking3:
It looks like I didn't quite get all of the compile-time branches, like OSSL_PACKAGE definition and version printout, and a curl-openssl.m4 section for detecting AWS-LC.

jeking3:
Regarding the feature window, that's fine, we'll let this bake appropriately then, and it will be better when it goes in.

jeking3:
That line length linter sure is pedantic!

jeking3:
Rebased to pick up CI improvements.

jeking3:
Rebased to resolve conflicts in READMEs.  I see we're down to two failing CI jobs (unrelated to these changes), so it looks like some of those were fixed (or removed) as well.

jeking3:
Rebased on master to resolve conflict and pick up CI fixes.

MarcelRaad (approved):
Looks good to me and the CI failures look unrelated.

docs/libcurl/symbols-in-versions
```diff
@@ -1015,6 +1015,7 @@ CURLSSH_AUTH_KEYBOARD           7.16.1
 CURLSSH_AUTH_NONE               7.16.1
 CURLSSH_AUTH_PASSWORD           7.16.1
 CURLSSH_AUTH_PUBLICKEY          7.16.1
+CURLSSLBACKEND_AWSLC            7.88.0
```
bagder, line 1018: This should probably become 8.1.0 as that's the next feature window version.

.github/scripts/spellcheck.words
```diff
@@ -42,6 +42,11 @@ Automake
 Autotools
 autotools
 AVR
+AWS
+AWS-LC
+AWSLC
+aws-lc
+awslc
```
bagder, line 49: Do we really need to spell it this many different ways? Is there not an official way we could stick to?

bagder (changes requested):

Issue #10321, opened 2023-01-19
openssl + wolfssl combo backend build broken


marekr:
This commit made a change to curl_ntlm_core https://github.com/curl/curl/commit/5fd7cd7302d3ea265661822b5eea200f97005133

It rearranged the includes in the curl_ntlm_core.c so that openssl headers are included before wolfssl if both backends are enabled.  However, the curl_ntlm_core.h header was not updated and it still includes wolfssl before openssl.

The problem is curl_ntlm_core.h since updated, if you attempt to build, the source file will include openssl and the header will include wolfssl. Wolfssl's compat types immediately clash into openssl types

The result is

```
FAILED: lib/CMakeFiles/libcurl.dir/curl_ntlm_core.c.obj 
[...]

PR #10322, opened 2023-01-19
tls: fixes for wolfssl + openssl combo builds

vszakats:
1. Add `USE_WOLFSSL` to the TLS backend priority list in `lib/curl_ntlm_core.c`.

2. Fix `lib/curl_ntlm_core.h` to respect TLS backend priority, bringing this in sync with the above list and `lib/curl_ntlm_core.c` itself.

   Reported-by: Mark Roszko
   Ref: https://github.com/curl/curl/issues/10321

3. Allow enabling both wolfSSL and OpenSSL at the same time in `lib/Makefile.mk`. Update the logic to select the crypto-specific lib for `ngtcp2`, where only one TLS backend is supported at the same time.

Closes #10322

### DIFF
--- a/lib/Makefile.mk
+++ b/lib/Makefile.mk
@@ -183,13 +183,6 @@ ifneq ($(findstring -ssl,$(CFG)),)
   OPENSSL_LIBS ?= -lssl -lcrypto
   _LIBS += $(OPENSSL_LIBS)
 
-  ifneq ($(wildcard $(OPENSSL_INCLUDE)/openssl/aead.h),)
-    OPENSSL := boringssl
-  else
-    # including libressl
-    OPENSSL := openssl
-  endif
-
   ifneq ($(findstring -srp,$(CFG)),)
     ifneq ($(wildcard $(OPENSSL_INCLUDE)/openssl/srp.h),)
       # OpenSSL 1.0.1 and later.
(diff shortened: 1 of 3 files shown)


jay:
The reason it was the other way around though was because if OpenSSL came first (like it does in this PR) then `include <openssl/ssl.h>` could include a wolfSSL include with that name (depending on build order; if OpenSSL and wolfSSL include paths are both used) if I understand @elms' commit right.

vszakats:
I tend to think that if this combo isn't supported in _all_ cases (which it isn't), we can hardly say it this is a supported combination. It also don't seem to be something much tested before. @elms commit says it adds support, but there's no mention of building in parallel with OpenSSL, so this maybe something not tested or considered. To me the order picked seems arbitrary, by just extending existing code. As for header confusion (picking the same header from the wrong project): if the header paths are setup correctly this should not happen because wolfSSL ones are referenced under `wolfssl/openssl/`.

(All of this is mostly guessing, my build attempts always failed with this combo.)

bagder:
wolfSSL provides an OpenSSL API, that curl partially uses to do some things.

The only way we can build curl to use *both* OpenSSL and wolfSSL simultaneously is if they the two libraries do not share any symbols/function names, which would mean that wolfSSL's OpenSSL API would have to be disabled. That would then make the wolfSSL backend less feature rich.

If someone insists on making that work, fine, but I'm not going to personally lose any sleep over us not supporting that.

vszakats:
I plan to commit this as-is. This should help adding further fixes if someone is interested in doing those. It also brings `.mk` builds on par with cmake/autotools, which also don't explicitly block this combination.

Issue #10305, opened 2023-01-16
OS400 compiling a C++ program with the include curl.h fails.


johnsherrill:
The include curl.h includes systems.h
systems.h has this section for OS400

~~~c
#elif defined(__OS400__)
#  if defined(__ILEC400__)
#    define CURL_TYPEOF_CURL_OFF_T     long long
#    define CURL_FORMAT_CURL_OFF_T     "lld"
#    define CURL_FORMAT_CURL_OFF_TU    "llu"
#    define CURL_SUFFIX_CURL_OFF_T     LL
#    define CURL_SUFFIX_CURL_OFF_TU    ULL
#    define CURL_TYPEOF_CURL_SOCKLEN_T socklen_t
#    define CURL_PULL_SYS_TYPES_H      1
#    define CURL_PULL_SYS_SOCKET_H     1
#  endif           
~~~

`__ILEC400__` is only defined if using the C compiler 
If you what to  use curl in C++ code the compile fails 
since `CURL_TYPEOF_CURL_OFF_T` is not defined so the below code is not used

~~~c
/* Data type definition of curl_off_t. */

#ifdef CURL_TYPEOF_CURL_OFF_T
  typedef CURL_TYPEOF_CURL_OFF_T curl_off_t;
#endif                           
~~~
which causes errors in curl.h since curl_off_t is not defined.

adding || (__cplusplus) to the "#if defined(__ILEC400) line fixes the problem.

jay:
> adding || (__cplusplus) to the "#if defined(__ILEC400) line fixes the problem.

@jonrumsey does this seem right to you?

johnsherrill:
I can state with __cplusplus  added I can compile C++ that includes curl.h. Without it fails with curl_off_t not defined.

jonrumsey:
Yes, `#if defined(__ILEC400__) || defined(__cplusplus)` would define those types for both C & C++ compilers, however you'll only be including the systems.h header when compiling a C/C++ module so just removing the check for `__ILEC400__` and leaving this under the `__OS400__` check is equally valid and perhaps a little clearer.

johnsherrill:
I agree that would be cleaner.

PR #10329, opened 2023-01-20
system.h: assume OS400 is always built with ILEC compiler

jay:
Prior to this change the OS400 types were only defined when `__ILEC400__`. That symbol is only defined by IBM's C compiler and not their C++ compiler, which led to missing types when users would compile a C++ application.

Assisted-by: Jon Rumsey
Reported-by: John Sherrill

Fixes https://github.com/curl/curl/issues/10305
Closes #xxxx

---

Can we be sure that these types are the only types used and that compiler is the only compiler used on OS400?

/cc @jonrumsey @johnsherrill

### DIFF
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -227,16 +227,14 @@
 #  define CURL_TYPEOF_CURL_SOCKLEN_T unsigned int
 
 #elif defined(__OS400__)
-#  if defined(__ILEC400__)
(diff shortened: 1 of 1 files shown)


jonrumsey:
The IBM C and C++ compilers are the only native compilers on the platform. In the IBM i PASE environment it is possible to use a broad range of other compilers like gcc, however compilers in this environment build and link against AIX binaries, not native OS400.

PR #10331, opened 2023-01-21
Add version info for NGTCP2 (similar t other libs)

MortenMacFly:
Similar to all other libraries the version of NGTCP2 would be accessible in the curl_version_info struct.
Not sure if it was intended not to do so until now though...
Still I hope you'll find this useful.


MortenMacFly:
I'm afraid I don't get what the root cause or the error is... :-/ I'll continue to try to understand and dig into it...
It was compiling on my machine nut of course it was just one compiler and not that many...
I am sorry if this causes any trouble...

MortenMacFly:
This is the error:

jay:
> The bitmask seems "full"? What to do???

I don't know... I don't remember discussing the CURL_VERSION macros specifically but it's the same problem as we have with 32 protocols, see https://curl.se/mail/lib-2021-09/0029.html. 

The CURL_VERSION are used for features, usually. Isn't ngtcp2 a supporting library the same as nghttp2?  We have CURL_VERSION_HTTP2 and CURL_VERSION_HTTP3 is that not enough?

MortenMacFly:
> The CURL_VERSION are used for features, usually. Isn't ngtcp2 a supporting library the same as nghttp2? We have CURL_VERSION_HTTP2 and CURL_VERSION_HTTP3 is that not enough?

Well actually these are 3 libs. Currently, (as of the last release v7.87.0_2) their version numbers are:
- NGHTTP2 = 1.51.0
- NGHTTP3 = 0.8.0
- NGTCP2  = 0.12.1
(The latter two come from QUIC on Windows.)

That is what I wanted to reflect. (Maybe its overdone through...)

After removing the feature flag ting it at least compiles on more machines, but still not all.

include/curl/curl.h
```diff
@@ -3128,6 +3132,7 @@ typedef struct curl_version_info_data curl_version_info_data;
 #define CURL_VERSION_HSTS         (1<<28) /* HSTS is supported */
 #define CURL_VERSION_GSASL        (1<<29) /* libgsasl is supported */
 #define CURL_VERSION_THREADSAFE   (1<<30) /* libcurl API is thread-safe */
+#define CURL_VERSION_NGTCP2       (1<<31) /* libngtcp2 is supported */
```
bagder: This is not really a "feature" for us. ngtcp2 is a library, HTTP3 is the feature.

docs/libcurl/curl_version_info.3
```diff
@@ -113,6 +113,9 @@ typedef struct {
   /* when 'age' is CURLVERSION_ELEVENTH or higher (>= 7.87.0), the members
      below exist */
   const char *const *feature_names; /* Feature names. */
+  /* when 'age' is CURLVERSION_TWELFTH or higher (>= 7.88.0), the members
+     below exist */
+  const char *ngtcp2_version; /* human readable string. */
```
bagder, line 118: when ngtcp2 is used, so is nghttp3 which is a separate and independent library. Shouldn't that too be reported then?

bagder (changes requested):
Some too long lines:
~~~
./version.c:256:95: warning: Longer than 79 columns (LONGLINE)
   msnprintf(ngtcp2_buf, sizeof(ngtcp2_buf), "libngtcp2/%s", ngtcp2_version_info->version_str);
./version.c:670:93: warning: Longer than 79 columns (LONGLINE)
     msnprintf(ngtcp2_buffer, sizeof(ngtcp2_buffer), "%s", ngtcp2_version_info->version_str);

docs/libcurl/curl_version_info.3
MortenMacFly: You are right. I did not see that. I'll try to update the code accordingly...

include/curl/curl.h
MortenMacFly: OK, I've removed this anyway. Thanks for the clarification.

docs/libcurl/curl_version_info.3
MortenMacFly: I've implemented the same for HTTP3 now to be consistent. Build queue still runs but hopefully with a success (some always fails though - you are the expert....).

lib/version.c
```diff
@@ -73,6 +73,14 @@
 #include <gsasl.h>
 #endif
 
+#ifdef USE_NGTCP2
+#include <ngtcp2.h>
+#endif
+
+#ifdef USE_NGHTTP3
+#include <nghttp3.h>
+#endif
```
bagder, line 82: This is not how these header files are included. They are located in subdirectories. But I also have a more generic thought on this.

bagder (commented):
Since we have three different quic/h3 backends already, should we not make these new struct fields more generic to work with them as well? Perhaps by calling them "quic_library" and "http3_library" and then call a backend specific function to fill them in appropriately?

lib/version.c
MortenMacFly: I am not entirely sure how to handle this. If USE_NGHTTP3 defined, the include folders point to the folder with this header file in it. Also, it is done similarly for other library header files, too. So I am not exactly sure what to do. But I take your second part of the remark as "I have a better idea" - which is totally fine with me. So this is just to confirm I read your comment but will/can nothing do about without further advise. :-)

bagder: 1. The function that outputs the quic and http3 versions should be in the `curl_ngtcp2.c` file, and correspondingly in `curl_msh3.c`  and `curl_quiche.c.
2. In `vquic.c` you create the generic quic function(s) that is called from version.c to output the version

See the TLS library version as an example.

bagder:
Wait, we already have `quic_version` version in there! Isn't that what you try to add?

MartinHalleTUHH:
> Wait, we already have `quic_version` version in there! Isn't that what you try to add?

Alright, I tried to read myself into the actual meaning of [quic](https://en.wikipedia.org/wiki/QUIC) As far as I understand now, [ngtcp2](https://github.com/ngtcp2/ngtcp2) builds _on top of that_ and may additionally need [nghttp3](https://github.com/ngtcp2/nghttp3) for HTTP/3, right? In that case if quic is the "core" and the others are features based on the core I assume this version info is enough, really. I was under the impression that its independent stuff.

So, fair enough, that makes this PR obsolete.

PR #10332, opened 2023-01-22
arm64 is supported

freshe:
Minor thing. Building an arm64 version works flawlessly with the VS arm64 toolset. So I guess it's supported, just not mentioned in the README.


jay:
Thanks

Issue #10336, opened 2023-01-23
Assertion failure in bio_cf_out_write if CURLOPT_HEADERFUNCTION callback gets CURLINFO_TLS_SSL_PTR


fujii:
I'm using libcurl 7.87.0 with libressl 3.7.0 for Windows WebKit.
After upgrading libcurl to 7.87.0 from 7.86, WebKit browser is crashing by loading https pages with a proxy.

With the debug libcurl.dll, I'm observing the following assertion failure.

~~~
Assertion failed!

Program: ...bkit\gb\WebKitLibraries\wincairo\bin64\libcurl.dll
D:\work\WinCairoReq\g\buildtrees\curl\src\7...\openssl.c
Line: 709

Expression: data
~~~

Callstack:

~~~
ucrtbased.dll!common_assert_to_message_box<wchar_t>(const wchar_t * const expression, const wchar_t * const file_name, const unsigned int line_number, void * const return_address) Line 388	C++
ucrtbased.dll!common_assert<wchar_t>(const wchar_t * const expression, const wchar_t * const file_name, const unsigned int line_number, void * const return_address) Line 424	C++
ucrtbased.dll!_wassert(const wchar_t * expression, const wchar_t * file_name, unsigned int line_number) Line 444	C++
libcurl.dll!bio_cf_out_write(bio_st * bio, const char * buf, int blen) Line 709	C
crypto-50.dll!BIO_write(bio_st * b, const void * in, int inl) Line 412	C
ssl-53.dll!tls13_legacy_wire_write(ssl_st * ssl, const unsigned char * buf, unsigned __int64 len) Line 75	C
ssl-53.dll!tls13_legacy_wire_write_cb(const void * buf, unsigned __int64 n, void * arg) Line 97	C
ssl-53.dll!tls13_record_send(tls13_record * rec, __int64(*)(const void *, unsigned __int64, void *) wire_write, void * wire_arg) Line 178	C
ssl-53.dll!tls13_record_layer_write_record(tls13_record_layer * rl, unsigned char content_type, const unsigned char * content, unsigned __int64 content_len) Line 1082	C
ssl-53.dll!tls13_record_layer_write_chunk(tls13_record_layer * rl, unsigned char content_type, const unsigned char * buf, unsigned __int64 n) Line 1101	C
ssl-53.dll!tls13_record_layer_write(tls13_record_layer * rl, unsigned char content_type, const unsigned char * buf, unsigned __int64 n) Line 1116	C
[...]

PR #10340, opened 2023-01-23
Nested cfilter calls, handling of current data

icing:
Manage handling of current easy handle in nested cfilter calls.

- refs #10336 where the previous implementation cleared `data` so the outer invocation lost its data.

### DIFF
--- a/lib/cfilters.h
+++ b/lib/cfilters.h
@@ -433,4 +433,50 @@ size_t Curl_conn_get_max_concurrent(struct Curl_easy *data,
                                     struct connectdata *conn,
                                     int sockindex);
 
+
+/**
+ * Types and macros used to keep the current easy handle in filter calls,
+ * allowing for nested invocations. See #10336.
+ *
+ * `cf_call_data` is intended to be a member of the cfilter's `ctx` type.
+ * A filter defines the macro `CF_CTX_CALL_DATA` to give access to that.
+ *
+ * With all values 0, the default, this indicates that there is no cfilter
+ * call with `data` ongoing.
+ * Macro `CF_DATA_SAVE` preserves the current `cf_call_data` in a local
+ * variable and sets the `data` given, incrementing the `depth` counter.
+ *
+ * Macro `CF_DATA_RESTORE` restores the old values from the local variable,
+ * while checking that `depth` values are as expected (debug build), catching
+ * cases where a "lower" RESTORE was not called.
+ *
+ * Finally, macro `CF_DATA_CURRENT` gives the easy handle of the current
+ * invocation.
+ */
+struct cf_call_data {
+  struct Curl_easy *data;
+  int depth;
+};
+
+/**
+ * define to access the `struct cf_call_data for a cfilter. Normally
+ * a member in the cfilter's `ctx`.
+ *
+ * #define CF_CTX_CALL_DATA(cf)   -> struct cf_call_data instance
+*/
+
+#define CF_DATA_SAVE(cf, data) \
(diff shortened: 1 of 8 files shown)


icing:
Apart from general trouble on latest macos header files, this seems to look good.

@jay  care to review and eventually merge? Thanks!

xquery (commented):
just a nitpick - you might consider relegating these unrelated changes to a separate PR

lib/cfilters.h
```diff
@@ -433,4 +433,50 @@ size_t Curl_conn_get_max_concurrent(struct Curl_easy *data,
...
+*/
+
+#define CF_DATA_SAVE(cf, data) \
+  struct cf_call_data cf_cd_save = CF_CTX_CALL_DATA(cf); \
+  DEBUGASSERT(cf_cd_save.data == NULL || cf_cd_save.depth > 0); \
+  CF_CTX_CALL_DATA(cf).data = (struct Curl_easy *)data; \
+  CF_CTX_CALL_DATA(cf).depth++
+
+#define CF_DATA_RESTORE(cf) \
+  DEBUGASSERT(CF_CTX_CALL_DATA(cf).depth == cf_cd_save.depth + 1); \
```
xquery: more a question of my ignorance ... are these DEBUGASSERT 'lit up' with current CI ?

jay:
@xquery What specifically are you referring to? This looks straightforward to me and I don't see how or why it should be broken up.

lib/cfilters.h
```diff
@@ -433,4 +433,50 @@ size_t Curl_conn_get_max_concurrent(struct Curl_easy *data,
...
+ * a member in the cfilter's `ctx`.
+ *
+ * #define CF_CTX_CALL_DATA(cf)   -> struct cf_call_data instance
+*/
+
+#define CF_DATA_SAVE(save, cf, data) \
+  (save) = CF_CTX_CALL_DATA(cf); \
+  DEBUGASSERT((save).data == NULL || (save).depth > 0); \
+  CF_CTX_CALL_DATA(cf).data = (struct Curl_easy *)data; \
+  CF_CTX_CALL_DATA(cf).depth++
```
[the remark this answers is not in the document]
icing: agreed.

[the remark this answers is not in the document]
icing: Just added this and also made non-debug versions that use less space.

lib/cfilters.h
```diff
@@ -433,4 +433,71 @@ size_t Curl_conn_get_max_concurrent(struct Curl_easy *data,
...
+  do { (save) = CF_CTX_CALL_DATA(cf); \
+    DEBUGASSERT((save).data == NULL || (save).depth > 0); \
+    CF_CTX_CALL_DATA(cf).depth++;  \
+    CF_CTX_CALL_DATA(cf).data = (struct Curl_easy *)data; \
+  } while(0)
+
+#define CF_DATA_RESTORE(cf, save) \
+  do { \
+    DEBUGASSERT(CF_CTX_CALL_DATA(cf).depth == (save).depth + 1); \
+    DEBUGASSERT((save).data == NULL || (save).depth > 0); \
```
jay, line 482: The depth is never decremented, why is that not necessary?

icing: Because RESTORE restores the previous value?

jay: > Because RESTORE restores the previous value? At least that is how I intend it to work...

I see it now. `save` is a copy of call_data; call_data's data and depth is updated; on restore save is assigned back to call_data. Ok.

jay:
Thanks

PR #10342, opened 2023-01-25
CI: Switch to ReviewPad to label pull requests

dfandrich:
GitHub tokens in actions do not allow the "write" capability when
running on pull requests originating from forked repositories, which is
a major source of pull requests for curl. Additionally, the
actions/labeler action does not currently support ensuring that all
files in the PR match at least one pattern. Either one of these
limitations make it unsuitable for curl.

ReviewPad is a GitHub application that solves both those limitations,
and provides additional functionality that could be useful to the curl
project in the future. This change switches the existing file patterns
to ReviewPad and improves them by labeling PRs with files that match
several patterns in one label.

If approved, I'll enable ReviewPad on the curl repo then submit this.


jay:
Ref: https://github.com/curl/curl/pull/10326

vszakats:
Is it worth it to channel through all commits/PRs through a third-party just for this functionality, while also giving that third-party (partial?) write access to the repo?

I sense that automatic labeling is inherently a hit or miss even with a strong set of filename rules. These rules will also need to be maintained and kept in sync with the codebase.

Is there any way to delete wrongly added or redundant labels? It seems they will be re-added after each PR commit.

Manual labeling on the other hand gives a small incentive to go through new (unlabeled) PR/Issues, read and understand them, and then label them, the latter being a relatively small extra effort, while also not having to fight with/against any bot.

dfandrich:
That's the question we'll have to answer. Right now, the only way to get this
functionality in a safe way is to use an app or run a bot ourselves. My feeling
is that we should automate as much as we reasonably can. Right now, we have to
label 100% of PRs manually, so if we can drop that to 50% (or 20%) then it's
human lives saved. The label will also appear instantly. My reading of the docs
is that the bot will do its labelling once per PR so we wouldn't be fighting
it.

There's more functionality I'd like to add if this goes ahead, like adding a
reference to the PR in the description (which I always forget to do) and
suggest reviewers based on the files changed.

bagder:
> suggest reviewers based on the files changed.

This particular thing is already managed by `.github/CODEOWNERS`.

bagder:
I seem to see "labeler" set labels correctly already even for PRs from forked repos so what exactly would reviewpad add for us?

dfandrich:
labeler has an extremely limited set of PR that it will label right now, namely
only PRs that contain a single file or those that contain more than one file
that all match a single glob (like a couple of documentation files).
Fortunately, most PRs we get seem to touch a single file so it's still working
in many cases.  I have a PR outstanding [1] on labeler to also allow labeling a
larger set of PRs in one category (specifically, those containing multiple
files that can each match any glob for a single label). But even with that
feature it can't catch a PR matching the common case of PRs that touch both a
source file and a man page.

ReviewPad has a more powerful syntax that would catch all of these cases as
well as the last one today.  But, I'm happily surprised by how many PRs labeler
is catching already so if there's concern about granting ReviewPad permissions
on curl then I'll withdraw this PR for now and see how the improved labeler
fares.

[1]: https://github.com/actions/labeler/pull/487

dfandrich:
I'm going to close this for now, since the current labeler is working surprisingly well, with hope for even better results with an update. We can always revisit this if we want the improved PR coverage that ReviewPad can provide.

PR #10353, opened 2023-01-28
CryptAcquireContext should be found

ALittleDruid:
After set(CMAKE_REQUIRED_INCLUDES "${CURL_SOURCE_DIR}/include")

image

image

"D:\vcpkg\installed\x64-windows\include" is correct

https://github.com/microsoft/vcpkg/issues/27978#issue-1462189230

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1155,6 +1155,46 @@ if(HAVE_FILE_OFFSET_BITS)
 endif()
 check_type_size("off_t"  SIZEOF_OFF_T)
 
+# For Windows, all compilers used by CMake should support large files
+if(WIN32)
+  set(USE_WIN32_LARGE_FILES ON)
+
+  # Use the manifest embedded in the Windows Resource
+  set(CMAKE_RC_FLAGS "${CMAKE_RC_FLAGS} -DCURL_EMBED_MANIFEST")
+
+  # Check if crypto functions in wincrypt.h are actually available
+  if(HAVE_WINCRYPT_H)
+    check_symbol_exists(CryptAcquireContext "${CURL_INCLUDES}" USE_WINCRYPT)
+  endif()
+  if(USE_WINCRYPT)
+    set(USE_WIN32_CRYPTO ON)
+  endif()
+
+  # Link required libraries for USE_WIN32_CRYPTO or USE_SCHANNEL
+  if(USE_WIN32_CRYPTO OR USE_SCHANNEL)
+    list(APPEND CURL_LIBS "advapi32" "crypt32")
+  endif()
+
+  # Matching logic used for Curl_win32_random()
+  if(MINGW)
+    check_c_source_compiles("
+      #include <_mingw.h>
+      #if defined(__MINGW64_VERSION_MAJOR)
+      #error
+      #endif
+      int main(void) {
+        return 0;
+      }"
+      HAVE_MINGW_ORIGINAL)
+  endif()
+
+  if(NOT HAVE_MINGW_ORIGINAL)
+    list(APPEND CURL_LIBS "bcrypt")
+  else()
+    set(HAVE_FTRUNCATE OFF)
+  endif()
+endif()
+
 # include this header to get the type
 set(CMAKE_REQUIRED_INCLUDES "${CURL_SOURCE_DIR}/include")
 set(CMAKE_EXTRA_INCLUDE_FILES "curl/system.h")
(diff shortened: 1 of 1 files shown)


jay:
Can any dev familiar with cmake explain how CMAKE_REQUIRED_INCLUDES is supposed to work? I notice in CMakeLists.txt it is sometimes appended to as a list and other times cleared, as it is before getting the type of curl_off_t, which seems to have caused this bug. The reporter's fix to move the CryptAcquireContext check before the curl_off_t check looks like it masks a larger problem.

jzakrzewski:
It's a list of include directories to use for all the checks that involve running a compiler. In my own projects I tend to re-set it after each check to make sure I'm not clobbering the next one, and to make sure the stuff I'm looking for is where I expect it, and not in some random place.

ALittleDruid:
> 
https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html

check_symbol_exists(CryptAcquireContext "${CURL_INCLUDES}" USE_WINCRYPT)

"${CURL_INCLUDES}" is windows.h;ws2tcpip.h;winsock2.h;wincrypt.h;sys/stat.h;sys/types.h;sys/utime.h;fcntl.h;idn2.h;io.h;locale.h;setjmp.h;signal.h;stdbool.h;stdlib.h;string.h;time.h;stddef.h

File CheckSymbolExists.c:
/* */
#include <windows.h>
#include <ws2tcpip.h>
#include <winsock2.h>
#include <wincrypt.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <sys/utime.h>
#include <fcntl.h>
#include <idn2.h>
#include <io.h>
#include <locale.h>
#include <setjmp.h>
#include <signal.h>
#include <stdbool.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>
#include <stddef.h>

int main(int argc, char** argv)
{
  (void)argv;
#ifndef CryptAcquireContext
  return ((int*)(&CryptAcquireContext))[argc];
#else
  (void)argc;
  return 0;
#endif
}

Another way to modify it: change "${CURL_INCLUDES}" to "windows.h"

check_symbol_exists(CryptAcquireContext "windows.h" USE_WINCRYPT)

jzakrzewski:
`CryptAcquireContext` is documented to live in `wincrypt.h`, so I suppose this is where we should be looking for it, and not "everywhere".

jay:
I looked at your latest changes and I don't understand how this works for you now if CMAKE_REQUIRED_INCLUDES does not need to be reset. Don't you have the same problem, where the previously set CMAKE_REQUIRED_INCLUDES uses the wrong include directory?

ALittleDruid:
> I looked at your latest changes and I don't understand how this works for you now if CMAKE_REQUIRED_INCLUDES does not need to be reset. Don't you have the same problem, where the previously set CMAKE_REQUIRED_INCLUDES uses the wrong include directory?

check_symbol_exists(<symbol> <files> <variable>)
check_symbol_exists(CryptAcquireContext "${CURL_INCLUDES}" USE_WINCRYPT)
<files> is windows.h;ws2tcpip.h;winsock2.h;wincrypt.h;sys/stat.h;sys/types.h;sys/utime.h;fcntl.h;idn2.h;io.h;locale.h;setjmp.h;signal.h;stdbool.h;stdlib.h;string.h;time.h;stddef.h

CMAKE_REQUIRED_INCLUDES  is ${CURL_SOURCE_DIR}/include

cmake try compile CheckSymbolExists.c

`#include <windows.h>
#include <ws2tcpip.h>
#include <winsock2.h>
#include <wincrypt.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <sys/utime.h>
#include <fcntl.h>
#include <idn2.h>
#include <io.h>
#include <locale.h>
#include <setjmp.h>
#include <signal.h>
#include <stdbool.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>
#include <stddef.h>

int main(int argc, char** argv)
{
  (void)argv;
#ifndef CryptAcquireContext
  return ((int*)(&CryptAcquireContext))[argc];
#else
  (void)argc;
  return 0;
#endif
}`

CheckSymbolExists.c include unnecessary header files and idn2.h not found in ${CURL_SOURCE_DIR}/include

The latest changes only include windows.h and wincrypt.h

jay:
> CheckSymbolExists.c include unnecessary header files and idn2.h not found in ${CURL_SOURCE_DIR}/include

So in that case is the reason the test fails that idn2.h isn't found?

ALittleDruid:
> > CheckSymbolExists.c include unnecessary header files and idn2.h not found in ${CURL_SOURCE_DIR}/include
> 
> So in that case is the reason the test fails that idn2.h isn't found?

Yes

Befor `set(CMAKE_REQUIRED_INCLUDES "${CURL_SOURCE_DIR}/include")`, CMAKE_REQUIRED_INCLUDES  is D:\vcpkg\installed\x64-windows\include and idn2.h can be found in that path

That's enough to check CryptAcquireContext include `windows.h` and `wincrypt.h `, the default include path is able to find these header files

jay:
Thanks

PR #10359, opened 2023-01-29
openssl: only use the CA_BLOB if verifypeer is set

### LINKED ISSUE #10351: Empty CURLOPT_CAINFO_BLOB causes error
### I did this

@pgroke-dt [reported](https://curl.se/mail/lib-2023-01/0070.html) on the mailing list that, for curl 7.87.0 w/ OpenSSL backend, setting CURLOPT_CAINFO_BLOB to an empty string causes an error when peer verification is disabled.

The CA blob import function expects "[one or more certificates](https://curl.se/libcurl/c/CURLOPT_CAINFO_BLOB.html)" and returns an error if no certificate is read:

https://github.com/curl/curl/blob/c12fb3ddaf48e709a7a4deaa55ec485e4df163ee/lib/vtls/openssl.c#L3077-L3078

In 7.86.0 ossl_connect_step1 calls load_cacert_from_memory directly and ignores the result if peer verification is disabled. Since then, certificate caching support was added in 3c16697e. Now in 7.87.0 ossl_connect_step1 calls Curl_ssl_setup_x509_store which calls load_cacert_from_memory, but does not ignore the result.

### I expected the following

My first impression is this isn't a bug however AFAICT all other SSL backends that support CA blob either skip loading or ignore a bad load when peer verification is disabled.

### curl/libcurl version

curl 7.87.1-DEV (i386-pc-win32) libcurl/7.87.1-DEV OpenSSL/1.1.1s nghttp2/1.50.0

<br>

/cc @gvollant @tlsa @icing

bagder:
Reported-by: Paul Groke
Bug: https://curl.se/mail/lib-2023-01/0070.html
Fixes #10351

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -3078,146 +3078,148 @@ static CURLcode populate_x509_store(struct Curl_cfilter *cf,
 
   if(verifypeer) {
 #if defined(USE_WIN32_CRYPTO)
-  /* Import certificates from the Windows root certificate store if requested.
-     https://stackoverflow.com/questions/9507184/
-     https://github.com/d3x0r/SACK/blob/master/src/netlib/ssl_layer.c#L1037
-     https://datatracker.ietf.org/doc/html/rfc5280 */
-  if(ssl_config->native_ca_store) {
-    HCERTSTORE hStore = CertOpenSystemStore(0, TEXT("ROOT"));
+    /* Import certificates from the Windows root certificate store if
+       requested.
+       https://stackoverflow.com/questions/9507184/
+       https://github.com/d3x0r/SACK/blob/master/src/netlib/ssl_layer.c#L1037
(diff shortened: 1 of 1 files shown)


lib/vtls/openssl.c
```diff
@@ -3076,218 +3076,214 @@ static CURLcode populate_x509_store(struct Curl_cfilter *cf,
...
 
-      if(result)
-        return result;
+        if(result)
+          return result;
+      }
+      if(imported_native_ca)
+        infof(data, "successfully imported Windows CA store");
+      else
+        infof(data, "failed importing Windows CA store");
```
jay, line 3220: IMO the old message is better and consistent. We use the same 'error, continuing anyway' format for both ca file and path.

bagder: I always disliked that way of phrasing it, but alright I'll restore it!

bagder:
Yeah, no appveyor fails now! :smile:

PR #10360, opened 2023-01-29
openssl: adapt to boringssl's error code type

bagder:
BoringSSL uses uint32_t, OpenSSL uses 'unsigned 'long'


bagder:
Unclear why the GHA job doesn't show these warnings while the zuul job did...

PR #10362, opened 2023-01-29
vtls: remove unnecessary null check

### LINKED ISSUE #10361: vtls: dereference before null check
This is a nit [found by coverity](https://scan5.scan.coverity.com/reports.htm#v16981/p10206/fileInstanceId=232562788&defectInstanceId=53088090&mergedDefectId=1520659&eventId=53088090-1) 

The code checks that `cf->ctx` is non-NULL on this line:

https://github.com/curl/curl/blob/c5987e1430bc01ab8c3a3aa6274d57d22557ca8e/lib/vtls/vtls.c#L1553

... but on the preceding line, the use of the `CF_DATA_SAVE()` macro has already dereferenced  `cf->ctx`. So either the NULL-check is unnecessary and we can remove it, or it is necessary and we need to move it it up a little.

icing:
Do not null check when we already assume cf-ctx exists. Fixes #10361

### DIFF
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -1550,7 +1550,7 @@ static bool ssl_cf_data_pending(struct Curl_cfilter *cf,
   bool result;
 
   CF_DATA_SAVE(save, cf, data);
-  if(cf->ctx && Curl_ssl->data_pending(cf, data))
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #10364, opened 2023-01-29
Fix symbols missing when static linking ngtcp2,nghttp3 and nghttp2 using cmake and MSVC on Windows.

### LINKED ISSUE #10363: Unresolved symbols when static linking ngtcp2,nghttp3,nghttp2 using cmake and MSVC on Windows.
### I did this

+ Building static libraries of https://github.com/quictls/openssl.
+ Building static libraries of ngtcp2,nghttp3 and nghttp2 with openssl with quic above.
+ Uisng cmake to build curl with `cmake -G 'Visual Studio 17 2022' -A x64 '-DUSE_NGTCP2=ON' '-DUSE_NGHTTP2=ON' '-DCMAKE_PREFIX_PATH=<where to install openssl, ngtcp2, ngtcp3, nghttp2>' '-DBUILD_SHARED_LIBS=OFF' '-DCMAKE_POSITION_INDEPENDENT_CODE=ON' '-DBUILD_TESTING=OFF'  '-DCURL_USE_OPENSSL=ON' '-DOPENSSL_ROOT_DIR=<where to install openssl>' ...`.

### I expected the following

Building and install success. but got link error with some unresolved symbols of these libraries and also some system library symbols depended by openssl.

### curl/libcurl version

7.87.0

### operating system

Windows

### Additional

I found nghttp2,nghttp3 and ngtcp2 use `NGHTTP2_STATICLIB` , `NGHTTP3_STATICLIB` and `NGTCP2_STATICLIB` to tell compiler to use static library, but the find script do not check them.

owent:
Fixes #10363 

+ Try to detect if we can use static library of nghttp2,nghttp3 and ngtcp2.
  + Define `NGHTTP3_STATICLIB`, `NGTCP2_STATICLIB`, `NGHTTP2_STATICLIB` as needed。
+ Add some common libraries needed by openssl.


CMakeLists.txt
```diff
@@ -535,6 +539,61 @@ if(USE_NGTCP2)
...
+    if(NOT HAVE_SIZEOF_SSIZE_T)
+      check_type_size("long" SIZEOF_LONG)
+      check_type_size("__int64" SIZEOF___INT64)
+      if(SIZEOF_LONG EQUAL SIZEOF_SIZE_T)
+        set(CMAKE_REQUIRED_DEFINITIONS "-Dssize_t=long")
+      elseif(SIZEOF___INT64 EQUAL SIZEOF_SIZE_T)
+        set(CMAKE_REQUIRED_DEFINITIONS "-Dssize_t=__int64")
+      else()
+        set(CMAKE_REQUIRED_DEFINITIONS "-Dssize_t=long long")
+      endif()
```
bagder, line 585: This type defines seem unrelated to a static/dynamic build. Why do you need those?

owent: nghttp2 use `ssize_t` but not define it.

owent: The failed jobs seem not a problem of this PR, could you please help me to find out what's the problem?

bagder: > nghttp2 use `ssize_t` but not define it.

We've built curl with nghttp2 for many years already without having to add any extra defines. Why does the need start now? And if we actually do need them, they are still unrelated to the static builds and should not be done in the same PR that fixes the linking of static libs.

owent: `ssize_t` is already defined before by https://github.com/curl/curl/blob/curl-7_87_0/CMakeLists.txt#L1013-L1026 and https://github.com/curl/curl/blob/curl-7_87_0/lib/curl_config.h.cmake#L776 . It's available for all the sources of curl , but unavailable for `check_symbol_exists()`. I tried to add the same definitions to `CMAKE_REQUIRED_DEFINITIONS` just to make sure `check_symbol_exists(nghttp2_version "nghttp2/nghttp2.h" HAVE_NGHTTP2_DYNAMICLIB)` and `check_symbol_exists(nghttp2_version "nghttp2/nghttp2.h" HAVE_NGHTTP2_STATICLIB)` can work correctly, this setting do not take any effect on curl building.

bagder: thanks for explaining, I think I understand now. It is necessary for this "hack" to work.

I will leave the judgement for if this is worth merging to someone who better understands and reads cmake.

owent: > thanks for explaining, I think I understand now. It is necessary for this "hack" to work.
> 
> I will leave the judgement for if this is worth merging to someone who better understands and reads cmake.

Thanks a lot.

bagder:
Sorry, I see no votes or voices in favor of this PR. Maybe because not many people build static on Windows with cmake.

PR #10366, opened 2023-01-30
WebSocket: fix bad assert, unstick shutdown

bagder:


lib/ws.c
```diff
@@ -479,7 +479,6 @@ CURL_EXTERN CURLcode curl_ws_recv(struct Curl_easy *data, void *buffer,
 
       /* update buffer and frame info */
       wsp->frame.offset += datalen;
-      DEBUGASSERT(wsp->frame.bytesleft >= (curl_off_t)datalen);
```
dfandrich, line 482: I don't think this assert is bad by itself, as long as the following two lines is there. The assert protects bytesleft from ever going negative in the subtraction two lines down. With the assert gone, it's possible for bytesleft to go negative which I don't think is the intention but shows that the assumptions are wrong somewhere else.

bagder: For this reproducer case, **bytelength** was zero at this point, so the assert() was clearly not doing a good at job that position at least. But sure, maybe it would work better in the conditional block just before the subtraction is done.

We can always add more asserts when we go about and continue to debug this functionality.

PR #10367, opened 2023-01-30
http2: no http2 debug by default

bagder:
Follow-up to 71b7e01610

Bug: https://curl.se/mail/archive-2023-01/0034.html


icing:
Agree to that. In #10349 comes the `LOG_CF` use so that debug logging can be enabled via the env var `CURL_DEBUG`.

bagder:
Ah right, which succeeds this. I can just hold off this for a while instead and things work out when #10349 goes in.,

PR #10368, opened 2023-01-30
zuul: stop using this CI service

bagder:
The important jobs have already transitioned. The remaining ones we can skip for now.


dfandrich:
Don't forget `scripts/cijobs.pl`

bagder:
I left it there on purpose for now as it is harmless and just won't find any zuul jobs now. But yes, I'll clean that one up as well.

PR #10372, opened 2023-01-30
[WIP] curl: add --progress-percent to display progress as percentage

IGLOU-EU:
Add an option to display only the percentage.    
The goal is to make it easier to use curl with shell scripts.

Like this
```bash
curl --progress-percent http://speed.hetzner.de/1GB.bin -o /dev/null 2>&1 | zenity --progress --title "Power of curl"
```

I saw that it is necessary to add the curl version at the introduction, but I do not know what to put :face_in_clouds:


dfandrich:
The progress display already shows the percentage. What is the use case where
that's all you'd want to see?

IGLOU-EU:
The actual progress display was absolutely not scripting friendly. You need to use tr, sed or awk.
And you can find a lot of thread like this one https://community.unix.com/t/function-to-run-a-progress-bar-in-zenity/

So the goal is to allow using curl directly and only see the percentage.

dfandrich:
And what's the problem with using those utilities? They're there to do this
kind of thing, after all.

The thing about adding --progress-percent is that tomorrow someone is going to
want --progress-total to get only the total amount transferred instead and the
next day --progress-received to get only the bytes received and then
--progress-percent-and-size to get the percentage but also the total number of
bytes and so on. And each option has a cost beyond development in ongoing
testing and maintenance and binary size, which are harder to justify when there
are other ways of accomplishing the task.

IMHO, if it's deemed the user experience is sufficiently improved that
the cost of something like this is worth bearing, a scheme like --write-out
that lets the user customize the progress bar to provide exactly what's needed
would be more future-proof and make scripting easier for more people.

bagder:
Amen to what @dfandrich says.

IGLOU-EU:
Interesting and true! I need some time, but I'll look into how 'write-out' works. Can I keep this PR open in WIP ?

jay:
Why isn't -# sufficient? I don't see a good reason to add this to curl.

IGLOU-EU:
As I said before, it avoids the use of other utilities and pipes, thus saving the launching of other processes and buffering.   
Another interesting thing, @dfandrich mentioned the fact that you can format your output as you want it to be.

And for me, it is more readable, simpler and more effective to do something like this
```bash
curl --myformat myurl -o /dev/null 2>&1 | zenity --progress --title "Power of curl"
```

Instead of this
```bash
curl -# myurl -o /dev/null 2>&1 | stdbuf -oL tr '\r' '\n' | grep -o '[0-9]*\.[0-9]'  | zenity --progress --title "Power of curl
```

jay:
I don't see many people using it when we already have a progress bar.

jay:
I suggest take this to the mailing list for further discussion.

adrelanos:
* https://github.com/curl/curl/discussions/12306

miaLDN:
Eye opening to see pure emotive performance for their gatekeeping.

PR #10374, opened 2023-01-30
Docs: Fix argument type of `curl_free`

### LINKED ISSUE #10373: curl_free man page: `curl_free` accepts `void*` instead of `char*`
The [documentation of `curl_free`](https://curl.se/libcurl/c/curl_free.html) says that it takes `char*` as its argument whereas it actually takes `void*`:

https://github.com/curl/curl/blob/be5f6c31247ff24db76e6ffbe49a1aa8ab3f9f5b/include/curl/curl.h#L2678

If you followed the documentation you would get an error in the following case:

`std::unique_ptr<char, void(*)(char*)> res {curl_easy_escape(nullptr, str.data(), static_cast<int>(str.size())), curl_free};`

Thomas1664:
Fixes #10373

### DIFF
--- a/docs/libcurl/curl_free.3
+++ b/docs/libcurl/curl_free.3
@@ -28,7 +28,7 @@ curl_free - reclaim memory that has been obtained through a libcurl call
 .nf
 #include <curl/curl.h>
 
-void curl_free(char *ptr);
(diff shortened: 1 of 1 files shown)


bagder:
thanks!

Issue #10356, opened 2023-01-28
cmake builds with gcc-11 on macOS now permafail


bagder:
These builds seem to have started to fail on Jan 21, 2023. Before that the cmake builds failed because of homebrew problems but on the 21st the problem changed to seemingly only not work for gcc-11 on mac. Before the homebrew problems (Jan 8),  those builds worked fine in the CI.

The three builds are known as

- **macOS / cmake gcc-11 OpenSSL**
- **macOS / cmake gcc-11 LibreSSL**
- **macOS / cmake gcc-11 libssh2**

Example build failure ([from here](https://github.com/curl/curl/actions/runs/3976406714/jobs/6816875496)):

~~~
[ 28%] Building C object lib/CMakeFiles/libcurl.dir/hmac.c.o
[ 28%] Building C object lib/CMakeFiles/libcurl.dir/hostasyn.c.o
[ 28%] Building C object lib/CMakeFiles/libcurl.dir/hostip.c.o
In file included from /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/dispatch.h:28,
                 from /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFStream.h:20,
                 from /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFPropertyList.h:17,
                 from /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CoreFoundation.h:66,
                 from /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/System/Library/Frameworks/SystemConfiguration.framework/Headers/SCDynamicStoreCopySpecific.h:30,
                 from /Users/runner/work/curl/curl/lib/hostip.c:70:
[...]

PR #10381, opened 2023-01-31
macos.yml: revert to Xcode 14.0.1 to avoid cmake/gcc-11 build errors

pheiduck:
Possible fix for: https://github.com/curl/curl/issues/10356

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -23,6 +23,9 @@ concurrency:
 
 permissions: {}
 
+env:
+  DEVELOPER_DIR: /Applications/Xcode_14.0.1.app/Contents/Developer
+
 jobs:
   autotools:
     name: ${{ matrix.build.name }}


jay:
Thanks, that seems to have done the trick. Is there something wrong with Xcode 14.2?

pheiduck:
> Thanks, that seems to have done the trick. Is there something wrong with Xcode 14.2?
I am currently still not sure if xcode is really the sloppy one for the error. Maybe it is also a problem in connection with gcc-11?
I will try it with gcc-12.

pheiduck:
Okay doesn't matter. It's xcode related

bagder:
I feel inclined to merge this to get away from the annoyances of https://github.com/curl/curl/issues/10356 

@jay @icing what do you think?

jay:
I'm fine with it. I think we can assume for now this is an xcode issue and try the next version when it's released as @pheiduck suggested.

bagder:
Thanks!

pheiduck:
@bagder xcode 14.3 is still in beta: https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes

so we can revert this PR, when GH CI is bumped to macOS 13 w/ xcode 14.3

PR #10385, opened 2023-02-01
urlapi: skip path checks if path is just "/"

bagder:
As a miniscule optimization, treat a path of the length 1 as the same as non-existing, as it can only be a single leading slash, and that's what we do for no paths as well.

### DIFF
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -1200,9 +1200,10 @@ static CURLUcode parseurl(const char *url, CURLU *u, unsigned int flags)
     path = u->path = Curl_dyn_ptr(&enc);
   }
 
-  if(!pathlen) {
-    /* there is no path left, unset */
+  if(pathlen <= 1) {
+    /* there is no path left or just the slash, unset */
     path = NULL;
+    pathlen = 0;
   }
   else {
     if(!u->path) {


lib/urlapi.c
```diff
@@ -1200,9 +1200,10 @@ static CURLUcode parseurl(const char *url, CURLU *u, unsigned int flags)
     path = u->path = Curl_dyn_ptr(&enc);
   }
 
-  if(!pathlen) {
-    /* there is no path left, unset */
+  if(pathlen <= 1) {
+    /* there is no path left or just the slash, unset */
     path = NULL;
+    pathlen = 0;
```
kvarec, line 1206: From clang static analysis:
```
urlapi.c:1206:5: warning: Value stored to 'pathlen' is never read [deadcode.DeadStores]
    pathlen = 0;
    ^         ~
```
https://dev.azure.com/daniel0244/curl/_build/results?buildId=13892&view=logs&j=e2e9a903-bd95-5516-8620-b8f4e5800a42&t=b96a1c9b-30a9-56e9-62cc-d046c220147f

bagder: ah, look at that. Thanks!

PR #10386, opened 2023-02-01
CODEOWNERS: remove the peeps mentioned as CI owners

bagder:
These owners do not have the bandwidth/energy to do the reviews which makes PRs stall and this ownership claim flawed.

Follow-up to c04c78ac87c4d46737934345a

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,13 +1,3 @@
 # Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
 #
 # SPDX-License-Identifier: curl
-
-# Code owners for CI configuration
-## GitHub Actions
-.github/workflows/ @cmeister2 @mback2k
-## AppVeyor
-appveyor.yml @MarcelRaad @mback2k
-tests/appveyor.pm @mback2k
-## Azure Pipelines
-.azure-pipelines.yml @cmeister2 @mback2k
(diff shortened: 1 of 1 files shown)


cmeister2:
Hmm. I wouldn't say that I don't have the energy here; I feel like I lack authority to be approving changes. There's also changes where I don't feel as though I'm an expert, and so don't comment.

I'm unsure how you want to proceed. If you feel like this experiment hadn't borne the fruit we hoped for, then it's your project: happy to approve this. Alternatively, happy to have this as a venue for discussion on adaptation?

bagder:
I don't think asking for a review is saying you need to have a mandate to do anything in particular other than to give your opinion and "judgement" about a proposed change, and perhaps in particular the way it is done.

The effect we have now is that we assign CI related PRs for reviews that then just sit there without anyone actually doing the reviews (and with two people already assigned the job, it deters others to do it since it looks like the job is already "taken"). I think it looks bad as well when eventually just override the non-provided reviews and go ahead and review it myself or just merge it with one or two pending reviews.

I don't think this process helps; I think it hurts.

For auto-assigning reviews, I would expect that the asignees at least usually fulfills the task as otherwise why do it automatically in the first place?

cmeister2 (approved):

cmeister2:
Fair point. I've approved.

PR #10390, opened 2023-02-01
[docs] Fix memory leak in example for `CURLOPT_WRITE_FUNCTION`

### LINKED ISSUE #10387: CURLOPT_WRITEFUNCTION man page: Memory leak in example
The callback in the example for [`CURLOPT_WRITEFUNCTION`](https://curl.se/libcurl/c/CURLOPT_WRITEFUNCTION.html) calls `realloc` but the memory is never freed in the calling code. Although it should be obvious I think it would be better to remind people that the buffer inside `memory` needs to be freed after using it.

I'd make a patch to fix this if you want.

Thomas1664:
Fixes #10387

### DIFF
--- a/docs/libcurl/opts/CURLOPT_WRITEFUNCTION.3
+++ b/docs/libcurl/opts/CURLOPT_WRITEFUNCTION.3
@@ -103,12 +103,16 @@ For all protocols
  }
 
  struct memory chunk = {0};
+ CURLcode res;
 
  /* send all data to this function  */
  curl_easy_setopt(curl_handle, CURLOPT_WRITEFUNCTION, cb);
 
  /* we pass our 'chunk' struct to the callback function */
  curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, (void *)&chunk);
+
+ /* send a request */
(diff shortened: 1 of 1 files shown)


docs/libcurl/opts/CURLOPT_WRITEFUNCTION.3
```diff
@@ -109,6 +109,9 @@ For all protocols
 
  /* we pass our 'chunk' struct to the callback function */
  curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, (void *)&chunk);
+
+ /* remember to free the buffer */
+ free(chunk.response)
```
bagder, line 114: Well, at this point in the code this `free()` call will do no good since it hasn't had the chance to allocate anything yet!

Shouldn't you then also perhaps invoke curl_easy_perform() before the free is called?

Thomas1664: Yes. I also fixed all remaining errors, i.e. `curl_handle` coming out of nowhere.

bagder (approved):

bagder:
Thanks!

PR #10393, opened 2023-02-01
GHA: cache libressl build output

andy5995:


pheiduck:
@andy5995 Can you explain why this is necessary?

andy5995:
> @andy5995 Can you explain why this is necessary?

@pheiduck Hi. I wouldn't say it's necessary, but downloading a single file seems more efficient than cloning the repo.

I'm actually looking into implementing [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) for this PR. So even that archive wouldn't have to be downloaded each time...

andy5995:
> I'm actually looking into implementing [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) for this PR. So even that archive wouldn't have to be downloaded each time...

Now I'm wondering if [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) would we be a better way to go.

Seems that instead of caching the libressl archive, it would be better to cache the build output for future use. What do you folks think?

I'd better first practice on one of my own repos though... This could be a lot of trial and error for me.

pheiduck:
Okay, let's wait and see what @bagder says about this, I know some maintainers (not of curl) don't like to cache dependencies as this doesn't necessarily speed up deploying.

bagder:
Also note that this fails the REUSE job because the added file does not contain copyright information. You can also add an ignore rule for it in `.reuse/dep5`.

andy5995:
> git clone or curl download within GHA from github itself I figure are likely going to be very snappy operations either way. Is there a measurable difference?

No there isn't. At least not by eyeballing the transfer. If I used `time` it might be measured, but I doubt it would be significant.

> 
> Caching would certainly be interesting, if it can key it to perhaps the version number or something so that it gets flushed when we bump it in the future.

In this case, with the cache enabled, it's using the sha256sum to determine if the cache needs updating.

Three questions: 

1 .Do you want me to update the libressl version in this PR? There was a new release in December.

2. for the libressl build, should I change `make install` to `make install-strip`?

3. Do you want me to change back to cloning the directory? We can hash some different file if you prefer. A text file with the version name should be just as adequate as the sha256sum for the hashing.

btw, the caching seems to be working:

image

pheiduck:
>1. Do you want me to update the libressl version in this PR? There was a new release in December. https://github.com/libressl/portable/tags

At least to the release from 02/07 **3.6.2**

bagder:
Let's try this. Can you please squash this PR into a more sensible number of commits?

andy5995:
The patch included with the libressl portable 3.6.2 release is currently broken due to a recent commit on one of it's subprojects:

https://github.com/libressl/openbsd/commit/fd731cbcd2764243ebbfc65db13bac00fcaa1839#diff-308090c74cb589fc9898236d8186238a33c38587ce2f2919315b5144a24a0cbbR369

This PR is now failing (https://github.com/curl/curl/actions/runs/4268376318/jobs/7430823941#step:7:65) when the patch is attempted (as part of `./autogen.sh` from the libressl package).

andy5995:
I switched to 3.7.0

andy5995:
The 3.7.0 failed also.

I was mistakenly getting the libressl archives from https://github.com/libressl/portable/tags 

I'm now using a mirror, as mentioned in the [libressl README](https://github.com/libressl/portable#compatibility-with-other-operating-systems). Seems to be working now.

andy5995:
I accidentally screwed up this branch. I'll try to fix it.

mback2k:
I cannot follow the changes proposed with this PR and it seems like the summary of changes doesn't include the actual intent anymore: https://github.com/curl/curl/pull/10393/files

andy5995:
I updated the title.

.github/workflows/linux.yml
```diff
@@ -170,10 +170,22 @@ jobs:
...
 
     - if: ${{ contains(matrix.build.install_steps, 'libressl') }}
+      run: echo "LIBRESSL_VER=3.6.1" >> $GITHUB_ENV
+
+    - if: ${{ contains(matrix.build.install_steps, 'libressl') }}
+      id: cache-libressl
+      uses: actions/cache@v3
+      with:
+        path: /home/runner/libressl
+        key: ${{ runner.os }}-build-${{ matrix.build.name }}-${{ env.LIBRESSL_VER }}
```
andy5995, line 180: When ${{ env.LIBRESSL_VER }} changes, cache.hit will == false

andy5995:
Here in the workflow you can see where the cache is ultimately saved:

image

andy5995:
I rebased and pushed. Here it restores the cache and skips the "install libressl" step.

image

bagder:
I would like feedback from @cmeister2 or @mback2k on this.

Caching seems like a good thing to add, but it also makes the yaml more advanced and I cannot say that I myself understand how this works so there's an additional risk of future errors and mess-ups. But **I** don't have to fully understand everything as long as others are onboard with the changes.

mback2k:
My question is: maybe there is a better way to reduce the YAML clutter/noise this adds just for one dependency by moving the logic into a custom GitHub script action that handles both execution of something and caching of the output. Because in the end I guess it would make sense to introduce this caching to all the dependencies, right?

andy5995:
> My question is: maybe there is a better way to reduce the YAML clutter/noise this adds just for one dependency by moving the logic into a custom GitHub script action that handles both execution of something and caching of the output. Because in the end I guess it would make sense to introduce this caching to all the dependencies, right?

@mback2k Right.

I don't know of a way to reduce the clutter/noise (other than [what I just changed](https://github.com/curl/curl/pull/10393/commits/c73b96429f8dae76022c26b86901a2193be1a483)). Maybe someone will make a PR for my PR to introduce some fancy new technique.

PR #10395, opened 2023-02-02
Add 'since version' to %{header_json} documentation

mdaloia:
The documentation of `%{header_json}` missed to mention since which version this variable for `--write-out` is present.

Based on commit https://github.com/curl/curl/commit/4133a69f2daa476bb6d902687f1dd6660ea9c3c5 we can determine from the tags were this commit is present that the first version to include it was `7.83.0`.
This could be also checked with: `git tag --contains 4133a69f2daa476bb6d902687f1dd6660ea9c3c5`


bagder (approved):

bagder:
Thanks!

PR #10397, opened 2023-02-02
doh: Optimization for disabled ENABLE_IPV6

kvarec:
For disabled IPv6 a condition (conn->ip_version != CURL_IPRESOLVE_V4) is always false.

### DIFF
--- a/lib/doh.c
+++ b/lib/doh.c
@@ -396,6 +396,7 @@ struct Curl_addrinfo *Curl_doh(struct Curl_easy *data,
     goto error;
   dohp->pending++;
 
+#ifdef ENABLE_IPV6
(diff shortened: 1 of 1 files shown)


bagder:
The title says "Code is unusable without ENABLE_IPV6" but it seems this is just an optimization?

kvarec:
> The title says "Code is unusable without ENABLE_IPV6" but it seems this is just an optimization?

kvarec:
> The title says "Code is unusable without ENABLE_IPV6" but it seems this is just an optimization?

fixed

bagder (approved):

bagder:
Thanks!

PR #10398, opened 2023-02-02
fopen: tempname is already NULL

kvarec:
[CWE-1164] V1048: The '* tempname' variable was assigned the same value.
https://pvs-studio.com/en/docs/warnings/v1048/


bagder (approved):

jay:
Thanks

PR #10399, opened 2023-02-02
http: fix CWE-570

kvarec:
[CWE-570] V560: A part of conditional expression is always false: conn->bits.authneg.
[CWE-570] V560: A part of conditional expression is always false: conn->handler->protocol & (0 | 0).

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -2396,8 +2396,7 @@ CURLcode Curl_http_bodysend(struct Curl_easy *data, struct connectdata *conn,
        we don't upload data chunked, as RFC2616 forbids us to set both
        kinds of headers (Transfer-Encoding: chunked and Content-Length) */
     if(http->postsize != -1 && !data->req.upload_chunky &&
-       (conn->bits.authneg ||
-        !Curl_checkheaders(data, STRCONST("Content-Length")))) {
+       (!Curl_checkheaders(data, STRCONST("Content-Length")))) {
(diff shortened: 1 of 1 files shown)


lib/http.c
```diff
@@ -2396,8 +2396,7 @@ CURLcode Curl_http_bodysend(struct Curl_easy *data, struct connectdata *conn,
        we don't upload data chunked, as RFC2616 forbids us to set both
        kinds of headers (Transfer-Encoding: chunked and Content-Length) */
     if(http->postsize != -1 && !data->req.upload_chunky &&
-       (conn->bits.authneg ||
-        !Curl_checkheaders(data, STRCONST("Content-Length")))) {
+       (!Curl_checkheaders(data, STRCONST("Content-Length")))) {
```
bagder, line 2399: Please explain why removing `conn->bits.authneg` from here is the correct move

bagder:
Also, what does the title mean? how does this "fix CWE-570" ?

lib/http.c
kvarec: conn->bits.authneg is always false here, because in the block above, with conn->bits.authneg == true, we do not get here

bagder: Ah, yes of course. Thanks!

bagder (approved):

bagder:
Thanks!

PR #10401, opened 2023-02-02
setopt: optimization for WIN32

kvarec:
[CWE-570] V547: Expression 'arg > 2147483647' is always false.
https://pvs-studio.com/en/docs/warnings/v547/

[CWE-1164] V1048: The 'uarg' variable was assigned the same value.
https://pvs-studio.com/en/docs/warnings/v1048/

### DIFF
--- a/lib/setopt.c
+++ b/lib/setopt.c
@@ -199,8 +199,10 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
     arg = va_arg(param, long);
     if(arg < -1)
       return CURLE_BAD_FUNCTION_ARGUMENT;
+#if SIZEOF_LONG > 4
     else if(arg > INT_MAX)
       arg = INT_MAX;
+#endif
 
     data->set.dns_cache_timeout = (int)arg;
     break;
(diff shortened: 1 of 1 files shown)


jay:
> [CWE-570] V547: Expression 'arg > 2147483647' is always false.

In my opinion that is too pedantic and I don't think it's worth it to add guards around every expression that causes it.

bagder:
I concur with @jay here. The compiler can even optimize away these statements by itself.

kvarec:
> I concur with @jay here. The compiler can even optimize away these statements by itself.

But already there is code with this macro check: https://github.com/curl/curl/blob/master/lib/setopt.c#L2631
And here:
```c
if(uarg >= UINT_MAX)
      uarg = UINT_MAX;
```
if `uarg == UINT_MAX`, why `uarg = UINT_MAX`?

jay:
> ```c
> if(uarg >= UINT_MAX)
>       uarg = UINT_MAX;
> ```
> if `uarg == UINT_MAX`, why `uarg = UINT_MAX`?

Ok that change is fine. 

> But already there is code with this macro check: https://github.com/curl/curl/blob/master/lib/setopt.c#L2631

https://github.com/curl/curl/blob/c12fb3ddaf48e709a7a4deaa55ec485e4df163ee/lib/setopt.c#L2605-L2619

I think that is also unnecessary. I don't think we have any analyzers that would warn on it.

/cc @tmilburn

bagder:
I made #10421 to fix the `>=` checks into `>`.

PR #10402, opened 2023-02-02
strtoofft: optimization with macro ISSPACEBASE

kvarec:
[CWE-570] V560: A part of conditional expression is always false: ((* str) == ' ').
[CWE-570] V560: A part of conditional expression is always false: ((* str) == '\t').

### DIFF
--- a/lib/strtoofft.c
+++ b/lib/strtoofft.c
@@ -225,7 +225,7 @@ CURLofft curlx_strtoofft(const char *str, char **endp, int base,
 
   while(*str && ISBLANK(*str))
     str++;
-  if(('-' == *str) || (ISSPACE(*str))) {
+  if(('-' == *str) || (ISSPACEBASE(*str))) {
     if(endp)
       *endp = (char *)str; /* didn't actually move */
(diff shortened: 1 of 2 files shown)


lib/strtoofft.c
```diff
@@ -225,7 +225,7 @@ CURLofft curlx_strtoofft(const char *str, char **endp, int base,
 
   while(*str && ISBLANK(*str))
     str++;
-  if(('-' == *str) || (ISSPACE(*str))) {
+  if(('-' == *str) || (ISSPACEBASE(*str))) {
```
bagder, line 228: I would rather suggest:
~~~c
if(!ISXDIGIT(x))
 ...

kvarec: done

kvarec: Instead of `if(('-' == *str) || (ISSPACEBASE(*str)))`?

bagder: Or maybe just leave it? This `ISSPACEBASE` does not increase readability.

bagder (changes requested):
This doesn't build

PR #10407, opened 2023-02-03
url: fix PVS warning V560

kvarec:
[CWE-571] V560: A part of conditional expression is always true: userp.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -2275,7 +2275,7 @@ static CURLcode parse_proxy(struct Curl_easy *data,
       goto error;
     }
     /* path will be "/", if no path was found */
-    if(strcmp("/", path)) {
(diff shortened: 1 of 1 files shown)


lib/url.c
```diff
@@ -2275,7 +2275,7 @@ static CURLcode parse_proxy(struct Curl_easy *data,
       goto error;
     }
     /* path will be "/", if no path was found */
-    if(strcmp("/", path)) {
+    if(strcmp("/", path) != 0) {
```
bagder, line 2278: No thanks, we compare against non-zero implicitly all through curl code on purpose. I don't think this increases readability.

kvarec: Ok. I canceled this change.

bagder (approved):

bagder:
Thanks!

PR #10408, opened 2023-02-03
urlapi: fix PVS warning V560

kvarec:
[CWE-571] V560: A part of conditional expression is always true: qlen.

### DIFF
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -1102,10 +1102,10 @@ static CURLUcode parseurl(const char *url, CURLU *u, unsigned int flags)
     hostp = p; /* host name starts here */
 
     /* find the end of the host name + port number */
-    while(p && *p && !HOSTNAME_END(*p))
+    while(*p && !HOSTNAME_END(*p))
       p++;
 
-    len = p ? p - hostp : 0;
+    len = p - hostp;
     if(len) {
       if(Curl_dyn_addn(&host, hostp, len)) {
         result = CURLUE_OUT_OF_MEMORY;
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -1102,10 +1102,10 @@ static CURLUcode parseurl(const char *url, CURLU *u, unsigned int flags)
     hostp = p; /* host name starts here */
 
     /* find the end of the host name + port number */
-    while(*p && !HOSTNAME_END(*p))
+    while(p && *p && !HOSTNAME_END(*p))
       p++;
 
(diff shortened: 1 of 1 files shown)


lib/urlapi.c
```diff
@@ -1102,10 +1102,10 @@ static CURLUcode parseurl(const char *url, CURLU *u, unsigned int flags)
     hostp = p; /* host name starts here */
 
     /* find the end of the host name + port number */
-    while(*p && !HOSTNAME_END(*p))
+    while(p && *p && !HOSTNAME_END(*p))
```
bagder, line 1105: How can `p` be NULL here?

kvarec: If schemelen == false, p can be not initialized...

kvarec: Maybe need also 'const char *p = NULL;'?

bagder: Isn't `p` then assigned on line 1100?

kvarec: Oh, see...
But url can be NULL? On release build checking `DEBUGASSERT(url)` is dummy and potential p can be NULL?

bagder: `url` should not be possible to be NULL as input, the assert is there to make sure it dies if that would ever happen anyway.

kvarec:
```c
#if defined(DEBUGBUILD)
#define DEBUGASSERT(x) assert(x)
#else
#define DEBUGASSERT(x) do { } while(0)
#endif
```
assert enabled only for DEBUGBUILD, and on release build if potentially `url == NULL` the application dies, or not?

bagder: Then how does `url` end up NULL  there?

kvarec: Someone will write a code with an bug that will result in `parseurl` call with `url == NULL`.

bagder: The ASSERT is there to catch such possible future mistake (in tests). `parseurl()` is only called from within a single function: `parseurl_and_replace()`.

`parseurl_and_replace()` is called on two places and the URL argument cannot be NULL in either invocation.

If `parseurl()` needs to handle a NULL url then that check should be added explicitly in the beginning of the function. But I don't think it needs to.

kvarec: Ok, i get it.

bagder (approved):

bagder:
Thanks!

PR #10409, opened 2023-02-03
curl_setup: Disable by default recv-before-send in Windows

jay:
Prior to this change a workaround for Windows to recv before every send was enabled by default. The way it works is a recv is called before every send and saves the received data, in case send fails because in Windows apparently that can wipe out the socket's internal received data buffer.

This feature has led to several bugs because the way libcurl operates it waits on a socket to read or to write, and may not at all times check for buffered receive data.

Two recent significant bugs this workaround caused:
- Broken Schannel TLS 1.3 connections (#9431)
- HTTP/2 arbitrary hangs (#10253)

The actual code remains though it is disabled by default. Though future changes to connection filter buffering could improve the situation IMO it's just not tenable to manage this behavior, though it could possibly be handled more correctly at a higher level (eg transfer handling checks for server error response before it has completed sending its request).

Ref: https://github.com/curl/curl/issues/9431
Ref: https://github.com/curl/curl/issues/10253

Closes #xxxx

---

/cc @Karlson2k

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -777,14 +777,16 @@ endings either CRLF or LF so 't' is appropriate.
 #define FOPEN_APPENDTEXT "a"
 #endif
 
-/* WinSock destroys recv() buffer when send() failed.
- * Enabled automatically for Windows and for Cygwin as Cygwin sockets are
- * wrappers for WinSock sockets. https://github.com/curl/curl/issues/657
- * Define DONT_USE_RECV_BEFORE_SEND_WORKAROUND to force disable workaround.
+/* Windows workaround to recv before every send, because apparently Winsock
+ * destroys destroys recv() buffer when send() failed.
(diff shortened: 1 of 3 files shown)


Karlson2k:
Maybe just disable recv-before-send for the handshakes and use it for the data transfers? 
It should be easy to implement and better than complete disable of this feature.

Karlson2k:
Of the other hand, this workaround was implemented for HTTP servers not following RFC requirements: https://www.rfc-editor.org/rfc/rfc9112#section-9.6-9
In the ideal world everything should work without this workaround, but in practice disabling it may lower compatibility with existing servers.

jay:
> In the ideal world everything should work without this workaround, but in practice disabling it may lower compatibility with existing servers.

Thanks for your feedback. Yes, unfortunately premature server responses may be lost on Windows if they're followed by RST.

PR #10420, opened 2023-02-06
ws: fix multiframe send handling

### LINKED ISSUE #10413: Websockets: send multi frames
### I did this
I want to send multiple frames (not fragments).
AFAIK 1st frame must not have CURLWS_CONT bit set and FIN bit aslo must be 0. All other frames must have CURLWS_CONT bit set, also last frame must have FIN bit set.

Now, in ws_packethead function, FIN bit is always set to 1 if CURLWS_CONT  bit is not set. So the server gets only first frame and ignores all other frames.

### curl/libcurl version

7.87.1-DEV

### operating system

Windows 11

bagder:
Fixes #10413

@mikeduglas, does it look ok?

### DIFF
--- a/docs/libcurl/curl_ws_send.3
+++ b/docs/libcurl/curl_ws_send.3
@@ -53,6 +53,8 @@ should be set to zero.
 If \fBCURLWS_RAW_MODE\fP is enabled in \fICURLOPT_WS_OPTIONS(3)\fP, the
 \fBflags\fP argument should be set to 0.
 
+To send a message consisting of multiple frames, set the \fICURLWS_CONT\fP bit
+in all frames except the final one.
 .SH FLAGS
 .IP CURLWS_TEXT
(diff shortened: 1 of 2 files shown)


mikeduglas:
No it doesn't, first if clause should look like this, because final frame in multiframe packet must have CONT opcode as well.

```
  if(!(flags & CURLWS_CONT)) {
    if (!ws->ws.contfragment)
      /* if not marked as continuing, assume this is the final fragment */
      firstbyte |= WSBIT_FIN | opcode;
    else
      /* if marked as continuing, assume this is the final fragment; keep a CONT opcode and set FIN bit */
      firstbyte |= WSBIT_FIN | WSBIT_OPCODE_CONT;
    ws->ws.contfragment = FALSE;
  }

```

bagder:
ok cool, updated!

roberte777:
Is there a reason this was closed? This is needed in order to properly parse fragmented websocket messages.

PR #10430, opened 2023-02-07
multi: stop sending empty HTTP/3 UDP datagrams on Windows

jay:
- Limit the 0-sized send procedure that is used to reset a SOCKET's FD_WRITE to TCP sockets only.

Prior to this change the reset was used on UDP sockets as well, but unlike TCP sockets a 0-sized send actually sends out out a datagram.

Assisted-by: Marc Hörsken

Ref: https://github.com/curl/curl/pull/9203

Fixes https://github.com/curl/curl/issues/9086
Closes #xxxx

---

@mback2k will this affect activity checks on the extra user fd's that could possibly? be neither TCP nor UDP?

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1110,6 +1110,22 @@ CURLMcode curl_multi_fdset(struct Curl_multi *multi,
   return CURLM_OK;
 }
 
+#ifdef USE_WINSOCK
+/* Reset FD_WRITE for TCP sockets. Nothing is actually sent. UDP sockets can't
+ * be reset this way because an empty datagram would be sent. #9203
+ *
+ * "On Windows the internal state of FD_WRITE as returned from
+ * WSAEnumNetworkEvents is only reset after successful send()."
+ */
+static void reset_socket_fdwrite(curl_socket_t s)
+{
+  int t;
+  int l = (int)sizeof(t);
+  if(!getsockopt(s, SOL_SOCKET, SO_TYPE, (char *)&t, &l) && t == SOCK_STREAM)
+    send(s, NULL, 0, 0);
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


lib/multi.c
```diff
@@ -1110,6 +1110,16 @@ CURLMcode curl_multi_fdset(struct Curl_multi *multi,
   return CURLM_OK;
 }
 
+#ifdef USE_WINSOCK
+static bool check_socktype_stream(SOCKET s)
```
bagder, line 1114: since this returns TRUE for TCP, maybe the function name could be named something llike `is_tcp_socket()` to make it more obvious below?

bagder: nevermind

lib/multi.c
```diff
@@ -1110,6 +1110,16 @@ CURLMcode curl_multi_fdset(struct Curl_multi *multi,
...
+{
+  int t;
+  int l = (int)sizeof(t);
+  return (!getsockopt(s, SOL_SOCKET, SO_TYPE, (char *)&t, &l) &&
+          t == SOCK_STREAM);
```
bagder, line 1119: maybe this function should also do the send(), instead of leaving that done in two places where this function is called?

jay: Ok done. I should note, I am still unsure how this is supposed to work with sockets other than TCP and UDP that are provided by the user, or if it's even appropriate to send 0 in those cases.

bagder (approved):

lib/multi.c
bagder: I don't think we will get to know that until we merge this fix and someone actually runs into a side-effect in a distant future. At least this fix has the potential to remove a lot of crazy from using h3 on Windows right now!

Issue #10389, opened 2023-02-01
Performance issue v7.87.0


nicolasbuch:
### I did this
Upgraded our application servers from alpine linux v3.16.3 (bundles curl v7.83.1) to v3.17.1 (bundles curl v7.87.0) and noticed our ElasticSearch server response time jump to roughly the double. Our ElasticSearch implementation uses curl as connection handler.
The problem does not seem to appear in curl v7.86.0.

### I expected the following
The same performance from curl v7.87.0 as i got from the previous versions.

### curl/libcurl version

curl 7.87.0 (aarch64-alpine-linux-musl) libcurl/7.87.0 OpenSSL/3.0.7 zlib/1.2.13 brotli/1.0.9 nghttp2/1.51.0
Release-Date: 2022-12-21
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets

### operating system
Linux d28e463d093f 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 Linux
&
Linux ip-172-xxx-xxx-xxx.eu-central-1.compute.internal 4.14.301-224.520.amzn2.x86_64 #1 SMP Fri Dec 9 09:57:03 UTC 2022 x86_64 Linux

### Easiest way to reproduce (docker)
```
# curl (v7.87.0) google.com and get response time
docker run --rm curlimages/curl:7.87.0 -o /dev/null -s -w 'Establish Connection: %{time_connect}s\nTTFB: %{time_starttransfer}s\nTotal: %{time_total}s\n' https://google.com

# Output
Establish Connection: 0.214716s
TTFB: 0.424541s
Total: 0.424690s
```
```
# curl (v7.86.0) google.com and get response time
docker run --rm curlimages/curl:7.86.0 -o /dev/null -s -w 'Establish Connection: %{time_connect}s\nTTFB: %{time_starttransfer}s\nTotal: %{time_total}s\n' https://google.com

# Output
Establish Connection: 0.018976s
TTFB: 0.220286s
Total: 0.220399s
```

bagder:
Suspected duplicate of #10157 and #10146 fixed in #10160

[...]

PR #10432, opened 2023-02-07
Openssl and HTTP/2 handshake improvements:

icing:
- move Curl_ssl_setup_x509_store() call after the first send (ClientHello) this gives time to parse CA anchors while waiting on the server reply
- add a small buffer to nghttp2 session sending in order to aggregate small SETTINGS/PRIO/WIN_UPDATE frames that nghttp2 "writes" to the callback individually.

Refs #10389


bagder (approved):

calvin2021y:
Is wolfSSL need this kind patch ?

bagder:
@calvin2021y the wolfSSL backend does not support CA caching so this exact change seems unlikely there, but there are probably other performance enhancing changes that can be done there. Oh, and CA caching would benefit wolfSSL users as well.

bagder:
Thanks!

PR #10441, opened 2023-02-08
ws: use %Ou for outputting curl_off_t with info()

### LINKED ISSUE #10439: Incorrect format %zu for curl_off_t in ws.c
### I did this

in ws.c there are several calls of infof() function. For curl_off_t variables %zu format is used which breaks the output. %llu format works well, at least here in Visual Studio 2022, win32 platform.

### curl/libcurl version

curl 7.87.1-DEV (i386-pc-win32) libcurl/7.87.1-DEV Schannel WinIDN

### operating system

Windows 11
Visual Strudio 2022

bagder:
Reported-by: Mike Duglas
Fixes #10439

### DIFF
--- a/lib/ws.c
+++ b/lib/ws.c
@@ -296,7 +296,7 @@ static CURLcode ws_decode(struct Curl_easy *data,
     *oleft = 0;             /* bytes yet to come (for this frame) */
   }
 
-  infof(data, "WS: received %zu bytes payload (%zu left, buflen was %zu)",
+  infof(data, "WS: received %Ou bytes payload (%Ou left, buflen was %zu)",
         payloadsize, *oleft, inlen);
   return CURLE_OK;
 }


lib/ws.c
```diff
@@ -296,7 +296,7 @@ static CURLcode ws_decode(struct Curl_easy *data,
     *oleft = 0;             /* bytes yet to come (for this frame) */
   }
 
-  infof(data, "WS: received %zu bytes payload (%zu left, buflen was %zu)",
+  infof(data, "WS: received %Ou bytes payload (%Ou left, buflen was %zu)",
```
jay, line 299: there's CURL_FORMAT_CURL_OFF_TU. Ou looks cleaner though.
```suggestion
  infof(data, "WS: received %" CURL_FORMAT_CURL_OFF_TU " bytes payload (%"
        CURL_FORMAT_CURL_OFF_TU " left, buflen was %zu)",
```

bagder: Yeah, when we know it is our own printf() I think we can use Ou, while as if we use an "external" printf we must use `CURL_FORMAT_CURL_OFF_TU`

Issue #10433, opened 2023-02-07
HTTP/2 request content upload stalls when response headers are received


LPardue:
### I did this

Scenario: a curl client POSTing a large file (say over 100 KB) to an HTTP/2 server that is configured to respond immediately before  a) waiting for all of the indicated content-length to be received or b) waiting for the request stream FIN. Note that this server also leaves the response side open until it detects the client closes its side.

I'm aware of reports that in such a scenario, once curl receives the response, the upload no longer progresses. For servers that have some kind of body read timeout (e.g. http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_send_timeout) this means that the stream or connection is eventually timed and reset or closed with an error

### I expected the following

When a response is received before curl completes a request upload, and the server has not closed its side, curl should continue the upload.

### curl/libcurl version

One confirmed report with an old verion

```
curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3
Release-Date: 2020-12-09
```

Uncomfirmed reports with later versions exhibit the same behaviour.

### operating system

```
Debian 5.10.158-2
```

bagder:
I believe @icing has a fix for this?

icing:
I have indeed. Will make a PR tomorrow pbly.

icing:
Fix and test case for this in https://github.com/curl/curl/pull/10443.

PR #10443, opened 2023-02-08
HTTP/2 and HTTP/3 upload handling

icing:
HTTP/2 and HTTP/3 upload handling, continue upload when state.drain is set.

- as reported in #10433, HTTP/2 uploads may stall when a response is received before the upload is done. This happens when the data->state.drain is set for such a transfer, as the special handling in transfer.c from then on only cared about downloads.
- add continuation of uploads, if applicable, in this case.
- add pytest case test_07_12_upload_seq_large to reproduce this scenario (although, current nghttp2 implementation is using drain less often)


bagder (approved):

bagder:
@icing fine to merge now I presume?

bagder:
Thanks!

PR #10444, opened 2023-02-08
nghttp2, small fixes.

icing:
- use memory buffer in full available size
- fail receive of reset/errored streams early

pytest:
- make test_05 error cases more reliable


bagder (approved):

bagder:
Thanks!

Issue #10449, opened 2023-02-09
build warning in old version of visual studio


jay:
build warning in old version of visual studio, in [cf_h2_send](https://github.com/curl/curl/blob/3de3ea6a646687de2bf4154ce647485a84e1c6f9/lib/http2.c#L1961) -1 is assigned to a size_t in a bunch of places, eg `len = -1`

..\..\..\..\lib\http2.c(1978): warning C4245: '=' : conversion from 'int' to 'size_t', signed/unsigned mismatch

I'm using a pre generated build configuration from the projects directory with enabled http2 added. Possibly needs `ssize_t sent` or I can just ignore it. I notice none of the CI builds show this warning.

@icing

icing:
Indeed, that does not look sound. I'll add a `ssize_t` for the return value.

icing:
Should be addressed in #10450.

PR #10450, opened 2023-02-09
nghttp2, better handling of cf_h2_send() return value.

icing:
- use explicit ssize_t return var instead of mis-using size_t len for -1 returns.

Refs #10449

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -2039,7 +2039,7 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
     }
 
 #ifdef DEBUG_HTTP2
-    if(!len) {
+    if(!nwritten) {
       infof(data, "http2_send: easy %p (stream %u) win %u/%u",
             data, stream->stream_id,
             nghttp2_session_get_remote_window_size(ctx->h2),
@@ -2048,12 +2048,14 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
         );
 
     }
-    infof(data, "http2_send returns %zu for stream %u", len,
+    infof(data, "http2_send returns %zd for stream %u", nwritten,
           stream->stream_id);
 #endif
+    /* handled writing BODY for open stream. */
     goto out;
   }
-
+  /* Stream has not been opened yet. `buf` is expected to contain
+   * request headers. */
   /* TODO: this assumes that the `buf` and `len` we are called with
    * is *all* HEADERs and no body. We have no way to determine here
    * if that is indeed the case. */
@@ -2123,7 +2125,7 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
   infof(data, "Using Stream ID: %u (easy handle %p)",
         stream_id, (void *)data);
   stream->stream_id = stream_id;
-  /* See comment above. We assume that the whole buf was consumed by
+  /* See TODO above. We assume that the whole buf was consumed by
    * generating the request headers. */
   nwritten = len;
 
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1977,6 +1977,7 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
   CURLcode result;
   struct h2h3req *hreq;
   struct cf_call_data save;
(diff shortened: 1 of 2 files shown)


lib/http2.c
```diff
@@ -1999,26 +2000,33 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
...
-    /* Nullify here because we call nghttp2_session_send() and they
-       might refer to the old buffer. */
+    /* The callback for stream input updates `upload_len` with what it
+     * has provided to nghttp2. If it has *not* been called, `upload_len`
+     * is unchanged and we should report a block on send. */
+    if(stream->upload_len == len) {
+      *err = CURLE_AGAIN;
+      nwritten = -1;
+      goto out;
+    }
```
jay, line 2020: I am not sure about this logic, I notice there's a check for !len that comes later in the `ifdef DEBUG_HTTP2` section. Also the two len references in that section should be updated to nwritten

lib/http2.c
```diff
@@ -2105,25 +2116,28 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
...
-    len = -1;
+    nwritten = -1;
     goto out;
   }
 
   infof(data, "Using Stream ID: %u (easy handle %p)",
         stream_id, (void *)data);
   stream->stream_id = stream_id;
+  /* See comment above. We assume that the whole buf was consumed by
+   * generating the request headers. */
```
jay, line 2127: 'see comment above' is too generic, maybe see TODO above or something so we don't lose track of it. not to mention that TODOs are easy to lose track of, it's better to open an issue if there is one than put it in a TODO

jay (changes requested):

lib/http2.c
icing: I'll change to TODO. Also agree that opening an issue would be good for use of `data->state.buffer` in HTTP/2 and HTTP/3.

lib/http2.c
icing: Good catch. The `len` use below should use `nwritten`.

icing: Updated in recent push.

lib/http2.c
icing: Updated in recent push.

lib/http2.c
jay: Thanks but I meant the new logic to return eagain rather than continue on and call resume, should_close_session etc and !nwritten eval (which will never be true now)

lib/http2.c
```diff
@@ -1999,26 +2000,33 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
...
+    /* The callback for stream input updates `upload_len` with what it
+     * has provided to nghttp2. If it has *not* been called, `upload_len`
+     * is unchanged and we should report a block on send. */
+    if(stream->upload_len == len) {
+      *err = CURLE_AGAIN;
+      nwritten = -1;
+      goto out;
+    }
+
+    nwritten = len - stream->upload_len;
```
jay, line 2022: I get no warning for this despite the unsigned/signed conversion

icing: the result of a `-` operation is signed.

jay: ? You are probably thinking of some other type or language. The result should be unsigned which is then assigned to signed type. VS of any version doesn't seem to warn on this at all (I've asked [here](https://stackoverflow.com/questions/75445208/can-visual-studio-warn-of-unsigned-signed-type-conversion) anyway) and gcc only warns when `-Wsign-conversion`, so it looks like a non issue.

icing: You are right. I am tempted to edit my stupid reply above, but will accept my shortcomings.

bagder:
@jay any updated comments on this before we merge?

jay:
Yes, this PR doesn't just fix the return value it changes the logic (see my comment [here](https://github.com/curl/curl/pull/10450#discussion_r1103754005)) . It bypasses checks for should_close_session, nghttp2_session_resume_data and the debug text is no longer shown when len is 0. I ran into what I think is a bug while checking this, but I don't know yet if it's related. I will check it more tomorrow.

icing:
Closing this, as the changes have become part of #10618

PR #10456, opened 2023-02-09
CURLOPT_PIPEWAIT, allow waited reuse also for subsequent connections.

icing:
As tesated in test_02_07, when firing off 200 urls with --parallel, 199 wait for the first connection to be established. if that is multiuse, urls are added up to its capacity.
The first url over capacity opens another connection. But subseqent urls found the same situation and open a connection too. They should have waited for the second connection to actually connect and make its capacity known.

This change fixes that by

- setting `connkeep()` early in the HTTP setup handler. as otherwise a new connection is marked as closeit by default and not considered for multiuse at all
- checking the "connected" status for a candidate always and continuing to PIPEWAIT if no alternative is found.

pytest:
- removed "skip" from test_02_07
- added test_02_07b to check that http/1.1 continues to work as before

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -233,6 +233,7 @@ static CURLcode http_setup_conn(struct Curl_easy *data,
 
   Curl_mime_initpart(&http->form);
   data->req.p.http = http;
+  connkeep(conn, "HTTP default");
 
   if((data->state.httpwant == CURL_HTTP_VERSION_3)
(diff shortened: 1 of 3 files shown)


bagder (approved):

bagder:
I'll play safe and merge this after the 7.88.0 release, to give this a little more time between merge and release.

bagder:
Thanks!

PR #10460, opened 2023-02-09
CI: Retry failed downloads to reduce spurious failures

dfandrich:
A temporary error with a remote server shouldn't cause a CI run to fail.
Also, put a cap on the time to download to fail faster on a misbehaving
server or connection and use HTTP compression where possible to reduce
download times.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,7 @@ commands:
     steps:
       - run:
           command: |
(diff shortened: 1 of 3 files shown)


bagder (approved):

dfandrich:
I spotted one I missed in the Intel compilers workflow as you were reviewing. I'll submit this once that one goes green.

PR #10462, opened 2023-02-09
no-clobber.d: only use long form options in man page text

### LINKED ISSUE #10461: gen.pl duplicates option text
As reported in #10459, the man page for the `--no-clobber` option has the text

> You can thus use --clobber to enforce the clobbering, even if <ins>-J,</ins>  --remote-header-name or -J is specified.

However, the source file no-clobber.d says

> You can thus use --clobber to enforce the clobbering, even if --remote-header-name or -J is specified.

Note the underlined text in the first case `-J,`. The erroneous text is created with the command

    perl ./gen.pl mainpage no-clobber.d remote-header-name.d

but it does not appear with the command

    perl ./gen.pl mainpage no-clobber.d

the difference being that adding the `remote-header-name.d` file includes an option referenced in no-clobber.d that appears in the area which the added text occurs.

bagder:
... since they are expanded and the short-form gets mentioned automatically so if the short form is mention as well, it gets repeated.

Fixes #10461
Reported-by: Dan Fandrich

### DIFF
--- a/docs/cmdline-opts/no-clobber.d
+++ b/docs/cmdline-opts/no-clobber.d
@@ -15,5 +15,5 @@ of the file that would be created, up to filename.100 after which it will not
 create any file.
 
 Note that this is the negated option name documented.  You can thus use
---clobber to enforce the clobbering, even if --remote-header-name or -J is
+--clobber to enforce the clobbering, even if --remote-header-name is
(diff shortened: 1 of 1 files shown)


dfandrich (approved):
You learn something new every day!

PR #10463, opened 2023-02-10
rustls: improve error messages

jsha:
Changes numeric error codes into verbose error codes in two places. Adds a prefix indicating that the error came from rustls, and in some places which function it came from.

Adds special handling for RUSTLS_RESULT_UNEXPECTED_EOF, since the default message of "Unexpected EOF" is insufficiently explanatory.


bagder (approved):

bagder:
Thanks!

Issue #10468, opened 2023-02-10
Issue while building curl with ngtcp2 + nghttp3


Karthikdasari0423:
Hi @bagder 

while trying to create curl with ngtcp2 and nghttp2,below error is seen

make[2]: Entering directory '/src/curl/lib'
  CC       vquic/libcurl_la-curl_ngtcp2.lo
vquic/curl_ngtcp2.c: In function ‘quic_settings’:
vquic/curl_ngtcp2.c:234:26: error: ‘NGTCP2_DEFAULT_HANDSHAKE_TIMEOUT’ undeclared (first use in this function); did you mean ‘NGTCP2_ERR_HANDSHAKE_TIMEOUT’?
  234 |   s->handshake_timeout = NGTCP2_DEFAULT_HANDSHAKE_TIMEOUT;
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                          NGTCP2_ERR_HANDSHAKE_TIMEOUT
vquic/curl_ngtcp2.c:234:26: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [Makefile:3211: vquic/libcurl_la-curl_ngtcp2.lo] Error 1
make[2]: Leaving directory '/src/curl/lib'

After chaging to vquic/curl_ngtcp2.c  curl ran sucessfully

Expected to raise and PR and change

root@ubuntu:/src/curl# curl -V
curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.9 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3
Release-Date: 2020-01-08
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: AsynchDNS brotli GSS-API HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets

### operating system

Ubuntu
Linux ubuntu 5.4.0-125-generic #141-Ubuntu SMP Wed Aug 10 13:42:03 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

icing:
[...]

PR #10469, opened 2023-02-10
build error due to change in ngtcp2 variables

Karthikdasari0423:
raised PR cause while build curl with ngtcp2 below error is noticed
https://github.com/curl/curl/issues/10468

hers is the PR to ngtcp2 which is causing issue
https://github.com/ngtcp2/ngtcp2/pull/665/

### DIFF
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -231,7 +231,7 @@ static void quic_settings(struct cf_ngtcp2_ctx *ctx,
 
   (void)data;
   s->initial_ts = timestamp();
-  s->handshake_timeout = NGTCP2_ERR_HANDSHAKE_TIMEOUT;
+  s->handshake_timeout = 10;
   s->max_window = 100 * stream_win_size;
   s->max_stream_window = stream_win_size;
 
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -231,7 +231,7 @@ static void quic_settings(struct cf_ngtcp2_ctx *ctx,
 
   (void)data;
   s->initial_ts = timestamp();
-  s->handshake_timeout = NGTCP2_DEFAULT_HANDSHAKE_TIMEOUT;
+  s->handshake_timeout = NGTCP2_ERR_HANDSHAKE_TIMEOUT;
   s->max_window = 100 * stream_win_size;
   s->max_stream_window = stream_win_size;
 


vszakats:
`NGTCP2_DEFAULT_HANDSHAKE_TIMEOUT` has indeed be deleted last week (but still [present](https://github.com/ngtcp2/ngtcp2/blob/61625319ffcb74892da23635bfa04967cbdf1f19/lib/includes/ngtcp2/ngtcp2.h#L483) in the current latest release 0.13.1): https://github.com/ngtcp2/ngtcp2/commit/edc47d5037eca213acb9ea437549e08754ed0788

The reason being: "current value (10 seconds) is arbitrary choice and no good reason behind it"

But, `NGTCP2_ERR_HANDSHAKE_TIMEOUT` is an error code with a negative value of -246:
https://github.com/ngtcp2/ngtcp2/blob/8a78123df44e50b64211f8981ef152dde52c8943/lib/includes/ngtcp2/ngtcp2.h#L790

This will silence the compiler error, but it doesn't seem like a useful timeout value.

Karthikdasari0423 (commented):
--handshake-timeout=<DURATION>
              Set the QUIC handshake timeout.
              Default: 10s

Karthikdasari0423:
> `NGTCP2_DEFAULT_HANDSHAKE_TIMEOUT` has indeed be deleted last week (but still [present](https://github.com/ngtcp2/ngtcp2/blob/61625319ffcb74892da23635bfa04967cbdf1f19/lib/includes/ngtcp2/ngtcp2.h#L483) in the current latest release 0.13.1): [ngtcp2/ngtcp2@edc47d5](https://github.com/ngtcp2/ngtcp2/commit/edc47d5037eca213acb9ea437549e08754ed0788)
> 
> The reason being: "current value (10 seconds) is arbitrary choice and no good reason behind it"
> 
> But, `NGTCP2_ERR_HANDSHAKE_TIMEOUT` is an error code with a negative value of -246: https://github.com/ngtcp2/ngtcp2/blob/8a78123df44e50b64211f8981ef152dde52c8943/lib/includes/ngtcp2/ngtcp2.h#L790
> 
> This will silence the compiler error, but it doesn't seem like a useful timeout value.

Building curl with ngtcp2 
https://github.com/curl/curl/blob/master/docs/HTTP3.md
as of this link,ngtcp2 will be always latest one and not released one

vszakats:
This will become a problem on the next ngtcp2 release either way, so definitely needs to be addressed, ideally before the curl release next week.

Karthikdasari0423:
> I think 10 needs to be multiplied by `NGTCP2_SECONDS`:
> 
> ```c
> #define NGTCP2_SECONDS ((ngtcp2_duration)1000000000ULL)
> ```
> 
> This restores the existing arbitrary value.
> 
> Is there an existing curl timeout that we could re-use here? Like `CURLINFO_CONNECT_TIME_T`?

sorry,i am not sure of this

bagder:
@tatsuhiro-t any advice?

bagder:
I propose we provide our own version of the removed define for now. I'm not sure if 10 seconds should be fixed like this, but we can work on that later:
~~~
#define QUIC_HANDSHAKE_TIMEOUT (10 * NGTCP2_SECONDS)
~~~

tatsuhiro-t:
It works.  If curl has "connect()" timeout, then it can be used here.
Alternatively, just ommit setting handshake_timeout, then QUIC idle timeout will terminate connection during handshake.

tatsuhiro-t:
https://github.com/curl/curl/blob/ca95d391ea7a6ef5d36d52fc46b8b13ae4454a36/lib/vquic/curl_ngtcp2.c#L92
https://github.com/curl/curl/blob/ca95d391ea7a6ef5d36d52fc46b8b13ae4454a36/lib/vquic/curl_ngtcp2.c#L244

60 seconds.

bagder:
While pondering if we should perhaps allow for more than 10 seconds and rather use the idle timeout, I propose #10474 which just restores the 10 second handshake timeout but also fixes another necessary change prompted by upstream changes.

PR #10471, opened 2023-02-10
Fix Brotli name

data-man:
For avoiding CMake's warning:
>CMake Warning (dev) at /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (BROTLI)
  does not match the name of the calling package (Brotli).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  CMake/FindBrotli.cmake:31 (find_package_handle_standard_args)
  CMakeLists.txt:713 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.


jay:
Thanks. Note recent CMake should always set the uppercase version (ie BROTLI_FOUND) even if the case is mixed.

PR #10472, opened 2023-02-10
Address undefined behaviour by checking for null before incrementing pointer at end of do/while loop.

jkeastbb:
This address undefined behaviour found using clang's UBsan:

curl/lib/setopt.c:177:14: runtime error: applying non-zero offset 1 to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior curl/lib/setopt.c:177:14 in

### DIFF
--- a/lib/setopt.c
+++ b/lib/setopt.c
@@ -174,7 +174,7 @@ static CURLcode protocol2num(const char *str, curl_prot_t *val)
 
       *val |= h->protocol;
     }
-  } while(str++);
+  } while(str && str++);
 
   if(!*val)
     /* no protocol listed */


dfandrich:
Does this actually fix a problem? The loop will already exit at the right time without this patch. Is this just a C spec compliance issue?

jkeastbb:
> 

I haven't observed this actually breaking but the issue was flagged by clang's Undefined Behaviour sanitizer: 

`
curl/lib/setopt.c:177:14: runtime error: applying non-zero offset 1 to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior curl/lib/setopt.c:177:14 in 

`
This change resolves the issue flagged.

dfandrich:
That looks fine to me then. Would you update the description to mention ubsan?

jkeastbb:
> That looks fine to me then. Would you update the description to mention ubsan?

Okay, thanks, I've done that now.  Sorry, I wasn't sure what details were expected.  Hope it is okay now.

PR #10475, opened 2023-02-11
windows: always use curl's basename() implementation

### LINKED ISSUE #10261: windows:curl -F filename error when i use chinese
I want to upload file by curl ,but i get the file name  error and missing file suffix

curl/libcurl version 7.87.0
```
curl  'http://10.128.6.26:3030/package/publish' --form 'projectName="electron-test"'  --form 'packageFile=@"C:\Users\Administrator\Desktop\test-upload\测试中文-0.1.3-win.exe"'
```
i get file name '测试中文-0.1.3-w'

In version 7.83.1

I get file name '测试中文-0.1.3-win.exe'

System:win10

vszakats:
The `basename()` \[1\]\[2\] implementation provided by mingw-w64 \[3\] makes assumptions about input encoding and may break with non-ASCII strings.

`basename()` was auto-detected with CMake, autotools and since 68fa9bf3f5d7b4fcbb57619f70cb4aabb79a51f6 (2022-10-13), also in `Makefile.mk` after syncing its behaviour with the mainline build methods. A similar patch for curl-for-win broke official Windows builds earlier, in release 7.83.1_4 (2022-06-15).

This patch forces all Windows builds to use curl's internal `basename()` implementation to avoid such problems.

\[1\]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html
\[2\]: https://www.man7.org/linux/man-pages/man3/basename.3.html
\[3\]: https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-crt/misc/basename.c

Reported-by: UnicornZhang on Github
Assisted-by: Cherish98 on Github

Fixes #10261
Closes #10475

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -838,6 +838,13 @@ int getpwuid_r(uid_t uid, struct passwd *pwd, char *buf,
 #define USE_HTTP3
 #endif
 
+/* Certain Windows implementations are not aligned with what curl expects,
+   so always use the local one on this platform. E.g. the mingw-w64
+   implementation can return wrong results for non-ASCII inputs. */
+#if defined(HAVE_BASENAME) && defined(WIN32)
+#undef HAVE_BASENAME
+#endif
+
 #if defined(USE_UNIX_SOCKETS) && defined(WIN32)
 #  if defined(__MINGW32__) && !defined(LUP_SECURE)
      typedef u_short ADDRESS_FAMILY; /* Classic mingw, 11y+ old mingw-w64 */


vszakats:
Unrelated Windows link failure in a test:
```
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: lib2305.o:lib2305.c:(.text+0x28): undefined reference to `curlx_win32_fopen'
collect2.exe: error: ld returned 1 exit status
```
Possible a regression after 5a9a04d556772b0281108a5455e0fa2aa5c9bbe3.

lib/curl_setup.h
```diff
@@ -838,6 +838,10 @@ int getpwuid_r(uid_t uid, struct passwd *pwd, char *buf,
 #define USE_HTTP3
 #endif
 
+#if defined(HAVE_BASENAME) && defined(WIN32)
```
jay, line 841: I suggest a brief comment explaining why. Also is this a bug in mingw-w64 or is it our bug?

bagder (approved):

lib/curl_setup.h
vszakats: Added a comment.

Sticking with the short example from https://github.com/curl/curl/issues/10261#issuecomment-1426679328, this seems to be a mingw-w64 issue. But hard to say for certain without reproducing/debugging this with that exact (raw byte!) input and locale (and possibly the OS env). Also the original bug report is a slightly different case.

[ Just by eyeballing the source, there seems to be one bug at least, where the `if (*refname)` condition is always true, so the `else` branch never executes, even though it should if the input has no pathsep. The `else` branch should normally be hit by the short example, but not by the original bug report. ]

PR #10478, opened 2023-02-12
CONTRIBUTE: More formally specify the commit description

dfandrich:
This codifies what people have actually used in git commits over the
past 6 years. I've left off some lesser-used headers that appear to
duplicate others and tried to describe a consistent use for several
others that were used more arbitrarily.

This makes it easier for new committers to find out the kinds of things
we want to acknowledge, makes it easier to perform statistical analysis
on commits, and opens the possibility of performing lint checks on
descriptions before submission.

QUESTIONS:

The number of endorsed header types could be reduced and simplified, e.g.:

- Use Reported-by: instead of Found-by: 
- Use Assisted-by: instead of Tested-by:
- Use Acked-by: instead of Approved-by:
- or use Reviewed-by: instead of Acked-by: and Approved-by:
- Use Authored-by: for each author instead of Co-Authored-by:

Endorsing them all does allow slightly more fine-grained information at the
price of slightly higher cognitive load reading them.

### DIFF
--- a/docs/CONTRIBUTE.md
+++ b/docs/CONTRIBUTE.md
@@ -188,34 +188,62 @@ to the list or better yet: change it to a pull request.
 
 ### Commit messages
 
-A short guide to how to write commit messages in the curl project.
+A short guide to how to write git commit messages in the curl project.
 
     ---- start ----
     [area]: [short line describing the main effect]
            -- empty line --
-    [full description, no wider than 72 columns that describe as much as
+    [full description, no wider than 72 columns that describes as much as
     possible as to why this change is made, and possibly what things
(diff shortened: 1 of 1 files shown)


jay:
I don't think we need every instance of -by to be explained, we already know and state whatever-else-by: is an attribution. Some of the more common terms could be suggested on a single line like "Some common attribution terms are Reviewed-by, Signed-off-by, etc". Acked-by and tested-by aren't used like ever AFAICT.

bagder (approved):

dfandrich:
Jay's suggestion makes the text considerably smaller and more approachable so I've done that.

PR #10487, opened 2023-02-12
more KNOWN_BUGS cleanups

bagder:
Issues no longer considered bugs, some are now just documented behavior. One issue moved to TODO.


vszakats:
Windows CI flakyness is much improved, but still flaky. Manually restarted AppVeyor tasks twice today. It's mostly 1-3 failed tasks with a single, seemingly random failed test (e.g. 660, 1149, 1217 2x, 1224, 1242, 1363 2x, 1450, 2003, 1509, 1526).

bagder:
> Windows CI flakyness is much improved, but still flaky

They are still flaky, but I think they are much less so than in the past. But yeah, let me keep that one.

PR #10489, opened 2023-02-12
ngtcp2-gnutls.yml: use github tarball instead of git clone

pheiduck:
better Version update control
bump gnutls to 3.8.0

### DIFF
--- a/.github/workflows/ngtcp2-gnutls.yml
+++ b/.github/workflows/ngtcp2-gnutls.yml
@@ -78,7 +78,6 @@ jobs:
         curl -Lo gnutls https://www.gnupg.org/ftp/gcrypt/gnutls/$GNUTLS_VER_TAG/gnutls-$GNUTLS_VER.tar.xz
         tar -xf gnutls
         cd gnutls-$GNUTLS_VER
-        ./bootstrap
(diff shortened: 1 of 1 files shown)


.github/workflows/ngtcp2-gnutls.yml
```diff
@@ -73,8 +73,11 @@ jobs:
       name: 'install nettle'
 
     - run: |
-        git clone --depth=1 -b 3.7.8 https://github.com/gnutls/gnutls.git
-        cd gnutls
+        GNUTLS_VER_TAG="v3.7"
+        GNUTLS_VER="3.7.9"
+        curl -Lo gnutls https://www.gnupg.org/ftp/gcrypt/gnutls/$GNUTLS_VER_TAG/gnutls-$GNUTLS_VER.tar.xz
+        tar -xf gnutls
+        cd gnutls-$GNUTLS_VER
```
bagder, line 80: this changes the site to hammer from github.com which presumably is close/fast/free for github actions to work against, to gnupg.org. I presume because this is faster?

Would it make sense to compare a download of the tarball from github.com?

pheiduck:
For some reasons it does not build. Should I close this and do a version bump via git clone?

PR #10495, opened 2023-02-13
header-api, tests with h2/h3

icing:
Adding test cases for HTTP/2 and HTTP/3 that check "header_json" output and therefore use of header-api.


bagder (approved):

bagder:
Thanks!

PR #10498, opened 2023-02-13
Test and fix for forward proxy handling (non-tunneling).

icing:
- adding pytest test_10 cases for proxy httpd setup tests
- fixing openssl bug in https: proxy hostname verification that used the hostname of the request and not the proxy name.


bagder (approved):

bagder:
Thanks!

PR #10509, opened 2023-02-14
Make the test telnet server shut down a socket gracefully

icing:
- test 1452 failed occasionally with ECONNRESET errnos in curl when the server closed the connection in an unclean state.
- changed server shutdown to make a final recv() with short timeout before shutting down the socket


bagder:
:heavy_check_mark: wow, all green! :green_heart: :green_heart:  :green_heart:

bagder:
Thanks!

PR #10513, opened 2023-02-14
test2600: try to detect correctly when ALARM_TIMEOUT is in use and ad…

icing:
…just for it.

- use higher timeout values > 1s
- skip duration checks

### DIFF
--- a/tests/unit/unit2600.c
+++ b/tests/unit/unit2600.c
@@ -23,10 +23,43 @@
  ***************************************************************************/
 #include "curlcheck.h"
 
+#ifdef HAVE_NETINET_IN_H
+#include <netinet/in.h>
+#endif
+#ifdef HAVE_NETINET_IN6_H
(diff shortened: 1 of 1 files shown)


tests/unit/unit2600.c
```diff
@@ -223,6 +256,7 @@ static void check_result(struct test_case *tc,
                   tc->id, tc->exp_cf6_creations, tr->cf6.creations);
     fail(msg);
   }
+#ifndef USE_ALARM_TIMEOUT
   duration_ms = Curl_timediff(tr->ended, tr->started);
```
MarcelRaad, line 260: GCC now complains that `duration_ms` is unused in the `USE_ALARM_TIMEOUT` case.

MarcelRaad (approved):
Works for me now on Solaris 11, thanks a lot! 👍

bagder:
Thanks, both of you!

PR #10516, opened 2023-02-15
build fix: include stdint.h more readily

piru:
stdin.h is needed for uint<n>_t etc types, and stdint.h wasn't included when needed.

### DIFF
--- a/lib/http.h
+++ b/lib/http.h
@@ -42,7 +42,7 @@ typedef enum {
 
 #ifndef CURL_DISABLE_HTTP
 
-#if defined(_WIN32) && (defined(ENABLE_QUIC) || defined(USE_NGHTTP2))
+#if defined(ENABLE_QUIC) || defined(USE_NGHTTP2)
 #include <stdint.h>
 #endif
 


lib/http.h
```diff
@@ -42,7 +42,7 @@ typedef enum {
 
 #ifndef CURL_DISABLE_HTTP
 
-#if defined(_WIN32) && (defined(ENABLE_QUIC) || defined(USE_NGHTTP2))
+#if defined(ENABLE_QUIC) || defined(USE_NGHTTP2) || defined(USE_MSH3)
```
bagder, line 45: This can be simplified to just:
~~
#ifdef ENABLE_QUIC
~~~

... since it will always be defined if nghttp2 or msh3 is defined.

bagder: building with any quic backend will define `ENABLE_QUIC` , so checking extra for `USE_MSH3` isn't necessary

bagder (approved):

bagder:
Thanks!

Issue #10514, opened 2023-02-14
curl aborts after half of the connection timeout


mkauf:
The connection timeout handling is broken since commit dbd16c3e256c6ce872829d1654785485361a0a78 (found this commit with `git bisect`).

### I did this

```
time curl --connect-timeout 10 http://my-server:11223/
```

`my-server` is a special system that drops everything on port 11223. iptables rule on this server: `iptables -A INPUT -p tcp --destination-port 11223 -j DROP`. It has an IPv4 address and an IPv6 address.

**Before the mentioned commit (curl <= 7.69.1):**
```
curl: (28) Connection timed out after 10001 milliseconds

real	0m10.004s
user	0m0.002s
sys	0m0.001s
```
-> that's expected

**After the mentioned commit**:

```
curl: (28) Failed to connect to my-server port 11223: Connection timed out

real	0m5.205s
user	0m0.000s
sys	0m0.003s
```

5.2 seconds = (connection timeout) / 2 + (happy eyeballs timeout)

**Current master:**
```
curl: (28) Failed to connect to my-server port 11223 after 5000 ms: Timeout was reached

real	0m5.002s
user	0m0.000s
sys	0m0.003s
```

### I expected the following

curl aborts after the connection timeout (10 seconds) and not after half of the connection timeout (5 seconds)

### curl/libcurl version

curl 7.87.1-DEV (x86_64-pc-linux-gnu) libcurl/7.87.1-DEV nghttp2/1.51.0
Release-Date: [unreleased]
Protocols: dict file ftp gopher http imap mqtt pop3 rtsp smtp telnet tftp
Features: alt-svc AsynchDNS HTTP2 IPv6 Largefile threadsafe UnixSockets

### operating system

Linux myhost 6.1.10-100.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Feb  6 19:58:39 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

icing:
@mkauf: would you be able to run  #10517 in your setup? I found cases where the effective connect timeout was halfed.

mkauf:
@icing : Yes, I have tested it. Your pull request fixes this bug :+1:

PR #10517, opened 2023-02-15
Fix connect timeout handling to use full duration in certain cases.

icing:
- refers to the observation in #10514
- connect timeout was used at half the configured value, if the destination had 1 ip version 4 and other version 6 addresses (or the other way around)
- extended test2600 to reproduce these cases


icing:
> @icing @mkauf you think this is good to merge for 7.88.1

You are brave to consider adding another change of mine...😌

bagder:
Onward and upwards!

bagder:
Thanks!

PR #10519, opened 2023-02-15
Adding more proxy tests 

icing:
Using httpd for direct and tunneling checks on http: and https: proxies and resources.


bagder:
Thanks!

PR #10523, opened 2023-02-15
http: Do not use Negotiate if user and pw are provided

chschenk:
If we provide username and password to curl and the server supports authentication via Kerberos (WWW-Authenticate: Negotiate) and basic authentication libcurl tries to authenticate via Kerberos instead of using the provided username and password to authenticate via basic auth. With this patch libcurl does not use Negotiate authentication method if a username and password are provided.

I really would like to add a test which tests this behavior.
But i'm afraid, that i do not really understand how to add a test like that.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -633,6 +633,10 @@ CURLcode Curl_http_auth_act(struct Curl_easy *data)
   CURLcode result = CURLE_OK;
   unsigned long authmask = ~0ul;
 
+  /* We don't have to try to negotiate if we have a supplied user/password */
+  if(data->state.aptr.user && data->state.aptr.passwd)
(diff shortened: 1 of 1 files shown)


bagder:
I suppose the only good way to test this is for a curl that has support for both methods get offered both and see that this change makes it pick Basic...

dfandrich:
I can see some backwards compatibility issues here, if a client sets a password
in case a Kerberos ticket isn't available but expects to use Kerberos if
possible. Also, what happens with this patch if the server *only* supports
Kerberos and not Basic auth?

bagder:
Very legitimate concerns there raised by @dfandrich ...

bagder:
No response, closing.

Issue #10525, opened 2023-02-15
7.88.0 HTTP/2 (multiplexing?) data corruption


bagder:
@eworm-de reported on IRC that running the arch installer with libcurl 7.88.0 caused corrupted/broken transfers.  Correct size, bad content.

The problem went away after reverting 8c762f59983a and also if just removing:

https://github.com/curl/curl/blob/3027611ca6d4cc5510d2d0fccc3e5a074e09a2fb/lib/http2.c#L1910

This looks like reason enough for a patch release "very soon", but we still want to produce a test case to reproduce and properly understand - as the change was done for a reason and just reverting the change just brings back the old inefficiency.

The breakage was detected when enabling parallel downloads using the arch package installer pacman.

icing:
After much testing and hair pulling, I think we have identified the problem. There are several things which must come together to make this happen. And I think with much less likelihood, this could even happen on older curl versions. The recent changes increased the chances significantly.

First, there is no out-of-bounds or use-after-free here. The corruption happens inside the response data and only involves response data.

What happens:

1. a stream download buffer, `data->state.buffer` becomes full when taking in a HTTP/2 DATA frame. This causes a `NGHTTP2_ERR_PAUSE` to be returned by the callback. The data and size is remembered in `stream->pausedata`.
2. eventually, the transfer reads from the paused stream
   * on 1st read, the `data->state.buffer` is drained
   * on 2nd read, the remembered `stream->pausedata` is copied to the passed `buf`. If it fits (and it does normally), the stream is "unpaused" and `process_pending_input()` is invoked to continue nghttp2's processing of the network data buffer (`ctx->inbuf`). The read returns.
3. the error happens, when there is another HTTP/2 DATA frame *for the same stream* in the network buffer. Then the call to  `process_pending_input()` will invoke the callback again. It will then store the data in `stream->mem` buffer as usual (that one is empty as it was drained in 1.). However `stream->mem` is the same as the `buf` where the pausedata was copied. And this was not yet returned to the caller. So the callback will overwrite it and the caller gets the a copy of "later" data, garbling the download.

The chances vary with the buffer and HTTP/2 frame sizes involved. Our tests with Apache httpd make it very unlikely for this to happen. Our network input buffer is 32K and Apache sends 16K DATA frames. Also, Apache alternates DATA beween streams, making it unlikely that 2 DATA frames for the same stream are there for processing.

`nginx` where the corruption is observed, sends 8K DATA frames and alternates differently.

[...]

PR #10529, opened 2023-02-15
HTTP/2 buffer/pausedata and outpu flush fix.

icing:
* do not process pending input data when copying pausedata to the caller
 * return CURLE_AGAIN if the output buffer could not be completely written out.

Refs #10525.


eworm-de:
Your commit message has a typo. 😜

eworm-de:
With this change applied on top of `7.88.0` I can no longer reproduce this issue.
Thanks a lot for investigating and fixing!

eworm-de:
The first commit still causes trouble... Will try with the second one tomorrow.

bagder:
Thanks!

PR #10530, opened 2023-02-16
http2: set drain on stream end

piru:
Ensure that `on_frame_recv()` stream end will trigger a read if there is pending data. Without this it could happen that the pending data is never consumed.

This combined with https://github.com/curl/curl/pull/10529 should fix https://github.com/curl/curl/issues/10525

Ref: https://github.com/curl/curl/issues/10525


bagder (approved):
nice!

icing:
Nice catch!

bagder:
Thanks!

PR #10534, opened 2023-02-16
Test for HTTP/2 corruption as reported in #10525

icing:
- adding test_02_20 for reproducing the situation
- using recently released mod_h2 Apache module
- skipping test if an older version is installed
- adding installation of current mod_h2 to github pytest workflow

This reproduces the error reliable (for me) on the lib/http2.c version of curl 7.88.0. And passes with the recent curl master.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -81,7 +81,6 @@ jobs:
       name: 'install nghttp2'
 
     - run: |
-        sudo rm -rf mod_h2
         git clone --depth=1 -b master https://github.com/icing/mod_h2
         cd mod_h2
         autoreconf -fi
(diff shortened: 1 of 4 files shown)


pheiduck:
libnghttp2 was not found.
https://github.com/curl/curl/actions/runs/4194306873/jobs/7272293483#step:7:111

pheiduck:
@icing it introduce a bug which need to be fixed on mod_h2 it self.
https://github.com/curl/curl/actions/runs/4194419967/jobs/7272559234#step:7:189

icing:
> @icing it introduce a bug which need to be fixed on mod_h2 it self.
> https://github.com/curl/curl/actions/runs/4194419967/jobs/7272559234#step:7:188

Thanks, currently trying to add support for building against old versions of httpd. ubuntu latest is a bit lagging behind here.

pheiduck:
@icing it needs to be removed after install

bagder:
Thanks!

PR #10540, opened 2023-02-16
socks: allow using DoH to resolve host names

### LINKED ISSUE #10537: DoH cannot be used with SOCKS5 proxy
curl manpage states that `--socks5` option resolves host names locally. I thought that *"locally"* meant *"not by the proxy server itself"*. When it comes to `--doh-url` option, it's assumed that the name resolving is done by the DoH server that may be local as well.

The result of combining these options together is weird: curl attempts to use the default system resolver and doesn't connect to the DoH server. If the default system resolver is not working or returns spurious IP addresses, then there is no way to reach the host.

### I did this

`curl.exe --doh-url https://cloudflare-dns.com/dns-query --socks5 127.0.0.1:1080 https://example.com/`

### I expected the following

curl utilizes Cloudflare's DNS server to resolve `example.com` and connects to the host using the local SOCKS5 proxy.

### curl/libcurl version

curl 7.88.0

### operating system

Windows

bagder:
For SOCKS modes where a local host resolve is done.

It was previously disabled in 12d655d4561, but a few local tests seem to indicate that it works fine.

Reported-by: roughtex on github
Fixes #10537

### DIFF
--- a/lib/socks.c
+++ b/lib/socks.c
@@ -316,7 +316,7 @@ static CURLproxycode do_SOCKS4(struct Curl_cfilter *cf,
     /* DNS resolve only for SOCKS4, not SOCKS4a */
     if(!protocol4a) {
       enum resolve_t rc =
-        Curl_resolv(data, sx->hostname, sx->remote_port, FALSE, &dns);
+        Curl_resolv(data, sx->hostname, sx->remote_port, TRUE, &dns);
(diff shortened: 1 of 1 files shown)


jay (approved):
The reason this works is because the connect phase is non-blocking since 4a4b63da

PR #10543, opened 2023-02-17
curl_setup: Suppress OpenSSL 3 deprecation warnings

jay:
- Define OPENSSL_SUPPRESS_DEPRECATED.

OpenSSL 3 has deprecated some of the functions libcurl uses such as those with DES, MD5 and ENGINE prefix. We don't have replacements for those functions so the warnings were disabled in autotools builds, but still showed in other builds.

Closes #xxxx


bagder (approved):
It is also added in `CMakeLists.txt` that you can remove now.

Issue #10538, opened 2023-02-16
--http3-only is using HTTP/1.1


Karthikdasari0423:
Hi @bagder 

I am using curl with --http3-only option to download file from nginx server.
From below curl man page and help page i came to know that using --http3 will allow to fall back ,--http3-only will not allow to fallback but seems to be with --http3-only also curl is falling back and using http1.1

man curl:-
=======

--http3-only
              sion on its own. Use --http3 for similar functionality with a fallback.
Instructs curl to use HTTP/3 to the host in the URL, with no fallback to earlier HTTP versions.
This option will make curl fail if a QUIC connection cannot be established, it will not attempt any other  HTTP  version on its own

       --http3
              Use --http3-only for similar functionality without a fallback.
Tells  curl  to try HTTP/3 to the host in the URL, but fallback to earlier HTTP versions if the HTTP/3 connection establishment fails

curl --help all :-
==========

 --http3              Use HTTP v3
     --http3-only         Use HTTP v3 only

root@ubuntu:~# curl -# -v -k --http3-only -o index.html https://127.0.0.1:443/
*   Trying 127.0.0.1:443...
* Connected to 127.0.0.1 (127.0.0.1) port 443 (#0)
* ALPN: offers http/1.1
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [88 bytes data]
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
} [1 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [155 bytes data]
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
{ [21 bytes data]
* TLSv1.3 (IN), TLS handshake, Certificate (11):
{ [768 bytes data]
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
{ [264 bytes data]
* TLSv1.3 (IN), TLS handshake, Finished (20):
{ [52 bytes data]
* TLSv1.3 (OUT), TLS handshake, Finished (20):
} [52 bytes data]
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN: server accepted http/1.1
* Server certificate:
*  subject: CN=ubuntu
[...]

PR #10544, opened 2023-02-17
setopt HTTP_VERSION fix: allowing HTTP3 when HTTP2 is not defined.

icing:
Refs #10538, where using --http3-only in a non http2 build has not effect.


bagder (approved):

bagder:
Thanks!

PR #10547, opened 2023-02-17
parsedate without sscanf

bagder:
sscanf() is complex and slow, and we can avoid doing copies


bagder:
I'm holding this until after 7.88.1

Issue #10539, opened 2023-02-16
OS400 build break in tftp.c, 5th argument of Curl_os400_sendto() should be const


jonrumsey:
### I did this
Build 7.88, failed in TFTP module with following const mismatch error at line 531;

CZM0280(30) Function argument assignment between types "struct sockaddr*" and "const struct sockaddr*" is not allowed.

```
    senddata = sendto(state->sockfd, (void *)state->spacket.data,
                      (SEND_TYPE_ARG3)sbytes, 0,
                      &data->conn->remote_addr->sa_addr,
                      data->conn->remote_addr->addrlen);
```

This appears to have been uncovered by https://github.com/curl/curl/pull/10213. 

The root cause is OS400 where in setup-os400.h, sendto() is hash defined to use Curl_os400_sendto(), however the prototype for the OS400 wrapper to deal with EBCDIC/ASCII conversions doesn't match the usual POSIX prototype of;

```
ssize_t sendto(int sockfd, const void *buf, size_t len, int flags,
               const struct sockaddr *dest_addr, socklen_t addrlen);
```

### I expected the following
Clean build

### curl/libcurl version
7.88

[curl -V output]

### operating system
OS400 V7R5M0

PR #10548, opened 2023-02-17
Correct Curl_os400_sendto()

jonrumsey:
Add const qualifier to 5th argument of Curl_os400_sendto()

Make OS400 wrapper for sendto match the normal prototype of sendto() with a const qualifier.

Addresses https://github.com/curl/curl/issues/10539


bagder (approved):

bagder:
Thanks!

PR #10549, opened 2023-02-17
Remove `now` from struct SingleRequest, since it is not needed.

icing:
The struct member is only used inside one function and always initialized first. There is no need to it to be member in the struct.


bagder (approved):

bagder:
Thanks!

PR #10550, opened 2023-02-17
cookie: parse without sscanf()

bagder:
Saves us from using 2*4096 bytes buffers on stack.


bagder:
I'm holding this until after 7.88.1

Issue #10533, opened 2023-02-16
"Content-Disposition" header is ignored


roughtex:
Try to download `https://www.xyplorer.com/xyfc/download/file.php?id=2599` using `--remote-header-name` option. The file is saved to `file.php` despite `Content-Disposition` header presence.

### I did this

`curl.exe --remote-name --remote-header-name https://www.xyplorer.com/xyfc/download/file.php?id=2599`

### I expected the following

`MicrosoftTermCollection-TBXY (PT, RU).zip`

or at least

`MicrosoftTermCollection-TBXY%20%28PT%2C%20RU%29.zip`

### curl/libcurl version

curl 7.88.0

### operating system

Windows

vszakats:
The relevant header is:
```
Content-Disposition: attachment; filename*=UTF-8''MicrosoftTermCollection-TBXY%20%28PT%2C%20RU%29.zip
```

Incidentally I just mentioned this exact issue earlier today in: https://github.com/curl/curl/discussions/10446#discussioncomment-4995045

bagder:
Yeah, percent decoding this into a specific encoding is a pain.

vszakats:
FWIW, my hack to resolve it was this (for UTF-8 only, bash, python and python lib-specific):
```
name="$(curl --fail --silent --show-error --head "${url}" \
  | grep -a -i '^Content-Disposition:' | cut -c 21- \
  | grep -a -i -o -E "filename\*=UTF-8''[A-Za-z0-9%_~\.-]+" | cut -c 18- | urldecode)"
name="$(basename "${name}")"
```
```python
#!/usr/bin/env python3

import sys
import urllib.parse as ul

sys.stdout.write(ul.unquote_plus(sys.stdin.read()))
```

I'm sure this can be improved/simplified quite a bit, but it did the job at the time.

bagder:
This is now documented in the man page. I think we can close this issue now.

PR #10555, opened 2023-02-18
remote-header-name.d: mention that filename* is not supported

bagder:
and that you can use --clobber to allow overwriting.

Ref: #10533

### DIFF
--- a/docs/cmdline-opts/remote-header-name.d
+++ b/docs/cmdline-opts/remote-header-name.d
@@ -20,11 +20,14 @@ The file is saved in the current directory, or in the directory specified with
 
 If the server specifies a file name and a file with that name already exists
 in the destination directory, it will not be overwritten and an error will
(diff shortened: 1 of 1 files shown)


docs/cmdline-opts/remote-header-name.d
```diff
@@ -20,12 +20,15 @@ The file is saved in the current directory, or in the directory specified with
...
-occur. If the server does not specify a file name then this option has no
-effect.
+occur - unless you allow it by using the --clobber option. If the server does
+not specify a file name then this option has no effect.
 
 There's no attempt to decode %-sequences (yet) in the provided file name, so
 this option may provide you with rather unexpected file names.
 
+This feature uses the name from the "filename" field, it does not yet support
+the "filename*" field.
```
jay, line 30: we should explain what it means because most people won't know

```suggestion
the "filename*" field (filenames with explicit character sets).
```

PR #10559, opened 2023-02-18
tool_operate: fix scanbuild compiler warning

jay:
Prior to this change Azure CI scanbuild warned of a potential NULL pointer string passed to strtol when CURLDEBUG enabled, even though the way the code was written it wouldn't have happened.

Bug: https://github.com/curl/curl/commit/5479d991#r101159711
Reported-by: Marcel Raad

Closes #xxxx


MarcelRaad (approved):
Thanks!

bagder (approved):

Issue #10144, opened 2022-12-22
MITM program breaks socketpair, causes getaddrinfo() thread fail to start


SerusDev:
### I did this
`curl https://www.google.com/`

### I expected the following
Instead of a success, curl always returns:
`curl: (6) getaddrinfo() thread failed to start`

### curl/libcurl version
```
curl 7.83.1 (Windows) libcurl/7.83.1 Schannel
Release-Date: 2022-05-13
Protocols: dict file ftp ftps http https imap imaps pop3 pop3s smtp smtps telnet tftp
Features: AsynchDNS HSTS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI UnixSockets
```

### operating system
Windows 10 x64

### possible cause
I was able to trace this problem to https://github.com/curl/curl/blob/master/lib/socketpair.c#L126

Apparently, curl on Windows simulates unix socket pair with a pair of regular TCP sockets on 127.0.0.1.
If the system (like in my case) runs a program that works as a proxy for this connection, socket addresses (ports) mismatch and Curl_socketpair fails.

I guess, the problem is widespread and cryptic. There is a number of antiviruses, firewalls, tunneling and monitoring software that proxy 127.0.0.1 connections and can cause this problem. At the same time, Curl_socketpair call does not provide any error messages. The top level getaddrinfo() error is very hard to track to this.
I experienced this problem when using git for Windows (curl based) together with Proxifier tool.

I propose to remove `a.inaddr.sin_port != a2.inaddr.sin_port` condition or perform peer validation with a chunk of a random data transmitted over the socket (recv/send calls).

bagder:
You could of course also argue that Proxifier is the one breaking curl here...

SerusDev:
> You could of course also argue that Proxifier is the one breaking curl here...

There is no program to blame. :)
The problem is not Proxifier specific. As you correctly stated any MITM tool including some antiviruses can cause this. The problem is that it is almost impossible to track the original error with the MITM tool.

The fix for this is very easy. Socketpair code should not check the source port. Honestly, I do not see any security or stability issues here because everything happens within loopback IP.
As far as I know, many other Windows programs (e.g., Firefox and VMWare) use the same loopback approach to emulate unix socketpair, but they do not check source port and work fine with a MITM program.

bagder:
> There is no program to blame. :)

I disagree, but that doesn't help much.

Someone needs to adjust the code so that users can keep running intrusive programs that MITM localhost connections.

PR #7144, opened 2021-05-28
socketpair: fix potential hangs on Windows

pgroke-dt:
Fixes potential hang in `accept` by using `select` + non-blocking `accept`.
Fixes potential hang in peer check by replacing the `send`/`recv` check with a `getsockname`/`getpeername` check.
Adds length check for returned `sockaddr` data.

---

We have, very rarely, seen `Curl_socketpair` hang in `accept` for more than 20 minutes (at which time our watchdog would write a minidump and restart the application). The callstacks were not always 100% identical but they all had in common that `Curl_socketpair` hangs in `accept`. Unfortunately we were not able to reproduce those hangs locally. After inspection of the code I found two weaknesses: 

1) AFAIU `accept` could block forever, for example if the OS drops the connection from the backlog for whatever reason. Also from my tests it looks like loopback connections really go through at least large parts of the TCP stack on Windows, so I think that other scenarios could also be possible. Maybe Windows drops loopback packets in certain situations? Or some kind of DoS guard in the OS kicks in even for loopback connections? IMO the easiest (and only) fix for this is to use a timeout with `accept`, so that's what I decided to do (`select` followed by non-blocking `accept`). Re. the use of `select`: Using `select` for a single socket in Windows-only code should not be an issue. The dreaded `select` issue with high FDs on POSIX systems doesn't affect Windows: the only limitation there is the _number_ of sockets in an `fd_set`, the _value_ of the socket handles doesn't matter.

2) The `send`+`recv` based peer validation could also lead to hangs. If a "foreign" socket connects first, trying to `recv` from the accepted connection could block forever. I decided to not use non-blocking IO here but replace the check with a `getsockname` + `getpeername` check instead -- because it's both easier and more secure. We have never seen hangs in `recv` but I thought if I'm already working on a patch for `Curl_socketpair` I should probably address that too.

I also thought it could make sense to add a check for the returned `addrlen` so I did that as well. (Might be a little paranoid, but I think it can't hurt.)

PR #10562, opened 2023-02-19
socketpair: allow EWOULDBLOCK when reading the pair check bytes

### LINKED ISSUE #10561: libcurl 7.88.0- getaddrinfo() thread failed to start
I use libcurl, compiled as a static library, in many of my programs.
Code written in C++, for Windows, using Visual Studio.
libcurl static library is compiled to call DLL Windows SSPI, and DLL WinIDN.
I have used all previous versions from 7.50.3 to 7.86.0, and all works great.

Recently I saw 7.88.0 is available, so I downloaded and compiled that one, using the same settings as all my previous builds.
Now all of my programs fail, with a message:
"getaddrinfo() thread failed to start".

So, I have gone back to using the previous version 7.86.0, and all my programs work again.

What might be causing this bug in 7.88.0?
Is there some extra curl setup required in this new version which I will have to change in my programs to make them work with 7.88.0?

bagder:
Reported-by: Gunamoi Software
Debuggged-by: Jay Satiro
Fixes #10561

### DIFF
--- a/lib/socketpair.c
+++ b/lib/socketpair.c
@@ -114,13 +114,42 @@ int Curl_socketpair(int domain, int type, int protocol,
   else {
     struct curltime check;
     struct curltime now = Curl_now();
+    char *p = (char *)&check;
+    size_t s = sizeof(check);
 
     /* write data to the socket */
     swrite(socks[0], &now, sizeof(now));
     /* verify that we read the correct data */
-    if((sizeof(now) != sread(socks[1], &check, sizeof(check)) ||
-        memcmp(&now, &check, sizeof(check))))
-      goto error;
+    do {
+      ssize_t nread = sread(socks[1], p, s);
+      if(nread == -1) {
+        int sockerr = SOCKERRNO;
+        if(
+#ifdef WSAEWOULDBLOCK
+          /* This is how Windows does it */
+          (WSAEWOULDBLOCK == sockerr)
+#else
+          /* errno may be EWOULDBLOCK or on some systems EAGAIN when it
+             returned due to its inability to send off data without
+             blocking. We therefore treat both error codes the same here */
+          (EWOULDBLOCK == sockerr) || (EAGAIN == sockerr) ||
+          (EINTR == sockerr) || (EINPROGRESS == sockerr)
+#endif
+          ) {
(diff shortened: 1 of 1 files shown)


bagder:
@jay this change is kind of made "blindly", it would be super cool if you could verify/adjust it

jay:
fyi I started running repeat.bat in this branch, 20 min no problems so far. I am going to increase MAX and let it run for a while.
~~~
git checkout bagder/socketpair-block
# build curl in visual studio config 'DLL Release - DLL Windows SSPI - DLL WinIDN'
# then checkout just repeat.bat from debug branch and run it from a command prompt
git checkout win32_debug_socketpair_init repeat.bat
~~~

lib/socketpair.c
```diff
@@ -114,13 +114,42 @@ int Curl_socketpair(int domain, int type, int protocol,
...
+#else
+          /* errno may be EWOULDBLOCK or on some systems EAGAIN when it
+             returned due to its inability to send off data without
+             blocking. We therefore treat both error codes the same here */
+          (EWOULDBLOCK == sockerr) || (EAGAIN == sockerr) ||
+          (EINTR == sockerr) || (EINPROGRESS == sockerr)
+#endif
+          ) {
+          /* busy-loop */
+          continue;
```
jay, line 140: I've been monitoring send results and when it returns wouldblock it often happens hundreds of times in a row, so now i'm testing a slight modification with a poll check beforehand

jay:
I have read through #7144 and #10144, and I don't know if there is a case for checking the source port. It seems to me like Proxifier is the problem. I do see that FF at one point a long time ago checked the source port in PR_NewTCPSocketPair but I don't see a similar check in the [current code](https://searchfox.org/mozilla-central/source/netwerk/base/PollableEvent.cpp#42).

bagder:
did you actually see a case where you need to allow **60** seconds of trying? I would've guessed just a few would be enough...

bagder:
I'll merge this in a sec and go for release

jay:
> did you actually see a case where you need to allow **60** seconds of trying? I would've guessed just a few would be enough...

no. i'm guessing how bad it could be but i didn't see anything like that.

PR #10568, opened 2023-02-20
HTTP/{2,3} test fixes.

Kangie:
- Rename `tests/httpserver.pl` -> `tests/http-server.pl` to be consistent with `http{2,3}-server.pl`
- add `tests/http3-server.pl` to distribution tarball
- replace `getcwd()` in `http{2,3}-server.pl` to fix out-of-tree tests.


Kangie:
I'm not going to touch it here, but the HTTP/2 and HTTP/3 proxy perl scripts could probably be consolidated - they're basically copy-pasted; a single `httpx-proxy.pl` with a conditional for the version of HTTP would probably be more sane than maintaining two (almost) identical invocations of the same proxy!

bagder (approved):

bagder:
Thanks!

PR #10584, opened 2023-02-21
curl.1: list global options

bagder:
- provide list of all global options in the top section
- mentioned for each global option

List generated in build, therefore kept up-to-date automatically.


bagder:
This of course creates some "fun" for non-autotools build systems to keep up. I should probably rework this to become a more "internal" gen.pl operation.

PR #10595, opened 2023-02-23
ngtcp2 fix file descriptor confusion on macOS

icing:
The QUIC implementation using ngtcp2 faultily closed file descriptor 0 during initialisation.

On macOS, the next `socket()` call then handed out file descriptor 0, but that could not be used successfully in further OS calls. With the unwanted close fixed, proper file descriptors are returned by `socket()`.


bagder (approved):

bagder:
Thanks!

PR #10601, opened 2023-02-24
hostip: avoid sscanf

bagder:


bagder:
The Microsoft C++ Code Analysis job sounds like it runs into some rate limit:
~~~
Error: Code Scanning could not process the submitted SARIF file:
rejecting SARIF, as there are more runs than allowed (1236 > 20)
Error: Code Scanning could not process the submitted SARIF file:
rejecting SARIF, as there are more runs than allowed (1236 > 20)

PR #10610, opened 2023-02-26
docs: install to standard location

andy5995:
I noticed in the [official Slackbuild script](https://mirrors.kernel.org/slackware/slackware64/source/n/curl/curl.SlackBuild) this is done manually. I thought it would be a good idea to incorporate it here to make it a little easier for distro maintainers.

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -33,6 +33,51 @@ GENHTMLPAGES = curl.html curl-config.html mk-ca-bundle.html
 PDFPAGES = curl.pdf curl-config.pdf mk-ca-bundle.pdf
 MANDISTPAGES = curl.1.dist curl-config.1.dist
 
+packextrasdir=@docdir@
+# Docs installed to @docdir@ (i.e. /usr/local/share/doc/curl by default;
+# configurable by passing '--docdir' to './configure')
+#
+# docs/README.md is omitted from installation because /README.md
+# will be installed instead.
+packextras_DATA = \
+ ALTSVC.md                                      \
+ BINDINGS.md                                    \
+ BUG-BOUNTY.md                                  \
+ BUGS.md                                        \
+ CIPHERS.md                                     \
+ CONNECTION-FILTERS.md                          \
+ CONTRIBUTE.md                                  \
+ DEPRECATE.md                                   \
+ EXPERIMENTAL.md                                \
+ FAQ                                            \
+ FEATURES.md                                    \
+ HELP-US.md                                     \
+ HISTORY.md                                     \
+ HSTS.md                                        \
+ HTTP2.md                                       \
+ HTTP3.md                                       \
+ HTTP-COOKIES.md                                \
+ HYPER.md                                       \
+ KNOWN_BUGS                                     \
+ MAIL-ETIQUETTE                                 \
+ MANUAL.md                                      \
+ MQTT.md                                        \
+ NEW-PROTOCOL.md                                \
+ options-in-versions                            \
+ PARALLEL-TRANSFERS.md                          \
+ RELEASE-PROCEDURE.md                           \
+ ROADMAP.md                                     \
+ RUSTLS.md                                      \
+ SECURITY-PROCESS.md                            \
+ SSLCERTS.md                                    \
+ SSL-PROBLEMS.md                                \
+ THANKS                                         \
+ TheArtOfHttpScripting.md                       \
+ TODO                                           \
+ URL-SYNTAX.md                                  \
+ VERSIONS.md                                    \
+ WEBSOCKET.md
+
 HTMLPAGES = $(GENHTMLPAGES)
 
 # Build targets in this file (.) before cmdline-opts to ensure that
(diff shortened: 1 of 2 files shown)


docs/Makefile.am
```diff
@@ -33,6 +33,21 @@ GENHTMLPAGES = curl.html curl-config.html mk-ca-bundle.html
...
+ BUGS.md \
+ CONTRIBUTE.md \
+ FAQ \
+ FEATURES.md \
+ INSTALL.md \
+ INTERNALS.md \
+ MANUAL.md \
+ README.md \
+ THANKS \
+ TODO
```
bagder, line 49: Just curious, what's the litmus test to use for adding a documentation file to this list?

andy5995: A fair question. I started with the ones I saw in the Slackbuild.

The [docs installed by Debian](https://salsa.debian.org/debian/curl/-/blob/master/debian/libcurl4-doc.docs) are somewhat different but also somewhat similar.

```
README
docs/*.md
docs/FAQ
docs/KNOWN_BUGS
docs/THANKS
docs/TODO
```

AFAIK, the docs installed to docdir can be any docs the upstream maintainers think would be useful and relevant. You could even add a complete html version of your website, if I'm not mistaken.

Sometimes the INSTALL is included, sometimes not (because it's not needed once the package has been installed, and if there's an INSTALL file in docdir, that implies the package has been installed).

andy5995: It would be nice if there was a page that indicated conventions for docs, rather than me just writing my own beliefs here.

I did some searching and found this on a [Fedora packaging guide](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_documentation)

>Any relevant documentation included in the source distribution should be included in the package in the proper documentation directory. Irrelevant documentation includes build instructions, the omnipresent INSTALL file containing generic build instructions, or example, and documentation for non-Linux systems, e.g. README.MSDOS. Also pay attention about which subpackage you include documentation in. For example API documentation belongs in the -devel subpackage, not the main one. Or if there’s a lot of documentation, consider putting it into a subpackage. In this case, it is recommended to use *-doc as the subpackage name.

andy5995: > I thought it would be a good idea to incorporate it here to make it a little easier for distro maintainers.

How this makes it easier for package maintainers is that if the docs go to the correct location using the upstream `make install`, then packagers don't have to add any special rules.

Example:

In this Slackbuild, I've specified the docdir during the configure (meson setup), and the docs will get installed to the tmp package directory.

[rmw Slackbuild](https://slackbuilds.org/slackbuilds/15.0/system/rmw/rmw.SlackBuild)

```
cd build
  CFLAGS="$SLKCFLAGS" \
  CXXFLAGS="$SLKCFLAGS" \
  meson .. \
    --buildtype=release \
    -Dstrip=true \
    -Db_sanitize=none \
    --mandir=/usr/man \
    --prefix=/usr \
    --localedir=/usr/share/locale \
    -Ddocdir=/usr/doc/$PRGNAM-$VERSION
  ninja
  DESTDIR=$PKG ninja install
cd ..
```

The package gets created soon after:

```
cd $PKG
/sbin/makepkg -l y -c n $OUTPUT/$PRGNAM-$VERSION-$ARCH-$BUILD$TAG.$PKGTYPE
```

bagder: in that case, then I think all docs should be installed except those that are specifically about specific internal details (like BUGREF, CURL_DISABLE and DYNBUF).

andy5995:
After installation, this is what the destination docdir contains:

```
/home/andy/src/curl/curl/build/foo_install/usr/local/share/doc/curl
[andy@prometheus curl](install-docs)$ ls -w 80
ALTSVC.md              GOVERNANCE.md          RELEASE-NOTES
BINDINGS.md            HELP-US.md             RELEASE-PROCEDURE.md
BUG-BOUNTY.md          HISTORY.md             ROADMAP.md
BUGS.md                HSTS.md                RUSTLS.md
CHANGES                HTTP2.md               SECURITY.md
CHECKSRC.md            HTTP3.md               SECURITY-PROCESS.md
CIPHERS.md             HTTP-COOKIES.md        SSLCERTS.md
CODE_OF_CONDUCT.md     HYPER.md               SSL-PROBLEMS.md
CONNECTION-FILTERS.md  KNOWN_BUGS             THANKS
CONTRIBUTE.md          MAIL-ETIQUETTE         TheArtOfHttpScripting.md
COPYING                MANUAL.md              TODO
DEPRECATE.md           MQTT.md                URL-SYNTAX.md
EARLY-RELEASE.md       NEW-PROTOCOL.md        VERSIONS.md
EXPERIMENTAL.md        options-in-versions    WEBSOCKET.md
FAQ                    PARALLEL-TRANSFERS.md
FEATURES.md            README.md
```

eli-schwartz:
Seems like several of these files are still pretty irrelevant to end users, even the ones looking at auxiliary documentation about interesting corners, known issues, etc.

They might have to add a special rule to remove the documentation for checksrc.pl, the code of conduct, the devdoc on "when to do an early release", the project governance docs, the source code for the website's posted etiquette rules for sending mailing list info, more devdocs on how a maintainer should handle publishing a release... :)

I think the list of things to install can be further fine-tuned, at least.

> How this makes it easier for package maintainers is that if the docs go to the correct location using the upstream `make install`, then packagers don't have to add any special rules.

I also wonder if maybe the best approach is to have packagers add one special rule: `--enable-doc-install`? Many people might find this documentation useful, but probably also many people find it a lot less useful than a user manual or API manual, such as manpages.

andy5995:
> Seems like several of these files are still pretty irrelevant to end users, even the ones looking at auxiliary documentation about interesting corners, known issues, etc.

Thank you for adding some perspective, @eli-schwartz .

@bagder I would defer to @eli-schwartz 's judgement about this. I don't know if you've worked with him before, but he's one of the lead developers of an immensely prolific project known as @mesonbuild ;)

He's also recognized as a trusted AUR packager for the [Arch User Repository](https://aur.archlinux.org/).

What he said makes sense to me, anyway. :)

> 
> They might have to add a special rule to remove the documentation for checksrc.pl, the code of conduct, the devdoc on "when to do an early release", the project governance docs, the source code for the website's posted etiquette rules for sending mailing list info, more devdocs on how a maintainer should handle publishing a release... :)
> 
> I think the list of things to install can be further fine-tuned, at least.

I totally agree and planned on accepting more feedback about this. I've made a few changes and pushed.

> 
> > How this makes it easier for package maintainers is that if the docs go to the correct location using the upstream `make install`, then packagers don't have to add any special rules.
> 
> I also wonder if maybe the best approach is to have packagers add one special rule: `--enable-doc-install`? Many people might find this documentation useful, but probably also many people find it a lot less useful than a user manual or API manual, such as manpages.

Good point. Personally, I rarely go to /usr/share/doc to find docs about a program. Either I use a man page or Internet search. But it seems like it's still fairly standard practice for projects to install some basic docs? (I have 334 project doc directories on my production system).

That might be one reason why Debian and Slackware only add a few basic docs for the curl package? On my Manjaro system, I see there are NO docs for curl in the docdir.

bagder:
> Seems like several of these files are still pretty irrelevant to end users

That, in combination with the fact that exactly **zero** users have asked for this makes me rather lean towards thinking this change is unnecessary.

PR #10611, opened 2023-02-26
transfer: limit Windows SO_SNDBUF updates to once a second

jay:
- Change readwrite_upload() to call win_update_buffer_size() no more than once a second to update SO_SNDBUF (send buffer limit).

Prior to this change during an upload readwrite_upload() could call win_update_buffer_size() anywhere from hundreds of times per second to an extreme test case of 100k per second (which is likely due to a bug, and that will be reported separately). In the latter case WPA profiler showed win_update_buffer_size was the highest capture count in readwrite_upload. In any case the calls were excessive and unnecessary.

Closes #xxxx

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -980,7 +980,15 @@ static CURLcode readwrite_upload(struct Curl_easy *data,
     if(result)
       return result;
 
-    win_update_buffer_size(conn->writesockfd);
+#if defined(WIN32) && defined(USE_WINSOCK)
+    {
+      struct curltime n = Curl_now();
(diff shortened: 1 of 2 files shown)


bagder (approved):
Sounds very reasonable!

icing:
Could we move this into the cf-socket.c and keep the update time in its context? I think this would be a better place where the actual network writes happen.

jay:
> Could we move this into the cf-socket.c

It was in readwrite_upload so it's only used on sockets that are being used by libcurl to send (large) request uploads

Issue #10618, opened 2023-02-27
HTTP/2 busy loop when upload size unknown


jay:
### I did this

Bisected to 71b7e016 connect/h2/h3 refactor (with 9932701 36dce556 compile fixes applied).

~~~
perl -e "print '0' x 52429310" > ultest

root@ubuntu1604-x64-vm:~/curl# cat ultest | time -p src/curl -o /dev/null -T - https://httpbin.org/put
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  100M  100 50.0M    0 50.0M   948k   948k  0:00:53  0:00:53 --:--:-- 8232k
real 54.03
user 20.45
sys 23.49
~~~

In a linux VM it is 100% reproducible and eats all the CPU. On the Windows host it is arbitrary  but when it happens it is for the entire transfer and consistently eats 8-10% CPU.

Profile taken in Linux using gprof (-fno-omit-frame-pointer -pg -g compiler flags in CFLAGS AND CXXFLAGS).
[output.png](https://user-images.githubusercontent.com/965580/221520046-3831de40-fadf-4b86-b700-7bd484e182d3.png) [output.txt](https://github.com/curl/curl/files/10837852/output.txt)

Profile in Windows host shows time is split between multi_poll and multi_perform. See [bad.png](https://user-images.githubusercontent.com/965580/221518027-b71e807c-b479-4930-82a1-31eba4ebe849.png) and [good.png](https://user-images.githubusercontent.com/965580/221518034-4ccad7ac-0b80-4d31-b6aa-ed42929d1c93.png) for comparison.

<details>
  <summary>Click to expand debugging notes from Windows runs (I took counts of calls to some functions)</summary>

~~~
git checkout 1c18f8d   # good

[33436] eagain count: 0
[33436] curl_multi_timeout_count: 0, curl_multi_wait_count: 0, curl_multi_poll_count: 29202, Curl_poll_count: 58399
[40020] eagain count: 0
[40020] curl_multi_timeout_count: 0, curl_multi_wait_count: 0, curl_multi_poll_count: 27257, Curl_poll_count: 54513
[39832] eagain count: 0
[39832] curl_multi_timeout_count: 0, curl_multi_wait_count: 0, curl_multi_poll_count: 27712, Curl_poll_count: 55419
~~~

~~~
git checkout 71b7e016103   # bad
[...]

PR #10627, opened 2023-02-27
H2 upload busy loop fix

icing:
Fixing busy loop on large uploads for HTTP/2.

- refs #10449 with improvements in h2_cf_send() and nwritten return code
  handling. Cherry picked and improved after review by @jay
- setting KEEP_SEND_PAUSE flags when exhausting remote HTTP/2 window
  size of a stream.
- clearing KEEP_SEND_PAUSE when receiving HTTP/2 window updates
  on a paused stream.

Refs #10618


jay:
Cursory tests in windows host and linux guest are ok. There appears to be some wild speed variations, especially on download, but I think it's the website because I get the same variations with --http1.1. I will try tomorrow to run the profilers to compare before the refactor against this fix.

bagder:
Even if there is "wild speed variations" that would be a separate issue, as this is about fixing https://github.com/curl/curl/issues/10618

jay:
Thanks! I landed it after moving logical operators at the beginning of a line to the end of the previous line, as described in [code style](https://github.com/curl/curl/blob/master/docs/CODE_STYLE.md#column-alignment).

> Even if there is "wild speed variations" that would be a separate issue, as this is about fixing #10618

At the time it was too soon to tell.

jay:
@bagder Can you take a look at this, because it looks possible that if KEEP_SEND_PAUSE is set outside of curl_easy_pause it may be possible that http2 can now unpause sending when the user has paused it with curl_easy_pause. Also, curl_easy_pause has some additional pause/unpause logic (like to [unpause mime sending](https://github.com/curl/curl/blob/d9ccc75b007aefb2b706ce1976ddadc39531fa23/lib/easy.c#L1096-L1102)).I see a KEEP_SEND_HOLD but it is not used anywhere.

icing:
@jay and @bagder: I agree that the semantics of this keepon handling are uncleear. Jay is correct to point out that the h2 busy loop fix might interfere with other uses of keepon.However, h2 needs flow control other than socket events (and h3 as well) and there seems to be no other mechanism in curl.Am 02.03.2023 um 04:43 schrieb Jay Satiro ***@***.***>:﻿
@bagder Can you take a look at this, because it looks possible that if KEEP_SEND_PAUSE is set outside of curl_easy_pause it may be possible that http2 can now unpause sending when the user has paused it with curl_easy_pause. Also, curl_easy_pause has some additional pause/unpause logic (like to unpause mime sending).I see a KEEP_SEND_HOLD but it is not used anywhere.

jay:
Do you know for certain that the cause of #10618 was a flow control issue, and do you know how the refactor caused the bug?

icing:
@jay I added test case `test_07_31_put_10m` to reproduce and the fix the issue.

The test uploads 10MB to a url that does a 10ms delay between accepting data chunks from the client. This causes the HTTP/2 upstream window for that stream to become full. And stay full for 10ms again and again.

On the curl side, this led to 100% cpu without progress, as the transfer loop was calling readwrite again and again only to get -1, CURLE_AGAIN all the time.

I am not claiming this change fixes all possible busy loops, but the one exposed in `test_07_31_put_10m`.

jay:
I think we should open it as an issue because curl8 is coming and we can't have curl_easy_pause sporadically unpausing, and also I'm concerned about the additional unpause logic in the function. I don't think it's going to work properly to also use it for flow control. I haven't had enough time to look into this. I will try tomorrow but unless I come up with something I will open it as a new issue.

PR #10632, opened 2023-02-27
sectransp: make read_cert() use a dynbuf when loading

bagder:


piru:
Using dynbuf makes more sense, indeed.

lib/vtls/sectransp.c
```diff
@@ -2150,50 +2150,39 @@ static long pem_to_der(const char *in, unsigned char **out, size_t *outlen)
...
+    if(!n)
+      break;
     if(n < 0) {
       close(fd);
-      free(data);
+      Curl_dyn_free(&certs);
       return -1;
     }
-    else if(n == 0) {
+    if(Curl_dyn_addn(&certs, buf, n)) {
```
jay, line 2177: #10645

PR #10636, opened 2023-02-27
Allocate the FTP wildcard struct on demand

bagder:
The feature is rarely used so this frees up data for the vast majority of easy handles that don't use it.


bagder:
I'll resubmit it against master to make more CI jobs run

PR #10637, opened 2023-02-27
src/tool_help: make curl -V show "outdated" if older than 350 days

bagder:
Food for thought. And discussion.

### DIFF
--- a/src/tool_help.c
+++ b/src/tool_help.c
@@ -160,13 +160,59 @@ static bool is_debug(void)
   return FALSE;
 }
 
+static void outdated(void)
+{
+  struct tm release;
+  const char *p;
+  unsigned long year, mon, day;
+  char *endp;
+  time_t relepoch;
+  time_t now;
+
+  p = LIBCURL_TIMESTAMP;
+#ifdef DEBUGBUILD
+  {
+    char *alt = getenv("CURL_TIMESTAMP");
+    if(alt)
+      p = alt;
+  }
+#endif
+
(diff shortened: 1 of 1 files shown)


bagder:
One of the primary flaws with this is possibly that even distros or others that patch their versions would get this.

dfandrich:
I like it! I'd make the # days before the message shows able to be overridden at compile time so long-term support distros can increase it to suit their own release schedules. If we make that easy to do, maybe they'll do that rather than just patching this entire code out, which I'm sure many will be tempted to do. Maybe even add a comment here that requests they do that instead?

350 days equals 6-7 releases, which seems like a lot but it's a balance between how long we should let people go before nagging them and how many security vulnerabilities they are exposed to. Our 5-year rate is 6.9 security vulnerabilities discovered every 350 days.

bagder:
Lots of people have remarked how this is a "timebombed" thing that will surprise users, will trigger for virtually all users using stable Linux distributions, probably will hurt the wrong people because many times the users of curl cannot themselves decide to upgrade etc. I am not going to push for this.

dfandrich:
I would expect Linux distros to either change the time threshold to match their
own release policies or remove the check altogether, so I wouldn't expect most
users to be affected unless they are actually using a version that is past the
support time frame. This should only hit users that are compiling themselves.
But, it might take a release or two for distros to figure out that this
new feature is there and they need to adjust it.

bagder:
I don't think we are ready for this drastic approach.

PR #10638, opened 2023-02-28
Single URLs make no sense in --rate example

jidanni:
Here somehow you need to put more than one URL in these examples, else they will make no sense, as --rate only affects the  second and beyond URLs. The first URL will always (OK, usually) finish the same time no matter what --rate is given.


bagder (approved):
Uh, fair point!

bagder:
Thanks!

Issue #10633, opened 2023-02-27
Since 7.88 https over unix socket's no longer connects to the server


jelly:
### I did this

```
[root@archlinux ~]# curl --verbose -k --unix-socket /run/cockpit/sock https://dummy
* Failed to connect to dummy port 443 after 0 ms: Couldn't connect to server
* Closing connection 0
curl: (7) Failed to connect to dummy port 443 after 0 ms: Couldn't connect to server
```

### I expected the following

```
[root@archlinux ~]# curl --version
curl 7.87.0 (x86_64-pc-linux-gnu) libcurl/7.87.0 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.10.0 nghttp2/1.52.0
Release-Date: 2022-12-21
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
[root@archlinux ~]# curl --verbose -k --unix-socket /run/cockpit/sock https://dummy
*   Trying /run/cockpit/sock:0...
* Connected to dummy (/run/cockpit/sock) port 443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* [CONN-0-0][CF-SSL] TLSv1.0 (OUT), TLS header, Certificate Status (22):
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS handshake, Client hello (1):
* [CONN-0-0][CF-SSL] TLSv1.2 (IN), TLS header, Certificate Status (22):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Server hello (2):
* [CONN-0-0][CF-SSL] TLSv1.2 (IN), TLS header, Finished (20):
* [CONN-0-0][CF-SSL] TLSv1.2 (IN), TLS header, Supplemental data (23):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
[...]

PR #10641, opened 2023-02-28
Fixing unix domain socket use in https connects.

icing:
- refs #10633, when h2/h3 eyeballing was involved, unix domain socket configurations were not honoured
- configuring --unix-socket will disable HTTP/3 as candidate for eyeballing
- combinatino of --unix-socket and --http3-only will fail during initialisation
- adding pytest test_11 to reproduce


bagder (approved):

bagder:
Thanks!

Issue #10622, opened 2023-02-27
Connection filter code attempts to resolve remote address for listen connection


piru:
### I did this

`curl -v --ftp-port "-" ftp://ftp.funet.fi`

In output I see a bogus error:
```
...
* sa_addr inet_ntop() failed with errno 47: Address family not supported by protocol family
...
```
The error is due to `set_remote_ip()` call in `Curl_conn_tcp_listen_set()` with unset `ctx->addr.sa_addr`. See https://github.com/curl/curl/blob/cf3e6ce92f1c9686129d0fd4a5d39c0fb38792ec/lib/cf-socket.c#L1816

### I expected the following

No superfluous errors being reported.

### curl/libcurl version

curl 7.88.1 (aarch64-apple-darwin22.3.0) libcurl/7.88.1 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) nghttp2/1.52.0
Release-Date: 2023-02-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB PSL SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

Darwin computername.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000 arm64

icing:
Thanks for the report. Please see #10642 for the fix.

piru:
This fixes the issue for me, thanks!

PR #10642, opened 2023-02-28
Fixes handling of remote addr for accepted tcp sockets.

icing:
- refs #10622, do not try to determine the remote address of a listen socket. There is none.
- Update remote address of an accepted socket by getpeername() if available.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1836,6 +1836,8 @@ static void set_accepted_remote_ip(struct Curl_cfilter *cf,
   struct Curl_sockaddr_storage ssrem;
   curl_socklen_t plen;
 
+  ctx->r_ip[0] = 0;
+  ctx->r_port = 0;
   plen = sizeof(ssrem);
   memset(&ssrem, 0, plen);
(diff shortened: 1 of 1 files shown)


lib/cf-socket.c
```diff
@@ -1828,6 +1827,36 @@ CURLcode Curl_conn_tcp_listen_set(struct Curl_easy *data,
...
 
+static void set_accepted_remote_ip(struct Curl_cfilter *cf,
+                                   struct Curl_easy *data)
+{
+  struct cf_socket_ctx *ctx = cf->ctx;
+#ifdef HAVE_GETPEERNAME
+  char buffer[STRERROR_LEN];
+  struct Curl_sockaddr_storage ssrem;
+  curl_socklen_t plen;
+
```
piru, line 1838: Minor thing, but maybe have:
```
  ctx->r_ip[0] = 0;
  ctx->r_port = 0;
```
...here to set some default values in case there's an error.

icing: Good suggestion. Added to the PR.

bagder (approved):

bagder:
Thanks!

Issue #10634, opened 2023-02-27
Error in the HTTP2 framing layer


rwmjones:
### I did this

I set up an Apache 2.4.55 server to serve a file over HTTP/2.  When connecting repeatedly to the server using the `curl` command, I get the following error (usually at random, but below it happened for the first URL):

```
$ ./src/curl --fail-early --head --http2-prior-knowledge http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img http://localhost/fedora-36.img
HTTP/2 200 
last-modified: Mon, 20 Feb 2023 10:47:51 GMT
etag: "180000000-5f51f6506fa85"
accept-ranges: bytes
content-length: 6442450944
content-type: application/octet-stream
date: Mon, 27 Feb 2023 21:56:39 GMT
server: Apache/2.4.55 (Fedora Linux) SVN/1.14.2

curl: (16) Error in the HTTP2 framing layer
```

### I expected the following

No error.

### curl/libcurl version

This happens with curl @ commit 788c6b38644c4fe761e39dbab00fb03906489254

I bisected the problem and it was introduced by commit 671158242db32032a41e8da89da33d773b481ef8

Reverting curl back to commit b7aaf074e56c4c6ef066da0d4d683628dcc2b225 (the commit before 67115824) fixes the problem.

```
$ ./src/curl -V
curl 8.0.0-DEV (x86_64-pc-linux-gnu) libcurl/8.0.0-DEV OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) nghttp2/1.52.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
[...]

PR #10643, opened 2023-02-28
Fix for http2-prior-knowledge when reusing connections.

icing:
- refs #10634 where errors in the HTTP/2 framing layer are observed.
- the bug was that on connection reuse, the code attempted to switch in yet another layer of HTTP/2 handling instead of detecting that this was already in place.
- added pytest testcase reproducing the issue.


rwmjones:
Fixed the problem for me, thanks!

bagder (approved):

bagder:
Thanks!

Issue #10521, opened 2023-02-15
http fails on 416 when resuming


Smackd0wn:
### I did this

On git master, `curl --fail -O -C - http://host/file` when the local `file` had already been completely downloaded.

Server response:

```
< HTTP/1.1 416 Requested Range Not Satisfiable
< Server: nginx/1.22.1
< Date: Wed, 15 Feb 2023 13:29:29 GMT
< Content-Type: text/html; charset=utf8
< Content-Length: 197
< Content-Range: bytes */1234567
```

### I expected the following

curl exits with code 0 (no error, even with `--fail`), because in `http_should_fail()`:

https://github.com/curl/curl/blob/3027611ca6d4cc5510d2d0fccc3e5a074e09a2fb/lib/http.c#L1161-L1167

However, when the requested range was unsatisfied (e.g., `Content-Range: */1234567`), `data->state.resume_from` is overwritten with `0` (see line 3479), and thus the check above always fail.

https://github.com/curl/curl/blob/3027611ca6d4cc5510d2d0fccc3e5a074e09a2fb/lib/http.c#L3452-L3480

piru:
It appears nginx returns: `content-range: bytes */<filesize>` in this case. For example Apache2 doesn't include the `content-range` header for `416` response.

This seems to fix this issue, but I am not entirely sure if this is the best way to fix it:
```
diff --git a/lib/http.c b/lib/http.c
index ee31aa2c6..caa3d50a1 100644
--- a/lib/http.c
+++ b/lib/http.c
@@ -3476,7 +3476,7 @@ CURLcode Curl_http_header(struct Curl_easy *data, struct connectdata *conn,
           k->content_range = TRUE;
       }
     }
-    else
+    else if(416 != k->httpcode)
       data->state.resume_from = 0; /* get everything */
   }
 #if !defined(CURL_DISABLE_COOKIES)
```
Comments are welcome.

hmeine:
I run into this problem as well. I tried running `curl -C -` to download a build artifact from an [artifactory](https://jfrog.com/de/artifactory/) server, hoping that repeated script runs would not download the large file multiple times.

The server replies with 
```
HTTP/1.1 416 Requested Range Not Satisfiable
Connection: keep-alive
Content-Length: 0
Content-Range: bytes */977880575
Date: Wed, 09 Aug 2023 09:57:07 GMT
Server: nginx/1.18.0 (Ubuntu)
X-Artifactory-Id: ***redacted***
[...]

PR #10644, opened 2023-02-28
http: ignore content-range for 416 responses

piru:
- fix for #10521
- someone should add a test for this as well (not included here)


piru:
There likely needs to be some handling for `-C -` vs `-C size` in the tool to avoid breaking existing behaviour.

bagder:
We need a different take.

PR #10645, opened 2023-03-01
DYNBUF.md: note Curl_dyn_add* calls Curl_dyn_free on failure

jay:
This is the existing behavior and it has been widely assumed in the codebase.

Closes #xxxx


jay:
The spellcheck CI is failing due to the word 'dyn' but I didn't add that word and I don't see where it is in the document. 'dynbuf' is already in the list of allowed words.

bagder:
I think it is because "Curl_dyn_free" is used without backticks. Write it like `Curl_dyn_free` instead.

bagder (approved):

PR #10648, opened 2023-03-01
If socket is already connected, return CURLE_OK

sjadhavar:
This will fix the issue mentioned at https://github.com/curl/curl/issues/10626

In 7.87.0, if callback method for CURLOPT_SOCKOPTFUNCTION returns CURL_SOCKOPT_ALREADY_CONNECTED then curl library used to return CURLE_OK. In 7.88.0, now even if callback returns CURL_SOCKOPT_ALREADY_CONNECTED, curl library still tries to connect to socket by invoking method do_connect().

This is regression caused by commit 
https://github.com/curl/curl/commit/71b7e0161032927cdfb4e75ea40f65b8898b3956#diff-96ac2b9e301765fa5e816e0c41282610e9c0193a5852cc9bb0ed053b1fc4ef7f

Fix: Check if we are already connected and return CURLE_OK.

PR #10651, opened 2023-03-01
test1960: verify CURL_SOCKOPT_ALREADY_CONNECTED

bagder:
When returned from the CURLOPT_SOCKOPTFUNCTION, like when we have a custom socket connected in the app, passed in to libcurl.

Verifies the fix in #10648

### DIFF
--- a/tests/libtest/lib1960.c
+++ b/tests/libtest/lib1960.c
@@ -24,12 +24,12 @@
 #include "test.h"
 
 #ifdef WIN32
+#undef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600 /* override mingw v1's value that prevents the
+                               inet_pton() handling */
 #include <winsock2.h>
 #include <ws2tcpip.h>
 #include <windows.h>
-
-/* because of mingw v1 */
-extern int inet_pton(int af, const char *src, void *dst);
 #endif
 
 #ifdef HAVE_NETINET_IN_H
--- a/tests/libtest/lib1960.c
+++ b/tests/libtest/lib1960.c
@@ -1,0 +1,143 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.haxx.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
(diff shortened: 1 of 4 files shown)


tests/libtest/lib1960.c
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#include "test.h"
```
MarcelRaad, line 26: test.h should be the first include because of the `_FILE_OFFSET_BITS` redifition, see https://github.com/curl/curl/pull/10182 and https://curl.se/dev/log.cgi?id=20230301111924-1540251#prob1.

tests/libtest/lib1960.c
```diff
@@ -0,0 +1,143 @@
...
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+#include "test.h"
+
+#ifdef WIN32
+#undef _WIN32_WINNT
```
MarcelRaad, line 27: Hmm, seems like my comment got applied only to the commit and not shown in the PR:
Or maybe just make HAVE_INET_PTON make a requirement for executing this test? The Windows headers might already be included indirectly by test.h, so this might be too late to redefine the value.

bagder: ah hurt by that *again* ...

bagder: but then, where/when is mingw.h included which is the header that apparently sets the "bad" define?

bagder: According to: https://github.com/msys2/MINGW-packages/issues/6191

bagder: While here: should we perhaps drop mingw v1 support and live a happier life?

bagder: trying a precheck to skip the test if inet_pton is not present

vszakats: Replying to https://curl.se/mail/lib-2023-03/0006.html:

"original" MinGW has been a constant source of issues :( (I've spent many days digging for issue unique to it) Its Windows headers are seriously outdated. The compiler as well. It is (and always was) difficult to install. It only support Intel 32-bit, which by itself is an almost dead platform. Now even its original website (mingw.org) is gone.

It's never an easy decision to drop a platform, compiler, feature, but in this case I'd vote to do so. It takes off a ton of complications, leaving largely a modern mingw-w64 and MSVC to worry about on Windows.

[ Meanwhile mingw-w64 can be installed via MSYS2 or standalone, or using most unixy package managers as well. Its Wine headers are up to date. It's actively developed. ]

bagder:
Success!! :rocket:

MarcelRaad (approved):

Issue #10646, opened 2023-03-01
libcurl 7.88.0 - Connection failed maybe due to the refactor of detecting if a connection is dead.


SendSonS:
With libcurl 7.88.0/7.88.1, we sometimes encountered this issue, that the sending request was 'Expire cleared', it tried to reuse connections but ' Connection died, tried 5 times before giving up'.

Refer to the attached logs:
[logs_libcurl_7_87_0.txt](https://github.com/curl/curl/files/10858698/logs_libcurl_7_87_0.txt)
[logs_libcurl_7_88_0.txt](https://github.com/curl/curl/files/10858699/logs_libcurl_7_88_0.txt)

There're 6 connections in the cache but they are all dead.
With 7.87.0, when handling a new request, it fount that the connection is already dead before sending data (_Connection 2 seems to be dead_), then it remove all dead connections.
With 7.88.0, when handling a new request, it did not remove any dead connection before sending data, but failed to send data (_we are done reading and this is set to close, stop send_), then it retry the next died connection (_Connection died, retrying a fresh connect (retry count: 1)_, finally it failed after retried with 5 times.

We checked the history and found that, before `create_conn` it will `prune_dead_connections`, the detecting method was changed at here:
https://github.com/curl/curl/commit/71b7e0161032927cdfb4e75ea40f65b8898b3956#diff-a8a54563608f8155973318f4ddb61d7328dab512b8ff2b5cc48cc76979d4204c
```
-      dead = SocketIsDead(conn->sock[FIRSTSOCKET]);
+      dead = !Curl_conn_is_alive(data, conn);
```

Looks like it cannot find the connection that is already dead before sending data.

icing:
@SendSonS what platform are you running on?

icing:
Thanks for the report and the logs.

Could your check that #10652 fixes the problem for you as well?

SendSonS:
> Thanks for the report and the logs.
> 
> Could your check that #10652 fixes the problem for you as well?

@icing We still see this issue with your patch.
The log is still as same as before, it re-tried to re-use 5 dead connections.

icing:
I assume the requests involved use `https:`?

icing:
Please find in #10646 my latest fix for putting a lid not this issue.

Background: the refactoring from 7.87 to 7.88 failed to properly preserve the handling of pending input on connection reuse. For `http` connections, having input arriving from a server is considered bad, since the server should not send anything unless there is some issue.

[...]

PR #10652, opened 2023-03-01
Fix for detecing a "dead" connection, e.g. not fit for reuse.

icing:
- refs #10646 where reuse was attempted on closed connections in the cache, leading to an exhaustion of retries on a transfer
- the mistake was that poll events like POLLHUP, POLLERR, etc were regarded as "not dead".
- change cf-socket filter check to regard such events as inidication of corpsiness.
- vtls filter checks: fixed interpretation of backend check result when inconclusive to interrogate status further down the filter chain.


bagder:
Thanks!

PR #10654, opened 2023-03-01
Renaming tests/tests-httpd to tests/http

icing:
- httpd is only one server we test with
- the suite coveres the HTTP protocol in general where the default test cases need a more beefy environment


bagder:
Maybe even just **http** as it already is in the subdirectory called tests?

icing:
> Maybe even just **http** as it already is in the subdirectory called tests?

Yeah, I tried that once. but `pytest` has some internal defaults that trigger on the dir matching "test*".

bagder (approved):

bagder:
thanks!

PR #10656, opened 2023-03-01
Give source files cf-http.* a better fitting name.

icing:
`cf-http.c` sounds like general http handling. But this filter is all about doing https connects.


bagder (approved):

bagder:
Thanks!

PR #10658, opened 2023-03-01
ftp: add HOST support

dorjoy03:
This PR adds HOST support for ftp both in curl cmdline tool and in libcurl (https://curl.se/docs/todo.html#HOST)
The command line option is `--ftp-virtual-host <hostname>` and libcurl option is `CURLOPT_FTP_VIRTUAL_HOST`. I added the virtual keyword in the options just to be clearer.

From what I see in rfc 7151, the HOST command should be sent immediately after the control connection is established and then everything should be like before. (The rfc does mention an exception where if the user-PI sends "server_name" in TLS client hello, then it could send the HOST command after the AUTH TLS command in section 3.2.2 and section 4).

I looked at the todo list in curl's website and thought it wouldn't be too hard to contribute this one. I would be happy to make necessary changes to make this PR ready to get merged for future release. 

Thanks.

**Edit:** The options have been modified to be `--ftp-send-host` for cmdline and `CURLOPT_FTP_SEND_HOST` for libcurl. By default the options will be false. If the options are given, curl will send the HOST command with hostname from the URL.


dorjoy03:
Seems like most build failures are due to test1139. When I ran locally it was failing too and then I ran the gen.pl script which modified some files and then the tests passed. So I committed everything in this branch. I am not sure if I should just remove some changes and force push? Any suggestions on which changes I should not have committed or what to do to make the builds successful?

bagder:
> --ftp-virtual-host [hostname]

Why would a typical user want to pass the name to this option? The typical virtual host is the name given in the URL, isn't it? And if you want to "play tricks" with the name, then we have `--resolve` already that lets you pass in another name in the URL and still connect to the "correct" host.

bagder:
Regarding the test 1139 failure, I think it is because you have not added `ftp-virtual-host.d` to the list of options in `docs/cmdline/Makefile.inc`.

dorjoy03:
> Why would a typical user want to pass the name to this option? The typical virtual host is the name given in the URL, isn't it? And if you want to "play tricks" with the name, then we have `--resolve` already that lets you pass in another name in the URL and still connect to the "correct" host.

I thought that what the ftp url is that is what we connect to and the virtual host parameter can be any private ip or other name may be that the ftp server can decide if it exists or not, right? What should be the approach here? I need some help here I think.



> Regarding the test 1139 failure, I think it is because you have not added `ftp-virtual-host.d` to the list of options in `docs/cmdline/Makefile.inc`

Thanks for this. I will fix this.

bagder:
> I thought that what the ftp url is that is what we connect to and the virtual host parameter can be any private ip or other name may be that the ftp server can decide if it exists or not, right?

Virtual name hosting is the thing we've done for HTTP since HTTP/1.1: we have multiple hosts for the same IP address:  `foobar` and `barfoo` both points to `192.168.0.1`. When the client speaks to the server on that IP address it tells the server which of the many names we want talk to. In HTTP we do it with the `Host:` headers and with HTTPS we do it with the SNI field in TLS. For FTP that feature has never been widespread but the way to do it is the `HOST` command.

The name we want is the name from the URL.

dorjoy03:
> The name we want is the name from the URL.

Got it. Thanks a lot.
Should I just re-do the PR to 'implicitly' send the HOST command with hostname from the URL and do a force push to make it reviewable? So, we will always be sending HOST command with the hostname from the URL and we don't really need the virtual host options added. Am I understanding it correctly?
The rfc 7151 does mention that literal ip address can also be sent as the hostname but it must match the server's literal ip address. We don't need to think about this. We can just send the hostname from the URL, right?

bagder:
I think we need to tread gently and only send the `HOST` command if the extra option is given. I'm pretty sure sending the command on *all* transfers would cause problems (and you would have to update hundreds of test cases!)

dorjoy03:
> I think we need to tread gently and only send the `HOST` command if the extra option is given. I'm pretty sure sending the command on _all_ transfers would cause problems (and you would have to update hundreds of test cases!)

You are right. So should I make the option 'boolean' (instead of being a string and requiring an argument) both in cmdline tool and libcurl so that if the option is given we send the HOST command with hostname from the URL?

dorjoy03:
> I think we need to tread gently and only send the `HOST` command if the extra option is given. I'm pretty sure sending the command on _all_ transfers would cause problems (and you would have to update hundreds of test cases!)

Thanks for your input. 
I have now updated the PR (did a force push). The cmdline option is `--ftp-send-host` and libcurl option is `CURLOPT_FTP_SEND_HOST`. By default, no HOST command will be sent. If the option is provided HOST command will be sent with hostname from the URL.

Some builds are failing. Not sure if I need some changes in my branch to make them pass.

bagder:
Have this functionality been verified against a real FTP server with support for `HOST` ?

dorjoy03:
@bagder Let me know if this PR needs more changes to make it reviewable. Also I need suggestions on how I can test with a real server with HOST support. Thanks.

bagder:
> I need suggestions on how I can test with a real server with HOST support

ProFTPD is the only software I've found that [claims to support this option](http://www.proftpd.org/docs/howto/FTP.html). It makes me realize and think adding support for `HOST` before any users actually want it for actual and real use cases is premature. It does not seem to be in use "out there".

dorjoy03:
> It makes me realize and think adding support for `HOST` before any users actually want it for actual and real use cases is premature. It does not seem to be in use "out there".

Yes. We can revisit/reopen this PR if it makes sense to add this support in curl then. Closing it for now. Thanks.

PR #10660, opened 2023-03-02
urlapi: parse IPv6 literals without ENABLE_IPV6

b-spencer:
As per [the discussion on the curl-library mailing list](https://curl.se/mail/lib-2023-02/0043.html), here is a patch to allow URLs with IPv6 literals to be parsed correctly even when `ENABLE_IPV6` is not configured, including when the target lacks the `AF_INET6` constant (and other IPv6 definitions).

It simply enables the existing curl IPv6 literal parsing capabilities when `ENABLE_IPV6` is not defined, and creates a limited-scope definition of `AF_INET6` that the URL and IP parsing code (only) use to communicate.  It seemed safer to not define `AF_INET6` more widely, and it seemed unnecessarily complex to switch curl to define its own constant and use it, so I took a simple pragmatic approach.  Feedback on this detail, including the resulting redefinition of `AF_INET6` in a three spots, is welcome.

For tests, I simply relaxed the restrictions on the existing tests to now require IPv6 support.  I do not know if any CI runs the tests this way (but I did manually).

I don't have a machine that lacks `AF_INET6`, so I also simulated that case by undefining it.  

All tests passed for me on Linux x86_64.

### DIFF
--- a/docs/libcurl/libcurl-url.3
+++ b/docs/libcurl/libcurl-url.3
@@ -137,6 +137,9 @@ Now the URL looks like
 .fi
 .SH AVAILABILITY
 The URL API was introduced in libcurl 7.62.0.
+
+A URL with a literal IPv6 address can be parsed even when IPv6 support is not
+enabled.
 .SH "SEE ALSO"
 .BR curl_url "(3), " curl_url_cleanup "(3), " curl_url_get "(3), "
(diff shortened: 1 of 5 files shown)


bagder (approved):

bagder:
Thanks!

PR #10662, opened 2023-03-02
Add test for socks proxy using unix sockets and connect to https

balki:
Similar to test1468 except using https instead of http

Connecting to http**s** via unix socks proxy was broken since [this](https://github.com/curl/curl/commit/671158242db32032a41e8da89da33d773b481ef8) commit (Feb 1 version 7.88) and was fixed [here](https://github.com/curl/curl/commit/a4d015e69f5ef4b6ab0e505257727c217801072e). Adding test to make sure it is not broken again.

### DIFF
--- a/tests/data/test1470
+++ b/tests/data/test1470
@@ -1,0 +1,64 @@
+<testcase>
+<info>
+<keywords>
+HTTPS
+HTTP
+HTTP GET
+SOCKS5
+SOCKS5h
+</keywords>
+</info>
+
+#
+# Server-side
+<reply>
+<data>
+HTTP/1.1 200 OK
+Date: Tue, 09 Nov 2010 14:49:00 GMT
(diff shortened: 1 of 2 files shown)


tests/data/test1470
```diff
@@ -0,0 +1,63 @@
...
+<features>
+proxy
+unix-sockets
+</features>
+<server>
+https
+socks5unix
+</server>
+ <name>
+HTTP GET with host name using SOCKS5h via unix sockets
```
bagder, line 42: maybe even `HTTPS GET`?

bagder (approved):
Very nice!

tests/data/test1470
balki: Changed. Thanks!

bagder:
Thanks!

Issue #10666, opened 2023-03-03
Issue with active FTPS connection after updating to libcurl 7.87.0/7.88.0


SandakovMM:
Hello
I am experiencing issues with FTP connection after updating to libcurl 7.87.0 and 7.88.0. Unable to reproduce the issue on 7.86.0.

### I did this
There is a simple C utility to reproduce the issue:
```C
int debug_curl(CURL* handle, curl_infotype type, char* data, size_t size, void* userptr)
{
    switch (type) {
    case CURLINFO_TEXT:
        printf("Curl output: * %s\n", data);
        break;
    case CURLINFO_HEADER_IN:
        printf("Curl output: < %s\n", data);
        break;
    case CURLINFO_HEADER_OUT:
        printf("Curl output: > %s\n", data);
        break;
    case CURLINFO_DATA_IN:
    case CURLINFO_DATA_OUT:
    case CURLINFO_SSL_DATA_IN:
    case CURLINFO_SSL_DATA_OUT:
    case CURLINFO_END:
        break;
    }
    return 0;
}

int result(char *buf, size_t size, size_t nmemb, void *context)
{
    printf("result: %s\n", buf);
    return 0;
}

int main(int argc, char **argv)
{
    char *url_with_user = argv[1];
    char *password = argv[2];

    CURL *curl = curl_easy_init();
    struct curl_slist *m_headers = NULL;
    if (!curl) {
        printf("unable to init curl");
        return 1;
    }

    curl_easy_setopt(curl, CURLOPT_TIMEOUT, 300);
    curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, 30);
    curl_easy_setopt(curl, CURLOPT_FTP_RESPONSE_TIMEOUT, 60);
    curl_easy_setopt(curl, CURLOPT_TCP_KEEPALIVE, 1);
    curl_easy_setopt(curl, CURLOPT_TCP_KEEPIDLE, 60);
    curl_easy_setopt(curl, CURLOPT_TCP_KEEPINTVL, 60);

    curl_easy_setopt(curl, CURLOPT_DEBUGFUNCTION, debug_curl);
    curl_easy_setopt(curl, CURLOPT_DEBUGDATA, NULL);
    curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);

    curl_easy_setopt(curl, CURLOPT_FTPPORT, "-");
    curl_easy_setopt(curl, CURLOPT_FTP_USE_EPRT, 1L);

[...]

PR #10669, opened 2023-03-03
ftp active mode with SSL, add the [...] filter

icing:
- since 7.87.0 we lost adding the SSL filter for an active FTP connection that uses SSL. This leads to hangers and timeouts as reported in #10666.


bagder:
Thanks!

Issue #10670, opened 2023-03-03
time_connect is 0 when error on 7.88.0/7.88.1


masterinspire:
# time_connect is 0 when error

### I did this

I can connect to `1.1.1.1`:

```
curl "https://1.1.1.1" --output nul --write-out "%{time_connect}" --no-progress-meter -slient
```

`time_connect` here is > 0.

**Deliberately** point `example.com` to `1.1.1.1` to cause a ssl error:

```
curl "https://example.com" --output nul --write-out "%{time_connect}" --no-progress-meter --resolve "example.com:443:1.1.1.1" -slient
```

**`time_connect` here is `0.000000` on `7.88`, > 0 on `7.87`.**

![image](https://user-images.githubusercontent.com/9047578/222750547-ccb8cc9e-4785-4df5-bdb9-c867793c6862.png)

### I expected the following

`time_connect` > 0.

### curl/libcurl version

curl 7.88.1 (x86_64-w64-mingw32) libcurl/7.88.1 OpenSSL/3.0.8 (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 WinIDN libssh2/1.10.0 nghttp2/1.52.0 ngtcp2/0.13.1 nghttp3/0.8.0 libgsasl/2.2.0
Release-Date: 2023-02-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe TLS-SRP UnixSockets zstd

### operating system

Microsoft Windows [Version 10.0.19045.2673]

PR #10671, opened 2023-03-03
Fix time_connect and time_appconnect timer statistics

icing:
- time_connect was not updated when the overall connection failed, e.g. when SSL verification was unsuccessful, refs #10670
- rework gather those values to interrogate involved filters, also from all eyeballing attempts, to report the maximum of those values.
- added 3 test cases in test_06 to check reported values on successful, partially failed and totally failed connections.


bagder:
Thanks!

PR #10672, opened 2023-03-04
rand: use arc4random as fallback when available

piru:
Normally curl uses cryptographically strong random provided by the selected SSL backend. If compiled without SSL support, a naive built-in function was used instead.

Generally this was okay, but it will result in some downsides for non- SSL builds, such as predictable temporary file names.

This change ensures that arc4random will be used instead, if available.


bagder (approved):

bagder:
Thanks!

Issue #10491, opened 2023-02-13
--stderr does not direct all stderr writes to file


u20221022:
### I did this
`curl.exe -I --stderr err.txt --write-out "%{stderr}%{json}" "https://curl.se/" > Nul`
{"content_type":"text/html","errormsg":null,"exitcode":0,"filename_effective":null,"ftp_entry_path":null,"http_code":200,"http_connect":0,"http_version":"2","local_ip":"?.?.?.?","local_port":?,"method":"HEAD","num_connects":1,"num_headers":21,"num_redirects":0,"proxy_ssl_verify_result":0,"redirect_url":null,"referer":null,"remote_ip":"151.101.65.91","remote_port":443,"response_code":200,"scheme":"HTTPS","size_download":0,"size_header":845,"size_request":70,"size_upload":0,"speed_download":0,"speed_upload":0,"ssl_verify_result":0,"time_appconnect":1.604500,"time_connect":0.221004,"time_namelookup":0.019257,"time_pretransfer":1.384094,"time_redirect":0.000000,"time_starttransfer":1.469586,"time_total":1.470412,"url":"https://curl.se/","url_effective":"https://curl.se/","urlnum":0,"curl_version":"libcurl/7.87.0 OpenSSL/3.0.8 (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 WinIDN libssh2/1.10.0 nghttp2/1.51.0 ngtcp2/0.13.1 nghttp3/0.8.0 libgsasl/2.2.0"}
`dir err.txt`
 Том в устройстве ? имеет метку ?
 Серийный номер тома: ?

 Содержимое папки ?:\?\curl-7.87.0_7-win64-mingw\bin

13.02.2023  05:00                 0 err.txt
               1 файлов              0 байт
               0 папок   ? байт свободно
`curl.exe -I --stderr err.txt --write-out "%{stderr}%{json}" "htps://curl.se/" > Nul`
[...]

PR #10673, opened 2023-03-04
tool: improve --stderr handling

jay:
- freopen stderr with the user-specified file (--stderr file) instead of using a separate 'errors' stream.

- Override stderr macro in tool_setup as global variable tool_stderr.

Both freopen and overriding the stderr macro are necessary because if the user-specified filename is "-" then stdout is assigned to tool_stderr and no freopen takes place.

Ref: https://github.com/curl/curl/issues/10491

Closes #xxxx

---

Note --stderr maps to libcurl's CURLOPT_STDERR but they are documented differently:

[--stderr](https://curl.se/docs/manpage.html#--stderr) :
"Redirect all writes to stderr"

[CURLOPT_STDERR](https://curl.se/libcurl/c/CURLOPT_STDERR.html) :
"when showing the progress meter and displaying CURLOPT_VERBOSE data"

If the curl tool has freopened stderr as a file (--stderr file) then libcurl stderr output will also go to the file. OTOH if stdout was specified (--stderr -) then libcurl writes to stderr go to actual stderr and just the criteria above is written to stdout. IMO the distinction is moot since in libcurl we only write to stderr for developer debug output.

### DIFF
--- a/src/Makefile.inc
+++ b/src/Makefile.inc
@@ -82,6 +82,7 @@ CURL_CFILES = \
   tool_paramhlp.c \
   tool_parsecfg.c \
   tool_progress.c \
+  tool_stderr.c \
   tool_strdup.c \
   tool_setopt.c \
   tool_sleep.c \
@@ -126,6 +127,7 @@ CURL_HFILES = \
   tool_setopt.h \
   tool_setup.h \
   tool_sleep.h \
+  tool_stderr.h \
   tool_strdup.h \
   tool_urlglob.h \
   tool_util.h \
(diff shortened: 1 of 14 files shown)


bagder:
I'm not sure I see how this simplifies or improves the current situation, but if you think so then I am fine with it.

jay:
It's an improvement to fprintf(stderr... rather than a separate error stream, when all stderr output is supposed to be redirected, since in the latter case we forgot at least once which led to a bug. It's failing some CI so I'm still working on it.

PR #10674, opened 2023-03-04
cmake: fix enabling LDAPS on Windows

### LINKED ISSUE #6284: [cmake] Can't use feature ldaps on Windows
Hi guys:

Recently, we received an issue about curl (https://github.com/microsoft/vcpkg/issues/14030), that reported that the feature `ldaps`  could not be enabled on Windows.
Since curl in vcpkg uses cmake to build, I checked the cmake configuration in curl and found something strange:
https://github.com/curl/curl/blob/abd846c374c5269809ce11740754bc6f00a0b193/CMakeLists.txt#L511-L520
Here we can use the system's ldaps on windows, but openldap is not disabled here, and:
https://github.com/curl/curl/blob/abd846c374c5269809ce11740754bc6f00a0b193/CMakeLists.txt#L1398-L1400
There is no judgment variable `USE_WIN32_LDAP` when judging whether to enable ldaps.

I think that's caused this issue.
But even if I use `USE_WIN32_LDAP` and explicitly enable ldaps in the configuration log, using curl -V does not explicitly enable ldaps.
Configure log:
```
-- Enabled protocols: DICT FILE FTP FTPS GOPHER HTTP HTTPS IMAP IMAPS LDAP LDAPS MQTT POP3 POP3S RTSP SMB SMBS SMTP SMTPS TELNET TFTP
```
Result:
```
curl 7.73.0-DEV (Windows) libcurl/7.73.0-DEV Schannel zlib/1.2.11
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher http https imap imaps ldap mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: AsynchDNS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI TrackMemory libz
```

### I expected the following
ldaps should be enabled on Windows.

### curl/libcurl version
7.73.0

[curl -V output]
```
curl 7.73.0-DEV (Windows) libcurl/7.73.0-DEV Schannel zlib/1.2.11
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher http https imap imaps ldap mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: AsynchDNS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI TrackMemory libz
```

### operating system
Windows 10, Visual Studio 2017, cmake 3.18.2

vszakats:
Before this patch, enabling LDAPS required a manual C flag: https://github.com/curl/curl-for-win/blob/c1cfc31cfc04f24f7a4f946564d6f0e1b4d7dd36/curl-cmake.sh#L105

Fix this and enable LDAPS automatically when using `wldap32` (and when not explicitly disabled). This matches autotools and `Makefile.mk` behavior. Also remove issue from KNOWN_BUGS.

Add workaround for this MSVS 2010 warning triggered by having LDAPS enabled in more CI tests:
`ldap.c(360): warning C4306: 'type cast' : conversion from 'int' to 'void *' of greater size`
Ref: https://ci.appveyor.com/project/curlorg/curl/builds/46408284/job/v8mwl9yfbmoeqwlr#L312

Reported-by: JackBoosY on github
Fixes #6284
Closes #10674


MarcelRaad (approved):

vszakats:
MSVS 2010 says:
```
ldap.c(360): error C2220: warning treated as error - no 'object' file generated [C:\projects\curl\lib\libcurl.vcxproj]
ldap.c(360): warning C4306: 'type cast' : conversion from 'int' to 'void *' of greater size [C:\projects\curl\lib\libcurl.vcxproj]
```

`LDAP_OPT_ON` is `((void *) 1)`

lib/ldap.c
```diff
@@ -357,7 +357,15 @@ static CURLcode ldap_do(struct Curl_easy *data, bool *done)
...
     server = ldap_sslinit(host, conn->port, 1);
+#if defined(_MSC_VER) && (_MSC_VER <= 1600)
+#pragma warning(push)
+/* 'type cast' : conversion from 'int' to 'void *' of greater size */
+#pragma warning(disable:4306)
+#endif
     ldap_set_option(server, LDAP_OPT_SSL, LDAP_OPT_ON);
+#if defined(_MSC_VER) && (_MSC_VER <= 1600)
+#pragma warning(pop)
+#endif
```
jay, line 368: the macros could be overridden instead

~~~diff
diff --git a/lib/ldap.c b/lib/ldap.c
index 5e53f4c..dcfd3c4 100644
--- a/lib/ldap.c
+++ b/lib/ldap.c
@@ -140,6 +140,13 @@ static void _ldap_free_urldesc(LDAPURLDesc *ludp);
 #define ldap_err2string ldap_err2stringA
 #endif
 
+#ifdef USE_WIN32_LDAP
+#undef LDAP_OPT_ON
+#undef LDAP_OPT_OFF
+#define LDAP_OPT_ON                 ((void *)(size_t)1)
+#define LDAP_OPT_OFF                ((void *)(size_t)0)
+#endif
 
 static CURLcode ldap_do(struct Curl_easy *data, bool *done);
 
~~~

jay (approved):
~~~
curl 8.0.0-DEV (Windows) libcurl/8.0.0-DEV Schannel
Release-Date: [unreleased]
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI threadsafe UnixSockets
~~~

lib/ldap.c
vszakats: This can work, though I'd prefer to guard it for the old MSVS version. This also made it clear why it's there and allowed easy removal once we drop MSVS support in the future.

PR #10675, opened 2023-03-04
tool: dump headers even if file is write-only

algorythmic:
The fixes in #10079 brought a (seemingly unrelated) change of open mode from `wb`/`ab` to `wb+`/`ab+` for the headerfile. This makes it no longer possible to write the header file to e.g. a pipe, like:

    curl -D >(grep ...) file:///dev/null

Which presently results in `Warning: Failed to open /dev/fd/63`

See #10079
Closes #10675


bagder (approved):

bagder:
Thanks!

PR #10687, opened 2023-03-06
docs: note '--data-urlencode' option

andy5995:


bagder:
I suppose we call it "URL encode" in plain English?

.github/scripts/spellcheck.words
```diff
@@ -810,6 +810,7 @@ URI
 URIs
 url
 URL's
+urlencode
```
bagder, line 813: I disagree. This is not an English word. You would write it as URL encode.

andy5995: Ok, I'll make the change. But 'urlencoded' is used elsewhere in the document and already on the whitelist.

bagder: Yeah, I noticed. But I think maybe those should rather be fixed as well at some point?

bagder (approved):

bagder:
Thanks!

PR #10690, opened 2023-03-06
Fix logic in connection reuse to deny reuse on "unclean" connections

icing:
- add parameter to `conn_is_alive()` cfilter method that returns if there is input data waiting on the connection
- refrain from re-using connnection from the cache that have input pending
- adapt http/2 and http/3 alive checks to digest pending input to check the connection state
- remove check_cxn method from openssl as that was just doing what the socket filter now does.
- add tests for connection reuse with special server configs

Refs #10646


bagder:
Thanks!

PR #10692, opened 2023-03-06
Use dynamic ports numbers in pytest suite

icing:
- necessary ports are bound at start of test suite and then given to server fixtures for use.
- this make parallel use of pytest (in separate directories), practically safe for use as OS tend to not reuse such port numbers for a while


bagder:
Thanks!

PR #10693, opened 2023-03-06
Fix HTTP/2 handling of RST and GOAWAY to recognize partial transfers

icing:
- a reset transfer (HTTP/2 RST) did not always lead to the proper error message on receiving its response, leading to wrong reports of a successful transfer
- test_05_02 was able to trigger this condition with increased transfer count. The simulated response errors did not carry a 'Content-Length' so only proper RST handling could detect the abort
- When doing such transfers in parallel, a connection could enter the state where a) it had been closed (GOAWAY received) b) the RST had not been "seen" for the transfer yet or c) the GOAWAY announced an error and the last successful stream id was not checked against ongoing transfers


bagder:
Thanks!

PR #10699, opened 2023-03-07
Adding the pytest suite to github workflows

icing:
Adding pytest to several github workflows

    - added to: ngtcp2-quictls, ngtcp2-gnutls and the linux varians
      quiche, bearssl, libressl, mbedtls, openssl3, rustls
    - added disabled in ngtcp2-wolfssl due to weird SSL_connect() errors
      not reproducable locally

    Improvements on pytest:

    -  handling of systems with nghttpx in $PATH
       - configure will seach $PATH got nghttpx used in pytest
       - pytest fixes for managing nghttpx without h3 support
       - ngtcp2-wolfssl: use a fully enabled wolfssl build

    - lower parallel count for http/1.1 tests, since we do not
       want to test excessive connections.
    - check built curl for HTTPS-proxy support in proxy tests
    - bearssl does not like one of our critical cert extensions, making
      it non-critical now
    - bearssl is too slow for test_12, skipping
    - making sure we do h3 tests only when curl and server support is there


bagder:
Thanks!

PR #10705, opened 2023-03-07
Refuse to resolve the `.onion` TLD.

Kangie:
RFC 7686 states that:

> Applications that do not implement the Tor
> protocol SHOULD generate an error upon the use of .onion and
> SHOULD NOT perform a DNS lookup.

Let's do that.

See curl/curl#543
https://www.rfc-editor.org/rfc/rfc7686#section-2

-----

I'm certain that this will inconvenience some people; I feel that any inconvenience is far outweighed by the benefits for those using (or trying to use) Tor.

I initially made this change into a default-on feature but decided against it. We're not Tor-aware, we should just refuse to resolve the `.onion` TLD.

If there's interest in making this a feature I'm willing to go back and do that so that anyone with a valid use case will be able to disable RFC 7686 protections while those that _need_ to be sure that we won't accidentally leak information can clearly see that their binary has this 'feature'.

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -652,6 +652,14 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
   CURLcode result;
   enum resolve_t rc = CURLRESOLV_ERROR; /* default to failure */
   struct connectdata *conn = data->conn;
+  /* We should intentionally error and not resolve .onion TLDs */
+  size_t hostname_len = strlen(hostname);
+  if(hostname_len >= 7 &&
+  (curl_strequal(&hostname[hostname_len-6], ".onion") ||
+  curl_strequal(&hostname[hostname_len-7], ".onion."))) {
+    failf(data, "Not resolving .onion address (RFC 7686)");
+    return CURLRESOLV_ERROR;
+  }
   *entry = NULL;
 #ifndef CURL_DISABLE_DOH
   conn->bits.doh = FALSE; /* default is not */
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -80,7 +80,6 @@ problems may have been fixed or changed somewhat since this was written.
  10.3 FTPS over SOCKS
 
  11. Internals
- 11.1 Curl leaks .onion hostnames in DNS
  11.2 error buffer not set if connection to multiple addresses fails
  11.4 HTTP test server 'connection-monitor' problems
  11.5 Connection information when using TCP Fast Open
@@ -524,14 +523,6 @@ problems may have been fixed or changed somewhat since this was written.
 
 
 11. Internals
-
-11.1 Curl leaks .onion hostnames in DNS
-
- Curl sends DNS requests for hostnames with a .onion TLD. This leaks
- information about what the user is attempting to access, and violates this
- requirement of RFC7686: https://datatracker.ietf.org/doc/html/rfc7686
-
- Issue: https://github.com/curl/curl/issues/543
 
 11.2 error buffer not set if connection to multiple addresses fails
 
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -186,8 +186,8 @@ test1432 test1433 test1434 test1435 test1436 test1437 test1438 test1439 \
 test1440 test1441 test1442 test1443 test1444 test1445 test1446 test1447 \
 test1448 test1449 test1450 test1451 test1452 test1453 test1454 test1455 \
 test1456 test1457 test1458 test1459 test1460 test1461 test1462 test1463 \
-test1464 test1465 test1466 test1467 test1468 test1469 test1470 \
-\
+test1464 test1465 test1466 test1467 test1468 test1469 test1470 test1471 \
+test1472 \
 test1500 test1501 test1502 test1503 test1504 test1505 test1506 test1507 \
 test1508 test1509 test1510 test1511 test1512 test1513 test1514 test1515 \
 test1516 test1517 test1518 test1519 test1520 test1521 test1522 test1523 \
--- a/tests/data/test1471
+++ b/tests/data/test1471
@@ -1,0 +1,39 @@
+<testcase>
+<info>
+<keywords>
+Onion
+Tor
+FAILURE
+</keywords>
+</info>
+#
+# Server-side
+<reply>
+</reply>
+
+#
+# Client-side
+<client>
+<server>
+none
+</server>
+<name>
+Fail to resolve .onion TLD
+</name>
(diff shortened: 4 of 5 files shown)


Kangie:
I'll fix up that CI after some sleep. :smile:

lib/hostip.c
```diff
@@ -659,6 +659,13 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
   (void)allowDOH;
 #endif
 
+  /* We should intentionally error and not resolve .onion TLDs */
+  char *last_dot = strrchr(hostname, '.');
+  if(last_dot && strcmp(".onion", last_dot) == 0) {
+    infof(data, "Not resolving .onion address (RFC 7686)");
```
bagder, line 665: 1. This should probably be failf() instead of infof().
2. We usually use `!strcmp()...` instead of comparing with `== 0`.
3. Shouldn't the check be case insensitive? Host names are.

danielgustafsson:
I haven't had time to read this one yet, but we've been down this rabbithole before so it's worth making sure we're not repeating issues discovered in #5159

Kangie:
The following is still technically an issue here:

> The main problem with this approach is that asking for a host foo.onion doesn't necessary equal a DNS resolve of that name, but could be a local host name that gets the domain appended in the resolve so it would ask for foo.onion.domain.tld

While that's strictly correct, how many people are _actually_ using two subdomains **and** relying on DNS search suffixes to fill in the rest? :stuck_out_tongue: On a more serious note, to actually be RFC compliant we should *never* request a DNS lookup for a request ending in `.onion`; it's a bit of an edge case but I'd make the argument that it's irrelevant in the real world.

A bug report from a user who reported that they could *sometimes* make curl resolve `.onion` TLDs in their somewhat kluged solution brought the original note in the KNOWN_BUGS and its associated issue to my attention. I figure that it's safer for everyone if curl isn't accidentally leaking `.onion` requests until someone decides to implement a proper Tor backend.

Kangie:
Results of my investigation from [Gentoo Bugzilla](https://bugs.gentoo.org/887287):

>... the adns/c-ares backend was updated at some point to be RFC 7686-compliant. A quick google confirms this happened in 2018:
>
> https://github.com/c-ares/c-ares/issues/196
>
>As this is the default in Gentoo, curl/libcurl will send a request to c-ares and get back "NXDOMAIN" as per the RFC:
>
>> Name Resolution APIs and Libraries (...) MUST either respond to requests for
>> .onion names by resolving them according to [tor-rendezvous] or by responding
>> with NXDOMAIN.
>
>When you disable that, (as you have here) curl simply sends the request directly using system DNS.

bagder:
> Bedtime thought, .onion. is a case that I suspect should to be handled as well, thoughts?

Yes!

Kangie:
If I understand test syntax correctly this is looking pretty good now. I'm not sure about checking the `strdup` for an OOM; we're at about the edge of my C knowledge at this point :smile: 

Happy to address any feedback you may have!

lib/hostip.c
```diff
@@ -652,13 +652,31 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
   CURLcode result;
   enum resolve_t rc = CURLRESOLV_ERROR; /* default to failure */
   struct connectdata *conn = data->conn;
+  /* We should intentionally error and not resolve .onion TLDs */
+  bool found_onion = FALSE;
+  char *last_dot = strrchr(hostname, '.');
+  if(last_dot) {
+    char *stripped_hostname = strdup(hostname);
```
danielgustafsson, line 659: Need to check `stripped_hostname` for NULL in case of OOM.

lib/hostip.c
```diff
@@ -652,13 +652,31 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
...
+      return CURLRESOLV_ERROR;
+    }
+  }
   *entry = NULL;
 #ifndef CURL_DISABLE_DOH
   conn->bits.doh = FALSE; /* default is not */
 #else
   (void)allowDOH;
 #endif
-
```
danielgustafsson, line 661: Please avoid unrelated whitespace changes.

lib/hostip.c
```diff
@@ -652,13 +652,31 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
...
+  bool found_onion = FALSE;
+  char *last_dot = strrchr(hostname, '.');
+  if(last_dot) {
+    char *stripped_hostname = strdup(hostname);
+    if(strlen(last_dot) == 1) {
+      /* We probably got the trailing dot on a FQDN */
+      stripped_hostname[strlen(hostname)-1] = '\0';
+      last_dot = strrchr(stripped_hostname, '.');
+    }
+    if(last_dot && curl_strequal(".onion", last_dot)) {
```
danielgustafsson, line 665: Are we at this point certain we don't have multibyte chars in `last_dot` such that `curl_strequal` works as intended? It might not be a problem in practice, but comparing apples to oranges have bitten code in the past.

lib/hostip.c
```diff
@@ -652,13 +652,31 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
...
   enum resolve_t rc = CURLRESOLV_ERROR; /* default to failure */
   struct connectdata *conn = data->conn;
+  /* We should intentionally error and not resolve .onion TLDs */
+  bool found_onion = FALSE;
+  char *last_dot = strrchr(hostname, '.');
+  if(last_dot) {
+    char *stripped_hostname = strdup(hostname);
+    if(strlen(last_dot) == 1) {
+      /* We probably got the trailing dot on a FQDN */
+      stripped_hostname[strlen(hostname)-1] = '\0';
```
danielgustafsson, line 662: Nitpick perhaps, but this should do `strlen(stripped_hostname)`

lib/hostip.c
danielgustafsson, line 659: Another solution to save an allocation would be to inspect `hostname` with `curl_strnequal()` and limiting the length in case of a trailing dot.

bagder: We like avoiding allocations!

Kangie:
I'll see what I can do about that over the weekend!

bagder:
Thanks!

catharsis71:
What if I'm using the ``--resolve`` option?  Shouldn't I be able to use .onion in that case since there's no DNS involved?

currently this is rejected:
``
curl -I http://aaa.onion/ --resolve "aaa.onion:80:127.0.0.1"
``

but ``curl -I http://127.0.0.1/ -H "Host: aaa.onion"`` (which does basically the same thing) is allowed and works fine

Issue #10694, opened 2023-03-06
 libcurl v7.88.1 crash on cleanup with missing cookie file


Keitagit-kun:
I get crush with libcurl v7.88.1 on call to curl_easy_cleanup() if cookies file is not exists originally. 
If I manually create blank text-file (with relevant filename) - there a no crush and file content successfully written (in my test case only header as there is no cookies).
Without CURLOPT_COOKIEJAR option set - there a no crush...
With libcurl v7.86.0 crush did not happen, but also cookies file was not created/written.
If I remember right, with some previous version cookiesfile was created without problems...
Cookes engine setup (Windows):
```
curl_easy_setopt(cURLHandle, CURLOPT_COOKIESESSION, 1L)
curl_easy_setopt(cURLHandle, CURLOPT_COOKIEFILE, sCookiesFilePath)
curl_easy_setopt(cURLHandle, CURLOPT_COOKIEJAR, sCookiesFilePath)
```
on beginning of connection log there a message:
WARNING: failed to open cookie file "* full file path to cookie file *"

### I expected the following
 I expected cookes file is created/written by libcurl...

bagder:
Thanks a lot for your report!

I've failed to reproduce this problem myself, so can I ask you to provide us with more details on how to go ahead and repeat this. Preferably with a command line or a stand-alone program we can run from our ends against a public URL to trigger the problem?

bagder:
From my reading of the description, this command line should trigger the problem?

~~~
$ ls -l foo 
ls: cannot access 'foo': No such file or directory
$ curl curl.se -c foo -b foo -o /dev/null

SergeyRyabinin:
Hello @bagder ,

We observe a similar issue (but a memory leak) with CURLOPT_COOKIEFILE in combination with curl_easy_reset.
In a recent change https://github.com/curl/curl/commit/af5999a6742ea90011e7fa08aade7eac9943b76a, there was an update to move cookies from `data->state` to `data->set`.

However, curl_easy_reset performs a non-conditional memset on `data->set`, however, [the documentation](https://curl.se/libcurl/c/curl_easy_reset.html) was not updated
> It does not change the following information kept in the handle: live connections, the Session ID cache, the DNS cache, the cookies, the shares or the alt-svc cache.

I believe that curl_easy_reset shall not cause a memory leak and do not call memset or store-and-restore cookie data to-from a temp variable.

Best regards,
Sergey

jay:
[...]

PR #10709, opened 2023-03-08
url: fix cookielist memleak if curl_easy_reset

jay:
- Free set.cookelist in Curl_freeset instead of Curl_close.

Prior to this change the cookielist linked list wasn't freed by curl_easy_reset which calls Curl_freeset to free all set.

Bug: https://github.com/curl/curl/issues/10694#issuecomment-1458619157
Reported-by: Sergey Ryabinin

Closes #xxxx

---

/cc @SergeyRyabinin


bagder:
Did you look into making a test case that reproduces/verifies?

PR #10711, opened 2023-03-08
urlapi: '%' is illegal in host names

bagder:
Update test 1560 to verify

Ref: #10708

### DIFF
--- a/lib/urlapi.c
+++ b/lib/urlapi.c
@@ -629,7 +629,7 @@ static CURLUcode hostname_check(struct Curl_URL *u, char *hostname,
   }
   else {
     /* letters from the second string are not ok */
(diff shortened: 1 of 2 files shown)


rcombs:
Percent signs in the host field can be desirable for e.g. IPv6 address literals: on some platforms, link-local addresses may be followed by `%[interface]` as an extension, e.g. `curl http://[fe80::…%en0]/` works in the current version on macOS and FreeBSD.

bagder:
That's not in a host name though, that's in an IPv6 address and we have tests verifying that to work fine still.

You can also use percent-encoding om host names and then percents are also present and that too is fine.

rcombs:
Ah, neat! Okay, sounds reasonable then.

PR #10713, opened 2023-03-08
test420: add cookies keyword

MarcelRaad:
It fails with `--disable-cookies`.

As seen in the autobuilds: https://curl.se/dev/log.cgi?id=20230308051817-71055#prob1


bagder:
> I wonder why I didn't do that earlier and still disable them via the label.

I presume we started out using the label approach before we made it a testable feature.

bagder (approved):

PR #10714, opened 2023-03-09
tls: remove "successful" messages about cert locations

bagder:
Pointless

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -3150,7 +3150,7 @@ static CURLcode populate_x509_store(struct Curl_cfilter *cf,
           return result;
       }
       if(imported_native_ca)
-        infof(data, "successfully imported Windows CA store");
+        infof(data, "imported Windows CA store");
       else
         infof(data, "error importing Windows CA store, continuing anyway");
     }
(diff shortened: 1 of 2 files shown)


jay:
I think if there is an error setting locations then that is useful information.

bagder:
> I think if there is an error setting locations then that is useful information.

Why is that useful knowledge when the user as asked to ignore the results of it?

bagder:
We don't normally output the reverse of error messages in curl. If things work fine, that's the expected behavior. If it does not, we output errors. Unless there are details in the success that could be good to reveal.

I couldn't find any such details here that are good to know in the event of success.

jay:
> Why is that useful knowledge when the user has asked to ignore the results of it?

It's not as simple as that. That code is part of a much larger block if(verifypeer) meaning you only see those verbose warnings if you have peer verification enabled and it can continue on using some other verification (iirc)

bagder:
Yes, but that's an optimization missing. In OpenSSL land we no longer even try to load them if no verification is being done and we should do the same for wolfSSL to gain valuable time when the user obviously doesn't care anyway.

jay:
I don't think we are on the same page. I'm talking specifically of removal of the verbose warnings when the code continues on error.

For example with OpenSSL we can import native CA certs in Windows in addition to the CA bundle. It could be peer verification is enabled and the native CA certs are imported but the CA bundle isn't due to error.

bagder:
ok, I still don't like the messages but this is not the hill for me

PR #10715, opened 2023-03-09
HTTP/2, fix error handling during parallel operations

icing:
RST and connection close were not handled correctly during parallel transfers, leading to aborted response bodies being reported complete.


bagder:
Thanks!

PR #10716, opened 2023-03-09
wolfSSL, resurrect the BIO `io_result`

icing:
In pytest'ing the situation occored that wolfSSL reported an IO error when the underlying BIO operation was returning an CURLE_AGAIN condition.

Readding the `io_result` filter context member to detect such situations.

Also, making sure that the returned CURLcode is initialized on all recv operations outcome.


bagder:
Thanks!

PR #10717, opened 2023-03-09
secure-transport, fixes recv return code handling

icing:
Return code handling of recv calls were not always correct when an error occured or the connection was closed.

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -3377,13 +3377,15 @@ static ssize_t sectransp_recv(struct Curl_cfilter *cf,
   DEBUGASSERT(backend);
 
   again:
+  *curlcode = CURLE_OK;
   err = SSLRead(backend->ssl_ctx, buf, buffersize, &processed);
 
   if(err != noErr) {
     switch(err) {
       case errSSLWouldBlock:  /* return how much we read (if anything) */
-        if(processed)
+        if(processed) {
           return (ssize_t)processed;
+        }
         *curlcode = CURLE_AGAIN;
         return -1L;
         break;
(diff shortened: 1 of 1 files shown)


lib/vtls/sectransp.c
```diff
@@ -3395,7 +3397,7 @@ static ssize_t sectransp_recv(struct Curl_cfilter *cf,
       case errSSLClosedGraceful:
       case errSSLClosedNoNotify:
         *curlcode = CURLE_OK;
-        return -1L;
+        return 0;
```
jay, line 3400: This recv function has returned -1 there for a long time, but neither way looks right to me. `errSSLClosedGraceful` seems to me like not an error, and `errSSLClosedNoNotify` seems to me like an error. I see fadf33c7 last made changes to that area a decade ago, but I have no idea if it's still relevant. @nickzman can you review

nickzman: Hi, I'm still here. Sorry I haven't had time for open source lately.

The difference between `errSSLClosedGraceful` and `errSSLClosedNoNotify` in Secure Transport is the former occurs when the server properly closes the TLS session prior to closing the TCP connection, and the latter occurs when the server closes the TCP connection without gracefully shutting down TLS first. Secure Transport makes that distinction in case there are any clients out there that are concerned about this, but we (and by we, I mean libcurl) really don't care.

jay: I'm still unclear on why if the intention was to ignore close notify for Secure Transport that you wrote it to return -1 for errSSLClosedGraceful and errSSLClosedNoNotify. It seems then like Stefan's change to return 0 for those would make sense?

icing: @jay, can you give any specific reason why this change should not be merged?

I am asking, because I think it is necessary for the next release and the longer we wait, the less feedback we can get from others before it.

bagder: I think `0` seems to be what was intended to be returned there and seems to be the correct fix. If we would change this to an error (for the NoNotify case), we should do that in a separate PR as then it should also not return `CURLE_OK`.

bagder:
Thanks!

PR #10719, opened 2023-03-09
docs/SECURITY-PROCESS.md: updates

bagder:
- allow Low+Medium issues to be managed through plain PRs
- update the bug-bounty part to reflect current reality


MarcelRaad (approved):
Ah, perfect, the fixup is just what I would have asked!

PR #10736, opened 2023-03-10
vlts recv, use full buffer size if possible

icing:
SSL backends like OpenSSL/wolfSSL and other return the content of one TLS record on read, but usually there are more available.

Change the vtls cfilter recv() function to fill the given buffer until a read would block.


bagder (approved):

bagder:
Thanks!

PR #10739, opened 2023-03-10
wolfssl: add quic/ngtcp2 detection in cmake, and fix builds

vszakats:
- Implement QUIC/ngtcp2 detection in CMake with wolfSSL.

  Because wolfSSL uses zlib if available, move compression detection before TLS detection.
  (OpenSSL might also need this in the future.)

- wolfSSL 5.5.0 started using C99 types in its `quic.h` header, but it doesn't #include the necessary C99 header itself, breaking builds (unless another dependency pulled it by chance.) Add local workaround for it. For this to work with all build tools, we had to fix our header detection first. PR: #10745

  Ref: https://github.com/curl/curl-for-win/commit/6ad5f6ecc15620c15625fc443476b3a1ecef4f3f

Closes #10739


bagder:
very green :heavy_check_mark: !

PR #10740, opened 2023-03-10
http: don't send 100-continue for short PUT requests

dfandrich:
This is already how curl is documented to behave in Everything curl, but in
actuality only short POSTs skip this. This should knock 30 seconds off a
full run of the test suite since the 100-continue timeout will no longer be
hit.

### DIFF
--- a/docs/FAQ
+++ b/docs/FAQ
@@ -990,11 +990,10 @@ FAQ
 
   4.16 My HTTP POST or PUT requests are slow
 
-  libcurl makes all POST and PUT requests (except for POST requests with a
-  tiny request body) use the "Expect: 100-continue" header. This header
-  allows the server to deny the operation early so that libcurl can bail out
-  before having to send any data. This is useful in authentication
-  cases and others.
+  libcurl makes all POST and PUT requests (except for requests with a small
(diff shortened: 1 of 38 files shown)


bagder (approved):

dfandrich:
I'm not sure what's up with tests 154 and 155. In some configurations they're failing, in others they're succeeding. The tests are both --anyauth tests with differences in the initial, auth-probing PUT. The failing tests are not logging the data in that initial request. Could it be that the server is closing the connection immediately after the headers are sent, when it recognizes that it's going to have to return 403 (and throw away the body)? Maybe there's no chance to log the body in that case.

bagder:
If you get stuck on this, I can dig in a little and see if I can figure it out after the weekend.

dfandrich:
I'm able to reproduce this locally occasionally, so there's hope.

dfandrich:
Looks like I was right. Both tests had the `auth_required` flag which forced the server to close the connection immediately after receipt of the request. I'm still not exactly sure why it didn't fail consistently, but (without looking at the server code) I'm guessing it had to do with timeing and how data was split in getting to the server.

Issue #10741, opened 2023-03-10
Schannel regression: CURLOPT_SSL_CIPHER_LIST has no effect on Windows 10 >= 10.0.17763


zhihaoy:
### I did this

Set `CURLOPT_SSL_CIPHER_LIST` on a connection that uses Schannel, TLS 1.2, and make a connection to an HTTPS server.

### I expected the following

Investigate its Hello message, the cipher suites should match my settings.

### curl/libcurl version

libcurl 7.85.0 to the latest

### operating system

Fails (`CURLOPT_SSL_CIPHER_LIST` has no effect) on Windows 10 build 17763 and above, such as 10 Enterprise 1809 and 20H1.
Works (`CURLOPT_SSL_CIPHER_LIST` is in effect) on Windows 10 build 14393, such as Windows Server 2016

### Cause analysis

The problem was introduced by https://github.com/curl/curl/commit/8beff4355956e3d18ceb3afc21c1f3edec82543c. The change moved this chunk of code: https://github.com/curl/curl/blob/801bd5138ce31aa0d906fa4e2eabfc599d74e793/lib/vtls/schannel.c#L503-L510 to https://github.com/curl/curl/blob/cd95ee9f771361acf241629d2fe5507e308082a2/lib/vtls/schannel.c#L1006-L1021 However, the new block was an else block started way above https://github.com/curl/curl/blob/cd95ee9f771361acf241629d2fe5507e308082a2/lib/vtls/schannel.c#L790-L792 Then the logic becomes: Since only Windows 10 build 17763 supports TLS 1.3, we configure *only* TLS 1.3 ciphers on those systems; if we are on a lower version of Windows, we go ahead and configure TLS 1.2 (or lower) ciphers. So the combination of "on a newer version of Windows 10" and "using TLS 1.2" falls into neither branch. Support for `CURLOPT_SSL_CIPHER_LIST` under that combined condition was lost.

jay:
I've submitted a fix in PR #10746 which should allow TLS < 1.3 ciphers to be set as long as TLS 1.3 ciphers are not set. I'm not sure if it's possible to combine the two because they use different credential structs and we are only allowed to pass a single credential struct to schannel. The PR is marked as draft for now. If anyone can investigate further that would be helpful.

/cc @wyattoday

vszakats:
[...]

PR #10746, opened 2023-03-12
schannel: fix user-set older TLS ciphers in Windows 10 & 11

jay:
- If CURLOPT_TLS13_CIPHERS is not set then use CURLOPT_SSL_CIPHER_LIST ciphers instead.

Prior to this change libcurl would ignore older ciphers (TLS <= 1.2) in Windows 10 1809 and later. For example, if the user set anything via CURLOPT_SSL_CIPHER_LIST it was ignored in favor of CURLOPT_TLS13_CIPHERS even if the latter wasn't set.

Unresolved: CURLOPT_TLS13_CIPHERS and CURLOPT_SSL_CIPHER_LIST are still mutually exclusive for Schannel. I'm not sure how to solve this. Since this commit, if CURLOPT_TLS13_CIPHERS is set then SCH_CREDENTIALS is used to set banned ciphers based on what is missing from the user's TLS 1.3 cipher list; however if CURLOPT_TLS13_CIPHERS is not set then SCH_CREDENTIALS is used to set allowed TLS <= 1.2 ciphers, if any, set via CURLOPT_SSL_CIPHER_LIST. Can they be combined somehow into a single credentials struct?

Reported-by: zhihaoy@users.noreply.github.com

Fixes https://github.com/curl/curl/issues/10741
Closes #xxxx

---

SCH_CREDENTIALS is what we use if CURLOPT_TLS13_CIPHERS is set.
https://learn.microsoft.com/en-us/windows/win32/api/schannel/ns-schannel-sch_credentials

SCHANNEL_CRED is what we use if CURLOPT_SSL_CIPHER_LIST is set, or if neither is set.
https://learn.microsoft.com/en-us/windows/win32/api/schannel/ns-schannel-schannel_cred

### DIFF
--- a/docs/CIPHERS.md
+++ b/docs/CIPHERS.md
@@ -346,9 +346,12 @@ maps them to the following case-insensitive names.
 ## Schannel
 
 Schannel allows the enabling and disabling of encryption algorithms, but not
-specific cipher suites. They are
+specific cipher suites, prior to TLS 1.3. The algorithms are
 [defined](https://docs.microsoft.com/windows/desktop/SecCrypto/alg-id) by
 Microsoft.
+
+The algorithms below are for TLS 1.2 and earlier. TLS 1.3 is covered in the
+next section.
 
 There is also the case that the selected algorithm is not supported by the
 protocol or does not match the ciphers offered by the server during the SSL
@@ -412,13 +415,23 @@ are running an outdated OS you might still be supporting weak ciphers.
 
 ### TLS 1.3 cipher suites
 
-(Note these ciphers are set with `CURLOPT_TLS13_CIPHERS` and `--tls13-ciphers`)
+You can set TLS 1.3 ciphers for Schannel by using `CURLOPT_TLS13_CIPHERS` or
+`--tls13-ciphers` with the names below.
+
+If TLS 1.3 cipher suites are set then libcurl will add or restrict Schannel TLS
+1.3 algorithms automatically. Essentially, libcurl is emulating support for
+individual TLS 1.3 cipher suites since Schannel does not support it directly.
 
 `TLS_AES_256_GCM_SHA384`
 `TLS_AES_128_GCM_SHA256`
 `TLS_CHACHA20_POLY1305_SHA256`
 `TLS_AES_128_CCM_8_SHA256`
 `TLS_AES_128_CCM_SHA256`
+
+Note if you set TLS 1.3 ciphers without also setting the minimum TLS version to
+1.3 then it's possible Schannel may negotiate an earlier TLS version and cipher
+suite if your libcurl and OS settings allow it. You can set the minimum TLS
+version by using `CURLOPT_SSLVERSION` or `--tlsv1.3`.
 
 ## BearSSL
 
--- a/docs/cmdline-opts/tls13-ciphers.d
+++ b/docs/cmdline-opts/tls13-ciphers.d
@@ -17,5 +17,5 @@ cipher suite details on this URL:
 https://curl.se/docs/ssl-ciphers.html
 
 This option is currently used only when curl is built to use OpenSSL 1.1.1 or
-later or Schannel. If you are using a different SSL backend you can try
+later, or Schannel. If you are using a different SSL backend you can try
 setting TLS 1.3 cipher suites by using the --ciphers option.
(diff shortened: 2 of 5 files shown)


wyattoday:
Yeah, there's not an easy way to solve this problem. Well, there's not an easy and *secure* way to solve this problem. And I don't see it as a problem.

The easy way to solve this problem is if the user specifies any of the billion old cipher list, then "downgrade" them to using a max of TLS 1.2 (regardless of what they specify).

But then they're using an old, less secure, version of TLS. Yes, TLS 1.2 isn't deprecated *yet*, but give it a couple of years.

My view on this problem is this: if you want to use old insecure ciphers, then stick to old insecure versions of Windows. If you want to use secure versions of Windows you'll also use secure ciphers.

That's the safest option (leaving it as-is).

Insecure options include:

1. Disabling the TLS 1.3 code-path if they specify any old-style cipher from that list of a billion ciphers. (as I said above). The downside is they cannot use TLS 1.3 or newer. Which is a security issue.
2. A second option (and just as bad security-wise) is to do 1, but to show a warning that TLS 1.3 is available but won't be used because they specified old-style ciphers. This is bad because libCurl users won't see the warning, and curl.exe users will likely just ignore the warning.


My opinion: **this isn't a bug.** If they want to use insecure ciphers they should use insecure versions of Windows too. Yes, this might annoy this 1 user (and maybe 1 or 2 others over the next few years), but it's a more secure option that's better for the health of the internet.

jay:
Thanks. It's a bug because the user can't set any legacy ciphers in newer versions of Windows. I'm likely to go with 1 or 2.

wyattoday:
@zhihaoy 

> When the protocol version isn't specified, I tend to say that the semantics of CURLOPT_SSL_CIPHER_LIST should be "these ciphers are okay, but I'm fine if they are not used," like when you're negotiating with servers. I.e., if we have an opportunity of establishing a connection using a more secure protocol, it's fine to kick some ciphers out.

Unfortunately, `CURLOPT_SSL_CIPHER_LIST` doesn't currently work like that (even before I added TLS 1.3 support to curl w/ schannel). Rather, it just takes your input and doesn't consider ejecting insecure options. Not providing a cipher list at all is the most secure option.

zhihaoy:
> My suggested fix is to leave the logic as-is but document that specifying the old-style cipher list will only work if they're on old Windows. Otherwise, they need to specify the new style cipher list.

Sounds like a user-hostile thing to do -- whether a potentially insecure option is honored now depends on your running environment. If outside Schannel, the behavior is the same (i.e., `CURLOPT_SSL_CIPHER_LIST` is ignored when using OpenSSL backend on newer versions of Windows), that would at least be consistent.

zhihaoy:
I talked with my team recently about this issue. We still want a code change.

The point was that, by retaining this regression, we lose the potential to have control over the ciphers to enforce stricter rules and stronger ciphers on most of our users. We were previously unsure whether a change was a good thing because the feature we had, which was broken by this regression, happens to be unimportant for other reasons. libcurl is foolish-proof by default, of course. Still, the knobs are there for a broader range of audiences, and we hope to see those knobs have more predictable behaviors so that library users can develop their applications for the good.

jay:
Please review the most recent commit:

**schannel: fix user-set legacy algorithms in Windows 10 & 11**

- If the user set a legacy algorithm list (CURLOPT_SSL_CIPHER_LIST) then use the SCHANNEL_CRED legacy structure to pass the list to Schannel.

- If the user set both a legacy algorithm list and a TLS 1.3 cipher list then abort.

Although MS doesn't document it, Schannel will not negotiate TLS 1.3 when SCHANNEL_CRED is used. That means setting a legacy algorithm list limits the user to earlier versions of TLS.

Prior to this change, since 8beff435 (precedes 7.85.0), libcurl would ignore legacy algorithms in Windows 10 1809 and later.

Reported-by: zhihaoy@users.noreply.github.com

Fixes https://github.com/curl/curl/issues/10741

jay:
Thanks to both of you for your feedback. I've landed the proposed change (https://github.com/curl/curl/pull/10746#issuecomment-1577936149) from last month which allows for the user to set legacy algorithms in Windows 10 and 11.

> I think the changes are good. However, I think a couple of errors should be added

I added a warning, for verbose mode only, when an algorithm cipher list is used and TLS 1.3 is an enabled protocol:

`* schannel: WARNING: This version of Schannel may negotiate a less-secure TLS version than TLS 1.3 because the user set an algorithm cipher list.`

Though currently Schannel does not negotiate TLS 1.3 when SCHANNEL_CRED is used, that is undocumented MS behavior and may change.

Issue #10743, opened 2023-03-11
Wasted CPU and slowed down transfer without CURLOPT_NOSIGNAL set to 1


jerstlouis:
### I did this
With a multi curl handle, a `curl_multi_perform()` loop wastes enormous amount of CPU and transfers way slower than it should unless `CURLOPT_NOSIGNAL` is set, because of `sigaction()` being called every single time continuously.

### I expected the following
I did not expect CPU to be wasted and transfer slowed down because of an obscure `CURLOPT_NOSIGNAL` option I had no idea about.

Setting `CURLOPT_NOSIGNAL` to 1 resulted in huge performance gain and no longer wasting CPU, but finding out about this was very costly. The cost associated with the default behavior is unexpected. Figuring this out required realizing that the cURL processing thread is busy most of the time calling _sigaction()_, and working backwards to _sigpipe_ignore()_, `no_signal`, and finally `CURLOPT_NOSIGNAL`. Not something that should be expected of developers using the library to get decent CPU usage / network performance.

### curl/libcurl version

curl 7.81.0 (x86_64-pc-linux-gnu)

### operating system

Linux hostname 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

bagder:
I'm sorry, what do you expect us to do about this issue?

jerstlouis:
@bagder Thank you. Well, how feasible is making `CURLOPT_NOSIGNAL` = 1 the default?

The problem is how expensive a call to `sigaction()` is from within `curl_multi_perform()`, which will typically be called at a high frequency. Whatever behavior is desired for `SIGPIPE` should be set *outside* of the loop.

At the very least, a mention of this performance issue in the [documentation page for curl_multi_perform()](https://curl.se/libcurl/c/curl_multi_perform.html) would help inform users about this problem.

bagder:
Can you qualify what "enormous amount of CPU" means in more specific terms? I seem to be able to download data very fast with libcurl in spite of this.

> how feasible is making CURLOPT_NOSIGNAL = 1 the default?

Not feasible at all, that would be a major change in a behavior and we would need to bump the SONAME number of that.

> mention of this performance issue

I'm not sure I agree with you yet that is is an issue...

jerstlouis:
@bagder Thank you for looking into this.

[...]

PR #10750, opened 2023-03-12
multi: make multi_perform ignore/unignore signals less often

bagder:
Ref: #10743


bagder:
too simple approach, need to rethink

jay:
> too simple approach, need to rethink

looks fine, what am i missing?

bagder:
Each easy handle has its own `CURLOPT_NOSIGNAL` set individually. We cannot switch off/on like this for all easy handles.

Issue #10751, opened 2023-03-13
HTTP/2 flow control (stream pause) can override curl_easy_pause


jay:
### I did this

Recent changes to HTTP/2 code from #10627 fixed an upload busy loop by creating sort of a flow control to pause the stream whenever the remote window size is 0 and unpause the stream whenever NGHTTP2_WINDOW_UPDATE is received. I approved and committed the fix, however [later it occurred to me](https://github.com/curl/curl/pull/10627#issuecomment-1451252647) that may be the wrong way to fix the issue because it uses the same mechanism to pause/unpause that the user can use via curl_easy_pause. The problem is that the transfer could be paused by the user and then unpaused internally by the HTTP/2 code, or the other way around. Also the curl_easy_pause function has additional pause/unpause code in it, so it's not exactly the same as the pausing in HTTP/2 code.

I think we should consider reverting the fix. By reverting that reintroduces the [HTTP/2 upload busy loop bug](https://github.com/curl/curl/issues/10618), which started in 71b7e016. I don't know what is causing that bug but the fix did work. However if the issue was flow control related then I don't know why it wasn't an issue before 71b7e016.

/cc @icing

bagder:
I don't think we should revert, but I think we should extend the pausing so that we use two bits for it. One that the user can set and one that we set internally, and we maintain the pause if either of the bits says pause.

bagder:
My reasoning for not reverting is that it is going to be unlikely that we pause an upload internally at exact the same time as the user pauses or unpauses it.

icing:
There is another change related to this: the handling of `data->state.drain` during uploads. In transfer.c:1088 we added

```c
#if defined(USE_HTTP2) || defined(USE_HTTP3)
  if(data->state.drain) {
    select_res |= CURL_CSELECT_IN;
    DEBUGF(infof(data, "Curl_readwrite: forcibly told to drain data"));
+    if((k->keepon & KEEP_SENDBITS) == KEEP_SEND)
+      select_res |= CURL_CSELECT_OUT;
  }
#endif
```

This code place overrides the socket event handling when `drain` is set. The addition was necessary since without it, uploads would stall when `drain` was set. This combination happens more likely on parallel transfers and especially when the server starts to reply before the upload is complete.

Admittedly, these hacks around the original, pure socket event based transfer handling for multi-connections show their age. Reverting any of those will break things. But, as @jay correctly states, the current fixes may clash with user pausing of uploads.

[...]

PR #10753, opened 2023-03-13
Use KEEP_SEND_HOLD for flow control in HTTP/2

icing:
- as mentioned in #10751, the use of KEEP_SEND_PAUSE interfers with pausing streams from the user side
- use the defined, but so far not used, KEEP_SEND_HOLD bit for flow control based suspend of sending in transfers


bagder (approved):
:+1: since this is tested and the fix is small and easy to grasp, I think we can merge this before the release once the CI says ok.

jay:
Thanks, I tested in Windows and Linux and see no discernable difference in actual time or CPU times (time -p).

PR #10762, opened 2023-03-14
multi: unlink handles in pending or msgsent states

bagder:


bagder:
I'll also make a test to move completed handles over to a new list so that they can be skipped in the main loop as well.

bagder:
The pytest failures are very curious. I need to dig into those...

bagder:
@icing I could use a little hand-holding to learn the best way to debug these pytest failures.

bagder:
> I think the problem is that you modify mult->easyp while iterating over it.

Ah yes indeed.

bagder:
@icing thanks for that hint, it seems that was the bug that held this back.

icing:
> @icing thanks for that hint, it seems that was the bug that held this back.

Yep, works now locally here as well.

PR #10769, opened 2023-03-15
Document that --next affects --write-out

jidanni:


docs/cmdline-opts/next.d
```diff
@@ -20,7 +20,7 @@ for each.
 
 --next will reset all local options and only global ones will have their
 values survive over to the operation following the --next instruction. Global
-options include --verbose, --trace, --trace-ascii and --fail-early.
+options include --write-out, --verbose, --trace, --trace-ascii and --fail-early.
```
bagder, line 23: This is not correct. To see, try:

    curl -w 'moo %{http_code}\n' localhost -so /dev/null --next localhost -so /dev/null -w 'moo %{http_code} again\n'

or just

    curl -w 'moo %{http_code}\n' localhost -so /dev/null --next localhost

bagder (changes requested):

docs/cmdline-opts/next.d
jidanni: OK, you are right and I made a 180 degree big mistake.

However in the process I discovered an actual documentation bug. The man page says:

>-: --next will reset all local options and only global ones will have their values survive over to the operation following the -:, --next instruction. Global options include -v, --verbose, --trace, --trace-ascii and --fail-early.

So --write-out is a local option, and it should get reset after the --next. But with

```
$ curl -w 'moo %{http_code}\n' \
	--next localhost -so /dev/null \
	--next localhost -so /dev/null \
	--next localhost -so /dev/null \
	--next localhost -so /dev/null
```
it still gets printed, yes only once, but still not zero times as the document makes one think!

bagder: This bug is now fixed via #10782

jidanni: There is a very real problem. If we are making a config file with a thousand entries, we previously could put a thousand of each item,
1,000 urls, etc., 
1000 of the word next.

Now we must be very careful and must only put 999 on the word next.

So you are fixing the problem I mentioned in the wrong way!

jidanni: It seems I posted my reply in
https://github.com/curl/curl/pull/10782#issuecomment-1474440036

PR #10770, opened 2023-03-15
Formally, not formerly

jidanni:


docs/cmdline-opts/write-out.d
```diff
@@ -134,7 +134,7 @@ IPv4 or IPv6. (Added in 7.29.0)
 The remote port number of the most recently done connection. (Added in 7.29.0)
 .TP
 .B response_code
-The numerical response code that was found in the last transfer (formerly
+The numerical response code that was found in the last transfer (formally
```
bagder, line 137: This means that the name was previously a different one. "Formerly" is how you write that. Formally is something completely different.

bagder (changes requested):

docs/cmdline-opts/write-out.d
jidanni: Well fine, you meant formerly.
But you mention the old one
$ man curl|col -b|grep -c http_code
3
$ man curl|col -b|grep -c response_code
1
more times than the new one.
So if the old one was old,
then you would have made a note about it
being old at each place you mention it.

Therefore we conclude that you probably mean that response_code is
merely an more formal alias for http_code.

And http_code is not being phased out, otherwise you would have warned
everywhere on the man page you mention http_code that it is an older way
of saying response_code, and it would be better to say response_code.

bagder: I maintain that what I is written is what is intended.

jidanni: Okay, then please put a note next to each of the http_code items, not just one, saying that they are deprecated or something.

PR #10771, opened 2023-03-15
HTTP/2 flow control and buffer improvements

icing:
HTTP/2 flow control and buffer improvements

    - use bufq for send/receive of network data
    - usd bufq for send/receive of stream data
    - use HTTP/2 flow control with no-auto updates to control the
      amount of data we are buffering for a stream
      HTTP/2 stream window set to 128K after local tests, defined
      code constant for now
    - elminiating PAUSEing nghttp2 processing when receiving data
      since a stream can now take in all DATA nghttp2 forwards

Based on  #10720

PR #10780, opened 2023-03-16
H2 proxy tunnel

icing:
HTTP/2 proxy tunnel implementation

based on #10720
based on #10771

- currently only on debug build and when env variable
   CURL_PROXY_TUNNEL_H2 is present.
- will ALPN negotiate with the proxy server and switch
  tunnel filter based on the protocol negotiated.
- http/1.1 tunnel code moved into cf-h1-proxy.[ch]
- http/2 tunnel code implemented in cf-h2-proxy.[ch]
- tunnel start and ALPN set remains in http_proxy.c
- moving all haproxy related code into cf-haproxy.[ch]

VTLS changes
- SSL filters rely solely on the "alpn" specification they
  are created with and no longer check conn->bits.tls_enable_alpn.
- checks on which ALPN specification to use (or none at all) are
  done in vtls.c when creating the filter.

Testing
- added a nghttpx forward proxy to the pytest setup that
  speaks HTTP/2 and forwards all requests to the Apache httpd
  forward proxy server.
- extending test coverage in test_10 cases
- adding proxy tests for direct/tunnel h1/h2 use of basic auth.
- adding test for http/1.1 and h2 proxy tunneling to pytest

### DIFF
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -108,6 +108,9 @@ LIB_CFILES =         \
   bufq.c             \
   bufref.c           \
   c-hyper.c          \
+  cf-h1-proxy.c      \
+  cf-h2-proxy.c      \
+  cf-haproxy.c       \
   cf-https-connect.c \
   cf-socket.c        \
   cfilters.c         \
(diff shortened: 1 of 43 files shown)


bagder:
Thanks!

PR #10783, opened 2023-03-17
Adding timeout to running curl in test cases

icing:
- we had a CI case once where `curl` seemingly did not return and it was hard to guess what happened.
- make curl execution in test cases time out after 60 seconds


bagder:
Thanks!

PR #10787, opened 2023-03-17
socket recv buffering

icing:
Adding socket recv buffering for most tcp cases

- use bufq as recv buffer, also for Windows pre-receive handling
- tweak bufq to properly handle some edge cases found


bagder:
Thanks!

PR #10798, opened 2023-03-20
GHA: run all linux test jobs with valgrind

bagder:


bagder:
New finding. These six tests leak memory in hyper builds: `1069 1072 1073 1500 1539 1554`

bagder:
The clang builds and the NSS build all show "unhandled dwarf2 abbrev form code" errors so I need to disable valgrind for those builds.

MarcelRaad:
According to e.g. https://bugs.kde.org/show_bug.cgi?id=457412, another solution would be to force DWARF v4 by passing `-gdwarf-4` to the compiler.

thesamesam:
> According to e.g. [bugs.kde.org/show_bug.cgi?id=457412](https://bugs.kde.org/show_bug.cgi?id=457412), another solution would be to force DWARF v4 by passing `-gdwarf-4` to the compiler.

or do Valgrind builds on a distro with >= 3.20.0. I haven't looked yet at what the distribution of compiler versions and such is but it might be useful aside from Valgrind.

bagder:
I think it is fine to run it on only the gcc builds for now. What I cannot explain though is how the hyper errors vanished in a rerun...

PR #10801, opened 2023-03-20
multi: remove PENDING + MSGSENT handles from the main linked list

bagder:
As they are not driving transfers or any socket activity, the main loop does not need to iterate over these handles. A performance improvement.

They are instead only held in their own separate lists.

'data->multi' is kept a pointer to the multi handle as long as the easy handle is actually part of it even when the handle is moved to the pending/msgsent lists. ->multi gets cleared by curl_multi_cleanup() which "orphans" all previously attached easy handles.

This is take 2. The first version was reverted for the 8.0.1 release.

Assisted-by: Stefan Eissing

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -3977,8 +3977,10 @@ static CURLcode wc_statemach(struct Curl_easy *data)
     case CURLWC_DONE:
     case CURLWC_ERROR:
     case CURLWC_CLEAR:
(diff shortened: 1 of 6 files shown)


jeroen:
I think a side effect of this is that `curl_multi_fdset()` no longer shows these connections, see https://github.com/jeroen/curl/issues/299

PR #10810, opened 2023-03-21
GHA-linux: add a sanitizer build

bagder:


bagder:
Made it work, I had added *both* address and memory sanitizers and apparently it can't do that. It can only do one of them at a time, so I will probably do two separate CI jobs.

PR #10812, opened 2023-03-22
Add asserts in DEBUG builds for GOOD_EASY_HANDLE and GOOD_MULTI_HANDL…

icing:
…E checks

- allow NULL pointers to "just" return an error as before
- fail hard on nun-NULL pointers that no longer show the MAGICs


bagder (approved):

bagder:
Thanks!

PR #10819, opened 2023-03-22
CMake config: version 8 is compatible with version 7

dg0yt:
AFAIU curl version 8.0.0 does bring any breaking API changes. However, the change of the major version has the effect that the CMake config is not considered compatible for `find_package(CURL 7.123 CONFIG)`, breaking downstream usage, e.g.
azure-core-cpp
~~~
CMake Error at D:/installed/x64-windows-static/share/curl/vcpkg-cmake-wrapper.cmake:11 (_find_package):
  Could not find a configuration file for package "CURL" that is compatible
  with requested version "7.44".

  The following configuration files were considered but not accepted:

    D:/installed/x64-windows-static/share/curl/CURLConfig.cmake, version: 8.0.0-DEV
~~~

kubernetes:
~~~
CMake Error at D:/installed/x64-windows-static/share/curl/vcpkg-cmake-wrapper.cmake:11 (_find_package):
  Could not find a configuration file for package "CURL" that is compatible
  with requested version "7.58.0".

  The following configuration files were considered but not accepted:

    D:/installed/x64-windows-static/share/curl/CURLConfig.cmake, version: 8.0.0-DEV
~~~

This PR changes the default version check to accept version 8 as a compatible major version if version 7 is requested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1609,6 +1609,15 @@ write_basic_package_version_file(
     VERSION ${CURL_VERSION}
     COMPATIBILITY SameMajorVersion
 )
+file(READ "${version_config}" generated_version_config)
+file(WRITE "${version_config}"
+"if(NOT PACKAGE_FIND_VERSION_RANGE AND PACKAGE_FIND_VERSION_MAJOR STREQUAL \"7\")
+    # Version 8 satisfies version 7... requirements
+    set(PACKAGE_FIND_VERSION_MAJOR 8)
+    set(PACKAGE_FIND_VERSION_COUNT 1)
+endif()
+${generated_version_config}"
+)
 
 # Use:
 # * TARGETS_EXPORT_NAME


bagder:
Someone needs to help me with this. I don't understand the description nor the patch.

dg0yt:
Doc is there: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-version-selection

For CMake config, it is up to the package to define what is a compatible version, through script code in the version cmake file. Based on the requested version from the `find_package` call, CMake sets some input variables for the version file, and it is up to the version file to decide.

CURL uses a CMake function to generate a version file with the "same major version" policy, i.e. any newer version is accepted if it is the same major version. e.g. 8.0.1 would satisfy the request for version 8.0.0 because 8.0.1 is newer and the major version is 8. But it wouldn't satisfy the request for 7.44 because it is a different major version.

This PR adds code before the actual version check which turns a request for `7.44` into a request for `8` by inspecting and modifying the version interface variables before passing control to the generated version check.

dg0yt:
Alternatively one could generate the version checking code with the `AnyNewerVersion` policy, but this would change behaviour for `find_package(CURL 6.12.3 CONFIG)`.

jzakrzewski:
Yeah, `AnyNewerVersion` is not the way to go as it may have unintended consequences.
I think this change make sense as it is.

Issue #10793, opened 2023-03-19
curl build error when ngtcp2 is installed but nghttp3 is not


chiouss:
Hi,

I installed ngtcp2 on Openwrt in order to build some small program with QUIC I wrote. But then I found curl will build failed after it finds ngtcp2 in system is usable but doesn't notice there's no nghttp3 in system.

I know I can add `--without-ngtcp2` in configure, but I add `--without-nghttp3` at the first time and found nothing changed. Then I realized that there's no checking of `USE_NGHTTP3` in code.

I think in configure.ac we need to check nghttp3 first then check ngtcp2 maybe?

```
In file included from quic.h:29,
                 from urldata.h:142,
                 from altsvc.c:30:
vquic/ngtcp2.h:30:10: fatal error: nghttp3/nghttp3.h: No such file or directory
 #include <nghttp3/nghttp3.h>
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
Makefile:1670: recipe for target 'libcurl_la-altsvc.lo' failed
```

### I did this
[...]

PR #10821, opened 2023-03-23
Adjust config and code defined checks for ngtcp2 without nghttp3

icing:
- make configure show on HTTP3 feature that both ngtcp2 and nghttp3 are in play
- define ENABLE_QUIC only when USE_NGTCP2 and USE_NGHTTP3 are defined
- add USE_NGHTTP3 in the ngtcp2 implementation
- refs #10793


bagder (approved):

bagder:
Thanks!

PR #10829, opened 2023-03-24
pytest improvements for suitable curl and error output

icing:
- will check built curl for http and https support and skip all tests if not there
- will dump stdout/stderr/trace output on errored responses


bagder:
merge conflict!

icing:
> merge conflict!

updated.

bagder:
Thanks!

PR #10835, opened 2023-03-25
doc: curl_mime_init() strong easy handle binding has been relaxed in 7.87.0

### LINKED ISSUE #10834: curl_mime_init man page: handle must be final target handle
> easy_handle is used for part separator randomization and error reporting. It does not need to be the final target handle.

This is incorrect, easy_handle must be the exact handle that the curl_mime will be given to, at least with HTTP.

To reproduce:
```c
#include <assert.h>
#include <curl/curl.h>

int main() {
	CURL *handle = curl_easy_init();
	CURL *cinit  = curl_easy_init();
	curl_mime *mime = curl_mime_init(cinit);
	curl_easy_setopt(handle, CURLOPT_URL, "https://pie.dev/post");
	curl_easy_setopt(handle, CURLOPT_POST, 1);
	assert(curl_easy_setopt(handle, CURLOPT_MIMEPOST, mime) == CURLE_OK); // this throws
	curl_easy_perform(handle);
	return 0;
}
```

If the example is changed to `curl_mime_init(handle)`, everything works as expected.

monnerat:
Fixes #10834

### DIFF
--- a/docs/libcurl/curl_mime_init.3
+++ b/docs/libcurl/curl_mime_init.3
@@ -31,14 +31,14 @@ curl_mime_init - create a mime handle
 curl_mime *curl_mime_init(CURL *easy_handle);
 .fi
 .SH DESCRIPTION
-\fIcurl_mime_init(3)\fP creates a handle to a new empty mime structure,
+\fIcurl_mime_init(3)\fP creates a handle to a new empty mime structure.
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #10838, opened 2023-03-25
ci/openssl-3.1.0+quic updates

pheiduck:
Breaking changes:
```
checking for SSL_is_quic... no
configure: openssl does not have QUIC interface, disabling it
configure: error: openssl was requested (--with-openssl) but not found
```
@tatsuhiro-t ngtcp2 needs to be updated to support `openssl 3.1.0`


tatsuhiro-t:
Wait for https://github.com/quictls/openssl/pull/111 to be merged.

PR #10849, opened 2023-03-28
docs/cmdline-opts: document the dotless config path

rpigott:
The real xdg config path is $XDG_CONFIG_HOME/curlrc, without the dot. The dotless name seems preferable, so let's match the documentation to the behavior.

---

curl 8.0.1 doesn't seem to care about `$XDG_CONFIG_HOME/.curlc` but I was happy to find it does find `$XDG_CONFIG_HOME/curlrc`. Let's document that.

### DIFF
--- a/docs/cmdline-opts/config.d
+++ b/docs/cmdline-opts/config.d
@@ -57,7 +57,7 @@ config file is checked for in the following places in this order:
 
 1) "$CURL_HOME/.curlrc"
 
-2) "$XDG_CONFIG_HOME/.curlrc" (Added in 7.73.0)
+2) "$XDG_CONFIG_HOME/curlrc" (Added in 7.73.0)
 
 3) "$HOME/.curlrc"
 


bagder:
Thanks!

keithbowes:
> Yeps, I believe that was just a typo. I don't think it ever used the dot-version in there.

Actually, in my distro's supplied version, 7.88.1, $XDG_CONFIG_HOME/.curlrc works but $XDG_CONFIG_HOME/curlrc doesn't. It seems that it originally used the dot-version and then was fixed later (see https://github.com/curl/curl/issues/12129). Therefore, the man page is technically incorrect; in fact $XDG_CONFIG_HOME/curlrc hasn't been used since 7.73.0, but:
* $XDG_CONFIG_HOME/.curlrc (7.73.0–7.81.0)
* $XDG_CONFIG_HOME/.curclrc (if $XDG_CONFIG_HOME is set) or $XDG_CONFIG_HOME/curlrc (if $XDG_CONFIG_HOME is not set) (7.82.0–7.88.1)
* $XDG_CONFIG_HOME/curlrc (8.0+)

rpigott:
I'm confused.

So curl changed the config path to include a dot contrary to the documentation? Shouldn't that just be reverted?

foomin10:
Probably keithbowes is right, and the document is wrong. In curl 7.73.0, $XDG_CONFIG_HOME/curlrc is not supported.
@rpigott @bch

PR #10852, opened 2023-03-28
telnet: simplify the implementation of str_is_nonascii()

kdudka:
There is no need to traverse the string twice.

### DIFF
--- a/lib/telnet.c
+++ b/lib/telnet.c
@@ -772,12 +772,11 @@ static void printsub(struct Curl_easy *data,
 
 static bool str_is_nonascii(const char *str)
 {
-  size_t len = strlen(str);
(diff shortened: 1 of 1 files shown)


bagder (approved):

pythontech:
Sorry to necro-post (and I don't have MSVC to hand to check), but couldn't this pragma ugliness be avoided by doing
```
  while((c = *str++) != '\0')
```
as is done in inet_pton.c ?

kdudka:
It could but it is a matter of taste.  I do not find it ugly myself at all.  The syntax I used is not new in curl code either.  You can find it in `lib/ftp.c`:
```c
while((slashPos = strchr(curPos, '/')))
```

PR #10853, opened 2023-03-28
tool_writeout: add URL component variables

bagder:
Output specific components from the used URL. The following `-w` variables are added for this purpose:

url.scheme, url.user, url.password, url.options, url.host, url.port, url.path, url.query, url.fragment, url.zoneid

and for getting the components out of the "effective URL":

urle.scheme, urle.user, urle.password, urle.options, urle.host, urle.port, urle.path, urle.query, urle.fragment, urle.zoneid

Added test 423 and 424 to verify.


src/tool_writeout.c
```diff
@@ -262,6 +364,33 @@ static int writeString(FILE *stream, const struct writeoutvar *wovar,
         valid = true;
       }
       break;
+    case VAR_INPUT_URLHOST:
+    case VAR_INPUT_URLPATH:
+    case VAR_INPUT_URLSCHEME:
```
rybak, line 369: Minor drive-by nitpick: would you mind sorting the branches of the switches in `writeString` here and in `urlpart` above, so that the order is the same as in array `static const struct writeoutvar variables[]` and in the enum in `tool_writeout.h`?

bagder: Sensible, yes.

docs/cmdline-opts/write-out.d
```diff
@@ -209,6 +209,70 @@ The total time, in seconds, that the full operation lasted.
...
+.TP
+.B urle.query
+The query part of the effective (last) URL that was fetched. (Added in 8.1.0)
+.TP
+.B urle.fragment
+The fragment part of the effective (last) URL that was fetched. (Added in 8.1.0)
+.TP
+.B urle.zoneid
+The zoneid part of the effective (last) URL that was fetched. (Added in 8.1.0)
+.TP
```
jay, line 275: This could all be combined in `.B url` and `.B url_effective` (imo makes more sense than urle) to say something like 

~~~
.B url
The URL that was fetched. (Added in 7.75.0)

Individual parts of the URL can written using format `url.<part>`. The
part names are: are fragment, host, options, password, path, port,
query, scheme, user, and zoneid. (Added in 8.1.0)
~~~

PR #10856, opened 2023-03-28
ws: handle reads before EAGAIN better

### LINKED ISSUE #10831: WebSocket Intermittent Infinite Loop
### I did this
I wrote code that created a WebSocket connection and used it do download data for multiple days.

### I expected the following
I expected the WebSocket connection to be stable for extended periods of time.  (Yes, I do know this is experimental.)

### What actually happened
The connection remained stable for a while, but then started ignoring data and using 100% cpu.  When this happened, curl_ws_recv was returning CURLE_OK, the recv parameter was set to 0, and the curl_ws_frame contained bytesleft of 0.  My code repeatedly processed these phantom frames and called curl_ws_recv again for the next one.

The verbose output leading up to the problem was:
* WS: received 160 bytes payload (0 left, buflen was 332)
* WS: 168 bytes left to decode
* WS:235 received FIN bit 1
* WS: received OPCODE TEXT
* WS: received 163 bytes payload (0 left, buflen was 168)
* WS: 1 bytes left to decode
* WS: plen == 1, EAGAIN

Following this, the last two lines were repeated indefinitely.

I dug into this behavior and believe I have found the source of the problem in ws.c. When curl_ws_recv (https://github.com/curl/curl/blob/master/lib/ws.c#L409) is called at the end of a frame (wsp->frame.bytesleft == 0) with one byte in the buffer (wsp->stillblen == 1):
- curl_easy_recv on line 431 is not called due to the condition on line 428.
- ws_decode on line 449 is called and returns CURLE_AGAIN since the buffer does not include enough bytes for a frame.
- The break on line 453 results in return value of CURLE_OK with a buffer of length zero.
- This repeats for each curl_ws_recv call.

I fixed this locally by updating:
https://github.com/curl/curl/blob/master/lib/ws.c#L453
`<        break;`
`>        continue;`
https://github.com/curl/curl/blob/master/lib/ws.c#L428
`<    if(!wsp->stillblen) {`
`>    if(!wsp->stillblen || result == CURLE_AGAIN) {`

So far, this has seemed to work, but I'm sure you can come up with something better.

### curl/libcurl version
curl 7.88.1 (x86_64-pc-linux-gnu) libcurl/7.88.1 OpenSSL/3.0.2 zlib/1.2.11
Release-Date: 2023-02-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets

### operating system
Linux virt-linux 5.19.0-35-generic #36~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb 17 15:17:25 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

bagder:
Reported-by: simplerobot on github
Fixes #10831

### DIFF
--- a/lib/ws.c
+++ b/lib/ws.c
@@ -425,10 +425,10 @@ CURL_EXTERN CURLcode curl_ws_recv(struct Curl_easy *data, void *buffer,
     size_t datalen;
     unsigned int recvflags;
 
-    if(!wsp->stillblen) {
+    if(!wsp->stillblen || (result == CURLE_AGAIN)) {
       /* try to get more data */
       size_t n;
-      result = curl_easy_recv(data, data->state.buffer,
+      result = curl_easy_recv(data, &data->state.buffer[wsp->stillblen],
                               data->set.buffer_size, &n);
       if(result)
         return result;
@@ -438,10 +438,10 @@ CURL_EXTERN CURLcode curl_ws_recv(struct Curl_easy *data, void *buffer,
         return CURLE_GOT_NOTHING;
       }
       wsp->stillb = data->state.buffer;
-      wsp->stillblen = n;
+      wsp->stillblen += n;
     }
 
-    infof(data, "WS: %u bytes left to decode", (int)wsp->stillblen);
+    infof(data, "WS: %zu bytes left to decode", wsp->stillblen);
     if(!wsp->frame.bytesleft) {
       size_t headlen;
       curl_off_t oleft;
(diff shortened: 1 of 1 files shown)


lib/ws.c
```diff
@@ -425,10 +425,10 @@ CURL_EXTERN CURLcode curl_ws_recv(struct Curl_easy *data, void *buffer,
     size_t datalen;
     unsigned int recvflags;
 
-    if(!wsp->stillblen) {
+    if(!wsp->stillblen || (result == CURLE_AGAIN)) {
       /* try to get more data */
       size_t n;
-      result = curl_easy_recv(data, data->state.buffer,
+      result = curl_easy_recv(data, &data->state.buffer[wsp->stillblen],
                               data->set.buffer_size, &n);
```
simplerobot, line 432: Nice catch.  This explains some weirdness I was just starting to look into.  Thanks for the quick reply.  I think the buffer size should be updated too:
      result = curl_easy_recv(data, &data->state.buffer[wsp->stillblen],
                              data->set.buffer_size - wsp->stillblen, &n);

bagder: Right, that was an oversight of me!

bagder:
> I'd be happy to open a new defect if you want to track this separately.

Please do!

PR #10857, opened 2023-03-28
cmake: picky linker fixes for openssl, ZLIB, H3 and others

### LINKED ISSUE #10832: undefined reference for zlib symbols when building with static dependencies and BUILD_SHARED_LIBS=OFF
### I did this

We build libcurl and all our other dependencies as static libraries.  Our build script for libcurl also builds the curl program, which in retrospect we don't need.  Anyhow, it runs this for our static build:
```
cmake .. \
  -D CMAKE_BUILD_TYPE=Release \
  -D CMAKE_POSITION_INDEPENDENT_CODE=ON \
  -D CURL_USE_OPENSSL=ON \
  -D OPENSSL_INCLUDE_DIR="/home/micah/.mussels/install/host-static/include" \
  -D OPENSSL_LIBRARIES="/home/micah/.mussels/install/host-static/lib" \
  -D OPENSSL_CRYPTO_LIBRARY="/home/micah/.mussels/install/host-static/lib/libcrypto.a" \
  -D OPENSSL_SSL_LIBRARY="/home/micah/.mussels/install/host-static/lib/libssl.a" \
  -D ZLIB_INCLUDE_DIR="/home/micah/.mussels/install/host-static/include" \
  -D ZLIB_LIBRARY_RELEASE="/home/micah/.mussels/install/host-static/lib/libz.a" \
  -D LIBSSH2_INCLUDE_DIR="/home/micah/.mussels/install/host-static/include" \
  -D LIBSSH2_LIBRARY="/home/micah/.mussels/install/host-static/lib/libssh2.a" \
  -D USE_NGHTTP2=ON \
  -D NGHTTP2_INCLUDE_DIR="/home/micah/.mussels/install/host-static/include" \
  -D NGHTTP2_LIBRARY="/home/micah/.mussels/install/host-static/lib/libnghttp2.a" \
  -D CMAKE_INSTALL_PREFIX="/home/micah/.mussels/install/host-static" \
  -D BUILD_SHARED_LIBS=OFF \
  -D CMAKE_INSTALL_LIBDIR=lib
```

This worked fine with curl-7.87.0, but failed when we updated to curl-8.0.1 with this:
```
❯ make
[ 76%] Built target libcurl
[ 76%] Linking C executable curl
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libcrypto.a(c_zlib.o): in function `zlib_stateful_compress_block':
c_zlib.c:(.text+0xba): undefined reference to `deflate'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libcrypto.a(c_zlib.o): in function `zlib_stateful_finish':
c_zlib.c:(.text+0xf6): undefined reference to `deflateEnd'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libcrypto.a(c_zlib.o): in function `zlib_stateful_init':
c_zlib.c:(.text+0x230): undefined reference to `deflateInit_'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libcrypto.a(c_zlib.o): in function `bio_zlib_ctrl':
c_zlib.c:(.text+0x394): undefined reference to `deflate'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libcrypto.a(c_zlib.o): in function `bio_zlib_write':
c_zlib.c:(.text+0x73a): undefined reference to `deflate'
/usr/bin/ld: c_zlib.c:(.text+0x85b): undefined reference to `deflateInit_'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libcrypto.a(c_zlib.o): in function `bio_zlib_free':
c_zlib.c:(.text+0xb07): undefined reference to `deflateEnd'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libssh2.a(comp.c.o): in function `comp_method_zlib_dtor':
comp.c:(.text+0xa1): undefined reference to `deflateEnd'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libssh2.a(comp.c.o): in function `comp_method_zlib_comp':
comp.c:(.text+0x272): undefined reference to `deflate'
/usr/bin/ld: /home/micah/.mussels/install/host-static/lib/libssh2.a(comp.c.o): in function `comp_method_zlib_init':
comp.c:(.text+0x355): undefined reference to `deflateInit_'
make[2]: *** [src/CMakeFiles/curl.dir/build.make:784: src/curl] Error 1
make[1]: *** [CMakeFiles/Makefile2:541: src/CMakeFiles/curl.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

Additional info, we're building with a tool called Mussels that I made to automate our dependency builds for linux/mac/windows/freebsd. So this came up in https://github.com/Cisco-Talos/clamav-mussels-cookbook/pull/32 where these are the full instructions used for that "host-static" build that failed: https://github.com/Cisco-Talos/clamav-mussels-cookbook/blob/157d31d952a832403895254220f3847a27fa0c61/recipes/libcurl-7.yaml#L150-L170

### I expected the following

I expected the build to succeed ;-).  Seems like the curl application is no longer being linked with `ZLIB::ZLIB` in 8.0.1 in this particular situation. 

### curl/libcurl version

8.0.1

### operating system

`Linux DESKTOP-N3L21K2 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

vszakats:
- fix HTTP/3 support detection with OpenSSL/quictls built with ZLIB.

- fix HTTP/3 support detection with OpenSSL/quictls/LibreSSL and
  `ld` linker on Windows.

- reposition ZLIB (and other compression) detection _after_ TLS
  detection, but before calling HTTP/3-support detection via
  `CheckQuicSupportInOpenSSL`.

  Regression from ebef55a61df0094b9790710a42f63c48e7de3c13
  May fix #10832                                    

  This also seems to fix an odd case, where OpenSSL/quictls is correctly
  detected, but its header path is not set while compiling, breaking build
  at `src/curl_ntlm_core.c`.
     
- satisfy "picky" linkers such as `ld` with MinGW, that are highly
  sensitive to lib order, by also adding brotli to the beginning of the
  liblist.

- satisfy "picky" linkers by adding certain Windows systems libs to
  the liblist for OpenSSL/LibreSSL. (Might need additional ones for
  other forks.)

Note: It'd make much sense to _always_ add `ws2_32`,
`crypt32` (except Windows App targets perhaps?), `bcrypt` (except old-mingw!) 
on Windows at this point. They are almost always required, and if some
aren't, they are just ignored by the linker with no effect on final binaries.

Closes #xxxxx                                       

---

With these, a CMake build with gcc/ld _almost_ works ;) Now `gsasl` lib is in the wrong position ¯\_(ツ)_/¯.


jay:
not against this but should we be making concessions for these users that are compiling static builds? things can always break as the static library changes dependencies. for example what if the reporter did not enable zlib for curl, then isn't the problem the same, his build will fail because his ssl still requires zlib but cmake does not know that?

shouldn't the burden be on him to specify the requirements if there's no `PKG_CONFIG="pkg-config --static"` or other autotools way that cmake can't use. even then something like nghttp2 at least on windows you have to specify CPPFLAGS="-DNGHTTP2_STATICLIB" as well. it seems static builds are a lot of trouble this way.

vszakats:
@jay: It'd definitely help to state that. Static builds are extremely fragile and labour-intensive, curl-for-win being the living proof for that. An infinite amount of time can be spent tracking the variations and their changes. E.g. BoringSSL still needs `phtread` and indeed, those extra macros here and there. Also, CMake and autotools are actively working _against_ making static linking work.

On Windows, I think it'd also help if we added the base set of Windows system libs that are almost always required. They are a common source of problems there.

bagder:
I am in favor of clearly documenting that doing working completely static curl builds is not the responsibility of the curl build scripts. To set the expectations right.

vszakats:
Here's one reply (and some more below that), to a static linking report/question received earlier: https://github.com/curl/curl-for-win/issues/39#issuecomment-1322331830
(This was Windows-specific, which is a particularly problematic platform for this, but some of it likely common to all platforms.)

And an unresolved discussion with a similar request: https://github.com/curl/curl-for-win/discussions/41

These were for final apps linking libcurl, which is kind of the same as linking curl itself.

vszakats:
Unrelated CI errors in `src/telnet.c`:
```
2>c:\projects\curl\lib\telnet.c(776) : error C2220: warning treated as error - no 'object' file generated
2>c:\projects\curl\lib\telnet.c(776) : warning C4706: assignment within conditional expression
```

CMakeLists.txt
```diff
@@ -546,6 +514,59 @@ if(CURL_USE_NSS)
...
+endif()
+
+option(CURL_BROTLI "Set to ON to enable building curl with brotli support." OFF)
+set(HAVE_BROTLI OFF)
+if(CURL_BROTLI)
+  find_package(Brotli QUIET)
+  if(BROTLI_FOUND)
+    set(HAVE_BROTLI ON)
+    list(PREPEND CURL_LIBS ${BROTLI_LIBRARIES})
+    list(APPEND CURL_LIBS ${BROTLI_LIBRARIES})
```
dg0yt, line 546: I doubt it ever made sense to PREPEND in addition to APPEND.

vszakats: This has been a while ago, and can't recall the exact test results for this specific case. But, it's unlikely I'd add such thing unless required to fix the picky-linker problem. (A clean solution would be `-Wl,--start-group` / `-Wl,--end-group`, but could not figure out how to use with CMake correctly.)

If a better solution to this exists, I'd be very glad to learn about it. It's been a two-decade problem with this linker and static linking.

dg0yt: The "picky-linker problem" is unknown to me, but I know that the order of static libs matters outside MSVC.
And I know that CMake may mess with the order unless it is properly defined by target dependencies. Subject to CMake version.
(And I know that meson still sorts link libs it gets from CMake...)

FTR I'm about  to patch out the PREPEND line in the vcpkg port. Vcpkg CI builds with static library linkage for windows, linux, osx, and android. And of course I'm just tracing a linking issue. I want to get rid of distractors.

vszakats: "Picky-linker" is my term, it applies to _binutils_ `ld` specifically. It may apply to others, but I haven't seen it anywhere else. Also llvm/clang's `lld` doesn't have this problem (neither does MSVC). My understanding is that the linker does no effort to resolve symbols in a second pass, so if a symbol is referenced earlier than defined, linking breaks. It must have a strong reason doing this for decades and without a fix.

Feel free to patch it out locally, though I'm not aware of any issues it may cause. You may also move it inside an `if(GCC)`, if it does.

dg0yt: > My understanding is that the linker does no effort to resolve symbols in a second pass, so if a symbol is referenced earlier than defined, linking breaks.

It is the other way round. First the dependent lib (`-lbrotlidec`), then the independent lib (`-lbrotlicommon`). 
If you get it wrong, then you need to repeat it.

And that's why extra *prepending* is still in the "never made sense" box for me.
(Hint: `-lz` and `-lm` are typically found at the end of the link libs.)

And that's why "sort" and "remove duplicates" do not work on link libs.

vszakats: Would this help with the root cause for brotli?:
```diff
diff --git a/CMake/FindBrotli.cmake b/CMake/FindBrotli.cmake
index 11ab7f825..dcab8f7cd 100644
--- a/CMake/FindBrotli.cmake
+++ b/CMake/FindBrotli.cmake
@@ -40,4 +40,4 @@ find_package_handle_standard_args(Brotli
 )
 
 set(BROTLI_INCLUDE_DIRS ${BROTLI_INCLUDE_DIR})
-set(BROTLI_LIBRARIES ${BROTLICOMMON_LIBRARY} ${BROTLIDEC_LIBRARY})
+set(BROTLI_LIBRARIES ${BROTLIDEC_LIBRARY} ${BROTLICOMMON_LIBRARY})
```

I'm trying to replicate it, but without knowing the exact trigger configuraiton, this might take some time. If it triggers in my current test setup at all.

PR #10859, opened 2023-03-28
telnet: make MSVC ignore warning for assignment within conditional

bagder:
Follow-up to d92a5007b60e0af7d


bagder:
Hm, for some reason this does not seem to work...

PR #10870, opened 2023-03-30
Relax check on test_07_02

icing:
As only 1 connection will be used when curl is slow. Happens when address sanitized in CI, for example.


bagder (approved):

bagder:
thanks!

Closed via 041cf77cbda1dbbe9677becc8f0712463899857b (which used the wrong PR number)

PR #10876, opened 2023-03-31
rustls, fix error in recv handling

icing:
- when rustls is told to recieve more TLS data and its internal plaintext buffers are full, it returns an IOERROR
- avoid receiving TLS data while plaintext is not read empty

pytest:
- increase curl run timeout when invoking pytest with higher verbosity


bagder:
Thanks!

PR #10883, opened 2023-04-03
Improved httpd detection for pytest

icing:
- better error messages when not found/complete
- handling of `--without-test-httpd`


bagder (approved):

bagder:
Thanks!

Issue #9841, opened 2022-11-01
BUG - broken UTF 8 encoded content output


OSPanel:

OSPanel:
I can't compile the binary, but I can check the test binary if needed.

vszakats:
@OSPanel: You can find fresh Unicode builds here: https://ci.appveyor.com/project/curlorg/curl-for-win/builds/45248995/artifacts

OSPanel:
Thanks. I downloaded and tested these x86+x64 builds (dated 10/26/2022), and this bug was found there :-(

vszakats:
I freshly made these builds today, right before posting here. They have Unicode enabled (check `curl -V`).

[ For reproducibility the dates are always the same for a given curl version number. ]

OSPanel:
I used this link: https://ci.appveyor.com/project/curlorg/curl-for-win/builds/45248995/artifacts
Release-Date: 2022-10-26
```
url -V
curl 7.86.0 (x86_64-w64-mingw32) libcurl/7.86.0 OpenSSL/3.0.5 (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.2 WinIDN libssh2/1.10.0 nghttp2/1.50.0 ngtcp2/0.10.0 nghttp3/0.7.1 libgsasl/2.2.0
Release-Date: 2022-10-26
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe TLS-SRP Unicode UnixSockets zstd
```

vszakats:
@OSPanel: That's the correct one indeed, and thanks for testing it!

Based on your tests, this issue is present in both Unicode and non-Unicode builds.

bagder:
I don't see an explanation exactly what the symptoms of this error is?

Is it perhaps code from this commit that causes it? 5bfaa86ceb3c2a9ac474a928e748c4a86a703b33

vszakats:
[...]

PR #10890, opened 2023-04-05
tool_cb_wrt: fix invalid unicode for windows console

jay:
- Suppress an incomplete UTF-8 sequence at the end of the buffer.

- Attempt to reconstruct incomplete UTF-8 sequence from prior call(s) in current call.

Prior to this change, in Windows console, UTF-8 sequences split between two or more calls to the write callback would cause invalid "replacement characters" U+FFFD to be printed instead of the actual Unicode character. This is because in Windows only UTF-16 encoded characters are printed to the console, therefore we convert the UTF-8 contents to UTF-16, which cannot be done with partial UTF-8 sequences.

Reported-by: Maksim Arhipov

Fixes https://github.com/curl/curl/issues/9841
Closes #xxx

---

Untested, WIP. Also I have no automated way to test for the issue this fixes because it only happens in the Windows console.

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -87,6 +87,12 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
   }
 #endif
 
+#ifdef WIN32
+  /* Discard incomplete UTF-8 sequence buffered from body */
+  if(outs->utf8seq[0])
+    memset(outs->utf8seq, 0, sizeof(outs->utf8seq));
+#endif
+
   /*
    * Write header data when curl option --dump-header (-D) is given.
    */
--- a/src/tool_cb_wrt.c
+++ b/src/tool_cb_wrt.c
@@ -233,35 +233,131 @@ size_t tool_write_cb(char *buffer, size_t sz, size_t nmemb, void *userdata)
 
 #ifdef WIN32
   fhnd = _get_osfhandle(fileno(outs->stream));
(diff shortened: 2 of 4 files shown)


OSPanel:
I can't compile the binary, but I can check the test binary if needed. Maybe there is a link to the compiled version?

jay:
You can download artifacts from the most recent appveyor CI jobs, for example

https://ci.appveyor.com/project/curlorg/curl/builds/46705722
[CMake, VS2022, Debug x64, Schannel, Static, Unicode](https://ci.appveyor.com/project/curlorg/curl/builds/46705722/job/pt13f6r5c0nlikwp/artifacts)
You'll also need vcruntime140d.dll and ucrtbased.dll from Visual C++ 2022. See [curl.zip](https://github.com/curl/curl/files/11340701/curl.zip)

There are other ones you can download that do not require Visual C++ debugging DLLs but you may need other DLLs like OpenSSL.

OSPanel:
Artifacts in this version have become permanent, previously they did not always appear and depended on receiving a response in portions.

OSPanel:
The bug is defeated :-) I hope we'll see the fix in the release soon, thanks!

OSPanel:
Why this fix was not added in the latest update ? Can no one cope with this?

jay:
This was held up because I don't have a test for it. I will give it another look to see if we can find some way to test for it. Our test apparatus doesn't cover situations like this. I think this is a worthwhile addition and that I may have to unfortunately add it without a test.

Issue #10888, opened 2023-04-04
Incompatible pointer type in cleanarg() function (Windows build)


lordmulder:
I think `cleanarg()` in "tool_getparam.c" is incorrect on platforms where `argv_item_t` is `wchar_t` (e.g. Windows).

This has been observed in curl 8.0.1 when building with MSYS2/mingw-w64 on Windows:
```
tool_getparam.c: In function 'cleanarg':
tool_getparam.c:555:25: warning: passing argument 1 of 'strlen' from incompatible pointer type [-Wincompatible-pointer-types]
  555 |     size_t len = strlen(str);
      |                         ^~~
      |                         |
      |                         wchar_t * {aka short unsigned int *}
In file included from C:/msys64/mingw32/include/guiddef.h:154,
                 from C:/msys64/mingw32/include/winnt.h:648,
                 from C:/msys64/mingw32/include/minwindef.h:163,
                 from C:/msys64/mingw32/include/windef.h:9,
                 from C:/msys64/mingw32/include/windows.h:69,
                 from ../lib/setup-win32.h:57,
                 from ../lib/curl_setup.h:248,
                 from tool_setup.h:38,
                 from tool_getparam.c:24:
C:/msys64/mingw32/include/string.h:64:37: note: expected 'const char *' but argument is of type 'wchar_t *' {aka 'short unsigned int *'}
   64 |   size_t __cdecl strlen(const char *_Str);
      |                         ~~~~~~~~~~~~^~~~

```

**Function in question:**
```
#ifdef HAVE_WRITABLE_ARGV
static void cleanarg(argv_item_t str)
{
  /* now that GetStr has copied the contents of nextarg, wipe the next
   * argument out so that the username:password isn't displayed in the
   * system process list */
  if(str) {
    size_t len = strlen(str);
    memset(str, ' ', len);
  }
}
#else
#define cleanarg(x)
#endif
```

**Suggested fix:**
```
#ifdef HAVE_WRITABLE_ARGV
static void cleanarg(argv_item_t str)
{
  /* now that GetStr has copied the contents of nextarg, wipe the next
   * argument out so that the username:password isn't displayed in the
   * system process list */
  if(str) {
#ifdef _UNICODE
    size_t len = wcslen(str);
    wmemset(str, L' ', len);
#else
    size_t len = strlen(str);
    memset(str, ' ', len);
#endif
  }
}
#else
#define cleanarg(x)
#endif
```

**Compare to definition of `argv_item_t`:**
```
[...]

PR #10896, opened 2023-04-06
configure: don't set HAVE_WRITABLE_ARGV on Windows

bagder:
Ref: #10888


bagder:
I can't spot this check being done in cmake at all... :open_mouth:

bagder:
thanks @jzakrzewski !

PR #10898, opened 2023-04-06
tests, increase sws timeout for more robust testing

icing:
- for https CONNECT forwarding, this was fixed at 5 seconds which led to spurious CI test failures
- add --keepalive parameter to sws to control this
- let httpserver use 30 seconds


bagder:
Thanks!

PR #10900, opened 2023-04-06
CURLPROXY_HTTPS2: for HTTPS proxy that may speak HTTP/2

bagder:
Setting this proxy type allows curl to negotiate and use HTTP/2 with HTTPS proxies.

(There is no way for the curl tool to set this yet)


icing (approved):

olife-cyber:
#

PR #10906, opened 2023-04-07
http2: remove check for !data as it was already dereferenced

bagder:
Pointed out by Coverity


dfandrich:
Evidently, gcc 11 in the Slackware-openssl-with-gssapi-gcc builds has a different opinion than Coverity:
```
http2.c: In function 'http2_data_setup.isra':
http2.c:225:18: error: null pointer dereference [-Werror=null-dereference]
  225 |   if(!data->req.p.http) {
      |       ~~~~~~~~~~~^~~~~
```

PR #10913, opened 2023-04-09
Clean command-line on Win32 to hide possible sensitive information fr…

lordmulder:
Clean command-line on Win32 to hide possible sensitive information from Task Manager et al., similar to what `cleanarg()` already does on other platforms. See #10888 for details.

### DIFF
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -763,6 +763,7 @@ static void init_terminal(void)
 }
 
 /* clean possible sensitive information from the win32 command-line */
+#ifdef NDEBUG
 #define _acmdln (*__p__acmdln())
 #define _wcmdln (*__p__wcmdln())
 _CRTIMP char    **__cdecl __p__acmdln(void);
@@ -772,6 +773,9 @@ static void clean_cmdln(void)
   memset(_acmdln, 'x', strlen(_acmdln));
   wmemset(_wcmdln, L'x', wcslen(_wcmdln));
 }
+#else
+#define clean_cmdln() ((void)0)
+#endif
 
 LARGE_INTEGER tool_freq;
 bool tool_isVistaOrGreater;
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -768,7 +768,7 @@ _CRTIMP char    **__cdecl __p__acmdln(void);
 _CRTIMP wchar_t **__cdecl __p__wcmdln(void);
 static void clean_cmdln(void)
 {
-  memset (_acmdln,  'x', strlen(_acmdln));
+  memset(_acmdln, 'x', strlen(_acmdln));
   wmemset(_wcmdln, L'x', wcslen(_wcmdln));
 }
 
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -32,6 +32,7 @@
 #ifdef WIN32
 #  include <stdlib.h>
 #  include <tlhelp32.h>
+#  include <wchar.h>
 #  include "tool_cfgable.h"
 #  include "tool_libinfo.h"
 #endif
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -761,6 +761,17 @@ static void init_terminal(void)
   }
 }
 
+/* clean possible sensitive information from the win32 command-line */
+#define _acmdln (*__p__acmdln())
+#define _wcmdln (*__p__wcmdln())
+_CRTIMP char    **__cdecl __p__acmdln(void);
+_CRTIMP wchar_t **__cdecl __p__wcmdln(void);
+static void clean_cmdln(void)
+{
+  memset (_acmdln,  'x', strlen(_acmdln));
+  wmemset(_wcmdln, L'x', wcslen(_wcmdln));
+}
+
 LARGE_INTEGER tool_freq;
 bool tool_isVistaOrGreater;
 
@@ -775,6 +786,8 @@ CURLcode win32_init(void)
     tool_isVistaOrGreater = false;
 
   QueryPerformanceFrequency(&tool_freq);
+
+  clean_cmdln();
 
   init_terminal();
 


lordmulder:
Not sure why some checks still fail. Seems unrelated to my code change.

MarcelRaad:
Yes, most failures look unrelated. This MinGW-w64 one doesn't though:
https://ci.appveyor.com/project/curlorg/curl/builds/46737844/job/nuiy4dt8r60byh57
```
CMakeFiles/curl.dir/objects.a(tool_doswin.c.obj): In function `clean_cmdln':
C:/projects/curl/src/tool_doswin.c:772: undefined reference to `__imp___p__acmdln'
C:/projects/curl/src/tool_doswin.c:772: undefined reference to `__imp___p__acmdln'
C:/projects/curl/src/tool_doswin.c:773: undefined reference to `__imp___p__wcmdln'
C:/projects/curl/src/tool_doswin.c:773: undefined reference to `__imp___p__wcmdln'
```

bagder:
This seems to wipe out the entire command line. On non-windows we only clean a number of specific arguments that are likely to contain sensitive data, because users might want to use the command line arguments to differentiate curl invocations. Aren't users likely to want to do that on Windows as well?

lordmulder:
> This seems to wipe out the entire command line. On non-windows we only clean a number of specific arguments that are likely to contain sensitive data, because users might want to use the command line arguments to differentiate curl invocations. Aren't users likely to want to do that on Windows as well?

On Windows, the whole command.-line is passed to the application as just a single string. In C/C++ programs, it's the C runtime library that breaks down the string into the `argv[]` array, before passing it to the `main()` or `wmain()`functions. But the memory where `argv[]` is stored is totally separate; the memory where the *original* command-line is stored is accessible in C/C++ via `_acmdln` and  `_wcmdln`. If you wanted to wipe out only certain `argv[]` elements, then you'd have to map back from the `argv[]` index to a particular sub-range of `_acmdln` and/or `_wcmdln`. I think this would be very difficult to do...

src/tool_doswin.c
```diff
@@ -763,6 +763,7 @@ static void init_terminal(void)
 }
 
 /* clean possible sensitive information from the win32 command-line */
+#ifdef NDEBUG
```
gvanem, line 766: A `NDEBUG` does not mean a *Release Build*. 
This better test for `#if !defined(_DEBUG)` instead.

bagder:
Do other windows command line programs "normally" wipe the command line like this?

src/tool_doswin.c
lordmulder: > A `NDEBUG` does not mean a _Release Build_. This better test for `#if !defined(_DEBUG)` instead.

I'm not sure about this. I think `NDBUG` is *standard* and disables debug assertions. So this is expected to be defined for "release" builds. `_DEBUG` is *MSVC-specicifc* and is defined by MSVC when linking against the "debug" CRT (option `/MDd` or `/Mtd`). Is `_DEBUG` set by (or does it have any significance for) other compilers, like MinGW/GCC or Clang?

https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170#remarks

> Do other windows command line programs "normally" wipe the command line like this?

Well, I cannot give you a concrete example. But from the [*The Old New Thing* article](https://devblogs.microsoft.com/oldnewthing/20091125-00/?p=15923) it should be clear that even though "snooping" another process' command-line is possible, you really shouldn't rely on it:

> Remember that from Win32’s point of view, the command line is just a string that is copied into the address space of the new process. How the launching process and the new process interpret this string is governed not by rules but by convention.

> What’s more, since the string is merely a “preinitialized variable”, a process could in principle (and many do in practice) write to the memory that holds the command line, in which case, if you go snooping around for it, you’ll see the modified command line. There is **no** secret hiding place where the kernel keeps the “real original command line”

> What does this mean for people who disregard this principle and go after the command line of another process? You have to understand what you are getting is non-authoritative information. In fact, it’s worse. It’s information the application itself may have changed in order to try to fool you, so do **not** use it to make important decisions.

jay:
> Do other windows command line programs "normally" wipe the command line like this?

The only Windows applications erasing the command line are malware. I understand that in Linux curl's cleanarg may wipe user/pass from the argument list, though I think that is more about convenience than security.

You can see another user's command line in Linux (unless hidepid and even then it may still be [accessible](https://access.redhat.com/solutions/6704531) via systemd) but cleanarg doesn't stop someone snatching up the full command line of other users before curl can modify them. IMO that is either an OS problem or a user problem, don't put sensitive information on the command line if you have an OS that is going to show it to other users. Windows does not allow one user to see another user's command line unless they are an admin or possibly have some restricted privilege.

I'm against this PR.

bagder:
> The only Windows applications erasing the command line are malware.

This is certainly an indication that we don't need to do this.

> I understand that in Linux curl's cleanarg may wipe user/pass from the argument list, though I think that is more about convenience than security.

Indeed it is. It is rather weak protection against anything so we keep it mostly because we added it once and it might help avoid some casual mistakes, but it is not to be considered a secure way to avoid secrets used on the command line to get seen/captured by other users running on the same system. Which is also why we don't consider [omissions from this to be a security issue](https://curl.se/dev/secprocess.html#visible-command-line-arguments).

Of course, a significant difference and explanation for the different approaches is that Windows has traditionally been a single-user systems while *nix systems have been multi-user since the beginning.

src/tool_doswin.c
gvanem: >  I think NDEBUG is standard and disables debug assertions. 

Disabling `assert()` is madness even for *Release mode* IMHO.

>  Is _DEBUG set by (or does it have any significance for) other compilers, like MinGW/GCC or Clang

`clang-cl` has it (`-MDd` or `-MTd`). MinGW does not have `_DEBUG` mode. Hence `!defined(_DEBUG)` is TRUE.
But why do you want `clean_cmdln()` only for *Release mode*?

lordmulder: ...because otherwise some of the checks fail 😞

jay:
Not to get too into the weeds on this but I am referring to an access token so yes if your access token has admin privileges or some other appropriate privilege then yes you can see the command line of other users. IMO that is exactly the way it should be.

> > The only Windows applications erasing the command line are malware
> 
> Any sources for this?

Yes, me. I've seen a few instances of malware change the command line, process name, etc. I've never seen a legit app do that.

lordmulder:
> Not to get too into the weeds on this but I am referring to an access token so yes if your access token has admin privileges or some other appropriate privilege then yes you can see the command line of other users. IMO that is exactly the way it should be.

Right. My point was: Pretty much **any** "normal" Windows _does_ have "full" (administrator) privileges. It's the default configuration. Technically, of course, you can create a user with **_limited_** privileges. And enterprises probably do this for their employees. But, apart from that, **no** Windows user usually sets up user accounts with **_limited_** privileges...

And, because even Microsoft knows that 99% of all users use "administrator" accounts for their daily work, UAC was invented: Ever since Vista, each user has **two** access tokens: One with "administrator" privileges, and one with **_limited_** privileges, the latter of which is used by default to create new processes – unless elevated privileges are requested explicitly...

bagder:
Thanks for your contribution, but I don't sense much love for this to get merged.

PR #10919, opened 2023-04-11
runtests: further refactoring on the path to parallel tests

dfandrich:
...with a few bug fixes discovered on the way.


dago:
I tried to reproduce the failing testcase and just tried with the code from 8.0.1 in my personal build directory and get this output:
```
test 1466...OK (1291 out of 1577, remaining: 04:48, took 0.658s, duration: 21:40)
socksd: socket path must be shorter than 108 chars
RUN: failed to start the SOCKS2 server
test 1469...OK (1294 out of 1577, remaining: 04:46, took 0.439s, duration: 21:50)
```

The relevant code in `socksd.c` is
```
#ifdef USE_UNIX_SOCKETS
        struct sockaddr_un sau;
        unix_socket = argv[arg];
        if(strlen(unix_socket) >= sizeof(sau.sun_path)) {
          fprintf(stderr,
                  "socksd: socket path must be shorter than %zu chars\n",
              sizeof(sau.sun_path));
          return 0;
        }
        socket_domain = AF_UNIX;
        socket_type = "unix";
#endif
```

The contents of `unix_socket` depends on the build path and that is
```
/home/dam/mgar/pkg/curl/trunk/work/solaris10-sparc/build-isa-sparcv8plus-features-full/curl-8.0.1/tests/socks21418.sock
```
which is 120 characters and therefore longer than 108 characters. For the buildbot environment this is
```
/export/home/buildbot-unstable10s/slave/curl-unthreaded-solaris10-sparc/build/tests/socks21418.sock
```
which is 100 characters and pretty close to the limit and with your latest change will exceed the limit.

I suggest two changes:

1. Print the string of the path with the calculated exceeding length in the error message in `socksd.c` making debugging this trivial
2. The error message outputted from `socksd` was not in the debug log. This would be helpful.
3. Allow a configurable directory for temporary files during testing to allow to circumvent this issue as it cannot be guaranteed in all circumstances to have a sufficiently short pathname - maybe an environment variable. This would allow me to set the path for test sockets to something reasonably short like `/tmp/curltest/`.

PR #10926, opened 2023-04-11
curl: add --proxy-http2

bagder:
For trying HTTP/2 with an HTTPS proxy.


bagder:
Nah, this is more user-facing and the protocol is named HTTP/2 after all, and there is an equivalent non-proxy option called `--http2`...

icing (approved):

PR #10928, opened 2023-04-11
vtls. Fix V220 for sizeof(buf->data)

kvarec:
V220 Suspicious sequence of types castings: memsize -> 32-bit integer -> memsize. The value being cast: 'sizeof (buf->data)'. curl\lib\vtls\vtls.c 2025

https://pvs-studio.com/en/docs/warnings/v220/

### DIFF
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -2022,7 +2022,7 @@ CURLcode Curl_alpn_to_proto_str(struct alpn_proto_buf *buf,
     len = strlen(spec->entries[i]);
     if(len >= ALPN_NAME_MAX)
       return CURLE_FAILED_INIT;
-    if(off + len + 2 >= (int)sizeof(buf->data))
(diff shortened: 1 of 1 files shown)


lib/vtls/vtls.c
```diff
@@ -2022,7 +2022,7 @@ CURLcode Curl_alpn_to_proto_str(struct alpn_proto_buf *buf,
     len = strlen(spec->entries[i]);
     if(len >= ALPN_NAME_MAX)
       return CURLE_FAILED_INIT;
-    if(off + len + 2 >= (int)sizeof(buf->data))
+    if(off + len + 2 >= sizeof(buf->data))
```
bagder, line 2025: I can just add that the sizeof() in question here is 14 bytes, so this typecast is not "Suspicious" even if superfluous.

bagder: Sorry, `40` is the correct size.

bagder:
Thanks!

PR #10934, opened 2023-04-11
Fix https connection setup to treat ssl_mode correctly.

icing:
- for HTTPS protocol, a disabled ssl should never be acceptables.


bagder:
Thanks!

PR #10949, opened 2023-04-13
GHA: suppress git clone output

pheiduck:
Follow-up: https://github.com/curl/curl/commit/8203aa6ed405ec832d2c62f18dfda293f89a23f9


bagder:
Thanks!

PR #10951, opened 2023-04-13
HTTTP/3 all backends, check stream_ctx more thoroughly

icing:
- callbacks and filter methods might be invoked at unexpected times, e.g. when the transfer's stream_ctx has not been initialized yet or, more likely, has already been taken down.
- check for existance of stream_ctx in such places and return an error or silently succeed the call.

### DIFF
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -1055,12 +1055,13 @@ static bool stream_is_writeable(struct Curl_cfilter *cf,
 {
   struct cf_quiche_ctx *ctx = cf->ctx;
   struct stream_ctx *stream = H3_STREAM_CTX(data);
+  quiche_stream_iter *qiter;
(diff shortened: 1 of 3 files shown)


lib/vquic/curl_quiche.c
```diff
@@ -344,6 +344,8 @@ static CURLcode write_resp_raw(struct Curl_cfilter *cf,
   ssize_t nwritten;
 
   (void)cf;
+  if(!stream)
```
pheiduck, line 347: Compiler complaines about: `quiche_stream_iter *qiter = quiche_conn_writable(ctx->qconn);`

May this will fix it?:

```
if(!stream) {
   return FALSE;
}
```

icing: Thanks! Just pused a fix.

bagder:
Thanks!

PR #10952, opened 2023-04-13
HTTP/2 proxy tunnel, fix ingress handling to stop too early

icing:
- progress ingress stopped too early, causing data from the underlying filters to not be processed and report that no tunnel data was available
- this lead to "hangers" where no socket activity was seen but data rested in buffers


bagder:
Thanks!

PR #10961, opened 2023-04-13
Disable socket receive buffer, unless USE_RECV_BEFORE_SEND_WORKAROUND…

icing:
… is in place

While we would like to use this, we have stalls in parallel transfers where not all buffered data is consumed and no socket events happen.


jay:
Thanks. This is a difficult and time consuming area.

PR #10973, opened 2023-04-16
cmake: speed up and extend picky clang/gcc options

vszakats:
Extend existing picky compiler options with ones missing compared to autotools builds. Also sync options between clang and gcc.

Redesign the way we enable these options to avoid the slow option detection almost completely.

This reduces the number of detections from 35 to zero for clang and 3 for gcc, even after adding a bunch of new options.

clang 3.0 (2011-11-29) and gcc 2.95 (1999-07-31) now required.

Also show enabled picky options.

Ref: https://github.com/libssh2/libssh2/pull/952

Closes #10973

### DIFF
--- a/CMake/PickyWarnings.cmake
+++ b/CMake/PickyWarnings.cmake
@@ -1,0 +1,173 @@
+include(CheckCCompilerFlag)
+
+if(PICKY_COMPILER)
+  if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+
+    # https://clang.llvm.org/docs/DiagnosticsReference.html
+    # https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
+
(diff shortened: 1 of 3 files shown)


vszakats:
This is unrelated and happens also without this patch:
```
tool_operhlp.c:112:5: warning: Value stored to 'uerr' is never read [deadcode.DeadStores]
    uerr = curl_url_get(uh, CURLUPART_QUERY, &query, 0);
    ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```
https://dev.azure.com/daniel0244/curl/_build/results?buildId=15247&view=logs&jobId=e2e9a903-bd95-5516-8620-b8f4e5800a42&j=e2e9a903-bd95-5516-8620-b8f4e5800a42&t=b96a1c9b-30a9-56e9-62cc-d046c220147f

bagder (approved):

PR #10975, opened 2023-04-16
cf-socket: turn off IPV6_V6ONLY on Windows if it is supported

lifenjoiner:
It makes support for IPv4-mapped IPv6 addresses.

IPV6_V6ONLY refs:
https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses
https://github.com/golang/go/blob/master/src/net/ipsock_posix.go
https://en.wikipedia.org/wiki/Unix-like
off:
https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html#proc-sys-net-ipv6-variables
https://man.netbsd.org/inet6.4
https://man.freebsd.org/cgi/man.cgi?query=inet6
https://github.com/apple-oss-distributions/xnu/blob/main/bsd/man/man4/inet6.4
on:
https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options
acts like off, but returns 1 and dummy setting:
https://man.dragonflybsd.org/?command=inet6
https://man.dragonflybsd.org/?command=ip6
unsupported and read-only returns 1:
https://man.openbsd.org/inet6.4

default value refs:
https://datatracker.ietf.org/doc/html/rfc3493#section-5.3
https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html#proc-sys-net-ipv6-variables

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -84,14 +84,16 @@
 /* It makes support for IPv4-mapped IPv6 addresses.
  * Linux kernel, NetBSD, FreeBSD and Darwin: default is off;
  * Windows Vista and later: default is on;
- * DragonFly BSD: off and dummy setting;
- * OpenBSD and previous Windows: unsupported.
+ * DragonFly BSD: acts like off, and dummy setting;
+ * OpenBSD and earlier Windows: unsupported.
  * Linux: controlled by /proc/sys/net/ipv6/bindv6only.
  */
 static void set_ipv6_v6only(curl_socket_t sockfd, int on)
 {
   (void)setsockopt(sockfd, IPPROTO_IPV6, IPV6_V6ONLY, (void *)&on, sizeof(on));
 }
+#else
+#define set_ipv6_v6only(x,y)
 #endif
 
 static void tcpnodelay(struct Curl_easy *data, curl_socket_t sockfd)
@@ -959,9 +961,7 @@ static CURLcode cf_socket_open(struct Curl_cfilter *cf,
 
 #ifdef ENABLE_IPV6
   if(ctx->addr.family == AF_INET6) {
-#if defined(IPV6_V6ONLY) && defined(WIN32)
     set_ipv6_v6only(ctx->sock, 0);
-#endif
     ipmsg = "  Trying [%s]:%d...";
   }
   else
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -80,6 +80,20 @@
 #include "memdebug.h"
 
 
+#if defined(ENABLE_IPV6) && defined(IPV6_V6ONLY) && defined(WIN32)
+/* It makes support for IPv4-mapped IPv6 addresses.
+ * Linux kernel, NetBSD, FreeBSD and Darwin: default is off;
+ * Windows Vista and later: default is on;
+ * DragonFly BSD: off and dummy setting;
+ * OpenBSD and previous Windows: unsupported.
+ * Linux: controlled by /proc/sys/net/ipv6/bindv6only.
+ */
+static void set_ipv6_v6only(curl_socket_t sockfd, int on)
+{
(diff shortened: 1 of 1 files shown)


bagder:
Quoting [section 5.3 of RFC 3493](https://www.rfc-editor.org/rfc/rfc3493#section-5.3):

> Note - This option has no effect on the use of IPv4 Mapped addresses

Which sounds like the opposite to the reason for this PR?

lifenjoiner:
The whole note in [section 5.3 of RFC 3493](https://www.rfc-editor.org/rfc/rfc3493#section-5.3):
>   Note - This option has no effect on the use of IPv4 Mapped addresses
>   which enter a node as a valid IPv6 addresses for IPv6 communications
>   as defined by Stateless IP/ICMP Translation Algorithm (SIIT) [5].

That is confusing. I guess it means _in separate translator "boxes"_ by the [SIIT (rfc2765) abstract section](https://www.rfc-editor.org/rfc/rfc2765):

>   The algorithm
>   translates between IPv4 and IPv6 packet headers (including ICMP
>   headers) in separate translator "boxes" in the network without
>   requiring any per-connection state in those "boxes".

Some facts on IPv4-mapped IPv6 addresses support:
* Turning off `IPV6_V6ONLY` does make it work on IPv4-mapped IPv6 addresses with dual-stack socket.
* Most OSes works on it by default. while Windows don't.
* Browsers like Chrome and Firefox can visit them, while IE doesn't, on Windows.
* Windows, `ping` works, while `nslookup` doesn't. `dig` does.

lifenjoiner:
The problem is Windows doesn't follow the RFC to set `IPV6_V6ONLY` off by default, but requires users to do it when they want to create dual-stack socket for IPv4-mapped IPv6 addresses support.
Search `IPV6_V6ONLY` in [the options table](https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options).

jay:
Does it really need a separate function?
~~~c
#if defined(IPV6_V6ONLY) && defined(WIN32)
  /* Allow IPv4 with AF_INET6 sockets on Windows. Other OSes allow this by default. */
  {
    int on = 0;
    setsockopt(ctx->sock, IPPROTO_IPV6, IPV6_V6ONLY, (void *)&on, sizeof(on));
  }
#endif
~~~

lifenjoiner:
> Does it really need a separate function?

Not really.

A function is for other place to reuse it. And, if not, the line is longer than 79 chars with `(void)` return. Compilers would optimize it for binaries :p

lifenjoiner:
> > Does it really need a separate function?
> 
> Not really.
> 
> A function is for other place to reuse it. And, if not, the line is longer than 79 chars with `(void)` return. Compilers would optimize it for binaries :p

The real reason is, I saw "we write C89 code", just recalled.
https://github.com/curl/curl/blob/6b1e4dc6cdd2c1fc1730dc64aa5e2d82615e5993/docs/CODE_STYLE.md?plain=1#L50

lib/cf-socket.c
```diff
@@ -80,6 +80,20 @@
...
+ * Windows Vista and later: default is on;
+ * DragonFly BSD: off and dummy setting;
+ * OpenBSD and previous Windows: unsupported.
+ * Linux: controlled by /proc/sys/net/ipv6/bindv6only.
+ */
+static void set_ipv6_v6only(curl_socket_t sockfd, int on)
+{
+  (void)setsockopt(sockfd, IPPROTO_IPV6, IPV6_V6ONLY, (void *)&on, sizeof(on));
+}
+#endif
```
bagder, line 95: If you do an `#else` here and provide an empty macro for that case, you can avoid the extra set of `#ifdef` later and make the code using this function easier on the eyes!

lifenjoiner: Updated.

bagder (approved):

bagder:
Thanks!

Issue #10976, opened 2023-04-16
test1097 failed:Address already in use (98)


eaglegai:
curl version: 7.79.1

runtests.pl(now servers.pm):
```
in sub runhttptlsserver {
    my $port = 24367;
    my ($httptlspid, $pid2);
    for (1 .. 10) {
        $port += int(rand(800));
        my $allflags = "--port $port $flags";

        my $cmd = "$httptlssrv $allflags > $logfile 2>&1";
        ($httptlspid, $pid2) = startnew($cmd, $pidfile, 10, 1);
```
the port range is (24367,25167)
```
in sub runhttpsserver {
    my $port = 24512; # start attempt
    for (1 .. 10) {
        $port += int(rand(600));
        my $options = "$flags --accept $port";

        my $cmd = "$perl $srcdir/secureserver.pl $options";
        ($httpspid, $pid2) = startnew($cmd, $pidfile, 15, 0);
```
the port range is (24512,25112)

in some case they whill use the same port,which causes test1097 failed by 'Address already in use'

eaglegai:
here is the log:
test319:
startnew: /usr/bin/gnutls-serv --port 24687 --http --priority NORMAL:+SRP --srppasswd ../../tests/certs/srp-verifier-db --srppasswdconf ../../tests/certs/srp-verifier-conf > log/httptls_server.log 2>&1
startnew: .httptls_server.pid faked with pid=1671741
RUN: HTTPTLS server PID 1671741 port **24687**
* pid httptls => 1671741 1671741
......

test1097:
startnew: perl -I../../tests ../../tests/httpserver.pl --pidfile ".http_server.pid" --logfile "log/http_server.log" --portfile .http_server.port --ipv4 --port 0 --srcdir "../../tests/data/.."
RUN: ../src/curl --max-time 13 --output log/http_verify.out --silent --verbose --globoff "http://127.0.0.1:33787/verifiedserver" 2>log/http_verify.log
CMD (0): ../src/curl --max-time 13 --output log/http_verify.out --silent --verbose --globoff "http://127.0.0.1:33787/verifiedserver" 2>log/http_verify.log
RUN: HTTP server is on PID 1682919 port 33787
* pid http => 1682919 1682919
[...]

PR #10977, opened 2023-04-16
tests: fix test1097 failed

eaglegai:
fix https://github.com/curl/curl/issues/10976

Signed-off-by: eaglegai <eaglegai@163.com>


eaglegai:
close

PR #10978, opened 2023-04-16
Use semicolons instead of commas

afalkenhahn:
I think we should use semicolons here instead of commas. Using commas causes a warning when compiling curl using the vbcc compiler.


MarcelRaad (approved):

bagder (approved):
indeed!

bagder:
Thanks!

PR #10980, opened 2023-04-16
ci/openssl-3.1.0+quic updates

pheiduck:


pheiduck:
test `2501` and `2502` aren't working until now.
Maybe when **openssl-3.1.0+quic** is out of _beta/dev_ state.

pheiduck:
@bagder is there a way to explicit disable tests for  ngtcp2 CIs?

bagder:
> @bagder is there a way to explicit disable tests for ngtcp2 CIs?

No... but I also don't think we should disable this test in a HTTP/3 build.

pheiduck:
Will do revert later than.

pheiduck:
if I understand it correctly, the connection with QUIC is not established here.

pheiduck:
@bagder, this PR is ready for review/merge.

PR #10983, opened 2023-04-16
tests/sshserver.pl: Define AddressFamily earlier

sevan:
As the comment states "Address family must be specified before ListenAddress", otherwise the tests fail to run `"failed starting SSH server" 52 times (582, 583, 600, 601, 602, 603, 604, 605, 606 and 43 more)`

### DIFF
--- a/tests/sshserver.pl
+++ b/tests/sshserver.pl
@@ -725,7 +725,7 @@ push @cfgarr, '#';
 #
 if(sshd_supports_opt('AddressFamily','any')) {
     # Address family must be specified before ListenAddress
-    splice @cfgarr, 14, 0, 'AddressFamily any';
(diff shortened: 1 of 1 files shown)


sevan:
Issue happened due to an ancient version of OpenSSH (v5.1) being used by accident on test system. Installing OpenSSH v9.3 and running the test suite is fine without this proposed change, so it's only needed for backwards compat testing.

bagder:
@dfandrich I assigned this review to you since you're all over the test scripts these days

dfandrich:
Thanks! I've submitted this with an extra no-op line added in a conditional earlier in the file so that this line gets added at exactly the same point in both a Windows and non-Windows config.

PR #10995, opened 2023-04-18
runtests refactoring

dfandrich:
Most has to to with positioning for parallel testing, but a number of
test bugs discovered on the way are fixed as well.

### DIFF
--- a/tests/servers.pm
+++ b/tests/servers.pm
@@ -61,6 +61,7 @@ BEGIN {
             serverfortest
             stopserver
             stopservers
+            subvariables
         )
     );
 }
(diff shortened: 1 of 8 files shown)


tests/servers.pm
```diff
@@ -1365,20 +1361,25 @@ sub runhttpsserver {
     my $pid2;
     my $httpspid;
     my $port = 24512; # start attempt
+    my %usedports = reverse %PORT;
     for (1 .. $serverstartretries) {
-        $port += int(rand(600));
+        $port += 1 + int(rand(600));
```
bagder, line 1366: Maybe all the `rand()` calls for the next port number could use the same number (900) ? I can't think of any good reason to make this different for different servers (even if I am the one who made them so to begin with).

dfandrich: I've reworked the port ranges used so there is now guaranteed to be no overlap between random port ranges used by different servers.

Issue #10988, opened 2023-04-17
HTTP2 connections stuck when paused and resumed multiple times


lemourin:
I'd reopen https://github.com/curl/curl/issues/10971 if I could. The fix for that bug introduced a new regression, please see https://github.com/curl/curl/issues/10971#issuecomment-1510308198.

I did more testing today with my real app [coro-cloudstorage](https://github.com/lemourin/coro-cloudstorage) and I noticed that the downloads are now getting stuck. I can provide more context on what the app does if necessary. I changed the [test program](https://pastebin.com/WwAZEgfA) a bit, it now resumes some transfers; with the memory leak fix one transfer gets stuck and returns "Stream error in the HTTP/2 framing layer" after ~5 minutes.

Without https://github.com/curl/curl/pull/10972 it leaks, with https://github.com/curl/curl/pull/10972 the connections get stuck.

lemourin:
The title might be a bit imprecise. The test program does a combination of abort / pause / resume:

1. fire 5 requests
2. pause them after 16mb received
3. fire 5 more requests after 3s
4. abort requests started in 3.
5. resume requests started in 1.
6. one of those 5 requests will get stuck

All requests are fired to the same URL. The same code works fine with HTTP 1.1.

bagder:
> one of those 5 requests will get stuck

Just one? Will the others go on? 

Are these h2 requests all done over the same connection?

lemourin:
> > one of those 5 requests will get stuck
> 
> Just one? Will the others go on?

It's not deterministic, anything from 0 to 5 will be stuck; it probably depends on timing.
> 
> Are these h2 requests all done over the same connection?

They should, no? All requests use the same multi handle.

lemourin:
I'm looking at the `multi.c` code. Shouldn't `data->state.tempwrite` be freed when the connection is closed instead of when a single transfer using that connection is aborted?

bagder:
`data` is the handle for a single transfer. `data->state.tempwrite` is data cached for that specific single transfer. When `multi_done()` is called, that single transfer is done, it has completed. There's no point in keeping that data around anymore at that point.

The freeing of that data is completely unrelated to the connection's lifetime or when it gets closed or not.

lemourin:
[...]

PR #10998, opened 2023-04-18
Set drain and expire transfer in HTTP/2 that is unpaused

icing:
- #10988 reports transfer stalling when pausing/unpausing
- if the servers send window is exhausted before the pause, it will not send more on unpausing until the transfer consumes its data.
- make sure unpaused transfers get drained and expire, so any data already buffered will be processed. Which triggers a WINDOW_UPDATE to the server and it will start sending again.


icing:
Closing, as superceded by #11005

PR #11002, opened 2023-04-19
http: store the password in the correct variable

bagder:
Typo from fc2f1e547a4a, detected by Coverity (because there's dead code due to this).


icing (approved):
Oopsie.

PR #11003, opened 2023-04-20
HTTP/3, quiche: terminate h1 response header when no body is sent

icing:
- fixes a failure in test2501 where a response without body was missing the final empty line


bagder (approved):

bagder:
Thanks!

PR #11006, opened 2023-04-21
http tests using libcurl, urlparser addition, h2 push, pausing

icing:
* adding `tests/http/clients` with specific clients using `libcurl`. these clients are built when you run `pytest`
* adding client `h2-serverpush` to test HTTP/2 PUSH feature
* adding internal method `curl_url_set_authority()` to process the `:authority` header of a PUSH_PROMISE
* adding `test_09_02` to verify PUSH handling
* adding client `h2-download` to verify parallel downloads via `libcurl` and with the possibility to pause/unpause transfers
* adding `test_02_21` to verify lib downloads and pausing behaviour 

Based on top of the #11005 changes.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -288,7 +288,6 @@ jobs:
     - if: ${{ contains(matrix.build.install_steps, 'pytest') }}
       run: |
         sudo apt-get install apache2 apache2-dev libnghttp2-dev
-        sudo python3 -m pip install impacket pytest cryptography multipart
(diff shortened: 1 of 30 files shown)


bagder:
Thanks!

dfandrich:
This PR breaks out-of-tree automake builds.  ws-data.c has `#include "../../../lib/curl_setup.h"` which make's VPATH loads from `srcdir`, but it in turn includes "curl_config.h" which isn't found because it's in `builddir` and therefore not in VPATH. I'll see what I can do about it.

PR #11021, opened 2023-04-25
urldata: CURLOPT_AWS_SIGV4 value is null-terminated

monnerat:
Without this fix, this option's value is not copied upon calling curl_easy_duphandle().


bagder (approved):

jay:
Thanks

PR #11024, opened 2023-04-25
Fix NULL pointer access

afalkenhahn:
We need to check against NULL here because `kbd_callback` can be called before `ssh_attach` stuffs the easy handle into the session's abstract storage.

### DIFF
--- a/lib/vssh/libssh2.c
+++ b/lib/vssh/libssh2.c
@@ -203,7 +203,7 @@ kbd_callback(const char *name, int name_len, const char *instruction,
   (void)instruction;
   (void)instruction_len;
 #endif  /* CURL_LIBSSH2_DEBUG */
-  if(num_prompts == 1) {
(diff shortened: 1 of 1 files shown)


bagder (approved):

jay:
How come this has never been an issue before? Shouldn't data always be valid there? The only way I can see this happening is the "abstract" (user-supplied pointer) parameter was NULL the last time it was set, and AFAICT we only set abstract on init when CURL_DEBUG but I'm not sure why:

https://github.com/curl/curl/blob/b16d1fa8ee567b52c09a0f89940b07d8491b881d/lib/vssh/libssh2.c#L3272-L3278

jay:
Closing in favor of #11026. @afalkenhahn please test

afalkenhahn:
I can confirm that #11026 fixes the issue.

jay:
Thanks

PR #11027, opened 2023-04-26
docs: use consistent code style in manual pages

emanuele6:
I was reading curl_unescape(3) and I noticed that there was an extra space after the open parenthesis in the SYNOPSIS; I removed the extra space.

I also ran a few  grep -r  commands to find and remove extra spaces after '(' in other files, and to find and replace uses of `T*` instead of `T *`.

And I also fixed some formatting inconsistencies: in curl_unescape(3), all function parameter were mentioned with bold text except length, that was mentioned as 'length'; and, in curl_easy_unescape(3), all parameters were mentioned in bold text except url that was italicised. Now they are all mentioned in bold.
Documentation is not very consistent in how function parameter are formatted: many pages italicise them, and others display them in bold text; but I think it makes sense to at least be consistent with formatting within the same page.

### DIFF
--- a/docs/libcurl/curl_easy_unescape.3
+++ b/docs/libcurl/curl_easy_unescape.3
@@ -39,7 +39,7 @@ that are URL encoded (%XX where XX is a two-digit hexadecimal number) are
 converted to their binary versions.
 
 If the \fBlength\fP argument is set to 0 (zero), \fIcurl_easy_unescape(3)\fP
-will use strlen() on the input \fIurl\fP string to find out the size.
+will use strlen() on the input \fBurl\fP string to find out the size.
 
 If \fBoutlength\fP is non-NULL, the function will write the length of the
 returned string in the integer it points to. This allows proper handling even
--- a/docs/libcurl/curl_easy_unescape.3
+++ b/docs/libcurl/curl_easy_unescape.3
@@ -33,13 +33,13 @@ char *curl_easy_unescape(CURL *curl, const char *input,
                          int inlength, int *outlength);
 .fi
 .SH DESCRIPTION
-This function converts the given URL encoded \fBinput\fP string to a "plain
-string" and returns that in an allocated memory area. All input characters
-that are URL encoded (%XX where XX is a two-digit hexadecimal number) are
-converted to their binary versions.
+This function converts the URL encoded string \fBinput\fP to a "plain string"
+and returns that in an allocated memory area. All input characters that are URL
+encoded (%XX where XX is a two-digit hexadecimal number) are converted to their
+binary versions.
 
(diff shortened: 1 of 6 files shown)


docs/libcurl/opts/CURLMOPT_SOCKETFUNCTION.3
```diff
@@ -87,8 +87,8 @@ All
 .nf
 static int sock_cb(CURL *e, curl_socket_t s, int what, void *cbp, void *sockp)
 {
-  GlobalInfo *g = (GlobalInfo*) cbp;
-  SockInfo *fdp = (SockInfo*) sockp;
+  GlobalInfo *g = (GlobalInfo *) cbp;
+  SockInfo *fdp = (SockInfo *) sockp;
```
emanuele6, line 91: I think those casts could just be removed since it's `void *` conversion.

docs/libcurl/curl_easy_unescape.3
```diff
@@ -39,7 +39,7 @@ that are URL encoded (%XX where XX is a two-digit hexadecimal number) are
 converted to their binary versions.
 
 If the \fBlength\fP argument is set to 0 (zero), \fIcurl_easy_unescape(3)\fP
-will use strlen() on the input \fIurl\fP string to find out the size.
+will use strlen() on the input \fBurl\fP string to find out the size.
```
bagder, line 42: The argument in the synopsis is called **input** now, not url! (renamed recently)

docs/libcurl/curl_unescape.3
```diff
@@ -28,7 +28,7 @@ curl_unescape - URL decodes the given string
 .nf
 #include <curl/curl.h>
 
-char *curl_unescape( const char *url, int length);
+char *curl_unescape(const char *url, int length);
```
bagder, line 31: I think this `url` should be renamed as well to `input`. Primarily to remove the risk that someone would think this operates on an actual URL.

docs/libcurl/curl_unescape.3
```diff
@@ -38,8 +38,8 @@ string" and return that as a new allocated string. All input characters that
 are URL encoded (%XX where XX is a two-digit hexadecimal number) will be
 converted to their plain text versions.
 
-If the 'length' argument is set to 0, \fIcurl_unescape(3)\fP will use strlen()
-on the input \fBurl\fP string to find out the size.
+If the \fBlength\fP argument is set to 0, \fIcurl_unescape(3)\fP will use
+strlen() on the input \fBurl\fP string to find out the size.
```
bagder, line 42: This this should be updated to match.

bagder (approved):

emanuele6:
Done; I also updated the "url" in `curl_escape(3)` to "string" to match `curl_easy_escape(3)`.

PR #11031, opened 2023-04-26
ngtcp2: use 0.15.0

bagder:
Adapt to new API calls

### DIFF
--- a/.github/workflows/ngtcp2-gnutls.yml
+++ b/.github/workflows/ngtcp2-gnutls.yml
@@ -101,7 +101,7 @@ jobs:
       name: 'install gnutls'
 
     - run: |
-        git clone --quiet --depth=1 -b v0.10.0 https://github.com/ngtcp2/nghttp3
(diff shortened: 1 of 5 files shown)


bagder:
We can't upgrade yet, because we need **nghttpx** and it is a part of nghttp2 and it still uses the deprecated ngtcp2 APIs  even in current git master so it fails to build.

bagder:
https://github.com/nghttp2/nghttp2/pull/1899

tatsuhiro-t:
nghttp2 v1.52 cannot be built with ngtcp2 v1.15.  Please wait for nghttp2 v1.53 to upgrade.

bagder:
@tatsuhiro-t any chance there is a nghttp2 release pending? We just ran into #11090 that seems to not happen on the latest ngtcp2 version, so we would like to take the jump...

tatsuhiro-t:
I will make new nghttp2 release tomorrow. I'm not sure the bug mentioned in #11090 has been fixed in the latest ngtcp2 release and main.  Will check.

pheiduck:
Is there a rebase planned here?
nghttp2 v1.53 released today.

PR #11039, opened 2023-04-27
libssh: tell it to use SFTP non-blocking

### LINKED ISSUE #11020: libssh: SFTP transfer: slow progress meter update
I've built curl  7.87.0 with libssh 0.10.4 for sftp support and I'm trying to download a 30 MiB file with a relatively slow bandwith of ~ 10 Mbit/s.

The update rate of the progress meter is then something like once every 4 seconds.
(To someone who doesn't know what's going on, the download appears to be stalling at first. No progress at all until ~ 6.5 MiB are transfered.)

It works fine when curl is built with libssh2. (Progress meter updates at least once every second.)

When looking into this I came to the conclusion, that the sftp transfer with libssh is done in blocking mode.
That means the do-while loop in readwrite_data() (lib/transfer.c)  runs until maxloops are exceeded. So depending on buffer size, there would be no updating on the progress meter for a long time.

Is there any specific reason why libssh is used in blocking mode for this?
Curl seems to prefer nonblocking mode in general.
I tried to fix this by adding `sftp_file_set_nonblocking(sshc->sftp_file)` after `sftp_open()` and as far as I can tell this works without a problem. (No thorough testing done yet, though.)

bagder:
Reported-by: Andreas Huebner
Fixes #11020

### DIFF
--- a/lib/vssh/libssh.c
+++ b/lib/vssh/libssh.c
@@ -1597,7 +1597,7 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
         MOVE_TO_SFTP_CLOSE_STATE();
         break;
       }
-
+      sftp_file_set_nonblocking(sshc->sftp_file);
       state(data, SSH_SFTP_DOWNLOAD_STAT);
       break;
 


bagder:
flaky CI day, but I think the PR is still fine

PR #11052, opened 2023-04-28
Do h2 flow window accounting for cancelled streams

icing:
- nghttp2 does not free connection level window flow for aborted streams
- when closing transfers, make sure that any buffered response data is "given back" to the flow control window
- add tests test_02_22 and test_02_23 to reproduce


bagder (approved):

bagder:
Thanks!

PR #11055, opened 2023-04-28
ci: set minimal permissions on workflow ngtcp2-quictls.yml

diogoteles08:
Hi! This PR brings a small security adjustment to limit the permissions of the workflow ngtcp2-quictls.yml, so that it follows the principle of least privilege. Exact same idea of the merged PR #9928

### DIFF
--- a/.github/workflows/ngtcp2-quictls.yml
+++ b/.github/workflows/ngtcp2-quictls.yml
@@ -42,11 +42,14 @@ concurrency:
   group: ngtcp2-openssl-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   autotools:
     name: ${{ matrix.build.name }}
     runs-on: 'ubuntu-latest'
     timeout-minutes: 60
+    permissions: read-all
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ngtcp2-quictls.yml
```diff
@@ -42,11 +42,14 @@ concurrency:
...
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   autotools:
     name: ${{ matrix.build.name }}
     runs-on: 'ubuntu-latest'
     timeout-minutes: 60
+    permissions: read-all
```
dfandrich, line 52: Is this one really necessary? None of the other workflows needs this, including ngtcp2-gnutls and ngtcp-wolfsshl.

diogoteles08: Good point, it's probably not. I just left it there because I was afraid of breaking some flow :grimacing: -- and read permissions should not bring any problem, since the repository is already public.
But I'll take it out

KarmaHaze (commented):
@KarmaHaze

dfandrich (approved):

dfandrich:
Thanks!

PR #11059, opened 2023-04-30
Fix CA certificate paths for AmiSSL and MorphOS

afalkenhahn:
AmiSSL stores certificates in `AmiSSL:Certs` and MorphOS stores them in `MOSSYS:Data/SSL/curl-ca-bundle.crt`. `config-amigaos.h` should respect that.


jay:
Have you tested this

afalkenhahn:
Of course.

jay:
Thanks

PR #11068, opened 2023-05-03
Disable quiche pacing while is not actually performed by curl

francoismichel:
Hi,

Just a one-line commit that disables pacing in the quiche library as currently curl does not
perform pacing when sending the QUIC packets.

When pacing is enabled, quiche assumes that packet are sent at the asked time,
see https://github.com/cloudflare/quiche/blob/6cef26b946695300b214cd837a92dc1774c0ba84/quiche/src/recovery/mod.rs#L376

If packets are sent earlier than the asked time, it can lead to wrong RTT estimations (undersestimations), as the RTT estimation will be based on the packet sent time. At its turn, this can have an impact on the hystart algorithm
afterwards that is currently enabled too, and quiche may exit the slow-start too early and underestimate the link bandwidth.

### DIFF
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -1258,6 +1258,7 @@ static CURLcode cf_connect_start(struct Curl_cfilter *cf,
     failf(data, "can't create quiche config");
     return CURLE_FAILED_INIT;
   }
+  quiche_config_enable_pacing(ctx->cfg, false);
   quiche_config_set_max_idle_timeout(ctx->cfg, QUIC_IDLE_TIMEOUT);
   quiche_config_set_initial_max_data(ctx->cfg, (1 * 1024 * 1024)
     /* (QUIC_MAX_STREAMS/2) * H3_STREAM_WINDOW_SIZE */);


bagder:
What would it take to do the reverse? I mean actually comply with the pacing?

francoismichel:
quiche stores the expected sending time of a packet in the `send_info` structure here:
https://github.com/curl/curl/blob/d8df0d6db7441b6e14920a7e16a10e32bdc9c7ae/lib/vquic/curl_quiche.c#L694-L694


In curl, the buffer containing the `send_info` seems regularly discarded and overwritten in a for loop in the call to `Curl_bufq_sipn` hereunder, so we lose this sending time information (the info of interest is stored in `readx.sendinfo`):

https://github.com/curl/curl/blob/d8df0d6db7441b6e14920a7e16a10e32bdc9c7ae/lib/vquic/curl_quiche.c#L739-L740


I don't know well all the internals of curl, but we would need a way to store and keep this `send_info`along with the packet. One way to do so would be to add a sending_time or similar information to be returned by `Curl_bufq_peek`, but this will probably imply quite a lot of code changes.
Another way would be to not queue quiche packets in the `bufq` structure while until their actual sending time, but this would require a new level of buffering in curl_quiche.c...

Maybe there are other simpler ways around this. I could have a look at it at some point but right now I am busy with a lot of urgent stuff. :-(

bagder (approved):
Seems sensible!

bagder:
Thanks!

Issue #10428, opened 2023-02-07
Hyperlink escape sequence causes bad output in Ubuntu 16 LTS terminal


jay:
### I did this

~~~
curl --head https://www.google.com
~~~

In Ubuntu 16 LTS terminal the location header has unrecognized escape sequences and duplicate URL.
~~~
location:]8;;https://www.google.com/\ https://www.google.com/
]8;;\content-type: text/html; charset=UTF-8
~~~

![Capture](https://user-images.githubusercontent.com/965580/217159919-ff897731-71a0-4a79-b837-e7a2562b4a67.PNG)

/cc @dfandrich

dfandrich:
What terminal emulator does this use? On all the ones I tried, the escape
sequence is either interpreted correctly or ignored and not displayed.

jay:
Gnome Terminal 3.18.3 + VTE 0.42.5

bagder:
@dfandrich is this only because of an older gnome terminal and we can just ask users to run a later version or is there anything more clever we can here?

dfandrich:
OSC 8 support was added to Gnome Terminal 6 years ago in ver. 3.25.1. There's a list tracking support in terminal emulators [on this page](/Alhadis/OSC8-Adoption/). The [OSC 8 spec](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#backward-compatibility) also says

>  Any terminal that correctly implements OSC parsing according to ECMA-48 is guaranteed not to suffer from compatibility issues. That is, even if explicit hyperlinks aren't supported, the target URI is silently ignored and the supposed-to-be-visible text is displayed, without artifacts.

It sounds like there may have been a bug in Gnome Terminal that didn't do this, but other terminal emulators, even those that don't support OSC 8, should ignore the extra data and that is what I found in my testing. That spec page lists four terminal emulators that still didn't work properly as of 3 years ago, but all of them are working (mostly) today in their latest versions.

I can't think how  we could easily and reliably detect whether a terminal emulator is buggy or not and dynamically enable this. xterm-style answerback strings might theoretically do it if those were still a thing, otherwise it would take linking with an X11 library  and even that wouldn't work all the time. My suggestion would be to just use --no-styled-output on those few cases where it doesn't work, or better yet, upgrade your terminal emulator since there aren't any known buggy implementations left according to that page.

jay:
[...]

PR #11071, opened 2023-05-03
tool_cb_hdr: Fix 'Location:' formatting for early VTE terminals

jay:
- Disable hyperlink formatting for the 'Location:' header value in VTE 0.48.1 and earlier, since it is buggy in those versions.

Prior to this change those terminals may show the location header value as gibberish or show it twice.

Ref: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#backward-compatibility

Fixes https://github.com/curl/curl/issues/10428
Closes #xxxx

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -344,6 +344,15 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
   char *copyloc = NULL, *locurl = NULL, *scheme = NULL, *finalurl = NULL;
   const char *loc = location;
   size_t llen = loclen;
+  char *vver = getenv("VTE_VERSION");
+
+  if(vver) {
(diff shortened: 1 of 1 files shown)


dfandrich (approved):
That's very convenient of VTE! This isn't completely reliable (e.g. when sshed into another host from a VTE terminal, under sudo, etc.), but it should work most of the time and the failure mode is pretty inconsequential. I suggest adding to the comment the year in which VTE was fixed so a code janitor coming through in 10 years can delete this. Also, I suggest treating an atoi value of 0 as supported so any future incompatible changes to VTE_VERSION that cause atoi to fail will fail open.

bagder (approved):

dfandrich (approved):

PR #11077, opened 2023-05-05
Corrected and Added Item in Simple Usage Section

nobedee:
Change Line Range - 21 to 27

PROPOSED CHANGE ONE:
****************************
Changed the description of an example in the "Simple Usage" section on line 21. The command original returned all matches for "curl" and did not return the definition.  

PROPOSED CHANGE TWO:
****************************
Added an example in the "Simple Usage" section that returns the definition of curl from a dictionary.


bagder (approved):

bagder:
Thanks!

PR #11106, opened 2023-05-12
url: fix null dispname for --connect-to option

smaeda-ks:
`parseurlandfillconn()` runs before `parse_connect_to_slist()` (which is where `conn->bits.conn_to_host` is set), so checking `conn->bits.conn_to_host` in `parseurlandfillconn()` would always be `FALSE`, and hence curl is always reporting `(nil)` dispname in output:

```sh
$ curl -svo /dev/null 'https://vercel.com/' --connect-to vercel.com:443:cname.vercel-dns.com 2>&1 | grep -i '* Connec'
* Connecting to hostname: cname.vercel-dns.com
* Connected to (nil) (76.76.21.142) port 443 (#0)
* Connection #0 to host (nil) left intact
```

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -1816,11 +1816,6 @@ static CURLcode parseurlandfillconn(struct Curl_easy *data,
   result = Curl_idnconvert_hostname(&conn->host);
   if(result)
     return result;
-  if(conn->bits.conn_to_host) {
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #11107, opened 2023-05-12
write-out.d: Use response_code in example

vitorgalvao:
The documentation for `response_code` in the manual states (emphasis mine):

> The numerical response code that was found in the last transfer (**formerly known as "http_code"**).

This suggests that `response_code` is preferred over `http_code`, yet the example uses the latter. Seeing as this code is likely to be one of the most common `write-out` options someone is looking for, using the newest/recommended name in the example should be more effective in getting people to use it.


bagder (approved):

bagder:
Thanks!

PR #11109, opened 2023-05-13
fix: Integration test

david4ilmor:
If something will be found during this test, I'll share with you the details ASAP.


bagder:
Please don't do this. It was not a fix of an integration test. It was you sending data, probably to a server of yours.

david4ilmor:
@bagder If any sensitive information was be fetched this way, the first thing I probably do is sending you the technical details using your security email. but nothing found here so I'm sorry for any inconvenience that may result from this test.

Have a great day

PR #11110, opened 2023-05-14
Add --with-openssl to GIT-INFO

osalbahr:
When I ran `./configure` I received the following error:

```
configure: error: select TLS backend(s) or disable TLS with --without-ssl.

Select from these:

  --with-amissl
  --with-bearssl
  --with-gnutls
  --with-mbedtls
  --with-nss
  --with-openssl (also works for BoringSSL and libressl)
  --with-rustls
  --with-schannel
  --with-secure-transport
  --with-wolfssl
```

I suggested `--with-openssl` according to https://curl.se/docs/install.html

### DIFF
--- a/GIT-INFO
+++ b/GIT-INFO
@@ -14,7 +14,7 @@ To build in environments that support configure, after having extracted
 everything from git, do this:
 
 autoreconf -fi
-./configure
+./configure --with-openssl
 make
 
   Daniel uses a ./configure line similar to this for easier development:


bagder (approved):
Well, you should select the TLS backend you prefer to use and we cannot assume that everyone wants OpenSSL, but I also figure that many readers of this command line that says "use OpenSSL" realize they might be able to select another one. So let's do this.

bagder:
Thanks!

osalbahr:
Thank you for explaining. Could you elaborate what you mean by "let's do this"?

PR #11113, opened 2023-05-15
drop linkcheck config file

pheiduck:
is no longer needed since links are always available over time.


bagder:
~~~
$ git grep mlc_config.json 
.reuse/dep5:Files: mlc_config.json
~~~
Remove it from there as well then I presume?

pheiduck:
Yep, will do a fixup.

bagder:
Thanks!

PR #11115, opened 2023-05-15
improved support for --limit-rate

icing:
The main change is to leave transfer loop when --limit-rate is in effect and that much data has been received. Mark the connection to simulate a POLLIN on next attempt, so buffered data will be picked up.

Then
- adjust HTTP/2 stream window size to --limit-rate plus some slack to make the server observe the pacing we want. I observer still an initial burst of incoming data, since we announce 512KB window on connection setup. But afterwards, the window is decreased and only the rate-limited amount is being sent when the buffers get drained.
- add h2/h3 test case to confirm behaviour on overall transfer duration

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -181,6 +181,7 @@ struct stream_ctx {
 
   int status_code; /* HTTP response status code */
   uint32_t error; /* stream error code */
+  uint32_t local_window_size; /* the local recv window size */
   bool closed; /* TRUE on stream close */
   bool reset;  /* TRUE on stream reset */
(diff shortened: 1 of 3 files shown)


bagder (approved):

bagder:
@icing: merge this now into 8.1.1 or wait until after you think?

icing:
My current self-evaluation says: WAIT! A couple of years at least!Am 18.05.2023 um 20:56 schrieb Daniel Stenberg ***@***.***>:﻿
@icing: merge this now into 8.1.1 or wait until after you think?

bagder:
Thanks!

PR #11126, opened 2023-05-17
Update `curl_easy_getinfo` doc

### LINKED ISSUE #11123: Missing info for the third arg in `curl_easy_getinfo` documentation
Hello there!
We are using libcurl in a C++ project, and following the [docs](https://curl.se/libcurl/c/curl_easy_getinfo.html) we used `curl_easy_getinfo` with only the specified types for the third argument: a pointer to `long`, `char *`, `curl_slist *` or `double`.
One of our users had a segmentation fault because of using `CURLINFO_SPEED_DOWNLOAD_T` with `curl_easy_getinfo` which requires a pointer to `curl_off_t`. The latter can be `long` or `long long` depending on the platform (cf. [PR](https://github.com/mamba-org/mamba/pull/2528)).
Do you confirm that the documentation should be updated to specify that the third argument can be a pointer to `curl_off_t` and `long long` as well? (I can submit a PR in that case if needed).
Thanks!

Hind-M:
Fix #11123

### DIFF
--- a/docs/libcurl/curl_easy_getinfo.3
+++ b/docs/libcurl/curl_easy_getinfo.3
@@ -33,13 +33,11 @@ CURLcode curl_easy_getinfo(CURL *curl, CURLINFO info, ... );
 .fi
 .SH DESCRIPTION
 Request internal information from the curl session with this function. The
-third argument \fBMUST\fP be a pointer to a long, a pointer to a curl_off_t, a
-pointer to a char *, a pointer to a double, a pointer to a struct curl_slist *,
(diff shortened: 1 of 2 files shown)


bagder (commented):
I think your edit here shows the insanity in listing all the types in this manner. The documentation as well as the header comment become unreadable wall of tests.

I think we should rather remove the types there and say that they can return types and that each option's type is documented in the corresponding man page for the option.

bagder (approved):

bagder:
Thanks!

PR #11130, opened 2023-05-17
Fix cmake error cross compiling ios, android

halx99:
refer to: https://github.com/axmolengine/buildware/actions/runs/5005991548/jobs/8970680345#step:3:15358


bagder (approved):
Makes sense!

PR #11131, opened 2023-05-17
urlapi: allow numerical parts in the host name

### LINKED ISSUE #11129: cURL 8.1.0 `URL using bad/illegal format or missing URL` for URL starting with digit
With latest release of cURL 8.1.0 now it impossible to use it for get/put requests with URLs starting with digits.

### I did this

curl http://10.xxx.xxx.xxx/composer/packages.json

### I expected the following

File was downloaded but got `URL using bad/illegal format or missing URL`

### curl/libcurl version

cURL 8.1.0

### operating system

Latest Alpine docker image

bagder:
It can only be an IPv4 address if all parts are all digits and no more than four parts, otherwise it is a host name. Even slightly wrong IPv4 will now be passed through as a host name.

Regression from 17a15d88467 shipped in 8.1.0

Extended test 1560 accordingly.

Reported-by: Pavel Kalyugin
Fixes #11129


bagder:
We're using invalid URLs in numerous test cases that this update detects and rejects!

dfandrich:
Won't this disallow a URL like http://www.37334/ ? My browser thinks that URL is fine when I have a matching entry in /etc/hosts. That ought to also work to reach the host www.37334.example.com if /etc/resolv.conf contains `search example.com`.

bagder:
Good point, thanks. I'll need to fix that.

bagder:
In fact, even `11.22.33.55.66` works as a name if provided in `/etc/hosts`.

bagder:
And so does `22.33.55.666`. I need to backpeddle my idea of verifying IPv4 addresses this strictly.

Issue #11138, opened 2023-05-18
[regression] Sending large header over http2 fails with a send error


emilio:
### I did this

https://github.com/curl/curl/commit/fc2f1e547a4a4b4bec5fd3c8bfde5136706488a1 introduced a hardcoded limit in max_line_length for http2 of 4kb, which is hit for requests with long headers.

Sending a request as the one in https://github.com/glandium/git-cinnabar/issues/314 fails now where it didn't fail before updating curl.

### I expected the following

The request is sent successfully.

### curl/libcurl version

```
curl 8.1.0 (x86_64-pc-linux-gnu) libcurl/8.1.0 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.10.0 nghttp2/1.53.0
Release-Date: 2023-05-17
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

```
Linux crimson 6.3.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 11 May 2023 16:40:42 +0000 x86_64 GNU/Linux
```

emilio:
Downgrading curl to:

```
curl 7.88.1-DEV (x86_64-pc-linux-gnu) libcurl/7.88.1-DEV OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.10.0 nghttp2/1.53.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

[...]

PR #11139, opened 2023-05-18
http2: Double http request parser max line length.

emilio:
This works around #11138, by doubling the limit, and should be a relatively safe fix.

Ideally the buffer would grow as needed and there would be no need for a limit? But that might be follow-up material.

Ref: #11138

### DIFF
--- a/lib/http1.h
+++ b/lib/http1.h
@@ -33,6 +33,8 @@
 #define H1_PARSE_OPT_NONE       (0)
 #define H1_PARSE_OPT_STRICT     (1 << 0)
 
+#define H1_PARSE_DEFAULT_MAX_LINE_LEN (8 * 1024)
+
 struct h1_req_parser {
   struct http_req *req;
   struct bufq scratch;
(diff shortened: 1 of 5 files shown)


emilio:
I don't think the appveyor etc failures are my fault but lmk if I missed something.

bagder:
> I don't think the appveyor etc failures are my fault but lmk if I missed something.

Don't be alarmed by the red CI build(s), that's just due to flaky/bad environments and not because of any flaw in your PR. Ignore them.

bagder:
> Ideally the buffer would grow as needed and there would be no need for a limit? But that might be follow-up material.

The buffer already grows to the needed size. We talk about the *maximum* size that the buffer is allowed to grow to here. *All* dynamic buffers in curl has a fixed maximum limit to avoid the risk of things going havoc due to mistakes or malice. (Some of them are of course using very large maximums.)

The question is what the maximum allowed size should be for this h2 header case.

bagder:
Thanks. Yes we can certainly discuss and work on improving this logic further. I merged this now to get the fix in first.

PR #11142, opened 2023-05-18
configure: fix --help indentation

emanuele6:
AC_ARG_ENABLE seems to only trim off whitespace from the start and end of its help-string argument, while prepending two spaces of indentation to all lines.

This means that the two spaces of indentation between the --enable-rtsp and the --disable-rtsp line were not removed causing ./configure --help to print:
```
Optional Features:
  [...]
  --enable-rtsp           Enable RTSP support
    --disable-rtsp          Disable RTSP support
```
I removed the indentation to fix the issue, now it prints:
```
Optional Features:
  [...]
  --enable-rtsp           Enable RTSP support
  --disable-rtsp          Disable RTSP support
```


bagder (approved):

bagder:
`--disable-hsts` seems to have the same problem, can you maybe address that one as well?

bagder (approved):

bagder:
Thanks!

PR #11143, opened 2023-05-18
select: avoid returning an error on EINTR from select() or poll()

### LINKED ISSUE #11135: `Curl_poll` may not ignore EINTR if 0 file descriptors are passed
I encountered this potential issue when trying to investigate a [user report](https://github.com/apache/arrow/issues/35318).

In the general case, `Curl_poll` will translate EINTR into a no-error return so that the caller may retry:
https://github.com/curl/curl/blob/a9f8fe28481fef7c28d85b4a12a3a35521408eaf/lib/select.c#L312-L318

However, if no valid fds are passed, `Curl_poll` delegates to `Curl_wait_ms` which doesn't try to eliminate EINTR errors:
https://github.com/curl/curl/blob/a9f8fe28481fef7c28d85b4a12a3a35521408eaf/lib/select.c#L94-L104

`Curl_poll` is used from `Curl_multi_wait` where it's not obvious whether a non-zero number of file descriptors is always [found](https://github.com/curl/curl/blob/a9f8fe28481fef7c28d85b4a12a3a35521408eaf/lib/multi.c#L1185-L1207).

This might be a red herring, but I suppose it wouldn't hurt fixing `Curl_wait_ms` (and/or `Curl_poll`) in any case.

pitrou:
This was already done for the poll() and select() calls
made directly from Curl_poll(), but was missed in
Curl_wait_ms(), which is called when there are no fds
to wait on.

Fixes #11135

### DIFF
--- a/lib/select.c
+++ b/lib/select.c
@@ -61,8 +61,8 @@
  * for the intended use of this function in the library.
  *
  * Return values:
- *   -1 = system call error, invalid timeout value, or interrupted
- *    0 = specified timeout has elapsed
+ *   -1 = system call error, or invalid timeout value
+ *    0 = specified timeout has elapsed, or interrupted
  */
 int Curl_wait_ms(timediff_t timeout_ms)
 {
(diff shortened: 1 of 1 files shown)


bagder (approved):

pitrou:
There are a couple CI failures but I'm sketical they are related. Windows system/socket routines can't return EINTR (Windows has no signals per se).

bagder:
Thanks!

PR #11144, opened 2023-05-18
cmake: avoid `list(PREPEND)` for compatibility

### LINKED ISSUE #11141: cmake error -  list does not recognize sub-command PREPEND
8.1.0 introdused new problem in cmake - at configure i get:
```
CMake Error at CMakeLists.txt:528 (list):
  list does not recognize sub-command PREPEND
```
culprit is `list(PREPEND CURL_LIBS ${BROTLI_LIBRARIES})` line added in [ cmake: picky-linker fixes for openssl, ZLIB, H3 and more #10857 ](https://github.com/curl/curl/pull/10857/commits/fbf4fa9162ed4663d61184da7458645ded82130d)
it not supported in cmake up to  3.10.1 and  3.12.4 (max I can use) while declared version is 3.7:
`cmake_minimum_required(VERSION 3.7...3.16 FATAL_ERROR)`
(my C compiler identification is GNU 7.2.0)

vszakats:
`list(PREPEND)` requires CMake v3.15, our minimum is v3.7.

Ref: https://cmake.org/cmake/help/latest/command/list.html#prepend

Regression from 1e3319a167d2f32d295603167486e9e88af9bb4e

Reported-by: Keitagit-kun on Github
Fixes #11141
Closes #11144


bagder (approved):
How about adding a comment about why  this does not use PREPEND to maybe save us a from doing the mistake again?

vszakats:
Added one. Though, the only sure cure is to lookup newly used CMake keywords in the manual. Which is trivial to miss sometimes.

Another option is to delete this line. I added it to help `ld` finding existing symbols, but when retested now, it was broken elsewhere. Had to restore the hack that fixes it globally: https://github.com/curl/curl-for-win/commit/8f38f1c34ca290d51e32e06b8fa161623c8d4ae0. This is apparently impossible to fix on this level. CMake, or even better `ld` could fix it. It's an ancient issue.

bagder:
> Another option is to delete this line

I have no opinion on this. I leave it up to you on decide the preferable route.

vszakats:
Let's keep this for now. I might try to address the newly popped-up issues in the future.

PR #11150, opened 2023-05-19
docs: document that curl_url_cleanup(NULL) is a safe no-op 

emanuele6:
This has always been the case, but it was not documented.

The paragraph was copied verbatim from curl_easy_cleanup.3


bagder (approved):

bagder:
Thanks!

Issue #11147, opened 2023-05-18
warning: -Wnull-dereference with ngtcp2 + gcc 13.1.0


vszakats:
### I did this

Built curl with gcc 13.1.0 and got this warning:
```
[...]

PR #11151, opened 2023-05-19
Fixing compiler warning about possible null-deref

icing:
- compiler analyzer did not include the call context for this static function where the condition had already been checked.
- eleminating the problem by making stream a call parameter
- refs #11147


bagder (approved):

bagder:
Thanks!

PR #11154, opened 2023-05-19
Remove "live" mod-h2 master from CI builds

icing:
- this will skip test_02_20_h2_small_frames until the CI has an Apache that supports the option

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -289,13 +289,7 @@ jobs:
       run: |
         sudo apt-get install apache2 apache2-dev libnghttp2-dev
         sudo python3 -m pip install -r tests/http/requirements.txt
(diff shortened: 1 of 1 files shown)


vszakats:
It may also be a stable solution to use a specific commit hash and build that, allowing to keep this test. That commit hash could be updated like any other version/tag we depend on in the CI.

bagder:
I forgot about this PR. Do we still need it? I don't see Apache related CI problems much?

icing:
I'll close if for now.

dfandrich:
No idea if it's related, but the pytests are failing quite frequently recently on the ngtcp2-gnutls build on GHA.

icing:
> No idea if it's related, but the pytests are failing quite frequently recently on the ngtcp2-gnutls build on GHA.

I think I found the cause of that in the just merged #11368.

PR #11159, opened 2023-05-19
sectransp.c: make the code c89 compatible

### LINKED ISSUE #11155: curl fails to compile with gcc, `--enable-debug=yes` and --with-secure-transport
### I did this

I cloned the repository and ran the following commands:

- `autoreconf --force --install`
- `./configure --disable-shared --enable-static --prefix=/ '--bindir=${prefix}/bin' '--sbindir=${prefix}/bin' '--libdir=${prefix}/lib' '--includedir=${prefix}/include' '--oldincludedir=${prefix}/include' --with-libidn2=no --with-librtmp=no --with-libpsl=no --with-libgsasl=no --with-schannel=no --with-secure-transport=yes --with-brotli=no --enable-shared=no --enable-static=yes --enable-dict=yes --enable-file=yes --enable-ftp=yes --enable-gopher=yes --enable-http=yes --enable-imap=yes --enable-ldap=no --enable-mqtt=yes --enable-pop3=yes --enable-rtsp=yes --enable-smb=yes --enable-smtp=yes --enable-telnet=yes --enable-tftp=yes --enable-debug=no --enable-ares=no --enable-threaded-resolver=yes --enable-cookies=yes --enable-ipv6=yes --enable-manual=no --enable-verbose=yes --enable-symbol-hiding=no --enable-unix-sockets=yes --with-zstd=no --without-openssl --without-wolfssl --without-libssh2 --without-nghttp2 --with-zlib=/Users/feignclaims/.conan2/p/b/zlib5dc1ed3312917/p --with-ca-fallback=nO`
- `make`

However, when running `make`, I got the following error:

```
vtls/sectransp.c: In function 'TLSCipherNameForNumber':
vtls/sectransp.c:907:3: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
  907 |   for(size_t i = SSL_FORTEZZA_DMS_WITH_FORTEZZA_CBC_SHA + 1;
      |   ^~~
vtls/sectransp.c:907:3: note: use option '-std=c99', '-std=gnu99', '-std=c11' or '-std=gnu11' to compile your code
```

I set `-enable-debug=no` and reconfigured it, keeping the other parameters the same, and it compiled successfully.

The issue might be caused by the `-std=gnu89` in CFLAGS, which is added when `want_warnings` is `yes` and the `compiler_id` is `GNU_C`.

### I expected the following

A successful compilation.

### curl/libcurl version

git master branch (with latest commits a64d7de61af5fb94b70972cd0981020e729c1499)

### operating system

Darwin FeignClaimsdeMacBook-Air.local 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 21:00:41 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8103 arm64

bagder:
Follow-up to dd2bb485521c2ec713001b3a

Reported-by: FeignClaims on github
Fixes #11155

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -83,6 +83,12 @@ jobs:
           install: nghttp2
           configure: --enable-debug --with-secure-transport
           macosx-version-min: 10.8
+        - name: gcc SecureTransport http2
+          install: nghttp2
+          CC: gcc-12
+          CXX: g++-12
+          configure: --enable-debug --with-secure-transport
+          macosx-version-min: 10.8
         - name: OpenSSL http2
           install: nghttp2 openssl
           configure: --enable-debug --with-openssl=/usr/local/opt/openssl
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -83,10 +83,9 @@ jobs:
           install: nghttp2
           configure: --enable-debug --with-secure-transport
           macosx-version-min: 10.8
-        - name: gcc SecureTransport http2
-          install: nghttp2
+        - name: gcc SecureTransport
           configure: CC=gcc-12 --enable-debug --with-secure-transport
-          macosx-version-min: 10.13
+          macosx-version-min: 10.8
         - name: OpenSSL http2
           install: nghttp2 openssl
           configure: --enable-debug --with-openssl=/usr/local/opt/openssl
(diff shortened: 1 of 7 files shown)


bagder:
I remember @nickzman saying something about some limitations using gcc + secure transport in the past. Can it affect the use of ALPN to enable h2 for example?

nickzman:
The limitation is with weak-linking symbols that may not be available in old OS versions. There are two ways to do weak-linking on Apple operating systems. First, there's the old-school way:
```
if(ThisOSSymbolFromMonterey != NULL) { }
```
This is deprecated, but it's supported by all compilers. The other way is with availability:
```
if(__builtin_available(iOS 15.0, macOS 12.0, *) { }
```
clang supports `__builtin_available`, but AFAIK, GCC & other compilers do not.

PR #11164, opened 2023-05-20
lib: rename struct 'http_req' to 'httpreq'

### LINKED ISSUE #11163: FreeBSD 14 "owns" struct http_req
(Found here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271526, reported by "O. Hartmann")

> Just updating ports: curl 8.0.1 -> 8.1

~~~

[...]
In file included from cf-socket.c:56:
In file included from ../lib/urldata.h:168:
../lib/http.h:263:8: error: redefinition of 'http_req'
struct http_req {
       ^
/usr/include/netinet/tcp.h:466:8: note: previous definition is here
struct http_req {
~~~

I would say this is a bug in their `netinet/tcp.h` header file, but I doubt that helps. We should rename our struct.

bagder:
Because FreeBSD 14 kidnapped the name.

Fixes #11163

### DIFF
--- a/lib/cf-h2-proxy.c
+++ b/lib/cf-h2-proxy.c
@@ -777,7 +777,7 @@ static CURLcode h2_submit(int32_t *pstream_id,
                           struct Curl_cfilter *cf,
                           struct Curl_easy *data,
                           nghttp2_session *h2,
-                          struct http_req *req,
+                          struct httpreq *req,
                           const nghttp2_priority_spec *pri_spec,
                           void *stream_user_data,
                           nghttp2_data_source_read_callback read_callback,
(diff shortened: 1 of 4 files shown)


emanuele6:
Wouldn't it be better to rename it to `curl_http_req` or some other name prefixed by `curl_`/`CURL`/`Curl_`?

bagder:
Maybe. It's not really practical to do *all* private structs using a `Curl_` prefix just *in case* some systems go bananas. I generally like shorter names because it makes them easier on the eye. We still have a huge amount of private structs that to not use `Curl_` prefixes and may of those have worked fine in many surroundings for a very long time.

In this specific case, it looks like FreeBSD did a mistake and likely will fix it so this would mostly be a change for the short term to make sure the name works on FreeBSD 14.

Issue #11157, opened 2023-05-19
cURL 8.1.0 POST fails, downgrading to 8.0.1 works as expected


user8446:
Hi,

After upgrading to 8.1.0, a simple POST will timeout. The URL does not start with numbers as [others have found](https://github.com/curl/curl/issues/11129).

API Call:

```
curl --request POST --max-time 60 --retry 4 --retry-delay 6 --progress-bar --output nul
    --url "https://api.trello.com/....."

curl: (28) Operation timed out after 60005 milliseconds with 0 bytes received
Warning: Problem : timeout. Will retry in 6 seconds. 4 retries left.
curl: (28) Operation timed out after 60006 milliseconds with 0 bytes received
Warning: Problem : timeout. Will retry in 6 seconds. 3 retries left.
curl: (28) Operation timed out after 60006 milliseconds with 0 bytes received
Warning: Problem : timeout. Will retry in 6 seconds. 2 retries left.
curl: (28) Operation timed out after 60002 milliseconds with 0 bytes received
Warning: Problem : timeout. Will retry in 6 seconds. 1 retries left.
curl: (28) Operation timed out after 60007 milliseconds with 0 bytes received
```

OS is Windows10

```
curl 8.1.0 (x86_64-w64-mingw32) libcurl/8.1.0 OpenSSL/3.1.0 (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.5 WinIDN libssh2/1.10.0 nghttp2/1.53.0 ngtcp2/0.15.0 nghttp3/0.11.0 libgsasl/2.2.0
Release-Date: 2023-05-17
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe UnixSockets zstd
```

Simply downgrading back to 8.0.1 resolves.

```
[...]

PR #11165, opened 2023-05-20
http/2, upload handling improvements

icing:
Make send buffer smaller to have progress and "upload done" reporting closer to reality. Fix handling of send "drain" condition to no longer trigger once the transfer loop reports it is done sending. Also do not trigger the send "drain" on RST streams.

Background:
- a upload stall was reported in #11157 that timed out
- test_07_33a reproduces a problem with such a stall if the server 404s the request and RSTs the stream.
- test_07_33b verifies a successful PUT, using the parameters from #11157 and checks success


bagder (approved):

bagder:
Thanks!

PR #11166, opened 2023-05-20
tests/servers: generate temp names in /tmp for unix domain sockets

### LINKED ISSUE #11152: curl 8.1.0: tests 1467 1468 failing with 'socket path too long'
When running the testsuite after updating from curl 8.0.1 to 8.1.0, tests 1467 and 1468 are failing, with this message on stderr:

```
curl: (6) Unix socket path too long: '/data/omnios-build/omniosorg/bloody/_build/curl-8.1.0/curl-8.1.0-build/build.i386/tests/log/server/socks.sock'
```

On my build system, that path is indeed > 107 characters (the size of `sockaddr_un.sun_path` being 108).

This seems to be a result of https://github.com/curl/curl/commit/89ede3953b8b64d8bf0d15d10989d003d2b30a12

```
CMD (0): ../src/curl --version 1>log/curlverout.log 2>log/curlvererr.log
********* System characteristics ********
* curl 8.1.0 (x86_64-pc-solaris2.11)
* libcurl/8.1.0 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.5 nghttp2/1.52.0
* Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets zstd
* Disabled: xattr
* Host: bloody
* System: SunOS bloody 5.11 omnios-master-42fe848aa52 i86pc i386 i86pc
* OS: solaris
* Env:
* Seed: 214771
* Servers: HTTP-IPv6 HTTP-unix FTP-IPv6
*****************************************
DISABLED: test 323
DISABLED: test 594
DISABLED: test 836
DISABLED: test 882
DISABLED: test 938
DISABLED: test 1182
DISABLED: test 1184
DISABLED: test 1209
DISABLED: test 1211
DISABLED: test 1307
DISABLED: test 1316
DISABLED: test 1510
DISABLED: test 1512
DISABLED: test 1801
DISABLED: test 2086
DISABLED: test 2043
startnew: perl -I../../../src/tests ../../../src/tests/http-server.pl --keepalive 30 --pidfile "log/server/http_server.pid" --logfile "log/http_server.log" --logdir "log" --portfile log/server/http_server.port --config log/ftpserver.cmd --ipv4 --port 0 --srcdir "../../../src/tests"
RUN: ../src/curl --max-time 13 --output log/http_verify.out --silent --verbose --globoff "http://127.0.0.1:38366/verifiedserver" 2>log/http_verify.log
CMD (0): ../src/curl --max-time 13 --output log/http_verify.out --silent --verbose --globoff "http://127.0.0.1:38366/verifiedserver" 2>log/http_verify.log
RUN: HTTP server is on PID 149 port 38366
* pid http => 149 149
startnew: server/socksd --pidfile log/server/socks2_server.pid --reqfile log/socksd-request.log --logfile log/socks2_server.log --unix-socket log/server/socks.sock --backend 127.0.0.1 --config log/ftpserver.cmd
RUN: SOCKS2 server is now running PID 158
* pid socks5unix => 158 158
test 1467...[HTTP GET via SOCKS5 proxy via unix sockets]
../src/curl --output log/curl1467.out  --include --trace-ascii log/trace1467 --trace-time --socks5 localhostlog/server/socks.sock http://127.0.0.1:38366/1467 > log/stdout1467 2> log/stderr1467
CMD (1536): ../src/curl --output log/curl1467.out  --include --trace-ascii log/trace1467 --trace-time --socks5 localhostlog/server/socks.sock http://127.0.0.1:38366/1467 > log/stdout1467 2> log/stderr1467

 1467: protocol FAILED!
 There was no content at all in the file log/server.input.
 Server glitch? Total curl failure? Returned: 6

 - abort tests
* kill pid for socks5unix => 158
* kill pid for http => 149
RUN: Process with pid 149 signalled to die
RUN: Process with pid 158 signalled to die

RUN: Process with pid 149 forced to die with SIGKILL
RUN: Process with pid 158 forced to die with SIGKILL
TESTDONE: 1 tests were considered during 2 seconds.

TESTFAIL: These test cases failed: 1467
```

```
bloody:omnios:% uname -a
SunOS bloody 5.11 omnios-master-42fe848aa52 i86pc i386 i86pc

bloody:omnios:% rg -A3 sockaddr_un /usr/include/sys/un.h
53:struct     sockaddr_un {
54-   sa_family_t     sun_family;             /* AF_UNIX */
55-   char            sun_path[108];          /* path name (gag) */
56-};
```

bagder:
... instead of putting them in the regular pid directories. Simply because system generally have strict length requirements on that path name to be less than 107 bytes and we easily hit that boundary otherwise.

The new concept generates two random names: one for the socks daemon and one for http.

Reported-by: Andy Fiddaman
Fixes #11152

### DIFF
--- a/tests/data/test1467
+++ b/tests/data/test1467
@@ -42,7 +42,7 @@ socks5unix
 HTTP GET via SOCKS5 proxy via unix sockets
  </name>
  <command>
---socks5 localhost%PWD/%SOCKSUNIXPATH http://%HOSTIP:%HTTPPORT/%TESTNUMBER
+--socks5 localhost%SOCKSUNIXPATH http://%HOSTIP:%HTTPPORT/%TESTNUMBER
 </command>
 </client>
 
--- a/tests/data/test1468
+++ b/tests/data/test1468
@@ -43,7 +43,7 @@ socks5unix
 HTTP GET with host name using SOCKS5h via unix sockets
  </name>
  <command>
-http://this.is.a.host.name:%HTTPPORT/%TESTNUMBER --proxy socks5h://localhost%PWD/%SOCKSUNIXPATH
+http://this.is.a.host.name:%HTTPPORT/%TESTNUMBER --proxy socks5h://localhost%SOCKSUNIXPATH
 </command>
 </client>
 
--- a/tests/data/test1470
+++ b/tests/data/test1470
@@ -45,7 +45,7 @@ socks5unix
 HTTPS GET with host name using SOCKS5h via unix sockets
  </name>
  <command>
-https://this.is.a.host.name:%HTTPSPORT/%TESTNUMBER -k --proxy socks5h://localhost%PWD/%SOCKSUNIXPATH
+https://this.is.a.host.name:%HTTPSPORT/%TESTNUMBER -k --proxy socks5h://localhost%SOCKSUNIXPATH
 </command>
 </client>
 
--- a/tests/servers.pm
+++ b/tests/servers.pm
@@ -180,11 +180,19 @@ sub getfreeport {
     return $server->sockport();
 }
 
+use File::Temp qw/ tempfile/;
+
 #######################################################################
 # Initialize configuration variables
 sub initserverconfig {
-    $SOCKSUNIXPATH = "$LOGDIR/$PIDDIR/socks.sock"; # SOCKS Unix domain socket
-    $HTTPUNIXPATH = "$LOGDIR/$PIDDIR/http.sock";   # HTTP Unix domain socket
+    my ($fh, $socks) = tempfile("/tmp/curl-socksd-XXXXXXXX");
+    close($fh);
+    unlink($socks);
+    my ($f2, $http) = tempfile("/tmp/curl-http-XXXXXXXX");
+    close($f2);
+    unlink($http);
+    $SOCKSUNIXPATH = $socks; # SOCKS Unix domain socket
+    $HTTPUNIXPATH = $http;   # HTTP Unix domain socket
     $stunnel = checkcmd("stunnel4") || checkcmd("tstunnel") || checkcmd("stunnel");
 
     # get the name of the current user


dfandrich (commented):
My biggest concern is that by putting the files into a public location, we suddenly need to be more concerned about security. The permissions suddenly become an issue. Also, the code as-is opens up a race condition in the creation of the socket file that might, at least theoretically, cause a security problem since an attacker could create his own socket between the time of unlink and bind. Even if file name were chosen randomly but not temporarily created by runtests, an attacker could still see the name passed on the command-line to socksd and create it before socksd gets a chance. Maybe this isn't actually a problem if socksd aborts if the socket inode already exists when it start.

My other objection is that these two sockets become special cases compared to the other test server files, but this is minor.

The file names should probably also have "curl' in the name so that it's more clear who's at fault if somebody's /tmp is suddenly littered with dozens of them.

bagder:
> My biggest concern is that by putting the files into a public location, we suddenly need to be more concerned about security

It is certainly not ideal but can we circumvent this issue in any other way than putting these files in `/tmp`? I mean another way is to just disable the tests if we think the path is too long for a particular system (we could add a check) - but I don't know if that is a better way. And I can't think of a better functional way to shorten the path than putting them in /tmp ...

Security for our test servers shouldn't be such a big deal, should it?

thesamesam:
Could you accept a `TMPDIR` env-var override? It'll then default to /tmp but it lets us override it in packaging.

bagder:
> Could you accept a `TMPDIR` env-var override? It'll then default to /tmp but it lets us override it in packaging.
> 
> EDIT: Looks like perl's `File::Temp` respects it already, so if we use that...

Good idea, and yes we use that for this. So we *should* be good for that!

PR #11169, opened 2023-05-21
Revert "urlapi: respect CURLU_ALLOW_SPACE and CURLU_NO_AUTHORITY for …

emanuele6:
…redirects"

This reverts commit df6c2f7b544f1f35f2a3e0be11f345affeb6fe9c. (It only keep the test case that checks redirection to an absolute URL without hostname and CURLU_NO_AUTHORITY).

I originally wanted to make CURLU_ALLOW_SPACE accept spaces in the hostname only because I thought
curl_url_set(CURLUPART_URL, CURLU_ALLOW_SPACE) was already accepting them, and they were only not being accepted in the hostname when curl_url_set(CURLUPART_URL) was used for a redirection.

That is not actually the case, urlapi never accepted hostnames with spaces, and a hostname with a space in it never makes sense. I probably misread the output of my original test when I they were normally accepted when using CURLU_ALLOW_SPACE, and not redirecting.

Some other URL parsers seems to allow space in the host part of the URL, e.g. both python3's urllib.parse module, and Chromium's javascript URL object allow spaces (chromium percent escapes the spaces with %20), (they also both ignore TABs, and other whitespace characters), but those URLs with spaces in the hostname are useless, neither python3's requests module nor Chromium's window.location can actually use them.

There is no reason to add support for URLs with spaces in the host, since it was not a inconsistency bug; let's revert that patch before it makes into release. Sorry about that.

I also reverted the extra check for CURLU_NO_AUTHORITY since that does not seem to be necessary, CURLU_NO_AUTHORITY already worked for redirects.


bagder (approved):

bagder:
Thanks!

PR #11170, opened 2023-05-21
New Protocol: Gemini

jecxjo:


docs/libcurl/symbols-in-versions
```diff
@@ -925,6 +925,7 @@ CURLPROTO_DICT                  7.19.4
 CURLPROTO_FILE                  7.19.4
 CURLPROTO_FTP                   7.19.4
 CURLPROTO_FTPS                  7.19.4
+CURLPROTO_GEMINI                8.0.2
```
bagder, line 928: It cannot be merged earlier than 8.2.0

include/curl/curl.h
```diff
@@ -1067,6 +1067,7 @@ typedef CURLSTScode (*curl_hstswrite_callback)(CURL *easy,
 #define CURLPROTO_SMBS   (1<<27)
 #define CURLPROTO_MQTT   (1<<28)
 #define CURLPROTO_GOPHERS (1<<29)
+#define CURLPROTO_GEMINI (1<<30)
```
bagder, line 1070: No can do. This set of defines is already full. To add a new, add it privately in `lib/urldata.h` similar to how we add WS and WSS there.

I presume there is only ever `gemini` and no different one for TLS vs non-TLS?

lib/gemini.c
```diff
@@ -0,0 +1,296 @@
...
+
+  /* url is guaranteed to be non-NULL and non-empty */
+  DEBUGASSERT(url);
+
+  /* gemini is guaranteed to be non-NULL */
+  DEBUGASSERT(gemini);
+
+  /* url must not start with 0xFEFF */
+  if((unsigned char)url[0] == 0xFF && (unsigned char)url[1] == 0xFE) {
+    failf(data, "URL starts with 0xFEFF");
```
bagder, line 70: This looks very specific. Shouldn't the URL be rejected already in the URL parser if this is an invalid UR? Why would a URL start with this byte combo and why is it so explicitly bad? Where is the spec for the gemini URL format?

lib/gemini.c
```diff
@@ -0,0 +1,296 @@
...
+    return CURLE_URL_MALFORMAT;
+  }
+
+  /* url must be 1024 bytes or less */
+  if(strlen(url) > 1024) {
+    failf(data, "URL is too long");
```
bagder, line 76: How can gemini as a protocol accept URL as a general concept and yet refuse to work on URLs longer than 1024 bytes? Shouldn't the protocol rather reject too long file names or something? Where is this restriction specified?

lib/gemini.c
```diff
@@ -0,0 +1,296 @@
...
+      return result;
+
+    gemini->reqindex += data->info.request_size;
+
+    /* If we haven't sent the entire request yet, return */
+    if(gemini->reqindex < gemini->reqlen)
+      return CURLE_OK;
+  }
+
+  /* Receive response header */
```
bagder, line 136: in the `doing` function, libcurl should be sending the request, it is not for receiving content. That's the PERFORM state.

bagder (commented):
We need (lots of) tests for this. Presumably best done by creating a gemini test server in the style of sws.c that we can control fine from the test suite and do both positive and negative testing with. We need documentation for how to use gemini with curl. And for the implementation, we need links and details to specs. For the protocol and for the URL format.

jecxjo:
Sounds good. I'll look into the standards you have set for your tests and get them in soon. Just wanted to revive getting Gemini support in the library.

lib/gemini.c
nytpu: It is a specific requirement that a response should not begin with a Unicode byte-order mark—uncommon but allowed for UTF-8, and IIRC very early-on it was a bug in a few Windows clients using a UTF-8 encoder with the default configuration. I think it *should* be caught as in valid by the URL parser but I suppose @jecxjo wanted to make it explicit. 

> <URL> is a UTF-8 encoded absolute URL, including a scheme, of maximum length 1024 bytes. The request MUST NOT begin with a U+FEFF byte order mark.
— [Gemini specification](https://gemini.circumlunar.space/docs/specification.gmi) § 2

bagder: A URL (as far as curl is concerned) starts with a scheme. A scheme can only be, in the words of RFC 3986:
~~~
scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

lib/gemini.c
jecxjo: The specification directly calls out this restriction (https://gemini.circumlunar.space/docs/specification.html Section 2: Gemini Request). I'm moving the check to the URL parsing and calling the reference to the spec.

bagder: I have since read the spec and I found that limit specified, thanks. I do have some additional questions on the UTF-8 encoding of the URL, because the limit is specified for the UTF-8 version and it's not at all clear to me what that means.

If there needs to be some encoding done, then the URL parser cannot enforce the limit since such an encoding process might change the length.

lib/gemini.c
```diff
@@ -0,0 +1,296 @@
...
+  if(gemini->reslen < rx_size) {
+
+    /* Not ready to read */
+    if(SOCKET_READABLE(sockfd, 0) <= 0)
+      return CURLE_OK;
+
+    result = Curl_read(data,
```
dimkr, line 143: Some Gemini servers still don't send close_notify although the spec says this is mandatory - maybe this is worth considering here.

kensanata: It's true. 😭 
https://github.com/mojolicious/mojo/pull/1875

bagder:
Is anyone working on taking this effort further? If not, it will be subject for closure soon as there's no point keeping around a half-baked PR with no activity.

ERnsTL:
@jecxjo Hi are you still there, would you like to continue with the Gemini pull request?

jecxjo:
Sorry been trying to create a similar test server like the one used for http but haven't had much time. Knowing what is needed I'm fine with closing the PR and once i get a test server working i can create a new PR.

bagder:
Thank you. I'm closing this then, but don't hesitate to create a new one when you feel read to take the next step.

zzo38:
About TOFU, I think that you can instead treat it like other TLS and use the `-k` switch if you want to bypass the security check like you do with any other protocols.

About charater encoding, I think that part about UTF-8 should just be ignored completely in the case of curl. It should just transmit the URL as-is, even if it contains invalid Unicode.

About close_notify, if it is believed to be necessary, a switch could be added to control this behaviour.

PR #11174, opened 2023-05-22
remove unused functions, make single-use functions static

bagder:


bagder:
@icing `Curl_cf_def_close` is a function that is *only* unit-tested and never used in libcurl itself. Isn't that odd?

icing:
> @icing `Curl_cf_def_close` is a function that is _only_ unit-tested and never used in libcurl itself. Isn't that odd?

If every filter we have adds its own method, it is an offering never used, indeed.

bagder:
> If every filter we have adds its own method, it is an offering never used, indeed.

I moved it into an `#ifdef DEBUGBUILD` to make the unit test remain functional without adding unused code to non-debug builds.

Issue #11175, opened 2023-05-22
Since 8.1.0 php-curl POST fails with HTTP/2 stream 1 was reset for {url}


mariandomanik:
We are using Sentry Laravel SDK https://docs.sentry.io/platforms/php/guides/laravel/
Since curl 8.1.0 both on prod and local, we were unable to connect to Sentry servers, always resulting with

`HTTP/2 stream 1 was reset for {our-sentry-dsn}`

This was called from PHP SDK and we were trying to follow the HTTP Exception, but this was the only message we were able to gather, with no HTTP error code and no additional information.

Downgrading to curl 7.86 fixed this issue.

Fails on both MacOSX 13.4 and 
5.10.178-162.673.amzn2.x86_64 #1 SMP Mon Apr 24 23:34:06 UTC 2023 x86_64 Linux

bagder:
This is next to impossible for us to act on. Can you please elaborate on how we can reproduce or you need to provide a lot more information about what exactly is happening!

bagder:
Where does the `...was reset for...` message come from? I can't seem to find that in libcurl 8.1.0.

stayallive:
Hello, maintainer from the Sentry SDK team here, we are already looking into this and we can confirm this happens on PHP with cURL 8.1.0, however using the cURL CLI directly seems to work fine so I'm inclined to lean that this is something on the PHP side or some other factor.

The CLI test I ran:

```bash
docker run --rm -ti curlimages/curl:8.1.0 \
  -X POST https://o447951.ingest.sentry.io/api/5572016/store/ \
  -H "Content-Type: application/json" \
  -H "X-Sentry-Auth: Sentry sentry_key=49bf4ae0f80a4871abb813507b2d7cae" \
  -d '{}'
```

Sentry PHP SDK issue: getsentry/sentry-php#1537.

icing:
My question would be if you see a RST on the stream do to a timeout on the server side? Or does the server reset immediately because it does not like what it sees.

In case of a timeout, please see #11176 where I fixed an issue that led to stalled curl uploads.

vszakats:
I've made fresh builds here:
https://ci.appveyor.com/project/curlorg/curl-for-win/builds/47111484/artifacts

I'll see if this can be made automated with reasonable effort.

bagder:
Please try 8.1.1 and see if the problem remains!

S3B-4:
[...]

PR #11176, opened 2023-05-22
http/2, unstuck uploads

icing:
- refs #11157 and #11175 where uploads get stuck or lead to RST streams
- fixes our h2 send behaviour to continue sending in the nghttp2 session as long as it wants to. This will empty our send buffer as long as the remote stream/connection window allows.
- in case the window is exhausted, the data remaining in the send buffer will wait for a WINDOW_UPDATE from the server. Which is a socket event that engages our transfer loop again
- the problem in the issue was that we did not exhaust the window, but left data in the sendbuffer and no further socket events did happen. The server was just waiting for us to send more.
- relatedly, there was an issue fixed that closing a stream with KEEP_HOLD set kept the transfer from shutting down - as it should have - leading to a timeout.

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1200,6 +1200,7 @@ static int on_stream_close(nghttp2_session *session, int32_t stream_id,
   stream->error = error_code;
   if(stream->error)
     stream->reset = TRUE;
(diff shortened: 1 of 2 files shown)


bagder:
Thanks!

PR #11182, opened 2023-05-23
configure: fix build with arbitrary CC and LD_LIBRARY_PATH

emanuele6:
Since ./configure and processes that inherit its environment variables are the only callers of the run-compiler script, we can just save the current value of the LD_LIBRARY_PATH and CC variables to another pair of environment variables, and make run-compiler a static script that simply restores CC and LD_LIBRARY_PATH to the saved value, and before running the compiler.

This avoids having to inject the values of the variables in the script, possibly causing problems if they contains spaces, quotes, and other special characters.

Also add exports in the script just in case LD_LIBRARY_PATH and CC are not already in the environment.

follow-up from 471dab2

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,9 +193,13 @@ dnl something different but only have that affect the execution of the results
 dnl of the compile, not change the libraries for the compiler itself.
 dnl
 compilersh="run-compiler"
-echo "CC=\"$CC\"" > $compilersh
-echo "LD_LIBRARY_PATH=\"$LD_LIBRARY_PATH\"" >> $compilersh
-echo 'exec $CC $@' >> $compilersh
+export "CURL_SAVED_CC=$CC"
+export "CURL_SAVED_LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+cat <<\EOF > "$compilersh"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -193,9 +193,13 @@ dnl something different but only have that affect the execution of the results
...
-echo "CC=\"$CC\"" > $compilersh
-echo "LD_LIBRARY_PATH=\"$LD_LIBRARY_PATH\"" >> $compilersh
-echo 'exec $CC $@' >> $compilersh
+export COMPILERSH_CC="$CC"
+export COMPILERSH_LD_LIBRARY_PATH="$LD_LIBRARY_PATH"
+cat <<\EOF > "$compilersh"
+export "CC=$COMPILERSH_CC"
+export "LD_LIBRARY_PATH=$COMPILERSH_LD_LIBRARY_PATH"
+exec $CC "$@"
+EOF
```
vszakats, line 202: I prefer the solution that doesn't use public env for this and expand at the time of creation. If we want to stick with this method, prefixing vars with `CURL_` would be a nice touch to avoid collisions.

bagder (approved):

bagder:
Thanks!

PR #11184, opened 2023-05-23
ngtcp2: build with 0.16.0 and nghttp3 0.12.0

bagder:
- moved to qlog_write
- crypto => encryption
- CRYPTO => ENCRYPTION
- removed "_is_"

### DIFF
--- a/.github/workflows/ngtcp2-gnutls.yml
+++ b/.github/workflows/ngtcp2-gnutls.yml
@@ -101,7 +101,7 @@ jobs:
       name: 'install gnutls'
 
     - run: |
-        git clone --quiet --depth=1 -b v0.11.0 https://github.com/ngtcp2/nghttp3
+        git clone --quiet --depth=1 -b v0.12.0 https://github.com/ngtcp2/nghttp3
         cd nghttp3
         autoreconf -fi
         ./configure --prefix=$HOME/all PKG_CONFIG_PATH="$HOME/all/lib/pkgconfig" --enable-lib-only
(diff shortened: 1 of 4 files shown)


bagder:
Needs a nghttp2 update as well, which has not been released yet.

pheiduck:
nghttp2 **1.54.0** is published now.

tatsuhiro-t:
Test failed with error  "curl: ngtcp2_conn.c:67: conn_update_timestamp: Assertion `conn->log.last_ts <= ts' failed.\n".

Starting ngtcp2 v0.16, we verify that timestamp an application passes to ngtcp2 library does not go back.  ngtcp2 entirely depends on that property.

The problem is, in cf_process_ingress, the timestamp is recorded in rctx.ts, and it is passed to ngtcp2_conn_read_pkt.
Inside the loop, cf_flush_egress is called, there we get fresh timestamp, and it could be larger than the value we recorded in rctx.ts.  Then next loop, ngtcp2_conn_read_pkt would cause assertion failure because rctx.ts points to back in time.

Perhaps, we need to pass rctx.ts to cf_flush_egress (or make it some internal sub-function that takes timestamp), so that the timestamp does not go back.

icing:
> Test failed with error "curl: ngtcp2_conn.c:67: conn_update_timestamp: Assertion `conn->log.last_ts <= ts' failed.\n".
> 
> Starting ngtcp2 v0.16, we verify that timestamp an application passes to ngtcp2 library does not go back. ngtcp2 entirely depends on that property.
> 
> The problem is, in cf_process_ingress, the timestamp is recorded in rctx.ts, and it is passed to ngtcp2_conn_read_pkt. Inside the loop, cf_flush_egress is called, there we get fresh timestamp, and it could be larger than the value we recorded in rctx.ts. Then next loop, ngtcp2_conn_read_pkt would cause assertion failure because rctx.ts points to back in time.
> 
> Perhaps, we need to pass rctx.ts to cf_flush_egress (or make it some internal sub-function that takes timestamp), so that the timestamp does not go back.

That can certainly be done. I'll look into this, probably tomorrow.

PR #11185, opened 2023-05-23
transfer numbers for logging, --trace-ids

icing:
Adding identifier to transfers and connections using the same connection cache (implicitly
the case when using the same multi handle). With type `curl_off_t` they are unique for all practical purposes.

Adding new `CURLINFO_XFER_ID` and `CURLINFO_CONN_ID` to the interface and documentation
to make them accessible for libcurl applications.

Adding `--trace-ids` to the curl command line to include the identifier in verbose log output. Example:

```
[4-x] * processing: https://curl.se/?4
...
[4-x] * STATE: INIT => CONNECT
...
[4-0] * STATE: PROTOCONNECT => DO
[4-0] * h2 [:method: GET]
...
[4-0] < HTTP/2 200
...
[4-x] * Expire cleared
```

This does not change any logging statements (yet). If we use these in the verbose output of `curl` itself, it would be suitable to change several statements that now contain connection id and a handle pointer.


lib/conncache.h
```diff
@@ -40,6 +40,7 @@ struct conncache {
   struct Curl_hash hash;
   size_t num_conn;
   long next_connection_id;
+  long next_easy_id;
```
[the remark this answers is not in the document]
icing: Both sound fine to me. With a 64 bit type, we reach uniqueness for all practical purposes, I guess (inside the same multi/conn cache).

The next question is then if we want to add this to all the `info()` statements or make this available to the logging functions. For the latter, it would sneak into the API and I am not sure that this is a good idea. But I do not know if someone is attached to the current log strings either.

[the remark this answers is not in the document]
icing: So, some more thinking between issue solving. A backwards compatible way to introduce this:

* go `curl_off_t` for transfer and connection id (do we have printf format for that one btw?)
* add `CURLINFO_EASY_ID` and `CURLINFO_CONN_ID` for the app to retrieve identifiers (-1 for unassigned, e.g. easy not part of multi yet or connection not assigned)
* add `--trace-id` to `curl` that adds these to its verbose/trace output

Do we provide a standard implementation for apps to set as `CURLOPT_DEBUGFUNCTION`? Something along the lines of:

```
CURL_LOGFMT_DEFAULT   (0)
CURL_LOGFMT_IDS    (1 << 0)
CURL_LOGFMT_TIME    (1 << 1)

/* set DEBUGFUNCTION/DEBUGDATA to log to FILE for this easy handle */
curl_easy_set_log(CURL*, FILE*, int fmt)
/* set DEBUGFUNCTION/DEBUGDATA to log to FILE for all easy handles in multi
    that have no other setting already */
curl_multi_set_log(CURLM*, FILE*, int fmt)

```

olife-cyber (commented):
Thank you

lib/curl_log.h
```diff
@@ -124,15 +124,15 @@ void Curl_log_cf_debug(struct Curl_easy *data, struct Curl_cfilter *cf,
  * [CONN-1-0][CF-SSL] this filter very much rocks
  * on connection #1 with sockindex 0 for filter of type "SSL". */
 #define DMSG(d,msg)  \
-  "[CONN-%ld] "msg, (d)->conn->connection_id
+  "[CONN-%zd] "msg, (d)->conn->connection_id
```
bagder, line 127: I think you need to use `"%" CURL_FORMAT_CURL_OFF_T` everywhere for outputting `curl_off_t` variables correctly. The `z` flag is for `size_t` and it can be different than `curl_off_t` !

lib/multi.c
```diff
@@ -194,7 +194,7 @@ static void mstate(struct Curl_easy *data, CURLMstate state
 #if defined(DEBUGBUILD) && !defined(CURL_DISABLE_VERBOSE_STRINGS)
   if(data->mstate >= MSTATE_PENDING &&
      data->mstate < MSTATE_COMPLETED) {
-    long connection_id = -5000;
+     curl_off_t connection_id = -5000;
```
bagder, line 197: Shouldn't this be `-1` now to match the documentation?

lib/multi.c
```diff
@@ -3690,7 +3704,7 @@ void Curl_expire_clear(struct Curl_easy *data)
     }
 
 #ifdef DEBUGBUILD
-    infof(data, "Expire cleared (transfer %p)", data);
+    infof(data, "Expire cleared (easy handle %p)", data);
```
bagder, line 3707: Maybe use the transfer id instead for consistency?

lib/curl_log.h
icing: I removed these macros.

bagder (approved):

Issue #11194, opened 2023-05-24
HTTP/2 POST reset/timeout after upgrade from 8.0.1 to 8.1.1


OlekMazur:
After upgrade from 8.0.1 to 8.1.1 libcurl is unable to POST anything to IBM Cloud (OpenWhisk) using HTTP/2.
I reproduced the problem using curl on Fedora. Unfortunately there seems to be no public endpoint to retest it.
I'm not sure if it is a duplicate of #11181.

### I did this

curl --verbose --header "Content-Type: text/csv" --data-binary @file --digest --user test:test https://eu-gb.functions.appdomain.cloud/api/v1/<my_secret_api_endpoint>

### I expected the following

```
*   Trying 172.65.198.79:443...
* Connected to eu-gb.functions.appdomain.cloud (172.65.198.79) port 443 (#0)
* ALPN: offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/pki/tls/certs/ca-bundle.crt
*  CApath: none
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=eu-gb.functions.appdomain.cloud
*  start date: May 13 05:16:57 2023 GMT
*  expire date: Aug 11 05:16:56 2023 GMT
*  subjectAltName: host "eu-gb.functions.appdomain.cloud" matched cert's "eu-gb.functions.appdomain.cloud"
*  issuer: C=US; O=Let's Encrypt; CN=R3
*  SSL certificate verify ok.
* using HTTP/2
* Server auth using Digest with user 'test'
* h2h3 [:method: POST]
* h2h3 [:path: /api/v1/<my_secret_api_endpoint>]
* h2h3 [:scheme: https]
* h2h3 [:authority: eu-gb.functions.appdomain.cloud]
* h2h3 [user-agent: curl/8.0.1]
[...]

PR #11200, opened 2023-05-24
http2: fix EOF handling on uploads with auth negotiation

icing:
- doing a POST with `--digest` does an override on the initial request with `Content-Length: 0`, but the http2 filter was unaware of that and expected the originally request body. It did therefore not send a final DATA frame with EOF flag to the server.
- The fix overrides any initial notion of post size when the `done_send` event is triggered by the transfer loop, leading to the EOF that is necessary.
- refs #11194. The fault did not happen in testing, as Apache httpd never tries to read the request body of the initial request, sends the 401 reply and closes the stream. The server used in the reported issue however tried to read the EOF and timed out on the request.

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1527,10 +1527,8 @@ static CURLcode http2_data_done_send(struct Curl_cfilter *cf,
   if(!stream->send_closed) {
     stream->send_closed = TRUE;
     if(stream->upload_left) {
-      /* If we operated with unknown length, we now know that everything
-       * that is buffered is all we have to send. */
-      if(stream->upload_left == -1)
-        stream->upload_left = Curl_bufq_len(&stream->sendbuf);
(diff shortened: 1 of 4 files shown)


dfandrich:
Does this affect all/most POSTs done with digest auth in HTTP/2 on non-Apache
servers? If so, then it sounds serious enough to warrant doing another point
release.

Note also the typo in Content-Length in the commit description.

icing:
> Does this affect all/most POSTs done with digest auth in HTTP/2 on non-Apache servers? If so, then it sounds serious enough to warrant doing another point release.

I do not know how many servers react this way.

bagder:
Thanks!

bagder:
Merged in 5c58cb0212bcf63 (typoed the keyword)

Issue #11205, opened 2023-05-25
Cloudflare DoH does not reply to query via HTTP/3


sfionov:
After updating to curl 8.1.1 I noticed that cloudflare DoH does not reply to my query sent via libcurl.

Seems a regression compared to 7.85 w/ ngtcp2/0.9.0 + nghttp3/0.7.1.

It works with --http2 but not with --http3.

### I did this

```bash
curl-8.1.1 % cat dns3.txt
0000012000010000000000010667697468756203636f6d00000100010000
291000000000000000
curl-8.1.1 % xxd -r -ps dns3.txt > dns3.bin
curl-8.1.1 % bin/curl --http3 -vk https://cloudflare-dns.com/dns-query -X POST -H "Accept: application/dns-message" -H "Content-Type: application/dns-message" --data-binary @./dns3.bin      
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying [2606:4700::6810:f8f9]:443...
* Skipped certificate verification
* Connected to cloudflare-dns.com (2606:4700::6810:f8f9) port 443 (#0)
* using HTTP/3
* Using HTTP/3 Stream ID: 0 (easy handle 0x156010408)
> POST /dns-query HTTP/3
> Host: cloudflare-dns.com
> User-Agent: curl/8.1.1-DEV
> Accept: application/dns-message
> Content-Type: application/dns-message
> Content-Length: 39
> 
* We are completely uploaded and fine
< HTTP/3 400 
< server: nginx
< date: Thu, 25 May 2023 11:44:39 GMT
< access-control-allow-origin: *
< content-length: 14
< cf-ray: 7ccd8719696b7207-IST
< 
* Connection #0 to host cloudflare-dns.com left intact
Invalid query.% 
```

### I expected the following
```
< HTTP/3 200 
< server: cloudflare
...
```

### curl/libcurl version

curl 8.1.1-DEV (Darwin) libcurl/8.1.1-DEV BoringSSL zlib/1.2.11 nghttp2/1.44.0 ngtcp2/0.15.0 nghttp3/0.11.0
Release-Date: [unreleased]
Protocols: http https
Features: alt-svc AsynchDNS HSTS HTTP2 HTTP3 HTTPS-proxy IPv6 Largefile libz NTLM SSL threadsafe TrackMemory UnixSockets

### operating system
macOS 13.3.1
```
[...]

PR #11207, opened 2023-05-25
http/3, send EOF indicator early as possible

icing:
- ngtcp2 and quiche implementations relied on the DONE_SEND event to forward the EOF for uploads to the libraries. This often result in a last 0 length EOF data. Tracking the amount of data left to upload allows EOF indication earlier.
- refs #11205 where CloudFlare DoH servers did not like to receive the initial upload DATA without EOF and returned a 400 Bad Request


bagder:
Thanks!

PR #11211, opened 2023-05-26
curl, --trace-time, cache value for a second

icing:
- caches HH:MM:SS computed and reuses it for logging during the same second.
- common function for plain log line start formatting


bagder (approved):

bagder:
Thanks!

PR #11212, opened 2023-05-26
remove trailing 'dnl' that causes this to break autoconf etc

foamdino:
This trailing `dnl` causes the following error when used as an automake extension
```
/usr/bin/m4:mk/m4ext/ax_libcurl.m4:275: Warning: end of file treated as newline
autom4te: error: /usr/bin/m4 failed with exit status: 1
aclocal: error: /usr/bin/autom4te failed with exit status: 1
autoreconf: error: aclocal failed with exit status: 1
```

### DIFF
--- a/docs/libcurl/libcurl.m4
+++ b/docs/libcurl/libcurl.m4
@@ -272,4 +272,4 @@ if (x) {;}
   fi
 
   unset _libcurl_with
-])dnl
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #11220, opened 2023-05-29
tests/servers.pm: pick unused port number with a server socket

bagder:
This change replaces the previous method of picking a port number at random to try to start servers on, then retrying up to ten times with new random numbers each time, with a function that creates a server socket on port zero, thereby getting a suitable random port set by the kernel. That server socket is then closed and that port number is used to setup the actual test server on.

There is a risk that *another* server can be started on the machine in the time gap, but the server verification feature will detect that.

### DIFF
--- a/tests/servers.pm
+++ b/tests/servers.pm
@@ -1796,7 +1796,6 @@ sub runsshserver {
     $flags .= "--ipv$ipvnum --addr \"$ip\" ";
     $flags .= "--user \"$USER\"";
 
-    my $wport = 0,
     my @tports;
     my $port = getfreeport($ipvnum);
 
@@ -1866,9 +1865,9 @@ sub runsshserver {
         die $msg;
     }
 
-    logmsg "RUN: $srvrname on PID $pid2 port $wport\n" if($verb);
+    logmsg "RUN: $srvrname on PID $pid2 port $port\n" if($verb);
 
-    return (0, $pid2, $sshpid, $wport);
+    return (0, $pid2, $sshpid, $port);
 }
 
 #######################################################################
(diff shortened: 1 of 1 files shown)


tests/servers.pm
```diff
@@ -1369,36 +1366,28 @@ sub runhttpsserver {
...
-        }
-        last;
+    if($httpspid <= 0 || !pidexists($httpspid)) {
+        # it is NOT alive
+        # don't call stopserver since that will also kill the dependent
+        # server that has already been started properly
+        $doesntrun{$pidfile} = 1;
+        $httpspid = $pid2 = 0;
+        logmsg "RUN: failed to start the $srvrname server\n";
+        return;
```
dfandrich, line 1381: This new error return path returns nothing here, while callers expect there to be 4 return values. I believe the missing first argument especially will cause an issue since it's not indicating an error. This goes for at last time call but probably the other similar bare return elsewhere in this PR.

dfandrich (commented):
This should definitely help with reliable test server startup, especially in a free port starvation situation. If all the test servers actually start with this PR, then we can assume that they are all already setting SO_REUSEADDR and therefore need no further change to that effect.

Some points that I noticed while reviewing:
- $serverstartretries can be deleted
- %usedports can be deleted

tests/servers.pm
bagder: I figured without return values they would all be nothing or equal zero, so they would be caught by the calling function as an error. Is that wrong?

dfandrich:
The first return value must be nonzero to mean error. e.g. 0+!$http2pid is 0 on
success (when $http2pid contains a PID) and 1 on error (when $http2pid is
zero). That's going to cause startservers() the return the wrong value on error.
I had to refactor startservers() in f82a38e0f3 to return a fine-grained error
in order to preserve the right behaviour after other refactoring.

PR #11228, opened 2023-05-30
Fix run-compiler for old /bin/sh

asedeno:
If you try to assign and export on the same line on some older /bin/sh implementations, it complains:

```
$ export "NAME=value"
NAME=value: is not an identifier
```

Also, as written, the name and value are quoted together. This also makes old /bin/sh sad:

```
$ "NAME=value"
NAME=value: not found
```

This commit rewrites run-compiler's assignments and exports to work with old /bin/sh, splitting assignment and export into two separate statements, and only quote the value. So now we have:

```
NAME="value"
export NAME
```

While we're here, make the same change to the two supporting assign+export lines preceeding the script to be consistent with how exports work throughout the rest of configure.ac.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,11 +193,15 @@ dnl something different but only have that affect the execution of the results
 dnl of the compile, not change the libraries for the compiler itself.
 dnl
 compilersh="run-compiler"
-export "CURL_SAVED_CC=$CC"
-export "CURL_SAVED_LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+CURL_SAVED_CC="$CC"
+export CURL_SAVED_CC
+CURL_SAVED_LD_LIBRARY_PATH="$LD_LIBRARY_PATH"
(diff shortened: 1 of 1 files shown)


emanuele6:
Splitting assignments and export into multiple commands to make the script work in older systems with an sh interpreters that doesn't comply with POSIX issue 6 or issue 7 is fine. :+1:

> Also, as written, the name and value are quoted together. This also makes old `/bin/sh` sad:
> ```
> $ "NAME=value"
> NAME=value: not found
> ```

That makes any sh sad, not only old sh. :(
Assignments only need to be quoted if they are an argument to a command like `export`.

asedeno:
> That makes any sh sad, not only old sh. :( Assignments only need to be quoted if they are an argument to a command like `export`.

Fair enough, we can rewrite the commit message to drop that note if we like.

bagder:
Thanks!

PR #11234, opened 2023-06-01
libssh2: do not use custom alloc functions for kbd_callback

bagder:
When libssh2_userauth_keyboard_interactive_ex() calls the callback to libcurl code, it allocates memory that is then managed and ultimately free()ed by libssh2.

It is then important that we don't use the normal strdup() call in libcurl, since this method will use a special memory system for debug builds and cause libssh2 to do a bad free().

If CURL_LIBSSH2_DEBUG is defined, we tell libssh2 to use the curl memory function and then it *should* instead use the curl function for strdup()!


bagder:
Never mind, I will solve this differently.

Issue #11111, opened 2023-05-14
I want to resolve onion addresses


kaizushi:
### I did this

Curl will not resolve onion addresses anymore on my system. This is because of RFC 7686, and the change was made because of DNS leakage. I have a good handle on any possible DNS leakage and do not need this feature, and it in fact breaks my system. I use transparent proxying, and I am certain DNS leakage is impossible here.

### curl/libcurl version

curl 8.0.1 (x86_64-pc-linux-gnu) libcurl/8.0.1 OpenSSL/1.1.1t zlib/1.2.13 libidn2/2.3.4 nghttp2/1.51.0
Release-Date: 2023-03-20
Protocols: dict file ftp ftps http https imap imaps mqtt pop3 pop3s rtsp smtp smtps tftp
Features: AsynchDNS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM SSL threadsafe TLS-SRP UnixSockets

### operating system

Gentoo Linux (default/linux/amd64/17.1/hardened/selinux)

bagder:
> Curl will not resolve onion addresses anymore on my system

That's a bit surprising since 0ae0abbe72514a75c10bfc4108d9f254f594c086 was merged *after* 8.0.1. `.onion` blocking is introduced in 8.1.0 for the first time. You are seeing something else or are running another version.

We could consider adding an option that allows a user to change this behavior.

kaizushi:
Gentoo literally implements this already, as I said this breaks my system which deals with DNS leakage other ways. It breaks all applications which on my system use Tor transparently, and end up using DNS for many reasons including onions. You really need to make this an optional feature at compile time. It has absolutely no purpose on my system and breaks scripts and software for many of my users. Please be reasonable, this is supposed to be a world of choice.

Running Tor on localhost on my system is an unacceptable compromise to security. Running a SOCKS port on the middlebox on my system and modifying torsocks and such on my system is an unacceptable compromise to security. This patch breaks systems with superior security to people running Tor on localhost, and it is important to let those of us with such systems opt out, anything otherwise is an unacceptable compromise to security.

dfandrich:
curl implemented this to follow RFC 7686 which states

  Applications that do not implement the Tor protocol SHOULD generate an error upon the use of .onion and SHOULD NOT perform a DNS lookup.

curl & libcurl do not implement the Tor protocol, so "Please be reasonable" to me means to follow the RFCs.

bagder:
[...]

PR #11236, opened 2023-06-01
hostip: add CURL_ALLOW_DOT_ONION to allow .onion lookups

omnivagant:
RFC 7686 states that:

> Applications that do not implement the Tor protocol SHOULD generate
> an error upon the use of .onion and SHOULD NOT perform a DNS lookup.

RFC 2119 states that:

> SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
> there may exist valid reasons in particular circumstances when the
> particular behavior is acceptable or even useful, but the full
> implications should be understood and the case carefully weighed
> before implementing any behavior described with this label.

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -682,13 +682,20 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
   CURLcode result;
   enum resolve_t rc = CURLRESOLV_ERROR; /* default to failure */
   struct connectdata *conn = data->conn;
-  /* We should intentionally error and not resolve .onion TLDs */
-  size_t hostname_len = strlen(hostname);
-  if(hostname_len >= 7 &&
-     (curl_strequal(&hostname[hostname_len - 6], ".onion") ||
-      curl_strequal(&hostname[hostname_len - 7], ".onion."))) {
-    failf(data, "Not resolving .onion address (RFC 7686)");
-    return CURLRESOLV_ERROR;
+
+  /* We should intentionally error and not resolve .onion TLDs, however,
+   * there are cases where a user may knowingly want to do so. */
+  char *allow_dot_onion = curl_getenv("CURL_ALLOW_DOT_ONION");
+  if(allow_dot_onion)
+    curl_free(allow_dot_onion);
+  else {
+    size_t hostname_len = strlen(hostname);
+    if(hostname_len >= 7 &&
+       (curl_strequal(&hostname[hostname_len - 6], ".onion") ||
+        curl_strequal(&hostname[hostname_len - 7], ".onion."))) {
+      failf(data, "Not resolving .onion address (RFC 7686)");
+      return CURLRESOLV_ERROR;
+    }
   }
   *entry = NULL;
 #ifndef CURL_DISABLE_DOH


omnivagant:
I have tested and it works as intended but I currently consider this WIP as it doesn't add any tests or to documentation and `ALLOW_ONION_LOOKUP` was just grabbed out of my behind. Still wanted to share for comments and input.

omnivagant:
Also, when curl is `./configure`d with `--enable-ares` this is will still fail since https://github.com/c-ares/c-ares/commit/955df983d7f643d1b00aa433f31ca3ec7f86b4bd with "Could not resolve host" but that is out-of-scope for this PR besides documentation, probably.

jzakrzewski:
If the env. name is not standardized (via a paper or de-facto), it should probably be namespaced (start with `CURL_`).

bagder:
I want to hear more about how this is used for legitimate purposes and not just actually for the purpose of snooping on `.onion` hosts being resolved - the exact thing this is meant to prevent.

In the previous discussion about this ask, it was mention how you could resolve a `.onion` name using a Tor DNS, but if you do, you would get a set of IP addresses back that can only be used within the Tor network so curl would still not be able to connect to those IP addresses normally surely?

omnivagant:
Transparent Tor Proxying (as mentioned in #11111) where you direct all traffic through Tor, and DNS requests (at least for .onion) to tor(1), where your applications do not need to know about Tor, is not that uncommon. There are multiple guides on how to set this up and several projects that will do this for you.

Usually with help of your systems firewall with NAT rules and sometimes additionally helped by dnsmasq or dnscrypt-proxy configuration for .onion addresses. Sometimes you do this locally on your host, sometimes in a chained network VM (like a Qubes OS "ProxyVM") and sometimes for a whole LAN.

omnivagant:
The, now namespace prefixed, toggle name change was inspired by Firefox that does follow RFC 7686 by having `network.dns.blockDotOnion` set to *true*, available to toggle through `about:config`.

bagder:
A follow-up question is: if we want to allow switching this off, how should we do that? Adding an environment variable for this purpose is a little surprising and unusual, possibly also a bit hard to find for users. We normally add options to `curl_easy_setopt()` for these things.

Alternatives that I can think of:

1. only environment variable support in libcurl (current PR)
2. environment variable + setopt in libcurl
3. environment variable + cmdline option in curl, setopt in libcurl
4. cmdline option in curl, setopt in libcurl

Where (4) is the standard for transfer options and we should only deviate from that path with very good reasons.

omnivagant:
I think I'm leaning toward (3) and not have the environment variable namespace prefixed, if the name is standardized, so that it can also be picked up by others. What would flag like `--allow-dot-onion` help if curl was built with `--enable-ares`?

I've opened https://gitlab.torproject.org/tpo/core/torspec/-/issues/202 to see what the Tor Project  people have to say.

Kangie:
Tor [do not believe](https://gitlab.torproject.org/tpo/core/torspec/-/issues/202#note_2909909) than an environment variable is the appropriate mechanism for 'disable rfc protection'.

IMO the RFC doesn't leave any wiggle-room there; adhere to it or don't; please don't implement a non-standard mechanism that changes the way DNS resolution works.

Instead, I suggest that this moves back into the discussion space. Some suggestions based on what we have so far from Tor:

1. Is Transparent Proxying best practice in 2023? Is there a better way to accomplish your goal that doesn't require doing anything client applications? Assume that RFC7686 is going to be adhered to by client applications.
2. How can cURL become aware that it's part of a Tor-enabled environment?

bagder:
> Tor [do not believe](https://gitlab.torproject.org/tpo/core/torspec/-/issues/202#note_2909909) than an environment variable is the appropriate mechanism for 'disable rfc protection'.

I am strongly against doing things in libcurl based on environment variables unless it is some kind of standard set somewhere. Which this is not.

milahu:
thanks! im using this workaround on nixos linux via https://github.com/milahu/nur-packages/commit/33b3ef8cc8f4d2b177f153464db2ba37a304b08c

milahu:
it would be nice to add a hint after the error message, something like

> hint: wrap your command with "CURL_ALLOW_DOT_ONION=1 torsocks"

markozajc:
Is there a workaround for this that doesn't involve recompiling curl?

milahu:
> Is there a workaround for this that doesn't involve recompiling curl?

https://github.com/curl/curl/discussions/11125#discussioncomment-6982136

markozajc:
> > Is there a workaround for this that doesn't involve recompiling curl?
> 
> [#11125 (comment)](https://github.com/curl/curl/discussions/11125#discussioncomment-6982136)

Thanks!

bagder (changes requested):
libcurl should not be controlled by environment variables like this. It makes it very inconvenient for applications. There is an API that should rather be used.

bagder:
This needs further discussion and deliberation.

PR #11241, opened 2023-06-02
system.h: Remove __IBMC__/__IBMCPP__ guards and apply to all z/OS (__MVS__) compiles

IgorTodorovskiIBM:
* z/OS now supports Clang where `__IBMC__`/`__IBMCPP__` macros are not set, yet the same system.h macro values apply.


bagder:
Thanks!

Issue #11242, opened 2023-06-03
curl consumes 100% CPU when sending a file with -F


l29ah:
### I did this
`curl -Ffile=@"VID_20230603_010956.mp4" https://0x0.st`

### I expected the following
Tiny CPU consumption appropriate for the pathetic ~200kB/s transfer speed on my i7 CPU.

### curl/libcurl version
```
curl 8.1.2 (x86_64-pc-linux-gnu) libcurl/8.1.2 GnuTLS/3.8.0 (OpenSSL/1.1.1u) zlib/1.2.13 c-ares/1.19.1 libidn2/2.3.4 nghttp2/1.52.0 librtmp/2.3
Release-Date: 2023-05-30
Protocols: dict file ftp ftps http https imap imaps mqtt pop3 pop3s rtmp rtsp smtp smtps tftp
Features: AsynchDNS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz MultiSSL NTLM SSL threadsafe TLS-SRP UnixSockets
```

### operating system
Gentoo Lignux
`Linux l29ah-x201 6.2.5+ #238 SMP PREEMPT_DYNAMIC Mon Mar 13 02:40:28 CET 2023 x86_64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz GenuineIntel GNU/Linux`

### top of perf report
```
# Children      Self  Command  Shared Object          Symbol                                          
# ........  ........  .......  .....................  ................................................
#
    27.96%     0.00%  curl     [unknown]              [.] 0x0000000000020003
            |
            ---0x20003
               |          
               |--4.68%--pthread_mutex_lock@@GLIBC_2.2.5
               |          
               |--3.98%--_gnutls_io_write_flush
               |          
               |--3.81%--_gnutls_record_overhead
               |          
               |--3.61%--_gnutls_epoch_get
               |          
               |--3.18%--record_overhead_rt
               |          
               |--2.50%--__GI___pthread_mutex_unlock_usercnt
               |          
               |--1.47%--gnutls_system_mutex_lock
               |          
               |--1.35%--_mbuffer_head_get_first
               |          
               |--1.06%--gnutls_system_mutex_unlock
               |          
                --0.91%--_gnutls_writev_emu.constprop.0

    12.93%     5.91%  curl     libc.so.6              [.] __send
            |          
            |--7.02%--__send
            |          
            |--0.60%--0x3c668934bc4645c5
            |          0x6a8
            |          |          
[...]

PR #11247, opened 2023-06-04
bufq: make write/pass methods more robust

icing:
- related to #11242 where curl enters busy loop when sending http2 data to the server


bagder (approved):

bagder:
Thanks!

PR #11248, opened 2023-06-04
ngtcp2: use 0.16.0

pheiduck:
- nghttp3 0.12.0
- nghttp2 1.54.0 (pending)


pheiduck:
WIP is done here: https://github.com/curl/curl/pull/11184

PR #7121, opened 2021-05-24
lib/hostip.c: make NAT64 address synthesis on macOS working

zajdee:
On macOS running in an IPv6-only access network where DNS64 and NAT64 functions are also deployed, the operating system is able to synthesize the NAT64 IPv6 address when an IPv4-only literal is used. This is described in `man 3 getaddrinfo` as follows:
```
     The current implementation supports synthesis of NAT64 mapped IPv6 addresses.  If hostname
     is a numeric string defining an IPv4 address (for example, ``192.0.2.1'' ) and ai_family is
     set to PF_UNSPEC or PF_INET6, getaddrinfo() will synthesize the appropriate IPv6
     address(es) (for example, ``64:ff9b::192.0.2.1'' ) if the current interface supports IPv6,
     NAT64 and DNS64 and does not support IPv4. If the AI_ADDRCONFIG flag is set, the IPv4
     address will be suppressed on those interfaces.  On non-qualifying interfaces,
     getaddrinfo() is guaranteed to return immediately without attempting any resolution, and
     will return the IPv4 address if ai_family is PF_UNSPEC or PF_INET. NAT64 address synthesis
     can be disabled by setting the AI_NUMERICHOST flag. To best support NAT64 networks, it is
     recommended to resolve all IP address literals with ai_family set to PF_UNSPEC and ai_flags
     set to AI_DEFAULT.

     Note that NAT64 address synthesis is always disabled for IPv4 addresses in the following
     ranges: 0.0.0.0/8, 127.0.0.0/8, 169.254.0.0/16, 192.0.0.0/29, 192.88.99.0/24, 224.0.0.0/4,
     and 255.255.255.255/32.  Additionally, NAT64 address synthesis is disabled when the network
     uses the well-known prefix (64:ff9b::/96) for IPv4 addresses in the following ranges:
     10.0.0.0/8, 100.64.0.0/10, 172.16.0.0/12, and 192.168.0.0/16.

     Historically, passing a host's own hostname to getaddrinfo() has been a popular technique
     for determining that host's IP address(es), but this is fragile, and doesn't work reliably
     in all cases. The appropriate way for software to discover the IP address(es) of the host
     it is running on is to use getifaddrs(3).
```
The operating system however does not perform this synthesis unless the `SCDynamicStoreCopyProxies` function from the `SystemConfiguration` framework is called.

[...]

PR #11254, opened 2023-06-05
src/hostip.c: Move macOS-specific calls into global init call

### LINKED ISSUE #11252: lib/hostip.c: SCDynamicStoreCopyProxies is not safe to run in a fork
### I did this

The Elasticsearch Ruby client uses the [`Typhoeus` Ruby gem](https://github.com/typhoeus/typhoeus), which uses `libcurl`. If you do something like this:

```ruby
require 'typhoeus'

fork {
  Typhoeus.head('https://wwww.google.com')
}
```

The process crashes with this warning:

```
objc[73067]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[73067]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

### I expected the following

No error.

For simple applications, you can work around the problem by making the `libcurl` call outside of the fork:

```ruby
require 'typhoeus'
 
Typhoeus.head('http://localhost')

fork {
  Typhoeus.head('http://localhost')
}
```

However, this doesn't always work. As discussed in the links below, these macOS system calls are not thread-safe. Under some circumstances calling `SCDynamicStoreCopyProxiesWithOptions` can crash with a backtrace as the following:

```
Target 0: (ruby) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x109be0b22)
  * frame #0: 0x0000000186ed66a0 libsystem_trace.dylib`_os_log_preferences_refresh + 64
    frame #1: 0x0000000186ed710c libsystem_trace.dylib`os_log_type_enabled + 712
    frame #2: 0x00000001871ea430 CoreFoundation`-[CFPrefsSearchListSource alreadylocked_copyValueForKey:] + 204
    frame #3: 0x00000001871ea344 CoreFoundation`-[CFPrefsSource copyValueForKey:] + 52
    frame #4: 0x00000001871ea2f8 CoreFoundation`__76-[_CFXPreferences copyAppValueForKey:identifier:container:configurationURL:]_block_invoke + 32
    frame #5: 0x00000001871e3260 CoreFoundation`__108-[_CFXPreferences(SearchListAdditions) withSearchListForIdentifier:container:cloudConfigurationURL:perform:]_block_invoke + 376
    frame #6: 0x000000018735fc78 CoreFoundation`-[_CFXPreferences withSearchListForIdentifier:container:cloudConfigurationURL:perform:] + 384
    frame #7: 0x00000001871e2b30 CoreFoundation`-[_CFXPreferences copyAppValueForKey:identifier:container:configurationURL:] + 168
    frame #8: 0x00000001871e2a4c CoreFoundation`_CFPreferencesCopyAppValueWithContainerAndConfiguration + 112
    frame #9: 0x0000000187e23f24 SystemConfiguration`SCDynamicStoreCopyProxiesWithOptions + 180
    frame #10: 0x000000019d29c01c libcurl.4.dylib`Curl_resolv + 288
    frame #11: 0x000000019d2cd40c libcurl.4.dylib`resolve_server + 320
    frame #12: 0x000000019d2cba80 libcurl.4.dylib`Curl_connect + 5764
    frame #13: 0x000000019d2b3344 libcurl.4.dylib`multi_runsingle + 580
    frame #14: 0x000000019d2b2fdc libcurl.4.dylib`curl_multi_perform + 124
    frame #15: 0x000000019d290bcc libcurl.4.dylib`curl_easy_perform + 276
    frame #16: 0x0000000111a68044 ffi_c.bundle`ffi_call_SYSV + 68
    frame #17: 0x0000000111a639fc ffi_c.bundle`ffi_call_int + 1560
    frame #18: 0x0000000111a633d8 ffi_c.bundle`ffi_call + 52
    frame #19: 0x0000000111a56ef0 ffi_c.bundle`call_blocking_function(data=<unavailable>) at Call.c:336:5
    frame #20: 0x0000000105167cd0 libruby.3.1.dylib`rb_nogvl + 268
    frame #21: 0x0000000111a56ec8 ffi_c.bundle`rbffi_do_blocking_call(data=<unavailable>) at Call.c:344:5
    frame #22: 0x0000000105019af8 libruby.3.1.dylib`rb_vrescue2 + 368
    frame #23: 0x0000000105019960 libruby.3.1.dylib`rb_rescue2 + 44
    frame #24: 0x0000000111a5715c ffi_c.bundle`rbffi_CallFunction(argc=<unavailable>, argv=<unavailable>, function=0x000000019d290ab8, fnInfo=0x00006000037bad60) at Call.c:387:9
    frame #25: 0x0000000111a5ace8 ffi_c.bundle`attached_method_invoke(cif=<unavailable>, mretval=0x000000016b2cdc90, parameters=<unavailable>, user_data=<unavailable>) at MethodHandle.c:174:26
    frame #26: 0x0000000111a63f08 ffi_c.bundle`ffi_closure_SYSV_inner + 988
    frame #27: 0x0000000111a681b4 ffi_c.bundle`.Ldo_closure + 20
    frame #28: 0x00000001051b64e4 libruby.3.1.dylib`vm_call_cfunc_with_frame + 232
    frame #29: 0x00000001051b8c08 libruby.3.1.dylib`vm_sendish + 1336
    frame #30: 0x000000010519b53c libruby.3.1.dylib`vm_exec_core + 8128
```

1. https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
1. https://bugs.python.org/issue33725

This system call was introduced in https://github.com/curl/curl/pull/7121 by @zajdee. I'm wondering: can we avoid calling this system call more than once? For example, I'm wondering if we can move this into `curl_global_init()` instead.

### curl/libcurl version

```
curl 7.88.1 (x86_64-apple-darwin22.0) libcurl/7.88.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0
Release-Date: 2023-02-20
```

### operating system

macOS Ventura 13.4

stanhu:
https://github.com/curl/curl/pull/7121 introduced a macOS system call to `SCDynamicStoreCopyProxies`, which is invoked every time an IP address needs to be resolved.

However, this system call is not thread-safe, and macOS will kill the process if the system call is run first in a fork. To make it possible for the parent process to call this once and prevent the crash, only invoke this system call in the global initialization routine.

In addition, this change is beneficial because it:

1. Avoids extra macOS system calls for every IP lookup.
2. Consolidates macOS-specific initialization in a separate file.

Closes #11252

### DIFF
--- a/lib/easy.c
+++ b/lib/easy.c
@@ -63,6 +63,7 @@
 #include "slist.h"
 #include "mime.h"
 #include "amigaos.h"
+#include "macos.h"
 #include "warnless.h"
 #include "sigpipe.h"
 #include "vssh/ssh.h"
@@ -180,6 +181,11 @@ static CURLcode global_init(long flags, bool memoryfuncs)
     goto fail;
   }
 #endif
+
+  if(Curl_macos_init()) {
+    DEBUGF(fprintf(stderr, "Error: Curl_macos_init failed\n"));
+    goto fail;
+  }
 
   if(Curl_resolver_global_init()) {
     DEBUGF(fprintf(stderr, "Error: resolver_global_init failed\n"));
--- a/lib/macos.c
+++ b/lib/macos.c
@@ -1,0 +1,62 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+
+#include "curl_setup.h"
+
+#if defined(__APPLE__)
+
+#if defined(TARGET_OS_OSX) && TARGET_OS_OSX
+
+#include <curl/curl.h>
+
+#include "macos.h"
+
+#if defined(ENABLE_IPV6) && defined(CURL_OSX_CALL_COPYPROXIES)
+#include <SystemConfiguration/SCDynamicStoreCopySpecific.h>
+#endif
+
+CURLcode Curl_macos_init(void)
+{
+#if defined(ENABLE_IPV6) && defined(CURL_OSX_CALL_COPYPROXIES)
+  {
+    /*
+     * The automagic conversion from IPv4 literals to IPv6 literals only
+     * works if the SCDynamicStoreCopyProxies system function gets called
+     * first. As Curl currently doesn't support system-wide HTTP proxies, we
+     * therefore don't use any value this function might return.
+     *
+     * This function is only available on a macOS and is not needed for
+     * IPv4-only builds, hence the conditions above.
+     */
+    CFDictionaryRef dict = SCDynamicStoreCopyProxies(NULL);
+    if(dict)
+      CFRelease(dict);
+  }
+#endif
+  return CURLE_OK;
+}
+
+#endif /* TARGET_OS_OSX */
+
+#endif /* __APPLE__ */
--- a/lib/macos.h
+++ b/lib/macos.h
@@ -1,0 +1,38 @@
+#ifndef HEADER_CURL_MACOS_H
+#define HEADER_CURL_MACOS_H
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+#include "curl_setup.h"
+
+#if defined(__APPLE__) && defined(TARGET_OS_OSX) && TARGET_OS_OSX
+
+CURLcode Curl_macos_init(void);
+
+#else
+
+#define Curl_macos_init() CURLE_OK
+
+#endif
+
+#endif /* HEADER_CURL_MACOS_H */
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -179,6 +179,7 @@ LIB_CFILES =         \
   krb5.c             \
   ldap.c             \
   llist.c            \
+  macos.c            \
   md4.c              \
   md5.c              \
   memdebug.c         \
@@ -315,6 +316,7 @@ LIB_HFILES =         \
   inet_ntop.h        \
   inet_pton.h        \
   llist.h            \
+  macos.h            \
   memdebug.h         \
   mime.h             \
   mqtt.h             \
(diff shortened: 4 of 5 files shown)


stanhu:
@zajdee Could you take a look and see if this IPv6 -> IPv4 mapping still works here?

zajdee:
It does look okay. Unfortunately I won't get to actually trying it sooner than tomorrow evening - would you be okay waiting until then?

stanhu:
@zajdee Sure, thanks.

stanhu:
@zajdee Would you mind confirming whether this change still works for the issue you were seeing with IPv6 mappings?

lib/macos.c
```diff
@@ -0,0 +1,62 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
```
bagder, line 8: Please remove the year range from here. If you check the other files we don't do the years anymore.

lib/easy.c
```diff
@@ -181,6 +182,13 @@ static CURLcode global_init(long flags, bool memoryfuncs)
   }
 #endif
 
+#if defined(__APPLE__)
+  if(Curl_macos_init()) {
+    DEBUGF(fprintf(stderr, "Error: Curl_macos_init failed\n"));
+    goto fail;
+  }
+#endif
```
bagder, line 190: You can remove the #ifdefs from this to make it easier on the eyes. Non-macOS builds will have this call optimized away since it can never be true then.

lib/macos.h
```diff
@@ -0,0 +1,38 @@
+#ifndef HEADER_CURL_MACOS_H
+#define HEADER_CURL_MACOS_H
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 1998 - 2022, Daniel Stenberg, <daniel@haxx.se>, et al.
```
bagder, line 10: Remove these years as well.

lib/macos.c
stanhu: Thanks. I was working on an older `master`.

lib/easy.c
stanhu: Makes sense, thanks.

zajdee:
I have finally managed to find some time to test your branch. Doing the simplest:

```
autoreconf -fi
./configure --with-secure-transport
make
```

results in a missing symbol:

```
$ ./src/curl -h
dyld[97017]: symbol not found in flat namespace '_Curl_macos_init'
Abort trap: 6
```

I will be investigating further and let you know.
(The simplest autoreconf/configure/make builds a working curl when using the main `curl` branch from the `curl/curl` repository.)

zajdee:
Static build works as expected. I will investigate the dynamic build again later today (to rule out potential library conflicts on my side).

```
autoreconf -fi; ./configure --with-secure-transport --disable-shared; make
```

(Note the IPv6 address being used instead of the IPv4 one. This is a result of the NAT64/DNS64-enabled environment w/ NAT64 prefix fd00:64::/96 and CopyProxies being correctly called by the code.)
```
$ curl -v https://1.1.1.1/
*   Trying fd00:64::101:101:443...
* Connected to 1.1.1.1 (fd00:64::101:101) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: C=US; ST=California; L=San Francisco; O=Cloudflare, Inc.; CN=cloudflare-dns.com
*  start date: Jan 12 00:00:00 2023 GMT
*  expire date: Jan 11 23:59:59 2024 GMT
*  subjectAltName: host "1.1.1.1" matched cert's IP address!
*  issuer: C=US; O=DigiCert Inc; CN=DigiCert TLS Hybrid ECC SHA384 2020 CA1
*  SSL certificate verify ok.
```

stanhu:
Thanks for confirming.

I repeated your steps and didn't run into the missing symbol issue.

stanhu:
@bagder Is there anything I can do to get this merged? Thanks so much for all your work on `curl`.

zajdee:
Hi @stanhu,
just to confirm, after cleaning up my environment even dynamic builds work as expected. Thank you.

@bagder I support the merge of this change.

bagder (approved):

bagder:
Thanks!

PR #11265, opened 2023-06-07
tests: fix error handling around UNIX sockets

### LINKED ISSUE #11258: Native Windows can't create UNIX sockets
Many CI builds on Windows show warnings like this during CI runs, such as this one from the Appveyor *CMake, mingw-w64, Debug x86, Schannel, Static* build:
```
RUN: failed to start the HTTP-unix server
== Contents of files in the log/ dir after test 1435
=== Start of file http_unix_server.log
 06:39:33.828000 Error creating socket: (10047) Unknown error
 06:39:33.828000 ========> sws quits
=== End of file http_unix_server.log
test 1435 SKIPPED: failed starting HTTP-unix server
```
However, msys2 builds on Appveyor like *autotools, msys2, Debug, no SSL* run the test fine (I don't know if this is because msys2 does some kind of emulation for unix sockets that doesn't happen in native builds).  Cirrus msys2 builds like *Windows 64-bit shared/release Schannel/SSPI/WinIDN/libssh2* also show the failure. The failures have been happening for at least a couple of months on Appveyor. I don't know if those non-msys2 builds have ever worked.

Because the test server works around it by skipping the tests, it could have easily stopped working at some point without it being noticed.

According to this [MS blog post](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/) Windows has supported `AF_UNIX` for more than 4 years, so I would expect this to be working on all Windows builds these days, assuming the CI services are using a modern-enough Windows.

dfandrich:
The wrong error code was checked on Windows, which may have caused all
UNIX sockets to be reported as error and the tests skipped.

Fixes #11258
Closes #11265

### DIFF
--- a/tests/runtests.pl
+++ b/tests/runtests.pl
@@ -2779,6 +2779,7 @@ while () {
     # a new one shortly. If all runners are busy, wait indefinitely for one to
     # finish.
     my $runnerwait = scalar(@runnersidle) && scalar(@runtests) ? 0 : undef;
+print STDERR "TESTING waiting indefinitely for test completion\n" if(!scalar(@runtests));
     my $ridready = runnerar_ready($runnerwait);
     if($ridready) {
         # This runner is ready to be serviced
--- a/tests/server/mqttd.c
+++ b/tests/server/mqttd.c
@@ -771,7 +771,7 @@ static bool incoming(curl_socket_t listenfd)
       if(CURL_SOCKET_BAD == newfd) {
         error = SOCKERRNO;
         logmsg("accept(%d, NULL, NULL) failed with error: (%d) %s",
-               sockfd, error, strerror(error));
+               sockfd, error, sstrerror(error));
       }
       else {
         logmsg("====> Client connect, fd %d. Read config from %s",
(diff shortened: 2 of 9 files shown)


dfandrich:
@gvanem  Would you mind taking a look?

gvanem:
@dfandrich No change. Same  `Error binding socket: (10045) Unknown error`. 
Some deeply hidden issue in `afunix.sys`?

dfandrich:
I'm not sure which versions of Windows is actually running on the CI services.
They might be running an older one that doesn't support UNIX sockets. In any
case, I'll submit this PR since the previous error checking was wrong.

tests/runtests.pl
```diff
@@ -2779,6 +2779,7 @@ sub displaylogs {
     # a new one shortly. If all runners are busy, wait indefinitely for one to
     # finish.
     my $runnerwait = scalar(@runnersidle) && scalar(@runtests) ? 0 : undef;
+print STDERR "TESTING waiting indefinitely for test completion\n" if(!scalar(@runtests));
```
bagder, line 2782: Is this print a debugging leftover or meant to actually be a plain STDERR output? It looks like an anomaly in the code.

dfandrich: This is just a debugging leftover. It won't be submitted.

PR #11271, opened 2023-06-08
Update tls13-ciphers.d to include Schannel.

wyattoday:
This was an oversight when I added TLS 1.3 support for Schannel (https://github.com/curl/curl/pull/8419 ).


bagder (approved):

bagder:
Thanks!

PR #11275, opened 2023-06-08
CI: fixup: future macos xcode issues for cmake gcc builds

pheiduck:
New way to fix: https://github.com/curl/curl/issues/10356

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -39,9 +39,6 @@ concurrency:
 
 permissions: {}
 
-env:
-  DEVELOPER_DIR: /Applications/Xcode_14.0.1.app/Contents/Developer
-
 jobs:
   autotools:
     name: ${{ matrix.build.name }}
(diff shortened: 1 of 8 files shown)


pheiduck:
Works as exacted, we need to solve the Copyright thing on reuse.

pheiduck:
@bagder, @icing, @jay, @mback2k and @dfandrich can someone take a look?
Also give me hints, where I can place the patches?
Not sure if they can be used on the makefile?

jay:
> New way to fix: #10356

This is a lot of code. Why is the current fix not enough? There is no context here for why all this code is necessary. What exactly causes the issue?

pheiduck:
@jay I don't know what exactly triggers this error, but it is not possible to build curl with cmake and gcc on recent xcode versions 14.1 < 14.3
I have locally examined the files and fixed the places that trigger the error.
Among other things, the error is present from 14.1 to 14.3.

pheiduck:
@bagder I have opened an issue be apple directly.

pheiduck:
@bagder
Apple told me that gcc causes the error and it is not Apple's fault they only support clang.

PR #11288, opened 2023-06-09
 HTTP/3, curl_ngtcp2, using ever increasing timestamp in io

icing:
- ngtcp2 v0.16.0 asserts that timestamps passed to its function
      will only ever increase.
- Use a context shared between ingress/egress operations that
      uses a shared timestamp, regularly updated during calls.
- based on #11184 that switched ngtcp2 and family version numbers

### DIFF
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -327,7 +327,7 @@ static void quic_settings(struct cf_ngtcp2_ctx *ctx,
   t->initial_max_streams_uni = QUIC_MAX_STREAMS;
   t->max_idle_timeout = QUIC_IDLE_TIMEOUT;
   if(ctx->qlogfd != -1) {
-    s->qlog.write = qlog_callback;
+    s->qlog_write = qlog_callback;
   }
 }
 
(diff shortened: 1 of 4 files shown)


lib/vquic/curl_ngtcp2.c
```diff
@@ -246,10 +246,37 @@ static void h3_data_done(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+  struct Curl_cfilter *cf;
+  struct Curl_easy *data;
+  ngtcp2_tstamp ts;
+  size_t pkt_count;
+  ngtcp2_path_storage ps;
+};
+
+static ngtcp2_tstamp timestamp(void)
+{
+  struct curltime ct = Curl_now();
```
piru, line 259: `Curl_now()` isn't guaranteed to always go forwards, in some situations it can jump back. For most modern platforms it is monotonic, however.

icing: The addition is a move, since the function is needed earlier in the code.

What is the alternative? Any recommendation?

piru: Unfortunately guaranteed monotonic clock adds platform dependency. There is no fully portable and widely available way to get truly monotonic clock.

Traditionally `clock_gettime(CLOCK_MONOTONIC, ..)` had been considered to be only increasing time stamp. However, for example macOS `CLOCK_MONOTONIC` isn't always monotonic and can actually jump back.

Depending on what is needed `CLOCK_MONOTONIC_RAW` might be suitable. This of course isn't available generally everywhere.

bagder:
We can merge this and work on the clock independently, I created #11291

PR #11291, opened 2023-06-09
timeval: use CLOCK_MONOTONIC_RAW if available

bagder:
Reported-by: Harry Sintonen
Ref: #11288

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -1072,6 +1072,38 @@ AC_DEFUN([CURL_CHECK_FUNC_CLOCK_GETTIME_MONOTONIC], [
   dnl until library linking and run-time checks for clock_gettime succeed.
 ])
 
+dnl CURL_CHECK_FUNC_CLOCK_GETTIME_MONOTONIC_RAW
+dnl -------------------------------------------------
+dnl Check if monotonic clock_gettime is available.
+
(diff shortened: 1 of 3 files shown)


dfandrich (commented):
Theoretically, CLOCK_MONOTONIC_RAW could exist and be found by autoconf but the system call could fail, while CLOCK_MONOTONIC would succeed. On Linux, this would happen if run on really old kernels (like 2.6.27) so it's probably not worth worrying about there, but what other systems support CLOCK_MONOTONIC_RAW? If there are others, or there could be others in the future, it might be worth calling both, with CLOCK_MONOTONIC only if CLOCK_MONOTONIC_RAW fails.

bagder:
macOS supports CLOCK_MONOTONIC_RAW. I don't know since which version...

bagder:
> Theoretically, CLOCK_MONOTONIC_RAW could exist and be found by autoconf but the system call could fail, while CLOCK_MONOTONIC would succeed

What weirdness would cause that to happen?

dfandrich:
Building on a recent kernel but running on an old one.

bagder:
@dfandrich what do you think of this approach?

dfandrich (approved):
Looks good to me. This should work fine now even on old (possibly ancient?) systems.

PR #11295, opened 2023-06-10
Update HTTP3.md and change the supported ngtcp2 and nghttp3 versions

Karthikdasari0423:


dfandrich:
Thanks for catching this!

PR #11299, opened 2023-06-11
docs: Can't typedef bool in c23

crrodriguez:
bool is a reserved keyword in c23.
Reserved keywords must not be redefined or UB ocurrs.

### DIFF
--- a/docs/examples/multi-debugcallback.c
+++ b/docs/examples/multi-debugcallback.c
@@ -36,7 +36,10 @@
 /* curl stuff */
 #include <curl/curl.h>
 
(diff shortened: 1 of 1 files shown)


docs/examples/multi-debugcallback.c
```diff
@@ -36,7 +36,10 @@
 /* curl stuff */
 #include <curl/curl.h>
 
+#if defined __STDC_VERSION__ && __STDC_VERSION__ > 201710L
```
bagder, line 39: Did you perchance mean `<= 201710L` ? 201710L seems to be the value that [identifies c18](https://github.com/cpredef/predef/blob/master/Standards.md)

PR #11300, opened 2023-06-11
docs: update CURLOPT_UPLOAD.3

oleg-jukovec:
The behavior of CURLOPT_UPLOAD differs from what is described in the documentation. The option automatically adds the
'Transfer-Encoding: chunked' header. It also ignores 'Content-Length' header at CURLOPT_HTTPHEADER.

The patch makes the documentation clearer.


bagder:
> It also ignores 'Content-Length' header at CURLOPT_HTTPHEADER

It *always* ignores that, in the sense that it doesn't read it nor parses it but just passes it along.

bagder (approved):

bagder:
Thanks!

PR #7666, opened 2021-09-03
curl: add --pass-file and --proxy-pass-file options

vszakats:
These two options have the same purpose as existing `--pass` and `--proxy-pass` options, but instead of expecting the secret string on the command-line, they allow them to be passed via a filename. This filename can point to an on-disk file, to STDIN, etc. This makes it possible to pass key passphrases securely to curl, without them appearing in the `ps` listing, command history, etc.

The passphrase is read from the filename as-is, up to the first zero byte or CR, LF characters.

Examples:
```shell
echo 'secret' | curl --user my-user: sftp://example.net/ \
  --key id_ed25519 --pubkey id_ed25519.pub --pass-file /dev/stdin

echo 'secret' | curl https://example.net/ \
  --cert-type P12 --cert client.p12 --pass-file /dev/stdin
```

Issue #6089, opened 2020-10-15
curl does not use encoding of CLI for Basic Authorization on Windows


c-kunz:
# I did this

```bash
chcp 936
curl -v -u X:Ü https://...
```

### I expected the following

Authorization header contains Base64 encoded credentials in the encoding of the codepage 936:

`Authorization: Basic WDo=`

### I've got instead:

`Authorization: Basic WDrDnA==`

This seems to be `UTF-8` encoded instead, despite the command line using CP936.

### curl/libcurl version

`curl --version`:

```
curl 7.73.0 (x86_64-pc-win32) libcurl/7.73.0 OpenSSL/1.1.1h (Schannel) zlib/1.2.11 brotli/1.0.9 zstd/1.4.5 WinIDN libssh2/1.9.0 nghttp2/1.41.0
Release-Date: 2020-10-14
```

### operating system

Windows 10 Version 2004 (Build 19041.450)

Happens when curl is invoked within `cmd.exe` as well as `PowerShell`.

### Hint

On Ubuntu 20.04 `curl` seems to always immediately respect the encoding of the terminal, which is what I would also expect to happen on Windows.

My Ubuntu 20.04 curl is:

```
curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3
Release-Date: 2020-01-08
```

bagder:
curl just uses what it gets as is. There's no conversion or encoding involved.

mkarg:
> curl just uses what it gets as is. There's no conversion or encoding involved.

Why does the conversion happen on Ubunto then?

c-kunz:
Indeed, I found out yesterday that curl in the following version on the same Windows machine:

```
curl 7.55.1 (Windows) libcurl/7.55.1 WinSSL
Release-Date: 2017-11-14, security patched: 2019-11-05
```

is using `ISO-8859-1` (CP850) to encode the credentials. This got me confused as well when upgrading to the latest curl which suddenly uses UTF-8, but nonetheless never respects the active codepage of the CLI.

MarcelRaad:
Probably caused by https://github.com/curl/curl/commit/9e5669f3880674de262000e772ef21ec0cc874be.

[...]

PR #11311, opened 2023-06-13
tool_parsecfg: support environment variables in config files

bagder:
Instances of `{{name}}` are replaced by the contents of the environment variable `name` at the time of the parsing. **Iff** the line is prefixed with `[expand]`.

The name must be shorter than 128 bytes and can only be `[A-Za-z0-9_]+`.

- [X] tests
- [X] documentation


emanuele6:
There should at least be a way to use a literal `$` in a parameter, maybe `\$`.

I think that programs that generate configs for curl should already be careful when generating configs with arguments that contain `\` or `"` or whitespace etc, so i don't think having to add a backslash also before `$` would be more complicated. The only concern may be that `\$` and `\$` will be interpreted differently in different versions, but I think config generators should already be replacing `\$` with `\\$`, so that is much of a problem in my opinion.

olifes-c:
Duplicate of #

bagder:
@emanuele6 good points indeed. Now we can specify `\${name}` to get `${name}` inserted without expansion.

vszakats:
To avoid undesired and potentially insecure macro expansions (leaking local env content to a remote system), we might consider enabling it only in the presence of a new option in the config file: `--allow-macros`, or something similar.

vszakats:
To cite real-life examples: It is recommended to never pass secrets via the command-line. Let's say a secret must be passed in the URL. Our only option is to pipe an `--url` option via a custom config like this:
```
exec <<EOF
--url "https://example.com/api/${secret_key}/method"
EOF
curl --config /dev/stdin
```

With this PR, this call becomes broken if the secret key happens to contain text like: `${...}`. If a malicious actor can force the value `${USER}` to appear in the secret, the local username is exfiltrated to the remote system (this is "just" PII. It can be worse.) `--config` is very flexible, and it's not unimaginable someone using `--user` or `--pass` with a user supplied username or password, which is similary problematic or perhaps even more common.

emanuele6:
This is a huge stretch. That code was already very broken. If the attack vector is that `$secret_key` you are unsafely injecting in the curl config, it was already possible to run to run with a `secret_key` that contains `"` and inject any arbitrary option, now there are only slightly more interesting things an attacker could inject..

bagder:
Thanks for the thoughtful comments.

A slight alternative take would be to require something different syntax-wise in the config line to enable the environment variable expansion feature, so that old config files will not expand and only those explicitly written for it, using the new thing, will.

**Brainstorming mode**. Like if we would prefix the line with something that can never be there now. Like a `$` letter:

`$url = "https://example.com/api/${secret_key}/method"`

or more verbose, prefix the option with `expand:`:

`expand:url = "https://example.com/api/${secret_key}/method"`

Or, the `--allow-macros` idea, but maybe only accept it *within* the config file itself. Maybe call it "expand-variables" ?

~~~
expand-variables
url = "https://example.com/api/${secret_key}/method"
~~~

vszakats:
It could even be a new, so far invalid option prefix, e.g. `---<option>`, or `--envexpand-<option>`, or `--expand-<option>`. It'd avoid using `$` because that is already overused.

vszakats:
@emanuele6 Not stricly on-topic, but you're welcome to suggest a better alternative. Being broken doesn't mean we should further expand on it. I'd rather welcome a safe way to pass secrets without touching the command line.

bagder:
It could be something that stands out:
~~~
[expand] url = "https://example.com/api/${secret_key}/method"
~~~

emanuele6:
> Being broken doesn't mean we should further expand on it.

@vszakats Exactly, you said this is a real-life problem. If you are concerned `$secret_key` is a possible attack vector, or possibly can contain `"` or `\`, you should definitely be informed about the fact your code is greately vulnerable, and you should fix it..

Unfortunately though, if you are concerned about that your solution is broken in so many ways.

One may think that it would be possible to prepare the string for injection in few steps using e.g. in bash you could use `tmp=${secret_key//\\/\\\\}` `${tmp//\"/\\\"}`, or directly `${secret_key//[\\\"]/\\&}` since bash 5.2 to prepend a `\\` in front of every backslash and double quote to prevent expansion.

But unfortunately, the format used by `--config' is way more complex and fiddly than that..

Consider:
```
curl -K - << 'EOF'
--url http://example.org/
--output /dev/null
--form "foo=abc
xyz hello"
EOF
```

Curl will not treat the literal newline between the quotes the same as it would treat `\n`, it will instead run `--form "foo=abc"` and print a warning for the `xyz hello"` line. If instead of that `xyz hello"` line you have `--form bar=boo"` it will also interpret `--form "bar=boo\"".

So you must also replace all literal newlines in your variable with `\n` and all carriage return with `\r`. 

In the end, you will need something like `tmp=${secret_key//\\/\\\\} tmp=${secret_key//\\/\\\"} tmp=${tmp//$'\n'/\\n} tmp=${tmp//$'\r'/\\r}` and then injecting `$tmp` in the curl config *"should"* be safe.

I think it's just a lost cause trying to use `-K` with a pipe heredoc (another note, heredocs are not pipes for all shells, and all versions of bash, even if you use `<< EOF` to pass the secret, it may be backed by a temporary regular file that will be accessible from disk.) to not pass secrets through the command line. (arguments and environment variables.) 

If we want to allow something like that, the `-K` syntax should probably be reworked a bit to allow simpler injections of arbitrary data.

---

I also want to make a suggestion regarding the new `${foo}` syntax.

Since it `${foo}` is not backwards compatible, and adds a new special meanings for `\$` and `$`, I don't think it is a good idea to use `$` as special sequence for those expansions.

I think using `\{foo}` instead would be more sensible.

If you want to have literal `\{foo}`, you use `\\{foo}` just like in previous versions of curl; and the previously unspecified `\{foo}`, now expands the value of the `foo` environment variable.

Instead of adding a new special symbol, it just reuses the `\` special symbol, and give meaning to `\{`; it will make things "simpler" to prepare for injection and it will be backwards compatible.

vszakats:
@emanuele6 Setting aside from my specific use-case, this is the generic problem of passing a secret (safely) to curl without using the command-line. Because passing through the command-line is known to be insecure and discouraged. I think it's useful that we're digging into this, thank you for your research. To me it looks like this isn't reasonably possible to do safely at this point, because `tmp=${secret_key//\\/\\\\} tmp=${secret_key//\\/\\\"} tmp=${tmp//$'\n'/\\n} tmp=${tmp//$'\r'/\\r}` doesn't seem like something we can expect from most users. I surmise very few if any users do this, and it's likely still not 100% safe. (Not to mention non-Unix shells.)

My point is that the original version of this PR opens up even more (I'd argue even worse) ways to inadvertently mess this up.

Expanding on @bagder's suggestion, we could as well allow a similar, but non-expanded, line-based syntax:
```
url = "https://example.com/api/${secret_key}/method"
```
where curl ensures that any `${secret_key}` value (substitued by the caller in this scenario) would end-up in the value as-is  (without adding new arguments or options to the command-line).

We may also revisit #7666, where I proposed `--pass-file` and `--proxy-pass-file` (though this solution isn't trivially applicable to `--url`), then dropped it in favour the `--config` suggestion.

bagder:
> I think using \{foo} instead would be more sensible.

We already support `{one,two,three}` etc globbing in URLs though, so `{one}` is a list of one (unless globbing is switched off). So even if `\{one}` is not identical and perhaps could be differentiated in code, I think they look too similar to a human eye.

vszakats:
I think `\{}` also present a lot of room for confusion, as the backlash is interpreted all over the place depending on platform/env, quoting, shell flavour. I don't see an issue with `${}` (as in `[expand] url = "https://example.com/api/${secret_key}/method"`), but if we opt using something else, perhaps `%{}` would be a better fit, or `{{}}` even.

bagder:
> perhaps %{}

We use that style already for the `-w` argument, so if you want to expand an environment in that string, it would be tricky if they use the same syntax:

`[expand] write-out = "%{response_code} %{HOME}`

Maybe `%ENV{}` ?

emanuele6:
> My point is that the original version of this PR opens up even more (I'd argue even worse) ways to inadvertently mess this up.

You keep insisting that giving access to environment variables is the worst thing ever, but this doesn't make much sense to me. How is it that bad? I don't understand what you think it is so much worse than being able to inject arbitrary options to make any a request to any site, or download a file to any path that may be auto executed as it is already possible in your example code.   

---

Also I just read the edited in part of your original post and I want to mention that your code is not correct.

```sh
exec <<EOF
--allow-envvars
--url "https://example.com/api/${SECRET_KEY_FROM_ENV}/method"
--pass "${password}"
EOF
curl --config /dev/stdin
```

That does not make sense because you are using an unquoted `EOF`, so `${SECRET_KEY_FROM_ENV}` and `${password}` are pre-injected by the shell, not by curl.


What you meant is probably `<< 'EOF'`, and in that case, the config is injection free, and fine. But it does not make any sense that you would have environment variables that contain secrets.

```sh
exec <<'EOF'
--allow-envvars
--url "https://example.com/api/${SECRET_KEY_FROM_ENV}/method"
--pass "${password}"
EOF
curl --config /dev/stdin
```

If you were ok with having enviornment variables with secrets in them, then you would just be able to run:

```sh
SECRET_KEY=$secret_key curl --config - << 'EOF'
--url https://example.com/api/${SECRET_KEY}/method
EOF
```

..with the new `${}` syntax and all would be fine.

Of course, that is not good, because environment variables are part of the command line just like arguments, and can be seen by any process.

The whole idea of making a process read credentials from a *pipe* file is that the content of *pipe* files can only be seen by the kernel and by the first process that reads it. So **if we assume** that curl will be the first process to read the credentials will be curl, the credentials will only be seen by curl, and not other processes. So that is why a injection is wanted, and you can't just use environment variables for this.

If your fear is that an attacker may get some of your other secrets because it has access to enviornment variables, I don't think that is a scenario that makes much sense since the whole point of you using that vulnerable code was not to pass secrets to the command line (environment variables and arguments)...

---

> Setting aside my specific use-case, this is the generic problem of passing a secret (safely) to curl without using the command-line. Because passing through the command-line is known to be insecure and discouraged. I think it's useful that we're digging into this, thank you for your research. To me it looks like this isn't reasonably possible to do safely at this point, because `tmp=${secret_key//\\/\\\\} tmp=${secret_key//\\/\\\"} tmp=${tmp//$'\n'/\\n} tmp=${tmp//$'\r'/\\r}` doesn't seem like something we can expect from most users. I surmise very few if any users do this, and it's likely still not 100% safe. (Not to mention non-Unix shells.)

Also again, you are assuming that the heredoc will be backed by a *pipe* file, but in reality most shells use a temporary regular file for that. And even bash has only started using pipes instead of regular files, and only for small heredocs, since a few recent versions. So the file would be accessible to the anyone while the shell is running.

An example of an easily accessible shell that uses temporary regular files for here docs is mksh (again bash will also do it if the file is reasonable enough. It does not choose to use a pipe for \*security\* reasons, it is just a optimisation, and you cannot expect the heredoc to be a pipe).

Let me show you that this approach will make it write the secret in clear text to a temporary file.

```
bash-5.1$ mksh
$ curl -K /dev/fd/3 3<< 'EOF'
> --form secret=wooo
> --url example.org
> --next
> # using -Ffile=@- to hang curl so you can check /proc/pid/fd/3
> --url example.org
> --form file=@-
> EOF
<hanging>
```

```
bash-5.1$ pidof curl
1003490
bash-5.1$ cat /proc/pid/1003490/fd/3
--form secret=wooo
--url example.org
--next
# using -Ffile=@- to hang curl so you can check /proc/pid/fd/3
--url example.org
--form file=@-
bash-5.1$ cat /proc/pid/1003490/fd/3
--form secret=wooo
--url example.org
--next
# using -Ffile=@- to hang curl so you can check /proc/pid/fd/3
--url example.org
--form file=@-
bash-5.1$
```

You must use `... | curl -K -` where `...` is shell code that uses shell builtins to print the desired content for the config file. You cannot/should not expect `<< EOF` to create a pipe, if you want a pipe you should be explicit.

Also if you are worrying about portability `echo`/`printf` are not builtins in all shells, so if the shell calls out to /bin/echo or /bin/printf the secrets will be briefly viewable as arguments to those processes.

So the entire approach you are using is flawed for many reasons already.

Instead of arguing against implementing this new feature, it would make more sense to advocate for a new suggested standardised way to pass secrets to curl safely, your problems are not related to the addition of this new features.

Sorry, if I insist, but really I don't think this features makes injections that much worse than they already are (as described above your attack scenario doesn't make much sense to me), and I don't think your complain is much related to the new feature.

Even if `curl` adds `[expand] ...` or `--expand-url` stuff, your code is still broken, that does not help you.

vszakats:
@bagder `%ENV{}` is fine, though a bit verbose. I'd give `{{}}` a second chance.

Weering off slightly from the brainstorm track, this would be a good way to support passing hex-encoded values:
```
[hex] pass = 01020304ff
```
(we have no option to do this with the curl tool at the moment. It can be useful to avoid codepage/encoding ambiguities #6089, or for binary password.)

vszakats:
@emanuele6 "You keep insisting that giving access to environment variables is the worst thing ever" please don't invent things that I didn't say!

emanuele6:
> @emanuele6 "You keep insisting that giving access to environment variables is the worst thing ever" please don't invent things that I didn't say!

You literally said "even worse".

vszakats:
@emanuele6 We disagree. I honestly don't understand why do you need to personally attack me for raising concerns. It certainly does make things worse.

I'm out of this discussion.

vszakats:
'even worse' != 'the worst thing ever'

emanuele6:
I was not personally attacking you.

I just want to insist that the issue is unrelated and only discussing not adding this feature will not solve the problem you are bringing up of users unsafely injecting data in config files.

And I don't think it makes sense to claim that having access to environment variables from curl config files suddenly makes the issue worse than it is. There should not be secrets in the environment variables so even if the attacker can access them it should not be much worse, if it is the problem is that you have secrets in your environment variables. With that example injection they can already replace a bashrc or similar file that is autorun..

In the mean time, I also mentioned that using `<< EOF` will not always pass the secrets through a pipe, and may cause the shell to generate a temporary file that will be accessible for the entire runtime of the curl process (shell's entire runtime if you use `exec << EOF` instead of just `curl << EOF`), so I told you, if you want to do something like this in general for other programs that are not curl you should be using `printf %s\\n "$hi" | foo` (ensuring that printf is a builtin command of the shell), and **not** `foo << EOF $hi EOF`. I thought it was important to mention it. And then I also pointed out the other problems I could see with the approach.

vszakats:
Accessing envvars is useful, no question about that, my point is adding this safely and without breaking exising uses. Setting aside security, the original proposal may break existing uses. The standard practice is to make such features opt-in. That's what I was proposing.

I wanted to focus on that, not on second-guessing, mis-representing and defending my intentions and dissecting these tiny examples to the point that makes no sense in this context. It makes no difference in these examples if we use `<<` or `|`. The point remains the same. Your posts come across as trying to convince me that every single bit of the shell and OS is insecure anyway, so there is not point in trying to do it right (and making it sure to point out that I'm personally doing it wrong, which is absolutely besides the point). FWIW in my particular (bog standard) envs `printf`/`echo` are builtins. Most of the time stdin is already used for passing data, so it's not always an option to use to pass secrets. Sometime there are multiple secrets that need to be passed, so stdin is not enough (not necessarily with curl though, but in such case `--config` is the only workaround with curl).

We're not here to fix the OS, or my examples, but to avoid the gotchas we can avoid within curl itself while adding a new feature.

Your argument seems to be to just add yet another insecure (but cool!) feature without giving it even a second thought, because everything is broken anyway. I don't agree with this approach and would be happier to see curl doing what it can to avoid these (or more of these) problems.

bagder:
# Updates

It now uses `{{syntax}}` and requires the line to be prefixed with `[expand]` for the environment variable expansion to happen. This instruction requirement prevents old config files from unexpectedly and accidentally get expansions. This way we have also created a new way to do config specific instructions, with expand being the first one.

When using `[expand]`, `\{{` will insert `{{` verbatim.

@vszakats mentioned a `[hex]` idea up there, but I didn't quite understand that.

bagder:
@vszakats: @emanuele6 and me brain-stormed away even further and we have crafted a slightly different take than what this PR does, which I feel might have even more merit and power. See the brainstorm wiki page for it here: https://github.com/curl/curl/wiki/command-line-variables

I bet we can polish that a little further

vszakats:
Thank you @bagder and @emanuele6. I trust you getting this perfectly right and the wiki is looking solid to me.

I'm guessing reading a variable from a pipe would look like this: `--variable secret@/dev/fd/3` (understanding that this is also not secure in certain cases)

On `hex` (disclaimer: I don't have the necessary time/energy/focus to present this in a "flawless" manner; only informally or sloppyly even). Sticking with passwords, but it may apply to any other string that may contain raw or binary (non-ASCII) strings (ref: https://github.com/curl/curl/wiki/libcurl-and-expected-string-encodings). Some may be using the password "ŰÈ" and expect it to work when passed to a server. Depending on how this was set, stored or verified on the server, the actual expected password might use UTF-8, UTF-16, a certain 8-bit codepage, etc. To get it right, the client must match this when sending the password over the wire. curl tool doesn't give total control over that, and sometimes (→ Windows Unicode builds) may alter the bytes making up the password string. By adding the ability to pass an exact stream of bytes, this can be worked around without worrying about any kind of encoding change happening in the background.

For example, this:
```
curl -v -u X:Ü https://...
```
could be written as (or similar):
```
curl -v -u '{{583adc:hex}}'
```
making sure that `Ü` is represented by its cp1252 codepoint for a server that expects that.

I'm guessing there is also a security angle here where this feature should be limited to certain values (e.g. "raw" and "binary" ones as in the codepage wiki page), to avoid sticking controls codes or newlines into values that don't normally hold those.

[ Detouring back to passing secrets via stdin / heredoc / pipe, after examining existing code, I realized that I avoided `printf '%s' "${var}" | curl ...` method because it leaks secrets to stdout with `set -x`, while heredoc doesn't. This may be the wrong solution or there may be solutions that fix both, but I did not find them. ]

bagder:
The #11346 is a different take that I think is better and covers more use cases. It implements what is being worked out in the wiki page: https://github.com/curl/curl/wiki/command-line-variables

PR #11321, opened 2023-06-16
lib: fix some format specifiers

jay:
- Use CURL_FORMAT_CURL_OFF_T where %zd was erroneously used for some curl_off_t variables.

- Use %zu where %zd was erroneously used for some size_t variables.

Prior to this change some of the Windows CI tests were failing because in Windows 32-bit targets have a 32-bit size_t and a 64-bit curl_off_t. When %zd was used for some curl_off_t variables then only the lower 32-bits was read and the upper 32-bits would be read for part or all of the next specifier.

Closes #xxxx

---

The CI failures are since e024d566. [Example](https://cirrus-ci.com/task/5617442999762944?logs=test#L2503):

~~~diff
-* Connection #3 to host server4.example.com left intact[CR][LF]
+* Connection #3 to host (nil) left intact[CR][LF]
~~~

Although there were a bunch of others it's not noticeable because there are no additional specifiers that cause the upper 32-bits to be read.

I did all of them manually since I don't get any warnings in the CI or my VS compiler which makes it hard to figure out where they all are. There might be a switch that will do it but I couldn't find it.

/cc @icing

### DIFF
--- a/lib/cf-h2-proxy.c
+++ b/lib/cf-h2-proxy.c
@@ -449,7 +449,7 @@ static CURLcode proxy_h2_progress_ingress(struct Curl_cfilter *cf,
 
   /* Process network input buffer fist */
   if(!Curl_bufq_is_empty(&ctx->inbufq)) {
-    DEBUGF(LOG_CF(data, cf, "Process %zd bytes in connection buffer",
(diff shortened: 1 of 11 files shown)


bagder (approved):
:+1: good job!

MarcelRaad (approved):

icing (approved):

PR #11329, opened 2023-06-17
Improve test and test harness reliability

dfandrich:
This is done by tweaking some tests and improving logging, and making the test
harness more careful about detecting problems with pipe operations. The latter
makes the test harness less likely to hang, especially on ^C.

### DIFF
--- a/tests/data/test1007
+++ b/tests/data/test1007
@@ -30,12 +30,14 @@ This data will not be sent
 <errorcode>
 69
 </errorcode>
+<strip>
+^timeout = [5-6]$
+</strip>
 <protocol>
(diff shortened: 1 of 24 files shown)


bagder:
An alternative to all those test filter updates would be to change the test TFTP server to use a fixed timeout value. I don't think the current logic that sometimes set it to 6 seconds instead of 5 actually does much good.

dfandrich:
As discussed offline, this timeout is set client-side and merely logged by the server. I didn't want to change all the client-side timeouts if possible since that's not really the point of these tests, but I did in test 285 just to expand test coverage.

PR #11330, opened 2023-06-17
http/2, workaround for stalled uploads

icing:
- refs #11242 where CPU busy loop is reported. Removed drain setting
   that caused this
- discovered stalled uploads using the server from #11242 on my slow
   internet
- added workaround on flushing the last upload chunk that, if not
   written fully, makes the transfer stall. This is not the fix we
   are looking for.

## Problem description

There is unavoidable buffering with HTTP/2 uploads. Even if we remove all our own buffers, nghttp2
still buffers the DATA frame if sending it would block.

The problem with buffered upload data arises on the very last chunk. If that write reports success, the transfer loop stops monitoring the connection for output and waits for a server reply. This will not come until the server times out the upload and RSTs the stream.

## Update

This PR is not it. I will make a new attempt next week. With the current transfer handling in curl, filters HAVE TO make sure date gets written out to the network or EGAIN the send call.

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -134,6 +134,7 @@ struct cf_h2_ctx {
   BIT(conn_closed);
   BIT(goaway);
   BIT(enable_push);
+  BIT(nw_out_blocked);
 };
 
 /* How to access `call_data` from a cf_h2 filter */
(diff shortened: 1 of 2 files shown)


lib/http2.c
```diff
@@ -2139,23 +2184,61 @@ static ssize_t cf_h2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
...
     DEBUGF(LOG_CF(data, cf, "[h2sid=%d] cf_send(len=%zu) -> %zd, %d, "
-                  "buffered=%zu, upload_left=%zu, stream-window=%d, "
-                  "connection-window=%d",
+                  "buffered=%zu, upload_left=%" CURL_FORMAT_CURL_OFF_T
+                  ", stream-window=%d, "
+                  "connection-window=%d, stream_send_buffer(%zu), "
+                  "nw_send_buffer(%zu)",
                   stream->id, len, nwritten, *err,
                   Curl_bufq_len(&stream->sendbuf),
                   (ssize_t)stream->upload_left,
```
jay, line 2227: the typecast needs to be removed if you are using CURL_FORMAT_CURL_OFF_T
```suggestion
                  stream->upload_left,
```

jay:
> Even if we remove all our own buffers, nghttp2 still buffers the DATA frame if sending it would block.

Is there some kind of signal from nghttp2 for this, wouldn't it return eagain or whatever its equivalent?

icing:
Closing and creating a new one with the real fix.

PR #11344, opened 2023-06-19
Unslowing hyper

icing:
- refs #11203 where hyper was reported as being slow
- fixes hyper_executor_poll to loop until it is out of tasks as advised by @seanmonstar in https://github.com/hyperium/hyper/issues/3237
- added a fix in hyper io handling for detecting EAGAIN
- added some debug logs to see IO results
- pytest http/1.1 test cases pass
- pytest h2 test cases fail on connection reuse. HTTP/2 connection reuse does not seem to work. Hyper submits a request on a reused connection, curl's IO works and thereafter hyper declares `Hyper: [1] operation was canceled: connection closed` on stderr without any error being logged before.

### DIFF
--- a/lib/c-hyper.c
+++ b/lib/c-hyper.c
@@ -71,9 +71,13 @@ size_t Curl_hyper_recv(void *userp, hyper_context *ctx,
   DEBUGASSERT(conn);
   (void)ctx;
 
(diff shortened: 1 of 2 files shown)


bagder (approved):
:+1:

gvanem:
@bagder You wrote you could confirm this slowness. But is it still present for you with this change?

PR #11346, opened 2023-06-19
tool: add command line variables

bagder:
Support command line variables. Set variables with `--variable name=content` or `--variable name@file` (where "file" can be stdin if set to a single dash (-)).

Variable contents can be expanded in option parameters using `{{name}}"` if the option name is prefixed with `--expand-`. This gets the contents of the variable `name` inserted, or a blank if the name does not exist as a variable. Insert `{{` verbatim in the string by prefixing it with a backslash, like `\{{`.

You an access and expand environment variables by first importing them. You can select to either require the environment variable to be set or you can provide a default value in case it is not already set. Plain `--variable %name` imports the variable called `name` but exits with an error if that environment variable is not already set. To provide a default value if it is not set, use `--variable %name=content` or `--variable %name@content`.

Example: get the `USER` environment variable into the URL. Fail if USER is not set:

    --variable '%USER'
    --expand-url = "https://example.com/api/{{USER}}/method"

When expanding variables, curl supports a set of functions that can make the variable contents more convenient to use. It can trim leading and trailing white space with `trim`, it can output the contents as a JSON quoted string with `json` and it can URL encode the string with `urlencode`. You apply the function(s) to a variable expansion, add them colon separated to the right side of the variable. They are then executed in a left to right order.

Example: get the contents of a file called $HOME/.secret into a variable called "fix". Make sure that the content is trimmed and percent-encoded sent as POST data. if HOME is not set, use "dummy" as default:

    --variable %HOME=dummy
    --expand-variable fix@{{HOME}}/.secret
    --expand-data "{{fix:trim:urlencode}}"
    https://example.com/

Documented. Eight new test cases.

- [X] Tests
- [X] Documentation
- [X] torture test safe
- [X] build + tests work on all platforms
- [X] embedded null-byte in content safe
- [X] positive user feedback


jay:
I think 'encode' or 'urlencode' is easier to understand than 'percent' or %. --expand-data "{{fix:urlencode}}", also wouldn't it be easier if it could all be done in one step

emanuele6:
@jay `percent` urlencodes only the value of the variable when expanding it, you use it when you want to a variable as a path component of a URL for example

```
--variable "foo=$foo" --expand-url 'https://example.org/{{foo:percent}}'
```

If you don't use `:percent` you will expand the value of `{{foo}}` directly potentially adding queries, fragments, slashes, etc

docs/cmdline-opts/variable.d
```diff
@@ -0,0 +1,35 @@
...
+
+The name must be unique within a command line invoke, setting the same
+variable name again will be ignored.
+
+The contents of a variable can be referenced in a later command line option
+when that option name is prefixed with "--expand-", and the name is used as
+"{{name}}" (without the quotes).
+
+When expanding variables, curl can also expand environment variables if the
+name is prefixed with "env:", like in "{{env:name}}".
```
ukaea-chah, line 25: Does `{{env:trim}}` mean the environment variable "trim", or the variable "env", trimmed?  A different choice of syntax would avoid ambiguity - perhaps a pipe character "|" to prefix conversions rather than ":"

bagder: Yes, thanks. Good point. @emanuele6 also brought up the fact that we right now cannot use a variable named `env`.

I avoided the pipe character so far because it has a special meaning to shells. Maybe that should not stand in the way. Using pipes for this purpose seems to also be more common in other situation featuring similar functionality.

Another alternative is to change the `env:` prefix and use something else to make it refer to an environment variable. Like `{{%name}}` or `{{!name}}` for example. I'm avoiding `$name` since then it *really* becomes hard to figure out if the shell expands the name or curl is!

bagder: I'll switch to `{{%name}}` for environment variables and we can get a feel for how that is. Then decide if that works or not.

bagder: `%` might not be a good choice for Windows...

emanuele6: It also might get confused with the `%` alias for `percent`.
But for that maybe we can just rename `:percent` to `:url` so it is shorter, and ditch the `:%` alias.

bagder:
After having slept on it, I believe I now favor the "fail with a fatal error" option. This will be the most helpful thing to users as it makes it immediately apparent what happens, as removing or replacing the null bytes will lead to more hard-to-debug errors to puzzle users.

If we want a replace option, we could instead implement a dedicated function for that purpose.

eloydegen:
I think the documentation should include some note about the security risks. If an attacker is able to modify an environment variable it might be possible to exfiltrate data which might otherwise not be possible.

bagder:
> I think the documentation should include some note about the security risks. If an attacker is able to modify an environment variable it might be possible to exfiltrate data which might otherwise not be possible.

Providing clear documentation is of course valuable and a priority. Do you have any proposed wording?

An attacker that can change environment variables for curl is a serious attack already before this, since curl will acknowledge several variables outside of this new feature.

Also, this new feature is only expanding environment variables that is explicitly asked for, which is a reason we add this `--expand-` prefix thing.

bagder:
Me and @emanuele6 brainstormed a bit more on IRC.

1. What if you want to force an environment variable to be set or
2. What if you want a default value in case the environment variable is not set?

We came up with this tweak to support both those things, that I think fits rather nicely:

- remove `{{%name}}` support, get back to a single name space where we need to "import" environment variables to be able to expand them

Import environment variables like this:

- `--variable %name` imports an environment variable, fails with an error if not set. It can then be expanded as normal with `{{name}}`.
- `--variable %name=default` imports an environment variable, or sets it to the default value. The default can then also be read from a file with `--variable %name@default`
- `name` can not be the same name as an existing variable

emanuele6:
> * name can not be the same name as an existing variable

Personally I think you should be allowed to redefine a variable.

While `--variable foo=bar --variable foo=baz` is likely a mistake, it is possible that the curlrc is defining with common names like `proxy`, `host` or `ua`, and if a script tries to use `--variable proxy=` unknowingly, it will fail unexpectedly.

Since variables are only used by the user for `--expand-` options, and not by curl internally, I think it should be fine to allow redefining them.

bagder:
thanks @jay!

PR #11347, opened 2023-06-19
CI: macos: use older gcc version

pheiduck:
Apple has no main gcc support.

So find out which version works best with xcode.

> Hi Philip,
> 
> after reviewing the issue you reported, we found that it should be investigated and fixed by gcc - the third-party software you are using. Contact them and ask for support.
> 
> Apple does not supply or support gcc. We only support clang.
> 
> When the problem is fixed, you can close this feedback. To do so, select "Close feedback" from the action button above. Thank you very much.


pheiduck:
downgrading also does not help here, I would contact gcc upstream later...

PR #11352, opened 2023-06-20
Fixe unused variable ctx in cf-socker.c

humbleAcolyte:
This PR fixes compiler warning (gcc 8.3 with flag -Wunused-variable) "unused variable ‘ctx’".
The variable is used only if HAVE_GETPEERNAME is defined, but variable defined before directive.


bagder:
Thanks!

PR #11356, opened 2023-06-21
Update curl.h Include <sys/select.h> for vxworks

sheshadriv32:


include/curl/curl.h
```diff
@@ -94,6 +94,7 @@
    (defined(__FreeBSD_version) && (__FreeBSD_version < 800000)) || \
    (defined(__MidnightBSD_version) && (__MidnightBSD_version < 100000)) || \
     defined(__sun__) || defined(__serenity__)
+    defined(__vxworks__)
```
bagder, line 97: This does not work without adding a trailing `|| \` on the line above, which then makes me suspect you changed this without testing?

bagder:
Still a draft, but I figure this can be merged as-is?

bagder:
Thanks!

Issue #11357, opened 2023-06-21
HTTP/2 downgrade to HTTP/1.1 not working


jay:
### I did this

Investigating #11034 which has a URL that when requested via HTTP/2 replies with RST_STREAM error HTTP_1_1_REQUIRED. libcurl should retry the transfer with HTTP/1.1 but doesn't. _(Note the reporter in that issue builds libcurl without HTTP/2 so it is an unrelated issue)._

~~~
curld -v https://wsnfse.vitoria.es.gov.br/producao/NotaFiscalService.asmx?WSDL

...
* HTTP/2 stream 1 was reset
* [CONN-0] readwrite_data() -> 56
* [CONN-0] Curl_readwrite() -> 56
* multi_done: status: 56 prem: 1 done: 0
* Connection #0 to host wsnfse.vitoria.es.gov.br left intact
* Expire cleared (transfer 0x526ee8)
curl: (56) HTTP/2 stream 1 was reset
~~~

Bisected to cab2d56e @icing

That commit changes http2_handle_stream_close to return a generic error before checking for CURLE_HTTP2_STREAM, which means CURLE_HTTP2_STREAM is never returned, the error is no longer properly output in verbose mode as HTTP_1_1_REQUIRED, and later code in libcurl to handle the downgrade is never reached:

https://github.com/curl/curl/blob/51f6a0dc1af2d597566d597695da54098b8e6119/lib/http2.c#L1579-L1598

https://github.com/curl/curl/blob/51f6a0dc1af2d597566d597695da54098b8e6119/lib/multi.c#L2479-L2500

It's unclear to me why `else if(stream->reset)` block needs to come before. Changing it back works but I don't know why it was changed in the first place there may be a good reason.

~~~diff
diff --git a/lib/http2.c b/lib/http2.c
index 9da3cae..0894561 100644
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1523,22 +1523,22 @@ static ssize_t http2_handle_stream_close(struct Curl_cfilter *cf,
     connclose(cf->conn, "REFUSED_STREAM"); /* don't use this anymore */
     data->state.refused_stream = TRUE;
     *err = CURLE_SEND_ERROR; /* trigger Curl_retry_request() later */
     return -1;
   }
-  else if(stream->reset) {
-    failf(data, "HTTP/2 stream %u was reset", stream->id);
-    *err = stream->bodystarted? CURLE_PARTIAL_FILE : CURLE_RECV_ERROR;
[...]

PR #11362, opened 2023-06-21
HTTP/2, error stream resets with code CURLE_HTTP2_STREAM

icing:
- refs #11357, where it was reported that HTTP/1.1 downgrades no longer works
- fixed with suggested change
- added test_05_03 and a new handler in the curltest module to reproduce that downgrades work


bagder (approved):

bagder:
Thanks!

PR #11371, opened 2023-06-22
quote.d: fix indentation of generated paragraphs

emanuele6:
quote.d was missing a .IP at the end which caused the paragraphs generated for See-also, Multi, and Example to not be indented correctly.

I also remove a redundant "This option can be used multiple times.", and replaced .IP "item" with .TP .B "item" to make more clear which lines are part of the list of commands and which aren't.


bagder:
Thanks!

PR #11376, opened 2023-06-22
Stop CMake from quietly ignoring missing Brotli

MichalPetryka:
For some reason the CMake project was set to `QUIET` for Brotli instead of `REQUIRED`. This makes builds unexpectedly ignore missing Brotli even when `CURL_BROTLI` is enabled.


MichalPetryka:
Discovered in https://github.com/microsoft/vcpkg/issues/32108.

bagder (approved):
Sounds reasonable to me!

bagder:
Thanks!

PR #11378, opened 2023-06-22
configure: the obsolete --without forms also trigger the error

divinity76:


configure.ac
```diff
@@ -438,7 +438,7 @@ Select from these:
 fi
 
 AC_ARG_WITH(darwinssl,,
-  AC_MSG_ERROR([--with-darwin-ssl no longer works!]))
+  AC_MSG_ERROR([--with-darwin-ssl and --without-darwin-ssl no longer works!]))
```
bagder, line 441: Since they are now two options, they no longer **work** ie the 's' should be removed as well.

divinity76: Thanks. I should subscribe to [Grammarly](https://www.grammarly.com/)

MarcelRaad (approved):

bagder:
Thanks!

Issue #11379, opened 2023-06-23
Crash in `h2_progress_egress` due to NULL stream


jbgoog:
### I did this

Seeing a [crash here](https://github.com/curl/curl/blob/curl-8_1_2/lib/http2.c#L1672) during `h2_progress_egress` call:
```
    rv = nghttp2_submit_priority(ctx->h2, NGHTTP2_FLAG_NONE,
                                 stream->id, &pri_spec);
```

`stream` is `NULL` so `stream->id` throws the following error:

```
EXC_BAD_ACCESS / KERN_INVALID_ADDRESS @0x00000000
```

### I expected the following
 - Application does not crash

### curl/libcurl version
 - 8.1.2

### operating system

iOS

dfandrich:
Can you give us more information? How do you reproduce this problem? What are
the versions of the dependencies? The output of curl -V would be a good start.

jbgoog:
We include libcurl as a dependency of our iOS app, so there is no commandline for `curl -V` unfortunately.

It looks like this line to get the stream can be `NULL`:
```
struct stream_ctx *stream = H2_STREAM_CTX(data);
```

```
#define H2_STREAM_CTX(d)    ((struct stream_ctx *)(((d) && (d)->req.p.http)? \
                             ((struct HTTP *)(d)->req.p.http)->h2_ctx \
                               : NULL))
```

From https://github.com/curl/curl/blob/curl-8_1_2/lib/http2.c#L1659

When that happens, `stream->id` aborts.

It looks like `stream` [here](https://github.com/curl/curl/blob/curl-8_1_2/lib/http2.c#L1659) is only used to get the `stream->id`
```
    DEBUGF(LOG_CF(data, cf, "[h2sid=%d] Queuing PRIORITY",
                  stream->id));
    DEBUGASSERT(stream->id != -1);
    rv = nghttp2_submit_priority(ctx->h2, NGHTTP2_FLAG_NONE,
                                 stream->id, &pri_spec);
```

Otherwise it's unused.

I believe replacing:
```
 struct stream_ctx *stream = H2_STREAM_CTX(data);
```
with
```
int32_t stream_id = H2_STREAM_ID(data);
```

and then using `stream_id` directly instead of `stream->id` will fix the issue.

Testing it now.

jbgoog:
Here's the stack trace:

```
(http2.c:1675)		h2_progress_egress
(http2.c:1670)		h2_progress_egress
(http2.c:2199)		cf_h2_connect
[...]

PR #11380, opened 2023-06-23
Replaced stream->id with H2_STREAM_ID macro

jbgoog:
### Changes
 - Replaces `stream_id` with `H2_STREAM_ID` macro
 - This should fix https://github.com/curl/curl/issues/11379

### Testing
 - Was able to rerun the iOS app through several iterations and can no longer get it to crash; HTTP2 requests are working as expected


jbgoog:
Declining in favor of https://github.com/curl/curl/pull/11384

PR #11384, opened 2023-06-26
HTTP/2, fix crash in handling stream weights

icing:
- refs #11379 where the crash was reported
- weights may change "on the run", which is why there are checks in general egress handling. These must not trigger when the stream has not been opened yet.
- add test2404 to reproduce and verify


jay:
This needs to go through CI again because the test wasn't in the makefile. I added that , cleaned up the commit message and force pushed to your branch.

jay:
Thanks

Issue #11377, opened 2023-06-22
Hyper CI build is running out of disk space


dfandrich:
The Hyper CI build on GHA has been failing regularly lately, and the [latest build](https://github.com/curl/curl/actions/runs/5348954251) provides a pretty good clue as to why:
```
hyper
The hosted runner: GitHub Actions 91 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

hyper
You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 0 MB
```
The build fails in the `run tests` step, well after Hyper and curl have been built, so it's probably sufficient to delete the Hyper sources after it's built to recover some space.

dfandrich:
I tried a [build](https://github.com/curl/curl/actions/runs/5366355632/jobs/9735735590) with code to clean up the Hyper sources and rust dependencies that it downloads before running the tests, but it didn't help with the disk space warnings. It may be that that the Hyper build has some super verbose logging enabled that rapidly fills up the disk, or it may be something completely different.

icing:
I see an infinite loop with hyper in tests, just filling up the trace log. Investigating.

PR #11385, opened 2023-06-26
Hyper, fix EOF handling on input

icing:
- refs #11377, where we ran out of disc space due to an infinite loop with debug logging


lib/c-hyper.c
```diff
@@ -73,8 +73,6 @@ size_t Curl_hyper_recv(void *userp, hyper_context *ctx,
 
   DEBUGF(infof(data, "Curl_hyper_recv(%zu)", buflen));
   result = Curl_read(data, conn->sockfd, (char *)buf, buflen, &nread);
-  if(!result && !nread)
-    result = CURLE_AGAIN;
```
jay, line 77: Does there need to be a similar revert for the same logic that was added in Curl_write? I notice #11203 was reopened because having this logic does not fix what was reported.

bagder (approved):

PR #11386, opened 2023-06-26
Refactor url_proto to tell why it failed.

markg85:
https://github.com/curl/curl/pull/8805 will need to know why `url_proto` fails.
That is a very self-contained change so might be better as it's own PR.

I did put the `struct URLProtoResult` just somewhere that worked for me, i'm not sure if `tool_cfgtable.h` is the appropriate file.

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -696,10 +696,12 @@ noretry:
 /*
  * Return the protocol token for the scheme used in the given URL
  */
-static const char *url_proto(char *url)
+static struct URLProtoResult url_proto(char *url)
 {
+  struct URLProtoResult res;
   CURLU *uh = curl_url();
-  const char *proto = NULL;
+  res.proto = NULL;
+  res.result = CURLE_URL_MALFORMAT;
 
   if(uh) {
     if(url) {
@@ -709,14 +711,24 @@ static const char *url_proto(char *url)
         if(!curl_url_get(uh, CURLUPART_SCHEME, &schemep,
                          CURLU_DEFAULT_SCHEME) &&
            schemep) {
-          proto = proto_token(schemep);
+          res.proto = proto_token(schemep);
+
+          if(!res.proto) {
+            res.result = CURLE_UNSUPPORTED_PROTOCOL;
+          }
+          else
+            res.result = CURLE_OK;
           curl_free(schemep);
         }
       }
     }
     curl_url_cleanup(uh);
   }
-  return proto? proto: "???";   /* Never match if not found. */
+
+  if(res.result)
+    res.proto = "???";
+
+  return res;
 }
 
 /* create the next (singular) transfer */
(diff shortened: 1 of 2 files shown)


markg85:
This PR is breaking a bunch of testcases because `url_proto` is now being checked whereas before it wasn't.
I checked a couple failed tests but lets take one in particular.
[test423](https://github.com/curl/curl/blob/master/tests/data/test423). It fails because:

- "h55p://hello2000:1/%TESTNUMBER?qqqq#ffff" start with a non-supported protocol (h55p)
- "local host" is also a non-supported protocol
- changing these to supported ones changes the output expectation

This is just one of the failing tests (423 424 686 1146 1268) but they all have protocol errors in their test. Yes, including testcase 1268 :) (tries to use the "hej" protocol, that's not a thing).

Do i fix the testcases with protocols that are supported (probably replacing all these protocol errors with `http` or `file`?

src/tool_operate.c
```diff
@@ -1260,6 +1267,9 @@ static CURLcode single_transfer(struct GlobalConfig *global,
           break;
 
         use_proto = url_proto(per->this_url);
+        result = use_proto.result;
+        if(result)
+          break;
```
bagder, line 1272: This is new behavior. We do not error on this in existing code because it allows libcurl to have the final say if the URL works or not.

markg85: Should i remove this check in this commit?
In https://github.com/curl/curl/pull/8805 i would then check for the protocol first (same style like `if(use_proto.proto == proto_scp || use_proto.proto == proto_sftp)`) and only in that case handle the return type?

In other terms: Handle return type on IPFS specific case, ignore otherwise?

markg85:
**Why** does this fail test 1268???
The code is now exactly the same as it was before yet on quite a few nodes that test just keeps on failing.
On many other nodes it's no issue. When i test it locally, there's also no issue.

markg85:
Yay, it passed (mostly).
The failing tests are nearly all brew errors. And there's one test error but that's because of some missing symbol in the ssl library there, nothing i can do about those.

markg85:
@bagder 

Anything left to change here? It's squashed, rebased and ready to go i think.

bagder:
I want to queue this up to be merged together with a first user of this functionality, since otherwise this just adds code with no obvious benefit.

markg85:
That's getting tricky on a branch level. As basing https://github.com/curl/curl/pull/8805 on this branch will close the current PR and open a new one.

While my intent was to split things up and make it easier, i think (unless you merge this as-is) it's easier now to re-integrate this change in https://github.com/curl/curl/pull/8805. That prevents a needless extra PR.

If you think this sounds fine, just +1 this and i'll go ahead with the above.

markg85:
5 days later.
I'm picking. Closing this one and re-integrating it in the PR that is using this.

PR #11392, opened 2023-06-28
Adding mod_h2 directive `H2ProxyRequests` needed for new master

icing:
Just updated master of mod_h2 which now requires `H2ProxyRequests` directives for forward proxying with HTTP/2 to work.


icing:
@jay or @dfandrich could you merge this when convenient? Thanks!

jay:
Thanks

PR #11393, opened 2023-06-28
Add libproxy support

janbrummer:
Updated MR initially made by dwmw2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,6 +224,13 @@ option(CURL_DISABLE_TFTP "disables TFTP" OFF)
 mark_as_advanced(CURL_DISABLE_TFTP)
 option(CURL_DISABLE_VERBOSE_STRINGS "disables verbose strings" OFF)
 mark_as_advanced(CURL_DISABLE_VERBOSE_STRINGS)
+option(CURL_DISABLE_LIBPROXY "disables Libproxy support" OFF)
+mark_as_advanced(CURL_DISABLE_LIBPROXY)
+
+if (NOT CONFIG_DISABLE_LIBPROXY)
+  find_package(libproxy REQUIRED)
+  list(APPEND CURL_LIBS ${LIBPROXY_LIBRARIES})
+endif()
 
 # Corresponds to HTTP_ONLY in lib/curl_setup.h
 option(HTTP_ONLY "disables all protocols except HTTP (This overrides all CURL_DISABLE_* options)" OFF)
@@ -1476,6 +1483,7 @@ _add_if("unicode"       ENABLE_UNICODE)
 _add_if("threadsafe"    HAVE_ATOMIC OR (WIN32 AND
                         HAVE_WIN32_WINNT GREATER_EQUAL 0x600))
 _add_if("PSL"           USE_LIBPSL)
+_add_if("libproxy"      NOT CURL_DISABLE_LIBPROXY)
 string(REPLACE ";" " " SUPPORT_FEATURES "${_items}")
 message(STATUS "Enabled features: ${SUPPORT_FEATURES}")
 
--- a/lib/curl_libproxy.c
+++ b/lib/curl_libproxy.c
@@ -1,0 +1,70 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Jan-Michael Brummer <jan-michael.brummer1@volkswagen.de>.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at http://curl.haxx.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ ***************************************************************************/
+
+#include "curl_setup.h"
+
+#ifndef CURL_DISABLE_LIBPROXY
+
+#include "urldata.h"
+
+#include <proxy.h>
+/* The last 3 #include files should be in this order */
+#include "curl_printf.h"
+#include "curl_memory.h"
+#include "memdebug.h"
+
+void Curl_libproxy_cleanup(pxProxyFactory *proxy_factory)
+{
+  if(proxy_factory)
+    px_proxy_factory_free(proxy_factory);
+}
+
+char *Curl_libproxy_detect_proxy(struct Curl_easy *data, const char *url)
+{
+  char *result = NULL;
+
+  /* Initialization of libproxy is done here, as otherwise libproxy and curl
+   * will have a cycle issue
+   */
+  if(!data->proxy_factory)
+    data->proxy_factory = px_proxy_factory_new();
+
+  if(data->proxy_factory) {
+    char **libproxy_results =
+        px_proxy_factory_get_proxies(data->proxy_factory, url);
+
+    if(libproxy_results) {
+      /*
+       * - We only cope with one; can't fall back on failure
+       * - direct:// access is returned as NULL
+       */
+      if(strcmp(libproxy_results[0], "direct://") != 0)
+        result = strdup(libproxy_results[0]);
+
+      px_proxy_factory_free_proxies(libproxy_results);
+    }
+  }
+
+  return result;
+}
+
+#endif /* CURL_DISABLE_LIBPROXY */
--- a/lib/url.c
+++ b/lib/url.c
@@ -457,6 +457,11 @@ CURLcode Curl_close(struct Curl_easy **datap)
   }
 #endif
 
+#ifndef CURL_DISABLE_LIBPROXY
+  Curl_libproxy_cleanup(data->proxy_factory);
+  data->proxy_factory = NULL;
+#endif
+
   /* destruct wildcard structures if it is needed */
   Curl_wildcard_dtor(&data->wildcard);
   Curl_freeset(data);
@@ -2113,6 +2118,11 @@ static char *detect_proxy(struct Curl_easy *data,
   }
   if(proxy)
     infof(data, "Uses proxy env variable %s == '%s'", envp, proxy);
+#ifndef CURL_DISABLE_LIBPROXY
+  if(!proxy) {
+    proxy = Curl_libproxy_detect_proxy(data, data->state.url);
+  }
+#endif
 
   return proxy;
 }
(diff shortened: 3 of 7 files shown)


jay:
Ref: #977

bagder:
also #9068 and my blog post about the dream: https://daniel.haxx.se/blog/2022/08/12/the-dream-of-auto-detecting-proxies/

janbrummer:
Correct and we have taken your review very serious and have addresses them. Volkswagen now takes care of libproxy and it's development.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,69 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) 2022-2023, Jan-Michael Brummer, <jan-michael.brummer1@volkswagen.de>.
```
bagder, line 8: We don't use year ranges in our copyright lines anymore.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,69 @@
...
+
+#include <proxy.h>
+/* The last 3 #include files should be in this order */
+#include "curl_printf.h"
+#include "curl_memory.h"
+#include "memdebug.h"
+
+void Curl_libproxy_cleanup(pxProxyFactory *proxy_factory)
+{
+  if (proxy_factory)
```
bagder, line 37: We don't use spaces after `if` before the `()` - which `make checksrc` will tell you.

lib/curl_libproxy.c
```diff
@@ -0,0 +1,69 @@
...
+  char *result = NULL;
+
+  /* Initialization of libproxy is done here, as otherwise libproxy and curl
+   * will have a cycle issue
+   */
+  if (!data->proxy_factory)
+    data->proxy_factory = px_proxy_factory_new();
+
+  if (data->proxy_factory) {
+    char **libproxy_results = px_proxy_factory_get_proxies(data->proxy_factory, url);
```
bagder, line 52: This function call is blocking and can potentially take seconds to complete, right? It seems very bad and can potentially disrupt a lot of other already ongoing transfers if that happens. This very basic API is simple, but I would call it **too simple** because of this.  We want it asynchronous to allow the rest of the library to continue while this single transfer figures out its proxy.

Also, this makes the libproxy library parse the provided URL (doesn't it?) and we know that having different URL parsers act on the same URL is a recipe for disaster and security problems. Will it use libcurl's URL parser?

lib/url.c
```diff
@@ -457,6 +457,11 @@ CURLcode Curl_close(struct Curl_easy **datap)
   }
 #endif
 
+#ifndef CURL_DISABLE_LIBPROXY
+  Curl_libproxy_cleanup (data->proxy_factory);
```
bagder, line 461: Another spaces we don't like.

bagder (changes requested):

lib/curl_libproxy.c
janbrummer: Used checksrc and fixed the nitpicks.

lib/curl_libproxy.c
janbrummer: Used checksrc and fixed the nitpicks.

lib/url.c
janbrummer: Used checksrc and fixed the nitpicks.

lib/curl_libproxy.c
janbrummer: libproxy is only making use of libcurl for it's download part and the download time is limited to 30sec (we can reduce this if required). Furthermore it is setting NOPROXY to ensure we are not cycling.

janbrummer: You are referring to your _multi_xx functions, right?

janbrummer: I can provide you a threaded version with a callback but i'm currently not sure how to connect it with curl in a sane way. Maybe you have an advice here?

janbrummer:
Regarding tests: Should it be disabled by default and we add a new action stage for e.g. Linux + libproxy. Or should it be enabled by default? Then i would need to change all actions.

lib/curl_libproxy.c
bagder: If threading is the only way to avoid blocking, then I believe we should keep the libproxy use out of libcurl. libcurl is meant to be able to run *thousands* of parallel transfers in a single thread. Adding threads for this purpose complicates matters a lot and makes libcurl use (much) more memory. Asynchronous non-blocking internal behaviors should be done without threads to the furthest extent possible (the threaded name resolving is the only exception).

janbrummer: What do you have in mind as an alternative? Polling for feedback? Actually we are free to implement the solution that suits curl.

bagder: Yes, polling. Other non-blocking APIs typically provide a file descriptor that can be used to wait for activity on (if it does not complete immediately). Once there is activity on the descriptor, libcurl calls the necessary function, which does as much as possible without blocking. Over and over until the action is complete.

janbrummer: Performance Test
----------------------

I thought it would be good to add some performance data to the discussion. Libproxy actually does the following steps to
identify the necessary proxy configuration:

- Lookup of it's configuration
- Downloading & parsing PAC files if necessary

For lookup I've done a n-runs for receiving the configuration using the different plugins on Linux. Result is:

KDE         : 0.000015 s
GNOME  : 0.000034 s
Env          : 0.000004 s
sysconfig: 0.000004 s

For **manual proxy configuration** we would be out right here. Next parts are for **automatic proxy** configuration:

Parsing of a larger PAC file (20k): 0.001124 s

The only variable time would be in case wpad or pac proxy servers are set in the configuration. In this case downloading the
PAC file is depending on libcurl to download it (with a hard limit of 30s). 

In my case using the company proxy i did have a this result:

Download: 0.001235 s

Please note that this file is downloaded once and then reused for the next calls. So in the worst initial case i do have measured
this value:

**0,002396 s**

And for the continuing runs:

**0.000966 s**

bagder: We need to design this feature (and others) to work proper in the **worst cases**. Like when it runs on on anciently slow hardware, the name resolve takes multiple seconds, the loading of the PAC is done over a slow VPN across the globe and the file is hundreds of kilobytes of complicated JavaScript.

nmoinvaz: We've been working on a user-mode proxy library that has threading support here: https://github.com/nmoinvaz/proxyres.

CMakeLists.txt
```diff
@@ -224,6 +224,13 @@ option(CURL_DISABLE_TFTP "disables TFTP" OFF)
 mark_as_advanced(CURL_DISABLE_TFTP)
 option(CURL_DISABLE_VERBOSE_STRINGS "disables verbose strings" OFF)
 mark_as_advanced(CURL_DISABLE_VERBOSE_STRINGS)
+option(CURL_DISABLE_LIBPROXY "disables Libproxy support" OFF)
```
nmoinvaz, line 227: This should default to `ON` since libproxy license is LGPL. It's possible users would likely not know they were compiling in incompatibly licensed code.

lib/curl_libproxy.c
bagder: @nmoinvaz I don't know what "threading support" means, but if you read my comments again here you see that I much prefer *not* using threads. Good old non-blocking API using the single thread.

nmoinvaz: I only meant that the library pushes the proxy resolution to a thread pool if the specific OS does not have asynchronous proxy resolution support. As far as the library API goes, it does not block.

CMakeLists.txt
AdrianVovk: > compiling in incompatibly licensed code

What are the implications of this? Distros will not be allowed to make use of this?

Why is it being linked together at all then if it's incompatibly licensed?

mcatanzaro: It's not incompatibly-licensed. The fact that the licenses are not the same does not make them incompatible. libproxy is fine except in environments where LPGLv2+ is banned (in which case, common dependencies like GLib are also banned and most of desktop Linux is off-limits).

bagder: libproxy being LGPL will make a problem for some, no doubt. We already support GnuTLS which also is LGPL, and for the same reason is maybe not a popular choice in numerous commercial and embedded solutions. (libcurl does not use GLib and in fact many users of libcurl don't use GLib elsewhere either)

I believe we should at least start out with *not* enabling it by default to get people to opt-in first to get usage, feedback and reports back first before we consider it enabled by default (if found).

MichalPetryka: Optionally, it could be enabled by default when building curl as a shared library but opt in when building a static one.

lib/curl_libproxy.c
bagder: I would not be against support *multiple* libraries for detecting proxy. Then the user can decide at build-time which one to use. We do this already for TLS, SSH, QUIC and more.

CMakeLists.txt
bagder: No, that would be a terrible and confusing way to do it. Besides, most builds of libcurl build both a dynamic and a static lib at once. The person who links an application with the library then decides which one to use.

bagder:
Since it still seems libproxy cannot provide a non-blocking API and there has been no movement here in over a month, it seems the time has not yet come for libproxy to be used by curl. Until me meet again, I close.

rhaberkorn:
> Since it still seems libproxy cannot provide a non-blocking API

I also don't think that libproxy could easily provide an end-to-end non-blocking API. The main blocking thing it does is downloading PAC files, which in itself could be done non-blocking - actually it's using libcurl for the purpose - but the PAC execution itself is fundamentally blocking. So IMHO the only way to make the operation non-blocking would be threading.

@bagder What non-blocking/asynchronous curl API would they have to support? The proxy detection in lib/url.c seems to be blocking/synchronous as well.

bagder:
> So IMHO the only way to make the operation non-blocking would be threading.

From a user's point of view, exactly how it is accomplished is not important. Of course curl could itself fire up a helper thread to execute libproxy's API from within, as an alternative take. libcurl itself is simply not allowed to block for that long.

> The proxy detection in lib/url.c seems to be blocking/synchronous as well.

There is no "detection" there to speak of, so there's nothing to block on.

nmoinvaz:
@rhaberkorn I don't know if you are looking for a solution, but we developed a cross-platform client side library (as opposed to libproxy which is a system library and GPL licensed) that handles proxy resolution requests asynchronously using threading and is available at https://github.com/snxd/proxyres.

rhaberkorn:
> > The proxy detection in lib/url.c seems to be blocking/synchronous as well.
> 
> There is no "detection" there to speak of, so there's nothing to block on.

That's one way to look at it. It's not necessary to expose any async API as long as it doesn't do anything that's potentially blocking. After reading some more and hopefully understanding how curl handles asynchronous operations, I think that in order to support asynchronous proxy resolution create_conn_helper_init_proxy() and detect_proxy() would merely have to carry through and set the "async" flag. So that wouldn't require huge refactoring and *should* be straight forward.

PR #11395, opened 2023-06-29
Digest md5: Use hostname to generate spn instead of realm

### LINKED ISSUE #11369: imap digest-md5 broken
I think the digest-md5 authentication for imap/smtp has been broken for some time.
Can someone check?

Same credentials with AUTH=LOGIN set with login options works.

kop316:
Fixes #11369

In https://www.rfc-editor.org/rfc/rfc2831#section-2.1.2

digest-uri-value should be serv-type "/" host , where host is

      The DNS host name or IP address for the service requested.  The
      DNS host name must be the fully-qualified canonical name of the
      host. The DNS host name is the preferred form; see notes on server
      processing of the digest-uri.

Realm may not be the host, so we must specify the host explicitly.

This fixes an issue I have seen where `DIGEST-MD5` does not work for the Visual Voicemail Server (and presumably fixes the other one references in the issue).


kop316:
I seem to keep getting:

```
15:43:23.019587 < "dXNlcm5hbWU9InVzZXIiLHJlYWxtPSJjdXJsIixub25jZT0iNTMwMGQxN2ExZDY5NWJkNDExZTRjZGY5NmY5NTQ4YzIzY2VkNjE3NSIsY25vbmNlPSIzNDMzMzIzMTM1MzMzMjMxMzYzMzMyMzEzNzMzMzIzMSIsbmM9IjAwMDAwMDAxIixkaWdlc3QtdXJpPSJpbWFwLzEyNy4wLjAuMSIscmVzcG9uc2U9YmU2MzgyNDkzNjJkN2FhMGUzNTM4NTA3Njc1MWFiNDgscW9wPWF1dGg="
15:43:23.019638 dXNlcm5hbWU9InVzZXIiLHJlYWxtPSJjdXJsIixub25jZT0iNTMwMGQxN2ExZDY5NWJkNDExZTRjZGY5NmY5NTQ4YzIzY2VkNjE3NSIsY25vbmNlPSIzNDMzMzIzMTM1MzMzMjMxMzYzMzMyMzEzNzMzMzIzMSIsbmM9IjAwMDAwMDAxIixkaWdlc3QtdXJpPSJpbWFwLzEyNy4wLjAuMSIscmVzcG9uc2U9YmU2MzgyNDkzNjJkN2FhMGUzNTM4NTA3Njc1MWFiNDgscW9wPWF1dGg= wasn't handled!
15:43:23.019672 > "A002 BAD dXNlcm5hbWU9InVzZXIiLHJlYWxtPSJjdXJsIixub25jZT0iNTMwMGQxN2ExZDY5NWJkNDExZTRjZGY5NmY5NTQ4YzIzY2VkNjE3NSIsY25vbmNlPSIzNDMzMzIzMTM1MzMzMjMxMzYzMzMyMzEzNzMzMzIzMSIsbmM9IjAwMDAwMDAxIixkaWdlc3QtdXJpPSJpbWFwLzEyNy4wLjAuMSIscmVzcG9uc2U9YmU2MzgyNDkzNjJkN2FhMGUzNTM4NTA3Njc1MWFiNDgscW9wPWF1dGg= is not dealt with![CR][LF]"
```

And I am not sure why this is occuring.

kop316:
I added the response auth that should be included per https://www.rfc-editor.org/rfc/rfc2831#section-2.1.3 . It looks like curl does not check that response auth.

bagder:
According to feedback in #11369, this does not solve the issue...

kop316:
> According to feedback in #11369, this does not solve the issue...

Yeah I unfortunately saw that. What I can say is this PR *does* fix DIGEST-MD5 login for me, so I have no further way of trying to diagnose/fix it for @MonkeybreadSoftware  . I gave a couple of ideas of how to fix it in https://github.com/curl/curl/issues/11369#issuecomment-1627804074 , maybe those will wok. 

I also saw this is identical to what is in `digest-sspi.c` https://github.com/curl/curl/blob/master/lib/vauth/digest_sspi.c#L140 (though I am admittedly not sure what the difference is between `digest_sspi.c` and `digest.c`. This also looks to be more correct versus what is there now (as I referred to here: https://github.com/curl/curl/pull/11395#issue-1781277378 ).

I welcome any other thoughts on what you think should/can be done.

jay:
> though I am admittedly not sure what the difference is between `digest_sspi.c` and `digest.c`

digest_sspi.c uses the Windows SSPI API that is enabled in some Windows builds. Both digest.c and digest_sspi.c should format the spn as "service/host" if passed service and host strings.

git blame puts it at 9feb2676 (check its comments) when, prior to that, to build spn there was only one parameter for host and realm called 'instance'. You can see since then for some calls the host is passed as the realm, but in this one case here the realm is passed as the host. I still don't know why that is. Essentially, that commit changed the format. @captain-caveman2k

kop316:
@jay If I am reading https://www.rfc-editor.org/rfc/rfc2831#section-2.1.2 correctly, `realm` is not the same as `host`, and `realm` goes into `realm-value` when calculating `A1`. the `digest-uri-value  = serv-type "/" host [ "/" serv-name ]` (here it is the `spn`) should be the host instead (which is what my PR here was attempting to correct). That was why I was confused about `digest_sspi.c` and `digest.c` being different, `digest_sspi.c` looks to correctly accumulate the spn value.

Actually when looking at https://github.com/curl/curl/commit/9feb2676a4e153eef7f5536f940678af2df2cf9e , almost every other one looks to use host as well.

Perhaps I am reading the RFC wrong?

kop316:
So for a different IMAP server for visual voicemail, they are able to successfully login with the current `DIGEST-MD5`.
the URL is ` imaps://vvm.mobile.att.net/INBOX?ALL`, realm is `bothwaacds03.attwireless.net`, digest-uri `imap/bothwaacds03.attwireless.net` . 

However, in Android's actual code, They appear to use the host for the `digest-uri` : https://cs.android.com/android/platform/superproject/+/master:packages/apps/Dialer/java/com/android/voicemail/impl/mail/store/imap/DigestMd5Utils.java;l=81

And as I annotated here: https://github.com/curl/curl/issues/11369#issuecomment-1613483953 the different realm is not functional for me in DIGEST-MD5.

So perhaps the realm is used some of the time for DIGEST-MD5, but I am just not sure?

jay:
My best guess as to what happened here is the dev assumed that the realm would be a more formal host (like maybe they assumed with a hostname of `p012384093214.mail.test.com` and a realm of `mail.test.com` that it would make more sense to send the latter, and did not consider a realm format of group@host, because as far as I can tell the RFC doesn't allow an spn of imap/group@host. I haven't found any discussion about it.

You said in the other thread you had a realm of `imap/realm@mavenir.com` and I'd be interested if `imap/mavenir.com` works for you instead. Regardless I think your change here is correct because it follows the RFC so I plan to commit it. If this breaks something for somebody possibly we'd have to change it to imap/<host>/<realmhost> if I'm reading it right...

> So for a different IMAP server for visual voicemail, they are able to successfully login with the current DIGEST-MD5.
the URL is  imaps://vvm.mobile.att.net/INBOX?ALL, realm is bothwaacds03.attwireless.net, digest-uri imap/bothwaacds03.attwireless.net .

In that case what happens if you send spn `imap/vvm.mobile.att.net` does it work?

kop316:
> You said in the other thread you had a realm of `imap/realm@mavenir.com` and I'd be interested if `imap/mavenir.com` works for you instead. 

I tried `imap/mavenir.com` as the digest-uri, that does not work.

>Regardless I think your change here is correct because it follows the RFC so I plan to commit it. If this breaks something for somebody possibly we'd have to change it to `imap/<host>/<realmhost>` if I'm reading it right...

I also tried this, and make the digest-uri `imap/e5.vvm.mstore.msg.t-mobile.com/mavenir.com`, that does not work as well

> 
> > So for a different IMAP server for visual voicemail, they are able to successfully login with the current DIGEST-MD5.
> > the URL is  imaps://vvm.mobile.att.net/INBOX?ALL, realm is bothwaacds03.attwireless.net, digest-uri imap/bothwaacds03.attwireless.net .
> 
> In that case what happens if you send spn `imap/vvm.mobile.att.net` does it work?

I actually worked with someone to try this out. I will see if I can get them to try it out and see if it works.

kop316:
@jay The user was successfully able to log into the IMAP server with spn `imap/vvm.mobile.att.net`. 

As an aside, AT&T built their own proprietary app for accessing their IMAP server. I sort of wonder if they use curl as a backend and just worked around this issue?

Issue #11405, opened 2023-07-06
curl: (55) Failed sending HTTP request when trying to send a request with large header size. >8.1.2


dimaboyko:
Hi, I am not certain this is a bug, perhaps this was mentioned somewhere, however, I did not discover it, looking at the changelog.
After updating curl from 8.0.1 I have started seeing requests failing to be sent. Further investigation showed, that this is only happening for requests with large header size (10K characters)

### I did this
Send a request with large header size. Example:
`curl -X GET -L -H "MyHeader: $(printf '%*s' 10000 | tr ' ' 'A')" https://postman-echo.com/get`
This is working ok on 8.0.1, but not on 8.1.2.
8.1.2 is returning `curl: (55) Failed sending HTTP request`
### I expected the following
Response from the server. 
### curl/libcurl version
Error happening with:
```shell
curl 8.1.2 (aarch64-alpine-linux-musl) libcurl/8.1.2 OpenSSL/1.1.1u zlib/1.2.12 brotli/1.0.9 nghttp2/1.47.0
Release-Date: 2023-05-30
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets
``` 

Request is sent successfully with:
```shell
curl 8.0.1 (aarch64-alpine-linux-musl) libcurl/8.0.1 OpenSSL/1.1.1u zlib/1.2.12 brotli/1.0.9 nghttp2/1.43.0
Release-Date: 2023-03-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets
```

### operating system
Alpine Linux 3.16 Docker image.
`Linux  5.15.49-linuxkit`

### Screenshot
<img width="1712" alt="image" src="https://github.com/curl/curl/assets/16645596/8b2c9673-3815-4994-9f2b-d62708c2959e">

dfandrich:
Could this be related to #10720 which changed internal header handling in 8.0.2?
[...]

PR #11407, opened 2023-07-07
HTTP/2, raise header limitations above and beyond

icing:
- not quite to infinity
- refs #11405
- rewrote the implementation of our internal HTTP/1.x request parsing to work with very large lines using dynbufs.
- new default limit is `DYN_HTTP_REQUEST`, aka 1MB, which is also the limit of curl's general HTTP request processing.

### DIFF
--- a/lib/http1.c
+++ b/lib/http1.c
@@ -38,124 +38,97 @@
 #include "memdebug.h"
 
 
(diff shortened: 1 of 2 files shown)


bagder (approved):

bagder:
Thanks!

PR #11416, opened 2023-07-09
curl: make %output{} in -w specify a file to write to

bagder:
It can be used multiple times.

Add docs. Test 990 and 991 verify.

Idea: #11400
Suggested-by: ed0d2b2ce19451f2


bagder:
I should probably also provide a way to append to a file

PR #11417, opened 2023-07-10
macOS: fix target detection

icing:
- TARGET_OS_OSX is not always defined on macOS
- this leads to missing symbol Curl_macos_init()
- TargetConditionals.h seems to define these only when dynamic targets are enabled (somewhere?)
- this PR fixes that on my macOS 13.4.1
- I have no clue why CI builds worked without it


xquery (approved):
LGTM

bagder:
Thanks!

Issue #6363, opened 2020-12-22
Excessive HTTP/2 packets with TCP_NODELAY


Hello71:
### Steps to reproduce

    wireshark &
    curl https://google.com
    curl --no-tcp-nodelay https://google.com

### Expected behavior

HTTP/2 GET is sent together with TLS Finished (optimally) or in one packet thereafter (realistically) in both cases.

### Actual behavior

Unless --no-tcp-nodelay is specified, HTTP/2 GET is sent in four separate TCP packets after TLS Finished: Magic, SETTINGS, WINDOW_UPDATE, HEADERS.

### Additional information

curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1i zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 nghttp2/1.42.0

Although it doesn't cause additional RTT, three packets is still (40+32+22)*3 = 282 bytes on IPv6, plus PHY overhead. Additionally, the server usually sends four ACKs back when one would suffice, eating up 216 bytes on the way back. That's a decent chunk on slow cellular devices, considering that as far as I know there's no good reason to send those bytes.

Firefox and Chromium send one Magic+SETTINGS+WINDOW_UPDATE packet, then a separate HEADERS packet. This seems reasonable, if suboptimal. Firefox also sends a large number of PRIORITY messages in the first packet, perhaps in an effort to "warm up" multiplexing.

bagder:
Does it look sane for HTTP/2 when `TCP_NODELAY` is not set then?

Hello71:
Yes, if --no-tcp-nodelay is specified, then all four are collapsed into one packet. However, that packet is delayed until the TLS Finished is acknowledged, which is one additional RTT.

bagder:
Ok, so then the current state is at least better than the opposite, as wasting a few bytes because of too many packets seems to generally be preferred to being slower. Personally, I don't think this issue gets on the top of my TODO list so I'll appreciate if someone else wants to dig in and polish this...

Hello71:
[...]

PR #11420, opened 2023-07-11
http2: send HEADER & DATA together if possible

peppapighs:
This PR is based on the conversation [here](https://curl.se/mail/lib-2023-06/0040.html).

When sending an HTTP/2 POST request, curl will always defer request body to a different TLS record (if SSL/TLS is enabled) and TCP packet. This problem is previously addressed in https://github.com/curl/curl/issues/6363.

Belows are network packets captured by Wireshark by running `SSLKEYLOGFILE=/tmp/sslkeylog.log src/curl -k "https://nghttp2.org" -d "moo"`:

image

We can combine the TCP packet by disabling `TCP_NODELAY`. Belows is a TCP packet sent by running `SSLKEYLOGFILE=/tmp/sslkeylog.log src/curl -k "https://nghttp2.org" -d "moo" --no-tcp-nodelay`:

image

As can be seen from the image above, curl still generates 2 TLS records for HEADER and DATA frames, respectively (ignore the first TLS record from MAGIC frame).

Therefore, this PR provides an optimization that allows request body to be sent together with request headers in a single TLS record and TCP packet if the body size is small enough. This is achieved by attaching request body to header data so that `h2_submit()` can write the body directly after it has created a stream.

Belows is a TCP packet sent by running `SSLKEYLOGFILE=/tmp/sslkeylog.log src/curl -k "https://nghttp2.org" -d "moo"` with this change:

image

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1893,7 +1893,8 @@ static ssize_t h2_submit(struct stream_ctx **pstream,
   struct h1_req_parser h1;
   struct dynhds h2_headers;
   nghttp2_nv *nva = NULL;
-  size_t nheader, i;
+  const void *body = NULL;
+  size_t nheader, bodylen, i;
   nghttp2_data_provider data_prd;
   int32_t stream_id;
   nghttp2_priority_spec pri_spec;
@@ -2011,6 +2012,20 @@ static ssize_t h2_submit(struct stream_ctx **pstream,
     }
   }
 
+  body = (const char *)buf + nwritten;
+  bodylen = len - nwritten;
+
+  if(bodylen > 0) {
+    /* We have request body to send in DATA frame */
+    ssize_t n = Curl_bufq_write(&stream->sendbuf, body, bodylen, err);
+    if(n < 0) {
+      *err = CURLE_SEND_ERROR;
+      nwritten = -1;
+      goto out;
+    }
+    nwritten += n;
+  }
+
 out:
   DEBUGF(LOG_CF(data, cf, "[h2sid=%d] submit -> %zd, %d",
          stream? stream->id : -1, nwritten, *err));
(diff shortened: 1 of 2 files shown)


icing (approved):
LGTM! Nicely done.

lib/http2.c
```diff
@@ -2011,6 +2012,20 @@ static ssize_t h2_submit(struct stream_ctx **pstream,
     }
   }
 
+  body = (const char *)buf + nwritten;
+  bodylen = len - nwritten;
+
+  if(bodylen > 0) {
+    /* We have request body to send in DATA frame */
+    ssize_t n = Curl_bufq_write(&stream->sendbuf, body, bodylen, err);
```
bagder, line 2020: Can this run into the bufq's max size? If not, I think the comment could mention this fact.

lib/http2.c
```diff
@@ -2011,6 +2012,20 @@ static ssize_t h2_submit(struct stream_ctx **pstream,
     }
   }
 
+  body = (const char *)buf + nwritten;
+  bodylen = len - nwritten;
+
+  if(bodylen > 0) {
```
bagder, line 2018: I think this should do `if(bodylen)` since anything non-zero will be larger than zero when the type is unsigned.

lib/http2.c
icing: In what way does it matter? Not sure i understand.

bagder: Oh right, it will deal with the number of bytes that it can and return that amount so it does not matter. Sorry, I did not think this through until now. Case closed. Thanks!

peppapighs: I believe it should not be possible since bufq should be empty at this point and the length of body cannot exceed `MAX_INITIAL_POST_SIZE`, which has the same value as `H2_STREAM_SEND_CHUNKS` * `H2_CHUNK_SIZE` (the max capacity of `stream->sendbuf`). However, since this fact is relying on different constant values that may or may not change in the future, I think maybe we should handle `CURLE_AGAIN` case by defer the DATA frame. What do you think?

icing: @peppapighs you accounted for such things by only incrementing `nwritten` by the amount acutally added to the `bufq`. The caller of the `send()` method has to be prepared that not all was written, as the same could happen to a HTTP/1.1 write onto a TCP socket.

peppapighs: Oh right. That makes sense.

bagder:
Thanks!

PR #11421, opened 2023-07-11
Enhanced Tracing in libcurl and curl

icing:
Making detailed logging availabel in libcurl and curl:

- curl_global_log_config() for setting log configration in libcurl
- `- --trace-config` command line option for curl
- man pages
- convert `DEBUGF(LOG_CF(..))` code to `CURL_LOG_CF()` that is available in non-debug builds
- protect `infof()` execution by verbosity check, saving some  cycles when not verbose
- test adjustments for new function

Examples:

Add HTTP/2 details to verbose output:

            curl -v --trace-config http/2 URL

Add HTTP/2, SSL and id details to verbose output:

            curl -v --trace-config http/2,ssl,ids URL

Enable all details in verbose output, including times and ids and all components in libcurl:

            curl -v --trace-config all URL

Enable all details, but not about SSL or TCP:

            curl -v --trace-config all,-ssl,-tcp URL


vszakats:
We'll need to update `libcurl.def` with the new API:
```diff
--- a/libcurl.def
+++ b/libcurl.def
@@ -28,6 +28,7 @@ curl_getenv
 curl_global_cleanup
 curl_global_init
 curl_global_init_mem
+curl_global_log_config
 curl_global_sslset
 curl_maprintf
 curl_mfprintf
```

bagder:
> We'll need to update `libcurl.def` with the new API:

Can't we generate that list? Or at worst make a CI job that checks that the list is up-to-date?

vszakats:
@bagder PR ready for checking it: #11570. I can't imagine a way to dynamically generate this from both CMake and `Makefile.mk` (and perhaps others that may be using this in the future). Converting it a raw list (without the `EXPORTS` header) and re-using it for more than `.def` purposes might work, so that a single list needs to be maintained, but it's non-trivial.

icing:
> ```diff
> curl_global_log_config
> ```

Added this in the recent push. Thanks for noticing

icing:
Some local measurement on my macOS dev machine. I see no real difference in performance between non-debug builds on master and this PR. Meaning the guarded log statements have no impact.

```
curl 8.3.0-DEV (x86_64-apple-darwin22.5.0) libcurl/8.3.0-DEV OpenSSL/3.0.9 zlib/1.2.11 brotli/1.0.9 zstd/1.5.5 c-ares/1.19.1 libidn2/2.3.4 libssh2/1.11.0 nghttp2/1.55.0 ngtcp2/0.17.0 nghttp3/0.13.0 librtmp/2.3

Downloads from httpd:
> python3 tests/http/scorecard.py -d --httpd h2
              Size       Single       Serial     Parallel
master:      100MB     703 MB/s     972 MB/s     934 MB/s
this PR:     100MB     710 MB/s     986 MB/s     959 MB/s

Small requests:
> python3 tests/http/scorecard.py -r --httpd h2

Requests, max in parallel
         Size        1            6           25           50          100
master:  10KB     1665 r/s     3428 r/s     3537 r/s     3522 r/s     3475 r/s
this PR: 10KB     1676 r/s     3439 r/s     3536 r/s     3515 r/s     3227 r/s

```

docs/libcurl/curl_global_log_config.3
```diff
@@ -0,0 +1,89 @@
...
+.\" *
+.\" * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+.\" * KIND, either express or implied.
+.\" *
+.\" * SPDX-License-Identifier: curl
+.\" *
+.\" **************************************************************************
+.TH curl_global_log_config 3 "01 August 2023" "libcurl" "libcurl"
+.SH NAME
+curl_global_log_config - Global libcurl logging configuration
```
bagder, line 26: What do you think of naming it `curl_global_trace()` ? It is shorter and I think that is slightly more fitting name that avoids the word "log" which might have a negative connotation. We all use the word *trace* for this for the command line option.

docs/libcurl/curl_global_log_config.3
```diff
@@ -0,0 +1,89 @@
...
+
+The \fIconfig\fP string is a list of comma-separated component names. Unknown
+names are ignored. The special name "all" applies to all components. Names
+may be prefixed with '+' or '-' to enable or disable detailed logging for
+a component.
+
+The list of component names is not part of curl's public API. Names may
+be added or disappear in future versions of libcurl. Since unknown names
+are silently ignored, outdated log configurations will not error when
+upgrading libcurl.
```
bagder, line 67: But we still need to document the existing available component names here, don't we? Otherwise it seems very hard to use this function!

bagder (commented):
Thought about how this can be tested, at least rudimentary?

docs/libcurl/curl_global_log_config.3
icing: Sure, can do.

docs/libcurl/curl_global_log_config.3
icing: There is a trade off with this, of course. Do we want to document these names as part of the public API? Do we keep the names then in future versions?

docs/libcurl/curl_global_log_config.3
icing: Changed in recent push.

icing:
> Thought about how this can be tested, at least rudimentary?

Added test cases in pytest.

bagder (approved):

icing:
Ready from my side.

PR #11428, opened 2023-07-12
Bump ngtcp2

tatsuhiro-t:


vszakats (approved):

bagder (approved):

bagder:
Thanks!

nono303:
⚠️ Also require this change for CMake
```patch
diff --git "a/CMakeLists.txt" "b/CMakeLists.txt"
index 49a44eabf..d25c9b388 100644
--- "a/CMakeLists.txt"
+++ "b/CMakeLists.txt"
@@ -602,7 +602,7 @@ if(USE_NGTCP2)
     elseif(HAVE_BORINGSSL)
       find_package(NGTCP2 REQUIRED BoringSSL)
     else()
-      find_package(NGTCP2 REQUIRED OpenSSL)
+      find_package(NGTCP2 REQUIRED QuicTLS)
     endif()
     CheckQuicSupportInOpenSSL()
   elseif(USE_GNUTLS)
diff --git "a/CMake/FindNGTCP2.cmake" "b/CMake/FindNGTCP2.cmake"
index ff0d49e33..5d307371b 100644
--- "a/CMake/FindNGTCP2.cmake"
+++ "b/CMake/FindNGTCP2.cmake"
@@ -71,7 +71,7 @@ endif()
 if(NGTCP2_FIND_COMPONENTS)
   set(NGTCP2_CRYPTO_BACKEND "")
   foreach(component IN LISTS NGTCP2_FIND_COMPONENTS)
-    if(component MATCHES "^(BoringSSL|OpenSSL|wolfSSL|GnuTLS)")
+    if(component MATCHES "^(BoringSSL|QuicTLS|wolfSSL|GnuTLS)")
       if(NGTCP2_CRYPTO_BACKEND)
         message(FATAL_ERROR "NGTCP2: Only one crypto library can be selected")
       endif()
```
...All theses breaking changes in latest ng modules without cmake regression validation (cf. https://github.com/ngtcp2/ngtcp2/issues/858) became hard to follow
(patch https://github.com/curl/curl/pull/11184 + https://github.com/curl/curl/pull/11428  for curl latest curl release)

vszakats:
So far we could use the OpenSSL build path with LibreSSL (3.7) as well. Would these changes keep compatibility with this scenario?

icing:
@vszakats you mean if `libngtcp2_crypto_quictls` can be build using libressl? Is that not more a question for the ngtcp2 project?

vszakats:
@icing: Oh sorry, it indeed is, confused this thread with the one in ngtcp2. In this case, it most likely will continue to work with LibreSSL under the new name.

Issue #11413, opened 2023-07-09
builds with libressl on macOS fail


bagder:
Our CI jobs that use cmake + libressl on macOS have all started to fail with this error output. Presumably they have changed libressl version that the curl build does not like?

The [configure build using libressl](https://github.com/curl/curl/actions/runs/5499059260/jobs/10020987667) fails as well, but only when building the examples(!)

[cmake example build log](https://github.com/curl/curl/actions/runs/5499059260/jobs/10020988270)

~~~
Undefined symbols for architecture x86_64:
  "_EVP_PKEY_get_bn_param", referenced from:
      _Curl_ossl_certchain in unity_0_c.c.o
  "_EVP_PKEY_get_id", referenced from:
      _Curl_ossl_certchain in unity_0_c.c.o
      _cert_stuff in unity_0_c.c.o
  "_OPENSSL_sk_num", referenced from:
      _Curl_ossl_certchain in unity_0_c.c.o
      _X509V3_ext in unity_0_c.c.o
      _cert_stuff in unity_0_c.c.o
      _ossl_verifyhost in unity_0_c.c.o
      _load_cacert_from_memory in unity_0_c.c.o
      _verifystatus in unity_0_c.c.o
  "_OPENSSL_sk_pop", referenced from:
      _cert_stuff in unity_0_c.c.o
  "_OPENSSL_sk_pop_free", referenced from:
      _cert_stuff in unity_0_c.c.o
      _load_cacert_from_memory in unity_0_c.c.o
  "_OPENSSL_sk_value", referenced from:
      _Curl_ossl_certchain in unity_0_c.c.o
      _X509V3_ext in unity_0_c.c.o
      _ossl_verifyhost in unity_0_c.c.o
      _load_cacert_from_memory in unity_0_c.c.o
      _verifystatus in unity_0_c.c.o
  "_SSL_CTX_set_options", referenced from:
      _ossl_connect_step1 in unity_0_c.c.o
  "_SSL_get1_peer_certificate", referenced from:
      _servercert in unity_0_c.c.o
      _verifystatus in unity_0_c.c.o
  "_X509_STORE_load_file", referenced from:
      _populate_x509_store in unity_0_c.c.o
  "_X509_STORE_load_path", referenced from:
      _populate_x509_store in unity_0_c.c.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
~~~

icing:
[...]

PR #11436, opened 2023-07-13
Brew fix for openssl in default path

icing:
- refs #11413
- if brew install/update links openssl into /usr/local, it will be found before anything we add with `-isystem path` to CPP/LDLFAGS. Get rid of that by unlinking the keg.


bagder:
:heavy_check_mark: !

bagder:
Thanks!

stefankiesz:
Hi, does anyone know what's the cause of this behavior? When I build curl using cmake, it finds the correct openssl version with find_package, but still fails because the linker tries to link with the brew installed openssl.

Is there a way to fix this without unlinking?

PR #11438, opened 2023-07-13
Try brew cleanup as advised by homebrew team

icing:
- as mentioned in https://github.com/orgs/Homebrew/discussions/4643 try method 1 to get rid of unwanted formulae


icing:
This method does seem to remove too much. Aborted.

PR #11444, opened 2023-07-16
wolfssl: Detect when TLS 1.2 support is not built into wolfssl

darktohka:
Right now, when building curl with a wolfssl build that has TLS 1.2 disabled, the build fails with curl trying to find the TLSv1_2_client_method symbol.

This PR adds a compile-time check to see if TLS 1.2 is disabled.


bagder (approved):

bagder:
Thanks!

PR #5368, opened 2020-05-10
CMake: Add runtests.pl targets, do not build test programs unless necessary

Lekensteyn:
Executive overview, see invididual commits for details:

 * Fix runtests.pl to work with CMake out-of-tree builds.
 * Add test, test-nonflaky, etc. targets that run runtests.pl (mirrors autotools functionality).

Minor details:

* Remove ctest support since there are no tests and it causes conflicts.
* Bump CMake minimum version from 3.0 to 3.2. It could be bumped further to 3.5 [with no loss of platform support](https://wiki.wireshark.org/Development/Support_library_version_tracking#CMake) if desired.

The next step is to run these tests for the CMake builds in CI, but that would introduce a conflict with my cmake-http3 branch that also modifies the travis config, so I'll wait for that to get merged first.

<details><summary>ninja test-nonflaky summary</summary>

```
TESTDONE: 1117 tests out of 1120 reported OK: 99%
TESTFAIL: These test cases failed: 1014 1026 1139 
TESTDONE: 1352 tests were considered during 1310 seconds.
TESTINFO: 232 tests were skipped due to these restraints:
TESTINFO: "configured as DISABLED" 19 times (323, 530, 584, 594, 836, 882, 938, 1209, 1211 and 10 more)
TESTINFO: "curl has threaded-resolver support" 1 times (506)
TESTINFO: "curl lacks debug support" 85 times (67, 68, 69, 81, 89, 90, 91, 150, 155 and 76 more)
TESTINFO: "curl lacks mqtt support" 7 times (1190, 1191, 1192, 1193, 1194, 1195, 1196)
TESTINFO: "curl lacks Metalink support" 16 times (2005, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015 and 7 more)
TESTINFO: "curl lacks TLS-SRP support" 4 times (320, 321, 322, 324)
TESTINFO: "curl lacks TrackMemory support" 2 times (96, 558)
TESTINFO: "Resolving IPv6 'ip6-localhost' didn't work" 2 times (241, 1083)
TESTINFO: "curl has ipv6 support" 1 times (1454)
TESTINFO: "curl lacks idn support" 14 times (165, 962, 963, 964, 965, 966, 967, 968, 969 and 5 more)
TESTINFO: "curl lacks WinSSL support" 2 times (2043, 2070)
TESTINFO: "disabled by keyword" 9 times (573, 587, 644, 1086, 1113, 1162, 1163, 1208, 2032)
[...]

PR #11446, opened 2023-07-16
cmake: add `libcurlu`/`libcurltool` for unit tests

aloisklink:
Add a `libcurlu`/`libcurltool` static library to the CMake build that is compiled only for unit tests. We use `EXCLUDE_FROM_ALL` to make sure that they're not built by default, they're only built if unit tests are built.

These libraries allow us to compile every unit test with CMake, rather than skipping some of them.

---

To make reviewing easier:
  - `libcurltool` is defined in the `Makefile.am` at: https://github.com/curl/curl/blob/ebd83bfbae31809a4525e953c8512aad797c38ad/src/Makefile.am#L77-L82
  - `libcurlu` is defined in the `Makefile.am` at: https://github.com/curl/curl/blob/ebd83bfbae31809a4525e953c8512aad797c38ad/lib/Makefile.am#L116-L118

You can also easily configure, build, and test everything using:

```bash
# configure the build/ directory for unit tests
cmake -B build/ -S . -DENABLE_THREADED_RESOLVER=OFF -DCMAKE_BUILD_TYPE=Debug -DENABLE_DEBUG=ON
# run `make testdeps all -j12`
cmake --build build/ --target=testdeps --target=all -j12
# run `make test-ci` (aka run tests)
# You may wish to modify tests/CMakeLists.txt to run tests with multiple cores
cmake --build build/ --target=test-ci
```

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -65,6 +65,14 @@ add_library(
   )
 
 add_library(
+  curlu # special libcurlu library just for unittests
+  STATIC
+  EXCLUDE_FROM_ALL
+  ${HHEADERS} ${CSOURCES}
+)
+target_compile_definitions(curlu PUBLIC UNITTESTS CURL_STATICLIB)
+
+add_library(
   ${PROJECT_NAME}::${LIB_NAME}
   ALIAS ${LIB_NAME}
   )
@@ -80,6 +88,7 @@ if(NOT BUILD_SHARED_LIBS)
 endif()
 
 target_link_libraries(${LIB_NAME} PRIVATE ${CURL_LIBS})
+target_link_libraries(curlu PRIVATE ${CURL_LIBS})
 
 transform_makefile_inc("Makefile.soname" "${CMAKE_CURRENT_BINARY_DIR}/Makefile.soname.cmake")
 include(${CMAKE_CURRENT_BINARY_DIR}/Makefile.soname.cmake)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -76,6 +76,14 @@ add_executable(
   ALIAS ${EXE_NAME}
   )
 
+add_library(
+  curltool # special libcurltool library just for unittests
+  STATIC
+  EXCLUDE_FROM_ALL
+  ${CURL_CFILES} ${CURLX_CFILES} ${CURL_HFILES}
+)
+target_compile_definitions(curltool PUBLIC UNITTESTS CURL_STATICLIB)
+
 if(CURL_HAS_LTO)
   set_target_properties(${EXE_NAME} PROPERTIES
     INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -33,18 +33,10 @@ include_directories(
   ${CURL_BINARY_DIR}/include      # To be able to reach "curl/curl.h"
 )
 
-# TODO build a special libcurlu library for unittests.
-# Until that happens, only build the unit tests when creating a static libcurl
-# or else they will fail to link. Some of the tests require the special libcurlu
-# build, so filter those out until we get libcurlu.
-list(FILTER UNITPROGS EXCLUDE REGEX
-  "unit1394|unit1395|unit1604|unit1608|unit1621|unit1650|unit1653|unit1655|unit1660|unit2600|unit2601|unit2602|unit2603")
-if(NOT BUILD_SHARED_LIBS)
+if (ENABLE_CURLDEBUG) # running unittests require curl to compiled with CURLDEBUG
   foreach(_testfile ${UNITPROGS})
     add_executable(${_testfile} EXCLUDE_FROM_ALL ${_testfile}.c ${UNITFILES})
     add_dependencies(testdeps ${_testfile})
-    target_link_libraries(${_testfile} libcurl ${CURL_LIBS})
-    set_target_properties(${_testfile}
-        PROPERTIES COMPILE_DEFINITIONS "UNITTESTS")
+    target_link_libraries(${_testfile} curltool curlu)
   endforeach()
 endif()


bagder:
any cmake person around who wants to give this PR their thumbs up? I wouldn't be able to say if this is done the right way or not.

aloisklink:
> any cmake person around who wants to give this PR their thumbs up?

I found two PRs that touched building unit tests in CMake, see:
 - https://github.com/curl/curl/pull/5368 by @Lekensteyn (reviewed by @jzakrzewski), and
 - https://github.com/curl/curl/pull/10722 by @dfandrich. @dfandrich seems pretty active in curl, so it's probably ideal to get a review from them, if possible!

dfandrich:
I like how this gets rid of the hack I added to build unit tests with a static library, but my knowledge of CMake isn't enough to tell if there are any issues with how it's done. Having said that, it looks fine to me, although I wouldn't have expected there to be any changes necessary to src/CMakeLists.txt

aloisklink:
> Having said that, it looks fine to me, although I wouldn't have expected there to be any changes necessary to `src/CMakeLists.txt`

I've mainly placed it there to match the definition of `libcurltool` in https://github.com/curl/curl/blob/ebd83bfbae31809a4525e953c8512aad797c38ad/src/Makefile.am#L75-L83

(some of the unit tests test functions in the `src/` folder)

We could move the definition to `tests/unit/CMakeLists.txt`, but my gut feeling is to keep the CMake implementation as similar to the existing Autotools implementation, unless there's a more common/better way to do things in CMake.

dfandrich:
I can't spot any objects from src/ in the libcurlu built by autoconf, and
there's no reference to it in the src/ directory. Are you sure this is needed?

aloisklink:
> I can't spot any objects from `src/` in the `libcurlu` built by autoconf, and there's no reference to it in the `src/` directory. Are you sure this is needed?

`libcurlu` doesn't use any objects from `src/`. There's a different library for the `src/` folder that's also needed for some unit tests called `libcurltool`. For example, `tests/unit/unit1394.c` tests a [`parse_cert_parameter()`](https://github.com/curl/curl/blob/2900c29218d2d24ab519853589da84caa850e8c7/tests/unit/unit1394.c#L84) which is defined in the `src/` folder in https://github.com/curl/curl/blob/2900c29218d2d24ab519853589da84caa850e8c7/src/tool_getparam.h#L63-L67

To be honest, most of the unit tests don't need `libcurltool`, but I've linked it to every unit test since it makes the CMake simpler, and that's what the current autoconf build does, see https://github.com/curl/curl/blob/2900c29218d2d24ab519853589da84caa850e8c7/tests/unit/Makefile.am#L48-L50

dfandrich:
I see it now. No objections to merging this.

jzakrzewski:
1. Those targets should probably be build conditionally - thay are not needed if `BUILD_TESTING` is OFF
2. I haven't tried, but it seems to me that `libcurlu` basically is the same as `libcurl` but static. Doesn't it compile the stuff twice? Can we avoid that?

aloisklink:
> 1. Those targets should probably be build conditionally - thay are not needed if `BUILD_TESTING` is OFF

Both the `curlu` and `curltool` library are marked with [`EXCLUDE_FROM_ALL`](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html#prop_tgt:EXCLUDE_FROM_ALL) so they won't be built by default:
https://github.com/curl/curl/blob/ed6ac110d2c850f87b627db837c9cca22e230262/lib/CMakeLists.txt#L70

Instead, they're built automatically whenever you build anything that depends on them. And **since the unit tests are the only targets that depend on `curlu` and `curltool`, they're only built when unit tests are enabled.**

We could hide them behind an `if(BUILD_TESTING AND ENABLE_CURLDEBUG)`, but:
  - it makes no functional difference, and
  - I want to change how `BUILD_TESTING` works in a future PR, since the way it currently works doesn't align with how [most CMake projects expect it work](https://cliutils.gitlab.io/modern-cmake/chapters/testing.html) (which makes it more difficult to use `libcurl` from another CMake project).

Maybe it's worth adding a comment? I think most CMake people will know what `EXCLUDE_FROM_ALL` does, but most curl contributors probably aren't CMake experts.

---

> 2. I haven't tried, but it seems to me that `libcurlu` basically is the same as `libcurl` but static. Doesn't it compile the stuff twice? Can we avoid that?

There is one other difference, it also defines `UNITTESTS`:

https://github.com/curl/curl/blob/ed6ac110d2c850f87b627db837c9cca22e230262/lib/CMakeLists.txt#L73

Defining `UNITTESTS` currently does nothing for `libcurlu` (but `git log -S UNITTESTS lib/` shows a bunch of times where it recently used to do things), but it does make quite a few differences for `libcurltool` in the `src/` folder, since there are a bunch of functions that are normally defined as `static` (aka internal linkage), that are defined without `static` so that they can be used in unit tests if `UNITTESTS` is defined:

https://github.com/curl/curl/blob/2900c29218d2d24ab519853589da84caa850e8c7/src/tool_xattr.c#L46-L51

So, we could avoid compiling everything twice for `libcurlu` (since `UNITTESTS` currently does nothing in `lib/`). My gut feeling is to leave it how it is though, since `git log -S UNITTESTS lib/` shows it being commonly used in the past. If we do want to change it, let me know! We should first make a PR for changing it in automake, though, to keep cmake aligned with automake, something like:

```diff
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -116 +116 @@ libcurl_la_CFLAGS = $(AM_CFLAGS) $(libcurl_la_CFLAGS_EXTRA)
-libcurlu_la_CPPFLAGS = $(AM_CPPFLAGS) -DCURL_STATICLIB -DUNITTESTS
+libcurlu_la_CPPFLAGS = $(AM_CPPFLAGS) -DCURL_STATICLIB
```

jzakrzewski:
> Both the `curlu` and `curltool` library are marked with [`EXCLUDE_FROM_ALL`](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html#prop_tgt:EXCLUDE_FROM_ALL) so they won't be built by default:

Damn, I totally missed the `EXCLUDE_FROM_ALL` there.

> My gut feeling is to leave it how it is though

After reading your analysis, I can agree with that. I feel like there could be some complicated way to re-use everything that does not rely on the additional macro, but I'm not sure if it's worth it.

So it's thumbs up from me.

bagder:
Thank you everyone and @aloisklink in particular. I'm merging!

Issue #11426, opened 2023-07-12
HTTP/2 connection uploads stall in 8.1.0 .. 8.1.2


laramiel:
### I did this

In tensorstore (http://github.com/google/tensorstore) we use curl for HTTP/2 uploads to GCS and other storage services. When trying to upgrade curl to 8.1.1 or later, uploads get stuck. 

This appears to trigger on a load-balanced server which closes the HTTP/2 connection. The client then leaves the connection in a CLOSE_WAIT state, which is never handled.

```
$ netstat -t
Active Internet connections (w/o servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State      
tcp        0      0 linux-repro.us-ce:44210 metadata.google.in:http ESTABLISHED
tcp        0      0 linux-repro.us-ce:44278 metadata.google.in:http ESTABLISHED
tcp        0      0 linux-repro.us-cent:ssh 172.217.41.166:38420    ESTABLISHED
tcp      200      0 linux-repro.us-ce:41612 il-in-f128.1e100.:https CLOSE_WAIT 
```

The process gets stuck with a backtrace in:

```
#0  0x00007fc0f05e596f in __GI___poll (fds=0x7fc0ed2eec00, nfds=1, timeout=2147483647) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00005642d1d8477d in Curl_poll.part.0 ()
#2  0x00005642d1d646bb in multi_wait.part ()
#3  0x00005642d1d64b99 in curl_multi_poll ()
#4  0x00005642d1d581e1 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<tensorstore::internal::Thread::Thread<tensorstore::internal_http::(anonymous namespace)::MultiTransportImpl::MultiTransportImpl(std::shared_ptr<tensorstore::internal_http::CurlHandleFactory>)::{lambda()#2}>(tensorstore::internal::Thread::private_t, tensorstore::internal::Thread::Options, tensorstore::internal_http::(anonymous namespace)::MultiTransportImpl::MultiTransportImpl(std::shared_ptr<tensorstore::internal_http::CurlHandleFactory>)::{lambda()#2}&&)::{lambda()#1}> > >::_M_run() ()
#5  0x00007fc0f07b2ed0 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007fc0f09fcea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
[...]

PR #11450, opened 2023-07-17
HTTP/2, treat initial SETTINGS as a WINDOW_UPDATE

icing:
- refs #11426 where spurious stalls on large POST requests are reported
- the issue seems to involve the following
  * first stream on connection adds up to 64KB of POST data, which is the max default HTTP/2 stream window size transfer is set to HOLD
  * initial SETTINGS from server arrive, enlarging the stream window. But no WINDOW_UPDATE is received.
  * curl stalls
- the fix un-HOLDs a stream on receiving SETTINGS, not relying on a WINDOW_UPDATE from lazy servers


lib/http2.c
```diff
@@ -1122,6 +1123,21 @@ static int on_frame_recv(nghttp2_session *session, const nghttp2_frame *frame,
                       ctx->max_concurrent_streams));
         multi_connchanged(data->multi);
       }
+      /* Since the initial stream window is 64K, a request might on on HOLD,
```
bagder, line 1126: "might on on HOLD" ?

lib/http2.c
```diff
@@ -1122,6 +1123,21 @@ static int on_frame_recv(nghttp2_session *session, const nghttp2_frame *frame,
...
+       * due to exhaustion. The (initial) SETTINGS may announce a much larger
+       * window and *assume* that we treat this like a WINDOW_UPDATE. Some
+       * servers send an explicit WINDOW_UPDATE, but not all seem to do that.
+       * To be safe, we UNHOLD a stream in order not to stall. */
+      if((data->req.keepon & KEEP_SEND_HOLD) &&
+         (data->req.keepon & KEEP_SEND)) {
+        data->req.keepon &= ~KEEP_SEND_HOLD;
+        stream = H2_STREAM_CTX(data);
```
bagder, line 1134: Maybe declare `stream` here as it is not use outside of this block?

icing: fixed.

PR #11452, opened 2023-07-17
wolfssl: support loading system CA certificates

darktohka:
The wolfssl backend currently does not support loading the system CA certificates. However, wolfSSL has built-in support for this functionality.

This PR ensures that system CA certs are loaded when the CURLSSLOPT_NATIVE_CA bit is set.

In case the system CA store cannot be read, the program will continue. If the system CA store was read successfully, but certificates could not be read either from memory or from disk, the program will still continue. This is the same behaviour that the OpenSSL backend exhibits.

### DIFF
--- a/.github/scripts/spellcheck.words
+++ b/.github/scripts/spellcheck.words
@@ -150,6 +150,7 @@ cyassl
 Cygwin
 daniel
 datatracker
+Debian
 decrypt
 deepcode
 DELE
@@ -216,6 +217,7 @@ Falkeborn
 Fandrich
 Fastly
 fcpp
+Fedora
 Feltzing
 ffi
 filesize
@@ -243,6 +245,7 @@ gcc
 GCM
 gdb
 Genode
+Gentoo
 Gergely
 getaddrinfo
 getenv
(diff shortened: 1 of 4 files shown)


bagder:
The wolfSSL function used for this purpose supports many more native CA stores, so maybe this should be reflected in the docs? See https://github.com/wolfSSL/wolfssl/pull/6629

darktohka:
Mentioning more native CA stores would mean that we'd have to sync the list of supported CA stores with the wolfSSL documentation. Is that okay?

bagder:
This PR unfortunately didn't make it into 8.2.0 due to lack of time. It will instead be targeted for 8.3.0 but could use that update in the docs.

darktohka:
Docs have been updated: https://github.com/curl/curl/pull/11452/commits/eb64b9d221e65bd7afc2ec203ee89d50a142da86

bagder:
Thanks!

Megamouse:
Sorry to post on a closed PR, but this does not necessarily compile on msvc unless you build wolfssl with WOLFSSL_SYS_CA_CERTS.
I don't know if that's just some random issue with my build setup or if it's an issue with the ifdefs used in curl/wolfssl.

bagder:
If this is an issue, then file an issue. Just commenting here has little effect.

PR #11454, opened 2023-07-17
quiche: avoid NULL deref in debug logging

bagder:
Coverity reported "Dereference after null check"

If stream is NULL and the function exits, the logging must not deref it.


icing (approved):
Also fine, in #11455 I opted for an early return when stream is NULL.

bagder:
> Also fine, in #11455 I opted for an early return when stream is NULL.

I believe they're both decent choices. As I already did it this way I think I'll let it be for now.

Issue #11449, opened 2023-07-17
Segmentation fault when sending http3-only request by multi interface


Cering:
### I did this
- Send `CURL_HTTP_VERSION_3ONLY` request parallelly by using multi interface
- After some requests finished, a segmentation fault occurred
- This segmentation fault occurred when using the **latest** release version(`libcurl/8.1.2+quiche/0.17.1`), but not found in previous version(`libcurl/8.0.1+quiche/0.16.0`)

### I expected the following
- Running the test code(see below) in `gdb`, the segmentation fault info: 
```
Program received signal SIGSEGV, Segmentation fault.
bufcp_take (pchunk=<synthetic pointer>, pool=0xc783a8) at bufq.c:191
191         pool->spare = chunk->next;
```

### Test code
```c++
#include <string>
#include <vector>
#include <sstream>
#include <unistd.h>
#include "curl/curl.h"

size_t mycurl_onrecv_body(char *ptr, size_t size, size_t nmemb, void *userdata)
{
    return size * nmemb;
}

void mycurl_process(CURLM *multi_handle)
{
    CURLMsg *curl_msg = nullptr;
    int msgs_left = 0;
    while((curl_msg = curl_multi_info_read(multi_handle, &msgs_left))) {
        if(curl_msg->msg == CURLMSG_DONE) {
            CURL *http_handle = curl_msg->easy_handle;
            CURLcode code = curl_msg->data.result;
            curl_multi_remove_handle(multi_handle, http_handle);
            printf("handle %p finished\n", http_handle);
            curl_easy_cleanup(http_handle);
        }
    }
    return;
}

int main()
{
    curl_global_init(CURL_GLOBAL_ALL);
    printf("%s\n", curl_version());

    CURLM *multi_handle = curl_multi_init();

    int reqid = 0;
    std::vector<std::string> url_list {
        "https://cloudflare-quic.com/",
        "https://cloudflare-quic.com/",
        "https://cloudflare-quic.com/",
    };

    int count = 5;
    while(count--) {
        int still_running = 0;
        for(auto url : url_list) {
            printf("New Request[%d]: %s\n", reqid++, url.c_str());
            CURL *http_handle = curl_easy_init();
            curl_easy_setopt(http_handle, CURLOPT_URL, url.c_str());
            curl_easy_setopt(http_handle, CURLOPT_VERBOSE, 1L);
            curl_easy_setopt(http_handle, CURLOPT_WRITEFUNCTION, mycurl_onrecv_body);
[...]

PR #11469, opened 2023-07-19
HTTP/3, quiche, fix segfault and other things

icing:
- refs #11449 where a segfault is reported when IP Eyeballing did not immediately connect but made several attempts
- The transfer initiating the eyeballing was initialized  too early, leadding to references to the filter instance that was then replaced in the subsequent eyeball attempts. That led to a use after free in the buffer handling for the transfer
- transfers are initiated now more lazy (like in the ngtcp2 filter), when the stream is actually opened
- suppress reporting on quiche event errors for "other" transfers than the current one to not fail a transfer due to faults in another one.
- revert recent return value handling for quiche_h3_recv_body() to not indicate an error but an EAGAIN situation. We wish quiche would document what functions return.

### DIFF
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -426,7 +426,7 @@ static ssize_t stream_resp_read(void *reader_ctx,
     return nread;
   }
   else {
-    *err = stream->resp_got_header? CURLE_PARTIAL_FILE : CURLE_RECV_ERROR;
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
You meant to reference #11449 right?

icing:
> You meant to reference #11449 right?

Yes.

PR #11474, opened 2023-07-19
bug_report: use issue forms instead of markdown template

jamacku:
Issue forms allow you to define "web-like" input forms using YAML syntax. It allows you to guide the reporter to get the required information.

An example is available here: https://github.com/jamacku/curl/issues/new?template=bug_report.yml
You can experiment using the link above.

---

Screenshot from 2023-07-19 16-39-13

Example of the report  filled using issue form:

Screenshot from 2023-07-19 16-40-28

---

When a user would like to report a bug using the plain issue, it's always possible to use: https://github.com/curl/curl/issues/new


jamacku:
To be transparent, issue forms are currently a GitHub Beta feature. But it's been like that for [2 years now](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/).

We have been using it since the beginning in [systemd upstream](https://github.com/systemd/systemd/issues/new/choose), and I haven't noticed any issues with this feature.

bagder (approved):
Let's try this!

bagder:
Thanks!

PR #11480, opened 2023-07-19
amigaos: fix connect() macro expansion clash

Futaura:
The new connection filter code contains a structure with an item called "connect". On AmigaOS the standard networking connect() function is implemented via a macro. Parentheses have been added to the ->connect() calls to prevent them triggering expansion of the connect() macro.

### DIFF
--- a/lib/cf-h1-proxy.c
+++ b/lib/cf-h1-proxy.c
@@ -1074,7 +1074,7 @@ static CURLcode cf_h1_proxy_connect(struct Curl_cfilter *cf,
   }
 
   DEBUGF(LOG_CF(data, cf, "connect"));
-  result = cf->next->cft->connect(cf->next, data, blocking, done);
+  result = (cf->next->cft->connect)(cf->next, data, blocking, done);
   if(result || !*done)
     return result;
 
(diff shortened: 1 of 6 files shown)


bagder:
That's a fricking lame set of headers. I propose you instead change the function pointer name so that we don't have to use the funny parentheses approach which is just asking for future breakage since you have no Amiga builds in the CI. Maybe just call it `cnnct` or something, and add a like explanation as a comment as to why it can't be called `connect`. Or do you have a better idea @icing ?

Futaura:
Looking at the rest of `struct Curl_cftype`, perhaps `do_connect` (and `do_close`) would be more appropriate, which would be consistent with the existing `do_recv` and `do_send`? I'm happy to create a new PR to do it this way.

An alternative solution would be to leave it as-is and add a `connect()` stub function in `lib/amigaos.c`, although that would make things more messy.

icing:
> Looking at the rest of `struct Curl_cftype`, perhaps `do_connect` (and `do_close`) would be more appropriate, which would be consistent with the existing `do_recv` and `do_send`? I'm happy to create a new PR to do it this way.
> 
> An alternative solution would be to leave it as-is and add a `connect()` stub function in `lib/amigaos.c`, although that would make things more messy.

I think renaming the function pointers to "do_*" is fine.

bagder:
Yeah, `do_*` is better.

Futaura:
Opened new PR #11491 to replace this one.

PR #11482, opened 2023-07-19
curl.rc: enable UTF-8 support for the (non-Unicode) ANSI API

### LINKED ISSUE #11461: When using the -o option, some characters changed
### I did this
cmd
"curl.exe" -o "ポ.html" "https://www.google.com/"
### I expected the following

I expect to run this command and save "ポ.html"
But what was actually saved was "ダ.html"
("ポ" reads po. "ダ" reads da in japanese)
When specifying a path with the -o option, if "ポ" is included, it seems to change to a different character for some reason.
I have confirmed that this bug occurs in version 8.1.2_5, but not in version 8.0.1.

### curl/libcurl version

curl 8.1.2 (x86_64-w64-mingw32) libcurl/8.1.2 OpenSSL/3.1.0 (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.5 WinIDN libssh2/1.11.0 nghttp2/1.55.1 ngtcp2/0.15.0 nghttp3/0.11.0 libgsasl/2.2.0
Release-Date: 2023-05-30
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe UnixSockets zstd

### operating system
Windows10 64bit

vszakats:
Enable the switch in the curl.exe manifest to use this feature.

This allows Windows in supported versions (Windows Version 1903
from 2019 May) to use UTF-8 encoded Unicode strings in its 8-bit
ANSI API flavour.

This patch also switches back to using `fwrite()` instead of 
`WriteConsoleW()` when writing to a console that has `CP_UTF8`
set as its "ANSI" codepage. This allows to drop an internal CP 
translation from UTF-8 to UTF-16 ("WIDE") when the "ANSI" API
already speaks UTF-8.

https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

Ref: https://github.com/curl/curl/issues/11461#issuecomment-1642393916
Fixes #11461
Closes #11482

---

Notice that this patch relies on the referenced Microsoft document and speculation. I have no Windows machine to test this on. Feedback is welcome if this resolves Unicode issues when using a non-Unicode `curl.exe`, and also whether there is any regression with older Windows versions, Unicode `curl.exe` builds or other build / environment combinations.


vszakats:
Notice that this patch only relies on the references Microsoft document and speculation. I have no Windows machine to test this. All feedback is welcome if this resolves Unicode issues when using a non-Unicode `curl.exe`, and also whether there is any regression with older Windows versions, Unicode `curl.exe` builds or other combinations.

src/tool_cb_wrt.c
```diff
@@ -234,7 +234,8 @@ size_t tool_write_cb(char *buffer, size_t sz, size_t nmemb, void *userdata)
 #ifdef WIN32
   fhnd = _get_osfhandle(fileno(outs->stream));
   if(isatty(fileno(outs->stream)) &&
-     GetConsoleScreenBufferInfo((HANDLE)fhnd, &console_info)) {
+     GetConsoleScreenBufferInfo((HANDLE)fhnd, &console_info) &&
+     GetACP() != CP_UTF8) {
```
MichalPetryka, line 238: This should probably use `GetConsoleOutputCP` instead.
I recall something about console corrupting UTF8 output though when a multibyte character lands on the end of the buffer, not sure if this is a good idea then.

vszakats: Good point, and quite possible. The goal would be to avoid the CRT doing any translations and just passing it to `WriteConsole()` internally. Reading the UCRT [source code](https://github.com/huangqinjin/ucrt), and still cannot decide which would be the one we need. It seems we'd need the handle opened in text mode for that, and we seem to be opening it in binary mode, so probably it's broken either way. We can resort to `WriteConsoleA()` in curl, but there is a chance this will cut multibyte chars into pieces. `fwrite()` seems to be getting this right, but hard to say without tests.

jay:
I think we should enable UTF-8 in the manifest only for Unicode builds. I don't know if there's an easy way to do that with a condition. And I would discard the changes for terminal writing, we can just use the code that converts it to UTF-16 and handles partial sequences.

vszakats:
As far as I understood it (maybe incorrectly) the point of this manifest setting is to help non-Unicode builds, aka the "ANSI" win32 API to handle UTF-8 transparently. Just like Unix did for a long time. This setting is not supposed to affect the "WIDE" API, which Unicode-enabled builds use. If this is true, the manifest setting is only necessary for non-Unicode builds and is a no-op for Unicode ones.

As for console output, I think the ultimate goal is to omit any CP/encoding translation inside curl on Windows, if we can avoid it, and let strings through transparently as we always did on non-Windows platforms. Doing translations would at best reduce performance, and pose some risk of doing a mistake there.

I haven't done any tests, so this is all based on the quoted MS document and possibly a dose of wishful thinking.

jay:
I think Unicode builds should be handling text in Unicode, not all builds. It would be unexpected to make curl builds without Unicode enabled use UTF-8 and break users who specify filenames or other things in local encodings.

As you know Unicode support is currently lacking because there are a number of places we don't know the encoding. We have _some_ support for Unicode in Unicode builds but it could be improved by making the ANSI versions of the API use UTF-8 encoding. That API if I understand correctly is still called in Unicode builds by the CRT. This way, for all the places we didn't want to get sucked into a maintenance rabbit hole, the CRT should (if I understand correctly) either handle it internally as utf-8 by its own locale settings or call the ANSI api functions whateverA which will use UTF-8.

While it would be great to have a curl Unicode build just need one manifest change rather than all the conversion we do, the  change is not supported everywhere. I see adding a manifest change for Unicode builds as an improvement and not a replacement.

vszakats:
Though I have never tested the CRT in Unicode mode (in another project we just choose to not use the CRT at all), my understanding [^1]  was that the `_UNICODE` macro instructs the CRT to take the "WIDE" codepath as well. If this is not the case here, the result is indeed a "hybrid" solution (aside: I'm not quite sure how this could result in an end-to-end Unicode app). In such case we may limit adding this manifest setting to non-Unicode builds.

As for the manifest setting: Again my understanding, is that it's saying to the OS, that this application can safely use UTF-8 as it's "ANSI" codepage. Hence the `*A()` win32 API calls are enabled to handle UTF-8 for that application. Actually setting the "ANSI" codepage to UTF-8 is up to the user, via a global Windows setting (as described in the article, also screenshot here: [^2]). Meaning: Adding this manifest setting will _not change_ the behaviour of a non-Unicode curl build for existing users with an 8-bit "ANSI" codepage configured. It will however make it handle UTF-8 transparently, as soon as the user enables this global Window setting.

[^1]: https://learn.microsoft.com/en-us/cpp/text/unicode-programming-summary
[^2]: via https://superuser.com/questions/269818/change-default-code-page-of-windows-console-to-utf-8: Vfm3a

MarcelRaad:
> As for the manifest setting: Again my understanding, is that it's saying to the OS, that this application can safely use UTF-8 as its "ANSI" codepage.

Hmm, my understanding of 
 https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#set-a-process-code-page-to-utf-8 is that it actually forces UTF-8, overriding the user's choice.

vszakats:
Right.. so the global Windows setting only affects "GDI" (meaning gdi32.dll APIs?), while the rest is always UTF-8. Weird. curl uses no GDI, meaning it's always on.

A next question, in what actual cases could an always-on UTF-8 break compatibility? Bearing in mind that Unicode builds already do break compatibility in certain cases. If UTF-8-enabled non-Unicode are overall better, it might still be a desirable solution.

[ I'm guessing that passing non-7-bit-ASCII strings via the command-line (or env) and expecting them to result in a certain byte-stream on the wire, might be one of the breaking ones, just like in Unicode-builds. (I find this one a reasonable trade-off, and can be mitigated by allowing to pass hex-encoded strings via the command-line at least.) ]

jay:
> _UNICODE macro instructs the CRT to take the "WIDE" codepath as well

Well sort of. `_UNICODE` makes it so the tchar macros map to wide character versions. See [here](https://learn.microsoft.com/en-us/cpp/text/generic-text-mappings-in-tchar-h?view=msvc-170). It's like how `UNICODE` maps the Win32 API function macros to the W suffixed function names. It's still possible to call either functions directly. For the CRT we may call the char version of functions and not the wchar version. And the same for Win32 API though if we're calling an A suffixed function directly it's pretty rare. This is why we do not have full Unicode support in Unicode builds.

> A next question, in what actual cases could an always-on UTF-8 break compatibility?

Any case where the user is passing some string in their current locale. Most likely a path, possibly authentication. It would (again, if I understand this correctly..) turn non-Unicode builds into Unicode.

vszakats:
@jay Thanks for the info, got it. It means that making the Unicode build actually Unicode we'd also need to swap all CRT calls to their Unicode-prepared (MS-specific) flavours with tchar or wchar, e.g. fopen() to _wfopen() or _tfopen(). Or possibly tap into the "ANSI"-as-UTF-8 feature and make sure to translate everything to UTF-8 before passing. If we already keep UTF-8 in memory in Unicode builds, having this manifest setting might actually help things there as well.

Back to the non-Unicode version, I believe any next step will need actual testing to clear the "if"s around this. This feature still seems the most feasible way to get curl reach a fully Unicode-capable build. IMO.

jay:
> Back to the non-Unicode version, I believe any next step will need actual testing to clear the "if"s around this. This feature still seems the most feasible way to get curl reach a fully Unicode-capable build. IMO.

What I'm saying is we have a Unicode build that it seems with this manifest change could be fully capable in Windows 10 and later. AFAICT the manifest change would fix the issues it has. It might also need a setlocale(LC_ALL, ""); for the CRT to call GetACP and then inherit the UTF-8 setting. I don't get why you want to turn non-Unicode builds into Unicode.

vszakats:
What you suggest would result in a hybrid build that does a bunch of fragile translations back and forth between encodings, _plus_ relies on this OS feature to do the rest. On older Windows versions, it would result in the same incomplete and undefined and incompatible solution that is the current Unicode build. Whereas the non-Unicode build with this setting would work just like it was on older Windowses, and would work in end-to-end Unicode mode on newer ones. This without all the unique code paths and complexity of the WIDE support (both in curl and in any dependency that it uses). Reducing complexity and using the same codepath for Windows and every other platform, is useful on its own.

I'd turn the question and ask, why would we want to use WIDE APIs here and there, when the ANSI API can perfectly do the job everywhere (including dependencies)?

jay:
> I'd turn the question and ask, why would we want to use WIDE APIs here and there, when the ANSI API can perfectly do the job everywhere (including dependencies)?

Legacy. Because we support Unicode in older versions of Windows, even if it's incomplete. If the locale in a Unicode build is forced to UTF-8 in Windows 10 version 1903 and later then there would still be conversions between UTF-8 and UTF-16. I don't see anything wrong with this and Windows does a lot of that itself. Either we are going to stop supporting older versions or that code is going to continue to stay. If it does, so what? The performance hit is minor and we have _some_ Unicode support in older versions and full Unicode support in later versions.

In my opinion your position is a bigger inconsistency because it's creating an executable that may or may not function as Unicode depending on what's running it. You are now proposing a non-Unicode build with a forced locale of UTF-8 encoding. That is curl will use the user's current locale prior to Windows 10 version 1903, and then later versions it would effectively function as Unicode.

vszakats:
While I believe the WIDE codepath is far from being usable as it is now, I don't see why you want to block this feature from non-Unicode builds?

By enabling UTF-8 for _both_ Unicode and non-Unicode builds, we end-up with a non-Unicode build that is actually supporting Unicode NOW, on newer Windows. And a Unicode build that is the exact same as before and possibly working better on newer Windows. Work can continue on the end-to-end WIDE support to eventually support older Windows versions.

If the concern is introducing incompatibilities for the non-Unicode builds on newer Windows, this is reasonable, but solvable by putting this behind a new build-time option, e.g. `CURL_WINDOWS_UTF8`. But, since we need to deal with same or similar issues with the Unicode build too, it's also an opportunity to fix those quicker, so I'm not too concerned about that downside.

vszakats:
This thread turned into a non-Unicode vs Unicode build contest, while IMO both could only benefit. My goal was to highlight this feature in the hope someone can make tests to turn this theory into actual test results, but otherwise I have no horses in this. Closing.

vszakats:
Related article:
https://nullprogram.com/blog/2021/12/30/

vszakats:
Report about using this solution successfully: https://mastodon.social/@berglerma/113804277697365905

PR #11483, opened 2023-07-20
socks: print IPv6 address in square brackets

Chilledheart:
IPv6 address with port number should be printed with square brackets.

> The [] style as expressed in [RFC3986] SHOULD be
> employed, and is the default unless otherwise specified.

refer: https://www.rfc-editor.org/rfc/rfc5952#section-6

with command line: `curl -v --proxy socks5://localhost:1083 -I http://localhost:12345`

previous: 
```
* SOCKS5 connect to IPv6 ::1:12345 (locally resolved)
* SOCKS5 request granted.
```

current:
```
* SOCKS5 connect to IPv6 [::1]:12345 (locally resolved)
* SOCKS5 request granted.
```

### DIFF
--- a/lib/socks.c
+++ b/lib/socks.c
@@ -830,9 +830,25 @@ CONNECT_RESOLVED:
       return CURLPX_RESOLVE_HOST;
     }
 
-    Curl_printable_address(hp, dest, sizeof(dest));
-    destlen = strlen(dest);
-    msnprintf(dest + destlen, sizeof(dest) - destlen, ":%d", sx->remote_port);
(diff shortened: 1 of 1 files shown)


bagder:
It seems unnecessary to add that much new (ifdef'ed) code when we already have most of the logic in place. I propose #11484 as an alternative fix for the same result.

Chilledheart:
Your pr is better(more consistent with other part which prints IPv6 address with brackets as well). Feel free to close this one if yours gets merged.

PR #11490, opened 2023-07-20
Ability to disable specific auths (Basic, Bearer, Digest, etc.) and non-standard like AWS.

wyattoday:
In the spirit of [tiny-curl](https://curl.se/tiny/) deprecated, optional, and non-standard code should be able to be compiled-out.

[AWS-SIG4](https://curl.se/libcurl/c/CURLOPT_AWS_SIGV4.html) is an organization-specific (namely, Amazon) chunk of code that is not applicable outside of communicating with Amazon servers. It is not standardized. Personally, I think this should be disabled by default in libCurl (and maybe even curl-cli too), but a good place to start is even adding the ability to disable it.

Digest Auth is actually a standard (unlike AWS-SIG4) but has [long-since become deprecated](https://www.rfc-editor.org/rfc/rfc6331.html) (both by standards organizations and by [end-user](https://learn.microsoft.com/en-us/windows/win32/secauthn/microsoft-digest-authentication) [corporations](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest) that had [previously used](https://en.wikipedia.org/wiki/Digest_access_authentication#Deprecations) Digest).

Lastly, Bearer, Basic, Negotiate, and Kerberos auth types are also configurably disabled. These are all safe authentication methods, but not every user of curl / libcurl will need them. Hence, reduce compiled binary size (and executed run-time checks) when they're not needed.

This PR continues that by adding 6 configure options / defines:

```
--disable-basic-auth
--disable-bearer-auth
--disable-digest-auth
--disable-kerberos-auth
--disable-negotiate-auth
--disable-aws
```

```
CURL_DISABLE_BASIC_AUTH
CURL_DISABLE_BEARER_AUTH
CURL_DISABLE_DIGEST_AUTH
CURL_DISABLE_KERBEROS_AUTH
CURL_DISABLE_NEGOTIATE_AUTH
CURL_DISABLE_AWS
```

This also removes the general `CURL_DISABLE_CRYPTO_AUTH` ([as suggested below](https://github.com/curl/curl/pull/11490#discussion_r1285272823) by @bagder ) since the specific auth-types can be disabled one-by-one.

**Personally, I think both Digest & AWS should be disabled by default. But this is a good start.**

We already deploy and use versions of libcurl with these auths (and AWS) disabled (and have for a few years now).


dfandrich:
--disable-crypto-auth is already there to disable them all, Basic notwithstanding.

wyattoday:
> --disable-crypto-auth is already there to disable them all, Basic notwithstanding.

Yeah, Basic isn't in there. And there's the case where you need NTLM but `--disable-crypto-auth` removes that ability.

Plus this removes AWS and give the ability for fine-grained disabling of individual auths that aren't needed.

configure.ac
```diff
@@ -3974,6 +3974,82 @@ AS_HELP_STRING([--disable-crypto-auth],[Disable cryptographic authentication]),
        AC_MSG_RESULT(yes)
 )
 
+dnl ************************************************************
+dnl disable basic authentication
+dnl
+AC_MSG_CHECKING([whether to enable basic authentication method])
+AC_ARG_ENABLE(basic-auth,
+AS_HELP_STRING([--enable-basic-auth],[Enable basic authentication])
```
vszakats, line 3982:
```suggestion
AS_HELP_STRING([--enable-basic-auth],[Enable basic authentication (default)])
```

Indicating which one is the default would be useful IMO. Same for the rest of the new settings.

vszakats: Existing options could also benefit greatly if their defaults would be known. It's also difficult to find it out after the fact. It'd be a good opportunity to start adding them with newly added options. The less work it will require to retrofit the missing ones later.

[IMO]

bagder:
> --disable-basic-auth
> --disable-bearer-auth
> --disable-digest-auth
> --disable-aws

Shouldn't this then also include ntlm and negotiate to the mix so that they all are independently controlled the same way?

wyattoday:
> > --disable-basic-auth
> > --disable-bearer-auth
> > --disable-digest-auth
> > --disable-aws
> 
> Shouldn't this then also include ntlm and negotiate to the mix so that they all are independently controlled the same way?

`CURL_DISABLE_NTLM` already exists, but you're right, I've added `CURL_DISABLE_NEGOTIATE`.

MarcelRaad:
Would it make sense to change `CURL_DISABLE_CRYPTO_AUTH` to only set these other defines and remove all checks for `CURL_DISABLE_CRYPTO_AUTH`?

Also, CMake is missing. Apart from that, this PR looks good to me!

configure.ac
```diff
@@ -3960,7 +3960,7 @@ dnl disable cryptographic authentication
 dnl
 AC_MSG_CHECKING([whether to enable cryptographic authentication methods])
 AC_ARG_ENABLE(crypto-auth,
-AS_HELP_STRING([--enable-crypto-auth],[Enable cryptographic authentication])
+AS_HELP_STRING([--enable-crypto-auth],[Enable cryptographic authentication (default)])
```
bagder, line 3963: I think you should drop the "crypto auth" part and go all in on the disabling them all one by one instead.

wyattoday: I agree. I'll do that. I'll have a couple of commits addressing this and CMake stuff later this week.

docs/CURL-DISABLE.md
```diff
@@ -12,6 +12,26 @@ Disable support for HTTP cookies.
...
+
+Disable support for the Digest authentication methods.
+
+## `CURL_DISABLE_NEGOTIATE_AUTH`
+
+Disable support for the negotiate authentication methods.
+
+## `CURL_DISABLE_AWS`
+
+Disable AWS-SIG4 support.
```
bagder, line 33:
```suggestion
Disable **AWS-SIG4** support.
```

I think this fixes the spelling complaint.

wyattoday: Yep, that fixed it.

wyattoday:
@MarcelRaad

> Also, CMake is missing and the spellcheck job is failing. Apart from that, this PR looks good to me!

Fixed the spelling and added the CMake changes.

@bagder

I'm all done making the requested changes. All the tests are passing (except for some failures due to the brittleness of the tests, not the code being tested).

Should be good for a final review and/or merge.

Let me know if there's anything else you want me to address.

configure.ac
wyattoday: Done!

bagder (approved):

bagder:
Thanks!

PR #11491, opened 2023-07-20
cfilters: rename close/connect functions to avoid clashes

Futaura:
Rename `close` and `connect` in `struct Curl_cftype` for consistency and to avoid clashes with macros of the same name (the standard AmigaOS networking connect() function is implemented via a macro).


MarcelRaad (approved):

bagder (approved):

bagder:
Thanks!

PR #11493, opened 2023-07-20
    Pedantry/semantic-alignment between functions, docs, comments

bch:
with respect to websocket protocol code; No functional change
    intended.

    * "totalsize", "framesize" becomes "fragsize" (we deal in frame fragments).

    * "sendflags" becomes "flags"

    * use canonical CURL *handle


bagder (approved):

bagder:
Thanks!

PR #11499, opened 2023-07-21
vtls/openssl: Support async certificate verification by user callback

pabusse:
# Overview
Update the OpenSSL connect state machine to handle `SSL_ERROR_WANT_RETRY_VERIFY`. This allows libcurl users to suspend processing while waiting for external I/O during certificate validation.

## Intended usage
Clients perform asynchronous certificate verification by installing a X.509 certificate verification function with `SSL_CTX_set_cert_verify_callback` from the SSL context callback (see `CURLOPT_SSL_CTX_FUNCTION`).

The verification function calls `SSL_set_retry_verify` before returning `0` to indicate that the outcome is still pending. Once verification completes, the client calls `curl_multi_wakeup` to re-enter the `SSL_connect` state machine and return the final verification outcome from the next call to the verification function.

This mechanism is useful for applications that use OpenSSL for TLS protocol handling and crypto while delegating certificate verification to the platform's native trust store.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -3864,6 +3864,12 @@ static CURLcode ossl_connect_step2(struct Curl_cfilter *cf,
       return CURLE_OK;
     }
 #endif
+#ifdef SSL_ERROR_WANT_RETRY_VERIFY
+    if(SSL_ERROR_WANT_RETRY_VERIFY == detail) {
+      connssl->connecting_state = ssl_connect_2;
+      return CURLE_OK;
+    }
+#endif
     else if(backend->io_result == CURLE_AGAIN) {
       return CURLE_OK;
     }


bagder:
> This allows libcurl users to suspend processing while waiting for external I/O during certificate validation

Shouldn't this then also be documented somewhere?

pabusse:
> Shouldn't this then also be documented somewhere?

Added a section on custom certificate verification with OpenSSL. Please take a look.

jay:
This is highly advanced and too niche to document in SSLCERTS. Instead, I suggest add a concise paragraph above the WARNING paragraphs in [CURLOPT_SSL_CTX_FUNCTION](https://curl.se/libcurl/c/CURLOPT_SSL_CTX_FUNCTION.html): "For OpenSSL, error code SSL_ERROR_WANT_RETRY_VERIFY is supported for custom certificate verification. (Added in 8.3.0)". Anyone who is using a custom CTX function it's up to them to investigate all of the details. There's such a thing as too much documentation.

docs/libcurl/opts/CURLOPT_SSL_CTX_FUNCTION.3
```diff
@@ -61,6 +61,9 @@ necessary. For example, you can use this function to call library-specific
 callbacks to add additional validation code for certificates, and even to
 change the actual URI of an HTTPS request.
 
+For OpenSSL, error code \fISSL_ERROR_WANT_RETRY_VERIFY\fP is supported for
+custom certificate verification. (Added in 8.3.0)
```
bagder, line 65: I don't understand.

The callback set with `CURLOPT_SSL_CTX_FUNCTION` is documented to use a `CURLcode` as return value. The man page example shows how it returns `CURLE_OK`.

How exactly is this "error code" set by the callback?

jay: Yes it could be misinterpreted. How about "For OpenSSL, asynchronous certificate verification via SSL_set_retry_verify is supported. (Added in 8.3.0)"

pabusse: Updated with your suggestion, thanks @jay!

OpenSSL's documentation of [SSL_set_retry_verify](https://www.openssl.org/docs/man3.1/man3/SSL_set_retry_verify.html) is a good entry point. It also points the reader to [SSL_CTX_set_cert_verify_callback](https://manpages.debian.org/testing/libssl-doc/SSL_CTX_set_cert_verify_callback.3ssl.en.html) which needs to be called from `CURLOPT_SSL_CTX_FUNCTION` to install the custom certificate verification function.

jay:
Thanks

PR #11503, opened 2023-07-23
src/mkhelp: strip off escape sequences

### LINKED ISSUE #11501: curl --manual includes ESC-sequences
### I did this

Original user bug-report in to the forum.ru-board [post](https://forum.ru-board.com/topic.cgi?forum=5&topic=39466&start=40#12), (translated) :

>**ozr1**
>
>I downloaded curl 8.2.0_1 from the project suite, I launched "curl --manual> curl.txt" and saw that the file was clogged with >ESC-sequence which was not noticed behind the previous ones (8.1.2 and below).
>
>Accordingly, the question is: how to read this "manual" now? Or, how to remove all these ESC-sequence from it from it?
>
>Sent: 17:14 22-07-2023"

try test use command:

curl --manual > manual.txt

then see this file in to any text editor, see example

### I expected the following

Clean ASCII text w/o ESC sequence's like cURL 8.1.2.

### curl/libcurl version

cURL 8.2.0

### operating system

Any (all) OS and platform's

bagder:
At some point the nroff command stopped stripping off escape sequences, so then this script needs to do the job instead.

Reported-by: VictorVG on github
Fixes #11501

### DIFF
--- a/src/mkhelp.pl
+++ b/src/mkhelp.pl
@@ -51,6 +51,7 @@ while (<STDIN>) {
     # remove trailing CR from line. msysgit checks out files as line+CRLF
     $line =~ s/\r$//;
 
+    $line =~ s/\x1b\x5b[0-9]+m//g; # escape sequence
     if($line =~ /^([ \t]*\n|curl)/i) {
         # cut off headers and empty lines
         $wline++; # count number of cut off lines


dfandrich:
Passing the -c option to nroff should fix it more directly.

bagder:
> Passing the -c option to nroff should fix it more directly.

Yes it might, but do you think that's better? It requires us to rely on that option working everywhere or checking for it, rather than controlling our own destiny in our script.

dfandrich:
But, aren't the escape sequences dependent on the terminal in use? So running
it on an oddball terminal could leave other control characters in-line that
won't be stripped.

bagder:
> aren't the escape sequences dependent on the terminal in use?

Oh, right good point. I bet there is an odd terminal somewhere that would prove you right...

bagder:
Oops, `nroff -c` does not fix this issue either. It just makes it some other highlight method:

~~~
$ nroff -man docs/curl.1 | hd | head -1
00000000  1b 5b 34 6d 63 75 72 6c  1b 5b 32 34 6d 28 31 29  |.[4mcurl.[24m(1)|
$ nroff -c -man docs/curl.1 | hd | head -1
00000000  5f 08 63 5f 08 75 5f 08  72 5f 08 6c 28 31 29 20  |_.c_.u_.r_.l(1) |

bagder:
> `nroff -c` does not fix this issue

Sticking to this PR for now. If we run into problems somewhere, let's take it then.

PR #11505, opened 2023-07-24
cmake: support building static and shared libcurl in one go

vszakats:
This patch adds the ability to build a static and shared libcurl library
in a single build session. It also adds an option to select which one to
use when building the curl executable.

New build options:
- `BUILD_STATIC_LIBS`. Default: `OFF`.
  Enabled automatically if `BUILD_SHARED_LIBS` is `OFF`.
- `BUILD_STATIC_CURL`. Default: `OFF`.
  Requires `BUILD_STATIC_LIBS` enabled.
  Enabled automatically if building static libcurl only.
- `STATIC_LIB_SUFFIX`. Default: empty.
- `IMPORT_LIB_SUFFIX`. Default: `_imp` if implib filename would collide
  with static lib name (typically with MSVC) in Windows builds.
  Otherwise empty.

Also:

- Stop setting the `CURL_STATICLIB` macro via `curl_config.h`, and pass
  it directly to the compiler. This also allows to delete a condition
  from `tests/server/CMakeLists.txt`.

- Complete a TODO by following the logic used in autotools (also for
  `LIBCURL_NO_SHARED`), and set `-DCURL_STATICLIB` in `Cflags:` of
  `libcurl.pc` for _static-only_ curl builds.

- Convert an existing CI test to build both shared and static libcurl.

Closes #11505


vszakats:
I replicated existing behaviour for compatibility where the implib is always named `libcurl_imp.lib` when building for MSVC (or —after this patch— any other compiler prone to static/implib name collision).

Though as discovered earlier, this past choice appears to be arbitrary. Also the `_imp` suffix is seldom used on Windows (AFAIK).

A more widely used convention would be using `libcurl_static.lib` for the static lib. This can be further limited to builds building both shared and static lib at the same time. (I implemented it this way in libssh2, also present in this PR's commit history.)

It's an odd problem with no standard and perfect solution. Let me know if we should do any breaking changes here.

PR #11513, opened 2023-07-25
docs: provide more see also for cipher options

bagder:
More cross references


bagder:
unfortunately this makes these weird troff warnings I don't understand:
~~~
troff:/home/daniel/src/curl/docs/curl.1:4883: warning: character with input code 226 not defined
troff:/home/daniel/src/curl/docs/curl.1:4883: warning: character with input code 226 not defined

PR #11516, opened 2023-07-25
macOS: Really fix target detection

### LINKED ISSUE #11502: dyld: lazy symbol binding failed: Symbol not found: _SCDynamicStoreCopyProxies
### I did this

Install curl 8.2.0 on OS X 10.8 (or on any Mac OS X or OS X version where curl was compiled with the macOS 10.11 SDK or earlier)

### I expected the following

curl compiles but fails to run, for example:

```
$ curl --version
dyld: lazy symbol binding failed: Symbol not found: _SCDynamicStoreCopyProxies
  Referenced from: /opt/local/lib/libcurl.4.dylib
  Expected in: flat namespace

dyld: Symbol not found: _SCDynamicStoreCopyProxies
  Referenced from: /opt/local/lib/libcurl.4.dylib
  Expected in: flat namespace

Trace/BPT trap: 5
```

This was reported to MacPorts here: https://trac.macports.org/ticket/67803

The problem seems to be that curl always wants to use `SCDynamicStoreCopyProxies` on macOS, and this is part of the SystemConfiguration framework, and has been ever since Mac OS X 10.1. However, curl only links with that framework when building against the macOS 10.12 SDK or later.

Normally a missing symbol would be a link-time error but by using the flags `-Wl,-undefined -Wl,dynamic_lookup` you're telling the linker to ignore the problem and that the symbol will exist at runtime. It doesn't, though, because you haven't linked in the framework that has the symbol.

This configure test is faulty:

```
checking whether to link macOS CoreFoundation and SystemConfiguration framework... no
```

It is checking whether `TARGET_OS_OSX` is defined. That define was newly added in the macOS 10.12 SDK.

It looks like an attempt to fix this issue was made in #11417 but it didn't fix it in m4/curl-sysconfig.m4.

`TARGET_OS_OSX` was introduced to help developers distinguish between macOS and other Darwin-derived operating systems (iOS, tvOS, watchOS). Is that distinction necessary here? If not, you can use `TARGET_OS_MAC` which encompasses all of them. Another way to identify any Darwin-based Apple OS is to check if both `__APPLE__` and `__MACH__` are defined.

### curl/libcurl version

8.2.0

### operating system

Darwin redacted 12.6.0 Darwin Kernel Version 12.6.0: Wed Mar 18 16:23:48 PDT 2015; root:xnu-2050.48.19~1/RELEASE_X86_64 x86_64

ryandesign:
`TARGET_OS_OSX` is only defined in the macOS 10.12 SDK and later and is only defined when dynamic targets are enabled. `TARGET_OS_MAC` is always defined but means any Darwin OS, including macOS, iOS, tvOS, watchOS. `TARGET_OS_IPHONE` means any Darwin OS other than macOS.

Follow-up to c73b2f82

Closes #11502

-----

I tested on macOS 12 building for OS X 10.8 and then for iOS 11.

And I tested on OS X 10.8 building for OS X 10.8 and then for iOS 7.

### DIFF
--- a/lib/macos.c
+++ b/lib/macos.c
@@ -24,21 +24,16 @@
 
 #include "curl_setup.h"
 
-#if defined(__APPLE__)
-
-#if !defined(TARGET_OS_OSX) || TARGET_OS_OSX
+#ifdef CURL_MACOS_CALL_COPYPROXIES
 
 #include <curl/curl.h>
 
 #include "macos.h"
 
-#if defined(ENABLE_IPV6) && defined(CURL_OSX_CALL_COPYPROXIES)
 #include <SystemConfiguration/SCDynamicStoreCopySpecific.h>
-#endif
 
 CURLcode Curl_macos_init(void)
 {
-#if defined(ENABLE_IPV6) && defined(CURL_OSX_CALL_COPYPROXIES)
   {
     /*
      * The automagic conversion from IPv4 literals to IPv6 literals only
@@ -46,17 +41,15 @@ CURLcode Curl_macos_init(void)
      * first. As Curl currently doesn't support system-wide HTTP proxies, we
      * therefore don't use any value this function might return.
      *
-     * This function is only available on a macOS and is not needed for
-     * IPv4-only builds, hence the conditions above.
+     * This function is only available on macOS and is not needed for
+     * IPv4-only builds, hence the conditions for defining
+     * CURL_MACOS_CALL_COPYPROXIES in curl_setup.h.
      */
     CFDictionaryRef dict = SCDynamicStoreCopyProxies(NULL);
     if(dict)
       CFRelease(dict);
   }
-#endif
   return CURLE_OK;
 }
 
-#endif /* TARGET_OS_OSX */
-
-#endif /* __APPLE__ */
+#endif
--- a/lib/macos.h
+++ b/lib/macos.h
@@ -23,9 +23,10 @@
  * SPDX-License-Identifier: curl
  *
  ***************************************************************************/
+
 #include "curl_setup.h"
 
-#if defined(__APPLE__) && (!defined(TARGET_OS_OSX) || TARGET_OS_OSX)
+#ifdef CURL_MACOS_CALL_COPYPROXIES
 
 CURLcode Curl_macos_init(void);
 
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -258,8 +258,9 @@
 #if defined(__APPLE__) && !defined(USE_ARES)
 #include <TargetConditionals.h>
 #define USE_RESOLVE_ON_IPS 1
-#  if !defined(TARGET_OS_OSX) || TARGET_OS_OSX
-#    define CURL_OSX_CALL_COPYPROXIES 1
+#  if TARGET_OS_MAC && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE) && \
+     defined(ENABLE_IPV6)
+#    define CURL_MACOS_CALL_COPYPROXIES 1
 #  endif
 #endif
 
--- a/m4/curl-sysconfig.m4
+++ b/m4/curl-sysconfig.m4
@@ -30,10 +30,10 @@ case $host_os in
       AC_LANG_PROGRAM([[
 #include <TargetConditionals.h>
       ]],[[
-#if (TARGET_OS_OSX)
+#if TARGET_OS_MAC && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
       return 0;
 #else
-#error Not a macOS
+#error Not macOS
 #endif
       ]])
     ],[


lib/macos.h
```diff
@@ -25,7 +25,7 @@
  ***************************************************************************/
 #include "curl_setup.h"
 
-#if defined(__APPLE__) && (!defined(TARGET_OS_OSX) || TARGET_OS_OSX)
+#if defined(__APPLE__) && TARGET_OS_MAC && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
```
ryandesign, line 28: Apparently you have line length restrictions:

```
./macos.h:28:92: warning: Longer than 79 columns (LONGLINE)
 #if defined(__APPLE__) && TARGET_OS_MAC && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
```

What alternate formatting does this project prefer to address this?

bagder: There are hundreds of examples present in the lib/ directory alone. Keep lines narrower than 80 columns:
~~~c
#if defined(blabla) && \
  defined(another)

lib/macos.h
```diff
@@ -25,7 +25,8 @@
  ***************************************************************************/
 #include "curl_setup.h"
 
-#if defined(__APPLE__) && (!defined(TARGET_OS_OSX) || TARGET_OS_OSX)
+#if defined(__APPLE__) && TARGET_OS_MAC && \
```
ryandesign, line 28: The c-ares build is failing with:

```
In file included from easy.c:66:
../lib/macos.h:28:27: error: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Werror,-Wundef]
#if defined(__APPLE__) && TARGET_OS_MAC && \
                          ^
```

I was expecting <TargetConditionals.h> to have been included by curl_setup.h but it only does that when not building for c-ares.

Maybe the nicest thing to do is to have curl_setup.h always include it (on Darwin). lib/vtls/sectransp.c also uses `TARGET_OS_` macros without first including <TargetConditionals.h>, probably expecting it to be included by the other Darwin headers it's already including.

Alternately, I can include <TargetConditionals.h> everywhere a `TARGET_OS_` macro is used.

lib/easy.c
```diff
@@ -182,10 +182,12 @@ static CURLcode global_init(long flags, bool memoryfuncs)
   }
 #endif
 
+#ifdef __APPLE__
   if(Curl_macos_init()) {
     DEBUGF(fprintf(stderr, "Error: Curl_macos_init failed\n"));
     goto fail;
   }
+#endif
```
bagder, line 190: The idea is to provide an empty macro for `Curl_macos_init()` when not needed so that this won't need an #ifdef...

lib/macos.h
bagder: > I was expecting <TargetConditionals.h> to have been included by curl_setup.h but it only does that when not building for c-ares.

I suppose because the `SCDynamicStoreCopyProxies` thingie is not needed for builds that use c-ares since then it does not use the regular macOS resolver functions.

lib/macos.c
```diff
@@ -24,9 +24,8 @@
 
 #include "curl_setup.h"
 
-#if defined(__APPLE__)
-
-#if !defined(TARGET_OS_OSX) || TARGET_OS_OSX
+#if defined(__APPLE__) && TARGET_OS_MAC && \
+  !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
```
bagder, line 28: Can't this be just `#ifdef CURL_OSX_CALL_COPYPROXIES` now?

lib/easy.c
ryandesign: I do realize that, but I was looking at the two blocks above this one, where `Curl_win32_init` is only called when `WIN32` is defined and `Curl_amiga_init` is only called when `__AMIGA__` is defined so I was trying to do something similar for `Curl_macos_init`. But I can remove the `#ifdef __APPLE__`…`#endif` again if you prefer.

lib/macos.c
ryandesign: Two things about that.

First, I think the intention was that `Curl_macos_init` would always be a real function on macOS and a macro that does nothing on iOS and every other OS. Currently, that real function calls `SCDynamicStoreCopyProxies` in some configurations and does nothing in others. So yes, I think that for the way this currently works it could be changed so that `Curl_macos_init` would only be a real function when `SCDynamicStoreCopyProxies` needs to be called and a macro that does nothing otherwise. But I think the reason it was set up this way is because at some time in the future you might want to have some other macOS-specific init code and if we change the condition under which `Curl_macos_init` is a real function now, that would have to be reverted at such a future time.

Second, everywhere that `CURL_OSX_CALL_COPYPROXIES` is referenced at present it is done this way:
```c
#if defined(ENABLE_IPV6) && defined(CURL_OSX_CALL_COPYPROXIES)
```
It seems like it would be better to define `CURL_OSX_CALL_COPYPROXIES` only if `ENABLE_IPV6` is also defined. Do you know any reason why that wasn't done?

lib/easy.c
bagder: Yes please, I will get rid of the other ones too: #11525

lib/macos.c
bagder: > I think the intention was that Curl_macos_init would always be a real function on macOS

I think no matter what the intention was, if there is nothing to do in the function then it doesn't need to exist. In fact, having an empty function just causes compiler warnings at times. I think it should only exist when it actually has something to do.

> at some time in the future you might want to...

If that ever happens, we can change and improve the logic *then*.

> It seems like it would be better to define CURL_OSX_CALL_COPYPROXIES only if ENABLE_IPV6 is also defined.

I completely agree!

> Do you know any reason why that wasn't done?

Probably just an oversight.

ryandesign:
I believe I've addressed all the issues. Let me know if anything else needs to be changed.

bagder (approved):

bagder:
Thanks!

Issue #11500, opened 2023-07-22
http2.c: ASSERT: len >= stream->upload_blocked_len


bagder:
### I did this

OSS-Fuzz triggered this assert when fuzzing HTTP. It happens on http.c:2080.

Here's a stack trace
~~~


+----------------------------------------Release Build Stacktrace----------------------------------------+
--
  | Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds-i386_curl_d1a7f12cc2e5055727a9c66d5eca203f3c8f5a6c/revisions/curl_fuzzer_http -rss_limit_mb=2560 -timeout=60 -runs=100 /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-8ae050ae9e02b07a06ae273720218a22cc08117a
  | Time ran: 0.10861730575561523
  |  
  | INFO: Running with entropic power schedule (0xFF, 100).
  | INFO: Seed: 2532200055
  | INFO: Loaded 1 modules   (104786 inline 8-bit counters): 104786 [0x8ea4f00, 0x8ebe852),
  | INFO: Loaded 1 PC tables (104786 PCs): 104786 [0x8ebe854,0x8f8b2e4),
  | /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds-i386_curl_d1a7f12cc2e5055727a9c66d5eca203f3c8f5a6c/revisions/curl_fuzzer_http: Running 1 inputs 100 time(s) each.
  | Running: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-8ae050ae9e02b07a06ae273720218a22cc08117a
  | curl_fuzzer_http: http2.c:2080: ssize_t cf_h2_send(struct Curl_cfilter *, struct Curl_easy *, const void *, size_t, CURLcode *): Assertion `len >= stream->upload_blocked_len' failed.
  | AddressSanitizer:DEADLYSIGNAL
  | =================================================================
  | ==613==ERROR: AddressSanitizer: ABRT on unknown address 0x00000265 (pc 0xf7f05509 bp 0xffd0fa1c sp 0xffd0fa00 T0)
  | SCARINESS: 10 (signal)
  | #0 0xf7f05509 in linux-gate.so.1
  | #1 0xf7be7275 in raise
  | #2 0xf7bcf3f6 in abort
[...]

PR #11519, opened 2023-07-25
Disable asssertion blocking OSSFuzz testing

icing:
- refs #11500
- not clear how this triggers and it blocks OSSFuzz testing other things. Since we handle the case with an error return, disabling the assertion for now seems the best way forward.


bagder:
I'd like to hold off a day or two first to see if if perhaps c76df46a195dff259f2a2716f9f8fb5b7eb89602 changes anything for OSS-Fuzz.

PR #11527, opened 2023-07-27
HTTP/2 proxy fix, CI workflow ngtcp2-linux with caches

icing:
Fix in h2 proxy tunnel: progress in ingress on sending

- depending on what is tunneled, the proxy may never get
    invoked for receiving data explicitly. Not progressing
    ingress may lead to stalls due to missed WINDOW_UPDATEs.

CI:
- add `ngtcp2` workflow, replacing the separate ngtcp2 variations
- use cache for common nghttpx / ngtcp2 / etc build
- use cache for building mod_h2
- uncached time: ~32 min  (longer than before since all tls libs are built)
- cached time: ~14 min


icing:
Ready for merge now.

bagder (approved):

PR #11532, opened 2023-07-28
CI caching for linux.yml workflow

icing:
Introducing caches for the third party modules that are downloaded and built in the different builds. Global version definition in `env` section as:

```
  bearssl-version: 0.6
  libressl-version: v3.6.1
  mbedtls-version: v3.3.0
  mod_h2-version: v2.0.21
  msh3-version: v0.6.0
  openssl3-version: openssl-3.1.1
  quictls-version: OpenSSL_1_1_1t+quic
  rustls-version: v0.10.0
```

OpenSSL was switched from master to 3.1.1, mod_h2 to v2.0.21, other versions are unchanged.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -44,6 +44,7 @@ env:
   bearssl-version: 0.6
   libressl-version: v3.6.1
   mbedtls-version: v3.3.0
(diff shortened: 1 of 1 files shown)


icing:
Ready for merge.

PR #11536, opened 2023-07-28
Revert "schannel: reverse the order of certinfo insertions"

nmoinvaz:
This reverts commit 8986df802db9b5338d9d50a54232ebae4dbcf6dd.

Windows does not guarantee a particular certificate ordering, even though TLS may have its own ordering/relationship guarantees. Recent versions of Windows 11 reversed the ordering of ceritifcates returned by `CertEnumCertificatesInStore`, therefore this commit no longer works as initially intended. libcurl makes no guarantees about certificate ordering if the operating system can't.

See issue #9706 for more details.

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -1665,7 +1665,6 @@ struct Adder_args
   struct Curl_easy *data;
   CURLcode result;
   int idx;
-  int certs_count;
 };
 
 static bool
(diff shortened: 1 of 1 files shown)


bagder (approved):

jay:
@RoguePointer80 do you have anything to add? the consensus is that order is not guaranteed

RoguePointer80:
I'm ok with this revert. I read the linked issue; my understanding is that it is no longer necessary since on newer versions of Windows (22h2) the order is now the same as OpenSSL.
I understand Microsoft's statement that "the order is not guaranteed", but I wouldn't put so much emphasis on it. I believe that statement means they reserve the right to change it, without a contractual binding to backward compatibility. But the order isn't "random", as shown from testing it is consistent.

As for the suggestion of checking Windows version, and choosing to reverse or not the order of certificates: seems too much trouble for something fragile, and since libCurl doesn't guarantee anything that check would amount to over-engineering.

✅ :pr-approved:

MarcelRaad (approved):

jay:
Thanks

Issue #11285, opened 2023-06-09
CMake FetchContent with ZLIB fails (>=8.0.0)


gvne:
### I did this

Include curl with zlib support in my CMake project using FetchContent.
Minimal CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.25)

project(test)

include(FetchContent)
FetchContent_Declare(zlib
  GIT_REPOSITORY https://github.com/madler/zlib.git
  GIT_TAG v1.2.13
)
FetchContent_MakeAvailable(zlib)
set(ZLIB_INCLUDE_DIRS ${zlib_SOURCE_DIR})
set(ZLIB_LIBRARY zlibstatic CACHE STRING "")
set(ZLIB_LIBRARIES ${ZLIB_LIBRARY})
set(ZLIB_FOUND 1)
add_library(ZLIB::ZLIB ALIAS zlibstatic)  # required for CMake > 3.4

FetchContent_Declare(curl
  URL https://github.com/curl/curl/releases/download/curl-8_1_2/curl-8.1.2.tar.xz
)
FetchContent_MakeAvailable(curl)
```

### I expected the following

The CMake configuration to run smoothly

### curl/libcurl version

8.1.2 (Issue started with 8.0.0)

### operating system

Linux pop-os 6.2.6-76060206-generic #202303130630~1683753207~22.04~77c1465 SMP PREEMPT_DYNAMIC Wed M x86_64 x86_64 x86_64 GNU/Linux 

### What I get

When configuring, I get the following error:
```
[cmake] -- Looking for idn2_lookup_ul in idn2;OpenSSL::SSL;OpenSSL::Crypto;ZLIB::ZLIB
[cmake] CMake Error at build/CMakeFiles/CMakeScratch/TryCompile-2M2kVT/CMakeLists.txt:25 (target_link_libraries):
[cmake]   Target "cmTC_70ab4" links to:
[cmake] 
[cmake]     ZLIB::ZLIB
[cmake] 
[cmake]   but the target was not found.  Possible reasons include:
[cmake] 
[cmake]     * There is a typo in the target name.
[cmake]     * A find_package call is missing for an IMPORTED target.
[cmake]     * An ALIAS target is missing.
```

When digging a little, I found something I don't understand in the [CMake macro `check_library_exists_concat`](https://github.com/curl/curl/blob/dacd25888fa7b2d7bb3d16672b986a69088ed292/CMake/Macros.cmake#L32)

[...]

PR #11537, opened 2023-07-28
cmake: don't check CMake target during library detection

nmoinvaz:
Keeps existing behavior which states that some system libraries may depend on others in order to do proper library detection. We remove CMake targets from required libraries test, because system libraries should not depend on CMake targets. Including them causes the library check to fail when using an ALIAS target.

See https://github.com/curl/curl/issues/11285#issuecomment-1629432834 for more details.

### DIFF
--- a/CMake/Macros.cmake
+++ b/CMake/Macros.cmake
@@ -29,8 +29,19 @@
 # order of least-to-most-dependent.  Some libraries depend on others
 # to link correctly.
 macro(check_library_exists_concat LIBRARY SYMBOL VARIABLE)
-  check_library_exists("${LIBRARY};${CURL_LIBS}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
+  # Enumerate CURL_LIBS and add each library to CMAKE_REQUIRED_LIBRARIES if it
+  # is not a CMake target. System libraries should not depend on CMake targets.
+  set(CMAKE_REQUIRED_LIBRARIES)
+  foreach(LIB ${CURL_LIBS})
+    if (NOT TARGET ${LIB})
+      set(CMAKE_REQUIRED_LIBRARIES ${LIB} ${CMAKE_REQUIRED_LIBRARIES})
+    endif()
+  endforeach()
+  # Using CMake targets in the call to check_library_exists() may cause it to fail
+  # unless the target is IMPORTED.
+  check_library_exists("${LIBRARY}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
     ${VARIABLE})
+  set(CMAKE_REQUIRED_LIBRARIES)
   if(${VARIABLE})
     set(CURL_LIBS ${LIBRARY} ${CURL_LIBS})
   endif()


nmoinvaz:
Alternative PR is #11558.

CMake/Macros.cmake
```diff
@@ -29,8 +29,19 @@
...
 macro(check_library_exists_concat LIBRARY SYMBOL VARIABLE)
-  check_library_exists("${LIBRARY};${CURL_LIBS}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
+  # Enumerate CURL_LIBS and add each library to CMAKE_REQUIRED_LIBRARIES if it
+  # is not a CMake target. System libraries should not depend on CMake targets.
+  set(CMAKE_REQUIRED_LIBRARIES)
+  foreach(LIB ${CURL_LIBS})
+    if (NOT TARGET ${LIB})
+      set(CMAKE_REQUIRED_LIBRARIES ${LIB} ${CMAKE_REQUIRED_LIBRARIES})
+    endif()
+  endforeach()
```
bradking, line 39: I think the problem is actually that `CURL_LIBS` is not maintained only by `check_library_exists_concat` anymore.  Other code paths are adding to that list too.  Instead of filtering the list here on every call, I think it will be better to refactor things so a separate list is used for accumulation by `check_library_exists_concat` calls.  One possibility is to keep `CURL_LIBS` here and update all other sites that add to it to use another list or directly call `target_link_libraries` for the dependents.

bradking: Yes, but curl's code is what adds `ZLIB::ZLIB` to `CURL_LIBS`.  It's not the result of a `check_library_exists_concat` call, and so should not be added to the list that `check_library_exists_concat` is accumulating.  It should be linked by other means.

nmoinvaz: Side note: `OpenSSL::SSL OpenSSL::Crypto` aliases are also used by curl.

vszakats: ldap+lber detection may depend on OpenSSL, and at that point we already detected OpenSSL. Thus it seems feasible to use `set(CMAKE_REQUIRED_LIBRARIES "${OPENSSL_LIBRARIES}")` when checking for ldap+lber.

Rest of checks don't need any dependency detected by curl.

nmoinvaz: In that case, if OpenSSL is an CMake `ALIAS`, how has that worked in the past? Seems like it would fail just like in our `ZLIB::ZLIB` case. I imagine nobody has ever run into that scenario because most people have some variety of OpenSSL already installed on their system (so why would they want to compile OpenSSL statically).

vszakats: Yes, I'm guessing in most cases OpenSSL is already built. I never tried building curl from an upper-level CMake (or make it build dependencies in a single run), but in such cases, projects need to be built in reverse topological order of dependency to work correctly.

nmoinvaz: @bradking I have opened a new PR for the changes you suggested. #11648

nmoinvaz:
Closed in favor of #11648.

DashaBochkar:
Good

PR #11546, opened 2023-07-30
cmake: add support for single libcurl compilation pass

vszakats:
Before this patch CMake builds used two separate compilation passes to
build the shared and static libcurl respectively. This patch allows to
reduce that to a single pass if the target platform and build settings
allow it.

This reduces CMake build times when building both static and shared
libcurl at the same time, making these dual builds an almost zero-cost
option.

Enable this feature for Windows builds, where the difference between the
two passes was the use of `__declspec(dllexport)` attribute for exported
API functions for the shared builds. This patch replaces this method
with the use of `libcurl.def` at DLL link time.

Also update `Makefile.mk` to use `libcurl.def` to export libcurl API
symbols on Windows. This simplifies (or fixes) this build method (e.g.
in curl-for-win, which generated a `libcurl.def` from `.h` files using
an elaborate set of transformations).

`libcurl.def` has the maintenance cost of keeping the list of public
libcurl API symbols up-to-date. This list seldom changes, so the cost
is low.

Closes #11546

---

- curl previously had a `libcurl.def` placed in the `lib` directory. I opted for the root directory, because in `lib` it may be overwritten by `.def` files generated by the DLL linker command when using certain build methods/options.

- single-pass libcurl builds can be freely extended to other platforms, if both shared/static libs have position-independent code enable and symbol exports can be fixed up via a linker option.

- it might be an option to make `libcurl.def` an OS-agnostic raw list of public API symbols (= without the top `EXPORTS` line) for re-use for other purposes. CMake could handle that with some extra lines. In `Makefile.mk` less trivially, but also seems doable.


dfandrich:
I thought the whole libcurl.def thing was unneeded because we're using
__declspec(dllexport) on all the relevant exports.

dfandrich:
It was removed the last time in commit 9bd7f004d

vszakats:
We do use `declspec()`, yes, and this is what forces separate compile passes for shared and static libs. The goal of this PR is to avoid that double work. Another positive side effect is that `Makefile.mk` builds can also use this `libcurl.def`, without having to generate it [manually](https://github.com/curl/curl-for-win/blob/1dd7524ef16ca7ea0ee3ca86b26b257e7b461eec/curl-gnumake.sh#L81-L102). `Makefile.mk` always used a single compile pass, so a `.def` file was always necessary for correct results.

The file was deleted because it wasn't used. This time it will be used by two build systems.

PR #11553, opened 2023-07-31
tests, fix h3 server check and parallel instances

icing:
- fix check for availability of nghttpx server
- add `tcp` frontend config for same port as quic, as without this, port 3000 is bound which clashes for parallel testing


bagder (approved):

tests/http3-server.pl
```diff
@@ -108,11 +108,12 @@
...
 my $cmdline="$nghttpx --http2-proxy --backend=$connect ".
+    "--frontend=\"*,$listenport\" ".
     "--frontend=\"*,$listenport;quic\" ".
     "--log-level=INFO ".
     "--pid-file=$pidfile ".
     "--errorlog-file=$logfile ".
     "--conf=$conf ".
     "$keyfile $certfile";
-print "RUN: $cmdline\n" if($verbose);
+print "RUN: $cmdline\n" if($verbose or 1);
```
dfandrich, line 118: "or 1" looks like a debugging leftover

dfandrich (changes requested):
I've never installed nghttpx so I always ignored those error messages—sorry.

tests/http3-server.pl
icing: Thanks for catching this!

bagder (approved):

PR #11555, opened 2023-07-31
cmake: detect `SSL_set0_wbio` in OpenSSL

vszakats:
Present in OpenSSL 1.1.0 and BoringSSL.
Missing from LibreSSL 3.8.0.

Follow-up to f39472ea9f4f4e12cfbc0500c4580a8d52ce4a59

While here, also fix `RAND_egd()` detection which was broken, likely all
along. This feature is probably broken with CMake builds and also
requires a sufficiently obsolete OpenSSL version, so this part of the
update was not tested.

Closes #11555

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@
 # https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package
 #
 # The following variables are available:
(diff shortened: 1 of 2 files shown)


vszakats:
Turns out this is too complicate without duplicating `function(CheckQuicSupportInOpenSSL)` or making it generic. It also turns out that `RAND_EDG` detection was probably always broken.

PR #11556, opened 2023-07-31
egd: delete feature detection and related source code

vszakats:
EGD is Entropy Gathering Daemon, a socket-based entropy source supported
by pre-OpenSSL v1.1 versions and now deprecated. curl also deprecated it
a while ago.

Its detection in CMake was broken all along because OpenSSL libs were
not linked at the point of feature check.

Delete detection from both cmake and autotools, along with the related
source snippet, and the `--with-egd-socket=` `./configure` option.

Closes #11556


bagder (approved):

vszakats:
Managed to fix this as a side-effect of adding `SSL_set0_wbio` detection in #11555, but it's probably best to delete it anyway, since nobody missed this for all these years.

bagder (approved):
Looks good, but I think there are a few more traces of `RAND_EGD` you can wipe out while at it:
~~~
$ git grep -l RAND_EGD  | wc -l
10

vszakats:
autotools detects `RAND_egd()` (correctly, I assumed) in `m4/curl-openssl.m4`, so did not want to touch it, but might make sense.

Do we want to also delete autotools detections bits? and the actual use inside `vtls/openssl.c`, too?

bagder:
> Do we want to also delete autotools detections bits? and the actual use inside `vtls/openssl.c`, too?

Yes let's do that. I think that was only for OpenSSL before 1.1 and already then I believe very few actually used this.

PR #11557, opened 2023-07-31
http2: avoid too early connection re-use/multiplexing

bagder:
HTTP/1 connections that are upgraded to HTTP/2 should not be picked up for reuse and multiplexing by other handles until the 101 switching process is completed.

Lots-of-debgging-by: Stefan Eissing
Reported-by: Richard W.M. Jones
Bug: https://curl.se/mail/lib-2023-07/0045.html


rwmjones:
Thanks for the quick patch.  I tested this and it fixes the problem for me.  Tested up to 1024 threads, where previously it would quickly fail with above ~ 200 threads.

icing (approved):
Tested with my local branch that includes a reproducing test case. Will add that after this PR has been merged.

PR #11558, opened 2023-07-31
cmake: don't add previous libraries to library detection

nmoinvaz:
Don't rely on previously found libraries for system library detection. In some cases the library list may contain a CMake target in which case `check_library_exists` may fail if the target is an ALIAS.

See https://github.com/curl/curl/issues/11285#issuecomment-1629432834 for more details.

### DIFF
--- a/CMake/Macros.cmake
+++ b/CMake/Macros.cmake
@@ -26,10 +26,9 @@
 # This macro checks if the symbol exists in the library and if it
 # does, it prepends library to the list.  It is intended to be called
 # multiple times with a sequence of possibly dependent libraries in
-# order of least-to-most-dependent.  Some libraries depend on others
-# to link correctly.
+# order of least-to-most-dependent.
 macro(check_library_exists_concat LIBRARY SYMBOL VARIABLE)
-  check_library_exists("${LIBRARY};${CURL_LIBS}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
+  check_library_exists("${LIBRARY}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
     ${VARIABLE})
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Alternative to #11537.

CMake/Macros.cmake
```diff
@@ -26,10 +26,9 @@
 # This macro checks if the symbol exists in the library and if it
 # does, it prepends library to the list.  It is intended to be called
 # multiple times with a sequence of possibly dependent libraries in
-# order of least-to-most-dependent.  Some libraries depend on others
-# to link correctly.
+# order of least-to-most-dependent.
 macro(check_library_exists_concat LIBRARY SYMBOL VARIABLE)
-  check_library_exists("${LIBRARY};${CURL_LIBS}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
+  check_library_exists("${LIBRARY}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
```
bradking, line 31: The `;${CURL_LIBS}` part is what distinguishes curl's `check_library_exists_concat` from CMake's `check_library_exists`.  The goal is to handle cases where system libraries depend on each other by checking for them in least-dependent-to-most-dependent order.

bradking (commented):
IMO this PR should be closed in favor of fixing the implementation of the existing semantics, as in #11537.

CMake/Macros.cmake
bradking: Ah, I see that was discussed in https://github.com/curl/curl/issues/11285#issuecomment-1629432834. Indeed the goal of `check_library_exists_concat` is to emulate the behavior in the autotools build system.

PR #11559, opened 2023-07-31
openssl: add CURLSSLOPT_ALLOW_UNSAFE_RENEG

bagder:
Setting this bit in `CURLOPT_SSL_OPTIONS` makes curl built to use OpenSSL allow unsafe renegotiation with OpenSSL 3+. This weakens security and should be a last resort.

Add `--ssl-allow-unsafe-reneg` and `--proxy-ssl-allow-unsafe-reneg`.


bagder:
There is a lack of positive feedback on this PR which makes me leave it for now. We might come back to it at a later point. Thanks everyone.

PR #11562, opened 2023-08-01
CI/spellcheck: build curl.1 to make it spellchecked

bagder:


bagder:
It spells! :green_heart:

PR #11563, opened 2023-08-01
H2 upgrade tests and mitigations

icing:
While issue of premature connection sharing with ongoing HTTP/2 Upgrades (https://curl.se/mail/lib-2023-07/0045.html) was fixed in d4618a3fc4d2753493555814c991dd048264e7e7, this PR adds a test case for the original problem and some additional mitigation.

- added in test_02_25 with special client to reproduce, based on sample code by Richard W.M. Jones
- added check in `http2.c` for being called with an unknown transfer, giving CURLE_HTTP2 instead of a crash
- added DEBUGBUILD library logging with transfer and connection ids for easier analysis


bagder (approved):

icing:
Ok, bearssl and libressl fail the new test with timeouts. Investigating...

icing:
I believe this is ready for merge now.

PR #11570, opened 2023-08-01
tests: ensure `libcurl.def` contains all exports

vszakats:
Add `test1279` to verify that `libcurl.def` lists all exported API
functions found in libcurl headers.

Also:

- extend test suite XML `stdout` tag with the `loadfile` attribute.

- fix `tests/extern-scan.pl` and `test1135` to include websocket API.

- use all headers (sorted) in `test1135` instead of a manual list.

- add options `--sort`, `--heading=` to `tests/extern-scan.pl`.

- add `libcurl.def` to the auto-labeler GHA task.

Follow-up to 2ebc74c36a19a1700af394c16855ce144d9878e3

Closes #11570


vszakats:
I'd preferred using `tests/extern-scan.pl` to extract the `CURL_EXTERN` list, but that script has extra bits in its output. Can those be deleted and the raw symbol names kept? It's also missing `websockets.h`.

bagder:
> Can those be deleted and the raw symbol names kept?

That seems reasonable. `extern-scan.pl` is (only) used by test 1135, so it's just a matter of making sure that is still happy.

vszakats:
Updated the Perl script to use raw symbol names, added websockets functions, then reused it from the GHA task.

bagder (approved):

bagder:
Hm, on second thoughts. Couldn't perhaps test 1135 get extended to do this check as well or something? It would be nice if we could detect this problem locally already when working on a new thing.

vszakats:
Forgot about the local aspect, but indeed, this would be useful. I just have idea how to implement it there. Would need to sort results, insert `EXPORTS` on top and compare with the external file.

vszakats:
@dfandrich Seems doable, it'd need a parser that extracts those extern symbols from `curl.inc.in`. Then sort it, along with `libcurl.def`, then compare them. There still remains a lot of details that can go out of sync with curl headers.

There is also a `gnv_libcurl_symbols.opt` for VMS. Similar to `libcurl.def`, with (much) more boilerplate and a hand-crafted symbol order for compatibility. Also outdated.

vszakats:
Settling on a version that keeps `libcurl.def` sorted and adds a separate test for it.

The next issue is that `tests/extern-scan.pl` needs a manually maintained list of libcurl headers. I might tackle that.

tests/FILEFORMAT.md
```diff
@@ -608,7 +608,7 @@ have a text/binary difference.
 If `nonewline` is set, we will cut off the trailing newline of this given data
 before comparing with the one actually received by the client
 
-### `<stdout [mode="text"] [nonewline="yes"] [crlf="yes"]>`
+### `<stdout [mode="text"] [nonewline="yes"] [crlf="yes"] [loadfile="filename"]>`
```
bagder, line 611: Wow, this was a cool idea. Me like!

tests/extern-scan.pl
```diff
@@ -28,18 +28,61 @@
...
+        "$root/include/curl/curl.h",
+        "$root/include/curl/easy.h",
+        "$root/include/curl/mprintf.h",
+        "$root/include/curl/multi.h",
+        "$root/include/curl/urlapi.h",
+        "$root/include/curl/options.h",
+        "$root/include/curl/header.h",
+        "$root/include/curl/websockets.h",
+        );
+}
```
bagder, line 85: Is there actually a use for the not-allheaders case?

vszakats: I didn't fully grasp the original intent, so I left it there, assuming its purpose is to somehow enforce an "append-only" growth of the list for VMS and OS/400 (based on the comment in `test1135`). If I misunderstood or that's not important, I agree we are better off deleting this case.

vszakats: I'm about deleting it, and that will change the order of expected results (because the manual header list was not alpha-sorted). If that's okay, it's ready for commit.

bagder: > assuming its purpose is to somehow enforce an "append-only" growth

Oh right, I forgot about that and I believe I had that in mind. I believe that is unnecessary because as you say, this list is just reading the headers in some specific order, it does not actually define in which order those builds use these symbols so any such append-only order would have to be defined elsewhere.

PR #11580, opened 2023-08-02
getinfo: remove header_size from info struct

bagder:
It is a duplicate of the size already kept in `req.headerbytecount`


bagder:
It's not a complete duplicate as `header_size` includes the size of the CONNECT response... I'll drop this attempt.

PR #11582, opened 2023-08-02
return error when receiving too large headers

bagder:
To avoid abuse. The limit is set to 257 KB. Incomplete research seems to suggest that some browsers use a 256 KB cap.

Test 497 verifies.


bagder:
I did some live tests with my Firefox and Chrome on Linux. Sending an endless stream of headers, when do they give up?

- Firefox: around 1MB
- Chrome: between 256KB and 300KB

Neither of them seem to have a fixed limit as repeated runs show different results.

Based on this, I think I should up the curl limit to 300KB.

bagder:
hyper has (rather annoyingly) its own limit which also returns a generic error which makes it very hard for us to catch specifically making us forced to return a rather bad error code there.

Also, I don't know what the hyper limit is but I have asked: https://github.com/hyperium/hyper/issues/3283

PR #11585, opened 2023-08-03
Use openssl 3.0.10+quic, nghttp3 0.14.0, ngtcp2 0.18.0

GrahamCampbell:
I'm not aware of any changes in these releases that would affect us. The CI in master is already failing, which makes things quite hard to check.


bagder:
> The CI in master is already failing

Yes, sorry about that. If you rebase and force-push now I think the state should be a little better.

bagder:
Thanks!

PR #11587, opened 2023-08-04
http: use %u for printfing int

bagder:
Follow-up to 3ee79c1674fd6f99e8efca5

I don't know how I missed this... :confused:


icing:
Getting now format checks on `infof()` and `failf()`?😌

bagder:
> Getting now format checks on `infof()` and `failf()`?

Yes we have that - when not checking for C90 compatibility. I must have missed that warning in my PR...

PR #11589, opened 2023-08-04
include/curl/mprintf.h: add __attribute__ for the prototypes

bagder:
- if gcc or clang is used
- if `__STDC_VERSION__ >= 199901L`, which means greater than C90
- if not using mingw
- only if `CURL_NO_FMT_CHECKS` is not defined

### DIFF
--- a/include/curl/mprintf.h
+++ b/include/curl/mprintf.h
@@ -32,18 +32,36 @@
 extern "C" {
 #endif
 
-CURL_EXTERN int curl_mprintf(const char *format, ...);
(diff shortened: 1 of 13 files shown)


bagder:
1. It might be a little "dangerous" to add this in the public header and thus impose this on "everyone"
2. the infof() and failf() format checks we have are disabled on mingw32. Do we need that here as well?

bagder:
2: yes, mingw will warn on use of the 'z' flag and we can't have that

bagder:
I'll take my chances and merge this. If it backfires, we can maybe make the mprintf.h defines only apply when building libcurl.

PR #11598, opened 2023-08-04
[GIT]  update .gitignore, ignore compdb

Simonhancrew:
update .gitignore, ignore compdb cache + compile_commands.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ libstandaloneengine.a
 tests/string
 tests/config
 
+# compdb
+.cache/
+compile_commands.json


.gitignore
```diff
@@ -66,3 +66,8 @@ libstandaloneengine.a
 tests/string
 tests/config
 
+# compdb
+.cache/
+compile_commands.json
+
+
```
vszakats, line 73:
```suggestion
```

vszakats, line 73: I suggest deleting the extra empty lines at the end of the file. (But keep the LF at the end of the last line)

Simonhancrew: Its done, thanks for your advice

bagder:
Why should curl care about "combdb" in its ignore file?

Simonhancrew:
I believe that many people use clangd as a code jumping and code hinting tool, which is why I wanted to add these. 

I think something that repo does not focus on should be divided into two parts, one is the compilation products, and the other is the dependency product of the code reading toolsets

Simonhancrew:
anyway, its only my opinion, hh

vszakats:
For non-curl-specific files, it seems better to ignore them in the global `.gitignore` that applies to all projects: `$XDG_CONFIG_HOME/git/ignore`.

bagder:
> it seems better to ignore them via the global ignore list

Right. Files created by curl files are fine to ignore by curl ignore files, but files and leftovers generated by random other helper tools are better ignored using the generic global ignore. Otherwise there will be no end to what we need to ignore.

Simonhancrew:
1, i got that,let me close this pr

PR #11602, opened 2023-08-05
Take again for main branch with default name on Curl package

lucianoCircelli:
I thought giving a different name in main branch to customize the package 
Your members will have to manually update their local environments. We'll let them know when they visit the repository, or you can share the following commands.


dfandrich:
You are posting these PRs on the curl repository, which clearly they are not intended to be!

lucianoCircelli:
Because Visual Studio would a pull request and it made it to Curl
packages and not in my forked and cloning repository of Curl

PR #11609, opened 2023-08-07
HTTP/3, ngtcp2, shorten handshake, trace cleanup

icing:
- shorten handshake timing by delayed x509 store load (OpenSSL) as we do for HTTP/2
- cleanup of trace output, align with HTTP/2 output

Sample handshake times on my machine

```
> python3 tests/http/scorecard.py -H h3

BEFORE
Handshakes               ipv4                      ipv6
  Host                   Connect    Handshake      Connect    Handshake
  curl.se                   0 ms        53 ms         0 ms       219 ms

AFTER
Handshakes               ipv4                      ipv6
  Host                   Connect    Handshake      Connect    Handshake
  curl.se                   0 ms        29 ms         0 ms       196 ms
```

**update**: my ipv6 unborked itself, now seeing 22ms/22ms on both ip version...*shrugs*

### DIFF
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -162,6 +162,9 @@ struct cf_ngtcp2_ctx {
   size_t max_stream_window;          /* max flow window for one stream */
   int qlogfd;
   BIT(got_first_byte);               /* if first byte was received */
+#ifdef USE_OPENSSL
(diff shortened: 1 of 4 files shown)


bagder:
From my dev machine to **curl.se**, this PR reduces my handshake times from 18ms/18ms to 5ms/5ms...

PR #11611, opened 2023-08-07
openssl: switch to modern init for LibreSSL 2.7.0+

vszakats:
LibreSSL 2.7.0 (2018-03-21) introduced automatic initialization, `OPENSSL_init_ssl()` function and deprecated the old, manual init method, as seen in OpenSSL 1.1.0. Switch to the modern method when available.

Ref: https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.7.0-relnotes.txt

Closes #11611

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -1691,7 +1691,7 @@ static int x509_name_oneline(X509_NAME *a, char *buf, size_t size)
 static int ossl_init(void)
 {
 #if (OPENSSL_VERSION_NUMBER >= 0x10100000L) &&  \
-  !defined(LIBRESSL_VERSION_NUMBER)
+  (!defined(LIBRESSL_VERSION_NUMBER) || LIBRESSL_VERSION_NUMBER >= 0x2070000fL)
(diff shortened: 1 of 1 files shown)


vszakats:
Looking to do the same in `ossl_cleanup()`, where the situation is murkier:

https://man.openbsd.org/OpenBSD-6.3/EVP_cleanup.3 - deprecated
https://man.openbsd.org/OpenBSD-6.3/ENGINE_cleanup.3 - not loading engines, so seems unnecessary
https://man.openbsd.org/OpenBSD-6.3/ERR_free_strings.3 - available, not a no-op
https://man.openbsd.org/OpenBSD-6.3/ERR_remove_thread_state.3 - available, not a no-op
https://man.openbsd.org/OpenBSD-6.3/ERR_remove_state.3 - deprecated
https://man.openbsd.org/OpenBSD-6.3/CONF_modules_free.3 - not loading CONF modules, so seems unnecessary
SSL_COMP_free_compression_methods - doesn't exist in LibreSSL

PR #10733, opened 2023-03-10
crypto: ensure crypto initialization works

danielgustafsson:
Make sure that context initialization during hash setup works to avoid going forward with the risk of a null pointer dereference.

Closes: xxxx
Reported-by: Philippe Antoine on HackerOne

PR #11614, opened 2023-08-07
crypto: ensure crypto initialization works

bagder:
Make sure that context initialization during hash setup works to avoid going forward with the risk of a null pointer dereference.

Reported-by: Philippe Antoine on HackerOne

Replaces #10733

### DIFF
--- a/lib/md4.c
+++ b/lib/md4.c
@@ -522,7 +522,6 @@ CURLcode Curl_md4it(unsigned char *output, const unsigned char *input,
   MD4_Update(&ctx, input, curlx_uztoui(len));
   MD4_Final(output, &ctx);
   return CURLE_OK;
-#endif
 }
 
 #endif /* USE_CURL_NTLM_CORE */
--- a/lib/md4.c
+++ b/lib/md4.c
@@ -92,9 +92,10 @@
 
 typedef struct md4_ctx MD4_CTX;
 
-static void MD4_Init(MD4_CTX *ctx)
+static CURLcode MD4_Init(MD4_CTX *ctx)
 {
   md4_init(ctx);
+  return CURLE_OK;
 }
 
 static void MD4_Update(MD4_CTX *ctx, const void *data, unsigned long size)
@@ -114,9 +115,10 @@ static void MD4_Final(unsigned char *result, MD4_CTX *ctx)
 #elif defined(AN_APPLE_OS)
 typedef CC_MD4_CTX MD4_CTX;
 
-static void MD4_Init(MD4_CTX *ctx)
+static CURLcode MD4_Init(MD4_CTX *ctx)
 {
   (void)CC_MD4_Init(ctx);
+  return CURLE_OK;
 }
 
 static void MD4_Update(MD4_CTX *ctx, const void *data, unsigned long size)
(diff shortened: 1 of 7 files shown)


lib/sha256.c
```diff
@@ -218,9 +221,13 @@ typedef struct sha256_ctx my_sha256_ctx;
...
 {
-  if(CryptAcquireContext(&ctx->hCryptProv, NULL, NULL, PROV_RSA_AES,
-                         CRYPT_VERIFYCONTEXT | CRYPT_SILENT)) {
-    CryptCreateHash(ctx->hCryptProv, CALG_SHA_256, 0, 0, &ctx->hHash);
+  if(!CryptAcquireContext(&ctx->hCryptProv, NULL, NULL, PROV_RSA_AES,
+                         CRYPT_VERIFYCONTEXT | CRYPT_SILENT))
+    return CURLE_OUT_OF_MEMORY;
+
+  if(!CryptCreateHash(ctx->hCryptProv, CALG_SHA_256, 0, 0, &ctx->hHash)) {
+    CryptReleaseContext(ctx->hCryptProv, 0);
```
jay, line 229: the insert code suggestion button is missing from this pr for some reason

lib/md4.c
```diff
@@ -93,10 +93,10 @@
 
 typedef struct md4_ctx MD4_CTX;
 
-static CURLcode MD4_Init(MD4_CTX *ctx)
+static int MD4_Init(MD4_CTX *ctx)
 {
   md4_init(ctx);
-  return CURLE_OK;
+  return 1;
```
jay, line 99: @danielgustafsson I've changed MD4_Init prototype because it has to match backends that provide that function such as OpenSSL. [See here](https://www.openssl.org/docs/manmaster/man3/MD4_Init.html)

Another way to do it would be the way we do for md5 and sha256 where for every backend we define a CURLcode my_md4_init that either wraps MD4_Init or does whatever the alternative MD4 init procedure.

Also, I am not sure about this one above, I cannot find anything about GnuTLS md4_init in documentation. If it returns a code then this will need to be changed.

lib/sha256.c
bagder: Possibly because I made the PR but @danielgustafsson did the commits?

lib/md4.c
bagder: GnuTLS uses nettle for crypto. nettle's md4_init function does not return anything:

https://git.lysator.liu.se/nettle/nettle/-/blob/master/md4.c#L58-74

bagder:
Thanks for the awesome assistance here @jay!

lib/sha256.c
jay: > Possibly because I made the PR but @danielgustafsson did the commits?

I see it now. I think what may have happened is I clicked on an individual commit instead of 'Files Changed', because when I did that just now I notice there's no suggestion button.

catenacyber:
Thanks Daniel, this was merged, right ?

bagder:
Correct, "closed this in https://github.com/curl/curl/commit/22eb9893bc8fa8dea874b0cad64fdca835e19a16" shows the commit which closed this PR when it was merged.

PR #11623, opened 2023-08-08
Debug ngtcp2-wolfssl ci fails

icing:
Increase trace info for test_02_07b where wolfssl sporadically returns -425 on connect


bagder:
Looks better, right? At least now, could also be just fluke in this run.

icing:
> Looks better, right? At least now, could also be just fluke in this run.

Yes, I'll push a small change, see if it repeats.

PR #11628, opened 2023-08-08
Update list-only.d adding SFTP as supported protocol

eppesuig:
This is a very small patch against the manual page.
The --list-only option works with SFTP protocol, as well.

### DIFF
--- a/docs/cmdline-opts/list-only.d
+++ b/docs/cmdline-opts/list-only.d
@@ -10,6 +10,7 @@ Example: --list-only ftp://example.com/dir/
 See-also: quote request
 Multi: boolean
 ---
+(FTP)
 When listing an FTP directory, this switch forces a name-only view. This is
 especially useful if the user wants to machine-parse the contents of an FTP
 directory since the normal directory view does not use a standard look or
@@ -18,6 +19,12 @@ the server instead of LIST.
 
 Note: Some FTP servers list only files in their response to NLST; they do not
 include sub-directories and symbolic links.
+
+(SFTP)
+When listing an SFTP directory, this switch forces a name-only view, one per line.
+This is especially useful if the user wants to machine-parse the contents of an
+SFTP directory since the normal directory view provides more information than just
+file names.
 
 (POP3)
 When retrieving a specific email from POP3, this switch forces a LIST command


docs/cmdline-opts/list-only.d
```diff
@@ -2,15 +2,15 @@ c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
...
 Help: List only mode
 Added: 4.0
-Category: ftp pop3
+Category: ftp pop3 sftp
 Example: --list-only ftp://example.com/dir/
 See-also: quote request
 Multi: boolean
 ---
-(FTP)
+(FTP SFTP)
```
bagder, line 13: You can remove this line, it's superfluous and just adds duplicated info.

bagder: I mean remove that line completely. It was wrong already before your change so you can fix that while there.

bagder (approved):

docs/cmdline-opts/list-only.d
eppesuig: OK, I hope I did it correctly.
I wonder if I should add a note about SFTP server listing . and .. as well. It this the same behavior as the FTP servers?

bagder: What behavior is that you mean? FTP directory listings are not very standardized so they can vary a little between server implementations.

docs/cmdline-opts/list-only.d
```diff
@@ -2,15 +2,14 @@ c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
...
+Protocols: FTP POP3 SFTP
 Help: List only mode
 Added: 4.0
-Category: ftp pop3
+Category: ftp pop3 sftp
 Example: --list-only ftp://example.com/dir/
 See-also: quote request
 Multi: boolean
 ---
-(FTP)
```
bagder, line 13: Actually, I was too early suggesting you remove this line. But just adding SFTP to it is also not enough. The paragraphs below this header explains FTP details that only partly overlaps with how it works for SFTP...

eppesuig: I changed the patch splitting the definition between FTP and SFTP.

bagder: :+1:

bagder:
Thanks!

PR #11629, opened 2023-08-08
cmake: fix to use variable for the curl namespace

vszakats:
Replace (wrong) literal with a variable to specify the curl
namespace.

Follow-up to 1199308dbc902c52be67fc805c72dd2582520d30 #11505

Reported-by: balikalina on Github
Fixes https://github.com/curl/curl/commit/1199308dbc902c52be67fc805c72dd2582520d30#r123923098
Closes #11629

### DIFF
--- a/CMake/curl-config.cmake.in
+++ b/CMake/curl-config.cmake.in
@@ -35,4 +35,4 @@ include("${CMAKE_CURRENT_LIST_DIR}/@TARGETS_EXPORT_NAME@.cmake")
 check_required_components("@PROJECT_NAME@")
 
 # Alias for either shared or static library
-add_library(curl::libcurl ALIAS curl::@LIB_SELECTED@)
+add_library(@PROJECT_NAME@::libcurl ALIAS @PROJECT_NAME@::@LIB_SELECTED@)


balikalina:
Some projects need to export alias
add_library(ALIAS) in config file in some cases cause an error
```
CMake Error at ***/build/parts/curl/install/lib64/cmake/CURL/CURLConfig.cmake:62 (add_library):
[25/26]   add_library cannot create ALIAS target "CURL::libcurl" because target
[25/26]   "CURL::libcurl_static" is imported but not globally visible.
[25/26] Call Stack (most recent call first):
[25/26]   /usr/share/cmake3/Modules/FindCURL.cmake:58 (find_package)
[25/26]   pull/tests/integration/CMakeLists.txt:29 (find_package)
[25/26] -- Found CURL: ***/build/parts/curl/install/lib64/cmake/CURL/CURLConfig.cmake (found version "8.3.0-DEV")  
[25/26] CMake Error at ***/build/parts/curl/install/lib64/cmake/CURL/CURLConfig.cmake:62 (add_library):
[25/26]   add_library cannot create ALIAS target "CURL::libcurl" because target
[25/26]   "CURL::libcurl_static" is imported but not globally visible.
[25/26] Call Stack (most recent call first):
[25/26]   /usr/share/cmake3/Modules/FindCURL.cmake:58 (find_package)
[25/26]   push/CMakeLists.txt:2 (find_package)
```

there is an issue about exporitng tha alias
https://gitlab.kitware.com/cmake/cmake/-/issues/18996

so could you consider please set property EXPORT_NAME instead of add_library(ALIAS)?


line 222 lib/CMakeLists.txt

```
add_library(${LIB_NAME} ALIAS ${LIB_SELECTED})
add_library(${PROJECT_NAME}::${LIB_NAME} ALIAS ${LIB_SELECTED})

#export libcurl alias as target name
set_property(TARGET ${LIB_SELECTED}
  PROPERTY
  EXPORT_NAME ${LIB_NAME})

if(CURL_ENABLE_EXPORT_TARGET)
```

if added, must remove the 
```
add_library(@PROJECT_NAME@::libcurl ALIAS @PROJECT_NAME@::@LIB_SELECTED@)
```
from CMake/curl-config.cmake.in

vszakats:
Thank you, is this what had in mind?: #11646

balikalina:
> Thank you, is this what you had in mind?: #11646

yes, exactly
if it do not crash any other builds for other build systems/os
may be should take some pause while waiting for other opinions

vszakats:
There was an issue reported with this solution over at libssh2: https://github.com/libssh2/libssh2/pull/1154#issuecomment-1673061358

With a difference suggestion for a fix.

balikalina:
> There was an issue reported with this solution over at libssh2: [libssh2/libssh2#1154 (comment)](https://github.com/libssh2/libssh2/pull/1154#issuecomment-1673061358)
> 
> With a difference suggestion for a fix.

if I move setting property EXPORT_NAME from lib/CMakeLists.txt to CMake/curl-config.cmake.in in a such way
```
include("${CMAKE_CURRENT_LIST_DIR}/@TARGETS_EXPORT_NAME@.cmake")
check_required_components("@PROJECT_NAME@")

# Alias for either shared or static library
#add_library(@PROJECT_NAME@::libcurl ALIAS @PROJECT_NAME@::@LIB_SELECTED@)
set_property(TARGET @PROJECT_NAME@::@LIB_SELECTED@ PROPERTY EXPORT_NAME @PROJECT_NAME@::libcurl)
```
receive error:

```
[25/26] CMake Error at ***/build/parts/curl/install/lib64/cmake/CURL/CURLConfig.cmake:63 (set_property):
[25/26]   EXPORT_NAME property can't be set on imported targets
[25/26]   ("CURL::libcurl_static")
```

balikalina:
both CMakefiles.txt add target library (example for static)
```
add_library(${LIB_STATIC} STATIC ${SOURCES})
```

then libcurl declare aliases
```
add_library(${LIB_NAME} ALIAS ${LIB_SELECTED})
add_library(${PROJECT_NAME}::${LIB_NAME} ALIAS ${LIB_SELECTED})

```
when ssh2 doesn't do this

and the last they both install target

so the only difference is the absence or ALIAS add_library() call in CMakeLists.txt

vszakats:
Thanks for investigating @balikalina and pointing to these bits.

I read docs and came up with a minimal test, but the bad news is that I couldn't reproduce, and all patch suggestions break the test, while the `master` build works OK. I tested this with `libssh2` only, because based on the CMake error message, this looks like the same issue in both libcurl and libssh2 (despite minor differences in their `CMakeLists.txt`s).

I closed the related PR, but feel free to discuss / investigate further in #11646.

PR #11632, opened 2023-08-08
schannel: use CERT_CONTEXT's pbCertEncoded to determine chain order

nmoinvaz:
`CERT_CONTEXT` from `SECPKG_ATTR_REMOTE_CERT_CONTEXT `contains end-entity/server certificate in `pbCertEncoded`. We can use this pointer to determine the order of certificates when enumerating `hCertStore` using `CertEnumCertificatesInStore`.

I tested this method of determining enumeration order on Windows 10 and Windows 11.

For background info see PR #9706 and this [comment](https://github.com/curl/curl/pull/10056#issuecomment-1669872136).

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -1687,6 +1687,7 @@ struct Adder_args
   struct Curl_easy *data;
   CURLcode result;
   int idx;
+  int certs_count;
 };
 
 static bool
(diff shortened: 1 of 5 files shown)


nmoinvaz:
Thanks, I think I got it now:
image

jay:
It looks like the test is not running in Windows CI jobs but it's running in Linux jobs. I don't see anything obvious that could be causing that.

nmoinvaz:
Which Windows CI jobs are you looking at? The ones I am seeing from AppVeyor only have a short build log. The only thing I can think of is maybe it doesn't like this:
```xml
# SSL with libraries supporting CURLOPT_CERTINFO
<features>
SSL
!bearssl
!mbedtls
!rustls
!wolfssl
</features>
```
Perhaps I need to specifically include openssl and schannel in that list..

jay:
> It looks like the test is not running in Windows CI jobs but it's running in Linux jobs.

It looks as though tests that use SSL are not running in Windows CI jobs because of an stunnel bug. mback2k/curl-docker-winbuildenv#2. I still do not know why some of the test results do not show a skipped test message for this test, which is what should happen if the test is skipped. It may be because of the way stunnel fails. The installation is valid and the file is executable but its version information can't be parsed. There's logic that lists SSL as a supported test type but then none of those tests actually run because stunnel technically isn't considered valid if the version info parsing fails.

nmoinvaz:
I reviewed those two PRs, and I can rebase after they are merged in.

nmoinvaz:
Rebased

nmoinvaz:
Rebased again.

jay:
Thanks

PR #11636, opened 2023-08-09
bug_report: require reporters to specify curl and os versions

jay:
- Change curl version and os sections from single-line input to multi-line textarea.

- Require curl version and os sections to be filled out before report can be submitted.

Closes #xxxx

---

Some reporters don't fill out the curl and os version fields, make it required so we don't have to follow up asking them. Also make the fields multi-line. Right now if you hit enter in the single-line fields it submits the issue. This just happened to me when what I was trying to do was to start a new line.


MarcelRaad (approved):

bagder (approved):
do it!

PR #11647, opened 2023-08-10
rustls: implement connect_blocking

jsha:
Previously I hadn't implemented connect_blocking because it was only used in FTP. But several of the tests that fail under rustls are FTP tests, so this is a necessary first step to make them pass. However, even with this change they don't pass yet so some additional changes may be necessary.

### DIFF
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -39,6 +39,7 @@
 #include "select.h"
 #include "strerror.h"
 #include "multiif.h"
+#include "connect.h" /* for the connect timeout */
 
 struct rustls_ssl_backend_data
 {
@@ -73,14 +74,6 @@ cr_data_pending(struct Curl_cfilter *cf, const struct Curl_easy *data)
   DEBUGASSERT(ctx && ctx->backend);
   backend = (struct rustls_ssl_backend_data *)ctx->backend;
   return backend->data_pending;
-}
-
-static CURLcode
-cr_connect(struct Curl_cfilter *cf UNUSED_PARAM,
-           struct Curl_easy *data UNUSED_PARAM)
-{
-  infof(data, "rustls_connect: unimplemented");
-  return CURLE_SSL_CONNECT_ERROR;
 }
 
 struct io_ctx {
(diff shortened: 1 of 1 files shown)


lib/vtls/rustls.c
```diff
@@ -589,6 +602,21 @@ cr_connect_nonblocking(struct Curl_cfilter *cf,
...
+                       struct Curl_easy *data, bool *done)
+{
+  cr_connect_common(cf, data, done, false);
+}
+
+static CURLcode
+cr_connect_blocking(struct Curl_cfilter *cf UNUSED_PARAM,
+           struct Curl_easy *data UNUSED_PARAM)
+{
+  bool done; // ignored
```
bagder, line 616:
~~~
vtls/rustls.c:616:14: error: C++ style comments are not allowed in ISO C90
  616 |   bool done; // ignored
      |              ^
vtls/rustls.c:616:14: note: (this will be reported only once per input file)

bagder: (it would be cool if checksrc could do this kind of comment itself...)

bagder:
any updates on this? it failed a lot of tests and now it has a merge-conflict as well...

jsha:
Ah, looks like I have some lint issues. Will fix soon. Thanks for the ping!

```
./vtls/rustls.c:517:88: warning: Longer than 79 columns (LONGLINE)
     infof(data, "loop - cr_connect_common handle(%d) blocking:%d", data->id, blocking);
./vtls/rustls.c:550:86: warning: Longer than 79 columns (LONGLINE)
     what = Curl_socket_check(readfd, CURL_SOCKET_BAD, writefd, socket_check_timeout);
./vtls/rustls.c:557:82: warning: Longer than 79 columns (LONGLINE)
       failf(data, "rustls connection timeout after %d ms", socket_check_timeout);
./vtls/rustls.c:615:14: warning: // comment (CPPCOMMENTS)
   bool done; // ignored
              ^
```

jsha:
I think I've fixed all the issues. The CI is green (except for AppVeyor, which I understand to be flaky). However, there are many fewer items in the CI than usual, which is confusing.

dfandrich:
It looks like the other tasks ran, but GitHub isn't showing them for some reason. There are some legitimate errors in the rustls build:
```
vtls/rustls.c: In function ‘cr_connect_common’:
vtls/rustls.c:546:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
mv -f vtls/.deps/libcurlu_la-openssl.Tpo vtls/.deps/libcurlu_la-openssl.Plo
  546 |     const timediff_t timeout_ms = Curl_timeleft(data, NULL, TRUE);
      |     ^~~~~
vtls/rustls.c:554:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
  554 |     const timediff_t socket_check_timeout = blocking?timeout_ms:0;
      |     ^~~~~
vtls/rustls.c: In function ‘cr_connect_nonblocking’:
vtls/rustls.c:617:1: error: no return statement in function returning non-void [-Werror=return-type]
  617 | }
      | ^
```

jsha:
Thanks @dfandrich! The errors you quote on lines 456 and 617 were fixed in my second-to-most-recent force push, but the line 554 error I only fixed in my most recent force push. I realized why I hadn't been able to reproduce these errors locally: I didn't have `--enable-warnings --enable-werror` in my local `./configure` invocation.

jsha:
Okay, now everything but FreeBSD is passing, and FreeBSD appears to be timing out.

bagder:
Thanks!

PR #11649, opened 2023-08-10
CI, ngtcp2-wolfssl, downgrade to wolfSSL v5.6.0

icing:
- due to failures in test_02_07b with v5.6.3, tests with master seem to indicate that the issue with session reuse is fixed there


icing:
All versions checked fail as well, closing.

PR #11650, opened 2023-08-10
tests: add a test to check translatable string options in OS400 wrapper

monnerat:
This test runs a perl script that checks all string options are properly translated by the OS400 character code conversion wrapper. It also verifies these options are listed in alphanumeric order in the wrapper switch statement.


dfandrich:
Otherwise looks fine to me.

monnerat:
> Otherwise looks fine to me.
Thanks for your review. All your remarks are now taken into account.

I've added a commit that removes the now redundant build time translatable string options check on OS/400: if kept, it would have only signaled false positives!

bagder (approved):

monnerat:
Thanks for merge :-)

PR #11654, opened 2023-08-11
quiche: fix quiche being unable to handle timeout events properly

trrui-huawei:
According to [quiche's document](https://github.com/cloudflare/quiche/blob/master/README.md#generating-outgoing-packets), it is essential to invoke `quiche_conn_on_timeout` when a timer expires. Failing to call this function can result in quiche being unable to manage various timeout events, including scenarios like retransmitting lost packets and handling idle timeout events.

> When packets are sent, the application is responsible for maintaining a timer to react to time-based connection events. The timer expiration can be obtained using the connection's [timeout()](https://docs.quic.tech/quiche/struct.Connection.html#method.timeout) method.

> The application is responsible for providing a timer implementation, which can be specific to the operating system or networking framework used. When a timer expires, the connection's [on_timeout()](https://docs.quic.tech/quiche/struct.Connection.html#method.on_timeout) method should be called, after which additional packets might need to be sent on the network:

### DIFF
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -56,7 +56,7 @@
 /* #define DEBUG_QUICHE */
 
 #define QUIC_MAX_STREAMS              (100)
-#define QUIC_IDLE_TIMEOUT        (5 * 1000) /* milliseconds */
+#define QUIC_IDLE_TIMEOUT        (60 * 1000) /* milliseconds */
 
 #define H3_STREAM_WINDOW_SIZE  (128 * 1024)
 #define H3_STREAM_CHUNK_SIZE    (16 * 1024)
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -751,6 +751,12 @@ static CURLcode cf_flush_egress(struct Curl_cfilter *cf,
   struct read_ctx readx;
   size_t pkt_count, gsolen;
 
+  quiche_conn_on_timeout(ctx->qconn);
+  if(quiche_conn_is_closed(ctx->qconn)) {
+    failf(data, "quiche_conn_on_timeout closed the connection");
+    return CURLE_SEND_ERROR;
+  }
+
   result = vquic_flush(cf, data, &ctx->q);
   if(result) {
     if(result == CURLE_AGAIN) {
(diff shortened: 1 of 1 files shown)


lib/vquic/curl_quiche.c
```diff
@@ -751,6 +751,12 @@ static CURLcode cf_flush_egress(struct Curl_cfilter *cf,
   struct read_ctx readx;
   size_t pkt_count, gsolen;
 
+  quiche_conn_on_timeout(ctx->qconn);
```
icing, line 754: This calls the function unconditionally on every egress handling. Which is not what the quiche example seem to mandate (guessing through their lack of documentation).

I think we need to handle timeouts more elaborate here. But we can do this based on this PR.

icing (approved):
This PR goes in the right direction. But needs to be built on to invoke `quiche_conn_on_timeout()` only in case the timeout really did happen.

lib/vquic/curl_quiche.c
trrui-huawei: Thanks! I have made additional modifications to the code. Please check it. :)

bagder: @LPardue @ghedo you think this makes sense?

LPardue: Few things:

* Calling quiche_conn_on_timeout() before any timeout elapsed is fine, there is no harm other than wasted cycles.
* It would be better to grab the timeout value once and only fire when expires, that query it every time and check for 0 (optimization)
* its not clear to me how this relates to the existing `timeout_ns` that also reads `quiche_conn_timeout()` for setting a Curl_expire

trrui-huawei: + That's right, using a variable to record the value of timeout can improve performance. However, in order to remain consistent with `curl_ngtcp2.c` and minimize code modifications, I chose not to do this. Maybe we can consider optimizing the implementations of `curl_ngtcp2` and `curl_quiche` together in the future.

+ Reading `quiche_conn_timeout` again is to promptly update the timing for the next timeout. After processing with `quiche_conn_on_timeout`, the timeout duration might change. Perhaps I can make a small optimization by placing the update of timeout after calling `quiche_conn_on_timeout`, which would require querying `quiche_conn_timeout` only when necessary.

bagder:
@icing you think we should merge this as-is and consider further improvements later? It seems like a step in the right direction at least to me.

bagder:
Thanks!

PR #11655, opened 2023-08-11
urlapi: add ability to convert back from punycode to IDN name

bagder:


bagder:
I wrote the windows code "blind" based on [this documentation](https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-idntounicode).

bagder:
So **4** Windows builds fail, the others work!? The test 1560 error output is awesomely weird and hard to detect. in [Circle CI](https://cirrus-ci.com/task/5422000227745792?logs=test#L2564) it looks like this:
~~~
 URL: -
 in: https://xn--rksmrgs-5wao1o.se
 wanted: https | [11] | [12] | [13] | räksmörgås.se | [15] | / | [16] | [17]
 got:    https | [11] | [12] | [13] | räksmörgås.se | [15] | / | [16] | [17]
~~~

But when I copy and pasted it here, it shows:

~~~
 URL: -
 in: https://xn--rksmrgs-5wao1o.se
 wanted: https | [11] | [12] | [13] | räksmörgås.se | [15] | / | [16] | [17]
 got:    https | [11] | [12] | [13] | räksmörgås.se | [15] | / | [16] | [17]
~~~

Assuming I get the correct trailing byte, it is 0x16.

jay:
> got:    https | [11] | [12] | [13] | räksmörgås.se� | [15] | / | [16] | [17]

I think you should have passed the length of `in_w` + 1. it says pass the count of characters of param lpASCIICharStr (which is in wide characters) and +1 is needed because of [this](https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-idntounicode#return-value): _"The retrieved string is null-terminated only if the input string is null-terminated."_

i don't know where -1 comes from maybe that was in earlier versions of the doc i will check

bagder:
I just guessed on the -1 since it seems to work for the conversion in the other direction...

jay:
> I just guessed on the -1 since it seems to work for the conversion in the other direction...

I just checked an old Windows SDK offline documentation from Windows 7 2009 and I do not see it documented there either so I don't know why we did it that way. I will change it.

bagder:
awesome @jay, that seems to have done it!

PR #11661, opened 2023-08-11
cookie-jar.d: enphasize that this option is ONLY writing cookies

bagder:
Ref: #11642

### DIFF
--- a/docs/cmdline-opts/cookie-jar.d
+++ b/docs/cmdline-opts/cookie-jar.d
@@ -19,9 +19,9 @@ written. The file will be written using the Netscape cookie file format. If
 you set the file name to a single dash, "-", the cookies will be written to
 stdout.
 
-This option specifies a file that cookies are save to. Not read from. To tell
-curl to read cookies from a file, use --cookie. These two options can specify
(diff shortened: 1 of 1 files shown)


docs/cmdline-opts/cookie-jar.d
```diff
@@ -19,9 +19,12 @@ written. The file will be written using the Netscape cookie file format. If
 you set the file name to a single dash, "-", the cookies will be written to
 stdout.
 
+This option specifies a file that cookies are save to. Not read from. To tell
```
jay, line 22: s/save/saved/ but I like the way --cookies says it better, and in reverse:

The file specified with [-c, --cookie-jar](https://curl.se/docs/manpage.html#-c) is only used for output. No cookies will be read from the file. To read cookies, use the [-b, --cookie](https://curl.se/docs/manpage.html#-b) option. Both options can specify the same file.

bagder: agreed!

jay (approved):
I would add `Reported-by: Dan Jacobson`

PR #11668, opened 2023-08-12
http_aws_sigv4: skip special user header entries

### LINKED ISSUE #11664: aws-sigv4 signature covers non-existent Expect: header
### I did this

```
curl -v
 --header 'expect:'
 --header 'Content-Type: application/json'
 --data '[ ... foo ... ]'
 --aws-sigv4 aws:amz:eu-central-1:appsync
 --user 'akid:secret'
 'https://xxx.appsync-api.eu-central-1.amazonaws.com/graphql'
```

### I expected the following

Correct AWS SigV4, i.e. `expect` is not part of curl's generated `SignedHeaders`, as it is not part of the emitted request headers, like this:
```
authorization: AWS4-HMAC-SHA256
 Credential=xxx/20230812/eu-central-1/appsync/aws4_request,
 SignedHeaders=content-type;host;x-amz-date,
 Signature=xxx
```

But what actually happens is this:
```
authorization: AWS4-HMAC-SHA256
 Credential=xxx/20230812/eu-central-1/appsync/aws4_request,
 SignedHeaders=content-type;expect;host;x-amz-date,
 Signature=xxx
```

Per man page, using `--header 'expect:'` should remove that header. The header is indeed not present, but it seems that by forcing the removal of this non-existent header, it somehow gets added to the list of headers to be considered for aws-sigv4.

Additional context:

Some AWS APIs just ignore this (e.g. `sts`), but others report failure, e.g. Appsync GraphQL with IAM Auth:
```
  "errors" : [ {
    "errorType" : "IncompleteSignatureException",
    "message" : "'expect' is named as a SignedHeader, but it does not exist in the HTTP request."
  } ]
```

This seems to happen in libcurl – I've discovered it while playing around with Hurl, which explicitly sets `Expect:`.

### curl/libcurl version

`curl 8.0.1 (aarch64-amazon-linux-gnu) libcurl/8.0.1 OpenSSL/3.0.8 zlib/1.2.11 libidn2/2.3.2 nghttp2/1.51.0`

### operating system

Amazon Linux 2023

jay:
- Do not include user header entries in the format 'name:' with no value in AWS SignedHeaders.

User header entries in that format are not actually used to specify a no-value header, instead they are used to signal that the user wants that internal header removed. For example if the user passes a header "Expect:" then that means they want libcurl to not send that header.

Prior to this change those header names were included in AWS SignedHeaders.

Reported-by: apparentorder@users.noreply.github.com

Fixes #11664
Closes #xxxx

---

Note I did not cover the case if the user actually does want a no-value header, which is specified as ending in a semicolon, eg `-H "foo;"` causes "foo:" header to be sent which means AWS SignedHeaders would show something like  `SignedHeaders=content-type;foo;;host;x-amz-date, Signature=...`. Someone familiar with AWS will need to see if that is correct behavior.

/cc @outscale-mgo

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -199,10 +199,41 @@ static CURLcode make_headers(struct Curl_easy *data,
     head = tmp_head;
   }
 
+  /* copy user headers to our header list. the logic is based on how http.c
+     handles user headers.
+
+     user headers in format 'name:' with no value are used to signal that an
+     internal header of that name should be removed. those user headers are not
+     added to this list.
+
+     user headers in format 'name;' with no value are used to signal that a
+     header of that name with no value should be sent. those user headers are
+     added to this list but in the format that they will be sent, ie the
+     semi-colon is changed to a colon for format 'name:'.
+
+     user headers with a value of whitespace only, or without a colon or
+     semi-colon, are not added to this list.
+     */
   for(l = data->set.headers; l; l = l->next) {
-    tmp_head = curl_slist_append(head, l->data);
-    if(!tmp_head)
+    char *dupdata, *ptr;
+    char *sep = strchr(l->data, ':');
+    if(!sep)
+      sep = strchr(l->data, ';');
+    if(!sep || (*sep == ':' && !*(sep + 1)))
+      continue;
+    for(ptr = sep + 1; ISSPACE(*ptr); ++ptr)
+      ;
+    if(!*ptr && ptr != sep + 1) /* a value of whitespace only */
+      continue;
+    dupdata = strdup(l->data);
+    if(!dupdata)
       goto fail;
+    dupdata[sep - l->data] = ':';
+    tmp_head = Curl_slist_append_nodup(head, dupdata);
+    if(!tmp_head) {
+      free(dupdata);
+      goto fail;
+    }
     head = tmp_head;
   }
 
--- a/tests/data/test1955
+++ b/tests/data/test1955
@@ -64,9 +64,10 @@ http://exam.ple.com:9000/aws_sigv4/testapi/test exam.ple.com:9000:%HOSTIP:%HTTPP
 <protocol>
 GET /aws_sigv4/testapi/test HTTP/1.1
 Host: exam.ple.com:9000
-Authorization: XXX4-HMAC-SHA256 Credential=xxx/19700101/ple/exam/xxx4_request, SignedHeaders=content-type;host;tesmixcase;test0;test1;test2;test_space;x-xxx-date, Signature=779a8ff876528aece8bf03b1296702af0644a4745aa5feabb6ebb1a7bb0d907e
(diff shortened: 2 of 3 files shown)


outscale-mgo:
I've reread this https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html

I think AWS tend to assume the header will be something like `"NAME: VALUE"`. and doesn't really specify what happen if VALUE is not present (I might have miss a part of the doc though).

So I guess `foo;;` is okay
But I think your code skip the header for the SignedHeader, but not for `Canonical header` computation here: 
https://github.com/curl/curl/pull/11668/files#diff-1d1fee636fd00dd0694ff6db673a8b5ab1ec1d3db540e748645f7b406cf64843R261
Which I guess should not include `'name:'` either. (maybe removing `name:` from the header list at the begin make_headers would be easier than spiking it every time)
Also a test would be nice :)
Thanks for improving signature v4 code.

lib/http_aws_sigv4.c
```diff
@@ -200,6 +200,15 @@ static CURLcode make_headers(struct Curl_easy *data,
   }
 
   for(l = data->set.headers; l; l = l->next) {
+    char *ptr = strchr(l->data, ':');
+    /* if the user header is in format 'name:' with no value then it is being
+       used to ignore an internal header so do not add it to the header list */
+    if(ptr) {
+      for(++ptr; ISSPACE(*ptr); ++ptr)
+        ;
```
bagder, line 208: I think we can insist on `name:` with no spaces at all when removing headers, so this check could be simplified a little.

jay:
So this turned out to be a bit more complicated than I thought. I redid the changes to handle both 'name:' and 'name;'. There is still a check for whitespace because I want to ignore value of whitespace only like `'name:   '` and `'name;   '` and missing separator 'name'  since those would not actually be sent. A weird quirk I noticed while testing this is the Host header if specified with only whitespace is sent by libcurl, unlike arbitrary headers. So for example
~~~
curl -v -H "foo;" -H "bar:" -H "baz:   " -H "Host:   " http://httpbin.org

> GET / HTTP/1.1
> Host:
> User-Agent: curl/8.3.0-DEV
> Accept: */*
> foo:
~~~
Since that is bad syntax I did not special case any further and the code ignores the value of whitespace only in all cases. Technically though if someone did a corner case like this they would get a bad signature, not that it would matter though because the request is already bad.

jay:
> But I think your code skip the header for the SignedHeader, but not for `Canonical header` computation here:

https://github.com/curl/curl/blob/c90c78333bbd15d171d26d3d4d005066ed152ca5/lib/http_aws_sigv4.c#L252-L270

true but it's iterating through the signed header list `head` (the just built list that skips the special user headers) and not through `data->set.headers` (the user headers unmodified)

> Also a test would be nice :)

I've updated and added comments to test 1955 about the special cases and I will need to recalculate the signature

bagder:
> Linux torture / torture Failing after 46m

^^

PR #11671, opened 2023-08-14
Added detection of custom FETCH and UID FETCH commands for IMAP

MonkeybreadSoftware:
Added detection of custom FETCH and UID FETCH commands for IMAP


MonkeybreadSoftware:
See https://github.com/curl/curl/pull/8659 for more details.

MonkeybreadSoftware:
To explain this. With current curl 7.82.0, we get this:

```
cs@MyMac curl-7.82.0 % /Users/cs/Temp/curl-7.82.0/src/.libs/curl -u "test@domain.com:password" -X "UID FETCH 107 BODY.PEEK[]" "imap://sslin.df.eu/INBOX" 
* 2 FETCH (UID 107 BODY[] {2211}
```

but we like to have this:

```
cs@MyMac curl-7.82.0 % curl -u "test@domain.com:password" "imap://sslin.df.eu/INBOX;UID=107"                 
Delivery-date: Wed, 30 Jan 2019 16:36:03 +0100
Received: from [80.67.18.7] (helo=mx07.ispgateway.de)
...
```

so the output of the transaction is the source code of the email and not the list line. A small change, but if you write an IMAP email application, it makes a huge difference.

lib/imap.c
```diff
@@ -663,8 +663,34 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
-  if(!result)
-    imap_state(data, IMAP_LIST);
+  if(!result) {
+    bool isFetch = false;
+
+    /* Switch state to be FETCH if custom command is a fetch command */
+    /* so output is handled correctly. */
+
+    if(imap->custom) {
+      if(strncasecompare(imap->custom, "FETCH", 5) == 1) {
```
bagder, line 673:
```suggestion
      if(strncasecompare(imap->custom, "FETCH", 5)) {
```

lib/imap.c
```diff
@@ -663,8 +663,34 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+        imap_state(data, IMAP_FETCH);
+        isFetch = true;
+      }
+
+      /* e.g. "UID FETCH 107 BODY.PEEK[]" */
+
+      if(!isFetch && imap->custom_params) {
+        if(strncasecompare(imap->custom, "UID", 3) == 1) {
```
bagder, line 681:
```suggestion
        if(strncasecompare(imap->custom, "UID", 3)) {
```

lib/imap.c
```diff
@@ -663,8 +663,34 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+          if(strncasecompare(imap->custom_params, " FETCH", 6) == 1) {
```
bagder, line 682:
```suggestion
          if(strncasecompare(imap->custom_params, " FETCH", 6)) {
```

lib/imap.c
bagder: ... as we don't compare with 1 for this kind of boolean functions anywhere.

bagder:
> Basically whether CURL makes a "FETCH" and you make a "FETCH" (or UID FETCH) via custom request, both are handled the same.

Why? If it already can do the transfer correctly without a custom request, why do you need to add support for doing the same transfer in a different way?

lib/imap.c
```diff
@@ -663,8 +663,34 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
     free(mailbox);
   }
 
-  if(!result)
-    imap_state(data, IMAP_LIST);
+  if(!result) {
+    bool isFetch = false;
```
bagder, line 667:
```suggestion
    bool isFetch = FALSE;
```

bagder: We use `TRUE` and `FALSE` everywhere.

MonkeybreadSoftware:
We need a little adjustment.

Such a request to list emails need to get LIST state, while fetch of email then needs FETCH state:

"UID FETCH 1:* (FLAGS INTERNALDATE RFC822.SIZE BODY.PEEK[HEADER.FIELDS (Message-Id DATE FROM SUBJECT TO SENDER REPLY-TO CC BCC)])"

MonkeybreadSoftware:
I added check to distinguish between a fetch of one email vs. a listing of a range of emails.

tests/test1665
```diff
@@ -0,0 +1,46 @@
...
+<verify>
+<protocol>
+A001 CAPABILITY
+A002 LOGIN user secret
+A003 SELECT %TESTNUMBER
+A004 FETCH 123 BODY[1]
+A005 LOGOUT
+</protocol>
+</verify>
+</testcase>
```
bagder, line 46: This is in the wrong directory. Move it into `data/`.

MonkeybreadSoftware: Done. Sorry for all my mistakes here 😳

lib/imap.c
```diff
@@ -663,8 +663,57 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+          SUBJECT TO SENDER REPLY-TO CC BCC)])
+          to get the list of emails
+        */
+        const char *positionColon   = strchr(imap->custom_params, ':');
+        const char *positionBracket = strchr(imap->custom_params, '(');
+
+        if(positionColon) {
+          if(!positionBracket || positionBracket > positionColon) {
+            isFetch = FALSE;
+          }
```
bagder, line 705:
```suggestion
        if(positionColon) {
          const char *paren = strchr(imap->custom_params, '(');
          if(!paren || paren > positionColon)
            isFetch = FALSE;
```

lib/imap.c
```diff
@@ -663,8 +663,52 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+          otherwise FETCH state, e.g.
+          FETCH 175 BODY.PEEK[]
+          to get one email
+          vs.
+          UID FETCH 1:* (FLAGS INTERNALDATE RFC822.SIZE
+          BODY.PEEK[HEADER.FIELDS (Message-Id DATE FROM
+          SUBJECT TO SENDER REPLY-TO CC BCC)])
+          to get the list of emails
+        */
+        const char *positionColon   = strchr(imap->custom_params, ':');
```
bagder, line 696: The scan-build warns for a (reasonable) possible NULL dereference here.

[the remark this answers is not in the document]
MonkeybreadSoftware: I've removed it. This was a left over.

MonkeybreadSoftware: Let me put custom_params in a local variable and put in "" if it s NULL.

lib/imap.c
```diff
@@ -663,8 +663,54 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+      }
+
+      /* e.g. "UID FETCH 107 BODY.PEEK[]" */
+
+      else if(strncasecompare(imap->custom, "UID", 3) &&
+              strncasecompare(imap->custom_params, " FETCH", 6)) {
+        isFetch = true;
+      }
+
+      if(isFetch) {
```
bagder, line 684:
```suggestion
      if(isFetch && imap->custom_params) {
```

lib/imap.c
```diff
@@ -663,8 +663,54 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+          to get one email
+          vs.
+          UID FETCH 1:* (FLAGS INTERNALDATE RFC822.SIZE
+          BODY.PEEK[HEADER.FIELDS (Message-Id DATE FROM
+          SUBJECT TO SENDER REPLY-TO CC BCC)])
+          to get the list of emails
+        */
+        const char *custom_params =
+          imap->custom_params ? imap->custom_params : "";
+        const char *positionColon = strchr(custom_params, ':');
```
bagder, line 698:
```suggestion
        const char *positionColon = strchr(imap->custom_params, ':');
```

... when combined with the change above.

lib/imap.c
```diff
@@ -663,8 +663,52 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+          SUBJECT TO SENDER REPLY-TO CC BCC)])
+          to get the list of emails
+        */
+        const char *positionColon   = strchr(imap->custom_params, ':');
+        if(positionColon) {
+          const char *paren = strchr(imap->custom_params, '(');
+          if(!paren || paren > positionColon)
+            isFetch = FALSE;
+        }
+      }
```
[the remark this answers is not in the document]
bagder: I'm sure there are several other IMAP users out there, but we rarely see anything from them apart from the infrequent issues filed...

MonkeybreadSoftware:
Thanks for little changes. Feels like we do pair programming on this.

lib/imap.c
```diff
@@ -663,8 +663,52 @@ static CURLcode imap_perform_list(struct Curl_easy *data)
...
+          to get one email
+          vs.
+          UID FETCH 1:* (FLAGS INTERNALDATE RFC822.SIZE
+          BODY.PEEK[HEADER.FIELDS (Message-Id DATE FROM
+          SUBJECT TO SENDER REPLY-TO CC BCC)])
+          to get the list of emails
+        */
+        const char *positionColon = strchr(imap->custom_params, ':');
+        if(positionColon) {
+          const char *paren = strchr(custom_params, '(');
```
bagder, line 698:
```suggestion
          const char *paren = strchr(imap->custom_params, '(');
```

bagder:
It seems this how makes a lot of CI jobs get stuck on test 647, which I cannot understand...

MonkeybreadSoftware:
I've added another NULL check to prevent crash.

bagder:
This still fails too many CI jobs.

MonkeybreadSoftware:
I see, but I can't find a reason why test 647 fails.
Can you run this test and see if you get an error or crash report?

bagder:
Failing CI jobs not updated since September, closing.

PR #11674, opened 2023-08-15
urlapi: return CURLUE_BAD_HOSTNAME if puny2idn encoding fails

bagder:
And document it

### DIFF
--- a/lib/idn.c
+++ b/lib/idn.c
@@ -68,9 +68,9 @@ WINBASEAPI int WINAPI IdnToUnicode(DWORD dwFlags,
 
 #define IDN_MAX_LENGTH 255
 
-bool Curl_win32_idn_to_ascii(const char *in, char **out)
+static CURLcode win32_idn_to_ascii(const char *in, char **out)
 {
-  bool success = FALSE;
+  CURLcode result = CURLE_OK;
 
   wchar_t *in_w = curlx_convert_UTF8_to_wchar(in);
   if(in_w) {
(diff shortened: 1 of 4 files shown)


bagder:
Alternatives:
1. have the flag be a "best effort" and just return it as-is if it fails (but that then doesn't tell the user if it worked)
2. introduce a new error code to make the IDN conversion failure more explicit
3. keep this PR

bagder:
@jacobmealey thoughts?

jacobmealey:
it looks like `idn2_to_unicode_8z8z(puny, &enc, 0)` is never throwing an error when passed some malformed punycode. It _should_ return [IDN2_ENCODING_ERROR](https://www.gnu.org/software/libidn/libidn2/reference/libidn2-idn2.h.html#IDN2-ENCODING-ERROR:CAPS) but its returning 0 (IDN2_OK) and an empty string. I've only done preliminary testing but that's what I've noticed. 

note that I'm using `libidn2.so.0.3.8`

bagder:
we could of course also remove the check for the `xn--` prefix and then trow a string at it that does not start with `xn--`...

bagder:
> it may be good to have curl_url_get return the same error for Curl_idn_decode

Ack. I'll make sure that it returns CURLUE_BAD_HOSTNAME for bad host names and CURLUE_OUT_OF_MEMORY only when it actually runs out of memory.

lib/idn.c
```diff
@@ -157,52 +173,68 @@ static char *idn_decode(const char *input)
...
+  if(!result)
+    *output = d;
+  return result;
 }
 
-char *Curl_idn_encode(const char *puny)
+CURLcode Curl_idn_encode(const char *puny, char **output)
 {
-  char *d = idn_encode(puny);
+  char *d;
```
jacobmealey, line 223: This either needs to be declared as NULL or the if statement on line 226 should be checking the result. Currently on a failed idn conversion this causes a `free(): invalid pointer` error on my machine. Changing the if statement or declaring d as NULL fixed it for me

bagder: :+1:

PR #11675, opened 2023-08-15
TLS session sharing test, BearSSL affection

icing:
- test TLS session sharing with special test client
- expect failure with wolfSSL
- disable flaky wolfSSL test_02_07b
- test apache: add suitable session cache directive

Other improvements
- generic vtls `get_select_socks()` improved
- BearSSL now with own `get_select_socks()` which fixes the socket polling during handshake. This results in reducing handshake times down to what other TLS backens achieve.


icing:
Ready for merge from my side.

bagder (approved):

PR #11677, opened 2023-08-15
tool: change some fopen failures from warnings to errors

jay:
- Error if a user-specified file cannot be opened.

- Change file2memory and file2string functions to return an error if passed a null FILE, instead of returning zero length data.

Prior to this change, if a user of the curl tool specified a file that could not be opened for reading data then curl would show a warning instead of an error and treat it as zero length data. For example, a POST using `--data @filenametypo` would cause a zero length POST which is probably not what the user intended.

Closes #xxxx


MarcelRaad (approved):

jay:
I haven't looked at all the CI results but I can see there are at least two tests failing 268 344 both of which expect curl to be able to handle non existent files. I can remove those tests but I think it's pretty clear now it was someone's intention to allow this behavior.

For example rather than create foo.txt and bar.txt beforehand if the open fails just treat it as empty

--etag-compare foo.txt --etag-save foo.txt

--cookie bar.txt --cookie-jar bar.txt

hm

jay:
I plan to dial this PR back somewhat since it seems there are some intended cases where it may be useful to allow non-existent files to be treated as empty.

jay:
Ok I limited this change to error on missing input files for --data, --data-binary, --data-urlencode, --header, --variable, --write-out. [test268](https://github.com/curl/curl/blob/curl-8_2_1/tests/data/test268) is going to fail because it tests to make sure --data will POST zero length content when passed a non-existent filename. IMO if the user passes a non-existent file for POST that's an accident not intentional. From 2005 639857ce @bagder

bagder:
I think trying to post a file that can't be read should rather cause an error.

icing:
It does not make work easier when we merge changes that break CI. Just saying.

PR #11679, opened 2023-08-15
http: close the connection after a late 417 received

### LINKED ISSUE #11678: curl reuses connection when non-chunked data has not entirely been sent
### I did this

I caught a case where test 357 hung, while running the test suite under CPU starvation conditions. What seems to happen is this:

1. curl sends a PUT request with `Expect: 100-continue` with HTTP/1.1
2. it waits a moment for a 100 Continue response which doesn't come
3. it starts sending its ~1 MiB payload in non-chunked mode
4. while sending that payload (in 64 KiB blocks) the server returns a 417 "expectation failed" code
5. curl *immediately* stops sending the remaining payload bytes, in order to start again without the `Expect: 100-continue` (again, in non-chunked mode)
6. curl improperly **reuses the existing connection** to send the new response
7. the server treats the start of the new request as part of the payload of the first request, and finishes reading the remaining bytes that it expected
8. the server then finds itself in the middle of the payload of the second request, gets confused and just sits there
9. curl completes sending the data of the second request and starts waiting for the HTTP response, which never comes

Here is an annotated excerpt from the trace log for the hanging test 357:

```
10:15:31.811080 == Info: !!! WARNING !!!
10:15:31.811167 == Info: This is a debug build of libcurl, do not use in production.
10:15:31.811171 == Info: processing: http://127.0.0.1:44697/we/want/357
10:15:31.811190 == Info: STATE: INIT => CONNECT handle 0x19dbde8; line 1974
10:15:31.811405 == Info: Added connection 0. The cache now contains 1 members
10:15:31.811451 == Info: STATE: CONNECT => CONNECTING handle 0x19dbde8; line 2027
10:15:31.811647 == Info:   Trying 127.0.0.1:44697...
10:15:31.811776 == Info: Connected to 127.0.0.1 (127.0.0.1) port 44697
10:15:31.811784 == Info: STATE: CONNECTING => PROTOCONNECT handle 0x19dbde8; line 2135
10:15:31.811789 == Info: STATE: PROTOCONNECT => DO handle 0x19dbde8; line 2165
10:15:31.811852 => Send header, 140 bytes (0x8c)
0000: PUT /we/want/357 HTTP/1.1
001b: Host: 127.0.0.1:44697
0032: User-Agent: curl/8.3.0-DEV
004e: Accept: */*
005b: Content-Length: 1053701
0074: Expect: 100-continue
008a:
10:15:31.811888 == Info: STATE: DO => DID handle 0x19dbde8; line 2259
10:15:31.811893 == Info: STATE: DID => PERFORMING handle 0x19dbde8; line 2377
10:15:32.889446 == Info: Done waiting for 100-continue
10:15:32.889863 => Send data, 65536 bytes (0x10000)
0000: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
0040: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
0080: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
(more of this, to reach exactly 65536 bytes of payload)
ff40: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
ff80: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
ffc0: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
That's 65536 bytes, when the server is expecting 1053701 bytes.
```
10:15:33.872300 == Info: HTTP 1.1 or later with persistent connection
10:15:33.872316 <= Recv header, 26 bytes (0x1a)
0000: HTTP/1.1 417 OK swsbounce.
10:15:33.872359 <= Recv header, 36 bytes (0x24)
0000: Date: Tue, 09 Nov 2010 14:49:00 GMT.
10:15:33.872374 <= Recv header, 25 bytes (0x19)
0000: Server: test-server/fake.
10:15:33.872385 <= Recv header, 18 bytes (0x12)
0000: Content-Length: 0.
10:15:33.872398 == Info: Got 417 while waiting for a 100
10:15:33.872404 <= Recv header, 1 bytes (0x1)
0000: .
10:15:33.872412 == Info: multi_done: status: 0 prem: 0 done: 0
10:15:33.872476 == Info: Connection #0 to host 127.0.0.1 left intact
10:15:33.872540 == Info: Issue another request to this URL: 'http://127.0.0.1:44697/we/want/357'
10:15:33.872545 == Info: STATE: PERFORMING => CONNECT handle 0x19dbde8; line 2553
10:15:33.872655 == Info: Found bundle for host: 0x19d77b8 [serially]
10:15:33.872659 == Info: Can not multiplex, even if we wanted to
10:15:33.872679 == Info: Re-using existing connection with host 127.0.0.1
```
That's the smoking gun ^^^
```
10:15:33.872685 == Info: STATE: CONNECT => CONNECTING handle 0x19dbde8; line 2027
10:15:33.872690 == Info: STATE: CONNECTING => PROTOCONNECT handle 0x19dbde8; line 2135
10:15:33.872694 == Info: STATE: PROTOCONNECT => DO handle 0x19dbde8; line 2154
10:15:33.873002 => Send header, 118 bytes (0x76)
0000: PUT /we/want/357 HTTP/1.1
001b: Host: 127.0.0.1:44697
0032: User-Agent: curl/8.3.0-DEV
004e: Accept: */*
005b: Content-Length: 1053701
0074:
10:15:33.873029 == Info: STATE: DO => DID handle 0x19dbde8; line 2259
10:15:33.873034 == Info: STATE: DID => PERFORMING handle 0x19dbde8; line 2377
10:15:33.873367 => Send data, 65536 bytes (0x10000)
0000: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
0040: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
…
```

### I expected the following

In step 6, curl should close the existing connection (since it is incomplete) and start a new one instead.

### curl/libcurl version

8.3.0-DEV

### operating system

Linux x86_64

dfandrich:
In this situation, only part of the data has been sent before aborting
so the connection is no longer usable.

Fixes #11678
Closes #11679

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -4270,7 +4270,18 @@ CURLcode Curl_http_readwrite_headers(struct Curl_easy *data,
               if((k->httpcode == 417) && data->state.expect100header) {
                 /* 417 Expectation Failed - try again without the Expect
                    header */
-                infof(data, "Got 417 while waiting for a 100");
+                if(!k->writebytecount &&
+                   k->exp100 == EXP100_AWAITING_CONTINUE) {
+                  infof(data, "Got HTTP failure 417 while waiting for a 100");
+                }
+                else {
+                  infof(data, "Got HTTP failure 417 while sending data");
+                  streamclose(conn,
+                              "Stop sending data before everything sent");
+                  result = http_perhapsrewind(data, conn);
+                  if(result)
+                    return result;
+                }
                 data->state.disableexpect = TRUE;
                 DEBUGASSERT(!data->req.newurl);
                 data->req.newurl = strdup(data->state.url);
--- a/tests/FILEFORMAT.md
+++ b/tests/FILEFORMAT.md
@@ -332,6 +332,7 @@ about to issue.
 
 - `auth_required` if this is set and a POST/PUT is made without auth, the
   server will NOT wait for the full request body to get sent
+- `delay: [msecs]` - delay this amount after connection
 - `idle` - do nothing after receiving the request, just "sit idle"
 - `stream` - continuously send data to the client, never-ending
 - `writedelay: [msecs]` delay this amount between reply packets
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -185,7 +185,7 @@ test1439 test1440 test1441 test1442 test1443 test1444 test1445 test1446 \
 test1447 test1448 test1449 test1450 test1451 test1452 test1453 test1454 \
 test1455 test1456 test1457 test1458 test1459 test1460 test1461 test1462 \
 test1463 test1464 test1465 test1466 test1467 test1468 test1469 test1470 \
-test1471 test1472 test1473 \
+test1471 test1472 test1473 test1474 \
 \
 test1500 test1501 test1502 test1503 test1504 test1505 test1506 test1507 \
 test1508 test1509 test1510 test1511 test1512 test1513 test1514 test1515 \
--- a/tests/data/test1328
+++ b/tests/data/test1328
@@ -10,7 +10,7 @@ globbing
 
 <reply>
 <data>
-HTTP/1.1 404 OK
+HTTP/1.1 404 NOT OK
 Date: Tue, 09 Nov 2010 14:49:00 GMT
 Content-Length: 6
 Funny-head: yesyes
--- a/tests/data/test1474
+++ b/tests/data/test1474
@@ -1,0 +1,111 @@
+<testcase>
+# This test is quite timing dependent and tricky to set up. The time line of
+# test operations looks like this:
+#
+# 1. curl sends a PUT request with Expect: 100-continue and waits only 1 msec
+#    for a 100 response.
+# 2. The HTTP server accepts the connection but waits 500 msec before acting
+#    on the request.
+# 3. curl doesn't receive the expected 100 response before its timeout expires,
+#    so it starts sending the body. It is throttled by a --limit-rate, so it
+#    sends the first 64 KiB then stops for 1000 msec due to this
+#    throttling.
+# 4. The server sends its 417 response while curl is throttled.
+# 5. curl responds to this 417 response by closing the connection (because it
+#    has a half-completed response outstanding) and starting a new one. This
+#    new request does not have an Expect: header so it is sent without delay.
+#    It's still throttled, however, so it takes about 16 seconds to finish
+#    sending.
+# 6. The server receives the response and this time acks it with 200.
+#
+# Because of the timing sensitivity (scheduling delays of 500 msec can cause
+# the test to fail), this test is marked flaky to avoid it being run in the CI
+# builds which are often run on overloaded servers.
+# Increasing the --limit-rate would decrease the test time, but at the cost of
+# becoming even more sensitive to delays (going from 500 msec to 250 msec or
+# less of accepted delay before failure).
+<info>
+<keywords>
+HTTP
+HTTP PUT
+Expect
+flaky
+</keywords>
+</info>
+# Server-side
+<reply>
+# 417 means the server didn't like the Expect header
+<data>
+HTTP/1.1 417 BAD swsbounce
+Date: Tue, 09 Nov 2010 14:49:00 GMT
+Server: test-server/fake
+Content-Length: 0
+
+</data>
+<data1>
+HTTP/1.1 200 OK
+Date: Tue, 09 Nov 2010 14:49:00 GMT
+Server: test-server/fake
+Content-Length: 10
+
+blablabla
+</data1>
+<datacheck>
+HTTP/1.1 417 BAD swsbounce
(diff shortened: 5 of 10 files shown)


lib/http.c
```diff
@@ -4270,6 +4270,7 @@ CURLcode Curl_http_readwrite_headers(struct Curl_easy *data,
                 data->state.disableexpect = TRUE;
                 DEBUGASSERT(!data->req.newurl);
                 data->req.newurl = strdup(data->state.url);
+                streamclose(conn, "Stop sending data before everything sent");
```
jay, line 4273: the info line for this block says it's handling a 417 while waiting for the 100, but your issue says that isn't what's happening. instead the time passed and data was partially sent. so is it correct to put it in this section specifically. like wouldn't data->req.writebytecount be nonzero for what you describe

dfandrich: It is still "hoping" for a 100, but it's past the point of sitting there idle waiting. It seems to handle the header (whether 100 or 417) as soon as it's received, so this looks to me like the right spot to put it. If it's not 417 the connection is closed in the same way in the block a few lines afterward. Yes, writebytecount is nonzero at this point in this failure scenario.

jay: what i mean is why not put some conditionals on it. isn't that info line misleading? aren't they mutually exclusive so if we receive 417 reply then we aren't going to receive 100 reply

~~~diff
diff --git a/lib/http.c b/lib/http.c
index 860843a..fd63dc0 100644
--- a/lib/http.c
+++ b/lib/http.c
@@ -4266,11 +4266,17 @@ CURLcode Curl_http_readwrite_headers(struct Curl_easy *data,
               if((k->httpcode == 417) && data->state.expect100header) {
                 /* 417 Expectation Failed - try again without the Expect
                    header */
-                infof(data, "Got 417 while waiting for a 100");
+                if(!k->writebytecount &&
+                   data->req.exp100 == EXP100_AWAITING_CONTINUE) {
+                  infof(data, "Got HTTP failure 417 while waiting for a 100");
+                }
+                else {
+                  infof(data, "Got HTTP failure 417 while sending data");
+                  streamclose(conn, "Stop sending data before everything sent");
+                }
                 data->state.disableexpect = TRUE;
                 DEBUGASSERT(!data->req.newurl);
                 data->req.newurl = strdup(data->state.url);
-                streamclose(conn, "Stop sending data before everything sent");
                 Curl_done_sending(data, k);
               }
               else if(data->set.http_keep_sending_on_error) {
~~~

dfandrich:
I had assumed that that precondition was already there for this block of code.
I'll see if that's the case or not.

dfandrich:
Looks like you were right. I've updated the change accordingly.

bagder (approved):

bagder:
Have you tried to trigger the problematic case more reliably in a (new) test case?

icing:
> Have you tried to trigger the problematic case more reliably in a (new) test case?

Or more general: does curl handle the HTTP/1.1 upload state correctly when a response comes in before the upload is complete? This does normally not happen, but HTTP does not forbid this case. Maybe I can reproduce such a scenario in the pytest with out special curl Apache module.

bagder:
> Or more general: does curl handle the HTTP/1.1 upload state correctly when a response comes in before the upload is complete?

It is expected to, yes. I believe we even have test case(s) for this.

icing:
> > Or more general: does curl handle the HTTP/1.1 upload state correctly when a response comes in before the upload is complete?
> 
> It is expected to, yes. I believe we even have test case(s) for this.

Apache httpd too smart. Sets "Connection: close" when the request body was not fully read.

dfandrich:
A lot of the mac builds seem to hanging (i.e. timing out), and they're not coughing up the logs. This might be tricky.

dfandrich:
I've tweaked the test by adding a "delay" command to the test HTTP server, as discussed offline. This should slightly reduce the tests susceptibility to pauses due to overloaded servers, but I still think it's prudent to mark it as "flaky" since it only takes a 500 msec delay to cause the test to fail. I'm happy enough to submit it like this.

bagder:
I'm totally fine with that. We can possibly try to un-flaky it later, but there's no need to stress that now.

PR #11682, opened 2023-08-16
lib: move mimepost data from HTTP to urldata->state

### LINKED ISSUE #11680: Using curl_formadd to post file could not rewind correctly when redirection happens
### I did this

I am using curl_formadd to post MIME data to server, and server replies redirection.  But it seems libcurl didn't rewind correctly. 
I think this phenomenon was cause by this commit 1b3973180b6b7f1023240a67d3acdb11b534bf03
In the version before 7.87.0, libcurl will do rewind immediately, but in version 7.87.0, libcurl will only make a mark, and then do the rewind operation in the next round of state machine driving. But at that time, the cached **req** member seems to be reinitialized again, causing **http->sendit** to be empty, and when sending data with curl_formadd, mimepost is also empty, and rewind is not executed correctly.

### I expected the following

I'm wondering if my understanding is correct or not. Would like get your confirmation.

### curl/libcurl version

curl 8.1.2

### operating system

Windows x86_64

bagder:
Fixes #11680

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -233,7 +233,6 @@ static CURLcode http_setup_conn(struct Curl_easy *data,
   if(!http)
     return CURLE_OUT_OF_MEMORY;
 
-  Curl_mime_initpart(&http->form);
   data->req.p.http = http;
   connkeep(conn, "HTTP default");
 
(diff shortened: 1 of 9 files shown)


bagder:
I edited my approach and now the converted data (which `curl_formadd()` style posts get, is kept in the easy handle for its lifetime - which should be better. I also edited test 650 verify that a redirect works with legacy formpost.

PR #11691, opened 2023-08-18
lib: lower the default connect timeout from 300 to 30 seconds

bagder:
It was always a very large value. 30 seconds is still a long time.


jay:
I disagree. This has been documented at 300 seconds and could be a breaking change for some users.

bagder:
> I disagree. This has been documented at 300 seconds and could be a breaking change for some users.

It could be a breaking change yes, there's no doubt about that.

But I also think there are a large number of scripts and command lines used where they unnecessary wait for those 300 seconds to end.

I don't think the fact that we once set a default value to X must mean that we can never change it. We just need to weigh the tradeoffs carefully.

How about a smaller change, to say 200 seconds? (Which could be a way for us to *gradually* lower this over a period of many years.)

jay:
> I don't think the fact that we once set a default value to X must mean that we can never change it.

Sure I agree.

> How about a smaller change, to say 200 seconds?

I think that of the two ideas proposed that is the better one but in my opinion I'd leave it.

PR #11694, opened 2023-08-19
tool/var: also error when expansion result starts with NUL

emanuele6:
Expansions whose output starts with NUL were being expanded to the empty string, and not being recognised as values that contain a NUL byte, and should error.

### DIFF
--- a/tests/data/test456
+++ b/tests/data/test456
@@ -1,0 +1,33 @@
+<testcase>
+<info>
+<keywords>
+HTTP
+variables
+</keywords>
+</info>
+
+#
+# Client-side
+<client>
+<server>
+http
+</server>
+<name>
+Variable output starting with null byte
+</name>
+<file name="%LOGDIR/junk">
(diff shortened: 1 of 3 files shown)


tests/data/test456
```diff
@@ -0,0 +1,33 @@
...
+#
+# Client-side
+<client>
+<server>
+http
+</server>
+<name>
+Variable output starting with null byte
+</name>
+<file name="%LOGDIR/junk">
```
emanuele6, line 18: I have used `%LOGDIR/junk` here because that is the file all the `test45?` tests are using, but I am now wondering if this is fine.
Wouldn't this possibly cause problems when running tests in parallel?

bagder: When run in parallel, each version gets its own %LOGDIR so I think it is fine.

bagder (approved):

bagder:
Thanks!

Issue #11451, opened 2023-07-17
Autobuilds: globbing tests failing


MarcelRaad:
### I did this

Look at the autobuilds at https://curl.se/dev/builds.html.

### I expected the following

Everything mostly green. Instead, the NetBSD 9.3 ARM and Solaris 10 SPARC runners both show failures in the following 5 tests:
0086 [HTTP, urlglob []-retrieval and -o #[num] usage]
0087 [urlglob with out of range -o #[num] usage]
1240 [glob [0-1] with stuff after range (7.33.0 regression)]
1283 [globbing range with same start and stop]
1328 [HTTP GET a globbed range with -f]

All of the failures are of the following pattern:
```
86: protocol FAILED:
--- log/check-expected 2023-07-16 04:09:16.958642306 +0200
+++ log/check-generated 2023-07-16 04:09:16.953213662 +0200
@@ -1,14 +1,14 @@
-GET /860001 HTTP/1.1[CR][LF]
+GET /0 HTTP/1.1[CR][LF]
```

### curl/libcurl version

master

### operating system

NetBSD 9.3 ARM and Solaris 10 SPARC (not i386!)

bagder:
@MarcelRaad this looks fixed now, right?

MarcelRaad:
Right, probably fixed by https://github.com/curl/curl/commit/20e3287aca994308cb040340d4d1865526cbf312, thanks!

bagder:
Fixed via 20e3287aca994308cb040340d4d1865526cbf312

PR #11698, opened 2023-08-20
tool_urlglob: use the correct format specifier for curl_off_t in msnprintf

majstar:
Hello.
I have encountered a problem with URL globbing functionality (incorrect output of values from number ranges).
On my system (BS2000/OSD) curl_off_t and unsigned long have different size, which leads to a problem when printing curl_off_t value using %lu format specifier.
My commit fixes the problem.


bagder (approved):

bagder:
I suspect this might also fix #11451

bagder:
Thanks!

PR #11706, opened 2023-08-21
Ci freebsd debug

icing:


bagder:
I figure we just pause these efforts and come back to trying to get this to work again in a few weeks or so?

bagder:
according to [Val Pakett](https://blahaj.zone/notes/9iorh7758l):

> someone already noticed this but your CI config is deleting rust right after installing it. pkg delete -y curl deletes rust because curl is a dependency of cargo. cargo uses libcurl to fetch dependencies.

> I'm not sure why you're upgrading pip with pip anyway, that seems like a waste of time in CI and it's not recommended to just pip as root overriding OS packages willy nilly

bagder:
FreeBSD then obviously provides libcurl in the curl package only, so when curl is removed it removes libcurl as well. This is not how Linux distros tend to work and, well, a weakness I would propose.

So we can't just remove the curl tool on FreeBSD without doing damage.

We can probably drop the removal, as it is only there for the occasional mistake when things accidentally fall back to run the system curl instead of the newly built one.

PR #11709, opened 2023-08-22
cirrus: do not delete curl

bagder:
- we need rust for the python crypto package
- since rust needs libcurl, deleting curl removes libcurl and rust
- remove one of the FreeBSD build to use less CPU


bagder:
Did not fix the problem. Scratch this.

Issue #11710, opened 2023-08-22
Wrong version mentioned for --disable (-q) in docs/options-in-versions


pszlazak:
### I did this

Having following settings in `~/.curlrc`:
```
--write-out \n
--netrc-optional
```
I executed:
```
curl --disable -v https://some.host/
```
In the output I noticed:
```
* Couldn't find host some.host in the .netrc file; using defaults
```

### I expected the following

I assumed config options from `~/.curlrc` will **not** be used, especially `--netrc-optional`.

According to [`docs/options-in-versions`](https://github.com/curl/curl/blob/master/docs/options-in-versions#L46C1-L46C15) [option to disable](https://curl.se/docs/manpage.html#-q) reading `curlrc` file was introduced in 5.0:
```
--disable (-q)                       5.0
```

This is a little bit confusing. I guess it could be like this instead:
```
--disable (-q)                       7.49.0 (5.0)
```

### curl/libcurl version

```
curl 7.29.0 (x86_64-redhat-linux-gnu) libcurl/7.29.0 NSS/3.44 zlib/1.2.7 libidn/1.28 libssh2/1.8.0
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp scp sftp smtp smtps telnet tftp
Features: AsynchDNS GSS-Negotiate IDN IPv6 Largefile NTLM NTLM_WB SSL libz unix-sockets
```

### operating system

`Red Hat Enterprise Linux Server release 7.5 (Maipo)`

jay:
curl 5.0 has --disable. It has to be used as the **first** curl parameter. Is that the actual command line you executed? Maybe you are seeing a bug that is not present in later versions. What happens if you use `-q` as the first parameter?

pszlazak:
I noticed that `--disable` was added as alias for `-q` in 7.49.0:  
https://github.com/curl/curl/compare/curl-7_48_0...curl-7_49_0 (click 'Files changed' tab and open diff for `scr/tool_operate.c`)

But I missed that there was one more correction in 7.50.0. It was even mentioned in [changelog](https://curl.se/changes.html) as:  
> [curl: fix -q [regression]](https://curl.se/bug/?i=842)

My main suggestion is to clearly state this in [`docs/options-in-versions`](https://github.com/curl/curl/blob/master/docs/options-in-versions#L46C1-L46C15):
```
--disable (-q)                       7.50.0 (5.0)
```

But that would be only option having appearance in two different versions...

[...]

PR #11712, opened 2023-08-22
disable.d: explain --disable not implemented prior to 7.50.0

jay:
Option -q/--disable was added in 5.0 but only -q was actually implemented. Later --disable was implemented in e200034 (precedes 7.49.0), but incorrectly, and fixed in 6dbc23c (precedes 7.50.0).

Reported-by: pszlazak@users.noreply.github.com

Fixes https://github.com/curl/curl/issues/11710
Closes #xxxx

### DIFF
--- a/docs/cmdline-opts/disable.d
+++ b/docs/cmdline-opts/disable.d
@@ -9,6 +9,9 @@ Added: 5.0
 See-also: config
 Multi: boolean
 ---
-If used as the first parameter on the command line, the *curlrc* config
+If used as the **first** parameter on the command line, the *curlrc* config
 file will not be read and used. See the --config for details on the default
 config file search path.
+
+Prior to 7.50.0 curl supported the short option name *q* but not the long
+option name *disable*.


jay:
I'm not sure how to stop the hyperlink conversion to both forms of the option when I want only the one form. I tried putting them in bold like `*--disable*` so it won't convert to `-q, --disable`.

bagder:
We might need to introduce some new syntax for that. Putting the asterisks around it seems to not be correct either.

docs/cmdline-opts/disable.d
```diff
@@ -9,6 +9,9 @@ Added: 5.0
 See-also: config
 Multi: boolean
 ---
-If used as the first parameter on the command line, the *curlrc* config
+If used as the **first** parameter on the command line, the *curlrc* config
 file will not be read and used. See the --config for details on the default
 config file search path.
+
+Prior to 7.50.0 curl supported the short option name *-q* but not the long
+option name *--disable*.
```
bagder, line 17: Maybe you can just avoid naming the long option in this last paragraph?

jay: I removed the dashes so it should show as this now:
Prior to 7.50.0 curl supported the short option name *q* but not the long option name *disable*.

bagder (approved):

PR #11721, opened 2023-08-23
secureserver.pl: fix stunnel path quoting

jay:
- Store the stunnel path in the private variable $stunnel unquoted and instead quote it in the command strings.

Prior to this change the quoted stunnel path was passed to perl's file operators which cannot handle quoted paths. For example:

~~~
$stunnel = "\"/C/Program Files (x86)/stunnel/bin/tstunnel\"";
if(-x $stunnel or -x "$stunnel")
# false even if path exists and is executable
~~~

Our other test scripts written in perl, unlike this one, use servers.pm which has a global $stunnel variable with the path stored unquoted and therefore those scripts don't have this problem.

Closes #xxxx

### DIFF
--- a/tests/secureserver.pl
+++ b/tests/secureserver.pl
@@ -123,12 +123,7 @@ while(@ARGV) {
     }
     elsif($ARGV[0] eq '--stunnel') {
         if($ARGV[1]) {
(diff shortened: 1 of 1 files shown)


nmoinvaz:
LGTM

PR #11723, opened 2023-08-24
docs/cmdline-opts: match the current output

eworm-de:
The release date has been added in output, reflect that in documentation.


bagder:
Thanks!

PR #11735, opened 2023-08-25
tool: avoid including leading spaces in the Location hyperlink

mssdvd:
This PR ensures that leading whitespaces are not marked as part of the Location URL.

Before:
image

After:
image

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -349,6 +349,7 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
   char *copyloc = NULL, *locurl = NULL, *scheme = NULL, *finalurl = NULL;
   const char *loc = location;
   size_t llen = loclen;
+  int space_skipped = 0;
   char *vver = getenv("VTE_VERSION");
 
   if(vver) {
@@ -360,9 +361,10 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
   }
 
   /* Strip leading whitespace of the redirect URL */
-  while(llen && *loc == ' ') {
+  while(llen && (*loc == ' ' || *loc == '\t')) {
     ++loc;
     --llen;
+    ++space_skipped;
   }
 
   /* Strip the trailing end-of-line characters, normally "\r\n" */
@@ -401,8 +403,10 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
      !strcmp("https", scheme) ||
      !strcmp("ftp", scheme) ||
      !strcmp("ftps", scheme)) {
-    fprintf(stream, LINK "%s" LINKST "%.*s" LINKOFF,
-            finalurl, (int)loclen, location);
+    fprintf(stream, "%.*s" LINK "%s" LINKST "%.*s" LINKOFF,
+            space_skipped, location,
+            finalurl,
+            (int)loclen - space_skipped, loc);
     goto locdone;
   }
 


src/tool_cb_hdr.c
```diff
@@ -401,8 +404,8 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
      !strcmp("https", scheme) ||
      !strcmp("ftp", scheme) ||
      !strcmp("ftps", scheme)) {
-    fprintf(stream, LINK "%s" LINKST "%.*s" LINKOFF,
-            finalurl, (int)loclen, location);
+    fprintf(stream, " " LINK "%s" LINKST "%.*s" LINKOFF,
+            finalurl, (int)prevllen, loc);
```
bagder, line 408: why should this use `prevllen` and not `locclen`? The first one includes whitespace and the latter does not.

bagder:
When I do `curl -I curl.se`, my terminal (Konsole) does not include the leading whitespace in the link.

mssdvd:
alacritty, foot and kitty include the space. I believe Konsole disables escape sequences for links by default.

bagder:
Ah, now I see the problem. Isn't this patch simpler?
~~~diff
diff --git a/src/tool_cb_hdr.c b/src/tool_cb_hdr.c
index aab96ae03..b1bad12e0 100644
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -399,12 +399,12 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
 
   if(!strcmp("http", scheme) ||
      !strcmp("https", scheme) ||
      !strcmp("ftp", scheme) ||
      !strcmp("ftps", scheme)) {
-    fprintf(stream, LINK "%s" LINKST "%.*s" LINKOFF,
-            finalurl, (int)loclen, location);
+    fprintf(stream, " " LINK "%s" LINKST "%.*s" LINKOFF,
+            finalurl, (int)loclen, loc);
     goto locdone;
   }
 
   /* Not a "safe" URL: don't linkify it */
 ~~~

> I believe Konsole disables escape sequences for links by default.

No it doesn't. At least I get it linkified just fine.

bagder:
Or maybe this is even better:
~~~diff
diff --git a/src/tool_cb_hdr.c b/src/tool_cb_hdr.c
index aab96ae03..878e8a563 100644
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -399,12 +399,12 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
 
   if(!strcmp("http", scheme) ||
      !strcmp("https", scheme) ||
      !strcmp("ftp", scheme) ||
      !strcmp("ftps", scheme)) {
-    fprintf(stream, LINK "%s" LINKST "%.*s" LINKOFF,
-            finalurl, (int)loclen, location);
+    fprintf(stream, " " LINK "%s" LINKST "%s" LINKOFF "\n",
+            finalurl, copyloc);
     goto locdone;
   }
 
   /* Not a "safe" URL: don't linkify it */

mssdvd:
I had thought of fixing it like that. Is there a change that not specifying the length will lead to bugs?

bagder:
Not that I can spot since `copyloc` is null-terminated string it should be fine. But I still would like to get @dfandrich to chime in, he wrote this code.

dfandrich:
The original intention was to preserve the whitespace so the user gets the same output with or without styling enabled (modula the styling, of course). The proposed patch strips leading whitespace and replaces it with a single character, which is often, but not always, the same. I didn't bother eliminating the leading whitespace from the linked section, but I can see how it could be annoying on some terminals.

What I would probably do instead is count the amount of whitespace stripped in the `/* Strip leading whitespace of the redirect URL */` loop by adding a `++space_skipped` in the loop, then change the fprintf to:

```
    fprintf(stream, "%.*s" LINK "%s" LINKST "%.*s" LINKOFF,
             space_skipped, location, finalurl, (int)loclen - space_skipped, loc);
```

This is untested but the idea is to print the leading whitespace as-is, then the rest of the linkification without that leading space.

dfandrich (approved):
It works for me!

src/tool_cb_hdr.c
```diff
@@ -363,6 +364,7 @@ void write_linked_location(CURL *curl, const char *location, size_t loclen,
   while(llen && *loc == ' ') {
```
dfandrich, line 364: It turns out that HTTP allows a tab character here as whitespace as well, so that should also be skipped.

dfandrich (changes requested):
skip tabs

dfandrich:
Thanks!

PR #11740, opened 2023-08-26
misc: remove unneeded duplicate words

jbampton:


docs/TheArtOfHttpScripting.md
```diff
@@ -156,7 +156,7 @@
     curl https://curl.se
 
  you get a web page returned in your terminal window. The entire HTML document
- that that URL holds.
+ that URL holds.
```
bagder, line 159: This was intentional. Just removing one makes it strange language. The second one could possibly be made a `the` to avoid the repitition.

bagder (changes requested):

docs/TheArtOfHttpScripting.md
emanuele6: I also like "The entire HTML document that that URL holds" better

bagder (approved):

bagder:
Thanks!

PR #11742, opened 2023-08-27
tool_paramhlp: improve str2num() by avoiding unnecessary call to strlen()

foutaise:
After reading @bagder last blog entry regarding CVE-2020-19909, I had a look to str2num() function and realized that a small improvement might be applied to avoid an unnecessary call to strlen().


bagder (approved):
:+1:

bagder:
Thanks!

PR #11743, opened 2023-08-27
altvc: accept and parse IPv6 address in response headers

### LINKED ISSUE #11737: Alt-svc: support IPv6 alt-authority
### I did this

Run `curl -v --alt-svc cache "https://caddy.pfad.fr"`

caddy.pfad.fr is a test server which replies with an `Alt-svc` header containing an IPv6 as alt-authority:
```
alt-svc: h3="[2a01:4f8:c0c:9a6d::42]:443"; ma=2592000
```

### I expected the following

The cache file to be populated.

Instead, I see:
```
* Excessive alt-svc host name, ignoring.
< alt-svc: h3="[2a01:4f8:c0c:9a6d::42]:443"; ma=2592000
```

Triggered by this code:

https://github.com/curl/curl/blob/c2212c05aa99bb31e4b99b0b66fc1747b7d01be6/lib/altsvc.c#L503-L507

Specifications:
- `alt-authority = quoted-string ; containing [ uri-host ] ":" port`  ([RFC7838](https://httpwg.org/specs/rfc7838.html#alt-svc))
- `uri-host      = <host, see [RFC3986], Section 3.2.2>`  [RFC7230](https://www.rfc-editor.org/rfc/rfc7230.html#section-2.7)
- `host        = IP-literal / IPv4address / reg-name` [RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2)
- `IP-literal = "[" ( IPv6address / IPvFuture  ) "]"` according to this same RFC

Hence I think that `h3="[2a01:4f8:c0c:9a6d::42]:443"; ma=2592000` is a valid `alt-svc` header value and should be correctly handled by curl.

---

Background information:
 
Each of my service listens on 1 dedicated IPv6 address on the same machine. The traffic to the shared IPv4 gets forwarded to the right service thanks to [snid](https://github.com/AGWA/snid/) (which uses SNI). However this setup can't handle UDP packets, hence I make http3 only available on IPv6.

Possible workaround: publish AAAA-only DNS record and use this domain as alt-svc (so IPv4 clients will waste a bit of there resources).

### curl/libcurl version

curl 8.2.1 (x86_64-pc-linux-gnu) libcurl/8.2.1 OpenSSL/3.1.2 zlib/1.3 brotli/1.0.9 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.11.0 nghttp2/1.55.1
Release-Date: 2023-07-26

### operating system

Arch Linux:
Linux 6.4.11-arch2-1 #1 SMP PREEMPT_DYNAMIC Sat, 19 Aug 2023 15:38:34 +0000 x86_64 GNU/Linux

bagder:
Store numerical IPv6 addresses in the alt-svc file with the brackets present.

Verify with test 437 and 438

Fixes #11737
Reported-by: oliverpool on github
Closes #.....


lib/altsvc.c
```diff
@@ -231,18 +253,40 @@ static CURLcode altsvc_load(struct altsvcinfo *asi, const char *file)
...
+  const char *dst6_pre = "";
+  const char *dst6_post = "";
+  const char *src6_pre = "";
+  const char *src6_post = "";
   CURLcode result = Curl_gmtime(as->expires, &stamp);
   if(result)
     return result;
-
+#ifdef ENABLE_IPV6
+  else {
```
jay, line 264: I think this would be better if you got rid of the else block and moved the declaration with the rest.

#ifdef ENABLE_IPV6
char ipv6_addr[16];
#endif

bagder: I sort of like reducing the number of #ifdefs rather than increasing them, which is why I took this approach...

jay:
commit area altvc typo should be altsvc

lib/altsvc.c
jay: Yes that's a benefit. When I read something like  if(foo) return foo; else { code block} I think unnecessary block but either way it's fine.

PR #11746, opened 2023-08-28
lib539.c: use curl_free() to free libcurl allocated data

bagder:
Reported-by: Nicholas Nethercote


jay:
this needs a url to the report if there is one

bagder:
> this needs a url to the report if there is one

There is none. This is based on an IRC conversation.

PR #11750, opened 2023-08-28
curl_multi_get_handles - get all easy handles in a multi handle

bagder:
Food for discussion. See https://curl.se/mail/lib-2023-08/0034.html

### DIFF
--- a/include/curl/multi.h
+++ b/include/curl/multi.h
@@ -426,6 +426,17 @@ CURL_EXTERN CURLMcode curl_multi_setopt(CURLM *multi_handle,
 CURL_EXTERN CURLMcode curl_multi_assign(CURLM *multi_handle,
                                         curl_socket_t sockfd, void *sockp);
 
+/*
+ * Name:    curl_multi_get_handles()
+ *
+ * Desc:    Returns an allocated array holding all handles currently added to
+ *          the multi handle. Marks the final entry with a NULL pointer. If
+ *          there is no easy handle added to the multi handle, this function
+ *          returns an array with the first entry as a NULL pointer.
+ *
+ * Returns: NULL on failure, otherwise a CURL **array pointer
+ */
+CURL_EXTERN CURL **curl_multi_get_handles(CURLM *multi_handle);
 
 /*
  * Name: curl_push_callback
--- a/docs/libcurl/Makefile.inc
+++ b/docs/libcurl/Makefile.inc
@@ -73,6 +73,7 @@ man_MANS = \
  curl_multi_assign.3 \
  curl_multi_cleanup.3 \
  curl_multi_fdset.3 \
+ curl_multi_get_handles.3 \
  curl_multi_info_read.3 \
  curl_multi_init.3 \
  curl_multi_perform.3 \
--- a/docs/libcurl/curl_multi_get_handles.3
+++ b/docs/libcurl/curl_multi_get_handles.3
@@ -1,0 +1,73 @@
+.\" **************************************************************************
+.\" *                                  _   _ ____  _
+.\" *  Project                     ___| | | |  _ \| |
+.\" *                             / __| | | | |_) | |
+.\" *                            | (__| |_| |  _ <| |___
+.\" *                             \___|\___/|_| \_\_____|
+.\" *
+.\" * Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+.\" *
+.\" * This software is licensed as described in the file COPYING, which
+.\" * you should have received as part of this distribution. The terms
+.\" * are also available at https://curl.se/docs/copyright.html.
+.\" *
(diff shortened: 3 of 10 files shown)


include/curl/multi.h
```diff
@@ -426,6 +426,17 @@ CURL_EXTERN CURLMcode curl_multi_setopt(CURLM *multi_handle,
...
 
+/*
+ * Name:    curl_multi_get_handles()
+ *
+ * Desc:    This function returns an allocated array holding all handles currently
+ *          added to the multi handle. It marks the end with a NULL pointer.
+ *          If there is no handles added, it returns an array with the first entry
+ *          a NULL pointer.
+ *
+ * Returns: NULL on failure, otherwise a CURL **array pointere
```
stof, line 437:
```suggestion
 * Returns: NULL on failure, otherwise a CURL **array pointer
```

MonkeybreadSoftware (commented):
Looks good to me.

bagder:
@jeroen you think this kind of API works for a binding such as yours, or would it be better served by an iterator-using one?

jeroen:
Yes this works for me, but my consideration is that this api is not extensible, if you would want to expose additional status information about each handle in the multi. For example, if the request is in queued, in progress, or done, and which connection it uses, incl http version, etc. 

There is a possibility to make this a more generic API to list the status of all the requests in the multi, which is something that can be useful e.g. to debug multiplex issues, for example what we discussed here: https://curl.se/mail/lib-2023-02/0024.html

A fun example to toy with might be to improve the [crawler](https://github.com/curl/curl/blob/master/docs/examples/crawler.c) to show detailed progress about the pending transfers, and be a bit smarter how to limit the number of requests based on what is already in the queue.

bagder:
> if you would want to expose additional status information about each handle in the multi. For example, if the request is in queued, in progress, or done, and which connection it uses, incl http version, etc.

If we want to provide that info - per easy handle - I think the *get_handles() API is not the right spot anyway. Then we should provide that via `curl_easy_getinfo()` so you'd iterate over the handles and query each handle for that info.

> There is a possibility to make this a more generic API

You mean by instead creating a `curl_multi_getinfo()` function and making this all-easy-handles  be a first "option" that this function can return?

bagder:
> - add a CURLHND parameter to curl_multi_get_handles so that the user can request only their (user) easy handles, or internal handles, or both,  in the multi.

Can we think of a use case where users actually want the internally created ones?

If they do, don't they then also want to know which kind each of them is? I mean, it seems like we should rather provide either non-internal or only-interal ones so that the caller knows what kind of handles they are.

include/curl/multi.h
```diff
@@ -436,7 +442,7 @@ CURL_EXTERN CURLMcode curl_multi_assign(CURLM *multi_handle,
  *
  * Returns: NULL on failure, otherwise a CURL **array pointer
  */
-CURL_EXTERN CURL **curl_multi_get_handles(CURLM *multi_handle);
+CURL_EXTERN CURL **curl_multi_get_handles(CURLM *multi_handle, CURLHND filter);
```
bagder, line 445: I prefer we call it 'flags' and make it more open ended. Probably as an 'int' instead of a typedef.

bagder: Or even unsigned int, to give us 32 bits of flags.

jay:
> Can we think of a use case where users actually want the internally created ones?

I can't. Maybe I overdid it I'll take away the filter parameter.

jay:
I should mention my hackish [ListEasyHandles](https://gist.github.com/jay/5b68c1bde985b1b2d23f2b583a26e5ad) example from several years ago that I just updated :)

rwmjones:
This would work for our use case.

bagder:
Any further comments before we merge this first version?

Should we mark it experimental in a first round to make it possible to tweak the API ?

rwmjones:
I read the updated patch series that you posted a few days ago and it still looks fine and would work for us.  I have no view on whether it should be experimental or not.

rwmjones:
> I have no view on whether it should be experimental or not.

We would use it right away, but conditionalize the code behind a `./configure`-time test, so for us it wouldn't matter.

bagder:
I intend to merge this as-is and give everyone a chance to try it out for a few weeks. If we have any issues or problems showing up, we can mark it experimental before we do the next release. Otherwise it can ship as-is.

This function is pretty simple and straight-forward. I don't think it should cause many problems.

rwmjones:
Thanks, here are the changes to nbdkit incorporating this new API (optionally):

https://gitlab.com/nbdkit/nbdkit/-/commit/025db8b7fea237fd64e5231a1ebd03b30ca2f49a
https://gitlab.com/nbdkit/nbdkit/-/commit/81340d68643b5c18861f95c6f113764305d10c53

PR #11751, opened 2023-08-28
Fix for c-ares timeout issue

sachusachin622:
Fix for https://github.com/c-ares/c-ares/issues/541
@bradh352 @bagder Please help me to review
Thanks

### DIFF
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -3238,4 +3238,5 @@ CURL_EXTERN CURLcode curl_easy_pause(CURL *handle, int bitmask);
 #endif /* __STDC__ >= 1 */
 #endif /* gcc >= 4.3 && !__cplusplus && !CURL_DISABLE_TYPECHECK */
 
+extern long curl_to_ares_timeout_ms;
 #endif /* CURLINC_CURL_H */
--- a/lib/asyn-ares.c
+++ b/lib/asyn-ares.c
@@ -173,6 +173,10 @@ CURLcode Curl_resolver_init(struct Curl_easy *easy, void **resolver)
   int optmask = ARES_OPT_SOCK_STATE_CB;
   options.sock_state_cb = sock_state_cb;
   options.sock_state_cb_data = easy;
+  if (curl_to_ares_timeout_ms) {
+    options.timeout = curl_to_ares_timeout_ms;
+    optmask |= ARES_OPT_TIMEOUTMS;
+  }
   status = ares_init_options((ares_channel*)resolver, &options, optmask);
   if(status != ARES_SUCCESS) {
     if(status == ARES_ENOMEM)
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -64,6 +64,7 @@
       }                                         \
     }                                           \
   } while(0)
+long curl_to_ares_timeout_ms;
 
 struct LongShort {
   const char *letter; /* short name option */
@@ -2276,6 +2277,7 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
     case 'm':
       /* specified max time */
       err = secs2ms(&config->timeout_ms, nextarg);
+      curl_to_ares_timeout_ms = config->timeout_ms;
       break;
     case 'M': /* M for manual, huge help */
       if(toggle) { /* --no-manual shows no manual... */


include/curl/curl.h
```diff
@@ -3238,4 +3238,5 @@ CURL_EXTERN CURLcode curl_easy_pause(CURL *handle, int bitmask);
 #endif /* __STDC__ >= 1 */
 #endif /* gcc >= 4.3 && !__cplusplus && !CURL_DISABLE_TYPECHECK */
 
+extern long curl_to_ares_timeout_ms;
```
bagder, line 3241: A hard no here. You cannot introduce global variables - anywhere - in curl or libcurl.

bagder (changes requested):
Why do you need to set the timeout for c-ares? libcurl manages its own timeout already.

sachusachin622:
> Why do you need to set the timeout for c-ares? libcurl manages its own timeout already.

In the context of using Curl, if a timeout of 20 seconds (-m 20 option) is set, while C-ares employs DEFAULT_TIMEOUT as 5000 (5 seconds) and DEFAULT_TRIES as 4, the outcome is as follows: C-ares initiates the sending of requests in a sequence, with intervals determined by an exponential progression. Specifically, C-ares transmits the first request after a 5-second interval, the second request after a 10-second interval, the third request after a 20-second interval, and the fourth request after a 40-second interval. The cumulative time taken to receive a partial query response (IPV4 address) from C-ares to pass it to Curl totals 75 seconds (1 minute and 15 seconds). During this period, Curl, having already elapsed the 20-second timeout, has concluded with a failure response.

**Solution**
From the above explanation of C-ares timeout logic, C-ares should return the partial query response to curl within curl timeout. So that we can derive the timeout of C-ares from timeout of curl.
Timeout of C-ares = ( Timeout of curl / 2^number of retries )

Dividing the curl timeout by 2^number of retries to get Timeout of C-ares because C-ares follows an exponential pattern for sending successive requests.

Example:
Consider curl timeout is 20 seconds and c-ares number of retries is 4 then C-ares timeout value is

Timeout of C-ares = (Timeout of curl/2^number of retries)

Timeout of C-ares = (20/2^4) = (20/16)

Timeout of C-ares = 1.25 seconds.

C-ares will send request every 1.25 seconds, 2.50 seconds , 5 seconds, and 10 seconds total 18.75 seconds required to send back response to curl.
C-ares able to send the partial query response back to curl before curl gets time's out, so that it can use the ipv4 address at least and continue with its operations.

Here is the packet capture for the above example:

_r_oot@[bd05f1ebe90d:/curl-8.2.1#](http://bd05f1ebe90d/curl-8.2.1) time curl -m20 -v http://www.google.com/

processing: http://www.google.com/
07:08:22.567944 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 58166+ A? [www.google.com](http://www.google.com/). (32)
07:08:22.568075 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
07:08:22.568758 IP 10.174.75.167.domain > bd05f1ebe90d.44543: 58166* 1/0/0 A 142.250.195.228 (49)
07:08:22.568767 IP 10.174.75.167.domain > bd05f1ebe90d.44543: 58166* 1/0/0 A 142.250.195.228 (49)
07:08:23.818355 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
07:08:26.319154 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
07:08:31.320595 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
Trying 142.250.195.228:80...
Connected to [www.google.com](http://www.google.com/) (142.250.195.228) port 80
GET / HTTP/1.1__

**Please guide me If anything wrong or missing here**

Thanks

include/curl/curl.h
sachusachin622: Hi @bagder ,
I want use this timeout_ms variable in Curl_resolver_init() function but not able to use. Can you please point me how can i access this variable in this Curl_resolver_init() function.

Thanks

bagder:
I made  #11753

include/curl/curl.h
bagder: Options passed to the library are done with `curl_easy_setopt`. There are 300+ options there already. You can check existing timeout options for how they are set form the tool and used by the library.

That said, I don't think users want to set this timeout separately.

PR #11753, opened 2023-08-28
asyn-ares: reduce timeout to 2000ms

bagder:
When UDP packets gets lost this makes for slightly faster retries.


bagder:
This is also done by @c-ares itself starting next release: https://github.com/c-ares/c-ares/pull/542

PR #11754, opened 2023-08-28
Fix aws sigv4 date twice

### LINKED ISSUE #11738: aws-sigv4 generates extra X-Amz-Date header (possibly different)
### I did this

```
curl -v --aws-sigv4 aws:amz:foo --header 'X-Amz-Date: 20230813T061344Z' --user foo:bar https://httpbin.org/post
```

### I expected the following

The provided `X-Amz-Date` is used for signature calculation and is emitted once.

But actually it's emitted twice:
```
> GET /post HTTP/2
> Host: httpbin.org
> authorization: AWS4-HMAC-SHA256 Credential=foo/20230825/foo/httpbin/aws4_request, SignedHeaders=host;x-amz-date, Signature=bcecfc18afb2e19fe031bfd2bbaa68faffb3ff3cc3e990b014217422f422860a
> x-amz-date: 20230825T202910Z
> user-agent: curl/7.81.0
> accept: */*
> x-amz-date: 20230813T061344Z
```

On some curl versions (tested: 7.81.0), the first `X-Amz-Date` is actually generated with a *different* date (the current date), and that current date is used for signature calculation instead of the `X-Amz-Date` given on the command line.

Other versions (tested: 8.1.2) emit the given `X-Amz-Date` twice as well, but both with the given value; in that case, the signature is calculated as expected.

As this is a rare / weird use-case, maybe it should just be added to the list of known bugs (?)

### curl/libcurl version

curl 7.81.0 (x86_64-pc-linux-gnu) libcurl/7.81.0 OpenSSL/3.0.2 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libidn2/2.3.2 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.43.0 librtmp/2.3 OpenLDAP/2.5.14

curl 8.1.2 (x86_64-apple-darwin22.0) libcurl/8.1.2 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0

### operating system

curl 8.1.2 on macOS 13.5, or
Darwin xxx 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:21:56 PDT 2023; root:xnu-8796.141.3~6/RELEASE_X86_64 x86_64

curl 7.81.0 on Ubuntu 22.04.2, or
Linux ip-10-0-0-148 5.19.0-1025-aws #26~22.04.1-Ubuntu SMP Mon Apr 24 01:58:15 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

outscale-mgo:
Fix #11738

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -214,14 +214,10 @@ static CURLcode make_headers(struct Curl_easy *data,
     if(!tmp_head)
       goto fail;
     head = tmp_head;
-    *date_header = curl_maprintf("%s: %s", date_hdr_key, timestamp);
+    *date_header = curl_maprintf("%s: %s\r\n", date_hdr_key, timestamp);
   }
   else {
     char *value;
-
-    *date_header = strdup(*date_header);
-    if(!*date_header)
-      goto fail;
 
     value = strchr(*date_header, ':');
     if(!value)
@@ -231,6 +227,7 @@ static CURLcode make_headers(struct Curl_easy *data,
       ++value;
     strncpy(timestamp, value, TIMESTAMP_SIZE - 1);
     timestamp[TIMESTAMP_SIZE - 1] = 0;
+    *date_header = NULL;
   }
 
   /* alpha-sort in a case sensitive manner */
@@ -612,14 +609,19 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
                                "Credential=%s/%s, "
                                "SignedHeaders=%s, "
                                "Signature=%s\r\n"
-                               "%s\r\n"
+                               /*
+                                * date_header is added here, only if it wasn't
+                                * pass as header (using CURLOPT_HTTPHEADER),
+                                * otherwise, date_header includes \r\n
+                                */
+                               "%s"
                                "%s", /* optional sha256 header includes \r\n */
                                provider0,
                                user,
                                credential_scope,
                                Curl_dyn_ptr(&signed_headers),
                                sha_hex,
-                               date_header,
+                               date_header ? date_header : "",
                                content_sha256_hdr);
   if(!auth_headers) {
     goto fail;
--- a/tests/libtest/lib1964.c
+++ b/tests/libtest/lib1964.c
@@ -1,0 +1,68 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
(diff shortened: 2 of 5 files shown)


tests/libtest/lib1964.c
```diff
@@ -0,0 +1,69 @@
...
+    fprintf(stderr, "curl_global_init() failed\n");
+    return TEST_ERR_MAJOR_BAD;
+  }
+
+  curl = curl_easy_init();
+  if(!curl) {
+    fprintf(stderr, "curl_easy_init() failed\n");
+    curl_global_cleanup();
+    return TEST_ERR_MAJOR_BAD;
+  }
```
jay, line 45: test.h has macros that do these checks

```suggestion
  global_init(CURL_GLOBAL_ALL);
  easy_init(curl);
```

also easy_setopt outputs the error and test_setopt doesn't so I'd switch that as well

and you don't need CURLOPT_HEADER, 0L); that's the default

lib/http_aws_sigv4.c
```diff
@@ -612,14 +609,14 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
                                "Credential=%s/%s, "
                                "SignedHeaders=%s, "
                                "Signature=%s\r\n"
-                               "%s\r\n"
+                               "%s" /* if providedd by user, must not be add */
```
bagder, line 612: This comment does not make sense (apart from the broken English). If this header is provided by the user, it is not added *here* no, but it is still added to the request.

outscale-mgo: I've just changed it to:
```C
                               /*                                                                                                                                                                           
                                * date_header is added here, only if it wasn't                                                                                                                              
                                * pass as header (using -H),                                                                                                                                           
                                * otherwise, date_header include \r\n                                                                                                                                       
                                */
```

lib/http_aws_sigv4.c
```diff
@@ -612,14 +609,19 @@ CURLcode Curl_output_aws_sigv4(struct Curl_easy *data, bool proxy)
                                "Credential=%s/%s, "
                                "SignedHeaders=%s, "
                                "Signature=%s\r\n"
-                               "%s\r\n"
+                               /*
+                                * date_header is added here, only if it wasn't
+                                * pass as header (using -H),
```
bagder, line 614: The header is actually set with `CURLOPT_HTTPHEADER` since this is library code.

jay:
Thanks

PR #11755, opened 2023-08-29
test/http, add tests for POSTing to redirect resources

icing:
adds tests cases for POSTing to 301, 302, 303 and 307 resources.


icing:
Will include this in another PR.

PR #11756, opened 2023-08-29
h2/h3, things around POST, slow network simulation

icing:
- added test cases for various code paths
- fixed handling of blocked write when stream had been closed inbetween attempts
- re-enabled DEBUGASSERT on send with smaller data size

slow network simulation, fixes

- in debug builds, environment variables can be set to simulate a slow
   network when sending data. cf-socket.c and vquic.c support
    * CURL_DBG_SOCK_WBLOCK: percentage of send() calls that should be
      answered with a EAGAIN. TCP/UNIX sockets.
      This is chosen randomly.
    * CURL_DBG_SOCK_WPARTIAL: percentage of data that shall be written
      to the network. TCP/UNIX sockets.
      Example: 80 means a send with 1000 bytes would only send 800
      This is applied to every send.
    * CURL_DBG_QUIC_WBLOCK: percentage of send() calls that should be
      answered with EAGAIN. QUIC only.
      This is chosen randomly.

- Fixes in http2.c and cf-h2-proxy.c to make test cases pass
- tests: some tests are skipped when the env vars are set, because
   they would time out just because of the amount of requests/data
- multi.c: on debug, print the state a transfer was in when it
    switched to done, e.g. was ended/aborted.

### DIFF
--- a/.github/workflows/ngtcp2-linux.yml
+++ b/.github/workflows/ngtcp2-linux.yml
@@ -267,3 +267,11 @@ jobs:
       name: 'run pytest'
       env:
         TFLAGS: "${{ matrix.build.tflags }}"
+
+    - run: pytest
+      name: 'run pytest with slowed network'
+      env:
+        # 33% of sends are EAGAINed
+        CURL_DBG_SOCK_WBLOCK: 33
+        # only 80% of data > 10 bytes is send
+        CURL_DBG_SOCK_WPARTIAL: 80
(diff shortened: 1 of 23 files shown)


jay:
Is this related to https://github.com/curl/curl/pull/11342#pullrequestreview-1597792624

bagder:
I believe it is work to polish code, which ideally makes the people in #11342 who still experience problems happy.

It brings back the assert that was removed because of #11500, so OSS-fuzz can help us verify this.

icing:
> Is this related to [#11342 (review)](https://github.com/curl/curl/pull/11342#pullrequestreview-1597792624)

This is part of the hunt for this issue, unsuccessful so far, but discovering other edge paths.

bagder (approved):

PR #11766, opened 2023-08-30
wolfssl, use master again in CI

icing:
- with the shared session update fix landed in master, it is time to use that in our CI again

### DIFF
--- a/.github/workflows/ngtcp2-linux.yml
+++ b/.github/workflows/ngtcp2-linux.yml
@@ -46,14 +46,12 @@ permissions: {}
 
 env:
   MAKEFLAGS: -j 3
-  openssl-version: 3.0.10+quic
+  quictls-version: 3.0.10+quic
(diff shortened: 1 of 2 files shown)


jay:
rebased on master due to ci fix

.github/workflows/ngtcp2-linux.yml
```diff
@@ -51,7 +51,7 @@ env:
   ngtcp2-version: v0.18.0
   nghttp2-version: v1.55.1
   gnutls-version: 3.8.0
-  wolfssl-version: v5.6.3-stable
+  wolfssl-version: master
```
bagder, line 54: Does this make the caching weird if we cache on master?

icing: > Does this make the caching weird if we cache on master?

Indeed, need to take that out. Thanks for noticing.

PR #11778, opened 2023-09-01
configure: trust pkg-config when it's used for zlib

dfandrich:
The library flags retrieved from pkg-config were later thrown out and
harded-coded, which negates the whole reason to use pkg-config.
Also, previously, the assumption was made that --libs-only-l and
--libs-only-L are the decomposition of --libs, which is untrue and would
not allow linking against a static zlib. The new approach is slightly
better in that it uses --libs, although only if --libs-only-l returns
nothing.

Reported-by: Randall
Closes #11778

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1351,9 +1351,14 @@ else
     CURL_CHECK_PKGCONFIG(zlib)
 
     if test "$PKGCONFIG" != "no" ; then
-      LIBS="`$PKGCONFIG --libs-only-l zlib` $LIBS"
-      LDFLAGS="$LDFLAGS `$PKGCONFIG --libs-only-L zlib`"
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Perhaps add this to the commit message:

Bug: https://curl.se/mail/lib-2023-08/0081.html

rsbeckerca:
> Perhaps add this to the commit message:
> 
> Bug: https://curl.se/mail/lib-2023-08/0081.html

FYI: You can use my GitHub id: `Randall S. Becker <randall.becker@nexbridge.ca>` for references like `Reported-by:`

dfandrich:
Too late for this one…

PR #11781, opened 2023-09-01
Ci alpine 32bit test

icing:


dfandrich:
FWIW, you can build 32-bit binaries in a regular Ubuntu container without the entire environment having to be 32-bit by installing the gcc-10-i686-linux-gnu package.

icing:
> FWIW, you can build 32-bit binaries in a regular Ubuntu container without the entire environment having to be 32-bit by installing the gcc-10-i686-linux-gnu package

Thanks, I try that locally on my debian.

dfandrich:
I'm working on a 32-bit Ubuntu cross build in #11799

icing:
Closed, abandoned this approach.

PR #11806, opened 2023-09-06
http_aws_sigv4: canonicalize the query

### LINKED ISSUE #11794: aws-sigv4 request signature does not match expected value
### I did this

I'm trying to sign `GET` requests to the SPARQL endpoint of an Amazon Neptune cluster where query string parameters are used to pass information. If the parameter values contain spaces, forward slashes `/`, or colons `:`, then the signature calculated by curl does not match what the server expects.

An example request:

```sh
curl "https://${NEPTUNE_HOSTNAME}:8182/sparql/gsp" \
  --url-query "graph=urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66" \
  --aws-sigv4 "aws:amz:${AWS_REGION}:neptune-db" \
  --user "${AWS_ACCESS_KEY_ID}:${AWS_SECRET_ACCESS_KEY}" \
  --header "x-amz-security-token: ${AWS_SESSION_TOKEN}"
```

Gets a response like this:

```json
{
  "code": "AccessDeniedException",
  "requestId": "96c53027-1a0c-96a1-5a8b-18a20bee9465",
  "detailedMessage": "The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details."
}
```

### I expected the following

The calculated signature should be correct.

### curl/libcurl version

curl 8.2.1 (x86_64-w64-mingw32) libcurl/8.2.1 OpenSSL/3.1.2 (Schannel) zlib/1.3 brotli/1.1.0 zstd/1.5.5 WinIDN libssh2/1.11.0 nghttp2/1.55.1 ngtcp2/0.19.0 nghttp3/0.15.0
Release-Date: 2023-07-26

### operating system

Windows 64-bit

bagder:
Percent encoding needs to be done using uppercase, and most non-alphanumerical must be percent-encoded.

Fixes #11794
Reported-by: John Walker

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -44,16 +44,16 @@
 
 #include "slist.h"
 
-#define HMAC_SHA256(k, kl, d, dl, o)        \
-  do {                                      \
-    ret = Curl_hmacit(Curl_HMAC_SHA256,     \
-                      (unsigned char *)k,   \
-                      kl,                   \
-                      (unsigned char *)d,   \
-                      dl, o);               \
-    if(ret) {                               \
-      goto fail;                            \
-    }                                       \
+#define HMAC_SHA256(k, kl, d, dl, o)           \
+  do {                                         \
+    result = Curl_hmacit(Curl_HMAC_SHA256,     \
+                         (unsigned char *)k,   \
+                         kl,                   \
+                         (unsigned char *)d,   \
+                         dl, o);               \
+    if(result) {                               \
+      goto fail;                               \
+    }                                          \
   } while(0)
 
 #define TIMESTAMP_SIZE 17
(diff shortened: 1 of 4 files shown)


outscale-mgo (commented):
The path need to be make canon too.
Also, the query string should be sort by query parameter name.

I still think there should be a way to disable this, as the "spec" how to canonize the query and path seems to change:
in 2021 it requires a double encoding for all APIs but s3: https://web.archive.org/web/20211022045219/https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
Now the documentation seems to be the same for all APIs: https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html

bagder:
> The path need to be make canon too.

Sure, but we can do that in a separate PR. This for the query.

> Also, the query string should be sort by query parameter name.

Added now

> I still think there should be a way to disable this

How would a user/anyone know when to enable/disable? It seems this authentication method is not really designed to be used...

outscale-mgo:
> How would a user/anyone know when to enable/disable? It seems this authentication method is not really designed to be used...

Could an environment variable do the trick ?, like `CURL_DISABLE_URL_ENCODING_SIGV4` ?

lib/http_aws_sigv4.c
```diff
@@ -369,9 +369,107 @@ static CURLcode calc_s3_payload_hash(struct Curl_easy *data,
...
+          }
+          else
+            /* '%' without a following two-digit hex, encode it */
+            result = Curl_dyn_addn(dq, "%25", 3);
+          break;
+        default: {
+          /* URL encode */
+          const char hex[] = "0123456789ABCDEF";
+          char out[3]={'%'};
+          out[1] = hex[*q>>4];
```
jay, line 453:
```suggestion
          out[1] = hex[((unsigned char)*q)>>4];
```

bagder:
> Could an environment variable do the trick ?, like CURL_DISABLE_URL_ENCODING_SIGV4 ?

I don't see how an environment variable helps much. How would a user know when to use that? When will you advice a user to set it?

bagder:
I added test 439 in an attempt to verify this feature, but the `Signature` part of the `Authorization`  header does not seem to be stable so the test fails every time. I haven't figured out what makes it so. This is debug-build only so it runs with a fake/set time in order to get a stable output.

jay:
I do not like environment variables in place of configuration options and also I'm not convinced it should be an option.

bagder:
> I haven't figured out what makes it so

Ah, the port number for the host!

outscale-mgo:
> > Could an environment variable do the trick ?, like CURL_DISABLE_URL_ENCODING_SIGV4 ?
> 
> I don't see how an environment variable helps much. How would a user know when to use that? When will you advice a user to set it?

My idea was to have a way to deactivate URL encoding, that would be documented as an option to enable users to encode the URL themselves, if encoding doesn't work.
It could be with an option or environment variable.

Anyway, I did think about it a little more, and it seems most implementation did follow what is presently in the documentation (and done in this PR), so maybe we should force the URL encoding as suggested by @bagde and @jay and think about a way to deactivate it, if, and once an issue is raise.

bagder:
> think about a way to deactivate it, if, and once an issue is raise.

I think this is a sensible approach. If we need to provide an option for this or otherwise act differently in some situation, let's deal with that if and when someone has such a case.

bagder:
The test 439 still fails on Windows and mac for some reason. I'm investigating.

PR #11811, opened 2023-09-07
libcurl-env.3: add CURL_DBG_ envvars

jay:
- Document debug variables CURL_DBG_SOCK_WBLOCK, CURL_DBG_SOCK_WPARTIAL and CURL_DBG_QUIC_WBLOCK.

Follow-up to 331b89a3 which added those variables.

Closes #xxxx

### DIFF
--- a/.github/scripts/spellcheck.words
+++ b/.github/scripts/spellcheck.words
@@ -14,6 +14,7 @@ allocator
 alnum
 ALPN
 Altera
+AltSvc
 ALTSVC
 amiga
 AmigaOS
@@ -139,6 +140,7 @@ cURL
 CURLcode
 CURLE
 CURLH
+CURLINFO
 curlimages
 curlrc
 curltest
@@ -151,7 +153,9 @@ cyassl
 Cygwin
 daniel
 datatracker
+dbg
 Debian
+DEBUGBUILD
 decrypt
 deepcode
 DELE
--- a/docs/libcurl/Makefile.inc
+++ b/docs/libcurl/Makefile.inc
@@ -111,6 +111,7 @@ man_MANS = \
  curl_ws_send.3 \
  libcurl-easy.3 \
  libcurl-env.3 \
+ libcurl-env-dbg.3 \
  libcurl-errors.3 \
  libcurl-multi.3 \
  libcurl-security.3 \
--- a/docs/libcurl/libcurl-env-dbg.3
+++ b/docs/libcurl/libcurl-env-dbg.3
@@ -94,4 +94,4 @@ OpenLDAP tracing is enabled if this variable exists and its value is 1 or
 greater. There's a number of debug levels, refer to \fIopenldap.c\fP comments.
 .RE
 .SH "SEE ALSO"
-.BR libcurl-env (3),
+.BR libcurl-env (3)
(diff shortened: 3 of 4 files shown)


bagder:
I think these should be documented somewhere in `tests/` instead, since they are for debug/development/testing only. Then we can document *all* the debug environment variables we have in that document.

jay:
> Then we can document _all_ the debug environment variables we have in that document.

This document already has a specific section for debug variables. IMO it's easier to keep all of them in the same document. That way we can refer to just https://curl.se/libcurl/c/libcurl-env.html

bagder:
okay, let's do that. We need to figure out the failed spell CI job though.

bagder (approved):

jay:
> IMO it's easier to keep all of them in the same document. That way we can refer to just https://curl.se/libcurl/c/libcurl-env.html

I've since documented more debug variables and split off into a separate document libcurl-env-dbg.3.

jay:
> We need to figure out the failed spell CI job though.

I added `AltSvc`, `CURLINFO` and `DEBUGBUILD` to spellcheck.words.

I don't know what to do about `dbg`, `DBG` and `WPARTIAL`.

`dbg` is seen only in the title. I could add it to spellcheck.words but it could also be considered a misspelling of debug.
~~~
.TH libcurl-env-dbg 3 "18 September 2023" "libcurl" "libcurl"
.SH NAME
libcurl-env-dbg \- environment variables libcurl DEBUGBUILD understands
~~~

`DBG` & `WPARTIAL` are seen only here:
~~~
.IP "CURL_DBG_SOCK_WPARTIAL"
Percentage of data that shall be written to the network. For TCP/UNIX sockets.
This is applied to every send. Example: CURL_DBG_SOCK_WPARTIAL=80 means a send
with 1000 bytes would only send 800.
~~~
I think I can ignore CURL_DBG_ vars by putting a regex in cleanspell? I'm not sure why only this CURL_DBG_ triggers the spellcheck but I'm guessing it's the example.

bagder:
You can use the fact that the spell checker ignores emphasized words:

    Example: CURL_DBG_SOCK_WPARTIAL=80

Could be made as

    Example: \fBCURL_DBG_SOCK_WPARTIAL\fP=80

... which usually helps making the man page clearer as well!

jay:
> \fBCURL_DBG_SOCK_WPARTIAL\fP=80

Thanks. I did it like \fICURL_DBG_SOCK_WPARTIAL=80\fP because I think changing the color looks better than bolding it (in my msys terminal that just makes it brighter). If you think it makes more sense as bold I'll change it.

Capture

bagder:
I'm fine with whichever, I can't say I am consistent when I use them...

jay:
@icing I added the two variables from #12035, please confirm

> CURL_DBG_SOCK_RBLOCK
> 
> The percentage of recv() calls that should be answered with a EAGAIN at random. For TCP/UNIX sockets.
> 
> CURL_DBG_SOCK_RMAX
> 
> The maximum data that shall be received from the network in one recv() call. For TCP/UNIX sockets. This is applied to every recv.
> 
> Example: CURL_DBG_SOCK_RMAX=400 means recv buffer size is limited to a maximum of 400 bytes.

icing:
> @icing I added the two variables from #12035, please confirm
> 
> > CURL_DBG_SOCK_RBLOCK
> > The percentage of recv() calls that should be answered with a EAGAIN at random. For TCP/UNIX sockets.
> > CURL_DBG_SOCK_RMAX
> > The maximum data that shall be received from the network in one recv() call. For TCP/UNIX sockets. This is applied to every recv.
> > Example: CURL_DBG_SOCK_RMAX=400 means recv buffer size is limited to a maximum of 400 bytes.

Thanks!

icing (approved):

Issue #11810, opened 2023-09-07
Curl does not limit output data size


ElliotKillick:
### I did this

Consider the following: `test="$(curl https://example.com)"`

`example.com` could cause curl to download a very large amount of data which could lead to an out of memory scenario when put into the shell variable `test`.

A similar DoS could happen if writing to the disk by filling up disk space to its entirety.

Both of these could completely lock up a system and cause other applications to abort (or crash if they're written poorly). For memory exhaustion, a system with OOM reaper may also eventually take action.

### I expected the following

Considering this is such a basic security issue, curl should have a built-in option to limit its output size from `stdout`/`-o`/`--output`. Once the hard limit for output size has been reached, curl should stop reading from the network and terminate with curl error `23` or probably better would be a new error code to indicate this exact error.

I propose one of `--max-size` (similar to `--max-time`), `--max-output`, or `--max-out-size` (similar to `--max-filesize` which doesn't serve as a hard limit) to stay aligned with option names already existing in curl.

### curl/libcurl version

curl 8.0.1

### operating system

Fedora Linux 38

dfandrich:
The Unix way is to pipe the output of curl into head or dd or some other
program to limit the maximum amount of data if that is an issue in your
application. --max-filesize will prevent a too-long download from even
starting, which is an optimization that a pipe can't (easily) do, but it's easy
to limit the data transferred without adding the 258th command-line option to
curl.

ElliotKillick:
`--max-filesize` won't work as a hard (security) limit due to the reason given in the curl manual `NOTE` for this option. Changing this option to also enforce a hard limit would be an acceptable solution though...

For Unix systems, a `head -c` will mostly work. Although because it was [only recently added to POSIX](https://austingroupbugs.net/view.php?id=407), not all Unix systems have this yet, for example, [OpenBSD](https://man.openbsd.org/head.1). `dd` would definitely work on all Unix systems in practice, however, due to how `dd` works it may come with a performance hit (not to mention it's not pretty).

I've indeed tried piping into `head -c`, however, in POSIX sh it can't easily be done when you also want to check curl's error code. This is due to the fact that `head` (the last command run in the pipeline) will conceal curl's error code. `set -o pipefail` can be used to propagate curl's error code but again, [only recently added to POSIX](https://www.austingroupbugs.net/view.php?id=789) and popular shells like Dash have yet to support it.

[...]

PR #11813, opened 2023-09-07
tool_cb_wrt: error on max filesize exceeded

jay:
- Return CURLE_FILESIZE_EXCEEDED when the received data exceeds the maximum file size.

This is to handle those cases where the filesize is not known in advance (eg chunked encoding) but the bytes received exceeds the max file size.

Prior to this change --max-filesize had no effect when the file size was not provided by the server.

Reported-by: Elliot Killick

Fixes https://github.com/curl/curl/issues/11810
Closes #xxxx

---

draft for how i think we should do it. if there are no objections i'll add a test and finalize


bagder:
It now struck me that this (too-side) implementation might be a bit too simplistic. Since we have `CURLOPT_MAXFILESIZE` we should probably have that work in the same way? And if so, the check should probably be moved to the library.

bagder:
I did #11820 for that, which replaces this PR.

jay:
Closing in favor of #11820

PR #11820, opened 2023-09-08
let the max filesize option stop too big transfers too

### LINKED ISSUE #11810: Curl does not limit output data size

bagder:
Fixes #11810

I believe the wording and functionality of the option allows us to do this minor tweak of functionality without causing users harm. I still mark it to get done in the feature window for extra precaution.

### DIFF
--- a/lib/progress.c
+++ b/lib/progress.c
@@ -317,9 +317,16 @@ timediff_t Curl_pgrsLimitWaitTime(curl_off_t cursize,
 /*
  * Set the number of downloaded bytes so far.
  */
-void Curl_pgrsSetDownloadCounter(struct Curl_easy *data, curl_off_t size)
+CURLcode Curl_pgrsSetDownloadCounter(struct Curl_easy *data, curl_off_t size)
 {
+  if(data->set.max_filesize && (size > data->set.max_filesize)) {
+    failf(data, "Exceeded the maximum allowed file size "
+          "(%" CURL_FORMAT_CURL_OFF_T ")",
+          data->set.max_filesize);
+    return CURLE_FILESIZE_EXCEEDED;
+  }
   data->progress.downloaded = size;
+  return CURLE_OK;
 }
 
 /*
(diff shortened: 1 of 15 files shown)


lib/progress.c
```diff
@@ -317,9 +317,15 @@ timediff_t Curl_pgrsLimitWaitTime(curl_off_t cursize,
 /*
  * Set the number of downloaded bytes so far.
  */
-void Curl_pgrsSetDownloadCounter(struct Curl_easy *data, curl_off_t size)
+CURLcode Curl_pgrsSetDownloadCounter(struct Curl_easy *data, curl_off_t size)
 {
+  if(data->set.max_filesize && (size > data->set.max_filesize)) {
+    failf(data, "Exceeded the maximum allowed file size (%zu)",
+          data->set.max_filesize);
```
jay, line 324:
```suggestion
    failf(data, "Exceeded the maximum allowed file size "
          "(%" CURL_FORMAT_CURL_OFF_T ")",
          data->set.max_filesize);
```

PR #11824, opened 2023-09-09
CI: run pytest with the -v option

dfandrich:
This lists of the test cases being run so it can be tracked over time.

Closes #11824

This lack can be currently seen as "truncated" test runs on my test summary page.


icing (approved):
Thought that shorter would be nice, but if -v works better for you, it's all fine.

PR #11825, opened 2023-09-10
configure: fix `HAVE_TIME_T_UNSIGNED` check

benoit-pierre:
The syntax was incorrect (need a proper main body), and the test condition was wrong (resulting in a signed `time_t` detected as unsigned).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3569,8 +3569,10 @@ CURL_RUN_IFELSE(
   [
   #include <time.h>
   #include <limits.h>
-  time_t t = -1;
-  return (t > 0);
+  int main(void) {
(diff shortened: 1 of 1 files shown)


bagder:
It is curious why it would fail in that direction for you. For me it failed and detects it as *not* unsigned...

bagder (approved):

bagder:
Thanks!

benoit-pierre:
What I meant is there were 2 errors: a syntax error, preventing the check of ever detecting an unsigned `time_t`, and an error in the test condition (`t > 0`), incorrectly flipping the result on my machine once the syntax issue was solved.

PR #11834, opened 2023-09-11
tests: improve reliability of some tests

dfandrich:
Also, tweak some test timings & metadata.


dfandrich:
This has turned into a bit of a grab bag of test improvements, but I think it's ready to submit.

PR #11835, opened 2023-09-11
Add --decode-remote-name option to url-decode incoming filenames

gamache:
This PR implements a command line option called `--decode-remote-name`, to URL-decode all incoming filenames when saving files with `-J, --remote-header-name` and/or `-O, --remote-name`.

For example,
* `curl -O http://example.com/some%20file.txt` leaves a file named `some%20file.txt` in your current directory, while
* `curl -O --decode-remote-name http://example.com/some%20file.txt` leaves a file named `some file.txt`.

The immediate inspiration for this feature came from the comments on a [recent Curl blog post](https://daniel.haxx.se/blog/2023/09/04/the-curl-wget-venn-diagram/), but I have also wanted it myself a few times over the years.

`curl -O --decode-remote-name` will URL-decode the filename as pulled from the URL.

`curl -OJ --decode-remote-name` will attempt to use and decode the `Content-Disposition` header's `filename*` field before falling back to the undecoded `filename` field (whose format does not support URL-encoding) then the decoded URL.

I added some tests to exercise the feature, namely:
```
test 3210...[-J and --decode-remote-name falls back to decoded filename from URL]
test 3211...[-J and --decode-remote-name decodes the Content-Disposition filename* (quoted)]
test 3212...[-J and --decode-remote-name decodes the Content-Disposition filename* (unquoted)]
test 3213...[-J and --decode-remote-name falls back to undecoded filename if filename* not present (quoted)]
test 3214...[-J and --decode-remote-name falls back to undecoded filename if filename* not present (unquoted)]
test 3215...[-J and --decode-remote-name removes pathname from filename* (literal \, quoted)]
test 3216...[-J and --decode-remote-name removes pathname from filename* (literal \, unquoted)]
test 3217...[-J and --decode-remote-name removes pathname from filename* (literal /, quoted)]
test 3218...[-J and --decode-remote-name removes pathname from filename* (literal /, unquoted)]
test 3219...[-J and --decode-remote-name removes pathname from filename* (encoded \, quoted)]
test 3220...[-J and --decode-remote-name removes pathname from filename* (encoded \, unquoted)]
test 3221...[-J and --decode-remote-name removes pathname from filename* (encoded /, quoted)]
test 3222...[-J and --decode-remote-name removes pathname from filename* (encoded /, unquoted)]
test 3223...[-O and --decode-remote-name decodes filename from URL]
test 3224...[-O and --decode-remote-name removes pathname (literal \)]
test 3225...[-O and --decode-remote-name removes pathname (encoded \)]
test 3226...[-O and --decode-remote-name removes pathname (literal /)]
test 3227...[-O and --decode-remote-name removes pathname (encoded /)]
```

No known bugs, nor changes to existing behavior.

Known limitations or quirks:
* `-OJ --decode-remote-name` does not do anything with the text encoding information present in the `Content-Disposition` header's `filename*` field; it simply copies bytes.
* The new `get_cd_field` function retains legacy quote-handling behavior, as demonstrated in `tests/data/test1312` and `tests/data/1313`. This behavior accepts input that does not adhere to [RFC 6266](https://datatracker.ietf.org/doc/html/rfc6266), such as broken quoting.

Thanks for any help!


bagder:
Does this not make an assumption that the file system uses UTF-8 for file names? Windows for example, does not, so will the decoding then not make the file names wrong there? 

Some source say that NTFS (on windows at least) uses UTF-16. For FAT file systems the situation seems to be even more confused. Both of these are also supported on Linux, but how does it it encode the file names then?

gamache:
I had to do some digging on this. Here is what I learned:

On Windows, there are two APIs for filenames: one based on `char *` that expects bytes in the local encoding (most frequently, ISO-8859-1), and another based on `wchar_t *` which expects UTF-16 codepoints. If Curl was built with `ENABLE_UNICODE=yes`, it uses the UTF-16 API, otherwise it uses the `char *` API.

This has several effects:
* Codepoints above U+FFFF are not representable in Windows filenames under any circumstances.
* Codepoints above U+00FF are not representable in Windows filenames unless `ENABLE_UNICODE=yes` was specified at build time.

But given these constraints, it's straightforward to convert the contents of the `filename*` field to whichever encoding is in use, so I am preparing a commit that performs this conversion.

gamache:
Translating into and out of UTF is too much of a rathole for a feature that is mostly there to allow spaces in filenames. :) I am going to close this PR and create another with a much more limited scope.

PR #11836, opened 2023-09-12
quic: set ciphers/curves the same way regular TLS does

### LINKED ISSUE #11796: Unable to set ciphers for quic traffic
### I did this

i tried to run below command with chacha20 ciphers

curl -v -# --http3-only -k -o /tmp/BPS.pdf https://127.0.0.1:8443/BPS.pdf --ciphers TLS_CHACHA20_POLY1305_SHA256

and file also downloaded sucessfully but nginx logs is not showing chacha20 as ciphers used

cat /var/log/nginx/access.log
127.0.0.1 - - [04/Sep/2023:17:50:14 +0000] "GET /BPS.pdf HTTP/3.0" 200 30182355 "-" "curl/8.3.0-DEV"
127.0.0.1 - - [04/Sep/2023:17:50:14 +0000] "HTTP/3.0" "GET /BPS.pdf HTTP/3.0" 200 30182355 "-" "curl/8.3.0-DEV" "h3" "-""GET https://127.0.0.1/BPS.pdf 9948104b97890658c6b4ccba6908c68d 177287 1693849814.723 0.167 - - - "/var/www/html/BPS.pdf" OKTLSv1.3/TLS_AES_128_GCM_SHA256

### I expected the following

i expect curl to use chacha20 ciphers

### curl/libcurl version

curl -V
curl 8.3.0-DEV (x86_64-pc-linux-gnu) libcurl/8.3.0-DEV OpenSSL/3.0.9 zlib/1.2.11 brotli/1.0.9 nghttp2/1.56.0-DEV ngtcp2/0.18.0 nghttp3/0.14.0
Release-Date: [unreleased]
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTP3 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets

### operating system

uname -a
Linux ubuntu 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

am i missing anything here cause with ngtcp2 and aioquic,i am able to see chacha20 ciphers in nginx logs

cat /var/log/nginx/access.log
127.0.0.1 - - [04/Sep/2023:17:45:26 +0000] "GET / HTTP/3.0" 200 10701 "-" "aioquic/0.9.21"
127.0.0.1 - - [04/Sep/2023:17:45:26 +0000] "HTTP/3.0" "GET / HTTP/3.0" 200 10701 "-" "aioquic/0.9.21" "h3" "-""GET https://localhost/ 3e68fabe6168c1cf6986e8f210b61280 176074 1693849526.229 0.000 - - - "/var/www/html/index.html" OKTLSv1.3/TLS_CHACHA20_POLY1305_SHA256

bagder:
for OpenSSL/BoringSSL/wolfSSL

Fixes #11796
Reported-by: Karthikdasari0423 on github

### DIFF
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -149,8 +149,8 @@ static CURLcode quic_x509_store_setup(struct Curl_cfilter *cf,
         SSL_CTX_set_verify(ctx->sslctx, SSL_VERIFY_PEER, NULL);
         /* tell OpenSSL where to find CA certificates that are used to verify
            the server's certificate. */
-        if(!SSL_CTX_load_verify_locations(
-              ctx->sslctx, ssl_cafile, ssl_capath)) {
+        if(!SSL_CTX_load_verify_locations(ctx->sslctx, ssl_cafile,
+                                          ssl_capath)) {
           /* Fail if we insist on successfully verifying the server. */
           failf(data, "error setting certificate verify locations:"
                 "  CAfile: %s CApath: %s",
@@ -213,6 +213,15 @@ static CURLcode quic_ssl_setup(struct Curl_cfilter *cf, struct Curl_easy *data)
       SSL_free(ctx->ssl);
       ctx->ssl = NULL;
       return CURLE_QUIC_CONNECT_ERROR;
+    }
+  }
+
+  {
+    struct connectdata *conn = data->conn;
+    const char *curves = conn->ssl_config.curves;
+    if(curves && !SSL_CTX_set1_curves_list(ctx->sslctx, curves)) {
+      failf(data, "failed setting curves list for QUIC: '%s'", curves);
+      return CURLE_SSL_CIPHER;
     }
   }
 
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -178,6 +178,8 @@ static CURLcode quic_ssl_setup(struct Curl_cfilter *cf, struct Curl_easy *data)
 {
   struct cf_quiche_ctx *ctx = cf->ctx;
   unsigned char checkip[16];
+  struct connectdata *conn = data->conn;
+  const char *curves = conn->ssl_config.curves;
 
   DEBUGASSERT(!ctx->sslctx);
   ctx->sslctx = SSL_CTX_new(TLS_method());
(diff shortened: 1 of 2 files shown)


lib/vquic/curl_quiche.c
```diff
@@ -216,6 +216,15 @@ static CURLcode quic_ssl_setup(struct Curl_cfilter *cf, struct Curl_easy *data)
...
   }
 
+  {
+    struct connectdata *conn = data->conn;
+    const char *curves = conn->ssl_config.curves;
+    if(curves && !SSL_CTX_set1_curves_list(ctx->sslctx, curves)) {
+      failf(data, "failed setting curves list for QUIC: '%s'", curves);
+      return CURLE_SSL_CIPHER;
+    }
+  }
```
jay, line 226: This has to be set higher up before the SSL object is created from the SSL_CTX. In other words above SSL_new. There is no similar code for ciphers here, is that intentional? If both these backends need similar CTX init or SSL init (eg same SNI code is now in both) maybe it would be better to have a common function rather than duplicate the code

Also I don't have quiche but looking at this it does not appear to free SSL_CTX and SSL in cf_quiche_ctx_clea? not sure if that's a bug or not but ngtcp2 clear there's similar code that does that.

bagder: > This has to be set higher up before the SSL object is created from the SSL_CTX. In other words above SSL_new

That sounds wrong. This functions uses the sslctx as argument so it needs to be done after that is created, which is done within this same function.

> There is no similar code for ciphers here, is that intentional?

Yes, this is using BoringSSL and I don't know how those are set other than this.

> looking at this it does not appear to free SSL_CTX and SSL in cf_quiche_ctx_clear

True, but that seems unrelated to this PR.

jay: > > This has to be set higher up before the SSL object is created from the SSL_CTX. In other words above SSL_new
> 
> That sounds wrong. This function uses the sslctx as argument so it needs to be done after that is created, which is done within this same function.

When an SSL object is created it points to the CTX used to create it, not a copy. In other words:

~~~c
ctx = SSL_CTX_new(TLS_client_method());
ssl = SSL_new(ctx);
assert(ctx == SSL_get_SSL_CTX(ssl));
~~~

The OpenSSL documentation [says](https://www.openssl.org/docs/man3.0/man3/SSL_CTX_new.html):

_"**An SSL_CTX object should not be changed after it is used to create any SSL objects** or from multiple threads concurrently, since the implementation does not provide serialization of access for these cases."_

You are doing this:
~~~c
ctx->ssl = SSL_new(ctx->sslctx);
...
SSL_CTX_set1_curves_list(ctx->sslctx, curves);
~~~

It may appear to work but it's not guaranteed. I am saying do this instead:
~~~c
SSL_CTX_set1_curves_list(ctx->sslctx, curves);
...
ctx->ssl = SSL_new(ctx->sslctx);
~~~

bagder: aah now I see what you mean. Pushed an update

PR #11838, opened 2023-09-12
misc: better random strings

piru:
Generate alphanumerical random strings.

Prior this change curl used to create random hex strings. This was mostly okay, but having alphanumerical random strings is better: The strings have more entropy in the same space.

The MIME multipart boundary used to be mere 64-bits of randomness due to being 16 hex chars. With these changes the boundary is 22 alphanumerical chars, or little over 130 bits of randomness.

### DIFF
--- a/lib/rand.c
+++ b/lib/rand.c
@@ -24,6 +24,8 @@
 
 #include "curl_setup.h"
 
+#include <limits.h>
+
 #ifdef HAVE_FCNTL_H
 #include <fcntl.h>
 #endif
@@ -262,6 +264,39 @@ CURLcode Curl_rand_hex(struct Curl_easy *data, unsigned char *rnd,
     *rnd++ = hex[*bufp & 0x0F];
     bufp++;
     num -= 2;
+  }
+  *rnd = 0;
+
+  return result;
+}
+
+/*
+ * Curl_rand_alnum() fills the 'rnd' buffer with a given 'num' size with random
+ * alphanumerical chars PLUS a null-terminating byte.
+ */
+
+static const char alnum[26 + 26 + 10] =
+  "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+CURLcode Curl_rand_alnum(struct Curl_easy *data, unsigned char *rnd,
+                         size_t num)
+{
+  CURLcode result = CURLE_OK;
+  const int alnumspace = sizeof(alnum);
+  unsigned int r;
+  DEBUGASSERT(num > 1);
+
+  num--; /* save one for null-termination */
+
+  while(num) {
+    do {
+      result = randit(data, &r);
+      if(result)
+        return result;
+    } while(r >= (UINT_MAX - UINT_MAX % alnumspace));
+
+    *rnd++ = alnum[r % alnumspace];
+    num--;
   }
   *rnd = 0;
 
--- a/lib/fopen.c
+++ b/lib/fopen.c
@@ -64,7 +64,7 @@ CURLcode Curl_fopen(struct Curl_easy *data, const char *filename,
   fclose(*fh);
   *fh = NULL;
 
-  result = Curl_rand_hex(data, randsuffix, sizeof(randsuffix));
+  result = Curl_rand_alnum(data, randsuffix, sizeof(randsuffix));
   if(result)
     goto fail;
 
--- a/lib/mime.c
+++ b/lib/mime.c
@@ -1289,9 +1289,9 @@ curl_mime *curl_mime_init(struct Curl_easy *easy)
     mime->lastpart = NULL;
 
     memset(mime->boundary, '-', MIME_BOUNDARY_DASHES);
-    if(Curl_rand_hex(easy,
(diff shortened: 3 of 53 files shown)


lib/rand.c
```diff
@@ -267,3 +269,36 @@ CURLcode Curl_rand_hex(struct Curl_easy *data, unsigned char *rnd,
...
+/*
+ * Curl_rand_alnum() fills the 'rnd' buffer with a given 'num' size with random
+ * alphanumerical chars PLUS a null-terminating byte.
+ */
+
+CURLcode Curl_rand_alnum(struct Curl_easy *data, unsigned char *rnd,
+                         size_t num)
+{
+  CURLcode result = CURLE_OK;
+  const char alnum[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
dfandrich, line 282: A microoptimization: use the same characters as base64encdec so the linker can save space and eliminate one copy. Assuming the extra three characters are allowed, of course. Also, this could be static const.

jay:
The restrictions may have been for legacy server compatibility? @monnerat

lib/rand.c
piru: @dfandrich While the extra chars might work in some context, there is no knowing where this function might end up being used. To remain useful I'd prefer sticking to plain alnum. I'll make it array that is not 0 terminated to fix this. Then the compiled can merge the arrays since the prefix is the same.

dfandrich:
Don't bother to spend much time on that. This is definitely micro.

piru:
Regarding the legacy server compatibility aspect: I did not make Digest Auth use alnum, even though `cnonce` could be made to use it as well. RFC 7616 https://datatracker.ietf.org/doc/html/rfc7616 says `The cnonce value is an opaque quoted ASCII-only string`. All examples just use hex string there, probably due to symmetry. Due to this I'm sure there might be some broken implementations that would b0rk on actual non-hex `cnonce`, and thus it likely is wise to avoid using one there.

For other things using alnum should be safe, I think.

monnerat:
> The restrictions may have been for legacy server compatibility?

No, I did not have such limitation in mind when I wrote mime: I used this procedure because it was available and already used for that purpose in the legacy form API code: https://github.com/curl/curl/blob/5bae72734b45a01c6337eb3b2c40020c4e904415/lib/formdata.c#L1555-L1567
In fact, the randomness is not very important in mime: there is no cryptographic and/or secrecy requirements and the only things we need are:
- No 2 identical nested boundaries,
- Not matching a real data line.

Both are already unlikely to happen.

piru:
No MIME RFC spells it out explicitly, but having very low entropy or even predictable boundary would lead to security impact where the attacker could inject arbitrary multiparts via crafted message contents. This of course wasn't the case here, 64-bit of strong randomness is already very difficult to target in any meaningful way.

bagder:
My recollection of the layout of the boundary separator:

It started out without the long sequence of dashes but was modified to look like this early on (20+ years ago) due to someone having problems with the first approach because - as always - users who wrote server side scripts (PHP?) did lots of assumptions about how to parse and detect that boundary based on how the boundaries that were used by the popular browsers at the time used.

We then adapted to this version that looks much more like the browsers' back then.

I don't know if this can still be a problem. I don't know if this change actually makes this *not* look like a browser separator. I don't know if the browsers perhaps have even modified their separators since then.

piru:
Some quick research on the separator browsers use these days:

- Chromium about charset to use: https://github.com/chromium/chromium/blob/466969ca58ae8686107d4a971c7918862158d5bc/net/base/mime_util.cc#L799
- Actual Chromium boundary generation: https://github.com/chromium/chromium/blob/466969ca58ae8686107d4a971c7918862158d5bc/net/base/mime_util.cc#L874
- It seems Mozilla uses 3 x random numbers so bunch of digits:
https://github.com/mozilla/gecko-dev/blob/1cae9ca7cc7d17cfc92088dfc68e5ff391128caa/dom/html/HTMLFormSubmission.cpp#L355

dfandrich:
This is causing a warning treated as an error on Windows VS2002 on [Appveyor](https://ci.appveyor.com/project/curlorg/curl/builds/48051881/job/strhd101u35cevji):
```
C:\projects\curl\lib\rand.c(278,19): error C2220: the following warning is treated as an error [C:\projects\curl\lib\libcurl_object.vcxproj]
C:\projects\curl\lib\rand.c(278,19): warning C4295: 'alnum': array is too small to include a terminating null character [C:\projects\curl\lib\libcurl_object.vcxproj]
```

bagder:
It's better to create a new issue for that, since this PR has already been merged it is unclear what it means to reopen it?

dfandrich:
The PR is not quite ready to submit? ☺

bagder:
> The PR is not quite ready to submit? ☺

But the commit was not reverted so this PR cannot be merged (again). The code that is merged needs a (different) fix, not this PR... It's not terribly important, I made #11870 to hopefully fix it.

piru:
Right. Let this right here be a warning for anyone trying to micro optimize. 😂

bagder:
Fixed

Issue #11767, opened 2023-08-30
AppVeyor CI msys2/cygwin debug jobs permafail for test2600 after #11690


vszakats:
### I did this

Pushed a PR. These AppVeyor CI jobs always fail for test2600:
- autotools, msys2, Debug, no Proxy, no SSL
- autotools, msys2, Debug, no SSL
- autotools, cygwin, Debug, no SSL

All 64-bit builds.
The MSYS2 _Release_ build is always OK however.
Release build has AsynchDNS enabled, while Debug ones don't.

Happening since:
https://ci.appveyor.com/project/curlorg/curl/builds/47840460

Original report: https://github.com/curl/curl/pull/11690#issuecomment-1687228076
Ref: 862741637 #11690

### I expected the following

AppVeyor CI jobs green. Or, green after a rebuild in case of flaky results.

### curl/libcurl version

master

### operating system

Windows MSYS2/Cygwin

jay:
I've looked at a few and afaics it's always the happy eyeballs subtests that fail.

~~~
 5: test case took 403ms
 unit2600.c:309 test failed: '5: expected ipv6 to start delayed after 150ms, instead first attempt made after 201ms'

...

 6: test case took 403ms
 unit2600.c:309 test failed: '6: expected ipv4 to start delayed after 150ms, instead first attempt made after 201ms'
~~~

The error message is misleading because USE_ALARM_TIMEOUT is defined in this case (which may be the issue?) and that means the happy eyeballs timeout is set to 1000ms (instead of he_timeout which is 150ms). So it's failing because it expects a delay of 1s and it's not getting it. As you noted in your original report these fails are all with synchronous dns on Windows.

https://github.com/curl/curl/blob/226d042a58a9ff9fedae44a9c962e7f8339207bf/tests/unit/unit2600.c#L329-L337

https://github.com/curl/curl/blob/226d042a58a9ff9fedae44a9c962e7f8339207bf/tests/unit/unit2600.c#L300-L310

/cc @icing

vszakats:
Closing due to no interest.

dfandrich:
There's definitely interest in dealing with failing tests. The problem is lack of developers with access to Windows. We can't have tests permanently failing.

dfandrich:
[...]

PR #11843, opened 2023-09-13
Fix the help text to the static build option in CMakeLists.txt

Mathias-Fuchs:
The second argument to the CMake "option" command is the help text and is displayed, for instance, as a tooltip next to the checkbox in the CMake GUI. When the user wants to build static libraries, they will have to check that checkbox. Accordingly, the help text should read "build static libraries", just the same way as the help text to the shared libraries option is "build shared libraries".


Mathias-Fuchs:
when changing a help text string results in unsuccessful tests ...

dfandrich:
Those three test jobs are permafailing right now.  #11767

vszakats (approved):

bagder:
Thanks!

PR #11847, opened 2023-09-14
quiche: fix build error with --with-ca-fallback

junhochoi:
Fix build error when curl is built with --with-quiche and --with-ca-fallback.


icing (approved):

jay:
Thanks

PR #11851, opened 2023-09-14
dfilters, a client writer stack per easy handle

icing:
`dfilters` are a chain of write filters kept at each curl easy handle (commonly named "data" internally). This chain handles the writing of body and meta data to the client application. 

### Goal

The eventual goal is to make response data fully handled by forwarding them to the dfilter chain. At least for body data, this seems very feasible. This will mean that connection filters like HTTP/2 and HTTP/3 will be able to directly forward incoming data without the need to copy them back and forth in the transfer loop.

### Design

As can be seen in `dfilters.h`, there is some similarity to connection filters. A dfilter writer has `do_init()` and `do_close()` callbacks for lifetime management. Data is written either via `do_meta()` or `do_body()`. There are methods `is_paused()` and `unpause()`, but no writer but `df-out` will really worry about those.

Protocol handlers may add their own writers to this chain as needed. The chain has the lifetime of a single request (e.g. `SingleRequest`, commonly `data->req`). At the end, the writers `do_close()` will be called, giving them the chance to de-allocate resources. protocol handlers do not need to care about this.

#### Phases

New in dfilters are the "phase" a writer is added to. The phase of writers determines in what order they are added to the chain. And therefore in what order they are called. The order of phases is

```
CONN --> TRANSCODE --> PROTOCOL --> DECODE --> APP
```

No matter what writers are already present in the chain, adding a `DECODE` writer will always make it the first being called in this phase. Adding a second `DECODE` will make that the first and the other one(s) being called thereafter.

This means, the order of adding writers matters only for the same phase. It matters not if `TRANSCODE` writers are added previous to `DECODE` ones or afterwards.

As an example, the HTTP protocol would use these as follows:

 * `TRANSCODE`: remove "chunked" encoding
 * `PROTOCOL`: collect HEADERs, check "Content-Length" sanity
 * `DECODE`: apply content decoders (gzip, brotli, etc.)
 
### content_encoding.*

The former `content_encoding.[ch]` source was already pretty close to what a dfilter writer is now. It was able to build a chain of decoders from a HTTP "Content-Encoding:" header. Response data was written through this chain to deflate/unzip/unbrotli the body.

This is now found in `df-http-enc.[ch]` as `DECODE`writers.

### Curl_client_write()

The implementation of `Curl_client_write()` in `sendf.c` was moved into the `cf-out.c` writer. The buffer for keeping data of a PAUSEd transfer are kept locally at the writer now. The handling of "unpausing" a transfer was moved from `easy.c` into the writer as well.

`Curl_client_write()` has been replaced with `Curl_client_write_body()` and `Curl_client_write_meta()` to have a more clear distinction between what is written. (FTP is the only one that writes a mix, being what it is.) These are now found in `dfilters.[ch]`.

### CRLF conversions

For FTP there was a special handling in `Curl_client_write()` that sometimes did conversions of CR+LF to just LF. This has been factored out into `df-crlf2lf.[ch]` as stand-alone `TRANSCODE` writer. That one is installed and controlled by the FTP protocol handler.

### HTTP header collection

Similar, the special code in `Curl_client_write()` that collected HTTP headers written via `Curl_headers_push()` was moved to `cf-http.[ch]` as `PROTOCOL` writer. Installed by the HTTP protocol handler and its WS friend.

### Request statistics and progress

The various places that update `data->req.bytecount` and call `Curl_pgrsSetDownloadCounter()` should be axed. A standard `df-req` `PROTOCOL` writer can do that more easily.

### HTTP chunked decoding

This should become its own dfilter `TRANSCODE` writer. This will replace the special handling in `transfer.c`. Together with a `df-req` writer, transfer will no longer worry about chunking or how many *real* response bytes have been written.


jay:
> The eventual goal is to make response data fully handled by forwarding them to the dfilter chain. At least for body data, this seems very feasible. This will mean that connection filters like HTTP/2 and HTTP/3 will be able to directly forward incoming data without the need to copy them back and forth in the transfer loop.

Though it may be slightly faster to not have to copy, this is a big structural change to make that happen and IMO looks more complex and not less.

icing:
> > The eventual goal is to make response data fully handled by forwarding them to the dfilter chain. At least for body data, this seems very feasible. This will mean that connection filters like HTTP/2 and HTTP/3 will be able to directly forward incoming data without the need to copy them back and forth in the transfer loop.
> 
> Though it may be slightly faster to not have to copy, this is a big structural change to make that happen and IMO looks more complex and not less.

I decided to introduce these changes in smaller steps, starting with #11885 to clarify the sematics of the `Curl_client_write()` arguments. That'll allow to make things a bit simpler.

I still believe the `dfilter` concept makes sense. It is an evolution of the `content_encoding.[ch]` code, where stacked filters have proven their worth already. There is lots to gain here which then no longer needs to be lifted in every protocol handler. Also, simplification of our transfer loop, the most complex curl part, is high on my list.

Closing this to be revisited in another PR.

Issue #6149, opened 2020-10-31
FTPS upload data loss with TLS 1.3


Bylon2:
### I did this
When uploading with FTP, the file is truncated towards the end although callbacks are called normally and curl_easy_perform exits with 0.

"read" callback have a default buffer of 64K in this libcurl version, and most of the time, the last 64K buffer of the file is missing.

That happens when using:
curl_easy_setopt(CURLOPT_USE_SSL, CURLUSESSL_ALL);

Doing the same ftp transfer (upload) with:
curl_easy_setopt(CURLOPT_USE_SSL, CURLUSESSL_CONTROL);

works perfectly fine.
(The 'control' part of the FTP connection is always encrypted so that the password is not shown in 'clear text')

### I expected the following
I expected both should work!

### curl/libcurl version
libcurl/7.68.0 (x86_64-pc-linux-gnu) OpenSSL/1.1.1f

### operating system
```
$ uname -a
Linux bylon-pc 5.4.0-52-generic #57-Ubuntu SMP Thu Oct 15 10:57:00 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

```

Since this is inside my fuse driver, I didn't narrow it down to a small program.
If the answer is that this is a "new" issue, I'll try to write the smallest possible code to show the issue.

_[Note]_ There might also be some "timing" issues involved. This test is part of my non regression test suite for my driver, and showed no error as long as I was using my ISP fiber-box (600Mbps upload). The box broke, and while I wait for a new one, I am now using a much slower backup Wifi connection that triggers this truncation.

**Suspecting:** the wifi connection (which is forwarded through my Raspberry Pi) is also not reliable. I am suspecting that at some point the SSL layer drops data and should be instructed to "retry". This is probably done as long as the file is uploading, but an "end flush" is not done. Last test dropped 196608 bytes at the end of the file (3 x 65536). The rest of the file is uploaded fine (diff compared on the uploaded part).

**Other tests and observations:**
In the test suite, I have other "write" operation just after the current one exhibiting the bug. Since the network is much slower, these operations start in parallel when writing the end of the file through FTP. I suspected there could be a non-safe OpenSSL parallel operation and added a long enough "sleep" after the first FTP upload. Uploads are not any more in parallel with the "sleep" but it did not fix the bug.

I also notice, running with Valgrind, that when I terminate the driver (fusermount -u since it is a fuse driver), I get:

```
==183296== Process terminating with default action of signal 13 (SIGPIPE)
[...]

PR #11858, opened 2023-09-15
SSL, invoke shutdown implementation when closing a connection

icing:
* fix http2 and h2-proxy filter to call the next filters on close
* make openssl's shutdown handle a one-time effort with added tracing for what it achieved

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -1876,15 +1876,45 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
 
   if(backend->handle) {
     if(cf->next && cf->next->connected) {
-      char buf[32];
+      char buf[1024];
+      int nread, err;
+      long sslerr;
+
       /* Maybe the server has already sent a close notify alert.
          Read it to avoid an RST on the TCP connection. */
       (void)SSL_read(backend->handle, buf, (int)sizeof(buf));
-
-      (void)SSL_shutdown(backend->handle);
+      ERR_clear_error();
+      if(SSL_shutdown(backend->handle) == 1) {
+        CURL_TRC_CF(data, cf, "SSL shutdown finished");
+      }
+      else {
+        nread = SSL_read(backend->handle, buf, (int)sizeof(buf));
+        err = SSL_get_error(backend->handle, nread);
+        switch(err) {
+        case SSL_ERROR_NONE: /* this is not an error */
+        case SSL_ERROR_ZERO_RETURN: /* no more data */
+          CURL_TRC_CF(data, cf, "SSL shutdown, EOF from server");
+          break;
+        case SSL_ERROR_WANT_READ:
+          /* SSL has send its notify and now wants to read the reply
+           * from the server. We are not really interested in that. */
+          CURL_TRC_CF(data, cf, "SSL shutdown sent");
+          break;
+        case SSL_ERROR_WANT_WRITE:
+          CURL_TRC_CF(data, cf, "SSL shutdown send blocked");
+          break;
+        default:
+          sslerr = ERR_get_error();
+          CURL_TRC_CF(data, cf, "SSL shutdown, error: '%s', errno %d",
+                      (sslerr ?
+                       ossl_strerror(sslerr, buf, sizeof(buf)) :
+                       SSL_ERROR_to_str(err)),
+                      SOCKERRNO);
+          break;
+        }
+      }
 
       ERR_clear_error();
-
       SSL_set_connect_state(backend->handle);
     }
 
--- a/lib/cf-h2-proxy.c
+++ b/lib/cf-h2-proxy.c
@@ -1199,6 +1199,8 @@ static void cf_h2_proxy_close(struct Curl_cfilter *cf, struct Curl_easy *data)
     cf_h2_proxy_ctx_clear(ctx);
     CF_DATA_RESTORE(cf, save);
   }
+  if(cf->next)
+    cf->next->cft->do_close(cf->next, data);
 }
 
 static void cf_h2_proxy_destroy(struct Curl_cfilter *cf,
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -2425,6 +2425,8 @@ static void cf_h2_close(struct Curl_cfilter *cf, struct Curl_easy *data)
     cf_h2_ctx_clear(ctx);
     CF_DATA_RESTORE(cf, save);
   }
+  if(cf->next)
+    cf->next->cft->do_close(cf->next, data);
 }
 
 static void cf_h2_destroy(struct Curl_cfilter *cf, struct Curl_easy *data)
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -1494,7 +1494,8 @@ static void ssl_cf_close(struct Curl_cfilter *cf,
 
   CF_DATA_SAVE(save, cf, data);
   cf_close(cf, data);
-  cf->next->cft->do_close(cf->next, data);
+  if(cf->next)
+    cf->next->cft->do_close(cf->next, data);
   CF_DATA_RESTORE(cf, save);
 }
 


icing:
Should be merged, imo.

jay:
While it may be necessary to wait for a shutdown in some cases, I don't like this fix because it's blocking (for 1+ seconds depending on time granularity) and not specific. Please see my comments [here](https://github.com/curl/curl/issues/6149#issuecomment-795080891).

icing:
> While it may be necessary to wait for a shutdown in some cases, I don't like this fix because it's blocking (for 1+ seconds depending on time granularity) and not specific. Please see my comments [here](https://github.com/curl/curl/issues/6149#issuecomment-795080891).

I agree that possibly blocking for a second on each connection individually is not good. I opened a discussion at #11878.

icing:
Amended PR with a simpler version, no delays.

bagder (approved):

lib/vtls/openssl.c
```diff
@@ -1878,15 +1878,42 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+      char buf[1024];
+      int nread, err;
+      long sslerr;
 
       ERR_clear_error();
+      if(SSL_shutdown(backend->handle)) {
+        CURL_TRC_CF(data, cf, "SSL shutdown finished");
+      }
+      else {
+        nread = SSL_read(backend->handle, buf, (int)sizeof(buf));
```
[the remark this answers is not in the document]
jay: I also wonder how this works for FTPS TLS 1.3 session tickets on the data connection as reported in #6149. There was a report that @mkauf's change to SSL_read first worked to read any unread tickets in the queue, which otherwise would not be read if data channel is upload only.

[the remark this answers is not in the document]
icing: @jay I assume you mean something like the change I added just now?

lib/vtls/openssl.c
```diff
@@ -1882,6 +1882,8 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
       int nread, err;
       long sslerr;
 
+      /* Do a read to catch any pending data in SSL or on the socket */
```
mkauf, line 1885: I think that the original comment was better - if something is read, it's probably a close notify alert, not just "data".

icing: Thanks. Resurrected the comment out of curl-7_86_0.

lib/vtls/openssl.c
```diff
@@ -1878,15 +1878,45 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+      long sslerr;
+
       /* Maybe the server has already sent a close notify alert.
          Read it to avoid an RST on the TCP connection. */
       (void)SSL_read(backend->handle, buf, (int)sizeof(buf));
-
-      (void)SSL_shutdown(backend->handle);
-
       ERR_clear_error();
+      if(SSL_shutdown(backend->handle)) {
```
jay, line 1889:
```suggestion
      if(SSL_shutdown(backend->handle) == 1) {
```

The manpage for [SSL_shutdown](https://www.openssl.org/docs/man3.1/man3/SSL_shutdown.html) says that 1 means finished.

lib/vtls/openssl.c
```diff
@@ -1878,15 +1878,45 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+        err = SSL_get_error(backend->handle, nread);
+        switch(err) {
+        case SSL_ERROR_NONE: /* this is not an error */
+        case SSL_ERROR_ZERO_RETURN: /* no more data */
+          CURL_TRC_CF(data, cf, "SSL shutdown, EOF from server");
+          break;
+        case SSL_ERROR_WANT_READ:
+          /* SSL has send its notify and now wants to read the reply
+           * from the server. We are not really interested in that. */
+          CURL_TRC_CF(data, cf, "SSL shutdown, sent, leaving");
```
jay, line 1903:
```suggestion
          CURL_TRC_CF(data, cf, "SSL shutdown, sent");
```

lib/vtls/openssl.c
```diff
@@ -1878,15 +1878,45 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+          break;
+        case SSL_ERROR_WANT_WRITE:
+          CURL_TRC_CF(data, cf, "SSL shutdown, blocked send, leaving");
```
jay, line 1906:
```suggestion
          CURL_TRC_CF(data, cf, "SSL shutdown, blocked send");
```

lib/vtls/openssl.c
```diff
@@ -1878,15 +1878,45 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+          break;
+        default:
+          sslerr = ERR_get_error();
+          CURL_TRC_CF(data, cf, "SSL shutdown, error: '%s', errno %d",
+                      (sslerr ?
+                        ossl_strerror(sslerr, buf, sizeof(buf)) :
+                        SSL_ERROR_to_str(err)),
```
jay, line 1913:
```suggestion
                       ossl_strerror(sslerr, buf, sizeof(buf)) :
                       SSL_ERROR_to_str(err)),
```
also why did you assign sslerr to ERR_get_error and err to SSL_get_error, wouldn't it make more sense to use err for ERR_get_error and sslerr for SSL_get_error?

lib/vtls/openssl.c
icing: And that `0`  means "not finished". Hmm, tricky API as usual in OpenSSL.

When a function is documented to have such a BOOLEAN return value, I prefer testing it like I did.

lib/vtls/openssl.c
icing: Sure, can do if you prefer this.

lib/vtls/openssl.c
icing: see above.

lib/vtls/openssl.c
icing: Will fix the indentation. 

As to the assigns: `SSL_get_error()` returns `int`, while `ERR_get_error()` returns `long`.

lib/vtls/openssl.c
jay: That's the mistake though it's not bool it's int and when it returns -1 the shutdown is not finished

lib/vtls/openssl.c
jay: > As to the assigns: `SSL_get_error()` returns `int`, while `ERR_get_error()` returns `long`.

yes I understand but what I'm saying is sslerr makes more sense representing SSL_get_error, so what I mean is wouldn't this make more sense:

long err;
int sslerr;
...
sslerr = SSL_get_error(backend->handle, nread);
...
err = ERR_get_error();

Also wouldn't they be retrieving the same error at this point?

lib/vtls/openssl.c
icing: Ah, reading the man page further reveals this.

lib/vtls/openssl.c
icing: I admit to copying this from other places in our code. If you would like to go over the TLS implementations and correct the naming of variables, please do.

As to your question "is this the same"? As far as my memory goes, the `long` is partially the same, but has in addition the OpenSSL component (CRYPTO, etc.) coded into the value. So it includes the `int` in a way that is internal to the OpenSSL implementation, afaik.

jay: > I admit to copying this from other places in our code.

Ok I see what you mean, other functions like ossl_shutdown use that naming as well. nevermind

jay:
I've squashed all the commits to prepare for merge however this does not work as I expected. The internal handle used by the connection cache to close connections (ie the closure_handle) has verbose mode false and CURL_LOG_LVL_NONE so none of those trace messages are output when `CURL_DEBUG=ssl`

icing:
> I've squashed all the commits to prepare for merge however the change does not work as I expected. The internal handle used by the connection cache to close connections (ie the closure_handle) has verbose mode false and CURL_LOG_LVL_NONE so none of those trace messages are output when `CURL_DEBUG=ssl`

Yes, that is a known current limitation of libcurl/curl as documented in #11878. I think we should enable `curl` to also trace actions on connection cache cleanup.

This is outside of this PR really, it applies to all `close()` handling, with or without SSL.

bagder:
> The internal handle used by the connection cache to close connections (ie the closure_handle) has verbose mode false and CURL_LOG_LVL_NONE so none of those trace messages are output when CURL_DEBUG=ssl

This is an old and long-standing issue that we need to fix, but it is separate and needs a larger take.

jay:
Thanks. Let's try this out and see how it goes.

PR #11861, opened 2023-09-15
ftp, make disconnect persevere more

icing:
- we test if FTP transfers send the final QUIT, but this is unreliable when the network is blocking
- this leads to test failures in CI, test1387 for example
- this PR retries sending the QUIT data in a loop with waits.
- this is not a very good soluition, but lets see if out CI tests like it.


icing:
Abandoning this. The approach is not corrent. nor reliable. Opened #11878 for discussion of the general shortcomings here.

PR #11866, opened 2023-09-16
curl-www: remove questionable text

jhauga:
Removed (page located at `http://www.formpost.com/getthis/`). Was tinkering and got weird popups after copying and pasting into browser. Anyway the site isn't active anymore. It's listed for sale.

### DIFF
--- a/docs/MANUAL.md
+++ b/docs/MANUAL.md
@@ -323,8 +323,6 @@ of the letter's ASCII code.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


bagder (changes requested):
I think it should rather be changed to use `example.com`, as there is a second reference to the domain in the example command line below the HTML.

bagder (approved):

bagder:
Thanks!

PR #11868, opened 2023-09-16
http_aws_sigv4: fix sorting with empty parts

### LINKED ISSUE #11855: AWS signature test fails on Solaris
### I did this

Running test 439 on a Solaris 11 SPARC-64 machine causes test 439 to fail. This can be seen in the Solaris [curl autobuilds](https://curl.se/dev/builds.html) and I've been able to reproduce it myself on another Solaris 11 machine. Both i386 and SPARC machine are affected, as are Solaris 10 and 11.

The test fails with `protocol FAILED` with this diff:

```diff
-Authorization: AWS4-HMAC-SHA256 Credential=user/19700101/us-east-2/es/aws4_request, SignedHeaders=host;x-amz-date, Signature=88884e3b3142133685b2092d29d8b522b785b1a9ec9e4a90cbea83e882f8dcb6[CR][LF]
+Authorization: AWS4-HMAC-SHA256 Credential=user/19700101/us-east-2/es/aws4_request, SignedHeaders=host;x-amz-date, Signature=873faed63dd4fff88adb7004aa31cf9584953b64f501c774c7f8494dfc471433[CR][LF]
```

### I expected the following

Test 439 should pass.

### curl/libcurl version

curl 8.3.0-DEV (sparc-sun-solaris2.11) libcurl/8.3.0-DEV OpenSSL/1.0.2u zlib/1.2.13 brotli/1.0.2 zstd/1.4.5 libidn2/2.0.4 libpsl/0.21.0 (+libidn2/2.0.4) nghttp2/1.30.0 librtmp/2.3 libgsasl/1.8.0 OpenLDAP/2.4.40
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc brotli Debug gsasl HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB PSL SSL threadsafe TLS-SRP TrackMemory UnixSockets zstd

### operating system

Solaris 10 & Solaris 11

dfandrich:
When comparing with an empty part, the non-empty one is always
considered greater-than. Previously, the two would be considered equal
which would randomly place empty parts amongst non-empty ones.

Fixes #11855
Closes #11868

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -409,6 +409,11 @@ static int compare_func(const void *a, const void *b)
 {
   const struct pair *aa = a;
   const struct pair *bb = b;
+  /* If one element is empty, the other is always sorted higher */
+  if(aa->len == 0)
+    return -1;
+  if(bb->len == 0)
+    return 1;
   return strncmp(aa->p, bb->p, aa->len < bb->len ? aa->len : bb->len);
 }
 


bagder (approved):
Nice catch!

lib/http_aws_sigv4.c
```diff
@@ -409,6 +409,11 @@ static int compare_func(const void *a, const void *b)
 {
   const struct pair *aa = a;
   const struct pair *bb = b;
+  /* If one element is empty, the other is always sorted higher */
+  if(aa->len == 0)
+    return -1;
+  if(bb->len == 0)
+    return 1;
```
bagder, line 416: For a second I thought this might need a special case for when both lengths are zero, but I doubt that is true. This code should make the sorting stable and working nonetheless.

PR #11879, opened 2023-09-18
websockets: Expose WebSocket FIN and RSV bits in public API

LinuxBrickie:
This commit is not intended to be merged yet as further work is required. I will add comments in the PR to some of the changes to explain what still needs to be done.

When receiving WebSocket data with curl_ws_recv there was no way to know when a fragmented message had been fully received. A fragmented message is one which is sent over multiple frames. The total size of such a message is not necessarily known a priori so the protocol provides a "FIN" header bit to mark the final frame of the message.

The existing libtest 2305 involved the test server sending a fragmented message over three frames but used the known message size to test for message completion. The test has been updated to test for the FIN bit. The test also did not fail if insufficient bytes were received so that has been fixed too.

### DIFF
--- a/include/curl/websockets.h
+++ b/include/curl/websockets.h
@@ -30,6 +30,7 @@ extern "C" {
 
 struct curl_ws_frame {
   int age;              /* zero */
+  int actual_flags;      /* See the CURLWS_FLAGS_* defines */
   int flags;            /* See the CURLWS_* defines */
   curl_off_t offset;    /* the offset of this data into the frame */
   curl_off_t bytesleft; /* number of pending bytes left of the payload */
@@ -37,6 +38,12 @@ struct curl_ws_frame {
 };
 
 /* flag bits */
+#define CURLWS_FLAGS_FIN   (1<<7)
+#define CURLWS_FLAGS_RSV1  (1<<6)
+#define CURLWS_FLAGS_RSV2  (1<<5)
+#define CURLWS_FLAGS_RSV3  (1<<4)
+
+/* flag bits - not actually flags, opcodes + CURLWS_OFFSET - TODO */
 #define CURLWS_TEXT       (1<<0)
 #define CURLWS_BINARY     (1<<1)
 #define CURLWS_CONT       (1<<2)
--- a/lib/ws.c
+++ b/lib/ws.c
@@ -45,7 +45,8 @@
 #include "memdebug.h"
 
 
-#define WSBIT_FIN 0x80
+#define WSBIT_FLAGS_MASK  (0xf0)
+
 #define WSBIT_OPCODE_CONT  0
 #define WSBIT_OPCODE_TEXT  (1)
 #define WSBIT_OPCODE_BIN   (2)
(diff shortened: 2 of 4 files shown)


include/curl/websockets.h
```diff
@@ -30,13 +30,20 @@ extern "C" {
...
   curl_off_t offset;    /* the offset of this data into the frame */
   curl_off_t bytesleft; /* number of pending bytes left of the payload */
   size_t len;           /* size of the current data chunk */
 };
 
 /* flag bits */
+#define CURLWS_FLAGS_FIN   (1<<7)
+#define CURLWS_FLAGS_RSV1  (1<<6)
+#define CURLWS_FLAGS_RSV2  (1<<5)
+#define CURLWS_FLAGS_RSV3  (1<<4)
```
LinuxBrickie, line 44: I've just gone with the values that ultimately appear in the header hence the bit shifting starting at 4.

include/curl/websockets.h
```diff
@@ -30,13 +30,20 @@ extern "C" {
 
 struct curl_ws_frame {
   int age;              /* zero */
+  int actual_flags;      /* See the CURLWS_FLAGS_* defines */
```
LinuxBrickie, line 33: Clearly this is not the final variable name. This is just to distinguish it from the existing `flags` variable.

What is currently in `flags` is actually the opcode and the offset flag. The opcode is not comprised of flags so I suggest we change the `CURLWS_TEXT`, `CURLWS_BINARY` and so on "flags" so be their actual opcode values. Having them as flags seems very confusing to me and makes callers think they can be ORed together which AFAICS they can't. In fact I think it is impossible to send a fragmented message but I'll go into that in a further email on the libcurl mailing list.

Further work then is to rename `flags` as `opcode` and deal with the `CURLWS_OFFSET` flag separately.

bagder:
> A fragmented message is one which is sent over multiple frames. The total size of such a message is not necessarily known a priori so the protocol provides a "FIN" header bit to mark the final frame of the message.

The `CURLWS_CONT` flag is currently documented to be set if it is *not* the final fragment, thus the complete inverse of `FIN` - so that should indeed be possible to know.

LinuxBrickie:
> The `CURLWS_CONT` flag is currently documented to be set if it is _not_ the final fragment, thus the complete inverse of `FIN` - so that should indeed be possible to know.

Yes, that's what I initially thought and tried but the CURLWS_CONT flag is not set in the first frame either. In my original email to the list I pointed out that CURLWS_CONT not being set for the first frame *is* what I'd expect if it has the same meaning as the WebSocket CONT opcode. Your [reply](https://curl.se/mail/lib-2023-07/0013.html) to that was that that was the intention

So, if the intention really is for CURLWS_CONT to match the meaning of the CONT opcode, it needs to be unset for the first frame and set for all  remaining frames. IMO if we try to use the same terminology but with different meanings I think it will lead to confusion for both callers and maintainers.

Also, bear in mind that the discussion here is not only about the FIN bit. The RSV* bits also need exposed as they can be used by extensions that the client and server agree upon. No idea why they called them "reserved", a better name would have been "extensions" and denoting them EXT* but hey.

bagder:
> So, if the intention really is for CURLWS_CONT to match the meaning of the CONT opcode, it needs to be unset for the first frame and set for all remaining frames

The way I see it, having `CURLWS_CONT` work in a consistent way should provide the same functionality as exposing `FIN`, shouldn't it? (I'm not a WebSocket expert. I might be getting this wrong.)

If it doesn't, then I think it should be fixed. And if it does, then why also expose `FIN` ? That would just make us provide two different signals for the same thing I see no good in that.

> bear in mind that the discussion here is not only about the FIN bit

Sure, but I don't see any issues with those

include/curl/websockets.h
bagder: One challenge: there are three RSV bits in every incoming WebSocket "frame", but the messages you get from libcurl may be a partial frame, a fragment, so we need to make sure that we return these bits set for all fragments of that frame in which the RSV bits are set.

LinuxBrickie:
> The way I see it, having `CURLWS_CONT` work in a consistent way should provide the same functionality as exposing `FIN`, shouldn't it? (I'm not a WebSocket expert. I might be getting this wrong.)

Not if it is consistent with the CONT opcode in the websocket protocol, no. If that were possible then the protocol would have no need of the FIN bit.

A simple example would be a 3-frame message. The protocol requires the frames to have the following OpCode and FIN bits in their headers:

| Frame no. | OpCode | FIN bit |
|--------|--------|--------|
| 1 | TEXT | 0 |
| 2 | CONT | 0 |
| 3 |CONT | 1 |

The OpCode is a 4-bit integer and is *not* a bit field. Confusingly the CURLWS_* defines suggest otherwise and I think that's a big mistake and possibly where the confisuion may be coming from.

bagder:
This PR has 69 failed jobs with no further updates and I have yet to understand which situations the current API cannot handle. Closing.

roberte777:
This is needed to be able to parse fragmented messages, CURLWS_CONT does not do what is required. Would it be possible for someone to pick this back up? 

| Frame No.   |      OpCode      |  FIN bit |
|----------|:-------------:|------:|
| 1 |  TEXT | 0 |
| 2 |    CONT   |   0 |
| 3 | CONT |    1 |

In the above example, in the first message, OpCode gives no indication that I should be expecting further fragments. I need to know if I should be expecting other fragments so I know if I should return the websocket message to the user or continue receiving for the rest of it. The FIN bit being zero tells me that I have not received the full message.

In the case of the last example, the OpCode is CONT. This gives no indication that I have received the full message and can return it to the user. On the contrary, seeing that the FIN bit is 1 would indicate I have received the full message and should return it to the user.

PR #11880, opened 2023-09-18
trace: fix C89 build with CURL_DISABLE_VERBOSE_STRINGS

hwti:
gcc supports variadic macros even in C89 mode, so Curl_trc_cf_infof is both defined as Curl_nop_stmt and as a stub function.
Moreover, curl_trc.h errors without variadic macro support, so the stub is unnecessary.

### DIFF
--- a/lib/curl_trc.c
+++ b/lib/curl_trc.c
@@ -238,14 +238,4 @@ CURLcode Curl_trc_init(void)
   return CURLE_OK;
 }
 
-#if !defined(__STDC_VERSION__) || (__STDC_VERSION__ < 199901L)
-void Curl_trc_cf_infof(struct Curl_easy *data, struct Curl_cfilter *cf,
-                       const char *fmt, ...)
-{
-  (void)data;
-  (void)cf;
-  (void)fmt;
-}
-#endif
-
 #endif /* !DEBUGBUILD */


bagder:
This does not help silence the compiler warnings for me. If I use `--enable-debug --disable-verbose` even after applying this patch I get warnings (errors with `--enable-werror`) that start like this:

~~~
curl_trc.h:137:15: error: anonymous variadic macros were introduced in C99 [-Wvariadic-macros]
  137 | #define infof(...)  Curl_nop_stmt
      |               ^~~
curl_trc.h:138:21: error: anonymous variadic macros were introduced in C99 [-Wvariadic-macros]
  138 | #define CURL_TRC_CF(...)  Curl_nop_stmt
      |                     ^~~
curl_trc.h:139:27: error: anonymous variadic macros were introduced in C99 [-Wvariadic-macros]
  139 | #define Curl_trc_cf_infof(...)  Curl_nop_stmt
      |                           ^~~

hwti:
> This does not help silence the compiler warnings for me. If I use `--enable-debug --disable-verbose` even after applying this patch I get warnings (errors with `--enable-werror`) that start like this:
> 

In fact there are more cases than I thought :
 - with `--enable-warnings` (or `--enable-debug`), `-std=gnu89` is added for gcc (that's how you see the warning), but not clang for example
 - else the autotools add `-std=gnu11` (or `-std=gnu99`, ...) if supported (done by `_AC_PROG_CC_STDC_EDITION` in `AC_PROG_CC`)
 - with CMake, nothing is added and `PICKY_COMPILER` is `ON` by default, so the warnings are present on any compiler which defaults to C89
 
 I was testing with CMake, `-DPICKY_COMPILER=OFF -DCURL_DISABLE_VERBOSE_STRINGS=ON`, and gcc 4.8 (gnu89 by default).
 
Maybe we could align the CMake behavior with the autotools (ie allowing more recent standards if detected, enabling `PICKY_COMPILER` by default only on debug build, and have it force C89).

But for the warning, what would be the preferred way ?

`-Wno-variadic-macros` is supported by gcc >= 4.0, older versions have no way to disable the warning without dropping `-pedantic` or switching to C99.

We could mirror the verbose string path with :
```
 #if defined(HAVE_VARIADIC_MACROS_C99) && \
    defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)
#define infof(...)  Curl_nop_stmt
#define CURL_TRC_CF(...)  Curl_nop_stmt
#define Curl_trc_cf_infof(...)  Curl_nop_stmt
#else
void Curl_nop_infof(struct Curl_easy *data, const char *fmt, ...);
#define infof Curl_nop_infof
#define CURL_TRC_CF Curl_trc_cf_infof
#endif
```
But :
 - the strings would actually be in the binary despite the `CURL_DISABLE_VERBOSE_STRINGS`
 - `HAVE_VARIADIC_MACROS_C99` and `HAVE_VARIADIC_MACROS_GCC` would have less meaning
 - a few locations would need to be fixed, when variables are under `#if !defined(CURL_DISABLE_VERBOSE_STRINGS)` but not the `infof()` using them

OlekMazur (commented):
This change fixes issue #11891 in my case.

hwti:
There are also a few unused variables / arguments warnings which appear with `CURL_DISABLE_VERBOSE_STRINGS`, since `info` and others are defined as `Curl_nop_stmt`.

We could keep the common macro definition, which would have a `if(FALSE)`.
At least with gcc, it allows to mark variables as used and check the printf format, without generating any code (even at `-O0` IIRC).
As with the current code, variadic macro support would be needed for `CURL_DISABLE_VERBOSE_STRINGS` (else there is no way to remove the strings from the library).
There would be a few cases in the code where the `CURL_DISABLE_VERBOSE_STRINGS` checks would need to be adjusted, to either keep the variables defined or to flag the `infof` call, or to define macros to replace functions.

```C
#if defined(HAVE_VARIADIC_MACROS_C99)

#define infof(data, ...) \
  do { if(Curl_trc_is_verbose(data)) \
         Curl_infof(data, __VA_ARGS__); } while(0)
#define CURL_TRC_CF(data, cf, ...) \
  do { if(Curl_trc_cf_is_verbose(cf, data)) \
         Curl_trc_cf_infof(data, cf, __VA_ARGS__); } while(0)

#elif defined(HAVE_VARIADIC_MACROS_GCC)

#define infof(data, args...) \
  do { if(Curl_trc_is_verbose(data)) \
         Curl_infof(data, args); } while(0)
#define CURL_TRC_CF(data, cf, args...) \
  do { if(Curl_trc_cf_is_verbose(cf, data)) \
         Curl_trc_cf_infof(data, cf, args); } while(0)

#else /* no variadic macro support */

#if !defined(CURL_DISABLE_VERBOSE_STRINGS)
#define infof Curl_infof
#define CURL_TRC_CF Curl_trc_cf_infof
#else
#error "No variadic macro support, CURL_DISABLE_VERBOSE_STRINGS is not supported"
#endif

#endif

#if !defined(CURL_DISABLE_VERBOSE_STRINGS)
/* informational messages enabled */

#define Curl_trc_is_verbose(data)    ((data) && (data)->set.verbose)
#define Curl_trc_cf_is_verbose(cf, data) \
                            ((data) && (data)->set.verbose && \
                            (cf) && (cf)->cft->log_level >= CURL_LOG_LVL_INFO)

#else /* !CURL_DISABLE_VERBOSE_STRINGS */
/* All informational messages are not compiled in for size savings */

#define Curl_trc_is_verbose(d)        ((void)(d), FALSE)
#define Curl_trc_cf_is_verbose(x,y)   ((void)(x), (void)(y), FALSE)

#endif /* CURL_DISABLE_VERBOSE_STRINGS */
```

hwti:
For the `-Wvariadic-macros`, they could be disabled around the macro definitions :
```C
#if defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wvariadic-macros"
#endif
...
#if defined(__GNUC__)
#pragma GCC diagnostic pop
#endif
```
But `#pragma GCC diagnostic push` only works for gcc >= 4.6, so maybe it would be better to add the `-Wno-variadic-macros` in the buildsystem.

vszakats:
In libssh2 we went through this earlier this year and settled with a pure C89 solution. Its only drawback is a little unusual syntax with double parenthesis, e.g.:
```c
        _libssh2_debug((session, LIBSSH2_TRACE_CONN,
                       "Cancelling tcpip-forward session for %s:%d",
                       listener->host, listener->port));
```

This seems much less of an issue than fighting non-C89 code with C89 compilers.

PR #11885, opened 2023-09-19
Curl_client_write, disambiguate flag semantics

icing:
- use CLIENTWRITE_BODY *only* when data is actually body data
- add CLIENTWRITE_INFO for meta data that is *not* a HEADER
- debug assertions that BODY/INFO/HEADER is not used mixed
- move `data->set.include_header` check into Curl_client_write so protocol handlers no longer have to care
- add special in FTP for `data->set.include_header` for historic, backward compatible reasons
- move unpausing of client writes from easy.c to sendf.c, so that code is in one place and can forward flags correctly

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -2067,6 +2067,31 @@ static bool ftp_213_date(const char *p, int *year, int *month, int *day,
   return TRUE;
 }
 
+static CURLcode client_write_header(struct Curl_easy *data,
+                                    char *buf, size_t blen)
+{
+  /* Some replies from an FTP server are written to the client
+   * as CLIENTWRITE_HEADER, formatted as if they came from a
+   * HTTP conversation.
+   * In all protocols, CLIENTWRITE_HEADER data is only passed to
+   * the body write callback when data->set.include_header is set
+   * via CURLOPT_HEADER.
+   * For historic reasons, FTP never played this game and expects
+   * all its HEADERs to do that always. Set that flag during the
+   * call to Curl_client_write() so it does the right thing.
+   *
+   * Notice that we cannot enable this flag for FTP in general,
+   * as an FTP transfer might involve a HTTP proxy connection and
+   * headers from CONNECT should not automatically be part of the
+   * output. */
+  CURLcode result;
+  int save = data->set.include_header;
+  data->set.include_header = TRUE;
+  result = Curl_client_write(data, CLIENTWRITE_HEADER, buf, blen);
+  data->set.include_header = save;
+  return result;
+}
+
 static CURLcode ftp_state_mdtm_resp(struct Curl_easy *data,
                                     int ftpcode)
 {
@@ -2120,8 +2145,7 @@ static CURLcode ftp_state_mdtm_resp(struct Curl_easy *data,
                   tm->tm_hour,
                   tm->tm_min,
                   tm->tm_sec);
-        result = Curl_client_write(data, CLIENTWRITE_BOTH, headerbuf,
-                                   headerbuflen);
+        result = client_write_header(data, headerbuf, headerbuflen);
         if(result)
           return result;
       } /* end of a ridiculous amount of conditionals */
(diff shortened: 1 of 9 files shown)


lib/ftp.c
```diff
@@ -2067,6 +2067,31 @@ static bool ftp_213_date(const char *p, int *year, int *month, int *day,
...
+   * all its HEADERs to do that always. Set that flag during the
+   * call to Curl_client_write() so it does the right thing.
+   *
+   * Notice that we cannot enable this flag for FTP in general,
+   * as an FTP transfer might involve a HTTP proxy connection and
+   * headers from CONNECT should not automatically be part of the
+   * output. */
+  CURLcode result;
+  int save = data->set.include_header;
+  data->set.include_header = TRUE;
```
bagder, line 2089: Where within `Curl_client_write()` is `data->set.include_header` checked? I can't find it, so I'm not following why it needs to be set like this!

bagder: Never mind. I see it now, this patch adds the check.

Would it not be nicer to pass in the "include_header" bool as a new argument to Curl_client_write  instead of changing the variable like this?

icing: > Never mind. I see it now, this patch adds the check.
> 
> Would it not be nicer to pass in the "include_header" bool as a new argument to Curl_client_write instead of changing the variable like this?

My thinking is that FTP is the sole exception here and that every other protocol does not have to care about this. I'd rather spare all others the "burden" to think about it than make FTP look nicer.

bagder: Ah right. Sounds fair.

lib/sendf.h
```diff
@@ -28,18 +28,33 @@
...
+ * - HEADER can have additional bits set (more than one)
+ *   - STATUS special "header", e.g. response status line in HTTP
+ *   - CONNECT header was received during proxying the connection
+ *   - 1XX header is part of an intermediate response, e.g. HTTP 1xx code
+ *   - TRAILER header is trailing response data, e.g. HTTP trailers
+ * BODY, INFO and HEADER should not be mixed, as this would lead to
+ * confusion on how to interpret/format/convert the data.
+ */
+#define CLIENTWRITE_BODY    (1<<0) /* non-meta information, BODY */
+#define CLIENTWRITE_INFO    (1<<1) /* meta information, not a HEADER */
```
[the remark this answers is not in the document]
icing: Yeah, that is what I though first as well. But then test cases failed. There are cases, where `data->req.include_header` is set, but pingpong never wrote with `CLIENTWRITE_BODY´. Other protocol handler parts did, though. I believe IMAP is such a case, if memory serves me.

So, before this PR, we had calls to `Curl_client_write()` that:

1. added `CLIENTWRITE_BODY` when `data->req.include_header` was TRUE
2. did not add `CLIENTWRITE_BODY` although `data->req.include_header` was TRUE
3. added `CLIENTWRITE_BODY` irregardless of `data->req.include_header`

Case 1 is now automatically handles in `Curl_client_write()`. Case 2 is now changed to `CLIENTWRITE_INFO`. Case 3 is handled via the special flag set in FTP (and only there because FTP via HTTP PROXY does not want to see the CONNECT headers).

Why do all this? Because with this PR, the flags say what the written data is - and no longer to which callback it shall be passed to. So `CLIENTWRITE_BODY` really is only used for body data.

When this holds true, it is possible to move several things into `Curl_client_write()`:

* content encoding writers, e.g. `data->req.write_stack`
* chunked decoding
* progress updates

making life for transfer loop and protocol handlers easier.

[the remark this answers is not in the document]
icing: To explain the case some more: we have this nice writer stack in `content_encoding.[ch]`. That is also called for `Transfer-Encoding` headers. But `chunked` is not implemented there. Why not?

Well, `chunked` changes the amount of bytes in a response. The chunk framing is not counted against `Content-Length` and download progress updates. In order to count the bytes correctly, it needs to know what of its buffer is *really* the content.

So, when writing received transfer data, one needs to `if()` check for chunked encoding, call it, get back an updated buffer position and length and use that for client writes and progress updates.

If `chunked` were a content_encoding writer, we could just write the received data through the writer stack and it would take care of all this.  Add a "progress" writer *after* "chunked" and updates just work for everyone.

bagder (approved):

bagder:
Thanks!

PR #11888, opened 2023-09-19
test1474: disable test on NetBSD, OpenBSD and Solaris 10

dfandrich:
These two kernels only send a fraction of the requested amount of the
first large block, invalidating the assumptions of the test and causing
it to fail.

Assisted-by: Christian Weisgerber
Ref: https://curl.se/mail/lib-2023-09/0021.html
Closes #11888


dfandrich:
I discovered the failure also frequently shows up in the Solaris 10 autobuilds, but not all the time and not on Solaris 11, so I've disabled the test there as well.

PR #11897, opened 2023-09-20
tftpd: workaround a stringop-overread warning

monnerat:
When optimizing, GCC 12 detects and reports a stringop-overread warning:

tftpd.c: In function ‘write_behind.isra’:
tftpd.c:485:12: warning: ‘write’ reading between 1 and 2147483647 bytes from a region of size 0 [-Wstringop-overread]
  485 |     return write(test->ofile, writebuf, count);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from tftpd.c:71:
/usr/include/arpa/tftp.h:58:30: note: source object ‘tu_data’ of size 0
   58 |                         char tu_data[0];        /* data or error string */
      |                              ^~~~~~~

This occurs because writebuf points to this field and the latter cannot be considered as being of dynamic length because it is not the last field in the structure. Thus it is bounded to its declared size.

This commit explicitly computes the address to unbind it from the declaration.


bagder:
Nice catch. But since we also need to have a working private header, can't we instead just use that always and then keep the existing approach since our struct has the field last?

IOW, can this maybe be an alternative fix?

~~~diff
diff --git a/tests/server/tftpd.c b/tests/server/tftpd.c
index 8bcd452d0..75de7f430 100644
--- a/tests/server/tftpd.c
+++ b/tests/server/tftpd.c
@@ -65,15 +65,12 @@
 #include <netinet/in.h>
 #endif
 #ifdef HAVE_ARPA_INET_H
 #include <arpa/inet.h>
 #endif
-#ifdef HAVE_ARPA_TFTP_H
-#include <arpa/tftp.h>
-#else
 #include "tftp.h"
-#endif
+
 #ifdef HAVE_NETDB_H
 #include <netdb.h>
 #endif
 #ifdef HAVE_SYS_FILIO_H
 /* FIONREAD on Solaris 7 */

monnerat:
> IOW, can this maybe be an alternative fix?

Probably, I have to check.
In general, I prefer using system-provided headers when they are some because they are subject to definition changes, but in the tftp case there is almost no chance this can occur !!!

monnerat:
> Probably, I have to check.

Checked: OK. No more warning, compiles and runs tftp tests successfully.
It can also be a candidate to drop `HAVE_ARPA_TFTP_H`.

I let you apply your preferred fix at your convenience.

bagder:
> Checked: OK. No more warning, compiles and runs tftp tests successfully. It can also be a candidate to drop `HAVE_ARPA_TFTP_H`.

Yeah, that's my preferred route to go with this.

monnerat:
The commit in this PR has been completely rewritten to always use curl's own tftp.h.

bagder:
Thanks!

PR #11903, opened 2023-09-21
add CURLSSLBACKEND_ALL enum

TedLyngmo:
[ssl]: add CURLSSLBACKEND_ALL enum for cleaner code and examples

This is to make code that now does `(curl_sslbackend)-1` clearer
and easier to read:
```
/* list the available ones */
const curl_ssl_backend **list;
curl_global_sslset(CURLSSLBACKEND_ALL, NULL, &list);
```


include/curl/curl.h
```diff
@@ -170,7 +170,8 @@ typedef enum {
   CURLSSLBACKEND_MBEDTLS = 11,
   CURLSSLBACKEND_MESALINK               CURL_DEPRECATED(7.82.0, "") = 12,
   CURLSSLBACKEND_BEARSSL = 13,
-  CURLSSLBACKEND_RUSTLS = 14
+  CURLSSLBACKEND_RUSTLS = 14,
+  CURLSSLBACKEND_ALL = ~0
```
bagder, line 174: I think it should be set to `-1` and be made first in the list.

bagder (commented):
Also, you did not update `symbols-in-versions` so test 1119 fails.

include/curl/curl.h
TedLyngmo: Ok, but why prefer `-1`? We want all bits set, which `~0` guarantees.

TedLyngmo:
> symbols-in-versions

Thanks! What will the next version be? 8.3.1?

include/curl/curl.h
bagder: It is not used like that. It is just supposed to be a number that does not match one of the existing (or future) TLS backends. See:

https://github.com/curl/curl/blob/3d53f211e5c907bb6e8aaddf05a3acdcae247e5d/lib/vtls/vtls.c#L1369-L1396

bagder:
> Thanks! What will the next version be? 8.3.1?

We are fairly sure it will become 8.4.0 so I think you can use that.

jay:
IMO `curl_global_sslset(CURLSSLBACKEND_ALL, ...` is confusing because it implies setting all backends. To me, it makes more sense as -1.

bagder:
> CURLSSLBACKEND_ALL

Yeah, maybe *_NONE instead?

TedLyngmo:
> > CURLSSLBACKEND_ALL
> 
> Yeah, maybe *_NONE instead?

`CURLSSLBACKEND_NONE` is already defined to `0`. I used `_ALL` because it seemed to fit into the documentation:

```
/* list the available ones */
const curl_ssl_backend **list;
curl_global_sslset((curl_sslbackend)-1, NULL, &list);
````

bagder:
> `CURLSSLBACKEND_NONE` is already defined to `0`. I used `_ALL` because it seemed to fit into the documentation:

Oh right. But maybe we should then rather change the documentation to use *NONE instead of -1? As far as I can see, that gives the same result...

TedLyngmo:
> > `CURLSSLBACKEND_NONE` is already defined to `0`. I used `_ALL` because it seemed to fit into the documentation:
> 
> Oh right. But maybe we should then rather change the documentation to use *NONE instead of -1? As far as I can see, that gives the same result...

Ah, well, that would certainly be even easier if it indeed does the same thing :-)

TedLyngmo:
Since #11909 has been approved, this PR is meaningless. Closing.

PR #11906, opened 2023-09-21
github/labeler: fix up more the labeler config format

dfandrich:
The labeler job doesn't like the formatting, despite the YAML parsers I
tested being fine with it.

Follow-up to dd12b452a
Closes #11906


dfandrich:
It looks like the PR job isn't picking up the job config change, so I'm pushing it anyway, hoping this fixes it.

PR #11908, opened 2023-09-21
Move `contenc_writer` infra to Curl_client_write()

icing:
Now that `CLIENTWRITE_BODY` is only used for response body bytes, move the infrastructure for `struct contenc_writer` into `Curl_client_write()`. Available to any protocol.

- move definitions from content_encoding.h to sendf.h
- move create/cleanup/add code into sendf.c
- installed content_encoding writers will always be called on Curl_client_write(CLIENTWRITE_BODY)
- Curl_client_cleanup() frees writers and tempbuffers from paused transfers, irregardless of protocol

A rename of the struct is planned later (naming!), but I wanted the move to be easier to read in this PR.


jay:
@monnerat

bagder (approved):
I like it!

PR #11914, opened 2023-09-22
cmake: lib `CURL_STATICLIB` fixes (Windows)

### LINKED ISSUE #11844: When building a static library project for Visual Studio with CMake, the preprocessor variable CURL_STATICLIB does not get set
### I did this

I used the CMake GUI on Windows and checked the checkbox next to BUILD_STATIC_LIBS, to generate Visual Studio solution and project files.

### I expected the following

I expected CMake to generate project files which define the preprocessor variable CURL_STATICLIB that appears on line 128 in curl.h. However, this preprocessor variable was not defined. The CMake variables ENABLE_STATIC and CPPFLAG_CURL_STATICLIB (which is the empty string if no static libs are to be built) don't seem to have an effect when using CMake for Visual Studio project generation - they only seem to have an effect on the autoconf files. Maybe, a line such as
``` CMake
add_definitions(${CPPFLAG_CURL_STATICLIB})
```
might have to be added to CMakeLists.txt. Will be happy to submit a pull request. This issue does not yet seem to appear under https://curl.se/docs/knownbugs.html#CMake. 

### curl/libcurl version

curl 8.3.0-DEV

### operating system

Windows 10

vszakats:
- always define `CURL_STATICLIB` when building libcurl for Windows.

  This disables `__declspec(dllexport)` for exported libcurl symbols.
  In normal mode (hide symbols) these exported symbols are specified
  via `libcurl.def`. When not hiding symbols, all symbols are exported
  by default.

  Regression from 1199308dbc902c52be67fc805c72dd2582520d30

  Fixes #11844

- fix to omit `libcurl.def` when not hiding private symbols.

  Regression from 2ebc74c36a19a1700af394c16855ce144d9878e3

- fix `ENABLED_DEBUG=ON` + shared curl tool Windows builds by also
  omitting `libcurl.def` in this case, and exporting all symbols
  instead. This ensures that a shared curl tool can access all debug
  functions which are not normally exported from libcurl DLL.

- delete `INTERFACE_COMPILE_DEFINITIONS "CURL_STATICLIB"` for "objects"
  target.

  Follow-up to 2ebc74c36a19a1700af394c16855ce144d9878e3

- delete duplicate `BUILDING_LIBCURL` definitions.

- fix `HIDES_CURL_PRIVATE_SYMBOLS` to not overwrite earlier build settings.

  Follow-up to 1199308dbc902c52be67fc805c72dd2582520d30

Closes #11914


vszakats:
A new issue came up: `memdebug.c` functions cannot be accessed in `ENABLED_DEBUG=ON` + shared `curl.exe` builds. Because these are missing from `libcurl.def` and after this PR `CURL_STATICLIB` leaves them unexported.

vszakats:
Solved the `ENABLED_DEBUG=ON` + shared tool case by introducing `libcurl-debug.def`.

vszakats:
Solved this one by extending `libcurl-debug.def` to include all standard libcurl API definitions as well. Plus a test for it. This fixes debug + shared tool builds with both older CMake and `Makefile.mk` builds.

This perhaps isn't the most elegant solution, but I cannot think of a better one. Other alternatives may be to:
- Split off debug functions to a separate lib and link that statically always (this would be likely more complex.)
- Find a CMake feature that is able to set list of exports with a regex (`curl_*`) for all supported compilers.
- Either drop Windows support from `Makefile.mk` or find an `ldd`/`ld` option to limit exports to a regex.

vszakats:
Next issues:
1. non-Windows symbol `curl_dbg_socketpair` is picked up by `tests/extern-scan.pl`.
2. `tests/extern-scan.pl` misses to pick up `curl_dbg_malloc`, `curl_dbg_calloc`, `curl_dbg_realloc`.

vszakats:
Tending to prefer the option 'Export all symbols for debug builds' as a reasonable compromise. This will make it unnecessary to roll a `libcurl-debug.def`.

vszakats:
Strange error from a single CI job:
```
=== Start of file stderr3203
 Unsuccessful open on filename containing newline at ../../tests/extern-scan.pl line 88.
 readline() on closed filehandle H at ../../tests/extern-scan.pl line 90.
 Unsuccessful open on filename containing newline at ../../tests/extern-scan.pl line 88.
 readline() on closed filehandle H at ../../tests/extern-scan.pl line 90.
=== End of file stderr3203
```
Ref: https://github.com/curl/curl/actions/runs/6281395226/job/17062947662?pr=11914#step:11:3161

lib/CMakeLists.txt
```diff
@@ -122,18 +122,22 @@ if(NOT DEFINED SHARE_LIB_OBJECT)
   endif()
 endif()
 
+if(WIN32)
+  # On Windows we handle exports via libcurl.def. Except when symbol hiding is
+  # disabled or debug mode is enabled. In those cases we export all symbols
+  # from libcurl DLL, and skip using the .def file.
```
jay, line 128: i suggest really spell it out, for example

```suggestion
  # from libcurl DLL, and skip using the .def file.
  # Define CURL_STATICLIB always, to disable __declspec(dllexport) for exported
  # libcurl symbols.
```

otherwise i think for most people reading it it's not immediately obvious how CURL_STATICLIB relates

vszakats: Agreed. How does this version look?:
```
  # Define CURL_STATICLIB always, to disable __declspec(dllexport) for exported
  # libcurl symbols. We handle exports via libcurl.def instead. Except when
  # symbol hiding is disabled or debug mode is enabled. In those cases we
  # export _all_ symbols from libcurl DLL, without using the .def file.
```

jay: looks good

PR #11916, opened 2023-09-22
http: make several functions static when not built to use hyper

bagder:
The need to be externally reachable when hyper is used since they are then used from c-hyper. Having them static otherwise reduces scope and allows for better compiler optimizations.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -112,113 +112,10 @@ static CURLcode ws_setup_conn(struct Curl_easy *data,
                               struct connectdata *conn);
 #endif
 
-/*
- * HTTP handler interface.
- */
-const struct Curl_handler Curl_handler_http = {
-  "HTTP",                               /* scheme */
-  http_setup_conn,                      /* setup_connection */
-  Curl_http,                            /* do_it */
-  Curl_http_done,                       /* done */
-  ZERO_NULL,                            /* do_more */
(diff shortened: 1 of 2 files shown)


dfandrich (commented):
Typo in "functions" in title.

Does this really make a noticable difference? None of the functions look like they're in the data path, and applications that need to squeeze every ounce of size/speed should be able to get the same thing with LTO. The net change is an additional 16 lines of code for this, which is an ongoing a cost to developers.

bagder:
Valid questions.

Most of the 16 added lines are just the single define for static, so I don't think they add burden to the code reader.

I've come to appreciate how compilers can find more problems and optimize better with static functions, and also properly identify when functions are not used. A rather small change with a small benefit, but I figured it is worth it.

LTO is good but lacks the same knowledge of the code as the compiler has.

If you feel this is a change not worth landing I will of course respect that.

bagder:
Addition:

We have `./scripts/singleuse.pl` which helps me identify functions in libcurl that are contenders for `static` or to get removed. Long term, I would like to use that in a CI job to help us identify such functions automatically. Then we need to make single-use within-file functions static, but we also need to provide a way to mark functions as "this is a false positive, never mind". But I have not worked out the details for this. Work in progress.

bagder:
I will drop this.

PR #11918, opened 2023-09-22
lib: use wrapper for curl_mime_data fseek callback

### LINKED ISSUE #11900: 32-bit ARM: test 0650...FAILED
### I did this

Built curl on a 64-bit ARM machine in a 32-bit environment:

```
root@aarch64:/tmp# mkdir x
root@aarch64:/tmp# curl -sO https://distfiles.adelielinux.org/adelie/1.0-beta5/iso/adelie-rootfs-mini-armv7-1.0-beta5-20230829.txz
root@aarch64:/tmp# tar -C x -xf adelie-rootfs-mini-armv7-1.0-beta5-20230829.txz 
root@aarch64:/tmp# mount -B /dev x/dev
root@aarch64:/tmp# mount -t proc proc x/proc
root@aarch64:/tmp# mount -t sysfs sysfs x/sys
root@aarch64:/tmp# setarch linux32 chroot x /bin/sh
# cd
# apk add -q build-tools openssl-dev
# curl -sO https://curl.se/download/curl-8.3.0.tar.gz
# tar -xf curl-8.3.0.tar.gz 
# cd curl-8.3.0
# ./configure --with-openssl
# make -j$(nproc)
# make -j$(nproc) check
...
test 0647...OK (570 out of 1403, remaining: 03:54, took 1.118s, duration: 02:40)
test 0648...OK (571 out of 1403, remaining: 03:53, took 0.088s, duration: 02:40)
test 0649...OK (572 out of 1403, remaining: 03:53, took 0.085s, duration: 02:40)
test 0650...FAILED
test 0651...OK (574 out of 1403, remaining: 03:53, took 0.192s, duration: 02:41)
test 0652...OK (575 out of 1403, remaining: 03:53, took 0.338s, duration: 02:42)
test 0653...OK (576 out of 1403, remaining: 03:53, took 0.253s, duration: 02:42)
...
```

More information (including test logs):
* https://git.adelielinux.org/adelie/packages/-/issues/1060

### I expected the following

All tests to pass.

### curl/libcurl version

```
$ ./src/curl -V
curl 8.3.0 (armv7-foxkit-linux-musleabihf) libcurl/8.3.0 OpenSSL/1.1.1v zlib/1.3 zstd/1.5.4 libssh2/1.10.0 nghttp2/1.52.0
Release-Date: 2023-09-13
Features: alt-svc AsynchDNS HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

Adélie Linux (musl-based).

```
Linux builder-arm 5.10.35-00045-g8510b2d4996d #1 SMP PREEMPT Wed Dec 21 12:13:49 UTC 2022 armv7l GNU/Linux
```

ncopa:
fseek uses long offset which does not match with curl_off_t. This leads to undefined behavior when calling the callback and caused failure on arm 32 bit.

Use a wrapper to solve this and use fseeko which uses off_t instead of long.

Thanks to the nice people at Libera IRC #musl for helping finding this out.

Closes #11882
Closes #11900


lib/formdata.c
```diff
@@ -789,6 +789,12 @@ static CURLcode setname(curl_mimepart *part, const char *name, size_t len)
   return res;
 }
 
+/* wrap call to fseeko so it matches the calling convetion of callback */
+static int fseeko_wrapper(void *stream, curl_off_t offset, int whence)
+{
+  return fseeko(stream, (off_t)offset, whence);
```
bagder, line 795: Isn't `fseeko()` going to cause portability problems? For example I don't think it exists on Windows...

ncopa: fseeko is [POSIX.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fseek.html), but I don't know if that helps on windows.

Lets see what the CI says...

dfandrich:
We don't really test old systems in CI, and 2008 is still 19 years newer than minimal compilers we support. This really needs a configure check to gate it.

bagder:
As discussed on IRC, this should probably be done so that:

fseeko is used if existing otherwise (POSIX)
_fseeki64 is used if existing otherwise (Windows)
fseek is used (legacy)

bagder:
Maybe like this:

~~~diff
From fef1c8dbbc2c9e7e69f00a38c56ca7d493c2bca0 Mon Sep 17 00:00:00 2001
Date: Mon, 25 Sep 2023 08:31:32 +0200
Subject: [PATCH] fixup add check for and use of _fseeki64 and fseek

---
 CMakeLists.txt          | 1 +
 configure.ac            | 1 +
 lib/curl_config.h.cmake | 3 +++
 lib/formdata.c          | 8 ++++++++
 4 files changed, 13 insertions(+)

index 0b3aed906..c9fbf2c2e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1120,10 +1120,11 @@ if(WIN32)
   set(HAVE_GETADDRINFO_THREADSAFE ${HAVE_GETADDRINFO})
 endif()
 check_symbol_exists(freeaddrinfo   "${CURL_INCLUDES}" HAVE_FREEADDRINFO)
 check_symbol_exists(pipe           "${CURL_INCLUDES}" HAVE_PIPE)
 check_symbol_exists(ftruncate      "${CURL_INCLUDES}" HAVE_FTRUNCATE)
+check_symbol_exists(fseeko         "${CURL_INCLUDES}" HAVE_FSEEKO)
 check_symbol_exists(getpeername    "${CURL_INCLUDES}" HAVE_GETPEERNAME)
 check_symbol_exists(getsockname    "${CURL_INCLUDES}" HAVE_GETSOCKNAME)
 check_symbol_exists(if_nametoindex "${CURL_INCLUDES}" HAVE_IF_NAMETOINDEX)
 check_symbol_exists(getrlimit      "${CURL_INCLUDES}" HAVE_GETRLIMIT)
 check_symbol_exists(setlocale      "${CURL_INCLUDES}" HAVE_SETLOCALE)
diff --git a/configure.ac b/configure.ac
index 2fc9f2f01..0a89f4ad1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -3584,10 +3584,11 @@ AC_CHECK_DECLS([getpwuid_r], [], [AC_DEFINE(HAVE_DECL_GETPWUID_R_MISSING, 1, "Se
 
 
 AC_CHECK_FUNCS([fnmatch \
   fchmod \
   fork \
+  fseeko \
   geteuid \
   getpass_r \
   getppid \
   getpwuid \
   getpwuid_r \
diff --git a/lib/curl_config.h.cmake b/lib/curl_config.h.cmake
index 328a77d73..a1cb9d212 100644
--- a/lib/curl_config.h.cmake
+++ b/lib/curl_config.h.cmake
@@ -196,10 +196,13 @@
 #cmakedefine HAVE_FREEADDRINFO 1
 
 /* Define to 1 if you have the ftruncate function. */
 #cmakedefine HAVE_FTRUNCATE 1
 
+/* Define to 1 if you have the fseeko function. */
+#cmakedefine HAVE_FSEEKO 1
+
 /* Define to 1 if you have a working getaddrinfo function. */
 #cmakedefine HAVE_GETADDRINFO 1
 
 /* Define to 1 if the getaddrinfo function is threadsafe. */
 #cmakedefine HAVE_GETADDRINFO_THREADSAFE 1
diff --git a/lib/formdata.c b/lib/formdata.c
index 3b9b8fee3..5311060a5 100644
--- a/lib/formdata.c
+++ b/lib/formdata.c
@@ -790,11 +790,19 @@ static CURLcode setname(curl_mimepart *part, const char *name, size_t len)
 }
 
 /* wrap call to fseeko so it matches the calling convetion of callback */
 static int fseeko_wrapper(void *stream, curl_off_t offset, int whence)
 {
+#ifdef HAVE_FSEEKO
   return fseeko(stream, (off_t)offset, whence);
+#elif defined(WIN32)
+  return _fseeki64(stream, offset, whence);
+#else
+  if(offset > LONG_MAX)
+    return -1;
+  return fseek(stream, (long)offset, whence);
+#endif
 }
 
 /*
  * Curl_getformdata() converts a linked list of "meta data" into a mime
  * structure. The input list is in 'post', while the output is stored in
-- 
2.40.1

ncopa:
I'd also check for _fseeki64 instead of relying on WIN32.

Good idea with `if (offset > LONG_MAX)`.

ncopa:
Looks like fseeko is not found on macos with cmake:
https://github.com/curl/curl/actions/runs/6298724899/job/17098112644?pr=11918#step:6:217

ncopa:
> Looks like `fseeko` is not found with cmake: https://github.com/curl/curl/actions/runs/6298724899/job/17098112644?pr=11918#step:6:217

Because we need also need to add stdio.h to `CURL_INCLUDES`.


```

**Review by bagder — APPROVED** [review]

**Comment by bagder** [comment]
Thanks! Awesome work.

File: CMakeLists.txt [hunk 2]
```diff
@@ -1037,6 +1037,7 @@ check_include_file_concat("signal.h"         HAVE_SIGNAL_H)
 check_include_file_concat("stdatomic.h"      HAVE_STDATOMIC_H)
 check_include_file_concat("stdbool.h"        HAVE_STDBOOL_H)
 check_include_file_concat("stdint.h"         HAVE_STDINT_H)
+check_include_file_concat("stdio.h"          HAVE_STDIO_H)
```
vszakats, line 1040: Why was this check added?

`HAVE_STDIO_H` is not used in the source. `stdio.h` is a standard C89 header and we use it everywhere unconditionally.

ncopa: yes, it is the Fseek detection that needs it.

bagder: Is that due to some funky cmakeism? We don't have to check for stdio.h, but maybe it needs to be set somehow so that it gets used for the fseeko check?

PR #11921, opened 2023-09-22
wolfssh: do cleanup in Curl_ssh_cleanup

dfandrich:
Closes: #11921

### DIFF
--- a/lib/easy.c
+++ b/lib/easy.c
@@ -303,9 +303,6 @@ void curl_global_cleanup(void)
 
   Curl_ssh_cleanup();
 
-#ifdef USE_WOLFSSH
-  (void)wolfSSH_Cleanup();
-#endif
 #ifdef DEBUGBUILD
   free(leakpointer);
 #endif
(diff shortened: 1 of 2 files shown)


bagder (commented):
I don't think this is correct. The wolfSSH init and [cleanup](https://www.wolfssl.com/documentation/manuals/wolfssh/chapter13.html#wolfssh_cleanup) functions seem to be global for the application:

"Cleans up the wolfSSH library when done. Should be called at before termination of the application. After calling, do not make any more calls to the library."

dfandrich:
So, are you saying the call should rather be moved to main_free() in the curl tool or similiar and dropped altogether from the library? Because this change as-is doesn't change when it's being called.

bagder:
I suggest they should be called from within curl_global_init and curl_global_cleanup

jay:
it's doing that he just changed it around
curl_global_cleanup -> Curl_ssh_cleanup -> wolfSSH_Cleanup
which is what we already do for the other ssh backends

bagder:
Ah yes, sorry I didn't read this properly from the beginning. Thanks for your patience. I'll go back to my corner and be quiet now.

bagder (approved):

PR #11922, opened 2023-09-23
ECH experimental

sftcd:
This is an (as-promised, on the mailing list) early pull request for adding HTTPS RR an ECH support to cURL, that has had so far minimal testing when using OpenSSL or wolfSSL as the TLS provider, but does seem to work. (There's a known outstanding [issue](https://github.com/wolfSSL/wolfssl/issues/6802) with the wolfSSL build.) The goal for now is to get some early but detailed review as to whether the direction of the current code changes looks good. It may well be that  this PR evolves into something that can be merged, or it may require a chunk more work first, either is OK.


lib/doh.c
```diff
@@ -182,6 +186,33 @@ doh_write_cb(const void *contents, size_t size, size_t nmemb, void *userp)
...
+  char string_buffer[2048], *ptr;
+
+  ptr = string_buffer;
+  if(len > 1000) {
+    infof(data, "\t:buffer too long %ld", len);
+    return;
+  }
+  for(j = 0; j < len; j ++) {
+    *ptr++ = BIN2HEXDIGIT(buf[j] / 16);
+    *ptr++ = BIN2HEXDIGIT(buf[j] % 16);
```
bagder, line 203:
```suggestion
    *ptr++ = BIN2HEXDIGIT(buf[j] >> 4);
    *ptr++ = BIN2HEXDIGIT(buf[j] & 0x0f);
```

Personally I find that better showing that we are rotating and masking out bits.

lib/doh.c
```diff
@@ -182,6 +186,33 @@ doh_write_cb(const void *contents, size_t size, size_t nmemb, void *userp)
...
+    if((ptr - string_buffer) > 2000) {
```
bagder, line 204: Can this even happen since you already check `len` before this loop starts?

lib/doh.c
```diff
@@ -182,6 +186,33 @@ doh_write_cb(const void *contents, size_t size, size_t nmemb, void *userp)
...
+#ifdef USE_HTTPSRR
+static void local_print_buf(struct Curl_easy *data,
+                            unsigned char *buf, size_t len)
+{
+  size_t j;
+  char string_buffer[2048], *ptr;
+
+  ptr = string_buffer;
+  if(len > 1000) {
+    infof(data, "\t:buffer too long %ld", len);
```
bagder, line 198: I don't think the tab should be there. We don't use tabs anywhere else in infof() calls. This goes for the other infof() calls as well.

Since this is more of an internal error than anything else, does it really need an error message at all?

lib/doh.c
```diff
@@ -182,6 +186,33 @@ doh_write_cb(const void *contents, size_t size, size_t nmemb, void *userp)
...
+    *ptr++ = BIN2HEXDIGIT(buf[j] / 16);
+    *ptr++ = BIN2HEXDIGIT(buf[j] % 16);
+    if((ptr - string_buffer) > 2000) {
+      *ptr = '\0';
+      infof(data, "\t%s - truncated", string_buffer);
+      return;
+    }
+  }
+  *ptr = '\0';
+  infof(data, "\t%s", string_buffer);
```
bagder, line 211: This should also explain what this buffer is if the infof() should be kept.

lib/doh.c
```diff
@@ -407,6 +441,29 @@ struct Curl_addrinfo *Curl_doh(struct Curl_easy *data,
...
+  }
+  else {
+   /* the 16 is the most we get with a 5-digit port */
+   size_t qnamelen = strlen(hostname) + 16;
+
+    qname = (char *)calloc(1, qnamelen);
+    if(!qname)
+      goto error;
+    curl_msnprintf(qname, qnamelen, "_%d._https.%s",
+                     port % 0xffff, hostname);
```
bagder, line 457: The port number will be legit, no need to `% 0xffff` it. Which is wrong anyway since 0xffff is a legitimate port number.

Also: I prefer using `aprintf()` directly for things like this instead of calloc + msnprintf, as then you don't have to calculate the max space like this.

lib/doh.c
```diff
@@ -182,6 +186,33 @@ doh_write_cb(const void *contents, size_t size, size_t nmemb, void *userp)
   return realsize;
 }
 
+#ifdef USE_HTTPSRR
+static void local_print_buf(struct Curl_easy *data,
+                            unsigned char *buf, size_t len)
```
bagder, line 191: Maybe this should also be within `#ifdef CURLDEBUG` ?

lib/doh.c
```diff
@@ -888,8 +996,238 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+ * https://tools.ietf.org/html/rfc1035#section-3.1
+ *
+ * The input buffer pointer will be modified so it points to
+ * just after the end of the DNS name encoding on output. (And
+ * that's why it's an "unsigned char **" :-)
+ */
+static int local_decode_rdata_name(unsigned char **buf, size_t *remaining,
+                                   char **dnsname)
+{
+    unsigned char *cp = NULL;
```
bagder, line 1024: Not two-space indentation.

lib/doh.c
```diff
@@ -923,9 +1267,9 @@ CURLcode Curl_doh_is_resolved(struct Curl_easy *data,
       if(!p->dnstype)
         continue;
       rc[slot] = doh_decode(Curl_dyn_uptr(&p->serverdoh),
-                            Curl_dyn_len(&p->serverdoh),
-                            p->dnstype,
-                            &de);
+                              Curl_dyn_len(&p->serverdoh),
+                              p->dnstype,
+                              &de);
```
bagder, line 1272: ?

lib/doh.c
```diff
@@ -888,8 +996,238 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+
+  *alpns = oval;
+  return CURLE_OK;
+err:
+  if(oval)
+    free(oval);
+  return CURLE_BAD_CONTENT_ENCODING;
 }
 
+static int Curl_doh_decode_httpsrr(unsigned char *rrval, size_t len,
```
bagder, line 1113: Return `CURLcode` here?

lib/doh.c
```diff
@@ -888,8 +996,238 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+      rem -= (clen + 1);
+      clen = *cp++;
+    }
+    *buf = cp;
+    *remaining = rem - 1;
+    *dnsname = thename;
+    return CURLE_OK;
+}
+
+static int local_decode_rdata_alpn(unsigned char *rrval, size_t len,
```
bagder, line 1065: Return `CURLcode` ?

lib/doh.c
```diff
@@ -888,8 +996,238 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+ * @return is 1 for success, error otherwise
+ *
+ * The encoding here is defined in
+ * https://tools.ietf.org/html/rfc1035#section-3.1
+ *
+ * The input buffer pointer will be modified so it points to
+ * just after the end of the DNS name encoding on output. (And
+ * that's why it's an "unsigned char **" :-)
+ */
+static int local_decode_rdata_name(unsigned char **buf, size_t *remaining,
```
bagder, line 1021: Return `CURLcode` ?

lib/doh.c
```diff
@@ -969,6 +1313,20 @@ CURLcode Curl_doh_is_resolved(struct Curl_easy *data,
     } /* address processing done */
 
     /* Now process any build-specific attributes retrieved from DNS */
+#ifdef USE_HTTPSRR
+    if(de.numhttps_rrs > 0 && result == CURLE_OK && *dnsp) {
+      int rv;
```
bagder, line 1318:
```suggestion
      CURLcode result;
```

We try to use the variable name `result` for the generic local variable that holds return code from functions returning `CURLcode`.

lib/doh.c
```diff
@@ -969,6 +1313,20 @@ CURLcode Curl_doh_is_resolved(struct Curl_easy *data,
...
+      struct Curl_https_rrinfo *hrr = NULL;
+      rv = Curl_doh_decode_httpsrr(de.https_rrs->val, de.https_rrs->len, &hrr);
+      if(rv != CURLE_OK) {
```
bagder, line 1321: We typically check for errors without comparing and just do it if `result` is non-zero:

```suggestion
      if(result) {
```

lib/doh.h
```diff
@@ -97,12 +105,44 @@ struct dohaddr {
...
+
+/*
+ * These may need escaping when found within an alpn string
+ * value.
+ */
+#define COMMA_CHAR                    ','
+#define BACKSLASH_CHAR                '\\'
+
+struct dohhttps_rr {
+  size_t len; /* raw encoded length */
```
bagder, line 131: What is the maximum length of these RR fields? Maybe we can save a few bytes by using something smaller than `size_t` ?

lib/doh.c
```diff
@@ -888,8 +996,238 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+    clen = *cp++;
+    if(clen == 0) {
+      /* special case - return "." as name */
+      thename[0] = '.';
+      thename[1] = 0x00;
+    }
+    while(clen) {
+      if(clen > rem) {
+        free(thename);
+          return CURLE_OUT_OF_MEMORY;
```
bagder, line 1046: Odd indentation

lib/doh.c
```diff
@@ -888,8 +996,238 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+   * spec here is as per draft-ietf-dnsop-svcb-https, section-7.1.1
+   * encoding is catenated list of strings each preceded by a one
+   * octet length
+   * output is comma-sep list of the strings
+   * implementations may or may not handle quoting of comma within
+   * string values, so we migth see a comma within the wire format
+   * version of a string, in which case we'll precede that by a
+   * backslash - same goes for a backslash character
+   */
+  int remaining = (int) len;
```
bagder, line 1078: Would this not avoid the need for several typecasts if you make it a `size_t`  here?

bagder:
Some of the CI test failures happen because `docs/options-in-versions` and `docs/libcurl/symbols-in-version` need updating

include/curl/curl.h
```diff
@@ -2210,6 +2214,18 @@ typedef enum {
   /* set a specific client IP for HAProxy PROXY protocol header? */
   CURLOPT(CURLOPT_HAPROXY_CLIENT_IP, CURLOPTTYPE_STRINGPOINT, 323),
 
+  /* ECH status: 1L to enable, 0L otherwise */
+  CURLOPT(CURLOPT_ECH_STATUS, CURLOPTTYPE_LONG, 324),
+
+  /* ECH hard status: 1L to enable, 0L otherwise */
+  CURLOPT(CURLOPT_ECH_HARD_STATUS, CURLOPTTYPE_LONG, 327),
```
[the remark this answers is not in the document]
sftcd: sure, if that's how it's preferred happy to do it

include/curl/curl.h
```diff
@@ -3063,7 +3079,7 @@ struct curl_version_info_data {
   const char *version;      /* LIBCURL_VERSION */
   unsigned int version_num; /* LIBCURL_VERSION_NUM */
   const char *host;         /* OS/host/cpu/machine when configured */
-  int features;             /* bitmask, see defines below */
+  long features;             /* bitmask, see defines below */
```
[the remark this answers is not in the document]
sftcd: ack, will look into ``feature_names``

lib/connect.c
```diff
@@ -693,7 +693,12 @@ static CURLcode is_connected(struct Curl_cfilter *cf,
   for(i = 0; i < sizeof(ctx->baller)/sizeof(ctx->baller[0]); i++) {
     struct eyeballer *baller = ctx->baller[i];
     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
+#ifdef USE_HTTPSRR
+                /* not an HTTPS RR issue but may as well fix */
+                baller?baller->name:"NULL",
+#else
                 baller?baller->name:NULL,
+#endif
```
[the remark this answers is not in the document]
sftcd: It just removed a compile warning for me. Now that I know better what warning's are what, happy to revert;-)

include/curl/curl.h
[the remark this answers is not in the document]
sftcd: think that's done, will be in next push

include/curl/curl.h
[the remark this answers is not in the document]
sftcd: I did that (I thnik) but get a warning about (1<<31) being too big, though things work ok. I guess I'm wondering if it's worthwhile have ECH as a "feature" like that anyway.

lib/doh.c
sftcd: sure, did that

lib/doh.c
sftcd: yep, could be cleaned up, but is there an existing function to use instead of adding a new ``local_print_buf()`` ? If so, I'd use that and get rid of this entirely.

lib/doh.c
sftcd: was helpful for me debugging to see the RR and ECHConfig values, but got rid of the tab and see above also

lib/doh.c
sftcd: ack

lib/doh.c
sftcd: fair point, did that

lib/doh.c
sftcd: did that, thanks

lib/doh.c
sftcd: done

lib/doh.c
sftcd: fixed

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.h
sftcd: 16  bits max as per [this](https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https#section-2.2) - I changed to ``uint16_t`` for the struct.

Is the concern to make the array of those structs smaller? If so, then the above is probably an ok change though those values get mapped to ``size_t`` in some API calls, which I'm guessing is ok?

lib/doh.c
sftcd: yep, fixed.

lib/doh.c
sftcd: I think ``int remaining`` is maybe safer here - with some badly encoded ALPN values, ``remaining``  might go negative so not sure the typecasts are avoidable though doubtless the code could look nicer.

sftcd:
> Some of the CI test failures happen because `docs/options-in-versions` and `docs/libcurl/symbols-in-versions` need updating

I'll push a version with the mostly trivial changes indicated above. (I hit the "resolve" button on those that seem most trivial:-) I'll then rebase to try make the CI situation better and push another.

Thanks for all the good comments meanwhile!

lib/connect.c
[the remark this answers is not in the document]
sftcd: As an FYI, the compile warning happens for me when I don't enable debug and then I get:
```
In file included from sendf.h:29,
                 from connect.c:57:
In function 'is_connected',
    inlined from 'cf_he_connect' at connect.c:906:16:
connect.c:695:27: warning: '%s' directive argument is null [-Wformat-overflow=]
  695 |     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
curl_trc.h:123:38: note: in definition of macro 'CURL_TRC_CF'
  123 |          Curl_trc_cf_infof(data, cf, __VA_ARGS__); } while(0)
      |                                      ^~~~~~~~~~~
connect.c: In function 'cf_he_connect':
connect.c:695:28: note: format string is defined here
  695 |     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
      |                            ^~
```

include/curl/curl.h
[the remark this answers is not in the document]
sftcd: Since that (1<<31) seems to make many of the CI builds unhappy, and since DoH and HTTPS RR support aren't listed as "features," I'll take that out for now. Can be added back if/when I figure out how to do it right:-)

sftcd:
So now that I'm clearing up some of the CI chaff, an interesting question arises - if we're not doing ECH, then when, if ever, should we make a DNS query for an HTTTPS RR? Probably, for now, never, but that'll likely change if/as support for other uses for HTTPS RRs is added.

sftcd:
The current DoH code seems to setup a new TLS session with the DoH-server for each DoH query. That probably makes no difference if the client is only making one request for one A record, but if we're now sometimes requesting A, AAAA and HTTPS RRs, it may be worth refactoring that code to use just one TLS session. Could do that but probably better as a separate PR sometime. Would need to understand though if it were possible/desirable to preserve the TLS session with the DoH-server across the runtime of an application using the library, which'd be more complex, but sometimes much more efficient.

lib/doh.c
bagder: libcurl does not have any generic "hex encode" function, but I agree it would make sense. I'll make one in a separate PR! See  #11990

lib/doh.c
```diff
@@ -411,6 +448,35 @@ struct Curl_addrinfo *Curl_doh(struct Curl_easy *data,
...
+   * request when ECH is enabled in the build and is being used for
+   * the curl operation.
+   */
+# ifdef USE_ECH
+  if(data->set.tls_enable_ech
+     || data->set.tls_enable_ech_hard) {
+    if(port == 443)
+      qname = strdup(hostname);
+    else
+      qname = aprintf("_%d._https.%s", port, hostname);
```
bagder, line 470: Remember that this can return NULL and you want to handle that error.

lib/doh.c
```diff
@@ -411,6 +448,35 @@ struct Curl_addrinfo *Curl_doh(struct Curl_easy *data,
...
+   * making this request breaks a bunch of DoH tests, e.g. test2100,
+   * where the addiitonal request doesn't match the pre-cooked data
+   * files, so there's a bit of work attached to making the request
+   * in a non-ECH use-case. For the present, we'll only make the
+   * request when ECH is enabled in the build and is being used for
+   * the curl operation.
+   */
+# ifdef USE_ECH
+  if(data->set.tls_enable_ech
+     || data->set.tls_enable_ech_hard) {
```
bagder, line 466: shouldn't hard imply that `data->set.tls_enable_ech` is set? Possibly even using the same variable but use different values for the different levels?

sftcd:
On 29/09/2023 17:13, Daniel Stenberg wrote:
> While you were working on this, we changed how these references should be formatted and they are now verified to follow this format by test 1173. Now they should be:
> 
> ~~~
> .BR CURLOPT_ECH_CONFIG (3),
> .BR CURLOPT_ECH_STATUS (3)

Aha! Thanks for that - it was doing my head in trying to
figure why the local tests differed from the CI builds:-)

I'll process the your other comments shortly, but mostly
they look like good things to do.

Cheers,
S.

docs/libcurl/opts/CURLOPT_ECH_PUBLIC.3
```diff
@@ -0,0 +1,61 @@
...
+  curl_easy_perform(curl);
+}
+.fi
+.SH AVAILABILITY
+Added in 8.4.0
+.SH RETURN VALUE
+Returns CURLE_OK on success or CURLE_OUT_OF_MEMORY if there was insufficient
+heap space.
+.SH "SEE ALSO"
+.BR CURLOPT_ECH_CONFIG "(3), " CURLOPT_ECH_STATUS "(3), "
```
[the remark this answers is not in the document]
sftcd: Done.

docs/cmdline-opts/ech.d
```diff
@@ -0,0 +1,15 @@
...
+Added: x.xx.x
+Category: tls ECH
+Example: --ech $URL
+See-also: echconfig echpublic ech-hard doh-url
+Multi: single
+---
+Enable Encrypted Client Hello (ECH) for use with the TLS session, if
+pssible, i.e. this is opportunistic.
+
+This option only works with TLS 1.3.
```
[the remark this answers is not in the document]
sftcd: done

docs/libcurl/symbols-in-versions
```diff
@@ -339,6 +339,8 @@ CURLE_URL_MALFORMAT_USER        7.1           7.17.0
 CURLE_USE_SSL_FAILED            7.17.0
 CURLE_WEIRD_SERVER_REPLY        7.51.0
 CURLE_WRITE_ERROR               7.1
+CURLECH_ENABLE                  8.4.0-DEV
+CURLECH_HARD                    8.4.0-DEV
```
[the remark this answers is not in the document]
sftcd: Done

docs/cmdline-opts/echpublic.d
```diff
@@ -0,0 +1,16 @@
+c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+SPDX-License-Identifier: curl
+Long: echpublic
+Arg: <name>
+Help: TLS Encrypted Client Hello (ECH)
+Added: x.xx.x
+Category: tls ECH
+Example: --echpubli $PUBLIC_NAME $URL
```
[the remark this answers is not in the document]
sftcd: see latest commit

docs/cmdline-opts/echconfig.d
```diff
@@ -0,0 +1,17 @@
+c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+SPDX-License-Identifier: curl
+Long: echconfig
```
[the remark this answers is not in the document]
sftcd: see latest commit

docs/cmdline-opts/ech-hard.d
```diff
@@ -0,0 +1,15 @@
...
+Arg: <name>
+Help: TLS Encrypted Client Hello (ECH)
+Added: x.xx.x
+Category: tls ECH
+Example: --ech-hard $URL
+See-also: echconfig echpublic ech doh-url
+Multi: single
+---
+Enable Encrypted Client Hello (ECH) for use with the TLS session, and
+fail if that's not pssible.
```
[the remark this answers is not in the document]
sftcd: see latest commit (along these lines)

docs/libcurl/curl_easy_setopt.3
```diff
@@ -717,6 +717,16 @@ if(curl) {
   curl_easy_cleanup(curl);
 }
 .fi
+.SH ENCRYPTED CLIENT HELLO OPTIONS
+.IP CURLOPT_ECH_STATUS
```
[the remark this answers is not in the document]
sftcd: ah, missed that one, might do it next time

sftcd:
So the latest commit https://github.com/curl/curl/pull/11922/commits/fb2b2cd286ff7d80837659feea50a03e92378e20 re-does the command line args as follows:

- ``--ech <config>`` - the ``config`` value can be one of:
    - ``false`` says to not attempt ECH
    - ``true`` says to attempt ECH, if possible
    - ``grease`` if attempting ECH is not possible, then send a GREASE'd ECH extension
    - ``hard`` hard-fail the connection if ECH cannot be attempted
    - a base64 encoded ECHConfigList, rather than one accessed from the DNS
- ``--echpublic`` over-ride the ``public_name`` from an ECHConfigList

I think that's along the lines suggested above. 

Still needs some testing and happy to tweak more.

sftcd:
Given there's a preference for fewer options, the [latest push](https://github.com/curl/curl/pull/11922/commits/73358484a57ead5064cc5349681c33178c097595) reduces the command line to one option only as in:

- ``--ech <config>`` - the ``config`` value can be one of:
    - ``false`` says to not attempt ECH
    - ``true`` says to attempt ECH, if possible
    - ``grease`` if attempting ECH is not possible, then send a GREASE ECH extension
    - ``hard`` hard-fail the connection if ECH cannot be attempted
    - ``ecl:<b64value>`` a base64 encoded ECHConfigList, rather than one accessed from the DNS
    - ``pn:<name>`` over-ride the ``public_name`` from an ECHConfigList

lib/doh.c
sftcd: Now making use of the #11990 function.

configure.ac
```diff
@@ -4378,18 +4391,28 @@ if test "x$want_ech" != "xno"; then
...
-    AC_CHECK_FUNCS(SSL_get_ech_status,
-      ECH_SUPPORT="ECH support available (OpenSSL with SSL_get_ech_status)"
+    AC_CHECK_FUNCS(SSL_ech_set1_echconfig,
+      ECH_SUPPORT="ECH support available via OpenSSL with SSL_ech_set1_echconfig"
+      ECH_ENABLED=1)
+  fi
+  dnl check for boringssl equivalent
+  if test "x$OPENSSL_ENABLED" = "x1"; then
+    AC_CHECK_FUNCS(SSL_set1_ech_config_list,
+      ECH_SUPPORT="ECH support available via boringssl with SSL_set1_ech_ech_config_list"
```
vszakats, line 4403:
```suggestion
      ECH_SUPPORT="ECH support available via boringssl with SSL_set1_ech_config_list"
```

sftcd: thanks, fixed.

vszakats:
I hope it's not too late, but I made this patch for CMake earlier:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 96a1aeb21..cf188437e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -732,6 +732,28 @@ if(USE_MSH3)
   list(APPEND CURL_LIBS ${MSH3_LIBRARIES})
 endif()
 
+option(USE_HTTPSRR "Enable HTTPS RR support for ECH (experimental)" OFF)
+option(USE_ECH "Enable ECH support" OFF)
+if(USE_ECH)
+  if(USE_OPENSSL OR USE_WOLFSSL)
+    # Be sure that the OpenSSL/wolfSSL library actually supports ECH.
+    if(NOT DEFINED HAVE_OPENSSL_ECH)
+      if(USE_OPENSSL AND HAVE_BORINGSSL)
+        openssl_check_symbol_exists(SSL_set1_ech_config_list "openssl/ssl.h" HAVE_OPENSSL_ECH)
+      elseif(USE_OPENSSL)
+        openssl_check_symbol_exists(SSL_ech_set1_echconfig "openssl/ssl.h" HAVE_OPENSSL_ECH)
+      elseif(USE_WOLFSSL)
+        openssl_check_symbol_exists(wolfSSL_CTX_GenerateEchConfig "wolfssl/options.h;wolfssl/ssl.h" HAVE_OPENSSL_ECH)
+      endif()
+    endif()
+    if(NOT HAVE_OPENSSL_ECH)
+      message(FATAL_ERROR "ECH support missing in OpenSSL/BoringSSL/wolfSSL")
+    endif()
+  else()
+    message(FATAL_ERROR "ECH requires OpenSSL, BoringSSL or wolfSSL")
+  endif()
+endif()
+
 if(NOT CURL_DISABLE_SRP AND (HAVE_GNUTLS_SRP OR HAVE_OPENSSL_SRP))
   set(USE_TLS_SRP 1)
 endif()
@@ -1563,6 +1585,7 @@ _add_if("TLS-SRP"       USE_TLS_SRP)
 # TODO option --with-nghttp2 tests for nghttp2 lib and nghttp2/nghttp2.h header
 _add_if("HTTP2"         USE_NGHTTP2)
 _add_if("HTTP3"         USE_NGTCP2 OR USE_QUICHE)
+_add_if("ECH"           USE_ECH)
 _add_if("MultiSSL"      CURL_WITH_MULTI_SSL)
 # TODO wolfSSL only support this from v5.0.0 onwards
 _add_if("HTTPS-proxy"   SSL_ENABLED AND (USE_OPENSSL OR USE_GNUTLS
diff --git a/lib/curl_config.h.cmake b/lib/curl_config.h.cmake
index 0bfb45796..b38a62359 100644
--- a/lib/curl_config.h.cmake
+++ b/lib/curl_config.h.cmake
@@ -814,3 +814,9 @@ ${SIZEOF_TIME_T_CODE}
 
 /* Define to 1 to enable TLS-SRP support. */
 #cmakedefine USE_TLS_SRP 1
+
+/* Define to 1 to force HTTPS RR support for ECH. */
+#cmakedefine USE_HTTPSRR 1
+
+/* Define to 1 if ECH support is available. */
+#cmakedefine USE_ECH 1
```

Short of an ECH-enabled TLS library, it's only lightly tested.

sftcd:
> I hope it's not too late, but I made this patch for CMake earlier:
> 
> Short of an ECH-enabled TLS library, it's only lightly tested.

Oh thanks, I'll compare it with what I just did.

BTW - does the cmake build work cleanly with the ``master`` branch? I saw some issues with that. (My flailing about notes are [here](https://github.com/sftcd/curl/blob/ECH-experimental/docs/ECH.md#building-with-cmake):-)

vszakats:
For local test builds, cmake works (for me) with master, just by omitting `USE_MANUAL`. In this case the build creates a dummy `tool_hugehelp.c` automatically, and curl doesn't have a manual embedded.

Enabling it can be tricky with out of tree builds.

To prevent rebuilding the manual I use `-DCMAKE_C_FLAGS=-DUSE_MANUAL=1` (instead of `-DUSE_MANUAL=ON`).

Then supply a pre-built or dummy `tool_hugehelp.c` copy myself, e.g. with `cp -p src/tool_hugehelp.c.cvs build/src/tool_hugehelp.c`:

https://github.com/curl/curl-for-win/blob/06edf3f3b6d81c1efc419a5933a93fe6ffa2816c/curl-cmake.sh#L352-L359

sftcd:
> I hope it's not too late, but I made this patch for CMake earlier:

Thanks again. Since yours was cleaner than mine, I've now gone more or less went with your suggested patch.

sftcd:
On 18/10/2023 16:47, 南宫雪珊 wrote:
> What is the meaning of `--ech false`, why not delete `--ech`?

In my local setup I have "ech=TRUE" in ``$HOME/.curlrc`` so this
provides a way to turn that off from the command line, if needed.

S.

vvb2060:
https://github.com/curl/curl/blob/9742dba5234ae8a717468856382a40991a18df74/docs/cmdline-opts/ech.d#L10

What if i want `hard` + base64 ECHConfigList?
My thoughts: --ech true/false/hard/grease[:string]
The string is optional, if it has a dot then it is public_name, otherwise it is base64 ECHConfigList

sftcd:
> https://github.com/curl/curl/blob/9742dba5234ae8a717468856382a40991a18df74/docs/cmdline-opts/ech.d#L10
> 
To be honest, I'm not sure what "Multi: single" means, I probably just copied from some other ``.d`` file.

> What if i want `hard` + base64 ECHConfigList? 

At present supplying ``--ech hard --ech ecl:<b64value>``  does that. Basically it'll use the "last" true/false/hard/grease value, then "last" public_name and the "last" base64-encoded ECHConfigList (if the latter two are supplied).

> My thoughts: --ech true/false/hard/grease[:string] The string is optional, if it has a dot then it is public_name, otherwise it is base64 ECHConfigList

Not sure what that'd be better? Happy to change to something that is better of course but would like to understand why first.  Switching ``[:string]`` based on the presence/absence of a dot also seems a bit brittle and who knows there might be a use-case for ``--ech pn:localhost`` or something else that isn't a FQDN.

sftcd:
Thanks for the comments btw - I take it from those that, at minimum, more text is needed in ``ech.d`` to clarify the above. I'll add something in a while.

docs/cmdline-opts/ech.d
```diff
@@ -0,0 +1,35 @@
...
+- "pn:<name>": a name to use to over-ride the public_name field of an
+   ECHConfigList
+
+When multiple ``--ech`` options are supplied then the most-recent
+value for true/false/hard/grease will be used, as will the most-recent
+``ecl:<b64string>`` value, and ``pn:<name>`` value, if either of
+those were provided.
+
+ECH only works with TLS 1.3 and also requires using
+DoH or else providing an ECHConfigList on the command line.
```
bagder, line 35:
```suggestion
DoH or providing an ECHConfigList on the command line.
```

docs/libcurl/opts/CURLOPT_ECH.3
```diff
@@ -0,0 +1,74 @@
...
+.\" * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+.\" * KIND, either express or implied.
+.\" *
+.\" * SPDX-License-Identifier: curl
+.\" *
+.\" **************************************************************************
+.\"
+.TH CURLOPT_ECH 3 "30 Sep 2023" libcurl libcurl
+.SH NAME
+CURLOPT_ECH \- control configuration for Encrypted Client Hello
```
bagder, line 27:
```suggestion
CURLOPT_ECH \- configuration for Encrypted Client Hello
```

docs/libcurl/opts/CURLOPT_ECH.3
```diff
@@ -0,0 +1,74 @@
...
+Turns off ECH.
+.IP grease
+Instructs client to emit a GREASE'd ECH extension.
+.IP true
+Instructs client to attempt ECH, if possible, but to not fail if attempting ECH is not possible.
+.IP hard
+Instructs client to attempt ECH and fail if if attempting ECH is not possible.
+.IP ecl:<base64-value>
+If the string starts with "ecl:" then the remainder of the string should be a base64-encoded
+ECHConfigList that will be used for ECH rather than attempting to download such a value from
```
bagder, line 48:
```suggestion
ECHConfigList that is used for ECH rather than attempting to download such a value from
```

docs/libcurl/opts/CURLOPT_ECH.3
```diff
@@ -0,0 +1,74 @@
...
+the DNS.
+.IP pn:<name>
+If the string starts with "pn:" then the remainder of the string should be a DNS/host name
+that will be used to over-ride the public_name field of the ECHConfigList that will be used
```
bagder, line 52:
```suggestion
that will be used to over-ride the public_name field of the ECHConfigList that is used
```

docs/libcurl/opts/CURLOPT_ECH.3
```diff
@@ -0,0 +1,74 @@
...
+for ECH.
+.SH DEFAULT
+There is no default value, meaning ECH is disabled if no
+configuration is supplied.
```
bagder, line 56:
```suggestion
NULL, meaning ECH is disabled.
```

docs/libcurl/opts/CURLOPT_ECH.3
```diff
@@ -0,0 +1,74 @@
...
+.SH PROTOCOLS
+TLS, and requires TLS1.3.
```
bagder, line 58: It does beg the question: how about QUIC?

docs/libcurl/opts/CURLOPT_ECH.3
```diff
@@ -0,0 +1,74 @@
...
+CURL *curl = curl_easy_init();
+
+const char *config ="ecl:AED+DQA87wAgACB/RuzUCsW3uBbSFI7mzD63TUXpI8sGDTnFTbFCDpa+CAAEAAEAAQANY292ZXIuZGVmby5pZQAA";
+if(curl) {
+  curl_easy_setopt(curl, CURLOPT_ECH, config);
+  curl_easy_perform(curl);
+}
+.fi
+.SH AVAILABILITY
+Added in 8.4.0
```
bagder, line 70: This seems likely to happen in 8.6.0 as things look.

include/curl/curl.h
```diff
@@ -640,6 +640,9 @@ typedef enum {
   CURLE_PROXY,                   /* 97 - proxy handshake error */
   CURLE_SSL_CLIENTCERT,          /* 98 - client-side certificate required */
   CURLE_UNRECOVERABLE_POLL,      /* 99 - poll/select returned fatal error */
+#ifdef USE_ECH
+  CURLE_ECH_REQUIRED      ,      /* 100 - ECH tried but failed */
```
bagder, line 644:
```suggestion
  CURLE_ECH_FAILED      ,      /* 100 - ECH tried but failed */
```

Mostly because this makes the error code spell out what actually happened.

lib/doh.c
```diff
@@ -42,9 +42,17 @@
 #include "curl_printf.h"
 #include "curl_memory.h"
 #include "memdebug.h"
+#ifdef USE_HTTPSRR
+# include "escape.h"
+#endif
```
bagder, line 47: No need to ifdef this for just an include. Minimizing the amount of ifdefs makes the code easier on the eye.

lib/doh.c
```diff
@@ -42,9 +42,17 @@
...
 
 #define DNS_CLASS_IN 0x01
 
+#ifdef USE_HTTPSRR
+/* local_print_buf truncates if the hex string will be more than this */
+# define LOCAL_PB_HEXMAX 400
+#endif
```
bagder, line 54: Same here, just for a define that is not used otherwise anyway I don't think ifdef is needed but instead just makes it harder to read.

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+      goto err;
+    memcpy(op, cp, tlen);
+    op += tlen;
+    if((size_t)(op - oval) >= ((size_t)2 * len))
+      goto err;
+    *op++ = COMMA_CHAR;
+    cp += tlen;
+    remaining -= (int)tlen;
+  }
+  oval[strlen(oval) - 1] = '\0';
```
bagder, line 1120: Can I propose that the above loop and decode is done into a dynbuf instead of this custom calloc + memcpy logic? dynbufs generally make it harder to mess up the memory fiddling.

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+        goto err;
+    }
+    if(pcode == HTTPS_RR_CODE_NO_DEF_ALPN)
+      lhrr->no_def_alpn = TRUE;
+    if(pcode == HTTPS_RR_CODE_IPV4) {
+      lhrr->ipv4hints = calloc(1, plen);
+      if(!lhrr->ipv4hints)
+        goto err;
+      lhrr->ipv4hints_len = (size_t)plen;
+      memcpy(lhrr->ipv4hints, cp, plen);
```
bagder, line 1173: Maybe use `Curl_memdup() ` here?

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+    }
+    if(pcode == HTTPS_RR_CODE_ECH) {
+      lhrr->echconfiglist = calloc(1, plen);
+      if(!lhrr->echconfiglist)
+        goto err;
+      lhrr->echconfiglist_len = (size_t)plen;
+      memcpy(lhrr->echconfiglist, cp, plen);
```
bagder, line 1180: and here

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+    }
+    if(pcode == HTTPS_RR_CODE_IPV6) {
+      lhrr->ipv6hints = calloc(1, plen);
+      if(!lhrr->ipv6hints)
+        goto err;
+      lhrr->ipv6hints_len = (size_t)plen;
+      memcpy(lhrr->ipv6hints, cp, plen);
```
bagder, line 1187: and here

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+    pcode = (uint16_t)((*cp << 8) + (*(cp + 1)));
+    cp += 2;
+    plen = (uint16_t)((*cp << 8) + (*(cp + 1)));
+    cp += 2;
+    remaining -= 4;
+    if(pcode == HTTPS_RR_CODE_ALPN) {
+      if(local_decode_rdata_alpn(cp, plen, &lhrr->alpns) != CURLE_OK)
+        goto err;
+    }
+    if(pcode == HTTPS_RR_CODE_NO_DEF_ALPN)
```
bagder, line 1166:
```suggestion
    else if(pcode == HTTPS_RR_CODE_NO_DEF_ALPN)
```

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+      lhrr->no_def_alpn = TRUE;
+    if(pcode == HTTPS_RR_CODE_IPV4) {
```
bagder, line 1168:
```suggestion
    else if(pcode == HTTPS_RR_CODE_IPV4) {
```

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+      lhrr->ipv4hints = calloc(1, plen);
+      if(!lhrr->ipv4hints)
+        goto err;
+      lhrr->ipv4hints_len = (size_t)plen;
+      memcpy(lhrr->ipv4hints, cp, plen);
+    }
+    if(pcode == HTTPS_RR_CODE_ECH) {
```
bagder, line 1175:
```suggestion
    else if(pcode == HTTPS_RR_CODE_ECH) {
```

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+      lhrr->echconfiglist = calloc(1, plen);
+      if(!lhrr->echconfiglist)
+        goto err;
+      lhrr->echconfiglist_len = (size_t)plen;
+      memcpy(lhrr->echconfiglist, cp, plen);
+    }
+    if(pcode == HTTPS_RR_CODE_IPV6) {
```
bagder, line 1182:
```suggestion
    else if(pcode == HTTPS_RR_CODE_IPV6) {
```

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+      if(!lhrr->ipv6hints)
+        goto err;
+      lhrr->ipv6hints_len = (size_t)plen;
+      memcpy(lhrr->ipv6hints, cp, plen);
+    }
+    if(plen > 0 && plen <= remaining) {
+      cp += plen;
+      remaining -= plen;
+    }
+  }
```
bagder, line 1193: If `remaining` is non-zero at this point, something is likely wrong in the parser, right? Maybe you can just add a `DEBUGASSERT(!remaining);` there to trigger a debug build to notice this for further research?

lib/doh.c
```diff
@@ -894,8 +1014,239 @@ UNITTEST void de_cleanup(struct dohentry *d)
...
+  return CURLE_OUT_OF_MEMORY;
+}
+
+# ifdef CURLDEBUG
+static void local_print_httpsrr(struct Curl_easy *data,
+                                struct Curl_https_rrinfo *hrr)
+{
+  if(!hrr) {
+    infof(data, "NULL HTTPS_RR provided!");
```
bagder, line 1214: Can this actually happen? It seems like this should be a `DEBUGASSERT()` ?

lib/doh.h
```diff
@@ -51,7 +54,12 @@ typedef enum {
   DNS_TYPE_NS = 2,
   DNS_TYPE_CNAME = 5,
   DNS_TYPE_AAAA = 28,
+#ifdef USE_HTTPSRR
+  DNS_TYPE_DNAME = 39,           /* RFC6672 */
+  DNS_TYPE_HTTPS = 65
+#else
   DNS_TYPE_DNAME = 39           /* RFC6672 */
+#endif
```
bagder, line 62: I think you can drop these ifdefs, the values can be there always.

lib/doh.h
```diff
@@ -88,6 +96,9 @@ int Curl_doh_getsock(struct connectdata *conn, curl_socket_t *socks);
 
 #define DOH_MAX_ADDR 24
 #define DOH_MAX_CNAME 4
+#ifdef USE_HTTPSRR
+#define DOH_MAX_HTTPS 4
+#endif
```
bagder, line 101: This define could be done unconditionally.

lib/doh.c
```diff
@@ -492,6 +557,26 @@ static DOHcode store_aaaa(const unsigned char *doh,
...
+                           struct dohentry *d,
+                           uint16_t len)
+{
+  /* silently ignore RRs over the limit */
+  if(d->numhttps_rrs < DOH_MAX_HTTPS) {
+    struct dohhttps_rr *h = &d->https_rrs[d->numhttps_rrs];
+    h->val = malloc(len);
+    if(!h->val)
+      return DOH_OUT_OF_MEM;
+    memcpy(h->val, &doh[index], len);
```
bagder, line 572: `Curl_memdup` maybe?

lib/doh.c
```diff
@@ -884,7 +991,20 @@ doh2ai(const struct dohentry *de, const char *hostname, int port)
...
+    case DNS_TYPE_A:
+      return "A";
+    case DNS_TYPE_AAAA:
+      return "AAAA";
+    case DNS_TYPE_HTTPS:
+      return "HTTPS";
+    default:
+       return "unknown";
+  }
+#endif
```
bagder, line 1007:
```suggestion
  switch(dnstype) {
    case DNS_TYPE_A:
      return "A";
    case DNS_TYPE_AAAA:
      return "AAAA";
#ifndef USE_HTTPSRR
    case DNS_TYPE_HTTPS:
      return "HTTPS";
#endif
    default:
       return "unknown";
  }
```

bagder (commented):
A bunch of comments of mine. Most of them are plain *opinions* so do take them as proposals, not strict requirements.

docs/cmdline-opts/ech.d
sftcd: done

docs/libcurl/opts/CURLOPT_ECH.3
sftcd: done

docs/libcurl/opts/CURLOPT_ECH.3
sftcd: done

docs/libcurl/opts/CURLOPT_ECH.3
sftcd: done

docs/libcurl/opts/CURLOPT_ECH.3
sftcd: done

docs/libcurl/opts/CURLOPT_ECH.3
sftcd: To be honest: no idea;-( 

Since QUIC is still very much in-development within OpenSSL (IIUC), I've not gone near that yet. It is a good question though I suspect will come up once I make an ECH PR for OpenSSL. I've also no real idea how ECH is or is not supported for QUIC with boringssl nor wolfSSL so that's also to be investigated.

So I think the current answer is that's a quite large TODO. I don't know if that'd be better handled via a separate PR (both for curl and OpenSSL), but that's my current guess. If that's an ok guess, then probably the best answer for now is: "later."

docs/libcurl/opts/CURLOPT_ECH.3
sftcd: done (for now:-)

include/curl/curl.h
sftcd: I need to look at this some more. The ECH spec defines the new "ech_required" alert that clients use to close the connection in this case, which is why I used that. OTOH, boringssl does differentiate between that and an "ech failed" API error code, so there's a possibly silly divergence between my OpenSSL fork and boringssl there. As @davidben is smarter than me he's probably correct there, so that's one I'll revisit, but not right now.

So, I've left this as-is for the moment.

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: sounds right, but I'll revisit later, and there's still an escaping TODO as well, so leaving this unresolved for now.

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done

lib/doh.h
sftcd: done

lib/doh.h
sftcd: done

lib/doh.c
sftcd: done

lib/doh.c
sftcd: done (well modulo ``s/ifndef/ifdef/`` in the middle of the switch)

lib/doh.c
sftcd: took a stab at that in https://github.com/curl/curl/pull/11922/commits/ecf5952a8b668d59f7da4d082c91d555f3689aec - could be better if ``dynbuf`` percolates upwards in the data structure but this seems at least to work:-)

include/curl/curl.h
sftcd: I had a look at this. While I could make a change like that, I reckon it's likely better to leave this as-is for now on the basis that once there's a PR submitted to the OpenSSL maintainers, I'm pretty sure they'll want changes in OpenSSL error codes, maybe that one, but probably others as well, and those changes would likely percolate up to the curl code too, so a change now probably isn't worth it for now.

sftcd:
A note for the curious: I've just rebased with the main branch and will be fixing things next few days so don't worry if the inevitable CI check fails occur for a while. Will update here with comments when this is suited for more review, with the target of offering it up as an experimental feature when the next feature window opens, which I guess will be sometime in Feb. (Reason to consider merging it then: the IANA codepoints for ECH have now been registered in the TLS registries so if anyone did go distributing/using a version of curl with this ECH experimental feature enabled that shouldn't cause any interop harms even in the longer term.)

vvb2060:
```
curl --ech true --doh-url https://1.0.0.1/dns-query https://crypto.cloudflare.com/cdn-cgi/trace -v

......

* a DoH request is completed, 2 to go
* a DoH request is completed, 1 to go
* Connection #1 to host 1.0.0.1 left intact
* a DoH request is completed, 0 to go
* DoH Host name: crypto.cloudflare.com
* TTL: 148 seconds
* DoH A: 162.159.138.85
* DoH A: 162.159.137.85
* DoH AAAA: 2606:4700:0007:0000:0000:0000:a29f:8a55
* DoH AAAA: 2606:4700:0007:0000:0000:0000:a29f:8955
* DoH HTTPS RR: length 142
* Some HTTPS RR to process
* Host crypto.cloudflare.com:443 was resolved.
* IPv6: 2606:4700:7::a29f:8a55, 2606:4700:7::a29f:8955
* IPv4: 162.159.138.85, 162.159.137.85
*   Trying 162.159.138.85:443...
* Connected to crypto.cloudflare.com (162.159.138.85) port 443
* ALPN: curl offers h2,http/1.1
* ECH: ECHConfig from DoH HTTPS RR
* ECH: imported ECHConfigList of length 71
* TLSv1.2 (OUT), TLS Unknown, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
*  CAfile: none
*  CApath: /system/etc/security/cacerts
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (OUT), TLS alert, ECH required (633):
* ECH: no retry_configs (rv = 0)
* BoringSSL: error:1000013f:SSL routines:OPENSSL_internal:ECH_REJECTED
* Closing connection
curl: (101) BoringSSL: error:1000013f:SSL routines:OPENSSL_internal:ECH_REJECTED
```

curl will not fall back to no ech when connection fails

vvb2060:
```
C/C++: [armeabi-v7a] Compile thumb  : curl_static <= openssl.c
C/C++: /home/runner/work/curl-android/curl-android/tool/src/main/native/../../../../curl/src/main/native/curl/lib/vtls/openssl.c:3904:70: warning: format specifies type 'long' but the argument has type 'size_t' (aka 'unsigned int') [-Wformat]
C/C++:             infof(data, "ECH: imported ECHConfigList of length %ld", elen);
C/C++:                                                                ~~~   ^~~~
C/C++:                                                                %zu
C/C++: /home/runner/work/curl-android/curl-android/tool/src/main/native/../../../../curl/src/main/native/curl/lib/curl_trc.h:76:27: note: expanded from macro 'infof'
C/C++:          Curl_infof(data, __VA_ARGS__); } while(0)
C/C++:                           ^~~~~~~~~~~
C/C++: 1 warning generated.
```

sftcd:
> ```
> curl --ech true --doh-url https://1.0.0.1/dns-query https://crypto.cloudflare.com/cdn-cgi/trace -v
> 
> ......

> curl will not fall back to no ech when connection fails

Odd. ECH should work for that URL. I just checked and it did for a current build with OpenSSL and and older one (Jan 15th) with boringssl. I've not done a boring build since I rebased with 8.6.1 but will. And I'll get back when I've done that...

Any other info about what platform you're on, whether a curl+ECH+OpenSSL build works etc?

vvb2060:
This is not a problem with boringssl, cloudflare network rejects ech when using warp.

sftcd:
> This is not a problem with boringssl, cloudflare network rejects ech when using warp.

Ah - didn't know that. Any more info on it?

So in this case, the client has found an ECHConfig in the DNS published for the web server, and has attempted to use that, but then see's a failure. I'm not sure falling back to cleartext SNI at that point is correct. (``--ech true`` would fall back to cleartext SNI if it had not found an ECHConfig in the DNS.)

vvb2060:
> Any more info on it?

No, cloudflare doesn't publicly document this behavior, I'm guessing because it's already on the cloudflare intranet, ech is pointless. You can test it by [install WARP client](https://developers.cloudflare.com/warp-client/get-started/).

I think `--ech true` should be similar to browser behavior, try when possible and then fall back.

sftcd:
> > Any more info on it?
> 
> No, cloudflare doesn't publicly document this behavior, I'm guessing because it's already on the cloudflare intranet, ech is pointless. You can test it by [install WARP client](https://developers.cloudflare.com/warp-client/get-started/).
> 

Interesting. Will try that when I get a chance (probably not today). Given the output you show above I guess it's the warp client that's causing the fail or do you know? I wonder if GREASEd ECH would also fail similarly.

> I think `--ech true` should be similar to browser behavior, try when possible and then fall back.

Not sure TBH. If we did that then any middlebox could force cleartext SNI without a user knowing (unless they'd chosen to use ``--ech hard``). I'll take a look and see if curl does such fallbacks (creating a new TLS session if some TLS setting asked for on the command line fails).

vvb2060:
> I wonder if GREASEd ECH would also fail similarly.

```
curl --ech GREASE --doh-url https://1.0.0.1/dns-query https://crypto.cloudflare.com/cdn-cgi/trace

fl=467f66
h=crypto.cloudflare.com
ip=
ts=1707312347.172
visit_scheme=https
uag=curl/8.6.1-DEV
colo=
sliver=none
http=http/2
loc=
tls=TLSv1.3
sni=plaintext
warp=plus
gateway=off
rbi=off
kex=X25519
```

sftcd:
So if GREASE'd ECH is working, that implies it's not blocking on the existence of an ECH extension in the client hello. Speculating, it could be that however CF have warp setup, the ``public_name`` for the ECHConfig causes a problem (on the client side?). If that's the case, you could (with the curl+ECH+OpenSSL build) over-ride that value using ``--ech pn:<foo>`` and see if some value of foo works. (That's assuming that ECH decryption will be attempted on the server-side should the client allow the ClientHello onto the n/w.)

If however, in the initial case reported, the client is sending out the ClientHello, then could be something else at play.

bagder (commented):
Nice work! Here is a first set of comments, questions and feedback!

docs/ECH.md
```diff
@@ -0,0 +1,499 @@
...
+    ...stuff...
+    make -j8
+    ...stuff (maybe go for coffee)...
+```
+
+To build our ECH-enabled curl fork, making use of the above:
+
+```bash
+    cd $HOME/code
+    git clone https://github.com/sftcd/curl
```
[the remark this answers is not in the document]
sftcd: indeed - now fixed locally (will commit in a bit)

docs/ECH.md
```diff
@@ -0,0 +1,499 @@
...
+    ...lots more output...
+```
+
+If you do not get that WARNING at the end of the ``configure`` command, then ECH
+is not enabled, so go back some steps and re-do whatever needs re-doing:-) If you
+want to debug curl then you should add ``--enable-debug`` to the ``configure``
+command.
+
+You may need to install some libraries to get ``configrure`` to
+work, e.g. I recently had to add  ``sudo apt install libpsl-dev``
```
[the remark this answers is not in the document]
sftcd: sure, deleted (will commit in a bit)

docs/ECH.md
```diff
@@ -0,0 +1,499 @@
...
+
+## Building with cmake
+
+This is not working cleanly yet, but to build with cmake:
+
+```bash
+    cd $HOME/code
+    git clone https://github.com/sftcd/curl
```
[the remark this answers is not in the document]
sftcd: as above

docs/libcurl/opts/CURLOPT_ECH.md
```diff
@@ -0,0 +1,72 @@
+---
+c: Copyright (C) Daniel Stenberg, <daniel.se>, et al.
+SPDX-License-Identifier: curl
+Title: CURLOPT_ECH
+Section: 3
+Source: libcurl
+See-also:
+  - CURLOPT_DOH_URL (3)
+Protocol:
+  - HTTPS
```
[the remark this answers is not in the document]
sftcd: At one level, TLS would be more correct. However, for now, ECH support that fetches ECHConfig values (the public keys needed) from the DNS (via DoH), only supports use of the HTTPS RR, which is intended only for HTTPS. At the command-line or library level, if an application finds and supplies the right ECHConfig value, then that should work ok, but I've only done tests of that with web servers on the other side, those being the only ECH-enabled servers with which I can currently test. (Well, haproxy is a little different, not being a web server, but the only protocol tested so far was still HTTPS.)

I'd be happy to test with some other protocol on top of TLS if/when possible.

In summary, I'd say for now, HTTPS is more accurate but it could sensibly change to TLS in future.

docs/libcurl/opts/CURLOPT_ECH.md
```diff
@@ -0,0 +1,72 @@
...
+#include <curl/curl.h>
+
+CURLcode curl_easy_setopt(CURL *handle, CURLOPT_ECH, char *config);
+~~~
+
+# DESCRIPTION
+
+Pass a string that specifies configuration details for ECH.
+In all cases, if ECH is attempted, it may fail for various reasons.
+The keywords supported are:
```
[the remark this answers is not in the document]
sftcd: Yeah, that's probably phrased too much like the related command line argument text. I'll take a look at making it better, and maybe make a test.

docs/libcurl/symbols-in-versions
```diff
@@ -340,6 +340,7 @@ CURLE_URL_MALFORMAT_USER        7.1           7.17.0
 CURLE_USE_SSL_FAILED            7.17.0
 CURLE_WEIRD_SERVER_REPLY        7.51.0
 CURLE_WRITE_ERROR               7.1
+CURLE_ECH_REQUIRED              8.6.1
```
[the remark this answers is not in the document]
sftcd: ack, fixed locally (will commit in a bit)

docs/ECH.md
```diff
@@ -0,0 +1,499 @@
...
+The ``--ech true`` option is opportunistic, so tries to do ECH but does not fail if
+the client for example cannot find any ECHConfig values. The ``--ech hard``
+option hard-fails if there is no ECHConfig found in DNS, so for now, that is not
+a good option to set as a default.
+
+## Code changes for ECH support when using DoH
+
+All code changes are in a new ``ECH-experimental`` branch of our fork
+([here](https://github.com/sftcd/curl/tree/ECH-experimental)) and are
+``#ifdef`` protected via ``USE_ECH`` or ``USE_HTTPSRR``:
```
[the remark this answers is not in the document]
sftcd: changed locally (will commit in a bit)

docs/libcurl/opts/CURLOPT_ECH.md
```diff
@@ -0,0 +1,72 @@
...
+if(curl) {
+  curl_easy_setopt(curl, CURLOPT_ECH, config);
+  curl_easy_perform(curl);
+}
+~~~
+
+
+# AVAILABILITY
+
+Added in 8.6.0
```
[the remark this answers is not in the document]
bagder: The feature window closes on April 27 for this cycle. So we have two long weeks! :grinning:

docs/libcurl/opts/CURLOPT_ECH.md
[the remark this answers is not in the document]
bagder: Ok, thanks for the additional info. There is however a subtle extra detail that I think still argues for this to me marked as `TLS` in this file: we have extra special magic for options that are for TLS, as they also need to specify for which TLS backends they work - which then by extension ends up visible in a [table of TLS options on the webpage](https://curl.se/libcurl/c/tls-options.html).

In that regard this is a TLS option that depends on the specific TLS backend if it is supported or not in curl.

The caveat that this feature probably only works for HTTPS for the moment should probably then be added in the description.

Of course, this feature has the additional complexity documentation wise that it also requires a patch on OpenSSL to work with it, but maybe we can add that in text and yet claim OpenSSL support in the meta-data...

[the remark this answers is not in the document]
sftcd: Sure, I'll make it say "TLS" and "OpenSSL, boringssl, WolfSSL" and try craft text to explain the actual status. I guess that's ok as it'd also appear as experimental on the web page you mention? Or can say that in text.

docs/libcurl/opts/CURLOPT_ECH.md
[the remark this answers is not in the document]
sftcd: Well, that timing's not too bad - after the 29th I'll be marking a lot of exam scripts so between now and then is a good time to try get it over the line:-)

I plan to try address all these comments and re-do all the tests (with boring/wolf etc) next few days so whether this makes it will likely depend on whether your or others' reviews find problems mostly.

lib/hostip.h
```diff
@@ -58,8 +62,32 @@ struct connectdata;
...
+  uint16_t priority;
+  char *target;
+  char *alpns;
+  bool no_def_alpn;
+  unsigned char *ipv4hints;
+  size_t ipv4hints_len;
+  unsigned char *echconfiglist;
+  size_t echconfiglist_len;
+  unsigned char *ipv6hints;
+  size_t ipv6hints_len;
```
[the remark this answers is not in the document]
sftcd: Not sure I'm following the comment tbh. 

Currently, the fields are the mandatory ones, then the optional ones in the order of their IANA codepoints. I've added a comment to that effect that may be a good enough reason to keep the ordering. (Might make debugging a teeny tiny bit easier maybe.)

But, happy to re-order if that makes sense, though in that case, could do with a better hint as to what order. (Do you consider a ``size_t`` bigger or smaller than an ``unsigned char *`` ?)

lib/urldata.h
```diff
@@ -55,6 +55,15 @@
 
 struct curl_trc_featt;
 
+#ifdef USE_ECH
+/* CURLECH_ bits for the tls_ech option */
+# define CURLECH_DISABLE    0
+# define CURLECH_GREASE     1
+# define CURLECH_ENABLE     2
+# define CURLECH_HARD       3
+# define CURLECH_CLA_CFG    4
```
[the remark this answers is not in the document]
sftcd: yep, a bitmask is right here - you can definitely get e.g. "hard" and have an ECHConfig provided from the command line (CLA_CFG). Will change to (1<<2) etc.

lib/setopt.c
```diff
@@ -3141,6 +3141,49 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
...
+      return result;
+    }
+    plen = strlen(argptr);
+    if(plen > CURL_MAX_INPUT_LENGTH) {
+      data->set.tls_ech = CURLECH_DISABLE;
+      result = CURLE_BAD_FUNCTION_ARGUMENT;
+      return result;
+    }
+    /* set tls_ech flag value, preserving CLA_CFG bit */
+    if(plen == 5 && strcasecompare(argptr, "FALSE"))
```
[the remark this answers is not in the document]
sftcd: fair 'nuff - changed to case sensitive 

The ``plen == 5`` stuff is just belt'n'braces - prefer wearing those when parsing inputs:-)

docs/ECH.md
```diff
@@ -0,0 +1,499 @@
...
+To build our ECH-enabled curl fork, making use of the above:
+
+```bash
+    cd $HOME/code
+    git clone https://github.com/sftcd/curl
+    cd curl
+    git checkout ECH-experimental
+    autoreconf -fi
+    export LD_LIBRARY_PATH=$HOME/code/openssl
+    LDFLAGS="-L$HOME/code/openssl" ./configure --with-ssl=$HOME/code/openssl --enable-ech --enable-httpsrr
```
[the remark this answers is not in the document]
sftcd: tried that just now but didn't immediately work, will come back to it

include/curl/curl.h
davidben: "Required" was named that way because failure to decrypt in ECH isn't immediately fatal to the protocol. You just handshake with ClientHelloOuter instead. That's what allows ECH GREASE to work. Of course, the client has an opinion as to which of ClientHelloOuter or ClientHelloInner it wants, but the server doesn't know that. So the intended read was "hey, you (server) didn't decrypt ECH and I (client) required it to accept this connection".

docs/libcurl/opts/CURLOPT_ECH.md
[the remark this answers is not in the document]
sftcd: made that change, will push shortly

docs/libcurl/opts/CURLOPT_ECH.md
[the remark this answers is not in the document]
sftcd: this is still TBD

lib/connect.c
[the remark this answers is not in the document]
sftcd: that seems to have gotten fixed along the way

docs/options-in-versions
```diff
@@ -55,6 +55,7 @@
 --doh-insecure                       7.76.0
 --doh-url                            7.62.0
 --dump-header (-D)                   5.7
+--ech                                8.6.1
```
bagder, line 58:
```suggestion
--ech                                8.8.0
```

bagder:
I think we can merge this as a first take once these minor nits are fixed. This is still **experimental** so we do not guarantee functionality or ABI/API compatibility, but getting the first version in git will allow for more and easier tests by a larger community.

One thing that needs fixing at some point is the configure/cmake checks: for multissl to work with ech, we need to allow **per-TLS ECH ability**, not a global boolean.

BTW, as do not think `LD_LIBRARY_PATH` is very nice to use, I prefer this build like this:

~~~
LDFLAGS="-Wl,-rpath,$HOME/build-openssl-ech/lib64" \
./configure  --with-ssl=$HOME/build-openssl-ech  --enable-ech --enable-httpsrr

docs/cmdline-opts/ech.md
```diff
@@ -0,0 +1,53 @@
+---
+c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+SPDX-License-Identifier: curl
+Long: ech
+Arg: <config>
+Help: Configure Encrypted Client Hello (ECH) for use with the TLS session
+Added: 8.6.1
+Category: tls ECH
```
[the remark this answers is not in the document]
sftcd: just did that

bagder:
Thanks a lot for your hard and patient work on this @sftcd . We can now continue polishing this in separate PRs.

sftcd:
> Thanks a lot for your hard and patient work on this @sftcd . We can now continue polishing this in separate PRs.

Excellent! Thanks for being so receptive for experimental stuff like this. Happy to help as useful in future too of course.

PR #11928, opened 2023-09-24
cmake: fix unity with Windows Unicode + TrackMemory

vszakats:
Found the root cause of the startup crash in unity builds with Unicode
and TrackMemory enabled at the same time.

We must make sure that the `memdebug.h` header doesn't apply to
`lib/curl_multibyte.c` (as even noted in a comment there.) In unity
builds all headers apply to all sources, including `curl_multibyte.c`.
This probably resulted in an infinite loop on startup.

Exclude this source from unity compilation with TrackMemory enabled,
in both libcurl and curl tool. Enable unity mode for a debug Unicode
CI job to keep it tested. Also delete the earlier workaround that
fully disabled unity for affected builds.

Follow-up to d82b080f6374433ce7c98241329189ad2d3976f8 #12005
Follow-up to 3f8fc25720900b14b7432f4bd93407ca15311719 #11095

Closes #11928

---

Crashing on startup example:
```
could not get curl version! at C:/projects/curl/tests/runtests.pl line 2439.
```
https://ci.appveyor.com/project/curlorg/curl/builds/48113110/job/3ou611asxeop6uvi#L2522

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,16 +144,6 @@ if(ENABLE_CURLDEBUG)
   set_property(DIRECTORY APPEND PROPERTY COMPILE_DEFINITIONS CURLDEBUG)
 endif()
 
-# This combination results in curl tool crashing on startup when run on native
-# Windows (it was seen to work when run under WINE though). We disable UNITY
-# mode as a workaround.
-if(ENABLE_UNICODE AND ENABLE_CURLDEBUG)
-  set(CMAKE_UNITY_BUILD OFF)
-  message(WARNING "Unity builds are not supported with Windows Unicode and "
-    "TrackMemory (-DCURLDEBUG) enabled at the same time. Unity mode has been "
-    "disabled.")
-endif()
-
 # For debug libs and exes, add "-d" postfix
 if(NOT DEFINED CMAKE_DEBUG_POSTFIX)
   set(CMAKE_DEBUG_POSTFIX "-d")
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -121,7 +121,7 @@ environment:
       DISABLED_TESTS: '!1139 !1501'
       ADD_PATH: 'C:\msys64\usr\bin'
     # generated CMake-based MSYS Makefiles builds (mingw cross-compiling)
-    - job_name: 'CMake, mingw-w64, gcc 8, Debug x64, Schannel, Static, Unicode'
+    - job_name: 'CMake, mingw-w64, gcc 8, Debug x64, Schannel, Static, Unicode, Unity'
       APPVEYOR_BUILD_WORKER_IMAGE: 'Visual Studio 2015'
       BUILD_SYSTEM: CMake
       PRJ_GEN: 'MSYS Makefiles'
@@ -134,6 +134,7 @@ environment:
       ADD_PATH: 'C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin;C:\msys64\usr\bin'
       MSYS2_ARG_CONV_EXCL: '/*'
       BUILD_OPT: -k
+      UNITY: 'ON'
     - job_name: 'CMake, mingw-w64, gcc 7, Debug x64, Schannel, Static, Unicode'
       APPVEYOR_BUILD_WORKER_IMAGE: 'Visual Studio 2017'
       BUILD_SYSTEM: CMake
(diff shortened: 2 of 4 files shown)


vszakats:
So, for one thing, libtests and curl tool wasn't finding libcurl DLL, that was the next issue. After fixing that, the non-unity builds started running as expected, but their unity counterparts kept failing. They run fine locally with wine64, but crash on startup on real Windows. [ Unity binaries are also much smaller, but I attribute that to the more concise debug info in the unity builds. Their size become similar after stripping. ]

So the next issue is to find what causes these shared/debug unity builds to fail on real Windows.

vszakats:
`curl_memory.h`, `curl_multibyte.h` and `memdebug.h` look suspect, but no evidence or test result.

I'm about to give up finding the issue, and instead disable UNITY for `src` when UNICODE and DEBUG is enabled. It doesn't look a terrible limitation.

MarcelRaad (approved):

jay:
`CMake, mingw-w64, gcc 8, Debug x64, Schannel, Static, Unicode, Unity` (the build you just enabled unity on) failed in ci due to an ftp test. ftp testing in windows has some flakiness (#12002 for example) so I suspect it's unrelated but I set the job to rerun

vszakats:
The way to track this down was not the most obvious. While doing the first tests I added a way to run `curl.exe` after each build. This turned out to be non-trivial with the MS-DOS batch, triggering a rewrite in PowerShell that is better, and going into an all new rabbit hole. But, it resulted in CI jobs able to detect this issue, a more clearly and much quicker than `./runtests.pl` failing with `could not get curl version!`. Locally, it needed a real Windows VM for running `curl.exe`. Strangely `WINE` was able to run these filaing `curl.exe`s without issues.

Then it took a long series of builds to narrow this down the mininum conditions.

When this cleared up, I had some guesses where to look, but eyeballing the source didn't help. I was also looking at headers, assuming that these could affect both libcurl and curl, so missed the comment in the affected source file.

At this point I created one build with and one without the issue and planned to compare disassembled binaries. For this I had to learn how to tell CMake to use `-O0` without enabling debug. This needed editing its own config file (unintuitively: `.../share/cmake/Modules/Compiler/GNU.cmake`, for llvm/clang) and not supported with regular CMake options or via `CMakeLists.txt`. But, this turned out to be a dead-end anyway. Even with non-debug `-O0`, the resulting objects have internal functions in different order, string optimization was still enabled and cross-references also made up too much difference in the `objdump -D` outputs.

Next I picked the compiler commands from the build logs and re-run them by replacing `-o <obj>` with `-E` for both builds, to compare the actual C code they're ending up with. It was interesting to realize that an almost minimal Windows curl build creates 600MB of compilation input (src + lib) without unity and 11MB (!) with unity. This explains the massive gain in compilation speed. And also the dramatic reduction in the debug info included in debug builds. A large portion of that is likely Windows headers (I haven't made tests with a Linux or macOS build):

```
-rw-r--r--  1 user  staff  465451081 Oct  3 10:46 lib-non_unity.c.E
-rw-r--r--  1 user  staff    8108339 Oct  3 10:44 lib-unity.c.E
-rw-r--r--  1 user  staff  134004891 Oct  3 10:47 src-non_unity.c.E
-rw-r--r--  1 user  staff    3659820 Oct  3 10:44 src-unity.c.E
```

Comparing texts this large was a no-go, so picked up the `[uninclude](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=contrib/uninclude;hb=HEAD)` script from GCC, to strip systems headers. Had to modify it to recognize mingw-w64 headers (`x86_64-w64-mingw32/include`). Since curl keeps no executable code in headers, I took a little risk and also stripped `include/curl` and `curl/lib` (this will bite a bit later).

This still had too much noise, so restarted the `-E` pass after making sure that the order of the non-unity build sources matches the ones in `Makefile.inc` files. (A PR is coming up fixing a few alpha-sorting issues in those.) Sorting was different because I left in `-j4` for the build runs that generated the logs with the compiler commands.

At this point it was a matter of sifting through the two diffs for `lib` and `src` respectively. Disregaring the remaining header-noise, these were reasonably identical, but not completely so. Notably `((void *)0)` might be `((void*)0)`, and `((NTSTATUS)0x00000000)` is also `((NTSTATUS)0x00000000L)` depending on unity and non-unity. I surmise because Windows headers' sensitivity to include order. `src` did not reveal anything suspicious. But `lib` _did_:

Screen Shot 2023-10-03 at 11 11 50

The risk I took to strip `curl/lib` stuff from the `.E` output deleted the `curl_multbyte.c` reference from the `src` diffs, otherwise I could have caught that as well.

The fix itself was trivial, after realizing that we also use `curl_multibyte.c` for the curl tool.

dfandrich:
Sounds like the outline to an unreleased Tom Clancy thriller! ☺

PR #11929, opened 2023-09-25
cmake: fix stderr initialization in unity builds

vszakats:
Before this patch, in certain build configurations the curl tool may
not have displayed anything (debug, macOS), or crashed at startup
(debug, Windows).

Follow-up to 3f8fc25720900b14b7432f4bd93407ca15311719
Necessary after 2f17a9b654121dd1ecf4fc043c6d08a9da3522db

Closes #11929

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -66,6 +66,11 @@ if(BUILD_STATIC_CURL)
   set(CURLX_CFILES ../lib/dynbuf.c ../lib/base64.c)
 endif()
 
+# Compile this source separately to avoid having `stderr` overridden in this
+# file. Necessary for successfully initializing stderr in unity builds.
+# (Not necessary for the libcurltool target.)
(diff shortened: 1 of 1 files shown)


vszakats:
This patch initially aimed to fix this CI error, which I attributed to a failing curl tool:
`could not get curl version! at C:/projects/curl/tests/runtests.pl line 2439.`

But it turns out this is unrelated, and also happens without unity:
https://ci.appveyor.com/project/curlorg/curl/builds/48113479

dfandrich:
"could not get curl version!" basically means the curl executable could not be run.

vszakats:
The binaries are working when downloaded locally, and there are no 3rd party DLLs required. This makes me think that the problem may be as simple as `curl.exe` not finding `libcurl.dll`. But I failed to come up with a DLL copy command that made it work, probably due to trivial mistake I'm making in `appveyor.yml`.

PR #11937, opened 2023-09-25
multi: fix small timeouts

hwti:
Since Curl_timediff rounds down to the millisecond, timeouts which expire in less than 1ms are considered as outdated and removed from the list.
We can use Curl_timediff_us instead, big timeouts could saturate but this is not an issue.


bagder (approved):

bagder:
Thanks!

PR #11938, opened 2023-09-25
multi: round the timeout up to prevent early wakeups

hwti:
Curl_timediff rounds down to the millisecond, so curl_multi_perform can be called too early, then we get a timeout of 0 and call it again.

The code already handled the case of timeouts which expired less than 1ms in the future.
By rounding up, we make sure we will never ask the platform to wake up too early.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -3422,20 +3422,10 @@ static CURLMcode multi_timeout(struct Curl_multi *multi,
 
     if(Curl_splaycomparekeys(multi->timetree->key, now) > 0) {
       /* some time left before expiration */
-      timediff_t diff = Curl_timediff(multi->timetree->key, now);
-      if(diff <= 0)
-        /*
-         * Since we only provide millisecond resolution on the returned value
-         * and the diff might be less than one millisecond here, we don't
-         * return zero as that may cause short bursts of busyloops on fast
-         * processors while the diff is still present but less than one
-         * millisecond! instead we return 1 until the time is ripe.
-         */
-        *timeout_ms = 1;
-      else
-        /* this should be safe even on 64 bit archs, as we don't use that
-           overly long timeouts */
-        *timeout_ms = (long)diff;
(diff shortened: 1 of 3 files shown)


lib/multi.c
```diff
@@ -3422,20 +3422,10 @@ static CURLMcode multi_timeout(struct Curl_multi *multi,
...
-         * processors while the diff is still present but less than one
-         * millisecond! instead we return 1 until the time is ripe.
-         */
-        *timeout_ms = 1;
-      else
-        /* this should be safe even on 64 bit archs, as we don't use that
-           overly long timeouts */
-        *timeout_ms = (long)diff;
+      timediff_t diff = Curl_timediff_ceil(multi->timetree->key, now);
+      /* this should be safe even on 64 bit archs, as we don't use that
```
bagder, line 3426: This comment looks wrong. I suppose it was meant to say "even on 32 bit", since only then and on Windows `long` is 32 bit and therefore smaller than `timediff_t`.

lib/timeval.h
```diff
@@ -43,6 +43,14 @@ struct curltime Curl_now(void);
  */
 timediff_t Curl_timediff(struct curltime t1, struct curltime t2);
 
+/*
+ * Make sure that the first argument (t1) is the more recent time and t2 is
+ * the older time, as otherwise you get a weird negative time-diff back...
+ *
+ * Returns: the time difference in number of milliseconds, rounded up
+ */
+timediff_t Curl_timediff_ceil(struct curltime t1, struct curltime t2);
```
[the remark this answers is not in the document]
hwti: Done, I adjusted the comments too (for `Curl_timediff_us`, the prototypes used newer/older, but the comment was still using t1/t2), but it makes the diff a little harder to read (due to the line wrap change).

lib/multi.c
hwti: done

bagder (approved):

bagder:
Thanks!

PR #11939, opened 2023-09-25
connect: only start the happy eyeballs timer when needed

hwti:
The timeout is only used when there is a second address family, for the delayed eyeballer.


icing (approved):
This looks correct.

bagder:
Thanks!

PR #11940, opened 2023-09-25
build: delete checks for C89 standard headers

vszakats:
Delete checks and guards for standard C89 headers and assume these are
available: `stdio.h`, `string.h`, `time.h`, `setjmp.h`, `stdlib.h`,
`stddef.h`, `signal.h`

Some of these we already used unconditionally, some others we only used
for feature checks.

Follow-up to 9c7165e96a3a9a2d0b7059c87c699b5ca8cdae93 #11918 (for `stdio.h` in CMake)

Closes #11940


bagder (approved):

ncopa:
`stdio.h` is needed to detect `fseeko`, both on MacOS and Linux.
```
-- Looking for fseeko
-- Looking for fseeko - not found
```

vszakats:
@ncop: Do you think this solution would also work:
```
check_symbol_exists(fseeko         "${CURL_INCLUDES};stdio.h" HAVE_FSEEKO)
check_symbol_exists(_fseeki64      "${CURL_INCLUDES};stdio.h" HAVE__FSEEKI64)
```
This also self-documents it clearly that stdio.h is necessary here for detection, without adding it for every other check by default.

vszakats:
@bagder: I extended this PR with other C89 headers that we're checking, most likely unnecessarily. I'm looking into `signal.h` and plan to add it shorly.

vszakats:
ea8fbb5233b24745177871a7e1e9105a137f94f9 (from 2008) suggests that a system called "Cell" might not actually come with a C89 `signal.h`. According to the [original report](https://curl.se/mail/lib-2008-11/0164.html) this is CellOS for Playstation 3. Could this be still relevant (or on other systems) to keep checking for `signal.h`?

jay:
[c89](http://port70.net/~nsz/c/c89/c89-draft.txt) 4.1.2 standard headers:
~~~
         <assert.h>                 <locale.h>                 <stddef.h>
         <ctype.h>                  <math.h>                   <stdio.h>
         <errno.h>                  <setjmp.h>                 <stdlib.h>
         <float.h>                  <signal.h>                 <string.h>
         <limits.h>                 <stdarg.h>                 <time.h>
~~~

regarding HAVE__FSEEKI64 see #11944

vszakats:
I plan to follow this up with a patch that moves the global standard headers (`string.h`, `stdlib.h`, `stddef.h`) to specific detections where they are actually needed (if they are) based on autotools logic. Also found an incomplete/broken detection in CMake and one missing compared to autotools.

ncopa:
> @ncop: Do you think this solution would also work?:
> 
> ```cmake
> check_symbol_exists(fseeko         "${CURL_INCLUDES};stdio.h" HAVE_FSEEKO)
> check_symbol_exists(_fseeki64      "${CURL_INCLUDES};stdio.h" HAVE__FSEEKI64)
> ```
> 
> This also self-documents that stdio.h is necessary for detection, without adding it for every other check by default.

Yes, I would expect that work work, and according the logs it does:
```
-- Looking for fseeko
-- Looking for fseeko - found
```

I agree that it is better this way.

ncopa:
> Also found an incomplete/broken detection in CMake and one missing compared to autotools.

I only discovered the broken test of fseeko, because I was looking at the logs. It is fragile because the fallback code also works, so a broken test can easily go undetected.

Maybe it would be an idea to add a test to CI to run both `configure` and `cmake` and compare the generated `curl_config.h`?

bagder:
> Maybe it would be an idea to add a test to CI to run both `configure` and `cmake` and compare the generated `curl_config.h`?

That's a good idea. It will of course only test the exact options used in the test and we can run a few different ones, but it should still help us.

vszakats:
Such detection comparison test would be very useful indeed.

m4/curl-functions.m4
```diff
@@ -298,15 +292,13 @@ curl_includes_string="\
...
-#ifdef HAVE_STRING_H
-#  include <string.h>
-#endif
+#include <string.h>
 #ifdef HAVE_STRINGS_H
 #  include <strings.h>
 #endif
 /* includes end */"
   AC_CHECK_HEADERS(
-    sys/types.h string.h strings.h,
```
MarcelRaad, line 309: This seems to break `strerror_r` detection on Solaris, visible as warnings in the autobuilds since yesterday:
https://curl.se/dev/log.cgi?id=20230927074238-1773871#prob1
```
configure: WARNING: cannot determine strerror_r() style: edit lib/curl_config.h manually.
```

vszakats: Is it possible Solaris doesn't have a string.h? Or it something else? Is there a way to pull or view config.log of a failed build?

vszakats: According to this, strerror_r is declared in string.h: https://docs.oracle.com/cd/E86824_01/html/E54766/strerror-r-3c.html

PR #11941, opened 2023-09-25
http: correct comment for http_should_fail

bagder:
Reported-by: Christian Schmitz
Ref: #11936


lib/http.c
```diff
@@ -1159,7 +1159,7 @@ CURLcode Curl_http_input_auth(struct Curl_easy *data, bool proxy,
  * http_should_fail() determines whether an HTTP response has gotten us
  * into an error state or not.
  *
- * @param conn all information about the current connection
+ * @param data information about the transfer
```
jay, line 1162: IMO this is not needed but if you are going to change it I would just call it the easy handle

bagder: We can probably just remove the entire line there.

PR #11942, opened 2023-09-25
docs: add see also curl_multi_get_handles to some man pages

bagder:


jay:
Isn't the format supposed to be `"(3),"` ? I've pushed that as a change that removes whitespace and adds the comma where necessary

bagder:
Maybe we should extend the man page format-checker to also check those...

bagder:
I did that in #11957

PR #11944, opened 2023-09-25
config-win32: define HAVE__FSEEKI64

jay:
Follow-up to 9c7165e9 which added an fseeko wrapper to the lib that calls _fseeki64 if it is available.

Closes #xxxx

---

Follow up to #11918. I'm not entirely sure about the logic here but I think we need to do something for builds that use the pre-generated win32 configuration file config-win32.h.

/cc @ncopa

### DIFF
--- a/lib/config-win32.h
+++ b/lib/config-win32.h
@@ -564,6 +564,10 @@ Vista
 #  endif
 #endif
 
+#ifdef USE_WIN32_LARGE_FILES
+#define HAVE__FSEEKI64
+#endif
+
 /* Define to the size of `off_t', as computed by sizeof. */
 #if defined(__MINGW32__) && \
   defined(_FILE_OFFSET_BITS) && (_FILE_OFFSET_BITS == 64)


MarcelRaad (approved):
Makes sense to me!

vszakats:
We could also enable this function via `CMake/Platforms/WindowsCache.cmake`.

Shouldn't we check for `USE_WIN32_LARGE_FILES` on use (in `lib/formdata.c`), rather than in `lib/config-win32.h`, to make it behave the same with all build systems? (with CMake and autotools, `HAVE__FSEEKI64` is enabled regardless of the  `USE_WIN32_LARGE_FILES` setting.)

ncopa:
I was not aware that curl supported pre-generated configs. `cmake` and `autotools` will detect if the function(s) are available regardless of platform and version. I think its better to run those tests instead of making assuptions based on OS/platform etc. But ofc, if you do support build with pre-generated configs (without running the tests), then sure.

jay:
> Shouldn't we check for `USE_WIN32_LARGE_FILES` on use (in `lib/formdata.c`), rather than in `lib/config-win32.h`, to make it behave the same with all build systems? (CMake and autotools enables `HAVE__FSEEKI64` regardless of the `USE_WIN32_LARGE_FILES` setting.)

That's probably not necessary. If _fseeki64 is available then it can be used even if large file support is not enabled, but on any modern system I don't know where that would happen. config-win32 supports some older systems that (I guess?) would not have _fseeki64 so I'm ok with it as is. I just compiled a test program in Visual Studio 2008 that calls _fseeki64 and it's available there, so I think we'd have to go really far back to find some time that it is not supported.

PR #11945, opened 2023-09-26
ci: switch old mingw MSYS1 jobs to MSYS2

vszakats:
Follow-up to 38029101e2d78ba125732b3bab6ec267b80a0e72

Closes #xxxxx


mback2k:
Please don't merge before I had a chance to check the details here. At first glance this doesn't look correct regarding the path changes and used container images I provide.

mback2k (approved):
Nevermind, this generally looks good to me, but I plan to follow up with another restructuring that will also include static builds formerly done on Cirrus CI.

mback2k:
But I am still a little disappointed that MinGW and msys1 support is gone for no reason.

bagder:
> gone for no reason

1. Exactly **nobody** voiced their support for keeping mingw v1 support during the 6+ months we had it marked for pending removal
2. Getting rid of this means reducing the number of special cases in the code, which is good

vszakats:
I've been building for Windows XP via mingw-w64 with no toolchain issues. It's not the default, but was doable at the time when I dropped this effort in mid-2022. It needed `-D_WIN32_WINNT=0x0501` and some other tweaks. Ref: https://github.com/curl/curl-for-win/commit/6976612160075c1e9ee967964d5dec1a25c5ac6c

vszakats:
It fails with two 32-bit jobs, with the error I was chasing for a few while (without success) already:
```
could not get curl version! at ./runtests.pl line 2439.
make[1]: *** [Makefile:826: ci-test] Error 127
```

mback2k:
> > gone for no reason
> 
> 1. Exactly **nobody** voiced their support for keeping mingw v1 support during the 6+ months we had it marked for pending removal

Sorry, I must have missed the initial pending removal marking.

> 2. Getting rid of this means reducing the number of special cases in the code, which is good

Understood and actually makes a lot of sense. 👍

PR #11946, opened 2023-09-26
Remove Curl_multi_dump declaration from headers

icing:
- as the function was removed previously
- as a cheap ploy to increase my commit count


jay:
Thanks

PR #11947, opened 2023-09-26
Update version.c to remove unneeded easy_lock.h include

MonkeybreadSoftware:
Removed unneeded easy_lock.h include.
As far as I see, it is not needed.


MonkeybreadSoftware:
never mind. The #if needs it.

PR #11948, opened 2023-09-26
Update hostip.c to add #if around easy_lock.h 

MonkeybreadSoftware:
easy_lock.h not always needed.


MonkeybreadSoftware:
never mind. The #if needs it.

PR #11956, opened 2023-09-26
HTTP2: support GOAWAY Callback

watsocao:
Some Applications need to be notified when underlined H2 connection receiving GOAWAY frame so that it could do some clean up job and prepare fresh connection. This patch is first attempt to support this option to CURLM object.


calvin2021y:
Maybe add a option, to allow auto reconnect and send pending request if there is GOAWAY frame

PR #11957, opened 2023-09-26
docs: clean up and polish the SEE ALSO sections of all man pages

bagder:


jay:
> Skip the quotes around the section number for example (3)

Is there a documented format? I thought it was `example "(3),"` but probably because that's the way we've done it in the past.

bagder:
> Is there a documented format? I thought it was `example "(3),"` but probably because that's the way we've done it in the past.

Sure the format is documented, but its a very diverse format that allows for many different ways to do this. I landed on this updated version (which I also think is simplified) after trying what works with the `man` command and a little bit what makes the man page manageable to edit for us humans and also easier to check by a script.

PR #11961, opened 2023-09-27
Client writer, part 2, accounting + logging

icing:
This PR has these changes:

Renaming of `unencode_*` to `cwriter`, e.g. client writers

- documentation of sendf.h functions
- move max decode stack checks back to content_encoding.c
- define writer `phase` which was used as `order` before
- introduce phases for monitoring inbetween decode phases
- offering default implementations for init/write/close

Add `type` paramter to client writer's `do_write()`

- always pass all writes through the writer stack
- writers who only care about `BODY` data will pass other writes unchanged

add RAW and PROTOCOL client writers

- RAW used for `Curl_debug()` logging of `CURLINFO_DATA_IN`
- PROTOCOL used for updates to `data->req.bytecount`, `max_filesize` checks and 
      `Curl_pgrsSetDownloadCounter()`
- remove all updates of `data->req.bytecount` and calls to
      `Curl_pgrsSetDownloadCounter()` and `Curl_debug()` from  other code
- adjust test457 expected output to no longer see the excess write


icing:
Now that rtp-parser has been merged, I'll replace this with a new, clean PR.

PR #2609, opened 2018-05-27
Android build fix & CMake fix

bwalle:
See the individual patches.

PR #11964, opened 2023-09-27
workflow to compare configure vs cmake outputs

bagder:
Details pending to fix/address:

- [X] HAVE_FNMATCH #11973
- [X] HAVE_SENDMSG #11973
- [X] HAVE_SCHED_YIELD #11973
- [x] USE_TLS_SRP #11967
- [X] TIME_WITH_SYS_TIME #11975
- [X] HAVE_SUSECONDS_T #11977
- [X] `CURL_SA_FAMILY_T` is worked out by `curl_setup.h` and not purely in cmake but should be fine
- [x] HAVE_WRITABLE_ARGV #11978
- [x] HAVE_GETADDRINFO_THREADSAFE #11979
- [x] HAVE_CLOCK_GETTIME_MONOTONIC_RAW #11981
- [X] HAVE_NETINET_UDP_H and HAVE_SYS_WAIT_H #11996
- [X] CURL_CA_PATH #11997
- [x] HAVE_IOCTL_FIONBIO, HAVE_IOCTL_SIOCGIFADDR #12008
- [x] HAVE_LDAP_SSL, HAVE_LDAP_URL_PARSE #12006 #12015
- [X] HAVE_SA_FAMILY_T missing from configure (ignored)
- [X] filter out (allowlist) symbols known to "not matter" for this comparison

### DIFF
--- a/.github/workflows/configure-vs-cmake.yml
+++ b/.github/workflows/configure-vs-cmake.yml
@@ -1,0 +1,45 @@
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+
+name: configure-vs-cmake
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - '*.ac'
+    - '**/*.m4'
+    - '**/CMakeLists.txt'
+    - 'lib/curl_config.h.cmake'
+    - 'scripts/cmp-config.pl'
+
+  pull_request:
+    branches:
+    - master
+    paths:
+    - '*.ac'
+    - '**/*.m4'
+    - '**/CMakeLists.txt'
+    - 'lib/curl_config.h.cmake'
+    - 'scripts/cmp-config.pl'
+
+permissions: {}
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: run configure --with-openssl
+      run: |
+         autoreconf -fi
+         ./configure --with-openssl
+
+    - name: run cmake
+      run: |
+         mkdir build && cd build && cmake ..
+
+    - name: compare generated curl_config.h files
+      run: ./scripts/cmp-config.pl lib/curl_config.h build/lib/curl_config.h
--- a/scripts/cmp-config.pl
+++ b/scripts/cmp-config.pl
@@ -1,0 +1,123 @@
+#!/usr/bin/env perl
+#***************************************************************************
+#                                  _   _ ____  _
+#  Project                     ___| | | |  _ \| |
+#                             / __| | | | |_) | |
+#                            | (__| |_| |  _ <| |___
+#                             \___|\___/|_| \_\_____|
+#
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# This software is licensed as described in the file COPYING, which
+# you should have received as part of this distribution. The terms
+# are also available at https://curl.se/docs/copyright.html.
+#
+# You may opt to use, copy, modify, merge, publish, distribute and/or sell
+# copies of the Software, and permit persons to whom the Software is
+# furnished to do so, under the terms of the COPYING file.
+#
+# This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+# KIND, either express or implied.
+#
+# SPDX-License-Identifier: curl
+#
+###########################################################################
+
+my $autotools = $ARGV[0];
+my $cmake = $ARGV[1];
+
+if(!$cmake) {
+    print "Usage: cmp-config <config1> <config2.h>\n";
+    exit;
+}
+
+# this lists complete lines that will be removed from the the output if
+# matching
+my %remove = (
+    '#define _FILE_OFFSET_BITS 64' => 1,
+    '#define CURL_EXTERN_SYMBOL' => 1,
+    '#define CURL_SA_FAMILY_T sa_family_t' => 1,
+    '#define GETHOSTNAME_TYPE_ARG2 size_t' => 1,
+    '#define HAVE_BROTLI 1' => 1,
+    '#define HAVE_BROTLI_DECODE_H 1' => 1,
+    '#define HAVE_DECL_GETPWUID_R 1' => 1,
+    '#define HAVE_DLFCN_H 1' => 1,
+    '#define HAVE_GETHOSTBYNAME 1' => 1,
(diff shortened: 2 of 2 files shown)


vszakats:
`arc4random` is not present on Windows systems. This saves a little cmake 'generate' time:
```diff
--- a/CMake/Platforms/WindowsCache.cmake
+++ b/CMake/Platforms/WindowsCache.cmake
@@ -52,6 +52,7 @@ if(NOT UNIX)
     set(HAVE_GETHOSTNAME 1)
     set(HAVE_LIBZ 0)
 
+    set(HAVE_ARC4RANDOM 0)
     set(HAVE_ARPA_INET_H 0)
     set(HAVE_FCNTL_H 1)
     set(HAVE_IFADDRS_H 0)
```

.github/workflows/configure-vs-cmake.yml
```diff
@@ -0,0 +1,40 @@
...
+
+name: configure-vs-cmake
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - '*.ac'
+    - 'm4/*'
+    - '**/CMakeLists.txt'
```
vszakats, line 13:
```suggestion
    - '**/CMakeLists.txt'
    - 'lib/curl_config.h.cmake'
```

.github/workflows/configure-vs-cmake.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+    branches:
+    - master
+    paths:
+    - '*.ac'
+    - 'm4/*'
+    - '**/CMakeLists.txt'
```
vszakats, line 20:
```suggestion
    - '**/CMakeLists.txt'
    - 'lib/curl_config.h.cmake'
```

vszakats:
Attempt to address TLS-SRP: #11967

bagder:
I moved the configure and cmake changes I did so far into #11973 so that we can merge fixes gradually even before this is ready to merge, leaving this PR for the script + CI job.

vszakats:
Which on is the "canonical" CA path?:
```
-#define CURL_CA_PATH "/etc/ssl/certs/"
+#define CURL_CA_PATH "/etc/ssl/certs"
```

bagder:
> Probably: `/etc/ssl/certs`

I think so.

vszakats:
- `HAVE_DECL_GETPWUID_R`: could find it in configure as `HAVE_DECL_GETPWUID_R_MISSING`. It aims to detect "old" Android where this function is present but the declaration commented out. Check added in 9c33813d835dde214c1a9449f789dd72ab3d9dc3 (2018-05-27) #2609. CMake was discussed, but didn't get the same check. It's possible the CMake check works only if the declaration is also present, I'm not sure.

- `HAVE_LDAP_URL_PARSE`: This tells to use LDAP API `ldap_url_parse` in place of a whole bunch of local code in `lib/ldap.c` (originally introduced for Windows LDAP, which appears to need it to this day). Macro added in d0edb478964cb5bf7446c34d5ac1e9d34dc5ca63 replacing `WIN32`. configure _does_ check for this. So the issue is that CMake doesn't, simply deriving its value from `NOT HAVE_STRICMP`. This was implemented in the initial CMake commit 4c5307b45655ba75ab066564afdc0c111a8b9291. → It needs a fix in CMake.

- configure makes public all detections, even those used only internally. Some of these don't seem to be used internally even. Some of them could be deleted, but can we somehow tell autotools to not publish results we don't need in the source? Also, configure checks for C89 headers (`HAVE_SETJMP_H`, `HAVE_STDIO_H`, `HAVE_STDLIB_H`, `HAVE_STRING_H`), which seems unnecessary, but I might be missing something. This is the list: `HAVE_GETHOSTBYNAME`, `HAVE_SYS_XATTR_H`, `HAVE_OPENSSL_*_H`, `HAVE_OPENSSL3`, `HAVE_LIBBROTLIDEC`, `HAVE_LIBSSL`, `HAVE_LIBZSTD`, `HAVE_ZSTD`, `HAVE_ZSTD_H`

- `HAVE_DLFCN_H`: only in autotools, but where is it coming from?

- `HAVE_IOCTL_FIONBIO`, `HAVE_IOCTL_SIOCGIFADDR`: detection present in both, but in configure they were dependent on a detection unused (`HAVE_IOCTL`): `CURL_CHECK_FUNC_IOCTL`, so I deleted it in 4d73854462f30948acab12984b611e9e33ee41e6 and c3456652a0c72d1845d08df9769667db7e159949. This was a mistake, as even if we don't use `HAVE_IOCTL`, we did use its dependencies. Ouch!

- `PACKAGE*`, `VERSION`, `LT_OBJDIR`: we need to exclude them from this comparison.

- `USE_MANUAL`: requires `-DUSE_MANUAL=ON` CMake config, which has the side-effect of rebuilding `tool_hugehelp.c`.

- `OS`: we can probably exclude it, or pass an accurate triplet to cmake via `-DCMAKE_C_COMPILER_TARGET=...`.

Rest, TODO:

- only in CMake: `HAVE_SA_FAMILY_T`, `SIZEOF_LONG_LONG`, `_FILE_OFFSET_BITS`

bagder:
> HAVE_SA_FAMILY_T missing from configure

This is currently sorted out in the header file so it does not actually need to be fixed.

vszakats:
Pushed #12015 that fixes `HAVE_LDAP_URL_PARSE` in my first LDAP patch. I forgot to delete the old logic.

There are more problems though: autotools detects LDAP and LDAPS (via libs?), but NOT the LDAP headers, which it checks separately. CMake only checks for headers and that fails. `HAVE_LDAP_SSL` is based on the non-header LDAPS detection in autotools, so it mismatches the simplistic method (= ! disabled + headers detected) CMake still uses.

I'm leaving this to somebody else, as the last weeks were exhausting and I'm starting to make mistakes. I'm also not familiar with LDAP and its history.

bagder:
@vszakats I also think that detection of individual third party libs are less important than getting "the generic" stuff more aligned and I think we have taken a huge step in that direction now.

vszakats:
Agreed, we can tackle the LDAP details later. With some (wildcard?) allowlisting, we're almost there.

It'd even be useful to include this check and ignore the diff exit code for now.

Suggestion: I think results would be easier to read if named `autotools` / `cmake` instead of `one` and `other`.

bagder:
@vszakats I kept the exit code now but have 20-something lines filtered off and now it runs green. I figure we can go live with this and see how it runs over time and adjust it as we go forward. You good with that?

The script outputs the filter lines that are not actually filtered out, so we can figure out if we want to clean the filter going forward.

PR #11966, opened 2023-09-27
autotools: fixup recent bad edits

vszakats:
Follow-up to 96c29900bcec (#11940)

Closes #11966


vszakats:
@bagder: good find! here is two more of the same error I made.

bagder (approved):

MarcelRaad:
Great, Solaris is warning-free again, thanks!
https://curl.se/dev/log.cgi?id=20230927131842-1995410

PR #11967, opened 2023-09-27
cmake: detect TLS-SRP in OpenSSL/wolfSSL/GnuTLS

vszakats:
With new option `CURL_DISABLE_SRP=ON` to force-disable it.
To match existing option and detection logic in autotools.

Also:
- fix detecting GnuTLS.
  We assume `nettle` as a GnuTLS dependency.
- add CMake GnuTLS CI job.
- bump AppVeyor CMake OpenSSL MSVC job to OpenSSL 1.1.1 (from 1.0.2)
  TLS-SRP fails to detect with 1.0.2 due to an OpenSSL header bug.
- fix compiler warning when building with GnuTLS and disabled TLS-SRP.
- fix comment typos, whitespace.

Ref: #11964

Closes #11967


vszakats:
[ This should probably be a 'found', if this OpenSSL 1.0.2 has SRP enabled: https://ci.appveyor.com/project/curlorg/curl/builds/48139282/job/7xbgl6gw8u7a17fy#L33 ]

PR #11973, opened 2023-09-28
Fix configure and cmake mistakes/differences

bagder:
Part of the cleanup work of #11964


CMake/Platforms/WindowsCache.cmake
```diff
@@ -52,6 +52,7 @@ if(NOT UNIX)
     set(HAVE_GETHOSTNAME 1)
     set(HAVE_LIBZ 0)
 
+    set(HAVE_ARC4RANDOM 0)
     set(HAVE_ARPA_INET_H 0)
```
vszakats, line 56:
```suggestion
    set(HAVE_ARPA_INET_H 0)
    set(HAVE_FNMATCH 0)
    set(HAVE_SCHED_YIELD 0)
```
We also cache these for Windows, they are not present there.

CMake/Platforms/WindowsCache.cmake
```diff
@@ -52,6 +52,7 @@ if(NOT UNIX)
     set(HAVE_GETHOSTNAME 1)
     set(HAVE_LIBZ 0)
 
+    set(HAVE_ARC4RANDOM 0)
```
vszakats, line 55:
```suggestion
    set(HAVE_ARC4RANDOM 0)
    set(HAVE_FNMATCH 0)
    set(HAVE_SCHED_YIELD 0)
```

PR #11976, opened 2023-09-28
appveyor: minor improvements

vszakats:
- run `curl -V` after builds to see if they run and with what features.
  Except for one job where a CRT DLL is missing. And ARM64 which should
  fail, but is silently not launched instead.

- copy libcurl DLL next to curl tool and tests binaries in shared mode.
  This makes it possible to run the tests. (We don't run tests after
  these builds yet.)

- list the DLLs and EXEs present after the builds.

- add `DEBUG` variable for CMake builds to allow disabling it, for
  testing non-debug builds. (currently enabled for all)

- add commented lines that dump CMake configuration logs for debugging
  build/auto-detection issues.

- add gcc version to jobs where missing.

- switch a job to the native MSYS2 mingw-w64 toolchain. This adds gcc 9
  to the build mix.

- make `SHARED=OFF` and `OPENSSL=OFF` defaults global.

- delete a duplicate backslash.

Closes #11976

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,6 +29,7 @@ version: 7.50.0.{build}
 
 environment:
     UNITY: "OFF"
+    SHARED: "OFF"
     matrix:
       # generated CMake-based Visual Studio Release builds
       - job_name: "CMake, VS2008, Release x86, Schannel"
@@ -40,7 +41,6 @@ environment:
         SCHANNEL: ON
         ENABLE_UNICODE: OFF
         HTTP_ONLY: OFF
-        DEBUG: ON
         TESTING: OFF
         SHARED: ON
         DISABLED_TESTS: ""
@@ -54,7 +54,6 @@ environment:
         SCHANNEL: OFF
         ENABLE_UNICODE: OFF
         HTTP_ONLY: OFF
-        DEBUG: ON
         TESTING: OFF
         SHARED: ON
         DISABLED_TESTS: ""
@@ -70,9 +69,8 @@ environment:
         SCHANNEL: ON
         ENABLE_UNICODE: OFF
         HTTP_ONLY: OFF
-        DEBUG: ON
+        DEBUG: "ON"
         TESTING: OFF
-        SHARED: OFF
         DISABLED_TESTS: ""
       # generated CMake-based Visual Studio Debug builds
       - job_name: "CMake, VS2010, Debug x64, no SSL, Static"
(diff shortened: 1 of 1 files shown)


vszakats:
```
/c/OpenSSL-v111-Win64/bin/capi.dll
/c/OpenSSL-v111-Win64/bin/dasync.dll
/c/OpenSSL-v111-Win64/bin/libcrypto-1_1-x64.dll
/c/OpenSSL-v111-Win64/bin/libssl-1_1-x64.dll
/c/OpenSSL-v111-Win64/bin/ossltest.dll
/c/OpenSSL-v111-Win64/bin/padlock.dll
/c/OpenSSL-v111-Win64/libcrypto-1_1-x64.dll
/c/OpenSSL-v111-Win64/libssl-1_1-x64.dll
/c/OpenSSL-v111-Win64/tests/fuzz/libcrypto-1_1-x64.dll
/c/OpenSSL-v111-Win64/tests/fuzz/libssl-1_1-x64.dll
/c/OpenSSL-v111-Win64/tests/libcrypto-1_1-x64.dll
/c/OpenSSL-v111-Win64/tests/libssl-1_1-x64.dll
```

jay:
On Windows you can use curl --dump-module-paths (undocumented) to see which DLLs are actually being loaded

vszakats:
`--dump-module-paths` needs `curl.exe` to start in the first place! There are some issues to fix to get there: missing libcurl DLL, missing OpenSSL DLLs, debug+unicode+unity combo crash bug, ARM64 exe, at least one unidentified case and maybe cases where the location of the built `curl.exe` isn't known.

jay:
> `--dump-module-paths` needs `curl.exe` to start in the first place

ah. dependency walker may work for that eg `depends /c /ot:outfile curld.exe` and make outfile an artifact that can be downloaded.

vszakats:
I didn't know about `depends`, thanks! With the mingw-w64 toolchain I use this:
```shell
x86_64-w64-mingw32-objdump --all-headers curl.exe | grep -a -F -i 'DLL Name' | sort -r -f
```

vszakats:
Managed to fix all cases to run `curl.exe` except `CMake, VS2008, Release x86, Schannel`, that is missing `MSVCR90.dll`. Run remains disabled for that job. Otherwise this PR is complete. I plan to follow-up with a yamllint/formatting/indenting patch.

vszakats:
Turns out the MS-DOS batch script is ending up not running `curl.exe` in most cases (hence the suspiciously silent ARM64 build), so I'm attempting a rewrite in bash. (if that fails, in PowerShell)

PR #11979, opened 2023-09-29
cmake: detect `HAVE_GETADDRINFO_THREADSAFE`

vszakats:
Based on existing autotools logic.

autotools checks for old versions of the allowlisted target OSes and
disables this feature when seeing them. In CMake we assume we're running
on newer systems and enable regardless of OS version.

autotools always runs all 3 probes for non-fast-tracked systems and
enables this feature if any one of them was successful. To save
configuration time,  CMake stops at the first successful check.

OpenBSD is not fast-tracked and then gets blocklisted as a generic BSD
system. I haven't double-checked if this is correct, but looks odd.

Ref: #11964 (effort to sync cmake detections with autotools)

Closes #11979

---

I might have misread or overlooked autotools logic. Please review.


CMake/OtherTests.cmake
```diff
@@ -132,3 +133,69 @@ if(NOT CMAKE_CROSSCOMPILING)
     }" HAVE_POLL_FINE)
   endif()
 endif()
+
+# Detect HAVE_GETADDRINFO_THREADSAFE
+
+if(WIN32)
+  set(HAVE_GETADDRINFO_THREADSAFE ${HAVE_GETADDRINFO})
+elseif(CMAKE_SYSTEM_NAME STREQUAL "AIX" OR
```
jay, line 141: note autotools only checks the OSes if HAVE_GETADDRINFO
https://github.com/curl/curl/blob/6fa1d817e5b1a00d7d0c8168091877476b499317/m4/curl-functions.m4#L2011-L2014

so it might need a check like 
```suggestion
elseif(NOT HAVE_GETADDRINFO)
  set(HAVE_GETADDRINFO_THREADSAFE FALSE)
elseif(CMAKE_SYSTEM_NAME STREQUAL "AIX" OR
```

vszakats: Thank you!

PR #11984, opened 2023-09-29
FAQ: How do I upgrade curl.exe in Windows?

bagder:
This is a growing question, better answer it here to get somewhere to point users to.

### DIFF
--- a/docs/FAQ
+++ b/docs/FAQ
@@ -441,6 +441,9 @@ FAQ
   managed and controlled entirely by Microsoft as owners of the operating
   system.
 
+  You can always download and install the latest version of curl for Windows
+  from https://curl.se/windows into a separate location.
+
   2.4 Does curl support SOCKS (RFC 1928) ?
 
   Yes, SOCKS 4 and 5 are supported.
--- a/docs/FAQ
+++ b/docs/FAQ
@@ -26,6 +26,7 @@ FAQ
  2. Install Related Problems
   2.1 configure fails when using static libraries
   2.2 Does curl work/build with other SSL libraries?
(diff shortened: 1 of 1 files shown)


docs/FAQ
```diff
@@ -427,6 +428,22 @@ FAQ
   have their pros and cons, and we try to maintain a comparison of them here:
   https://curl.se/docs/ssl-compared.html
 
+  2.3 How do I upgrade curl.exe in Windows?
+
+  The curl tool that is shipped as an integrated component of Windows 10 and
+  Windows 11 is managed by Microsoft. If you were to delete the file or
+  replace it with a newer version downloaded from https://curl.se/windows,
+  then Windows Update will cease to work on your system.
```
jay, line 436: Saying "Windows Update will cease to work on your system" seems like overkill, I'd think it would update the tool or not, I'm not sure which I've never tried it. Anyone know what happens?

bagder: It completely stops windows update because it detects windows as having been tampered with. I think my phrasing here is correct. It will not update *anything* after you do that.

jay: What OS? I tried it in Windows 10 and I cannot reproduce. I resumed an old VM image running 20H2, took ownership of curl.exe and deleted it, then updated Windows. The first few updates were successful and curl.exe was not added back. Then I installed the cumulative update, also successful, and it had readded curl.exe (but a later version 7.79.1).

bagder: I have not done this myself, but I have had multiple persons contacting me about it.

jay: Ok fair enough.

Issue #11982, opened 2023-09-29
libcurl 8.3.0 erroneously closes paused transfers with unknown content-length


piru:
### I did this

```
/* Proof of Concept for libcurl 8.3.0 bug where paused downloads
 * with unknown Content-Length suddendly get errornously reported
 * as completed.
 *
 * Issue discovered and PoC written by Harry Sintonen <sintonen@iki.fi>
 *
 * Loosely based on https://curl.se/libcurl/c/multi-app.html
 */

#include <stdio.h>
#include <string.h>
#include <sys/time.h>
#include <unistd.h>
#include <stdlib.h>
#include <curl/curl.h>

#define HANDLECOUNT 2

int err(void)
{
  fprintf(stderr, "something unexpected went wrong - bailing out!\n");
  exit(2);
}

struct handle
{
  int idx;
  int paused;
  CURL *h;
};

static size_t cb(void *data, size_t size, size_t nmemb, void *clientp)
{
  size_t realsize = size * nmemb;
  struct handle *handle = (struct handle *) clientp;
  curl_off_t totalsize;

  printf("handle %d write %lu bytes\n", handle->idx, realsize);

  if(curl_easy_getinfo(handle->h, CURLINFO_CONTENT_LENGTH_DOWNLOAD_T, &totalsize) == CURLE_OK)
    printf("handle %d Content-Length %"CURL_FORMAT_CURL_OFF_T"\n", handle->idx, totalsize);

  handle->paused = 1;

  return CURL_WRITEFUNC_PAUSE;
}

int main(void)
{
  struct handle handles[HANDLECOUNT];
  CURLM *multi_handle;
  int i, still_running = 1, msgs_left, numfds;
  CURLMsg *msg;
  int rounds = 0;
  int rc = 1;

  curl_global_init(CURL_GLOBAL_DEFAULT);

  for(i = 0; i<HANDLECOUNT; i++) {
    handles[i].idx = i;
    handles[i].paused = 0;
    handles[i].h = curl_easy_init();
    if(!handles[i].h ||
      curl_easy_setopt(handles[i].h, CURLOPT_WRITEFUNCTION, cb) != CURLE_OK ||
      curl_easy_setopt(handles[i].h, CURLOPT_WRITEDATA, &handles[i]) != CURLE_OK ||
      curl_easy_setopt(handles[i].h, CURLOPT_FOLLOWLOCATION, 1L) != CURLE_OK ||
[...]

PR #11989, opened 2023-09-29
Testcase and fix for pausing h2 streams

icing:
- refs #11982 where it was noted that paused transfers may close successfully without delivering the complete data
- made sample poc into tests/http/client/h2-pausing.c and added test_02_27 to reproduce

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -1046,6 +1046,19 @@ static CURLcode readwrite_upload(struct Curl_easy *data,
   return CURLE_OK;
 }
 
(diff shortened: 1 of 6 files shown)


icing:
Linux event-based is timing out after 90 min.

bagder:
@icing it happens already without this PR though, like in #11990  ...

bagder (approved):

bagder:
@icing I presume this is ready to land?

icing:
> @icing I presume this is ready to land?

Yes. I was just hesitant because of the event-based thing. But if that is caused by another merge, this one is ready.

PR #11993, opened 2023-09-29
Shawn.webb/features/tld blocklist

lattera:
Provide plumbing for general domain blocklisting

This provides the plumbing for a more dynamic approach to domain blocklisting. Using a dynamic approach will allow developers and users alike to choose their security preferences.

As part of this work, remove the prohibition on the `.onion` TLD. Users who want to block Tor Onion Services can add `.onion` to the blocklist. Alternatively, if the curl project feels strongly about keeping the prohibition, re-adding `.onion` and `.onion.` to the blocklist is a simple and easy development task.

Siged-off-by: Shawn Webb <shawn.webb@hardenedbsd.org>


dfandrich:
What's the real-world use case for this, besides .onion?

gvanem:
This is great! 
*PS*. how about support for blocked domains in `.curlrc` ?

lattera:
> What's the real-world use case for this, besides .onion?

I could see how one might want to maintain a blocklist to prevent DNS leaks. Or to selectively fetch resources, especially in a recursive fetching context. Or other reasons, of course.

lattera:
> This is great! _PS_. how about support for blocked domains in `.curlrc` ?

Good suggestion! If I have some spare cycles, I'll definitely look into that. If someone else beats me to the punch, then I'd be happy to incorporate their commit(s) into this PR.

bagder:
I want to second @dfandrich's question: Please talk about real-world use cases for this.

And when you do, please elaborate on why libcurl needs to do this and not the application itself. We offer a URL parser API, the application can easily block domains by itself.

lattera:
> I want to second @dfandrich's question: Please talk about real-world use cases for this.
> 
> And when you do, please elaborate on why libcurl needs to do this and not the application itself. We offer a URL parser API, the application can easily block domains by itself.

Should the hardcoded prohibition on `.onion` be removed from libcurl? I feel this is a good compromise between a hardcoded prohibition (an extremely and overly broad one at that, given libcurl's wide adoption) on `.onion` and providing the means for opting in to (or out of) a blockade.

This patch solves the real-world use case scenario of providing HardenedBSD's Tor Onion Service users with the ability to still fetch packages and OS updates from behind a fully Tor-ified network. Without this patch, an entire part of the public internet is blocked with no method to opt out of the blockade/prohibition.

This patch is by no means completely finished. I'm disinterested in re-adding the prohibition on `.onion` in this pull request. I suspect the curl project would want the prohibition re-added (and that's their choice to make), but at least this pull request provides the plumbing to make that prohibition a choice to be opted in to or out of, rather than hardcoded.

jay:
I don't know whether or not this is needed. If it is then the way it is implemented will need to be fixed to handle a list instead of items added to a list internally. See [CURLOPT_RESOLVE](https://curl.se/libcurl/c/CURLOPT_RESOLVE.html) for example.

lattera:
> I don't know whether or not this is needed. If it is then the way it is implemented will need to be fixed to handle a list instead of items added to a list internally. See [CURLOPT_RESOLVE](https://curl.se/libcurl/c/CURLOPT_RESOLVE.html) for example.

I'm happy to take care of that. I'll follow up with a commit either later this evening or over the weekend.

bagder:
RFC 7686 is not about "blocking domains", it is about not leaking `.onion` names over DNS. Most libcurl users won't even know about 7686 nor `.onion` and I presume that is exactly why the RFC is written the way it is: it says we should prevent the name from being resolved with DNS.

I understand what you're after, but I don't think adding a custom domain-blocking option is the correct way to let the whole world go back to the `.onion` leak.

lattera:
> RFC 7686 is not about "blocking domains", it is about not leaking `.onion` names over DNS. Most libcurl users won't even know about 7686 nor `.onion` and I presume that is exactly why the RFC is written the way it is: it says we should prevent the name from being resolved with DNS.

I completely understand that the RFC's intent is not to block `.onion`. However, the current implementation in libcurl has that effect. The problem here is the implementation in libcurl of the relevant section of the RFC and not the RFC itself.

> I understand what you're after, but I don't think adding a custom domain-blocking option is the correct way to let the whole world go back to the `.onion` leak.

Do you have any other suggestions for a compromise I could implement?

dfandrich:
Join the discussion on #11125

bagder:
> I completely understand that the RFC's intent is not to block .onion. However, the current implementation in libcurl has that effect

... for the subset of users who use transparent and not explicit proxies.

> Do you have any other suggestions for a compromise I could implement?

First: this is not all about "blocking domains", so I think we can start with changing the terminology, The currently existing logic only prevents `.onion` names from being resolved locally. Such host names can still be used perfectly fine over proxies: HTTP/HTTPS/socks4/socks5.

We've been talking about this subject back and forth in #11125 for a while but so far a "final solution" has not been seen.

lattera:
> > I completely understand that the RFC's intent is not to block .onion. However, the current implementation in libcurl has that effect
> 
> ... for the subset of users who use transparent and not explicit proxies.
> 
> > Do you have any other suggestions for a compromise I could implement?
> 
> First: this is not all about "blocking domains", so I think we can start with changing the terminology, The currently existing logic only prevents `.onion` names from being resolved locally. Such host names can still be used perfectly fine over proxies: HTTP/HTTPS/socks4/socks5.

Environments exist where extreme care must be taken for all traffic, not just proxy-able traffic. Entire devices must not be immediately aware that they're behind a non-traditional network (Tor transparent proxy as a prime example.)

These environments exist for specific reasons, usually with regards to human life. Forcing a technology that cannot be used (a SOCKS proxy in this case) for fear of real-world safety (or even technical capability as not all devices support setting a device-wide SOCKS proxy) seems counter-intuitive.

> 
> We've been talking about this subject back and forth in #11125 for a while but so far a "final solution" has not been seen.

What does a "final solution" to this issue look like to you?

bagder:
> Forcing a technology

I would argue that curl does not force any technology onto anyone. It's just a piece of open source code. You can use something else.

> What does a "final solution" to this issue look like to you?

I don't know. I haven't seen it yet.

jay:
> It's just a piece of open source code. You can use something else.

how about a patch for his distro

~~~diff
diff --git a/lib/hostip.c b/lib/hostip.c
index acb4b51..c3ff9ec 100644
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -683,14 +683,6 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
   CURLcode result;
   enum resolve_t rc = CURLRESOLV_ERROR; /* default to failure */
-  /* We should intentionally error and not resolve .onion TLDs */
-  size_t hostname_len = strlen(hostname);
-  if(hostname_len >= 7 &&
-     (curl_strequal(&hostname[hostname_len - 6], ".onion") ||
-      curl_strequal(&hostname[hostname_len - 7], ".onion."))) {
-    failf(data, "Not resolving .onion address (RFC 7686)");
-    return CURLRESOLV_ERROR;
-  }
   *entry = NULL;
 #ifndef CURL_DISABLE_DOH
   conn->bits.doh = FALSE; /* default is not */
~~~

lattera:
> > It's just a piece of open source code. You can use something else.
> 
> how about a patch for his distro
> 
> ```diff
> diff --git a/lib/hostip.c b/lib/hostip.c
> index acb4b51..c3ff9ec 100644
> --- a/lib/hostip.c
> +++ b/lib/hostip.c
> @@ -683,14 +683,6 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
>    CURLcode result;
>    enum resolve_t rc = CURLRESOLV_ERROR; /* default to failure */
>    struct connectdata *conn = data->conn;
> -  /* We should intentionally error and not resolve .onion TLDs */
> -  size_t hostname_len = strlen(hostname);
> -  if(hostname_len >= 7 &&
> -     (curl_strequal(&hostname[hostname_len - 6], ".onion") ||
> -      curl_strequal(&hostname[hostname_len - 7], ".onion."))) {
> -    failf(data, "Not resolving .onion address (RFC 7686)");
> -    return CURLRESOLV_ERROR;
> -  }
>    *entry = NULL;
>  #ifndef CURL_DISABLE_DOH
>    conn->bits.doh = FALSE; /* default is not */
> ```

We've already done that in HardenedBSD. But that doesn't help non-HardenedBSD systems living inside a Tor-ified network.

piru:
I see no problem with this feature as long as curl keeps blocking `.onion` resolving by default.

piru:
What I have problem with is deliberately confrontational attitude and rudeness.

lib/hostip.c
```diff
@@ -652,6 +667,35 @@ static bool tailmatch(const char *full, const char *part)
   return strncasecompare(part, &full[flen - plen], plen);
 }
 
+static bool is_hostname_blocked(struct Curl_easy *data, const char *hostname)
```
piru, line 670: This doesn't scale very well. In many situations (such as blocking spam) you could have hundred of thousands of entries in the blocklist. I'd much rather see a solution using a bloom filter or similar.

piru:
Generally speaking far more flexible solution is to use adblockplus (abp) https://adblockplus.org/filter-cheatsheet compatible filters lists in the applications. There are number of different implementations which use bloom filters, making them very scalable. It's however debatable if such feature belong to libcurl in the first place.

lib/setopt.c
```diff
@@ -3170,6 +3170,16 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
   case CURLOPT_QUICK_EXIT:
     data->set.quick_exit = (0 != va_arg(param, long)) ? 1L:0L;
     break;
+  case CURLOPT_BLOCK_DOMAIN:
+    /*
+     * A string that defines a domain to block
+     */
+    argptr = va_arg(param, char *);
```
piru, line 3177: Generally libcurl doesn't require the caller to keep the passed parameters such as strings in memory after `curl_easy_setopt`. Thus this is not correct in a way that it doesn't clone the passed string.

```Strings passed to libcurl as 'char *' arguments, are copied by the library; the string storage associated to the pointer argument may be discarded or reused after curl_easy_setopt returns.```
source: https://curl.se/libcurl/c/curl_easy_setopt.html

Pachin0:
Thank you everyone who had the sensibility of understanding that giving users choice is something we should strive for.

lib/hostip.c
lattera: I thought about that, too. However, I think implementing something more complex might drive this pull request out-of-scope for its intended purpose: provide the plumbing for making dynamic the blocklisting of domains.

I suspect the underlying data structures and algorithms could be improved upon at a later stage of development. I also think this tiny little feature won't be used much, or with large numbers of entries, in its early days.

vszakats:
I can see value in app-level blocking if it's able to block complete URLs (not just domains), especially if it's compatible with existing rule standards, as mentioned by @piru.

bagder:
I would expect that a "block domains" or "block URLs" feature would mean that they would be blocked completely from transfers and not be allowed over proxies. Which is what the .onion handling allows today.  So, not quite the same thing.

vszakats:
> I would expect that a "block domains" or "block URLs" feature would mean that they would be blocked completely from transfers and not be allowed over proxies. Which is what the .onion handling allows today. So, not quite the same thing.

Yes. I intentionally didn't mention `.onion` as I believe it has nothing to do with a generic blocking solution, this PR included. Domains can already be blocked via `--doh-url` or custom DNS servers such as pi-hole. This is more efficient because it can be applied to whole machines or the complete network. Full URLs however can only be seen by curl. But, even there the value is not so high, as the URLs accessed are pretty much what the user passes to curl, except for redirections and perhaps globbing. It means in most cases the caller can filter before passing them to curl.

lattera:
I believe I've addressed all technical issues within scope of this pull request. Please let me know what additional changes need to be made, if any.

bagder:
Mostly, this wants to add a feature nobody has asked for so I'm reluctant to spend time on doing a detailed review. But if you insist, here are some remaining problems:

- there is no docs
- there is no tests
- this leaks `.onion` by default
- it causes numerous CI failures
- it does not "block domain" so the naming is wrong
- it does not accept an slist in style of other similar options

Pachin0:
There is a long thread of people specifically wanting this feature.

bagder:
> There is a long thread of people specifically wanting this feature.

Can you direct me to where people are asking for a feature to block selected domains from being resolved with DNS? The only thread I've seen contains a few people asking to be able resolve `.onion` host names. Which is more like an opposite of this proposal.

lattera:
> Mostly, this wants to add a feature nobody has asked for so I'm reluctant to spend time on doing a detailed review. But if you insist, here are some remaining problems:
> 
>     * there is no docs

I'm happy to write those missing docs once we determine a better name for the feature (see my comment about the name below).

> 
>     * there is no tests

I'm happy to write those missing tests.

> 
>     * this leaks `.onion` by default

Again, the goal of this pull request is to make a generic, re-usable method for blocking DNS resolution for certain records. Given my disinterest in adding that prohibition back in, a future pull request could be made by interested parties to put that prohibition back in place; however, using a more dynamic approach like what this pull request offers.

This pull request, then, could serve as a basis for supporting the proposed `--ignore-onion` and `--no-ignore-onion` option. Just append two entries to the blocklist if the decision is made to prohibit resolution of Tor Onion Services: `.onion` and `.onion.`.

Hopefully this patch may be even more useful to someone who might have other use cases--ones I couldn't think of. Usually making a concept generic and providing it to a wider set of use cases helps both developers and users alike.

> 
>     * it causes numerous CI failures

I've been paying attention to the CI failures and have been addressing them. I do need to take a deeper look, and if this pull request is the cause of the failures, I will indeed address them.

> 
>     * it does not "block domain" so the naming is wrong

I do admittedly suck at naming things. Would you have a more preferred name for the feature? I have no preference on the name.

> 
>     * it does not accept an slist in style of other similar options

I thought I addressed this, but perhaps not fully (or not correctly.) Can you help me understand what I need to change with regards to the slist?

bagder:
> a future pull request could be made by interested parties to put that prohibition back in place

We prevent `.onion` names from getting leaked over DNS today. We can discuss providing options for how that can be modified at run-time, but I remain convinced that is the sensible default. According to me, in order to maintain behavior, a way that re-introduces .oinon leakage needs to be opt-in.

lattera:
> > a future pull request could be made by interested parties to put that prohibition back in place
> 
> We prevent `.onion` names from getting leaked over DNS today. We can discuss providing options for how that can be modified at run-time, but I remain convinced that is the sensible default. According to me, in order to maintain behavior, a way that re-introduces .oinon leakage needs to be opt-in.

This can be addressed by someone else in a separate pull request. This pull request is not meant to make a decision as to whether Tor Onion Service access via Tor transparent proxy should be prohibited. This pull request is ONLY meant to provide easy dynamic plumbing for implementing those kinds of decisions.

bagder:
> This can be addressed by someone else in a separate pull request

Sorry, but we cannot merge a PR that breaks behavior and assigns work onto mr Someone Else. That's an unfinished PR.

bagder:
As I have repeatedly said: we need to work out a solution that we think works, then we can write the code for that solution. Such a solution should probably be discussed, critiqued and vetted by the peeps in #11125.

I do not have a magic solution. Then I would have explained that already.

PR #12006, opened 2023-10-02
cmake: fix `HAVE_LDAP_SSL`, `HAVE_LDAP_URL_PARSE` on non-Windows

vszakats:
- set `HAVE_LDAP_URL_PARSE` if `ldap_url_parse` function exists.
  Before this patch we set it based it on the presence of `stricmp`,
  which correctly enabled it on e.g. Windows, but was inaccurate for
  other platforms.

- always set `HAVE_LDAP_SSL` if an LDAP backend is detected and
  LDAPS is not explicitly disabled. This mimics autotools behaviour.
  Previously we set it only for Windows LDAP. After this fix, LDAPS is
  correctly enabled e.g. in default macOS builds.

- enable LDAP[S] for a CMake macOS CI job. Target OS X 10.9 (Mavericks)
  to avoid deprecation warnings for LDAP API.

- always detect `HAVE_LDAP_SSL_H`, even with LDAPS explicitly disabled,
  to sync behaviour with autotools.

- fix benign typo in internal variable name.

- eliminate an internal variable.

Ref: #11964 (effort to sync cmake detections with autotools)

Closes #12006


vszakats:
Merged this one minor cleanup too early. The follow-up is in 4e8a3a1fc83b98c8803eb81124d2e7d9ca0b5974.

PR #12011, opened 2023-10-03
Client writer, part3, eliminate `k->str`

icing:
This PR is on top of https://github.com/curl/curl/pull/11961 and in addition has these changes:

- changed header/chunk/handler->readwrite prototypes to
      accept `buf`, `blen` and a `pconsumed` pointer. They
      now get the buffer to work on and report back how many
      bytes they consumed
- this eliminates `k->str` in SingleRequest, a pointer to *somewhere* in `data->state.buffer` (hopefully) and ideally 0-terminated (maybe)
- improved excess data handling to properly calculate with any body data left in the `headerb` buffer
- eliminated `k->badheader` enum differences for it to be a `bool`


icing:
will rebase and remake once part2 is merged

PR #12012, opened 2023-10-03
Client writer, part4, correct excess data checks

icing:
Move excess body data handling into `download` cwriter (`PROTOCOL` phase)
- means the size is checked after all `Transfer-Encoding`s have been decoded, thus counting the *actual* content length
- resolves some quirky dances for trying to get `chunked`, `buffer` and `headerb` accounted for in `transfer.c`

Make Curl_readwrite_headers() set `data->req.download_done` instead of having a `bool*` arg
- move k->keepon modification from http methods into transfer loop


icing:
will rebase, remake once part3 is merged

PR #12025, opened 2023-10-04
Improve reliability and debuggability of tests

dfandrich:
- Use more test macros to catch errors in libtests early
- Log more in case of errors
- Adjust timeouts for improved reliability
- ignore "timing-dependent" test in CMake CI builds

### DIFF
--- a/tests/data/test1130
+++ b/tests/data/test1130
@@ -52,7 +52,7 @@ http
 HTTP POST forced expect 100-continue with a 404
 </name>
  <command option="no-output">
--d @%LOGDIR/file%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER0001 -H "Expect: 100-continue" --expect100-timeout 999
+-d @%LOGDIR/file%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER0001 -H "Expect: 100-continue" --expect100-timeout 99
 </command>
 </client>
 
--- a/tests/libtest/lib1518.c
+++ b/tests/libtest/lib1518.c
@@ -81,7 +81,7 @@ int test(char *URL)
   curl_easy_getinfo(curl, CURLINFO_REDIRECT_COUNT, &curlRedirectCount);
   curl_easy_getinfo(curl, CURLINFO_EFFECTIVE_URL, &effectiveUrl);
   curl_easy_getinfo(curl, CURLINFO_REDIRECT_URL, &redirectUrl);
-  res = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writecb);
+  test_setopt(curl, CURLOPT_WRITEFUNCTION, writecb);
 
   printf("res %d\n"
          "status %d\n"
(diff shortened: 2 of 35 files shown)


tests/data/test1130
```diff
@@ -52,7 +52,7 @@ http
 HTTP POST forced expect 100-continue with a 404
  </name>
  <command option="no-output">
--d @%LOGDIR/file%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER0001 -H "Expect: 100-continue" --expect100-timeout 999
+-d @%LOGDIR/file%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER http://%HOSTIP:%HTTPPORT/%TESTNUMBER0001 -H "Expect: 100-continue" --expect100-timeout 99
```
jay, line 55: why change these tests to have a shorter expect timeout? isn't the longer timeout more reliable?

tests/libtest/lib1518.c
```diff
@@ -81,7 +81,7 @@ int test(char *URL)
   curl_easy_getinfo(curl, CURLINFO_REDIRECT_COUNT, &curlRedirectCount);
   curl_easy_getinfo(curl, CURLINFO_EFFECTIVE_URL, &effectiveUrl);
   curl_easy_getinfo(curl, CURLINFO_REDIRECT_URL, &redirectUrl);
-  res = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writecb);
+  test_setopt(curl, CURLOPT_WRITEFUNCTION, writecb);
```
jay, line 84: I've noticed, not specific to this PR, that some of the tests use macro [test_setopt](https://github.com/curl/curl/blob/curl-8_3_0/tests/libtest/test.h#L49-L51) and some use macro [easy_setopt](https://github.com/curl/curl/blob/curl-8_3_0/tests/libtest/test.h#L173-L193). 

tests dir:
~~~
$ git grep '[^_]easy_setopt' |wc -l
    239

$ git grep test_setopt |wc -l
    676
~~~

What is the purpose of using macro test_setopt instead of macro easy_setopt? If the option can't be set then the former doesn't set an error message while the latter does. Both quit (goto test_cleanup). Wouldn't reliability be improved by always using the easy_setopt?

tests/data/test1130
dfandrich: This is a) for consistency, and b) for expediency. If something goes horribly wrong and the timeout needs to be used to end the test, this will cause a 16 minute delay before the test exits, while 99 seconds is way more than enough already.

tests/libtest/lib1518.c
dfandrich: I didn't notice that there were two macros and I thought I *was* using `easy_setopt`. `easy_setopt` is a bit broken, though, in that it will exit if a *previous* function set res to an error (`easy_setopt` does not always set `res`). These macros should probably be unified (and that problem fixed), but maybe there's a case where both need to exist.

dfandrich:
This change makes test 1086 run again (with the result ignored) which is showing a strange problem in the test harness on some Windows builds: `Warning: unable to close filehandle DWRITE properly: Broken pipe at C:/projects/curl/tests/ftpserver.pl line 527`  I'm Just going to make that test stop running again in those environments again rather than debug it now.

PR #12034, opened 2023-10-05
CURLOPT_DEBUGFUNCTION.3: warn about internal handles

jay:
- Warn that the user's debug callback may be called with the handle parameter set to an internal handle.

Without this warning the user may assume that the only handles their debug callback receives are the easy handles on which they set CURLOPT_DEBUGFUNCTION.

This is a follow-up to f8cee8cc which changed DoH handles to inherit the debug callback function set in the user's easy handle. As a result those handles are now passed to the user's debug callback function.

Closes #xxxx

---

I did not mention DoH specifically in case we want to expand on it later, like for example in the future we may want to make the connection closure internal handle inherit the user's debug function and verbosity. (See #11878)

### DIFF
--- a/docs/libcurl/opts/CURLOPT_DEBUGFUNCTION.3
+++ b/docs/libcurl/opts/CURLOPT_DEBUGFUNCTION.3
@@ -62,6 +62,7 @@ not null-terminated, but is exactly of the \fIsize\fP as told by the
 The \fIclientp\fP argument is the pointer set with \fICURLOPT_DEBUGDATA(3)\fP.
 
 Available \fBcurl_infotype\fP values:
+.RS
 .IP CURLINFO_TEXT
 The data is informational text.
 .IP CURLINFO_HEADER_IN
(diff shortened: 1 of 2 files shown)


docs/libcurl/opts/CURLOPT_DEBUGFUNCTION.3
```diff
@@ -79,6 +80,11 @@ The data is protocol data sent to the peer.
 The data is SSL/TLS (binary) data sent to the peer.
 .IP CURLINFO_SSL_DATA_IN
 The data is SSL/TLS (binary) data received from the peer.
+.RE
+
+WARNING: This callback may be called with the curl \fIhandle\fP set to an
+internal handle. If you need to distinguish your handle from internal handles
+then set \fICURLOPT_PRIVATE(3)\fP on your handle. (Added in 8.4.0)
```
bagder, line 87: I think you should move the `(Added)` part to the first sentence, as otherwise it could be read as if the PRIVATE thing is the new addition.

jay: Done. I made it two separate paragraphs. Also I added a `DEBUGASSERT(!data->set.private_data);` before the callback is called.

bagder: If the application only wants to output what is sent/received, it may not care to differentiate which handle is doing it and then it does not need to set `CURLOPT_PRIVATE` at all. So isn't the assert asserting a bit too much?

jay: The assert only checks that private data is not set for an internal handle, the code is 
~~~c
      if(data->internal)
        DEBUGASSERT(!data->set.private_data);
~~~

bagder: Ah, yes that makes sense!

bagder (approved):

PR #12044, opened 2023-10-06
cmake: pre-fill rest of detection values for Windows

vszakats:
The goal of this patch is to avoid unnecessary feature detection work
when doing Windows builds with CMake. Do this by pre-filling well-known
detection results for Windows and specifically for mingw-w64 and MSVC
compilers. Also limit feature checks to platforms where the results are
actually used. Drop a few redundant ones. And some tidying up.

- pre-fill remaining detection values in Windows CMake builds.

  Based on actual detection results observed in CI runs, preceding
  similar work over libssh2 and matching up values with
  `lib/config-win32.h`.

  This brings down CMake configuration time from 58 to 14 seconds on the
  same local machine.

  On AppVeyor CI this translates to:
  - 128 seconds -> 50 seconds VS2022 MSVC with OpenSSL (per CMake job):
    https://ci.appveyor.com/project/curlorg/curl/builds/48208419/job/4gw66ecrjpy7necb#L296
    https://ci.appveyor.com/project/curlorg/curl/builds/48217440/job/8m4fwrr2fe249uo8#L186
  - 62 seconds -> 16 seconds VS2017 MINGW (per CMake job):
    https://ci.appveyor.com/project/curlorg/curl/builds/48208419/job/s1y8q5ivlcs7ub29?fullLog=true#L290
    https://ci.appveyor.com/project/curlorg/curl/builds/48217440/job/pchpxyjsyc9kl13a?fullLog=true#L194

  The formula is about 1-3 seconds delay for each detection. Almost all
  of these trigger a full compile-link cycle behind the scenes, slow
  even today, both cross and native, mingw-w64 and apparently MSVC too.
  Enabling .map files or other custom build features slows it down
  further. (Similar is expected for autotools configure.)

- stop detecting `idn2.h` if idn2 was deselected.
  autotools does this.

- stop detecting `idn2.h` if idn2 was not found.
  This deviates from autotools. Source code requires both header and
  lib, so this is still correct, but faster.

- limit `ADDRESS_FAMILY` detection to Windows.

- normalize `HAVE_WIN32_WINNT` value to lowercase `0x0a12` format.

- pre-fill `HAVE_WIN32_WINNT`-dependent detection results.
  Saving 4 (slow) feature-detections in most builds: `getaddrinfo`,
  `freeaddrinfo`, `inet_ntop`, `inet_pton`

- fix pre-filled `HAVE_SYS_TIME_H`, `HAVE_SYS_PARAM_H`,
  `HAVE_GETTIMEOFDAY` for mingw-w64.
  Luckily this do not change build results, as `WIN32` took
  priority over `HAVE_GETTIMEOFDAY` with the current source
  code.

- limit `HAVE_CLOCK_GETTIME_MONOTONIC_RAW` and
  `HAVE_CLOCK_GETTIME_MONOTONIC` detections to non-Windows.
  We're not using these in the source code for Windows.

- reduce compiler warning noise in CMake internal logs:
  - fix to include `winsock2.h` before `windows.h`.
    Apply it to autotools test snippets too.
  - delete previous `-D_WINSOCKAPI_=` hack that aimed to fix the above.
  - cleanup `CMake/CurlTests.c` to emit less warnings.

- delete redundant `HAVE_MACRO_SIGSETJMP` feature check.
  It was the same check as `HAVE_SIGSETJMP`.

- delete "experimental" marking from `CURL_USE_OPENSSL`.

- show CMake version via `CMakeLists.txt`.
  Credit to the `zlib-ng` project for the idea:
  https://github.com/zlib-ng/zlib-ng/blob/61e181c8ae93dbf56040336179c9954078bd1399/CMakeLists.txt#L7

- make `CMake/CurlTests.c` pass `checksrc`.

- `CMake/WindowsCache.cmake` tidy-ups.

- replace `WIN32` guard with `_WIN32` in `CMake/CurlTests.c`.

Closes #12044

---

Clearer without whitespace changes: https://github.com/curl/curl/pull/12044/files?w=1


vszakats:
Also tested an experimental configure-vs-cmake feature-detection comparison.
Besides differently detected dependencies and some usual suspects, they were
identical, except `#define CURL_CA_PATH "/etc/ssl/certs"` which was picked up
by CMake, but missed by autotools:
https://github.com/curl/curl/actions/runs/6431509560/job/17464559513?pr=12044

macOS configure-vs-cmake patch:
```diff
diff --git a/.github/workflows/configure-vs-cmake.yml b/.github/workflows/configure-vs-cmake.yml
index 3131bc128..0f20801a3 100644
--- a/.github/workflows/configure-vs-cmake.yml
+++ b/.github/workflows/configure-vs-cmake.yml
@@ -27,7 +27,7 @@ on:
 permissions: {}
 
 jobs:
-  check:
+  check-linux:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
@@ -39,7 +39,28 @@ jobs:
 
     - name: run cmake
       run: |
-         mkdir build && cd build && cmake ..
+         cmake -B build
+
+    - name: compare generated curl_config.h files
+      run: ./scripts/cmp-config.pl lib/curl_config.h build/lib/curl_config.h
+
+  check-macos:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: install packages
+      run: |
+         while [[ $? == 0 ]]; do for i in 1 2 3; do brew update && brew install libtool autoconf automake && break 2 || { echo Error: wait to try again; sleep 10; } done; false Too many retries; done
+
+    - name: run configure --with-openssl
+      run: |
+         autoreconf -fi
+         ./configure --with-openssl
+
+    - name: run cmake
+      run: |
+         cmake -B build
 
     - name: compare generated curl_config.h files
       run: ./scripts/cmp-config.pl lib/curl_config.h build/lib/curl_config.h
```

Issue #12040, opened 2023-10-05
TFTP retries too quickly on mingw-w64 (test 286)


dfandrich:
### I did this

This [CI run](https://ci.appveyor.com/project/curlorg/curl/builds/48201178/job/fk8w48y63orr2rqx) shows test 286 (*TFTP send of boundary case 512 byte file*) failing with a timeout. The logs shows it sending the write request and (presumably) getting the response, then sending the first (and only) data block and waiting for the ACK, but the ACK is missing (possibly due to the server being slow). It then retries and goes through this in a loop 50 times before giving up.

This is what I expect should happen in this case except for one thing: it only waits about 40 µsec between retries. The entire retry attempt (50 retries) takes about 3 msec. Something is clearly wrong with the retry timer here.

What might be relevant here is that this is the very first data block whose ACK is missing. Could it be the right retry time hasn't been set yet and so it retries for 0 seconds?

### I expected the following

curl should way 6 seconds between retries. That should be enough for even an extremely slow server to respond with an ACK, and the test should succeed.

### curl/libcurl version

8.4.0-DEV

### operating system

MSYS_NT-10.0-14393 APPVYR-WIN 3.0.7-338.x86_64 2019-07-11 10:58 UTC x86_64 Msys

bagder:
Created #12047 for researching this

dfandrich:
I tried applying this patch to delay the first data block ACK to see if my speculation about whether that is the trigger was true, but on Linux everything worked as expected and test 286 succeeded. It would be great if someone could try this on msys2 and see if it triggers the bug or not.

```patch
diff --git a/tests/server/tftpd.c b/tests/server/tftpd.c
index 670897c0d..c41aaf3e4 100644
--- a/tests/server/tftpd.c
+++ b/tests/server/tftpd.c
@@ -1271,6 +1271,7 @@ send_ack:
       goto abort;
     }
     write_behind(test, pf->f_convert);
+    sleep(90); /* DEBUGGING #12040 */
     for(;;) {
 #ifdef HAVE_ALARM
       alarm(rexmtval);
```

dfandrich:
I caught [this test failing](https://ci.appveyor.com/project/curlorg/curl/builds/48254111/job/3l9j1l5uac5txweq) (CMake, mingw-w64, gcc 9, Debug x64, Schannel, Static, Unity) with your extra logging added. Here's the start of the log (it continues in the same way for 2 msec until the retries are up):
```
[...]

PR #12047, opened 2023-10-06
tftp: add debug logging for issue 12040

bagder:
see #12040


bagder:
Happens too irregularly.

PR #12053, opened 2023-10-07
.cirrus.yml: freebsd 14

pheiduck:
ensure curl works on latest freebsd version

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -43,9 +43,9 @@ freebsd_task:
   name: FreeBSD
 
   matrix:
-    - name: FreeBSD 13.2
+    - name: FreeBSD 14.0
       freebsd_instance:
-        image_family: freebsd-13-2
+        image_family: freebsd-14-0
 
   env:
     CIRRUS_CLONE_DEPTH: 10
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -54,7 +54,7 @@ freebsd_task:
 
   pkginstall_script:
     - pkg update -f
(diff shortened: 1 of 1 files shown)


dfandrich:
I wouldn't expect 14 to be released for another 4 months or so. We've had
issues with stability of prerelease versions before and had to disable them I
would prefer to wait another few months before adding it. The CI builds are
unstable enough as it is.

pheiduck:
> I wouldn't expect 14 to be released for another 4 months or so. We've had issues with stability of prerelease versions before and had to disable them I would prefer to wait another few months before adding it. The CI builds are unstable enough as it is.

Thanks for your feedback @dfandrich sure definitely we should wait until it is released.

vszakats:
Released today: https://www.freebsd.org/releases/14.0R/relnotes/

pheiduck:
I may need to find a solution for this:
```
pkg: No packages available to install matching 'openssh-portable' have been found in the repositories
```

vszakats:
Have you tried with `openssl-quictls`?:

https://cgit.freebsd.org/ports/tree/security/openssl-quictls

pheiduck:
> Have you tried with `openssl-quictls`?:
> 
> https://cgit.freebsd.org/ports/tree/security/openssl-quictls

nope, will try :)

vszakats:
Oh sorry, the build is green now, but I misread openssh as openssl... Regarding openssh-portable, these are the options according to package search:
https://ports.freebsd.org/cgi/ports.cgi?query=openssh-portable

pheiduck:
> Oh sorry, the build is green now, but I misread openssh as openssl... Regarding openssh-portable, package search knows about it: https://ports.freebsd.org/cgi/ports.cgi?query=openssh-portable
> 
> I assume we use `openssh-portable` for tests. Tests are green too. I haven't checked ssh-based tests specifically, but perhaps openssh come with the base install?

Commit is reverted.

vszakats (approved):

bagder (approved):

PR #12060, opened 2023-10-08
setopt: ignore CURLOPT_PRIVATE on internal handles

jay:
- If CURLOPT_PRIVATE is set on an internal handle then ignore it. Also throw an assertion for debug builds.

It's been documented that the user may set CURLOPT_PRIVATE on their handles to distinguish them from private handles, therefore we cannot allow CURLOPT_PRIVATE to be set on internal handles.

Ref: https://github.com/curl/curl/blob/0dc40b2a/docs/libcurl/opts/CURLOPT_DEBUGFUNCTION.3#L85-L89
Ref: https://github.com/curl/curl/blob/0dc40b2a/docs/libcurl/opts/CURLOPT_SSL_CTX_FUNCTION.3#L73-L79

Closes #xxxx

### DIFF
--- a/lib/setopt.c
+++ b/lib/setopt.c
@@ -2370,7 +2370,15 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
     /*
      * Set private data pointer.
      */
-    data->set.private_data = va_arg(param, void *);
+    if(!data->internal)
+      data->set.private_data = va_arg(param, void *);
+    else {
+      /* Ignore setting private data on internal handles since the user is
+         allowed to distinguish their handles from internal handles by setting
+         private data. */
+      DEBUGASSERT(0);
+      return CURLE_BAD_FUNCTION_ARGUMENT;
+    }
     break;
 
   case CURLOPT_MAXFILESIZE:


bagder:
A user would still be the one that would set the *PRIVATE pointer, so even if we would allow it for an internal handle it would still allow the user to differentiate the handles using that option.

Don't we just complicate things for ourselves by explicitly forbidding this?

jay:
> A user would still be the one that would set the *PRIVATE pointer, so even if we would allow it for an internal handle it would still allow the user to differentiate the handles using that option.

To make sure we are on the same page consider this:

~~~c
int debug_callback(CURL *handle, curl_infotype type, char *data, size_t size, void *clientp)
{
  void *private = NULL;
  curl_easy_getinfo(handle, CURLINFO_PRIVATE, &private);
  if(!private) {
    /* handle is internal since user set CURLOPT_PRIVATE on their handles */
    /* what would be the advantage of user setting CURLOPT_PRIVATE here? */
  }
}
~~~

I already added debugasserts that asserts private data is not set on an internal handle before the debug callback is called so I was looking for some way to improve on that. The purpose of the debugassert is to notify us internally that we set CURLOPT_PRIVATE on an internal handle, since I can't think of a way to reliably test for that. So what you are saying is a fair point I just can't think of a better way yet....

bagder:
> The purpose of the debugassert is to notify us internally that we set CURLOPT_PRIVATE on an internal handle

internally we never set `CURLOPT_PRIVATE` on any handle since the function is provided for the API user, not internals. We should never set it ourselves for any handle.

We allow `curl_easy_setopt()` from a callback, and we have always allowed `CURLOPT_PRIVATE` for all handles, so why not also allow it for "internal handles" if the user thinks of a use for this? They could for example set that pointer as a way to differentiate this handle from another. Perhaps using the http2 push callback when a new easy handle is created for the new transfer.

Preventing `CURLOPT_PRIVATE` now actually is a minor ABI breakage.

jay:
You raise some good points. I'm closing this in favor of a different take uses checksrc to bans assignment to private_data in the source. See #12081

Also note http2 push handles that are created internally are not marked as internal because though they're created internally the ownership is passed to the user.

https://github.com/curl/curl/blob/09e25b9d94f4106eac8ca3a43b221bdc66f405e4/lib/urldata.h#L2017-L2019

PR #12062, opened 2023-10-08
http: avoid Expect: 100-continue if Upgrade: is used

### LINKED ISSUE #12022: Curl doesn't send request body when the server responds with HTTP 101
### I did this

Using a local Liberty server and curl 8.3.0 client, I sent a HTTP POST request over plain HTTP1 as follows:
```
curl --http2 -v -H"Expect:100-continue" http://localhost:9080/system/greeter/hello --data "foo"
```
The server sent the following 2 response headers in quick succession:
```
HTTP/1.1 100 Continue
Content-Length: 0
Date: Tue, 03 Oct 2023 19:06:00 GMT

HTTP/1.1 101 Switching Protocols
Date: Tue, 03 Oct 2023 19:06:00 GMT
Upgrade: h2c
Connection: Upgrade
Content-Length: 0
```
In response to that, Curl sent the HTTP2 connection preface without sending the data.
It doesn't fail every time; sometimes when the HTTP 101 header is late enough, curl manages to send the request body correctly.

### I expected the following

I expected curl to send the request data before sending the HTTP2 preface

### curl/libcurl version

curl 8.3.0 (x86_64-w64-mingw32) libcurl/8.3.0 OpenSSL/3.1.2 (Schannel) zlib/1.3 brotli/1.1.0 zstd/1.5.5 WinIDN libssh2/1.11.0 nghttp2/1.56.0 ngtcp2/0.19.1 nghttp3/0.15.0

### operating system

Windows 11

bagder:
Reported-by: Daniel Jelinski
Fixes #12022

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -1678,8 +1678,6 @@ static CURLcode expect100(struct Curl_easy *data,
                           struct dynbuf *req)
 {
   CURLcode result = CURLE_OK;
-  data->state.expect100header = FALSE; /* default to false unless it is set
-                                          to TRUE below */
   if(!data->state.disableexpect && Curl_use_http_1_1plus(data, conn) &&
      (conn->httpversion < 20)) {
     /* if not doing HTTP 1.0 or version 2, or disabled explicitly, we add an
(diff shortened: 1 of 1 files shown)


icing:
I don't think this fixes the issue. The problem is not really the 100 response, but that the server will expect the complete upload before replying with a 101. Do we do that?

The alternative is to disable the `Ugrade:` when having a request body.

bagder:
> I don't think this fixes the issue. The problem is not really the 100 response, but that the server will expect the complete upload before replying with a 101.

Right, but without (waiting for) a 100 curl will do the upload normally "at once", to which the server then presumably returns 101 *after* the request is sent. Right?

>Maybe we do that, but do we have a server to test that with?

Getting the 101 after the request seems like the standard scenario though. I would imagine that possibly there is a risk if the server returns the 101 before the entire request body has been sent but that's nothing that this PR changes.

Issue #2634, opened 2018-06-05
BoringSSL with Schannel


gvanem:
Compiling libcurl with the combination `-DUSE_SCHANNEL`, `-DHAVE_BORINGSSL` and 
`-DUSE_OPENSSL`, breaks in several places because of (from clang-cl):
```
In file included from vtls/openssl.c:52:
In file included from f:/MingW32/src/inet/Crypto/BoringSSL/include\openssl/ssl.h:145:
f:/MingW32/src/inet/Crypto/BoringSSL/include\openssl/base.h(275,29):  error: expected ')'
typedef struct X509_name_st X509_NAME;
                            ^
f:\ProgramFiler-x86\WindowsKits\Include\10.0.16299.0\um\wincrypt.h(2897,55):  note: expanded from macro 'X509_NAME'
#define X509_NAME                           ((LPCSTR) 7)
                                                      ^
f:/MingW32/src/inet/Crypto/BoringSSL/include\openssl/base.h(275,29):  note: to match this '('
f:\ProgramFiler-x86\WindowsKits\Include\10.0.16299.0\um\wincrypt.h(2897,45):  note: expanded from macro 'X509_NAME'
#define X509_NAME                           ((LPCSTR) 7)
                                            ^
```
A fix is already in [ldap.c](https://github.com/curl/curl/blob/master/lib/ldap.c#L60):
```c
#ifdef HAVE_BORINGSSL
# undef X509_NAME
# undef X509_CERT_PAIR
# undef X509_EXTENSIONS
#endif
```
which IMHO should be moved to another place. `urldata.h`?

bagder:
Yeah, clearly something like that should be done. Did you try it? I'm thinking maybe it will cause problems with `vtls/openssl.c` ?

gvanem:
I did the `#undef` to `vtls/schannel.h` instead. Since that file usually is responsible for pulling in `<wincrypt.h>`. In `ldap.c`, it's pulled in via `<winldap.h>`. And `ldap.c` never includes any BoringSSL headers AFAICS. So I came up with this patch:
```diff
--- a/lib/ldap.c 2018-05-09 19:58:09
+++ b/ldap.c     2018-06-07 11:08:05
@@ -54,15 +54,6 @@
 # endif /* HAVE_LDAP_SSL && HAVE_LDAP_SSL_H */
 #endif

-/* These are macros in both <wincrypt.h> (in above <winldap.h>) and typedefs
[...]

PR #1610, opened 2017-06-24
Fix boringssl inclusion in winbuild build system

BurningEnlightenment:
Compile with `WIN32_LEAN_AND_MEAN` which prevents `windows.h` from including too much clutter including `wincrypt.h` which in turn contains some preprocessor macros that clash with boringssl symbols.

Detect boringssl by checking the existance of `is_boringssl.h` and set the corresponding `HAVE_BORINGSSL` for compilation which is used in `ldap.c` to undefine the evil macros.

Fixes #1562

PR #12065, opened 2023-10-08
build: do not publish `HAVE_BORINGSSL`, `HAVE_AWSLC` macros

vszakats:
Syncing this up with CMake.

Source code uses the built-in `OPENSSL_IS_AWSLC` and
`OPENSSL_IS_BORINSSL` macros to detect BoringSSL and AWS-LC. No help is
necessary from the build tools.

The one use of `HAVE_BORINGSSL` in the source turned out to be no longer
necessary for warning-free BoringSSL + Schannel builds. Ref: #1610 #2634

autotools detects this anyway for display purposes.
CMake detects this to decide whether to use the BoringSSL-specific
crypto lib with ngtcp2. It detects AWS-LC, but doesn't use the detection
result just yet (planned in #12066).

Ref: #11964

Closes #12065


bagder (approved):

jay (approved):

vszakats:
When the single use of `HAVE_BORINGSSL` was introduced in cd34ffa614b04ed6ab6533fa3b8661f0ef9abfa9, it was used in `lib/ldap.c`, later moved to `schannel.h` in 274940d7438af7ef92ce3e11d75620db94675932. It was used to avoid `wincrypt.h` redefinitions with BoringSSL + Schannel MultiSSL builds. One side effect of using `HAVE_BORINGSSL` (compared to `OPENSSL_IS_BORINGSSL) is that it's present before including `openssl.h`.

Making tests now to see if this trick is still necessary.

Issue 1: BoringSSL no longer compiles with curl_ngtcp2.c, after introducing `SSL_CTX_set_ciphersuites` into it unconditionally in aa9a6a177017e4b74d33cdf85a3594900f4a7f81.

bagder:
> BoringSSL no longer compiles with curl_ngtcp2

I tried to copy the logic from openssl.c when I did that, but then clearly I failed...

jay:
https://github.com/curl/curl/blob/6fa1d817e5b1a00d7d0c8168091877476b499317/lib/vtls/openssl.c#L203-L217
and then `#ifdef HAVE_SSL_CTX_SET_CIPHERSUITES`

jay:
though for ngtcp2 maybe we don't need as much legacy check

~~~diff
diff --git a/lib/vquic/curl_ngtcp2.c b/lib/vquic/curl_ngtcp2.c
index 27711ef..ccd933f 100644
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -433,11 +433,13 @@ static CURLcode quic_ssl_ctx(SSL_CTX **pssl_ctx,
   {
     const char *ciphers13 = conn->ssl_config.cipher_list13 ?
       conn->ssl_config.cipher_list13 : QUIC_CIPHERS;
+#ifndef OPENSSL_IS_BORINGSSL
     if(SSL_CTX_set_ciphersuites(ssl_ctx, ciphers13) != 1) {
       failf(data, "failed setting QUIC cipher suite: %s", ciphers13);
       return CURLE_SSL_CIPHER;
     }
     infof(data, "QUIC cipher selection: %s", ciphers13);
+#endif
   }
 
   /* Open the file if a TLS or QUIC backend has not done this before. */
~~~

vszakats:
@jay: Agreed, H3 assumes LibreSSL 3.7 or quictls 3.0 (or AWC-LC, not sure when it added this function but 0.15.0 has it).

vszakats:
The `HAVE_BORINGSSL` trick is no longer needed, confirmed with a warning-free BoringSSL + Schannel build:
```
curl 8.4.0-DEV (x86_64-w64-mingw32) libcurl/8.4.0-DEV BoringSSL/b98ce18c (Schannel) zlib/1.2.13 brotli/1.1.0 zstd/1.5.4 WinIDN nghttp2/1.52.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe Unicode UnixSockets zstd
```

PR #12070, opened 2023-10-09
cmake: replace `check_library_exists_concat()`

### LINKED ISSUE #11285: CMake FetchContent with ZLIB fails (>=8.0.0)

vszakats:
The idea of `check_library_exists_concat()` is that it detects an
optional component and adds it to the list of libs that we also use in
subsequent component checks. This caused problems when detecting
components with unnecessary dependencies that were not yet built.

CMake offers the `CMAKE_REQUIRED_LIBRARIES` variable to set libs used
for component checks, which we already use in most cases. That left 4
uses of `check_library_exists_concat()`. Only one of these actually
needed the 'concat' feature (ldap/lber).

Delete this function and replace it with standard
`check_library_exists()` and manual management of our `CURL_LIBS`
list we use when linking build targets. And special logic to handle the
ldap/lber case.

(We have a similar function for headers: `check_include_file_concat()`.
It works, but problematic for performance reasons and because it hides
the actual headers required in `check_symbol_exists()` calls.)

Ref: #11537 #11558
Fixes #11285
Fixes #11648
Closes #12070

### DIFF
--- a/CMake/Macros.cmake
+++ b/CMake/Macros.cmake
@@ -23,19 +23,6 @@
 ###########################################################################
 #File defines convenience macros for available feature testing
 
-# This macro checks if the symbol exists in the library and if it
-# does, it prepends library to the list.  It is intended to be called
-# multiple times with a sequence of possibly dependent libraries in
-# order of least-to-most-dependent.  Some libraries depend on others
-# to link correctly.
-macro(check_library_exists_concat LIBRARY SYMBOL VARIABLE)
-  check_library_exists("${LIBRARY};${CURL_LIBS}" ${SYMBOL} "${CMAKE_LIBRARY_PATH}"
-    ${VARIABLE})
-  if(${VARIABLE})
-    set(CURL_LIBS ${LIBRARY} ${CURL_LIBS})
-  endif()
-endmacro()
-
 # Check if header file exists and add it to the list.
 # This macro is intended to be called multiple times with a sequence of
 # possibly dependent header files.  Some headers depend on others to be
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -378,7 +378,10 @@ if(ENABLE_THREADED_RESOLVER)
 endif()
 
 # Check for all needed libraries
-check_library_exists_concat("socket" connect      HAVE_LIBSOCKET)
+check_library_exists("socket" "connect" "" HAVE_LIBSOCKET)
+if(HAVE_LIBSOCKET)
+  set(CURL_LIBS "socket;${CURL_LIBS}")
+endif()
 
 check_function_exists(gethostname HAVE_GETHOSTNAME)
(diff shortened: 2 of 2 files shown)


vszakats:
@nmoinvaz: Does this fix the issues for you?

nmoinvaz:
I am on an older version of the `CMakeLists.txt` but I have applied your changes and they work for me. Thanks for your help.

PR #12075, opened 2023-10-10
CI, pytest slow network tests, remove

icing:
- remove these tests as they are currently not reliable in our CI setups.


jay:
By not reliable do you mean curl can't handle the simulated congestion due to bugs or the CI can't? Isn't it better to have this additional testing to uncover cases where it fails?

icing:
> By not reliable do you mean curl can't handle the simulated congestion due to bugs or the CI can't? Isn't it better to have this additional testing to uncover cases where it fails?

Sorry to not have been precise. curl handles the test cases, but CI sometimes fails on these due to additional conditions. I would like to add an additional CI job in the future, specific to these, instead of mixing this in.

Issue #12064, opened 2023-10-08
http3 ngtcp2: QUIC connection left intact after Operation timed out 


calvin2021y:
### I did this

build libcurl with master branch and  openssl-3.0.10+quic, lastest ngtcp2 and nghttp3 release. 

set CURLOPT_MAXAGE_CONN more than 110, with multi interface.

with step:

1) send request to https://1.1.1.1 with http3 protocol,  get 200 response.

2) wait 110 secends

3) send request to https://1.1.1.1 with http3 protocol again with same multi handle , get timeout error

4) wait 1~ 10 secends

5)  send request to https://1.1.1.1 with http3 protocol again with same multi handle , get timeout error

repeat 4, 5 will keep get timeout error.

there is the trace logs:

```sh
*   Trying 1.1.1.1:443...
* QUIC cipher selection: TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_CCM_SHA256
*  subjectAltName: host "1.1.1.1" matched cert's IP address!
* Verified certificate just fine
* Connected to 1.1.1.1 (1.1.1.1) port 443
* using HTTP/3
* [HTTP/3] [0] OPENED stream for https://1.1.1.1/
* [HTTP/3] [0] [:method: GET]
* [HTTP/3] [0] [:scheme: https]
* [HTTP/3] [0] [:authority: 1.1.1.1]
* [HTTP/3] [0] [:path: /]
* [HTTP/3] [0] [accept: */*]
* [HTTP/3] [0] [user-agent: curl/8.3.0]
> GET / HTTP/3
Host: 1.1.1.1
Accept: */*
User-Agent: curl/8.3.0

< HTTP/3 200 
< date: Sun, 08 Oct 2023 15:15:08 GMT
< content-type: text/html
< report-to: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=dJkPRYeedFYxg%2B%2FLhPQwNqz8QX2WcEnECCVIwKTGahkSIF0uR%2FimwiBuCzRrKQEwrQge2f8rz%2F5nfsdc9rb2JLlKx6%2BWDHosS8FsksBclFdstpgMnKCvseU%3D"}],"group":"cf-nel","max_age":604800}
< nel: {"report_to":"cf-nel","max_age":604800}
< last-modified: Fri, 21 Jul 2023 21:11:33 GMT
[...]

PR #12077, opened 2023-10-10
QUIC, manage connection idle timeouts

icing:
- configure a 120s idle timeout on our side of the connection
- track the timestamp when actual socket IO happens
- check IO timestamp to our *and* the peer's idle timeouts in "is this connection alive" checks
- refs #12064

### DIFF
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -1630,7 +1630,7 @@ static bool cf_quiche_conn_is_alive(struct Curl_cfilter *cf,
   if(!ctx->qconn)
     return FALSE;
 
-  /* Both sides of the QUIC connection announce they max idle time in
+  /* Both sides of the QUIC connection announce they max idle times in
    * the transport parameters. Look at the minimum of both and if
    * we exceed this, regard the connection as dead. The other side
    * may have completely purged it and will no longer respond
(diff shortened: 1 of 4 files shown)


lib/vquic/curl_quiche.c
```diff
@@ -1617,9 +1623,31 @@ static bool cf_quiche_conn_is_alive(struct Curl_cfilter *cf,
...
                                     bool *input_pending)
 {
+  struct cf_quiche_ctx *ctx = cf->ctx;
   bool alive = TRUE;
 
   *input_pending = FALSE;
+  if(!ctx->qconn)
+    return FALSE;
+
+  /* Both sides of the QUIC connection announce they max idle time in
```
bagder, line 1633: *their* max idle *times*

bagder:
> configure a 120s idle timeout on our side of the connection

Should this be aligned with `CURLOPT_MAXAGE_CONN`  ? If curl won't use connections older than N seconds, maybe it makes sense to set the idle timeout to N +1 or something?

icing:
> > configure a 120s idle timeout on our side of the connection
> 
> Should this be aligned with `CURLOPT_MAXAGE_CONN` ? If curl won't use connections older than N seconds, maybe it makes sense to set the idle timeout to N +1 or something?

From the transfer that opened the connection or which one?

bagder:
> From the transfer that opened the connection or which one?

Hm. That's a good question!

Let's stick it to 120 for now and we can get back to that if we figure out a better way...

PR #12081, opened 2023-10-10
checksrc: ban internal assignment of user's private_data

jay:
- Add a rule 'PRIVATEDATA' in checksrc to disallow use of private_data member assignment.

- Remove assertion that private_data is not set on an internal handle, since the user may set it on the handle via CURLOPT_PRIVATE.

This is my second take on a proactive ban to ensure that we do not set private_data on internal handles for our own use, since those handles may at some point be exposed to the user (eg debug callback).

We already tell the user that "If you need to distinguish your curl handle from internal handles then set CURLOPT_PRIVATE on your handle."

Therefore, private_data must not be assigned to for internal handles by libcurl but may later be assigned to by the user (via CURLOPT_PRIVATE).

Ref: https://github.com/curl/curl/pull/12060
Ref: https://github.com/curl/curl/blob/0dc40b2a/docs/libcurl/opts/CURLOPT_DEBUGFUNCTION.3#L85-L89
Ref: https://github.com/curl/curl/blob/0dc40b2a/docs/libcurl/opts/CURLOPT_SSL_CTX_FUNCTION.3#L73-L79

Closes #xxxx

### DIFF
--- a/lib/curl_trc.c
+++ b/lib/curl_trc.c
@@ -61,10 +61,6 @@ void Curl_debug(struct Curl_easy *data, curl_infotype type,
       "* ", "< ", "> ", "{ ", "} ", "{ ", "} " };
     if(data->set.fdebug) {
       bool inCallback = Curl_is_in_callback(data);
-      /* CURLOPT_DEBUGFUNCTION doc says the user may set CURLOPT_PRIVATE to
-         distinguish their handle from internal handles. */
-      if(data->internal)
-        DEBUGASSERT(!data->set.private_data);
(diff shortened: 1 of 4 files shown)


bagder:
I'm against this.

No internal code at all should ever set any value in `data->set.*`, that is just wrong (but we have done several such mistakes in the past). That struct is reserved for user-set values set with `curl_easy_setopt()`.  Further, I think using checksrc for this protection is too clumsy for this kind of check.

jay:
Ok, closing in favor of my third take #12104

PR #10284, opened 2023-01-12
connect: fix access of pointer before NULL check

bagder:
Detected by Coverity CID 1518992

PR #12082, opened 2023-10-11
lib: fix gcc warning

vszakats:
Do not pass NULL to printf %s.

Seen with gcc 13.2.0 on Debian:
```
.../curl/lib/connect.c:696:27: warning: '%s' directive argument is null [-Wformat-overflow=]
```
Ref: https://github.com/curl/curl-for-win/actions/runs/6476161689/job/17584426483#step:3:11104

Ref: #10284
Co-authored-by: Jay Satiro
Closes #12082

---

```
In function 'is_connected',
    inlined from 'cf_he_connect' at /home/runner/work/curl-for-win/curl-for-win/curl/lib/connect.c:906:16:
/home/runner/work/curl-for-win/curl-for-win/curl/lib/connect.c:696:27: warning: '%s' directive argument is null [-Wformat-overflow=]
  696 |     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/curl-for-win/curl-for-win/curl/lib/curl_trc.h:123:38: note: in definition of macro 'CURL_TRC_CF'
  123 |          Curl_trc_cf_infof(data, cf, __VA_ARGS__); } while(0)
      |                                      ^~~~~~~~~~~
In file included from /home/runner/work/curl-for-win/curl-for-win/curl/bld/lib/CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c:49:
/home/runner/work/curl-for-win/curl-for-win/curl/lib/connect.c: In function 'cf_he_connect':
/home/runner/work/curl-for-win/curl-for-win/curl/lib/connect.c:696:28: note: format string is defined here
  696 |     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
      |                            ^~
make[2]: Leaving directory '/home/runner/work/curl-for-win/curl-for-win/curl/bld'
```

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -693,11 +693,11 @@ evaluate:
   result = CURLE_COULDNT_CONNECT;
   for(i = 0; i < sizeof(ctx->baller)/sizeof(ctx->baller[0]); i++) {
     struct eyeballer *baller = ctx->baller[i];
+    if(!baller)
+      continue;
     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
-                baller?baller->name:NULL,
-                baller?baller->has_started:0,
-                baller?baller->result:0);
-    if(baller && baller->has_started && baller->result) {
+                baller->name, baller->has_started, baller->result);
+    if(baller->has_started && baller->result) {
       result = baller->result;
       break;
     }


lib/connect.c
```diff
@@ -694,7 +694,7 @@ static CURLcode is_connected(struct Curl_cfilter *cf,
   for(i = 0; i < sizeof(ctx->baller)/sizeof(ctx->baller[0]); i++) {
     struct eyeballer *baller = ctx->baller[i];
     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
-                baller?baller->name:NULL,
+                baller?baller->name:"?",
```
jay, line 697: I would abstain from including this dozen line warning in the commit message. It's simply that printf-like functions %s format specifier have undefined behavior for NULL. ([technically](https://stackoverflow.com/questions/11589342/what-is-the-behavior-of-printing-null-with-printfs-s-specifier))

All the other loops do an if(!baller) continue; so I have an alternate proposal

~~~diff
diff --git a/lib/connect.c b/lib/connect.c
index c7ba3e2..92eebf0 100644
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -693,11 +693,11 @@ evaluate:
   result = CURLE_COULDNT_CONNECT;
   for(i = 0; i < sizeof(ctx->baller)/sizeof(ctx->baller[0]); i++) {
     struct eyeballer *baller = ctx->baller[i];
+    if(!baller)
+      continue;
     CURL_TRC_CF(data, cf, "%s assess started=%d, result=%d",
-                baller?baller->name:NULL,
-                baller?baller->has_started:0,
-                baller?baller->result:0);
-    if(baller && baller->has_started && baller->result) {
+                baller->name, baller->has_started, baller->result);
+    if(baller->has_started && baller->result) {
       result = baller->result;
       break;
     }
~~~

Ref: #10284

vszakats: You're right. I'll amend the patch too.

I'm seeing many warnings and issues (not necessarily code or even curl-specific) that I stuck to the habit of keeping a copy and a log reference to keep track of them and look them up later if necessary. Surely for this case it will be enough to keep it in the PR, not in the commit.

jay (approved):
Since the warning text is not in this one I would add a brief explanation of it or the action taken, example

~~~
lib: fix gcc warning

- Do not pass NULL to printf %s.

...
~~~

The problem with the warnings in some cases like this one is they are confusing and take extra cognitive processing so that's why I was suggesting a summary of sorts so it's easy to see what the warning is actually for or what is being fixed. I agree it is good to keep the actual warning in the PR so we have a record of what it actually said. The CI links are all ephemeral so in the commit message they are only good for so long.

PR #12085, opened 2023-10-11
build: add `src/.checksrc` to source tarball

### LINKED ISSUE #12084: Curl 8.4.0 fails to build because of a checksrc error. File src/.checksrc is missing from release tarball
### I did this

Hi,

See my comments in #11958, it seems that the release tarball for curl 8.4.0 misses the file `src/.checksrc`. When we try to build it from sources, we get this when checksrc is ran:

```
09:20:40  ./tool_stderr.c:62:1: warning: STDERR already disabled from line 35 (BADCOMMAND)
09:20:40     /* !checksrc! disable STDERR 1 */
09:20:40   ^
09:20:40  ./tool_stderr.c:69:1: warning: STDERR already disabled from line 35 (BADCOMMAND)
09:20:40     /* !checksrc! disable STDERR 1 */
09:20:40   ^
09:20:40  ./tool_stderr.c:35:18: warning: Unused ignore: STDERR (UNUSEDIGNORE)
09:20:40     /* !checksrc! disable STDERR 1 */
09:20:40                    ^
09:20:40  checksrc: 0 errors and 3 warnings
09:20:40  checksrc: 0 errors and 10 warnings suppressed
```

We have downloaded this tarball https://curl.se/download/curl-8.4.0.tar.gz where apparently the newly added file `src/.checksrc` is missing. If we apply this patch to add it back, it works:

```
This file is committed in git, but missing from the curl 8.4.0 tarball, and makes the build fail if missing.

--- /dev/null
+++ src/.checksrc
@@ -0,0 +1 @@
+enable STDERR
```

So... shall there be a new 8.4.1 release with the file included in the tarball ?

Cheers,
Romain

### I expected the following

Build works from the official tarball.

### curl/libcurl version

curl 8.4.0

### operating system

Ubuntu 22.04.

vszakats:
Regression from e5bb88b8f824ed87620bd923552534c83c2a516e #11958

Bug: https://github.com/curl/curl/pull/11958#issuecomment-1757079071
Reported-by: Romain Geissler
Fixes #12084
Closes #12085

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -89,7 +89,7 @@ CLEANFILES = tool_hugehelp.c
 NROFF=env LC_ALL=C @NROFF@ @MANOPT@ 2>/dev/null # figured out by the configure script
 
 EXTRA_DIST = mkhelp.pl \
- Makefile.mk curl.rc Makefile.inc CMakeLists.txt
+ Makefile.mk curl.rc Makefile.inc CMakeLists.txt .checksrc
 
 # Use absolute directory to disable VPATH
 MANPAGE=$(abs_top_builddir)/docs/curl.1


Romain-Geissler-1A:
Ah thanks ! I also reported it as #12084.

MarcelRaad:
Ah, that's also visible in the autobuilds. We need to look at them more :-(

MarcelRaad (approved):

vszakats:
@MarcelRaad: Can we make that test fail if it detects a distcheck issue?

MarcelRaad:
I'm not familiar with the packaging process, unfortunately. Probably @bagder can answer that? We have that issue quite regularly, so that would be great.

vszakats:
Unless I'm missing something, we might miss a `make distcheck`
in the distcheck GHA. It needs to be an out-of-tree build to catch
this (and cover dot files, which it probably does). This needs to
be tested in practice. Existing tests try to build from a freshly made
source tarball, but one of them is CMake, which doesn't run checksrc
on its own to catch this particular case, the other is autotools, which
is likely not a debug build to break on this either.

I'm testing this in #12088.

PR #12088, opened 2023-10-11
CI: add autotools, out-of-tree, debug build to distro check job

vszakats:
Add a job that builds curl from a generated source tarball sample, with
autotools, out-of-tree, in debug mode.

Ref: #12085
Closes #12088


vszakats:
```
make[3]: *** No rule to make target 'Makefile.dist', needed by 'distdir-am'.  Stop.
```
https://github.com/curl/curl/actions/runs/6480792976/job/17597019192#step:7:3863

vszakats:
Stuck at:
```
make[3]: *** No rule to make target 'Makefile.dist', needed by 'distdir-am'.  Stop.
```
https://github.com/curl/curl/actions/runs/6480934827/job/17597469241#step:7:2658

vszakats:
This was "easy": `Makefile.dist` is referenced from `Makefile.am` in `EXTRA_DIST`, while apparently it shouldn't be there. First because `distcheck` break on this, second because this file isn't part of the of the source tarball (not with this filename anyway).

Next problem: `ERROR: files left in build directory after distclean:` followed by a list of files created during the build process.

Next, next problem: `distcheck` actually doesn't complain about a missing `src/.checksrc` at all. Perhaps this isn't even what it's meant to do, though I swear to have been warned about missing files in e.g. libssh2, when running distcheck.

vszakats:
Dropped the `make distcheck` idea, though in theory it's probably the way to fix this class of problem.
Added a debug out-of-tree configure job instead.

Example of it catching the missing `src/.checksrc`:
https://github.com/curl/curl/actions/runs/6483454571/job/17605256636#step:3:4907

vszakats:
Now it's this:
/usr/bin/install: cannot create regular file '/usr/local/lib/libcurl.so.4.8.0': Permission denied

I'm giving up!

vszakats:
Okay, let's try without `install` first...

bagder:
> because distcheck breaks on this, second because this file isn't part of the of the source tarball (not with this filename anyway).

`make dist` makes all `*.dist` files get included in the tarball without the `.dest` extension. So the `Makefile.dist` file ends up as just `Makefile` in the tarball.

vszakats:
Phew, green now:
https://github.com/curl/curl/actions/runs/6483854257/job/17606626203?pr=12088

This is already rebased on master with the `src/.checksrc` fix.

vszakats:
@bagder: Do you think we should delete `Makefile.dist` off the `EXTRA_DIST` list, or should is stay there? (this patch deletes it at the moment)

dfandrich:
> /usr/bin/install: cannot create regular file '/usr/local/lib/libcurl.so.4.8.0':
> Permission denied

You can add --prefix=$PWD/install to install it somewhere safe.

vszakats:
That did the trick, thanks Dan!

bagder:
> @bagder: Do you think we should delete `Makefile.dist` off the `EXTRA_DIST` list, or should is stay there? (this patch deletes it at the moment)

I believe the reasoning for including `Makefile.dist` in the tarball is that it lets a user make a subsequent release with `maketgz` based on only the files provided in the tarball, and I believe that is a sensible policy. This, since the root `Makefile` might get overwritten if for example configure is invoked from within the source directory.

So yes: I think it can remain in EXTRA_DIST.

vszakats:
@bagder: Thanks, I'll revert this, but `Makefile.dist` is not present in the tarball, which might need another fix.

bagder:
> @bagder: Thanks, I'll revert this, but `Makefile.dist` is not present in the release tarball (I'm guessing due to the error showcased above), which might need another fix.

Ah yes, and that's a slightly tricky one since we have that hook that puts `.dist` files into the tarball without that extension:

https://github.com/curl/curl/blob/a20d7bd974d04065ee1d7e3de21a87e449d1c246/Makefile.am#L171-L178

PR #12090, opened 2023-10-11
easy_lock: add a pthread_mutex_t fallback

hwti:
This allows to keep the init threadsafe with gcc < 4.9.0 (no C11 atomics).


bagder (approved):
Oh, this looks like a pretty clever approach when pthreads is already in use!

jay:
Thanks

PR #12105, opened 2023-10-13
trace: replace variadic macros with C89 syntax

### LINKED ISSUE #11891: Compilation error - function/macro conflict in lib/curl_trc
### I did this

./configure --with-openssl --disable-verbose --enable-warnings
make

### I expected the following

Expected success like in previous releases. Now there's an error:
```
In file included from curl_trc.c:25:
curl_setup.h:715:25: error: expected identifier or '(' before 'do'
  715 | #  define Curl_nop_stmt do { } while(0)
      |                         ^~
curl_trc.h:139:33: note: in expansion of macro 'Curl_nop_stmt'
  139 | #define Curl_trc_cf_infof(...)  Curl_nop_stmt
      |                                 ^~~~~~~~~~~~~
curl_trc.c:242:6: note: in expansion of macro 'Curl_trc_cf_infof'
  242 | void Curl_trc_cf_infof(struct Curl_easy *data, struct Curl_cfilter *cf,
      |      ^~~~~~~~~~~~~~~~~
curl_setup.h:715:32: error: expected identifier or '(' before 'while'
  715 | #  define Curl_nop_stmt do { } while(0)
      |                                ^~~~~
curl_trc.h:139:33: note: in expansion of macro 'Curl_nop_stmt'
  139 | #define Curl_trc_cf_infof(...)  Curl_nop_stmt
      |                                 ^~~~~~~~~~~~~
curl_trc.c:242:6: note: in expansion of macro 'Curl_trc_cf_infof'
  242 | void Curl_trc_cf_infof(struct Curl_easy *data, struct Curl_cfilter *cf,
      |      ^~~~~~~~~~~~~~~~~
make[2]: *** [Makefile:2388: libcurl_la-curl_trc.lo] Błąd 1
```
Looks like the definition of Curl_trc_cf_infof in curl_trc.c:242 is in conflict with macros in curl_trc.h:139/142 - maybe it's just redundant.

### curl/libcurl version

curl 8.3.0 (cannot paste output of curl -V because it failed to compile)

### operating system

Linux 6.4.13-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Aug 30 17:07:31 UTC 2023 x86_64 GNU/Linux

### LINKED ISSUE #12083: 8.4.0 fails to build with CURL_DISABLE_VERBOSE_STRINGS=ON
### I did this

The cmake command line I used to generate Visual Studio solution file:
```
cmake .. -G "Visual Studio 16 2019" -DCURL_TARGET_WINDOWS_VERSION=0x0600 -DBUILD_CURL_EXE=OFF -DBUILD_SHARED_LIBS=OFF -DCURL_ZLIB=OFF -DCURL_DISABLE_ALTSVC=ON -DCURL_DISABLE_COOKIES=ON -DCURL_DISABLE_DOH=ON -DCURL_DISABLE_GETOPTIONS=ON -DCURL_DISABLE_HSTS=ON -DCURL_DISABLE_LIBCURL_OPTION=ON -DCURL_DISABLE_MIME=ON -DCURL_DISABLE_NETRC=ON -DCURL_DISABLE_NTLM=ON -DCURL_DISABLE_PARSEDATE=ON -DCURL_DISABLE_PROGRESS_METER=ON -DCURL_DISABLE_SHUFFLE_DNS=ON -DCURL_DISABLE_SOCKETPAIR=ON -DCURL_DISABLE_VERBOSE_STRINGS=ON -DHTTP_ONLY=ON -DCURL_ENABLE_SSL=OFF -DUSE_LIBIDN2=OFF -DCURL_USE_LIBPSL=OFF -DCURL_USE_LIBSSH2=OFF -DENABLE_UNIX_SOCKETS=OFF
```

It generated a solution file without issues, but then it failed to compile with this error:
```
curl_trc.c(247,1): error C2059: syntax error: 'do'`
```

Removing `Curl_trc_cf_infof` definition in `curl_trc.c` fixed the build for me.

### I expected the following

curl should build without errors when CURL_DISABLE_VERBOSE_STRINGS=ON

### curl/libcurl version

curl 8.4.0

### operating system

Windows

vszakats:
Replace `infof()` and `CURL_TRC_CF()` macros using C99 variadic
macros with a solution using C89 syntax.

Also:
- fix compiler warnings in `CURL_DISABLE_VERBOSE_STRINGS` builds.
  Co-authored-by: Stefan Eissing (via #12164)
- delete redundant NULL check in `Curl_trc_cf_infof()`.
- fix `#if` comments in `lib/curl_trc.*`.
- fix indentation.
 
Ref: https://github.com/curl/curl/pull/11922#discussion_r1336507080
Fixes #11880
Fixes #11891
Fixes #12083
Closes #12105

---

- [x] add C89-compliant alternative macros
- [x] upgrade single line calls with new macros
- [x] finish converting multiline calls (went with the libssh2 style)
- [x] reflow lines breaking checksrc line limit
- [x] delete variadic macros (pending PRs will need updates!)
- [x] delete variadic macro feature detections (autotools, cmake)
- [x] rename back `infog` → `infof`, `CURL_TRC_CG` → `CURL_TRC_CF`

→ in another PR: ~~possibly improve runtime perf by pointing macro to short inline function that checks verbose option and calls into the actual logging as needed. Needs 'inline' feature detection (less problematic than variadic, and also optional).~~ Vararg functions aren't inlined. Another solution is passing `data` (and `cf`) twice, then (like before) do the verbosity check in a short macro and call the logging function from there.


jay:
Sorry I did not realize what I was asking for. I thought your change was just a few lines for CURL_TRC_CF. I didn't know we'd have to write infof(( and debugf(( like that. I think you should wait for more feedback before continuing.

What about if we replace CURL_TRC_CF with a tracef( like we can already do infof( and debugf( ? Would that work or am I misunderstanding?

vszakats:
`infof()` also uses variadic macro at the moment. I'm not aware of another solution to shortcut this in C89.

It isn't all very complicated. The manual work for multiline calls is tedious but mechanical. Even that should be easy with a more capable regex search and replace than my editor has (or using some awk magic).

Either way, let's discuss. I'm holding it off.

lib/curl_trc.h
```diff
@@ -145,6 +150,9 @@ void Curl_trc_cf_infof(struct Curl_easy *data, struct Curl_cfilter *cf,
 #error "missing VARIADIC macro define, fix and rebuild!"
 #endif
 
+#define infog(...)  Curl_nop_stmt
```
bagder, line 153: Can't we just make `infof()` require double parentheses? I would prefer sticking to the same name we have used for so long (introduced before year 2000), it is now stuck in my muscle memory.

vszakats: It was just temporary to allow upgrading them gradually (and allow testing with the first batch of changes).
Now that I finished all of them, it can be renamed back to its usual name. I'll do that soon.

icing:
If I read this correctly, it means that the `do { if(enabled())} while(0)` pattern is gone. If that is so, I am against this change as all traces that were basically missed branches before are now function calls.

vszakats:
I believe that can be reintroduced with an inline stub that does this check and delegates the actual logging in the separate call. Inline seems to be easier to detect and enable and also doesn't break compilers (like the current solution does) if some of them doesn't support it or we don't yet detect correctly and enable. (see this item in the PR description)

The other solution is introducing thread local storage for `data` (and cf), and keep the current PP trick. That seems more complicated.

icing:
Looking at the referenced comment, are we doing all this to avoid a warning when passing `NULL` to a `%s` arg?

bagder:
@icing: I believe #11880 is the primary problem this addresses.

vszakats:
@icing: This is unrelated to the NULL %s issue. This one is a fix for the problems with various C89 compilers not supporting variadic macros. The PR description links to the issues and reports I was aware of.

icing:
Not having many cpu cycles right now. But could we not without variadic macros #else define static functions in the header file?

vszakats:
Defining static functions in the header would need another hack to make them work in unity builds (otherwise the duplicate copies would clash on compile). AFAICS vararg functions are neither _implicitly_ inlined by popular compilers [^1][^2]. If this would help perf by keeping function calls local, it _may_ make sense for non-unity builds. I'd address that (or other optimizations) in a separate PR though. This will need measurements and/or someone with intricate knowledge on the topic.

[^1]: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=10980
[^2]: https://en.wikipedia.org/wiki/Inline_function#Restrictions (It's the best source, I know.)

icing:
I'd thought unity builds would evalulate the #idndef protected curl_trc.h only once. But I have no expierence with that...

jzakrzewski:
@vszakats Wouldn't the header guards take care of the duplicate definitions?

vszakats:
@jzakrzewski Oh, you're right, they would indeed!

vszakats:
Pushed a commit to test the static-function-in-header option.

vszakats:
The header option results in large amount of these kind of warnings:
```
In file included from sendf.h:29,
                 from altsvc.c:37:
curl_trc.h:125:13: error: ‘Curl_trc_cf_infof’ defined but not used [-Werror=unused-function]
  125 | static void Curl_trc_cf_infof(struct Curl_easy *data, struct Curl_cfilter *cf,
      |             ^~~~~~~~~~~~~~~~~
```

First because `curl_trc.h` gets pulled via `sendf.h` even into files with no
verbose calls. This can be untangled (but carefully, as we need to keep
`memdebug.h` last, which gets now pulled via `curl_trc.h`, so the latter
has to come last too). But then comes those files which include `curl_trc.h` 
directly, but only use only one of `infof()` / `CURL_TRC_CF()`, not both.
This needs splitting the header into two.

Perhaps there is an order emerging from this at the end, but definitely 
looks like something to do in a separate PR.

bagder:
A static function in the header file is a pain because of that warning. I propose to rather stick to the simple condition the code in master uses.

vszakats:
@bagder By simple codition in master I assume you mean `defined(CURL_DISABLE_VERBOSE_STRINGS`, is that right?

That said I just did a few rounds of `curl_trc.h` #include cleanups and it caught dupes, wrong include orders, and such. It also simplified the include tree by dropping `curl_trc.h` from `sendf.h`. In effect it replaces the trailing trio of `curl_printf.h`, `curl_memory.h`, `memdebug.h` with `curl_trc.h`. We might want to keep this work in a separate PR without the statics-in-header bit.

bagder:
> By simple codition in master I assume you mean defined(CURL_DISABLE_VERBOSE_STRINGS), is that right?

I meant in the same vein as this:

~~~c
#define infof(data, ...) \
  do { if(Curl_trc_is_verbose(data)) \
         Curl_infof(data, __VA_ARGS__); } while(0)
#define CURL_TRC_CF(data, cf, ...) \
  do { if(Curl_trc_cf_is_verbose(cf, data)) \
         Curl_trc_cf_infof(data, cf, __VA_ARGS__); } while(0)
~~~

Simple checks that avoid the function calls if verbose is not set.

Maybe like:
~~~c
#define infof(args) \
  do { if(Curl_trc_is_verbose(data)) \
         Curl_infof args; } while(0)
#define CURL_TRC_CF(args) \
  do { if(Curl_trc_cf_is_verbose(cf, data)) \
         Curl_trc_cf_infof args; } while(0)
~~~

... which "only" requires us to do those function calls with double parentheses.

vszakats:
@bagder That'd need to use something like `CURL_TRC_CF(data, cf, (arg1, arg2))`. We could then use `data` and `cf` to make the `Curl_trc_cf_is_verbose()` call. But we could _not_ pass then down to `Curl_trc_cf_infof()` via its argument list, even though this function also needs them. The argument list is a single, opaque list of expressions: `(arg1, arg2)`. We'd need to store `data` and `cf` it in per-thread variables and load them from there in `Curl_trc_cf_infof()`.

bagder:
Oh right. Unless the macro *requires* that `data` and `cf` are existing local variables of course. But that kind of macro is ugly and so annoying to understand when you get compiler errors down the line...

bagder:
Drastic alternative:

Document the requirement for C99 varargs support if you want to build with verbose disabled?

vszakats:
Are the issues only with verbose disabled though? (I haven't gone into the reports in detail. Remembering libssh2, it was a mess to navigate the CI warnings/errors there.)

IMO if we advertise curl as C89 compatible, it is a good thing if it actually is.

icing:
> Are the issues only with verbose disabled though? (I haven't gone into the reports in detail. Remembering libssh2, it was a mess to navigate the CI warnings/errors there.)
> 
> IMO if we advertise curl as C89 compatible, it is a good thing if it actually is.
> 
> PS: I could tolerate the doubly-passed variables as the drastic alternative.

We have to balance the cost against the purity of approach.

* I presume we are talking about a very niche build situation
* We do not want tracing (which is a non-debug feature now) to become significanlty more costly for the majority of our users. And we'd like to add more tracing points to our code, simply because it makes analysis of reported issues much easier/less effort for us
* We also would like the compiler to check format strings of our trace calls. Does that work with the change?

vszakats:
printf format check continue to work, yes.

bagder (approved):
I think we can stick with this solution. The biggest different to the previous might be that we had an `if()` in the macro and now we instead need to make a function call first to evaluate the `if()` even when nothing is logged. But even on the slowest CPUs curl ever runs on, that should be insignificant enough to not be a blocker.

icing:
> I think we can stick with this solution. The biggest difference to the previous might be that we had an `if()` in the macro and now we instead need to make a function call first to evaluate the `if()` even when nothing is logged. But even on the slowest CPUs curl ever runs on, that should be insignificant enough to not be a blocker.

hmm, it hurts. it hurts my soul that we do that for such a compile edge case...

icing:
Code maintenance is my concern. It was common practise up until we introduced these TRC calls to add `fprintf(stderr, ...)` while debugging an issue and, once fixed, remove these again from the code. That is not only a waste of effort, but a problem when people have unexplained behaviour on a deployment that we cannot easily reproduce. This happens all the time when curl is used outside the public internet.

Guarding those with `if()` or `#ifdef DEBUG_THIS_PROTO` reduces code readability.

We have already been in the world were trace calls were expensive/added only for local debugging. I would like us to move away from that.

vszakats:
If `infof(data, ((data, arg1, arg2))` looks interesting/acceptable, I'm OK to do
some tests with it to see if it could be useful.

icing:
How does #12164 look to you?

vszakats:
#12164: To me it seems cleaner to address the root problem then trying to silence warnings. I'm not saying it's impossible to do it, but I didn't explore this, so can't tell how effective it is. The overall notion of using this feature seems complex (two types of feature detections in two build tools, compiler-specific warning suppressions, STDC-specific code, etc). My impression is that this will continue to cause problems depending on compiler and/or version.

The unused-variable warning fixes are more solid than in this patch. I'll adapt those a bit later.

icing:
> #12164: To me it seems cleaner to address the root problem then trying to silence warnings. I'm not saying it's impossible to do it, but I didn't explore this, so can't tell how effective it is. The overall notion of using this feature seems complex (two types of feature detections in two build tools, compiler-specific warning suppressions, STDC-specific code, etc). My impression is that this will continue to cause problems depending on compiler and/or version.
> 
> The unused-variable warning fixes are more solid than in this patch. I'll adapt those a bit later.

Maybe we have just painted ourselves into a corner. In Apache httpd, we simply do:

```
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
#define AP_HAVE_C99
#endif

...
#ifdef AP_HAVE_C99
/* need additional step to expand APLOG_MARK first */
#define ap_log_error(...) ap_log_error__(__VA_ARGS__)
/* need server_rec *sr = ... for the case if s is verbatim NULL */
#define ap_log_error__(file, line, mi, level, status, s, ...)           \
    do { const server_rec *sr__ = s; if (APLOG_MODULE_IS_LEVEL(sr__, mi, level)) \
             ap_log_error_(file, line, mi, level, status, sr__, __VA_ARGS__);    \
    } while(0)
#else
#define ap_log_error ap_log_error_
#endif
AP_DECLARE(void) ap_log_error_(const char *file, int line, int module_index,
                               int level, apr_status_t status,
                               const server_rec *s, const char *fmt, ...)
                              __attribute__((format(printf,7,8)));
```

icing:
made #12167 to check the apache style approach.

On `--disable-verbose` static functions are declared that do not do anything and should be optimized away easily. The difference to before is that all args to these calls are checked by the compiler.

vszakats:
Reshuffled commits to do the same job in fewer steps. Then rebased to master (this PR rots _fast_).
Also merged warning updates from #12164. Except for http2.c which has seen an update since then in #12172.

PR #12109, opened 2023-10-13
curl.h: only include "osreldate.h" on pre v8 FreeBSD

### LINKED ISSUE #12107: osreldate.h doesn't exist on PlayStation
### I did this

Build fails on PlayStation due to missing `osreldate.h`

### I expected the following

_No response_

### curl/libcurl version

curl 8.4.0

### operating system

PlayStation

bagder:
Follow-up to b89789d82fdb2df. This header does not exist in all modern versions.

Fixes #12107
Reported-by: Faraz Fallahi

### DIFF
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -65,7 +65,8 @@
 #include <stdio.h>
 #include <limits.h>
 
-#if (defined(__FreeBSD__) && (__FreeBSD__ >= 2)) || defined(__MidnightBSD__)
+#if (defined(__FreeBSD__) && (__FreeBSD__ >= 2) && (__FreeBSD__ < 8)) || \
+  defined(__MidnightBSD__)
 /* Needed for __FreeBSD_version or __MidnightBSD_version symbol definition */
 #include <osreldate.h>
 #endif


include/curl/curl.h
```diff
@@ -65,7 +65,8 @@
 #include <stdio.h>
 #include <limits.h>
 
-#if (defined(__FreeBSD__) && (__FreeBSD__ >= 2)) || defined(__MidnightBSD__)
+#if (defined(__FreeBSD__) && (__FreeBSD__ >= 2) && (__FreeBSD__ < 8)) || \
+  defined(__MidnightBSD__)
```
jay, line 69:
```suggestion
#if defined(__FreeBSD__) && (__FreeBSD__ >= 8)
#include <sys/param.h>
#elif (defined(__FreeBSD__) && (__FreeBSD__ >= 2) && (__FreeBSD__ < 8)) || \
  defined(__MidnightBSD__)
```

from the issue

> FreeBSD manual [says](https://docs.freebsd.org/en/books/porters-handbook/porting-dads/#porting-versions) use `#include <sys/param.h>`.

It may apply to Midnight BSD as well but I can't find any documentation on it.

Nobody else has reported this which is surprising. I think maybe the reporter had some bad cross compile, and maybe osreldate still exists. It seems like we would've heard about it already.

MarcelRaad:
Do we need to include any header at all? We now use `__FreeBSD__` to be able to check for `__FreeBSD_version`. As we only check for a major version with `__FreeBSD_version`, can't we just use `__FreeBSD__` and remove the include?

bagder:
A significant problem here is that we just edit and have ideas about this blindly, None of us actually builds on (really old) FreeBSD or MidnightBSD...

Maybe we can remove the `__FreeBSD_version` check. I just can't see how we can figure that out without just going ahead and do it and then see if someone yells at us later on.

include/curl/curl.h
emaste: osreldate.h still exists on modern FreeBSD versions; it hasn't been removed.

bagder:
Based on feedback this seems to be based on a wrong assumption. Closing.

PR #12115, opened 2023-10-13
tests: Fix Windows test helper tool search & use it for handle64

dfandrich:
The checkcmd() and checktestcmd() functions would not have worked on
Windows due to hard-coding the UNIX PATH separator character and not
adding .exe file extension. This meant that tools like stunnel, valgrind
and nghttpx would not have been found and used on Windows, and
inspection of previous test runs show none of those being found in pure
Windows CI builds.

With this fixed, they can be used to detect the handle64.exe program
before attempting to use it. When handle64.exe was called
unconditionally without it existing, it caused perl to abort the test
run with the error

    The running command stopped because the preference variable
    "ErrorActionPreference" or common parameter is set to Stop:
    sh: handle64.exe: command not found

Closes #12115

### DIFF
--- a/tests/pathhelp.pm
+++ b/tests/pathhelp.pm
@@ -789,6 +789,7 @@ sub exe_ext {
         $^O eq 'dos' || $^O eq 'os2') {
         return '.exe';
     }
+    return '';
 }
 
 1;    # End of module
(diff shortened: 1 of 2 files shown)


bagder (approved):

jay:
I think it would be worthwhile to rebase this on a bad commit (that actually causes handle64 to be invoked) and run it through CI to make sure it functions as expected.

dfandrich:
Testing it is the trick. I'll try pushing a build to the CI that randomly clears locks some percentage of the time to see how it goes.

PR #12116, opened 2023-10-13
tool_getparam: Limit --rate to be smaller than number of ms

sohomdatta1:
Currently, curl allows users to specify absurd request rates that might be higher than the number of milliseconds in the unit (ex: `curl --rate 3600050/h http://localhost:8080` does not error out despite there being only 3600000ms in a hour).

This PR adds a conditional check before the millisecond calculation making sure that the number is not higher than the numerator (the unit) If the number is higher, curl errors out with `PARAM_NUMBER_TOO_LARGE`

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -1043,6 +1043,12 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
             break;
           }
         }
+
+        if(denominator > numerator) {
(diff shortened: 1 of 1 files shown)


sohomdatta1:
In the current case, the result of the division is zero for numbers higher than the number of ms in the unit is always, zero, so it doesn't really make sense to keep that case around

bagder (approved):

bagder:
Thanks!

Issue #12125, opened 2023-10-15
Test 1014 failure on FreeBSD 6.2


dfandrich:
### I did this

This test failues to to curl reporting `threadsafe` but not `curl-config`
```
Mismatch in features lists:
curl:        alt-svc asynchdns ipv6 largefile libz threadsafe unixsockets
curl-config: alt-svc asynchdns ipv6 largefile libz unixsockets
test 1014...[Compare curl --version with curl-config --features]
 1014: postcheck FAILED
== Contents of files in the log/1/ dir after test 1014
=== Start of file server.cmd
 Testnum 1014
=== End of file server.cmd
=== Start of file stdout1014
 curl 8.4.1-DEV (i586-unknown-freebsd6.2) libcurl/8.4.1-DEV zlib/1.2.3
 Release-Date: [unreleased]
 Protocols: dict file ftp gopher http imap mqtt pop3 rtsp smtp telnet tftp
 Features: alt-svc AsynchDNS IPv6 Largefile libz threadsafe UnixSockets
=== End of file stdout1014
```
The configure test `checking for stdatomic.h` returns `no`, which is correct—it doesn't exist. However, `pthread.h` does exist.

This may be related to #12090.

### I expected the following

No test errors.

### curl/libcurl version

curl-8.4.0-dev

### operating system

FreeBSD 6.2 i386

jay:
Please try #12127 in autotools and cmake and see if it says threadsafe in the build results

PR #12127, opened 2023-10-15
build: fix 'threadsafe' feature detection for older gcc

jay:
- Add 'threadsafe' to the feature list shown during build if POSIX threads are being used.

This is a follow-up to 5adb6000 which added support for building a thread-safe libcurl with older versions of gcc where atomic is not available but pthread is.

Reported-by: Dan Fandrich

Fixes https://github.com/curl/curl/issues/12125
Closes #xxxx

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1548,8 +1548,9 @@ _add_if("HTTPS-proxy"   SSL_ENABLED AND (USE_OPENSSL OR USE_GNUTLS
                         OR USE_SCHANNEL OR USE_RUSTLS OR USE_BEARSSL OR
                         USE_MBEDTLS OR USE_SECTRANSP))
 _add_if("unicode"       ENABLE_UNICODE)
-_add_if("threadsafe"    HAVE_ATOMIC OR (WIN32 AND
-                        HAVE_WIN32_WINNT GREATER_EQUAL 0x600))
+_add_if("threadsafe"    HAVE_ATOMIC OR
+                        (USE_THREADS_POSIX AND HAVE_PTHREAD_H) OR
+                        (WIN32 AND HAVE_WIN32_WINNT GREATER_EQUAL 0x600))
 _add_if("PSL"           USE_LIBPSL)
 string(REPLACE ";" " " SUPPORT_FEATURES "${_items}")
 message(STATUS "Enabled features: ${SUPPORT_FEATURES}")
--- a/configure.ac
+++ b/configure.ac
@@ -4612,6 +4612,9 @@ fi
 
 if test "$tst_atomic" = "yes"; then
   SUPPORT_FEATURES="$SUPPORT_FEATURES threadsafe"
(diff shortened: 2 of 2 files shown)


dfandrich:
I don't know how to easily get cmake going on this ancient FreeBSD, but I tested the autoconf version but it didn't work. The problem is that `HAVE_PTHREAD_H` isn't set as an environment variable anywhere. Replacing that clause with `x"$ac_cv_header_pthread_h" = x"yes"`  (the superfluous `x`s are in there for compatibility with ancient shells that don't handle empty arguments well) fixes it. Thanks!

jay:
> I tested the autoconf version but it didn't work. The problem is that `HAVE_PTHREAD_H` isn't set as an environment variable anywhere.

That is confusing because it is supposed to be defined when pthreads is used isn't it?

https://github.com/curl/curl/blob/d755a5f7c009dd63a61b2c745180d8ba937cbfeb/configure.ac#L3741-L3744

> Replacing that clause with `x"$ac_cv_header_pthread_h" = x"yes"` (the superfluous `x`s are in there for compatibility with ancient shells that don't handle empty arguments well) fixes it. Thanks!

Thanks I have incorporated your feedback. Please try the most recently pushed version. I updated both conditionals to prepend the x but I notice in the file it's done both ways. When is one more appropriate than the other? If I understand correctly `test $foo = "bar"` would be a problem if $foo was some command line switch like --qux because then test would interpret it wrong, but in practice when is that actually going to happen with the variables that we set? Or if $foo was unset then wouldn't we have noticed it by now with all the conditionals elsewhere that don't prepend the x?

dfandrich:
> That is confusing because it is supposed to be defined when pthreads is used
> isn't it?

Those definitions live in a different namespace and aren't available later in
the configure script. I'll try your change tomorrow.

dfandrich:
I tested the update patch with autotools and it works fine. I simulated a missing stdatomic.h in a cmake build on another system and that worked as well.

dfandrich:
Just noticed the `&&` in the test command. I'm pretty sure that's a GNUism and should be changed to `-a` for portability.

jay:
Thanks. I have added your change, please take another look.

dfandrich:
I mis-parsed the statement. It's fine as-is since the shell spawns a second test for the second test clause, but it would avoid a fork if `&& test` is replaced with `-a`.

dfandrich:
Yes, that's exactly what I was thinking.

dfandrich (approved):

dfandrich:
W.r.t. the x's in the `x"$FOO" = xyz` question, yes, we likely would have seen fallout already from all the cases where we don't do that. It's mostly just really, really old shells that have an issue with testing against an empty string. POSIX requires it so it's only pre-POSIX environments where it could make a difference, but since POSIX predates C89 so there could theoretically be a system where it's needed. But, I wouldn't lose too much sleep over it.

Issue #12124, opened 2023-10-15
Compile error on old gcc


dfandrich:
### I did this

Compiling on FreeBSD 6.2 on gcc 3.4.6 (from over 16 years ago) gives this error:
```
tool_urlglob.c:77 error: implicit declaration of function `__builtin_mul_overflow'
```
Presumably, this was added to gcc some time between 3.4 and 6 and the #ifdef needs to check the gcc version. Since we're supporting C89, one can't complain that this is too old a compiler.

### I expected the following

This old compiler should work (and after fixing the above, it does).

### curl/libcurl version

curl 8.4.0-dev

### operating system

FreeBSD 6.2 i386

PR #12128, opened 2023-10-15
tool_urlglob: fix build for old gcc versions

jay:
- Don't use __builtin_mul_overflow for GCC 4 and earlier.

The function was added in GCC 5.

Ref: https://gcc.gnu.org/gcc-5/changes.html

Reported-by: Dan Fandrich

Fixes https://github.com/curl/curl/issues/12124
Closes #xxxx


dfandrich:
That patch works. Thanks!

PR #12130, opened 2023-10-15
Added missing parameter 'lpNumberOfCharsWritten' for WriteConsoleW()

lordmulder:
This is documented as "optional" by Microsoft, I know. But, apparently, this was *not* optional on older Windows versions! On those version, the function verifiably fails with error `ERROR_INVALID_ACCESS`, if parameter is `NULL`.

As a result, we get error:
```
curl: (23) Failure writing output to destination
```

Probably regression from:
https://github.com/curl/curl/commit/af3f4e419b9f339790de281c871640a773c391c0

This PR fixes the error 👍

### DIFF
--- a/src/tool_cb_wrt.c
+++ b/src/tool_cb_wrt.c
@@ -237,7 +237,7 @@ size_t tool_write_cb(char *buffer, size_t sz, size_t nmemb, void *userdata)
   if(isatty(fileno(outs->stream)) &&
      GetConsoleScreenBufferInfo((HANDLE)fhnd, &console_info)) {
     wchar_t *wc_buf;
-    DWORD wc_len;
+    DWORD wc_len, chars_written;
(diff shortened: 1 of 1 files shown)


jay:
> This is documented as "optional" by Microsoft, I know. But, apparently, this was _not_ optional on older Windows versions!

I can confirm. Here's a screenshot from the Windows 7 SDK local help file:

Capture

They should not have changed the documentation unless they were correcting it and it was always optional.

lordmulder:
I guess, if it's optional on all Windows versions _that M$ still officially supports_, then they don't bother to reflect the old behavior in their documentation... 🙄

jay:
> they don't bother to reflect the old behavior or requirements in their documentation

I reported it to MS, refer to MicrosoftDocs/Console-Docs#299 for further discussion.

jay:
Thanks

PR #12132, opened 2023-10-15
GHA: bump ngtcp2 and nghttp3 to 1.0.x, nghttp2 to 1.58.0

bagder:
"No more incompatible changes."

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -47,7 +47,7 @@ env:
   mod_h2-version: v2.0.25
   msh3-version: v0.6.0
   openssl3-version: openssl-3.1.3
-  quictls-version: OpenSSL_1_1_1w+quic
+  quictls-version: 3.1.4+quic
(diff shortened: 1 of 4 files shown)


bagder:
The lastest nghttp2 release (1.57.0) cannot build nghttpx with these versions, so we probably need a new version of that as well.

vszakats:
nghttp2 1.78.0 is out: https://github.com/nghttp2/nghttp2/releases/tag/v1.58.0

icing:
ngtcp2 has a v1.0.1 released. currently testing locally

vszakats:
If we're at this, quictls also got an update yesterday to 3.0.12+quic, though 3.1.4+quic is also available.

bagder:
> If we're at this, quictls also got an update yesterday to 3.0.12+quic, and 3.1.4+quic is also available.

Yeps, might as well do that while we're there.

Issue #12129, opened 2023-10-15
curl tool won't search for curlrc (no leading dot) in XDG_CONFIG_HOME


jay:
### I did this

@b-jazz reported in [#8208 (comment)](https://github.com/curl/curl/issues/8208#issuecomment-1763283082) that when curl is searching for the .curlrc file, it searches XDG_CONFIG_HOME for .curlrc but does not search it for curlrc (no leading dot).

The original reporter of #8208 reported that as well. When that issue was fixed by 764e4f06 it did not fix that part of the issue. Instead it was written in a specific way to not do that. findfile will search for curlrc (no leading dot) in $CURL_HOME/.config/curlrc and $HOME/.config/curlrc, but only if XDG_CONFIG_HOME is not set and .curlrc has not been found anywhere else.

Here's two typical setups and how curl currently behaves:

If CURL_HOME is not set, and XDG_CONFIG_HOME is set, and HOME is set, then the order is:

- $XDG_CONFIG_HOME/.curlrc
- $HOME/.curlrc
- [`pw_dir`](https://pubs.opengroup.org/onlinepubs/009604499/basedefs/pwd.h.html)/.curlrc

If CURL_HOME is not set, and XDG_CONFIG_HOME is **not** set, and HOME is set, then the order is:

- $HOME/.curlrc
- $HOME/.config/curlrc _(this path is checked because XDG_CONFIG_HOME is not set)_
- [`pw_dir`](https://pubs.opengroup.org/onlinepubs/009604499/basedefs/pwd.h.html)/.curlrc

Note pw_dir is usually the same as $HOME.

Note for brevity I've excluded how we also search for _curlrc when .curlrc is not found, since that logic just complicates understanding this and AFAICT it's not relevant here.

### I expected the following

I'm not sure if this is a bug or not since the commit was written specifically not to strip the dot from curlrc when checking XDG_CONFIG_HOME, and I can't figure from the discussion why it was done that way.

### curl/libcurl version

curl 8.4.0

### operating system

Linux

jay:
[-K, --config](https://curl.se/docs/manpage.html#-K) doc says it should be doing `$XDG_CONFIG_HOME/curlrc` so then is there not supposed to be any `$XDG_CONFIG_HOME/.curlrc`? If that's the case fixing it won't be as simple as turning on `withoutdot` member for XDG_CONFIG_HOME because there's logic that skips XDG_CONFIG_HOME in that case.

b-jazz:
[...]

PR #12135, opened 2023-10-16
tool_findfile: detect files in XDG home without leading dot

jay:
WIP

The order is now:

- $XDG_CONFIG_HOME/curlrc
- $XDG_CONFIG_HOME/.curlrc
- $XDG_CONFIG_HOME/_curlrc (if dotscore is true)

Fixes https://github.com/curl/curl/issues/12129
Closes #xxxx

Prior to this change there was no check for the dotless version of the filename.

### DIFF
--- a/docs/cmdline-opts/config.d
+++ b/docs/cmdline-opts/config.d
@@ -58,7 +58,11 @@ config file is checked for in the following places in this order:
 
 1) **"$CURL_HOME/.curlrc"**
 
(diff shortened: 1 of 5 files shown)


jay:
Any opinions on this?

dfandrich:
I haven't reviewed the code, but conceptually, this seems like the right thing to do. There's no need for a leading dot in files in ~/.config, and practically no programs do it.

bagder:
Stalled and merge conflicts, closing for now.

PR #12137, opened 2023-10-16
gnutls: support CURLSSLOPT_NATIVE_CA

bagder:
Remove the CURL_CA_FALLBACK logic. That build option was added to allow primarily OpenSSL to use the default paths for loading the CA certs. For GnuTLS it was instead made to load the "system certs", which is different and not desirable.

The native CA store loading is now asked for with this option.

Follow-up to 7b55279d1d856


jay:
also, unlike openssl, gnutls does not check verifypeer before importing certificates. if there is not a good reason for that we could put all the imports behind a if(verifypeer) like we do for openssl

bagder:
> if there is not a good reason for that we could put all the imports behind a if(verifypeer) like we do for openssl

Oh right, I think we should!

bagder:
Nice!

PR #12140, opened 2023-10-17
PlayStation doesn't have getenv()

fffaraz:
Build fails on PlayStation since it doesn't have `getenv()`

### DIFF
--- a/lib/getenv.c
+++ b/lib/getenv.c
@@ -31,7 +31,8 @@
 
 static char *GetEnv(const char *variable)
 {
-#if defined(_WIN32_WCE) || defined(CURL_WINDOWS_APP)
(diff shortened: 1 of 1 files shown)


lib/getenv.c
```diff
@@ -31,7 +31,7 @@
 
 static char *GetEnv(const char *variable)
 {
-#if defined(_WIN32_WCE) || defined(CURL_WINDOWS_APP)
+#if defined(_WIN32_WCE) || defined(CURL_WINDOWS_APP) || defined(__ORBIS__) || defined(__PROSPERO__)
```
bagder, line 34: This is a too long line that will make the CI builds cry.,

But also `__ORBIS__`  and `__PROSPERO__` are not widely known (missing from [predef](https://github.com/cpredef/predef)) so they require explanations in a comment.

PR #12141, opened 2023-10-17
cookie: branchless comparison for sorting by creation time

visitorckw:
In the cookie module, when sorting cookies by creation time, the comparison logic was previously implemented using a ternary operator, which had room for optimization. This change replaces the ternary operator with a direct subtraction, making the comparison branchless and reducing the number of generated assembly instructions.

### DIFF
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -1320,7 +1320,7 @@ static int cookie_sort(const void *p1, const void *p2)
     return (l2 > l1) ? 1 : -1;
 
   /* 4 - compare cookie creation time */
-  return (c2->creationtime > c1->creationtime) ? 1 : -1;
+  return c2->creationtime - c1->creationtime;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


jay:
we can't do it that way because signed integer overflow is undefined behavior. also the compiler will optimize statements like that, for example https://godbolt.org/z/GWY3e7jWx
~~~asm
xor     ecx, ecx
cmp     edi, eax
setg    cl
~~~

visitorckw:
Thank you for your response.
It's possible that I misunderstood. I initially assumed that ```creationtime``` starts from 0 and increments by 1 each time, so there would be no negative values, and thus no potential for integer overflow when using subtraction. On the other hand, it was my mistake to conduct the test on x86-64 gcc 9.4.0 using the default -O0 flag, which led to the observation of the compiler generating branch instructions. I should have conducted the test using the -O2 flag, where the reduction in instructions might be minimal, possibly just 1 or 2 instructions.

PR #12142, opened 2023-10-17
multi: use pipe instead of socketpair to *wakeup()

bagder:
If pipe() is present. Less overhead.


bagder:
Ultimately, we can probably switch to use pipe() instead of socketpair() for the other two uses as well: asyn-thread and ntlm_wb.

PR #12144, opened 2023-10-17
tracing defaults for multi and share handles

icing:
Configure a multi handle to provide defaults for added easy handles when it comes to tracing:
- CURLMOPT_DEBUGDATA, CURLMOPT_DEBUGFUNCTION, CURLMOPT_VERBOSE

Configure a shared handle to provide defaults for easy handles that use the share when it comes to tracing:
- CURLSHOPT_DEBUGDATA, CURLSHOPT_DEBUGFUNCTION, CURLSHOPT_VERBOSE

Defaults for DEBUGDATA and DEBUGFUNCTION apply to easy handles where none of them are already set.
Default to VERBOSE act only as enabler. It will never turn off verbosity in an easy handle that has it set.

This is now used in the `curl` tool itself.

- when using curl verbose, the closing of connections in the pool become visible
- setting tracing on the easy handles could be removed, however `--libcurl` would then no longer generate the code for it. That is why I left it in.


docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,70 @@
...
+CURLcode curl_multi_setopt(CURLM *handle, CURLOPT_DEBUGFUNCTION,
+                          debug_callback);
+.SH DESCRIPTION
+Pass a pointer to your callback function, which should match the prototype
+that is defined for \fICURLOPT_DEBUGFUNCTION(3)\fP.
+
+\fICURLMOPT_DEBUGFUNCTION(3)\fP and \fICURLMOPT_DEBUGDATA\fP can be used
+as defaults for all easy handles that are added to the multi. They are also
+used for internal handles, for examples when closing cached connections.
+They take effect only for easy handles added to the multi handle afterwards.
```
bagder, line 41: I did not get this when I first read it. I propose adding a few more words:

```suggestion
These options take effect and are used only for the easy handles that are added to the multi handle after these options have been set for the multi handle.
```

docs/libcurl/opts/CURLMOPT_VERBOSE.3
```diff
@@ -0,0 +1,61 @@
...
+.fi
+.SH DESCRIPTION
+Set the \fIonoff\fP parameter to 1 to make the library display a lot of
+verbose information about its operations on \fIhandle\fPs added to
+this multi.
+
+\fICURLMOPT_VERBOSE(3)\fP can be used as default for all easy handles
+that are added to the multi. They are also used for internal handles,
+for examples when closing cached connections.
+They take effect only for easy handles added to the multi handle afterwards.
```
bagder, line 42:
```suggestion
This option take effect and is used only for the easy handles that are added to the multi handle after this option is set for the multi handle.
```

docs/libcurl/opts/CURLMOPT_VERBOSE.3
```diff
@@ -0,0 +1,61 @@
...
+.\" **************************************************************************
+.\"
+.TH CURLMOPT_VERBOSE 3 "17 Oct 2023" libcurl libcurl
+.SH NAME
+CURLMOPT_VERBOSE \- verbose mode
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+
+CURLcode curl_multi_setopt(CURLM *handle, CURLOPT_VERBOSE, long onoff);
```
bagder, line 32:
```suggestion
CURLcode curl_multi_setopt(CURLM *handle, CURLMOPT_VERBOSE, long onoff);
```

docs/libcurl/opts/CURLMOPT_VERBOSE.3
```diff
@@ -0,0 +1,61 @@
...
+that are added to the multi. They are also used for internal handles,
+for examples when closing cached connections.
+They take effect only for easy handles added to the multi handle afterwards.
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+main()
+{
+  GlobalInfo setup;
```
bagder, line 49: Maybe drop this line?

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,70 @@
...
+NULL
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+main()
+{
+  struct data my_tracedata;
+
+  GlobalInfo setup;
```
bagder, line 55: Maybe remove this?

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,70 @@
...
+  /* ... use debug callback and custom pointer */
+  curl_multi_setopt(multi, CURLMOPT_DEBUGFUNCTION, my_trace);
+  curl_multi_setopt(multi, CURLMOPT_DEBUGDATA, &my_tracedata);
+}
+.fi
+.SH AVAILABILITY
+Always
```
bagder, line 62:
```suggestion
This option was added in libcurl 8.5.0.
```

docs/libcurl/opts/CURLMOPT_VERBOSE.3
```diff
@@ -0,0 +1,61 @@
...
+  /* ... use debug callback and custom pointer */
+  curl_multi_setopt(multi, CURLMOPT_VERBOSE, 1L);
+}
+.fi
+.SH AVAILABILITY
+Always
```
bagder, line 55:
```suggestion
This option was added in libcurl 8.5.0.
```

docs/libcurl/opts/CURLSHOPT_DEBUGDATA.3
```diff
@@ -0,0 +1,71 @@
...
+{
+  struct data my_tracedata;
+
+  share = curl_share_init();
+  curl_share_setopt(share, CURLSHOPT_DEBUGFUNCTION, my_trace);
+  curl_share_setopt(share, CURLSHOPT_DEBUGDATA, &my_tracedata);
+}
+.fi
+.SH AVAILABILITY
+Always
```
bagder, line 64:
```suggestion
This option was added in libcurl 8.5.0.
```

docs/libcurl/opts/CURLMOPT_DEBUGDATA.3
```diff
@@ -0,0 +1,70 @@
...
+  struct data my_tracedata;
+
+  GlobalInfo setup;
+  /* ... use debug callback and custom pointer */
+  curl_multi_setopt(multi, CURLMOPT_DEBUGFUNCTION, my_trace);
+  curl_multi_setopt(multi, CURLMOPT_DEBUGDATA, &my_tracedata);
+}
+.fi
+.SH AVAILABILITY
+Always
```
bagder, line 63:
```suggestion
This option was added in libcurl 8.5.0.
```

docs/libcurl/opts/CURLSHOPT_VERBOSE.3
```diff
@@ -0,0 +1,62 @@
...
+main()
+{
+  struct data my_tracedata;
+
+  share = curl_share_init();
+  curl_share_setopt(share, CURLSHOPT_VERBOSE, 1L);
+}
+.fi
+.SH AVAILABILITY
+Always
```
bagder, line 56:
```suggestion
This option was added in libcurl 8.5.0.
```

docs/libcurl/opts/CURLSHOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,71 @@
...
+{
+  struct data my_tracedata;
+
+  share = curl_share_init();
+  curl_share_setopt(share, CURLSHOPT_DEBUGFUNCTION, my_trace);
+  curl_share_setopt(share, CURLSHOPT_DEBUGDATA, &my_tracedata);
+}
+.fi
+.SH AVAILABILITY
+Always
```
bagder, line 62:
```suggestion
This option was added in libcurl 8.5.0.
```

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
icing: ack

docs/libcurl/opts/CURLMOPT_VERBOSE.3
icing: ack

docs/libcurl/opts/CURLMOPT_VERBOSE.3
icing: ack

docs/libcurl/opts/CURLMOPT_VERBOSE.3
icing: ack

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
icing: ack

docs/libcurl/opts/CURLMOPT_VERBOSE.3
icing: ack

docs/libcurl/opts/CURLSHOPT_DEBUGDATA.3
icing: ack

docs/libcurl/opts/CURLMOPT_DEBUGDATA.3
icing: ack

docs/libcurl/opts/CURLSHOPT_VERBOSE.3
icing: ack

docs/libcurl/opts/CURLSHOPT_DEBUGFUNCTION.3
icing: ack

docs/libcurl/opts/CURLMOPT_DEBUGDATA.3
```diff
@@ -0,0 +1,72 @@
...
+CURLcode curl_multi_setopt(CURLM *handle, CURLMOPT_DEBUGDATA, void *pointer);
+.fi
+.SH DESCRIPTION
+Pass a \fIpointer\fP to whatever you want passed in to your
+\fICURLMOPT_DEBUGFUNCTION(3)\fP in the last void * argument. This pointer is
+not used by libcurl, it is only passed to the callback.
+
+\fICURLMOPT_DEBUGFUNCTION(3)\fP and \fICURLMOPT_DEBUGDATA\fP can be used
+as defaults for all easy handles that are added to the multi. They are also
+used for internal handles, for examples when closing cached connections.
```
bagder, line 41:
```suggestion
used for internal handles, for example when closing cached connections.
```

docs/libcurl/opts/CURLMOPT_DEBUGDATA.3
```diff
@@ -0,0 +1,72 @@
...
+NULL
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+main()
+{
+  struct data my_tracedata;
+
+  GlobalInfo setup;
```
bagder, line 58:
```suggestion
```

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,66 @@
...
+
+CURLcode curl_multi_setopt(CURLM *handle, CURLOPT_DEBUGFUNCTION,
+                          debug_callback);
+.SH DESCRIPTION
+Pass a pointer to your callback function, which should match the prototype
+that is defined for \fICURLOPT_DEBUGFUNCTION(3)\fP.
+
+\fICURLMOPT_DEBUGFUNCTION(3)\fP and \fICURLMOPT_DEBUGDATA\fP can be used
+as defaults for all easy handles that are added to the multi. They are also
+used for internal handles, for examples when closing cached connections.
```
bagder, line 40:
```suggestion
used for internal handles, for example when closing cached connections.
```

docs/libcurl/opts/CURLMOPT_VERBOSE.3
```diff
@@ -0,0 +1,57 @@
...
+CURLcode curl_multi_setopt(CURLM *handle, CURLMOPT_VERBOSE, long onoff);
+.fi
+.SH DESCRIPTION
+Set the \fIonoff\fP parameter to 1 to make the library display a lot of
+verbose information about its operations on \fIhandle\fPs added to
+this multi.
+
+\fICURLMOPT_VERBOSE(3)\fP can be used as default for all easy handles
+that are added to the multi. They are also used for internal handles,
+for examples when closing cached connections.
```
bagder, line 41:
```suggestion
for example when closing cached connections.
```

docs/libcurl/opts/CURLMOPT_VERBOSE.3
```diff
@@ -0,0 +1,57 @@
...
+This option takes effect and is used only for the easy handles that are
+added to the multi handle after this option is set for the multi handle.
+.SH PROTOCOLS
```
bagder, line 44:
```suggestion
.SH DEFAULT
This is treated as if set to 0 by default.
.SH PROTOCOLS
```

docs/libcurl/opts/CURLSHOPT_DEBUGDATA.3
```diff
@@ -0,0 +1,71 @@
...
+                             void *pointer);
+.fi
+.SH DESCRIPTION
+Pass a \fIpointer\fP to whatever you want passed in to your
+\fICURLSHOPT_DEBUGFUNCTION(3)\fP in the last void * argument. This pointer is
+not used by libcurl, it is only passed to the callback.
+
+\fICURLSHOPT_DEBUGFUNCTION(3)\fP and \fICURLSHOPT_DEBUGDATA\fP can be used
+as defaults for all easy handles that use the same share. They are also
+used for internal handles, for examples when closing cached connections.
```
bagder, line 42:
```suggestion
used for internal handles, for example when closing cached connections.
```

docs/libcurl/opts/CURLSHOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,71 @@
...
+
+CURLSHcode curl_share_setopt(CURLSH *share, CURLSHOPT_DEBUGFUNCTION,
+                             debug_callback);
+.SH DESCRIPTION
+Pass a pointer to your callback function, which should match the prototype
+that is defined for \fICURLOPT_DEBUGFUNCTION(3)\fP.
+
+\fICURLSHOPT_DEBUGFUNCTION(3)\fP and \fICURLSHOPT_DEBUGDATA\fP can be used
+as defaults for all easy handles that use the same share. They are also
+used for internal handles, for examples when closing cached connections.
```
bagder, line 40:
```suggestion
used for internal handles, for example when closing cached connections.
```

docs/libcurl/opts/CURLSHOPT_VERBOSE.3
```diff
@@ -0,0 +1,62 @@
...
+                             long onoff);
+.fi
+.SH DESCRIPTION
+Set the \fIonoff\fP parameter to 1 to make the library display a lot of
+verbose information about its operations on \fIhandle\fPs used with
+this share.
+
+\fICURLSHOPT_VERBOSE(3)\fP can be used as default for all easy handles
+that use this share via \fICURLOPT_SHARE(3)\fP. It is also used for
+internal handles, for examples when closing cached connections.
```
bagder, line 42:
```suggestion
internal handles, for example when closing cached connections.
```

docs/libcurl/opts/CURLSHOPT_VERBOSE.3
```diff
@@ -0,0 +1,62 @@
...
+.SH PROTOCOLS
```
bagder, line 43:
```suggestion
.SH DEFAULT
0
.SH PROTOCOLS
```

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,66 @@
...
+These options take effect and are used only for the easy handles that are
+added to the multi handle after these options have been set for the multi
+handle.
+
+Values have the same restrictions and use patterns as defined for
+\fICURLOPT_DEBUGFUNCTION(3)\fP and \fICURLOPT_DEBUGDATA(3)\fP.
```
bagder, line 46:
```suggestion
\fICURLOPT_DEBUGFUNCTION(3)\fP and \fICURLOPT_DEBUGDATA(3)\fP.

Setting this option to NULL makes the function not be used from the share object and instead the value set in the easy handle dictates what function to call.
```

docs/libcurl/opts/CURLSHOPT_DEBUGFUNCTION.3
```diff
@@ -0,0 +1,71 @@
...
+They take effect only for easy handles when this share is set for them
+using \fICURLOPT_SHARE(3)\fP.
+
+Values have the same restrictions and use patterns as defined for
+\fICURLOPT_DEBUGFUNCTION(3)\fP and \fICURLOPT_DEBUGDATA(3)\fP.
```
bagder, line 45:
```suggestion
\fICURLOPT_DEBUGFUNCTION(3)\fP and \fICURLOPT_DEBUGDATA(3)\fP.

Setting this option to NULL makes the debug function not be used from the multi object and instead the value set in the easy handle dictates what function to call.
```

bagder (commented):
I went over it again and I have some polish to propose!

jay:
I've added CURLSHOPT_STDERR and CURLMOPT_STDERR, which is used by verbose, although I have to wonder if all these extra options are overkill and maybe there is a better way to solve this like a CURLOPT_DEBUGINHERIT or something so that the caller knows their CURLOPT_STDERR and CURLOPT_DEBUGFUNCTION/DATA can be used after the easy handle is cleaned up.

docs/libcurl/opts/CURLMOPT_DEBUGFUNCTION.3
icing: Both options would need to be set to NULL in order to not override the ease settings. I'll adapt the suggested wording.

docs/libcurl/opts/CURLSHOPT_DEBUGFUNCTION.3
icing: see above

icing:
Jay, you changed the way these variables are now inherited, seemingly in a way not intended by me. Also, the test cases hard crash for me.

Not a happy Monday morning...

jay:
> Jay, you changed the way these variables are now inherited, seemingly in a way not intended by me. Also, the test cases hard crash for me.

It's a WIP. How did you intend? My changes were in line with this:

> Defaults for DEBUGDATA and DEBUGFUNCTION apply to easy handles where none of them are already set.
> Default to VERBOSE act only as enabler. It will never turn off verbosity in an easy handle that has it set.

icing:
Had to force push my latest changes. Thanks for addressing the NULL-deref in the meantime.

jay:
I think the inheritance issue complicates things. If someone sets easy handle stderr, and not easy handle debug function, and sets multi debug function, then it would make sense to for multi debug function to be inherited but if I'm reading this right then right now that doesn't happen.

In the next few days I am going to write up an alternate proposal to do a CURLOPT_DEBUGINTERNAL to set the easy handle's debug settings on the conn cache handle and we can see how that looks.

icing:
> I think the inheritance issue complicates things. If someone sets easy handle stderr, and not easy handle debug function, and sets multi debug function, then it would make sense to for multi debug function to be inherited but if I'm reading this right then that doesn't happen.

The design in this PR is intended to make the full set of DEBUG/VERBOSE inheritable only. This is intended to give the "least surprise" to code that is setting these properties on easy handles. They will function as before.
 
> In the next few days I am going to write up an alternate proposal to do a CURLOPT_DEBUGINTERNAL to set the easy handle's debug settings on the conn cache handle and we can see how that looks.

The thing you seem to have in mind, and pushed into this PR before I reverted, follows another idea. Namely, that each individual property should be inherited. This gives rise to the problem of how to manage the case when the application does not want a property to be inherited, e.g. have DEBUGFUNCTION not set on an easy handle.

I find that leads to more "complicated" management and therefore decided not to do that. Looking forward to your proposal on how to manage this.

icing:
Did not make the cut. We are still looking for the optimal approach.

PR #12145, opened 2023-10-17
hostip: show the list of IPs when resolving is done

bagder:
Getting 'curl.se' today then gets this verbose output which might help debugging connectivity related matters.

* Host curl.se:80 was resolved.
* IPv6: 2a04:4e42:800::347, 2a04:4e42:600::347, 2a04:4e42:400::347, 2a04:4e42:200::347, 2a04:4e42::347, 2a04:4e42:e00::347, 2a04:4e42:c00::347, 2a04:4e42:a00::347
* IPv4: 151.101.193.91, 151.101.129.91, 151.101.65.91, 151.101.1.91

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -117,6 +117,13 @@
 
 static void freednsentry(void *freethis);
 
+#ifndef CURL_DISABLE_VERBOSE_STRINGS
+static void show_resolve_info(struct Curl_easy *data,
+                              struct Curl_dns_entry *dns);
+#else
+#define show_resolve_info(x,y) Curl_nop_stmt
+#endif
+
 /*
  * Curl_printable_address() stores a printable version of the 1st address
  * given in the 'ai' argument. The result will be stored in the buf that is
@@ -481,9 +488,11 @@ Curl_cache_addr(struct Curl_easy *data,
       return NULL;
   }
 #endif
+  if(!hostlen)
+    hostlen = strlen(hostname);
 
   /* Create a new cache entry */
-  dns = calloc(1, sizeof(struct Curl_dns_entry));
+  dns = calloc(1, sizeof(struct Curl_dns_entry) + hostlen);
   if(!dns) {
     return NULL;
   }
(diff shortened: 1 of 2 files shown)


bagder:
excellent, thanks!

jay:
It would be better if it showed hostname but there's no easy way to determine which is the hostname passed to resolve so I prpose add it in Curl_dns_entry. see most recent commit. I also added the 'none' to clarify if none were found:

~~~
* STATE: CONNECT => RESOLVING handle 0x747208; line 1985
* Host google.com:80 was resolved.
* IPv6: (none)
* IPv4: 142.250.81.238
* STATE: RESOLVING => CONNECTING handle 0x747208; line 2059
*   Trying 142.250.81.238:80...
~~~

bagder:
> +1 byte for hostname null terminator

That byte is already in the struct by default. Now you add one extra byte...

jay:
oops forgot about the hostname[1]. reset to 'fixup avoid warning with strcpy'. i saw some malloc crashes and i figured that was the problem.

bagder:
it is the strcpy that is wrong because the host name can be longer than hostlen...

lib/hostip.c
```diff
@@ -1285,18 +1299,89 @@ CURLcode Curl_loadhostpairs(struct Curl_easy *data)
...
+#ifdef CURLRES_IPV6
+  Curl_dyn_init(&out[1], 1024);
+#endif
+
+  while(a) {
+    if(
+#ifdef CURLRES_IPV6
+       (a->ai_family == PF_INET6) ||
+#endif
+       (a->ai_family == PF_INET)) {
```
fffaraz, line 1336: Error when compiling without IPv6 support:
```
error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
note: remove extraneous parentheses around the comparison to silence this warning
```

jay: Thanks but we don't use that warning.

PR #12146, opened 2023-10-17
asyn-thread: use pipe instead of socketpair for IPC when available

bagder:
If pipe() is present. Less overhead.

### DIFF
--- a/lib/socketpair.h
+++ b/lib/socketpair.h
@@ -40,7 +40,7 @@
 #define wakeup_close     sclose
 #define wakeup_create(p) Curl_socketpair(AF_UNIX, SOCK_STREAM, 0, p)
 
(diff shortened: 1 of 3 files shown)


lib/socketpair.h
```diff
@@ -25,6 +25,23 @@
...
+#define wakeup_create pipe
+
+#else /* HAVE_PIPE */
+
+#define wakeup_write     swrite
+#define wakeup_read      sread
+#define wakeup_close     sclose
+#define wakeup_create(p) Curl_socketpair(AF_UNIX, SOCK_STREAM, 0, p)
+
+#endif /* __HAVE_PIPE__ */
```
vszakats, line 43:
```suggestion
#endif /* HAVE_PIPE */
```
minor nit

Issue #12148, opened 2023-10-17
IPFS/IPNS URLs ignore the path


Stebalien:
### I did this

I run:

`curl -v ipns://ipfs.tech/community`

I get:

```
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET /ipns/ipfs.tech HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Access-Control-Allow-Headers: Content-Type
< Access-Control-Allow-Headers: Range
< Access-Control-Allow-Headers: User-Agent
< Access-Control-Allow-Headers: X-Requested-With
< Access-Control-Allow-Methods: GET
< Access-Control-Allow-Methods: HEAD
< Access-Control-Allow-Methods: OPTIONS
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Content-Length
< Access-Control-Expose-Headers: Content-Range
< Access-Control-Expose-Headers: X-Chunked-Output
< Access-Control-Expose-Headers: X-Ipfs-Path
< Access-Control-Expose-Headers: X-Ipfs-Roots
< Access-Control-Expose-Headers: X-Stream-Output
< Content-Type: text/html; charset=utf-8
< Location: /ipns/ipfs.tech/
< X-Ipfs-Path: /ipns/ipfs.tech
< X-Ipfs-Roots: QmbMTqnawMAbD4Fzgd5NM9GpKT67FdNZ1EsH8D4r52PTeT
< Date: Tue, 17 Oct 2023 18:26:48 GMT
< Content-Length: 51
< 
<a href="/ipns/ipfs.tech/">Moved Permanently</a>.
```

The same goes for all `ipfs://` urls:

`curl -v ipfs://bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze/wiki/cURL`

Output:

```
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET /ipfs/bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Access-Control-Allow-Headers: Content-Type
< Access-Control-Allow-Headers: Range
[...]

PR #12152, opened 2023-10-18
Broader IPFS and IPNS URL support like http/s

markg85:
Commit msg:
```
Previously just ipfs://<cid> and ipns://<cid> was supported. Which is too strict for some usecases.

This patch allows paths and query arguments to be used too. Making this work according to normal http semantics: ipfs://<cid>/foo/bar?key=val
ipns://<cid>/foo/bar?key=val

The gateway url support is changed.
It now only supports gateways in the form of:
http://<gateway>/foo/bar
http://<gateway>

Query arguments here are explicitly not allowed and trigger an intended malformed url error.

There also was a crash when IPFS_PATH was set with a non trailing forward slash. This has been fixed.

Lastly, a load of test cases have been added to verify the above.
```

This fixes https://github.com/curl/curl/issues/12148 and a couple more issues that had been found.

I've added a lot of testcases to verify the fixes.
Still, the testing isn't super elaborate (say everything for ipfs, then another time for ipns and then everything 3x for the 3 different ways to set a gateway). The reasoning is that much code is following the same path so an issue popping up for ipns will be there for ipfs too. As long as that stays as-is then those current tests cover a lot :)

cc @lidel @Stebalien

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -697,6 +697,32 @@ noretry:
   return result;
 }
 
+/* helper function to ensure input ends in val_to_ensure */
+static CURLcode ensure_trailing(char **input, const char val_to_ensure)
+{
+  if(*input && **input) {
+    size_t len = strlen(*input);
+    if(((*input)[len - 1] != val_to_ensure)) {
+      struct curlx_dynbuf dyn;
+      curlx_dyn_init(&dyn, len + 2);
+
+      if(curlx_dyn_addn(&dyn, *input, len)) {
+        Curl_safefree(*input);
+        return CURLE_OUT_OF_MEMORY;
+      }
+
+      Curl_safefree(*input);
+
+      if(curlx_dyn_addn(&dyn, &val_to_ensure, 1))
+        return CURLE_OUT_OF_MEMORY;
+
+      *input = curlx_dyn_ptr(&dyn);
+    }
+  }
+
+  return CURLE_OK;
+}
+
 static char *ipfs_gateway(void)
 {
   char *gateway = NULL;
@@ -704,37 +730,40 @@ static char *ipfs_gateway(void)
   char *gateway_composed_file_path = NULL;
   FILE *gateway_file = NULL;
 
-  gateway = getenv("IPFS_GATEWAY");
+  gateway = curlx_getenv("IPFS_GATEWAY");
 
   /* Gateway is found from environment variable. */
-  if(gateway && *gateway) {
-    char *composed_gateway = NULL;
-    bool add_slash = (gateway[strlen(gateway) - 1] != '/');
-    composed_gateway = aprintf("%s%s", gateway, (add_slash) ? "/" : "");
-    if(composed_gateway) {
-      gateway = aprintf("%s", composed_gateway);
-      Curl_safefree(composed_gateway);
+  if(gateway) {
+    if(ensure_trailing(&gateway, '/')) {
+      Curl_safefree(gateway);
+      return NULL;
     }
     return gateway;
   }
-  else
-    /* a blank string does not count */
-    gateway = NULL;
 
   /* Try to find the gateway in the IPFS data folder. */
-  ipfs_path = getenv("IPFS_PATH");
+  ipfs_path = curlx_getenv("IPFS_PATH");
 
   if(!ipfs_path) {
-    char *home = getenv("HOME");
+    char *home = curlx_getenv("HOME");
     if(home && *home)
       ipfs_path = aprintf("%s/.ipfs/", home);
     /* fallback to "~/.ipfs", as that's the default location. */
+
+    Curl_safefree(home);
   }
 
   if(!ipfs_path) {
     Curl_safefree(gateway);
     Curl_safefree(ipfs_path);
     return NULL;
+  }
+  else {
+    if(ensure_trailing(&ipfs_path, '/')) {
+      Curl_safefree(gateway);
+      Curl_safefree(ipfs_path);
+      return NULL;
+    }
   }
 
   gateway_composed_file_path = aprintf("%sgateway", ipfs_path);
@@ -749,24 +778,32 @@ static char *ipfs_gateway(void)
   Curl_safefree(gateway_composed_file_path);
 
   if(gateway_file) {
-    char *buf = NULL;
+    int c;
+    struct curlx_dynbuf dyn;
+    curlx_dyn_init(&dyn, INT_MAX);
 
-    if((PARAM_OK == file2string(&buf, gateway_file)) && buf && *buf) {
-      bool add_slash = (buf[strlen(buf) - 1] != '/');
-      gateway = aprintf("%s%s", buf, (add_slash) ? "/" : "");
+    /* get the first line of the gateway file, ignore the rest */
+    while((c = getc(gateway_file)) != EOF && c != '\n' && c != '\r') {
+      if(curlx_dyn_addn(&dyn, &c, 1))
+        break;
     }
-    Curl_safefree(buf);
 
     if(gateway_file)
       fclose(gateway_file);
 
+    if(curlx_dyn_len(&dyn) > 0)
+      gateway = curlx_dyn_ptr(&dyn);
+
+    if(gateway)
+      ensure_trailing(&gateway, '/');
+
     if(!gateway) {
-      Curl_safefree(gateway);
       Curl_safefree(ipfs_path);
       return NULL;
     }
 
     Curl_safefree(ipfs_path);
+
     return gateway;
   }
 
@@ -783,20 +820,26 @@ static CURLcode ipfs_url_rewrite(CURLU *uh, const char *protocol, char **url,
                                  struct OperationConfig *config)
 {
   CURLcode result = CURLE_URL_MALFORMAT;
-  CURLUcode urlGetResult;
+  CURLUcode getResult;
   char *gateway = NULL;
+  char *gatewayhost = NULL;
+  char *gatewaypath = NULL;
+  char *gatewayquery = NULL;
+  char *inputpath = NULL;
+  char *gatewayscheme = NULL;
+  char *gatewayport = NULL;
   char *cid = NULL;
   char *pathbuffer = NULL;
-  CURLU *ipfsurl = curl_url();
+  CURLU *gatewaysurl = curl_url();
 
-  if(!ipfsurl) {
+  if(!gatewaysurl) {
     result = CURLE_FAILED_INIT;
     goto clean;
   }
 
-  urlGetResult = curl_url_get(uh, CURLUPART_HOST, &cid, CURLU_URLDECODE);
+  getResult = curl_url_get(uh, CURLUPART_HOST, &cid, CURLU_URLDECODE);
 
-  if(urlGetResult) {
+  if(getResult) {
     goto clean;
   }
 
@@ -808,9 +851,14 @@ static CURLcode ipfs_url_rewrite(CURLU *uh, const char *protocol, char **url,
    * if we do have something but if it's an invalid url.
    */
   if(config->ipfs_gateway) {
-    if(curl_url_set(ipfsurl, CURLUPART_URL, config->ipfs_gateway,
-                    CURLU_GUESS_SCHEME)
-                    == CURLUE_OK) {
+    /* ensure the gateway ends in a trailing / */
+    if(ensure_trailing(&config->ipfs_gateway, '/') != CURLE_OK) {
+      result = CURLE_OUT_OF_MEMORY;
+      goto clean;
+    }
+
+    if(!curl_url_set(gatewaysurl, CURLUPART_URL, config->ipfs_gateway,
+                    CURLU_GUESS_SCHEME)) {
       gateway = strdup(config->ipfs_gateway);
       if(!gateway) {
         result = CURLE_URL_MALFORMAT;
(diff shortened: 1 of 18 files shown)


src/tool_operate.c
```diff
@@ -697,45 +697,59 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
+{
+  if(*input && **input) {
+    size_t len = strlen(*input);
+    if(((*input)[len - 1] != val_to_ensure)) {
+      char *composed_input = aprintf("%s%c", *input, val_to_ensure);
+      if(composed_input) {
+        Curl_safefree(*input);
+        *input = aprintf("%s", composed_input);
+        Curl_safefree(composed_input);
+      }
```
Stebalien, line 711:
```suggestion
      char *composed_input = aprintf("%s%c", *input, val_to_ensure);
      Curl_safefree(*input);
      *input = composed_input;
```

(should be equivalent?)

Or even:

```c
      *input = Curl_saferealloc(*input, len+2); // needs some casting
      if(*input) {
         input[len] = val_to_ensure;
         input[len+1] = 0;
      }
```

But, I'm not sure if that violates some rule of curl's coding style.

src/tool_operate.c
```diff
@@ -751,10 +765,9 @@ static char *ipfs_gateway(void)
   if(gateway_file) {
     char *buf = NULL;
 
-    if((PARAM_OK == file2string(&buf, gateway_file)) && buf && *buf) {
-      bool add_slash = (buf[strlen(buf) - 1] != '/');
-      gateway = aprintf("%s%s", buf, (add_slash) ? "/" : "");
-    }
+    if((PARAM_OK == file2string(&buf, gateway_file)) && buf && *buf)
+      gateway = aprintf("%s", buf);
```
Stebalien, line 769: seems like a useless copy?

Stebalien (commented):
Basic feedback but I'm not familiar with curl's coding guidelines.

src/tool_operate.c
```diff
@@ -697,45 +697,59 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
+/* helper function to ensure input ends in val_to_ensure */
+static void ensure_trailing(char **input, const char val_to_ensure)
+{
+  if(*input && **input) {
+    size_t len = strlen(*input);
+    if(((*input)[len - 1] != val_to_ensure)) {
+      char *composed_input = aprintf("%s%c", *input, val_to_ensure);
+      if(composed_input) {
+        Curl_safefree(*input);
+        *input = aprintf("%s", composed_input);
```
bagder, line 709: Save a useless copy:

```suggestion
        *input = composed_input;
```

src/tool_operate.c
```diff
@@ -697,45 +697,59 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
 }
 
+/* helper function to ensure input ends in val_to_ensure */
+static void ensure_trailing(char **input, const char val_to_ensure)
+{
+  if(*input && **input) {
+    size_t len = strlen(*input);
+    if(((*input)[len - 1] != val_to_ensure)) {
+      char *composed_input = aprintf("%s%c", *input, val_to_ensure);
+      if(composed_input) {
```
bagder, line 707: When this `aprintf()` call fails, shouldn't `*input` be freed and pointed to `""` or NULL or something?

src/tool_operate.c
```diff
@@ -697,45 +697,59 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
   /* Gateway is found from environment variable. */
   if(gateway && *gateway) {
-    char *composed_gateway = NULL;
-    bool add_slash = (gateway[strlen(gateway) - 1] != '/');
-    composed_gateway = aprintf("%s%s", gateway, (add_slash) ? "/" : "");
-    if(composed_gateway) {
-      gateway = aprintf("%s", composed_gateway);
-      Curl_safefree(composed_gateway);
-    }
+    ensure_trailing(&gateway, '/');
```
bagder, line 727: If `gateway` is NULL when this function returns, should it still return that NULL here or continue? And if it returns NULL here, should it not provide an error message to help the user why it stops?

src/tool_operate.c
markg85: That does mean every usage of this function needs to be checked `if(!ensure_trailing(..)) // error handling`.
I'll turn it into a function returning a CURLcode and checking for that on each invocation of this function.

src/tool_operate.c
markg85: Fixed by restructuring it with the idea below from Stebalien.

src/tool_operate.c
markg85: I think that's fine, i see more of those constructions in curl.
Thank you for this hint!  This idea works just fine and saves a temp copy!

src/tool_operate.c
markg85: That's not entirely the case...

It gets the file content and checks it for non-null. If that is non-null than `gateway` is filled with it's value.
If the file would exist but is empty, this line would make `gateway` a non-null but empty variable.

That being said, i will restructure the checks to after the call which would clean it up nicely too.

markg85:
@bagder 

```
[100%] Linking C executable curl
/usr/bin/ld: CMakeFiles/curl.dir/tool_operate.c.o: in function `ensure_trailing':
tool_operate.c:(.text+0x1384): undefined reference to `Curl_saferealloc'
gmake[2]: *** [src/CMakeFiles/curl.dir/build.make:873: src/curl] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:526: src/CMakeFiles/curl.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```
Any idea why many compilers can't seem to find `Curl_saferealloc`? I did add the header for it `strdup.h`. Is this function exclusive to libcurl and can't be used in the tool?

For reference, on my local pc (gcc (GCC) 13.2.1 20230801) it works just fine.

bagder:
> I'll just use the normal realloc instead.

I recommend using a dynbuf instead. Because it is less easy to mess up.

src/tool_operate.c
Stebalien: No, but copying `buf` to a new `gateway` buffer with `aprintf` is definitely a useless copy.

src/tool_operate.c
```diff
@@ -824,32 +878,95 @@ static CURLcode ipfs_url_rewrite(CURLU *uh, const char *protocol, char **url,
...
+
+  if(curl_url_get(gatewaysurl, CURLUPART_PATH,
+                  &gatewaypath, CURLU_URLDECODE) != CURLUE_OK) {
+    /* we need it if it's there, no error if it's not */
+  }
+
+  /* get the path from user input */
+  if(curl_url_get(uh, CURLUPART_PATH,
+                  &inputpath, CURLU_URLDECODE) != CURLUE_OK) {
+    inputpath = aprintf("%s", "");
```
bagder, line 925:
```suggestion
    inputpath = strdup("");
    if(!inputpath)
       goto clean;
```

src/tool_operate.c
```diff
@@ -824,32 +878,95 @@ static CURLcode ipfs_url_rewrite(CURLU *uh, const char *protocol, char **url,
...
+  if(curl_url_set(uh, CURLUPART_PORT,
+                  gatewayport, CURLU_URLENCODE) != CURLUE_OK) {
+    goto clean;
+  }
+
+  /* if the input path is just a slash, clear it */
+  if(inputpath && *inputpath && strlen(inputpath) == 1) {
+    if(*inputpath == '/') {
+      Curl_safefree(inputpath);
+      inputpath = aprintf("%s", "");
```
bagder, line 948:
```suggestion
      inputpath = strdup("");
      if(!inputpath)
         goto clean;
```

tests/data/DISABLED
```diff
@@ -70,9 +70,19 @@
 266
 579
 587
+# most of the IPFS tests, hyper doesn't like them
```
bagder, line 73: Why doesn't hyper like the IPFS? This seems like something we should fix, not hide.

src/tool_operate.c
```diff
@@ -697,45 +697,75 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
+      curlx_dyn_init(&dyn, len + 2);
+
+      if(curlx_dyn_add(&dyn, *input)) {
+        Curl_safefree(*input);
+        return CURLE_OUT_OF_MEMORY;
+      }
+
+      Curl_safefree(*input);
+
+      if(curlx_dyn_addf(&dyn, "%c", val_to_ensure)) {
```
bagder, line 716: Leaner:

```suggestion
      if(curlx_dyn_addn(&dyn, &val_to_ensure, 1)) {      
```

src/tool_operate.c
```diff
@@ -697,45 +697,75 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
+/* helper function to ensure input ends in val_to_ensure */
+static CURLcode ensure_trailing(char **input, const char val_to_ensure)
+{
+  if(*input && **input) {
+    size_t len = strlen(*input);
+    if(((*input)[len - 1] != val_to_ensure)) {
+      struct curlx_dynbuf dyn;
+      curlx_dyn_init(&dyn, len + 2);
+
+      if(curlx_dyn_add(&dyn, *input)) {
```
bagder, line 709: Save a strlen:

```suggestion
      if(curlx_dyn_addn(&dyn, *input, len)) {
```

tests/data/DISABLED
markg85: It's the CRLF vs LF line endings. I just can't get those to work.
If you have more luck, please do. Last time i spend days trying to make hyper like it, nothing worked.

src/tool_operate.c
```diff
@@ -749,24 +779,36 @@ static char *ipfs_gateway(void)
...
-      gateway = aprintf("%s%s", buf, (add_slash) ? "/" : "");
+    int c;
+    struct curlx_dynbuf dyn;
+    curlx_dyn_init(&dyn, INT_MAX);
+
+    /* get the first line of the gateway file, ignore the rest */
+    while((c = getc(gateway_file)) != EOF && c != '\n' && c != '\r') {
+      if(curlx_dyn_addf(&dyn, "%c", (char) c)) {
+        /* reset the buffer, 0 length is going to trigger to cleanup below */
+        curlx_dyn_reset(&dyn);
```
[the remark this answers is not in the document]
markg85: Nice! Fixed.

tests/data/DISABLED
markg85: Hyper is fixed. I just needed to add `nocheck="yes"` to all tests that contain server data. I somehow completely missed that before. No IPFS tests are disabled for hyper now. Yay :)

src/tool_operate.c
```diff
@@ -749,24 +778,33 @@ static char *ipfs_gateway(void)
...
       fclose(gateway_file);
 
+    if(curlx_dyn_len(&dyn) > 0)
+      gateway = curlx_dyn_ptr(&dyn);
+
+    if(gateway)
+      ensure_trailing(&gateway, '/');
+
     if(!gateway) {
       Curl_safefree(gateway);
```
bagder, line 801: if gateway is already NULL, there is not much point in freeing it...

markg85: Nice! Removed.

markg85:
Hi @bagder,

Just flattened it all to one commit.
It's probably ready to get merged now.

If it's ok with you, i'd prefer to wait merging till a second functional-review has been done by @lidel.
Is merging on Monday still fine to make it in the 8.5.0 release?

bagder:
> Is merging on Monday still fine to make it in the 8.5.0 release?

This is a bugfix, we can merge such up until December 6...

src/tool_operate.c
```diff
@@ -697,45 +697,74 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
   /* Gateway is found from environment variable. */
-  if(gateway && *gateway) {
-    char *composed_gateway = NULL;
-    bool add_slash = (gateway[strlen(gateway) - 1] != '/');
-    composed_gateway = aprintf("%s%s", gateway, (add_slash) ? "/" : "");
-    if(composed_gateway) {
-      gateway = aprintf("%s", composed_gateway);
-      Curl_safefree(composed_gateway);
+  if(gateway) {
+    if(ensure_trailing(&gateway, '/') != CURLE_OK) {
```
bagder, line 737: We usually go with just `if(ensure_trailing(&gateway, '/')) {` when checking CURLcode functions since we know `CURLE_OK` is zero. That saves us from comparing with `CURLE_OK` in checks. It's not a *must*, but it is a standard pattern in curl code. You could simplify several checks here like this.

src/tool_operate.c
```diff
@@ -808,7 +851,13 @@ static CURLcode ipfs_url_rewrite(CURLU *uh, const char *protocol, char **url,
...
-    if(curl_url_set(ipfsurl, CURLUPART_URL, config->ipfs_gateway,
+    /* ensure the gateway ends in a trailing / */
+    if(ensure_trailing(&config->ipfs_gateway, '/') != CURLE_OK) {
+      result = CURLE_OUT_OF_MEMORY;
+      goto clean;
+    }
+
+    if(curl_url_set(gatewaysurl, CURLUPART_URL, config->ipfs_gateway,
                     CURLU_GUESS_SCHEME)
                     == CURLUE_OK) {
```
bagder, line 862: *I* would write these conditions like this to make them slightly shorter and in my view easier to read:

```suggestion
    if(!curl_url_set(gatewaysurl, CURLUPART_URL, config->ipfs_gateway,
                     CURLU_GUESS_SCHEME)) {
```

src/tool_operate.c
```diff
@@ -824,32 +873,92 @@ static CURLcode ipfs_url_rewrite(CURLU *uh, const char *protocol, char **url,
...
+    goto clean;
+  }
+
+  /* if the input path is just a slash, clear it */
+  if(inputpath && *inputpath && strlen(inputpath) == 1) {
+    if(*inputpath == '/') {
+      Curl_safefree(inputpath);
+      inputpath = strdup("");
+      if(!inputpath)
+        goto clean;
```
bagder, line 940: Since the memory is already allocated and a single byte that you want to not be there. Maybe just overwrite it?

```suggestion
      *inputpath = '\0';
```

src/tool_operate.c
markg85: This + comment below. Cleans it up nicely! Thanx :)

For reference. Some jobs back i was "forced" into writing checks verbose. The reasoning there was some people didn't like to have to pay attention to the single exclamation mark and wanted a policy where the `!` at the beginning of an if-statement wasn't allowed... That - now a habit of sorts - kinda stayed with me. Not that verbose is wrong, but short and still readable is better imho.

I like your style better :)

markg85:
Hi @bagder,

I just updated the patch with a little off-site review feedback. Only thing changed is the CID in the testcases (from `Qm...`, base58) to be of the "v1"  (`bafyt...` base32) format. It's only to discourage the `Qm...` format and not keep copying it over in new testcases.

Assuming it passes all the currently running tests, feel free to merge it!

bagder (approved):

mkauf:
There are failed tests for this change in the autobuilds. E. g. test 724: "Malformed URL". https://curl.se/dev/log.cgi?id=20231107030851-1844341#prob1 or https://curl.se/dev/log.cgi?id=20231107033909-1867421#prob1

bagder:
I did some further tidying up in 01d9b8bc7ceb56dc116 so it may things it a little.

markg85:
I'm happy some testcases fail, means I've added enough cases to catch edits that might change the behavior :)

mkauf:
@markg85 / @bagder 

* There is an endianness bug. Proposed bugfix:
```
diff --git a/src/tool_ipfs.c b/src/tool_ipfs.c
index 271fff2cd..435d1697c 100644
--- a/src/tool_ipfs.c
+++ b/src/tool_ipfs.c
@@ -104,7 +104,8 @@ static char *ipfs_gateway(void)

     /* get the first line of the gateway file, ignore the rest */
     while((c = getc(gateway_file)) != EOF && c != '\n' && c != '\r') {
-      if(curlx_dyn_addn(&dyn, &c, 1))
+      char c_char = (char)c;
+      if(curlx_dyn_addn(&dyn, &c_char, 1))
         goto fail;
     }
```

* Some tests set the environment variable `IPFS_DATA`. But this variable is not checked at all in the curl tool code. Probably you wanted to set the environment variable `IPFS_PATH`.

* Typo in the tests: Change "traling" to "trailing"

* Wrong link in the documentation: Change "#Automatic-gateway-detection" to "#automatic-gateway-detection" in `docs/IPFS.md`

markg85:
Hi @mkauf 

Thank you for that review, that's much appreciated!

> There is an endianness bug. Proposed bugfix:
I'll fix it with that suggestion.
Meanwhile, could you explain the reasoning why adding an extra temp variable with `c` casts as `char`  magically fixes endianess? Internally - in curlx_dyn_addn - the input is casted to `unsigned char *`. I'd assume that cast + the size of 1 (1 byte) would all be enough to catch endian differences, no?

> Some tests set the environment variable IPFS_DATA. But this variable is not checked at all in the curl tool code. Probably you wanted to set the environment variable IPFS_PATH.
Very good catch! While `IPFS_DATA` [does exist](https://docs.ipfs.tech/install/run-ipfs-inside-docker/), the intended use here is `IPFS_PATH`. Will fix the testcases.
The real thing that is being tested there though (tests 736, 737 and 738) isn't even the environment variable. It's making sure IPFS knows where to look for it's config file. Which by default is in $HOME/.ipfs. In other terms, tests 736 and 738 aren't testing properly and 737 just tests what's the default already.

> Typo in the tests: Change "traling" to "trailing"
Fixed.

> Wrong link in the documentation: Change "#Automatic-gateway-detection" to "#automatic-gateway-detection" in docs/IPFS.md
Fixed.

PR incoming.

mkauf:
@markg85

>  I'd assume that cast + the size of 1 (1 byte) would all be enough to catch endian differences, no?

No, because the pointer still points to an `int`. It still points to the same memory address as before.

These answers explain it in more detail:
* https://stackoverflow.com/a/21975205
* https://stackoverflow.com/a/21975272

PR #12153, opened 2023-10-19
openssl: use OPENSSL_INIT_NO_ATEXIT

rcmcdonald91:
https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_INIT_set_config_file_flags.html
> OPENSSL_INIT_NO_ATEXIT
> By default OpenSSL will attempt to clean itself up when the process exits via an “atexit” handler. Using this option suppresses that behaviour. This means that the application will have to clean up OpenSSL explicitly using OPENSSL_cleanup().

https://developers.redhat.com/articles/2022/10/31/best-practices-application-shutdown-openssl#recommended_initialization_and_shutdown_procedure
> Recommended initialization and shutdown procedure
> The best option is a properly serialized deinitialization, during which the atexit hooks are executed in the correct order, all the application threads terminate except the main one, etc. This sequence requires you to be thorough in your coding, but it's worth the effort.
> 
> If your application has to shut down manually, OpenSSL authors suggest a procedure with two parts.
> 
> First, the application should use the OPENSSL_INIT_NO_ATEXIT option in the OPENSSL_init_crypto call at OpenSSL initialization, to avoid implicit installation of the OpenSSL cleanup handler. After that, when the application terminates and no further OpenSSL functions will be called, invoke OPENSSL_cleanup explicitly to free resources on shutdown.
> 
> In conclusion, a manual shutdown option is useful for some libraries, but a predictable order of initialization and deinitialization is preferable.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -1705,6 +1705,9 @@ static int ossl_init(void)
 #else
     OPENSSL_INIT_LOAD_CONFIG |
 #endif
+#ifdef OPENSSL_INIT_NO_ATEXIT
+    OPENSSL_INIT_NO_ATEXIT |
+#endif
     0;
   OPENSSL_init_ssl(flags, NULL);
 #else
@@ -1748,6 +1751,7 @@ static void ossl_cleanup(void)
   !defined(LIBRESSL_VERSION_NUMBER)
   /* OpenSSL 1.1 deprecates all these cleanup functions and
      turns them into no-ops in OpenSSL 1.0 compatibility mode */
+  OPENSSL_cleanup();
 #else
   /* Free ciphers and digests lists */
   EVP_cleanup();


dfandrich:
Can you provide a bit more context? This sounds like a backward compatibility
break, so it might need to be enabled by a bit in curl_global_init.

rcmcdonald91:
> Can you provide a bit more context? This sounds like a backward compatibility break, so it might need to be enabled by a bit in curl_global_init.

Over in FreeBSD land, the pkg package manager now uses libcurl. I am seeing a segv identical to https://github.com/openssl/openssl/issues/17312#issuecomment-1313950807. The conclusion there it seems was that consumers should explicitly opt-out of OpenSSL default deinitialization especially when performing it manually already. libcurl is not following the OpenSSL documentation in this regard.

rcmcdonald91:
Here is another real life example https://github.com/nmap/nmap/issues/2482

jay:
I don't see how this is our problem. It looks to me like the atexit registered OpenSSL_cleanup is called, then some other library's atexit registered function is called, but that function calls into OpenSSL which has already cleaned up.

bagder:
Here we go again. OpenSSL init and cleanup clearly is impossible to get right for all situations. We've adjusted it every once in a while ever since we added support for it, in early 1999.

What I'm curious about is *why now?* We have never set this flag and presumably there already exist a few applications how there with multiple OpenSSL users within the same process and they have not reported problems with that. (and as @jay says, it would seem like an OpenSSL problem then and not the users' of it)

rcmcdonald91:
> I don't see how this is our problem. It looks to me like the atexit registered OpenSSL_cleanup is called, then some other library's atexit registered function is called, but that function calls into OpenSSL which has already cleaned up.

Because that second and thus incorrect call to OpenSSL_cleanup is from calling say curl_multi_cleanup.

if you break on context_deinit (which is openssl code) and curl_multi_cleanup , you end up hitting context_deinit first (coming from the OPENSSL_cleanup atexit call):
```
Thread 1 hit Breakpoint 2, context_deinit (ctx=0x800c1c9c0)
    at /usr/src/crypto/openssl/crypto/context.c:122
122	   if (ctx == NULL)
(gdb) bt
#0  context_deinit (ctx=0x800c1c9c0)
    at /usr/src/crypto/openssl/crypto/context.c:122
#1  0x000000080086c060 in ossl_lib_ctx_default_deinit ()
    at /usr/src/crypto/openssl/crypto/context.c:160
#2  0x000000080087188a in OPENSSL_cleanup ()
    at /usr/src/crypto/openssl/crypto/init.c:427
#3  0x0000000801011e8b in __cxa_finalize (dso=dso@entry=0x0)
    at /usr/src/lib/libc/stdlib/atexit.c:238
#4  0x00000008010123b1 in exit (status=0)
    at /usr/src/lib/libc/stdlib/exit.c:72
#5  0x0000000800f351d1 in __libc_start1 (argc=3, argv=0x7fffffffea58, 
    env=0x7fffffffea78, cleanup=<optimized out>, mainX=0x2e1280 <main>)
    at /usr/src/lib/libc/csu/libc_start1.c:157
#6  0x00000000002d3320 in _start ()
    at /usr/src/lib/csu/amd64/crt1_s.S:83i
```


and then later on when pkg_shutdown is called (another atexit registered by pkg), you end up calling curl_multi_cleanup:
```
(gdb) c
Continuing.

Thread 1 hit Breakpoint 1, curl_multi_cleanup (multi=0x8016ec000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/multi.c:2791
2791	/wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/multi.c: No such file or directory.
(gdb) bt
#0  curl_multi_cleanup (multi=0x8016ec000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/multi.c:2791
#1  0x00000000005a7de8 in curl_cleanup (repo=0x80164f000)
    at fetch_libcurl.c:520
#2  0x000000000055e1b4 in pkg_repo_free (r=0x80164f000)
    at pkg_config.c:1490
#3  0x000000000055e08f in pkg_shutdown () at pkg_config.c:1506
#4  0x0000000801011e8b in __cxa_finalize (dso=dso@entry=0x0)
    at /usr/src/lib/libc/stdlib/atexit.c:238
#5  0x00000008010123b1 in exit (status=0)
    at /usr/src/lib/libc/stdlib/exit.c:72
#6  0x0000000800f351d1 in __libc_start1 (argc=3, argv=0x7fffffffea58, 
    env=0x7fffffffea78, cleanup=<optimized out>, mainX=0x2e1280 <main>)
    at /usr/src/lib/libc/csu/libc_start1.c:157
#7  0x00000000002d3320 in _start ()
    at /usr/src/lib/csu/amd64/crt1_s.S:83
```

which then puts you down another path that leads to calling OPENSSL_cleanup again.
```
(gdb) c
Continuing.

Thread 1 received signal SIGSEGV, Segmentation fault.
Address not mapped to object.
rwlock_rdlock_common (rwlock=0x0, abstime=0x0)
    at /usr/src/lib/libthr/thread/thr_rwlock.c:177
177		ret = check_and_init_rwlock(rwlock, &prwlock);
(gdb) bt
#0  rwlock_rdlock_common (rwlock=0x0, abstime=0x0)
    at /usr/src/lib/libthr/thread/thr_rwlock.c:177
#1  0x0000000800886425 in CRYPTO_THREAD_read_lock (lock=0x0)
    at /usr/src/crypto/openssl/crypto/threads_pthread.c:95
#2  0x000000080086c83c in ossl_lib_ctx_get_data (ctx=0x800c1c9c0, 
    index=1, meth=0x800bbfb90)
    at /usr/src/crypto/openssl/crypto/context.c:377
#3  0x0000000800880819 in get_provider_store (libctx=0x800c1c9c0)
    at /usr/src/crypto/openssl/crypto/provider_core.c:339
#4  0x0000000800884591 in ossl_provider_deregister_child_cb (
    handle=0x80164f700)
    at /usr/src/crypto/openssl/crypto/provider_core.c:1806
#5  0x000000080087f626 in ossl_provider_deinit_child (ctx=0x801639200)
    at /usr/src/crypto/openssl/crypto/provider_child.c:281
#6  0x000000080086c474 in OSSL_LIB_CTX_free (ctx=0x801639200)
    at /usr/src/crypto/openssl/crypto/context.c:248
#7  0x000000080134116d in legacy_teardown (provctx=0x8016e8e60)
    at /usr/src/crypto/openssl/providers/legacyprov.c:168
#8  0x0000000800881868 in ossl_provider_teardown (prov=0x80164f700)
    at /usr/src/crypto/openssl/crypto/provider_core.c:1558
#9  0x0000000800880e5c in ossl_provider_free (prov=0x80164f700)
    at /usr/src/crypto/openssl/crypto/provider_core.c:696
#10 0x0000000800a21fb2 in evp_cipher_free_int (cipher=0x801715700)
    at /usr/src/crypto/openssl/crypto/evp/evp_enc.c:1651
#11 0x0000000800a1d44f in EVP_CIPHER_free (cipher=0x801715700)
    at /usr/src/crypto/openssl/crypto/evp/evp_enc.c:1666
#12 0x000000080066b67a in ssl_evp_cipher_free (cipher=0x801715700)
    at /usr/src/crypto/openssl/ssl/ssl_lib.c:5967
#13 SSL_CTX_free (a=0x801612e00)
    at /usr/src/crypto/openssl/ssl/ssl_lib.c:3492
#14 0x000000000051e66b in ossl_close (cf=0x8016206f0, data=0x801662000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/vtls/openssl.c:1927
#15 0x000000000051a0e0 in cf_close (cf=0x8016206f0, data=0x801662000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/vtls/vtls.c:1424
#16 0x00000000005186a8 in ssl_cf_close (cf=0x8016206f0, 
    data=0x801662000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/vtls/vtls.c:1496
#17 0x00000000004c1e93 in cf_setup_close (cf=0x801620780, 
    data=0x801662000)
--Type <RET> for more, q to quit, c to continue without paging--
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/connect.c:1299
#18 0x00000000004b7e7a in cf_hc_close (cf=0x801620630, data=0x801662000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/cf-https-connect.c:435
#19 0x00000000004bdfcc in Curl_conn_close (data=0x801662000, index=0)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/cfilters.c:164
#20 0x0000000000508ab9 in conn_shutdown (data=0x801662000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/url.c:697
#21 0x0000000000508a20 in Curl_disconnect (data=0x801662000, 
    conn=0x80163c800, dead_connection=false)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/url.c:809
#22 0x00000000004c0520 in Curl_conncache_close_all_connections (
    connc=0x8016ec110)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/conncache.c:533
#23 0x00000000004eafc2 in curl_multi_cleanup (multi=0x8016ec000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/multi.c:2826
#24 0x00000000005a7de8 in curl_cleanup (repo=0x80164f000)
    at fetch_libcurl.c:520
#25 0x000000000055e1b4 in pkg_repo_free (r=0x80164f000)
    at pkg_config.c:1490
#26 0x000000000055e08f in pkg_shutdown () at pkg_config.c:1506
#27 0x0000000801011e8b in __cxa_finalize (dso=dso@entry=0x0)
    at /usr/src/lib/libc/stdlib/atexit.c:238
#28 0x00000008010123b1 in exit (status=0)
    at /usr/src/lib/libc/stdlib/exit.c:72
#29 0x0000000800f351d1 in __libc_start1 (argc=3, argv=0x7fffffffea58, 
    env=0x7fffffffea78, cleanup=<optimized out>, mainX=0x2e1280 <main>)
    at /usr/src/lib/libc/csu/libc_start1.c:157
#30 0x00000000002d3320 in _start ()
    at /usr/src/lib/csu/amd64/crt1_s.S:83
```

bagder:
> Because that second and thus incorrect call to OpenSSL_cleanup is from calling say curl_multi_cleanup.

I don't see that in your stack trace. I see libcurl calling `SSL_CTX_free()`.

rsbeckerca:
Question on this case: Is there a mixture of static and dynamic linkage for OpenSSL? We have seen issues in OpenSSL when that is done. `atexit()` semantics with DLLs can vary from one platform to another, and static variables in a DLL do not always share addresses with statically linked modules, so anything used in `atexit()` code as semaphores become non-portable in such a mix. Not sure it applies here, but I have seen similar evidence elsewhere.

bagder:
But why is `OPENSSL_cleanup` called as a result of `SSL_CTX_free`? Surely that's super strange. libcurl certainly intends to continue using OpenSSL even after that...

rcmcdonald91:
I don't see that it is, `OpenSSL_cleanup` is being called first at exit and THEN later on during `curl_multi_cleanup` we land on a path that takes us through `SSL_CTX_free`. Once `OpenSSL_cleanup` is called, any future calls are UB. As far as I understand it at least.

rsbeckerca:
> > > Because that second and thus incorrect call to OpenSSL_cleanup is from calling say curl_multi_cleanup.
> > 
> > 
> > I don't see that in your stack trace. I see libcurl calling `SSL_CTX_free()`.
> 
> Ah yes thanks for that observation, but that is more to the crux of the problem...there are two paths to teardown OpenSSL.
> 
> `OpenSSL_cleanup` is guarded by the `stopped` variable that ensures it can only ever run once, either from the atexit handler or explicitly: https://github.com/openssl/openssl/blob/098f27f9ef8be2a418f76896ee3c824e8709fcf7/crypto/init.c#L364C1-L367C17
> 
> The path through `SSL_CTX_free` has no such safety belt.

Actually, it is the stopped variable that is the problem when simultaneously statically and dynamically linked, because stopped has two incarnations on some platforms (not all), so the value of the semaphore during atexit processing is indeterminate.

rcmcdonald91:
> > I don't see that it is, `OpenSSL_cleanup` is being called first at exit and THEN later on during `curl_multi_cleanup` we land on a path that takes us through `SSL_CTX_free`. Once `OpenSSL_cleanup` is called, any future calls are UB.
> 
> If `OpenSSL_cleanup` is being called at **exit**, how is `curl_multi_cleanup` being called after that?

Also atexit:

```(gdb) c
Continuing.

Thread 1 received signal SIGSEGV, Segmentation fault.
Address not mapped to object.
rwlock_rdlock_common (rwlock=0x0, abstime=0x0)
    at /usr/src/lib/libthr/thread/thr_rwlock.c:177
177		ret = check_and_init_rwlock(rwlock, &prwlock);
(gdb) bt
#0  rwlock_rdlock_common (rwlock=0x0, abstime=0x0)
    at /usr/src/lib/libthr/thread/thr_rwlock.c:177
#1  0x0000000800886425 in CRYPTO_THREAD_read_lock (lock=0x0)
    at /usr/src/crypto/openssl/crypto/threads_pthread.c:95
#2  0x000000080086c83c in ossl_lib_ctx_get_data (ctx=0x800c1c9c0, 
    index=1, meth=0x800bbfb90)
    at /usr/src/crypto/openssl/crypto/context.c:377
#3  0x0000000800880819 in get_provider_store (libctx=0x800c1c9c0)
    at /usr/src/crypto/openssl/crypto/provider_core.c:339
#4  0x0000000800884591 in ossl_provider_deregister_child_cb (
    handle=0x80164f700)
    at /usr/src/crypto/openssl/crypto/provider_core.c:1806
#5  0x000000080087f626 in ossl_provider_deinit_child (ctx=0x801639200)
    at /usr/src/crypto/openssl/crypto/provider_child.c:281
#6  0x000000080086c474 in OSSL_LIB_CTX_free (ctx=0x801639200)
    at /usr/src/crypto/openssl/crypto/context.c:248
#7  0x000000080134116d in legacy_teardown (provctx=0x8016e8e60)
    at /usr/src/crypto/openssl/providers/legacyprov.c:168
#8  0x0000000800881868 in ossl_provider_teardown (prov=0x80164f700)
    at /usr/src/crypto/openssl/crypto/provider_core.c:1558
#9  0x0000000800880e5c in ossl_provider_free (prov=0x80164f700)
    at /usr/src/crypto/openssl/crypto/provider_core.c:696
#10 0x0000000800a21fb2 in evp_cipher_free_int (cipher=0x801715700)
    at /usr/src/crypto/openssl/crypto/evp/evp_enc.c:1651
#11 0x0000000800a1d44f in EVP_CIPHER_free (cipher=0x801715700)
    at /usr/src/crypto/openssl/crypto/evp/evp_enc.c:1666
#12 0x000000080066b67a in ssl_evp_cipher_free (cipher=0x801715700)
    at /usr/src/crypto/openssl/ssl/ssl_lib.c:5967
#13 SSL_CTX_free (a=0x801612e00)
...
>>>> #23 0x00000000004eafc2 in curl_multi_cleanup (multi=0x8016ec000)
    at /wrkdirs/usr/ports/ports-mgmt/pkg/work/pkg-1.20.8/external/curl/lib/multi.c:2826
#24 0x00000000005a7de8 in curl_cleanup (repo=0x80164f000)
    at fetch_libcurl.c:520
#25 0x000000000055e1b4 in pkg_repo_free (r=0x80164f000)
    at pkg_config.c:1490
>>>> #26 0x000000000055e08f in pkg_shutdown () at pkg_config.c:1506
>>>> #27 0x0000000801011e8b in __cxa_finalize (dso=dso@entry=0x0)
    at /usr/src/lib/libc/stdlib/atexit.c:238**
#28 0x00000008010123b1 in exit (status=0)
    at /usr/src/lib/libc/stdlib/exit.c:72
#29 0x0000000800f351d1 in __libc_start1 (argc=3, argv=0x7fffffffea58, 
    env=0x7fffffffea78, cleanup=<optimized out>, mainX=0x2e1280 <main>)
    at /usr/src/lib/libc/csu/libc_start1.c:157
#30 0x00000000002d3320 in _start ()
    at /usr/src/lib/csu/amd64/crt1_s.S:83
```

rsbeckerca:
> > I don't see that it is, `OpenSSL_cleanup` is being called first at exit and THEN later on during `curl_multi_cleanup` we land on a path that takes us through `SSL_CTX_free`. Once `OpenSSL_cleanup` is called, any future calls are UB.
> 
> If `OpenSSL_cleanup` is being called at **exit**, how is `curl_multi_cleanup` being called after that?

Exit processing can occur either when the DLL unloads (`dlclose()`) or when the program exits, or both. The `dlclose()` sometimes precedes program at exit processing.

bagder:
> Exit processing can occur

I just don't think I have ever seen someone clean up libcurl in an atexit call before so I just didn't think it was done on purpose...

bagder:
I still think this is an OpenSSL flaw of course, but I guess that doesn't make anyone happier.

bagder:
Doing the libcurl cleanup in the atexit will of course end up in a race with OpenSSL so if OpenSSL cleans up first, then libcurl will/might call OpenSSL after its cleanup and that is what we see here. To fix that, we probably have to merge this PR.

As I said before, the init and cleanup of OpenSSL is an open wound that we just continuously keep poking at and it scares me every time we return to this area and change the sequences because the (unexpected and not at all understood) consequences tend to hide and bounce back up only months or years later when someone else runs into another edge case...

rcmcdonald91:
> I don't think I have ever seen someone clean up libcurl in an atexit call before so I just didn't think it was done on purpose...

Utilizing the OpenSSL atexit cleanup process is fine for say applications that directly consume OpenSSL, and where the atexit ordering is (more) well-defined and well-understood. I think the problem here is that libcurl is library for other applications and so you introduce the opportunity for races.

Generally speaking, library consumers of OpenSSL (like libcurl) should probably avoid using the atexit OpenSSL_cleanup and handle this explicitly, at least then you can be more sure about the order of operations.

> Doing the libcurl cleanup in the atexit will of course end up in a race with OpenSSL so if OpenSSL cleans up first, then libcurl will/might call OpenSSL after its cleanup and that is what we see here. To fix that, we probably have to merge this PR.
> 
> As I said before, the init and cleanup of OpenSSL is an open wound that we just continuously keep poking at and it scares me every time we return to this area and change the sequences because the (unexpected and not at all understood) consequences tend to hide and bounce back up only months or years later when someone else runs into another edge case...

Yea I completely understand that concern.

piru:
In general use of `atexit` handlers in library is a bad idea to begin with, and here is a perfect example why.

Some earlier discussion about the problem:
- https://github.com/openssl/openssl/issues/9581
- https://github.com/openssl/openssl/issues/17469

rsbeckerca:
If this is actually resulting from `atexit` processing in OpenSSL, and dynamic + static linking is not used, an issue should probably be opened at OpenSSL.

jay:
> and then later on when pkg_shutdown is called (another atexit registered by pkg), you end up calling curl_multi_cleanup:

I see, so you are saying that library pkg's atexit function cleans up libcurl and that happens after OpenSSL_cleanup. If library pkg called curl_global_init to initialize it then I would think at exit that OpenSSL_cleanup would be called after that library pkg cleaned up libcurl. Instead, maybe the first implict/explicit call to OpenSSL init function happened later so its cleanup was added to the atexit list later, and since atexit list is LIFO it called OpenSSL_cleanup before library pkg cleanup.

rsbeckerca:
> > and then later on when pkg_shutdown is called (another atexit registered by pkg), you end up calling curl_multi_cleanup:
> 
> I see, so you are saying that library pkg's atexit function cleans up libcurl and that happens after OpenSSL_cleanup. If library pkg called curl_global_init to initialize it then I would think at exit that OpenSSL_cleanup would be called after that library pkg cleaned up libcurl. Instead, maybe the first implict/explicit call to OpenSSL init function happened later so its cleanup was added to the atexit list later, and since atexit list is LIFO it called OpenSSL_cleanup before library pkg cleanup.

`atexit()` is LIFO from a certain point of view. When `dlclose()` executes, any `atexit()` done by the DLL is executed LIFO, individually for each DLL being closed. The same for the main program. However, and this is the rub, the `atexit()` list is not guaranteed to be cleared by a `dlclose()`, so the main's `atexit()` list could reinvoke something registered by the DLL on the list that is gone - why using `atexit()` from a DLL is bad. POSIX is ambiguous (according to some platform developers when I opened cases). It is not globally LIFO

bagder:
A notable difference when using this flag is of course that valgrind detects huge amounts of memory "in use at exit", compared to nothing before:

~~~
$ valgrind  ./src/curl -sv https://curl.se -o /dev/null
....
==1347922== 
==1347922== HEAP SUMMARY:
==1347922==     in use at exit: 303,063 bytes in 6,229 blocks
==1347922==   total heap usage: 143,687 allocs, 137,458 frees, 6,929,858 bytes allocated
==1347922== 
==1347922== LEAK SUMMARY:
==1347922==    definitely lost: 0 bytes in 0 blocks
==1347922==    indirectly lost: 0 bytes in 0 blocks
==1347922==      possibly lost: 0 bytes in 0 blocks
==1347922==    still reachable: 303,063 bytes in 6,229 blocks
==1347922==         suppressed: 0 bytes in 0 blocks
==1347922== Rerun with --leak-check=full to see details of leaked memory
==1347922== 
==1347922== For lists of detected and suppressed errors, rerun with: -s
==1347922== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)

rcmcdonald91:
Here are my results:
```
$ uname -a
FreeBSD freebsd 15.0-CURRENT FreeBSD 15.0-CURRENT #24 main-n265934-7ca90db2f3ee-dirty: Tue Oct 17 11:47:39 EDT 2023     root@freebsd:/usr/obj/usr/src/amd64.amd64/sys/GENERIC amd64
```

before patch:
```
$ valgrind  curl -sv https://curl.se -o /dev/null
==80788== 
==80788== HEAP SUMMARY:
==80788==     in use at exit: 135,546 bytes in 13 blocks
==80788==   total heap usage: 135,462 allocs, 135,449 frees, 6,687,536 bytes allocated
==80788== 
==80788== LEAK SUMMARY:
==80788==    definitely lost: 0 bytes in 0 blocks
==80788==    indirectly lost: 0 bytes in 0 blocks
==80788==      possibly lost: 0 bytes in 0 blocks
==80788==    still reachable: 135,546 bytes in 13 blocks
==80788==         suppressed: 0 bytes in 0 blocks
==80788== Rerun with --leak-check=full to see details of leaked memory
==80788== 
==80788== For lists of detected and suppressed errors, rerun with: -s
==80788== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
after patch:
```
==80806== 
==80806== HEAP SUMMARY:
==80806==     in use at exit: 444,537 bytes in 6,513 blocks
==80806==   total heap usage: 135,459 allocs, 128,946 frees, 6,670,472 bytes allocated
==80806== 
==80806== LEAK SUMMARY:
==80806==    definitely lost: 0 bytes in 0 blocks
==80806==    indirectly lost: 0 bytes in 0 blocks
==80806==      possibly lost: 0 bytes in 0 blocks
==80806==    still reachable: 444,537 bytes in 6,513 blocks
==80806==         suppressed: 0 bytes in 0 blocks
==80806== Rerun with --leak-check=full to see details of leaked memory
==80806== 
==80806== For lists of detected and suppressed errors, rerun with: -s
==80806== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
This sort of makes sense if `OPENSSL_cleanup` is no longer be called at all.

bagder:
Setting this bit for OpenSSL init in libcurl still does not solve the problem completely since *every* individual OpenSSL-using library would need to do the same in order to completely fix this for any random application. This makes me convinced this remains an OpenSSL problem.

Is it not a better approach to do an OpenSSL init in the application before any other has a chance to do it, that sets this bit, thus making sure no other library gets to do it the wrong way first? Then it wouldn't matter how many other OpenSSL users there are nor how they they init OpenSSL.

Optionally, another work-around/fix could be to completely avoid atexit and leave that only for OpenSSL itself, as presumably it works better if they are left to their own devices.

fche:
Has it been explained somewhere why openssl cleanup of any sort needs to be done at exit?  What would happen if openssl_cleanup were a nop?  Or if it were not called at all at a process exit?

jay:
I'm still not convinced this is our problem

ngie-eign:
My group consumes the cURL and OpenSSL libraries and ran into this problem internally with one of our applications.
One interesting point looking at this is the fact that MD4 (which cURL uses) is supplied by the [legacy provider](https://docs.openssl.org/3.0/man7/OSSL_PROVIDER-legacy/) on 3.x.

It looks like some structures are allocated to which point to MD4* in vtls.c -- I wonder if the issue seen here is that cURL attempts to tear down resources in use in tls.c after the legacy provider--previously provided by libcrypto--is being torn down.

I'm guessing threads aren't necessarily at play because freebsd/pkg doesn't make use of threading directly (it makes use of threads indirectly in the libarchive compressors).

PR #12138, opened 2023-10-16
Chore: Typo fixed in FAQ

SusheelThapa:
This PR fix the typo in `FAQ` file

```pseudo
htpt > http
```

PR #12154, opened 2023-10-19
fixed a typo in FAQ

parikshit-adhikari:
in line number 757 of file: FAQ, typo has been fixed by changing 'htpt' to 'http'


bagder:
#12138 submitted the exact same thing.

If you read the text, you see that the typo is there on purpose to illustrate a point.

PR #12160, opened 2023-10-19
cmake: USE_QUICHE uses openssl_check_symbol_exists

cbodley:
the `CheckQuicSupportInOpenSSL` function was removed from `USE_NGTCP2` in dee310d54261f9a8416e87d50bccfe2cbe404949, but the `USE_QUICHE` section wasn't updated accordingly. this results in cmake errors when `USE_QUICHE` is enabled:
```
-- Found QUICHE: /path/to/quiche/target/release/libquiche.a
CMake Error at CMakeLists.txt:712 (CheckQuicSupportInOpenSSL):
  Unknown CMake command "CheckQuicSupportInOpenSSL".
```


cbodley:
cc @vszakats as this relates to https://github.com/curl/curl/pull/11555

vszakats:
@cbodley Thanks for reporting this. How could I miss that use!

Can you test if #12162 fixes the problem?

vszakats:
Ops, did not notice this is a PR, not a report, sorry.

My patch is a little different, by avoid duplicating the quic-checking logic and also allowing to pair quiche with wolfSSL.

cbodley:
thanks, closing in favor of https://github.com/curl/curl/pull/12162

PR #12162, opened 2023-10-19
cmake: fix OpenSSL quic detection in quiche builds

vszakats:
An orphan call to `CheckQuicSupportInOpenSSL()` remained after a recent
update when checking QUIC for quiche. Move back QUIC detection to
a function and fixup callers to use that. Also make sure that quiche
gets QUIC from BoringSSL, because it doesn't support other forks at this
time.

Regression from dee310d54261f9a8416e87d50bccfe2cbe404949 #11555

Reported-by: Casey Bodley <cbodley@redhat.com>
Fixes #12160
Closes #12162


CMakeLists.txt
```diff
@@ -650,6 +650,20 @@ macro(openssl_check_symbol_exists SYMBOL FILES VARIABLE)
   cmake_pop_check_state()
 endmacro()
 
+# Ensure that the OpenSSL fork actually supports QUIC.
+macro(openssl_check_quic)
+  if(NOT DEFINED HAVE_SSL_CTX_SET_QUIC_METHOD)
+    if(USE_OPENSSL)
+      openssl_check_symbol_exists(SSL_CTX_set_quic_method "openssl/ssl.h" HAVE_SSL_CTX_SET_QUIC_METHOD)
+    elseif(USE_WOLFSSL)
+      openssl_check_symbol_exists(wolfSSL_set_quic_method "wolfssl/options.h;wolfssl/openssl/ssl.h" HAVE_SSL_CTX_SET_QUIC_METHOD)
```
cbodley, line 659: do you know whether quiche actually supports wolfssl? afaik it depends on boringssl: https://github.com/cloudflare/quiche#building

so this cmake configure step may succeed with USE_WOLFSSL and USE_QUICHE, but fail during compilation

the USE_QUICHE block should probably include a check for HAVE_BORINGSSL to make sure that openssl_check_quic() actually checked for the correct SSL_CTX_set_quic_method symbol here

cbodley (commented):
thanks, this does fix the issue with USE_QUICHE and boringssl:
```
-- Looking for OPENSSL_IS_BORINGSSL
-- Looking for OPENSSL_IS_BORINGSSL - found
-- Looking for OPENSSL_IS_AWSLC
-- Looking for OPENSSL_IS_AWSLC - not found
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.13")
-- Looking for SSL_set0_wbio
-- Looking for SSL_set0_wbio - found
-- Looking for SSL_CTX_set_srp_username
-- Looking for SSL_CTX_set_srp_username - not found
-- Checking for one of the modules 'quiche'
-- Found QUICHE: /home/cbodley/ceph/build/src/rgw/quiche/target/release/libquiche.a
-- Looking for SSL_CTX_set_quic_method
-- Looking for SSL_CTX_set_quic_method - found
-- Looking for quiche_conn_set_qlog_fd
-- Looking for quiche_conn_set_qlog_fd - found
```

CMakeLists.txt
vszakats: No, I haven't made tests with quiche yet.

If quiche depends specifically on BoringSSL, it's sensible 
indeed to check for that. Pushed a commit to do this.

Do you know if quiche works with AWS-LC as well? It's Amazon's BoringSSL fork.

cbodley: > No, I haven't made tests with quiche yet.
> 
> If quiche depends specifically on BoringSSL, it's sensible indeed to check for that. Pushed a commit to do this.

thanks, updated branch still works for me 👍 

> 
> Do you know if quiche works with AWS-LC as well? It's Amazon's BoringSSL fork.

i don't see any discussion about AWS-LC on their github. i checked out/built AWS-LC, but when i plugged it into quiche i got undefined references:
```
/path/to/quiche/quiche/src/tls.rs:672: undefined reference to `SSL_get_curve_id'
/usr/bin/ld: /path/to/quiche/quiche/src/tls.rs:677: undefined reference to `SSL_get_curve_name'
```

vszakats: Thanks for testing these.

BoringSSL it is then. We can extend this when quiche adds support for more forks (or vice versa.)

cbodley: who knows, maybe openssl will support quic someday 🙃

vszakats: by _some_ definition of "quic" :)

bagder: > do you know whether quiche actually supports wolfssl

I'm pretty sure it does not. It is written with BoringSSL in mind.

bagder: > maybe openssl will support quic someday

They already do, but they implement their own QUIC stack and not the "standard" API for it. curl has no support for OpenSSL's API and AFAIK, neither does nghttp3 so we are quite far off an OpenSSL backend for h3.

vszakats: Yeah, their completely different, newly developed quic API is now in 3.2-alpha.

For extra fun, there was some symbol re-use, but maybe not too bad:
https://github.com/openssl/openssl/issues/22204

PR #12164, opened 2023-10-20
variadic macros with --disable-verbose

icing:
- add pragmas for gcc and clang to ignore variadic macros pendanteries in C89 mode
- (void)some vars that are not used in --disable-verbose
- ifdef some functions not used in --disable-verbose
- refs #11891


icing:
abandoned in favor of #12167

PR #12167, opened 2023-10-20
apache style infof and trace macros/functions

icing:
- test for a simplified C99 variadic check
- args to infof() in --disable-verbose are no longer disregarded but must compile.

### DIFF
--- a/lib/arpa_telnet.h
+++ b/lib/arpa_telnet.h
@@ -56,12 +56,14 @@ static const char * const telnetoptions[]=
   "TERM SPEED",  "LFLOW",          "LINEMODE",      "XDISPLOC",
   "OLD-ENVIRON", "AUTHENTICATION", "ENCRYPT",       "NEW-ENVIRON"
 };
+#define CURL_TELOPT(x)    telnetoptions[x]
+#else
+#define CURL_TELOPT(x)    ""
 #endif
 
 #define CURL_TELOPT_MAXIMUM CURL_TELOPT_NEW_ENVIRON
 
 #define CURL_TELOPT_OK(x) ((x) <= CURL_TELOPT_MAXIMUM)
-#define CURL_TELOPT(x)    telnetoptions[x]
 
 #define CURL_NTELOPTS 40
 
@@ -103,7 +105,12 @@ static const char * const telnetcmds[]=
 
 #define CURL_TELCMD_OK(x) ( ((unsigned int)(x) >= CURL_TELCMD_MINIMUM) && \
                        ((unsigned int)(x) <= CURL_TELCMD_MAXIMUM) )
+
+#ifndef CURL_DISABLE_VERBOSE_STRINGS
 #define CURL_TELCMD(x)    telnetcmds[(x)-CURL_TELCMD_MINIMUM]
+#else
+#define CURL_TELCMD(x)    ""
+#endif
 
 #endif /* CURL_DISABLE_TELNET */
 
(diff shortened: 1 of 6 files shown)


bagder:
This looks like my favorite take so far on the c89 issue!

bagder:
@vszakats what's your opinion on this approach?

vszakats:
If I'm reading the diff correctly this patch will:
1. move feature detection to C (from build systems). Assuming this is correct also for all curl build envs (vs httpd), it's a definite improvement and less moving parts.
2. replace the noop hack with an empty function. Downside is with -O0 builds with disabled verbose strings. Probably not an issue in most build cases.
3. keep using variadic macros, so the root problem remains with us. Difficult to estimate any fallouts.

If we want to keep using non-C89 syntax, this can improve things a bit
and address some pending issues. (The envs I'm using weren't affected
by those, so can't confirm this with tests.)

bagder:
> keep using variadic macros, so the root problem remains with us. Difficult to estimate any fallouts.

The variadic macros are only used if the compiler supports them, though. Which admittedly is going to be almost every compiler used to build curl.

icing:
> If I'm reading the diff correctly this patch will:
> 
> ...
> 3. keep using variadic macros, so the root problem remains with us. Difficult to estimate any fallouts.

Could you elaborate what the "root" problem is if the compiler claims to support C99? AFAIK, this has been in use for over a decade in Apache httpd on various platforms.

vszakats:
The root issue is C99 syntax present in a C89 project. This can be hacked around forever, it may even work, in most cases or in all cases, if the necessary tricks are put in place. I cannot answer what tricks are necessary to make this work "well enough". 

This and the fact that we have pending issues as a result of this (and seeing other projects with these exact issues), pushed me to the approach to make the source compliant.

I proposed solutions to avoid the assumed (but never measured) performance drop that this may cause. I can work on that if somebody is interested.

Everything is a balancing act, but in general I'd prefer sticking to C89 code first, then start optimizing for performance, if necessary.

icing:
> The root issue is C99 syntax present in a C89 project. This can be hacked around forever, it may even work, in most cases or in all cases, if the necessary tricks are put in place. I cannot answer what tricks are necessary to make this work "well enough".

Thanks. I would not call the proposed solution a "trick", but it certainly breaks out of the C89 restrictions where C99 is available. Admittedly, this keeps it "non-pure-C89" (as it was before).

> Everything is a balancing act, but in general I'd prefer sticking to C89 code first, then start optimizing for performance, if necessary.

The problems, as I perceive them,  were the optimizations around `--disable-verbose` that tried to trick C89 compilers into not generating code for what is a valid function call. That led to problems because this is not achievable in a portable, compliant way. That "trick" is being removed in this PR at the cost of a very dumb or restricted compiler not optimizing the static calls.

I personally am not a fan of the alternate `infof((args))` double parenthesis calls. A matter of taste. 

Both ways seem to work.

vszakats:
No worries, thanks for this patch and all your input. I'm also not a super fan of double parentheses, but from a style POV I don't mind them (probably got used to them in another project). That aside, new log calls are typically copy-pasted from existing ones, so rarely will someone type it from scratch. Even if so, mistakes are caught instantly on compile. My concern is perf drop if any. The double arg passing (if we want to go there), well, that I also need to get used to, but probably still passable, and not affecting the final binaries negatively.

I also wonder if there is any further C89-trick that could be applied here.

bagder:
Fixes #12083
Fixes #11880
Fixes #11891

PR #12176, opened 2023-10-21
http: consider resume with CURLOPT_FAILONERRROR and 416 to be fine

### LINKED ISSUE #10521: http fails on 416 when resuming

### LINKED ISSUE #12174: Curl is failing when downloading a already downloaded file with resume download and --fail error set
### I did this

I tried to download a file and it got downloaded and then again i tried to download with resume download and --fail. 

# ran it twice
curl -O -C - -f "https://github.com/robots.txt"
curl -O -C - -f "https://github.com/robots.txt"

### I expected the following

curl: (22) The requested URL returned error: 416

The above was the error recieved which shouldn,t have happened as this was fixed already in https://github.com/curl/curl/issues/6740 as part of Fixed in 7.76.0 - March 31 2021 release

### curl/libcurl version

curl 7.88.1 (x86_64-pc-linux-gnu) libcurl/7.88.1 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.3 libpsl/0.21.2 (+libidn2/2.3.3) libssh/0.10.4/openssl/zlib nghttp2/1.52.0 librtmp/2.3
Release-Date: 2023-02-20

### operating system

Linux ar 6.2.0-34-generic #34-Ubuntu SMP PREEMPT_DYNAMIC Mon Sep  4 13:06:55 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

bagder:
Finding a 'Content-Range:' in the response changed the handling.

Add test case 1475 to verify -C - with 416 and Content-Range: header, which is almost exactly like test 194 which instead uses a fixed -C offset. Adjusted test 194 to also be considered fine.

Fixes #10521
Reported-by: Smackd0wn
Fixes #12174
Reported-by: Anubhav Rai

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -3663,7 +3663,7 @@ CURLcode Curl_http_header(struct Curl_easy *data, struct connectdata *conn,
           k->content_range = TRUE;
       }
     }
-    else
+    else if(k->httpcode < 300)
       data->state.resume_from = 0; /* get everything */
   }
 #if !defined(CURL_DISABLE_COOKIES)
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -185,7 +185,7 @@ test1439 test1440 test1441 test1442 test1443 test1444 test1445 test1446 \
 test1447 test1448 test1449 test1450 test1451 test1452 test1453 test1454 \
 test1455 test1456 test1457 test1458 test1459 test1460 test1461 test1462 \
 test1463 test1464 test1465 test1466 test1467 test1468 test1469 test1470 \
-test1471 test1472 test1473 test1474 \
+test1471 test1472 test1473 test1474 test1475 \
 \
 test1500 test1501 test1502 test1503 test1504 test1505 test1506 test1507 \
 test1508 test1509 test1510 test1511 test1512 test1513 test1514 test1515 \
--- a/tests/data/test1475
+++ b/tests/data/test1475
@@ -1,0 +1,83 @@
+<testcase>
+# also verified by 1156 in libcurl API terms
+
+<info>
+<keywords>
+HTTP
+HTTP GET
+Resume
+</keywords>
+</info>
+
+# Server-side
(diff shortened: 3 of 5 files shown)


Anubhav-Rai:
Hi @bagder , I see it has passed all test cases and i have tested on many systems and it is failing . What could be the issue. are the test cases proper?

bagder:
> Hi @bagder , I see it has passed all test cases and i have tested on many systems where it failed . What could be the issue. are the test cases proper?

Are you saying you have applied this patch, rebuilt curl and it still shows the problem? If so, then please provide a lot more details.

Anubhav-Rai:
No that is not what i meant . I am saying that this issue was fixed in 2021 right and it still reappears. Is this a real issue or a problem from my side ? 
Thank you

bagder:
The case you report here was not fixed in 2021, which is why I'm proposing a fix now.

PR #12178, opened 2023-10-21
tool/-T: perform non-chunked transfer for stdin if it is a regular file

### LINKED ISSUE #12171: `-T -` always uploads with "Transfer-Encoding: chunked"
### I did this

```bash
curl -T - localhost:40000 < ~/.bashrc
```
```console
$ nc -l 40000
PUT / HTTP/1.1
Host: localhost:40000
User-Agent: curl/8.4.0
Accept: */*
Transfer-Encoding: chunked
Expect: 100-continue

30e
[...]
```

### I expected the following

```console
$ nc -l 40000
PUT / HTTP/1.1
Host: localhost:40000
User-Agent: curl/8.4.0
Accept: */*
Content-Length: 782

[...]
```

`curl` should realise that stdin is a regular file, and that it can do a non-chunked upload, like it would do if you used `-T ~/.bashrc`.
~~`curl` already does that when you use `@-` or `<-` with `-F`/`-d` and similar, but not when you use `-T`.~~ (EDIT: Actually, that is not true; I was misremembering; `curl` only knows the content length for stdin with `-F`/`-d` because it always loads the entire file into memory before sending it)

The problem is that `curl -T` never stats stdin to figure if it is a regular file when you use `-` as argument.
On Linux and BSDs, a possible workround is `-T /dev/stdin` (with `--request-target /`); since a path was provided, `curl` will stat it and realise it is a regular file, but it would be nicer if `curl` would just do that on its own.

I wanted to create a patch to fix this bug when I first learned about it, but the code that deals with stat-ing the file for upload (in `src/tool_operate.c`) is doing something special on `#define __VMS` systems to figure out the *real* upload size of regular files that involes opening the file, and reading all its data to count bytes. I didn't know why it did that, or how to do that for stdin if the system is VMS (maybe using `lseek` to go to the start of stdin), and anyway I would not be able to test the code I write for VMS on VMS, so I gave up trying to write the patch myself to avoid accidentally breaking the VMS port.

I remembered about this bug today, and noticed no one created an issue for it.

Ref: https://github.com/curl/curl/pull/11396#issuecomment-1613684489

### curl/libcurl version

curl 8.4.0
```
curl 8.4.0 (x86_64-pc-linux-gnu) libcurl/8.4.0 OpenSSL/3.1.3 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.11.0 nghttp2/1.57.0
Release-Date: 2023-10-11
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zst
```

### operating system

Arch Linux
```
Linux t420 6.1.57-1-lts #1 SMP PREEMPT_DYNAMIC Wed, 11 Oct 2023 05:10:50 +0000 x86_64 GNU/Linux
```

### LINKED ISSUE #12177: `-T /dev/stdin`/`-T /dev/fd/$n` may upload with an incorrect content length on BSDs
### I did this

`-T path` will stat `path` to figure out its size in bytes to use it as `Content-Length` if it is a regular file.

The problem with that is that, on BSDs and some other UNIXes (not Linux), `open(path)` may not give you a file descriptor with a 0 offset from the start of the file.

On BSDs, if the path given to open is to a `/dev/fd/*` file or `/dev/stdin`, `/dev/stdout`, `/dev/stderr`, that will trigger a special behaviour for `open`: `open` on BSD, instead of reopening the file related to that file descriptor with different open modes like in Linux, will actually duplicate the file descriptor specified by that path if its open modes are compatible with the ones requested by `open`, and otherwise fail.

Since `-T /dev/fd/4` for example, will effectively just be a `dup(4)`, when `curl` will open the path it thinks is for a regular file (because fd 4 is indeed a file descriptor for a regular file, `myfile`), it will not find itself at byte offset 0 into the file, and will send less data than `Content-Length`.

```sh
printf 'NAME\tJOB\nTom\tCarpenter\nAlice\tMechanic\n' > myfile
{
   read -r skip_firstline <&4 &&
   curl -T /dev/fd/4 localhost:43000
} 4< myfile
```
```console
$ nc -l localhost:43000
PUT /4 HTTP/1.1
Host: localhost:43000
User-Agent: curl/8.2.1
Accept: */*
Content-Length: 38

Tom     Carpenter
Alice   Mechanic
```

### I expected the following

I expected:
```console
$ nc -l 43000
PUT /4 HTTP/1.1
Host: localhost:43000
User-Agent: curl/8.2.1
Accept: */*
Content-Length: 29

Tom     Carpenter
Alice   Mechanic
```
`curl` should checked the current offset into the regular file and use its `size - offset` as content length instead of just using its size.

---
Other possible alternatives:
* check if the offset into the file is not 0 after operning it, and perform a chunked upload in that case
```console
$ nc -l 43000
PUT /4 HTTP/1.1
Host: localhost:43000
User-Agent: curl/8.2.1
Accept: */*
Transfer-Encoding: chunked
Expect: 100-continue

1d
Tom     Carpenter
Alice   Mechanic

0
[...]
```
* seek to offset 0 after opening the regular file and send the entire file:
```console
$ nc -l 43000
PUT /4 HTTP/1.1
Host: localhost:43000
User-Agent: curl/8.2.1
Accept: */*
Content-Length: 38

NAME    JOB
Tom     Carpenter
Alice   Mechanic
```

---
Those solutions are all fairly easy to implement. I, personally, would prefer either the first or the second approach, I don't really like the idea of `curl` seeking back to 0 after opening.

### curl/libcurl version

curl 8.2.1 (I am using my <https://sdf.org/> shell account to test this)
```
curl 8.2.1 (x86_64--netbsd) libcurl/8.2.1 OpenSSL/1.1.1u zlib/1.2.13 libidn2/2.3.4 nghttp2/1.52.0
Release-Date: 2023-07-26
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL threadsafe TLS-SRP UnixSocket
```

### operating system

```
NetBSD sdf 9.3 NetBSD 9.3 (GENERIC) #0: Thu Aug  4 15:30:37 UTC 2022  mkrepro@mkrepro.NetBSD.org:/usr/src/sys/arch/amd64/compile/GENERIC amd64
```

emanuele6:
curl will now also compute the content-length of the transfer if stdin is the file to upload and stdin is a regular file, using its file size.

Since, while being a regular file, stdin could not have its offset at the start of the file, curl will now also get the current offset into the upload file's file descriptor and use (filesize - offset) as content-length for transfer instead of just using the full filesize. This also fixes a bug on BSDs where open("/dev/fd/N") behaves like dup(N), so, if N is a file descriptor to a regular file, the file offset of the file descriptor returned by open() may not have been at the start of the file despite curl's previous assumption.

Since I don't know anything about VMS systems, I left the behaviour for VMS unchanged; on VMS, curl will still perform a chunked transfer if the upload file is stdin.

Fixes #12171
Fixes #12177

---

```console
$ wc -c ~/.bashrc
782
$ sed 1d ~/.bashrc | wc -c
767
```
```console
$ # src/curl -T - localhost:40400 < ~/.bashrc
$ nc -l 40400
PUT / HTTP/1.1
Host: localhost:40400
User-Agent: curl/8.3.0-DEV
Accept: */*
Content-Length: 782

#!/bin/bash --
#  _               _
# | |__   __ _ ___| |__  _ __ ___
# | '_ \ / _` / __| '_ \| '__/ __|
# | |_) | (_| \__ \ | | | | | (__
# |_.__/ \__,_|___/_| |_|_|  \___|
[...]
```
```console
$ # cat ~/.bashrc | src/curl -T - localhost:40400
$ nc -l 40400
PUT / HTTP/1.1
Host: localhost:40400
User-Agent: curl/8.3.0-DEV
Accept: */*
Transfer-Encoding: chunked
Expect: 100-continue

30e
#!/bin/bash --
#  _               _
# | |__   __ _ ___| |__  _ __ ___
# | '_ \ / _` / __| '_ \| '__/ __|
# | |_) | (_| \__ \ | | | | | (__
# |_.__/ \__,_|___/_| |_|_|  \___|
[...]
```
```console
$ # { read -r x && src/curl -T - localhost:40400 ;} < ~/.bashrc
$ nc -l 40400
PUT / HTTP/1.1
Host: localhost:40400
User-Agent: curl/8.3.0-DEV
Accept: */*
Content-Length: 767

#  _               _
# | |__   __ _ ___| |__  _ __ ___
# | '_ \ / _` / __| '_ \| '__/ __|
# | |_) | (_| \__ \ | | | | | (__
# |_.__/ \__,_|___/_| |_|_|  \___|
[...]
```

---
I didn't add a test to the test suite because I don't know how to write tests like those that has a open file descriptor to a regular file with a non-zero offset.
```

### CODE REVIEW

**Comment by emanuele6** [comment]
oops, I forgot to rebase

File: src/tool_operate.c [hunk 1]
```diff
@@ -315,7 +317,22 @@ static CURLcode pre_transfer(struct GlobalConfig *global,
 
     /* we ignore file size for char/block devices, sockets, etc. */
     if(S_ISREG(fileinfo.st_mode))
+#ifdef __VMS
       uploadfilesize = fileinfo.st_size;
+#else
+    {
+      /* When the upload file is stdin, or when the upload file is
+         /dev/std{in,out,err} or /dev/fd/N on BSDs, the offset may not
+         be 0 */
```
jay, line 326: can you give an example of this happening

emanuele6: Sure!
For stdin as upload file (`-T -`), it is the third example in the PR description: `{ read -r x && src/curl -T - localhost:40400 ;} < ~/.bashrc`.
For an example of `-T /dev/fd/N` etc on BSDs, see #12177.

bagder:
~~~
TESTFAIL: These test cases failed: 60 98 1068 1069 1073 
~~~

These tests are assumed to not be able to get the size from stdin, so probably we need to change how `runtests.pl` invokes those with data on stdin. Right now they are made to run `< log/stdin` , which thus with this PR is treated differently now.

emanuele6:
Hmm, yeah, the test runner is doing `$cmdargs .= " <$stdinfile";` to run curl with a specified stdin, and `$stdinfile` is always a regular file, but those tests assume stdin will be a pipe basically.

src/tool_operate.c
```diff
@@ -300,6 +300,8 @@ static CURLcode pre_transfer(struct GlobalConfig *global,
     }
     if(per->infd == -1)
 #else
+  if(per->uploadfile) {
+    if(!stdin_upload(per->uploadfile))
       per->infd = open(per->uploadfile, O_RDONLY | O_BINARY);
     if((per->infd == -1) || fstat(per->infd, &fileinfo))
```
emanuele6, line 306: I need to refactor this a bit better, or this will close stdin if `fstat` fails here

emanuele6: Also `per->infdopen` should be set to true only if uploadfile is not stdin

emanuele6:
I was reading this source file for another reason, and I found this block of code, which is doing pretty much what I am doing in this PR, but for `-F@`, using `stdio.h`.
https://github.com/curl/curl/blob/b6e6d4ff8f253c8b8055bab9d4d6a10f9be109f3/src/tool_formparse.c#L104-L123

I think that that `sbuf.st_fab_rfm != FAB$C_VAR && sbuf.st_fab_rfm != FAB$C_VFC &&` for VMS is supposed to tell you when it is ok to use `statsize - offset` for stdin on VMS, but I am not sure if I should add that to make `curl` not need to do a chunked transfer for stdin on VMS in some extra cases, since I have no idea what that actually does.

@monnerat You wrote that snippet of code, could you please tell me if this looks correct to you? Are you be able to test the patch of VMS?
```diff
diff --git a/src/tool_operate.c b/src/tool_operate.c
index 96c7ac745..4d7642ede 100644
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -306,12 +306,11 @@ static CURLcode pre_transfer(struct GlobalConfig *global,
 #endif
     }
 
+    if (per->infd == -1 || (
 #ifdef __VMS
-    if (per->infd == -1)
-#else
-    if (per->infd == -1 || fstat(per->infd, &fileinfo))
+        isstdin &&
 #endif
-    {
+        fstat(per->infd, &fileinfo))) {
       helpf(tool_stderr, "Can't open '%s'", per->uploadfile);
       if(!isstdin && per->infd != -1) {
         close(per->infd);
@@ -322,11 +321,11 @@ static CURLcode pre_transfer(struct GlobalConfig *global,
     per->infdopen = isstdin;
 
     /* we ignore file size for char/block devices, sockets, etc. */
-    if(S_ISREG(fileinfo.st_mode))
+    if(S_ISREG(fileinfo.st_mode)
 #ifdef __VMS
-      uploadfilesize = fileinfo.st_size;
+       && sbuf.st_fab_rfm != FAB$C_VAR && sbuf.st_fab_rfm != FAB$C_VFC
 #else
-    {
+    ) {
       /* When the upload file is stdin, or when the upload file is
          /dev/std{in,out,err} or /dev/fd/N on BSDs, the offset may not
          be 0 */
```

tests/data/test1073
```diff
@@ -47,7 +47,7 @@ which is impossible in HTTP/1.0
 # Verify data after the test has been "shot"
 <verify>
 <errorcode>
-25
+65
```
emanuele6, line 50: I had to change the error code to `65` because I think now the problem is that curl tries to rewind the file descriptor to resend it after the redirect, but it can't because stdin is not seekable

emanuele6:
@bagder Now you can specify `<stdin pipe="yes">` if you want stdin to be a pipe in a test.

tests/data/test1072
```diff
@@ -53,7 +53,7 @@ which is impossible in HTTP/1.0
 # Verify data after the test has been "shot"
 <verify>
 <errorcode>
-25
```
emanuele6, line 56: ditto

emanuele6:
If I want to add a test that checks that `-T -` (with regular file stdin) does not perform a *chunky upload*, what number should it be?

bagder:
> If I want to add a test that checks that `-T -` (with regular file stdin) does not perform a _chunky upload_, what number should it be?

You can pick any available number. I usually pick one that is the next available counting upward from the one I clone when I made a new one... But you can do whatever. Don't go crazy and pick one way outside of the current range though. :grin: 

If there still ends up a number collision at merge time because another PR has had the bad idea and picked *the same* number, renumbering a single test case should be just a matter of renaming it  as we use variables for the test number within the files.

emanuele6:
Do I need to somehow disable the test on `VMS` or is that not necessary because `runtests.pl` can only run on UNIX-like systems?

bagder:
> Do I need to somehow disable the test on `VMS` or is that not necessary because `runtests.pl` can only run on UNIX-like systems?

runtests should run on most systems, but I am not sure if it works/worked on VMS. We generally don't have any special stuff for VMS in the test suite, and I suspect that is because it is never executed on that OS...

Since we don't have any regulars around using VMS, we have to do as as good as we can and simply guess and then let people on VMS sweep up the pieces later and fix that we broke.

monnerat:
> I think that that sbuf.st_fab_rfm != FAB$C_VAR && sbuf.st_fab_rfm != FAB$C_VFC && for VMS is supposed to tell you when it is ok to use statsize - offset for stdin on VMS, but I am not sure if I should add that to make curl not need to do a chunked transfer for stdin on VMS in some cases, since I have no idea what that actually does.

As you say, I wrote the MIME api version of `tool_formparse.c`, but I've never put my hands on a VMS keyboard: I just came to the same conclusion as you by reading function `VmsSpecialSize`() in file `tool_operate.c` as I needed to know the file size.

This should probably be confirmed/adjusted/refined/completed by some VMS guy, but there hasn't been any complaint about it in 6 years :-)

If it can help you, I just found this: http://odl.sysworks.biz/disk$axpdocmar021/opsys/vmsos73/vmsos73/4523/4523pro_006.html.

emanuele6:
@monnerat Hmm, ok. Thanks anyway.

emanuele6:
The windows CIs are failing because `1072` and `1073` are still returning 25 on windows; the return 65 on linux. :/

bagder:
> The windows CIs are failing because `1072` and `1073` are still returning 25 on windows; they return 65 on linux. :/

Is this because Windows does not treat the file the same thing when this is used? We need to manage this somehow...

emanuele6:
I think the problem may have something to do with `--anyauth` always printing a warning when the upload file is stdin; it should rather only warn if the file is a non-regular file in general, not just only stdin always. I will try to take a look later.

bagder:
Do you plan to resurrect this work @emanuele6 ?

emanuele6:
The only conflict seems to be that the test number that I was using has been taken for something else; the patch was working for UNIX-like systems; as far as I remember, the only problem was that this changed some exit codes in Windows because it was causing curl to sometimes fail earlier in a different way.
I cannot test Windows locally, so that was blocking me a bit.

PR #12182, opened 2023-10-23
BINDINGS: add V binding

ttytm:
Astonishing how good libcurl works. Thanks a lot for the work that so many projects can build on.

The PR adds a link to a V request lib that uses libcurl.


bagder (approved):

bagder:
Thanks!

PR #12184, opened 2023-10-23
Client writer, part 2, accounting + logging

icing:
This PR has these changes:

Renaming of unencode_* to cwriter, e.g. client writers
- documentation of sendf.h functions
- move max decode stack checks back to content_encoding.c
- define writer phase which was used as order before
- introduce phases for monitoring inbetween decode phases
- offering default implementations for init/write/close

Add type paramter to client writer's do_write()
- always pass all writes through the writer stack
- writers who only care about BODY data will pass other writes unchanged

add RAW and PROTOCOL client writers
- RAW used for Curl_debug() logging of CURLINFO_DATA_IN
- PROTOCOL used for updates to data->req.bytecount, max_filesize checks and Curl_pgrsSetDownloadCounter()
- remove all updates of data->req.bytecount and calls to Curl_pgrsSetDownloadCounter() and Curl_debug() from other code
- adjust test457 expected output to no longer see the excess write

### DIFF
--- a/tests/data/test457
+++ b/tests/data/test457
@@ -20,7 +20,8 @@ aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 30
 bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
 21;heresatest=moooo
-cccccccccccccccccccccccccccccccc
+cccccccccccccccccccccccccccccc
+c
 
 0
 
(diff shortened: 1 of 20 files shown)


tests/data/test457
```diff
@@ -20,7 +20,8 @@ aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 30
 bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
 21;heresatest=moooo
-cccccccccccccccccccccccccccccccc
+cccccccccccccccccccccccccccccc
+c
```
bagder, line 24: I'm curious: was this change needed to make the test work now? If so: why?

icing: Ah, indeed. We used to write excess body data, e.g. data exceeding `--max-filesize` and handle the error afterwards. With the PR, we write only the max amount and then error.

I had to do some trickery with the `\n`, so that the expected data ended with `\n`.

PR #12186, opened 2023-10-23
cmake: allow `HAVE_POLL_FINE` when cross-compiling

hwti:
The autotools build skips the run test in this case.
Since the CMake build doesn't have the simpler compile tests, we keep the existing run test, but when cross-compiling we just check if it compiles.

### DIFF
--- a/CMake/OtherTests.cmake
+++ b/CMake/OtherTests.cmake
@@ -86,51 +86,53 @@ endif()
 
 unset(CMAKE_TRY_COMPILE_TARGET_TYPE)
 
-if(NOT CMAKE_CROSSCOMPILING)
-  if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "iOS")
-    # only try this on non-apple platforms
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "iOS")
+  # only try this on non-apple platforms
 
-    # if not cross-compilation...
-    set(CMAKE_REQUIRED_FLAGS "")
-    if(HAVE_SYS_POLL_H)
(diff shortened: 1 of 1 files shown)


CMake/OtherTests.cmake
```diff
@@ -86,51 +86,54 @@ endif()
...
-      #ifdef HAVE_SYS_POLL_H
-      #  include <sys/poll.h>
-      #elif  HAVE_POLL_H
-      #  include <poll.h>
-      #endif
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "iOS")
+  # only try this on non-apple platforms
+
+  # if not cross-compilation...
+  set(CMAKE_REQUIRED_FLAGS "")
```
vszakats, line 93:
```suggestion
  # only try this on non-Apple platforms

  set(CMAKE_REQUIRED_FLAGS "")
```
Strip the comment that becomes obsolete with this patch.

hwti:
> I don't think this is what we want. The "good poll" cannot be detected only by compiling. It needs to run to evaluate how poll() works. If it can't run, it can't check. Which is why it does no attempt for cross-compiled builds.

But the autotools build assumes it is a "good poll" when cross-compiling : https://github.com/curl/curl/blob/93d8e3564e85588796792aa6e3ba0c37de9ac752/m4/curl-functions.m4#L4081 (`tst_works_poll` is kept at `unknown`, and later the tests are `"$tst_works_poll" != "no"`).
So why shouldn't the CMake build do the same ?

Since most of the time poll works correctly, I think it makes sense to have a way to use it when cross-compiling.

vszakats:
If I understand correctly, cmake _disables_ poll when unable to detect, while autotools _enables_ it when unable to detect?

If so, I think it's sensible to sync their behaviour, but then which default is the "more correct" one? I'd err on the side of caution and disable when unsure (in cross-builds), like cmake does now. Then cross-build targets with well-known support can be allowlisted. Otherwise it's up the builder to enable it manually.

Issue #12189, opened 2023-10-24
Description of EXIT CODE 25 in man page is not fully correct


emanuele6:
### I did this

`curl.1` describes exit code 25 as:
```
       25     FTP  could  not STOR file. The server denied the STOR operation,
              used for FTP uploading.
```
Suggesting it is a FTP only error.

### I expected the following

`curl` can return 25 if any upload fails (e.g. HTTP PUT):
```sh
$ cat .bashrc | curl -sST - -0 example.org
curl: (25) Chunky upload is not supported by HTTP 1.0
```
`libcurl-errors.3` describes error 25 (`CURLE_UPLOAD_FAILED`) as:
```
       CURLE_UPLOAD_FAILED (25)
              Failed starting the upload. For FTP, the server typically denied
              the STOR command. The error buffer usually contains the server's
              explanation for this.
```

### curl/libcurl version

curl 8.4.0
```
curl 8.4.0 (x86_64-pc-linux-gnu) libcurl/8.4.0 OpenSSL/3.1.3 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.11.0 nghttp2/1.57.0
Release-Date: 2023-10-11
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

ArchLinux
```
Linux t420 6.1.57-1-lts #1 SMP PREEMPT_DYNAMIC Wed, 11 Oct 2023 05:10:50 +0000 x86_64 GNU/Linux
```

PR #12190, opened 2023-10-24
page-footer: clarify exit code 25

jay:
- Clarify that curl tool exit code 25 means an upload failed to start.

Exit code 25 is equivalent to CURLE_UPLOAD_FAILED (25). Prior to this change the documentation only mentioned the case of FTP STOR failing.

Reported-by: Emanuele Torre

Ref: https://github.com/curl/curl/blob/curl-8_4_0/docs/libcurl/libcurl-errors.3#L113-L115

Fixes https://github.com/curl/curl/issues/12189
Closes #xxxx

### DIFF
--- a/docs/cmdline-opts/page-footer
+++ b/docs/cmdline-opts/page-footer
@@ -173,8 +173,8 @@ appears if --fail is used.
 .IP 23
 Write error. Curl could not write data to a local filesystem or similar.
 .IP 25
-FTP could not STOR file. The server denied the STOR operation, used for FTP
-uploading.
+Failed starting the upload. For FTP, the server typically denied the STOR
+command.
 .IP 26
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
[Everything curl](https://everything.curl.dev/usingcurl/returns) documents it this way:

> Upload failed. The server refused to accept or store the file that curl tried to send to it. This is usually due to wrong access rights on the server but can also happen due to out of disk space or other resource constraints. This error can happen for many protocols.

jay:
That's a bit more general because it does not say that the upload failed to start, just that it failed. libcurl strerror text and libcurl-errors.3 both say failed to start. AFAICT it's only when it failed to start. I think either way is fine though.

PR #12198, opened 2023-10-25
Move async resolver state from easy handle to connectdata

icing:
- resolving is done for a connection, not for every transfer
- save create/dup/free of a cares channel for each transfer
- check values of setopt calls against a local channel if no connection has been attached yet, when needed.


bagder:
Unfortunately this introduced a torture test fail. On my machine this crashes:

`$ ./runtests.pl -t41 -n 534`

bagder:
in a c-ares build I should add

bagder:
Follow-up in #12219

PR #12199, opened 2023-10-25
docs: preserve the modification date when copying the prebuilt man page

mkauf:
The previously built man page "curl.1" must be copied with the original modification date, otherwise the man page is never updated.

This fixes a bug that has been introduced with commit 2568441cab.

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -111,7 +111,8 @@ SUFFIXES = .1 .html .pdf
 # have changed.
 $(abs_builddir)/curl.1:
 	if test "$(top_builddir)x" != "$(top_srcdir)x" -a -e "$(srcdir)/curl.1"; then \
-		$(INSTALL_DATA) "$(srcdir)/curl.1" $@; fi
+		$(INSTALL_DATA) "$(srcdir)/curl.1" $@ \
+		&& touch -r "$(srcdir)/curl.1" $@; fi
 	cd cmdline-opts && $(MAKE)
 
 html: $(HTMLPAGES)


mkauf:
Fix this autobuild problem: https://curl.se/dev/log.cgi?id=20231025023847-2167007#prob3

CC @dfandrich

dfandrich:
-p is nonstandard, so it won't be completely portable. The main platforms
support it, but there's no way to override the option on platforms where it
doesn't work, which isn't ideal. But, I'm not sure if this is anymore more than
a theoretical concern in this day and age.

mkauf:
A different bugfix that uses the `touch` command:
```
diff --git a/docs/Makefile.am b/docs/Makefile.am
index 5454e8330..bb8241e72 100644
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -111,7 +111,8 @@ SUFFIXES = .1 .html .pdf
 # have changed.
 $(abs_builddir)/curl.1:
        if test "$(top_builddir)x" != "$(top_srcdir)x" -a -e "$(srcdir)/curl.1"; then \
-               $(INSTALL_DATA) "$(srcdir)/curl.1" $@; fi
+               $(INSTALL_DATA) "$(srcdir)/curl.1" $@; \
+               touch -r "$(srcdir)/curl.1" $@; fi
        cd cmdline-opts && $(MAKE)
 
 html: $(HTMLPAGES)
```

bagder:
> A different bugfix that uses the touch command:

I would expect touch -r have the same problems or perhaps worse than cp -p ...

dfandrich:
cp -p and touch -r are POSIX though; install -p isn't (that I can find).

dfandrich:
I think the touch proposal three comments up has the best chance of working. We're not using touch anywhere else in the build system yet but it's a standard POSIX command so it should be available everywhere.

mkauf:
Ok, I have updated the pull request, now it uses "touch -r".

mkauf:
> I would expect touch -r have the same problems or perhaps worse than cp -p ...

`cp -p` also changes file ownership information, `touch -r` only changes the file's timestamp. So I think that it should work.

dfandrich:
It was using cp -p until commit 2568441cab which switched to install and
introduced the problem this PR fixes. The switch away from cp -p was done
because cp errors out when it can't change the UID and GID, but we don't care
about those.

dfandrich (approved):

mkauf:
@bagder OK to merge this?

bagder (approved):

PR #5453, opened 2020-05-25
Adding zstd decoding support

gvollant:
Zstd is a very fast compressor, and give better speed vs compression ratio than gzip/deflate or brotli.

This pull request add zstd support on curl and libcurl

to test
curl --compressed https://de-de.facebook.com/
curl --compressed https://de-de.facebook.com/ --verbose

Somes links:
https://www.iana.org/assignments/http-parameters/http-parameters.xhtml

https://datatracker.ietf.org/doc/draft-kucherawy-rfc8478bis/

https://tools.ietf.org/html/rfc8478

https://facebook.github.io/zstd/

PR #12200, opened 2023-10-25
cmake: speed up zstd detection

vszakats:
Before this patch we detected the presence of a specific zstd API to
see if we can use the library. zstd published that API in its first
stable release: v1.0.0 (2016-08-31).

Replace that method by detecting the zstd library version instead and
accepting if it's v1.0.0 or newer. Also display this detected version
and display a warning if the zstd found is unfit for curl.

We use the same version detection method as zstd itself, via its public
C header.

This deviates from autotools which keeps using the slow method of
looking for the API by building a test program. The outcome is the same
as long as zstd keeps offering this API.

Ref: https://github.com/facebook/zstd/commit/5a0c8e24395079f8e8cdc90aa1659cd5ab1b7427 (2016-08-12, committed)
Ref: https://github.com/facebook/zstd/releases/tag/v0.8.1 (2016-08-18, first released)
Ref: https://github.com/facebook/zstd/releases/tag/v1.0.0

Closes #12200

---

```console
-- Checking for one of the modules 'libzstd'
-- Found Zstd: .../zstd/lib/libzstd.a (found version "1.5.5") 
```

### DIFF
--- a/CMake/FindZstd.cmake
+++ b/CMake/FindZstd.cmake
@@ -56,11 +56,18 @@ find_library(Zstd_LIBRARY NAMES zstd
     ${PC_Zstd_LIBRARY_DIRS}
 )
 
+if(Zstd_INCLUDE_DIR)
+  file(READ "${Zstd_INCLUDE_DIR}/zstd.h" _zstd_header)
+  string(REGEX MATCH ".*define ZSTD_VERSION_MAJOR *([0-9]+).*define ZSTD_VERSION_MINOR *([0-9]+).*define ZSTD_VERSION_RELEASE *([0-9]+)" _zstd_ver "${_zstd_header}")
+  set(Zstd_VERSION "${CMAKE_MATCH_1}.${CMAKE_MATCH_2}.${CMAKE_MATCH_3}")
+endif()
+
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(Zstd
   REQUIRED_VARS
     Zstd_LIBRARY
     Zstd_INCLUDE_DIR
+  VERSION_VAR Zstd_VERSION
 )
 
 if(Zstd_FOUND)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -613,6 +613,8 @@ if(CURL_ZSTD)
     set(HAVE_ZSTD ON)
     list(APPEND CURL_LIBS ${Zstd_LIBRARIES})
     include_directories(${Zstd_INCLUDE_DIRS})
+  else()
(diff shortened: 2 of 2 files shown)


bagder:
> This deviates from autotools which keeps using the slow method of
> looking for the API by building a test program

Right, but that's how configure always detects symbols/functions. It makes sure we can link with them. I don't think that is problematic.

On my Linux dev machine, a configure at current master run that detects a lot of third party components including HTTP/3, zstd, brotli and more, configure completes in less than 7 seconds.

A bare "cmake" run at the same commit on the same machine takes 0.8 seconds longer while detecting fewer 3rd party libs.

vszakats:
On my machine and on most CI machines, configure takes a lot more than 7 seconds. It's 82 seconds for a Windows cross-build on macOS for example and 52 seconds for pure macOS (_without_ running `autoreconf`). It's so slow that it makes it in practice unusable for local (and CI) iterations involving the build systems.

For curl-for-win specifically, this slow configuration stage needs to run twice. This performance has been traditionally the case with all Windows (and apparently macOS) machines and Windows cross-builds, as opposed to pure Linux where the necessary operations are fast. Surely with a bleeding edge computer this could also be made faster, but those are not even an option in free CI tiers.

I'm not attempting to make autotools fast, because I reckon its mojo is to do and redo all the detection work on every run. This also has a valid case on *nix systems that have wide and unpredicable variations on what's available (also true for CMake).

But, in many cases, these detection results are repeating the exact same work with predictable results. Speaking of Windows, almost all of these results are well-known. Same case with zstd: the result is the function of its version (which in this case means ALL existing stable releases), and detecting the version is much faster then triaging a full build. I understand this might not catch issues where zstd cannot be linked for other reasons (say a broken zstd build), or a broken toolchain setup, but these are rather rare. Also, toolchain issues are caught earlier anyway and even a broken zstd is caught at link stage latest.

I think this is safe. I can fast-track this manually for my builds, but why not make all builds benefit?

vszakats:
The zstd feature check was added for Ubuntu 16.04 LTS Xenial, according to https://github.com/curl/curl/pull/5453#issuecomment-634907456. Though Xenial is still officially supported till 2026, Ubuntu's package search page no longer supports it. According to one source, it came with v0.5.1, according to others, with v1.3.1. Either way, I verified and the version detection method works with v0.5.1:
```console
-- Checking for one of the modules 'libzstd'
-- Found Zstd: .../zstd-0.5.1/lib/pkg/usr/local/lib/libzstd.a (found version "0.5.1") 
```

bagder (approved):

PR #12204, opened 2023-10-26
SSL, some tidy ups

icing:
Move all handling of connection SSL config into vtls.c

- init defaults, cleanup
- inheriting from easy handle
- updating when easy handle is changed
- use Curl_ssl_cf_get_primary_config() in QUIC filters
- remove `Curl_ssl_get_config()`, no longer needed


bagder (approved):
Me like!

bagder (approved):

PR #12207, opened 2023-10-26
Provide call to tell connection to forget about a socket

icing:
- fixed libssh.c workaround for a socket being closed by the library
- eliminate the terrible hack in cf-socket.c to guess when this happened and try not closing the socket again.
- fixes race in eyeballing when socket could have failed to be closed for a discarded connect attempt

### DIFF
--- a/lib/vssh/libssh.c
+++ b/lib/vssh/libssh.c
@@ -1965,8 +1965,7 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
         /* conn->sock[FIRSTSOCKET] is closed by ssh_disconnect behind our back,
            explicitly mark it as closed with the memdebug macro. This libssh
            bug is fixed in 0.10.0. */
-        fake_sclose(conn->sock[FIRSTSOCKET]);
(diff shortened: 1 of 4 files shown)


lib/vssh/libssh.c
```diff
@@ -1965,8 +1965,7 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
         /* conn->sock[FIRSTSOCKET] is closed by ssh_disconnect behind our back,
            explicitly mark it as closed with the memdebug macro. This libssh
            bug is fixed in 0.10.0. */
-        fake_sclose(conn->sock[FIRSTSOCKET]);
-        conn->sock[FIRSTSOCKET] = CURL_SOCKET_BAD;
+        Curl_conn_forget_socket(data, FIRSTSOCKET);
```
bagder, line 1968: should the comment just before this, that explains the use of **fake_** sclose here not also be carried over to the new function? Otherwise it is hard to follow...

icing: You want me to change even more lines? Hmm...okay then!

Issue #12201, opened 2023-10-26
Return CURLE_WRITE_ERROR when CURLOPT_WRITEFUNCTION callback returns CURL_WRITEFUNC_ERROR


enwillyado:
### I did this

Make a `CURLOPT_WRITEFUNCTION` callback that returns `CURL_WRITEFUNC_ERROR` in order to **abort** the curl session.

For instance, if a large download need to be aborted without errors by user.

### I expected the following

![imagen](https://github.com/curl/curl/assets/16733813/527f8189-c053-4a4e-885c-e7f7873a9038)

The cURL session must return `CURLE_ABORTED_BY_CALLBACK`.

### curl/libcurl version

lib environment, last version

### operating system

Unix ARM.

icing:
So, what did you see instead? What is the failure you'd like to report?

enwillyado:
See the `chop_write` function:
https://github.com/curl/curl/blob/6ec70a9dd305e17b017f1283a034dfa1c03bbd7c/lib/sendf.c#L334

When a `CURLOPT_WRITEFUNCTION` callback called, the return size was stored in `wrote` variable. This function checks if the returned value was `CURL_WRITEFUNC_PAUSE`. If the special `CURL_WRITEFUNC_ERROR` return negative value is returned, a generic `CURLE_WRITE_ERROR`. Get the specific `CURLE_ABORTED_BY_CALLBACK` error is better in order to distinguish from a generic error.

enwillyado:
Probably this report should be a suggestion to have a `CURL_WRITEFUNC_ABORT` special returned value, right?

bagder:
I still don't understand. Your callback returns `CURL_WRITEFUNC_ERROR`.

What are you saying is returned and what are you saying you think *should* have been returned?

enwillyado:
Yes, literally if the `CURLOPT_WRITEFUNCTION` callback needs to return a abort, now only `CURL_WRITEFUNC_ERROR` can used. Then, if this special value was returned, the curl sessions ends with `CURLE_WRITE_ERROR` final code. This is the same final code if callbacks get an error (not memory or something like this)... in this case, by _returning a amount that differs from the amount passed_ by curl, is enough.

[...]

PR #12208, opened 2023-10-26
lib: CURL_WRITEFUNC_ABORT is a new write callback error

bagder:
... which makes libcurl return CURLE_ABORTED_BY_CALLBACK as error instead of CURLE_WRITE_ERROR.

Ref: #12201


docs/libcurl/opts/CURLOPT_WRITEFUNCTION.3
```diff
@@ -63,8 +63,11 @@ that amount differs from the amount passed to your callback function, it
 signals an error condition to the library. This causes the transfer to get
 aborted and the libcurl function used returns \fICURLE_WRITE_ERROR\fP.
 
-You can also abort the transfer by returning CURL_WRITEFUNC_ERROR (added in
-7.87.0), which makes \fICURLE_WRITE_ERROR\fP get returned.
+You can stop the transfer by returning CURL_WRITEFUNC_ERROR (added in 7.87.0),
+which makes \fICURLE_WRITE_ERROR\fP get returned.
+
+You can abort the transfer by returning CURL_WRITEFUNC_ABORT (added in 8.5.0),
```
jay, line 69: i think users may find this confusing. if stopping and aborting are the same thing except for the error code then i would refer to both of them one way or the other

bagder:
I'm having seconds thoughts on this PR. Does this really add enough value?

jay:
> I'm having seconds thoughts on this PR. Does this really add enough value?

As I understand it the user that requested this feature wants to be able to distinguish between a callback error and a write error caused by any other libcurl function. In practice I doubt that's much of an issue, judging by the results of `git grep CURLE_WRITE_ERROR lib`. OTOH I like it for consistency. The read function has CURL_READFUNC_ABORT => CURLE_ABORTED_BY_CALLBACK and so it makes sense to have CURL_WRITEFUNC_ABORT => CURLE_ABORTED_BY_CALLBACK.

I tagged this as needs-votes and we can wait for more feedback.

bagder:
No one has voted in favor, and I think most users already provide their own error reason in a private struct field already if they want one. I think this PR only makes adds new code path combinations for very little extra gain or use.

PR #12211, opened 2023-10-27
url: fix build when form API is disabled

mscdex:
I checked other places in the codebase where this same cleanup is done and they all seem to be guarded by the same #ifndef so I also changed that to match.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -455,8 +455,8 @@ CURLcode Curl_close(struct Curl_easy **datap)
   }
 #endif
 
+#ifndef CURL_DISABLE_FORM_API
   Curl_mime_cleanpart(data->state.formp);
-#ifndef CURL_DISABLE_HTTP
   Curl_safefree(data->state.formp);
 #endif
 


lib/url.c
```diff
@@ -455,8 +455,8 @@ CURLcode Curl_close(struct Curl_easy **datap)
   }
 #endif
 
+#ifndef CURL_DISABLE_FORM_API
```
bagder, line 458: Should this not rather check `CURL_DISABLE_MIME` ?

mscdex: At least with `./configure` if you try to build with the form API enabled but with MIME disabled, it fails with:

```
configure: error: MIME support needs to be enabled in order to enable form API support
```

So I don't think there is an issue in this case.

bagder: Ah yes, because the `formp` pointer is for the form API only.

However, this breaks the **linux ubuntu disable_http_smtp_imap) ** build because you removed the `CURL_DISABLE_HTTP` check.

bagder:
No response but still not building cleanly.

PR #12214, opened 2023-10-27
DOH, use PIPEWAIT when HTTP/2 is attempted

icing:
It this is more/less performant than using 2 connection (A and AAAA records) depends largely on network conditions and handshake time. But using only a single connection seems beneficial in any case.

This does not make any difference once the DOH connection is live in the cache, of course.


bagder (approved):
Good catch!

PR #12215, opened 2023-10-27
hostip: silence compiler warning `-Wparentheses-equality`

vszakats:
Seen with LLVM 17.

```
hostip.c:1336:22: warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
 1336 |        (a->ai_family == PF_INET)) {
      |         ~~~~~~~~~~~~~^~~~~~~~~~
hostip.c:1336:22: note: remove extraneous parentheses around the comparison to silence this warning
 1336 |        (a->ai_family == PF_INET)) {
      |        ~             ^         ~
hostip.c:1336:22: note: use '=' to turn this equality comparison into an assignment
 1336 |        (a->ai_family == PF_INET)) {
      |                      ^~
      |                      =
1 warning generated.
```

Follow-up to b651aba0962bb31353f55de4dc35f745952a1b10 #12145

Closes #12215

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -1333,9 +1333,9 @@ static void show_resolve_info(struct Curl_easy *data,
   while(a) {
     if(
 #ifdef CURLRES_IPV6
-       (a->ai_family == PF_INET6) ||
+       a->ai_family == PF_INET6 ||
 #endif
-       (a->ai_family == PF_INET)) {
+       a->ai_family == PF_INET) {
       char buf[MAX_IPADR_LEN];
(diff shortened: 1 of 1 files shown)


bagder (approved):

jay:
nothing against this change but imo extraneous parentheses warning is too strict a warning to be useful and i don't think we should have to adjust code that way. extra parentheses are subjective. they may or may not be helpful. for example they are sometimes helpful for readability or maybe macros use it so you end up with extra. i thought we weren't using the warning.

PR #12217, opened 2023-10-27
Makefile.mk: restore `_mingw.h` for default `_WIN32_WINNT`

### LINKED ISSUE #12134: Could not resolve IPv6 address of host

vszakats:
In 8.4.0 we deleted `_mingw.h` as part of purging old-mingw support.
Turns out `_mingw.h` had the side-effect of setting a default
`_WIN32_WINNT` value expected by `lib/config-win32.h` to enable
`getaddrinfo` support in `Makefile.mk` mingw-w64 builds. This caused
disabling support for this unless specifying the value manually.

Restore this header and update its comment to tell why we continue
to need it.

This triggered a regression in official Windows curl builds starting
with 8.4.0_1. Fixed in 8.4.0_6. (8.5.0 will be using CMake.)

Regression from 38029101e2d78ba125732b3bab6ec267b80a0e72 #11625

Reported-by: zhengqwe on github
Helped-by: Nico Rieck
Fixes #12134
Fixes #12136
Closes #12217

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -28,6 +28,11 @@
 #define CURL_NO_OLDIES
 #endif
 
+/* Set default _WIN32_WINNT */
+#ifdef __MINGW32__
+#include <_mingw.h>
+#endif
+
 /*
  * Disable Visual Studio warnings:
  * 4127 "conditional expression is constant"


bagder:
Awesome work on this @vszakats !

vszakats:
Thanks @bagder.

After [more investigation](https://github.com/curl/curl/pull/12136#issuecomment-1783288531) this seems to be the correct fix. An even more correct fix would be to include `windows.h` or similar, and make this work for all compilers. But this would probably be a regression for build performance. `windows.h` in mingw-w64 begins with #including `_mingw.h`, and ultimately ignoring `sdkddkver.h`.

PR #12218, opened 2023-10-27
wolfssl: add default case for wolfssl_connect_step1 switch

kareem-wolfssl:
Fixes ZD#16824.
Customer is using a strict compiler which requires default cases for all switch statements.


bagder (approved):

bagder:
Thanks!

PR #12224, opened 2023-10-28
Makefile.mk: drop Windows support

vszakats:
And DLL-support with it. This leaves `Makefile.mk` for MS-DOS and Amiga.

We recommend CMake instead. With unity mode it's much faster, and about
the same without.

Ref: https://github.com/curl/curl/pull/12221#issuecomment-1783761806
Closes #12224

---

- [x] Hold off till curl v8.6.0.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -481,7 +481,6 @@ Windows:
   - all:
       - changed-files:
           - any-glob-to-all-files:
-              - '**/Makefile.mk'
               - 'appveyor.yml'
               - 'CMake/Platforms/WindowsCache.cmake'
               - 'lib/*win32*'
--- a/Makefile.dist
+++ b/Makefile.dist
@@ -30,27 +30,6 @@ ssl:
 	./configure --with-openssl
 	make
 
-mingw32:
-	$(MAKE) -C lib -f Makefile.mk
-	$(MAKE) -C src -f Makefile.mk
-
-mingw32-clean:
-	$(MAKE) -C lib -f Makefile.mk clean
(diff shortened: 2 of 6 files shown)


bagder (approved):

jay:
@gvanem does this affect how you build curl?

gvanem:
@jay No and I have no objections to remove support for these.

firstrose:
Excuse me, but could you tell me how to build curl by mingw with cmake under windows now? Start from 8.6.0, the makefile doesn't work anymore. It confuses me before I see this. Thanks.

jay:
> Excuse me, but could you tell me how to build curl by mingw with cmake under windows now? Start from 8.6.0, the makefile doesn't work anymore. It confuses me before I see this. Thanks.

Original mingw is no longer supported. It may still work I haven't tried. mingw-w64 is supported. To have cmake generate mingw build files use `-G "MinGW Makefiles"`, for example

git clone https://github.com/curl/curl.git
mkdir curlbuild
cd curlbuild
cmake ../curl -G "MinGW Makefiles" -DCURL_USE_SCHANNEL=ON
cmake --build .

you will of course need the prerequisites for mingw gcc and make installed beforehand which should be as simple as open a MSYS2 MINGW64 shell prompt and `pacman -S mingw-w64-x86_64-toolchain`

also you can set the install path with `-DCMAKE_INSTALL_PREFIX:PATH=yourpathhere`

see also [INSTALL-CMAKE.md](https://github.com/curl/curl/blob/curl-8_7_1/docs/INSTALL-CMAKE.md) for cmake info

PR #12225, opened 2023-10-28
build: require Windows XP or newer

vszakats:
After this patch we assume availability of `getaddrinfo` and
`freeaddrinfo`, first introduced in Windows XP. Meaning curl
now requires building for Windows XP as a minimum.

TODO: assume these also in autotools.

Ref: https://github.com/curl/curl/pull/12221#issuecomment-1783761806
Closes #12225

---

https://github.com/curl/curl/pull/12225/files?w=1

### DIFF
--- a/CMake/Platforms/WindowsCache.cmake
+++ b/CMake/Platforms/WindowsCache.cmake
@@ -82,6 +82,10 @@ else()
   endif()
 endif()
 
+# Available in Windows XP and newer
+set(HAVE_GETADDRINFO 1)
+set(HAVE_FREEADDRINFO 1)
+
 set(HAVE_FCHMOD 0)
 set(HAVE_SOCKETPAIR 0)
 set(HAVE_SENDMSG 0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1087,22 +1087,24 @@ if(WIN32)
   # avoid storing HAVE_WIN32_WINNT in CMake cache
   unset(HAVE_WIN32_WINNT CACHE)
 
-  if(HAVE_WIN32_WINNT AND HAVE_WIN32_WINNT STRLESS "0x0501")
-    # Windows XP is required for freeaddrinfo, getaddrinfo
-    message(FATAL_ERROR "Building for Windows XP or newer is required.")
-  endif()
-
-  # pre-fill detection results that are based on target OS version
-  if(HAVE_WIN32_WINNT AND (MINGW OR MSVC))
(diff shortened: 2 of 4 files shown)


jay:
I think this is reasonable but here is some context. There is some version of Windows CE that was still supported by MS until earlier this month. curl survey says for the lesser used versions of Windows (page 11) that they are hardly used:

| Version        | 2023           |
| -------------- | -------------- |
|Windows XP|4.6%|
|Windows Server 2019|2.9%|
|Windows Server 2008|2.6%|
|Windows Vista|2.3%|
|Windows 2000|1.4%|
|Windows Server 2003|1.1%|
|Windows Server 2022|0.9%|
|Windows CE/Embedded|0.6%|
|Windows 98|0.6%|
|Windows 95|0.3%|

The sample size at 600+ is not definitive but is useful. Note the respondents may have selected multiple versions.

I suspect anyone using very old versions of Windows is probably using very old versions of curl, just because we don't get reports for anything older than XP. I wonder how many of those users are using a curl from 5 years ago because everything "just works".

For next year's survey I think it would be good to ask which version of curl the user is using. curl 8, curl 7, both, the latest, multiple versions, etc. maybe for those who build how often they build as well (I may have overlooked this if it's there already). There's a similar suggestion in the survey "How often do you update cURL and how do you receive updates?". @bagder

bagder:
@jay excellent suggestions. I have recorded these notes now, let's just make sure they are remembered when the user survey time comes again.

bagder (approved):
I'm fine with bumping the Windows requirement to XP.

vszakats:
Not directly related, but, may I ask if curl indends to support Windows CE? [ esp. in the light that it has been deprecated by MS as @jay mentioned. Code has references to it, but it's rarely if ever mentioned. ]

vszakats:
Merged.

I'd be interesed in the Window CE support status regardless!

Also this leaves a pending TODO for autotools, where we might also avoid detecting these features for Windows and assume them.

jay:
> I'd be interesed in the Window CE support status regardless!

AFAICT we don't have a standard build process for it. My guess is it's embedded device manufacturers that use it and don't contribute back. Maybe they build it using Visual Studio 2008 and winbuild? It appears possible to build [OpenSSL](https://qiita.com/souju/items/94117c024862f57459c3) and [wolfSSL](https://www.wolfssl.com/forums/topic1193-how-to-compile-for-windows-ce-5.html) (at least in 2019) for Windows CE as well.

The problem is we don't know how it's being done, if it's even still being done or what to support.

vszakats:
Thanks @jay!

PR #12227, opened 2023-10-29
build: fix compiler warning with auths disabled

vszakats:
```
./curl/lib/http.c:979:12: warning: unused function 'is_valid_auth_separator' [-Wunused-function]
static int is_valid_auth_separator(char ch)
           ^
5 warnings generated.
```

Follow-up to e92edfbef64448ef461117769881f3ed776dec4e #11490

Closes #12227


vszakats:
This first iteration is also an incorrect fix for controlling the if-tree.

PR #12233, opened 2023-10-30
libcurl-thread.3: simplify the TLS section

bagder:
All TLS libraries curl can use are threadsafe since OpenSSL 1.1.x, August 2016.

### DIFF
--- a/docs/libcurl/libcurl-thread.3
+++ b/docs/libcurl/libcurl-thread.3
@@ -42,8 +42,10 @@ interface but you must provide your own locking and set
 Note that some items are specifically documented as not thread-safe in the
 share API (the connection pool and HSTS cache for example).
 .SH TLS
-All current TLS libraries libcurl supports are thread-safe. That was not
-always the case in the past.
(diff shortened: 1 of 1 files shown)


docs/libcurl/libcurl-thread.3
```diff
@@ -42,37 +42,8 @@ interface but you must provide your own locking and set
...
-engine is used by libcurl in a way that is fully thread-safe.
-
-https://www.openssl.org/docs/man1.1.0/man3/CRYPTO_THREAD_run_once.html#DESCRIPTION
-
-OpenSSL <= 1.0.2 the user must set callbacks.
-
-https://www.openssl.org/docs/man1.0.2/man3/CRYPTO_set_locking_callback.html#DESCRIPTION
-
-https://curl.se/libcurl/c/opensslthreadlock.html
-
```
jay, line 61: I think we should leave the OpenSSL section. The main paragraph could be rewritten "All current TLS libraries libcurl supports are thread-safe. For OpenSSL there are some caveats:

PR #12237, opened 2023-10-31
late clone of connection ssl config

icing:
- perform connection cache matching against `data->set.ssl.primary` and proxy counterpart
- fully clone connection ssl config only when connection is used


bagder (approved):
Makes sense. Please take a look at the merge conflict too!

Issue #10129, opened 2022-12-21
aws-sigv4 not correctly signing requests with empty URL query params


hjmallon:
### I did this

```
URL="https://${BUCKET}.s3.${REGION}.amazonaws.com/?versioning"
curl -v \
     --aws-sigv4 "aws:amz:${REGION}:s3" \
     --get "${URL}" \
     --user "${aws_access_key_id}:${aws_secret_access_key}" \
     --header "X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" \
     --header "X-Amz-Security-Token: ${aws_session_token}"
```

### I expected the following

With the proper setup of `BUCKET`, `REGION`, `aws_access_key_id`, `aws_secret_access_key` and `aws_session_token` I expect `200 OK` and data. See https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketVersioning.html

However I got `403 Forbidden` and an error returned `SignatureDoesNotMatch`.

I have tracked it down a bit further and it it related to the `?versioning` parameter. If I set it to `?versioning=` then it works. AWS CLI seems to be doing that automatically, expanding `versioning` into `versioning=`.

So this works

```
URL="https://${BUCKET}.s3.${REGION}.amazonaws.com/?versioning="
curl -v \
     --aws-sigv4 "aws:amz:${REGION}:s3" \
     --get "${URL}" \
     --user "${aws_access_key_id}:${aws_secret_access_key}" \
     --header "X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" \
     --header "X-Amz-Security-Token: ${aws_session_token}"
```

### curl/libcurl version

```
curl 7.87.0-DEV (x86_64-apple-darwin22.2.0) libcurl/7.87.0-DEV (SecureTransport) OpenSSL/1.1.1s zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.4 libssh2/1.10.0 nghttp2/1.51.0 librtmp/2.3
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
[...]

Issue #8107, opened 2021-12-06
--aws-sigv4 doesn't sign requests with query string parameters with no value correctly


dzaslavskiy:
This request gives an error:
```
curl --user "<key>:<secret>" --aws-sigv4 aws:amz:<region>:es -k -X GET 'https://<es-host>/_cat/indices?v'

{"message":"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details."}
```

Removing the query string or or adding a value to the query parameter fixes the issue 
```
curl --user "<key>:<secret>" --aws-sigv4 aws:amz:<region>:es -k -X GET 'https://<es-host>/_cat/indices'

green open index-name index-uuid 5 1       16    0 285.4kb 142.7kb
```
```
curl --user "<key>:<secret>" --aws-sigv4 aws:amz:<region>:es -k -X GET 'https://<es-host>/_cat/indices?v=true'

green open index-name index-uuid 5 1       16    0 285.4kb 142.7kb
```

### curl/libcurl version

curl 7.79.1 (x86_64-w64-mingw32) libcurl/7.79.1 OpenSSL/1.1.1l (Schannel) zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.1 libssh2/1.10.0 nghttp2/1.45.1
Release-Date: 2021-09-22
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI TLS-SRP zstd

### operating system

windows 10; git-bash

outscale-mgo:
I'm presently trying to find the bug, out of curiosity, which call exactly did you made (having the full URL could held debug I guess) ?

dzaslavskiy:
```
$ ./curl --user "..." --aws-sigv4 aws:amz:us-east-1:es -X GET "https://search-curl-test-yk7vggewbojld4psjkwqnjifry.us-east-1.es.amazonaws.com/_cat/indices?v=true"

health status index     uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   .kibana_1 Q9Me-vJiQXGZwplpTsTxpg   1   0          1            0        5kb            5kb

[...]

Issue #7559, opened 2021-08-12
--aws-sigv4 doesn't sign requests with `*` correctly


poborin:
### I did this
my idiomatic request to AWS Elasticsearch Service looks as following:

```shell
curl --request GET "https://${ES_DOMAIN_ENDPOIN}/tempo_data-*/event/_mapping" \
--user $AWS_ACCESS_KEY:$AWS_SECRET_KEY \
--aws-sigv4 "aws:amz:ap-southeast-2:es"
```

here is the verbose output:
```text
*   Trying 13.55.161.62:443...
* Connected to ES_DOMAIN_ENDPOIN (13.55.161.62) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=*.ap-southeast-2.es.amazonaws.com
*  start date: Feb 18 00:00:00 2021 GMT
*  expire date: Mar 19 23:59:59 2022 GMT
*  subjectAltName: host "ES_DOMAIN_ENDPOIN" matched cert's "*.ap-southeast-2.es.amazonaws.com"
*  issuer: C=US; O=Amazon; OU=Server CA 1B; CN=Amazon
*  SSL certificate verify ok.
* Using HTTP2, server supports multiplexing
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Server auth using AWS_SIGV4 with user 'AWS_ACCESS_KEY'
* Using Stream ID: 1 (easy handle 0x7f83b4813e00)
> GET /tempo_data-*/event/_mapping HTTP/2
> Host: ES_DOMAIN_ENDPOIN
[...]

PR #12244, opened 2023-11-01
http_aws_sigv4: canonicalise valueless query params

hjmallon:
Query params with `?novalparam` (i.e. no =) need to be given an empty value while canonicalising

From https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

```
When a request targets a subresource, the corresponding query parameter value will be an empty string (""). For example, the following URI identifies the ACL subresource on the examplebucket bucket:

http://s3.amazonaws.com/examplebucket?acl 

The CanonicalQueryString in this case is as follows:

UriEncode("acl") + "=" + ""
```

This fixes my issue https://github.com/curl/curl/issues/10129 and I think this other issue is a dupe (?) https://github.com/curl/curl/issues/8107.

In this branch I have also cleaned up some other things in `KNOWN_BUGS` (I can make a different MR if that is easier) ~~ ~and also [http_aws_sigv4: reading too far in canon_query](https://github.com/curl/curl/commit/81272ab30104cc71f13814070a29e8d3c5ac3aee) is a missing length check (I think, but there might be some other guarantee somewhere that we can always read 2 chars after a `%` that I haven't seen.~

EDIT: Removed extra commit with length check as it is not relevant!

### DIFF
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -101,7 +101,6 @@ problems may have been fixed or changed somewhat since this was written.
 
  16. aws-sigv4
  16.1 aws-sigv4 does not sign requests with * correctly
- 16.3 aws-sigv4 is missing the amz-content-sha256 header
  16.6 aws-sigv4 does not behave well with AWS VPC Lattice
 
  17. HTTP/2
@@ -574,10 +573,6 @@ problems may have been fixed or changed somewhat since this was written.
 16.1 aws-sigv4 does not sign requests with * correctly
 
  https://github.com/curl/curl/issues/7559
-
-16.3 aws-sigv4 is missing the amz-content-sha256 header
-
- https://github.com/curl/curl/issues/8810
 
 16.6 aws-sigv4 does not behave well with AWS VPC Lattice
 
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -103,7 +103,6 @@ problems may have been fixed or changed somewhat since this was written.
  16.1 aws-sigv4 does not sign requests with * correctly
  16.2 aws-sigv4 does not sign requests with valueless queries correctly
  16.3 aws-sigv4 is missing the amz-content-sha256 header
- 16.4 aws-sigv4 does not sort query string parameters before signing
  16.5 aws-sigv4 does not sign requests with empty URL query correctly
  16.6 aws-sigv4 does not behave well with AWS VPC Lattice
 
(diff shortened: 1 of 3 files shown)


lib/http_aws_sigv4.c
```diff
@@ -497,6 +502,10 @@ static CURLcode canon_query(struct Curl_easy *data,
         }
       }
     }
+    if(!found_equals) {
+      /* queries without value still need an equals */
+      result = Curl_dyn_addn(dq, "=", 1);
```
bagder, line 507:
```suggestion
      result = Curl_dyn_addn(dq, "=", 1);
      if(result)
        return result:
```

... because otherwise it might be overwritten below and still return OK in an error situation.

hjmallon: The loop checks the result every time and bails if it had failed, should we continue trying to add chars in the case of failure? If not maybe the patch should be 

```
    if(!found_equals && !result) {
      /* queries without value still need an equals */
      result = Curl_dyn_addn(dq, "=", 1);
    }
    if(i < entry - 1 && !result) {
      /* insert ampersands between query pairs */
      result = Curl_dyn_addn(dq, "&", 1);
    }
```

hjmallon: Could the existing code overwrite an error from the loop in the `/* insert ampersands between query pairs */` case?

docs/KNOWN_BUGS
```diff
@@ -103,7 +103,6 @@ problems may have been fixed or changed somewhat since this was written.
  16.1 aws-sigv4 does not sign requests with * correctly
  16.2 aws-sigv4 does not sign requests with valueless queries correctly
  16.3 aws-sigv4 is missing the amz-content-sha256 header
```
outscale-mgo, line 104: This one should be removed too, as it was fix here: https://github.com/curl/curl/pull/9995/files
note that sigv4 require amz-content-sha256 only with s3 (and maybe a few other API), but not always, so adding it all the time would have been a mistake.

outscale-mgo: Also, we could add a line:
`aws-sigv4 does not canonize URL path`

hjmallon: I have removed 16.3, but haven't added the new one (they seem to need refs to issues etc, maybe you have one?)

outscale-mgo: I think the doc say that path needs to be canon: https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html
but I guess, that what happen here: https://github.com/curl/curl/issues/7559 just that the problem doesn't end with '*'.
Also, sorry if my comments are not pertinent, I'm slightly sick right now.

lib/http_aws_sigv4.c
bagder: yes, it can, so your proposed fix looks fine

docs/KNOWN_BUGS
hjmallon: Hope you feel better soon! I'm sure its a valid issue, but as I haven't looked into it I would rather not add the `KNOWN_ISSUE` in this MR (where commit history etc might cause someone to ask me about it 😀)

bagder (approved):

bagder:
Thanks!

PR #12246, opened 2023-11-02
lib: add CURLOPT_DEBUG_CONNCACHE

jay:
- Add CURLOPT_DEBUG_CONNCACHE to debug the connection cache.

This option enables debugging the internal connection cache by copying debug settings to the connection cache when the easy handle is added to the multi handle.

The following settings are inherited by the connection cache handle:

CURLOPT_DEBUGFUNCTION, CURLOPT_DEBUGDATA, CURLOPT_STDERR and CURLOPT_VERBOSE.

The copied settings remain even if the option is later disabled. The set data must survive until any multi and share handles used by the easy handle are cleaned up and the easy handle is cleaned up.

Ref: https://github.com/curl/curl/discussions/11878
Ref: https://github.com/curl/curl/pull/12144

Closes #xxxx


docs/libcurl/opts/CURLOPT_DEBUG_CONNCACHE.3
```diff
@@ -0,0 +1,98 @@
...
+.\" * furnished to do so, under the terms of the COPYING file.
+.\" *
+.\" * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+.\" * KIND, either express or implied.
+.\" *
+.\" * SPDX-License-Identifier: curl
+.\" *
+.\" **************************************************************************
+.\"
+.TH CURLOPT_DEBUG_CONNCACHE 3 "31 Oct 2023" libcurl libcurl
```
bagder, line 25: I think the option name can use some improvement. This option sets debug options for the internally created easy handle used to shut down connections when there is no active easy handle to use.

`CURLOPT_DEBUG_CONNCLOSE` maybe?

docs/libcurl/opts/CURLOPT_DEBUG_CONNCACHE.3
```diff
@@ -0,0 +1,98 @@
...
+.SH DEFAULT
+0
+.SH PROTOCOLS
+All
+.SH EXAMPLE
+.nf
+CURL *curl = curl_easy_init();
+if(curl) {
+  /* CURLOPT_DEBUG_CONNCACHE copies debug settings to every
+   * internal connection cache used by the easy handle:
```
bagder, line 73: The connection cache is always just there when there is a handle, I don't think we should speak of it getting created post-init. It is created when the easy or multi handle is created.

This option makes libcurl copy debug properties/settings from the easy handle to the closure handle when this easy handle is added to the multi handle (when either `curl_multi_add_handle` or `curl_easy_perform` are called)

bagder:
Reading this PR another idea popped into my head. What if we instead could extract the internal closure handle? Then we could just do ordinary *setopt() calls on that to set the debug options?

~~~c
CURL *internal;
curl_easy_getinfo(handle, CURLINFO_INTERNAL_HANDLE_FOR_CLOSING_CONNETIONS, &internal);
curl_easy_setopt(internal, CURLOPT_VERBOSE, 1L);

docs/libcurl/opts/CURLOPT_DEBUG_CONNCACHE.3
```diff
@@ -0,0 +1,98 @@
...
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+
+CURLcode curl_easy_setopt(CURL *handle, CURLOPT_DEBUG_CONNCACHE, long onoff);
+.fi
+.SH DESCRIPTION
+This option enables debugging the internal connection cache associated with
+your easy handle. Several debug settings are copied from \fIhandle\fP to every
+associated internal connection cache:
```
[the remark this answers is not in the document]
jay: > I don't quite understand this description: _every connection cache_. There is only one connection cache per easy handle.

Yes but over the lifetime of the easy handle it may be associated with multiple connection caches and they're all going to receive the copied settings. I take your point though here and your other comments, the wording could use some improvement.

jay:
> Reading this PR another idea popped into my head. What if we instead could extract the internal closure handle? Then we could just do ordinary *setopt() calls on that to set the debug options?

That puts the burden on the user _when_ to have to do that, and since the closure handle may change I don't think it would be easy.

docs/libcurl/opts/CURLOPT_DEBUG_CONNCACHE.3
icing: Just a side note: the conncache's internal handle is now created when the conncache is set up, but it is only used during closedown. It could just exist during the closedown call, so that for 99% of the multi operations, it would not be there at all.

icing:
> > Reading this PR another idea popped into my head. What if we instead could extract the internal closure handle? Then we could just do ordinary *setopt() calls on that to set the debug options?
> 
> That puts the burden on the user _when_ to have to do that, and since the closure handle may change I don't think it would be easy.

The people using this with `DEBUGDATA` of their own need to be aware that their `void*` needs to exist at least until the multi handle/share is cleaned up. And that the `easy` handle appearing in their `DEBUGFUNCTION` may not be one of their own.

Lifetime handling is all a bit messy with conncache. The internal handle is a crutch. I am not sure if @bagder 's idea of making the internal handle accessible would help. It is "simpler", but then there are millions of things one is not supposed to do with the internal handle. I have the feeling that we would be fixing stuff several times afterwards due to people being creative.

docs/libcurl/opts/CURLOPT_DEBUG_CONNCACHE.3
bagder: > It could just exist during the closedown call, so that for 99% of the multi operations, it would not be there at all.

Agreed. In fact, I would not mind experimenting with only creating it if truly needed, as if there are easy handles existing when `curl_multi_cleanup()` is called, one of those handles could perhaps be used for driving the shutdowns. Instead of creating an extra internal easy handle just for that purpose.

bagder:
Yeah, my idea does not really hold any closer scrutiny..

docs/libcurl/opts/CURLOPT_DEBUG_CONNCACHE.3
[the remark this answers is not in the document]
bagder: Since it is still always only one cache at a time I think "the cache" is more apt than "every cache".

bagder:
I think the biggest downside with this proposal is the strange life time effect. It feel like it is just a bit too hard for a user/application to figure out how long data needs to live. I think Stefan's #12144 has a better story there where it is more defined, but with the downside that it requires quite a lot of new options.

During discussions today, another idea came up that might be worth exploring:

## An explicit default/inherit handle

We create an API where the application can create an easy handle, set options to it and then set *that* handle as a "default handle". Then the closure handle could inherit its debug options and we could use that to set other options' default values.

~~~c
CURL *easy = curl_easy_init();
curl_easy_setopt(easy, CURLOPT_VERBOSE, 1L);
curl_multi_setopt(multi, CURLMOPT_DEFAULT_HANDLE, easy); /* name TBD */
~~~

This can then be used internally as a closure handle...

bagder:
I think we are still looking for the best take on solving this problem.

PR #12254, opened 2023-11-02
strdup: don't allow Curl_strndup to read past a null terminator

jay:
- Use malloc + strncpy instead of Curl_memdup to dupe the string before null terminating it.

Prior to this change if Curl_strndup was passed a length longer than the allocated string then it could copy out of bounds.

This change is for posterity. Curl_strndup was just introduced today and none of the current calls to it pass a length longer than allocated length of the input.

Follow-up to d3b3ba35.

Closes #xxxx

### DIFF
--- a/lib/strdup.c
+++ b/lib/strdup.c
@@ -103,18 +103,20 @@ void *Curl_memdup(const void *src, size_t length)
  *
  * Curl_strndup(source, length)
  *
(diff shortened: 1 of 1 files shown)


bagder:
I selected this functionality on purpose since this is the use case we have in existing source code. I don't think we currently need a "stop at the null byte" feature anywhere... But I don't object to this change.

PR #12180, opened 2023-10-22
quiche: use a quiche_conn_peer_transport_params()

junhochoi:
- There was quiche API changes separating quiche_stats and quiche_transport_params(). Use a new API.
- Bump quiche version to master to test it in CI.

PR #12260, opened 2023-11-03
Quiche compile fix

gaby64:
Compiles with quiche github master, I dont expect this to be merged, just making it available to those looking for a solution if they encounter the same compiling issues.


bagder:
Duplicate of #12180 

We still await a quiche release that uses this before we merge. So that we can build with the latest release.

PR #12261, opened 2023-11-03
schannel: CA file and memory blob cache

ajax16384:
add schannel CA file and memory blob cache support
add schannel CURLOPT_CA_CACHE_TIMEOUT support

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -2810,13 +2810,8 @@ HCERTSTORE Curl_schannel_get_cached_cert_store(struct Curl_cfilter *cf,
     }
   }
   else {
-    if(!conn_config->CAfile) {
-      return NULL;
-    }
-    if(!mbackend->CAfile) {
-      return NULL;
-    }
-    if(strcmp(mbackend->CAfile, conn_config->CAfile)) {
+    if(!conn_config->CAfile || !mbackend->CAfile ||
+       strcmp(mbackend->CAfile, conn_config->CAfile)) {
       return NULL;
     }
   }
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -2799,7 +2799,7 @@ HCERTSTORE Curl_schannel_get_cached_cert_store(struct Curl_cfilter *cf,
     if(mbackend->CAinfo_blob_size != ca_info_blob->len) {
       return NULL;
     }
-    schannel_sha256sum((const char *)ca_info_blob->data,
+    schannel_sha256sum((const unsigned char *)ca_info_blob->data,
                        ca_info_blob->len,
                        info_blob_digest,
                        CURL_SHA256_DIGEST_LENGTH);
@@ -2858,7 +2858,7 @@ bool Curl_schannel_set_cached_cert_store(struct Curl_cfilter *cf,
     if(!CAinfo_blob_digest) {
       return false;
     }
-    schannel_sha256sum((const char *)ca_info_blob->data,
+    schannel_sha256sum((const unsigned char *)ca_info_blob->data,
                        ca_info_blob->len,
                        CAinfo_blob_digest,
                        CURL_SHA256_DIGEST_LENGTH);
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -2754,10 +2754,20 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
   return &backend->ctxt->ctxt_handle;
 }
 
-HCERTSTORE schannel_get_cached_cert_store(struct Curl_cfilter *cf,
-                                          const struct Curl_easy *data)
+HCERTSTORE Curl_schannel_get_cached_cert_store(struct Curl_cfilter *cf,
+                                               const struct Curl_easy *data)
 {
+  struct ssl_primary_config *conn_config = Curl_ssl_cf_get_primary_config(cf);
   struct Curl_multi *multi = data->multi_easy ? data->multi_easy : data->multi;
+  const struct curl_blob *ca_info_blob = conn_config->ca_info_blob;
+  struct schannel_multi_ssl_backend_data *mbackend;
+  const struct ssl_general_config *cfg = &data->set.general_ssl;
+  timediff_t timeout_ms;
+  timediff_t elapsed_ms;
+  struct curltime now;
+  unsigned char info_blob_digest[CURL_SHA256_DIGEST_LENGTH];
+
+  DEBUGASSERT(multi);
 
   if(!multi) {
     return NULL;
@@ -2767,25 +2777,20 @@ HCERTSTORE schannel_get_cached_cert_store(struct Curl_cfilter *cf,
     return NULL;
   }
 
-  struct schannel_multi_ssl_backend_data *mbackend;
   mbackend = (struct schannel_multi_ssl_backend_data *)multi->ssl_backend_data;
   if(!mbackend->cert_store) {
     return NULL;
   }
 
   /* use cached_x509_store_expired timediff calculation pattern */
-  const struct ssl_general_config *cfg = &data->set.general_ssl;
-  timediff_t timeout_ms = cfg->ca_cache_timeout * (timediff_t)1000;
+  timeout_ms = cfg->ca_cache_timeout * (timediff_t)1000;
   if(timeout_ms >= 0) {
-    struct curltime now = Curl_now();
-    timediff_t elapsed_ms = Curl_timediff(now, mbackend->time);
+    now = Curl_now();
+    elapsed_ms = Curl_timediff(now, mbackend->time);
     if(elapsed_ms >= timeout_ms) {
       return NULL;
     }
   }
-
-  struct ssl_primary_config *conn_config = Curl_ssl_cf_get_primary_config(cf);
-  const struct curl_blob *ca_info_blob = conn_config->ca_info_blob;
 
   if(ca_info_blob) {
     if(!mbackend->CAinfo_blob_digest) {
@@ -2794,7 +2799,6 @@ HCERTSTORE schannel_get_cached_cert_store(struct Curl_cfilter *cf,
     if(mbackend->CAinfo_blob_size != ca_info_blob->len) {
       return NULL;
     }
-    unsigned char info_blob_digest[CURL_SHA256_DIGEST_LENGTH];
     schannel_sha256sum((const char *)ca_info_blob->data,
                        ca_info_blob->len,
                        info_blob_digest,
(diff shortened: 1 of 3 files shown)


lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+    }
+    if(strcmp(mbackend->CAfile, conn_config->CAfile)) {
+      return NULL;
+    }
+  }
+
+  return mbackend->cert_store;
+}
+
+bool schannel_set_cached_cert_store(struct Curl_cfilter *cf,
```
bagder, line 2823: This is not a static function, then it needs to be prefixed with `Curl_`.

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
   return &backend->ctxt->ctxt_handle;
 }
 
+HCERTSTORE schannel_get_cached_cert_store(struct Curl_cfilter *cf,
```
bagder, line 2757: As this is not a static function, then it needs to be prefixed with `Curl_`.

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+
+  if(!multi) {
+    return NULL;
+  }
+
+  if(!multi->ssl_backend_data) {
+    return NULL;
+  }
+
+  struct schannel_multi_ssl_backend_data *mbackend;
```
bagder, line 2770: We write C89 code, so we prefer the variable declarations at the top of every code block. Even for this code, I presume there is no compiler that is going to warn about it in this windows specific section.

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+    }
+    unsigned char info_blob_digest[CURL_SHA256_DIGEST_LENGTH];
+    schannel_sha256sum((const char *)ca_info_blob->data,
+                       ca_info_blob->len,
+                       info_blob_digest,
+                       CURL_SHA256_DIGEST_LENGTH);
+    if(memcmp(mbackend->CAinfo_blob_digest,
+              info_blob_digest,
+              CURL_SHA256_DIGEST_LENGTH)) {
+        return NULL;
```
bagder, line 2805: What is the purpose of this check?

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+  return mbackend->cert_store;
+}
+
+bool schannel_set_cached_cert_store(struct Curl_cfilter *cf,
+                                    const struct Curl_easy *data,
+                                    HCERTSTORE cert_store)
+{
+  struct Curl_multi *multi = data->multi_easy ? data->multi_easy : data->multi;
+
+  if(!multi) {
```
bagder, line 2829: I propose a DEBUGASSERT() here as well for the same reason. This should not be possible to happen.

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+                       ca_info_blob->len,
+                       CAinfo_blob_digest,
+                       CURL_SHA256_DIGEST_LENGTH);
+    CAinfo_blob_size = ca_info_blob->len;
+  }
+  else {
+    if(conn_config->CAfile) {
+      CAfile = strdup(conn_config->CAfile);
+      if(!CAfile) {
+        return false;
```
bagder, line 2866: In this and the other early returns, do they not leak memory allocated previously in this function?

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
   return &backend->ctxt->ctxt_handle;
 }
 
+HCERTSTORE schannel_get_cached_cert_store(struct Curl_cfilter *cf,
+                                          const struct Curl_easy *data)
+{
+  struct Curl_multi *multi = data->multi_easy ? data->multi_easy : data->multi;
+
```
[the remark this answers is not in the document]
ajax16384: used same pattern from https://github.com/curl/curl/blob/3e6254f819070eb82c72e94c4469d9baaddfe2f4/lib/vtls/openssl.c#L3377

as it checked multi for null without assert.

lib/vtls/schannel.c
ajax16384: OK

lib/vtls/schannel.c
ajax16384: OK

lib/vtls/schannel.c
ajax16384: OK

lib/vtls/schannel.c
ajax16384: memcmp compares hash of previous mem block with current, if it's not equal then cert_store is invalid

lib/vtls/schannel.c
ajax16384: all possible allocations will be freed at backend free method (free_multi_ssl_backend_data)

lib/vtls/schannel.c
ajax16384: OK

lib/vtls/schannel.c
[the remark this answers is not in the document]
bagder: it is now!

bagder:
[warnings](https://ci.appveyor.com/project/curlorg/curl/builds/48438286/job/ulprlnp5ijgmh4e0#L330)

~~~
C:\projects\curl\lib\vtls\schannel.c(2802,50): warning C4057: 'function': 'const unsigned char *' differs in indirection to slightly different base types from 'const char *' [C:\projects\curl\lib\libcurl_object.vcxproj]
C:\projects\curl\lib\vtls\schannel.c(2861,50): warning C4057: 'function': 'const unsigned char *' differs in indirection to slightly different base types from 'const char *' [C:\projects\curl\lib\libcurl_object.vcxproj]

ajax16384:
> [warnings](https://ci.appveyor.com/project/curlorg/curl/builds/48438286/job/ulprlnp5ijgmh4e0#L330)
> 
> ```
> C:\projects\curl\lib\vtls\schannel.c(2802,50): warning C4057: 'function': 'const unsigned char *' differs in indirection to slightly different base types from 'const char *' [C:\projects\curl\lib\libcurl_object.vcxproj]
> C:\projects\curl\lib\vtls\schannel.c(2861,50): warning C4057: 'function': 'const unsigned char *' differs in indirection to slightly different base types from 'const char *' [C:\projects\curl\lib\libcurl_object.vcxproj]
> ```

thanks, fixed

lib/vtls/schannel.c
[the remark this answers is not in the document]
ajax16384: OK

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,153 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+  if(ca_info_blob) {
+    CAinfo_blob_digest = malloc(CURL_SHA256_DIGEST_LENGTH);
+    if(!CAinfo_blob_digest) {
+      return false;
+    }
+    schannel_sha256sum((const unsigned char *)ca_info_blob->data,
+                       ca_info_blob->len,
+                       CAinfo_blob_digest,
+                       CURL_SHA256_DIGEST_LENGTH);
```
jay, line 2864: is this really faster than storing the original blob and comparing?

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,153 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+  else {
+    if(!conn_config->CAfile) {
+      return NULL;
+    }
+    if(!mbackend->CAfile) {
+      return NULL;
+    }
+    if(strcmp(mbackend->CAfile, conn_config->CAfile)) {
+      return NULL;
+    }
```
jay, line 2821:
```suggestion
    if(!conn_config->CAfile || !mbackend->CAfile ||
       strcmp(mbackend->CAfile, conn_config->CAfile))
      return NULL;
```
several lines like this should be consolidated

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,153 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+
+  /* free old cache data */
+  if(mbackend->cert_store) {
+    CertCloseStore(mbackend->cert_store, 0);
+  }
+  free(mbackend->CAinfo_blob_digest);
+  free(mbackend->CAfile);
+
+  mbackend->time = Curl_now();
+  mbackend->cert_store = cert_store;
```
jay, line 2884: I notice that unlike openssl code there's no increase to a reference count when this is added

ajax16384: Here(schannnel*.c) HCERTSTORE resource use non refcount contructor/destrctor pattern. Just after successfull Curl_schannel_set_cached_cert_store call from Curl_verify_certificate - cache record became this resource owner. Lifetime of cert_store do not exceed get/store* methods.

lib/vtls/schannel.c
ajax16384: blob might be memory expensive
(like https://curl.se/docs/caextract.html - 200KB)

lib/vtls/schannel.c
ajax16384: OK

lib/vtls/schannel_verify.c
```diff
@@ -630,31 +631,46 @@ CURLcode Curl_verify_certificate(struct Curl_cfilter *cf,
...
-                                  (HCRYPTPROV)NULL,
-                                  CERT_STORE_CREATE_NEW_FLAG,
-                                  NULL);
-      if(!trust_store) {
-        char buffer[STRERROR_LEN];
-        failf(data, "schannel: failed to create certificate store: %s",
-              Curl_winapi_strerror(GetLastError(), buffer, sizeof(buffer)));
-        result = CURLE_SSL_CACERT_BADFILE;
+      /* try cache */
+      trust_store = Curl_schannel_get_cached_cert_store(cf, data);
```
jay, line 635: openssl get/set of cert store has more conditionals, see `cache_criteria_met`. it looks like most? of those don't apply here but what about data->set.general_ssl.ca_cache_timeout

https://github.com/curl/curl/blob/3e6254f819070eb82c72e94c4469d9baaddfe2f4/lib/vtls/openssl.c#L3439-L3460

lib/vtls/schannel.c
```diff
@@ -2754,6 +2754,148 @@ static void *schannel_get_internals(struct ssl_connect_data *connssl,
...
+  struct multi_ssl_backend_data *msbd)
+{
+  struct schannel_multi_ssl_backend_data *mbackend =
+    (struct schannel_multi_ssl_backend_data*)msbd;
+  if(mbackend->cert_store) {
+    CertCloseStore(mbackend->cert_store, 0);
+  }
+  free(mbackend->CAinfo_blob_digest);
+  free(mbackend->CAfile);
+  free(mbackend);
```
jay, line 2896: is mbackend guaranteed to exist when this function is called?

lib/vtls/schannel.c
jay: @tlsa do you happen to remember why you used ref counting for the openssl store

lib/vtls/schannel.c
jay: > is mbackend guaranteed to exist when this function is called?

nm it looks like openssl does the same thing because the vtls function checks that mbackend exists before calling

lib/vtls/schannel.c
jay: In a release configuration in an average of over 100 runs a memcmp of the 200k memory takes 24us and a sha256sum comparison takes 3200us (1600us x 2). That is basically 3 ms difference between the two. I did not account for the memdup time in the first case though I forgot about that. So first case might actually be slower because of crt locks needed to copy the memory. My conclusion is it's not worth copying the entire blob.

lib/vtls/schannel_verify.c
ajax16384: Thanks. Zero cache timeout should skip cache usage at all. Will fix with next commit.

jay:
Thanks

Issue #12228, opened 2023-10-29
compiler warnings in http.c when all auth methods disabled


vszakats:
### I did this

Built non-Windows (e.g. Linux) curl with CMake options:
```
-DCURL_DISABLE_BASIC_AUTH=ON
-DCURL_DISABLE_BEARER_AUTH=ON
-DCURL_DISABLE_DIGEST_AUTH=ON
-DCURL_DISABLE_NEGOTIATE_AUTH=ON
-DCURL_DISABLE_NTLM=ON
```

```
./curl/lib/http.c:737:12: warning: unused variable 'result' [-Wunused-variable]
  CURLcode result = CURLE_OK;
           ^
./curl/lib/http.c:995:18: warning: variable 'availp' set but not used [-Wunused-but-set-variable]
  unsigned long *availp;
                 ^
./curl/lib/http.c:996:16: warning: variable 'authp' set but not used [-Wunused-but-set-variable]
  struct auth *authp;
               ^
```

Ref:
0d3956b8c6558b3077fc7aabbf853b86058becd1 #11895
e92edfbef64448ef461117769881f3ed776dec4e #11490

This one fixed by: #12262
```
./curl/lib/http.c:1143:12: warning: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Wextra-semi-stmt]
           ;
           ^
```

### I expected the following

No compiler warnings.

### curl/libcurl version

8.5.0-DEV master

### operating system

any

bagder:
Fixed in #12262

vszakats:
#12262 fixed one out of 4. #12272 will fix the remaining 3. (though it's fine to continue tracking this there)

vszakats:
Reopening because of the unfixed warnings.

vszakats:
Sorry, this is a more correct set of options to trigger the warnings:
```
-DCURL_DISABLE_BASIC_AUTH=ON
-DCURL_DISABLE_BEARER_AUTH=ON
-DCURL_DISABLE_DIGEST_AUTH=ON
-DCURL_DISABLE_KERBEROS_AUTH=ON
-DCURL_DISABLE_NEGOTIATE_AUTH=ON
-DCURL_DISABLE_AWS=ON
-DCURL_DISABLE_NTLM=ON
```
This is on macOS.

PR #12262, opened 2023-11-03
HTTP: fix empty-body warning

Gottox:
This change fixes a compiler warning with gcc-12.2.0 when `-DCURL_DISABLE_BEARER_AUTH=ON` is used.

    /home/tox/src/curl/lib/http.c: In function 'Curl_http_input_auth':
    /home/tox/src/curl/lib/http.c:1147:12: warning: suggest braces around empty body in an 'else' statement [-Wempty-body]
     1147 |            ;
          |            ^

This is useful as it fixes an issue when libcurl is compiled with `-Werror`.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -1144,7 +1144,14 @@ CURLcode Curl_http_input_auth(struct Curl_easy *data, bool proxy,
               }
             }
 #else
-           ;
+            {
+              /*
+               * Empty block to terminate the if-else chain correctly.
+               *
+               * A semicolon would yield the same result here, but can cause a
+               * compiler warning when -Wextra is enabled.
+               */
+            }
 #endif
 
     /* there may be multiple methods on one line, so keep reading */


vszakats:
Ref, this fixes one of the 4 warnings reported in #12228.

Gottox:
Let me fix the other 4 too in this PR.

lib/http.c
```diff
@@ -1144,7 +1144,7 @@ CURLcode Curl_http_input_auth(struct Curl_easy *data, bool proxy,
               }
             }
 #else
-           ;
+           {}
```
bagder, line 1147: Seeing this is an unusual construct for curl code, can you add a comment here explaining why plain `;` is not enough?

Gottox: Simply put it: Compilers. [gcc warns](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wempty-body) you about orphaned `;` in the source code. This warning is also enabled by the `-Wextra` flag. I'm not sure about the reason, I would guess it prevents errors like this:

```c
if (foo);
   do_stuff();
```

bagder: Yes, and I want *that* as a comment in the code so that future readers understand that is the reason why this construct is used...

Gottox: Misread, will add!

Gottox: Tbh, I personally find this macro spaghetti quite hard to understand. Would you be fine if I refactor this into static functions too?

Gottox: Something like this: (pseudocode)
```c
static void handle_ntlm(const char *auth, ...) {
#ifdef USE_NTLM
  if (!checkprefix("Negotiate", auth) || !is_valid_auth_separator(auth[9])) {
    return;
  }
  // NTLM auth code
#else
  return;
#endif
}

// ...

CURLcode Curl_http_input_auth(struct Curl_easy *data, bool proxy,
                              const char *auth) /* the first non-space */
{
// ...
  if (handle_ntlm(auth, ...)) {
  } else if (handle_basic(auth, ...)) {
  } else if (handle_...(auth, ...)) {
  } else ...
```

The compiler would still eliminate empty functions, but I find it less noisy to read. Do you think this could help readability?

bagder: Yes I think it would. Reducing ifdef-mazes is a worthy effort.

Gottox: The PR is finished as is, I'll open a new one for the refactoring.

bagder (approved):

Issue #12268, opened 2023-11-04
URLs over 1MB give "Out of memory" after connection has been made


jeroen:
### I did this

A user of the R bindings [has reported](https://github.com/jeroen/curl/issues/318) that URLs over 1MB in length always generate an oom error, after the request has completed.  He claims this started happening starting libcurl 7.71.0.

This does seem related to https://github.com/curl/curl/issues/6681 but it is maybe unexpected that this errors after the connection has been initiated? Perhaps if this is expected the error message can be improved.

Below a reprex in C:

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <curl/curl.h>

int main(void)
{
  CURL *curl = curl_easy_init();
  char * url = calloc(1, 1300000);
  strcpy(url, "https://fakerapi.it/api/v1/custom?");
  for(int i = strlen(url); i < 1200000; i++){
    url[i] = 'x';
  }
  curl_easy_setopt(curl, CURLOPT_URL, url);
  curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
  curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);
  CURLcode res = curl_easy_perform(curl);
  if(res != CURLE_OK)
    fprintf(stderr, "curl_easy_perform() failed: %s\n",
            curl_easy_strerror(res));
  return 0;
}
```

### curl/libcurl version

libcurl 7.71.0 and up.

### operating system

Anwy

bagder:
It is a case of crappy error code + no good error message.

This code runs into the maximum size allowed for a HTTP request: 1024*1024 bytes (`DYN_HTTP_REQUEST`).

anthonynorth:
Libcurl 7.70.0 and earlier had no URL length limit.

I agree the error message is misleading (thank you for fixing), but any error from curl here is unexpected. _Full URL syntax with no length limit_ is still documented as one of libcurl's features.

bagder:
That phrase should probably updated/clarified.

Virtually no HTTP server will accept a request that large, which is one reason why libcurl no longer allows them. Also of course, no browser will ever use anything close to the size libcurl supports. libcurl does not actually limit the size of URLs but rather in some ways they are used, like HTTP. It can do longer requests for other protocols.

bagder:
[...]

PR #12269, opened 2023-11-04
curl.h: add CURLE_TOO_LARGE

bagder:
This adds a new error code that can be used when an internal field grows too large, like when a dynbuf reaches its maximum. Previously it would return CURLE_OUT_OF_MEMORY for this, which is highly misleading.

Ref: #12268


dfandrich:
Would CURLE_BAD_FUNCTION_ARGUMENT work instead? Or is that too oblique?

bagder:
> Would CURLE_BAD_FUNCTION_ARGUMENT work instead? Or is that too oblique?

I think that error code should be used when there is a more direct and immediately correlation between an input the result.

jay:
hyper ci job test 1154 is returning CURLE_OUT_OF_MEMORY instead of CURLE_TOO_LARGE

PR #12271, opened 2023-11-04
strdup: do Curl_strndup without strncpy

### LINKED ISSUE #12258: `-Wstringop-truncation` warnings with gcc 13 unity
### I did this

Did a curl-for-win build:
```
In file included from ./curl/bld-cmake-gcc-x64-ucrt/lib/CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c:346:
In function 'Curl_strndup',
    inlined from 'strstore' at ./curl/lib/cookie.c:368:10,
    inlined from 'sanitize_cookie_path' at ./curl/lib/cookie.c:314:5:
./curl/lib/strdup.c:117:3: warning: 'strncpy' output truncated before terminating nul copying 1 byte from a string of the same length [-Wstringop-truncation]
  117 |   strncpy(buf, src, length);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~
In function 'Curl_strndup',
    inlined from 'strstore' at ./curl/lib/cookie.c:368:10,
    inlined from 'pathmatch' at ./curl/lib/cookie.c:180:5,
    inlined from 'Curl_cookie_getlist' at ./curl/lib/cookie.c:1419:26:
./curl/lib/strdup.c:117:3: warning: 'strncpy' output truncated before terminating nul copying 1 byte from a string of the same length [-Wstringop-truncation]
  117 |   strncpy(buf, src, length);                          
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~                           
In function 'Curl_strndup',                                   
    inlined from 'curl_url_get' at ./curl/lib/urlapi.c:1595:13:
./curl/lib/strdup.c:117:3: warning: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
  117 |   strncpy(buf, src, length);                          
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~                           
In file included from ./curl/bld-cmake-gcc-x64-ucrt/lib/CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c:379:
./curl/lib/urlapi.c: In function 'curl_url_get':              
./curl/lib/urlapi.c:1593:22: note: length computed here       
 1593 |     size_t partlen = strlen(ptr);                     
      |                      ^~~~~~~~~~~                      
[ 12%] Built target libcurl_object                              
```

Possibly related:
4855debd8a2c1cbd0b0dbbb8319b1743c4644873 #12254
d3b3ba35a5c4f862df2aa3e12de983e5457b429a #12251

### I expected the following

No warnings.

### curl/libcurl version

8.5.0-DEV

### operating system

any

bagder:
To avoid (false positive) gcc-13 compiler warnings.

Follow-up to 4855debd8a2c1cb

Idea-by: Jay Satiro
Reported-by: Viktor Szakats
Fixes #12258


jay:
nm i think your way is better, note posix allows both https://pubs.opengroup.org/onlinepubs/9699919799/functions/strdup.html#tag_16_573_07

bagder:
Now when this function actually follows POSIX, we can make the code use the native `strndup` on systems that have one...

jay:
It's unclear to me if posix strndup allows for reading past than `size` to determine length. I would think not but I found this proposed implementation in a mingw [issue](https://github.com/msys2/MINGW-packages/issues/4999):

~~~c
char * strndup(const char * src, size_t size) {
  size_t len = strlen(src);
  len = len < size ? len : size;
~~~

POSIX says it expects src to be a string so unlike our implementation we couldn't use strndup on memory that may not be null terminated. If that's going to be an issue then we should stick with our own strndup instead of posix

bagder:
> we should stick with our own strndup instead of posix

Good research. Yeah, let's stick to our own.

bagder:
Let's not overload this PR. This fixes the function for the warning reason. I don't think we need to make it a curlx_ one at this point. I don't think we should call it plain `strndup` if it isn't identical in functionality.

jay (approved):

bagder:
Closed via ac57e69b589010 (I missed the closes keyword)

PR #12272, opened 2023-11-05
http: refactor http auth into static functions

Gottox:
Based on https://github.com/curl/curl/pull/12262 I factored the auth functions out into its own functions. While doing that I noticed that the auth code could be turned into a abstract interface with multiple implementors. Are you peps interested in this?

The changes I currently did are purely cosmetic (given, that the compiler is smart enough to do proper inlining/dead code elimination), but if I'm implementing an abstract interface, this would introduce an indirection between the http code and the http-auth code.

This PR is opened as a draft for now to discuss if my approach makes sense.


lib/http.c
```diff
@@ -980,6 +983,169 @@ static int is_valid_auth_separator(char ch)
...
+            return CURLE_OUT_OF_MEMORY;
+          authinfo->data->state.authproblem = FALSE;
+          /* we received a GSS auth token and we dealt with it fine */
+          *authinfo->negstate = GSS_AUTHRECV;
+        }
+        else
+          authinfo->data->state.authproblem = TRUE;
+      }
+    }
+    return true;
```
bagder, line 1024: We use `TRUE` and `FALSE` (uppercase).

lib/http.c
```diff
@@ -1027,125 +1193,11 @@ CURLcode Curl_http_input_auth(struct Curl_easy *data, bool proxy,
...
-              }
-            }
-#else
-           ;
-#endif
+    if(handle_spnego(&authinfo)) {}
+    else if(handle_ntlm(&authinfo)) {}
+    else if(handle_digest(&authinfo)) {}
+    else if(handle_basic(&authinfo)) {}
+    else if(handle_bearer(&authinfo)) {}
```
bagder, line 1200:
```suggestion
    if(handle_spnego(&authinfo) || 
       handle_ntlm(&authinfo) ||
       handle_digest(&authinfo) ||
       handle_basic(&authinfo) ||
       handle_bearer(&authinfo))
       {}
```

Saves having to use `{}` so many times.

bagder (commented):
1. Needs a rebase
2. Seems to have broken test 1204

lib/http.c
Gottox: in that case, even the `if` is not needed:

```
handle_spnego(&authinfo) || 
       handle_ntlm(&authinfo) ||
       handle_digest(&authinfo) ||
       handle_basic(&authinfo) ||
       handle_bearer(&authinfo);
```

Gottox:
@bagder Generally speaking would you support factoring out all the authentication specific code into its own interface?

For start, I'm thinking of a simple interface like this: (pseudocode)

```c
struct authimpl {
  const char *name;
  bool (*handle)(struct authdata *);
};

// ...

static bool handle_auth_basic(struct authdata *) {
  // ...
}

static struct authimpl authimpls[] {
  { "Basic", 5, handle_auth_basic },
  { "Bearer", 5, handle_auth_bearer },
  // ...
}

// ...

struct authdata ad = { ... };
for(int i = 0; i < LENGTH(authimpls); i++) {
  if (checkauth(authimpls[i].name, auth)) {
    authimpls[i].handle(&ad);
    break;
  }
}
```

In the second iteration I'd like to move the auth methods to their own files and let them build conditionally. This would reduce `#ifdef` clutter even more at the cost compile time inlining.

lib/http.c
```diff
@@ -980,6 +983,169 @@ static int is_valid_auth_separator(char ch)
...
+      if(authinfo->authp->picked == CURLAUTH_NEGOTIATE) {
+        CURLcode result = Curl_input_negotiate(authinfo->data,
+                                               authinfo->conn,
+                                               authinfo->proxy,
+                                               authinfo->auth);
+        if(!result) {
+          free(authinfo->data->req.newurl);
+          authinfo->data->req.newurl = strdup(authinfo->data->state.url);
+          if(!authinfo->data->req.newurl)
+            return CURLE_OUT_OF_MEMORY;
```
bagder, line 1015: This returns an error code, not a boolean! You probably need to propagate the error code.

bagder:
> @bagder Generally speaking would you support factoring out all the authentication specific code into its own interface?

I would. I think that could clean things up nicely!

bagder:
Stalled

PR #12279, opened 2023-11-06
lib: accept 100MB input strings and 100MB HTTP request

### LINKED ISSUE #12268: URLs over 1MB give "Out of memory" after connection has been made

bagder:
Reported-by: Jeroen Ooms and Anthony North
Fixes #12268

### DIFF
--- a/lib/dynbuf.h
+++ b/lib/dynbuf.h
@@ -80,7 +80,7 @@ int Curl_dyn_vprintf(struct dynbuf *dyn, const char *format, va_list ap_save);
 #define DYN_DOH_CNAME       256
 #define DYN_PAUSE_BUFFER    (64 * 1024 * 1024)
 #define DYN_HAXPROXY        2048
-#define DYN_HTTP_REQUEST    (1024*1024)
+#define DYN_HTTP_REQUEST    (100*1024*1024)
 #define DYN_APRINTF         8000000
 #define DYN_RTSP_REQ_HEADER (64*1024)
 #define DYN_TRAILERS        (64*1024)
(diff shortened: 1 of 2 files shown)


bagder:
This PR is on hold because we have not received feedback from anyone saying this is a good idea or that they would like to see this change merged.

GrahamCampbell:
My gut feeling is that this is a bad idea, and that keeping these limits low is a great security feature for the majority of users, helping them maintain availability and catch bugs. I d'no if it's against the curl design principals to allow this to be configured at compile time. Maybe that is an out for the small number of people who might need this? Another possibly overkill thing could be to have this as a runtime option?

PR #12282, opened 2023-11-06
Windows VC14.20 project missing

dsv123:
Windows projects included VC14, VC14.10, VC14.30 but not VC14.20.
OpenSSL and WolfSSL bat scripts mention VC14.20 so I don't suspect an underlying problem with this platform toolset.

Updated the templates to produce a VC14.20 project. Project opens (and builds) in Visual Studio 2019 as expected.
The VC14.30 solution contained text "Visual Studio 15" instead of "Visual Studio 17", updated.


bagder (approved):

bagder:
Thanks!

bagder:
@dsv123 this does not add any code to `Makefile.am`  in the style of 14_30 so I'm pretty sure this then does not create working 14.20 files in the tarballs. See:

https://github.com/curl/curl/blob/28287092cc5a6d6ef875ce157f9cd16fc7723305/Makefile.am#L600-L622

bagder:
I realized this when I made #12288 just now

bagder:
Anyway, that should be fixed in a separate PR, just wanted to highlight this.

PR #12287, opened 2023-11-06
cmake: option to disable install & drop `curlu` target when unused

i1qh2n:
This patch makes the following changes: 
 + adds the option `CURL_DISABLE_INSTALL` - to disable 'install' targets.
 + Removes the target `curlu` when the option `BUILD_TESTING` is set to `OFF` - to prevent it from being loaded in Visual Studio.


CMakeLists.txt
```diff
@@ -1523,259 +1526,263 @@ if(BUILD_TESTING)
...
 
-# Helper to populate a list (_items) with a label when conditions (the remaining
-# args) are satisfied
-macro(_add_if label)
-  # needs to be a macro to allow this indirection
-  if(${ARGN})
-    set(_items ${_items} "${label}")
-  endif()
-endmacro()
+if (NOT CURL_DISABLE_INSTALL)
```
vszakats, line 1529:
```suggestion
if(NOT CURL_DISABLE_INSTALL)
```

CMakeLists.txt
```diff
@@ -1523,259 +1526,263 @@ if(BUILD_TESTING)
...
 
-  add_custom_target(curl_uninstall
-      COMMAND ${CMAKE_COMMAND} -P
-      ${CMAKE_CURRENT_BINARY_DIR}/CMake/cmake_uninstall.cmake)
-endif()
+    add_custom_target(curl_uninstall
+        COMMAND ${CMAKE_COMMAND} -P
+        ${CMAKE_CURRENT_BINARY_DIR}/CMake/cmake_uninstall.cmake)
+  endif()
+endif()
```
vszakats, line 1788: Please re-add the closing linefeed.

lib/CMakeLists.txt
```diff
@@ -47,20 +47,24 @@ if(USE_ARES)
...
-target_compile_definitions(curlu PUBLIC UNITTESTS CURL_STATICLIB)
+if (BUILD_TESTING)
+    add_library(
+      curlu # special libcurlu library just for unittests
+      STATIC
+      EXCLUDE_FROM_ALL
+      ${HHEADERS} ${CSOURCES}
+    )
+    target_compile_definitions(curlu PUBLIC UNITTESTS CURL_STATICLIB)
+endif()
```
vszakats, line 58:
```suggestion
if(BUILD_TESTING)
  add_library(
    curlu # special libcurlu library just for unittests
    STATIC
    EXCLUDE_FROM_ALL
    ${HHEADERS} ${CSOURCES}
  )
  target_compile_definitions(curlu PUBLIC UNITTESTS CURL_STATICLIB)
endif()
```
Let's keep indenting by 2 spaces.

lib/CMakeLists.txt
```diff
@@ -47,20 +47,24 @@ if(USE_ARES)
...
 
 if(ENABLE_CURLDEBUG)
   # We must compile these sources separately to avoid memdebug.h redefinitions
   # applying to them.
   set_source_files_properties(memdebug.c curl_multibyte.c PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
 endif()
-target_link_libraries(curlu PRIVATE ${CURL_LIBS})
+if (BUILD_TESTING)
+    target_link_libraries(curlu PRIVATE ${CURL_LIBS})
+endif()
```
vszakats, line 67:
```suggestion
if(BUILD_TESTING)
  target_link_libraries(curlu PRIVATE ${CURL_LIBS})
endif()
```
Here too, plus formatting.

CMakeLists.txt
```diff
@@ -105,6 +105,9 @@ option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 option(BUILD_STATIC_LIBS "Build static libraries" OFF)
 option(BUILD_STATIC_CURL "Build curl executable with static libcurl" OFF)
 option(ENABLE_ARES "Set to ON to enable c-ares support" OFF)
+option(CURL_DISABLE_INSTALL "Set to ON to disable instalation targets" OFF)
+
+
```
vszakats, line 110: One empty line should be enough above.

vszakats:
Can you add a few words about why the new `if(BUILD_TESTING)` conditions are necessary to disable install?

i1qh2n:
It is my understanding that the `curlu` library is only needed in when building tests. With that it falls under the same category as installation targets. In that it is not needed when building inline unless the option `BUILD_TESTING` is set to `ON`

vszakats:
Did those targets get built for you with `BUILD_TESTING=OFF`? If so, can you post the cmake settings to trigger it?

i1qh2n:
With the default configuration on this repository:
Setting `BUILD_TESTING` to `OFF` produces the projects
```
ALL_BUILD
curl 
curl_uninstall
curltool
INSTALL
libcurl_object
libcurl_shared
ZERO_CHECK
``` 

Setting `BUILD_TESTING` to `ON` produces the projects
```
ALL_BUILD
curl 
curl_uninstall
curltool
curlu
INSTALL
libcurl_object
libcurl_shared
Test lib ...
Test lib ...
ZERO_CHECK
``` 

I have this repository embedded in an experimental project that uses curl as a submodule. 

The initial example is what I am using to lock the options I need in place before recursing into this repository’s scripts.    

Toggling `BUILD_TESTING=ON|OFF` from that level achieves the same result as if I were toggling it on the command line.  

For instance, if I wanted to include this project’s tests as part of my project testing, I would conditionally switch that variable to `ON` at the scope before `add_subdirectory(libcurl)`. That in turn would toggle on the `curlu` project as well as all the `Test lib …` projects to include as part of my build. 

I added the conditional BUILD_TESTING statement around `curlu` because it was not needed (outside of testing) to link and use the main output libraries.

CMakeLists.txt
```diff
@@ -1523,259 +1525,263 @@ if(BUILD_TESTING)
...
-set(libdir                  "${CMAKE_INSTALL_PREFIX}/lib")
-foreach(_lib ${CMAKE_C_IMPLICIT_LINK_LIBRARIES} ${CURL_LIBS})
-  if(TARGET "${_lib}")
-    set(_libname "${_lib}")
-    get_target_property(_imported "${_libname}" IMPORTED)
-    if(NOT _imported)
-      # Reading the LOCATION property on non-imported target will error out.
-      # Assume the user won't need this information in the .pc file.
-      continue()
+
```
vszakats, line 1530: Let's delete this extra empty line, too.

vszakats:
Your results say that `BUILD_TESTING` is working as expected, meaning `curlu` (static lib) is not built when `BUILD_TESTING` is set to `OFF`. In that case, do you think we still need the extra guards added in this patch?

i1qh2n:
That is a miscommunication on my end, the above settings are only _with_ the extra guards in place.
From a fresh clone, setting `BUILD_TESTING` to `OFF` produces the projects:
```
ALL_BUILD
curl 
curl_uninstall
curltool
curlu
INSTALL
libcurl_object
libcurl_shared
ZERO_CHECK
``` 
With `curlu` being present in both.

vszakats:
What do you mean by 'project' in this context?

The reason I'm insisting is because this test:
```
cmake . -B bld -DBUILD_TESTING=OFF
```
when built, does not produce a `curlu` library. I'm assuming the reason for this is that even though `curlu` is defined as a target, it never gets used because `BUILD_TESTING=ON` is required to build things dependent on it under `tests`. Without building `tests`, `curlu` as a target is not referenced, thus also not built.

I'm looking for an example which builds `curlu` with `BUILD_TESTING=OFF`.

i1qh2n:
I see what you are saying.  By ‘project’ I am referring to visual studio project files.  

Even if it is not referenced in the actual build, the project file is still generated. In turn, it is loaded in the visual studio solution.

By default, the ALL_BUILD project will build both it and either `libcurl_static` or `libcurl_shared` depending on settings. 

The purpose of this patch is to add the option to toggle off the installation projects in the generated solution.  With the `BUILD_TESTING` condition , it removes the unreferenced project when disabled, or adds it back when enabled.  

It allows a strict filter to only generate and include the core libraries needed to build, link and use curl.

vszakats:
Thanks for explaining!

If I got it correctly this patch does these two things:

1. add option `CURL_DISABLE_INSTALL` to disable 'install'.
2. disable tests targets more when `BUILD_TESTING` is `OFF`, to avoid them appearing in Visual Studio.

Would you mind updating the PR message to clarify?

i1qh2n:
> Would you mind updating the PR message to clarify?

Done. 

Thank you for your time!

vszakats:
Thank you, LGTM!

vszakats (approved):

PR #12288, opened 2023-11-07
Makefile.am: drop vc10, vc11 and vc12 from the dist tarball

bagder:
They are end of life products. Support for generating them remain in the repo for a while but this change drops them from distribution.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -275,9 +275,7 @@ checksrc:
 
 .PHONY: vc-ide
 
-vc-ide: $(VC10_LIBVCXPROJ_DEPS) $(VC10_SRCVCXPROJ_DEPS)                  \
- $(VC11_LIBVCXPROJ_DEPS) $(VC11_SRCVCXPROJ_DEPS) $(VC12_LIBVCXPROJ_DEPS) \
- $(VC12_SRCVCXPROJ_DEPS) $(VC14_LIBVCXPROJ_DEPS) $(VC14_SRCVCXPROJ_DEPS) \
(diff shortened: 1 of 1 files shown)


MarcelRaad:
These old versions were actually the only use case I still saw for the project files at all as CMake removed support for generating them for VC 10 and 11, and deprecated VC 12. Newer Visual Studio versions even come with CMake.

bagder:
No one would be happier than me if we would remove them all from both git and the dist, but are our users ready for that?

jay:
I'm still using the vc10 and vc12 projects on some older computers. I switch between builds a lot and SSL libraries using the configuration drop down. I can re-generate if necessary but a lot of the time it isn't necessary. Like for example if no file is added I usually don't have to regenerate. If I want to switch between OpenSSL and WinSSL (Schannel) I don't have to regenerate. If I want to switch between 64/32 or release/debug I don't have to regenerate. And so on. I just choose from the drop down. The project configurations are very basic and had to I write a script to [enable http2](https://gist.github.com/jay/431854a56e0079203f28). As for users that may be using the old versions I don't know if that's happening. I see no reason to disable generating them for distribution if they're going to remain in the repo. I guess we'll find out if anyone makes a report about it.

bagder:
> I can re-generate if necessary but a lot of the time it isn't necessary.

I would imagine that most curl developers need that rather than to have them shipped in the tarball as we do change the source file set every once in a while.

> I see no reason to disable generating them for distribution if they're going to remain in the repo. 

It's a gradual thing. Let's first see if anyone even notices this step.

Issue #4776, opened 2020-01-03
Reporting of PRETRANSFER_TIME and STARTTRANSFER_TIME when using curl_multi_perform()


lonezor:
### I did this
I'm seeing this in libcurl-7.67.0 using HTTP/2 backend when sending POST in
custom build environment, but it can also be seen with HTTP 1.1 GET so
I'm reproducing it with Ubuntu package libcurl3-gnutls:amd64.

The example multi-single.c has been modified to run a larger number
of transfers to localhost's default Ubuntu page (12 kB HTML)

To make it more readable I'm exaggerating the test by running 3000
transfers over two re-used connections:

```
0001 code 200 name_lookup 0.010411 connect 0.010475 app_connect 0.000000 pre_transfer 0.010502 start_transfer 0.011029 total_time 0.011038
0002 code 200 name_lookup 0.010935 connect 0.010957 app_connect 0.000000 pre_transfer 0.010974 start_transfer 0.018878 total_time 0.019320
0003 code 200 name_lookup 0.000012 connect 0.000012 app_connect 0.000012 pre_transfer 0.000043 start_transfer 0.008355 total_time 0.019235
0004 code 200 name_lookup 0.000005 connect 0.000005 app_connect 0.000005 pre_transfer 0.000022 start_transfer 0.008036 total_time 0.027650
0005 code 200 name_lookup 0.000004 connect 0.000004 app_connect 0.000004 pre_transfer 0.000012 start_transfer 0.008417 total_time 0.027657
...
0500 code 200 name_lookup 0.000005 connect 0.000005 app_connect 0.000005 pre_transfer 0.000020 start_transfer 0.006393 total_time 1.919017
0501 code 200 name_lookup 0.000021 connect 0.000021 app_connect 0.000021 pre_transfer 0.000028 start_transfer 0.006649 total_time 1.919024
0502 code 200 name_lookup 0.000004 connect 0.000004 app_connect 0.000004 pre_transfer 0.000025 start_transfer 0.006327 total_time 1.925707
0503 code 200 name_lookup 0.000021 connect 0.000021 app_connect 0.000021 pre_transfer 0.000028 start_transfer 0.006667 total_time 1.925714
0504 code 200 name_lookup 0.000004 connect 0.000005 app_connect 0.000005 pre_transfer 0.000016 start_transfer 0.006313 total_time 1.932292
[...]

PR #12293, opened 2023-11-07
Fix for StartTransfer Time

zousar:
Ensure that StartTransfer time for a request doesn't include the time spent waiting for a connection when we are constraining concurrent connection quantity.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1938,8 +1938,7 @@ static CURLMcode multi_runsingle(struct Curl_multi *multi,
         result = CURLE_OK;
         break;
       }
-      else
-      {
+      else {
         /* update the start time when we have left the
            CURLE_NO_CONNECTION_AVAILABLE state */
         Curl_pgrsTimeWas(data, TIMER_STARTSINGLE, *nowp);
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1940,9 +1940,10 @@ static CURLMcode multi_runsingle(struct Curl_multi *multi,
       }
       else
       {
-        /* update the start time when we have left the CURLE_NO_CONNECTION_AVAILABLE state */
+        /* update the start time when we have left the
+           CURLE_NO_CONNECTION_AVAILABLE state */
         Curl_pgrsTimeWas(data, TIMER_STARTSINGLE, *nowp);
-        
+
         if(data->state.previouslypending) {
           /* this transfer comes from the pending queue so try move another */
           infof(data, "Transfer was pending, now try another");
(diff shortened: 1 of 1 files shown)


icing:
Not sure I understand the fix here. `*nowp` is the value that was already set some lines above. `Curl_connect()` does not wait, so how does setting the timestamp again help? (I must overlook something)

zousar:
> Not sure I understand the fix here. `*nowp` is the value that was already set some lines above. `Curl_connect()` does not wait, so how does setting the timestamp again help? (I must overlook something)

`Curl_pgrsTimeWas` does not alter `*nowp` in the added line.  The purpose of it being called in the "else" block is to have it update the `data->progress.t_startsingle` time with the timestamp stored in `*nowp`.  By updating that `data->progress.t_startsingle` value in the "else" as we cease to be in the `CURLE_NO_CONNECTION_AVAILABLE` state, it ensures that the `CURLINFO_STARTTRANSFER_TIME` metric doesn't include the time spent waiting for a connection to be available.  The `CURLINFO_TOTAL_TIME` metric still includes the time spent  waiting for a connection to be available, but the StartTransfer time does not.

As was noted in this comment in the past:
https://github.com/curl/curl/issues/4776#issuecomment-570890159
it doesn't seem correct to include connection wait time in StartTransfer time.

For my use case, I use StartTransfer - Connect as a measure of network latency, and having StartTransfer include the time spent waiting for a connection when we are limiting concurrent connections undermines that use case.

bagder:
I think the main reason this is not the right/sufficient change is that this *drastically* modifies the connection timeout logic. Current it sets a maximum time that you allow the transfer to spend until it has connected. Period. Independent of if it sits in the pending queue waiting for the chance to create a connection or not. The timeout means it fails if it spends longer than N milliseconds to connect.

This change now resets that timer when the handle gets moved out from the pending queue.

Let's take this scenario:

- Max connect timeout set to 7 seconds.
- 5 seconds are spent in the pending queue waiting for a chance to start
- 2 more seconds is spent trying to connect, and then the timeout hits and the transfer aborts. Because it has now spent its 7 seconds.

With this proposed edit, it seems to rather:

- Max connect timeout set to 7 seconds.
- 5 seconds are spent in the pending queue waiting for a chance to start
- the timer is set again, clearing the timeout counter
- 7 *more* seconds can be spent trying to connect, and then the timeout hits and the transfer aborts. Having spent a total of 12 seconds...

Or am I wrong?

zousar:
I believe you are correct. If the timeout logic must include time spent
queued waiting for a connection to become available (when they are
limited), then I suspect the only solution to the original problem is to
leave the current time marker (startsingle) alone and introduce a new one
for use by the computation of the StartTransfer metric. Is this something
you want me to attempt as a PR or would you rather attempt it (or a
different approach)?

bagder:
See  #12368

PR #12298, opened 2023-11-08
tidy-up: use `OPENSSL_VERSION_NUMBER`

vszakats:
Uniformly use `OPENSSL_VERSION_NUMBER` to check for OpenSSL version.
Before this patch some places used `OPENSSL_VERSION_MAJOR`.

Also fix `lib/md4.c`, which included `opensslconf.h`, but that doesn't
define any version number in these implementations: BoringSSL, AWS-LC,
LibreSSL, wolfSSL. (Only in mainline OpenSSL/quictls). Switch that to
`opensslv.h`. This wasn't causing a deeper problem because the code is
looking for v3, which is only provided by OpenSSL/quictls as of now.

According to https://github.com/openssl/openssl/issues/17517, the macro
`OPENSSL_VERSION_NUMBER` is safe to use and not deprecated.

Closes #12298

### DIFF
--- a/lib/md4.c
+++ b/lib/md4.c
@@ -32,7 +32,7 @@
 #include "warnless.h"
 
 #ifdef USE_OPENSSL
-#include <openssl/opensslconf.h>
+#include <openssl/opensslv.h>
 #if (OPENSSL_VERSION_NUMBER >= 0x30000000L) && !defined(USE_AMISSL)
 /* OpenSSL 3.0.0 marks the MD4 functions as deprecated */
 #define OPENSSL_NO_MD4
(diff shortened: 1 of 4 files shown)


MarcelRaad:
I think the reason for avoiding `OPENSSL_VERSION_NUMBER` when checking for OpenSSL v3 or later was that it used to be deprecated in v3.0, but that seems to be reverted:
https://github.com/openssl/openssl/issues/17517

MarcelRaad (approved):

vszakats:
Thank you! I was wondering about this, good to have this clarified.

PR #12311, opened 2023-11-13
allow longer http methods for service-specific custom methods

gfx:
Hi, team! Looks like https://github.com/curl/curl/pull/10720 has accidentally added a new limitation to the length of HTTP methods to `11`, which affects our internal use of curl(1).

This PR doubles the limitation, which is large enough for our usage. Can you take a look at it? 

* [x] Checked whether it works for our usage

### DIFF
--- a/lib/http.h
+++ b/lib/http.h
@@ -263,7 +263,7 @@ CURLcode Curl_http_decode_status(int *pstatus, const char *s, size_t len);
  * All about a core HTTP request, excluding body and trailers
  */
 struct httpreq {
-  char method[12];
+  char method[24];
   char *scheme;
   char *authority;
   char *path;


jay:
I think this a good idea. The longest [registered](https://www.iana.org/assignments/http-methods/http-methods.xhtml) method [UPDATEREDIRECTREF](https://www.rfc-editor.org/rfc/rfc4437.html#section-7) is 17 characters long. By custom methods do you mean methods not registered?

gfx:
@jay 

Thank you for the additional motivation to change the current limitation! 

Our custom methods won't be registered because they are used only in our internal services. The max length of our methods is 17 bytes as well, so I think 24 is large enough for those who use custom, unregistered methods.

jay:
Thanks

PR #12319, opened 2023-11-13
ancestor: provide a handle to inherit from/use

bagder:
Add `CURLMOPT_ANCESTOR` and `CURLSHOPT_ANCESTOR`, for DoH and closing down connections.

This is an alternative to #12246 and #12144, for discussion and brainstorming.

/cc @icing and @jay


bagder:
I think we could make good answers to and solutions for those questions if we want to.

jay:
I think all of the ideas we've come up with are sort of ok but have problems, they are going to confuse the user and probably us which will make them hard to maintain.

bagder:
> I think all of the ideas we've come up with are sort of ok but have problems, they are going to confuse the user and probably us which will make them hard to maintain.

I suspect you're entirely correct. We probably need to give this some more thoughts and time.

At least this PR of mine exercises the duphandle logic internally for virtually every created handle which has made me find and fix four or five issues with it!

bagder:
Let's ponder on this some more.

PR #12321, opened 2023-11-14
tool: support bold headers in Windows

jay:
- If virtual terminal processing is enabled in Windows then use ANSI escape codes Esc[1m and Esc[22m to turn bold on and off.

Suggested-by: Gisle Vanem

Ref: https://github.com/curl/curl/discussions/11770

Closes #xxxx

---

prompt

/cc @gvanem @dfandrich

### DIFF
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -714,6 +714,8 @@ static struct TerminalSettings {
 #define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x0004
 #endif
 
+bool tool_isVirtualTerminal;
+
 static void restore_terminal(void)
 {
   if(InterlockedExchange(&TerminalSettings.valid, (LONG)FALSE))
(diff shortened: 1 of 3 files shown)


src/tool_doswin.c
```diff
@@ -714,6 +714,8 @@ static struct TerminalSettings {
 #define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x0004
 #endif
 
+bool tool_isVirtualTerminal;
```
bagder, line 717: Why the `virtual` part of the name? The bold is output when the terminal is real as your screenshot shows. The virtual part puzzles me.

gvanem: Users have good reasons to turn VTS off even in Win-10 build 16299.
It simply means `(TerminalSettings.dwOutputMode & ENABLE_VIRTUAL_TERMINAL_PROCESSING) == TRUE`.

bagder: Okay, that's the technical name for the feature in the OS but for people reading the code, wouldn't a better name be something like `terminal_has_bold_enabled` ? It is the implied connection between "virtual" and "it can do styled output" in the variable name that I reacted to. Not that the the check if the styled output works actually needs to check with something that is called a virtual terminal.

jay: Ok I changed the name to `tool_term_has_bold`

gvanem:
What could be done for the link-text on Windows? AFCAIK, it's possible with underlined text under VTS,
but I'm not sure that's the same as for Unix. How does a `LINK <url>` look and work under Unix? Click on it and it opens the
default browser?

bagder:
> How does a LINK <url> look and work under Unix? Click on it and it opens the default browser?

Click on it and the "default app" for that link type opens it, yes. Presumably typically a browser.

gvanem:
> Click on it and the "default app" for that link type opens it, yes. 

That would be hard to accomplish on Windows I guess.

bagder:
> With [Windows Terminal](https://github.com/microsoft/terminal), you can Ctrl-click URLs to open them in your default browser. 🙂

Yes, but with some of the other-systems terminals, we can also "mark up" relative URLs used in the headers and spell out the full URL for when someone clicks on them. Ie use more meta-data than what is visible on screen.

DHowett:
If you end up using OSC 8 (the standard for hyperlinks in terminals) that will work on Windows as well!

PR #12330, opened 2023-11-15
test348: suppress uploaded data check

monnerat:
The test is about reaching a disk full condition upon ftp upload, so the uploaded data is supposed to be incomplete.


bagder:
I disagree. We know the upload works fine since we control the server for this. This test makes the test server say the disk is full, but it does this after the upload has been completed. Thus, we know that the upload should have been successful and the verification of the data should work.

Do you see errors when running this test?

monnerat:
> Do you see errors when running this test?

Not in the current master branch. I reached a problem about it while trying to adjust test system for ManageSieve that can have multiline commands without intermediate server replies. I guess I have to find another adjustment.

The proper uploaded ftp data check is already performed by test107.

Issue #4092, opened 2019-07-04
curl --compressed not decompressing data successfully


ericcurtin:
### I did this

I wrote a simple http server:

https://github.com/ericcurtin/asio/commit/1d37a1d225d1e812a747b595c02f9770ebd75dd0

When I request content from the server and pipe the output to gunzip all works fine

```
curl -x "" 127.0.0.1:5000/main.cpp --output - | gunzip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   758  100   758    0     0   740k      0 --:--:-- --:--:-- --:--:--  740k
// g++ -O0 main.cpp server.cpp connection_manager.cpp request_handler.cpp
// connection.cpp reply.cpp mime_types.cpp request_parser.cpp -lboost_system
// -lpthread -lz
//
// run like: ./a.out 0.0.0.0 5000 .
//
// main.cpp
// ~~~~~~~~
//
// Copyright (c) 2003-2017 Christopher M. Kohlhoff (chris at kohlhoff dot com)
//
// Distributed under the Boost Software License, Version 1.0. (See accompanying
// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
//

#include <iostream>
#include <string>
#include <boost/asio.hpp>
#include "server.hpp"

int main(int argc, char* argv[])
{
  try
  {
    // Check command line arguments.
    if (argc != 4)
    {
      std::cerr << "Usage: http_server <address> <port> <doc_root>\n";
      std::cerr << "  For IPv4, try:\n";
      std::cerr << "    receiver 0.0.0.0 80 .\n";
      std::cerr << "  For IPv6, try:\n";
      std::cerr << "    receiver 0::0 80 .\n";
      return 1;
    }

    // Initialise the server.
    http::server::server s(argv[1], argv[2], argv[3]);

    // Run the server until stopped.
    s.run();
  }
  catch (std::exception& e)
  {
    std::cerr << "exception: " << e.what() << "\n";
  }

  return 0;
}
```

But when I try and decompress via --compressed curl flag it fails after the first 512 bytes:

```
curl -x "" 127.0.0.1:5000/main.cpp --compressed
// g++ -O0 main.cpp server.cpp connection_manager.cpp request_handler.cpp
// connection.cpp reply.cpp mime_types.cpp request_parser.cpp -lboost_system
// -lpthread -lz
//
[...]

Issue #3796, opened 2019-04-20
GET requests with gzip encoding fail with CURLE_WRITE_ERROR


jay:
### I did this

Dmitri Trembovetski [reported on the mailing list](https://curl.haxx.se/mail/lib-2019-04/0060.html) that when he enables gzip his requests fail with CURLE_WRITE_ERROR. He traced it to process_trailer() in lib/content_encoding.c line 149:

https://github.com/curl/curl/blob/521bbbe29928f9bc1c61306df612e856d45cbe5a/lib/content_encoding.c#L136-L157

He provided a trace.log which I constructed into the server response, get it here [trace.zip](https://github.com/curl/curl/files/3100680/trace.zip)

First window:
`while true; do cat response.bin | nc -4l localhost 8000; done`

Second window:
`curld -v --compressed -o out http://localhost:8000`

While inflating the response inflate returns Z_STREAM_END. The [zlib manual](https://www.zlib.net/manual.html) says _"inflate() should normally be called until it returns Z_STREAM_END or an error."_ And following that libcurl calls process_trailer to handle the extra data. Interestingly that's kind of a workaround because most modern zlibs should process the trailer automatically, and it appears that's what's happened here, as expected. What is not expected is that extra data after the stream ends (z->avail_in == 23) but no trailer is expected because as noted modern zlib handles that (zp->trailerlen == 0). 

The 23 bytes that are unexpected:

`1F 8B 08 00 00 00 00 00 00 FF 01 00 00 FF FF 00 00 00 00 00 00 00 00`

Another interesting thing, it starts with a gzip header. gzip RFC 1952 [says in 2.2 file format](https://tools.ietf.org/html/rfc1952#page-5):
~~~
      A gzip file consists of a series of "members" (compressed data
      sets).  The format of each member is specified in the following
      section.  The members simply appear one after another in the file,
      with no additional information before, between, or after them.
~~~

So shouldn't we allow it as a separate member and not treat it as an unexpected trailer?

### I expected the following

A good time

### curl/libcurl version

master from a few days ago 538db66 2019-04-16

~~~
curl 7.65.0-DEV (i386-pc-win32) libcurl/7.65.0-DEV OpenSSL/1.0.2r zlib/1.2.11 nghttp2/1.37.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
[...]

PR #12340, opened 2023-11-16
Fixed "Failed writing received data to disk/application" error when multiple compressed streams are present

sokurenko:
Failed writing received data to disk/application error is displayed when using curl and there are multiple compressed streams in response. For example:

curl https://www.fust.ch/ --compressed

Attached data with multiple compression for test:
[output.gz](https://github.com/curl/curl/files/13377925/output.gz)

Also please see:
https://support.zabbix.com/browse/ZBX-23702

### DIFF
--- a/lib/content_encoding.c
+++ b/lib/content_encoding.c
@@ -239,6 +239,21 @@ static CURLcode inflate_stream(struct Curl_easy *data,
       /* No more data to flush: just exit loop. */
       break;
     case Z_STREAM_END:
+      if(!zp->trailerlen && z->avail_in)
+      {
+        uInt avail_in = z->avail_in;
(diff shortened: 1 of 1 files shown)


jay:
There should be one stream not multiple streams. See #4092 and #3796

sokurenko:
> There should be one stream not multiple streams. See #4092 and #3796

I understand, but it might be better to continue instead of failing or decompressing first stream and finishing without error.

jay:
It's almost unheard of. In my opinion we should keep it an error but I will leave this open for a while to gather more feedback.

jay:
Thanks for your proposal but we're going to continue to treat this as an error.

PR #12345, opened 2023-11-16
cmake: add `CURL_DISABLE_BINDLOCAL` option

vszakats:
To match similar autotools option.

Default is `ON`.

Closes #12345


bagder (approved):
We should extend test 1165 to also verify that the cmake build is in sync.

PR #12346, opened 2023-11-16
autotools: stop setting `-std=gnu89` with `--enable-warnings`

vszakats:
Do not alter the C standard when building with `--enable-warnings` when
building with gcc.

On one hand this alters warning results compared to a default build.
On the other, it may produce different binaries, which is unexpected.

Also fix new warnings that appeared after removing `-std=gnu89`:

- include: fix public curl headers to use the correct printf mask for
  `CURL_FORMAT_CURL_OFF_T` and `CURL_FORMAT_CURL_OFF_TU` with mingw-w64
  and Visual Studio 2013 and newer. This fixes the printf mask warnings
  in examples and tests. E.g. [1]

- conncache: fix printf format string [2].

- http2: fix potential null pointer dereference [3].
  (seen on Slackware with gcc 11.)

- libssh: fix printf format string in SFTP code [4].
  Also make MSVC builds compatible with old CRT versions.

- libssh2: fix printf format string in SFTP code for MSVC.
  Applying the same fix as for libssh above.

- unit1395: fix `argument is null` and related issues [5]:
  - stop calling `strcmp()` with NULL to avoid undefined behaviour.
  - fix checking results if some of them were NULL.
  - do not pass NULL to printf `%s`.

- ci: keep a build job with `-std=gnu89` to continue testing for
  C89-compliance. We can apply this to other gcc jobs as needed.
  Ref: b23ce2cee7329bbf425f18b49973b7a5f23dfcb4 (2022-09-23) #9542

[1] https://dev.azure.com/daniel0244/curl/_build/results?buildId=18581&view=logs&jobId=ccf9cc6d-2ef1-5cf2-2c09-30f0c14f923b
[2] https://github.com/curl/curl/actions/runs/6896854263/job/18763831142?pr=12346#step:6:67
[3] https://github.com/curl/curl/actions/runs/6896854253/job/18763839238?pr=12346#step:30:214
[4] https://github.com/curl/curl/actions/runs/6896854253/job/18763838007?pr=12346#step:29:895
[5] https://github.com/curl/curl/actions/runs/6896854253/job/18763836775?pr=12346#step:33:1689

Closes #12346

### DIFF
--- a/docs/examples/ftpgetinfo.c
+++ b/docs/examples/ftpgetinfo.c
@@ -26,6 +26,12 @@
 
 #include <curl/curl.h>
 
+#if defined(__MINGW32__)
+#include <inttypes.h>
+#undef CURL_FORMAT_CURL_OFF_T
+#define CURL_FORMAT_CURL_OFF_T PRId64
+#endif
+
 /* <DESC>
  * Checks a single file's size and mtime from an FTP server.
  * </DESC>
--- a/docs/examples/ftpgetinfo.c
+++ b/docs/examples/ftpgetinfo.c
@@ -26,12 +26,6 @@
 
 #include <curl/curl.h>
 
-#if defined(__MINGW32__)
-#include <inttypes.h>
-#undef CURL_FORMAT_CURL_OFF_T
-#define CURL_FORMAT_CURL_OFF_T PRId64
-#endif
-
 /* <DESC>
  * Checks a single file's size and mtime from an FTP server.
  * </DESC>
--- a/lib/vssh/libssh.c
+++ b/lib/vssh/libssh.c
@@ -1159,13 +1159,23 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
         break;
       }
       else if(statvfs) {
+        #ifdef _MSC_VER
+        #define LIBSSH_VFS_SIZE_MASK "llu"
+        #else
+        #define LIBSSH_VFS_SIZE_MASK PRIu64
+        #endif
         char *tmp = aprintf("statvfs:\n"
-                            "f_bsize: %llu\n" "f_frsize: %llu\n"
-                            "f_blocks: %llu\n" "f_bfree: %llu\n"
-                            "f_bavail: %llu\n" "f_files: %llu\n"
-                            "f_ffree: %llu\n" "f_favail: %llu\n"
-                            "f_fsid: %llu\n" "f_flag: %llu\n"
-                            "f_namemax: %llu\n",
+                            "f_bsize: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_frsize: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_blocks: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_bfree: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_bavail: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_files: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_ffree: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_favail: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_fsid: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_flag: %" LIBSSH_VFS_SIZE_MASK "\n"
+                            "f_namemax: %" LIBSSH_VFS_SIZE_MASK "\n",
                             statvfs->f_bsize, statvfs->f_frsize,
                             statvfs->f_blocks, statvfs->f_bfree,
                             statvfs->f_bavail, statvfs->f_files,
--- a/lib/vssh/libssh.c
+++ b/lib/vssh/libssh.c
@@ -1160,7 +1160,7 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
       }
       else if(statvfs) {
         #ifdef _MSC_VER
-        #define LIBSSH_VFS_SIZE_MASK "llu"
+        #define LIBSSH_VFS_SIZE_MASK "I64u"
         #else
         #define LIBSSH_VFS_SIZE_MASK PRIu64
         #endif
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -105,7 +105,7 @@ jobs:
         - name: openssl3
           install_packages: zlib1g-dev valgrind
           install_steps: gcc-11 openssl3 pytest
-          configure: LDFLAGS="-Wl,-rpath,$HOME/openssl3/lib64" --with-openssl=$HOME/openssl3 --enable-debug --enable-websockets
+          configure: CFLAGS=-std=gnu89 LDFLAGS="-Wl,-rpath,$HOME/openssl3/lib64" --with-openssl=$HOME/openssl3 --enable-debug --enable-websockets
           singleuse: --unit
 
         - name: openssl3-O3
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -201,9 +201,10 @@
 #  define CURL_TYPEOF_CURL_SOCKLEN_T int
 
 #elif defined(__MINGW32__)
+#  include <inttypes.h>
 #  define CURL_TYPEOF_CURL_OFF_T     long long
-#  define CURL_FORMAT_CURL_OFF_T     "I64d"
-#  define CURL_FORMAT_CURL_OFF_TU    "I64u"
+#  define CURL_FORMAT_CURL_OFF_T     PRId64
+#  define CURL_FORMAT_CURL_OFF_TU    PRIu64
 #  define CURL_SUFFIX_CURL_OFF_T     LL
 #  define CURL_SUFFIX_CURL_OFF_TU    ULL
 #  define CURL_TYPEOF_CURL_SOCKLEN_T socklen_t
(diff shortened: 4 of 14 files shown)


docs/examples/ftpgetinfo.c
```diff
@@ -26,6 +26,12 @@
 
 #include <curl/curl.h>
 
+#if defined(__MINGW32__)
+#include <inttypes.h>
+#undef CURL_FORMAT_CURL_OFF_T
+#define CURL_FORMAT_CURL_OFF_T PRId64
+#endif
```
bagder, line 33: Is this not something we can/should fix in curl/system.h ?

lib/vssh/libssh.c
```diff
@@ -1159,13 +1159,23 @@ static CURLcode myssh_statemach_act(struct Curl_easy *data, bool *block)
         break;
       }
       else if(statvfs) {
+        #ifdef _MSC_VER
+        #define LIBSSH_VFS_SIZE_MASK "llu"
+        #else
+        #define LIBSSH_VFS_SIZE_MASK PRIu64
+        #endif
```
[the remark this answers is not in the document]
MarcelRaad: It's a C99 thing:
https://en.cppreference.com/w/c/types/integer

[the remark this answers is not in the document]
vszakats: In `libssh.h` v0.10.4 there is this, though it has changed since and [now](https://gitlab.com/libssh/libssh-mirror/-/blob/2df232463802203e2d675d4f99c65776a6f6b3c4/include/libssh/libssh.h#L54) includes `inttypes.h`, making it safe (probably) to use `PRIu64` everywhere:
```c
#ifdef _MSC_VER
  /* Visual Studio hasn't inttypes.h so it doesn't know uint32_t */
  typedef int int32_t;
  typedef unsigned int uint32_t;
  typedef unsigned short uint16_t;
  typedef unsigned char uint8_t;
  typedef unsigned long long uint64_t;
  typedef int mode_t;
#else /* _MSC_VER */
  #include <unistd.h>
  #include <inttypes.h>
  #include <sys/types.h>
#endif /* _MSC_VER */
```

Though for `_MSC_VER` I just left the value unchanged, as I didn't want to disturb what we had there (and cannot test it. CI also doesn't test it.).

docs/examples/ftpgetinfo.c
vszakats: We probably should. But it'd mean to add `inttypes.h` to our public headers, which isn't C89, and I didn't want to make such decision for the sake of fixing these warnings.

Also it's not so simple because this header requires `Visual Studio 2013`, while are still testing for 2010 and even 2008. I'm not sure what we support officially, maybe 2010, which we offer build files for in `projects/Windows/VC10`.

MarcelRaad: <inttypes.h> and the C99 types do actually exist in VS 2010. They were only added for C++ later (via <cstdint> in the std namespace).

vszakats: I based the VC 2013 cutoff version on CMake feature detection results, and it failed to detect `inttypes.h` with VC 2010, e.g.: https://ci.appveyor.com/project/libssh2org/libssh2/builds/46925869/job/wf6g47pff8xg55c3#L84

[ It's possible it's available for VC 2012, too. That version wasn't CI tested, so I went for the safest next version, 2013. I might have made special tests for it, can't remember now. ]

Does it need anything special to use it with 2010?

MarcelRaad: Ah, sorry, you're right! The C99 types in stdint.h were available in VS 2010, but not inttypes.h. I confused the two headers.

vszakats: No worries, thanks for clarifying!

vszakats: Actually the bigger concern is not MSVC here, but mingw. For internal purposes we know that we have mingw-w64, which does have `inttypes.h` support. I have no idea if this is also true for old-mingw.

Let's see how it goes in CI.

vszakats: Updated the public headers, with the caveat that I have no idea about `inttypes.h` support in old/classic/legacy mingw.

Let's see how it goes in CI.

vszakats:
Looked up the history of this option. Added in b23ce2cee7329bbf425f18b49973b7a5f23dfcb4 (last year) via #9542, with the goal of testing for C89 compliance. I think this would be better achieved by passing `-std=C89` via `CFLAGS` in specific CI jobs where we want to test for this.

vszakats:
Restored `-std=gnu89` as `CFLAGS` in the linux / openssl3 CI job.

vszakats:
The old old-mingw header package (from 2010) contains `inttypes.h`:
https://sourceforge.net/projects/mingw/files/MinGW/Base/mingwrt/mingwrt-3.18/ in `mingwrt-3.18-mingw32-dev.tar.gz`

It means old-mingw likely continues to work fine with public headers, though I haven't actually tried.

lib/vssh/libssh.c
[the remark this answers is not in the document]
jay: _msc_ver anything that can build curl will have __int64 (and unsigned __int64) and `%I64d` (and `%I64u`). `%llu` is not in older versions of the crt 

https://github.com/curl/curl/blob/d755a5f7c009dd63a61b2c745180d8ba937cbfeb/include/curl/system.h#L372-L385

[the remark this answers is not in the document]
vszakats: Meaning this mask was broken with MSVC all along?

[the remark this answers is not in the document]
jay: > Meaning this mask was broken with (old) MSVC all along?

Yes. It will compile without warning in older versions of visual studio but its output is the same as %lu

vszakats:
I couldn't replicate the same issue with libssh2 [1](https://github.com/curl/curl/actions/runs/6916746209/job/18817148474#step:29:897). I cannot grasp it for libssh as well, as it seems rooted from the fact that `uint64_t` somehow becomes `long unsigned int` ["uint64_t {aka long unsigned int}"](https://github.com/curl/curl/actions/runs/6896854253/job/18763838007?pr=12346#step:29:895) in that CI job.

Anyway, for libssh2 the MSVC-issue seems to apply too:
https://github.com/curl/curl/blob/a9fd0d0083a534ef1efa82e7546846776cbd7373/lib/vssh/libssh2.c#L1965-L1971

Any thoughts to include this MSVC fix for libssh2 too?:
```diff
diff --git a/lib/vssh/libssh2.c b/lib/vssh/libssh2.c
index 543b6fb43..4eb5cf63f 100644
--- a/lib/vssh/libssh2.c
+++ b/lib/vssh/libssh2.c
@@ -1962,13 +1962,23 @@ static CURLcode ssh_statemach_act(struct Curl_easy *data, bool *block)
         break;
       }
       else if(rc == 0) {
+        #ifdef _MSC_VER
+        #define LIBSSH2_VFS_SIZE_MASK "I64u"
+        #else
+        #define LIBSSH2_VFS_SIZE_MASK "llu"
+        #endif
         char *tmp = aprintf("statvfs:\n"
-                            "f_bsize: %llu\n" "f_frsize: %llu\n"
-                            "f_blocks: %llu\n" "f_bfree: %llu\n"
-                            "f_bavail: %llu\n" "f_files: %llu\n"
-                            "f_ffree: %llu\n" "f_favail: %llu\n"
-                            "f_fsid: %llu\n" "f_flag: %llu\n"
-                            "f_namemax: %llu\n",
+                            "f_bsize: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_frsize: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_blocks: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_bfree: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_bavail: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_files: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_ffree: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_favail: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_fsid: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_flag: %" LIBSSH2_VFS_SIZE_MASK "\n"
+                            "f_namemax: %" LIBSSH2_VFS_SIZE_MASK "\n",
                             statvfs.f_bsize, statvfs.f_frsize,
                             statvfs.f_blocks, statvfs.f_bfree,
                             statvfs.f_bavail, statvfs.f_files,
```

vszakats:
To make libssh/libssh2 masks the same and future-proof, I'm thinking of moving from:
libssh:
```
        #ifdef _MSC_VER
        #define LIBSSH_VFS_SIZE_MASK "I64u"
        #else
        #define LIBSSH_VFS_SIZE_MASK PRIu64
        #endif
```
libssh2:
```
        #ifdef _MSC_VER
        #define LIBSSH2_VFS_SIZE_MASK "I64u"
        #else
        #define LIBSSH2_VFS_SIZE_MASK "llu"
        #endif
```
to this for both:
```
        #ifdef PRIu64
        #define LIBSSH_VFS_SIZE_MASK PRIu64
        #elif defined(_MSC_VER)
        #define LIBSSH_VFS_SIZE_MASK "I64u"
        #else
        #define LIBSSH_VFS_SIZE_MASK "llu"
        #endif
```
This would use the official `uint64_t` mask whenever available from `inttypes.h` either in via `libssh.h` (next release will include it for all targets) or `curl/system.h`. And fall back to the universal MSVC value, or the universal standard one for others.

PR #12349, opened 2023-11-17
vtls/vquic, keep peer name information together

icing:
- add `struct ssl_peer` to keep hostname, dispname and sni for a filter
- allocate `sni` for use in VTLS backend
- eliminate `Curl_ssl_snihost()` and its use of the download buffer
- use ssl_peer in SSL and QUIC filters

### DIFF
--- a/lib/urldata.h
+++ b/lib/urldata.h
@@ -266,6 +266,13 @@ typedef enum {
 /* SSL backend-specific data; declared differently by each SSL backend */
 struct ssl_backend_data;
 
+struct ssl_peer {
+  char *hostname;        /* hostname for verification */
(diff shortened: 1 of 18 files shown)


bagder (approved):

vszakats:
After this patch we have a new warning in this CI job:
https://ci.appveyor.com/project/curlorg/curl/builds/48559761/job/y4v0u7r42kqj9auw#L213
```
2>.\vtls\schannel.c(1067) : error C2220: warning treated as error - no 'object' file generated
2>.\vtls\schannel.c(1067) : warning C4189: 'hostname' : local variable is initialized but not reference
```

PR #12358, opened 2023-11-19
autotools: show detected clang/gcc/icc compiler version

vszakats:
Also show if we detected Apple clang.

Closes #12358


vszakats:
Merging this into #12362.

PR #12359, opened 2023-11-19
build: check for C99 extensions with clang

vszakats:
Use `-Wc99-extensions` to warn for non-C89 syntax.

Follow-up to b23ce2cee7329bbf425f18b49973b7a5f23dfcb4 #9542

Closes #12359

---

Expected it to detect `bool` use, but it didn't.

Like here, which was a result of a compiler detection mixup, ending up enabling this option and detecting `bool` uses: https://github.com/curl/curl-for-win/actions/runs/6831206570/job/18580363613#step:3:6876


bagder:
> Expected it to detect bool use, but it didn't.

We have our own `bool` typedef as a fallback in `lib/curl_setup_once.h`, it does not need C99.

vszakats:
Thanks, looking good then, closing the experiment.

PR #12362, opened 2023-11-19
autotools: fix/improve gcc and Apple clang version detection

vszakats:
- Before this patch we expected `n.n` `-dumpversion` output, but Ubuntu
  may return `n-win32` (also with `-dumpfullversion`). Causing these
  errors and failing to enable picky warnings:
  ```
  ../configure: line 23845: test: : integer expression expected
  ```
  Ref: https://github.com/libssh2/libssh2/actions/runs/6263453828/job/17007893718#step:5:143

  Fix that by stripping any dash-suffix and handling a dotless (major-only)
  version number by assuming `.0` in that case.

  `9.3-posix`, `9.3-win32`, `6`, `9.3.0`, `11`, `11.2`, `11.2.0`
  Ref: https://github.com/mamedev/mame/pull/9767

- fix Apple clang version detection for releases between
  'Apple LLVM version 7.3.0' and 'Apple LLVM version 10.0.1' where the
  version was under-detected as 3.7 llvm/clang equivalent.

- fix Apple clang version detection for 'Apple clang version 11.0.0'
  and newer where the Apple clang version was detected, instead of its
  llvm/clang equivalent.

- display detected clang/gcc/icc compiler version.

Via libssh2:
- https://github.com/libssh2/libssh2/commit/00a3b88c51cdb407fbbb347a2e38c5c7d89875ad
  https://github.com/libssh2/libssh2/pull/1187
- https://github.com/libssh2/libssh2/commit/89ccc83c7da73e7ca3a112e350008131942b592e
  https://github.com/libssh2/libssh2/pull/1232

Closes #12362

### DIFF
--- a/m4/curl-compilers.m4
+++ b/m4/curl-compilers.m4
@@ -93,18 +93,37 @@ AC_DEFUN([CURL_CHECK_COMPILER_CLANG], [
     fi
     AC_MSG_CHECKING([compiler version])
     fullclangver=`$CC -v 2>&1 | grep version`
+    if echo $fullclangver | grep 'Apple' >/dev/null; then
+      appleclang=1
+    else
+      appleclang=0
+    fi
     clangver=`echo $fullclangver | grep "based on LLVM " | "$SED" 's/.*(based on LLVM \(@<:@0-9@:>@*\.@<:@0-9@:>@*\).*)/\1/'`
     if test -z "$clangver"; then
-      if echo $fullclangver | grep "Apple LLVM version " >/dev/null; then
-        dnl Starting with Xcode 7 / clang 3.7, Apple clang won't tell its upstream version
-        clangver="3.7"
-      else
-        clangver=`echo $fullclangver | "$SED" 's/.*version \(@<:@0-9@:>@*\.@<:@0-9@:>@*\).*/\1/'`
-      fi
+      clangver=`echo $fullclangver | "$SED" 's/.*version \(@<:@0-9@:>@*\.@<:@0-9@:>@*\).*/\1/'`
+      oldapple=0
+    else
+      oldapple=1
     fi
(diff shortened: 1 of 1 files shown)


bagder:
> gcc version detection is still half broken because we translate '10' to '10.10' because cut -d. -f2 returns the first word if the delimiter missing.

Is cut's `-s` portable enough to be used in the `-f2` invocation ? *"do not print lines not containing delimiters"*

vszakats:
`-s` is there on macOS and GNU, and part of POSIX.2: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cut.html

But, we need to fill the blank with a zero, so this is what I was experiment with instead:
```diff
--- a/m4/curl-compilers.m4
+++ b/m4/curl-compilers.m4
@@ -160,7 +160,11 @@ AC_DEFUN([CURL_CHECK_COMPILER_GNU_C], [
     compiler_id="GNU_C"
     gccver=`$CC -dumpversion`
     gccvhi=`echo $gccver | cut -d . -f1`
-    gccvlo=`echo $gccver | cut -d . -f2`
+    if echo $gccver | grep -F "." >/dev/null; then
+      gccvlo=`echo $gccver | cut -d . -f2`
+    else
+      gccvlo="0"
+    fi
     compiler_num=`(expr $gccvhi "*" 100 + $gccvlo) 2>/dev/null`
     flags_dbg_yes="-g"
     flags_opt_all="-O -O0 -O1 -O2 -O3 -Os -Og -Ofast"
```

vszakats:
I'm also testing a fix for Xcode version detection: https://github.com/libssh2/libssh2/pull/1232

vszakats:
Merged #12358 into this PR, and also applied the Xcode version fixes from https://github.com/libssh2/libssh2/pull/1232.

PR #12363, opened 2023-11-20
transfer: avoid calling the read callback again after EOF

bagder:
Regression since 7f43f3dc5994d01b12 (7.84.0)

Bug: https://curl.se/mail/lib-2023-11/0017.html

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -231,10 +231,16 @@ CURLcode Curl_fillreadbuffer(struct Curl_easy *data, size_t bytes,
     extra_data = data->state.in;
   }
 
-  Curl_set_in_callback(data, true);
-  nread = readfunc(data->req.upload_fromhere, 1,
-                   buffersize, extra_data);
-  Curl_set_in_callback(data, false);
+  if(!data->req.fread_func_eof) {
+    Curl_set_in_callback(data, true);
+    nread = readfunc(data->req.upload_fromhere, 1,
+                     buffersize, extra_data);
+    Curl_set_in_callback(data, false);
+    /* make sure the callback is not called again after EOF */
+    data->req.fread_func_eof = !nread;
+  }
+  else
+    nread = 0;
 
   if(nread == CURL_READFUNC_ABORT) {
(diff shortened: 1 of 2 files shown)


lib/transfer.c
```diff
@@ -231,10 +231,16 @@ CURLcode Curl_fillreadbuffer(struct Curl_easy *data, size_t bytes,
...
-  nread = readfunc(data->req.upload_fromhere, 1,
-                   buffersize, extra_data);
-  Curl_set_in_callback(data, false);
+  if(!data->req.fread_func_eof) {
+    Curl_set_in_callback(data, true);
+    nread = readfunc(data->req.upload_fromhere, 1,
+                     buffersize, extra_data);
+    Curl_set_in_callback(data, false);
+    /* make sure the callback is not called again after EOF */
+    data->req.fread_func_eof = (nread != buffersize);
```
jay, line 240: that is going to set it for situations other than eof, i'm curious why you didn't do `data->req.fread_func_eof = !nread;`. like what if there's a pause for example

bagder: Ah right, pause.... 

Well I did it of course because returning an error should not call it again either.

jay: it's not just that there are other scenarios like libcurl requests 100 bytes and 50 bytes or something is returned

lib/urldata.h
```diff
@@ -753,6 +753,7 @@ struct SingleRequest {
                         specific upload buffers. See readmoredata() in http.c
                         for details. */
   BIT(no_body);      /* the response has no body */
+  BIT(fread_func_eof); /* the read callback has returned EOF */
```
jay, line 756: are there any scenarios where this needs to cleared like code that calls the seek function

bagder: Good thinking. I would say there is. I'll check what test coverage we have for this. Ideally this PR should fail a test where we rewind...

bagder: The `SingleRequest` struct is cleared first thing in `Curl_connect` so it seems good.

PR #4855, opened 2020-01-27
curl.h: define CURL_WIN32 on windows

bagder:
... so that the subsequent logic below can use a single known define to know
when built on Windows (as we don't define WIN32 anymore).

Follow-up to 1adebe7886ddf20b

Reported-by: crazydef on github
Fixes #4854

Issue #4854, opened 2020-01-27
Removing WIN32 define breaks Windows builds


crazydef:
### operating system: Windows 10

The commit that removes the WIN32 define didn't also remove all references to WIN32 in other places, so now nothing builds on Windows.

I understand the desire to remove WIN32, but existing usage of that macro should have been changed to use the compiler defined _WIN32 at the same time.  As things now stand, this non-standard macro needs to be declared externally when building *any project* that includes curl.h on Windows.

bagder:
> now nothing builds on Windows.

I'm afraid that's wrong. All of our CI builds on Windows still build perfectly. You need to be much more specific and explain exactly what doesn't build and preferably why.

`WIN32` should not be defined by curl headers.

crazydef:
Your CI builds probably still build because your toolchain (cmake, by any chance?) defines WIN32 itself.  Other build systems (like Visual Studio, for example) do not.  Any project which includes curl.h is now forced to define WIN32 on Windows.

crazydef:
I agree WIN32 shouldn't be defined by Curl headers.  In the same way I agree that the Curl headers shouldn't try to rely on non-standard, user-defined macros to detect the platform.

bagder:
> Other build systems (like Visual Studio, for example) do not

You mean using the files under projects/Windows ? Then we should fix those builds.

> In the same way I agree that the Curl headers shouldn't try to rely on non-standard, user-defined macros to detect the platform.

Feel free to suggest and provide improvements to our headers and other code. We only do the best we can and there are always room for improvements.

crazydef:
No, I mean projects that *use* curl.  Not any code in the curl directory.  (Although that might be affected too.)

I'm not talking about building curl, I'm talking about including curl.h in my projects.

bagder:
Why should curl's headers be responsible for identifying win32 for your projects?

crazydef:
Because curl requires WIN32 to be declared.  It used to declare it itself if _WIN32 was declared, now it assumes it's declared externally.  Curl shouldn't require the user to define the platform.  That's the job of compiler writers and OS vendors.  Curl should be using _WIN32 (or whatever) to detect the platform.  I.e, the standard platform preprocessor definitions.

As for a recommendation, In system.h I would add CURL_WIN32, CURL_MACOS, etc platform definitions and use those throughout the rest of the curl source code.

bagder:
Le me try to decipher this.

[...]

PR #12376, opened 2023-11-21
windows: use built-in `_WIN32` macro to detect Windows

vszakats:
Windows compilers define `_WIN32` automatically. Windows SDK headers
or build env defines `WIN32`, or we have to take care of it. The
agreement seems to be that `_WIN32` is the preferred practice here.
Make the source code rely on that to detect we're building for Windows.

Public `curl.h` was using `WIN32`, `__WIN32__` and `CURL_WIN32` for
Windows detection, next to the official `_WIN32`. After this patch it
only uses `_WIN32` for this. Also, make it stop defining `CURL_WIN32`.

There is a slight chance these break compatibility with Windows
compilers that fail to define `_WIN32`. I'm not aware of any obsolete
or modern compiler affected, but in case there is one, one possible
solution is to define this macro manually.

grepping for `WIN32` remains useful to discover Windows-specific code.

Also:

- extend `checksrc` to ensure we're not using `WIN32` anymore.

- apply minor formatting here and there.

- delete unnecessary checks for `!MSDOS` when `_WIN32` is present.

Closes #12376

### DIFF
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -57,7 +57,9 @@
  * Define CURL_WIN32 when build target is Win32 API
  */
 
-#if (defined(_WIN32) || defined(__WIN32__) || defined(WIN32)) &&        \
+#if (defined(_WIN32) || \
+     defined(WIN32) /* !checksrc! disable BANNEDPREPROC 1 */ || \
+     defined(__WIN32__)) && \
   !defined(__SYMBIAN32__)
 #define CURL_WIN32
 #endif
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -53,17 +53,6 @@
 #include "curlver.h"         /* libcurl version defines   */
 #include "system.h"          /* determine things run-time */
 
-/*
- * Define CURL_WIN32 when build target is Win32 API
- */
-
-#if (defined(_WIN32) || \
-     defined(WIN32) /* !checksrc! disable BANNEDPREPROC 1 */ || \
-     defined(__WIN32__)) && \
-  !defined(__SYMBIAN32__)
-#define CURL_WIN32
-#endif
-
 #include <stdio.h>
 #include <limits.h>
 
@@ -76,7 +65,7 @@
 #include <sys/types.h>
 #include <time.h>
 
-#if defined(CURL_WIN32) && !defined(_WIN32_WCE) && !defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(_WIN32_WCE) && !defined(__CYGWIN__)
 #if !(defined(_WINSOCKAPI_) || defined(_WINSOCK_H) || \
       defined(__LWIP_OPT_H__) || defined(LWIP_HDR_OPT_H))
 /* The check above prevents the winsock2 inclusion if winsock.h already was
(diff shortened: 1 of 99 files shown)


jay:
My initial reaction to this was it's unnecessary, even though it is "more correct". We already use WIN32 and it works fine. However it is, yes, more correct. My biggest concern I think is that we will occasionally forget WIN32 is now _WIN32 and make the mistake of using the former. To that end I've added a checksrc warning category BANNEDPREPROC that checks for banned symbols (eg WIN32) on preprocessor lines.

vszakats:
Good idea to make `checksrc` check for this.

I'm thinking to sync `curl/curl.h` with this and replace
`CURL_WIN32` with `_WIN32`, to avoid the minor chance of
it disagreeing with what 'win32' is in the internals and causing
issues.

`CURL_WIN32` seems to be a documented feature of `curl.h` 
since 7.69.0, so we might want to keep it around for compatibility,
if we're convinced this is useful or necessary. I tend to think it isn't.

This would be a reboot of 1adebe7886ddf20b0733bf9ccbae4ed4866dcfb6
and 8bd863f97b6c79f561bc063e634cecdf4badf776 + #4854 + #4855, but
instead of keeping around `CURL_WIN32` as a replacement for our local
`WIN32` we just rely on `_WIN32`, which works universally out of the box.

What do you think?

jay:
I struggle to think of when _WIN32 isn't available. Let's see if anyone else has anything to say on it...

Ref: https://github.com/curl/curl/pull/4855

vszakats:
Some references:
https://web.archive.org/web/20190205162154/nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#WindowsCygwinnonPOSIXandMinGW
https://github.com/cpredef/predef/blob/master/OperatingSystems.md#windows

One list mentions Borland C++, so I made a test with 
its 5.5.1 version from 2000, and it _does_ define `_WIN32`.
I also tried Open Watcom 1.9 from 2010 and it does too.
Pelles C most definitely had it because it was emulating
MSVC in most aspects.

Can't think of more niche compilers that are still around.

But if there are, they can almost certainly set `-D_WIN32`
manually at build time. Not just for curl.

include/curl/curl.h
```diff
@@ -57,7 +57,9 @@
  * Define CURL_WIN32 when build target is Win32 API
  */
 
-#if (defined(_WIN32) || defined(__WIN32__) || defined(WIN32)) &&        \
+#if (defined(_WIN32) || \
+     defined(WIN32) /* !checksrc! disable BANNEDPREPROC 1 */ || \
+     defined(__WIN32__)) && \
   !defined(__SYMBIAN32__)
```
vszakats, line 63: It's curious that `_WIN32` or `__WIN32__` could end up together with `__SYMBIAN32__`,
so that it had to be guarded-out.

Added in 1960eebc2d021ecf5ffc3f6d4e935d54aa592c72 (2008). Anyhow, Symbian 
was deprecated in curl in 3d64031fa7a80ac4ae3fd09a5939196268b92f81 (2020), so
this no longer a concern.

vszakats: Ref: #12378

dfandrich: > It's curious that `_WIN32` or `__WIN32__` could end up together with `__SYMBIAN32__`, so that it had to be guarded-out.

It's been too long for me to recall for certain, but it may have been for x86 builds targeting the Symbian emulator running on Win32.

bagder (approved):
I'm fine with also removing `CURL_WIN32` from the public header. We never documented it for any purpose. It is mentioned in `symbols-in-version` simply because it is present in `curl.h` but I don't think we break any promises (API/ABI) by taking it away again.

PR #12377, opened 2023-11-21
ci: Add minimal permissions to distcheck.yml

diogoteles08:
Hi! Bringing a simple security change, similar to the one on #11055.

Additionally, I'd like to suggest that you take a look into changing your repository settings to [set read-only permission as default permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository) for newly created workflows. If you make sure the default permissions are restricted, there will be no need to worry if we forget to explicitly list the permissions for any newly created workflow =)
After this change, any workflow that require write permissions will need to have the write permissions explicitly defined in the workflow file.

### DIFF
--- a/.github/workflows/distcheck.yml
+++ b/.github/workflows/distcheck.yml
@@ -17,6 +17,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   maketgz-and-verify-in-tree:
     runs-on: ubuntu-latest


bagder:
AFAICS, it is already set to read-only:

Screenshot 2023-11-21 at 23-58-21 General · curl_curl

diogoteles08:
Great! Thanks for the information (it's not public so I couldn't retrieve it by myself)

Feel free to close this PR then. There is still some value of having the permissions documented on the workflows, but it's not a security need anymore.

Issue #12380, opened 2023-11-22
curl 8.3 build failing with OPENSSL 3 due to absence of DES functions


devdatt9129:
### I did this

I tried to configure and build curl 8.3 with openssl 3, but i am getting the following error
```
curl_ntlm_core.c:113:4: error: #error "Can't compile NTLM support without a crypto library with DES."
make: *** [libcurl_la-curl_ntlm_core.lo] Error 1
```

### curl/libcurl version

curl 8.3.0

### operating system

RHEL 6

bagder:
It seems you have an OpenSSL build with DES support disabled(?), which is not what OpenSSL 3 does by default. Since NTLM support in curl needs DES, it fails to build.

I don't think we can make curl's configure adapt to every possible OpenSSL build combination. If you want to build curl to not use DES, I think you need to disable NTLM from your build.

devdatt9129:
I built my openssl using this command, i havent specfied no-des anywhere
perl Configure --prefix=/usr/local/lib64 --openssldir=/usr/local/lib64 enable-fips no-shared -m64 -Wa,--noexecstack linux-x86_64 threads no-idea no-mdc2 no-dsa no-rc5 no-ssl3 enable-asm enable-camellia enable-seed enable-rfc3779 enable-cms --api=3.0.0 no-deprecated  

does this have anything to do with the error

bagder:
I suspect maybe `no-deprecated` does it?

devdatt9129:
Okay ill try removing it once
Also, will disabling ntlm have any effect in runtime on my solution?

bagder:
>  Also, will disabling ntlm have any effect in runtime on my solution?

If you need to use NTLM sure, otherwise no.

vszakats:
I've been making `no-deprecated` tests since yesterday and with curl-for-win settings `lib/curl_ntlm_core.c` [compiles fine](https://ci.appveyor.com/project/curlorg/curl-for-win/builds/48587418?fullLog=true#L7297). This source has protections to auto-disable with no DES or 3.0 deprecated functions. It seems this isn't enough for all build configs, but I wonder why.

On the other hand my build bumps into a different error:
```
../../lib/vtls/openssl.c:4097:3: error: call to undeclared function 'OpenSSL_add_all_algorithms'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  OpenSSL_add_all_algorithms();
  ^
[...]

PR #12384, opened 2023-11-22
openssl: fix building with v3 `no-deprecated` + add CI test

vszakats:
- build quictls with `no-deprecated` in CI to have test coverage for
  this OpenSSL 3 configuration.

- don't call `OpenSSL_add_all_algorithms()`, `OpenSSL_add_all_digests()`.
  The caller code is meant for OpenSSL 3, while these two functions were
  only necessary before OpenSSL 1.1.0. They are missing from OpenSSL 3
  if built with option `no-deprecated`, causing build errors:
  ```
  vtls/openssl.c:4097:3: error: call to undeclared function 'OpenSSL_add_all_algorithms'; ISO C99 and later do not   support implicit function declarations [-Wimplicit-function-declaration]
  vtls/openssl.c:4098:3: error: call to undeclared function 'OpenSSL_add_all_digests'; ISO C99 and later do not   support implicit function declarations [-Wimplicit-function-declaration]
  ```
  Ref: https://ci.appveyor.com/project/curlorg/curl-for-win/builds/48587418?fullLog=true#L7667

  Regression from b6e6d4ff8f253c8b8055bab9d4d6a10f9be109f3 #12030
  Bug: https://github.com/curl/curl/issues/12380#issuecomment-1822944669

- vquic/curl_ngtcp2: fix using `SSL_get_peer_certificate` with
  `no-deprecated` quictls 3 builds.
  Do it by moving an existing solution for this from `vtls/openssl.c`
  to `vtls/openssl.h` and adjusting caller code.
  ```
  vquic/curl_ngtcp2.c:1950:19: error: implicit declaration of function 'SSL_get_peer_certificate'; did you mean   'SSL_get1_peer_certificate'? [-Wimplicit-function-declaration]
  ```
  Ref: https://github.com/curl/curl/actions/runs/6960723097/job/18940818625#step:24:1178

- curl_ntlm_core: fix `-Wunused-parameter`, `-Wunused-variable` and
  `-Wunused-function` when trying to build curl with NTLM enabled but
  without the necessary TLS backend (with DES) support.

Closes #12384

---

- curl_ntlm_core: allow NTLM with `no-deprecated` OpenSSL 3 builds.
  Though OpenSSL 3 deprecated the DES functions needed for the NTLM
  feature, its `no-deprecated` build option doesn't actually disable
  these DES functions. It means we can use them for NTLM even if
  `OPENSSL_NO_DEPRECATED_3_0` is set. Adjust our logic to allow it.
  This failed tests due to some types missing. Dropping this.
  https://github.com/curl/curl/actions/runs/6961417865/job/18942957524?pr=12384#step:24:132

### DIFF
--- a/.github/workflows/ngtcp2-linux.yml
+++ b/.github/workflows/ngtcp2-linux.yml
@@ -107,7 +107,7 @@ jobs:
         cd $HOME
         git clone --quiet --depth=1 -b openssl-${{ env.quictls-version }} https://github.com/quictls/openssl quictls
         cd quictls
-        ./config --prefix=$HOME/nghttpx --libdir=$HOME/nghttpx/lib
+        ./config no-deprecated --prefix=$HOME/nghttpx --libdir=$HOME/nghttpx/lib
         make
       name: 'build quictls'
 
--- a/.github/workflows/ngtcp2-linux.yml
+++ b/.github/workflows/ngtcp2-linux.yml
@@ -66,7 +66,7 @@ jobs:
         - name: quictls
           configure: >-
             PKG_CONFIG_PATH="$HOME/nghttpx/lib/pkgconfig" LDFLAGS="-Wl,-rpath,$HOME/nghttpx/lib"
-            --with-ngtcp2=$HOME/nghttpx --enable-warnings --enable-werror --enable-debug
+            --with-ngtcp2=$HOME/nghttpx --enable-warnings --enable-werror --enable-debug --disable-ntlm
             --with-test-nghttpx="$HOME/nghttpx/bin/nghttpx"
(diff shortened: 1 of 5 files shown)


ajbozarth:
Thanks for the catch, I missed that those were deprecated when we added the version checks

vszakats:
@ajbozarth: These calls seem redundant, but can you confirm their removal didn't break this feature?

ajbozarth (commented):
I've checkout and run the code and it's still fully functional

LGTM (can't make this an approve cause of perms)

vszakats:
Thank you @ajbozarth!

Issue #11832, opened 2023-09-11
http3: Can't fallback to http2 on Windows (CURLE_BAD_FUNCTION_ARGUMENT)


vvb2060:
### I did this

```
curl.exe --http3 https://crypto.cloudflare.com/cdn-cgi/trace --trace-ascii -
== Info: processing: https://crypto.cloudflare.com/cdn-cgi/trace
== Info:   Trying [2606:4700:7::a29f:8955]:443...
== Info:  CAfile: curl-ca-bundle.crt
== Info:  CApath: none
== Info:   Trying 162.159.137.85:443...
== Info:   Trying [2606:4700:7::a29f:8955]:443...
== Info:   Trying [2606:4700:7::a29f:8955]:443...
== Info:   Trying 162.159.137.85:443...
== Info: Connected to crypto.cloudflare.com (2606:4700:7::a29f:8955) port 443
== Info: ALPN: offers h2,http/1.1
=> Send SSL data, 5 bytes (0x5)
0000: .....
== Info: TLSv1.3 (OUT), TLS handshake, Client hello (1):
=> Send SSL data, 512 bytes (0x200)
0000: ........C`...x.6/f..x.L.....%..".m.[.. h...'.....{...[.......*QN
0040: c.a.....>.......,.0.........+./...$.(.k.#.'.g.....9.....3.....=.
0080: <.5./.....u.........crypto.cloudflare.com.......................
00c0: ..................h2.http/1.1.........1.....". .................
0100: ................+........-.....3.&.$... <......))......U...H4!(.
0140: ..y.$A.o........................................................
0180: ................................................................
01c0: ................................................................
== Info: Closing connection
curl: (43) A libcurl function was given a bad argument

```
```
curl.exe --http3-only https://crypto.cloudflare.com/cdn-cgi/trace --trace-ascii -
== Info: processing: https://crypto.cloudflare.com/cdn-cgi/trace
== Info:   Trying [2606:4700:7::a29f:8955]:443...
== Info:  CAfile: curl-ca-bundle.crt
== Info:  CApath: none
== Info:   Trying 162.159.137.85:443...
== Info:   Trying [2606:4700:7::a29f:8955]:443...
== Info:   Trying 162.159.137.85:443...
== Info: Closing connection
curl: (43) A libcurl function was given a bad argument
```

crypto.cloudflare.com does not support h3, but curl does not fallback to h2.

### I expected the following

Test on android:
```
[...]

PR #12400, opened 2023-11-24
QUIC eyeballers connect retries

icing:
- when a connect immediately goes into DRAINING state, do not attempt retries in the QUIC connection filter. Instead, return CURLE_WEIRD_SERVER_REPLY
- When eyeballing, interpret CURLE_WEIRD_SERVER_REPLY as an inconclusive answer. When all addresses have been attempted, rewind the address list once on an inconclusive answer.
- refs #11832 where connects were retried indefinitely until the overall timeout fired

### DIFF
--- a/lib/connect.c
+++ b/lib/connect.c
@@ -363,7 +363,7 @@ struct eyeballer {
   expire_id timeout_id;              /* ID for Curl_expire() */
   CURLcode result;
   int error;
-  int rewinds;                       /* how often we rewinded the addr list */
+  BIT(rewinded);                     /* if we rewinded the addr list */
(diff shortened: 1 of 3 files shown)


lib/connect.c
```diff
@@ -362,9 +363,12 @@ struct eyeballer {
   expire_id timeout_id;              /* ID for Curl_expire() */
   CURLcode result;
   int error;
+  int rewinds;                       /* how often we rewinded the addr list */
```
bagder, line 366: Since you only allow rewinding *once*, couldn't this be a single bit instead?

icing: changed

bagder (approved):

PR #12401, opened 2023-11-24
curl: output warning if curl_easy_setopt returns error

bagder:
... to help pinpoint the issue better.


bagder:
Doesn't really work for general use.

Issue #12327, opened 2023-11-15
libcurl.dll memory leak 


southernedge:
### I did this

Hello, I would like to contact the developers.
I noticed a memory leak in libcurl.dll (libcurl-x64.dll) a long time ago. This is version independent.
Here is a small [MS Visual Studio C test project](https://drive.google.com/file/d/1ydOD_vbKBYPtOjxIV4l4sVgtvSWzwYVc/view?usp=sharing) showing this problem.
Downloading a file occurs in an endless loop. To speed up the manifestation of memory leaks, the download occurs in 64 threads.
Leave the program running for 15-30 minutes and you will see how memory consumption will increase many times over.
If there is my error and I do not release curl resources in a timely manner, then please suggest a solution to this problem.
Thank you very much for your work, attention and time spent!

### curl/libcurl version

8.4.0

### operating system

Windows any

bagder:
Please provide a stand-alone source code file that reproduces the problem, ideally without Windows specific code so that we can try it elsewhere.

You say it leaks in **any** libcurl version and I find it hard to believe that you tried most of them. Which versions did you use?

southernedge:
For this test I used 8.4.0

bagder:
If you cannot make a smaller and less Windows-centric recipe (I don't use Windows), then perhaps you can instead give us much more details about this claimed leak. What memory is leaking? Where is it allocated? How much?

(We test for memory leaks extensively, in all versions and all commits all the time.)

gvanem:
You  do not seems to call `CInet::curl_global_cleanup()` anywhere and `LoadDll()` loads multiple libcurls.
And this:
```c
CInet::~CInet( void )
{
}
```

should at least call `FreeLibrary()`.

southernedge:
> If you cannot make a smaller and less Windows-centric recipe (I don't use Windows), then perhaps you can instead give us much more details about this claimed leak. What memory is leaking? Where is it allocated? How much?
> 
> (We test for memory leaks extensively, in all versions and all commits all the time.)

I cannot give you an exact answer to your questions. Memory is leaking inside the dll module. The Visual Studio debugger does not show memory leaks inside external dll modules.
I see the memory increase after each file download using Windows Task Manager. The memory grows with each file download and does not return to its begin state. 
The test program does not allocate memory at all. It's running idle.
It would be good if you took a quick look at the text of the module to see if there is an error on my part.

southernedge:
[...]

PR #12408, opened 2023-11-27
dllmain: Call OpenSSL thread cleanup for Windows and Cygwin

jay:
- Call OPENSSL_thread_stop on thread termination (DLL_THREAD_DETACH) to prevent a memory leak in case OpenSSL is linked statically.

- Warn in libcurl-thread.3 that in some cases OpenSSL may need a call to OPENSSL_thread_stop before thread termination in order to stop a memory leak.

OpenSSL may need per-thread cleanup to stop a memory leak. For Windows and Cygwin if libcurl was built as a DLL then we can do that for the user by calling OPENSSL_thread_stop on thread termination. However, if libcurl was built statically then we do not have notification of thread termination and cannot do that for the user.

Also, there are several other unusual cases where it may be necessary for the user to call OPENSSL_thread_stop, so in the libcurl-thread warning I added a link to the OpenSSL documentation.

Reported-by: southernedge@users.noreply.github.com
Reported-by: zmcx16@users.noreply.github.com

Ref: https://www.openssl.org/docs/man3.0/man3/OPENSSL_thread_stop.html#NOTES

Fixes https://github.com/curl/curl/issues/12327
Closes #xxxx

### DIFF
--- a/lib/dllmain.c
+++ b/lib/dllmain.c
@@ -1,0 +1,81 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+
+#include "curl_setup.h"
+
+#ifdef USE_OPENSSL
+#include <openssl/crypto.h>
+#endif
+
+/* The fourth-to-last include */
+#ifdef __CYGWIN__
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#ifdef _WIN32
+#undef _WIN32
+#endif
+#endif
+
+/* The last 3 #include files should be in this order */
+#include "curl_printf.h"
+#include "curl_memory.h"
+#include "memdebug.h"
+
+/* DllMain() must only be defined for Windows and Cygwin DLL builds. */
+#if (defined(_WIN32) || defined(__CYGWIN__)) && !defined(CURL_STATICLIB)
+
+#if defined(USE_OPENSSL) && \
+    !defined(OPENSSL_IS_AWSLC) && \
+    !defined(OPENSSL_IS_BORINGSSL) && \
+    !defined(LIBRESSL_VERSION_NUMBER) && \
+    (OPENSSL_VERSION_NUMBER >= 0x10100000L)
+#define PREVENT_OPENSSL_MEMLEAK
+#endif
+
+#ifdef PREVENT_OPENSSL_MEMLEAK
+BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved);
+BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
+{
+  (void)hinstDLL;
+  (void)lpvReserved;
+
+  switch(fdwReason) {
+  case DLL_PROCESS_ATTACH:
+    break;
+  case DLL_PROCESS_DETACH:
+    break;
+  case DLL_THREAD_ATTACH:
+    break;
+  case DLL_THREAD_DETACH:
+    /* Call OPENSSL_thread_stop to prevent a memory leak in case OpenSSL is
+       linked statically.
+       https://github.com/curl/curl/issues/12327#issuecomment-1826405944 */
+    OPENSSL_thread_stop();
+    break;
+  }
+  return TRUE;
+}
+#endif /* OpenSSL */
+
+#endif /* DLL build */
--- a/docs/libcurl/curl_global_cleanup.md
+++ b/docs/libcurl/curl_global_cleanup.md
@@ -54,6 +54,11 @@ still running then your program may crash or other corruption may occur. We
 recommend you do not run libcurl from any module that may be unloaded
 dynamically. This behavior may be addressed in the future.
 
+libcurl may not be able to fully clean up after multi-threaded OpenSSL
+depending on how OpenSSL was built and loaded as a library. It is possible in
+some rare circumstances a memory leak could occur unless you implement your own
+OpenSSL thread cleanup. Refer to libcurl-thread(3).
+
 # EXAMPLE
 
 ~~~c
--- a/docs/libcurl/libcurl-thread.md
+++ b/docs/libcurl/libcurl-thread.md
@@ -37,10 +37,28 @@ share API (the connection pool and HSTS cache for example).
 
 # TLS
 
-All current TLS libraries libcurl supports are thread-safe. OpenSSL 1.1.0+ can
-be safely used in multi-threaded applications provided that support for the
-underlying OS threading API is built-in. For older versions of OpenSSL, the
-user must set mutex callbacks.
+All current TLS libraries libcurl supports are thread-safe.
+
+## OpenSSL
+
+OpenSSL 1.1.0+ can be safely used in multi-threaded applications provided that
+support for the underlying OS threading API is built-in. For older versions of
+OpenSSL, the user must set mutex callbacks.
+
+libcurl may not be able to fully clean up after multi-threaded OpenSSL
+depending on how OpenSSL was built and loaded as a library. It is possible in
+some rare circumstances a memory leak could occur unless you implement your own
+OpenSSL thread cleanup.
+
+For example, on Windows if both libcurl and OpenSSL are linked statically to a
+DLL or application then OpenSSL may leak memory unless the DLL or application
+calls OPENSSL_thread_stop() before each thread terminates. If OpenSSL is built
+as a DLL then it does this cleanup automatically and there is no leak. If
+libcurl is built as a DLL and OpenSSL is linked statically to it then libcurl
+does this cleanup automatically and there is no leak (added in libcurl 8.7.0).
+
+Please review the OpenSSL documentation for a full list of circumstances:
+https://www.openssl.org/docs/man3.0/man3/OPENSSL_thread_stop.html#NOTES
 
 # Signals
 
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -119,11 +119,12 @@ if(SHARE_LIB_OBJECT)
   set(LIB_OBJECT "libcurl_object")
   add_library(${LIB_OBJECT} OBJECT ${HHEADERS} ${CSOURCES})
   if(WIN32)
-    # Define CURL_STATICLIB always, to disable __declspec(dllexport) for exported
-    # libcurl symbols. We handle exports via libcurl.def instead. Except with
-    # symbol hiding disabled or debug mode enabled, when we export _all_ symbols
-    # from libcurl DLL, without using libcurl.def.
-    set_property(TARGET ${LIB_OBJECT} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_STATICLIB")
+    # Define CURL_STATICLIB always, to disable __declspec(dllexport) for
+    # exported libcurl symbols. We handle exports via libcurl.def instead.
+    # Except with symbol hiding disabled or debug mode enabled, when we export
+    # _all_ symbols from libcurl DLL, without using libcurl.def.
+    set_property(TARGET ${LIB_OBJECT} APPEND
+      PROPERTY COMPILE_DEFINITIONS "CURL_STATICLIB")
   endif()
   target_link_libraries(${LIB_OBJECT} PRIVATE ${CURL_LIBS})
   set_target_properties(${LIB_OBJECT} PROPERTIES
@@ -153,7 +154,8 @@ if(BUILD_STATIC_LIBS)
   add_library(${LIB_STATIC} STATIC ${LIB_SOURCE})
   add_library(${PROJECT_NAME}::${LIB_STATIC} ALIAS ${LIB_STATIC})
   if(WIN32)
-    set_property(TARGET ${LIB_OBJECT} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_STATICLIB")
+    set_property(TARGET ${LIB_OBJECT} APPEND
+      PROPERTY COMPILE_DEFINITIONS "CURL_STATICLIB")
   endif()
   target_link_libraries(${LIB_STATIC} PRIVATE ${CURL_LIBS})
   # Remove the "lib" prefix since the library is already named "libcurl".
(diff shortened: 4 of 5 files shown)


bagder (approved):

jay:
This is on hold because we're still investigating #12327

jay:
@vszakats can you take a look at the unity logic please. In dllmain.c I needed to include windows.h for cygwin but I don't want that being included by other units (like if they were combined in a unity build) so I added some logic to exempt it in cmake, but I'm not sure if anything else is needed for other build systems.

vszakats:
@jay: `#ifdef CURL_STATICLIB` probably doesn't work for this because we have it enabled for both static and shared builds now. With CMake the approach we use with `libcurl.rc` might work here too. With autotools the `_DLL` (IIRC) macro tells if it's compiling for DLL (I've found no way to make the list of objects different for static and shared).

bagder:
`CURL_STATICLIB` is about libcurl being static, not OpenSSL. What if you build libcurl dynamically with a static OpenSSL?

vszakats:
IMO `CURL_STATICLIB` itself is a hack, and I think it shouldn't be there in the first place when compiling the lib itself. If we depend on this for anything that changes generated code, it prevents building libcurl in a single pass, IOW it forces building libcurl _twice_. (It's on my longer term TODO list to delete it, though I'm not quite sure yet if it's possible without breaking build scripts.)

Using `CURL_DISALLOW_EXTERN` as a replacement doesn't help with this.

At this point I wouldn't touch this in `CMakeLists.txt` because it would undo months of work and will force going back to `Makefile.mk` for builds with acceptable performance.

I think a better approach for CMake would be to limit compiling `dllmain.c` to the `${LIB_SHARED}` target:
```
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -182,7 +182,7 @@ if(BUILD_SHARED_LIBS)
   add_library(${LIB_SHARED} SHARED ${LIB_SOURCE})
   add_library(${PROJECT_NAME}::${LIB_SHARED} ALIAS ${LIB_SHARED})
   if(WIN32)
-    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES libcurl.rc)
+    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES libcurl.rc dllmain.c)
     if(HIDES_CURL_PRIVATE_SYMBOLS)
       set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES "${CURL_SOURCE_DIR}/libcurl.def")
     endif()
```

The next problem is how to enable it inside `dllmain.c`. For autotools we need
to enable it when `DLL_EXPORT` is defined. I don't know what is the equivalent
for MSVC builds (or if it compiles in a single pass, like CMake does now). For CMake
builds, we can append `DLL_EXPORT` to `COMPILE_DEFINITIONS` for the
`${LIB_SHARED}` target and have this solved. At the same time we should stop 
relying on `CURL_STATICLIB` for this. (I haven't tested these.)

vszakats:
Here's a rough patch building seemingly fine with CMake (while also keeping autotools working, but, not tested).

Except this:
```
lib/dllmain.c:41:13: warning: no previous prototype for 'DllMain' [-Wmissing-prototypes]
   41 | BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
      |             ^~~~~~~
```

```diff
diff --git a/lib/CMakeLists.txt b/lib/CMakeLists.txt
index 51d52578e..2e99c7861 100644
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -181,10 +181,17 @@ if(BUILD_SHARED_LIBS)
   list(APPEND libcurl_export ${LIB_SHARED})
   add_library(${LIB_SHARED} SHARED ${LIB_SOURCE})
   add_library(${PROJECT_NAME}::${LIB_SHARED} ALIAS ${LIB_SHARED})
+  unset(_shared_compile_definitions)
+  unset(_shared_compile_sources)
+  if(WIN32 OR CYGWIN)
+    set_source_files_properties(${LIB_DLLMAIN} PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+    list(APPEND _shared_compile_sources ${LIB_DLLMAIN})
+    list(APPEND _shared_compile_definitions "DLL_EXPORT")
+  endif()
   if(WIN32)
-    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES libcurl.rc)
+    list(APPEND _shared_compile_sources ${LIB_RCFILES})
     if(HIDES_CURL_PRIVATE_SYMBOLS)
-      set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES "${CURL_SOURCE_DIR}/libcurl.def")
+      list(APPEND _shared_compile_sources "${CURL_SOURCE_DIR}/libcurl.def")
     endif()
   endif()
   target_link_libraries(${LIB_SHARED} PRIVATE ${CURL_LIBS})
@@ -195,7 +202,13 @@ if(BUILD_SHARED_LIBS)
     POSITION_INDEPENDENT_CODE ON)
   if(HIDES_CURL_PRIVATE_SYMBOLS)
     set_property(TARGET ${LIB_SHARED} APPEND PROPERTY COMPILE_FLAGS "${CURL_CFLAG_SYMBOLS_HIDE}")
-    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_HIDDEN_SYMBOLS")
+    list(APPEND _shared_compile_definitions "CURL_HIDDEN_SYMBOLS")
+  endif()
+  if(_shared_compile_definitions)
+    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY COMPILE_DEFINITIONS ${_shared_compile_definitions})
+  endif()
+  if(_shared_compile_sources)
+    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES ${_shared_compile_sources})
   endif()
   if(CURL_HAS_LTO)
     set_target_properties(${LIB_SHARED} PROPERTIES
diff --git a/lib/Makefile.am b/lib/Makefile.am
index 1237c8e99..5ce7c5650 100644
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -70,6 +70,8 @@ include Makefile.inc
 libcurl_la_SOURCES = $(CSOURCES) $(HHEADERS)
 libcurlu_la_SOURCES = $(CSOURCES) $(HHEADERS)
 
+libcurl_la_SOURCES += $(LIB_DLLMAIN)
+
 libcurl_la_CPPFLAGS_EXTRA =
 libcurl_la_LDFLAGS_EXTRA =
 libcurl_la_CFLAGS_EXTRA =
diff --git a/lib/Makefile.inc b/lib/Makefile.inc
index e568ef953..46af66273 100644
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -362,6 +362,8 @@ LIB_HFILES =         \
   warnless.h         \
   ws.h
 
+LIB_DLLMAIN = dllmain.c
+
 LIB_RCFILES = libcurl.rc
 
 CSOURCES = $(LIB_CFILES) $(LIB_VAUTH_CFILES) $(LIB_VTLS_CFILES) \
--- dllmain.c	2023-11-28 20:45:00.383279467 +0000
+++ dllmain.c--	2023-11-28 20:42:18.552910188 +0000
@@ -24,17 +24,15 @@
 
 #include "curl_setup.h"
 
-#if (defined(_WIN32) || defined(__CYGWIN__)) && !defined(CURL_STATICLIB)
+#if (defined(_WIN32) || defined(__CYGWIN__)) && defined(DLL_EXPORT)
 
 /* WARNING: Including Cygwin windows.h may define _WIN32 in old versions or
    may have unexpected behavior in unity builds (where all source files are
    bundled into a single unit). For that reason, this source file must be
    compiled separately for Cygwin unity builds. */
 
-#ifdef __CYGWIN__
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
-#endif
 
 #ifdef USE_OPENSSL
 #include <openssl/crypto.h>
@@ -64,4 +62,4 @@
   return TRUE;
 }
 
-#endif /* (_WIN32 || __CYGWIN__) && !CURL_STATICLIB */
+#endif /* (_WIN32 || __CYGWIN__) && !DLL_EXPORT */
```

jay:
> IMO CURL_STATICLIB itself is a hack, and I think it shouldn't be there in the first place when compiling the lib itself. If we depend on this for anything that changes generated code, it prevents building libcurl in a single pass, IOW it forces building libcurl twice. (It's on my longer term TODO list to delete it, though I'm not quite sure yet if it's possible without breaking build scripts.)

It's great that it's possible to do a single pass with CMake for both shared and static but we have to account for other build systems. Also we tell the user to define CURL_STATICLIB if libcurl is linked statically.

> At this point I wouldn't touch this in CMakeLists.txt because it would undo months of work and will force going back to Makefile.mk for builds with acceptable performance.

Alright let's say we leave it win32 CURL_STATICLIB always for cmake. I'm sure there is a way to still allow cmake to do a single pass while allowing the other build systems to do their thing.

> Here's a rough patch building seemingly fine with CMake (while also keeping autotools working, but not tested).

Couldn't this be addressed by leaving dllmain check as I did it (ie check for !CURL_STATICLIB) but then also telling cmake not to compile dllmain.c for static builds? That preserves the behavior you added without needing changes to other build systems.

vszakats:
Here's one take reusing `CURL_STATICLIB`. Works with the happy-path, buts need (much) more testing before merging:
(works with the unmodified `dllmain.c`)
```
diff --git a/lib/CMakeLists.txt b/lib/CMakeLists.txt
index 51d52578e..e0ce3eacd 100644
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -114,23 +114,26 @@ if(NOT DEFINED SHARE_LIB_OBJECT)
   endif()
 endif()
 
-if(WIN32)
-  # Define CURL_STATICLIB always, to disable __declspec(dllexport) for exported
-  # libcurl symbols. We handle exports via libcurl.def instead. Except with
-  # symbol hiding disabled or debug mode enabled, when we export _all_ symbols
-  # from libcurl DLL, without using libcurl.def.
-  add_definitions("-DCURL_STATICLIB")
-endif()
-
 if(SHARE_LIB_OBJECT)
   set(LIB_OBJECT "libcurl_object")
   add_library(${LIB_OBJECT} OBJECT ${HHEADERS} ${CSOURCES})
+  unset(_object_compile_definitions)
+  if(WIN32)
+    # Define CURL_STATICLIB always, to disable __declspec(dllexport) for exported
+    # libcurl symbols. We handle exports via libcurl.def instead. Except with
+    # symbol hiding disabled or debug mode enabled, when we export _all_ symbols
+    # from libcurl DLL, without using libcurl.def.
+    list(APPEND _object_compile_definitions "CURL_STATICLIB")
+  endif()
   target_link_libraries(${LIB_OBJECT} PRIVATE ${CURL_LIBS})
   set_target_properties(${LIB_OBJECT} PROPERTIES
     POSITION_INDEPENDENT_CODE ON)
   if(HIDES_CURL_PRIVATE_SYMBOLS)
     set_property(TARGET ${LIB_OBJECT} APPEND PROPERTY COMPILE_FLAGS "${CURL_CFLAG_SYMBOLS_HIDE}")
-    set_property(TARGET ${LIB_OBJECT} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_HIDDEN_SYMBOLS")
+    list(APPEND _object_compile_definitions "CURL_HIDDEN_SYMBOLS")
+  endif()
+  if(_object_compile_definitions)
+    set_property(TARGET ${LIB_OBJECT} APPEND PROPERTY COMPILE_DEFINITIONS ${_object_compile_definitions})
   endif()
   if(CURL_HAS_LTO)
     set_target_properties(${LIB_OBJECT} PROPERTIES
@@ -152,6 +155,10 @@ if(BUILD_STATIC_LIBS)
   list(APPEND libcurl_export ${LIB_STATIC})
   add_library(${LIB_STATIC} STATIC ${LIB_SOURCE})
   add_library(${PROJECT_NAME}::${LIB_STATIC} ALIAS ${LIB_STATIC})
+  unset(_static_compile_definitions)
+  if(WIN32)
+    list(APPEND _static_compile_definitions "CURL_STATICLIB")
+  endif()
   target_link_libraries(${LIB_STATIC} PRIVATE ${CURL_LIBS})
   # Remove the "lib" prefix since the library is already named "libcurl".
   set_target_properties(${LIB_STATIC} PROPERTIES
@@ -160,7 +167,10 @@ if(BUILD_STATIC_LIBS)
     INTERFACE_COMPILE_DEFINITIONS "CURL_STATICLIB")
   if(HIDES_CURL_PRIVATE_SYMBOLS)
     set_property(TARGET ${LIB_STATIC} APPEND PROPERTY COMPILE_FLAGS "${CURL_CFLAG_SYMBOLS_HIDE}")
-    set_property(TARGET ${LIB_STATIC} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_HIDDEN_SYMBOLS")
+    list(APPEND _static_compile_definitions "CURL_HIDDEN_SYMBOLS")
+  endif()
+  if(_static_compile_definitions)
+    set_property(TARGET ${LIB_STATIC} APPEND PROPERTY COMPILE_DEFINITIONS ${_static_compile_definitions})
   endif()
   if(CURL_HAS_LTO)
     set_target_properties(${LIB_STATIC} PROPERTIES
@@ -181,10 +191,15 @@ if(BUILD_SHARED_LIBS)
   list(APPEND libcurl_export ${LIB_SHARED})
   add_library(${LIB_SHARED} SHARED ${LIB_SOURCE})
   add_library(${PROJECT_NAME}::${LIB_SHARED} ALIAS ${LIB_SHARED})
+  unset(_shared_compile_sources)
+  if(WIN32 OR CYGWIN)
+    set_source_files_properties(${LIB_DLLMAIN} PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+    list(APPEND _shared_compile_sources ${LIB_DLLMAIN})
+  endif()
   if(WIN32)
-    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES libcurl.rc)
+    list(APPEND _shared_compile_sources ${LIB_RCFILES})
     if(HIDES_CURL_PRIVATE_SYMBOLS)
-      set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES "${CURL_SOURCE_DIR}/libcurl.def")
+      list(APPEND _shared_compile_sources "${CURL_SOURCE_DIR}/libcurl.def")
     endif()
   endif()
   target_link_libraries(${LIB_SHARED} PRIVATE ${CURL_LIBS})
@@ -197,6 +212,9 @@ if(BUILD_SHARED_LIBS)
     set_property(TARGET ${LIB_SHARED} APPEND PROPERTY COMPILE_FLAGS "${CURL_CFLAG_SYMBOLS_HIDE}")
     set_property(TARGET ${LIB_SHARED} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_HIDDEN_SYMBOLS")
   endif()
+  if(_shared_compile_sources)
+    set_property(TARGET ${LIB_SHARED} APPEND PROPERTY SOURCES ${_shared_compile_sources})
+  endif()
   if(CURL_HAS_LTO)
     set_target_properties(${LIB_SHARED} PROPERTIES
       INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE
```

vszakats:
`OPENSSL_thread_stop` is missing from these forks (as of their current latest versions): LibreSSL, BoringSSL, AWS-LC. And also wolfSSL.

jay:
Some of these word checks are snobby. It doesn't like the word `will`. It doesn't like the word `very`. It doesn't even like `It's`. I mean who doesn't like `It's`.

lib/dllmain.c
```diff
@@ -0,0 +1,82 @@
...
+  case DLL_PROCESS_DETACH:
+    break;
+  case DLL_THREAD_ATTACH:
+    break;
+  case DLL_THREAD_DETACH:
+#if defined(USE_OPENSSL) && \
+    !defined(OPENSSL_IS_AWSLC) && \
+    !defined(OPENSSL_IS_BORINGSSL) && \
+    !defined(LIBRESSL_VERSION_NUMBER) && \
+    (OPENSSL_VERSION_NUMBER >= 0x10100000L)
```
vszakats, line 71: What do you think of excluding the whole `DllMain` for backends that don't need this call?

jay: good idea

BrianInglis:
Local patch applied by Cygwin OpenSSL maintainer to test release; see [Test: openssl 3.0.16-2, mingw64-{i686,x86_64}-openssl-3.0.16-2](https://cygwin.com/pipermail/cygwin-announce/2025-May/012341.html) and 
[openssl and libcurl issue with cargo](https://cygwin.com/pipermail/cygwin/2025-May/258203.html)

Issue #12410, opened 2023-11-27
Crashed at Curl_http_bodysend


yushicheng7788:
### I did this

I use CURLOPT_HTTPPOST to send files, but curl crashed. I think it is caused by the following part of the code.
```
/* Convert the form structure into a mime structure, then keep
       the conversion */
    if(!data->state.formp) {
      data->state.formp = calloc(sizeof(curl_mimepart), 1);
      if(!data->state.formp)
        return CURLE_OUT_OF_MEMORY;
      Curl_mime_cleanpart(data->state.formp);
      result = Curl_getformdata(data, data->state.formp, data->set.httppost,
                                data->state.fread_func, data->state.seek_func);
      if(result)
        return result;
      data->state.mimepost = data->state.formp;
    }
```
If Curl_getformdata() returns any error, which will cause data->state.mimepost = data->state.formp; not executed. But if I sent this request again, data->state.formp is not NULL, and data->state.mimepost is NULL, which cause a crash. Is there any optimization method in this case?

### I expected the following

Curl will not crash in this case.

### curl/libcurl version

8.4.0

### operating system

MacOS

yushicheng7788:
Can we call Curl_safefree(data->state.formp); when Curl_getformdata return non-zero value to fix this crash?

bagder:
Can you provide us an example that reproduces this problem?

yushicheng7788:
```
int main(int argc, char** argv)
{
    CURLM *cm;
    CURLMsg *msg;
    unsigned int transfers = 0;
    int msgs_left = -1;
    int left = 1;

    curl_global_init(CURL_GLOBAL_ALL);
    cm = curl_multi_init();
    
    CURL *eh = curl_easy_init();
    curl_easy_setopt(eh, CURLOPT_URL, _T("http://localhost:9093/micro/all"));
    struct curl_httppost *lastptr=NULL;
    struct curl_httppost *m_formpost=NULL;
    curl_formadd(&m_formpost,
        &lastptr,
        CURLFORM_COPYNAME, "file",
        CURLFORM_FILE, CMM_T2A(_T("asdwq/usr/bin/zip")),
        CURLFORM_END);
    curl_easy_setopt(eh, CURLOPT_HTTPPOST, m_formpost);
    curl_multi_add_handle(cm, eh);
    
    do {
        int still_alive = 1;
        curl_multi_perform(cm, &still_alive);
        while((msg = curl_multi_info_read(cm, &msgs_left))) {
[...]

PR #12419, opened 2023-11-28
Free data->state.formp when Curl_getformdata return non-zero value

yushicheng7788:
This pull request solves [this issue](https://github.com/curl/curl/issues/12410). When Curl_getformdata returns a non-zero value, data->state.formp needs to be released to ensure that easy handle will not crash when reused.

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -3260,7 +3260,11 @@ CURLcode Curl_http(struct Curl_easy *data, bool *done)
 
   result = Curl_http_body(data, conn, httpreq, &te);
   if(result)
+  {
+    if (data->state.formp)
+      Curl_safefree(data->state.formp);
     return result;
+  }
 
   p_accept = Curl_checkheaders(data,
                                STRCONST("Accept"))?NULL:"Accept: */*\r\n";
--- a/lib/http.c
+++ b/lib/http.c
@@ -2439,8 +2439,10 @@ CURLcode Curl_http_body(struct Curl_easy *data, struct connectdata *conn,
       Curl_mime_cleanpart(data->state.formp);
       result = Curl_getformdata(data, data->state.formp, data->set.httppost,
                                 data->state.fread_func);
-      if(result)
+      if(result) {
+        Curl_safefree(data->state.formp);
         return result;
+      }
       data->state.mimepost = data->state.formp;
     }
     break;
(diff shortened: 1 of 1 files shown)


lib/http.c
```diff
@@ -3260,7 +3260,11 @@ CURLcode Curl_http(struct Curl_easy *data, bool *done)
 
   result = Curl_http_body(data, conn, httpreq, &te);
   if(result)
+  {
+    if (data->state.formp)
```
bagder, line 3264: No need for an `if()` here.

bagder:
I propose this slightly different version, which frees the pointer close to where the error happens:
~~~diff
From 5b2a4caf7b3cff535df3f146eeec6bc7cfdc6259 Mon Sep 17 00:00:00 2001
From: yushicheng7788 <lance.yu@zoom.us>
Date: Tue, 28 Nov 2023 15:49:49 +0800
Subject: [PATCH] Curl_http_body: Free data->state.formp when Curl_getformdata
 errors

Fixes #12410
---
 lib/http.c | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git a/lib/http.c b/lib/http.c
index 45748dd29..be6d442e8 100644
--- a/lib/http.c
+++ b/lib/http.c
@@ -2437,12 +2437,14 @@ CURLcode Curl_http_body(struct Curl_easy *data, struct connectdata *conn,
       if(!data->state.formp)
         return CURLE_OUT_OF_MEMORY;
       Curl_mime_cleanpart(data->state.formp);
       result = Curl_getformdata(data, data->state.formp, data->set.httppost,
                                 data->state.fread_func);
-      if(result)
+      if(result) {
+        Curl_safefree(data->state.formp);
         return result;
+      }
       data->state.mimepost = data->state.formp;
     }
     break;
 #endif
   default:
-- 
2.43.0

yushicheng7788:
Thanks for your review, already changed. And I'm new to curl. I will read more docs later. Thanks for your work again!

bagder:
Based on your fine recipe on reproducing this problem, I wrote a test case that verified this fix which then led me to do some further fixes as this would cause a memory-leak. My slightly larger take is in  #12421

bagder:
I merged #12421 instead, thanks!

PR #12427, opened 2023-11-29
tcp socket, trace local address used in connect

icing:
adds a cfilter trace of the bound local address during connect attempt. QUIC connects already did that.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1158,6 +1158,9 @@ static CURLcode cf_tcp_connect(struct Curl_cfilter *cf,
 
     /* Connect TCP socket */
     rc = do_connect(cf, data, cf->conn->bits.tcp_fastopen);
+    set_local_ip(cf, data);
+    CURL_TRC_CF(data, cf, "local address %s port %d...",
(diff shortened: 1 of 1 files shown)


lib/cf-socket.c
```diff
@@ -1158,13 +1158,13 @@ static CURLcode cf_tcp_connect(struct Curl_cfilter *cf,
...
-    set_local_ip(cf, data);
-    CURL_TRC_CF(data, cf, "local address %s port %d...",
-                ctx->l_ip, ctx->l_port);
     if(-1 == rc) {
       result = socket_connect_result(data, ctx->r_ip, SOCKERRNO);
       goto out;
     }
+    set_local_ip(cf, data);
+    CURL_TRC_CF(data, cf, "local address %s port %d...",
+                ctx->l_ip, ctx->l_port);
```
bagder, line 1167: When is this displayed?

icing: When `tcp` tracing is active. Do you want it on `infof()` instead?

bagder (approved):

PR #12428, opened 2023-11-29
Aborting pausing send when connection is marked for closing.

dkarpov1970:
This handles cases of some bi-directional "upgrade" scenarios (i.e. WebSockets) where sending is paused until some "upgrade" handshake is completed, but server rejects the handshake and closes the connection.

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -733,6 +733,7 @@ static CURLcode readwrite_data(struct Curl_easy *data,
        on from our side, we need to stop that immediately. */
     infof(data, "we are done reading and this is set to close, stop send");
     k->keepon &= ~KEEP_SEND; /* no writing anymore either */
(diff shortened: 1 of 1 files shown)


dkarpov1970:
I don't think that the "AppVeyor / CMake, mingw-w64, gcc 13, Debug, x64, Schannel, Static, Unicode, Unity " test failure is real.
Seems like some false positive as I don't see how my very minor change could do such failure. 
Unfortunately, I don't know how to get more details about what was causing the failure.

bagder:
Don't be alarmed by the red CI build(s), that's just due to flaky/bad environments and not because of any flaw in your PR. Ignore them.

bagder (approved):

bagder:
Thanks!

PR #12430, opened 2023-11-30
multi: during ratelimit multi_getsock should return no sockets

bagder:
... as there is nothing to wait for then, it just waits. Otherwise, this causes much more CPU work and updates than necessary during ratelimit periods.

Ref: https://curl.se/mail/lib-2023-11/0056.html


icing (approved):
Running `curl --silent --limit-rate 1000 https://curl.se/?[0-4] -o /dev/null -w '%{speed_download}\n'` gives:

```
908
921
921
908
1103
```

adding `-Z` gives:

```
1088
1052
1024
1024
915
```

I'd say it looks good!

PR #12434, opened 2023-12-01
tool_writeout_json: fix JSON encoding of non-ascii bytes

emanuele6:
When the JSON encoding code was refactored to be reused for `{{var:json}}` variables were redeclared as `char*` instead of `unsigned char*`.

char variables if unspecified can be either signed or unsigned depending
on the platform according to the C standard; in most platforms, they are
signed.

This meant that the `*i<32` were always true for bytes with the top bit set. So they were always getting encoded as `\uXXXX`, and then since they were also signed negative, they were getting extended with 1s causing `'\xe2'` to be expanded to `\uffffffe2`, for example:

    $ curl --variable 'v=“' --expand-write-out '{{v:json}}\n' file:///dev/null
    \uffffffe2\uffffff80\uffffff9c

I fixed this bug by making the code use explicitly `unsigned char*` variables instead of `char*` variables.

I also added an explicit `(unsigned)` cast for correctness since the `curlx_dyn_addf()` function takes a `va_list`, and expects a `unsigned` while `*i` is an `unsigned char`.

Reported by iconoclast\_hero on the #bash IRC channel of libera.chat.


emanuele6:
Actually this has always been bugged, ever since that function was added to support `-w %{json}`
04c03416e68fd635a15cae8201872f5c29fdcca8

I assumed it used to not be bugged because the `jsonString()` used by `trurl` correctly uses `unsigned` variables, and I thought it was copied from curl.

src/tool_writeout_json.c
```diff
@@ -70,7 +70,7 @@ int jsonquoted(const char *in, size_t len,
       break;
     default:
       if(*i < 32)
-        result = curlx_dyn_addf(out, "\\u%04x", *i);
+        result = curlx_dyn_addf(out, "\\u%04x", (unsigned)*i);
```
jay, line 73: the cast is unnecessary now. however if `*i` was type char (assume `char` is signed) then you'd have to cast to `unsigned char` not `unsigned`
~~~
char x = 'é'; // decimal: -23, hex: e9
printf("%x", (unsigned)x); // ffffffe9
printf("%x", (unsigned char)x); // e9
~~~

emanuele6: @jay It is technically necessary because `sizeof (usnigned char) != sizeof (unsigned)` and `curlx_dyn_addf` takes arguments with a variadic list `...`.

emanuele6: According to <https://cppreference.com/> casting to `int`/`unsigned` is unnecessary for types smaller than `int` because they get converted automatically.
https://en.cppreference.com/w/cpp/language/variadic_arguments

But I don't think that is specified in `C89`, I can't find it.

emanuele6:
Is this a bug in the windows version?

```
  test 0268...[JSON encoding of unicode string]
  
   268: protocol FAILED:
  --- log/check-expected	2023-12-01 01:44:12.016402600 +0000
  +++ log/check-generated	2023-12-01 01:44:12.016402600 +0000
  @@ -2,7 +2,7 @@
   Host: 127.0.0.1:1060[CR][LF]
   User-Agent: curl/8.5.0-DEV[CR][LF]
   Accept: */*[CR][LF]
  -Content-Length: 3[CR][LF]
  +Content-Length: 1[CR][LF]
   Content-Type: application/x-www-form-urlencoded[CR][LF]
   [CR][LF]
  -%e2%80%9c
  +%93
```

I need to rewrite the test to read the data from a file, because it seems `cmd.exe` is passing `\x93` to curl instead of `\xe2\x80\x9c` on Windows.

src/tool_writeout_json.c
jay: it's called default argument promotions. for arguments not prototyped by the function there's integer promotions  and float becomes double. note old c standard c89 refers to integer promotions as "integral promotions".

[3.3.2.2 Function calls](https://port70.net/~nsz/c/c89/c89-draft.html#3.3.2.2)

> _If the expression that denotes the called function has a type that does not include a prototype, the integral promotions are performed on each argument and arguments that have type float are promoted to double. These are called the default argument promotions._

[3.2.1.1 Characters and integers](https://port70.net/~nsz/c/c89/c89-draft.html#3.2.1.1)

> _A char, a short int, or an int bit-field, or their signed or unsigned varieties, or an object that has enumeration type, may be used in an expression wherever an int or unsigned int may be used. If an int can represent all values of the original type, the value is converted to an int; otherwise it is converted to an unsigned int. These are called the integral promotions._

jay:
For windows you may have to read the data from a file instead of the command line. The reason is because in Windows you can build curl with or without unicode support. Even with unicode builds I get weird output with your pr. for example:
~~~
> curld --variable "v=®" --expand-write-out "{{v:json}}\n" file:///NUL
┬«
~~~
that should output an ® but it doesn't

bagder (approved):

bagder:
Thanks!

PR #12439, opened 2023-12-01
vtls: consistently use typedef names for OpenSSL structs

davidben:
The foo_st names don't appear in OpenSSL public API documentation. The FOO typedefs are more common. This header was already referencing SSL_CTX via <openssl/ssl.h>. There is a comment about avoiding <openssl/x509v3.h>, but OpenSSL actually declares all the typedefs in <openssl/ossl_typ.h>, which is already included by <openssl/ssl.h> (and every other OpenSSL header), so just use that. Though I've included it just to be explicit.

(I'm also fairly sure including <openssl/ssl.h> already triggers the Schannel conflicts anyway. The comment was probably just out of date.)

### DIFF
--- a/lib/vtls/openssl.h
+++ b/lib/vtls/openssl.h
@@ -31,6 +31,7 @@
  * This header should only be needed to get included by vtls.c, openssl.c
  * and ngtcp2.c
  */
+#include <openssl/ossl_typ.h>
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #12448, opened 2023-12-03
make all libcurl man page examples compile - and verify them in CI

bagder:
- the gcc invocation uses the flags ` -Wunused -Werror -Wno-unused-but-set-variable`
- also run `checksrc,pl` on the code to make them follow our code style


bagder:
I think this is pretty much done for merge.

PR #12452, opened 2023-12-04
examples/rtsp-options.c: add

bagder:
Just a bare bones RTSP example using CURLOPT_RTSP_SESSION_ID and CURLOPT_RTSP_REQUEST set to CURL_RTSPREQ_OPTIONS.

### DIFF
--- a/docs/examples/Makefile.inc
+++ b/docs/examples/Makefile.inc
@@ -108,6 +108,7 @@ check_PROGRAMS = \
   progressfunc \
   protofeats \
   resolve \
+  rstp-options \
   sendrecv \
   sepheaders \
   sftpget \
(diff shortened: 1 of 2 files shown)


gvanem:
How come `-DUSE_HYPER` and RTSP is not possible?
And I found no mention of `"Hyper"` in any `CURLINFO_RTSP_*.3` man-pages.

bagder:
> How come -DUSE_HYPER and RTSP is not possible?

It's a documented restriction. Because we implement RTSP mostly with HTTP code, making hyper support that "trick" is perhaps possible but a complication we have not dealt with.

>  And I found no mention of "Hyper" in any CURL*_RTSP_*.3 man-pages.

It is also still experimental. Maybe we can make it support RTSP before we un-experimental it...

PR #12453, opened 2023-12-04
use strndup/memdup instead of malloc, memcpy and null-terminate etc

bagder:


bagder:
@jay: I want to remove the `memchr` stuff again from strndup() because I think it potentially breaks several of these strndup cases.

lib/strdup.c
```diff
@@ -104,18 +104,14 @@ void *Curl_memdup(const void *src, size_t length)
...
  ***************************************************************************/
 void *Curl_strndup(const char *src, size_t length)
 {
-  char *buf = memchr(src, '\0', length);
-  if(buf)
-    length = buf - src;
-  buf = malloc(length + 1);
+  char *buf = malloc(length + 1);
   if(!buf)
     return NULL;
```
jay, line 116: > I want to remove the memchr stuff again from strndup() because I think it potentially breaks several of these strndup cases.

How is it breaking? If we don't check if the length of src is shorter than the passed in length then the memcpy can read past the end of src. If the full passed in length needs to be allocated then it could be adjusted after the allocation but memchr would need to be used:

```suggestion
  char *end;
  if(!buf)
    return NULL;
  end = memchr(src, '\0', length);
  if(end)
    length = end - src;
```

bagder: Because nowhere in the code does it assume that it should only add the substring up to a null terminator. They all want the specified subtring allocated, with a null terminator so that it can be treated as a string. 

In some cases the substring might contain a null byte and then the added null terminator is of course unnecessary but that's how the code already works. If the memchr() logic remains , this use case breaks.

jay: Can you give me an example? I still don't get it.

lib/bufref.c
```diff
@@ -116,12 +117,9 @@ CURLcode Curl_bufref_memdup(struct bufref *br, const void *ptr, size_t len)
   DEBUGASSERT(len <= CURL_MAX_INPUT_LENGTH);
 
   if(ptr) {
-    cpy = malloc(len + 1);
+    cpy = Curl_strndup(ptr, len);
```
monnerat, line 120: I think this is wrong: the function is supposed to duplicate the given byte count, but this will stop at the first null byte.
The original code behaves more like memdup, but appends a null byte as a sentinel in case the data is effectively a string.
Maybe we should have a such a generic dual-use duplication function.

bagder: No, Curl_strndup() **does not** stop at the first null byte (any more). That is exactly what I mentioned above in this PR.

monnerat: My bad! I've been abused by the Curl_strndup() original code and the usage semantics associated with this name.
Sorry for the noise.

bagder: I've been pondering on maybe renaming this function just to make it less likely that one would think that it has this functionality...

icing:
Curl_memdup0 ?Am 08.12.2023 um 14:05 schrieb Daniel Stenberg ***@***.***>:﻿
@bagder commented on this pull request.



In lib/bufref.c:
> @@ -116,12 +117,9 @@ CURLcode Curl_bufref_memdup(struct bufref *br, const void *ptr, size_t len)
   DEBUGASSERT(len <= CURL_MAX_INPUT_LENGTH);
 
   if(ptr) {
-    cpy = malloc(len + 1);
+    cpy = Curl_strndup(ptr, len);

I've been pondering on maybe renaming this function just to make it less likely that one would think that it has this functionality...

lib/bufref.c
monnerat: > ... one would think that it has this functionality...

This happened to me: will probably affect others.

bagder:
I made #12490

PR #12458, opened 2023-12-05
github/labeler: switch from the beta to labeler v5

dfandrich:
Some keys were renamed and the dot option was made default.

Closes #12458


dfandrich:
Looks like I can't test this in a PR so I'll submit and see what happens.

PR #12459, opened 2023-12-05
Remove the Curl_cft_ssl_proxy object if the CURL_DISABLE_PROXY flag was defined during compilation.

MAntoniak:


bagder:
Thanks!

PR #12475, opened 2023-12-06
tidy-up: fix yamllint whitespace issues in labeler.yml

vszakats:
Follow-up to bda212911457c6fadfbba50be61afc4ca513fa56 #12466

Closes #12475


dfandrich (approved):
You get whiplash reading the file now with all the extra whitespace, but consistency is probably more important.

vszakats:
Yeah, can't say I like it either. Looks an odd choice for the data structure this needs to represent.

Issue #12477, opened 2023-12-06
[autotools] NTLM_WB support cannot be enabled


benoit-pierre:
### I did this

Using `./configure --with-openssl --enable-ntlm --enable-ntlm-wb` (with `ntlm_auth` in the standard location) does  result in a config with NTLM_WB support disabled.

### I expected the following

Support should be enabled.

It looks like the issue is checking for `fork` was removed in 2ba804942fda50bbd385c743309a7ec8aca2cf47, resulting in `ac_cv_func_fork` never being set to `yes`, so [this test in `m4/curl-confopts.m4`](https://github.com/curl/curl/blob/master/m4/curl-confopts.m4#L620) always fail.

### curl/libcurl version

curl 8.5.0

### operating system

Arch Linux.

jay:
#12479

Also please see https://curl.se/dev/deprecate.html#ntlm_wb-auth

benoit-pierre:
Thanks, I know about the deprecation. In fact, I'm not really interested in the support, I just noticed the issue while updating the Meson wrap for cURL to 8.5.0.

Additionally, there's actually one test for the feature (1310), which does not seem to work (tested on Arch Linux and Ubuntu 22.04).

jay:
Since NTLM_WB is deprecated there are no plans to fix this. Thanks

PR #12479, opened 2023-12-06
configure: resurrect check for fork()

jay:
The check was removed in 2ba804942fda50bbd385c743309a7ec8aca2cf47 but it appears that broke support for NTLM winbind (NTLM_WB).

Reported-by: Benoit Pierre

Fixes https://github.com/curl/curl/issues/12477
Closes #xxxx

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3582,6 +3582,7 @@ AC_CHECK_FUNCS([\
   _fseeki64 \
   arc4random \
   fnmatch \
+  fork \
   fseeko \
   geteuid \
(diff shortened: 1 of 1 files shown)


bagder:
Reverting f76fcd6f4de1d838f008aa1d052f21d1e944e441 and d4314cdf65aee295db627016934bd9eb621ab077 makes test 1310 work again (combined with this fix).

jay:
> Maybe we let this remain broken for now?

I was debating that as well. Ok.

PR #12482, opened 2023-12-07
asyn-thread: use GetAddrInfoExW on >= Windows 8

pps83:
Use GetAddrInfoExW if available starting from win8 instead of starting a thread for each DNS request

### DIFF
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -646,15 +646,17 @@ static bool init_resolve_thread(struct Curl_easy *data,
       td->tsd.w8.hints.ai_family = hints->ai_family;
       td->tsd.w8.hints.ai_socktype = hints->ai_socktype;
       td->complete_ev = CreateEvent(NULL, TRUE, FALSE, NULL);
-      /* fall back to old thread code if something failed */
-      if(td->complete_ev) {
-        err = Curl_GetAddrInfoExW(td->tsd.w8.name, td->tsd.w8.port, NS_DNS,
-          NULL, &td->tsd.w8.hints, &td->tsd.w8.res, NULL,
-          &td->tsd.w8.overlapped, &query_complete, &td->tsd.w8.cancel_ev);
-        if(err != WSA_IO_PENDING)
-          query_complete(err, 0, &td->tsd.w8.overlapped);
-        return TRUE;
+      if(!td->complete_ev) {
+          /* failed to start, mark it as done here for proper cleanup. */
+          td->tsd.done = 1;
+          goto err_exit;
       }
+      err = Curl_GetAddrInfoExW(td->tsd.w8.name, td->tsd.w8.port, NS_DNS,
+        NULL, &td->tsd.w8.hints, &td->tsd.w8.res, NULL,
+        &td->tsd.w8.overlapped, &query_complete, &td->tsd.w8.cancel_ev);
+      if(err != WSA_IO_PENDING)
+        query_complete(err, 0, &td->tsd.w8.overlapped);
+      return TRUE;
     }
   }
 #endif
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -54,6 +54,7 @@
 #  define RESOLVER_ENOMEM  ENOMEM
 #endif
 
+#include "system_win32.h"
 #include "urldata.h"
 #include "sendf.h"
 #include "hostip.h"
@@ -68,12 +69,6 @@
 #include "curl_printf.h"
 #include "curl_memory.h"
 #include "memdebug.h"
-
-#ifdef _WIN32
-#include "system_win32.h"
-/* set in win32_init() */
-extern bool Curl_isWindows8OrGreater;
-#endif
 
 struct resdata {
   struct curltime start;
@@ -361,8 +356,9 @@ query_complete(DWORD err, DWORD bytes, LPWSAOVERLAPPED overlapped)
       memcpy(ca->ai_addr, ai->ai_addr, ss_size);
 
       if(namelen) {
+        size_t i;
         ca->ai_canonname = (void *)((char *)ca->ai_addr + ss_size);
-        for(size_t i = 0; i < namelen; ++i) /* convert wide string to ascii */
+        for(i = 0; i < namelen; ++i) /* convert wide string to ascii */
           ca->ai_canonname[i] = (char)ai->ai_canonname[i];
         ca->ai_canonname[namelen] = '\0';
       }
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -44,6 +44,15 @@
 #include <inet.h>
 #endif
 
+#ifdef _WIN32
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#include <windows.h>
+#define MAX_URL_LENGTH 256
+/* set in win32_init() */
+extern bool Curl_isWindows8OrGreater;
+#endif
+
 #if defined(USE_THREADS_POSIX) && defined(HAVE_PTHREAD_H)
 #  include <pthread.h>
 #endif
@@ -144,9 +153,24 @@ static bool init_resolve_thread(struct Curl_easy *data,
                                 const char *hostname, int port,
                                 const struct addrinfo *hints);
 
+#if (_WIN32_WINNT >= 0x0600)
+/* Thread sync data used for GetAddrInfoExW  for win8+ */
+struct thread_sync_data_w8
+{
+    OVERLAPPED overlapped;
+    ADDRINFOEXW *res;
+    HANDLE cancel_ev;
+    WCHAR name[256]; /* max domain name is 253 chars */
+    WCHAR service_name[20]; /* port */
+    ADDRINFOEXW hints;
+};
+#endif
 
 /* Data for synchronization between resolver thread and its parent */
 struct thread_sync_data {
+#if (_WIN32_WINNT >= 0x0600)
+  struct thread_sync_data_w8 w8;
+#endif
   curl_mutex_t *mtx;
   int done;
   int port;
@@ -165,6 +189,9 @@ struct thread_sync_data {
 };
 
 struct thread_data {
+#if (_WIN32_WINNT >= 0x0600)
+  HANDLE complete_ev;
+#endif
   curl_thread_t thread_hnd;
   unsigned int poll_interval;
   timediff_t interval_end;
@@ -276,6 +303,144 @@ static CURLcode getaddrinfo_complete(struct Curl_easy *data)
   return result;
 }
 
+
+#if (_WIN32_WINNT >= 0x0600)
+static VOID WINAPI
+query_complete(DWORD err, DWORD bytes, LPOVERLAPPED overlapped)
+{
+  (void)bytes;
+#ifndef CURL_DISABLE_SOCKETPAIR
+  char buf[1];
+#endif
+  const ADDRINFOEXW* ai;
+  struct Curl_addrinfo* cafirst = NULL;
+  struct Curl_addrinfo* calast = NULL;
+  struct Curl_addrinfo* ca;
+  size_t ss_size;
+  int error = (int)err;
+  struct thread_sync_data* tsd =
+    CONTAINING_RECORD(overlapped, struct thread_sync_data, w8.overlapped);
+  struct thread_data* td = tsd->td;
+  const ADDRINFOEXW* res = tsd->w8.res;
+
+  if(error == ERROR_SUCCESS) {
+    /* traverse the addrinfo list */
+  
+    for(ai = res; ai != NULL; ai = ai->ai_next) {
+      size_t namelen = ai->ai_canonname ? wcslen(ai->ai_canonname) + 1 : 0;
+      /* ignore elements with unsupported address family, */
+      /* settle family-specific sockaddr structure size.  */
+      if(ai->ai_family == AF_INET)
+        ss_size = sizeof(struct sockaddr_in);
+#ifdef ENABLE_IPV6
+      else if(ai->ai_family == AF_INET6)
+        ss_size = sizeof(struct sockaddr_in6);
+#endif
+      else
+        continue;
+  
+      /* ignore elements without required address info */
+      if(!ai->ai_addr || !(ai->ai_addrlen > 0))
+        continue;
+  
+      /* ignore elements with bogus address size */
+      if((size_t)ai->ai_addrlen < ss_size)
+        continue;
+  
+      ca = malloc(sizeof(struct Curl_addrinfo) + ss_size + namelen);
+      if(!ca) {
+        error = EAI_MEMORY;
+        break;
+      }
+  
+      /* copy each structure member individually, member ordering, */
+      /* size, or padding might be different for each platform.    */
+      ca->ai_flags     = ai->ai_flags;
+      ca->ai_family    = ai->ai_family;
+      ca->ai_socktype  = ai->ai_socktype;
+      ca->ai_protocol  = ai->ai_protocol;
+      ca->ai_addrlen   = (curl_socklen_t)ss_size;
+      ca->ai_addr      = NULL;
+      ca->ai_canonname = NULL;
+      ca->ai_next      = NULL;
+  
+      ca->ai_addr = (void *)((char *)ca + sizeof(struct Curl_addrinfo));
+      memcpy(ca->ai_addr, ai->ai_addr, ss_size);
+  
+      if(namelen) {
+        ca->ai_canonname = (void *)((char *)ca->ai_addr + ss_size);
+        sprintf(ca->ai_canonname, "%S", ai->ai_canonname);
+      }
+  
+      /* if the return list is empty, this becomes the first element */
+      if(!cafirst)
+        cafirst = ca;
+  
+      /* add this element last in the return list */
+      if(calast)
+        calast->ai_next = ca;
+      calast = ca;
+    }
+  
+    /* if we failed, also destroy the Curl_addrinfo list */
+    if(error) {
+      Curl_freeaddrinfo(cafirst);
+      cafirst = NULL;
+    }
+    else if(!cafirst) {
+#ifdef EAI_NONAME
+      /* rfc3493 conformant */
+      error = EAI_NONAME;
+#else
+      /* rfc3493 obsoleted */
+      error = EAI_NODATA;
+#endif
+#ifdef USE_WINSOCK
+      SET_SOCKERRNO(error);
+#endif
+    }
+    tsd->res = cafirst;
+  }
+
+  if(tsd->w8.res) {
+    FreeAddrInfoExW(tsd->w8.res);
+    tsd->w8.res = NULL;
+  }
+
+  if(error) {
+    tsd->sock_error = SOCKERRNO?SOCKERRNO:error;
+    if(tsd->sock_error == 0)
+      tsd->sock_error = RESOLVER_ENOMEM;
+  }
+  else {
+    Curl_addrinfo_set_port(tsd->res, tsd->port);
+  }
+
+  Curl_mutex_acquire(tsd->mtx);
+  if(tsd->done) {
+    /* too late, gotta clean up the mess */
+    Curl_mutex_release(tsd->mtx);
+    destroy_thread_sync_data(tsd);
+    free(td);
+  }
+  else {
+#ifndef CURL_DISABLE_SOCKETPAIR
+    if(tsd->sock_pair[1] != CURL_SOCKET_BAD) {
+      /* DNS has been resolved, signal client task */
+      buf[0] = 1;
+      if(swrite(tsd->sock_pair[1],  buf, sizeof(buf)) < 0) {
+        /* update sock_erro to errno */
+        tsd->sock_error = SOCKERRNO;
+      }
+    }
+#endif
+    tsd->done = 1;
+    Curl_mutex_release(tsd->mtx);
+    if(td->complete_ev != NULL)
+      SetEvent(td->complete_ev); /* Notify caller that the query completed */
+  }
+}
+#endif
 
 #ifdef HAVE_GETADDRINFO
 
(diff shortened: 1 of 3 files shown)


pps83:
When I wrote the code, in my tests using GetAddrInfoExW I was able to do roughly 30% more requests (or request time was 30% less or something like that).

pps83:
perhaps asyn-thread.c should also be renamed as part of another commit.

bagder:
checksrc has some remarks:
~~~
./asyn-thread.c:316:24: warning: space after declarative asterisk (ASTERISKSPACE)
   struct Curl_addrinfo* cafirst = NULL;
                        ^
./asyn-thread.c:317:24: warning: space after declarative asterisk (ASTERISKSPACE)
   struct Curl_addrinfo* calast = NULL;
                        ^
./asyn-thread.c:318:24: warning: space after declarative asterisk (ASTERISKSPACE)
   struct Curl_addrinfo* ca;
                        ^
./asyn-thread.c:321:27: warning: space after declarative asterisk (ASTERISKSPACE)
   struct thread_sync_data* tsd =
                           ^
./asyn-thread.c:323:22: warning: space after declarative asterisk (ASTERISKSPACE)
   struct thread_data* td = tsd->td;
                      ^
./asyn-thread.c:328:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:341:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:345:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:349:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:355:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:366:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:369:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:372:9: warning: use of sprintf is banned (BANNEDFUNC)
         sprintf(ca->ai_canonname, "%S", ai->ai_canonname);
         ^
./asyn-thread.c:374:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:378:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:384:2: warning: Trailing whitespace (TRAILINGSPACE)
   
  ^
./asyn-thread.c:439:24: warning: we prefer !variable instead of "== NULL" comparisons (EQUALSNULL)
     if(td->complete_ev != NULL)
                        ^
./asyn-thread.c:560:26: warning: we prefer !variable instead of "== NULL" comparisons (EQUALSNULL)
       if(td->complete_ev != NULL)
                          ^
./asyn-thread.c:568:26: warning: we prefer !variable instead of "== NULL" comparisons (EQUALSNULL)
       if(td->complete_ev != NULL) {
                          ^
./asyn-thread.c:644:24: warning: we prefer !variable instead of "== NULL" comparisons (EQUALSNULL)
     if(td->complete_ev != NULL) {
                        ^
./asyn-thread.c:695:22: warning: we prefer !variable instead of "== NULL" comparisons (EQUALSNULL)
   if(td->complete_ev != NULL) {
                      ^
checksrc: 0 errors and 21 warnings
checksrc: 0 errors and 7 warnings suppressed

lib/asyn-thread.c
```diff
@@ -44,6 +44,15 @@
 #include <inet.h>
 #endif
 
+#ifdef _WIN32
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#include <windows.h>
```
vszakats, line 50: Are these needed? We already pull them via `setup_win32.h`?

vszakats:
Two suggestions:
- to use GetAddrInfoExA (and matching code) for non-Unicode builds and the current W flavour for Unicode builds.
- to convert GetAddrInfoEx[AW] to dynamic so libcurl can use this feature when run on Win8, while also staying compatible with older OS versions (IOW without building specifically for Win8 or newer). It means dropping `_WIN32_WINNT >= 0x0600` conditions from the source and using.

jay:
> * to use `GetAddrInfoExA` (and matching code) for non-Unicode builds and the current `W` flavour for Unicode builds.

Only the GetAddrInfoExW version works for overlapped requests in Windows 8, see **lpOverlapped** in [parameters section](https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfoexw#parameters) of doc:

_"On Windows 8 and Windows Server 2012 whenever the UNICODE or \_UNICODE macro is not defined, this parameter is currently reserved and must be set to NULL."_

lib/asyn-thread.c
```diff
@@ -144,9 +153,24 @@ static bool init_resolve_thread(struct Curl_easy *data,
...
 
+#if (_WIN32_WINNT >= 0x0600)
+/* Thread sync data used for GetAddrInfoExW  for win8+ */
+struct thread_sync_data_w8
+{
+    OVERLAPPED overlapped;
+    ADDRINFOEXW *res;
+    HANDLE cancel_ev;
+    WCHAR name[256]; /* max domain name is 253 chars */
+    WCHAR service_name[20]; /* port */
```
bagder, line 164: This field is for holding the decimal 16 bit port number, 5 digits maximum (plus null byte). Why space for 20 wide chars?

lib/asyn-thread.c
```diff
@@ -454,6 +634,24 @@ static bool init_resolve_thread(struct Curl_easy *data,
   /* The thread will set this to 1 when complete. */
   td->tsd.done = 0;
 
+#if (_WIN32_WINNT >= 0x0600)
+  if(Curl_isWindows8OrGreater) {
+    wsprintfW(td->tsd.w8.name, L"%S", hostname);
```
bagder, line 639: This must first check that the host name fits the target buffer as curl does not generally restrict host name lengths to 255 bytes.

bagder (changes requested):

pps83:
> > * to use `GetAddrInfoExA` (and matching code) for non-Unicode builds and the current `W` flavour for Unicode builds.
> 
> Only the GetAddrInfoExW version works for overlapped requests in Windows 8, see **lpOverlapped** in [parameters section](https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfoexw#parameters) of doc:
> 
> _"On Windows 8 and Windows Server 2012 whenever the UNICODE or _UNICODE macro is not defined, this parameter is currently reserved and must be set to NULL."_

Yes, I know, what's the point? The code does use GetAddrInfoExW

lib/asyn-thread.c
bagder: I'm not familiar with wsprintfW(), but does this convert the name somehow? Or put another way: is it necessary or could this function be skipped?

jay: > I'm not familiar with wsprintfW(), but does this convert the name somehow?

wsprintfW (notice the 'W' suffix for wide char) and %S (notice the uppercase 'S') means write a char string (eg hostname) as part of a wide char string. (The MS documentation is confusing. I tried to clarify it at MicrosoftDocs/sdk-api#1713). It's only safe if the wide character buffer is big enough.

curlx_convert_UTF8_to_wchar could be used instead but since the call to GetAddrInfoExW is asynchronous I would free the wide character copy of hostname only after the completion routine is called.

pps83:
I updated the code to dynamically load FreeAddrInfoExW, GetAddrInfoExCancel, and GetAddrInfoExW.\
I didn't remove `#if (_WIN32_WINNT >= 0x0600)` guards to avoid defining all these internal structs. These are vista+ checks, to make it compile, but at runtime it won't lock to vista+. Code itself requires win8.

lib/asyn-thread.c
pps83: i updated the code use `MultiByteToWideChar` (similar code used by `curlx_convert_UTF8_to_wchar` under the hood)

pps83:
@bagder one test fails, but I've seen it pass once if I'm not mistaken. Any ideas what's the reason, is it related?

lib/system_win32.c
```diff
@@ -38,13 +38,24 @@
...
+
 /* Pointer to the if_nametoindex function */
 IF_NAMETOINDEX_FN Curl_if_nametoindex = NULL;
 
+void(WSAAPI *ptrFreeAddrInfoExW)(ADDRINFOEXW_ *pAddrInfoEx) = NULL;
+INT(WSAAPI *ptrGetAddrInfoExCancel)(LPHANDLE lpHandle) = NULL;
+INT(WSAAPI *ptrGetAddrInfoExW)(PCWSTR pName, PCWSTR pServiceName,
+  DWORD dwNameSpace, LPGUID lpNspId, const ADDRINFOEXW_ *hints,
+  ADDRINFOEXW_ **ppResult, struct timeval *timeout, LPOVERLAPPED lpOverlapped,
+  LOOKUP_COMPLETION lpCompletionRoutine, LPHANDLE lpHandle) = NULL;
```
bagder, line 57: Odd indenting. Global variables need to be `Curl_` prefixed if not static.

lib/system_win32.c
```diff
@@ -104,6 +115,16 @@ CURLcode Curl_win32_init(long flags)
...
 
+  s_ws2_32Dll = Curl_load_library(TEXT("ws2_32.dll"));
+  if(s_ws2_32Dll) {
+    *(FARPROC*)&ptrFreeAddrInfoExW = GetProcAddress(s_ws2_32Dll,
+      "FreeAddrInfoExW");
+    *(FARPROC*)&ptrGetAddrInfoExCancel = GetProcAddress(s_ws2_32Dll,
+      "GetAddrInfoExCancel");
+    *(FARPROC*)&ptrGetAddrInfoExW = GetProcAddress(s_ws2_32Dll,
+      "GetAddrInfoExW");
+  }
```
bagder, line 126: an `if` implies it can fail. What happens if the DLL loading fails?

lib/asyn-thread.c
```diff
@@ -276,6 +300,141 @@ static CURLcode getaddrinfo_complete(struct Curl_easy *data)
...
+      ca->ai_addr      = NULL;
+      ca->ai_canonname = NULL;
+      ca->ai_next      = NULL;
+
+      ca->ai_addr = (void *)((char *)ca + sizeof(struct Curl_addrinfo));
+      memcpy(ca->ai_addr, ai->ai_addr, ss_size);
+
+      if(namelen) {
+        ca->ai_canonname = (void *)((char *)ca->ai_addr + ss_size);
+        curl_msprintf(ca->ai_canonname, "%S", ai->ai_canonname);
```
bagder, line 365: Should this really use upper case `%S` here? Quotes around the name?

lib/asyn-thread.c
```diff
@@ -454,6 +628,34 @@ static bool init_resolve_thread(struct Curl_easy *data,
...
 
+#ifdef _WIN32
+  if(Curl_isWindows8OrGreater && ptrFreeAddrInfoExW &&
+      ptrGetAddrInfoExCancel && ptrGetAddrInfoExW) {
+    int hostname_w_len = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS,
+        hostname, -1, NULL, 0);
+    int name_sz = sizeof(td->tsd.w8.name) / sizeof(td->tsd.w8.name[0]);
+    if(hostname_w_len > 0 && hostname_w_len < name_sz)
+      hostname_w_len = MultiByteToWideChar(CP_UTF8, 0, hostname, -1,
+        td->tsd.w8.name, hostname_w_len);
```
bagder, line 639: Maybe a comment to explain the logic and reason for this second call?

lib/asyn-thread.c
```diff
@@ -454,6 +628,34 @@ static bool init_resolve_thread(struct Curl_easy *data,
...
+    if(hostname_w_len > 0 && hostname_w_len < name_sz) {
+      swprintf(td->tsd.w8.port,
+        sizeof(td->tsd.w8.port) / sizeof(td->tsd.w8.port[0]),
+        L"%d", port);
+      td->tsd.w8.hints.ai_family = hints->ai_family;
+      td->tsd.w8.hints.ai_socktype = hints->ai_socktype;
+      td->complete_ev = CreateEvent(NULL, TRUE, FALSE, NULL);
```
bagder, line 646: if `CreateEvent` returns NULL here, is that not an out of memory error or similar that should rather cause an error to be returned?

lib/asyn-thread.c
bagder: Also, `curl_msnprintf` is encouraged.

lib/system_win32.c
pps83: then these will be NULL. When the GetAddrInfoExW is attempted to be used all these are checked in asyn-thread.c line 632

lib/system_win32.c
pps83: updated.

lib/asyn-thread.c
pps83: `td->complete_ev != NULL` indicates that GetAddrInfoExW code is to be used instead of manual thread. If it failed there, then it simply falls back to using regular thread code.

lib/asyn-thread.c
pps83: first one calculates the length needed, second one converts (the code is copied from curlx_convert_UTF8_to_wchar) without doing any allocations

lib/asyn-thread.c
pps83: this simply converts from wide chars to regular chars.

pps83: updated to use curl_msnprintf here

lib/asyn-thread.c
pps83: added comments

lib/system_win32.c
jay: Use Winsock only if it is already loaded. I would guard the block with USE_WINSOCK and then GetModuleHandleA("ws2_32") instead and get rid of the loadlibrary/freelibrary.

lib/asyn-thread.c
bagder: Shouldn't the second argument be the same in both invocations then so that they behave the same?

lib/asyn-thread.c
bagder: No, `curl_msnprintf()` does not convert from wide chars. That's an internal printf() implementation and `%S` makes the output string get enclosed in double quotes. No conversion.

pps83: `%S` is the same as `%s`, but expects source to be `wchar_t` string

bagder: `curl_msnprintf` is an internal printf implementation. It does **no** wchar conversion at all.

lib/asyn-thread.c
pps83: updated to error out

lib/asyn-thread.c
pps83: updated to manually convert wide string to narrow (as I understand sprintf/snprintf are disabled in curl code, so I manually loop over the string).

lib/asyn-thread.c
```diff
@@ -69,6 +69,12 @@
 #include "curl_memory.h"
 #include "memdebug.h"
 
+#ifdef _WIN32
+#include "system_win32.h"
+/* set in win32_init() */
+extern bool Curl_isWindows8OrGreater;
```
bagder, line 75: This should be an extern placed in `system_win32.h` to match the variable which is declared in `system_win32.c`.

You don't need to protect this include with an `#ifdef` as the header itself deals with that.

Finally: you placed this include *below* the includes just above which is said to be used as the last include files in this file: move this include to above those.

pps83: updated

pps83:
updated

lib/asyn-thread.c
```diff
@@ -454,6 +626,37 @@ static bool init_resolve_thread(struct Curl_easy *data,
   /* The thread will set this to 1 when complete. */
   td->tsd.done = 0;
 
+#ifdef _WIN32
+  if(Curl_isWindows8OrGreater && Curl_FreeAddrInfoExW &&
+      Curl_GetAddrInfoExCancel && Curl_GetAddrInfoExW) {
+    int hostname_w_len = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS,
+        hostname, -1, NULL, 0); /* calculate required length */
```
bagder, line 633: BTW, this host name is known to be ascii here. Does that change anything? Do you still need to do this funny wide character dance?

If the name was provided an an IDN, it was already converted before this and the host name is the punycode (all ASCII) version.

pps83:
I'm constantly seeing this test fail. Is it something known?
image

lib/asyn-thread.c
pps83: I could simply loop there instead.

lib/system_win32.c
pps83: the problem is then if the winsock gets unloaded/reloaded during runtime the pointers might end up dangling. LoadLibrary is the right approach here.

bagder:
> I'm constantly seeing this test fail. Is it something known?

Those appveyor ones are unfortunately known to be flaky. I think you can ignore those.

lib/asyn-thread.c
jay: If the name was not IDN though wouldn't it pass through here? It seems safer to leave it

bagder (approved):
@jay anything else we need to do before we merge?

lib/asyn-thread.c
```diff
@@ -454,6 +626,37 @@ static bool init_resolve_thread(struct Curl_easy *data,
...
+        L"%d", port);
+      td->tsd.w8.hints.ai_family = hints->ai_family;
+      td->tsd.w8.hints.ai_socktype = hints->ai_socktype;
+      td->complete_ev = CreateEvent(NULL, TRUE, FALSE, NULL);
+      if(!td->complete_ev) {
+          /* failed to start, mark it as done here for proper cleanup. */
+          td->tsd.done = 1;
+          goto err_exit;
+      }
+      err = Curl_GetAddrInfoExW(td->tsd.w8.name, td->tsd.w8.port, NS_DNS,
```
bagder, line 650: Do the name + port buffers really need to be allocated and stored and kept after this function has been called? Could they not just be on the stack?

bagder: I'm thinking like this:
~~~c
#ifdef _WIN32
  if(Curl_isWindows8OrGreater && Curl_FreeAddrInfoExW &&
     Curl_GetAddrInfoExCancel && Curl_GetAddrInfoExW) {
    WCHAR namebuf[256]; /* max domain name is 253 chars */
    WCHAR portbuf[8]; /* port */
    /* calculate required length */
    int w_len = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS,
                                    hostname, -1, NULL, 0);
    int name_sz = sizeof(namebuf) / sizeof(namebuf[0]);
    if((w_len > 0) && (w_len < name_sz))
      /* do utf8 conversion */
      w_len = MultiByteToWideChar(CP_UTF8, 0, hostname, -1, namebuf, w_len);
    if((w_len > 0) && (w_len < name_sz)) {
      swprintf(portbuf, sizeof(portbuf) / sizeof(portbuf[0]),
               L"%d", port);
      td->tsd.w8.hints.ai_family = hints->ai_family;
      td->tsd.w8.hints.ai_socktype = hints->ai_socktype;
      td->complete_ev = CreateEvent(NULL, TRUE, FALSE, NULL);
      if(!td->complete_ev) {
        /* failed to start, mark it as done here for proper cleanup. */
        td->tsd.done = 1;
        goto err_exit;
      }
      err = Curl_GetAddrInfoExW(namebuf, portbuf, NS_DNS,
                                NULL, &td->tsd.w8.hints, &td->tsd.w8.res, NULL,
                                &td->tsd.w8.overlapped, &query_complete,
                                &td->tsd.w8.cancel_ev);
      if(err != WSA_IO_PENDING)
        query_complete(err, 0, &td->tsd.w8.overlapped);
      return TRUE;
    }
  }
#endif

bagder: Slightly more polished:
~~~c
#ifdef _WIN32
  if(Curl_isWindows8OrGreater && Curl_FreeAddrInfoExW &&
     Curl_GetAddrInfoExCancel && Curl_GetAddrInfoExW) {
#define MAX_NAME_LEN 256
#define MAX_PORT_LEN 8
    WCHAR namebuf[MAX_NAME_LEN];
    WCHAR portbuf[MAX_NAME_LEN];
    /* calculate required length */
    int w_len = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, hostname,
                                    -1, NULL, 0);
    if((w_len > 0) && (w_len < MAX_NAME_LEN)) {
      /* do utf8 conversion */
      w_len = MultiByteToWideChar(CP_UTF8, 0, hostname, -1, namebuf, w_len);
      if((w_len > 0) && (w_len < MAX_NAME_LEN)) {
        swprintf(portbuf, MAX_PORT_LEN, L"%d", port);
        td->tsd.w8.hints.ai_family = hints->ai_family;
        td->tsd.w8.hints.ai_socktype = hints->ai_socktype;
        td->complete_ev = CreateEvent(NULL, TRUE, FALSE, NULL);
        if(!td->complete_ev) {
          /* failed to start, mark it as done here for proper cleanup. */
          td->tsd.done = 1;
          goto err_exit;
        }
        err = Curl_GetAddrInfoExW(namebuf, portbuf, NS_DNS,
                                  NULL, &td->tsd.w8.hints, &td->tsd.w8.res,
                                  NULL, &td->tsd.w8.overlapped,
                                  &query_complete, &td->tsd.w8.cancel_ev);
        if(err != WSA_IO_PENDING)
          query_complete(err, 0, &td->tsd.w8.overlapped);
        return TRUE;
      }
    }
  }
#endif

bagder: Testing this take in #12555

jay:
> @jay anything else we need to do before we merge?

I'd like to look at this again before merging, but I haven't had time yet. There's been a few changes since I last looked.

lib/asyn-thread.c
bagder: worked perfectly. So there's no need to keep that memory around...

pps83: added this update

pps83:
@bagder #12555 is missing some of the changes, while all your suggestions/updates are applied in this PR.

pps83:
@bagder the PR should be good to go. Please merge

bagder:
Thanks a lot @pps83 !

vszakats:
After this patch, this warning appeared (clang 17 + latest mingw-w64):
```
./lib/asyn-thread.c:306:5: warning: cast from 'PCHAR' (aka 'char *') to 'struct thread_sync_data *' increases required alignment from 1 to 8 [-Wcast-align]
  306 |     CONTAINING_RECORD(overlapped, struct thread_sync_data, w8.overlapped);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../aarch64-w64-mingw32/include/winnt.h:717:48: note: expanded from macro 'CONTAINING_RECORD'
  717 | #define CONTAINING_RECORD(address,type,field) ((type *)((PCHAR)(address) - (ULONG_PTR)(&((type *)0)->field)))
      |                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

PR #12485, opened 2023-12-08
Fix MSVC warning C4189: 'htype': local variable is initialized but not referenced when CURL_DISABLE_HEADERS_API is defined.

MAntoniak:


bagder:
Thanks!

PR #12490, opened 2023-12-08
lib: rename Curl_strndup to Curl_memdup0 to avoid misunderstanding

bagder:
Since the copy does not stop at a null byte, let's not call it anything that makes you think it works like the common strndup() function.

Based on feedback from Jay Satiro, Stefan Eissing and Patrick Monnerat

### DIFF
--- a/lib/altsvc.c
+++ b/lib/altsvc.c
@@ -123,11 +123,11 @@ static struct altsvc *altsvc_createid(const char *srchost,
     dlen -= 2;
   }
 
-  as->src.host = Curl_memdump0(srchost, hlen);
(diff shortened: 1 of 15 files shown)


lib/altsvc.c
```diff
@@ -123,11 +123,11 @@ static struct altsvc *altsvc_createid(const char *srchost,
     dlen -= 2;
   }
 
-  as->src.host = Curl_strndup(srchost, hlen);
+  as->src.host = Curl_memdump0(srchost, hlen);
```
monnerat: dump or dup? probably a typo.

PR #12491, opened 2023-12-08
doh: remove unused local variable

danielgustafsson:
The nurl variable is no longer used during probing following a refactoring, so remove.


bagder (approved):
Interesting that no compiler warned about this...

danielgustafsson:
> Interesting that no compiler warned about this...

Agreed, I'd expect static analyzers would pick this one out.

PR #12492, opened 2023-12-09
build: fix some `-Wsign-conversion`/`-Warith-conversion` warnings

vszakats:
- enable `-Wsign-conversion` warnings, but also setting them to not
  raise errors.
- fix `-Warith-conversion` warnings seen in CI.
  These are triggered by `-Wsign-converion` and causing errors unless
  explicitly silenced. It makes more sense to fix them, there just a few
  of them.
- fix some `-Wsign-conversion` warnings.
- hide `-Wsign-conversion` warnings with a `#pragma`.
- add macro `CURL_WARN_SIGN_CONVERSION` to unhide them on a per-build
  basis.
- update a CI job to unhide them with the above macro:
  https://github.com/curl/curl/actions/workflows/linux.yml -> OpenSSL -O3

---

TODO:
- [x] re-silence these warnings before merge.
- [x] fix all `-Wsign-conversion` warnings. [FUTURE]
- [x] delete `CURL_WARN_SIGN_CONVERSION` logic and delete `-Wno-error=sign-conversion` options. [FUTURE]

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -111,7 +111,7 @@ jobs:
           - name: openssl3-O3
             install_packages: zlib1g-dev valgrind
             install_steps: gcc-11 openssl3
-            configure: CFLAGS=-O3 LDFLAGS="-Wl,-rpath,$HOME/openssl3/lib64" --with-openssl=$HOME/openssl3 --enable-debug --enable-websockets
+            configure: CPPFLAGS=-DCURL_WARN_SIGN_CONVERSION CFLAGS=-O3 LDFLAGS="-Wl,-rpath,$HOME/openssl3/lib64" --with-openssl=$HOME/openssl3 --enable-debug --enable-websockets
             singleuse: --unit
 
           - name: openssl3-clang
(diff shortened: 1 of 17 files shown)


vszakats:
`-Wsign-conversion` generates a bunch of warnings. These are made visible in this PRs CI run.

E.g.:
- Linux: https://github.com/curl/curl/actions/runs/7232826763/job/19707427557?pr=12492#step:29:26
- Windows: https://ci.appveyor.com/project/curlorg/curl/builds/48773301/job/kldc93c44ylw715r?fullLog=true

510-520 unique warnings in each, probably with large overlap.

There are a few options to proceed with this:
- Enable it with a `-Wno-error` exception, so that the warnings appear, but don't block. This might not play well with PowerShell in `appveyor.ci`, but it also has the undesired effect of polluting all logs with these warnings all the time, making actual errors difficult to spot and need searching for `error:` in the logs.
- Enable it in cmake / autotools, then silence it globally via a `#pragma` in `curl_setup.h`, with an option to unsilence. This isn't useful on the get go, but allows working on these warnings locally, or in a dedicated CI job. This allows to move around the `#pragma` as ideally more and more code would have these warnings fixed, making it unnecessary at one point.
- Not enable this warning at all.

`-Wsign-conversion` is one of the few recommended compiler hardening options suggested by OpenSSF:
https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++#-Wsign-conversion

Issue #12488, opened 2023-12-08
Repeat WRITEFUNCTION but not call PROGRESSFUNCTION until transmission complete


Cering:
### I did this

- There are some block i/o operation in my `CURLOPT_WRITEFUNCTION`, such as save data into disk, or copy it to another thread.
- Also set `CURLOPT_PROGRESSFUNCTION` to display the download percentage progress.
- The test code please see below.

### I expected the following

- It works well in `curl-7.85.0`, the percentage progress like: 0% -> 1% -> 2% -> ... -> 99% -> 100%.
- But when update to `curl-8.0.1`, I found the `CURLOPT_WRITEFUNCTION` repeat in loop, and not call `CURLOPT_PROGRESSFUNCTION` immediately after receive a buffer. When received first response from server, `CURLOPT_WRITEFUNCTION` will never called until transmission complete, and the percentage progress will be: 0% -> 100%. 

### Test Code
```c
#include <stdlib.h>
#include <unistd.h>
#include "curl/curl.h"

static size_t mycurl_onrecv_body(char *ptr, size_t size,
                                 size_t nmemb, void *userdata)
{
    printf("mycurl_onrecv_body: %u\n", size * nmemb);
    // sleep 20ms, simulate block i/o operation
    usleep(20 * 1000);
    return size * nmemb;
}

static int mycurl_debug(CURL *handle, curl_infotype type, char *data,
                        size_t size, void *userptr)
{
    if(type == CURLINFO_TEXT)
        printf("mycurl_debug[%d]: %s", type, data);
    return 0;
}

static int mycurl_progress(void *clientp, double dltotal, double dlnow,
                           double ultotal, double ulnow)
{
    double percent = dltotal > 0 ? 100.0*dlnow/dltotal : 0;
    printf("mycurl_progress[%.2f%%]: %.2f / %.2f\n", percent, dlnow, dltotal);
    return 0;
}

int main(void)
{
    CURL *http_handle;
    const char *url =
        "http://baiducdncmn2.inter.iqiyi.com/videos/v0/20150311/qycs0001.f4v";

    curl_global_init(CURL_GLOBAL_ALL);
    printf("%s\n", curl_version());

    http_handle = curl_easy_init();
    curl_easy_setopt(http_handle, CURLOPT_URL, url);
    curl_easy_setopt(http_handle, CURLOPT_VERBOSE, 1L);
    curl_easy_setopt(http_handle, CURLOPT_NOPROGRESS, 0L);
[...]

PR #12504, opened 2023-12-11
readwrite_data: loop less

bagder:
This function is made to loop in order to drain incoming data faster. Completely removing the loop has a measerably negative impact on transfer speeds.

Downsides with the looping include

- it might call the progress callback much more seldom. Especially if the write callback is slow.

- rate limiting becomes less exact

- a single transfer might "starve out" other parallel transfers

- QUIC timers for other connections can't be maintained correctly

The long term fix should be to remove the loop and optimize coming back to avoid the transfer speed penalty.

This fix lower the max loop count to reduce the starvation problem, and avoids the loop completely for when rate-limiting is in progress.

Ref: #12488
Ref: https://curl.se/mail/lib-2023-12/0012.html


jay:
Not against this but I think to fully fix the issue the pr could be improved by adding a time constraint. Otherwise the servicing of other transfers may suffer. I'd like to know what you and @icing think about the 1ms limit I proposed in the issue thread. In theory if the data is backed up and available then it be gobbled in less than a millisecond. If a system call or the user's callback holds things up the time constraint would kick in.

bagder:
> think about the 1ms limit I proposed

I wanted this PR to be a *simple* (and temporary?) fix that would limit the impact of the current design without sacrificing too much.

I  figured such a time-limiting approach or calling the progress meter from within this callback etc to be more complicated for a stop-gap solution, when the actual and best fix is to avoid the looping altogether.

A time check is not bad per se, I just figure maybe an even simpler take can work for now and we then instead work on a "proper" fix going forward.

jay:
Ok

PR #12505, opened 2023-12-11
gnutls: fix build with --disable-verbose

baruchsiach:
infof() parameters must be defined event with --disable-verbose since commit dac293cfb702 ("lib: apache style infof and trace macros/functions").

### DIFF
--- a/lib/vtls/gtls.c
+++ b/lib/vtls/gtls.c
@@ -818,16 +818,17 @@ Curl_gtls_verifyserver(struct Curl_easy *data,
   char certname[65] = ""; /* limited to 64 chars by ASN.1 */
   size_t size;
   time_t certclock;
-  const char *ptr;
   int rc;
   CURLcode result = CURLE_OK;
 #ifndef CURL_DISABLE_VERBOSE_STRINGS
+  const char *ptr;
   unsigned int algo;
   unsigned int bits;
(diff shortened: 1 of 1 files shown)


lib/vtls/gtls.c
```diff
@@ -833,8 +833,10 @@ Curl_gtls_verifyserver(struct Curl_easy *data,
                                      gnutls_cipher_get(session),
                                      gnutls_mac_get(session));
 
+#ifndef CURL_DISABLE_VERBOSE_STRINGS
   infof(data, "SSL connection using %s / %s",
         gnutls_protocol_get_name(version), ptr);
+#endif
```
vszakats, line 839: Should we also deal with the `ptr` value is assigned by unused with `--disabled-verbose`? Similar cases often cause a `-Wunused-variable`.

bagder: Right, this function should not declare and assign `ptr` when not using it. @baruchsiach can you add that into this fix?

baruchsiach: > Right, this function should not declare and assign `ptr` when not using it. @baruchsiach can you add that into this fix?

Done.

vszakats (approved):

bagder (approved):

bagder:
Thanks!

PR #12507, opened 2023-12-12
Documentation about client writer design and use

icing:


bagder (approved):
:+1: Don't forget to also add this file name to the `EXTRA_DIST` list in `docs/Makefile.am` to include it in the dist tarball

icing:
Thanks for the heads up. done.

PR #1288, opened 2017-02-25
cmake: Build manual pages (curl.1 and libcurl (3) manpages)

Lekensteyn:
Also make Perl mandatory to allow building the docs.

While CMakeLists.txt could probably read the list of manual pages from
Makefile.am, actually putting those in CMakeLists.txt is cleaner so that
is what is done here.

Fixes #1230
___
cmake: add support for building HTML and PDF docs

Note that for some reason there is this warning (that also exists with
autotools, added since curl-7_15_1-94-ga718cb05f):

    docs/libcurl/curl_multi_socket_all.3:1: can't open `man3/curl_multi_socket.3': No such file or directory

Additionally, adjust the roffit --mandir option to support creating
links when doing out-of-tree builds.
___
(NOTE: this does not install manpages yet)

PR #12509, opened 2023-12-12
cmake: delete obsolete TODOs more [ci skip]

vszakats:
- manual completed: 898b012a9bf388590c4be7f526815b5ab74feca1 #1288
- soname completed: 5de6848f104d7cb0017080e31216265ac19d0dde #10023
- bunch of others that are completed
- `NTLM_WB_ENABLED` is implemented in a basic form, and now also
  scheduled for removal, so a TODO at this point isn't useful.

And this 'to-check' item:

Q: "The cmake build selected to run gcc with -fPIC on my box while the
   plain configure script did not."

A: With CMake, since 2ebc74c36a19a1700af394c16855ce144d9878e3 #11546
   and fc9bfb14520712672b4784e8b48256fb29204011 #11627, we explicitly
   enable PIC for libcurl shared lib. Or when building libcurl for
   shared and static lib in a single pass. We do this by default for
   Windows or when enabled by the user via `SHARE_LIB_OBJECT`.
   Otherwise we don't touch this setting. Meaning the default set by
   CMake (if any) or the toolchain is used. On Debian Bookworm, this
   means that PIC is disabled for static libs by default. Some platforms
   (like macOS), has PIC enabled by default.
   autotools supports the double-pass mode only, and in that case
   CMake seems to match PIC behaviour now (as tested on Linux with gcc.)

Follow-up to 5d5dfdbd1a6c40bd75e982b66f49e1fa3a7eeae7 #12500

Closes #12509

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,15 +21,7 @@
 # SPDX-License-Identifier: curl
 #
 ###########################################################################
-# curl/libcurl CMake script
 # by Tetetest and Sukender (Benoit Neil)
-
-# TODO:
-# The output .so file lacks the soname number which we currently have within the lib/Makefile.am file
-# Add documentation subproject
-#
-# To check:
-# (From Daniel Stenberg) The cmake build selected to run gcc with -fPIC on my box while the plain configure script did not.
 
 # Note: By default this CMake build script detects the version of some
 # dependencies using `check_symbol_exists`.  Those checks do not work
@@ -1524,7 +1516,6 @@ if(NOT CURL_DISABLE_INSTALL)
   endmacro()
 
   # NTLM support requires crypto function adaptions from various SSL libs
-  # TODO alternative SSL libs tests for SSP1, GnuTLS, NSS
   if(NOT (CURL_DISABLE_NTLM) AND
       (USE_OPENSSL OR USE_MBEDTLS OR USE_DARWINSSL OR USE_WIN32_CRYPTO OR USE_GNUTLS))
     set(use_curl_ntlm_core ON)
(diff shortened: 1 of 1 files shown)


jay:
> Q: "The cmake build selected to run gcc with -fPIC on my box while the plain configure script did not."
> 
> A: With CMake, since [2ebc74c](https://github.com/curl/curl/commit/2ebc74c36a19a1700af394c16855ce144d9878e3) #11546 and [fc9bfb1](https://github.com/curl/curl/commit/fc9bfb14520712672b4784e8b48256fb29204011) #11627, we explicitly enable PIC for libcurl shared lib. Or when building libcurl for shared and static lib in a single pass. We do this by default for Windows or when enabled by the user via `SHARE_LIB_OBJECT`. Otherwise we don't touch this setting. Meaning the default set by CMake (if any) or the toolchain is used. On Debian Bookworm, this means that PIC is disabled for static libs by default. Some platforms (like macOS), has PIC enabled by default. autotools supports the double-pass mode only, and in that case CMake seems to match PIC behaviour now (as tested on Linux with gcc.)

Ideally this should be noted somewhere other than the commit message

vszakats:
I think the point is that it now works as it should (if my tests were accurate), and therefore need no special documentation. If it'd help, I'm happy to include it in this commit, but not volunteering to write it myself.

jay:
Alright well do you think something like this is worth it:
~~~diff
diff --git a/docs/INSTALL.cmake b/docs/INSTALL.cmake
index 4e7f706..b103da5 100644
--- a/docs/INSTALL.cmake
+++ b/docs/INSTALL.cmake
@@ -87,3 +87,11 @@ cmake-gui
         GUI.  Once you have selected all the options you want, click the
         "Generate" button.
         6. Run the native build tool that you used CMake to generate.
+
+Notes
+=====
+   If you build curl as a shared library then the position indepenent code
+   option is enabled if the compiler supports it. If you build curl as a shared
+   library and a static library at the same time,
+   like `-DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=ON`, then the position
+   independent code option is enabled for both if the compiler supports it.
~~~

vszakats:
With `-DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=ON`, PIC is enabled for both only when doing a single-pass build with `SHARE_LIB_OBJECT=YES`. This is the default for Windows. In other cases (`SHARE_LIB_OBJECT=NO`) the two lib flavours are built separately, shared with PIC, static without.

jay:
Nevermind. Though it may be unexpected researching it just now I don't think adding PIC is a cause for concern.

jay (approved):

vszakats:
PIC is an almost cost-free on 64-bit systems [^1]. It's also a prerequisite for ASLR. Hence on Windows I think it is safe to have enabled by default. On macOS/iOS store apps it has been a requirement since 2011. It's also the default in Xcode, so even there it'd be safe to enable by default. On *nix systems the picture is less clear but most (non-embedded) systems are likely shifting into this direction or already there, I don't know. In curl-for-win Linux builds I enabled it.

[ I'm in favour of PIC, unless there is a specific reason not to use it. ]

[^1]: <https://nebelwelt.net/publications/files/12TRpie.pdf>

PR #12512, opened 2023-12-13
Eliminate `conn->cselect_bits` and one parameter

icing:
- use `data->state.dselect_bits` everywhere instead
- remove `bool *comeback` parameter as non-zero `data->state.dselect_bits` will indicate that IO is incomplete.

### DIFF
--- a/lib/easy.c
+++ b/lib/easy.c
@@ -1111,7 +1111,7 @@ CURLcode curl_easy_pause(struct Curl_easy *data, int action)
     if(!data->state.tempcount)
       /* if not pausing again, force a recv/send check of this connection as
          the data might've been read off the socket already */
-      data->conn->cselect_bits = CURL_CSELECT_IN | CURL_CSELECT_OUT;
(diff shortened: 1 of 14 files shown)


lib/easy.c
```diff
@@ -1111,7 +1111,7 @@ CURLcode curl_easy_pause(struct Curl_easy *data, int action)
     if(!data->state.tempcount)
       /* if not pausing again, force a recv/send check of this connection as
          the data might've been read off the socket already */
-      data->conn->cselect_bits = CURL_CSELECT_IN | CURL_CSELECT_OUT;
+      data->state.dselect_bits = CURL_CSELECT_IN | CURL_CSELECT_OUT;
```
bagder, line 1114: What do you think about renaming this struct field to plain `select_bits` now when there is only one? The `d` prefix seems mostly arbitrary anyway.

icing: Sure. done.

PR #12515, opened 2023-12-14
cmake: fix generation for system name iOS

marksinkovics:
This PR fixes a problem that happens during CMake configuration when the `CMAKE_SYSTEM_NAME` set to `iOS` and not `Darwin`. This value is available (as far as I remember) version 3.14. The final solution (thanks to @vszakats) is to use `APPLE` which contains all the Apple platforms https://cmake.org/cmake/help/latest/variable/APPLE.html.

This issue was found when during vcpkg installation. Running command `vcpkg install curl:arm64-ios` and `vcpkg install curl:x64-ios` failed with message 
```
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   HAVE_H_ERRNO_ASSIGNABLE_EXITCODE (advanced)
```
After this fix, I was able to compile the compile the binary without any issue.

In addition to that fix, this PR also contains an simplification to check if the platform is not APPLE.

Thank you very much for reviewing!


CMake/OtherTests.cmake
```diff
@@ -142,6 +142,7 @@ elseif(NOT HAVE_GETADDRINFO)
   set(HAVE_GETADDRINFO_THREADSAFE FALSE)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "AIX" OR
        CMAKE_SYSTEM_NAME STREQUAL "Darwin" OR
+       CMAKE_SYSTEM_NAME STREQUAL "iOS" OR
        CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" OR
```
vszakats, line 146:
```suggestion
elseif(APPLE OR
       CMAKE_SYSTEM_NAME STREQUAL "AIX" OR
       CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" OR
```
Would there be any downside to use `APPLE`? It includes these two flavours and the remaining ones.

vszakats: There is a similar case in `CMake/OtherTests.cmake`. This is the patch I have pending for it; we could include it in this PR, if correct:
```diff
--- a/CMake/OtherTests.cmake
+++ b/CMake/OtherTests.cmake
@@ -86,9 +86,7 @@ endif()
 unset(CMAKE_TRY_COMPILE_TARGET_TYPE)
.
 if(NOT CMAKE_CROSSCOMPILING)
-  if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "iOS")
-    # only try this on non-apple platforms
-
+  if(NOT APPLE)
     # if not cross-compilation...
     set(CMAKE_REQUIRED_FLAGS "")
     if(HAVE_SYS_POLL_H)
```

marksinkovics: Thank you very much for the suggestion! Using APPLE is even a better solution and with that all the Apple platforms are covered: https://cmake.org/cmake/help/latest/variable/APPLE.html 

I updated my PR with your suggestion, and I also added change the other line you mentioned. I tested locally and  it worked properly on my side.

Can you please check it again?

vszakats (approved):

CMake/OtherTests.cmake
vszakats: LGTM!

CMake/OtherTests.cmake
BoscoBecker, line 1: nice

vszakats:
Thank you, merged now!

PR #12521, opened 2023-12-14
lib: move async resolve struct to Curl_easy from connectdata

bagder:
The name resolve is done *for* the connection and not *in* it. By keeping the data in the easy handle, we can better reuse the c-ares channel which should be a performance boost instead of creating it and killing it for every new connection.


bagder:
I now realize this move (easy => conn) was just done in #12198... Hm. Time to think.

PR #12538, opened 2023-12-16
cmake: whitespace tidy-up in `OtherTests.cmake`

vszakats:
Closes #12538


Kelly0661:
Markdown is supported

PR #12540, opened 2023-12-16
build: more `-Wformat` fixes

vszakats:
- memdebug: update to not trigger `-Wformat-nonliteral` warnings.
- imap: mark `imap_sendf()` with  `CURL_PRINTF()`.
- tool_msgs: mark static function with `CURL_PRINTF()`.

Closes #12540


vszakats:
Existing memdebug code sometimes used `%ld` for the socket type, but the `CURL_FORMAT_SOCKET_T` macro doesn't, so this might need more tweak.

PR #12543, opened 2023-12-17
Append the flags for the polling notification of libevent, following …

TheBetterSolution:
Append the flags for the polling notification of libevent, following the advice from: http://lists.danga.com/pipermail/memcached/2003-October/000336.html.


bagder:
I don't see how the details of the referenced post motivates the provided code change. In particular because we have seen any bug reported in this area. Please tell us more about why we want this change.

PR #12545, opened 2023-12-17
tests: Respect $TMPDIR when creating sockets

yedayak:
When running on termux, where $TMPDIR isn't /tmp, running the tests failed, since the server config tried creating sockets in /tmp, without checking the temp dir config. Use the TMPDIR variable that makes it find the correct directory everywhere [0]

[0] https://perldoc.perl.org/File::Temp#tempfile


bagder (approved):

bagder:
Thanks!

PR #12547, opened 2023-12-18
runtests: support -gl. Like -g but for lldb.

bagder:
Invokes the test case via lldb instead of gdb. Since using gdb is such a pain on mac, using lldb is sometimes less quirky.


bagder:
-g was introduced in 63b5748eb656c632266ecad1521bfa5012ad3676

PR #12567, opened 2023-12-20
Makefile.am: fix the MSVC project generation

### LINKED ISSUE #12564: Windows vcxproj files are missing
### I did this

Heyya

I just tried to updated to curl stable 8.5.0, and saw that the libcurl.vcxproj files are missing.
they are also missing here on git:
https://github.com/curl/curl/tree/master/projects/Windows/VC14.30/lib

They where there until 8.4.0.

would be nice if those can be added back

### I expected the following

-

### curl/libcurl version

curl 8.5.0, git

### operating system

Windows

Edit:
to clarify:
https://curl.se/download.html - download curl-8.5.0.zip from there, and the files are missing
https://curl.se/download/ - download curl-8.4.0.zip from here, and the files are there

bagder:
It made the vcxproj files not get included in dist tarballs.

Regression since 74423b5df4c8117891eb89 (8.5.0)

Reported-by: iAroc on github
Fixes #12564

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -127,8 +127,9 @@ EXTRA_DIST = CHANGES COPYING maketgz Makefile.dist curl-config.in            \
  $(VC_DIST) $(WINBUILD_DIST) $(PLAN9_DIST) lib/libcurl.vers.in buildconf.bat \
  libcurl.def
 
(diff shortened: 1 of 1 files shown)


bagder:
With this PR applied I ran `maketgz` and it included the vcxproj files in the tarballs.

PR #12570, opened 2023-12-21
lib: fix the Memory leak: as

zengwei2000:
lib: fix the Memory leak: as

Signed-off-by: zengwei zengwei1@uniontech.com


bagder (approved):
Nice catch! This really should never happen though, as the asserts would imply. Have you seen this leak occur for real ?

PR #12572, opened 2023-12-21
appveyor: replace PowerShell with bash + parallel autotools

### LINKED ISSUE #12560: AppVeyor build failures do not always show in log output
I'm reviewing some CI jobs from a [recent PR](https://github.com/curl/curl/pull/12482) and I noticed that a bunch failed in appveyor. For example [this one](https://ci.appveyor.com/project/curlorg/curl/builds/48795111/job/03e930dvk7t6swi9?fullLog=true). When they fail the last line of the log always looks something like:

`The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: ../../../docs/examples/10-at-a-time.c: In function 'main':`

but that's it, that's the last line. It doesn't actually show the compilation error so I don't know what needs to be fixed.

vszakats:
PowerShell works (after a steep development curve), but one property of
it stuck and kept causing unresolvable usability issues: With
`$ErrorActionPreference=Stop`, it does abort on failures, but shows only
the first line of the error message. In `Continue` mode, it shows the
full error message, but doesn't stop on all errors. Another issue is
PowerShell considering any stderr output as if the command failed (this
has been improved in 7.2 (2021-Nov), but fixed versions aren't running
in CI and will not be for a long time in all test images.)

Thus, we're going with bash.

Also:
- use `-j2` with autotools tests, making them finish 5-15 minutes per
  job faster.
- omit `POSIX_PATH_PREFIX`.
- use `WINDIR`.
- prefer forward slashes.

Follow-up to: 75078a415d9c769419aed4153d3d525a8eba95af #11999
Ref: #12444

Fixes #12560
Closes #12572


appveyor.sh
```diff
@@ -0,0 +1,134 @@
...
+# This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+# KIND, either express or implied.
+#
+# SPDX-License-Identifier: curl
+#
+###########################################################################
+
+# build_script:
+
+set -x
```
jay, line 28: should this be `-e` as well so that the script aborts on error,  also don't the calls to batch files need to be like `cmd /c foo.bat`

vszakats: Yeah, this was an early version of this script. Fixed now, along with other things.

PR #12574, opened 2023-12-21
Added curl_multi_pollfds() function allowing to get "waitids" for polling from the multi handle.

dkarpov1970:
This function extracts `curl_waitfd` structures which are similar to poll(2)'s pollfd structure from a given multi_handle.

These structures can be used for polling on multi_handle file descriptors in a fashion similar to curl_multi_poll(3). The curl_multi_perform(3) function should be called as soon as one of them is ready to be read from or written to.

### DIFF
--- a/docs/examples/10-at-a-time.c
+++ b/docs/examples/10-at-a-time.c
@@ -95,7 +95,7 @@ static size_t write_cb(char *data, size_t n, size_t l, void *userp)
   return n*l;
 }
 
-static void add_transfer(CURLM *cm, int i, int *left)
+static void add_transfer(CURLM *cm, unsigned int i, int *left)
 {
   CURL *eh = curl_easy_init();
   curl_easy_setopt(eh, CURLOPT_WRITEFUNCTION, write_cb);
--- a/docs/examples/anyauthput.c
+++ b/docs/examples/anyauthput.c
@@ -69,17 +69,15 @@ static int my_seek(void *userp, curl_off_t offset, int origin)
 /* read callback function, fread() look alike */
 static size_t read_callback(char *ptr, size_t size, size_t nmemb, void *stream)
 {
-  ssize_t retcode;
-  unsigned long nread;
+  size_t nread;
 
-  retcode = fread(ptr, size, nmemb, stream);
+  nread = fread(ptr, size, nmemb, stream);
 
-  if(retcode > 0) {
-    nread = (unsigned long)retcode;
-    fprintf(stderr, "*** We read %lu bytes from file\n", nread);
+  if(nread > 0) {
+    fprintf(stderr, "*** We read %lu bytes from file\n", (unsigned long)nread);
   }
 
-  return retcode;
+  return nread;
 }
 
 int main(int argc, char **argv)
--- a/docs/examples/ftpget.c
+++ b/docs/examples/ftpget.c
@@ -42,7 +42,7 @@ static size_t my_fwrite(void *buffer, size_t size, size_t nmemb, void *stream)
     /* open file for writing */
     out->stream = fopen(out->filename, "wb");
     if(!out->stream)
-      return -1; /* failure, cannot open file to write */
+      return 0; /* failure, cannot open file to write */
   }
   return fwrite(buffer, size, nmemb, out->stream);
 }
--- a/docs/examples/ftpsget.c
+++ b/docs/examples/ftpsget.c
@@ -44,7 +44,7 @@ static size_t my_fwrite(void *buffer, size_t size, size_t nmemb,
     /* open file for writing */
     out->stream = fopen(out->filename, "wb");
     if(!out->stream)
-      return -1; /* failure, cannot open file to write */
+      return 0; /* failure, cannot open file to write */
   }
   return fwrite(buffer, size, nmemb, out->stream);
 }
--- a/docs/examples/http2-download.c
+++ b/docs/examples/http2-download.c
@@ -54,7 +54,7 @@ struct transfer {
 #define NUM_HANDLES 1000
 
 static
-void dump(const char *text, int num, unsigned char *ptr, size_t size,
+void dump(const char *text, unsigned int num, unsigned char *ptr, size_t size,
           char nohex)
 {
   size_t i;
@@ -66,7 +66,7 @@ void dump(const char *text, int num, unsigned char *ptr, size_t size,
     /* without the hex output, we can fit more on screen */
     width = 0x40;
 
-  fprintf(stderr, "%d %s, %lu bytes (0x%lx)\n",
+  fprintf(stderr, "%u %s, %lu bytes (0x%lx)\n",
           num, text, (unsigned long)size, (unsigned long)size);
 
   for(i = 0; i<size; i += width) {
--- a/docs/examples/sendrecv.c
+++ b/docs/examples/sendrecv.c
@@ -44,6 +44,13 @@ static int wait_on_socket(curl_socket_t sockfd, int for_recv, long timeout_ms)
   FD_ZERO(&outfd);
   FD_ZERO(&errfd);
 
+/* Avoid this warning with pre-2020 Cygwin/MSYS releases:
+ * warning: conversion to 'long unsigned int' from 'curl_socket_t' {aka 'int'} may change the sign of the result [-Wsign-conversion]
+ */
+#if defined(__GNUC__) && defined(__CYGWIN__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+#endif
   FD_SET(sockfd, &errfd); /* always check for error */
 
   if(for_recv) {
@@ -52,6 +59,9 @@ static int wait_on_socket(curl_socket_t sockfd, int for_recv, long timeout_ms)
   else {
     FD_SET(sockfd, &outfd);
   }
+#if defined(__GNUC__) && defined(__CYGWIN__)
+#pragma GCC diagnostic pop
+#endif
 
   /* select() returns the number of signalled sockets or -1 */
   res = select((int)sockfd + 1, &infd, &outfd, &errfd, &tv);
--- a/docs/examples/sftpget.c
+++ b/docs/examples/sftpget.c
@@ -53,7 +53,7 @@ static size_t my_fwrite(void *buffer, size_t size, size_t nmemb,
     /* open file for writing */
     out->stream = fopen(out->filename, "wb");
     if(!out->stream)
-      return -1; /* failure, cannot open file to write */
+      return 0; /* failure, cannot open file to write */
   }
   return fwrite(buffer, size, nmemb, out->stream);
 }
--- a/tests/http/clients/h2-download.c
+++ b/tests/http/clients/h2-download.c
@@ -112,11 +112,11 @@ static size_t my_write_cb(char *buf, size_t nitems, size_t buflen,
                           void *userdata)
 {
   struct transfer *t = userdata;
-  ssize_t nwritten;
+  size_t nwritten;
 
   if(!t->resumed &&
      t->recv_size < t->pause_at &&
-     ((curl_off_t)(t->recv_size + (nitems * buflen)) >= t->pause_at)) {
+     ((t->recv_size + (curl_off_t)(nitems * buflen)) >= t->pause_at)) {
     fprintf(stderr, "[t-%d] PAUSE\n", t->idx);
     t->paused = 1;
     return CURL_WRITEFUNC_PAUSE;
@@ -131,11 +131,11 @@ static size_t my_write_cb(char *buf, size_t nitems, size_t buflen,
   }
 
   nwritten = fwrite(buf, nitems, buflen, t->out);
-  if(nwritten < 0) {
+  if(nwritten < buflen) {
     fprintf(stderr, "[t-%d] write failure\n", t->idx);
     return 0;
   }
-  t->recv_size += nwritten;
+  t->recv_size += (curl_off_t)nwritten;
   return (size_t)nwritten;
 }
 
@@ -171,7 +171,7 @@ static void usage(const char *msg)
     "  download a url with following options:\n"
     "  -m number  max parallel downloads\n"
     "  -n number  total downloads\n"
-    "  -p number  pause transfer after `number` response bytes\n"
+    "  -P number  pause transfer after `number` response bytes\n"
   );
 }
 
@@ -185,7 +185,7 @@ int main(int argc, char *argv[])
   const char *url;
   size_t i, n, max_parallel = 1;
   size_t active_transfers;
-  long pause_offset = 0;
+  size_t pause_offset = 0;
   int abort_paused = 0;
   struct transfer *t;
   int ch;
@@ -205,7 +205,7 @@ int main(int argc, char *argv[])
       transfer_count = (size_t)strtol(optarg, NULL, 10);
       break;
     case 'P':
-      pause_offset = strtol(optarg, NULL, 10);
+      pause_offset = (size_t)strtol(optarg, NULL, 10);
       break;
     default:
      usage("invalid option");
@@ -234,7 +234,7 @@ int main(int argc, char *argv[])
   for(i = 0; i < transfer_count; ++i) {
     t = &transfers[i];
     t->idx = (int)i;
-    t->pause_at = (curl_off_t)pause_offset * i;
+    t->pause_at = (curl_off_t)(pause_offset * i);
   }
 
   n = (max_parallel < transfer_count)? max_parallel : transfer_count;
--- a/tests/server/mqttd.c
+++ b/tests/server/mqttd.c
@@ -497,7 +497,7 @@ static curl_socket_t mqttit(curl_socket_t fd)
   unsigned short packet_id;
   size_t payload_len;
   size_t client_id_length;
-  unsigned int topic_len;
+  size_t topic_len;
   size_t remaining_length = 0;
   size_t bytes = 0; /* remaining length field size in bytes */
   char client_id[MAX_CLIENT_ID_LENGTH];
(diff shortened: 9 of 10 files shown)


bagder:
I think you need this to fix singleuse:
~~~diff
diff --git a/scripts/singleuse.pl b/scripts/singleuse.pl
index b8a57f8d8..6145cb997 100755
--- a/scripts/singleuse.pl
+++ b/scripts/singleuse.pl
@@ -104,10 +104,11 @@ my %api = (
     'curl_multi_setopt' => 'API',
     'curl_multi_socket' => 'API',
     'curl_multi_socket_action' => 'API',
     'curl_multi_socket_all' => 'API',
     'curl_multi_poll' => 'API',
+    'curl_multi_pollfds' => 'API',
     'curl_multi_strerror' => 'API',
     'curl_multi_timeout' => 'API',
     'curl_multi_wait' => 'API',
     'curl_multi_wakeup' => 'API',
     'curl_mvaprintf' => 'API',

bagder:
As I mentioned on the mailing list, this function iterates over all transfers to get their sockets, and since more than one transfer may use the same socket this created pollset may then add the same socket multiple times, potentially even with different events.

I believe it might be better to instead iterate over the sockets in the socket hash and ignore which transfers they belong to. Perhaps something like this:

~~~c
CURLMcode curl_multi_pollfds(struct Curl_multi *multi,
                             struct curl_waitfd *ufds,
                             unsigned int size,
                             unsigned int *fd_count)
{
  unsigned int out = 0;
  struct Curl_hash *h;
  struct Curl_hash_iterator iter;
  struct Curl_hash_element *he;

  if(!GOOD_MULTI_HANDLE(multi))
    return CURLM_BAD_HANDLE;

  if(multi->in_callback)
    return CURLM_RECURSIVE_API_CALL;

  /* The multi handles holds N number of transfers using M sockets. Since more
   * than one transfer might use the same sockets, this function iterates over
   * the entries in the socket hash rather than over all transfers, so that it
   * only ever adds each unique socket once to the output array. */

  h = &multi->sockhash;
  DEBUGASSERT(h);

  /* Now populate the output array with the sockets and the actions they wait
   * for. */
  h = &multi->sockhash;
  Curl_hash_start_iterate(h, &iter);
  for(he = Curl_hash_next_element(&iter); he;
      he = Curl_hash_next_element(&iter)) {
    if(out >= size)
      return CURLM_OUT_OF_MEMORY;
    else {
      struct Curl_sh_entry *sh = (struct Curl_sh_entry *)he->ptr;
      struct curl_waitfd *p = &ufds[out++];
      p->fd = sh->fd;
      p->events =
        ((sh->action & CURL_POLL_IN) ? CURL_WAIT_POLLIN : 0) |
        ((sh->action & CURL_POLL_OUT) ? CURL_WAIT_POLLOUT : 0);
      p->revents = 0;
    }
  }

  *fd_count = out;

  return CURLM_OK;
}

dkarpov1970:
I am a bit confused. I copied iteration logic from the multi_wait() function. 
Doesn't the multi_wait() has the same issue when the same socket may belong to multiple transfers? Or it handles it differently?

bagder:
> Doesn't the multi_wait() has the same issue when the same socket may belong to multiple transfers

I'm pretty sure it has!

dkarpov1970:
Hmm... I tried your function (I had to replace `p->fd = sh->fd;` by `p->fd = *((curl_socket_t*)he->key);` to make it compile) but somehow it doesn't work as expected.

By some reason, I can never enter the for-loop during transfers execution as the Curl_hash_next_element() never gives me a not null  element.
As far as I can see, the new entries are added to the `multi->sockhash` in the `singlesocket()` function , but I see the `singlesocket()` is called in my test (H2 transfers sharing the same socket) only after some transfer is completed and I call `curl_multi_remove_handle()`.

I am using the `curl_multi_perform() `, so it seems that the `multi->sockhash` is not filled when it is used.
Maybe it  may work only for `curl_multi_socket_action` ?

dkarpov1970:
I came up with a simple linear search to filter out already added descriptors:

`CURLMcode curl_multi_pollfds(struct Curl_multi *multi,
                             struct curl_waitfd *ufds,
                             unsigned int size,
                             unsigned int *fd_count)
{
  struct Curl_easy *data;
  unsigned int nfds = 0;
  struct easy_pollset ps;
  unsigned int i;
  CURLMcode result = CURLM_OK;
  struct curl_waitfd* ufd;
  unsigned int j;

  if(!GOOD_MULTI_HANDLE(multi))
    return CURLM_BAD_HANDLE;

  if(multi->in_callback)
    return CURLM_RECURSIVE_API_CALL;

  memset(&ps, 0, sizeof(ps));
  for(data = multi->easyp; data; data = data->next) {
    multi_getsock(data, &ps);

    for(i = 0; i < ps.num; i++) {
      if(nfds < size) {
        curl_socket_t fd = ps.sockets[i];
        int fd_idx = -1;

        /* Simple linear search to skip addition of already added descriptor */
        for(j = 0; j < nfds && fd_idx < 0; j++) {
          if(ufds[j].fd == fd) {
            fd_idx = (int)j;
            break;
          }
        }
        
        if (fd_idx < 0) {
          ufd = &ufds[nfds++];
          ufd->fd = ps.sockets[i];
          ufd->events = 0;
        } else
          ufd = &ufds[fd_idx];

        if(ps.actions[i] & CURL_POLL_IN)
          ufd->events |= CURL_WAIT_POLLIN;
        if(ps.actions[i] & CURL_POLL_OUT)
          ufd->events |= CURL_WAIT_POLLOUT;
      }
      else if(size > 0) {
        result = CURLM_OUT_OF_MEMORY;
        if(!fd_count)
          return result;
      }
    }
  }

  if(fd_count)
    *fd_count = nfds;
  return result;
}`

dkarpov1970:
I can see that most of the failures are like: TESTFAIL: These test cases failed: 1135 .
I am wondering how I should fix it? 
Should I just modify tests\data\test1135 and add curl_multi_pollfds in the "stdout" section?

dfandrich:
[test1135](https://github.com/curl/curl/blob/master/tests/data/test1135) says that new externs must be added to the end of the list to maintain backward compatibility. That implies that the new declaration must also be added after all the others in the headers.

dkarpov1970:
I guess, the "list" means the list of external functions in the `libcurl.def`, right?
So, to fix the test1135, I will need to move the declaration of  the `curl_multi_pollfds`  in the libcurl.def  down to the end of the list (because it is a new function), and update the "stdout" section accordingly. Is it correct?

bagder:
The list it compares with is the list in `test1135`. The test extracts all functions from the header files and compare that it is the known list as provided in the test.

dkarpov1970:
Got it, thanks! Corrected the order.

dkarpov1970:
Somehow, I can't make the Test1279 (Verify libcurl.def against CURL_EXTERN declarations) pass.
It seems to rely on some sorted order in the `libcurl.def`, which should match the list of "extern" extracted from the header files.

I tried both sorted order in the `libcurl.def` with `curl_multi_pollfds()` and unsorted (placing it in the same place as in the `multi.h` header) but it didn't work. 
Would appreciate an advice where I should put the `curl_multi_pollfds()` in the `libcurl.def` to make this test pass.

libcurl.def
```diff
@@ -91,4 +92,4 @@ curl_version
 curl_version_info
 curl_ws_meta
 curl_ws_recv
-curl_ws_send
+curl_ws_send
```
vszakats, line 95: I suggest keeping the ending linefeed.

dkarpov1970: Yeah, it was linefeed ending missing, indeed! Thanks a lot!
The remaining test failures don't seem to be caused by my changes.

lib/multi.c
```diff
@@ -1116,6 +1116,68 @@ CURLMcode curl_multi_fdset(struct Curl_multi *multi,
...
+        }
+        else
+          ufd = &ufds[fd_idx];
+
+        if(ps.actions[i] & CURL_POLL_IN)
+          ufd->events |= CURL_WAIT_POLLIN;
+        if(ps.actions[i] & CURL_POLL_OUT)
+          ufd->events |= CURL_WAIT_POLLOUT;
+      }
+      else if(size > 0) {
```
bagder, line 1168: Why this check? Surely size 0 is also too small?

include/curl/multi.h
```diff
@@ -464,6 +464,20 @@ typedef int (*curl_push_callback)(CURL *parent,
                                   struct curl_pushheaders *headers,
                                   void *userp);
 
+/*
+ * Name:    curl_multi_pollfds()
```
bagder, line 468: Since this function populates `curl_waitfd` structs, should it perhaps rather be called `curl_multi_waitfds` ? I mean, the data it returns is not carved out to be used by `poll()` only but can be used by "anything".

lib/multi.c
```diff
@@ -1116,6 +1116,68 @@ CURLMcode curl_multi_fdset(struct Curl_multi *multi,
...
+  memset(&ps, 0, sizeof(ps));
+  for(data = multi->easyp; data; data = data->next) {
+    multi_getsock(data, &ps);
+
+    for(i = 0; i < ps.num; i++) {
+      if(nfds < size) {
+        curl_socket_t fd = ps.sockets[i];
+        int fd_idx = -1;
+
+        /* Simple linear search to skip addition of already added descriptor */
```
bagder, line 1147:
```suggestion
        /* Simple linear search to skip addition of an already added descriptor */
```

bagder (commented):
We also need a man page for the new function.

lib/multi.c
dkarpov1970: Sorry, I was off for a while.
This check allows just to get a number of descriptors when the passed 'size' is 0 and not return an error when the size is too small (which for size 0 will happen on the first descriptor).

In other words, it provides client an option to pass the `size 0` to get the number of "wait" descriptors, allocate necessary storage for them, and get them into that storage in the next call. This helps to avoid memory reallocations if they are critical.

include/curl/multi.h
dkarpov1970: Yes, I agree that the `curl_multi_waitfds` name indeed better represents what this function does.

lib/multi.c
dkarpov1970: Sure, I will try to create a man page for this function using the previous man pages (i.e. curl_multi_fdset.3), but I may need some help as I am not familiar with the man page format.

dkarpov1970: @bagder Could you please look at the man page and tell me what is wrong, so I can fix it?

I don't understand why I get the spelling error:
Misspelled words:
<htmlcontent> docs/libcurl/curl_multi_waitfds.md: html>body>p
--------------------------------------------------------------------------------
waitfds
--------------------------------------------------------------------------------

I looked at the other pages, and I don't see why I am getting this error.

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+.\" * SPDX-License-Identifier: curl
+.\" *
+.\" **************************************************************************
+.TH curl_multi_waitfds 3 "8 Jan 2024" "libcurl" "libcurl"
+.SH NAME
+curl_multi_waitfds - extracts file descriptor information from a multi handle
+.SH SYNOPSIS
+.nf
+#include <curl/curl.h>
+#include <stdlib.h>
```
bagder, line 30: `<stdlib.h>` is not really needed here, is it?

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+.SH DESCRIPTION
+This function extracts \fIcurl_waitfd\fP structures which are similar to
+\fIpoll(2)\fP's \fIpollfd\fP structure from a given multi_handle.
+
+These structures can be used for polling on multi_handle file descriptors in a
+fashion similar to \fIcurl_multi_poll(3)\fP. The \fIcurl_multi_perform(3)\fP
+function should be called as soon as one of them is ready to be read from or
+written to.
+
+libcurl fills provided \fIufds\fP array up to the \fIsize\fP.
```
bagder, line 46:
```suggestion
libcurl populates the provided \fIufds\fP array up to the \fIsize\fP.
```

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+                             unsigned int size,
+                             unsigned int *fd_count);
+.fi
+.SH DESCRIPTION
+This function extracts \fIcurl_waitfd\fP structures which are similar to
+\fIpoll(2)\fP's \fIpollfd\fP structure from a given multi_handle.
+
+These structures can be used for polling on multi_handle file descriptors in a
+fashion similar to \fIcurl_multi_poll(3)\fP. The \fIcurl_multi_perform(3)\fP
+function should be called as soon as one of them is ready to be read from or
```
bagder, line 43:
```suggestion
function should be called as soon as one of the file descriptors is ready to be read from or
```

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+written to.
+
+libcurl fills provided \fIufds\fP array up to the \fIsize\fP.
+If a number of descriptors used by the multi_handle is greater than the
```
bagder, line 47:
```suggestion
If the number of descriptors used by the multi_handle is greater than the
```

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+\fIsize\fP parameter then libcurl will fill the \fIufds\fP up to the \fIsize\fP
```
bagder, line 48:
```suggestion
\fIsize\fP parameter, libcurl fills the \fIufds\fP array up to the \fIsize\fP
```

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+and will return CURLM_OUT_OF_MEMORY error.
```
bagder, line 49:
```suggestion
and returns CURLM_OUT_OF_MEMORY.
```

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+
+If the \fIfd_count\fP argument is not a null pointer, it points to a variable
+that on returns specifies the number of descriptors used by the multi_handle to
+be checked for being ready to read or write.
+
+The client code can pass \fIsize\fP equal to zero just to get the number of the
+descriptors and allocate appropriate storage for them to be used in a
+subsequent function call.
+.SH EXAMPLE
+.nf
```
bagder, line 59:
~~~c
#include <stdlib.h>

docs/libcurl/curl_multi_waitfds.3
```diff
@@ -0,0 +1,114 @@
...
+    if(mc != CURLM_OK) {
+      fprintf(stderr, "curl_multi_waitfds() failed, code %d.\\n", mc);
+      free(ufds);
+      break;
+    }
+
+    /* Do polling on descriptors in ufds */
+
+    free(ufds);
+  } while (!mc);
```
bagder, line 101:
```suggestion
  } while(!mc);
```

lib/multi.c
bagder: Try this:
~~~diff
diff --git a/.github/scripts/cleanspell.pl b/.github/scripts/cleanspell.pl
index 329a97240..d8475a819 100755
--- a/.github/scripts/cleanspell.pl
+++ b/.github/scripts/cleanspell.pl
@@ -63,11 +63,11 @@ while(<F>) {
         $_ =~ s/CURLUPART_[A-Z0-9_]*//g;
         $_ =~ s/CURLUSESSL_[A-Z0-9_]*//g;
         $_ =~ s/curl_global_(init_mem|sslset|cleanup)//g;
         $_ =~ s/curl_(strequal|strnequal|formadd|waitfd|formget|getdate|formfree)//g;
         $_ =~ s/curl_easy_(nextheader|duphandle)//g;
-        $_ =~ s/curl_multi_fdset//g;
+        $_ =~ s/curl_multi_[a-z]*//g;
         $_ =~ s/curl_mime_(subparts|addpart|filedata|data_cb)//g;
         $_ =~ s/curl_ws_(send|recv|meta)//g;
         $_ =~ s/curl_url_(dup)//g;
         $_ =~ s/curl_pushheader_by(name|num)//g;
         $_ =~ s/libcurl-(env|ws)//g;

dkarpov1970: @bagder Thanks! I think I resolved all the issue, except one, which is related to documentation.
The failing check log says:
```
...
 curl_multi_strerror.3
 curl_multi_timeout.3
 curl_multi_wait.3
-curl_multi_waitfds.3
 curl_multi_wakeup.3
 curl_pushheader_byname.3
 curl_pushheader_bynum.3
ERROR: installed libcurl docs mismatch

```
So, I think I need to modify one more script and add `curl_multi_waitfds` somewhere, right?

dkarpov1970: @bagder As I can see these changes didn't get into 8.6.0,  I am wondering if it is possible to  get them into the next release?

bagder:
@dkarpov1970 can you take a look at the conflict and I think we should be ready for merge...

bagder:
Also, the new man page needs to be added to the list in `docs/libcurl/Makefile.inc`

dkarpov1970:
@bagder I think all conflicts have been resolved. But some checks are still failing.
Not sure, if they are false negative or not.

The "Cirrus CI / FreeBSD 14.0" check fails with error:
"make[5]: make[5]: don't know how to make curl_multi_waitfds.3. Stop"

which is strange because, as I understand, the `docs/libcurl/Makefile.inc` should take care of building man pages and no other build fails with that error. Could you please have a look and let me know what else I should correct?

bagder:
> **This branch cannot be rebased due to conflicts**

Please rebase against master and force-push.

dkarpov1970:
Closed this one as it became too messy.
Created a new one after rebasing with the latest master: https://github.com/curl/curl/pull/12976

PR #12588, opened 2023-12-23
Documentation: Fix incorrect C string escape in CURLOPT_POSTFIELDS

haydaralaidrus:
The string literal incorrectly formatted. The backslashes in the JSON string not properly escaped, leading to invalid C syntax:

```c
const char *json = "{"name": "daniel"}";
```

Where it should be:

```c
const char *json = "{\"name\": \"daniel\"}";
```


jay:
Thanks. I landed it with two backslashes as the escape because one backslash escapes the other and it shows as a single backslash.

haydaralaidrus:
@jay Got it, thank you.

PR #12589, opened 2023-12-23
verify-examples.pl: fail verification on unescaped backslash

jay:
- Check that all backslashes in EXAMPLE are properly escaped.

eg manpage must always use `\\n` never `\n`.

This is because the manpage requires we always double blackslash to show a single backslash. Prior to this change an erroneous single backslash would pass through and compile even though it would not show correctly in the manpage.

Ref: https://github.com/curl/curl/pull/12588

Closes #xxxx

---

Note I also added `die` for failed `open` of input/output.

### DIFF
--- a/.github/scripts/verify-examples.pl
+++ b/.github/scripts/verify-examples.pl
@@ -69,7 +69,7 @@ sub extract {
             if(/^.fi/) {
                 last;
             }
-            if(/\\[^\\]/) {
+            if(/(?<!\\)(?:\\{2})*\\(?!\\)/) {
                 print STDERR
                   "Error while processing file $f line $iline:\n$_" .
                   "Error: Single backslashes \\ are not properly shown in " .
--- a/.github/scripts/verify-examples.pl
+++ b/.github/scripts/verify-examples.pl
@@ -26,6 +26,7 @@
 my @files = @ARGV;
 my $cfile = "test.c";
 my $check = "./scripts/checksrc.pl";
+my $error;
 
 if($files[0] eq "-h") {
     print "Usage: verify-synopsis [man pages]\n";
@@ -75,6 +76,7 @@ sub extract {
                   "Error: Single backslashes \\ are not properly shown in " .
                   "manpage EXAMPLE output unless they are escaped \\\\.\n";
                 $fail = 1;
+                $error = 1;
                 last;
             }
             # two backslashes become one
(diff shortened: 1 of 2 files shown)


bagder (approved):

jay:
For some reason the CI job `manpage examples / verify (pull_request)` -- which runs verify-examples.pl to verify the manpage examples -- isn't running on the commits in this PR.

bagder:
Does it need this?
~~~diff
diff --git a/.github/workflows/man-examples.yml b/.github/workflows/man-examples.yml
index 080cb4c2d..640e3716c 100644
--- a/.github/workflows/man-examples.yml
+++ b/.github/workflows/man-examples.yml
@@ -10,16 +10,18 @@ on:
       - master
       - '*/ci'
     paths:
       - 'docs/libcurl/curl_*.3'
       - 'docs/libcurl/opts/*.3'
+      - '.github/scripts/verify-examples.pl'
   pull_request:
     branches:
       - master
     paths:
       - 'docs/libcurl/curl_*.3'
       - 'docs/libcurl/opts/*.3'
+      - '.github/scripts/verify-examples.pl'
 
 jobs:
   verify:
     runs-on: ubuntu-latest
     steps:

jay:
> Does it need this?

oh maybe. let's find out...

.github/scripts/verify-examples.pl
```diff
@@ -88,5 +97,8 @@ sub extract {
       $error |= testcompile($m);
       $error |= checksrc($m);
     }
+    else {
+      $error = 1;
```
bagder, line 101: This will cause an error for the cases where the script does not extract any example, which does not have to be the same as examples lacking double-backslash... and if so, this is done with no message output so we can't tell which file this is. I suspect maybe the template?

bagder:
How about this take?
~~~diff
diff --git a/.github/scripts/verify-examples.pl b/.github/scripts/verify-examples.pl
index 8c8fed11b..4b4ff5810 100755
--- a/.github/scripts/verify-examples.pl
+++ b/.github/scripts/verify-examples.pl
@@ -73,10 +73,11 @@ sub extract {
                 print STDERR
                   "Error while processing file $f line $iline:\n$_" .
                   "Error: Single backslashes \\ are not properly shown in " .
                   "manpage EXAMPLE output unless they are escaped \\\\.\n";
                 $fail = 1;
+                $error++;
                 last;
             }
             # two backslashes become one
             $_ =~ s/\\\\/\\/g;
             print O $_;
@@ -95,10 +96,7 @@ for my $m (@files) {
     my $out = extract($m);
     if($out) {
       $error |= testcompile($m);
       $error |= checksrc($m);
     }
-    else {
-      $error = 1;
-    }
 }
 exit $error;

Issue #622, opened 2016-01-30
curl returns correct LDAP result but after that hangs


alliquot:
Hello,

I faced a strange issue while querying LDAP Active Directory via curl: directly and via R-language wrapper - RCurl.

When I run RCurl under R session on Windows, my query is evaluated and result is returned immediately. Everything is fine.

When I do the same under R on Linux Debian, it hangs. So I've run the query via curl from the Linux command line. It turned out that the result is returned correctly (all information is correct), but curl is waiting for something. Only Ctrl+C can stop that.

My query:

     LDAP://x.x.x.x/dc=company,dc=com?userPrincipalName,msExchShadowGivenName,msExchShadowSn,sAMAccountName?sub?(&(objectCategory=person)(msExchShadowProxyAddresses=*some@email.com*))

My call:

    curl -u user@domain:pwd "ldap://ip/......"

Why the query on Windows is evaluated successfully and curl returns results immediately, while the same query on Linux makes curl hanged? Should I set some additional options?

This may be important: Debian is running under Hyper-V virtual machine.
The version of curl I'm using is: 7.38.0

Thanks in advance.

EDIT: 
adding --max-time 1 partially solves this issue, but this completely unprofessional solution. If the response will come in 2 seconds, it will be lost. And setting the max-time to 2 seconds will make the process waiting *always* 2 seconds.

monnerat:
I think you've hit the same problem discussed 10 days ago on the mailing list. Please look at thread http://curl.haxx.se/mail/lib-2016-01/0101.html.
This is probably the case if:
- The ldap backend used by your curl installation is openldap (unless the generic ldap backend also has the same problem: not tested),
- Queries are made to a server that returns referrals (probably an M$AD server?),
- External ldap configuration allows automatic referrals chasing.

There is no easy way to support automatic referrals chasing in curl yet.
As a workaround, I can suggest you disable this feature by ldap configuration ("REFERRALS off" in ldap.conf): this will release the hang. Of course, this will also miss the results from referrals, but in the current curl's state, they won't be returned anyway.
I started to investigate to fix this problem, but this is a huge work, so don't expect to have a real fix soon...

alliquot:
[...]

PR #12591, opened 2023-12-24
Fix LDAP hang by reading more

OzanCansel:
This patch fixes known bug `12.4`.

#### Bug
The problem is that libcurl only reads one `ldap` message at a time. So `LDAP_RES_SEARCH_RESULT` msg is never read and hangs until the timeout occurs.

#### Fix
Changed the code to read until there are no messages left or an error occurred in a non-blocking way.

### Loop Condition
`oldap_recv_single` only parses one message at a time even though there might be multiple `ldap` messages in the read buffer. So `oldap_recv_single` parses a single message and outputs the `CURLE_AGAIN` over `err` variable if a message is parsed and no error occurred but its type is not `LDAP_RES_SEARCH_RESULT`.

`LDAP_RES_SEARCH_RESULT` is the query end indicator. In a nutshell, `CURLE_AGAIN` constitutes that there might be remaining messages because query end indicator is not reached yet. So loop continues if a message is read and no error occurred.

`oldap_recv` return code is `CURLE_AGAIN` if at least one message is parsed and no error occurred. Otherwise the actual error is returned.

#### Wireshark trace
[ldap-wireshark.zip](https://github.com/curl/curl/files/13764277/ldap-wireshark.zip)

#### Before fix
https://github.com/curl/curl/assets/35337781/3f80644e-b745-44a2-881e-d77fbdd8468c

#### After fix
https://github.com/curl/curl/assets/35337781/8090e32a-ea71-40fa-ab20-d730067a7d50

### DIFF
--- a/lib/openldap.c
+++ b/lib/openldap.c
@@ -963,11 +963,10 @@ static CURLcode client_write(struct Curl_easy *data,
   return result;
 }
 
-static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
-                          size_t len, CURLcode *err)
+static int oldap_recv_single(struct Curl_easy *data, int sockindex, char *buf,
+                             size_t len, CURLcode *err)
 {
-  struct connectdata *conn = data->conn;
-  struct ldapconninfo *li = conn->proto.ldapc;
+  struct ldapconninfo *li = data->conn->proto.ldapc;
   struct ldapreqinfo *lr = data->req.p.ldap;
   int rc;
   LDAPMessage *msg = NULL;
@@ -975,7 +974,6 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
   struct timeval tv = {0, 0};
   struct berval bv, *bvals;
   int binary = 0;
-  CURLcode result = CURLE_AGAIN;
   int code;
   char *info = NULL;
 
@@ -986,16 +984,17 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
   rc = ldap_result(li->ld, lr->msgid, LDAP_MSG_ONE, &tv, &msg);
   if(rc < 0) {
     failf(data, "LDAP local: search ldap_result %s", ldap_err2string(rc));
-    result = CURLE_RECV_ERROR;
+    *err = CURLE_RECV_ERROR;
+
+    return 0;
   }
 
-  *err = result;
-
   /* error or timed out */
-  if(!msg)
-    return -1;
+  if(!msg) {
+    *err = CURLE_AGAIN;
 
-  result = CURLE_OK;
+    return 0;
+  }
 
   switch(ldap_msgtype(msg)) {
   case LDAP_RES_SEARCH_RESULT:
@@ -1004,9 +1003,12 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
     if(rc) {
       failf(data, "LDAP local: search ldap_parse_result %s",
             ldap_err2string(rc));
-      result = CURLE_LDAP_SEARCH_FAILED;
+      *err = CURLE_LDAP_SEARCH_FAILED;
       break;
     }
+
+    /* LDAP_RES_SEARCH_RESULT indicates the end of the query. */
+    *err = CURLE_OK;
 
     /* store the latest code for later retrieval */
     data->info.httpcode = code;
@@ -1021,9 +1023,10 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
     default:
       failf(data, "LDAP remote: search failed %s %s", ldap_err2string(code),
             info ? info : "");
-      result = CURLE_LDAP_SEARCH_FAILED;
+      *err = CURLE_LDAP_SEARCH_FAILED;
       break;
     }
+
     if(info)
       ldap_memfree(info);
     break;
@@ -1031,13 +1034,13 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
     lr->nument++;
     rc = ldap_get_dn_ber(li->ld, msg, &ber, &bv);
     if(rc < 0) {
-      result = CURLE_RECV_ERROR;
+      *err = CURLE_RECV_ERROR;
       break;
     }
 
-    result = client_write(data, STRCONST("DN: "), bv.bv_val, bv.bv_len,
-                          STRCONST("\n"));
-    if(result)
+    *err = client_write(data, STRCONST("DN: "), bv.bv_val, bv.bv_len,
+                        STRCONST("\n"));
+    if(*err)
       break;
 
     for(rc = ldap_get_attribute_ber(li->ld, msg, ber, &bv, &bvals);
@@ -1049,10 +1052,11 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
         break;
 
       if(!bvals) {
-        result = client_write(data, STRCONST("\t"), bv.bv_val, bv.bv_len,
-                              STRCONST(":\n"));
-        if(result)
+        *err = client_write(data, STRCONST("\t"), bv.bv_val, bv.bv_len,
+                            STRCONST(":\n"));
+        if(*err)
           break;
+
         continue;
       }
 
@@ -1062,9 +1066,9 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
       for(i = 0; bvals[i].bv_val != NULL; i++) {
         int binval = 0;
 
-        result = client_write(data, STRCONST("\t"), bv.bv_val, bv.bv_len,
-                              STRCONST(":"));
-        if(result)
+        *err = client_write(data, STRCONST("\t"), bv.bv_val, bv.bv_len,
+                            STRCONST(":"));
+        if(*err)
           break;
 
         if(!binary) {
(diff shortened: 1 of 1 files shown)


OzanCansel:
@monnerat Could you review my PR? I think you are experienced with this part.

lib/openldap.c
```diff
@@ -1088,41 +1092,71 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
...
+
+  read_count = 0;
+
+  /* There might be multiple messages so read all the available messages. */
+  do {
+    msg_read = oldap_recv_single(data, sockindex, buf, len, err);
+
+    if(msg_read)
+      read_count++;
+  } while(*err == CURLE_AGAIN && msg_read);
```
bagder, line 1149: Can you explain the logic here? Why does it loop because of this?

lib/openldap.c
```diff
@@ -1088,41 +1092,71 @@ static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
...
+}
+
+static ssize_t oldap_recv(struct Curl_easy *data, int sockindex, char *buf,
+                          size_t len, CURLcode *err)
+{
+  int read_count;
+  int msg_read;
+  (void)len;
+  (void)buf;
+  (void)sockindex;
```
bagder, line 1139: Why (void) these when they are used below?

monnerat:
> The problem is that libcurl only reads one ldap message at a time. So `LDAP_RES_SEARCH_RESULT` msg is never read and hangs until the timeout occurs.

I can't see why: as long as an `LDAP_RES_SEARCH_ENTRY` is seen, oldap_recv() retuns with `CURLE_AGAIN` (or an error). It is therefore recalled later after other handles get a chance to do something, and this is why you see all entries before it hangs. It should always see `LDAP_RES_SEARCH_RESULT` as well and this never returns `CURLE_AGAIN`.Other types of messages also terminates the transfer. We probably missed something in the protocol exchange that tricks us. Can you produce a wireshark trace? This would be very useful.

I neither can't see a notable protocol handling difference between your code and the original. The problem is the former tries to read all messages before a timeout or data end occurs without letting a chance to handles scheduling: this is what I wanted to avoid by reading one message at a time :-(

Unlike the known bug 12.1 targets, your fix does not deal with referrals: did you try https://github.com/curl/curl/issues/622#issuecomment-308722961 with the unpatched version ? Does it work for you ?

lib/openldap.c
OzanCansel: Fixed.

lib/openldap.c
OzanCansel: Added to the description since it is related with the logic.

OzanCansel:
> > The problem is that libcurl only reads one ldap message at a time. So `LDAP_RES_SEARCH_RESULT` msg is never read and hangs until the timeout occurs.
> 
> I can't see why: as long as an `LDAP_RES_SEARCH_ENTRY` is seen, oldap_recv() retuns with `CURLE_AGAIN` (or an error). It is therefore recalled later after other handles get a chance to do something, and this is why you see all entries before it hangs. It should always see `LDAP_RES_SEARCH_RESULT` as well and this never returns `CURLE_AGAIN`.Other types of messages also terminates the transfer. We probably missed something in the protocol exchange that tricks us. Can you produce a wireshark trace? This would be very useful.
> 
> I neither can't see a notable protocol handling difference between your code and the original. The problem is the former tries to read all messages before a timeout or data end occurs without letting a chance to handles scheduling: this is what I wanted to avoid by reading one message at a time :-(
> 
> Unlike the known bug 12.1 targets, your fix does not deal with referrals: did you try [#622 (comment)](https://github.com/curl/curl/issues/622#issuecomment-308722961) with the unpatched version ? Does it work for you ?

@monnerat
I can't reply in place so I drop a comment as respond to you :) 

- `readwrite_data` is not recalled again because poll tells that there are no new bytes available. It is exactly at [this line](https://github.com/curl/curl/blob/master/lib/transfer.c#L1097). `select_bits` is zero so condition fails. Therefore libcurl doesn't recall the `readwrite_data` which calls `oldap_recv`.

- Only one message is parsed (all entries are parsed in single message, at least, it is what it is for my case), `oldap_recv` is called only once and hangs. If more bytes will come it will continue but this parsing mechanism doesn't guarantee that all messages are parsed.

- I wrote `12.1` but it actually fixes `12.4` so treat like a typo.

Btw, if my claim is correct it will partially fix all the known bugs related with `ldap`.

OzanCansel:
@bagder `quiche / quiche (pull_request)` job fails but i don't think that it is related with my change. Is it caused by something else or should I fix it?

monnerat:
> readwrite_data is not recalled again because poll tells that there are no new bytes available. It is exactly at [this line](https://github.com/curl/curl/blob/master/lib/transfer.c#L1097). select_bits is zero so condition fails. Therefore libcurl doesn't recall the readwrite_data which calls oldap_recv.

Mmm, I see. Many thanks for the explanation.
So I think it's a `data_pending()` problem: ldap is  one of our few protocols that directly deals with a file descriptor and can buffer data into external library structures, like SCP and SFTP. `data_pending()` has been introduced with the connection filters architecture, so versions previous to this should not have this bug (@icing ?).
There definitively should be a protocol-level method to implement data_pending(). In the meantime, I rather would see a fix on https://github.com/curl/curl/blob/master/lib/transfer.c#L380 as something like
```
  return conn->handler->protocol&(CURLPROTO_SCP|CURLPROTO_SFTP|CURLPROTO_LDAP|CURLPROTO_LDAPS) ||
    Curl_conn_data_pending(data, FIRSTSOCKET);
```
Can you try it (I do not have an ActiveDirectory server to test with)? Thanks.

>I wrote 12.1 but it actually fixes 12.4 so treat it like a typo.

Thanks for the precision. 12.1 should have been fixed by https://github.com/curl/curl/commit/cdc1da912066535680f02eb31e6e7c1015bc92ff. At least it should not block anymore, but I don't think referrals are returned within LDIF data.

OzanCansel:
> > readwrite_data is not recalled again because poll tells that there are no new bytes available. It is exactly at [this line](https://github.com/curl/curl/blob/master/lib/transfer.c#L1097). select_bits is zero so condition fails. Therefore libcurl doesn't recall the readwrite_data which calls oldap_recv.
> 
> Mmm, I see. Many thanks for the explanation. So I think it's a `data_pending()` problem: ldap is one of our few protocols that directly deals with a file descriptor and can buffer data into external library structures, like SCP and SFTP. `data_pending()` has been introduced with the connection filters architecture, so versions previous to this should not have this bug (@icing ?). There definitively should be a protocol-level method to implement data_pending(). In the meantime, I rather would see a fix on https://github.com/curl/curl/blob/master/lib/transfer.c#L380 as something like
> 
> ```
>   return conn->handler->protocol&(CURLPROTO_SCP|CURLPROTO_SFTP|CURLPROTO_LDAP|CURLPROTO_LDAPS) ||
>     Curl_conn_data_pending(data, FIRSTSOCKET);
> ```
> 
> Can you try it (I do not have an ActiveDirectory server to test with)? Thanks.
> 
> > I wrote 12.1 but it actually fixes 12.4 so treat it like a typo.
> 
> Thanks for the precision. 12.1 should have been fixed by [cdc1da9](https://github.com/curl/curl/commit/cdc1da912066535680f02eb31e6e7c1015bc92ff). At least it should not block anymore, but I don't think referrals are returned within LDIF data.

I tried the fix but it didn't solve. The change you proposed is made in `data_pending` but the problem is that `readwrite_data` doesn't get called so the fix is not effective.

bagder:
> the problem is that `readwrite_data` doesn't get called

Not at all? That seems... wrong.

Isn't the LDAP transfer handled in the PERFORMING state?

OzanCansel:
> > the problem is that `readwrite_data` doesn't get called
> 
> Not at all? That seems... wrong.
> 
> Isn't the LDAP transfer handled in the PERFORMING state?

Let me explain the other way around. Assume that ldap server has sent 100 bytes as a response. Assume that, those 100 bytes corresponds to 5 messages. If this scenario occurs(probably %99 of cases), what current `openldap.c` does is it just parses the 1 message and expects `libcurl` to call `readwrite_data` which calls `oldap_recv` again but it doesn't happen because all bytes are read and nothing new. Therefore all transfers have been made but only 1 message is parsed out of 5 messages and hangs. My fix guarantees that all available messages are parsed.

bagder:
> it doesn't happen because all bytes are read and nothing new

That's the problem fixing `data_pending()` should address. So the question I then have is why fixing data_sending is not good enough?

OzanCansel:
> > it doesn't happen because all bytes are read and nothing new
> 
> That's the problem fixing `data_pending()` should address. So the question I then have is why fixing data_sending is not good enough?

``` patch
diff --git a/lib/transfer.c b/lib/transfer.c
index 59a3f05cb..9c71225db 100644
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -377,7 +377,7 @@ static int data_pending(struct Curl_easy *data)
 
   /* in the case of libssh2, we can never be really sure that we have emptied
      its internal buffers so we MUST always try until we get EAGAIN back */
-  return conn->handler->protocol&(CURLPROTO_SCP|CURLPROTO_SFTP) ||
+  return conn->handler->protocol&(CURLPROTO_SCP|CURLPROTO_SFTP|CURLPROTO_LDAP|CURLPROTO_LDAPS) ||
     Curl_conn_data_pending(data, FIRSTSOCKET);
 }
 
```

I made the change above but it doesn't solve the problem. Because `Curl_read` returns `CURLE_AGAIN` [here](https://github.com/curl/curl/blob/master/lib/transfer.c#L479), so it doesn't even call `data_pending`. At next attempt, `readwrite_data` is not called because `select_bits` is `0`, see [here](https://github.com/curl/curl/blob/master/lib/transfer.c#L1098). idk mb im lack of knowledge about curl internals but this is what I observed.

bagder:
> I made the change above but it didn't solve the problem. Because Curl_read returns CURLE_AGAIN [here](https://github.com/curl/curl/blob/master/lib/transfer.c#L479), so it doesn't even call data_pending

We're going in circles.

If the LDAP transfer is being done in the PERFORMING state then I cannot understand how what you say can happen, which leads me to suspect that the transfer is done in the wrong state, which is why I asked about this. Then it seems worth to address *that* problem first so that data_pending can get a chance to do its job.

Unfortunately we don't have LDAP test cases. :disappointed:

OzanCansel:
@bagder 
I tested `ldaps` query with apache studio server. It doesn't hang with it. So more likely there is a difference ldaps servers. I observed hang problem with `Windows Active Directory`.

bagder:
@OzanCansel can you build curl debug-enabled, run it against this server with `-v` and show us the verbose output?

bagder:
This PR will not go further until we get more more information and make sure we get things done the right way.

bagder:
Stalled

PR #12596, opened 2023-12-26
add `--secure` option and `CURL_SECURE` env

vszakats:
The idea of these new options is to undo the effect of `--insecure` / `-k`,
`--doh-insecure`, `--proxy-insecure`, and `CURLOPT_SSL_VERIFYPEER=0`,
`CURLOPT_SSL_VERIFYHOST=0` in libcurl. Thus forcing security for curl
connections when apps and scripts would try to enable insecure mode.

Closes #12596


dfandrich:
What if the user really actually wants an insecure connection? Do we need then
an --actually-insecure option to negate --secure? But what if
--actually-insecure is given but the user really, truly wants a secure
connection? How many levels is enough?

jay:
I'm against this, I think we've discussed this before and basically my take is we should not offer a way to override the intent. If someone uses insecure then that's what they want.

vszakats:
The problem this is attempting to solve is that apps and scripts are
prone to generously apply `--insecure` in cases. (according to a quick
GitHub search: in _many_ cases). There is at the moment no way to
rectify this as an end-user, short of auditing and rebuilding all 3rd-party
code involved (which is not a feasible solution in practice). This option
gives back some control over this to the end-user who wants to use 
TLS as designed: with proper authentication of the other end.

jay:
I don't think that's our problem or responsibility. If a command has `--insecure` then the author's intention was certainly to skip verification. What about --tlsv1.0 or any other number of options that we or the user might think are bad form but may be needed to make the transfer work? For example the user has --tlsv1.2 in their curlrc and a command line has --tlsv1.0. curlrc is not policy as the behavior can be changed by the command.

bagder:
This is pretty much exactly what I have proposed in the past (I failed to find the proposal just now) . It did not find many supporters back then.

vszakats:
For sure this isn't the single security issue that might happen while passing data over the internet, but I'm less worried about `--tlsv1.0` specifically. This option is much less often used in the wild and majority of servers no longer support TLSv1.0, with a trend that its support is steadily shrinking.

The insecure option however is widely used and often mis-used. It is breaking TLS (including every modern and accurately configured servers). It also causes that apps fall back using it for convencience, instead of e.g. offering a proper way to specify a custom certificate, or fixing / using a certificiate bundle.

A quick search revealed 1000 hits for `--tlsv1.0` and 100000 for `--insecure` + `curl`, and that's even without `-k` and the libcurl options.

This is a many-year old issue, and even back then there were hundreds of thousands of hits on GitHub for these keywords.
I agree it should be fixed in the apps, but apparently many times this doesn't happen.

Another variant of the `--secure` option would be one reporting such issues while letting the app continue with broken TLS. This would help with the audit part.

PR #12599, opened 2023-12-27
vtls: fix missing multissl version info

jay:
- Fix erroneous buffer copy logic from ff74cef5.

Prior to this change the MultiSSL version info returned to the user was empty.

Closes #xxxx

---
before: `libcurl/8.6.0-DEV  nghttp2/1.52.0`
after: `libcurl/8.6.0-DEV OpenSSL/3.0.8 (Schannel) nghttp2/1.52.0`

### DIFF
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -1413,10 +1413,12 @@ static size_t multissl_version(char *buffer, size_t size)
     backends_len = p - backends;
   }
 
-  if(size && (size < backends_len))
-    strcpy(buffer, backends);
-  else
-    *buffer = 0; /* did not fit */
+  if(size) {
+    if(backends_len < size)
+      strcpy(buffer, backends);
+    else
+      *buffer = 0; /* did not fit */
+  }
   return 0;
 }
 


jay:
If truncation is preferred for inadequate buffers then another way to do this would be revert that file:

~~~diff
diff --git a/lib/vtls/vtls.c b/lib/vtls/vtls.c
index 256b8fa..dd68a06 100644
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -1413,11 +1413,17 @@ static size_t multissl_version(char *buffer, size_t size)
     backends_len = p - backends;
   }
 
-  if(size && (size < backends_len))
-    strcpy(buffer, backends);
-  else
-    *buffer = 0; /* did not fit */
-  return 0;
+  if(!size)
+    return 0;
+
+  if(size <= backends_len) {
+    strncpy(buffer, backends, size - 1);
+    buffer[size - 1] = '\0';
+    return size - 1;
+  }
+
+  strcpy(buffer, backends);
+  return backends_len;
 }
 
 static int multissl_setup(const struct Curl_ssl *backend)
~~~

bagder (approved):

PR #12602, opened 2023-12-28
system_win32: fix a function pointer assignment warning

jay:
- Use CURLX_FUNCTION_CAST to suppress a function pointer assignment warning.

a6bbc87f added lookups of some Windows API functions and then cast them like `*(FARPROC*)&Curl_funcname = address`. Some versions of gcc warn about that as breaking strict-aliasing rules so this PR changes those assignments to use CURLX_FUNCTION_CAST.

Bug: https://github.com/curl/curl/pull/12581#issuecomment-1869804317
Reported-by: Marcel Raad

Closes #xxxx

### DIFF
--- a/lib/system_win32.c
+++ b/lib/system_win32.c
@@ -43,21 +43,17 @@ bool Curl_isWindows8OrGreater;
 /* Handle of iphlpapp.dll */
 static HMODULE s_hIpHlpApiDll = NULL;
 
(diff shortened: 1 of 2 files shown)


MarcelRaad (approved):
Thanks!

vszakats (approved):

PR #12606, opened 2023-12-28
Enable IDN on Windows when building with clang

mraunak:
This PR aims to enable IDN on the Windows platform with CLANG compiler version 17 which is required to successfully 
build TensorFlow-CPU on the Windows
Changed https://github.com/curl/curl/blob/master/lib/idn.c to use an inbuilt Windows function IdnToAscii on the Clang Compiler

Below is the screenshot of the error:

image

### DIFF
--- a/lib/idn.c
+++ b/lib/idn.c
@@ -53,7 +53,7 @@
 #ifdef USE_WIN32_IDN
 /* using Windows kernel32 and normaliz libraries. */
 
-#if !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x600
+#if defined(__clang__) || !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x600
 WINBASEAPI int WINAPI IdnToAscii(DWORD dwFlags,
                                  const WCHAR *lpUnicodeCharStr,
                                  int cchUnicodeChar,
--- a/lib/idn.c
+++ b/lib/idn.c
@@ -53,7 +53,7 @@
 #ifdef USE_WIN32_IDN
 /* using Windows kernel32 and normaliz libraries. */
 
-#if defined(__clang__) || !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x600
+#if !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x600 || WINVER < 0x600
 WINBASEAPI int WINAPI IdnToAscii(DWORD dwFlags,
                                  const WCHAR *lpUnicodeCharStr,
                                  int cchUnicodeChar,


gvanem:
So why isn't `-D_WIN32_WINNT >= 0x600` defined for Clang?

vszakats:
It doesn't look right to enable Windows SDK-specific code based on compiler vendor.
The official Windows binaries are built with clang 17 + IDN without this issues.

What needs to be seen is why this build env is missing these Windows declarations
or why it sets `USE_WIN32_IDN` without `_WIN32`, or why its Windows headers
are missing this declarations without a matching `_WIN32_WINNT`?

What build tool, build setup, curl version is this happening with?

bagder:
@mraunak: this PR awaits info from you...

mraunak:
Hi Team, Thank you for your input. 

Build tool: Bazel version 6.4.0

**Build config:** 
ERROR: D:/ockagzuo/external/curl/BUILD.bazel:31:11: Compiling lib/idn.c failed: (Exit 1): clang-cl.exe failed: error executing command (from target @curl//:curl) 
  cd /d D:/ockagzuo/execroot/org_tensorflow
  SET INCLUDE=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.38.33130\include;C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.38.33130\ATLMFC\include;C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\VS\include;C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\um;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\shared;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\winrt;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\cppwinrt;C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um;D:\user\mraunak\LLVM\lib\clang\17\include
    SET PATH=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.38.33130\bin\HostX64\x64;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\VC\VCPackages;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\TestWindow;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer;C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\bin\Roslyn;C:\Program Files\Microsoft Visual Studio\2022\Community\Team Tools\Performance Tools\x64;C:\Program Files\Microsoft Visual Studio\2022\Community\Team Tools\Performance Tools;C:\Program Files (x86)\Microsoft Visual Studio\Shared\Common\VSPerfCollectionTools\vs2019\\x64;C:\Program Files (x86)\Microsoft Visual Studio\Shared\Common\VSPerfCollectionTools\vs2019\;C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\x64\;C:\Program Files\Microsoft Visual Studio\2022\Community\Team Tools\DiagnosticsHub\Collector;C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\\x64;C:\Program Files (x86)\Windows Kits\10\bin\\x64;C:\Program Files\Microsoft Visual Studio\2022\Community\\MSBuild\Current\Bin\amd64;C:\Windows\Microsoft.NET\Framework64\v4.0.30319;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\;;C:\Windows\system32;C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja;C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\VC\Linux\bin\ConnectionManagerExe
    SET PWD=/proc/self/cwd
    SET PYTHON_BIN_PATH=D:/user/mraunak/venv310_new1/Scripts/python.exe
    SET PYTHON_LIB_PATH=D:/user/mraunak/venv310_new1/lib/site-packages
    SET TEMP=C:\Users\mraunak\AppData\Local\Temp\2
    SET TF2_BEHAVIOR=1
    SET TMP=C:\Users\mraunak\AppData\Local\Temp\2
    SET VSLANG=1033
  D:\user\mraunak\LLVM\bin\clang-cl.exe @bazel-out/x64_windows-opt/bin/external/curl/_objs/curl/idn.obj.params
# Configuration: d9c220ac8112bcb55e4eecee5f7704306287d5558f345559f8f07e36e16a82c0
# Execution platform: //tensorflow/tools/toolchains/win:x64_windows-clang-cl


**Curl version**: curl 8.4.0 (Windows) libcurl/8.4.0 Schannel WinIDN
Release-Date: 2023-10-11

**Clang version**: 17.0.6

dfandrich:
Could you show the actual compile error? That's not included in that snippet.

mraunak:
[clang-17.txt](https://github.com/curl/curl/files/13867263/clang-17.txt)

Hi @dfandrich, please find the attached log with the error. Please let me know if any further information is needed from my end

dfandrich:
The underlying errors seem to be the same as in your initial screenshot:
```
error: call to undeclared function 'IdnToUnicode';
error: call to undeclared function 'IdnToAscii
```
So, the question from vszakats still stands: why this build env is missing these Windows declarations?

mraunak:
Hi Team, IdnToUnicode and IdnToASCII function is handled by WinNls.h
https://github.com/tpn/winsdk-10/blob/master/Include/10.0.10240.0/um/WinNls.h#L2445
https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170. we need to update As per the above documentation WINVER <0x600 was updated to fix the issue. Please let me know if it is an acceptable solution

vszakats:
If this is caused by mismatched `_WIN32_WINNT` and `WINVER` values setup by the build environment, I'd vote to fix this in the build configuration.

vszakats:
Is this correct though?
https://github.com/tensorflow/tensorflow/blob/83b3fbfa32de0ff95f52e86a1370d7e3156acac0/third_party/curl.BUILD#L17-L21

According to the comment it's trying to fix something in curl
(would be useful to know what), but with this particular setup,
which isn't SDK 7.1, but SDK 10 and not MSVC but clang-cl, it
might rather be breaking things.

The Tensorflow build hack triggers some logic in curl, which
messes with `WINVER`:
https://github.com/curl/curl/blob/24ae4a07f3496ddb1506a79e34b969cc2e026091/lib/config-win32.h#L320-L380

This logic in curl seems dangerous and possibly obsolete, but
since it's MSVC-specific, I haven't looked at it. Touching either
`_WIN32_WINNT` or `WINVER` by any project is generally 
best to avoid and these are up to the builders to set (or leave as is).
If there is a combination that is broken, I'd vote to put out an
`#error` rather than trying to fix it locally.

I haven't tried but this problem might be fixed by deleting either
the Tensorflow hack or the curl hack that it triggers. Or both.

The other logic in curl that touches `WINVER` is this:
https://github.com/curl/curl/blob/24ae4a07f3496ddb1506a79e34b969cc2e026091/configure.ac#L2471-L2488

Added as part of a commit fixing legacy-mingw issues in 37f1c21cb9c809ec870803fc40e1ed2afd9534ac.
The commit message doesn't mention IDN while this logic specifically
tries to address some issues with it. If this is legacy-mingw specific,
I think it'd be best to delete it, because we no longer support it.

mraunak:
Hi Team, removing the hack https://github.com/tensorflow/tensorflow/blob/83b3fbfa32de0ff95f52e86a1370d7e3156acac0/third_party/curl.BUILD#L17-L21
works.

jay:
> This logic in curl seems dangerous and possibly obsolete, but
> since it's MSVC-specific, I haven't looked at it. Touching either
> `_WIN32_WINNT` or `WINVER` by any project is generally
> best to avoid and these are up to the builders to set (or leave as is).
> If there is a combination that is broken, I'd vote to put out an
> `#error` rather than trying to fix it locally.

I will check this with some old Visual Studio versions. I think it was primarily meant as workarounds for old versions (which are still supported via winbuild and pregenerated projects). I don't know why it would be applied to new versions. Regardless that tensor hack is fishy.

> Added as part of a commit fixing legacy-mingw issues in [37f1c21](https://github.com/curl/curl/commit/37f1c21cb9c809ec870803fc40e1ed2afd9534ac).
> The commit message doesn't mention IDN while this logic specifically
> tries to address some issues with it. If this is legacy-mingw specific,
> I think it'd be best to delete it, because we no longer support legacy-mingw.

#12684

bagder:
@mraunak  I take it this PR is no longer actually necessary or wanted?

mraunak:
Hi @bagder, yes, we can close it. Thank you!

bagder:
Thanks!

PR #12607, opened 2023-12-29
fix implicit int compile error in ipv6 detection

a-n-n-a-l-e-e:
ipv6 auto detection fails on clang 16 with
```console
error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
```

fix the test stub to declare a return type and return a value.


bagder (approved):

bagder:
Thanks!

PR #12610, opened 2023-12-30
openldap: fix STARTTLS

monnerat:
It was not working anymore since introduction of connection filters.

Also do not attempt to recover from a failing TLS negotiation with CURLUSESSL_TRY.


icing (approved):

bagder:
Thanks!

PR #12617, opened 2024-01-02
tool: prepend output_dir in header callback 

### LINKED ISSUE #12614: --remote-time (-R) doesn't work with --output-dir
### I did this

 curl --create-dirs -JRO --output-dir "some/path/" "https://someurl.com/somefile/download"

### I expected the following

The file to be created at `some/path` with the remote timestamp set

### curl/libcurl version

curl 8.5.0 (x86_64-pc-cygwin) libcurl/8.5.0 OpenSSL/3.2.0 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.11.0 nghttp2/1.58.0
Release-Date: 2023-12-06
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

Windows. curl download from MSYS2's MinGW64 repository.

bagder:
When Content-Disposition parsing is used and an output dir is prepended, make sure to store that new file name correctly so that it can be used for setting the file timestamp when --remote-time is used.

Extended test 3012 to verify.

Reported-by: hgdagon on github
Fixes #12614

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -178,10 +178,18 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
           return CURL_WRITEFUNC_ERROR;
         }
 
+        if(per->config->output_dir) {
+          outs->filename = aprintf("%s/%s", per->config->output_dir, filename);
+          free(filename);
+          if(!outs->filename)
+            return CURL_WRITEFUNC_ERROR;
+        }
+        else
+          outs->filename = filename;
+
         outs->is_cd_filename = TRUE;
         outs->s_isreg = TRUE;
         outs->fopened = FALSE;
-        outs->filename = filename;
         outs->alloc_filename = TRUE;
         hdrcbdata->honor_cd_filename = FALSE; /* done now! */
         if(!tool_create_output_file(outs, per->config))
--- a/src/tool_cb_wrt.c
+++ b/src/tool_cb_wrt.c
@@ -57,22 +57,12 @@ bool tool_create_output_file(struct OutStruct *outs,
   struct GlobalConfig *global;
   FILE *file = NULL;
   char *fname = outs->filename;
-  char *aname = NULL;
   DEBUGASSERT(outs);
   DEBUGASSERT(config);
   global = config->global;
   if(!fname || !*fname) {
     warnf(global, "Remote filename has no length");
     return FALSE;
-  }
-
-  if(config->output_dir && outs->is_cd_filename) {
-    aname = aprintf("%s/%s", config->output_dir, fname);
-    if(!aname) {
-      errorf(global, "out of memory");
-      return FALSE;
-    }
-    fname = aname;
   }
 
   if(config->file_clobber_mode == CLOBBER_ALWAYS ||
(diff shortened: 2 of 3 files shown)


jay:
afaict the outfile automatically has its outdir prepended [elsewhere](https://github.com/curl/curl/blob/curl-8_5_0/src/tool_operate.c#L1113-L1121) but the filename may change because it was extracted from content disposition (is_cd_filename true). in the latter case the outdir is later prepended in tool_create_output_file and this pr saves that filename with outdir prepended. makes sense except if no clobber code in the previous block changes the filename. i think the prepend should be done and saved earlier, how about in the header callback:

~~~diff
diff --git a/src/tool_cb_hdr.c b/src/tool_cb_hdr.c
index 198a8d0..6f4cdf5 100644
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -178,10 +178,19 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
           return CURL_WRITEFUNC_ERROR;
         }
 
+        if(per->config->output_dir && *per->config->output_dir) {
+          outs->filename = aprintf("%s/%s", per->config->output_dir, filename);
+          if(!outs->filename) {
+            free(filename);
+            return CURL_WRITEFUNC_ERROR;
+          }
+        }
+        else
+          outs->filename = filename;
+
         outs->is_cd_filename = TRUE;
         outs->s_isreg = TRUE;
         outs->fopened = FALSE;
-        outs->filename = filename;
         outs->alloc_filename = TRUE;
         hdrcbdata->honor_cd_filename = FALSE; /* done now! */
         if(!tool_create_output_file(outs, per->config))
diff --git a/src/tool_cb_wrt.c b/src/tool_cb_wrt.c
index 98063c3..143cba2 100644
--- a/src/tool_cb_wrt.c
+++ b/src/tool_cb_wrt.c
@@ -57,7 +57,6 @@ bool tool_create_output_file(struct OutStruct *outs,
   struct GlobalConfig *global;
   FILE *file = NULL;
   char *fname = outs->filename;
-  char *aname = NULL;
   DEBUGASSERT(outs);
   DEBUGASSERT(config);
   global = config->global;
@@ -66,15 +65,6 @@ bool tool_create_output_file(struct OutStruct *outs,
     return FALSE;
   }
 
-  if(config->output_dir && outs->is_cd_filename) {
-    aname = aprintf("%s/%s", config->output_dir, fname);
-    if(!aname) {
-      errorf(global, "out of memory");
-      return FALSE;
-    }
-    fname = aname;
-  }
-
   if(config->file_clobber_mode == CLOBBER_ALWAYS ||
      (config->file_clobber_mode == CLOBBER_DEFAULT &&
       !outs->is_cd_filename)) {
@@ -94,14 +84,12 @@ bool tool_create_output_file(struct OutStruct *outs,
       char *newname;
       /* Guard against wraparound in new filename */
       if(newlen < len) {
-        free(aname);
         errorf(global, "overflow in filename generation");
         return FALSE;
       }
       newname = malloc(newlen);
       if(!newname) {
         errorf(global, "out of memory");
-        free(aname);
         return FALSE;
       }
       memcpy(newname, fname, len);
@@ -135,10 +123,8 @@ bool tool_create_output_file(struct OutStruct *outs,
   if(!file) {
     warnf(global, "Failed to open the file %s: %s", fname,
           strerror(errno));
-    free(aname);
     return FALSE;
   }
-  free(aname);
   outs->s_isreg = TRUE;
   outs->fopened = TRUE;
   outs->stream = file;
~~~

jay:
It occurs to me that debug builds using CURL_TESTDIR may already set the directory regardless of if the filename is from the URL or content disposition. I think outdir prepend should be ignored in those cases or give an assertion, what do you think?

PR #12620, opened 2024-01-02
tool: make parser reject blank arguments if not supported

bagder:
Already in the getstr() function that clones the input argument.


src/tool_helpers.c
```diff
@@ -78,6 +78,8 @@ const char *param2text(int res)
     return "error encountered when reading a file";
   case PARAM_EXPAND_ERROR:
     return "variable expansion failure";
+  case PARAM_BLANK_STRING:
+    return "blank argument where content were expected";
```
qhill, line 82: s/were/was/

bagder: Thanks, I think I will even go with **is**.

PR #12631, opened 2024-01-03
tool_getparam: cmdline option parser refactor

bagder:
- the option names are now alpha sorted and lookup is a lot faster

- use case sensitive matching. It was previously case insensitive, but that was not documented nor tested.

- remove "partial match" feature. It was not documented, not tested and was always fragile as existing use could break when we add a new option

- lookup short options via a table

### DIFF
--- a/src/tool_cfgable.c
+++ b/src/tool_cfgable.c
@@ -25,6 +25,7 @@
 
 #include "tool_cfgable.h"
 #include "tool_formparse.h"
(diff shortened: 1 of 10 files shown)


bagder:
The most relevant speed tests I did with a manually added loop around the `getparameter()` call, trying the function in master 10 million times vs the one in this branch the same amount of times for something basic such as `--verbose`. This branch proved to execute more than a magnitude faster.

Issue #12632, opened 2024-01-03
libcurl 8.5.0 breaks mpd http streaming


bubbleguuum:
### I did this

libcurl  v8.5.0 is causing issues in mpd http streaming, for http streams whose size is > 128 MB (which causes mpd to not fully buffer the stream in memory). Shortly after playing such stream, a curl API inexplicably returns CURLE_OUT_OF_MEMORY, causing mpd to immediately stop playback. libcurl v8.4.0 and prior work just fine.

I reported all the details and findings about it and how to reproduce it on this [mpd issue](https://github.com/MusicPlayerDaemon/MPD/issues/1952).

The problematic commit has been found to be https://github.com/curl/curl/commit/47f5b1a37f3e8055dfea446236c05478087858dc via bissecting.

After analysis, the mpd developer thinks this is a curl issue, hence this report.

Interestingly, mpd running with git libcurl (as of the date of this post) do not cause an OOM but instead will make the mpd "io" thread (making curl API calls) to take 100% CPU.

Any insights for fixing this issue would be appreciated.

### curl/libcurl version

libcurl 8.5.0

### operating system

openSUSE Tumbleed (this issue is not OS dependent).

bagder:
Pointing to a cryptic issue in another application is not helping us much. Please describe a lot more about the actual problem and how we can reproduce it - without running or debugging someone else's complicated application.

bubbleguuum:
mpd developer: not my problem
curl developer: not my problem

After spending several hours on this issue to narrow it down and writing reports, I'm sick of it. Feel free to close.
Let's keep mpd crashing.

bagder:
A problem with curl *is* a problem for us, but for us to be able to work on a reported issue we need some guidance - a detailed description to start with. If neither you nor any other mpd user/developer want to help with that, then I cannot see how we can act on this report. Then it will be closed, yes. We are not mind-readers, unfortunately.

bubbleguuum:
If you want more details, ask mpd developer MaxKellerman, assuming he his willing to help, which I doubt.
I've done my part at this point.

bagder:
Sorry, but if nobody can describe **a curl problem** to us then there is no curl problem as far as I am concerned.

bubbleguuum:
Did you read this analysis from the mpd developer wit the change in behavior between 6.4.0 and 6.5.0:

https://github.com/MusicPlayerDaemon/MPD/issues/1952#issuecomment-1875905999

bagder:
[...]

PR #12633, opened 2024-01-04
adjust_pollset fix

icing:
- do not add a socket for POLLIN when the transfer does not want to send (for example is paused).
- refs #12632


MaxKellermann:
I confirm that this fixes https://github.com/MusicPlayerDaemon/MPD/issues/1952 (repeating my reply https://github.com/curl/curl/issues/12632#issuecomment-1876756935)

PR #12634, opened 2024-01-04
wolfssl: Allow use of certificate chain

lealem47:
Fixes ZD#17158.
Customer is using a certificate chain. Leaf cert + intermediate cert.

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -583,12 +583,25 @@ wolfssl_connect_step1(struct Curl_cfilter *cf, struct Curl_easy *data)
   if(ssl_config->primary.clientcert && ssl_config->key) {
     int file_type = do_file_type(ssl_config->cert_type);
 
-    if(wolfSSL_CTX_use_certificate_file(backend->ctx,
-                                        ssl_config->primary.clientcert,
-                                        file_type) != 1) {
-      failf(data, "unable to use client certificate (no key or wrong pass"
-            " phrase?)");
-      return CURLE_SSL_CONNECT_ERROR;
+    if(file_type == WOLFSSL_FILETYPE_PEM) {
+      if(wolfSSL_CTX_use_certificate_chain_file(backend->ctx,
+                                                ssl_config->primary.clientcert)
+         != 1) {
+        failf(data, "unable to use client certificate");
+        return CURLE_SSL_CONNECT_ERROR;
+      }
+    }
+    else if(file_type == WOLFSSL_FILETYPE_ASN1) {
+      if(wolfSSL_CTX_use_certificate_file(backend->ctx,
+                                          ssl_config->primary.clientcert,
+                                          file_type) != 1) {
+        failf(data, "unable to use client certificate");
+        return CURLE_SSL_CONNECT_ERROR;
+      }
+    }
+    else {
+      failf(data, "unknown cert type");
+      return CURLE_BAD_FUNCTION_ARGUMENT;
     }
 
     file_type = do_file_type(ssl_config->key_type);


bagder:
Please elaborate on what this change does and why we want it!

Why does the type of the certificate change what function to call?

lealem47:
`wolfSSL_CTX_use_certificate_chain_file` should be used instead of `wolfSSL_CTX_use_certificate_file` with PEM files so that a user can load in a certificate chain and have it be properly utilized by the wolfSSL library. Failure to do so can cause an authentication error since the peer will need have the intermediate certs to verify a leaf cert (see ZD#17158). Concatenated certificate chains are not valid in DER format, so `wolfSSL_CTX_use_certificate_file` is still used for that case.

jay:
It's probably ok. It's similar to what we do for client certs when using OpenSSL, see [here](https://github.com/curl/curl/blob/curl-8_5_0/lib/vtls/openssl.c#L1294-L1331).

I say probably because I don't know if wolfSSL_CTX_use_certificate_chain_file has exactly the same behavior as OpenSSL. OpenSSL's SSL_CTX_use_certificate_chain_file apparently has an undocumented parsing that allows private keys in the chain file but wolfSSL may not; I found [this bug](https://github.com/civetweb/civetweb/issues/437) when searching but it's several years old.

bagder:
Use of this function is the solution suggested by wolfSSL engineers. However, I would like to offer this slightly cleaned up alternative take:

~~~diff
diff --git a/lib/vtls/wolfssl.c b/lib/vtls/wolfssl.c
index 5250e091b..a3c017cea 100644
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -581,24 +581,29 @@ wolfssl_connect_step1(struct Curl_cfilter *cf, struct Curl_easy *data)
 
   /* Load the client certificate, and private key */
   if(ssl_config->primary.clientcert && ssl_config->key) {
     int file_type = do_file_type(ssl_config->cert_type);
 
-    if(file_type == WOLFSSL_FILETYPE_PEM &&
-            wolfSSL_CTX_use_certificate_chain_file(backend->ctx,
-                                        ssl_config->primary.clientcert) != 1) {
-      failf(data, "unable to use client certificate (no key or wrong pass"
-            " phrase?)");
-      return CURLE_SSL_CONNECT_ERROR;
+    if(file_type == WOLFSSL_FILETYPE_PEM) {
+      if(wolfSSL_CTX_use_certificate_chain_file(backend->ctx,
+                                                ssl_config->primary.clientcert)
+         != 1) {
+        failf(data, "unable to use client certificate");
+        return CURLE_SSL_CONNECT_ERROR;
+      }
     }
-    else if(file_type == WOLFSSL_FILETYPE_ASN1 &&
-            wolfSSL_CTX_use_certificate_file(backend->ctx,
-                                        ssl_config->primary.clientcert,
-                                        file_type) != 1) {
-      failf(data, "unable to use client certificate (no key or wrong pass"
-            " phrase?)");
-      return CURLE_SSL_CONNECT_ERROR;
+    else if(file_type == WOLFSSL_FILETYPE_ASN1) {
+      if(wolfSSL_CTX_use_certificate_file(backend->ctx,
+                                          ssl_config->primary.clientcert,
+                                          file_type) != 1) {
+        failf(data, "unable to use client certificate");
+        return CURLE_SSL_CONNECT_ERROR;
+      }
+    }
+    else {
+      failf(data, "unknown cert type");
+      return CURLE_BAD_FUNCTION_ARGUMENT;
     }
 
     file_type = do_file_type(ssl_config->key_type);
     if(wolfSSL_CTX_use_PrivateKey_file(backend->ctx, ssl_config->key,
                                        file_type) != 1) {

jay:
That looks fine as well.

lealem47:
Yes that looks good to me

bagder (approved):

bagder:
Thanks!

PR #12635, opened 2024-01-04
Revert "CI: run Circle macOS builds on x86 for now"

dfandrich:
This reverts commit 2683de3078eadc86d9b182e7417f4ee75a247e2c.
ARM resources are now available in Circle CI, so run these builds on ARM
again.

Follow-up to 2683de30

Closes #12635

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,42 +54,42 @@ commands:
       - run:
           command: |
             autoreconf -fi
-            ./configure --enable-warnings --enable-websockets --without-ssl CFLAGS='-Wno-vla -mmacosx-version-min=10.9'
+            ./configure --enable-warnings --enable-websockets --without-ssl CFLAGS='-Wno-vla -mmacosx-version-min=10.9' CPPFLAGS="-I$(brew --prefix libpsl)/include" LDFLAGS="-L$(brew --prefix libpsl)/lib -L$(brew --prefix icu4c)/lib" LIBS="-licuuc -licudata"
 
   configure-macos-debug:
     steps:
       - run:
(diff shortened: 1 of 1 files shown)


bagder (approved):

dfandrich:
A spot check showed that at least one "ARM" build was actually still running on x86_64. I'll take a look when I get a chance.

dfandrich:
Try again now, all the expected builds are running on ARM, but most faily because configure can't find libpsl. It's being installed by Brew and it is found fine on x86, but on ARM libpsl is simply not found. Is it installed in a different location on ARM that isn't being searched?

dfandrich:
I've figured out the incantation to libpsl working (additions to CPPFLAGS, LDFLAGS and LIBS). However, macos-arm-http-torture-ftp is failing due to unknown reasons on ARM.  The tests are running fine, then they just stop after test 250 with `make: *** [test-ci] Terminated: 15`.

dfandrich:
The problem was that test 250 is a slow test and Circle CI killed the test suite because it looked like it hung. I've explicitly disabled that test now and the PR should (finally) be ready to submit.

dfandrich:
I worked around the problem by adding `LIBS="-licuuc -licudata"` to configure [on the M1 builds in CI](https://github.com/curl/curl/commit/d7332e3e46c3ef401b34e6a1a129eb4dd846c452). So, I still don't know if there's a curl problem in the way configure is using pkg-config or if it's actually a Brew problem. But, if it's only happening in CI it would be pretty tedious to debug further. It was painful enough coming up with a workaround that worked.

PR #12652, opened 2024-01-07
telnet: use dynbuf instad of malloc for escape buffer

bagder:
Previously, send_telnet_data() would malloc + free a buffer every time for escaping IAC codes. Now, it reuses a dynbuf for this purpose.


jay:
LGTM. Note a bunch of jobs are failing MQTT testing, not sure why that is.

PR #12662, opened 2024-01-09
slist.c: make get_list_last be O(1)

rockeet:
Time complexity of `get_list_last` is O(n),  `Curl_slist_append_nodup` and `curl_slist_append` are also O(n).

This PR add an extra field `tail` on list head node, and update `tail` on each append, thus time complexity is O(1).

All existing tests passed, no new tests are required!

### DIFF
--- a/lib/slist.c
+++ b/lib/slist.c
@@ -32,21 +32,19 @@
 #include "curl_memory.h"
 #include "memdebug.h"
 
+struct curl_slist_head {
+   struct curl_slist head;
+   struct curl_slist *tail;
+};
+
 /* returns last node in linked list */
 static struct curl_slist *slist_get_last(struct curl_slist *list)
 {
-  struct curl_slist     *item;
-
   /* if caller passed us a NULL, return now */
   if(!list)
     return NULL;
-
-  /* loop through to find the last item */
-  item = list;
-  while(item->next) {
-    item = item->next;
-  }
-  return item;
+  else
+    return ((struct curl_slist_head*)list)->tail;
 }
 
 /*
@@ -65,7 +63,11 @@ struct curl_slist *Curl_slist_append_nodup(struct curl_slist *list, char *data)
 
   DEBUGASSERT(data);
 
-  new_item = malloc(sizeof(struct curl_slist));
+  if(list) {
+    new_item = malloc(sizeof(struct curl_slist));
+  } else {
+    new_item = malloc(sizeof(struct curl_slist_head));
+  }
   if(!new_item)
     return NULL;
 
@@ -73,11 +75,14 @@ struct curl_slist *Curl_slist_append_nodup(struct curl_slist *list, char *data)
   new_item->data = data;
 
   /* if this is the first item, then new_item *is* the list */
-  if(!list)
+  if(!list) {
+    ((struct curl_slist_head*)new_item)->tail = new_item;
     return new_item;
+  }
 
   last = slist_get_last(list);
   last->next = new_item;
+  ((struct curl_slist_head*)list)->tail = new_item;
   return list;
 }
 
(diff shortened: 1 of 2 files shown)


lib/slist.c
```diff
@@ -65,19 +53,26 @@ struct curl_slist *Curl_slist_append_nodup(struct curl_slist *list, char *data)
...
-  if(!list)
+  if(!list) {
+    ((struct curl_slist_head*)new_item)->tail = new_item;
     return new_item;
+  }
 
-  last = slist_get_last(list);
+  last = ((struct curl_slist_head*)list)->tail;
   last->next = new_item;
+  ((struct curl_slist_head*)list)->tail = new_item;
```
emanuele6, line 75: `head*` => `head *` same above

lib/slist.c
```diff
@@ -65,19 +53,26 @@ struct curl_slist *Curl_slist_append_nodup(struct curl_slist *list, char *data)
...
 
   /* if this is the first item, then new_item *is* the list */
-  if(!list)
+  if(!list) {
+    ((struct curl_slist_head*)new_item)->tail = new_item;
     return new_item;
+  }
 
-  last = slist_get_last(list);
+  last = ((struct curl_slist_head*)list)->tail;
```
emanuele6, line 73: The `ASAN` job detects an invalid read here because the `tests/http/clients/tls-session-reuse.c` uses a stack-allocated zero-initialised `curl_slist`, and calls `curl_slist_append(&slist, string)` on it instead of just calling `curl_slist_append(NULL, string)` for some reason. :/

https://github.com/curl/curl/blob/912d80c68019d3d9a4ceb9993a596dff8009f4d0/tests/http/clients/tls-session-reuse.c#L188

This patch may be backward incompatible, though that test file is fairly recent (5 months old), and it is the only instance of that pattern in the `curl` repo:
```console
$ git grep 'struct[^*]*curl_slist[^*]*;'
lib/slist.c:68:  new_item = malloc(sizeof(struct curl_slist));
tests/http/clients/tls-session-reuse.c:188:  struct curl_slist resolve;
```

bagder:
`Curl_slist_append_nodup` is used by `curl_slist_append()`. This latter function is a public API. It accepts a `struct curl_slist *` as first argument and it returns a `struct curl_slist *`. You cannot easily change the struct and maintain compatibility.

rockeet:
> `Curl_slist_append_nodup` is used by `curl_slist_append()`. This latter function is a public API. It accepts a `struct curl_slist *` as first argument and it returns a `struct curl_slist *`. You cannot easily change the struct and maintain compatibility.

The new `struct curl_slist_head` is compatible to `struct curl_slist`, unless users allocate a `struct curl_slist` themselves without using `curl_slist_append()`(like [tls-session-reuse.c](https://github.com/curl/curl/blob/912d80c68019d3d9a4ceb9993a596dff8009f4d0/tests/http/clients/tls-session-reuse.c#L188) does). I think this is not an issue

bagder:
> unless users allocate a struct curl_slist themselves

The thing with supporting the public struct and the existing API is that we cannot limit who allocates the struct or how it is allocated. The function should still work. Otherwise it is not compatible.

tests/http/clients/tls-session-reuse.c
```diff
@@ -185,7 +185,7 @@ int main(int argc, char *argv[])
   CURLMsg *msg;
   CURLSH *share;
   CURLU *cu;
-  struct curl_slist resolve;
+  struct curl_slist* resolve = curl_slist_append(NULL, "");
```
bagder, line 188: This change should not be necessary. If it is, the change is not compatible...

bagder:
Stalled.

PR #12665, opened 2024-01-09
multi: fix incorrect multi_easy logic

jay:
- Use data->multi and not data->multi_easy to refer to the active multi.

Prior to this change there were a few places that incorrectly treated data->multi_easy, if != NULL, as the easy handle's active multi.

Background:

The easy handle's active multi is always data->multi. It points to either the user-created multi (if the multi interface is used) or the internally created multi (if the easy interface is used). The internally created multi is always data->multi_easy and is retained until the easy handle is cleaned up.

If an easy handle is used with the easy interface and then the multi interface, data->multi_easy and data->multi will both exist and be different.

Closes #xxxxx

---

There's one other that I didn't change because I'm not sure if the multi_easy should be checked before multi:

https://github.com/curl/curl/blob/8edcfedc1a144f438bd1cdf814a0016cbe678aaf/lib/connect.c#L270-L287

Some logic is used to get what should be the connection cache used by the easy handle. I notice `data->state.conn_cache` is not checked so I assume it's not available at that point.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -3706,19 +3706,13 @@ static void process_pending_handles(struct Curl_multi *multi)
 
 void Curl_set_in_callback(struct Curl_easy *data, bool value)
 {
-  /* might get called when there is no data pointer! */
-  if(data) {
-    if(data->multi_easy)
-      data->multi_easy->in_callback = value;
-    else if(data->multi)
-      data->multi->in_callback = value;
-  }
+  if(data && data->multi)
+    data->multi->in_callback = value;
 }
 
-bool Curl_is_in_callback(struct Curl_easy *easy)
+bool Curl_is_in_callback(struct Curl_easy *data)
 {
-  return ((easy->multi && easy->multi->in_callback) ||
-          (easy->multi_easy && easy->multi_easy->in_callback));
+  return (data && data->multi && data->multi->in_callback);
 }
 
 unsigned int Curl_multi_max_concurrent_streams(struct Curl_multi *multi)
--- a/lib/easy.c
+++ b/lib/easy.c
@@ -1305,7 +1305,7 @@ static int conn_upkeep(struct Curl_easy *data,
     conn->handler->connection_check(data, conn, CONNCHECK_KEEPALIVE);
   }
   else {
-    /* Do the generic action on the FIRSTSOCKE filter chain */
+    /* Do the generic action on the FIRSTSOCKET filter chain */
     Curl_conn_keep_alive(data, conn, FIRSTSOCKET);
   }
   Curl_detach_connection(data);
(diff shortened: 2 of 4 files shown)


bagder (approved):
:+1:

bagder:
@jay why not merge this?

jay:
> why not merge this?

I was unsure about Curl_getconnectinfo but I think I figured it out. Please review the squashme commit.

lib/multi.c
```diff
@@ -898,7 +898,7 @@ CURLMcode curl_multi_remove_handle(struct Curl_multi *multi,
   /* Remove the association between the connection and the handle */
   Curl_detach_connection(data);
 
-  if(data->set.connect_only && !data->multi_easy) {
+  if(data->set.connect_only && (data->multi != data->multi_easy)) {
```
bagder, line 901: Why this change? If the easy interface is not used, the `multi_easy` pointer is NULL so isn't both the old and the new code achieving the same thing?

lib/connect.c
```diff
@@ -282,9 +282,9 @@ curl_socket_t Curl_getconnectinfo(struct Curl_easy *data,
                            data->share && (data->share->specifier
                            & (1<< CURL_LOCK_DATA_CONNECT))?
                            &data->share->conn_cache:
-                           data->multi_easy?
-                           &data->multi_easy->conn_cache:
-                           &data->multi->conn_cache, &find, conn_is_conn);
+                           data->multi?
+                           &data->multi->conn_cache:
+                           &data->multi_easy->conn_cache, &find, conn_is_conn);
```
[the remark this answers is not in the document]
bagder: I mean, it can hardly reach this code with `data->multi` being NULL, can it ?

lib/multi.c
jay: > If the easy interface is not used, the `multi_easy` pointer is NULL

sometimes. the rub is if an easy handle is used with the easy interface and then the multi interface. in that case they point to different multis. this pr addresses that issue by prioritizing multi over multi_easy  because the former is always the in-use multi. my hang up with Curl_getconnectinfo (which is used in this logic) is, following this scenario, what happens when the easy handle is removed by the user via curl_multi_remove_handle, because that would seem to cause disconnection of connections in multi_easy and not multi. afaict, this change fixes a bug by checking that they are different and Curl_getconnectinfo prioritizes multi over multi_easy

bagder:
@jay I propose  #12992 so that each easy handle only has one multi handle to relate to.

bagder:
Now merged. With this, an easy handle can only ever be associated with a single multi handle.

bagder (changes requested):
This can now be adjusted

bagder:
@jay do you want me to continue this?

I believe `Curl_close()` can also be simplified somewhat (when the easy handle is found to be part of a multi handle).

jay:
> do you want me to continue this?
> 
> I believe `Curl_close()` can also be simplified somewhat (when the easy handle is found to be part of a multi handle).

I dropped the fixup commit and changed the other commit's message, if more needs to be done have at it!

jay:
@bagder can you have another look, I don't think there's anything left to add

bagder (approved):
Looks good!

PR #12675, opened 2024-01-10
headers: make sure the trailing newline is not stored

bagder:
extended test1940 to verify blank header fields too

Bug: https://curl.se/mail/lib-2024-01/0019.html
Reported-by: Dmitry Karpov


gad-imageinsight:
Looks like a possible typo at line 38 of tests/libtest/lib1940.c. It reads “blank” [sic] whereas in tests/data/test1940 it is “Blank”, with an uppercase “B”. “Blank2” appears that way in both locations.

bagder:
I actually made that on purpose. Headers arrive with a specific casing but the lookup is case insensitive.

PR #12676, opened 2024-01-10
hsts: remove assert for zero length domain

bagder:
A zero length domain can happen if the HSTS parser is given invalid input data which is not unheard of and is done by the fuzzer.

Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65661


bagder:
Follow-up from cfe7902111ae547873

PR #12678, opened 2024-01-10
QUIC TLS consolidation and some small fixes

icing:
Refactor `ngtcp2` and `quiche` TLS setup code into a common source file:
- add `vquic/vquic-tls.[ch]` for setting up TLS contexts for OpenSSL (and variants), GnuTLS and wolfSSL
- change `curl_ngtcp2.c` to use this
- change `curl_quiche.c` to use this

Fix QUIC connect time reporting in curl property `time_connect` to count the time until the first byte from the server is seen:
- add test_01_06 to check that this works for all HTTP versions
- fix reporting in `ngtcp2` and `quiche`

In `cf-socket.c` remember if opened socket is connected or not. Also, discard extracting the peer address from the socket again, since this has already been stored when it was opened.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -788,6 +788,7 @@ struct cf_socket_ctx {
 #endif
   BIT(got_first_byte);               /* if first byte was received */
   BIT(accepted);                     /* socket was accepted, not connected */
+  BIT(sock_connected);               /* socket is "connected", e.g. in UDP */
   BIT(active);
   BIT(buffer_recv);
 };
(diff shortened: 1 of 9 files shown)


lib/cf-socket.c
```diff
@@ -788,6 +788,7 @@ struct cf_socket_ctx {
 #endif
   BIT(got_first_byte);               /* if first byte was received */
   BIT(accepted);                     /* socket was accepted, not connected */
+  BIT(sock_connected);               /* socket is "connected", e.g. in UDP */
```
bagder, line 791: I did not spot any *user* of this info, so I presume this is something you plan to use coming up?

icing: I found it interesting to keep, since the OpenSSL issue with connected socket requires, for now, a workaround. But we can also remove that in this PR and shift it to later.

bagder: I just wanted to make sure. Let's keep it there for now.

bagder (approved):

PR #12683, opened 2024-01-11
configure: when enabling QUIC, check that TLS supports QUIC

bagder:
Most importantly perhaps is when using OpenSSL that the used build/flavor has the QUIC API: the vanilla OpenSSL does not, only BoringSSL, libressl, AWS-LC and quictls do.

Ref: https://github.com/curl/curl/commit/5d044ad9480a9f556f4b6a252d7533b1ba7fe57e#r136780413

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2776,6 +2776,11 @@ esac
 
 curl_tcp2_msg="no      (--with-ngtcp2)"
 if test X"$want_tcp2" != Xno; then
+
+  if test "$QUIC_ENABLED" != "yes"; then
+    AC_MSG_ERROR([the detected TLS library does not support QUIC, making --with-ngtcp2 a no-no])
+  fi
+
   dnl backup the pre-ngtcp2 variables
   CLEANLDFLAGS="$LDFLAGS"
   CLEANCPPFLAGS="$CPPFLAGS"
@@ -3030,6 +3035,11 @@ esac
 
 curl_http3_msg="no      (--with-nghttp3)"
 if test X"$want_nghttp3" != Xno; then
+
+  if test "$NGTCP2_ENABLED" != "1"; then
+    AC_MSG_ERROR([--with-nghttp3 also requires --with-ntcp2])
+  fi
+
   dnl backup the pre-nghttp3 variables
   CLEANLDFLAGS="$LDFLAGS"
   CLEANCPPFLAGS="$CPPFLAGS"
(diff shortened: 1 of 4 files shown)


bagder:
I did not make any elaborate checks for GnuTLS and wolfSSL as I figure they are much likely to be a problem for users since they support QUIC "from scratch".

icing:
Nice. One problem remains. `vquic-tls.c` should guard against `!USE_MSH3`, because that is a QUIC built that "works" with a plain OpenSSL.

bagder:
Oh right, that complicates matters a little...

gvanem:
> because that is a QUIC built that "works" with a plain OpenSSL.

From Msh3's *About*: *Minimal HTTP/3 library on top of MsQuic*.
So not necessarily *plain OpenSSL*. Ref: https://github.com/microsoft/msquic/blob/main/docs/Platforms.md.

gvanem:
So what about `USE_MSH3` on Windows? 
Now I need to add `CSOURCES := $(filter-out vquic/vquic-tls.c, $(CSOURCES))` to my tweaked Makefile 
since the warning/error is still the same:
```c
vquic/vquic-tls.c(219,3): warning C4013: 'SSL_set_quic_use_legacy_codepoint' undefined; assuming extern returning int
  SSL_set_quic_use_legacy_codepoint(ctx->ssl, 0);
  ^
```

bagder:
That needs a different take than this PR, yes. It probably needs adjustments in the code.

bagder:
We should probably have CI jobs doing HTTTP/3 on Windows too...

gvanem:
Or should it say:
```diff
--- a/lib/vquic/vquic-tls.c 2024-01-11 11:17:26
+++ b/lib/vquic/vquic-tls.c 2024-01-12 10:26:47
@@ -24,7 +24,7 @@

 #include "curl_setup.h"

-#if defined(ENABLE_QUIC) && \
+#if defined(ENABLE_QUIC) && defined(USE_NGTCP2) && defined(USE_NGHTTP3) \
   (defined(USE_OPENSSL) || defined(USE_GNUTLS) || defined(USE_WOLFSSL))

 #ifdef USE_OPENSSL
```

bagder:
Isn't it used for quiche too?

bagder:
So maybe?
~~~c
#if defined(ENABLE_QUIC) && !defined(USE_MSH3) && \
  (defined(USE_OPENSSL) || defined(USE_GNUTLS) || defined(USE_WOLFSSL))

bagder:
or wait, isn't this still needed for msh3 on linux?

icing:
We use this code only for ngtcp2 and quiche now.  Sp

```
#if (defined(USE_NGTCP2) || defined(USE_QUICHE)) && \
  (defined(USE_OPENSSL) || defined(USE_GNUTLS) || defined(USE_WOLFSSL))

```

should do it?

bagder:
right, the latter might be enough

bagder:
If we later need it for msh3 on linux, I figure we could set a define for each specific backend that needs the init instead, like `#define VQUIC_TLS_INIT 1`  and only check for that.

PR #12684, opened 2024-01-12
configure.ac: error if --with-winidn and target OS too old

jay:
- If the minimum Windows target OS is pre-Vista (_WIN32_WINNT < 0x600) then error.

- Stop comparing against WINVER when making the minimum OS determinations (ie check only _WIN32_WINNT instead of both since that's the one that matters).

Prior to this change, if --with-winidn was used in such a case configure would change the minimum target OS to Vista (0x600) so the build would compile. That was done primarily as a workaround for original MinGW which we no longer support.

Bug: https://github.com/curl/curl/pull/12606#issuecomment-1885381038
Reported-by: Viktor Szakats

Closes #xxxxx

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2473,17 +2473,14 @@ if test "$want_winidn" = "yes"; then
     AC_LANG_PROGRAM([[
       #include <windows.h>
     ]],[[
-      #if (WINVER < 0x600) && (_WIN32_WINNT < 0x600)
+      #if (_WIN32_WINNT < 0x600)
       #error
       #endif
     ]])
   ],[
   ],[
-     CFLAGS=`echo $CFLAGS | $SED -e 's/-DWINVER=[[^ ]]*//g'`
-     CFLAGS=`echo $CFLAGS | $SED -e 's/-D_WIN32_WINNT=[[^ ]]*//g'`
-     CPPFLAGS=`echo $CPPFLAGS | $SED -e 's/-DWINVER=[[^ ]]*//g'`
-     CPPFLAGS=`echo $CPPFLAGS | $SED -e 's/-D_WIN32_WINNT=[[^ ]]*//g'`
-     WINIDN_CPPFLAGS="$WINIDN_CPPFLAGS -DWINVER=0x0600"
+    AC_MSG_RESULT([no])
+    AC_MSG_ERROR([--with-winidn specified but minimum OS is too old])
   ])
   #
   CPPFLAGS="$CPPFLAGS $WINIDN_CPPFLAGS"
@@ -4608,7 +4605,7 @@ else
     AC_LANG_PROGRAM([[
       #include <windows.h>
     ]],[[
-      #if (WINVER < 0x600) && (_WIN32_WINNT < 0x600)
+      #if (_WIN32_WINNT < 0x600)
       #error
       #endif
     ]])


vszakats (approved):

vszakats:
Thinking about this, if we require Vista, we'd need some additional tweaks for consistency. In `lib/idn.c` we have logic to declare these WinIDN APIs ourselves when building for < 0x0600. (Which is _kind of_ valid, since WinIDN _could_ also work on XP, though it needed `normaliz.dll`, which used to be a separate download, but Microsoft deleted that download a long time ago. So in practice it became problematic.)

Anyway, if we require Vista for IDN, the `lib/idn.c` trick is no longer necessary and it'd probably be better to replace it with something like this in `lib/curl-setup.h` and CMake could also use an update:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3a6ae3d42..332c08186 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -841,13 +841,6 @@ else()
   set(HAVE_LIBIDN2 OFF)
 endif()
 
-if(WIN32)
-  option(USE_WIN32_IDN "Use WinIDN for IDN support" OFF)
-  if(USE_WIN32_IDN)
-    list(APPEND CURL_LIBS "normaliz")
-  endif()
-endif()
-
 #libpsl
 option(CURL_USE_LIBPSL "Use libPSL" ON)
 mark_as_advanced(CURL_USE_LIBPSL)
@@ -1085,6 +1078,14 @@ if(WIN32)
       unset(HAVE_INET_PTON CACHE)
     endif()
   endif()
+
+  option(USE_WIN32_IDN "Use WinIDN for IDN support" OFF)
+  if(USE_WIN32_IDN)
+    if(HAVE_WIN32_WINNT AND HAVE_WIN32_WINNT STRLESS "0x600")
+      message(FATAL_ERROR "WinIDN requires building for Windows Vista or newer.")
+    endif()
+    list(APPEND CURL_LIBS "normaliz")
+  endif()
 endif()
 
 check_include_file_concat("sys/filio.h"      HAVE_SYS_FILIO_H)
diff --git a/lib/curl_setup.h b/lib/curl_setup.h
index 551243e27..794d42058 100644
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -623,6 +623,14 @@
 
 /* ---------------------------------------------------------------- */
 
+#if !defined(_WIN32) && defined(USE_WIN32_IDN)
+#undef USE_WIN32_IDN
+#endif
+
+#if defined(USE_WIN32_IDN) && (_WIN32_WINNT < 0x0600)
+#error "WinIDN requires building for Windows Vista or newer."
+#endif
+
 #if defined(HAVE_LIBIDN2) && defined(HAVE_IDN2_H) && !defined(USE_WIN32_IDN)
 /* The lib and header are present */
 #define USE_LIBIDN2
diff --git a/lib/idn.c b/lib/idn.c
index 81a177f8c..c4167db18 100644
--- a/lib/idn.c
+++ b/lib/idn.c
@@ -51,20 +51,6 @@
 #include "memdebug.h"
 
 #ifdef USE_WIN32_IDN
-/* using Windows kernel32 and normaliz libraries. */
-
-#if !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x600
-WINBASEAPI int WINAPI IdnToAscii(DWORD dwFlags,
-                                 const WCHAR *lpUnicodeCharStr,
-                                 int cchUnicodeChar,
-                                 WCHAR *lpASCIICharStr,
-                                 int cchASCIIChar);
-WINBASEAPI int WINAPI IdnToUnicode(DWORD dwFlags,
-                                   const WCHAR *lpASCIICharStr,
-                                   int cchASCIIChar,
-                                   WCHAR *lpUnicodeCharStr,
-                                   int cchUnicodeChar);
-#endif
 
 #define IDN_MAX_LENGTH 255
```

If want to keep support WinIDN for XP, maybe the autotools detection could probe building with `IdnToAscii()` with normaliz lib, but that's a little shaky because an XP target env might not have the DLL.

bagder:
Stalled.

PR #12688, opened 2024-01-12
header comments re: DNS option functionality only available to c-ares

bch:
note re: c-ares -only options for those browsing curl.h

no functional change

### DIFF
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -1846,7 +1846,8 @@ typedef enum {
   /* allow GSSAPI credential delegation */
   CURLOPT(CURLOPT_GSSAPI_DELEGATION, CURLOPTTYPE_VALUES, 210),
 
-  /* Set the name servers to use for DNS resolution */
+  /* Set the name servers to use for DNS resolution.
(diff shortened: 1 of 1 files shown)


include/curl/curl.h
```diff
@@ -2051,7 +2052,8 @@ typedef enum {
   /* Head start in milliseconds to give happy eyeballs. */
   CURLOPT(CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS, CURLOPTTYPE_LONG, 271),
 
-  /* Function that will be called before a resolver request is made */
+  /* Function that will be called before a resolver request is made.
+   * Only supported by the c-ares DNS backend */
```
bagder, line 2056: `CURLOPT_RESOLVER_START_FUNCTION` is not c-ares specific!

bagder (changes requested):

include/curl/curl.h
bch: Will review.

bch:
will resubmit

PR #12695, opened 2024-01-14
Header comment wrt c-ares -only option (redux).

bch:
note in header re: c-ares -only option. For consistency with other related options,  and correctness. No functional change.


bch:
ah - my previous PR confusion was misreading "[...]only the ares resolver backend has a resolver state" in CURLOPT_RESOLVER_START_FUNCTION(3)

bagder (approved):

bagder:
Thanks!

PR #12703, opened 2024-01-15
asyn-ares: resolver timeout follows the c-ares mechanism

sunlin7:
Hi,
The curl should set c-ares's timeout only  for `c-ares` < `1.20.0`.

curl use a fixed  timeout=2s for `c-ares` from commit:
  https://github.com/curl/curl/commit/a181b4a053339c0713f9c31c724f15ec5beb8d56, 
actually  `c-ares` used default timeout to 2s since `c-ares-1.20.0` release (also mentioned on prevous commit).
  https://github.com/c-ares/c-ares/pull/542
and the timeout=2 from curl will overwrite the `option: timeout` from the `/etc/resolv.conf`, which supported since `c-ares-1.23.0`:
  https://github.com/c-ares/c-ares/pull/632/files

So setting c-ares timeout to 2s should only apply to old c-ares < 1.20.0, won't overwrite timeout for c-ares > 1.23.0.

### DIFF
--- a/lib/asyn-ares.c
+++ b/lib/asyn-ares.c
@@ -173,10 +173,16 @@ CURLcode Curl_resolver_init(struct Curl_easy *easy, void **resolver)
   int status;
   struct ares_options options;
   int optmask = ARES_OPT_SOCK_STATE_CB;
+  static int ares_ver = 0;
   options.sock_state_cb = sock_state_cb;
   options.sock_state_cb_data = easy;
-  options.timeout = CARES_TIMEOUT_PER_ATTEMPT;
-  optmask |= ARES_OPT_TIMEOUTMS;
+  if(ares_ver == 0) {
+    ares_version(&ares_ver);
+  }
+  if(ares_ver < 0x011400) { /* c-ares included similar change since 1.20.0 */
+    options.timeout = CARES_TIMEOUT_PER_ATTEMPT;
+    optmask |= ARES_OPT_TIMEOUTMS;
+  }
 
   status = ares_init_options((ares_channel*)resolver, &options, optmask);
   if(status != ARES_SUCCESS) {


bagder:
> The curl should set c-ares's timeout only for c-ares < 1.20.0.

Why? It risks/makes the timeouts to be different with different c-ares versions. Why do curl users want that?

Also, this patch checks the version of c--ares used when *building* curl, not the version that is actually used in *run-time*. That might not be the same...

sunlin7:
> Why? It risks/makes the timeouts to be different with different c-ares versions. Why do curl users want that?
curl failed example (curl 8.4+, c-ares 1.24+):
1. user config DNS timeout in `/etc/resolv.conf` with `options timeout:1` for DNS timeout 1 second
2. first DNS service down (or un-accessable IP address eg: 10.10.10.10, to trigger timeout case)
3. `time ./curl http://www.google.com` then  will timeout for 2 seconds.
Expected timeout is 1 second, for user had set timout in `resolv.conf` file.
This PR will let curl follow user timeout setting.

> Also, this patch checks the version of c--ares used when building curl, not the version that is actually used in run-time. That might not be the same...
It's already there: curl-8.3 with c-ares the DNS timeout is 5 seconds; while curl-8.4 with c-ares timeout is 2 seconds.

I like to change curl to get the version from loaded libcares dynamicly, give me minutes.

bagder:
> user config DNS timeout in /etc/resolv.conf with options timeout:1 for DNS timeout 1 second

I did not even know that thing existed.

This then begs the question: what if we want to set the limit to **N** seconds but allow /etc/resolv.conf to override if set. Is that then not possible, you know?

sunlin7:
Hi @bagder 
This change is strong enough:
if libares < 1.20.0:  curl set timeout to 2s;
if libares >= 1.20.0 it already has the timeout to 2s, curl don't need to set the timeout;
if libares >= 1.24.0, user can set the timeout via /etc/resolv.conf.

> This then begs the question: what if we want to set the limit to N seconds but allow /etc/resolv.conf to override if set. Is that then not possible, you know?

Currently no grace way to do that (unless we parse the /etc/resolv.conf, but that's not curl's scope.)

bagder:
thanks!

Issue #12663, opened 2024-01-09
multiple ssl backends don't work as expected


notjaywu:
### I did this

I built curl locally by following https://curl.se/docs/install.html

I have the following variables for both cases:
```
export ARCH=arm64
export SDK=macosx
export DEPLOYMENT_TARGET=10.15
export CFLAGS="-arch $ARCH -isysroot $(xcrun -sdk $SDK --show-sdk-path) -m$SDK-version-min=$DEPLOYMENT_TARGET"
```

## Case 1
* curl is built with `./configure --host=$ARCH-apple-darwin --prefix=$(pwd)/artifacts --with-secure-transport`
* and run `./src/curl -v https://localhost:8005`
 
I can receive the response correctly from the local TLS server.
```
* Host localhost:8005 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8005...
* Connected to localhost (::1) port 8005
* ALPN: curl offers http/1.1
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate: **@**.local (**)
* Server certificate: mkcert **@**.local (**)
* using HTTP/1.x
> GET / HTTP/1.1
> Host: localhost:8005
> User-Agent: curl/8.5.0-DEV
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Tue, 09 Jan 2024 07:57:11 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< Transfer-Encoding: chunked
<
hello world
* Connection #0 to host localhost left intact
```

## Case 2
* curl is built with `./configure --host=$ARCH-apple-darwin --prefix=$(pwd)/artifacts --with-ssl=/opt/homebrew/opt/openssl@3.0 --with-secure-transport --with-default-ssl-backend=openssl`
* run `CURL_SSL_BACKEND=secure-transport ./src/curl -v https://localhost:8005`
 
I will get the following error:
```
* Host localhost:8005 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8005...
* Connected to localhost (::1) port 8005
* ALPN: curl offers http/1.1
* SSL: peer not verified:  RecoverableTrustFailure
* Closing connection
curl: (60) SSL: peer not verified:  RecoverableTrustFailure
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
[...]

PR #12704, opened 2024-01-15
fix: securetransport doesn't work as expected when curl is built with multiple ssl backends

notjaywu:
This PR is to fix the [issue#12663](https://github.com/curl/curl/issues/12663)

Given curl is built with multiple SSL backends, `CAfile` is set incorrectly when `SecureTransport` is the selected TLS backend.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -434,11 +434,13 @@ CURLcode Curl_init_userdefined(struct Curl_easy *data)
 
   /* Set the default CA cert bundle/path detected/specified at build time.
    *
-   * If Schannel is the selected SSL backend then these locations are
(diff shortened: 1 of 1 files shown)


bagder (approved):

jay:
Let's hear from another dev on mac before we do this. Also we have code in tool_operate that exempts Schannel from SSL environment variables. Would it be a good idea to do the same change needed there?

https://github.com/curl/curl/blob/7161cb17c01dcff1dc5bf89a18437d9d729f1ecd/src/tool_operate.c#L2590-L2645

@nickzman

bagder:
> Let's hear from another dev on mac before we do this

I suspect @notjaywu is about the first person ever to use multissl with Secure Transport...

PR #12710, opened 2024-01-15
tool_operate: make --remove-on-error only remove "real" files

bagder:
Reported-by: Harry Sintonen

### DIFF
--- a/docs/cmdline-opts/remove-on-error.d
+++ b/docs/cmdline-opts/remove-on-error.d
@@ -12,4 +12,4 @@ When curl returns an error when told to save output in a local file, this
 option removes that saved file before exiting. This prevents curl from
 leaving a partial file in the case of an error during transfer.
 
-If the output is not a file, this option has no effect.
+If the output is not a regular file, this option has no effect.
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -637,8 +637,14 @@ noretry:
       errorf(config->global, "curl: (%d) Failed writing body", result);
     }
     if(result && config->rm_partial) {
-      notef(global, "Removing output file: %s", outs->filename);
-      unlink(outs->filename);
+      struct_stat st;
+      if(!stat(outs->filename, &st) &&
+         S_ISREG(st.st_mode) &&
+         !unlink(outs->filename)) {
+        notef(global, "Removed output file: %s", outs->filename);
+      }
+      else
+        warnf(global, "Failed removing: %s", outs->filename);
     }
   }
 
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -639,12 +639,15 @@ noretry:
     if(result && config->rm_partial) {
       struct_stat st;
       if(!stat(outs->filename, &st) &&
-         S_ISREG(st.st_mode) &&
-         !unlink(outs->filename)) {
-        notef(global, "Removed output file: %s", outs->filename);
+         S_ISREG(st.st_mode)) {
+        if(!unlink(outs->filename))
+          notef(global, "Removed output file: %s", outs->filename);
+        else
+          warnf(global, "Failed removing: %s", outs->filename);
       }
       else
-        warnf(global, "Failed removing: %s", outs->filename);
+        warnf(global, "Skipping removal; not a regular file: %s",
+              outs->filename);
     }
   }
 


src/tool_operate.c
```diff
@@ -653,8 +653,12 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
       errorf(config->global, "curl: (%d) Failed writing body", result);
     }
     if(result && config->rm_partial) {
-      notef(global, "Removing output file: %s", outs->filename);
-      unlink(outs->filename);
+      struct_stat st;
+      stat(outs->filename, &st);
+      if(S_ISREG(st.st_mode)) {
```
dfandrich, line 658: The stat error code should be checked or this could be working on uninitialized data.

src/tool_operate.c
```diff
@@ -653,8 +653,12 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
+        notef(global, "Removing output file: %s", outs->filename);
+        unlink(outs->filename);
```
dfandrich, line 660: A failure here should at least show a warning, if not return an error code from the program. And possibly a notef if it does successfully delete the file.

docs/cmdline-opts/remove-on-error.d
```diff
@@ -12,4 +12,4 @@ When curl returns an error when told to save output in a local file, this
 option removes that saved file before exiting. This prevents curl from
 leaving a partial file in the case of an error during transfer.
 
-If the output is not a file, this option has no effect.
+If the output is not a regular file, this option cannot do its job.
```
dfandrich, line 15: "cannot" makes it seems like it will try to do its job but won't succeed. I prefer the original working (but say "regular file") with this change.

dfandrich (changes requested):

src/tool_operate.c
bagder: In this case, it already has an error and that is why this function is executed. It would be a bit strange to change the error code I think. Right?

src/tool_operate.c
bagder: Indeed!

docs/cmdline-opts/remove-on-error.d
```diff
@@ -12,4 +12,4 @@ When curl returns an error when told to save output in a local file, this
 option removes that saved file before exiting. This prevents curl from
 leaving a partial file in the case of an error during transfer.
 
-If the output is not a regular file, this option cannot do its job.
+If the output is not a regular file, this option has not effect.
```
dfandrich, line 15: "no" effect

src/tool_operate.c
dfandrich: If it's going to return an error anyway, even in the absence of --fail, then that seems fine.

src/tool_operate.c
```diff
@@ -654,11 +654,13 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
...
       struct_stat st;
-      stat(outs->filename, &st);
-      if(S_ISREG(st.st_mode)) {
+      if(!stat(outs->filename, &st) &&
+         S_ISREG(st.st_mode)) {
         notef(global, "Removing output file: %s", outs->filename);
         unlink(outs->filename);
       }
+      else
+        warnf(global, "Failed removing: %s", outs->filename);
```
dfandrich, line 663: Technically, this isn't going to be displayed when removal fails, but only when the removal isn't even attempted.

src/tool_operate.c
```diff
@@ -654,11 +654,13 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
     }
     if(result && config->rm_partial) {
       struct_stat st;
-      stat(outs->filename, &st);
-      if(S_ISREG(st.st_mode)) {
+      if(!stat(outs->filename, &st) &&
+         S_ISREG(st.st_mode)) {
         notef(global, "Removing output file: %s", outs->filename);
         unlink(outs->filename);
```
dfandrich, line 660: I still think it's worth writing a message if the unlink fails.

src/tool_operate.c
bagder: Well, the conditions prevented it from trying, so in that sense it failed to do it.

Any suggestion on a better wording?

dfandrich: Maybe "Skipping removal; not a regular file: %s"

src/tool_operate.c
dfandrich: See other message.

dfandrich: Why can't I resolve this? I'm fine with this.

PR #12711, opened 2024-01-15
cmdline-opts/page-footer: mention env vars not used by schannel

bagder:
Ref: #12704


docs/cmdline-opts/page-footer
```diff
@@ -49,7 +49,8 @@ If set, the specified number of characters is used as the terminal width when
 the alternative progress-bar is shown. If not set, curl tries to figure it out
 using other ways.
 .IP "CURL_CA_BUNDLE <file>"
-If set, it is used as the --cacert value.
+If set, it is used as the --cacert value. (Unless Schannel is used as TLS
+backend.)
```
jay, line 53:
```suggestion
If set, it is used as the --cacert value. This environment variable is ignored
if Schannel is used as the TLS backend.
```

PR #12719, opened 2024-01-16
CURLINFO_USED_PROXY: return a bool whether the proxy was used

bagder:
Idea:  #12669

- [X] - test case 536 verifies
- [X] - `%{proxy_used}` for -w


include/curl/curl.h
```diff
@@ -2938,6 +2938,7 @@ typedef enum {
   CURLINFO_XFER_ID          = CURLINFO_OFF_T + 63,
   CURLINFO_CONN_ID          = CURLINFO_OFF_T + 64,
   CURLINFO_QUEUE_TIME_T     = CURLINFO_OFF_T + 65,
+  CURLINFO_WAS_PROXY        = CURLINFO_LONG + 66,
   CURLINFO_LASTONE          = 65
```
genotrance, line 2942: Is this a problem? Last is still 65.

bagder: It could become one. I'll fix.

bagder: Thanks!

PR #12722, opened 2024-01-16
test2307: mark as flaky

bagder:
Added in 49ca84144edf9790 (#12707) the test fails in numerous (unrelated) CI builds.

### DIFF
--- a/tests/data/test2307
+++ b/tests/data/test2307
@@ -2,6 +2,7 @@
 <info>
 <keywords>
 WebSockets
+flaky
 </keywords>
 </info>
 


bagder:
Actually, we need to fix this better. This test permafails on my machine.

@icing: the websocket function returns `CURLE_RECV_ERROR` on the too-large value but the test case expects `CURLE_WRITE_ERROR` to be returned. Where is that conversion done and is that really expected?

icing:
Probably a collision with the ws refactoring. I look into it.

PR #12730, opened 2024-01-17
introducing "curldown" for libcurl man page format

bagder:
## overview

curldown is this new file format for libcurl man pages. It is markdown inspired with differences:

- Each file has a set of leading headers with meta-data
- Supports a small subset of markdown
- Uses .md file extensions for editors/IDE/GitHub to treat them nicely
- Generates man pages very similar to the previous ones
- Generates man pages that still convert nicely to HTML on the website
- Detects and highlights mentions of curl symbols automatically (when
  their man page section is specified)

## tools:

- cd2nroff: converts from curldown to nroff man page
- nroff2cd: convert an (old) nroff man page to curldown
- cdall: convert many nroff pages to curldown versions
- cd2cd: verifies and updates a curldown to latest curldown

This setup generates .3 versions of all the curldown versions at build time.

## CI

Since the documentation is now technically markdown in the eyes of many
things, the CI runs many more tests and checks on this documentation,
including proselint, link checkers and tests that make sure we capitalize the
first letter after a period...

## tasks

- [X] **all** used `.XX` instructions in nroff files are handled by nroff2cd
- [X] actually switch to curldown files in the repository (merge those, remove the nroff versions)
- [X] make maketgz generate the man pages from the `.md` sources
- [X] dropped the .3 files from the dist tarball
- [X] `docs/CURLDOWN.md` documents the format and involved tools
- [X] render the docs when building with autotools
- [X] supports See-also: as a yaml list
- [X] new nroff converted files are >99% identical to the old nroff files (excluding the cleanups)
- [X] use `.md` file extensions
- [X] working out-of-tree builds
- [X] make all CI jobs work again
- [X] make proselint happy about all ~500 new markdown files
- [X] render the docs when building with cmake

The tests that rely on the man pages to be present will be disabled for the CI jobs that use cmake on Appveyor for now.


scripts/nroff2cd
```diff
@@ -0,0 +1,172 @@
...
+            if($d =~ /.so (.*)/) {
+                # this is basically an include, so do that
+                my $f = $1;
+                # remove leading directory
+                $f =~ s/(.*?\/)//;
+                close(F);
+                open(F, "<:crlf", "$f") || return 1;
+            }
+            if($d =~ /^\.TH ([^ ]*) (\d) \"(.*?)\" ([^ ]*)/) {
+                # header, this neesd to be the first thing after leading comments
```
martinpeck, line 70: typo: `neesd` should be `needs`

bagder: :+1:

bagder:
The `CURLDOWN.md` file within the PR is the embryo of documentation for the format and how to use it.

bagder:
I cannot figure out what the REUSE job complains about! :cry:

bagder:
@mxmehl is this a REUSE bug or something? It reports `cd2nroff` has no license and copyright, but the header is the same header used in numerous other files and should be fine. Or what am I missing?

bagder:
I'm considering switching to using `.md` extensions instead so that IDEs/editors, GitHub etc show and handle the files a little nicer. It is after all a subset of markdown. I just don't want to mislead users...

docs/CURLDOWN.md
```diff
@@ -0,0 +1,111 @@
...
+
+The `nroff2cd` tool generates code style quotes for all `.fi` sections since
+the nroff format does not carry a distinction.
+
+# Format
+
+Each curldown starts with a header with meta-data:
+
+    ---
+    c: Copyright (C) Daniel Stenberg, <daniel.se>, et al.
```
cmeister2, line 61: Searching for "markdown preamble" seems to indicate that most similar code here has the data in YAML format. Have you considered this? I think the only change is that your "See-also" section becomes a list.

vszakats: It's also referred to as "front matter": https://jekyllrb.com/docs/front-matter/

bagder: Yeah, I based the current layout on how I saw a few others do it.

Would we actually gain anything in particular by doing it yaml compliant as compared to the current?

bagder:
@mxmehl many thanks, got it!

bagder:
An interesting side-effect of switching to `.md` is that all of a sudden, the proselint CI job got a lot of more text to complain about... :laughing:

bagder:
If someone could help me add the .3 generation logic for cmake I would appreciate it. (the existing html and PDF generation stuff can be removed)

bagder:
The missing support for building this with cmake is now the biggest remaining obstacle. I intend to take a stab at it, and as a first step I did #12742 since it kind of blocked documentation generation in general.

bagder:
I'm right now leaning on disabling the documentation tests on CI builds using cmake. I simply cannot figure out how to make them generated with cmake and it drives me mad.

vszakats:
Regarding CMake and doc generation, in my experience it never worked correctly:
- The official CMake-level option `-DENABLE_MANUAL=ON` does nothing.
- `-DUSE_MANUAL=ON` CMake option _did_ trigger a rebuild, but it did so unconditionally, so it also rebuilt them when they were already present e.g. from the official source tarball.
- This rebuild results in an empty `tool_hugehelp.c`, so it was unusable anyway.
- It also broke reproducibility by destroying the already present output (from official source tarball) and updating the timestamps (IMO these should retain those of the inputs for reproducibility).
- When generation does work in different sceanrios, the output was different than the source tarball, because it somehow has dependencies of the environment (can't recall the exact details.).
- To workaround these, while still providing docs in curl-for-win I'm using `CPPFLAGS=-DUSE_MANUAL=1`. This skips generation, and with a bunch of more tricks it also works with non-release-tarballs but with empty docs.

https://github.com/curl/curl-for-win/blob/c0383eee039ecf4f449e9912a42dabf7ef5f6c72/curl-cmake.sh#L389-L393
https://github.com/curl/curl-for-win/blob/c0383eee039ecf4f449e9912a42dabf7ef5f6c72/curl-cmake.sh#L443-L448

bagder:
Since #12742, USE_MANUAL at least does something. Even if maybe not 100% correct it builds the curl.1 file.

But yeah, it is certainly not ideal. I have struggled quite a lot to figure out how to generate these files with cmake but I have given up for this time and will move forward with a lacking solution for cmake.

> It also broke reproducibility by overwriting the already present output

Reproducibility should be about the contents not the time stamps, shouldn't it? So it would mostly be a question of using the same fixed date to produce the same fixed output? Setting the `SOURCE_DATE_EPOCH` environment variable when building the file fixes this.

vszakats:
> Since #12742, USE_MANUAL at least does something. Even if maybe not 100% correct it builds the curl.1 file.

Tried it and now both `curl.1` and `tool_hugehelp.c` are generated
with content.

Space differences compared to the release tarball are still present:
Screen Shot 2024-01-21
(left: original, right: as generated locally)

> > It also broke reproducibility by overwriting the already present output
> 
> Reproducibility should be about the contents not the time stamps, shouldn't it? So it would mostly be a question of using the same fixed date to produce the same fixed output? Setting the `SOURCE_DATE_EPOCH` environment variable when building the file fixes this.

It's about both, current timestamps can end up in release tarballs and make them different. It's one nasty part of reproducible builds. Thanks for the tip about `SOURCE_DATE_EPOCH`, I wasn't aware these scripts supported it.

bagder:
Right, those space differences are caused by nroff so not really a cmake vs autotools thing. Slightly related to this: I have a separate thing in progress that might lead to us soon being able to drop nroff from the building of that source file.

levitte:
@bagder caught my interest in this, specifically for generating the documentation with cmake.  I have a proposed patch, will generate a PR against this branch in a few.

levitte:
@bagder, you might be interested in the installation instructions I added in my PR as well.

bagder:
I believe this is PR working cmake wise with @levitte's fixes but it still fails four tests on appveyor due to windows-related reasons, such as no shell expansion on wildcards and something else: 1140 1173 1177 1477

It will disable them on appveyor and move forward if that is the only remaining nit.

scripts/cd2nroff
```diff
@@ -0,0 +1,328 @@
...
+        print STDERR "$f:$line:1:ERROR: no header present\n";
+        return 2;
+    }
+
+    my @desc;
+    my $quote = 0;
+    my $blankline = 0;
+    my $header = 0;
+
+    push @desc, ".\\\" generated by cd2nroff $cd2nroff from $f\n";
```
vszakats, line 189: `$f` remains empty in a local test case:
```man
.\" generated by cd2nroff 0.1 from 
.TH CURLINFO_APPCONNECT_TIME 3 "December 14 2023" libcurl
```

bagder: what is the full command line used when `cd2nroff` is invoked?

bagder: Oh, I see why

bagder:
~~~diff
diff --git a/docs/libcurl/CMakeLists.txt b/docs/libcurl/CMakeLists.txt
index fc1283a53..0d52a7e56 100644
--- a/docs/libcurl/CMakeLists.txt
+++ b/docs/libcurl/CMakeLists.txt
@@ -34,11 +34,11 @@ function(add_manual_pages _listname)
     else()
       string(REPLACE ".3" ".md" _mdfile "${CMAKE_CURRENT_SOURCE_DIR}/${_file}")
     endif()
 
     add_custom_command(OUTPUT "${_rofffile}"
-      COMMAND ${PROJECT_SOURCE_DIR}/scripts/cd2nroff < ${_mdfile} > ${_rofffile}
+      COMMAND ${PROJECT_SOURCE_DIR}/scripts/cd2nroff ${_mdfile} > ${_rofffile}
       DEPENDS "${_mdfile}"
       VERBATIM
     )
   endforeach()
 endfunction()

bagder:
They are built with autotools by default and there is no option to switch that off...

scripts/cd2nroff
vszakats: They look like this:
```
cd ./curl/_x64-win-ucrt-cmake-llvm-bld/docs/libcurl/opts && \
  ./curl/scripts/cd2nroff \
  < ./curl/docs/libcurl/opts/CURLINFO_APPCONNECT_TIME.md \
  > ./curl/_x64-win-ucrt-cmake-llvm-bld/docs/libcurl/opts/CURLINFO_APPCONNECT_TIME.3
```

bagder:
On my machine, building all libcurl ~500 man pages takes 800 milliseconds. It is a pretty quick operation, even if my machine might be above average CPU and disk speed wise.

scripts/cd2nroff
bagder: Right, my diff above changes that command line, to pass the file name instead `< [filename]` which is what made it blank there.

vszakats:
I takes almost as long as the build itself here, but I'm on fairly old gear. I'd appreciate an option to turn this off, as this will increase greatly the time for each build. When doing them by the hundreds, it adds up. (Even building curl.1/hugehelp is takes a fair amount, but I can tweak that locally.) For some reason autotools never built man pages with the curl-for-win script, I don't know why, with `--enable-manual` set.

bagder:
Fair enough!

vszakats:
Re: autotools, it did have this working before this patch, but was unnoticably fast (being just a handful of `install` commands). It may be noticable after this patch. Making a `--disable-docs` option useful there as well.

But, for now autotools builds quit with this error (with this PR applied), so I couldn't actually test it:
```
make[3]: *** No rule to make target `curl-config.1', needed by `all-am'.  Stop.
```

bagder:
> Edit: with this PR applied, autotools quits with:

We use autotools in numerous CI builds, including windows (on appveyor) and none of them fail like that... I'm not saying there is no problem left, but I might not consider that a blocker for merging this.

scripts/cd2nroff
vszakats: Thanks! That fixed it ...but now there is another issue: The filename embedded contains the whole (absolute) path of the source filename, which breaks reproducibility and may also leak local info.

```man
.\" generated by cd2nroff 0.1 from /<home-dir>/<path-to>/curl/docs/libcurl/opts/CURLINFO_CONNECT_TIME_T.md
.TH CURLINFO_CONNECT_TIME_T 3 "December 14 2023" libcurl
```

Perhaps the simplest fix is to omit this info from the generated output? Or strip all path from it?

bagder: It does not break reproducibility, you just have to rebuild with the same paths. But yes, I'll make the script cut off the dir name. I want the file name mentioned there to help users find the source.

bagder:
This is the biggest PR I/we have done in a long time. I completely expect that there will be some issues popping up once we merge this, but I don't think they are worse than we can just fix them. I believe this is a good PR.

vszakats:
No worries, re-run it in a clean sandbox, and that resolved it.

scripts/cd2nroff
vszakats: Patch to drop the source filename:
```diff
diff --git a/scripts/cd2nroff b/scripts/cd2nroff
index 08fbd768e..734cbf8f3 100755
--- a/scripts/cd2nroff
+++ b/scripts/cd2nroff
@@ -186,7 +186,7 @@ sub single {
     my $blankline = 0;
     my $header = 0;
 
-    push @desc, ".\\\" generated by cd2nroff $cd2nroff from $f\n";
+    push @desc, ".\\\" generated by cd2nroff $cd2nroff\n";
     push @desc, ".TH $title $section \"$date\" $source\n";
     while(<$fh>) {
         $line++;
```

To strip path:
```diff
diff --git a/scripts/cd2nroff b/scripts/cd2nroff
index 08fbd768e..d9938b6e9 100755
--- a/scripts/cd2nroff
+++ b/scripts/cd2nroff
@@ -30,6 +30,8 @@ Converts a curldown file to nroff (man page).
 =end comment
 =cut
 
+use File::Basename;
+
 my $cd2nroff = "0.1"; # to keep check
 my $dir;
 my $extension;
@@ -186,7 +188,7 @@ sub single {
     my $blankline = 0;
     my $header = 0;
 
-    push @desc, ".\\\" generated by cd2nroff $cd2nroff from $f\n";
+    push @desc, ".\\\" generated by cd2nroff $cd2nroff from " . basename($f) . "\n";
     push @desc, ".TH $title $section \"$date\" $source\n";
     while(<$fh>) {
         $line++;
```

gvanem:
This `scripts/cd2nroff` file yet is another Windows unfriendly Perl script IMHO.
My Cygwin Perl bitches on this command:
```
f:/Cygwin64/bin/perl.exe ../scripts/cd2nroff ../docs/libcurl/ABI.md > ../docs/libcurl/ABI.3
../docs/libcurl/ABI.md:68:1:ERROR: no header present
```

Bit this works fine:
```
f:/Cygwin64/bin/perl.exe ../scripts/cd2nroff < ../docs/libcurl/ABI.md > ../docs/libcurl/ABI.3
```

And so does my good old StrawBerry Perl (v5.18.4 from 2014). 
A newline-issue on `if(/^---/) {` once again. Arg!

gvanem:
>  several Windows CI jobs

Since all those CI jobs checks out with `git` and `config.autocrlf=false` presumably?
I do not  know muchPerl, but this `open($fh, "<:crlf", "$f")` opens in text-mode?

bagder:
> Since all those CI jobs checks out with git and config.autocrlf=false presumably?

No, I can't see anything that sets `config.autocrlf=false` or sets it specifically for `.md` files. Maybe we should rather set them to `eol=lf` ?

>  I do not know muchPerl, but this open($fh, "<:crlf", "$f") opens in text-mode?

Whenever I try to guess how something works with text files on Windows, I seem to guess wrong. So I don't know.

gvanem:
Another thing I feel is strange. From the bottom of e.g. `man CURLOPT_RESUME_FROM`:
curl-man

English text, but a Norwegian date (!)

bagder:
I bet you can fix that by changing/setting some locale environment variable(s)...

gvanem:
> ... some locale environment variable(s)...

Sure. `LC_ALL=C` or something? But `man curl` looks what I'd expect:
curl-man

PR #12732, opened 2024-01-17
fix misspelling

Man2Dev:


bagder:
Thanks, I had this fixed locally as well, and it is now part of the updated PR.

PR #12740, opened 2024-01-19
transfer: make the select_bits_paused condition check both directions

bagder:
If there is activity in a direction that is not paused, return false.

Reported-by: Sergey Bronnikov
Bug: https://curl.se/mail/lib-2024-01/0049.html


icing (approved):
That looks about right after I wrapped by head around it.

PR #12741, opened 2024-01-19
TODO: update broken link

Man2Dev:
original URL leads to "HTTP Status Code: 404 Page Not Found" page


bagder:
The draft has since changed name and this seems to link to the general page for the drafts, don't you think is a better URL? https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/

Man2Dev:
yes, this would be better. I will fix it right now

PR #12742, opened 2024-01-19
cmake: when USE_MANUAL=YES, build the curl.1 man page

bagder:
Fixes KNOWN_BUG 15.4


bagder:
If this works, we can consider enabling this by default later on.

PR #12745, opened 2024-01-20
url: notify unexpected input pending on live connection

jay:
- When pruning dead connections from the connection cache show a verbose message if a pruned connection was actually alive but had unexpected input pending and therefore was marked as dead.

Ref: https://github.com/curl/curl/discussions/12715#discussioncomment-8152360

Closes #xxxx


jay:
Well I can't decide on the wording for this but I think saying a connection with unexpected input is being "treated as" dead is probably accurate enough.

bagder:
I think this just risks being more or just differently misleading. When a connection closes, it becomes readable. Then the mere readability signals that it is dead. We don't actually know if there is "input" or not, so saying that there is might make users believe "input" means data.

jay:
> I think this just risks being more or just differently misleading. When a connection closes, it becomes readable. Then the mere readability signals that it is dead. We don't actually know if there is "input" or not, so saying that there is might make users believe "input" means data.

Yes I suppose it could be misinterpreted. hm.

jay:
abandoning this since it may confuse the user

PR #12746, opened 2024-01-20
url: show connection id in connection max-age verbose messages

jay:
.. for consistency with other verbose messages shown during the extraction of dead connections.

Closes #xxxx


bagder:
Doesn't `--trace-ids` already make this visible?

bagder:
`--trace-ids` already provide this functionality, for all messages.

PR #12757, opened 2024-01-22
pingpong: stop using the download buffer

bagder:
The pingpong logic now uses its own dynbuf for receiving command response data.


bagder:
Still 5 failing checks but they are all AppVeyor. It is concerning that they fail on 1206 and 1207, which are two pingpong-using tests and thus changed with this PR. I can't make that test fail in my local environment though...

bagder:
That seems to have been the final nit, the only CI fail is now unrelated to this change. I will merge.

PR #12758, opened 2024-01-22
CI, remove unnecessary `enable-tls13`

icing:
and specify libdir for smaller curl config. Just a tidy up.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -105,19 +105,19 @@ jobs:
           - name: openssl3
             install_packages: zlib1g-dev valgrind
             install_steps: gcc-11 openssl3 pytest
-            configure: CFLAGS=-std=gnu89 LDFLAGS="-Wl,-rpath,$HOME/openssl3/lib64" --with-openssl=$HOME/openssl3 --enable-debug --enable-websockets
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Is this "draft" on purpose/for a reason?

icing:
> Is this "draft" on purpose/for a reason?

I started doing drafts to save you time while I check if it is really working.😌

jay:
I'm going to land this but not with the rpath removals because that causes the system ossl to be loaded

ossl rpath:
* libcurl/8.6.0-DEV OpenSSL/3.1.3 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libpsl/0.21.0

ossl no rpath:
* libcurl/8.6.0-DEV OpenSSL/3.0.2 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libpsl/0.21.0

jay:
It seems to me the GHA cache of the OpenSSL 3.1.3 build would be affected so I [deleted it](https://github.com/curl/curl/actions/caches) from the cache before landing this PR in master. I will check the related jobs to make sure it's regenerated and succeeds.

PR #12762, opened 2024-01-23
cmake: speed up curldown processing, enable by default

vszakats:
- cmake: enable `BUILD_DOCS` by default (this controls converting and
  installing `.3` files from `.md` sources)

- cmake: speed up generating `.3` files by using a single command per
  directory, instead of a single command per file. This reduces external
  commands by about a thousand. (There remains some CMake logic kicking
  in resulting in 500 -one per file- external `-E touch_nocreate` calls.)

- cd2nroff: add ability to process multiple input files.

- cd2nroff: add `-k` option to use the source filename to form the
  output filename. (instead of the default in-file `Title:` line.)

Closes #12762

### DIFF
--- a/scripts/cd2nroff
+++ b/scripts/cd2nroff
@@ -30,8 +30,6 @@ Converts a curldown file to nroff (man page).
 =end comment
 =cut
 
-use File::Basename;
-
 my $cd2nroff = "0.1"; # to keep check
 my $dir;
 my $extension;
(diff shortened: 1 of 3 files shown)


scripts/cd2nroff
```diff
@@ -189,7 +189,7 @@ sub single {
     # cut off the leading path from the file name, if any
     $f =~ s/^(.*[\\\/])//;
 
-    push @desc, ".\\\" generated by cd2nroff $cd2nroff from $f\n";
+    push @desc, ".\\\" generated by cd2nroff $cd2nroff\n";
```
bagder, line 192: Two lines above it removes the path from the name, how can it be absolute?

vszakats: Oh sorry, this has changed compared to the last tested version before going to sleep. I'm deleting this.

vszakats:
Please double-check my broken Perl-fu.

PR #12771, opened 2024-01-24
cmake: better logic around docs building

levitte:
- make sure that docs are built when BUILD_DOCS is enabled
- have USE_MANUAL depend on BUILD_DOCS too, and display a warning when it's
  not enabled in spite of ENABLE_MANUAL being true


CMakeLists.txt
```diff
@@ -309,10 +309,15 @@ find_package(Perl)
...
-    set(USE_MANUAL ON)
+if(ENABLE_MANUAL)
+  if(BUILD_DOCS AND PERL_FOUND)
+    curl_nroff_check()
+    if(NROFF_USEFUL)
+      set(USE_MANUAL ON)
+    endif()
+  endif()
+  if(NOT USE_MANUAL)
+    message(WARNING "BUILD_DOCS disabled, or perl not found, or nroff not useful.  Will not provide built-in manual")
```
vszakats, line 320: This seems to also disable building curl.1 and tool_hugehelp.c (aka the 'manual') when
`BUILD_DOCS=OFF`. The point of adding `BUILD_DOCS` was to have a separate
control for the rest of docs (the individual man pages.), which this patch appears to 
break.

levitte: That's quite confusing, especially since `BUILD_DOCS` is documented as "to build manual pages", with no other qualification...  so what makes curl.1 a manual page that gets built even though building docs has been disabled.  This makes no sense at all.

We can see it another way, though: if it were me, I would remove `BUILD_DOCS`, i.e. always build the docs.  Would that be a suitable option?

vszakats: The wording is confusing indeed, but note that up until a few days ago none of these options worked correctly with CMake.

We have a manual page (and a built-in copy) for the curl tool and manual pages for libcurl.

I added `BUILD_DOCS` because building them took a lot of time. I've made it faster, but it still takes quite a bit of extra time. Also some builds (e.g. curl-for-win) doesn't distribute or need the man pages, so building them is just wasting CI time. I'd like to keep this option. It's also enabled by default.

Here's my proposal to separate these two settings, and to make the wording clearer: #12773

vszakats:
A modification of similar intent was merged via #12773, so I think it's safe to close this now.

PR #12773, opened 2024-01-24
cmake: rework options to enable curl and libcurl docs

vszakats:
Rework CMake options for building/using curl tool and libcurl manuals.

- rename `ENABLE_MANUAL` to `ENABLE_CURL_MANUAL`, meaning:
  to build man page and built-in manual for curl tool.

- rename `BUILD_DOCS` to `BUILD_LIBCURL_DOCS`, meaning:
  to build man pages for libcurl.

- `BUILD_LIBCURL_DOCS` now works without having to enable
  `ENABLE_CURL_MANUAL` too.

- drop support for existing CMake-level `USE_MANUAL` option to avoid
  confusion. (It used to work with the effect of current
  `ENABLE_CURL_MANUAL`, but only by accident.)

Ref: #12771
Closes #12773


CMakeLists.txt
```diff
@@ -306,13 +306,18 @@ endif()
...
-  if(NROFF_USEFUL)
-    set(USE_MANUAL ON)
+option(BUILD_DOCS "to build libcurl man pages" ON)
+option(ENABLE_MANUAL "to provide the built-in manual and man page for curl tool" OFF)
+
+if(ENABLE_MANUAL OR BUILD_DOCS)
+  if(PERL_FOUND)
+    curl_nroff_check()
+    if(NROFF_USEFUL)
+      set(HAVE_MANUAL_TOOLS ON)
```
levitte, line 316:
```suggestion
      set(HAVE_MANUAL_TOOLS ON)
      set(USE_MANUAL ON)
```

CMakeLists.txt
```diff
@@ -306,13 +306,18 @@ endif()
...
+    endif()
+  endif()
+  if(NOT HAVE_MANUAL_TOOLS)
+    message(WARNING "Perl not found, or nroff not useful. Will not provide built-in manual.")
```
levitte, line 320: Since this stops *any* documentation from being built, I'd say that.  So:

```suggestion
    message(WARNING "Perl not found, or nroff not useful. Will not build manuals.")
```

lib/curl_config.h.cmake
```diff
@@ -698,9 +698,6 @@ ${SIZEOF_TIME_T_CODE}
 /* if libPSL is in use */
 #cmakedefine USE_LIBPSL 1
 
-/* If you want to build curl with the built-in manual */
-#cmakedefine USE_MANUAL 1
-
```
levitte, line 703: .. undo this

src/CMakeLists.txt
```diff
@@ -24,7 +24,8 @@
 set(EXE_NAME curl)
 add_definitions(-DBUILDING_CURL)
 
-if(USE_MANUAL)
+if(HAVE_MANUAL_TOOLS AND ENABLE_MANUAL)
+  add_definitions("-DUSE_MANUAL")
```
levitte, line 28: ... undo this

levitte (commented):
If I may, I think that it's still useful to have USE_MANUAL up front, including in curl_config.h.  I believe these changes acheive that

CMakeLists.txt
levitte: This was supposed to be part of my proposal below

vszakats: `USE_MANUAL` isn't use any longer by CMake after this patch. The necessary C macro is defined in `src/CMakeList.txt`, so the above is unnecessary unless I missed something.

levitte:
I gotta admit, I still find it a bit weird that curl.1 wouldn't be built as part of building the docs, regardless of if `ENABLE_MANUAL` is set or not.

However, there is also `BUILD_CURL_EXE`, and I can fully understand that if this is disabled, building the corresponding manual would be a bit weird.  Could it be that this and `ENABLE_MANUAL` are a bit confounded?

CMakeLists.txt
vszakats: Updated the commit message to reflect that.

levitte: See my proposal below, where I suggest restoring its definition in curl_config.h.

vszakats: Why do you think that would be better?

It'd be two lines vs. one, and a `set(USE_MANUAL ON)` doesn't
readily conveys that we mean this for `curl_config.h`, and only
for that.

vszakats:
`curl.1` comes with the official source release tarball, so anyone using that
doesn't need `ENABLE_MANUAL=ON` to have `curl.1`. Same for `tool_hugehelp.c`.

`ENABLE_MANUAL=ON` has the effect of unconditionally rebuilding (I haven't
retested that after recent changes) `curl.1`, which breaks reproducibiliy in some
build envs. Not enabling it by default is a workaround for that.

We can rename `BUILD_DOCS` to `BUILD_LIBCURL_DOCS` to better reflect 
its function. (possibly even `ENABLE_MANUAL` to `ENABLE_CURL_MANUAL`.)

As for `BUILD_CURL_EXE`, what about doing this?:
```diff
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -25,6 +25,6 @@
 if(BUILD_DOCS)
   add_subdirectory(libcurl)
 endif()
-if(ENABLE_MANUAL)
+if(ENABLE_MANUAL AND BUILD_CURL_EXE)
   add_subdirectory(cmdline-opts)
 endif()
```

levitte:
> We can rename `BUILD_DOCS` to `BUILD_LIBCURL_DOCS` to better reflect its function. (possibly even `ENABLE_MANUAL` to `ENABLE_CURL_MANUAL`.)

That would be much less confusing, so yes please

> As for `BUILD_CURL_EXE`, what about doing this?:
> 
> ```diff
> --- a/docs/CMakeLists.txt
> +++ b/docs/CMakeLists.txt
> @@ -25,6 +25,6 @@
>  if(BUILD_DOCS)
>    add_subdirectory(libcurl)
>  endif()
> -if(ENABLE_MANUAL)
> +if(ENABLE_MANUAL AND BUILD_CURL_EXE)
>    add_subdirectory(cmdline-opts)
>  endif()
> ```

Looks good to me

CMakeLists.txt
levitte: Most of all, since the `USE_MANUAL` macro is checked in many places in the source, it's kinda odd not to have it being part of the collection of other `USE_` macros.

vszakats: It's used in 4 places, all inside `src`. Defining it via CMake and `lib/curl_config.h.cmake` makes the logic more difficult to follow IMO. It would also allow to set this accidentally and breaking the build in weird ways. Esp considering that `cmake -DUSE_MANUAL=ON` was a workaround for the broken `-DENABLE_MANUAL=ON` option for all previous releases. Also having the C-level macro set for `lib` is unncessary and a source for confusion.

vszakats:
Thanks, committed these updates.

src/CMakeLists.txt
```diff
@@ -24,7 +24,8 @@
 set(EXE_NAME curl)
 add_definitions(-DBUILDING_CURL)
 
-if(USE_MANUAL)
+if(ENABLE_CURL_MANUAL AND HAVE_MANUAL_TOOLS)
+  add_definitions("-DUSE_MANUAL")
```
vszakats, line 28: Notice this is the only valid case we should set this C-level macro. (Missing this was causing the initial CI failures with builds having `BUILD_DOCS=ON` + `ENABLE_MANUAL=OFF`.

vszakats:
> `curl.1` comes with the official source release tarball, so anyone using that doesn't need `ENABLE_MANUAL=ON` to have `curl.1`. Same for `tool_hugehelp.c`.
> 
> `ENABLE_MANUAL=ON` has the effect of unconditionally rebuilding (I haven't retested that after recent changes) `curl.1`, which breaks reproducibiliy in some build envs. Not enabling it by default is a workaround for that (and not doing this extra step for release builds, which is presumably _most_ builds out there.).

Reviewing curl-for-win script again, this remains having another
downside: To avoid rebuilding existing `curl.1` / `tool_hugehelp.c`
yet still *embedding* it into curl, the CPPFLAGS `-DUSE_MANUAL` 
must be passed manually. This was so before this week's changes
and still true.

Doing the rebuild only when `curl.1` (and/or `tool_hugehelp.c`)
are missing or outdated would be one solution; making the output
deterministic would be another. This would require dropping `nroff`,
which is on bagder's plans. Ideally both.
C `USE_MANUAL` also has the requirement that `tool_hugehelp.c`
does actually offer a `hugehelp` symbol in it, otherwise the build 
breaks. Using `src/tool_hugehelp.c.cvs` satisfies this, but not the 
empty stub generated by CMake (or autotools for that matter).

Perhaps an approach where we have a static `tool_hugehelp.c` in
the source tree, while expecting the manual content inside an 
generated `tool_hugehelp.h` (depending on `USE_MANUAL`)
could help to avoid some failure modes. But touching any of that looks
like a deep rabbit-hole.

In any case this is left for a separate PR.

CMakeLists.txt
levitte: You know, I'm stepping down from this...  it's starting to become a bike shed, if you know what I mean, and not really that important.

levitte:
Re this discussion, I clearly don't have a deeper understanding of all the ramifications, so I defer to your knwoledge.

CMakeLists.txt
vszakats: I do, no worries and thanks for your feedback!

PR #12783, opened 2024-01-25
appveyor: set the job "version" to use the branch name

bagder:
... might make the field a tad bit more useful.


bagder:
Not really, it then just uses 'master'

PR #12788, opened 2024-01-25
socks using its own buffer instead of data->state.buffer

icing:


bagder:
> docs/libcurl/libcurl-symbols.md

That's a generated file... :grimacing:

lib/socks.c
```diff
@@ -74,6 +74,7 @@ enum connect_t {
 struct socks_state {
   enum connect_t state;
   ssize_t outstanding;  /* send this many bytes more */
+  unsigned char buffer[8*1024]; /* more than enough */
```
bagder, line 77: How did you pick the size?

icing: I have a crystal skull next to my dropping candles here in my lair.

As I read the code, one socks packets always seem to be smaller than 1k, but I might be wrong...therefore the assertions.

icing: I started replacing the fixes buffer with a bufq, but that quickly escalated all over the file and I was not sure if it would be worth it.

bagder:
Line 434 in the original, 435 in your update, references `data->set.buffer_size` to check if the data fits, which is now wrong.

lib/socks.c
bagder: I believe username, password and hostname are all limited to 256 bytes max, so 1K should probably be enough.

icing:
> Line 434 in the original, 435 in your update, references `data->set.buffer_size` to check if the data fits, which is now wrong.

Nice catch! Push a fix.

bagder:
line 288: DEBUGASSERT(READBUFFER_MIN >= 600);

This check verifies the the smallest download buffer is big enough. It  should now probably check that `buffer` is at least this big.

bagder:
I did #12797 to help us verify that a ~600 byte buffer will work fine, but we can also adjust the size after the initial merge.

PR #12789, opened 2024-01-25
ftp+file, use stack buffer for resume handling

icing:
instead of data->state.buffer


bagder:
file.c still uses the state.buffer for downloads in `file_do`, did you mean to leave that part out?

icing:
> file.c still uses the state.buffer for downloads in `file_do`, did you mean to leave that part out?

Added now.

PR #12790, opened 2024-01-25
sasl: use http authentication mechanisms as a default, not a preset

### LINKED ISSUE #10259: SMTP Auth is not using XOAUTH2 even when specified with --login-options 'AUTH=XOAUTH2' \
### I did this
I want to send an E-Mail over SMTP using OAUTH2.

```
curl --ssl-reqd -v \
--url 'smtp://smtp.gmail.com' \
--user $USERNAME \
--login-options 'AUTH=XOAUTH2' \
--oauth2-bearer $ACCESSTOKEN \
--mail-from from@example.org \
--mail-rcpt to@example.org \
--upload-file mail.txt
```

For GMail this fails with 
```
< 250-smtp.gmail.com at your service, [2001:9e8:3c3:7c00:8d8:c156:791a:89a]
< 250-SIZE 35882577
< 250-8BITMIME
< 250-AUTH LOGIN PLAIN XOAUTH2 PLAIN-CLIENTTOKEN OAUTHBEARER XOAUTH
< 250-ENHANCEDSTATUSCODES
< 250-PIPELINING
< 250-CHUNKING
< 250 SMTPUTF8
} [5 bytes data]
> AUTH OAUTHBEARER
{ [5 bytes data]
< 334
} [5 bytes data]
> TOKEN
{ [5 bytes data]
< 334 eyJzdGF0dXMiOiJpbnZhbGlkX3JlcXVlc3QiLCJzY29wZSI6Imh0dHBzOi8vbWFpbC5nb29nbGUuY29tLyJ9 -> {"status":"invalid_request","scope":"https://mail.google.com/"}
} [5 bytes data]
> AQ==
{ [5 bytes data]
< 535-5.7.8 Username and Password not accepted. Learn more at
< 535 5.7.8  https://support.google.com/mail/?p=BadCredentials b6-20020aa7d486000000b0048447efe3fcsm2477734edr.84 - gsmt
curl: (67) Login denied
```
 
The bearer token is sent with `AUTH OAUTHBEARER` and formatted according to the format on line 72:
https://github.com/curl/curl/blob/5a9a5e171e186da14dc9d209a17304ed03c6cac8/lib/vauth/oauth2.c#L68-L73

Which should be supported according to: `< 250-AUTH LOGIN PLAIN XOAUTH2 PLAIN-CLIENTTOKEN OAUTHBEARER XOAUTH` 
But since I specified `--login-options 'AUTH=XOAUTH2'` this method should be used.

### I expected the following
According to Google the bearer token should be sent with `AUTH XOAUTH2`
https://developers.google.com/gmail/imap/xoauth2-protocol

And this token format should be used:
https://github.com/curl/curl/blob/5a9a5e171e186da14dc9d209a17304ed03c6cac8/lib/vauth/oauth2.c#L95-L100

Microsoft describes the same standard.
https://learn.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth
According to this site these are the main provider using SASL XOAUTH2 as the authentication mechanism:
https://mailtrap.io/blog/smtp-auth/

Since I never got it working, I don't know if this is the reason for the login the problem. But curl should use the specified login mechanism.

Thank you for all your work.

### curl/libcurl version

```
$ curl -V
curl 7.86.0 (x86_64-w64-mingw32) libcurl/7.86.0 OpenSSL/1.1.1s (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.3 libssh2/1.10.0 nghttp2/1.51.0
Release-Date: 2022-10-26
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe TLS-SRP UnixSockets zstd
```

**AND**

```
$version
[1] "7.64.1"

$ssl_version
[1] "(OpenSSL/1.1.1k) Schannel"

$libz_version
[1] "1.2.12"

$libssh_version
[1] "libssh2/1.9.0"

$libidn_version
[1] NA

$host
[1] "x86_64-w64-mingw32"

$protocols
 [1] "dict"   "file"   "ftp"    "ftps"   "gopher" "http"   "https"  "imap"   "imaps"  "ldap"   "ldaps"  "pop3"   "pop3s" 
[14] "rtsp"   "scp"    "sftp"   "smtp"   "smtps"  "telnet" "tftp"  

$ipv6
[1] TRUE

$http2
[1] FALSE

$idn
[1] TRUE
```

### operating system
Window 10 22H2 Build 19045.2364

monnerat:
If some are given, sasl maps http authentication options to sasl ones and merges them to the login options.
This may cause problems with the cli tool that sets the http login option CURLAUTH_BEARER as a side-effect of --oauth2-bearer, because this flag maps to more than one sasl mechanisms and the latter cannot be cleared individually by the login options string.
The solution retained here is to consider http authentication options as a default (overriding the hardcoded default mask for the protocol) that is ignored if a login option string is given.

New test 992 checks this.

Fixes #10259

### DIFF
--- a/lib/curl_sasl.c
+++ b/lib/curl_sasl.c
@@ -205,18 +205,23 @@ void Curl_sasl_init(struct SASL *sasl, struct Curl_easy *data,
   sasl->force_ir = FALSE;          /* Respect external option */
 
   if(auth != CURLAUTH_BASIC) {
-    sasl->resetprefs = FALSE;
-    sasl->prefmech = SASL_AUTH_NONE;
+    unsigned short mechs = SASL_AUTH_NONE;
+
+    /* If some usable http authentication options have been set, determine
+       new defaults from them. */
     if(auth & CURLAUTH_BASIC)
-      sasl->prefmech |= SASL_MECH_PLAIN | SASL_MECH_LOGIN;
+      mechs |= SASL_MECH_PLAIN | SASL_MECH_LOGIN;
     if(auth & CURLAUTH_DIGEST)
-      sasl->prefmech |= SASL_MECH_DIGEST_MD5;
+      mechs |= SASL_MECH_DIGEST_MD5;
     if(auth & CURLAUTH_NTLM)
-      sasl->prefmech |= SASL_MECH_NTLM;
+      mechs |= SASL_MECH_NTLM;
     if(auth & CURLAUTH_BEARER)
-      sasl->prefmech |= SASL_MECH_OAUTHBEARER | SASL_MECH_XOAUTH2;
+      mechs |= SASL_MECH_OAUTHBEARER | SASL_MECH_XOAUTH2;
     if(auth & CURLAUTH_GSSAPI)
-      sasl->prefmech |= SASL_MECH_GSSAPI;
+      mechs |= SASL_MECH_GSSAPI;
+
+    if(mechs != SASL_AUTH_NONE)
+      sasl->prefmech = mechs;
   }
 }
 
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -125,7 +125,7 @@ test952 test953 test954 test955 test956 test957 test958 test959 test960 \
 test961 test962 test963 test964 test965 test966 test967 test968 test969 \
 test970 test971 test972 test973 test974 test975 test976 test977 test978 \
 test979 test980 test981 test982 test983 test984 test985 test986 test987 \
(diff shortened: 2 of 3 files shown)


jay:
Thanks

PR #12791, opened 2024-01-25
http, use stack scratch buffer

icing:
- instead of data->state.buffer


bagder:
Remove `docs/libcurl/libcurl-symbols.md` from this!

bagder (approved):

PR #12795, opened 2024-01-25
git ignore our new generated friend `libcurl-symbols.md`

icing:


bagder:
It needs REUSE attention

icing:
> It needs REUSE attention

Added.

PR #12798, opened 2024-01-26
socks: reduce the buffer size to 600 (from 8K)

bagder:
This is malloc'ed memory and it does not more. Test 742 helps us verify this.


bagder:
Closed by a3a940702a7f (but it used the wrong PR number in the commit)

PR #12799, opened 2024-01-26
mqtt, remove remaining use of data->state.buffer

icing:


bagder:
scan-build suspects something:
~~~
mqtt.c:577:13: warning: Assigned value is garbage or undefined [core.uninitialized.Assign]
    encoded = buf[i];

icing:
> scan-build suspects something:
> 
> ```
> mqtt.c:577:13: warning: Assigned value is garbage or undefined [core.uninitialized.Assign]
>     encoded = buf[i];
> ```

A correct analysis. The buffer there needs to survive calls. Pushed a fix.

bagder (approved):

PR #12801, opened 2024-01-26
vtls, receive max buffer

icing:
- do not only receive one TLS record, but try to fill the passed buffer
- consider <4K remaning space is "filled".


bagder (approved):
Lets do this

PR #12803, opened 2024-01-26
krb5: ignore clang warnings on mvsnprintf()

bagder:
"error: format string is not a string literal"

### DIFF
--- a/lib/krb5.c
+++ b/lib/krb5.c
@@ -435,6 +435,9 @@ static char level_to_char(int level)
 /* Send an FTP command defined by |message| and the optional arguments. The
    function returns the ftp_code. If an error occurs, -1 is returned. */
 static int ftp_send_command(struct Curl_easy *data, const char *message, ...)
+  CURL_PRINTF(2, 3);
+
+static int ftp_send_command(struct Curl_easy *data, const char *message, ...)
 {
   int ftp_code;
   ssize_t nread = 0;


vszakats:
Proposing this, to have the local printf wrapper verified for format issues:
```diff
--- a/lib/krb5.c
+++ b/lib/krb5.c
@@ -434,6 +434,9 @@ static char level_to_char(int level)
 
 /* Send an FTP command defined by |message| and the optional arguments. The
    function returns the ftp_code. If an error occurs, -1 is returned. */
+static int ftp_send_command(struct Curl_easy *data, const char *message, ...)
+  CURL_PRINTF(2, 3);
+
 static int ftp_send_command(struct Curl_easy *data, const char *message, ...)
 {
   int ftp_code;
```

bagder:
Much better!

vszakats:
Oops, I meant the `CURL_PRINTF()` patch as an extra over the initial patch. The warning suppression is still necessary. Sorry for being unclear.

bagder:
I don't get any warnings with clang anymore after this!

vszakats:
That's interesting. It seems the compiler is clever enough to apply this knowledge to the function itself. It also means that some of these warning suppressions added earlier (called out "looking shit" in a comment), may be redundant after all.

vszakats:
Following-up on this: #12812

Turns out most of them became redundant after adding `CURL_PRINTF()`.
Could also eliminate two more with minor code changes.

PR #12805, opened 2024-01-26
Add xfer_buf to multi handle

icing:
- can be borrowed by transfer during recv-write operation
- needs to be released before borrowing again
- adjustis size to `data->set.buffer_size`
- used in transfer.c readwrite_data()

### DIFF
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -395,7 +395,6 @@ bool Curl_conncache_return_conn(struct Curl_easy *data,
          important that details from this (unrelated) disconnect does not
          taint meta-data in the data handle. */
       struct conncache *connc = data->state.conn_cache;
-      connc->closure_handle->state.buffer = data->state.buffer;
       connc->closure_handle->set.buffer_size = data->set.buffer_size;
       Curl_disconnect(connc->closure_handle, conn_candidate,
                       /* dead_connection */ FALSE);
@@ -522,11 +521,9 @@ Curl_conncache_extract_oldest(struct Curl_easy *data)
 void Curl_conncache_close_all_connections(struct conncache *connc)
 {
   struct connectdata *conn;
-  char buffer[READBUFFER_MIN + 1];
   SIGPIPE_VARIABLE(pipe_st);
   if(!connc->closure_handle)
     return;
-  connc->closure_handle->state.buffer = buffer;
   connc->closure_handle->set.buffer_size = READBUFFER_MIN;
 
   conn = conncache_find_first_connection(connc);
@@ -541,7 +538,6 @@ void Curl_conncache_close_all_connections(struct conncache *connc)
     conn = conncache_find_first_connection(connc);
   }
 
-  connc->closure_handle->state.buffer = NULL;
   sigpipe_ignore(connc->closure_handle, &pipe_st);
 
   Curl_hostcache_clean(connc->closure_handle,
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -395,7 +395,6 @@ bool Curl_conncache_return_conn(struct Curl_easy *data,
          important that details from this (unrelated) disconnect does not
          taint meta-data in the data handle. */
       struct conncache *connc = data->state.conn_cache;
-      connc->closure_handle->set.buffer_size = data->set.buffer_size;
       Curl_disconnect(connc->closure_handle, conn_candidate,
                       /* dead_connection */ FALSE);
     }
@@ -524,7 +523,6 @@ void Curl_conncache_close_all_connections(struct conncache *connc)
   SIGPIPE_VARIABLE(pipe_st);
   if(!connc->closure_handle)
     return;
-  connc->closure_handle->set.buffer_size = READBUFFER_MIN;
 
   conn = conncache_find_first_connection(connc);
   while(conn) {
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -784,7 +784,6 @@ static CURLcode multi_done(struct Curl_easy *data,
       data->state.lastconnect_id = -1;
   }
 
-  Curl_safefree(data->state.buffer);
   return result;
 }
 
@@ -1891,12 +1890,9 @@ static CURLcode readrewind(struct Curl_easy *data)
  */
 CURLcode Curl_preconnect(struct Curl_easy *data)
 {
-  if(!data->state.buffer) {
-    data->state.buffer = malloc(data->set.buffer_size + 1);
-    if(!data->state.buffer)
-      return CURLE_OUT_OF_MEMORY;
-  }
-
+  /* this used to do data->state.buffer allocation,
+     maybe remove completely now? */
+  (void)data;
   return CURLE_OK;
 }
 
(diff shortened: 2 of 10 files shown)


bagder:
Shouldn't this also remove the buffer from the easy handle?

icing:
> Shouldn't this also remove the buffer from the easy handle?

When I rebase on the other merges...

icing:
now with the download buffer removed everywhere. Curious if this survives the torture tests, because conncache shutdown *might* now try to allocate the xfer_buf, if it had not been used before.

bagder:
> conncache shutdown _might_ now try to allocate the xfer_buf, if it had not been used before.

Yes that sounds possible possible. Like if you only do `curl -I ftp://host`, as then it never actually did a transfer but the connection will be put in the cache.

But;: this is a *transfer* buffer now. Surely the connection shutdown no longer needs the buffer?

icing:
> > conncache shutdown _might_ now try to allocate the xfer_buf, if it had not been used before.
> 
> Yes that sounds possible. Like if you only do `curl -I ftp://host`, as then it never actually did a transfer but the connection will be put in the cache.
> 
> But;: this is a _transfer_ buffer now. Surely the connection shutdown no longer needs the buffer?

True, good point.

lib/conncache.c
```diff
@@ -395,7 +395,6 @@ bool Curl_conncache_return_conn(struct Curl_easy *data,
          important that details from this (unrelated) disconnect does not
          taint meta-data in the data handle. */
       struct conncache *connc = data->state.conn_cache;
-      connc->closure_handle->state.buffer = data->state.buffer;
       connc->closure_handle->set.buffer_size = data->set.buffer_size;
```
bagder, line 398: Maybe `closure->buffer_size` should not be assigned at all here, as then I believe it will remain at zero which should be a better value to detect the error if this is actually used when shutting down?

lib/conncache.c
```diff
@@ -522,11 +521,9 @@ Curl_conncache_extract_oldest(struct Curl_easy *data)
 void Curl_conncache_close_all_connections(struct conncache *connc)
 {
   struct connectdata *conn;
-  char buffer[READBUFFER_MIN + 1];
   SIGPIPE_VARIABLE(pipe_st);
   if(!connc->closure_handle)
     return;
-  connc->closure_handle->state.buffer = buffer;
   connc->closure_handle->set.buffer_size = READBUFFER_MIN;
```
bagder, line 527: Here as well.

lib/multi.c
```diff
@@ -3819,3 +3815,56 @@ struct Curl_easy **curl_multi_get_handles(struct Curl_multi *multi)
...
+      failf(data, "could not allocate xfer_buf of %zu bytes", xfer_buf_len);
+      return CURLE_OUT_OF_MEMORY;
+    }
+    data->multi->xfer_buf_len = xfer_buf_len;
+  }
+
+  data->multi->xfer_buf_borrowed = TRUE;
+  /* buf may be larger, but return length that easy wants */
+  *pbuf = data->multi->xfer_buf;
+  *pbuflen = xfer_buf_len;
```
bagder, line 3859: Wouldn't it be better to return the actual size so that the transfer can make use of the full buffer size? Is there actually ever a downside to that?

icing: Well, if someone only wants 2K, I thought this would be more backward compatible. I am not opposed to changing that, was just not sure it is the right thing.

lib/conncache.c
icing: I can do that and error the borrow if the transfer has size 0?

icing: Changed as requested, transfer with buffer_size 0 trying to borrow errors now.

lib/conncache.c
icing: removed

lib/multi.c
icing: changed to always return the size that is there

bagder:
First: I  want to wait with merging this until after 8.6.0. For safety.

Then:

I was thinking: before this PR, the download buffer is only allocated while in use. It is freed again from the easy handle after the transfer is complete (in the DONE state), which makes keeping N easy handles around not use much memory, even if those handles are used with the easy interface.

This PR puts the freeing of the buffer back to when the easy handle is cleaned up, making easy handles use much more memory post a completed transfer.

Do you think it would make sense to have the multi handle free the download buffer once there is no transfer in progress anymore? It would risk having to re-alloc it again immediately when new transfers are added, but it would make the easy interface use much more like before and might even make sense for the multi interface use cases too: simply use less memory when no transfers are running.

icing:
> First: I want to wait with merging this until after 8.6.0. For safety.

Agree.

> Then:
> 
> I was thinking: before this PR, the download buffer is only allocated while in use. It is freed again from the easy handle after the transfer is complete (in the DONE state), which makes keeping N easy handles around not use much memory, even if those handles are used with the easy interface.
> 
> This PR puts the freeing of the buffer back to when the easy handle is cleaned up, making easy handles use much more memory post a completed transfer.
> 
> Do you think it would make sense to have the multi handle free the download buffer once there is no transfer in progress anymore? It would risk having to re-alloc it again immediately when new transfers are added, but it would make the easy interface use much more like before and might even make sense for the multi interface use cases too: simply use less memory when no transfers are running.

We could, when entering `DONE` for a transfer, check if there are then no more active transfers at the multi handle and then free the buffer. That would put the memory situation for single, easy transfers back to how it was before.

icing:
Added `xfer_buf` free when `multi_done()` detects there aren't any active transfers any more.

lib/multi.c
```diff
@@ -476,6 +477,16 @@ static void link_easy(struct Curl_multi *multi,
   }
 }
 
+static bool multi_has_active_easy(struct Curl_multi *multi)
+{
+  struct Curl_easy *data;
+  for(data = multi->easyp; data; data = data->next) {
+    if(data->mstate < MSTATE_DONE)
+      return TRUE;
+  }
```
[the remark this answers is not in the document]
icing: Changed it to trigger on `multi->num_alive` drop when switching a transfer to MSTATE_COMPLETED.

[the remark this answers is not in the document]
bagder: :+1:

bagder (approved):

bagder:
Maybe add something like this?
~~~diff
diff --git a/docs/libcurl/opts/CURLOPT_BUFFERSIZE.md b/docs/libcurl/opts/CURLOPT_BUFFERSIZE.md
index 1faebeef5..55ea34fa8 100644
--- a/docs/libcurl/opts/CURLOPT_BUFFERSIZE.md
+++ b/docs/libcurl/opts/CURLOPT_BUFFERSIZE.md
@@ -40,10 +40,14 @@ buffer size allowed to be set is 1024.
 DO NOT set this option on a handle that is currently used for an active
 transfer as that may lead to unintended consequences.
 
 The maximum size was 512kB until 7.88.0.
 
+Starting in libcurl 8.7.0, there is just a single buffer used by all easy
+handles added to a multi handle no matter how many parallel transfers there
+are. The buffer is allocated as long as there are active transfers.
+
 # DEFAULT
 
 CURL_MAX_WRITE_SIZE (16kB)
 
 # PROTOCOLS

icing:
Added this, thanks! Also rebased without problems.

Issue #12806, opened 2024-01-26
Memory-leak with HTTP3


gvanem:
### I did this

Building with `-DUSE_NGHTTP3`, `-DCURLDEBUG` etc. And issuing these commands (in a .bat-file):
```
  set CURL_MEMDEBUG=c:\temp\memdebug.curl
  set CURL_DEBUG=http2,http3,setup
  set CURL_SSL_BACKEND=rustls
  curl.exe --http3 -v4oNUL https://www.google.com
  ...
  *   Trying 142.250.74.132:443...
  * QUIC cipher selection: TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_CCM_SHA256
  *  CAfile: f:\MingW32\src\inet\curl\ca-bundle.pem  << !! leak at 3392?
    ...
  * Connected to www.google.com (142.250.74.132) port 443
  * ALPN: curl offers h2,http/1.1
  * rustls_connection wants us to write_tls.
  * Curl_socket_check: reading would block
  * Curl_socket_check: reading would block
  * Curl_socket_check: reading would block
  * Curl_socket_check: reading would block
  * Curl_socket_check: reading would block
  *  subjectAltName: host "www.google.com" matched cert's "*.google.com"
  * Verified certificate just fine
  * Connected to www.google.com (142.250.74.132) port 443
  * [SETUP] close
  * [SETUP] destroy
  * using HTTP/3
  * STATE: CONNECTING => PROTOCONNECT handle 0x26d5ae2a328; line 2144
  * STATE: PROTOCONNECT => DO handle 0x26d5ae2a328; line 2176
  * Server auth using Basic with user 'anonymous'
  * [HTTP/3] [0] OPENED stream for https://www.google.com/
  * [HTTP/3] [0] [:method: GET]
  * [HTTP/3] [0] [:scheme: https]
  * [HTTP/3] [0] [:authority: www.google.com]
  * [HTTP/3] [0] [:path: /]
  * [HTTP/3] [0] [authorization: Basic YW5vbnltb3VzOnlvdXJAZW1haWwuYWRkcmVzcw==]
   ...
```

shows some leaks from `perl tests/memanalyze.pl %CURL_MEMDEBUG%`:
```
  Leak detected: memory still allocated: 71 bytes
  At 1c572a9a4a8, there's 39 bytes.
   allocated by vtls/openssl.c:3392
[...]

PR #12807, opened 2024-01-26
configure: error out on multissl + http3

bagder:
Since the QUIC/h3 code has not knowledge or handling of multissl it might bring unintended consequences if we allow it.

Ref: #12806

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4691,6 +4691,9 @@ fi
 
 if test "x$USE_NGTCP2_H3" = "x1" -o "x$USE_QUICHE" = "x1" \
     -o "x$USE_OPENSSL_H3" = "x1" -o "x$USE_MSH3" = "x1"; then
+  if test "x$CURL_WITH_MULTI_SSL" = "x1"; then
+    AC_MSG_ERROR([MultiSSL cannot be enabled with HTTP/3 and vice versa])
+  fi
   SUPPORT_FEATURES="$SUPPORT_FEATURES HTTP3"
 fi
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -720,6 +720,10 @@ if(USE_MSH3)
   list(APPEND CURL_LIBS ${MSH3_LIBRARIES})
 endif()
 
+if(CURL_WITH_MULTI_SSL AND (USE_NGTCP2 OR USE_QUICHE OR USE_MSH3))
+  message(FATAL_ERROR "MultiSSL cannot be enabled with HTTP/3 and vice versa.")
+endif()
+
 if(NOT CURL_DISABLE_SRP AND (HAVE_GNUTLS_SRP OR HAVE_OPENSSL_SRP))
   set(USE_TLS_SRP 1)
 endif()
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -829,6 +829,11 @@ int getpwuid_r(uid_t uid, struct passwd *pwd, char *buf,
 #if (defined(USE_NGTCP2) && defined(USE_NGHTTP3)) || \
     (defined(USE_OPENSSL_QUIC) && defined(USE_NGHTTP3)) || \
     defined(USE_QUICHE) || defined(USE_MSH3)
+
+#ifdef CURL_WITH_MULTI_SSL
+#error "Multi-SSL combined with QUIC is not supported"
+#endif
+
 #define ENABLE_QUIC
 #define USE_HTTP3
 #endif


configure.ac
```diff
@@ -4691,6 +4691,9 @@ fi
 
 if test "x$USE_NGTCP2_H3" = "x1" -o "x$USE_QUICHE" = "x1" \
     -o "x$USE_OPENSSL_H3" = "x1" -o "x$USE_MSH3" = "x1"; then
+  if test "x$CURL_WITH_MULTI_SSL" = "x1"; then
+    AC_MSG_ERROR([MultiSSL cannot be enabled with HTTP/3 and vice versa])
```
vszakats, line 4695: HTTP3 + Schannel/SecureTransport are working fine and the above disallows them too. It'd be useful to let these combos through. (curl-for-win uses them in default builds.)

bagder: > are working fine

In what combinations have you tested this? We don't test anything of this in CI and the h3 code has no multissl awareness. It feels brittle and risky.

vszakats: I'm not doing extra testing with these builds, but they were built like this (OpenSSL + Schannel) since MultiSSL became available.

Since the introduction of HTTP/3 (around 2022-05-23), I haven't re-evaluated that, but quictls/LibreSSL + Schannel continued to build and appeared to work fine in basic tests and without any obvious issues or bug reports. They also have fairly distinct codepaths, except some crypto backend stuff unrelated to HTTP/3.

It'd be a big shock if this combo was broken all along.

...testing with offical 8.5.0_5 under WINE (with script #12806, but with just `-v`, curl.se and `--disable`), I see no leaks, but schannel doesn't use H2, only H1 (also true with pure schannel build) and it _does actually touch the QUIC codepath_ (with `QUIC connect to`), but works. LibreSSL also works with no leaks, in H3 mode. (`-v4` resets H3 to H2. Is it a valid curl option?)

[ If so, or if unsupported, curl-for-win will need a reshuffle. Back then without Schannel, it was not possible to download `cacert.pem` (to bootstrap OpenSSL) via HTTPS because of the lack of native CA support. Maybe `--ca-native`/`CURLSSLOPT_NATIVE_CA` options fix that now? Schannel is also useful to do cross-checks for bug reports, and also to keep the Schannel codepath build-tested with curl-for-win options. I'd also expect at least some users relying on this in some form or another. ]

gvanem: Shouldn't there also be something like this:
```c
#if defined(CURL_WITH_MULTI_SSL) && defined(USE_NGHTTP3)
#error "Multi-SSL with Nghttp3 is unsupported"
#endif
```

Ref. the snippet in `curl_setup.h`:
```c
#if defined(USE_LIBIDN2) && defined(USE_WIN32_IDN)
#error "Both libidn2 and WinIDN are enabled, choose one."
#endif
```

bagder: The QUIC code has no awareness of multissl. Nobody has worked on that yet. The QUIC backends can't change TLS at run-time. It is not about nghttp3 specifically. It goes for all QUIC related code in libcurl.

If we in the multssl code initialize the selected TLS, it might be a *separate* TLS library used there than is used in QUIC which then is *not* gets initialized properly before use.

I have not actually tried this myself, but I can see how things like this can happen. I rather be cautious  and say that QUIC/h3 is not tested or confirmed to work with multissl.

I like to believe that multissl is less needed these days since the introduction of `--ca-native`, but I don't know exactly for which other purposes people previously have selected to use the Schannel backend.

vszakats: I guess in that case, the best is to shut down MultiSSL.

Just confirmed on a real Windows machine with network, that `--ca-native` works, so Schannel is no longer necessary for using HTTPS (without having a hand-rolled CA bundle that is.) curl-for-win also ships with a CA bundle.

As for macOS, `--ca-native` doesn't work, but there is a CA bundle at `/etc/ssl/cert.pem` (and also one provided for example by Homebrew.) I'm guessing these are still not native (and thus missing custom entries added via Keychain), but good enough to boot.

vszakats:
Same for CMake:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 6c7bd5625..1b5ea67c2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -720,6 +720,10 @@ if(USE_MSH3)
   list(APPEND CURL_LIBS ${MSH3_LIBRARIES})
 endif()
 
+if(CURL_WITH_MULTI_SSL AND (USE_NGTCP2 OR USE_QUICHE OR USE_MSH3))
+  message(FATAL_ERROR "MultiSSL cannot be enabled with HTTP/3 and vice versa.")
+endif()
+
 if(NOT CURL_DISABLE_SRP AND (HAVE_GNUTLS_SRP OR HAVE_OPENSSL_SRP))
   set(USE_TLS_SRP 1)
 endif()
```

bagder:
thanks @vszakats and @gvanem for your assistance!

PR #12811, opened 2024-01-27
http: only act on 101 responses when they are HTTP/1.1

bagder:
For 101 responses claiming to be any other protocol, bail out. This would previously trigger an assert.

Add test 1704 to verify.

Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66184


icing (approved):
That looks reasonable. WEIRD_SERVER_REPLY is exactly what I would expect.

PR #12818, opened 2024-01-29
cd2nroff: fix duplicate output issue

bagder:
Reported-by: Jay Satiro @jay 
Fixes https://github.com/curl/curl-www/issues/321

### DIFF
--- a/scripts/cd2nroff
+++ b/scripts/cd2nroff
@@ -339,12 +339,11 @@ sub single {
 
 $f = $ARGV[0];
 if(defined($f)) {
-    while($f) {
-        $r = single($f);
+    for my $f (@ARGV) {
+        my $r = single($f);
         if($r) {
             exit $r;
         }
-        $f = shift @ARGV;
     }
 }
 else {
--- a/scripts/cd2nroff
+++ b/scripts/cd2nroff
@@ -337,8 +337,7 @@ sub single {
     return $errors;
 }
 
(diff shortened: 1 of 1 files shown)


jay:
LGTM. I made a slight modification to check `@ARGV` directly. I tested both ways in bash in Windows:

~~~
/x/j/curl/curl/scripts/cd2nroff curl_escape.md | roffit >/tmp/temp.html
start /tmp/temp.html

/x/j/curl/curl/scripts/cd2nroff < curl_escape.md | roffit >/tmp/temp.html
start /tmp/temp.html
~~~

vszakats:
Thanks @jay, @bagder. I really [wasn't sure](https://github.com/curl/curl/pull/12762#issuecomment-1905626112) if my code was solid. (and yet to figure out how the fix works!)

jay:
> I really [wasn't sure](https://github.com/curl/curl/pull/12762#issuecomment-1905626112) if my code was solid. (and yet to figure out how the fix works!)

The bug was `$f = shift @ARGV;` because that is assigning $ARGV[0] to $f a second time since [`shift`](https://perldoc.perl.org/functions/shift) returns the argument that's removed. I've submitted #12819 which adds strict and warnings pragmas to help catch some things but it wouldn't have caught that.

vszakats:
Oh, right! Rookie mistake, nothing to do with the language. Glad to see more warnings in any case.

PR #12822, opened 2024-01-31
Correct example for --form-string

verhovsky:
```
$  curl --form-string "data" example.com
Warning: Illegally formatted input field
curl: option --form-string: is badly used here
curl: try 'curl --help' for more information
$  curl --form-string "name=data" example.com
<!doctype html>
<html>
[...]
```


bagder (approved):

bagder:
Thanks!

PR #12825, opened 2024-01-31
ntml_wb, fix buffer type

icing:
Follow-up to 8e9c304f4a350a822ca4138
Reported-by: tranzystorekk

### DIFF
--- a/lib/curl_ntlm_wb.c
+++ b/lib/curl_ntlm_wb.c
@@ -266,7 +266,7 @@ static CURLcode ntlm_wb_response(struct Curl_easy *data, struct ntlmdata *ntlm,
   size_t len_in = strlen(input), len_out = 0;
   struct dynbuf b;
   char *ptr = NULL;
-  unsigned char buf[1024]
+  unsigned char buf[1024];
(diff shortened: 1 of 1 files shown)


bagder (approved):

lib/curl_ntlm_wb.c
```diff
@@ -266,7 +266,7 @@ static CURLcode ntlm_wb_response(struct Curl_easy *data, struct ntlmdata *ntlm,
   size_t len_in = strlen(input), len_out = 0;
   struct dynbuf b;
   char *ptr = NULL;
-  usigned char buf[1024]
+  unsigned char buf[1024]
```
tranzystorekk, line 269: missing semicolon at the end as well 😅

bagder:
This mistake was not detected by CI because currently the configure script is broken and cannot enable NTLM_WB since the check requires a detected `fork` but that check has been removed.

A fix for that configure mistake exists in #12479 but we have held off on merging that on purpose since NTLM_WB is scheduled to get removed in June 2024.

lib/curl_ntlm_wb.c
bagder, line 269:
```suggestion
  unsigned char buf[1024];
```

PR #12826, opened 2024-01-31
OS400: avoid using awk in the build scripts

monnerat:
Awk is a PASE program and its use may cause a failure depending on the CCSID of the calling script (IBM bug?).

For this reason, revert to an sed-only solution to extract the exported symbols from the header files.


monnerat:
Thanks for merge!

PR #12837, opened 2024-02-01
tool: Add option --retry-ignore-server-time

jay:
- Add option --retry-ignore-server-time to ignore the server's requested retry delay time (aka the optional Retry-After response header).

This option stops the server from overriding curl's retry delay time before retrying a transient HTTP error (eg Too Many Requests, Gateway Timeout, etc). It is not polite to ignore the server time however it may be necessary for some network conditions.

For example, a server may set a retry time of 1 second even though that may not be practical depending on how busy the server is. The user can use this option to ignore the server time in favor of --retry-delay time (or the default algorithmic retry delay time), thereby giving their transfers a greater chance of success.

Bug: https://curl.se/mail/archive-2024-01/0022.html
Reported-by: Dirk Hünniger

Closes #xxxx


bagder:
I propose the simpler take: use `Retry-After` only if it is longer than the otherwise used value:
~~~
index 86936ba59..5a5fc8745 100644
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -552,11 +552,11 @@ static CURLcode post_per_transfer(struct GlobalConfig *global,
         curl_easy_getinfo(curl, CURLINFO_RETRY_AFTER, &retry_after);
         if(retry_after) {
           /* store in a 'long', make sure it doesn't overflow */
           if(retry_after > LONG_MAX/1000)
             sleeptime = LONG_MAX;
-          else
+          else if((retry_after * 1000) > sleeptime)
             sleeptime = (long)retry_after * 1000; /* milliseconds */
 
           /* if adding retry_after seconds to the process would exceed the
              maximum time allowed for retrying, then exit the retries right
              away */

jay:
> I propose the simpler take: use `Retry-After` only if it is longer than the otherwise used value:

Ok. Closing in favor of #12871

Issue #12844, opened 2024-02-01
Problem using --http1.0 on archlinux with curl 8.6.0


lepapareil:
### I did this

In curl 8.5.0 doing:

```
curl --location --http1.0 --head https://google.fr
```

Results on:

```
HTTP/1.0 301 Moved Permanently
Location: https://www.google.fr/
Content-Type: text/html; charset=UTF-8
...
```

But with curl 8.6.0, it results on:

```
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0
```

### I expected the following

```
HTTP/1.0 301 Moved Permanently
Location: https://www.google.fr/
Content-Type: text/html; charset=UTF-8
...
```

### curl/libcurl version

```
curl 8.6.0 (x86_64-pc-linux-gnu) libcurl/8.6.0 OpenSSL/3.2.0 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 libssh2/1.11.0 nghttp2/1.59.0
Release-Date: 2024-01-31
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

Archlinux docker image 20240101.0.204074

bagder:
This does not reproduce with curl 8.6.0 in debian:

~~~
curl 8.6.0 (x86_64-pc-linux-gnu) libcurl/8.6.0 OpenSSL/3.1.4 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.7 libpsl/0.21.2 libssh2/1.11.0 nghttp2/1.58.0 librtmp/2.3 OpenLDAP/2.5.13
Release-Date: 2024-01-31, security patched: 8.6.0-1
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
[...]

PR #12848, opened 2024-02-02
vtls: revert greedy read, add test case for unclean shutdown

icing:
Related to #12844, we see an SSL error reported when the server closes the connection without a proper TLS shutdown message. This PR fixes the consequences of such an error in the following way:

In addition, we see #12885 with "hanging" response handling due to the greedy receives in `vtls.c`.

* `vtls/vtls.c` no longer does "greedy", reverts to the state before 9a90c9dd64d2f03601833a70786d485851bd1b53.
* add test_05_04 to verify behaviour of http/1.0, http/1.1 and h2 against an Apache resource that does an unclean TLS shutdown (no shutdown notice, just closing TCP)

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -305,6 +305,8 @@ struct ossl_ssl_backend_data {
   X509*    server_cert;
   BIO_METHOD *bio_method;
   CURLcode io_result;       /* result of last BIO cfilter operation */
+  size_t io_read_iterations;
+  bool io_delay_subsequent_close;
 #ifndef HAVE_KEYLOG_CALLBACK
   /* Set to true once a valid keylog entry has been created to avoid dupes. */
   bool     keylog_done;
@@ -769,6 +771,18 @@ static int ossl_bio_cf_in_read(BIO *bio, char *buf, int blen)
     if(CURLE_AGAIN == result)
       BIO_set_retry_read(bio);
   }
+  else if(nread == 0 &&
+          backend->io_delay_subsequent_close &&
+          backend->io_read_iterations) {
+    /* connection closed. We are not in the first read iteration (OpenSSL
+     * may call us several times during one SSL_read()) and we are
+     * asked to delay the closing to the next SSL_read(). */
+    backend->io_result = CURLE_AGAIN;
+    backend->io_delay_subsequent_close = FALSE;
+    BIO_set_retry_read(bio);
+    nread = -1;
+    Curl_expire(data, 0, EXPIRE_RUN_NOW);
+  }
 
   /* Before returning server replies to the SSL instance, we need
    * to have setup the x509 store or verification will fail. */
@@ -780,6 +794,8 @@ static int ossl_bio_cf_in_read(BIO *bio, char *buf, int blen)
     }
     backend->x509_store_setup = TRUE;
   }
+  /* count our read invocations */
+  backend->io_read_iterations++;
 
   return (int)nread;
 }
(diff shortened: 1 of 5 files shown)


jay:
I've asked the reporter to see if reverting 9a90c9dd works as a fix. Regarding the other part I don't see why we should hide the syscall error from OpenSSL. We knew a long time ago that eventually OpenSSL would require a close notify (if that's even what's happening; syscall can mean many things) and I even noted it somewhere in the code. Frankly I thought we gave it a lot longer than we should have by not requiring it ourselves. For the Schannel backend we already do it, require close notify if there's no known termination point and we've done that for years.

icing:
The reason for hiding the error on multiplexed connections is that those protocols are not vulnerable to a truncation attack. Masking the error as a normal close allows them to aggressively receive data and process it.

The idea that a missing shutdown is a protocol violation is true, but when we see servers from google doing it, we should ask ourselves if erroring on it is in our users interest when the transfer would have been fine otherwise.

jay:
I disagree that syscall is a normal error. I don't have any objection to so called greedy reading or decrypting where you want to bundle records but that should not cause a transfer to error, and if it does there's an underlying problem there. As long as curl can read to the end point of the transfer that should not be an error. An error is only supposed to propagate when the end point is unknown because then curl doesn't know whether or not it has received all the data. Whether this is possible with HTTP/2 DATA frame without content length I don't know, I assume an END stream or something would be required as the end point so then curl would know.

icing:
Not every `SSL_ERROR_SYSCALL` is treated as normal error. The check is:

```
      if(backend->io_read_zero && (err == SSL_ERROR_SYSCALL)) {
        if(Curl_conn_is_multiplex(cf->conn, cf->sockindex)) {

```

If the BIO reads a normal connection close, length 0, and OpenSSL returns `SSL_ERROR_SYSCALL`, then this is taken as indicator that a TLS shutdown was missing. AND only is this error silenced for connections that are multiplexed, e.g. HTTP/2 and forwarded as a unerrored connection close.

HTTP/2 needs to deal with *any* connection close and error on streams that have not been closed property at the HTTP/2 protocol level. For HTTP/2 it is irrelevant, if a TLS shutdown is done or not, only the HTTP/2 state of things matters. Other protocols do not have that luxury and there we need the TLS shutdown. That is why the error suppression is only done on multiplexed connections.

bagder:
With HTTP/2 and HTTP/3 curl always knows when the data has arrived and the close-notify does not provide any helpful hints to us. On the contrary, since the close is on the connection and the transfers are per-stream, the closure is not even exactly related to a specific transfer. Any premature disconnect when doing h2 or h3 is detected already, without specific close-notify awareness.

icing:
I revert my foolish opinion from above. No special workaround for suppressing TLS close errors  in this PR.

Having added the test case the reproduces the problem with http/1.0 from  #12844, I found no way to reproduce an error in other protocols transfer processing. Therefore, I consider an additional workaround unnecessary since we are unable to verify that it does any good.

bagder (approved):

ThaFireDragonOfDeath:
The changes from this pull request fixes also a hang problem I encountered with curl as FTP client and vsftpd as FTP server (I wasn't able to reproduce this hang on ProFTPD or Windows Server FTP servers).
If I execute the command "curl --ssl-reqd --insecure --user user:password ftp://server_host_name_or_ip/random_existing_directoy_name" (listing content of a directory over FTP with SSL) multiple times with little to no delay after completion curl hanged until the FTP server cancels with "421 Data timeout. Reconnect. Sorry." and the client exits with "curl: (28) control connection looks dead".
After applying the changes of vtls.c from this pull request to the released 8.6.0 sources the hang wasn't reproducible anymore.

PR #12852, opened 2024-02-02
Correct a typo in the ALPN documentation

larsks:
The ALPN documentation erroneously referred to a "host number" instead of a
"port number".


bagder (approved):

bagder:
Thanks!

PR #12862, opened 2024-02-05
config parser errors should exit

bagder:


bagder:
For some unknown reason, *some* Windows jobs consider the blank variable missing but some others do not...

bagder:
In **4** of the Windows Appveyor CI builds, `getenv()` returns NULL when the environment variable `BLANK` has zero length content. When that happens, test 428 fails. There are also **8** other jobs where it returns a blank string correctly.

I'm struggling to find the pattern...

jay:
I haven't tried this yet but to be clear you are setting `BLANK=` in your test to [remove it](https://github.com/curl/curl/blob/5ce164e0e9290c96eb7d502173426c0a135ec008/tests/runner.pm#L646-L658
) from the environment block in perl, which means the processes that inherit that environment block should not have a BLANK environment variable and therefore getenv("BLANK") is supposed to return NULL... right?

https://ci.appveyor.com/project/curlorg/curl/builds/49123802/job/2a4w7i21npys9pd3?fullLog=true#L2894

bagder:
> you are setting BLANK= in your test to remove it from the environment block in perl

No. I have changed that so that we remove them using a different syntax. `BLANK=` (starting in this PR) sets the variable to a zero length string.

bagder:
In this commit test 428 is skipped on Windows.

jay:
In Windows AFAIK setting no-content environment variables can only be done by using Windows API SetEnvironmentVariable. If MS C runtime [_putenv](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-wputenv?view=msvc-170) is used to set a no-content environment variable then it removes the variable:

_"You can remove a variable from the environment by specifying an empty value_string, or in other words, by specifying only varname=."_

bagder:
I would have expected **all** Windows CI jobs to have failed test 428 if that's the case!

jay:
Yeah that's weird. Even if perl sets an environment variable to no-content using SetEnvironmentVariable I wouldn't expect a child process calling getenv to be able retrieve it.

PR #12872, opened 2024-02-06
version: allow building with ancient libpsl

bagder:
The psl_check_version_number() API was added in libpsl 0.11.0. CentOS 7 ships with version 0.7.0 which lacks this API. Revert to using the older versioning API if we detect an old libpsl version.

Follow-up to 72bd88adde0e8cf6e63644a7d6df1da01a399db4 Bug: https://curl.se/mail/archive-2024-02/0004.html Reported-by: Scott Mutter

### DIFF
--- a/lib/version.c
+++ b/lib/version.c
@@ -212,9 +212,13 @@ char *curl_version(void)
 
 #ifdef USE_LIBPSL
   {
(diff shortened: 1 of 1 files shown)


bagder:
This is actually just supporting the old API. I believe that old version of the lib can still be built using an updated database. But even so, I believe an outdated database is probably better than none at all.

PR #12887, opened 2024-02-06
linux.yml: add arch linux container

pheiduck:
configuration is the same as alpine.
Let me know if we should change this.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -201,7 +201,8 @@ jobs:
 
       - if: startsWith(matrix.build.container, 'archlinux/archlinux')
         run: |
-          pacman -Sy autoconf automake libtool perl openssl libssh2 zlib brotli zstd libidn2 openldap heimdall libpsl impacket libtasn1 python-pycryptodomex perl openssh stunnel git --noconfirm
+          pacman -Syyu autoconf automake libtool perl openssl libssh2 zlib brotli zstd libidn2 openldap heimdall libpsl impacket libtasn1 python-pycryptodomex perl openssh stunnel git --noconfirm
+          locale-gen "en_US.UTF-8"
         name: 'install dependencies'
 
       - uses: actions/checkout@v4
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -202,7 +202,7 @@ jobs:
       - if: startsWith(matrix.build.container, 'archlinux/archlinux')
         run: |
           pacman -Syyu openldap heimdall impacket python-pycryptodomex openssh stunnel git --noconfirm
-          pacman -R perl --noconfirm
+          PERL_BADLANG=0
         name: 'install dependencies'
 
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


dfandrich:
What's the purpose of this build? Alpine gives us a build against MUSL which we
don't get anywhere else. How does this build improve CI coverage?

pheiduck:
> What's the purpose of this build? Alpine gives us a build against MUSL which we don't get anywhere else. How does this build improve CI coverage?

Not much but it will build on latest upstream packages.

dfandrich:
There are too many distros to try to get them all. Each new one is a new opportunity for their maintainers to make some backwards-incompatible change that causes our builds to fail until one of us manually debugs and fixes the problem. If the breaking change is due to some dependency that has changed that curl needs to adapt to, it's worth it, but it's just as likely that some infrastructure gets changed that only affects the CI.

I do think we should have a build on a rolling distro to get our code exposed to new dependencies ASAP. I'm also quite sure that it will be painful for us, since we'll know immediately when some dependency is bumped incompatibly and must fix it to get builds green again. As it is now, we often get patches after somebody else builds curl on a rolling release, so we're somewhat shielded from that initial shock.

I'm not that familiar with Arch but my understanding is that it's a rolling release so that would satisfy this. Other likely contenders are OpenSUSE Tumbleweed and Debian Testing, where I'm guessing the latter probably has the most familiarity among the three by curl developers. I don't think we need or want more than one.

If we do take this PR, shouldn't there be a `pacman -U` to make sure everything is upgraded and we're truly running the latest?

bagder:
Several tests fail because of some setlocale issue?

~~~
sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)

pheiduck:
@bagder
```
```
It is suggested to use `CFLAGS=-Wno-error`, but that doesn't really make sense?

pheiduck:
This is happened on `make tests`:
```
lib582.c: In function ‘addFd’:
lib582.c:92:24: error: pointer ‘oldptr’ may be used after ‘realloc’ [-Werror=use-after-free]
   92 |       sockets->sockets = oldptr;
      |       ~~~~~~~~~~~~~~~~~^~~~~~~~
lib582.c:88:24: note: call to ‘realloc’ here
   88 |     sockets->sockets = realloc(oldptr, sizeof(curl_socket_t) *
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   89 |                                (sockets->max_count + 20));
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

pheiduck:
CFLAGS=-Wno-error makes no difference so I might revert this later

bagder:
> It is suggested to use `CFLAGS=-Wno-error`, but that doesn't really make sense?

Who is suggesting this and why?

pheiduck:
> > It is suggested to use `CFLAGS=-Wno-error`, but that doesn't really make sense?
> 
> Who is suggesting this and why?

Someone on Stackoverflow to not be warned about this error.
But it turns out this make no difference.

pheiduck:
I think I have to switch to debian [testing/unstable] and open a new PR for it. :/

pheiduck:
I will start another PR, with Fedora, I guess it's more stable for CI stuff. :)

dfandrich:
Fedora is not a rolling release distro, though.  It wouldn't add much value in comparison.

PR #12889, opened 2024-02-07
Typo fix in CURLOPT_WRITEFUNCTION.md

danielsz:
> The maximum amount of body data that is **be** passed to the write callback is defined in the curl.h header file

Removed superfluous **be**


bagder (approved):

bagder:
Thanks!

PR #12897, opened 2024-02-08
SHA-512/256 real implementation

Karlson2k:
This PR adds real support for SHA-512/256 hashing algorithm.
Currently Digest Auth fakes SHA-512/256 by using SHA-256 algorithm which is completely wrong.

The new implementation added as an optional and can be disabled if needed (the related tests are automatically disabled). The disabling machinery is here, but the support for disabling in `configure` and CMake is not implemented and can be added when/if needed.

### DIFF
--- a/lib/curl_sha512_256.c
+++ b/lib/curl_sha512_256.c
@@ -1,0 +1,601 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Evgeny Grin (Karlson2k), <k2k@narod.ru>.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ * SPDX-License-Identifier: curl
+ *
+ ***************************************************************************/
+
+#include "curl_setup.h"
+
+#if !defined(CURL_DISABLE_DIGEST_AUTH) && !defined(CURL_DISABLE_SHA512_256)
+
+#include "curl_sha512_256.h"
+#include "warnless.h"
+
+/* ** This implementation of SHA-512/256 hash calculation was originally ** *
+ * ** written by Evgeny Grin (Karlson2k) for GNU libmicrohttpd.          ** *
+ * ** The author ported the code to libcurl. The ported code is provided ** *
+ * ** under curl license.                                                ** *
+ * ** This is a minimal version with minimal optimisations. Performance  ** *
+ * ** can be significantly improved. Big-endian store and load macros    ** *
+ * ** are obvious targets for optimisation.                              ** */
+
+#ifdef __GNUC__
+#  if defined(__has_attribute) && defined(__STDC_VERSION__)
+#    if __has_attribute(always_inline) && __STDC_VERSION__ >= 199901
+#      define MHDX_INLINE inline __attribute__((always_inline))
+#    endif
+#  endif
+#endif
+
+#if !defined(MHDX_INLINE) && \
+  defined(_MSC_VER) && !defined(__GNUC__) && !defined(__clang__)
+#  if _MSC_VER >= 1400
+#    define MHDX_INLINE __forceinline
+#  else
+#    define MHDX_INLINE /* empty */
+#  endif
+#endif
+
+#if !defined(MHDX_INLINE)
+#  if defined(inline)
+     /* Assume that 'inline' macro was already defined correctly by
+      * the build system. */
+#    define MHDX_INLINE inline
+#  elif defined(__cplusplus)
+     /* The code is compiled with C++ compiler.
+      * C++ always supports 'inline'. */
+#    define MHDX_INLINE inline
+#  elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901
+     /* C99 (and later) supports 'inline' keyword */
+#    define MHDX_INLINE inline
+#  elif defined(__GNUC__) && __GNUC__ >= 3
+     /* GCC supports '__inline__' as an extension */
+#    define MHDX_INLINE __inline__
+#  else
+#    define MHDX_INLINE /* empty */
+#  endif
+#endif
+
+/* Bits manipulation macros and functions.
+   Can be moved to other headers to reuse. */
+
+#define MHDX_GET_64BIT_BE(ptr)                                  \
+  ( ((curl_uint64_t)(((const unsigned char*)(ptr))[0]) << 56) | \
+    ((curl_uint64_t)(((const unsigned char*)(ptr))[1]) << 48) | \
+    ((curl_uint64_t)(((const unsigned char*)(ptr))[2]) << 40) | \
+    ((curl_uint64_t)(((const unsigned char*)(ptr))[3]) << 32) | \
+    ((curl_uint64_t)(((const unsigned char*)(ptr))[4]) << 24) | \
+    ((curl_uint64_t)(((const unsigned char*)(ptr))[5]) << 16) | \
+    ((curl_uint64_t)(((const unsigned char*)(ptr))[6]) << 8)  | \
+    (curl_uint64_t)(((const unsigned char*)(ptr))[7]) )
+
+#define MHDX_PUT_64BIT_BE(ptr,val) do {                                     \
+  ((unsigned char*)(ptr))[7]=(unsigned char)((curl_uint64_t)(val));         \
+  ((unsigned char*)(ptr))[6]=(unsigned char)(((curl_uint64_t)(val)) >> 8);  \
+  ((unsigned char*)(ptr))[5]=(unsigned char)(((curl_uint64_t)(val)) >> 16); \
+  ((unsigned char*)(ptr))[4]=(unsigned char)(((curl_uint64_t)(val)) >> 24); \
+  ((unsigned char*)(ptr))[3]=(unsigned char)(((curl_uint64_t)(val)) >> 32); \
+  ((unsigned char*)(ptr))[2]=(unsigned char)(((curl_uint64_t)(val)) >> 40); \
+  ((unsigned char*)(ptr))[1]=(unsigned char)(((curl_uint64_t)(val)) >> 48); \
+  ((unsigned char*)(ptr))[0]=(unsigned char)(((curl_uint64_t)(val)) >> 56); \
+} while(0)
+
+/* Defined as a function. The macro version may duplicate the binary code
+ * size as each argument is used twice, so if any calculation is used
+ * as an argument, the calculation could be done twice. */
+static MHDX_INLINE curl_uint64_t
+MHDx_rotr64(curl_uint64_t value, unsigned int bits)
+{
+  bits %= 64;
+  if(0 == bits)
+    return value;
+  /* Defined in a form which modern compiler could optimise. */
+  return (value >> bits) | (value << (64 - bits));
+}
+
+/* SHA-512/256 specific data */
+
+/**
+ * Number of bits in single SHA-512/256 word.
+ */
+#define SHA512_256_WORD_SIZE_BITS 64
+
+/**
+ * Number of bytes in single SHA-512/256 word.
+ */
+#define SHA512_256_BYTES_IN_WORD (SHA512_256_WORD_SIZE_BITS / 8)
+
+/**
+ * Hash is kept internally as 8 64-bit words.
+ * This is intermediate hash size, used during computing the final digest.
+ */
+#define SHA512_256_HASH_SIZE_WORDS 8
+
+/**
+ * Size of SHA-512/256 resulting digest in bytes.
+ * This is the final digest size, not intermediate hash.
+ */
+#define SHA512_256_DIGEST_SIZE_WORDS (SHA512_256_HASH_SIZE_WORDS  / 2)
+
+/**
+ * Size of SHA-512/256 resulting digest in bytes
+ * This is the final digest size, not intermediate hash.
+ */
+#define SHA512_256_DIGEST_SIZE \
+  (SHA512_256_DIGEST_SIZE_WORDS * SHA512_256_BYTES_IN_WORD)
+
+/**
+ * Size of SHA-512/256 single processing block in bits.
+ */
+#define SHA512_256_BLOCK_SIZE_BITS 1024
+
+/**
+ * Size of SHA-512/256 single processing block in bytes.
+ */
+#define SHA512_256_BLOCK_SIZE (SHA512_256_BLOCK_SIZE_BITS / 8)
+
+/**
+ * Size of SHA-512/256 single processing block in words.
+ */
+#define SHA512_256_BLOCK_SIZE_WORDS \
+ (SHA512_256_BLOCK_SIZE_BITS / SHA512_256_WORD_SIZE_BITS)
+
+
+/**
+ * SHA-512/256 calculation context
+ */
+struct Sha512_256Ctx
+{
+  /**
+   * Intermediate hash value
+   * The variable is properly aligned. Smart compiler
+   * may automatically use fast load/store instruction
+   * for big endian data on little endian machine.
+   */
+  curl_uint64_t H[SHA512_256_HASH_SIZE_WORDS];
+  /**
+   * SHA-512/256 input data buffer
+   * The buffer is properly aligned. Smart compiler
+   * may automatically use fast load/store instruction
+   * for big endian data on little endian machine.
+   */
+  curl_uint64_t buffer[SHA512_256_BLOCK_SIZE_WORDS];
+  /**
+   * The number of bytes, lower part
+   */
+  curl_uint64_t count;
+  /**
+   * The number of bits, high part.
+   * Unlike lower part, this counts the number of bits, not bytes.
+   */
+  curl_uint64_t count_bits_hi;
+};
+
+
+/**
+ * Initialise structure for SHA-512/256 calculation.
+ *
+ * @param context the calculation context
+ * @return always CURLE_OK
+ */
+static CURLcode
+MHDx_sha512_256_init(void *context)
+{
+  struct Sha512_256Ctx *const ctx = (struct Sha512_256Ctx *) context;
+
+  /* Check whether the header and this file use the same numbers */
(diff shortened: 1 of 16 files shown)


lib/curl_sha512_256.c
```diff
@@ -0,0 +1,596 @@
...
+/* ** This implementation of SHA-512/256 hash calculation was originally ** *
+ * ** written by Evgeny Grin (Karlson2k) for GNU libmicrohttpd.          ** *
+ * ** The author ported the code to libcurl. The ported code is provided ** *
+ * ** under curl license.                                                ** *
+ * ** This is a minimal version with minimal optimisations. Performance  ** *
+ * ** can be significantly improved. Big-endian store and load macros    ** *
+ * ** are obvious targets for optimisation.                              ** *
+ * ** The code is designed in a way so any wide enough type (64 bits or  ** *
+ * ** more) can be used for calculations. Potentially it can be larger   ** *
+ * ** then 64 bits.                                                      ** */
```
bagder, line 41: This comment is not adding much for the reader and does not need to be provided.

```suggestion
```

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,596 @@
...
+
+#ifdef SIZEOF_INT
+#if SIZEOF_INT == 8
+#define CURLX_UINT64 unsigned int
```
bagder, line 45: We use the `curlx_` prefix for some functions we provide source access to, for the curl tool. In this code you seem to use it for most things but for a different purpose. I would appreciate if you did not. For symbols that are private in this file, you don't really need any special prefix.

bagder (commented):
Awesome! Just some nits from me...

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,596 @@
...
+#define CURLX_UINT64_C(val) val##U
+#endif /* SIZEOF_INT == 8 */
+#endif /* SIZEOF_INT */
```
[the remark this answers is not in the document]
Karlson2k: Not at the moment.
The code needs to be future-proof. 
The same for long long.

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,596 @@
...
+#define CURLX_UINT64_C(val) val##UL
+#endif /* SIZEOF_LONG == 8 */
+#endif /* !CURLX_UINT64 && SIZEOF_LONG */
+
+#if !defined(CURLX_UINT64) && defined(HAVE_LONGLONG)
+#define CURLX_UINT64 unsigned long long
+#define CURLX_UINT64_C(val) val##ULL
+/* Trim possibly larger value to 64 bits */
+#define CURLX_UINT64_TRIM(x) \
+ ((CURLX_UINT64) ((CURLX_UINT64)(x) & CURLX_UINT64_C(0xFFFFFFFFFFFFFFFF)))
```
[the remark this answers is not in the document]
Karlson2k: Yes, for the current moment.
However we already see __int128, so at some moment long long will be probably expanded.

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,596 @@
...
+#endif /* !CURLX_UINT64 && SIZEOF_LONG */
+
+#if !defined(CURLX_UINT64)
+#define CURLX_UINT64 __int64
```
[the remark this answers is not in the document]
Karlson2k: You are right. Thank you. I'll correct.

lib/curl_sha512_256.c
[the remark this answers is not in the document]
bagder: you have not tested on any such platform have you, so how do you *know* it is future-proof? You're only adding code that is never reached/used. Lots of code everywhere else in curl is going to break miserably long before his code is reached if anyone would ever introduce such weird stuff.

lib/curl_sha512_256.c
Karlson2k: The same (or quite similar) code was already published under LGPL-2.1+ license.
As I wrote the code alone, I can apply another license.

Similar comments exist in [md5.c](https://github.com/curl/curl/blob/5a4b2f93392c0fb1171dd17b67b05f39c6d45abc/lib/md5.c#L254) and [md4.c](https://github.com/curl/curl/blob/5a4b2f93392c0fb1171dd17b67b05f39c6d45abc/lib/md4.c#L228) code algorithms. I decided to add comment here to solve any possible questions about getting the code from incompatible license (`curl` license is compatible with `LGPL` but `LGPL` code cannot be re-used directly under `curl` license, unless the original author changed the license).

lib/curl_sha512_256.c
Karlson2k: OK.
These macros and functions could be moved to another file to be re-used.
I typically use some prefix to form a kind of namespace in C. This way I ensure that code will not conflict with any other code, both from the same application/library and from any third-party dynamic/static libs. Especially important for portability, when compiler and linker cannot hide non-exportable symbols (or file format does not allow symbols hiding).
Also this prevents any potential shadowing warning.
But since the prefix is reserved for other needs, I'll change it.

bagder:
Is there a particular reason you did not use the TLS library's crypto functions for this? OpenSSL etc provide functions for this.

lib/curl_sha512_256.c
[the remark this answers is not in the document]
Karlson2k: > `int` is never 8 bytes on any platform

Not correct.
ILP64 data model already exists.
See https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models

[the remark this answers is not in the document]
bagder: Read again. There are no platforms with `int` larger than 32 bits in current use where anyone would build curl. It's the same kind of argument saying that `char` does not have to be 8 bits etc. They are just theoretical proofs that it is *possible*, but they are not used.

Since the code is not tested on any such machines anyway, claiming that it works on such machines is just wishful thinking.

Karlson2k:
> Is there a particular reason you did not use the TLS library's crypto functions for this? OpenSSL etc provide functions for this.

Frankly, I didn't check the latest (read "five years old") OpenSSL versions. Previously OpenSSL supported only SHA-512, which is quite different.
Other TLS backends typically still support only SHA-512 (like GnuTLS).
Anyway, TLS backends support can be easily added, while hash computation algorithm is required in `libcurl` always, when Digest Auth is not disabled, so I believe it is most important part.

Karlson2k:
The only failing test is 1165, reported sync CURL_DISABLE not in sync.
Before fixing it, let me know what is preferred:
* new configure and cmake parameter to disable SHA-512/256
* remove option to disable SHA-512/256
* add something to mute test error without actual interface for disabling SHA-512/256

Karlson2k:
Unrelated: regexps in `test1165.pl` are incorrect.
The are `/(CURL_DISABLE_[A-Z_]+)/`
So `CURL_DISABLE_SHA` is detected by this tool instead of `CURL_DISABLE_SHA512_256`.

Karlson2k:
> Unrelated: regexps in `test1165.pl` are incorrect. The are `/(CURL_DISABLE_[A-Z_]+)/` So `CURL_DISABLE_SHA` is detected by this tool instead of `CURL_DISABLE_SHA512_256`.

PR #12903 published

dfandrich:
Are there any crypto libraries used by libcurl that still lack SHA2 functions? SHA-256 support was added to libcurl before OpenSSL supported it, but I don't see a need to carry a local version of this calculation today. There are optimized functions already available that curl could use instead.

lib/curl_sha512_256.c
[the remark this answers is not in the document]
Karlson2k: OK, I'll remove it. No problem.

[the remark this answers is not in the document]
Karlson2k: Just wanted to help you have one less place to fix when it happens. :)

Karlson2k:
@bagder Everything was corrected except one comment.
Let me know what to do with it.

Also CURL_DISABLE_xxx macro check was muted in unpolite way. Please give me [some hit what to do](https://github.com/curl/curl/pull/12897#issuecomment-1934496293).

Karlson2k:
VS2010 and VS2008 failures are related.
I don't have any of the to reproduce.
Looks like MHDX_UINT64 gets defined to empty. Quite strange.

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,588 @@
...
+#if !defined(MHDX_UINT64)
+#define MHDX_UINT64 unsigned __int64
+#define MHDX_UINT64_C(val) ((MHDX_UINT64)val)
+#endif /* !MHDX_UINT64 && SIZEOF_LONG */
+
+#ifndef MHDX_UINT64_TRIM
+/* Trim possibly larger value to 64 bits.
+   Assuming that no type is larger than 64 bits. */
+#define MHDX_UINT64_TRIM(x) ((MHDX_UINT64)(x))
+#endif /* !MHDX_UINT64_TRIM */
```
bagder, line 61: I propose you instead use the fact that we already have a 64 bit type figured out:

```suggestion
typedef unsigned CURL_TYPEOF_CURL_OFF_T mdhx_t
```

We could probably even consider making this type available for the entire libcurl, maybe as `curl_offu_t` or something?

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,588 @@
...
+
+/* Bits manipulation macros and functions.
+   Can be moved to other headers to reuse. */
+
+#define MHDX_GET_64BIT_BE(ptr)                                \
+  MHDX_UINT64_TRIM(                                           \
```
bagder, line 67: When you OR 64 bits together, it can't be larger than 64 bits so there's no need to trim this down to 64 bits. Even if the type could potentially hold something larger.

lib/curl_sha512_256.c
```diff
@@ -0,0 +1,588 @@
...
+ * size as each argument is used twice, so if any calculation is used
+ * as an argument, the calculation could be done twice. */
+static inline MHDX_UINT64
+MHDx_rotr64(MHDX_UINT64 value, unsigned int bits)
+{
+  bits %= 64;
+  if(0 == bits)
+    return value;
+  /* Defined in a form which modern compiler could optimise. */
+  return MHDX_UINT64_TRIM((value >> bits) | (value << (64 - bits)));
```
bagder, line 98: This does not need trimming, it can't be larger than 64 bit.

lib/curl_sha512_256.c
Karlson2k: Right. I'll correct it.

lib/curl_sha512_256.c
Karlson2k: Initial idea was to support types larger that 64 bits.
I current form it's not needed. I'll correct.

lib/curl_sha512_256.c
Karlson2k: If `CURL_TYPEOF_CURL_OFF_T` is always a macro and never typedef it will work.
I'd suggest to reproduce standard `stdint.h`, 
``` C
typedef unsigned CURL_TYPEOF_CURL_OFF_T curl_uint64_t
#define CURL_UINT64_SUFFIX CURL_SUFFIX_CURL_OFF_TU
#define CURL_UINT64_C(val)  val##CURL_UINT64_SUFFIX
```
It would be quite easy to read and understand by newcomers.

Karlson2k: Note: for external API libcurl *may* use `CURL_TYPEOF_CURL_OFF_T` as 32 bit.
See https://github.com/curl/curl/blob/476adfeac019edec71d67e144f184f4dbf1a46fb/include/curl/system.h#L364 and https://github.com/curl/curl/blob/476adfeac019edec71d67e144f184f4dbf1a46fb/include/curl/system.h#L406
A assume that internally libcurl uses the same types.
So probably it would make sense to add something like 
``` C
#ifdef CURL_HAVE_CULT_OFF_T_64BIT
  typedef unsigned CURL_TYPEOF_CURL_OFF_T curl_uint64_t
#  define CURL_UINT64_SUFFIX CURL_SUFFIX_CURL_OFF_TU
#  define CURL_UINT64_C(val)  val##CURL_UINT64_SUFFIX
#endif

Comment (Karlson2k) [reply lib/curl_sha512_256.c:61]: Of maybe even
``` C
#ifdef CURL_HAVE_CULT_OFF_T_64BIT
  typedef unsigned CURL_TYPEOF_CURL_OFF_T curl_uint64_t
#  define CURL_HAVE_UINT64
#  define CURL_UINT64_SUFFIX CURL_SUFFIX_CURL_OFF_TU
#  define CURL_UINT64_C(val)  val##CURL_UINT64_SUFFIX
#endif
````
The code requiring 64 bits can be disabled by testing macro CURL_HAVE_UINT64

**Comment by Karlson2k** [comment]
@bagder I found that `inline` is not correctly supported by curl build system.
While configure has correct test and substitution for `inline` keyword (see https://github.com/curl/curl/blob/07e5b3ea6012ec26ec6c9f4c01040e3b60640c8a/configure.ac#L500), CMake and others does not support it properly. This leads to build failures on MSVC 2008 and 2010.
I would suggest to add CMake code like
``` CMake
check_c_source_compiles("
  inline int foo_func(void){return 0;}
  int main(void)
  {
    return foo_func();
  }" HAVE_KEYWORD_INLINE)
if(not HAVE_KEYWORD_INLINE)
  check_c_source_compiles("
    __inline int foo_func(void){return 0;}
    int main(void)
    {
      return foo_func();
    }" HAVE_KEYWORD___INLINE)
endif()
```
The if both are not supported, add `-Dinlint=`, if only the second form is supported then `-Dinline=__inline`.

Also, for VS projects to work with old compilers, add somewhere in the code (some header included globally):
```C
#if !deinfed(inline) && defined(_MSC_VER)
#  if _MSC_VER < 1900
#    define inline __inline
#  endif
#endif
```

Currently curl code has `inline` keyword only for code that never compiled by MSVC.

Karlson2k:
Rebased and updated as my another PR #12903 has been merged.

lib/curl_sha512_256.c
bagder: > Note: for external API libcurl may use CURL_TYPEOF_CURL_OFF_T as 32 bit.

No, it only appears so (because the `system.h` header has not been updated). We have checks in the build that makes it fail if `curl_off_t` is not 64 bit. We can write code and rely on it being at least 64 bit. We require a working 64 bit datatype for building curl since a while back.

`curl_uint64_t` sounds excellent to my ears!

Karlson2k: > We can write code and rely on it being at least 64 bit. We require a working 64 bit datatype for building curl since a while back.
> 
> `curl_uint64_t` sounds excellent to my ears!

OK. Where to put this typedef? Which header is the best?
I think it should be internal-only and not to be exposed for public API.

bagder: Right, let's not make it public until there is a need for that. I think you can put it in `lib/curl_setup.h`, which is a pretty generic and widely included header file for the library build.

Karlson2k:
Tests failures seem to be unrelated.
@bagder Updated as you requested.

vszakats:
Regarding `inline`, unless there are strong reasons to spread this to all build-systems with detection and the rest (= complexity), wouldn't this code (or similar) make it work well-enough for most envs to start:
```c
#ifdef __GNUC__
#undef inline
#define inline __inline__
#elif defined(_MSC_VER)
#undef inline
#define inline __inline
#endif
```
Taken from the libssh2 source code.

Karlson2k:
> Regarding `inline`, unless there are strong reasons to spread this to all build-systems with detection and the rest (= complexity and build perf regression), wouldn't this code (or similar) make it work well-enough for most envs to start:
> 
> ```c
> #ifdef __GNUC__
> #undef inline
> #define inline __inline__
> #elif defined(_MSC_VER)
> #undef inline
> #define inline __inline
> #endif
> ```
> 
> Taken from the libssh2 source code.

No need to undef `inline` since it was detected correctly by `configure`. It is like breaking thing that was fixed.
Force inline should work better at this function is used too many times.

vszakats:
Yeah, so delete `#undef inline` and guard with `#ifndef inline`, or whatever is indicating that the thing has been manually detected upstream already.

vszakats:
This could work for both VS and CMake builds (maybe for autotools even):
```
#if !defined(inline)
#  ifdef __GNUC__
#    define inline __inline__
#  elif defined(_MSC_VER) && _MSC_VER < 1900
#    define inline __inline
#  endif
#endif
```

Karlson2k:
> This could work for both VS and CMake builds (maybe for autotools even):

Yep, for simple `inline` it would work OK.
I'd add CMake detection, like it's done in autotools.

However, this fix is a general fix for libcurl/curl.

For this particular code force inline is better. It can be also made a part of global header, if needed.
The current version works, tested on all supported compilers and platforms, reliable and future-proof (more future-proof than use of `__inline__` which is supported for backward-compatibility).

vszakats:
I'm against adding this to cmake if there is a reasonably good in-source solution for it, because adding each of these detections cripple build performance and spread redundant complexity to the build-tool level. Many targets need an in-source solution anyway.

If performance of this code is so critical to require `__forceinline`, it may be worth using an existing optimized implemention in the TLS backends. Otherwise letting the compiler do its best based on `inline` may be enough IMO.

Karlson2k:
It would be nice to use TLS-backend implementation of `SHA-512/256`, however. as I wrote earlier [here](https://github.com/curl/curl/pull/12897#issuecomment-1934462763) and [here](https://github.com/curl/curl/pull/12897#issuecomment-1934730920) there are just a few TLS-backends supporting `SHA-512/256`.

### I'd emphasize: `SHA-512/256` is rarely implemented by various libraries. `SHA-512` is widely supported, but it is quite different from `SHA-512/256`.

This PR does not introduce any global changes, expect `curl_uint64_t` type as it was requested by @bagder.
This PR **does not** require CMake changes. The suggested CMake changes are required for correct support of `inline` which is broken in libcurl globally (I think mainly because of old MS VC compilers) and not used by code in this PR.

Support of "force inline" is implemented carefully, in a safe way. It is used only on specific (whitelisted) compilers and will not break other compilers by falling back to simple static function.

Karlson2k:
Updated with even more careful and correct detection of supported `inline` keywords.
No changes to CMake, everything in the source code.

Karlson2k:
Note: this code may break some preprocessors
``` C
#if defined(SOME_MACRO) && SOME_MACRO >= 3
#define SOMETHING something
#endif
```
The correct way is
``` C
#ifdef SOME_MACRO
#if SOME_MACRO >= 3
#define SOMETHING something
#endif
#endif
```

Making comparison against C-like constants is an extension.
Incorrect:
``` C
#if SOME_VALUE >= 5000000L
#define SOMETHING something
#endif
```
Correct:
``` C
#if SOME_VALUE >= 5000000
#define SOMETHING something
#endif
```

bagder:
> Note: this code may break some preprocessors

I would argue that is a broken preprocessor. We use this construct in several places in the code already and so far no one has reported a problem with this.

Karlson2k:
> > Note: this code may break some preprocessors
> 
> I would argue that is a broken preprocessor. We use this construct in several places in the code already and so far no one has reported a problem with this.

Or can be just undetected...

No problem, let me know if you prefer a shorted merged preprocessor constructs and I'll adapt the code.

Karlson2k:
OK, macros are merged and simplified.

bagder:
Thanks!

PR #12898, opened 2024-02-08
Client Writer cw-out refactor

icing:
Refactoring of the client writer that passes the data to the client/application's callback functions.

- split out into own source `cw-out.[ch]` from `sendf.c`
- move `tempwrite` and `tempcount` from `data->state` into the context of the client writer
- redesign the 3 `tempwrite` dynbufs as a linked list of dynbufs. On paused transfers, this allows to "record" interleaved HEADER/BODY chunks to be "played back" in the same order on unpausing.
- keep the overall size limit of all buffered data to `DYN_PAUSE_BUFFER`. On exceeding that, return `CURLE_TOO_LARGE` instead of `CURLE_OUT_OF_MEMORY` as before.
- add method to be called when a transfer is `DONE` to allow writing of any data still buffered
- when paused, record HEADER writes exactly as they come for later playback. HEADERs are documented to be written one-by-one.

### body write buffering

The implementation is capable of buffering body output and aggregate date into larger writes. This is disabled since applications expect to see data at the time it arrives. Should the need from apps to
have curl aggregate date first, we'd need to add a `CURLOPT_` for this.

### DIFF
--- a/lib/Makefile.inc
+++ b/lib/Makefile.inc
@@ -138,6 +138,7 @@ LIB_CFILES =         \
   curl_sspi.c        \
   curl_threads.c     \
   curl_trc.c         \
+  cw-out.c           \
   dict.c             \
   doh.c              \
   dynbuf.c           \
@@ -283,6 +284,7 @@ LIB_HFILES =         \
   curl_threads.h     \
   curl_trc.h         \
   curlx.h            \
+  cw-out.h           \
   dict.h             \
   doh.h              \
   dynbuf.h           \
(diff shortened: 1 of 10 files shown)


jay:
All data should be passed to the client immediately if the transfer is not paused. There are clients that process HTTP responses line by line (usually sent by chunk) and stay connected for hours. Even in other cases I'd never want to delay writing to the client.

icing:
@jay thanks for your input. That was my suspicion. I will change this PR, not doing this by default and think about adding an option for applications that prefer larger writes.

PR #12903, opened 2024-02-08
test 1165: improve patter matching

Karlson2k:
* Fix possible part of the symbol match.
* Fix excluded digits at the end of the symbols (`CURL_DISABLE_POP3` was checked as `CURL_DISABLE_POP`)

Discovered while working on PR #12897

### DIFF
--- a/tests/test1165.pl
+++ b/tests/test1165.pl
@@ -44,7 +44,7 @@ sub scanconf {
     my ($f)=@_;
     open S, "<$f";
     while(<S>) {
-        if(/(CURL_DISABLE_[A-Z_]+)/g) {
+        if(/(CURL_DISABLE_[A-Z0-9_]+)/g) {
             my ($sym)=($1);
             $disable{$sym} = 1;
         }
@@ -67,9 +67,9 @@ sub scanconf_cmake {
     my ($f)=@_;
     open S, "<$f";
     while(<S>) {
-        if(/(CURL_DISABLE_[A-Z_]+)/g) {
+        if(/(CURL_DISABLE_[A-Z0-9_]+)/g) {
             my ($sym)=($1);
-            if(not $sym =~ /(CURL_DISABLE_INSTALL|CURL_DISABLE_TESTS|CURL_DISABLE_SRP)/) {
+            if(not $sym =~ /^(CURL_DISABLE_INSTALL|CURL_DISABLE_TESTS|CURL_DISABLE_SRP)$/) {
                 $disable_cmake{$sym} = 1;
             }
         }
@@ -85,7 +85,7 @@ sub scan_file {
     my ($source)=@_;
     open F, "<$source";
     while(<F>) {
-        while(s/(CURL_DISABLE_[A-Z_]+)//) {
+        while(s/(CURL_DISABLE_[A-Z0-9_]+)//) {
             my ($sym)=($1);
             $file{$sym} = $source;
         }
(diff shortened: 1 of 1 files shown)


Karlson2k:
Two failed checks seems to be unrelated.

tests/test1165.pl
```diff
@@ -44,7 +44,7 @@ sub scanconf {
     my ($f)=@_;
     open S, "<$f";
     while(<S>) {
-        if(/(CURL_DISABLE_[A-Z_]+)/g) {
+        if(/(?:^|[^A-Za-z0-9\-])(CURL_DISABLE_[A-Z0-9_]+)(?:$|[^A-Za-z0-9\-])/g) {
```
bagder, line 47: This cannot be correct, since it puts the wrong part into `$1`.

Why not just extend the pattern? Surely it doesn't do any false positive matches elsewhere?
~~~perl
 if(/(CURL_DISABLE_[A-Z0-9_]+)/g) {

tests/test1165.pl
```diff
@@ -67,9 +67,9 @@ sub scanconf_cmake {
...
```
bagder, line 70: Same here

tests/test1165.pl
```diff
@@ -85,7 +85,7 @@ sub scan_file {
     my ($source)=@_;
     open F, "<$source";
     while(<F>) {
-        while(s/(CURL_DISABLE_[A-Z_]+)//) {
+        while(s/(?:^|[^A-Za-z0-9\-])(CURL_DISABLE_[A-Z0-9_]+)(?:$|[^A-Za-z0-9\-])//) {
```
bagder, line 88: and here

tests/test1165.pl
Karlson2k: Nope. :)
It is non-capturing group because of `?:`
`(?:^|[^A-Za-z0-9\-])` is *matched* but not *captured*.
The same for `(?:$|[^A-Za-z0-9\-])`.

It works correctly, I tested it with modified script version with printing all matched results.
Right now there are no false positives here even without the start and end patterns. But this is the test, it **must** catch incorrect code changes.

tests/test1165.pl
Karlson2k: It works as well.
I don't create PRs without proper local testing.
I added debug printing of the results and results are fine.

tests/test1165.pl
Karlson2k: This one is correct too.

tests/test1165.pl
bagder: Ok thanks, that goes beyond what I knew. I still think it's overly complicated for no good reason.

Karlson2k:
@bagder The code works correctly.
You don't have questions about documentation pattern matching. It must get all values from documentation. As tests are passed everywhere, it means that matching symbols were found in autoconf, cmake and sources. Otherwise tests would fail.
These non-capturing groups are actually a poor-man replacement for `\b` (word boundary) which was added in Perl 5.22 As a bonus - it does not depend on used locale.

Let me know if you still thing that code should be simplified by removing non-capturing groups.

Karlson2k:
Force-pushed with correct commit text (and rebased)

bagder:
> Let me know, please, if you still think that code should be simplified by removing non-capturing groups.

I think so. The regexes should be as simple as possible to help readers understand then.

Karlson2k:
Updated with non-capturing groups removed, as requested.

Karlson2k:
Two build failures seem to be unrelated.

bagder (approved):

bagder:
Thanks!

PR #12907, opened 2024-02-08
linux.yml: add fedora container

pheiduck:
this will replace: https://github.com/curl/curl/pull/12887

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -183,7 +183,7 @@ jobs:
             singleuse: --unit
 
           - name: Fedora Latest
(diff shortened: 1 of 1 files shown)


dfandrich:
My comment from #12887 stands here as well: what does this buy us for the cost to maintain it?

pheiduck:
> My comment from #12887 stands here as well: what does this buy us for the cost to maintain it?

Coverage of tests on other Distros?
But yes maybe this make less sense due people can report there issues if there have some on there distro.

dfandrich:
There are too many distro for this to scale.  I still think adding one rolling release would be a net benefit.

PR #12922, opened 2024-02-10
cmake: fix syntax for older CMake versions

### LINKED ISSUE #12920: Build failure with cmake 3.19.2
### I did this

Trying to configure 8.6.0 (or master) using cmake 3.19.2 results in the error:
```
-- Enabled SSL backends: OpenSSL
-- Configuring done
CMake Error:
  Error evaluating generator expression:

    $<LIST:TRANSFORM,curl_easy_cleanup.3;curl_easy_duphandle.3;curl_easy_escape.3;curl_easy_getinfo.3;curl_easy_header.3;curl_easy_init.3;curl_easy_nextheader.3;curl_easy_option_by_id.3;curl_easy_option_by_name.3;curl_easy_option_next.3;curl_easy_pause.3;curl_easy_perform.3;curl_easy_recv.3;curl_easy_reset.3;curl_easy_send.3;curl_easy_setopt.3;curl_easy_strerror.3;curl_easy_unescape.3;curl_easy_upkeep.3;curl_escape.3;curl_formadd.3;curl_formfree.3;curl_formget.3;curl_free.3;curl_getdate.3;curl_getenv.3;curl_global_cleanup.3;curl_global_init.3;curl_global_init_mem.3;curl_global_trace.3;curl_global_sslset.3;curl_mime_addpart.3;curl_mime_data.3;curl_mime_data_cb.3;curl_mime_encoder.3;curl_mime_filedata.3;curl_mime_filename.3;curl_mime_free.3;curl_mime_headers.3;curl_mime_init.3;curl_mime_name.3;curl_mime_subparts.3;curl_mime_type.3;curl_mprintf.3;curl_multi_add_handle.3;curl_multi_assign.3;curl_multi_cleanup.3;curl_multi_fdset.3;curl_multi_get_handles.3;curl_multi_info_read.3;curl_multi_init.3;curl_multi_perform.3;curl_multi_poll.3;curl_multi_remove_handle.3;curl_multi_setopt.3;curl_multi_socket.3;curl_multi_socket_action.3;curl_multi_socket_all.3;curl_multi_strerror.3;curl_multi_timeout.3;curl_multi_wakeup.3;curl_multi_wait.3;curl_pushheader_bynum.3;curl_pushheader_byname.3;curl_share_cleanup.3;curl_share_init.3;curl_share_setopt.3;curl_share_strerror.3;curl_slist_append.3;curl_slist_free_all.3;curl_strequal.3;curl_strnequal.3;curl_unescape.3;curl_url.3;curl_url_cleanup.3;curl_url_dup.3;curl_url_get.3;curl_url_set.3;curl_url_strerror.3;curl_version.3;curl_version_info.3;curl_ws_meta.3;curl_ws_recv.3;curl_ws_send.3;libcurl-easy.3;libcurl-env.3;libcurl-env-dbg.3;libcurl-errors.3;libcurl-multi.3;libcurl-security.3;libcurl-share.3;libcurl-symbols.3;libcurl-thread.3;libcurl-tutorial.3;libcurl-url.3;libcurl-ws.3;libcurl.3,PREPEND,/home/dan/src/curl-git/build/docs/libcurl/>

  Expression did not evaluate to a known generator expression
```

~~Bisecting shows commit a808aab06 is at fault~~. This version of cmake is only a couple of years old and the documentation says cmake 3.7 is the earliest supported version, so I would expect this to work.

### I expected the following

cmake should run without error.

### curl/libcurl version

curl 8.6.0

### operating system

Linux x86_64

levitte:
A generator expression was added in some CMakeLists.txts, that was a bit
too modern.  Forunately, a foreach() loop can produce the same result
relatively easily.

Fixes #12920


levitte:
Actually, please hold on a bit.  The fix may be even easier...

levitte:
Nope, never mind.  This is good to go.

dfandrich:
There's another LIST:TRANSFORM in docs/libcurl/opts/CMakeLists.txt

levitte:
> There's another LIST:TRANSFORM in docs/libcurl/opts/CMakeLists.txt

Thanks for the reminder.  I fixed that in the same way

vszakats:
Proposing to move the transformation into a separate loop to keep a single `install()` invocation:
```diff
--- a/docs/libcurl/CMakeLists.txt
+++ b/docs/libcurl/CMakeLists.txt
@@ -62,7 +62,11 @@ add_custom_command(OUTPUT libcurl-symbols.md
 add_manual_pages(man_MANS)
 add_custom_target(man ALL DEPENDS ${man_MANS})
 if(NOT CURL_DISABLE_INSTALL)
-  install(FILES "$<LIST:TRANSFORM,${man_MANS},PREPEND,${CMAKE_CURRENT_BINARY_DIR}/>"
+  unset(_src)
+  foreach(_f ${man_MANS})
+    list(APPEND _src "${CMAKE_CURRENT_BINARY_DIR}/${_f}")
+  endforeach()
+  install(FILES ${_src}
           DESTINATION ${CMAKE_INSTALL_MANDIR}/man3)
 endif()
 
diff --git a/docs/libcurl/opts/CMakeLists.txt b/docs/libcurl/opts/CMakeLists.txt
index 82844a225..a4e2ec103 100644
--- a/docs/libcurl/opts/CMakeLists.txt
+++ b/docs/libcurl/opts/CMakeLists.txt
@@ -29,6 +29,10 @@ add_manual_pages(man_MANS)
 add_custom_target(opts-man DEPENDS ${man_MANS})
 add_dependencies(man opts-man)
 if(NOT CURL_DISABLE_INSTALL)
-  install(FILES "$<LIST:TRANSFORM,${man_MANS},PREPEND,${CMAKE_CURRENT_BINARY_DIR}/>"
+  unset(_src)
+  foreach(_f ${man_MANS})
+    list(APPEND _src "${CMAKE_CURRENT_BINARY_DIR}/${_f}")
+  endforeach()
+  install(FILES ${_src}
           DESTINATION ${CMAKE_INSTALL_MANDIR}/man3)
 endif()
```

With the GNU-make generator there is no perceivable performance difference and the log output is the same, but in general I find it better to bundle up file operations if possible.

levitte:
@vszakats, judging from my `_build/docs/libcurl/cmake_install.cmake`, your proposal would make no difference, 'cause with this PR *as it is*, I see this:

``` cmake
if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT
)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man3" TYPE FILE FI
LES
    "/home/levitte/gitwrk/github.com/curl/curl/docs/libcurl/curl_easy_cleanup.3"
    "/home/levitte/gitwrk/github.com/curl/curl/docs/libcurl/curl_easy_duphandle.3"
    ...
```

So I'm going to assume that CMake is pretty good at collecting information of files to be installed per DESTINATION, and generate sensible installation scripts.

Now granted, this is with a modern CMake (3.28), and I don't know for sure how far back CMake was this kind of smart, but generally speaking, I would avoid that sort of optimization if not necessary.

levitte:
I took a moment and made an installation of CMake 3.7.2, and lo'n'behold, it was less smart.  `_build/docs/libcurl/cmake_install.cmake` now looks like this for the same thing:

``` cmake
if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man3" TYPE FILE FILES "/home/levitte/gitwrk/github.com/curl/curl/_build/docs/libcurl/curl_easy_cleanup.3")
endif()

if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man3" TYPE FILE FILES "/home/levitte/gitwrk/github.com/curl/curl/_build/docs/libcurl/curl_easy_duphandle.3")
endif()

```

> ... but in general I find it better to bundle up file operations if possible.

Yeah ok.  I'm not sure I agree that making CMakeLists.txt a bit uglier to read is worth it, but sure.

levitte:
This PR now has @vszakats's suggestion

vszakats (approved):

jay:
Thanks

PR #12924, opened 2024-02-10
.gitattributes: set .checksrc as LF

Karlson2k:
This PR fixes debug builds on Windows.

CRLF line ending in `.checksrc` files is not handled by `scripts/checksrc.pl`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,6 +9,7 @@ configure.ac eol=lf
 *.in eol=lf
 *.am eol=lf
 *.sh eol=lf
+.checksrc eol=lf
 *.[ch] whitespace=tab-in-indent
 
 # Batch files (bat,btm,cmd) must be run with CRLF line endings.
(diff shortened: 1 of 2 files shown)


jay:
I notice all other file reads in checksrc are handled with `$windows_os ? $_ =~ s/\r?\n$// : chomp;` so I propose we do that instead:

~~~diff
diff --git a/scripts/checksrc.pl b/scripts/checksrc.pl
index 76f4660..be085e0 100755
--- a/scripts/checksrc.pl
+++ b/scripts/checksrc.pl
@@ -120,6 +120,7 @@ sub readlocalfile {
     open(my $rcfile, "<", "$dir/.checksrc") or return;
 
     while(<$rcfile>) {
+        $windows_os ? $_ =~ s/\r?\n$// : chomp;
         $i++;
 
         # Lines starting with '#' are considered comments
~~~

Karlson2k:
We can use both fixes.
There is no reason to keep `.checksrc` with CRLF ending of Windows.

jay:
Thanks. I landed this but omitted the gitattributes change. IMO there's no reason to have a special exception since it can now be handled like the other input files.

Issue #12921, opened 2024-02-10
Man page installation and distribution


michaelforney:
### I did this

Recently there's been a lot of changes around man pages concerning generation, installation, distribution, and build-by-default.

First, here's my understanding of some automake variables used in curl:
- `man_MANS`: Man pages listed here will get built by default and installed, but not distributed
- `dist_man_MANS`: Man pages listed here will get built by default, installed, and distributed.
- `noinst_man_MANS`: This doesn't actually do anything. I don't see any reference to it in automake or the generated Makefile.
- `nodist_MANS`: This doesn't do anything either. Looking at automake source, there *is* `nodist_man_MANS`, which is the same as `man_MANS` since man pages aren't distributed by default (see https://www.gnu.org/software/automake/manual/html_node/Man-Pages.html). The comment in the source says "We handle nodist_ for uniformity.  man pages aren't distributed by default so it isn't actually very important".
- `EXTRA_DIST`: Files here are distributed, but not installed or built by the default target.

Here's a summary of the recent history of a few man pages:
- Before 8.6.0, `mk-ca-bundle.1` was not generated, but it was distributed since it was in `noinst_man_MANS`, which was present in `EXTRA_DIST`.
  * 8.6.0 included eefcc1bda4bccd800f5a56a0fe17a2f44a96e88b, which added it to `man_MANS`, making it get installed and built by default. Since `noinst_man_MANS` was dropped from `EXTRA_DIST`, it no longer got distributed.
  * a911f4ffe66b42b4bbb4fae0e6b79484cbe80071 removed it from `man_MANS`, so it was no longer installed or built by default. The only reference left to `mk-ca-bundle.1` at this point was through `noinst_man_MANS`, a bogus variable not used by anything, and `CLEANFILES`.
  * 755b31da10dacca94e0d00029a7306e09e3fe913 added it to `dist_man_MANS`, making it get installed once again, distributed, and built by default. This is the current state.

  It seems to me that the intent is for it to be built by default and distributed, but not installed. I think the way to do this is to remove `mk-ca-bundle.1` from `dist_man_MANS`, add it to `EXTRA_DIST` to distribute it, and add a dependency `all-local: $(MK_CA_DOCS)` to build it by default.

- ~~Before 8.6.0, `curl.1` was distributed since it was present in `noinst_man_MANS`, which was added to `EXTRA_DIST`.~~
[...]

PR #12925, opened 2024-02-11
 #12921-Man page installation and distribution

sounar97:
Tried to understand and resolve the "Man page installation and distribution #12921"issue
Don't know if it's the right way

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -34,7 +34,6 @@ endif
 # EXTRA_DIST breaks with $(abs_builddir) so build it using this variable
 # but distribute it (using the relative file name) in the next variable
 man_MANS = $(abs_builddir)/curl.1
-noinst_man_MANS = $(MK_CA_DOCS)
 dist_man_MANS = $(CURLCONF_DOCS) $(MK_CA_DOCS)
 CURLPAGES = curl-config.md mk-ca-bundle.md
 
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -43,7 +43,7 @@ SUBDIRS = . cmdline-opts libcurl
 DIST_SUBDIRS = $(SUBDIRS) examples
 
 CLEANFILES = $(man_MANS) curl.1 $(CURLCONF_DOCS) $(MK_CA_DOCS)
-nodist_MANS = $(CLEANFILES)
+
 
 EXTRA_DIST =                                    \
  $(CURLPAGES)                                   \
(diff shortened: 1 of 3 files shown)


docs/Makefile.am
```diff
@@ -25,16 +25,15 @@
 AUTOMAKE_OPTIONS = foreign no-dependencies
 
 if BUILD_DOCS
```
michaelforney, line 27: Based on @bagder's last point in https://github.com/curl/curl/issues/12921#issuecomment-1937005261, I think this conditional should be dropped. We always want to distribute `mk-ca-bundle.1` and we always want to install and distribute `curl-config.1`.

I'm not sure if we want to add the `all-local` line or not. I mentioned it in the issue, since https://github.com/curl/curl/commit/755b31da10dacca94e0d00029a7306e09e3fe913 said "which also makes it get built by default". If being built upon `make dist` is enough, then it is not needed.

There is already an `EXTRA_DIST` assignment later on in the file, so `mk-ca-bundle.1` should be moved there (or perhaps `CURLPAGES`?), otherwise this one just gets overriden.

docs/Makefile.am
```diff
@@ -25,16 +25,15 @@
...
+all-local: $(MK_CA_DOCS)
 CURLCONF_DOCS = curl-config.1
 endif
 
 
 # EXTRA_DIST breaks with $(abs_builddir) so build it using this variable
 # but distribute it (using the relative file name) in the next variable
 man_MANS = $(abs_builddir)/curl.1
-noinst_man_MANS = $(MK_CA_DOCS)
 dist_man_MANS = $(CURLCONF_DOCS) $(MK_CA_DOCS)
```
michaelforney, line 37: We don't want to install `mk-ca-bundle.1`, so `$(MK_CA_DOCS)` should be removed here. It will be built by `all-local`, and distributed via `EXTRA_DIST`.

Also, distributing and installing `curl-config.1` should be done always, so we can replace `$(CURLCONF_DOCS)` here with `curl-config.1`.

michaelforney (commented):
We also need to change the `install-data-hook` in `Makefile.am` to unconditionally install in `docs`. Disabled `BUILD_DOCS` should only prevent installation in `docs/libcurl`

bagder:
Argh. I'm sorry I did not pay proper attention on your work here. I have now made #12959 which I believe fixes this issue (and a little more).

bagder:
Fixed in #12959

PR #11510, opened 2023-07-24
CI: enable parallel testing in CI builds

dfandrich:
The test-ci target now uses 2 processes by default, but the amount of
parallelism is tuned for each CI service and build environment based on
results of a number of test runs.  Some CI services use super-
oversubscribed build machines that can barely run the curl tests
already with no parallelism without frequently failing with
timing-induced failures. These continue to be run without parallelism.
Other services provide two fast, unloaded cores and these run with 14
processes, which is a good default for this kind of environment.

Here's a summary of the number of test processes by CI service:

  Appveyor - 2 (Windows MSVC), 1 (others)
  Azure - 2
  Circle CI - 14
  Cirrus - 28 (macOS), 14 (Linux), 7 (FreeBSD), 5 (macOS torture), 2 (Windows)
  GitHub Actions - 3 (macOS), 2 (Linux)

Some of these are a bit conservative to keep timing-induced flakiness down.

The net result is that the first test results should arrive only
3 minutes after a commit submission.

Changes merged via separate commits:
- 2a7c8b27fdd266894fe24d6d0f5d853ea843dff4 #14171
- 72341068a2d8f1b94a26add16830c725cc4054e6
- efce544418971b064d9fc8183a6caa582c54559b #14244
- c6cf411bacf97f230db160e543d0cc3bbe5b9aba

Ref: #10818
Closes #11510

PR #12927, opened 2024-02-12
GHA: adjust parallel job counts

Ayesh:
Adjusts the `make -j` flag to match the latest GitHub-hosted runner
hardware specs[^1]:

 - `ubuntu-latest` on 4 CPU cores
 - `macos-latest` on 3 CPU cores

The processor count is ideally obtained from `nproc`, but setting env
vars from the current CI yaml files is not possible because they expect
literal strings.

[^1]: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Closes #12927

### DIFF
--- a/.github/workflows/awslc.yml
+++ b/.github/workflows/awslc.yml
@@ -41,7 +41,7 @@ concurrency:
 permissions: {}
 
 env:
-  MAKEFLAGS: -j 3
+  MAKEFLAGS: -j 4
 
   # renovate: datasource=github-tags depName=awslabs/aws-lc versioning=semver registryUrl=https://github.com
   awslc-version: 1.28.0
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -40,7 +40,7 @@ concurrency:
 permissions: {}
 
 env:
-  MAKEFLAGS: -j 3
+  MAKEFLAGS: -j 4
   # unhandled
   bearssl-version: 0.6
   # renovate: datasource=github-tags depName=libressl-portable/portable versioning=semver registryUrl=https://github.com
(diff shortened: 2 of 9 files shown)


pheiduck:
Why not using `$(nproc)`?

Ayesh:
The root level `env` keys cannot have variables expanded, so they have to be hardcoded or taken from the context. At least I couldn't figure it out. 

We could setup the env var before calling `make`, but I wanted to keep the change set minimal.

dfandrich:
See PR #11510. This hasn't been enabled already because of reliability issues
on the CI systems. But, there have been a few issues fixed since I tried last,
so it will be interesting to see CI results of this PR.

Ayesh:
Thank you @dfandrich - I saw your commit that added parallel CI jobs and the discussions around it.
 
GitHub recently [doubled](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/) the specs of their hosted runners, so this one tries to take advantage of it. I could run this locally with `make -j16`, and the PR seems to pass CI too, so perhaps the issues are now ironed out?

bagder:
@dfandrich so do we close this as a duplicate of #11510 ?

dfandrich:
Sorry, I missed responding to the comment but more importantly missed that this was adjusting *build* jobs and not *test* jobs. This change should be conceptually fine. I've found that GHA machines can be a bit overloaded sometimes, so I could see how increasing the number of build processes might not always reduce the build time, but it should probably be done anyway to take advantage of times of low load.

vszakats:
> This change should be conceptually fine. I've found that GHA machines can be a bit overloaded sometimes, so I could see how increasing the number of build processes might not always reduce the build time, but it should probably be done anyway to take advantage of times of low load.

Agreed.

vszakats:
Thank you @Ayesh, merged now!

Issue #12831, opened 2024-01-31
curl verifies incorrect subject alternative name when contacting https proxy


HsiehYuho:
### I did this

I tried 
```
curl -x https://[proxy:ipv6]:8082 https://destination.hostname
```
It gives me SAN match failure. 
```
SSL:no alternative certificate subject name matches target hostname [proxy:ipv6]
```
However, the [proxy:ipv6] is in the proxy cert SAN field. 
When switching to 
```
curl -x https://[proxy:ipv6]:8082 https://[destination.ipv6]
```
the curl succeeds. 

We (my colleagues and me) are suspecting checks around here https://github.com/curl/curl/blob/master/lib/vtls/openssl.c#L2152 lead to that curl only checks the proxy SAN type that is same as the destination type 

### I expected the following

```
curl -x https://[proxy:ipv6]:8082 https://destination.hostname 
curl -x https://[proxy:ipv6]:8082 https://[destination.ipv6]
```
should at least give me the same result

### curl/libcurl version

curl 7.61.1 (x86_64-redhat-linux-gnu)

### operating system

Linux, but our company (meta) dev our own linux ..

HsiehYuho:
I didn't spend time working on a repro but we fairly sure that there is at least inconsistent output where there shouldn't be.

dfandrich:
That curl version is 5.4 years out of date. Does this reproduce with a recent
version?  What TLS library are you using?

HsiehYuho:
Hi Dan, 
Yes, it is reproducible using the latest curl 
```
./src/.libs/curl --version
curl 8.6.0-DEV (x86_64-pc-linux-gnu) 
```
we are using openssl.

icing:
@HsiehYuho I added a test case using '127.0.0.1' for the proxy and gave it a certificate with a x590 IPAddress in the alt names. This test works with curl 8.6.0.

Could you recheck your findings? Is there anything specific in your ipv6 address spec that we may not account for?

knekritz:
@icing I think the issue is specific to IPv6 because of the `conn->bits.ipv6_ip` check at https://github.com/curl/curl/blob/f8513ee62ec72af996139d2756beeac8a828afd8/lib/vtls/openssl.c#L2154

icing:
@knekritz thanks, that is exactly the right spot. Would you be able to verify that #12931 fixes the issue for you as well?

HsiehYuho:
[...]

PR #12931, opened 2024-02-13
proxy peer verification, use correct type

icing:
- refs #12831
- when verifying a proxy certificate for an ip address, use the correct ip family. That may be different from the "connection" ip family.

### DIFF
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -1574,6 +1574,11 @@ CURLcode Curl_ssl_peer_init(struct ssl_peer *peer, struct Curl_cfilter *cf)
     eport = cf->conn->remote_port;
   }
 
+  /* There is some inherited weirdness here. We should never get
+   * a hostname that is the empty string (I believe). We check
+   * for it, but it should not be necessary. Let's catch those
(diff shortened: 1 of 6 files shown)


lib/vtls/vtls.c
```diff
@@ -1589,8 +1594,8 @@ CURLcode Curl_ssl_peer_init(struct ssl_peer *peer, struct Curl_cfilter *cf)
     }
 
     peer->sni = NULL;
-    peer->is_ip_address = is_ip_address(peer->hostname)? TRUE : FALSE;
-    if(peer->hostname[0] && !peer->is_ip_address) {
+    peer->type = get_peer_type(peer->hostname);
+    if(peer->type == CURL_SSL_PEER_DNS) {
```
jay, line 1598: I notice you removed the 0 length hostname check for SNI, I assume because it's not possible

icing: Good catch. I re-added the check. I *believe* is is not necessary, but stranger things have happened before. Added a DEBUGASSERT to catch those situations in debug builds.

jay:
Thanks. I removed the comments above the DEBUGASSERT because it's implied in the assert.

Issue #12750, opened 2024-01-21
curl on Windows is slow when output destination is on SMB


CaptainFlint:
### I did this

Make sure you have a file URL with high download speed; a local disk, and a network SMB resource with high upload speed (either mapped as a disk letter, or accessible via network share URI).
First, try to download the file with `--output` option pointing onto a file located on the local disk, and mark the speed (might need a few seconds for it to stabilize).
Second, try the same, but with `--output` pointing to the SMB share, and compare the speed.

In my example, I set up a local HTTP server using `python -mhttp.server` for a directory with some Linux ISO distributions on my local hard drive (max speed for downloading is 250-300 Mbytes/s); and for the destination I used an SMB share on my locally running CentOS 7 virtual machine (transfer speed varies, but is about 100-250 Mbytes/s). This way I made sure I'm not throttled by my ISP, or the upstream server. But my friend found this issue when downloading a file from Internet via 500 Mbps connection onto his NAS, so it doesn't have to be all local.

My results:
```
C:\Programs\curl>curl "http://127.0.0.1:8000/ubuntu-22.04.2-desktop-amd64.iso" -o c:\Temp\a.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 4699M  100 4699M    0     0   244M      0  0:00:19  0:00:19 --:--:--  255M
```
```
C:\Programs\curl>curl "http://127.0.0.1:8000/ubuntu-22.04.2-desktop-amd64.iso" -o \\192.168.56.101\flint\tmp\a.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 4699M  100 4699M    0     0  22.6M      0  0:03:27  0:03:27 --:--:-- 22.7M
```
In other words, the speed is over 10 times lower when saving onto SMB. To confirm that the SMB itself is not slow, here is the measurement of direct copy of the same file to the same SMB location:
```
C:\Programs\curl>dir D:\Distribs\OS\ubuntu-20.04.6-live-server-amd64.iso & echo %TIME% & copy D:\Distribs\OS\ubuntu-20.04.6-live-server-amd64.iso \\192.168.56.101\flint\tmp\a.iso & cmd /v:on /c "echo !TIME!"
 Volume in drive D has no label.
 Volume Serial Number is 2461-9C46

 Directory of D:\Distribs\OS

[...]

PR #12932, opened 2024-02-13
file, use xfer buf for transfers

icing:
- refs #12750
- borrow multi handle's transfer buffer for up- and downloads of file: urls.
- restores the buffer size behaviour of curl 8.5.0


jay:
Thanks

PR #12933, opened 2024-02-13
WIP, OpenSSL QUIC, adaptation for v3.3.x

icing:
- set our idle timeout as transport parameter
- query negotiated idle timeout for connection alive checks
- query number of available bidi streams on a connection
- use write_ex2 with SSL_WRITE_FLAG_CONCLUDE to signal EOF on last chunk write, so stream close does not require an additional QUIC packet


icing:
Since this continues to work with OpenSSL v3.2.x, this can be merged.

Issue #12937, opened 2024-02-14
Spurious message: "Leftovers after chunking: X bytes"


mkauf:
### I did this

curl 8.6.0 shows the verbose message "Leftovers after chunking: X bytes" when it receives a response with "Transfer-Encoding: chunked".

Example:
```
$ curl -s -v --http1.1 https://echo.free.beeceptor.com > /dev/null
[...]
< HTTP/1.1 200 OK
< Alt-Svc: h3=":443"; ma=2592000
< Content-Type: application/json
< Date: Wed, 14 Feb 2024 14:40:48 GMT
< Vary: Accept-Encoding
< Transfer-Encoding: chunked
< 
{ [5 bytes data]
* Leftovers after chunking: 5 bytes
* Connection #0 to host echo.free.beeceptor.com left intact
```

"git bisect" found commit d7b6ce64ce0ad787ad2ed3ee05c94938a6b4f551

### I expected the following

No message "Leftovers after chunking".

curl 8.2.1 does not have the bug:
```
curl -s -v --http1.1 https://echo.free.beeceptor.com > /dev/null
[...]
< HTTP/1.1 200 OK
< Alt-Svc: h3=":443"; ma=2592000
< Content-Type: application/json
< Date: Wed, 14 Feb 2024 14:47:05 GMT
< Vary: Accept-Encoding
< Transfer-Encoding: chunked
< 
{ [5 bytes data]
* Connection #0 to host echo.free.beeceptor.com left intact
```

### curl/libcurl version

curl 8.6.0

### operating system

Fedora Linux 39

icing:
I do not see that, given your example, on my machine, but it may be some timing dependant thing. Will investigate further.

mkauf:
@icing Great, thanks! I have tested https://github.com/curl/curl/pull/12939, the message does not appear anymore :-)

jay:
Thanks

PR #12939, opened 2024-02-14
chunked decoding, fix accounting

icing:
- refs #12937
- chunks were handled correctly, but accounting was borked so that "left" bytes were reported on the last chunk written.


jay:
This looks finished to me but I notice it's in draft, is there a reason for that?

icing:
> This looks finished to me but I notice it's in draft, is there a reason for that?

No, just that github uses the setting from the last PR and I did miss that.

jay:
Thanks

PR #12945, opened 2024-02-15
websocket, fix curl_ws_recv()

icing:
- when data arrived in several chunks, the collection into the passed buffer always started at offset 0, overwriting the data already there.

### DIFF
--- a/lib/ws.c
+++ b/lib/ws.c
@@ -886,7 +886,7 @@ static ssize_t ws_client_collect(const unsigned char *buf, size_t buflen,
       return -1;
     }
     *err = CURLE_OK;
-    memcpy(ctx->buffer, buf, nwritten);
+    memcpy(ctx->buffer + ctx->bufidx, buf, nwritten);
     ctx->bufidx += nwritten;
   }
   return nwritten;


calvin2021y:
maybe a unit tests can prevent this kind error.

icing:
> maybe a unit tests can prevent this kind error.

It was triggered in pytest `test_20_06`, but it depended on how data arrived from the server. Any suggestion for a reliable test appreciated.

calvin2021y:
maybe use `netcat` and `pv` to limit the send speed of a exists tcp stream

```sh
nc -l -p 8888 | pv -L 512  | nc example.com 80
```

in this example pv limit the stream speed at 512 byte per second.

jay:
> It was triggered in pytest `test_20_06`, but it depended on how data arrived from the server.

Would [CURL_DBG_SOCK_RMAX](https://curl.se/libcurl/c/libcurl-env-dbg.html) work for that

icing:
> > It was triggered in pytest `test_20_06`, but it depended on how data arrived from the server.
> 
> Would [CURL_DBG_SOCK_RMAX](https://curl.se/libcurl/c/libcurl-env-dbg.html) work for that

Added a new debug environment var `CURL_WS_CHUNK_SIZE` to influence the buffer dimensions used in en-/decoding for websockets. This exposes the bug in test_20-07 and verifies the fix.

bagder (approved):

PR #12946, opened 2024-02-15
Readwrite rework, SingleRequest reshuffling

icing:
### Curl_read/Curl_write clarifications

- replace `Curl_read()`, `Curl_write()` and `Curl_nwrite()` to
      clarify when and at what level they operate
- send/recv of transfer related data is now done via
      `Curl_xfer_send()/Curl_xfer_recv()` which no longer has
      socket/socketindex as parameter. It decides on the transfer
      setup of `conn->sockfd` and `conn->writesockfd` on which
      connection filter chain to operate.
- send/recv on a specific connection filter chain is done via
      `Curl_conn_send()/Curl_conn_recv()` which get the socket index
      as parameter.
- rename `Curl_setup_transfer()` to `Curl_xfer_setup()` for
      naming consistency
- clarify that the special CURLE_AGAIN handling to return
      `CURLE_OK` with length 0 only applies to `Curl_xfer_send()`
      and CURLE_AGAIN is returned by all other send() variants.

### SingleRequest reshuffling

- move functions into request.[ch]
- differentiate between reset and free
- add Curl_req_done() to perform last actions
- add a send bufq to SingleRequest for future
      use in keeping upload data


icing:
Closed. Replaved with #12963.

PR #12947, opened 2024-02-15
configure.ac: find libpsl with pkg-config

ffontaine:
Find libpsl with pkg-config to avoid the following static build failure:

```
configure:28830: /home/autobuild/autobuild/instance-6/output-1/host/bin/aarch64_be-buildroot-linux-musl-gcc -o conftest -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -O3 -g0 -static -Werror-implicit-function-declaration -Wno-system-headers -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -I/home/autobuild/autobuild/instance-6/output-1/host/aarch64_be-buildroot-linux-musl/sysroot/usr/include  -static -L/home/autobuild/autobuild/instance-6/output-1/host/bin/../aarch64_be-buildroot-linux-musl/sysroot/usr/lib  -L/home/autobuild/autobuild/instance-6/output-1/host/aarch64_be-buildroot-linux-musl/sysroot/usr/lib conftest.c -lpsl  -lmbedtls -lmbedx509 -lmbedcrypto -lz  -latomic >&5
/home/autobuild/autobuild/instance-6/output-1/host/lib/gcc/aarch64_be-buildroot-linux-musl/11.4.0/../../../../aarch64_be-buildroot-linux-musl/bin/ld: /home/autobuild/autobuild/instance-6/output-1/host/bin/../aarch64_be-buildroot-linux-musl/sysroot/usr/lib/libpsl.a(psl.c.o): in function `is_public_suffix':
psl.c:(.text+0x2a8): undefined reference to `idn2_lookup_u8'
```

[...]

```
checking for library containing psl_builtin... no
configure: error: libpsl was not found
```

Fixes:
 - http://autobuild.buildroot.org/results/1fb15e1a99472c403d0d3b1a688902f32e78d002

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2075,19 +2075,74 @@ dnl **********************************************************************
 dnl Check for libpsl
 dnl **********************************************************************
 
-AC_ARG_WITH(libpsl,
-           AS_HELP_STRING([--without-libpsl],
-           [disable support for libpsl]),
-           with_libpsl=$withval,
-           with_libpsl=yes)
-curl_psl_msg="no      (libpsl disabled)"
-if test $with_libpsl != "no"; then
-  AC_SEARCH_LIBS(psl_builtin, psl,
-    [curl_psl_msg="enabled";
-     AC_DEFINE([USE_LIBPSL], [1], [PSL support enabled])
-     ],
-    [AC_MSG_ERROR([libpsl was not found]) ]
+dnl Default to compiler & linker defaults for LIBPSL files & libraries.
+OPT_LIBPSL=off
+AC_ARG_WITH(libpsl,dnl
+AS_HELP_STRING([--with-libpsl=PATH],[Where to look for libpsl, PATH points to the LIBPSL installation; when possible, set the PKG_CONFIG_PATH environment variable instead of using this option])
+AS_HELP_STRING([--without-libpsl], [disable LIBPSL]),
+  OPT_LIBPSL=$withval)
+
+if test X"$OPT_LIBPSL" != Xno; then
+  dnl backup the pre-libpsl variables
+  CLEANLDFLAGS="$LDFLAGS"
+  CLEANCPPFLAGS="$CPPFLAGS"
+  CLEANLIBS="$LIBS"
+
+  case "$OPT_LIBPSL" in
+  yes)
+    dnl --with-libpsl (without path) used
+    CURL_CHECK_PKGCONFIG(libpsl)
+
+    if test "$PKGCONFIG" != "no" ; then
+      LIB_PSL=`$PKGCONFIG --libs-only-l libpsl`
+      LD_PSL=`$PKGCONFIG --libs-only-L libpsl`
+      CPP_PSL=`$PKGCONFIG --cflags-only-I libpsl`
+    else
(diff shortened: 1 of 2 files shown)


dfandrich:
The test failures look relevant, probably related to removing `AM_CONDITIONAL`.

ffontaine:
Thanks for your review, PR was updated

configure.ac
```diff
@@ -2075,19 +2075,75 @@ dnl **********************************************************************
...
-if test $with_libpsl != "no"; then
-  AC_SEARCH_LIBS(psl_builtin, psl,
-    [curl_psl_msg="enabled";
-     AC_DEFINE([USE_LIBPSL], [1], [PSL support enabled])
-     ],
-    [AC_MSG_ERROR([libpsl was not found]) ]
+dnl Default to compiler & linker defaults for LIBPSL files & libraries.
+OPT_LIBPSL=off
+AC_ARG_WITH(libpsl,dnl
+AS_HELP_STRING([--with-libpsl=PATH],[Where to look for libpsl, PATH points to the LIBPSL installation; when possible, set the PKG_CONFIG_PATH environment variable instead of using this option])
```
jay, line 2081: you can change this to "where to look for libpsl, PATH points to the installation root". it's already known you can use PKG_CONFIG_PATH with autotools, we don't say it in the help text of every option

jay: > we don't say it in the help text of every option

disregard. actually we do say it for some options where pkg-config is preferred, i didn't expect that.

configure.ac
```diff
@@ -2075,19 +2075,75 @@ dnl **********************************************************************
...
+    CURL_CHECK_PKGCONFIG(libpsl)
+
+    if test "$PKGCONFIG" != "no" ; then
+      LIB_PSL=`$PKGCONFIG --libs-only-l libpsl`
+      LD_PSL=`$PKGCONFIG --libs-only-L libpsl`
+      CPP_PSL=`$PKGCONFIG --cflags-only-I libpsl`
+    else
+      dnl To avoid link errors, we do not allow --libpsl without
+      dnl a pkgconfig file
+      AC_MSG_ERROR([--libpsl was specified but could not find libpsl pkgconfig file.])
```
jay, line 2103: I assume you mean --with-libpsl and why would we not want to allow it without pkg-config?

ffontaine: I'll rename `--libpsl` to `--with-libpsl`.
If `--with-libpsl` is provided, but pkg-config file is not found, an error is raised as already done by libcurl when handling librtmp, nghttp2, ngtcp2, nghttp3 and quiche. However, it seems that libcurl is not very consistent as other options don't raise such kind of failure.

bagder: It is not consistent as to whether that is an error or not, correct. Like in this case, curl has built fine with libpsl for decades without pkg-config, so maybe it could be worth keeping that ability just to make existing setups keep working?

ffontaine: PR was updated

bagder:
Thanks!

PR #12948, opened 2024-02-15
Remove curl_mimepart object from UserDefined structure when CURL_DISABLE_MIME flag is active

MAntoniak:
Reduce size of UserDefined structure. For the x86 architecture, the structure size decreased by 368 bytes, for x64 440 bytes. It is therefore a significant reduction in the size of the structure.

Also remove unreachable code: when CURL_DISABLE_MIME is set, httpreq can never have HTTPREQ_POST_MIME value and the same goes for the CURL_DISABLE_FORM_API flag and the HTTPREQ_POST_FORM value.

As for the name of the function I created in smtp.c and imap.c I am open to suggestions.

### DIFF
--- a/lib/imap.c
+++ b/lib/imap.c
@@ -770,6 +770,7 @@ static CURLcode imap_perform_append(struct Curl_easy *data)
     return CURLE_URL_MALFORMAT;
   }
 
+#ifndef CURL_DISABLE_MIME
   /* Prepare the mime data if some. */
   if(data->set.mimepost.kind != MIMEKIND_NONE) {
     /* Use the whole structure as data. */
@@ -798,6 +799,7 @@ static CURLcode imap_perform_append(struct Curl_easy *data)
     data->state.fread_func = (curl_read_callback) Curl_mime_read;
     data->state.in = (void *) &data->set.mimepost;
   }
+#endif
 
   /* Check we know the size of the upload */
   if(data->state.infilesize < 0) {
@@ -1513,10 +1515,10 @@ static CURLcode imap_done(struct Curl_easy *data, CURLcode status,
   }
   else if(!data->set.connect_only && !imap->custom &&
           (imap->uid || imap->mindex || data->state.upload ||
-          data->set.mimepost.kind != MIMEKIND_NONE)) {
+          IS_MIME_POST(data))) {
     /* Handle responses after FETCH or APPEND transfer has finished */
 
-    if(!data->state.upload && data->set.mimepost.kind == MIMEKIND_NONE)
+    if(!data->state.upload && !IS_MIME_POST(data))
       imap_state(data, IMAP_FETCH_FINAL);
     else {
       /* End the APPEND command first by sending an empty line */
@@ -1582,7 +1584,7 @@ static CURLcode imap_perform(struct Curl_easy *data, bool *connected,
     selected = TRUE;
 
   /* Start the first command in the DO phase */
-  if(data->state.upload || data->set.mimepost.kind != MIMEKIND_NONE)
+  if(data->state.upload || IS_MIME_POST(data))
     /* APPEND can be executed directly */
     result = imap_perform_append(data);
   else if(imap->custom && (selected || !imap->mailbox))
--- a/lib/http.c
+++ b/lib/http.c
@@ -2344,9 +2344,11 @@ CURLcode Curl_http_body(struct Curl_easy *data, struct connectdata *conn,
   http->postsize = 0;
 
   switch(httpreq) {
+#ifndef CURL_DISABLE_MIME
   case HTTPREQ_POST_MIME:
     data->state.mimepost = &data->set.mimepost;
     break;
+#endif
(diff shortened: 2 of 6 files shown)


lib/imap.c
```diff
@@ -1513,10 +1522,10 @@ static CURLcode imap_done(struct Curl_easy *data, CURLcode status,
   }
   else if(!data->set.connect_only && !imap->custom &&
           (imap->uid || imap->mindex || data->state.upload ||
-          data->set.mimepost.kind != MIMEKIND_NONE)) {
+            IMAP_MIMEKIND_IS_NOT_NONE(data))) {
```
jay, line 1525: couldn't this be a single shorter macro somewhere like urldata and make it different for when mime is disabled like 

~~~c
#ifndef CURL_DISABLE_MIME
#define IS_MIME_POST(a) ((a)->set.mimepost.kind != MIMEKIND_NONE)
#else
#define IS_MIME_POST(a) FALSE
#endif
~~~

MAntoniak: I've moved the macro definition to urldata.h.

jay:
Did you test this? Though it passes the CI (since we don't test CURL_DISABLE_MIME) there are warnings and it fails to compile:

~~~
..\..\..\..\lib\formdata.c(575): warning C4013: 'Curl_mime_contenttype' undefined; assuming extern returning int
..\..\..\..\lib\formdata.c(575): warning C4047: '=' : 'const char *' differs in levels of indirection from 'int'
..\..\..\..\lib\formdata.c(726): error C2064: term does not evaluate to a function taking 305 arguments
~~~

formdata.c(726):
~~~c
    size_t nread = Curl_mime_read(buffer, 1, sizeof(buffer), &toppart);
~~~

mime.h(169):
~~~c
#define Curl_mime_read NULL
~~~

It should be a function-like macro such as
~~~c
#define Curl_mime_read(a,b,c,d) NULL
~~~

but that causes different errors because of code that passes the function as a pointer even when CURL_DISABLE_MIME, which doesn't seem right. for example http.c(2582):
~~~c
    data->state.fread_func = (curl_read_callback) Curl_mime_read;
~~~

MAntoniak:
Yes, I saw the warning and error. They ocur, when user defines CURL_DISABLE_MIME flag without CURL_DISABLE_FORM_API. In the project I'm working on I have both of these flags applied and only tested this scenario. This problems occur also without this PR.

The problem with Curl_mime_contenttype is easy to solve. Simply write

`#define Curl_mime_contenttype NULL`

in mime.h in section where CURL_DISABLE_MIME is defined.

More complex task is solve the problem with Curl_mime_read. I don't quite know how to solve it. It is the public function. I didn't analyze what is the right behavour when CURL_DISABLE_MIME is defined.

MAntoniak:
If I understand code correctly the Curl_getformdata function return CURLE_OUT_OF_MEMORY and also curl_formget return the same code error.  Maybe the solution is:
```
--- a/src/lib/curl/lib/formdata.c
+++ b/src/lib/curl/lib/formdata.c
@@ -712,6 +712,7 @@ CURLFORMcode curl_formadd(struct curl_httppost **httppost,
 int curl_formget(struct curl_httppost *form, void *arg,
                  curl_formget_callback append)
 {
+#ifndef CURL_DEFINED_MIME
   CURLcode result;
   curl_mimepart toppart;
 
@@ -737,6 +738,9 @@ int curl_formget(struct curl_httppost *form, void *arg,
 
   Curl_mime_cleanpart(&toppart);
   return (int) result;
+#else
+  return CURLE_OUT_OF_MEMORY;
+#endif
 }
```

jay:
@monnerat

monnerat:
> They ocur, when user defines CURL_DISABLE_MIME flag without CURL_DISABLE_FORM_API

This case is a nonsense since the form api uses the mime api. There already is a guard against it in configure.ac and a test for it ought to be be added in CMakeLists.txt by someone who masters cmake. Introducing macros and/or conditionals for that situation only decreases the code readability.

If the ultimate goal of this PR is to spare room in the UserDefined structure, I've had a good result by simply shortening the curl_mimedata structure to an empty one:
```
-- a/lib/mime.h
+++ b/lib/mime.h
@@ -108,6 +108,9 @@ struct curl_mime {
 
 /* A mime part. */
 struct curl_mimepart {
+#ifdef CURL_DISABLE_MIME
+  char dummy[0];
+#else
   curl_mime *parent;               /* Parent mime structure. */
   curl_mimepart *nextpart;         /* Forward linked list. */
   enum mimekind kind;              /* The part kind. */
@@ -128,6 +131,7 @@ struct curl_mimepart {
   const struct mime_encoder *encoder; /* Content data encoder. */
   struct mime_encoder_state encstate; /* Data encoder state. */
   size_t lastreadstatus;           /* Last read callback returned status. */
+#endif
 };
 
 CURLcode Curl_mime_add_header(struct curl_slist **slp, const char *fmt, ...)
```

Some code bytes are saved by this PR though.

bagder:
Thanks!

PR #12949, opened 2024-02-16
move authneg bit

icing:
- from `conn->bits.authneg` to `data->req.authneg`
- AFAICT, this is a property of the request about to be made and not a property of the connection
- in multiuse connections, transfer could step on each others toes here potentially.

### DIFF
--- a/lib/c-hyper.c
+++ b/lib/c-hyper.c
@@ -710,7 +710,7 @@ static int uploadstreamed(void *userdata, hyper_context *ctx,
     return HYPER_POLL_PENDING;
   }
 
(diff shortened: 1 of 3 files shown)


jay:
hm. I feel like there was probably a good reason authneg was put on the connection level... maybe because of authentications that are connection level and not request level like NTLM and that is why?

icing:
AFAICT, it was only used to make the special tweaks for an empty PUT/POST when auth negotiation was still ongoing.

jay:
@iboukris

bagder (approved):
I believe this is a correct change.

Issue #12944, opened 2024-02-15
Function parsefmt cannot parse "I64d" when macro MP_HAVE_INT_EXTENSIONS is not defined (by default it is not defined)!


nased0:
### I did this

I compiled libcurl statically with OpenSSL 1.1.1w (linked dynamically) by using CMake and Embarcadero bcc32 v. 7.0 compiler.
To do this I had to make some minor corrections in the code that are not related with the issue:

In file curl_setup.h I defined:

```
struct _stati64
{  
   unsigned int	st_dev;
   short	st_ino; 	/* Always zero ? */			     
   unsigned short	st_mode;	/* See above constants */		     
   short 	st_nlink;	/* Number of links. */			     
   short 	st_uid; 	/* User: Maybe significant on NT ? */	     
   short 	st_gid; 	/* Group: Ditto */			     
   unsigned int	st_rdev;	/* Seems useless (not even filled in) */     
  __int64	st_size;	/* File size in bytes */		     
  time_t	st_atime;	/* Access time (always 00:00 on FAT) */	     
  time_t	st_mtime;	/* Modified time */			     
  time_t	st_ctime;	/* Creation time */			     
};

```
In file vtls.c I added

`#undef random
`
before function Curl_ssl_random.

in file openssl.c I added pragmas for linker:

```
#pragma comment(lib, "bcrypt.lib")
#pragma comment(lib, "Wldap32.lib")
#pragma comment(lib, "c:\Libs\common\libssle-1_1.lib")
#pragma comment(lib, "c:\Libs\common\libcryptoe-1_1.lib")

```
where openssl libs were converted to bcc32 OMF linker format by commands:
```
implib -a libcryptoe-1_1.lib libcrypto-1_1.dll 
implib -a libssle-1_1.lib libssl-1_1.dll 
```

In files objects1.rsp (generated by CMake) I replaced UNIX path separators / by Windows path separators \

### I expected the following

I expected my app to download a car report from historiapojazdu.gov.pl (polish Central Vehicle Register). But it crashed with code C0000005. 
I managed to find the cause: function parsefmt defined in file mprintf.c cannot parse the substring "%I64d" when macro MP_HAVE_INT_EXTENSIONS is not defined. And by default, it is not defined! 
So this function ignored "%I64d" and set the type FORMAT_STRING in va_input.type, then wrote value hex: 7FFFFFFF (cookie expiration date) to va_input.val.str by the code:
```
    switch(iptr->type) {
    case FORMAT_STRING:
[...]

PR #12950, opened 2024-02-16
mprintf: fix format prefix I32/I64 for all windows compilers

jay:
- Support I32 & I64 (eg: %I64d) for all Win32 builds.

Prior to this change mprintf support for the I format prefix, which is a Microsoft extension, was dependent on the compiler used.

When Borland support was removed in fd7ef00f the prefix was then no longer supported for that compiler; however since it's still possible to build with Borland I'm restoring support for the prefix in this way.

Reported-by: Paweł Witas

Fixes https://github.com/curl/curl/issues/12944
Closes #xxxx

### DIFF
--- a/lib/mprintf.c
+++ b/lib/mprintf.c
@@ -49,16 +49,6 @@
 #endif
 
 /*
- * Non-ANSI integer extensions
- */
-
-#if (defined(_WIN32_WCE)) || \
-    (defined(__MINGW32__)) || \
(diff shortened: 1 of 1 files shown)


bagder (approved):
Looks sensible!

bagder:
The commit message should probably say that this is for all windows compilers and not specifically highlight borland.

PR #12952, opened 2024-02-17
Fix typo on config file location

icp1994:
`curlrc` -> `.curlrc`

### DIFF
--- a/docs/cmdline-opts/config.md
+++ b/docs/cmdline-opts/config.md
@@ -64,7 +64,7 @@ config file is checked for in the following places in this order:
 
 1) **"$CURL_HOME/.curlrc"**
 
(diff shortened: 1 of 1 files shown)


dfandrich:
This fix looke like it matches the current behaviour, but I suspect the code is not quite right. I think it should be looking for the non-dot version of the file in the XDG_CONFIG_HOME location (which is how other programs do it), but there's explicit code in findfile.c that disables that check for some reason.

jay:
> This fix looke like it matches the current behaviour, but I suspect the code is not quite right.

See #12129. I proposed a fix in #12135 but I've softened on it. Maybe it is better to do it this way and document the way it has been working.

bagder:
I think we should fix the code rather than to document the bug.

PR #12959, opened 2024-02-19
docs: dist curl*.1 and install without perl

### LINKED ISSUE #12921: Man page installation and distribution

bagder:
Fixes #12921
Reported-by: Michael Forney

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -30,12 +30,7 @@ MK_CA_DOCS = mk-ca-bundle.1
 CURLCONF_DOCS = curl-config.1
 endif
 
-
-# EXTRA_DIST breaks with $(abs_builddir) so build it using this variable
-# but distribute it (using the relative file name) in the next variable
-man_MANS = $(abs_builddir)/curl.1
-noinst_man_MANS = $(MK_CA_DOCS)
-dist_man_MANS = $(CURLCONF_DOCS) $(MK_CA_DOCS)
+man_MANS = $(abs_builddir)/curl.1 curl-config.1
 CURLPAGES = curl-config.md mk-ca-bundle.md
 
 # Build targets in this file (.) before cmdline-opts to ensure that
@@ -43,11 +38,13 @@ CURLPAGES = curl-config.md mk-ca-bundle.md
 SUBDIRS = . cmdline-opts libcurl
 DIST_SUBDIRS = $(SUBDIRS) examples
 
-CLEANFILES = $(man_MANS) curl.1 $(CURLCONF_DOCS) $(MK_CA_DOCS)
-nodist_MANS = $(CLEANFILES)
+if BUILD_DOCS
+CLEANFILES = curl.1 mk-ca-bundle.1 curl-config.1
+endif
 
 EXTRA_DIST =                                    \
  $(CURLPAGES)                                   \
+ $(CURLCONF_DOCS)                               \
  ALTSVC.md                                      \
  BINDINGS.md                                    \
  BUFREF.md                                      \
@@ -112,6 +109,8 @@ CD2_ = $(CD2_0)
 
 SUFFIXES = .1 .md
 
+all: $(MK_CA_DOCS)
+
 # $(abs_builddir) is to disable VPATH when searching for this file, which
 # would otherwise find the copy in $(srcdir) which breaks the $(HUGE)
 # rule in src/Makefile.am in out-of-tree builds that references the file in the
@@ -130,6 +129,8 @@ $(abs_builddir)/curl.1:
 	$(CD2)$(CD2NROFF)
 
 curl-config.1: curl-config.md
+
+mk-ca-bundle.1: mk-ca-bundle.md
 
 distclean:
 	rm -f $(CLEANFILES)
--- a/Makefile.am
+++ b/Makefile.am
@@ -249,6 +249,7 @@ install-data-hook:
 else
 install-data-hook:
 	(cd include && $(MAKE) install)
+	(cd docs && $(MAKE) install)
 endif
 
 # We extend the standard uninstall with a custom hook:
--- a/docs/libcurl/Makefile.am
+++ b/docs/libcurl/Makefile.am
@@ -36,7 +36,6 @@ m4macrodir = $(datadir)/aclocal
 dist_m4macro_DATA = libcurl.m4
 
 CLEANFILES = $(man_MANS) libcurl-symbols.md
-nodist_MANS = $(man_MANS)
 
 EXTRA_DIST = $(CURLPAGES) ABI.md symbols-in-versions symbols.pl  \
   mksymbolsmanpage.pl CMakeLists.txt
--- a/docs/libcurl/opts/Makefile.am
+++ b/docs/libcurl/opts/Makefile.am
@@ -30,7 +30,6 @@ include Makefile.inc
 CURLPAGES = $(man_MANS:.3=.md)
 endif
 CLEANFILES = $(man_MANS)
-nodist_MANS = $(man_MANS)
 
 EXTRA_DIST = $(CURLPAGES) CMakeLists.txt
 


bagder:
@michaelforney would prefer I merge this and you can verify using the following daily snapshots?

docs/Makefile.am
```diff
@@ -33,21 +33,20 @@ endif
 
 # EXTRA_DIST breaks with $(abs_builddir) so build it using this variable
 # but distribute it (using the relative file name) in the next variable
-man_MANS = $(abs_builddir)/curl.1
-noinst_man_MANS = $(MK_CA_DOCS)
-dist_man_MANS = $(CURLCONF_DOCS) $(MK_CA_DOCS)
+man_MANS = $(abs_builddir)/curl.1 curl-config.1
+noinst_MANS = mk-ca-bundle.1
```
michaelforney, line 37: This variable doesn't do anything (you can look at the generated Makefile.in to confirm). The only thing I could find about it in automake source is this comment:

```
## We don't really need this, but we use it in case we ever want to
## support noinst_MANS.
MANS = %MANS%
```

docs/Makefile.am
```diff
@@ -33,21 +33,20 @@ endif
 
 # EXTRA_DIST breaks with $(abs_builddir) so build it using this variable
 # but distribute it (using the relative file name) in the next variable
```
michaelforney, line 35: I think this comment should be updated, since curl.1 is now distributed via dist-hook.

docs/Makefile.am
```diff
@@ -112,6 +111,8 @@ CD2_ = $(CD2_0)
 
 SUFFIXES = .1 .md
 
+all: $(man_MANS) $(MK_CA_DOCS)
```
michaelforney, line 114: `all: $(man_MANS)` is unnecessary; automake does this automatically.

docs/Makefile.am
```diff
@@ -33,21 +33,20 @@ endif
...
 SUBDIRS = . cmdline-opts libcurl
 DIST_SUBDIRS = $(SUBDIRS) examples
 
-CLEANFILES = $(man_MANS) curl.1 $(CURLCONF_DOCS) $(MK_CA_DOCS)
-nodist_MANS = $(CLEANFILES)
+CLEANFILES = curl.1
 
 EXTRA_DIST =                                    \
  $(CURLPAGES)                                   \
+ $(CURLCONF_DOCS)                               \
```
michaelforney, line 49: Should mk-ca-bundle.1 be distributed as well, or no?

michaelforney (commented):
Thanks for looking at this. I confirmed that it fixes installation of man pages without perl, which was my main concern.

I commented about a few things that I think could still be cleaned up, and I'm not sure if you intend for mk-ca-bundle.1 to be distributed or not (doesn't matter to me either way).

Also, there are still uses of `nodist_MANS` in `docs/libcurl/Makefile.am` and `docs/libcurl/opts/Makefile.am`, which should be removed.

docs/Makefile.am
bagder: But it does not depend on/build `$(MK_CA_DOCS)` automatically without this

docs/Makefile.am
bagder: Should it? It's not installed, and somewhere I think we should draw the line. You can easily build it....

docs/Makefile.am
michaelforney: Correct. `all: $(MK_CA_DOCS)` is still needed.

docs/Makefile.am
michaelforney: I wasn't sure, that's why I asked. It is in current master, so maybe the commit message should mention that it drops mk-ca-bundle.1 from dist.

docs/Makefile.am
bagder: :+1:

docs/Makefile.am
bagder: :+1:

michaelforney:
Thanks! This looks good to me.

PR #12968, opened 2024-02-20
Time handling:

sounar97:
If time_t is not sufficient for storing dates beyond the year 2038, considered using a larger data type like curl_off_t or a dedicated time library in hstc.c file.

### DIFF
--- a/docs/BUFQ.md
+++ b/docs/BUFQ.md
@@ -144,5 +144,5 @@ The pool gets the size and the mount of spares to keep. The `bufq` gets the pool
 
 A pool can be shared between many `bufq`s, as long as all of them operate in the same thread. In curl that would be true for all transfers using the same multi handle. The advantages of a pool are:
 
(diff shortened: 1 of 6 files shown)


rsbeckerca:
> If time_t is not sufficient for storing dates beyond the year 2038, considered using a larger data type like curl_off_t or a dedicated time library in hstc.c file.

I can confirm that time_t is not sufficient for any 32-bit OS with a signed time_t. Those do exist at least currently in production. Some platforms use time64_t (S/390) instead.

bagder:
This PR looks completely broken and has lots of unrelated changes.

If `time_t` is 32 bits, then the system is completely lost very soon anyway (long before 2038 since date math will start to break as well) and curl changing some few types will not save it. It is not problem for curl to fix. The system **MUST** use a 64 bit time_t or provide some other system mechanisms to work around the problem.

PR #12969, opened 2024-02-21
Send Rework, Part 2

### LINKED ISSUE #12938: curl does not send trailers for small requests
### I did this

Trailers are sent only for "big" requests (approx. >= 64 KB). For "small" requests, trailers are not sent, and the function set by `CURLOPT_TRAILERFUNCTION` is not called.

**How to reproduce:**

For testing, I have added trailer support to the curl tool: [curl-trailers.patch.txt](https://github.com/curl/curl/files/14281359/curl-trailers.patch.txt) - build curl with this patch.

Create test data with:
```
( for N in $(seq 1000); do printf "param%d=%d&" $N $N; done ) > numbers-small.txt
( for N in $(seq 5000); do printf "param%d=%d&" $N $N; done ) > numbers.txt
```

Start netcat:
```
netcat -l localhost 8000
```

With this request, trailers are not sent:
```
curl -v -s --data @numbers-small.txt --header 'Transfer-Encoding: chunked' --header 'Trailer: MyTrailer' --trailer 'MyTrailer: My Trailer Value' http://localhost:8000/
```
Netcat output:
```
[...]
aram996=996&param997=997&param998=998&param999=999&param1000=1000&
0

```

When using the bigger "numbers.txt", trailers are sent:
```
curl -v -s --data @numbers.txt --header 'Transfer-Encoding: chunked' --header 'Trailer: MyTrailer' --trailer 'MyTrailer: My Trailer Value' http://localhost:8000/
```
Netcat output:
```
[...]
6&param4997=4997&param4998=4998&param4999=4999&param5000=5000&
0
MyTrailer: My Trailer Value
```

### I expected the following

Trailers are sent also for small requests.

### curl/libcurl version

curl master

### operating system

Fedora Linux 39

icing:
This is the second part of reworking curl's "send" or "upload" handling. The changes will arrive in a series of PRs, based on top of each other, similar to the client writer ones. For part 0 see #12963, part 1 is #12964.

### Client Reader

Added as documented [in CLIENT-READER.md](https://github.com/curl/curl/blob/5b1f31dfbab8aef467c419c68aa06dc738cb75d4/docs/CLIENT-READERS.md).

- old `Curl_buffer_send()` completely replaced by new `Curl_req_send()`
- old `Curl_fillreadbuffer()` replaced with `Curl_client_read()`
- HTTP chunked uploads are now formatted in a client reader added when needed.
- FTP line-end conversions are done in a client reader added when needed.
- when sending requests headers, remaining buffer space is filled with body data for sending in "one go". This is independent of the request body size. Resolves  #12938 as now small and large requests have the same code path.

Changes done to test cases:
- test513: now fails before sending request headers as this initial "client read" triggers the setup fault. Behaves now the same as in hyper build
- test547, test555, test1620: fix the length check in the lib code to only fail for reads *smaller* than expected. This was a bug in the test code that never triggered in the old implementation.

### DIFF
--- a/docs/CLIENT-READERS.md
+++ b/docs/CLIENT-READERS.md
@@ -19,7 +19,7 @@ CURLcode Curl_client_read(struct Curl_easy *data, char *buf, size_t blen,
 ```
 The arguments are the transfer to read for, a buffer to hold the read data, its length, the actual number of bytes placed into the buffer and the `eos` flag indicating that no more data will be available. The `eos` flag may be set for a read amount, if that amount was the last. That way curl can avoid to read an additional time.
 
-The implementation of `Curl_client_read()` uses a chain of *client reader* instances to get the data. This is very similar to the design of *client writers*. The chain of readers allows processing of the data to send.
+The implementation of `Curl_client_read()` uses a chain of *client reader* instances to get the data. This is similar to the design of *client writers*. The chain of readers allows processing of the data to send.
 
 The definition of a reader is:
 
--- a/docs/CLIENT-READERS.md
+++ b/docs/CLIENT-READERS.md
@@ -11,13 +11,13 @@ With this naming established, client readers are concerned with providing data f
 
 ## Invoking
 
-The transfer loop that sends and receives, will use `Curl_client_read()` to get more data to send for a transfer. If no specific reader has been installed yet, the default one that uses `CURLOPT_READFUNCTION`, will be added. The prototype is
+The transfer loop that sends and receives, is using `Curl_client_read()` to get more data to send for a transfer. If no specific reader has been installed yet, the default one that uses `CURLOPT_READFUNCTION` is added. The prototype is
 
 ```
 CURLcode Curl_client_read(struct Curl_easy *data, char *buf, size_t blen,
                           size_t *nread, bool *eos);
 ```
-The arguments are the transfer to read for, a buffer to hold the read data, its length, the actual number of bytes placed into the buffer and the `eos` flag indicating that no more data will be available. The `eos` flag may be set for a read amount, if that amount was the last. That way curl can avoid to read an additional time.
+The arguments are the transfer to read for, a buffer to hold the read data, its length, the actual number of bytes placed into the buffer and the `eos` flag indicating that no more data is available. The `eos` flag may be set for a read amount, if that amount was the last. That way curl can avoid to read an additional time.
 
 The implementation of `Curl_client_read()` uses a chain of *client reader* instances to get the data. This is similar to the design of *client writers*. The chain of readers allows processing of the data to send.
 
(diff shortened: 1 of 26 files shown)


jay:
Please add documentation after all your rework changes are settled, assuming they are accepted, so that contributors can know how the system operates.

icing:
> Please add documentation after all your rework changes are settled, assuming they are accepted, so that contributors can know how the system operates.

I'd be happy to do that, once all the piece have been worked out.

docs/CLIENT-READERS.md
```diff
@@ -0,0 +1,93 @@
+# curl client readers
+
+Client readers is a design in the internals of libcurl, not visible in its public API. They were started
+in curl v8.7.0. This document describes the concepts, its high level implementation and the motivations.
+
+## Naming
+
+`libcurl` operates between clients and servers. A *client* is the application using libcurl, like the command line tool `curl` itself. Data to be uploaded to a server is **read** from the client and **send** to the server, the servers response is **received** by `libcurl` and then **written** to the client.
```
bagder, line 8:
```suggestion
`libcurl` operates between clients and servers. A *client* is the application using libcurl, like the command line tool `curl` itself. Data to be uploaded to a server is **read** from the client and **sent** to the server, the servers response is **received** by `libcurl` and then **written** to the client.
```

docs/CLIENT-READERS.md
```diff
@@ -0,0 +1,93 @@
...
+
+With this naming established, client readers are concerned with providing data from the application to the server. Applications register callbacks via `CURLOPT_READFUNCTION`, data via `CURLOPT_POSTFIELDS` and other options to be used by `libcurl` when the request is send.
```
bagder, line 10:
```suggestion
With this naming established, client readers are concerned with providing data from the application to the server. Applications register callbacks via `CURLOPT_READFUNCTION`, data via `CURLOPT_POSTFIELDS` and other options to be used by `libcurl` when the request is sent.
```

docs/CLIENT-READERS.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Invoking
+
+The transfer loop that sends and receives, is using `Curl_client_read()` to get more data to send for a transfer. If no specific reader has been installed yet, the default one that uses `CURLOPT_READFUNCTION` is added. The prototype is
+
+```
+CURLcode Curl_client_read(struct Curl_easy *data, char *buf, size_t blen,
+                          size_t *nread, bool *eos);
+```
+The arguments are the transfer to read for, a buffer to hold the read data, its length, the actual number of bytes placed into the buffer and the `eos` flag indicating that no more data is available. The `eos` flag may be set for a read amount, if that amount was the last. That way curl can avoid to read an additional time.
```
bagder, line 20:
```suggestion
The arguments are the transfer to read for, a buffer to hold the read data, its length, the actual number of bytes placed into the buffer and the `eos` (*end of stream*) flag indicating that no more data is available. The `eos` flag may be set for a read amount, if that amount was the last. That way curl can avoid to read an additional time.
```

docs/CLIENT-READERS.md
```diff
@@ -0,0 +1,93 @@
...
+};
+
+struct Curl_creader {
+  const struct Curl_crtype *crt;  /* type implementation */
+  struct Curl_creader *next;  /* Downstream reader. */
+  Curl_creader_phase phase; /* phase at which it operates */
+};
+```
+
+`Curl_creader` is a reader instance with a `next`pointer to form the chain. It as a type `crt`which provides the implementation. The main callback is `do_read()` which provides the data to the caller. The others are for setup and tear down. `needs_rewind()` is explained further below.
```
bagder, line 43:
```suggestion
`Curl_creader` is a reader instance with a `next` pointer to form the chain. It as a type `crt` which provides the implementation. The main callback is `do_read()` which provides the data to the caller. The others are for setup and tear down. `needs_rewind()` is explained further below.
```

lib/bufq.c
```diff
@@ -417,6 +417,17 @@ ssize_t Curl_bufq_write(struct bufq *q,
   return nwritten;
 }
 
+CURLcode Curl_bufq_cwrite(struct bufq *q,
+                         const char *buf, size_t len,
+                         size_t *pnwritten)
```
bagder, line 422:
```suggestion
CURLcode Curl_bufq_cwrite(struct bufq *q,
                          const char *buf, size_t len,
                          size_t *pnwritten)
```

bagder (approved):
Just some nits/suggestions

docs/CLIENT-READERS.md
icing: fixed

docs/CLIENT-READERS.md
icing: Changed.

docs/CLIENT-READERS.md
icing: Fixed.

lib/bufq.c
icing: Fixed.

docs/CLIENT-READERS.md
icing: Added.

Issue #12971, opened 2024-02-22
HTTP2 requests through multi interface might not properly finish


5533asdg:
### I did this

The issue is observed after upgrading curl from 8.5.0 to 8.6.0.
Specifically, the behavior change is seen from https://github.com/curl/curl/commit/d7b6ce64ce0ad787ad2ed3ee05c94938a6b4f551.

When issuing queries, we might not get the state change from PERFORMING to DONE.

The currently seen behavior on failure is*:
1. HTTPS/2 request is sent through `curl_multi_add_handle(...)`.
2. The request is completed and `drain_steam(...)` (https://github.com/curl/curl/blob/master/lib/http2.c#L212) is called.
    - `data->state.select_bits` is set to `CURL_CSELECT_IN`.
    - `Curl_expire(data, 0, EXPIRE_RUN_NOW);` is called.
3. Client's `epoll` returns the socket to be writable but not readable.
4. `curl_multi_socket_action(..., CURL_CSELECT_OUT, ...)` is called by the client.
    - `data->state.select_bits` is set to `CURL_CSELECT_OUT` (overriding libcurl operation on step 2).
5. `CURLMOPT_TIMERFUNCTION` is called as a result of step 2.
6. `readwrite_data(...)` is not called because `data->state.select_bits == CURL_CSELECT_OUT` and `k->keepon == KEEP_RECV`.

The currently seen behavior on success is*:
1. HTTPS/2 request is sent through `curl_multi_add_handle(...)`.
2. The request is completed and `drain_steam(...)` (https://github.com/curl/curl/blob/master/lib/http2.c#L212) is called.
    - `data->state.select_bits` is set to `CURL_CSELECT_IN`.
    - `Curl_expire(data, 0, EXPIRE_RUN_NOW);` is called.
3. `CURLMOPT_TIMERFUNCTION` is called as a result of step 2.
4. `readwrite_data(...)` is called because `data->state.select_bits == CURL_CSELECT_IN` and `k->keepon == KEEP_RECV`.
    - Curl gets `EndOfSteam` (EOS) / `nread == 0` from  `readwrite_data(...)`.
5. Curl updated the state from PERFORMING to DONE.

*The difference is whether `curl_multi_socket_action(..., CURL_CSELECT_OUT, ...)` or `CURLMOPT_TIMERFUNCTION` is called first.

The previous behavior is (< 8.6.0):
1. HTTPS/2 request is sent through `curl_multi_add_handle(...)`.
[...]

PR #12972, opened 2024-02-22
fix multi_sock handling of select_bits

icing:
- refs #12971
- add the event bitmask to data->state.select_bits instead of overwriting them. They are cleared again on use.


bagder:
Thanks!

PR #12973, opened 2024-02-22
fopen: fix narrowing conversion warning on 32-bit Android

MarcelRaad:
This was fixed in commit 06dc599405f, but came back in commit 03cb1ff4d62.

When building for 32-bit ARM or x86 Android, `st_mode` is defined as `unsigned int` instead of `mode_t`, resulting in a -Wimplicit-int-conversion clang warning because `mode_t` is `unsigned short`. Add a cast to silence the warning.

Ref: https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r25c/libc/include/sys/stat.h#86

### DIFF
--- a/lib/fopen.c
+++ b/lib/fopen.c
@@ -129,7 +129,7 @@ CURLcode Curl_fopen(struct Curl_easy *data, const char *filename,
   }
 
   result = CURLE_WRITE_ERROR;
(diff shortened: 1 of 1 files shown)


MarcelRaad:
Wait, this doesn't work for MSVC...

MarcelRaad:
Closing in favor of https://github.com/curl/curl/pull/12998.

PR #12976, opened 2024-02-22
Added curl_multi_waitfds.

dkarpov1970:
This function extracts curl_waitfd structures which are similar to poll(2)'s pollfd structure from a given multi_handle.

These structures can be used for polling on multi_handle file descriptors in a fashion similar to curl_multi_poll(3). The curl_multi_perform(3) function should be called as soon as one of them is ready to be read from or written to.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -3243,7 +3243,7 @@ static CURLMcode multi_socket(struct Curl_multi *multi,
 
         if(data->conn && !(data->conn->handler->flags & PROTOPT_DIRLOCK))
           /* set socket event bitmask if they're not locked */
-          data->state.select_bits = (unsigned char)ev_bitmask;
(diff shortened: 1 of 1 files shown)


bagder:
This looks excellent. Now I think we only need a test case or three that verify the functionality.

dkarpov1970:
I am not very familiar with how to write test cases for libcurl, so it will probably take a while for me to add the tests.
Can we do it in a separate PR to avoid things like rebasing/merging conflicts etc?

bagder:
> not sure why daniel wants multiple tests

We want the function tested. If we can do that sufficiently in a single test, then that is fine. If it needs more than one test, then that's good.

dkarpov1970:
I will create a new PR when I have tests ready.

PR #12980, opened 2024-02-24
examples/websocket.c: check for CURLE_AGAIN on recv_pong()

Theldus:
Description
-----------
The example file [docs/examples/websocket.c] does not work out of the box, because the recv_pong() function does not consider that the `curl_ws_recv()` function might return CURLE_AGAIN if there is no data to read, causing the program to terminate prematurely.

This commit addresses this by using `select(2)` to make receiving pong blocking, and thus being able to continue with the main loop.

This also occurs in the example file [sendrecv.c], which also makes use of select(2) for sending and receiving data, which I based on for this commit.

[docs/examples/websocket.c]: https://github.com/curl/curl/blob/master/docs/examples/websocket.c
[sendrecv.c]: https://github.com/curl/curl/blob/master/docs/examples/sendrecv.c

### DIFF
--- a/docs/examples/websocket.c
+++ b/docs/examples/websocket.c
@@ -36,6 +36,13 @@
  */
 #ifdef __GNUC__
 #pragma GCC diagnostic ignored "-Wsign-conversion"
+#ifdef __GNUC__
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+#ifdef __DJGPP__
+#pragma GCC diagnostic ignored "-Warith-conversion"
+#endif
+#elif defined(_MSC_VER)
+#pragma warning(disable:4127)  /* conditional expression is constant */
 #endif
 
 /* Auxiliary function that waits on the socket. */


docs/examples/websocket.c
```diff
@@ -30,6 +30,39 @@
...
+ * warning: conversion to 'long unsigned int' from 'curl_socket_t' {aka 'int'}
+ * may change the sign of the result [-Wsign-conversion]
+ */
+#if defined(__GNUC__) && defined(__CYGWIN__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+#endif
+  FD_SET(sockfd, &errfd); /* always check for error */
+  FD_SET(sockfd, &infd);
+#if defined(__GNUC__) && defined(__CYGWIN__)
```
vszakats, line 57:
```suggestion
#if defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wsign-conversion"
#endif
  FD_SET(sockfd, &errfd); /* always check for error */
  FD_SET(sockfd, &infd);
#if defined(__GNUC__)
```
This hack is also necessary on some non-Cygwin systems. (OmniOS for example)

Theldus: @vszakats I've forced pushed my changes.
I was just following the same pattern found in [docs/examples/sendrecv.c](https://github.com/curl/curl/blob/master/docs/examples/sendrecv.c), which now was changed too.

vszakats: Thank you @Theldus.

docs/examples/websocket.c
```diff
@@ -30,6 +30,39 @@
...
+ */
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+#endif
+  FD_SET(sockfd, &errfd); /* always check for error */
+  FD_SET(sockfd, &infd);
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
```
vszakats, line 59:
```suggestion
/* Avoid warning in FD_SET() with pre-2020 Cygwin/MSYS releases:
 * warning: conversion to 'long unsigned int' from 'curl_socket_t' {aka 'int'}
 * may change the sign of the result [-Wsign-conversion]
 */
#ifdef __GNUC__
#pragma GCC diagnostic ignored "-Wsign-conversion"
#ifdef __DJGPP__
#pragma GCC diagnostic ignored "-Warith-conversion"
#endif
#elif defined(_MSC_VER)
#pragma warning(disable:4127)  /* conditional expression is constant */
#endif

/* Auxiliary function that waits on the socket. */
static int wait_on_socket(curl_socket_t sockfd, long timeout_ms)
{
  struct timeval tv;
  fd_set infd, outfd, errfd;
  int res;

  tv.tv_sec = timeout_ms / 1000;
  tv.tv_usec = (int)(timeout_ms % 1000) * 1000;

  FD_ZERO(&infd);
  FD_ZERO(&outfd);
  FD_ZERO(&errfd);

  FD_SET(sockfd, &errfd); /* always check for error */
  FD_SET(sockfd, &infd);
```
To sync up with 7e814c8717939393d4436d75f5f0c3ffa98c8c53 for GCC 4.x compatibility.

vszakats:
websocket example's ping received retries and waits in fa0ed1fe4f7de6519dfc5519728e3710bfe077fc #17860,
implementing what seems to be the goal of this PR.

Closing this in favor of the above.

In any case if you'd like continue this effort, please rebase it, or create a new PR from the fresh source.

---

diff without whitespace changes: https://github.com/curl/curl/pull/12980/files?w=1

PR #12990, opened 2024-02-25
strtoofft: fix the overflow check

bagder:
... to not rely on wrapping, since it is an undefined behavior that is not what always might happen.

Reported-by: vulnerabilityspotter on hackerone

### DIFF
--- a/lib/strtoofft.c
+++ b/lib/strtoofft.c
@@ -79,11 +79,10 @@ static int get_char(char c, int base);
 static curl_off_t strtooff(const char *nptr, char **endptr, int base)
 {
   char *end;
-  int is_negative = 0;
-  int overflow;
+  bool is_negative = FALSE;
+  bool overflow = FALSE;
   int i;
   curl_off_t value = 0;
-  curl_off_t newval;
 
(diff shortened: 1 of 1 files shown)


lib/strtoofft.c
```diff
@@ -129,19 +128,15 @@ static curl_off_t strtooff(const char *nptr, char **endptr, int base)
...
-  overflow = 0;
   for(i = get_char(end[0], base);
       i != -1;
       end++, i = get_char(end[0], base)) {
-    newval = base * value + i;
-    if(newval < value) {
-      /* We've overflowed. */
-      overflow = 1;
+
+    if(value > CURL_OFF_T_MAX / base - i) {
```
vulnerabilityspotter, line 135: I believe the correct test is the following otherwise you compare ``(value + i) * base > CURL_OFF_T_MAX`` instead of ``value * base + i > CURL_OFF_T_MAX``  :
```suggestion
    if(value > (CURL_OFF_T_MAX - i) / base) {
```

bagder: Ah, yes thanks.

I also suspect we have no CI build that actually use this code...

PR #12992, opened 2024-02-26
multi: make add_handle free any multi_easy

bagder:
If the easy handle that is being added to a multi handle has previously been used for curl_easy_perform(), there is a private multi handle here that we can kill off. While it flushes some caches etc for the easy handle would it be used for an easy interface transfer again after being used in the multi stack, this cleanup simplifies behavior and uses less memory.


bagder:
any thoughts on this @jay ?

PR #12998, opened 2024-02-26
fopen: fix narrowing conversion warning on 32-bit Android

kiefer-andreas:
This was fixed in commit 06dc599405fd8a87147eddcdb89dab62ce200d9b, but came back in commit 03cb1ff4d629d6110dab787c75c187626d58323d.

When building for 32-bit ARM or x86 Android, `st_mode` is defined as `unsigned int` instead of `mode_t`, resulting in a `-Wimplicit-int-conversion` clang warning because `mode_t` is `unsigned short`. Add a cast to silence the warning, but only for Android builds, because other architectures and platforms are not affected.

Ref: https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r25c/libc/include/sys/stat.h#86

### DIFF
--- a/lib/fopen.c
+++ b/lib/fopen.c
@@ -129,7 +129,12 @@ CURLcode Curl_fopen(struct Curl_easy *data, const char *filename,
   }
 
   result = CURLE_WRITE_ERROR;
+#if (defined(ANDROID) || defined(__ANDROID__)) && \
+    (defined(__i386__) || defined(__arm__))
+  fd = open(tempstore, O_WRONLY | O_CREAT | O_EXCL, (mode_t)(0600|sb.st_mode));
+#else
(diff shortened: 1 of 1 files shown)


bagder:
The description says "on 32-bit Android" but the code is clearly done unconditionally for all Android builds. I presume that is what is intended even though there is no mention of non-32 bit?

kiefer-andreas:
Ah, yes. Sorry. For 64-bit `st_mode` is defined as `mode_t`, so the cast is not necessary. Will update the commit.

bagder (approved):

MarcelRaad (approved):

MarcelRaad:
Thanks Andreas!

PR #13000, opened 2024-02-26
asyn-ares: remove a data race

RMMoreton:
If two easy curl handles are initialized from different threads at the same time, they can race to check/set the ares_ver variable. I think the consequences are minimal, but it's causing test failures when run under sanitizers.

I have not been able to write a short (and shareable) reproduction of this, but it's happening daily. I can keep trying to write a reproduction if that's valuable.

### DIFF
--- a/lib/asyn-ares.c
+++ b/lib/asyn-ares.c
@@ -122,6 +122,8 @@ struct thread_data {
 
 #define CARES_TIMEOUT_PER_ATTEMPT 2000
 
+static int ares_ver = 0;
+
 /*
  * Curl_resolver_global_init() - the generic low-level asynchronous name
  * resolve API.  Called from curl_global_init() to initialize global resolver
@@ -134,6 +136,7 @@ int Curl_resolver_global_init(void)
     return CURLE_FAILED_INIT;
   }
 #endif
+  ares_version(&ares_ver);
   return CURLE_OK;
 }
 
@@ -173,16 +176,8 @@ CURLcode Curl_resolver_init(struct Curl_easy *easy, void **resolver)
   int status;
   struct ares_options options;
   int optmask = ARES_OPT_SOCK_STATE_CB;
-  static int ares_ver = 0;
   options.sock_state_cb = sock_state_cb;
   options.sock_state_cb_data = easy;
-  if(ares_ver == 0)
-    ares_version(&ares_ver);
-
-  if(ares_ver < 0x011400) { /* c-ares included similar change since 1.20.0 */
-    options.timeout = CARES_TIMEOUT_PER_ATTEMPT;
-    optmask |= ARES_OPT_TIMEOUTMS;
-  }
 
   /*
      if c ares < 1.20.0: curl set timeout to CARES_TIMEOUT_PER_ATTEMPT (2s)
@@ -193,6 +188,11 @@ CURLcode Curl_resolver_init(struct Curl_easy *easy, void **resolver)
      if c-ares >= 1.24.0, user can set the timeout via /etc/resolv.conf to
      overwrite c-ares' timeout.
   */
+  DEBUGASSERT(ares_ver);
+  if(ares_ver < 0x011400) {
+    options.timeout = CARES_TIMEOUT_PER_ATTEMPT;
+    optmask |= ARES_OPT_TIMEOUTMS;
+  }
 
   status = ares_init_options((ares_channel*)resolver, &options, optmask);
   if(status != ARES_SUCCESS) {


bagder:
The downside with this is that you moved a run-time version check into a build-time one.

bagder:
What do you think of this as an alternative take?
~~~diff
diff --git a/lib/asyn-ares.c b/lib/asyn-ares.c
index 76efba78ad..4cbe2b95b2 100644
--- a/lib/asyn-ares.c
+++ b/lib/asyn-ares.c
@@ -57,10 +57,11 @@
 #include "inet_pton.h"
 #include "connect.h"
 #include "select.h"
 #include "progress.h"
 #include "timediff.h"
+#include "easy_lock.h"
 
 #if defined(CURL_STATICLIB) && !defined(CARES_STATICLIB) &&   \
   defined(_WIN32)
 #  define CARES_STATICLIB
 #endif
@@ -159,10 +160,23 @@ static void sock_state_cb(void *data, ares_socket_t socket_fd,
     DEBUGASSERT(easy);
     Curl_multi_closed(easy, socket_fd);
   }
 }
 
+#ifdef GLOBAL_INIT_IS_THREADSAFE
+
+static curl_simple_lock ares_lock = CURL_SIMPLE_LOCK_INIT;
+#define arescheck_lock() curl_simple_lock_lock(&ares_lock)
+#define arescheck_unlock() curl_simple_lock_unlock(&ares_lock)
+
+#else
+
+#define arescheck_lock()
+#define arescheck_unlock()
+
+#endif
+
 /*
  * Curl_resolver_init()
  *
  * Called from curl_easy_init() -> Curl_open() to initialize resolver
  * URL-state specific environment ('resolver' member of the UrlState
@@ -174,13 +188,16 @@ CURLcode Curl_resolver_init(struct Curl_easy *easy, void **resolver)
   struct ares_options options;
   int optmask = ARES_OPT_SOCK_STATE_CB;
   static int ares_ver = 0;
   options.sock_state_cb = sock_state_cb;
   options.sock_state_cb_data = easy;
-  if(ares_ver == 0)
-    ares_version(&ares_ver);
 
+  if(ares_ver == 0) {
+    arescheck_lock();
+    ares_version(&ares_ver);
+    arescheck_unlock();
+  }
   if(ares_ver < 0x011400) { /* c-ares included similar change since 1.20.0 */
     options.timeout = CARES_TIMEOUT_PER_ATTEMPT;
     optmask |= ARES_OPT_TIMEOUTMS;
   }

RMMoreton:
Oh, because of the other build-time version checks in the file I thought that a build-time check would be preferred. Why is a run-time check preferred in this case?

The race I'm concerned about still exists in your diff:

1. Thread 1 acquires the lock
2. Thread 1 enters `ares_version()`
3. Thread 2 loads `ares_ver` to use in the `if` statement
4. Thread 1 sets `ares_ver`
5. Thread 2 uses its outdated loaded value to check `if(ares_ver == 0)`

If we put the lock before the `if` then the race is gone:

```diff
+  arescheck_lock();
+  if(ares_ver == 0) {
+    ares_version(&ares_ver);
+  }
+  arescheck_unlock();
```

but I don't know if it's a good idea to serialize every thread here.

What do you think of using a file-level static variable, and using `Curl_resolver_global_init()` to initialize it? Something like:

```diff
diff --git a/lib/asyn-ares.c b/lib/asyn-ares.c
index 76efba78a..e15d12afa 100644
--- a/lib/asyn-ares.c
+++ b/lib/asyn-ares.c
@@ -122,6 +122,8 @@ struct thread_data {
 
 #define CARES_TIMEOUT_PER_ATTEMPT 2000
 
+static int ares_ver = 0;
+
 /*
  * Curl_resolver_global_init() - the generic low-level asynchronous name
  * resolve API.  Called from curl_global_init() to initialize global resolver
@@ -134,6 +136,7 @@ int Curl_resolver_global_init(void)
     return CURLE_FAILED_INIT;
   }
 #endif
+  ares_version(&ares_ver);
   return CURLE_OK;
 }
 
@@ -173,11 +176,8 @@ CURLcode Curl_resolver_init(struct Curl_easy *easy, void **resolver)
   int status;
   struct ares_options options;
   int optmask = ARES_OPT_SOCK_STATE_CB;
-  static int ares_ver = 0;
   options.sock_state_cb = sock_state_cb;
   options.sock_state_cb_data = easy;
-  if(ares_ver == 0)
-    ares_version(&ares_ver);
 
   if(ares_ver < 0x011400) { /* c-ares included similar change since 1.20.0 */
     options.timeout = CARES_TIMEOUT_PER_ATTEMPT;
```

jay:
> What do you think of using a file-level static variable, and using `Curl_resolver_global_init()` to initialize it?

That's fine. I updated your branch to do it that way.

bagder (approved):

bagder:
Thanks!

PR #13004, opened 2024-02-27
setopt: Fix disabling all protocols

danielgustafsson:
When disabling all protocols without enabling any, the resulting set of allowed protocols remained the default set.  Clearing the allowed set before inspecting the passed value from --proto make the set empty even in the errorpath of no protocols enabled.

Reported-by: Dan Fandrich <dan@telarity.com>
Closes: #xxxx


bagder (approved):

danielgustafsson:
Pushed with a comment typo-fix, thanks!

PR #13010, opened 2024-02-28
send-part-3

icing:
Further changes in send/upload handling

- Move all the "upload_done" handling to request.c
      - add possibility to abort sending of a request
      - add `Curl_req_done_sending()` for checks
      - transfer.c: readwrite_upload() now clean
- removing data->state.ulbuf and data->req.upload_fromhere
      - as well as data->req.upload_present
      - set data->req.upload_done on having read all from
        the client and completely flushed the send buffer
- tftp, remove setting of data->req.upload_fromhere
      - serves no purpose as `upload_present` is not set
        and the data itself is directly `sendto()` anyway
- smtp, make upload EOB conversion a client reader
- xfer_ulbuf addition
      - add xfer_ulbuf for borrowing, similar to xfer_buf
      - use in file upload
      - use in c-hyper body sending
- h1-proxy, remove init of data->state.ulbuf that is never used
- smb, add own send_buf instead of using data->state.ulbuf
- remove members of struct HTTP no longer used
- init null client reader on postfields if length is not positive

### DIFF
--- a/lib/c-hyper.c
+++ b/lib/c-hyper.c
@@ -730,7 +730,6 @@ static CURLcode bodysend(struct Curl_easy *data,
                          hyper_request *hyperreq,
                          Curl_HttpReq httpreq)
 {
-  struct HTTP *http = data->req.p.http;
   CURLcode result = CURLE_OK;
   struct dynbuf req;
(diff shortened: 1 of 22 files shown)


icing:
The failing hyper test is fixed by #13016.

PR #13012, opened 2024-02-29
editorconfig: line ending and indentation settings

fowl2:
This file will automatically configure supported code editors with the correct line endings and tab size as per `CODESTYLE`.

`max_line_length` seems less supported, but it looks like eg. VS Code might support it one day -https://github.com/microsoft/vscode/issues/111251

See https://editorconfig.org/ for reference.


dfandrich:
The policy so far has been to avoid editor-specific configuration files in the repository because it just doesn't scale.

fowl2:
I think the world has pretty much coalesced about just this one 😅

dfandrich:
It's an interesting proposal, with what looks like pretty wide support that solves the scaling issue I referred to. A few additions that are valid for curl:
* charset = utf-8
* spelling_language = en-US
The above should be valid everywhere except in a couple of `tests/data/test*` files that don't always conform (by design). Those files should probably be left at their default settings for these options to avoid erroneous conversion:
* charset, end_of_line, trim_trailing_whitespace, max_line_length, spelling_language
Also, I'd add a comment linking to https://editorconfig.org/ to help people wondering about the file

dfandrich:
Also, it needs a REUSE comment.

fowl2:
@dfandrich I think I've addressed your feedback, except I'm not sure what "REUSE comment" is

bagder:
> I'm not sure what "REUSE comment" is

Every file needs a copyright line and a license identifier. Maybe `.gitconfig` for example can serve as an example. See the top comments in that file.

fowl2:
Ahh, of course. Done.

bagder:
The pytest CI failures are fixed in #13037

.editorconfig
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) James May, <fowl@fowlsmurf.net>, et al.
+# SPDX-License-Identifier: curl
+
+# EditorConfig: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
```
bagder, line 9: This section contradicts `.gitattributes` which lists several extensions that use crlf.

Maybe using `[*.md]` here would be safer?

bagder:
No response, closing.

jay:
I think if this is ever resurrected that forcing the line endings lf is not a good idea because git on windows (by default?) converts the line endings to crlf when it checks out the files, and then when the files are committed it converts them back to lf.  there are some exceptions like we have in gitattributes.

PR #13022, opened 2024-03-01
configure: Don't make shell completions without perl

dfandrich:
The code that attempted to skip building the shell completions didn't
work properly and tried to build them even if perl wasn't available.
This step, as well as the install step, is now properly skipped without
perl.

Follow-up to 89733e2dd

Closes #13022


danielgustafsson (approved):
LGTM

Issue #13020, opened 2024-03-01
Torture test failure in test 56


dfandrich:
### I did this

`make test TEST_Q='-a -p -n -t146 56'`

The test failed with:

```
test 0056...[HTTP POST with *HUGE* request and chunked transfer-encoding]
 166 functions to make fail
 system() returned 34304
 56: torture FAILED: function number 146 in test.
 invoke with "-t146" to repeat this single case.
```
 When run under gdb, the message:

`lt-curl: /home/dan/src/curl-git/lib/request.c:240: req_send_buffer_add: Assertion `(size_t)n == blen' failed.`

appears. The failure can also be seen in [this log](https://github.com/curl/curl/actions/runs/8105176164/job/22153092741). This line was last changed in commit 9369c30cd8 PR #12969 @icing 

### I expected the following

The test should pass the torture tests.

### curl/libcurl version

curl 8.7.0-DEV commit 89733e2dd

### operating system

Linux x86_64 glibc

icing:
Nice find! Please see #13023 for the fix.

PR #13023, opened 2024-03-01
bufq, writing into a softlimit queue cannot be partial

icing:
- refs #13020
- when unable to obtain a new chunk on a softlimit bufq, this is an allocation error and needs to be reported as such.
- writes into a soflimit bufq never must be partial success

### DIFF
--- a/lib/bufq.c
+++ b/lib/bufq.c
@@ -396,7 +396,7 @@ ssize_t Curl_bufq_write(struct bufq *q,
   while(len) {
     tail = get_non_full_tail(q);
     if(!tail) {
-      if(q->chunk_count < q->max_chunks) {
+      if((q->chunk_count < q->max_chunks) || (q->opts & BUFQ_OPT_SOFT_LIMIT)) {
         *err = CURLE_OUT_OF_MEMORY;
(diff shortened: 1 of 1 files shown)


dfandrich:
The patch makes sense as far as I understand it, which unfortunately isn't very far.

icing:
> The patch makes sense as far as I understand it, which unfortunately isn't very far.

Happy to elaborate.

By default, a `bufq` is intialized with a chunk size and a max number of chunks which gives the total size it can hold. And write after that result in CURLE_AGAIN. But a write can also be partial, if there is some space left, but not enough to take in all bytes. This places then the burden on the caller to deal with that situation.

Initialized with flag `SOFTLIMIT` a `bufq` can take in data beyond the configured size. Then writes either fail with an error (when out of memory) or the complete passed array is written in. This is good for callers that have no way to deal with partial writes. Of course, the caller has the responsibility to not overdo it.

The patch fixes the situation where a malloc fails on a SOFTLIMIT bufq. This needs to be reported as failure of the write and not become a partial write.

bagder (approved):

Issue #13028, opened 2024-03-02
cmake does not build curl.1


dfandrich:
### I did this

`cmake -Bbuild . && make -C build -j9`

This builds the man pages in build/docs/libcurl/ but not  build/docs/curl.1  build/docs/curl-config.1 or build/docs/mk-ca-bundle.1  This causes tests 1140 and 1173 to fail.

### I expected the following

All the man pages should be built by cmake.

### curl/libcurl version

curl-8.7.0-DEV commit 9e2ee7049

### operating system

Linux x86_64

bagder:
So we have no CI builds verifying this or how did this problem sneak in?

bagder:
We have several cmake CI jobs in GHA macos.yml, but none of them runs the test suite.

The AppVeyor jobs that use cmake and run tests, have the tests 1140 and 1173 (and more) disabled.

cmake clearly is a second tier build system for building curl...

bagder:
This probably (?) never worked.

vszakats:
This was broken until maybe 8.6.0 or 8.5.0. Then we fixed it, but `ENABLE_CURL_MANUAL` stayed `OFF` by default, because the source tarball already provides these files (possibly even `mk-ca-bundle.1`, but I haven't followed that). So for most people it makes no sense to rebuild them (which also breaks reproducibility).

Solution: Maybe skip these tests if the necessary files are missing, then make sure to have perl/nroff around and set `ENABLE_CURL_MANUAL=ON` where we want these tests to run in CI?

dfandrich:
I've confirmed that `ENABLE_CURL_MANUAL=ON` causes curl.1 to be built and the tests to pass. But, this is inconsistent because
1. All 496 of the other man pages are built with `ENABLE_CURL_MANUAL=OFF`
2. automake builds build the manual by default (if all dependencies are there)

vszakats:
There is a separate control for the 496 libcurl man pages, called `BUILD_LIBCURL_DOCS`. `ON` by default. The source tarball doesn't ship them, so building them by default makes sense. Such option was recenly added for autotools too, called `--enable/disable-docs`, in 541321507e386744c3ea1200cc83cc98338e27d7.

IMO the autotools default is wrong in rebuilding the existing manuals even if they exist. This is double work (for possibly almost everyone outside curl's own CI), and breaks reproducibility.

One option would be to verify if the curl manuals exists and set the default accordingly in both cmake and autotools. I've been doing this in curl-for-win builds.

[...]

PR #13029, opened 2024-03-02
CI: Add CI build on Debian stretch to test old support

dfandrich:
This version still has ELTS support and contains some old versions of
key components like cmake to help prevent us from breaking that support.

Closes #13029

This currently fails tests due to #13028

### DIFF
--- a/.github/workflows/linux-old.yml
+++ b/.github/workflows/linux-old.yml
@@ -1,0 +1,95 @@
+# Copyright (C) Daniel Fandrich, <dan@coneharvesters.com>, et al.
+#
+# SPDX-License-Identifier: curl
+#
+# Compile on an old version of Linux that has barely the minimal build
+# requirements for CMake. This tests that curl is still usable on really
+# outdated systems.
+#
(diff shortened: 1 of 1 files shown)


dfandrich:
Debian stretch seems like the best old distro to use for this. See [#12920](12920#issuecomment-1937111591).

bagder:
#13028 is fixed. Rebase and try again?

dfandrich:
Tests 1140 and 1173 are still failing, despite the logs showing curl.1 is being built now. I thought it was related to out-of-tree builds, but that's not it either. It looks like curl.1 is being docs/cmdline-opts/curl.1 but the tests expect it in docs/curl.1. Copying it manually before running the tests makes it all work!

bagder:
> It looks like curl.1 is being docs/cmdline-opts/curl.1 but the tests expect it in docs/curl.1

Test 1140 and 1173 check for files with a wildcard that is wrong, but that just makes it *not* verify the file correctly. Which is probably also why these tests do not fail anywhere else.

dfandrich:
This build succeeds with PR #13197 in place.

PR #13033, opened 2024-03-03
cookie: if psl fails, reject the cookie

bagder:
A libpsl install without data and no built-in database is now considered bad enough to reject all cookies since they cannot be checked. It is somewhat of a user error, but still.

Reported-by: Dan Fandrich

### DIFF
--- a/lib/cookie.c
+++ b/lib/cookie.c
@@ -426,6 +426,7 @@ static void remove_expired(struct CookieInfo *cookies)
   }
 }
 
+#ifndef USE_LIBPSL
 /* Make sure domain contains a dot or is localhost. */
 static bool bad_domain(const char *domain, size_t len)
 {
@@ -443,6 +444,7 @@ static bool bad_domain(const char *domain, size_t len)
   }
   return TRUE;
 }
+#endif
 
 /*
   RFC 6265 section 4.1.1 says a server should accept this range:
@@ -1040,7 +1042,7 @@ Curl_cookie_add(struct Curl_easy *data,
         Curl_psl_release(data);
       }
       else
-        acceptable = !bad_domain(domain, strlen(domain));
+        infof(data, "libpsl problem, rejecting cookie for satety");
     }
 
     if(!acceptable) {


rsbeckerca:
Dan,

Are there instructions for obtaining valid data and/or setting up a built-in database?

Thanks,
Randall


From: Daniel Stenberg ***@***.***>
Sent: Sunday, March 3, 2024 12:44 PM
To: curl/curl ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [curl/curl] cookie: if psl fails, reject the cookie (PR #13033)


A libpsl install without data and no built-in database is now considered bad enough to reject all cookies since they cannot be checked. It is somewhat of a user error, but still.

Reported-by: Dan Fandrich

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/curl/curl/pull/13033

Commit Summary

  *   d424c8f<https://github.com/curl/curl/pull/13033/commits/d424c8f7e5c3716d6ad504fe2df982f3f299d212> cookie: if psl fails, reject the cookie

File Changes

(1 file<https://github.com/curl/curl/pull/13033/files>)

  *   M lib/cookie.c<https://github.com/curl/curl/pull/13033/files#diff-66dd8965a3e15c2dd1927e7e9d8ad8505b9458cf265c635cf630c627b74f084b> (4)

Patch Links:

  *   https://github.com/curl/curl/pull/13033.patch
  *   https://github.com/curl/curl/pull/13033.diff

bagder:
> Are there instructions for obtaining valid data and/or setting up a built-in database?

That's really a question for the libpsl team.

A default libpsl gets built with a built-in default PSL data set, which basically makes it impossible to fail - if there is no dynamic one to load from disk, it will fall back to the built-in set.

rsbeckerca:
> > Are there instructions for obtaining valid data and/or setting up a built-in database?
> 
> That's really a question for the libpsl team.
> 
> A default libpsl gets built with a built-in default PSL data set, which basically makes it impossible to fail - if there is no dynamic one to load from disk, it will fall back to the built-in set.

Thanks. Will go there for answers. Appreciate the reply.

PR #13045, opened 2024-03-04
http: better error message for HTTP/1.x response without status line

mkauf:
If a response without a status line is received, and the connection is known to use HTTP/1.x (not HTTP/0.9), report the error "Invalid status line" instead of "Received HTTP/0.9 when not allowed".


bagder:
The first line coming from a HTTP/1.0 or HTTP/1.1 server is called [the Status Line](https://www.rfc-editor.org/rfc/rfc9112.html#name-status-line). The headers start after that initial line.

mkauf:
@bagder : Thanks, I have updated this pull request.

bagder (approved):

bagder:
Thanks!

PR #13047, opened 2024-03-04
docs: ascii version of manpage without nroff

bagder:
# create ascii version of the manpage without nroff
 
 - build src/tool_hugegelp.c from the ascii manpage
 - the newly generated tool_hugehelp is 3K lines shorter and 50K smaller
 - move the the manpage and the ascii version build to docs/cmdline-opts
 - remove all use of nroff from the build process
 - should make the build entirely reproducible (by avoiding nroff)
    
## The ascii version of the manpage:
    
  - is built with gen.pl, just like the manpage is
  - has a right-justified column making the appearance similar to the previous version
  - uses a 4-space indent per level (instead of the old version's 7)
  - does not do hyphenation of words (which nroff does)

## history

We first made the curl build use nroff for building the hugehelp file in December 1998, for curl 5.2.

## `curl -M` sample

This is what the top of new version looks like in my terminal:

curl-txt

### DIFF
--- a/CMake/Macros.cmake
+++ b/CMake/Macros.cmake
@@ -68,35 +68,6 @@ macro(curl_internal_test CURL_TEST)
   endif()
 endmacro()
 
-macro(curl_nroff_check)
-  find_program(NROFF NAMES gnroff nroff)
-  if(NROFF)
-    # Need a way to write to stdin, this will do
-    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/nroff-input.txt" "test")
-    # Tests for a valid nroff option to generate a manpage
-    foreach(_MANOPT "-man" "-mandoc")
-      execute_process(COMMAND "${NROFF}" ${_MANOPT}
-        OUTPUT_VARIABLE NROFF_MANOPT_OUTPUT
-        INPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/nroff-input.txt"
-        ERROR_QUIET)
-      # Save the option if it was valid
-      if(NROFF_MANOPT_OUTPUT)
-        message("Found *nroff option: -- ${_MANOPT}")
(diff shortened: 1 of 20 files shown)


bagder:
I'm trying hard to figure out what exactly it is that fails in the cmake + MSVC jobs on AppVeyor... :confused: And nowhere else.

bagder:
The error message from cd2nroff: `Failed to open CURLOPT_SOCKS5_GSSPI_SERVICE.md : No such file or directory` is puzzling, as it has lost am `A` in the middle of the file name.

The command is invoked by cmake where it provides a *long* list of `*.md` file names for cd2nroff to convert to nroff.

bagder:
The latest test in this PR that seems to have made the build work again is a partial revert of 2620aa930bc73a that @vszakats did. It seems as if the very long command line from cmake is problematic? What I can't explain is if why that suddenly would become a problem in this PR, nor why it does not fail more reliably.

The command line in question is over 10,000 characters long and Windows is often said to have an 8K limit (?). The change in this PR converts that make to single script invokes for each conversion.

bagder:
> To avoid that, maybe `cd2nroff` could be extended to accept a list in file (e.g. `@inputs.txt`)? That could be generated by CMake dynamically. Going back to single invocations is significantly slower.

I can't see why not. Generating them all in a single invoke makes it ineffective dependency wise as then it will rebuild them all even if only one is changed, but especially on Windows generating them all like that will likely be noticeably faster when all of them need to be rendered.

I will however not try to do this within this PR as it seems unrelated. We could of course argue that my revert is also unrelated, but I could not get this PR to build clean on appveyor without that!

vszakats:
So far so good: https://github.com/curl/curl-for-win/actions/runs/8176058376

I'll re-compare the generated vs distributed `--manual` content when the daily snapshot tarball gets its update tomorrow.

PR #13053, opened 2024-03-05
Add support for OpenSSL and GnuTLS for SHA-512/256

Karlson2k:
This is a follow-up for PR #12897.

This PR adds support for SHA-512/256 digest calculation by TLS backends.
Currently only OpenSSL and GnuTLS (actually, nettle) support SHA-512/256.

### DIFF
--- a/lib/curl_sha512_256.c
+++ b/lib/curl_sha512_256.c
@@ -29,15 +29,132 @@
 #include "curl_sha512_256.h"
 #include "warnless.h"
 
-#if defined(USE_GNUTLS)
+/* The recommended order of the TLS backends:
+ * * OpenSSL
+ * * GnuTLS
+ * * wolfSSL
+ * * Schannel SSPI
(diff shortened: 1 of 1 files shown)


Karlson2k:
Updated to support LibreSSL, which added support for SHA-512/256 only in May 2023 while declaring compatibility with OpenSSL "2.0.0".

Karlson2k:
@bagder Hopefully I was able to follow the code style, and the PR is less intrusive than previous one.
The PR is ready for review (and merge 😄).

I added a couple of cosmetic fixes, not really related to the new implementation.
Let me know, please, whether I should remove some comments or change something else in this PR.

bagder (approved):

Karlson2k:
It has been temporarily marked as "draft" to avoid merging while the alternative PR #13070 is being published.

bagder:
Let's go with #13070

PR #13054, opened 2024-03-06
http2, push headers cleanup

icing:
- provide common cleanup method for push headers

### DIFF
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -276,8 +276,7 @@ static void free_push_headers(struct h2_stream_ctx *stream)
   size_t i;
   for(i = 0; i<stream->push_headers_used; i++)
     free(stream->push_headers[i]);
-  free(stream->push_headers);
-  stream->push_headers = NULL;
+  Curl_safefree(stream->push_headers);
   stream->push_headers_used = 0;
 }
 
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1431,10 +1431,10 @@ static int on_header(nghttp2_session *session, const nghttp2_frame *frame,
         return NGHTTP2_ERR_TEMPORAL_CALLBACK_FAILURE;
       }
       stream->push_headers_alloc *= 2;
-      headp = Curl_saferealloc(stream->push_headers,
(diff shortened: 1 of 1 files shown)


lib/http2.c
```diff
@@ -271,6 +271,16 @@ static CURLcode http2_data_setup(struct Curl_cfilter *cf,
   return CURLE_OK;
 }
 
+static void free_push_headers(struct h2_stream_ctx *stream)
+{
+  size_t i;
+  for(i = 0; i<stream->push_headers_used; i++)
+    free(stream->push_headers[i]);
+  free(stream->push_headers);
```
danielgustafsson, line 279: nitpick, but wouldn't `Curl_safefree()` be more appropriate here?

bagder: :+1:

icing: Changed.

danielgustafsson (approved):
LGTM

bagder:
I think it also needs this:
~~~diff
diff --git a/lib/http2.c b/lib/http2.c
index 2112545c5..725ec742c 100644
--- a/lib/http2.c
+++ b/lib/http2.c
@@ -1431,14 +1431,14 @@ static int on_header(nghttp2_session *session, const nghttp2_frame *frame,
         failf(data_s, "Too many PUSH_PROMISE headers");
         free_push_headers(stream);
         return NGHTTP2_ERR_TEMPORAL_CALLBACK_FAILURE;
       }
       stream->push_headers_alloc *= 2;
-      headp = Curl_saferealloc(stream->push_headers,
-                               stream->push_headers_alloc * sizeof(char *));
+      headp = realloc(stream->push_headers,
+                      stream->push_headers_alloc * sizeof(char *));
       if(!headp) {
-        stream->push_headers = NULL;
+        free_push_headers(stream);
         return NGHTTP2_ERR_TEMPORAL_CALLBACK_FAILURE;
       }
       stream->push_headers = headp;
     }
     h = aprintf("%s:%s", name, value);

icing:
> I think it also needs this:

Correct. Just added it. Thanks for finding.

Issue #13065, opened 2024-03-06
`Curl_resolver_init` is not thread-safe


nickitat:
### I did this

Hi!

We start getting sanitiser reports like [this](https://s3.amazonaws.com/clickhouse-test-reports/60462/c8e41dc07ca5448b2f007493cb7cc58916ad22b0/integration_tests__tsan__[3_6]//home/ubuntu/actions-runner/_work/_temp/test/output_dir/integration_run_parallel2_0.log) in our CI.

I assume that this code is supposed to be thread-safe, please correct me if I'm wrong.

If so, seems like the problematic code was introduced recently here:

https://github.com/curl/curl/commit/4224d6e0f3cc8ddf2e9e1aa82c356d9b88b53e43

so static variable `ares_ver` could be simultaneously read and modified by two threads. 
it seems enough to replace l. 176 with 

``` c
static int ares_ver = get_ares_version();
```

`get_ares_version()` will do the same call, but now compiler will guarantee single initialisation (btw, I'm not a C expert, do we have this guarantee in C?).

### I expected the following

no sanitiser complains

### curl/libcurl version

curl 8.6.0+

### operating system

linux

jay:
I don't see get_ares_version in our source or the c-ares api. See #13066 for proposed fix

bagder:
I'm pretty sure this is still thread safe, even if we should make it done nicer.

nickitat:
thank you, guys
when could we expect the next release?

bagder:
The next release ships on **March 27, 2024** and will become version **8.7.0**. The pending release notes for the coming release are always available on this URL: https://curl.se/dev/release-notes.html

nickitat:
I'd also love to get your advice on how we should proceed on our end. is it safe to use master version instead of the last release? to get the fix early. or maybe we could back-port the fix to the last release branch?

jay:
> is it safe to use master version instead of the last release? to get the fix earlier. or maybe we could back-port the fix to the last release branch?

Do not use master in production. master is our development branch. I suggest use git to cherry pick the fixes you need.

bagder:
> is it safe to use master version instead of the last release?

I believe sticking to the previous version is fine until the next release ships. I think this *not thread-safe* warning was a warning and it should be and was fixed, but the effects of the problem are totally harmless.

PR #13066, opened 2024-03-06
asyn-ares: fix data race warning

jay:
- Don't use a static variable to store the ares version.

Prior to this change several threads could write the same data to a static int variable at the same time. Though in practice it's not a problem ThreadSanitizer may warn.

Reported-by: Nikita Taranov

Fixes https://github.com/curl/curl/issues/13065
Closes #xxxxx


bagder:
See also #13000 which is another take on the same problem.

jay:
> See also #13000 which is another take on the same problem.

Ok. Closing in favor of #13000.

PR #13071, opened 2024-03-07
mbedtls: properly cleanup the thread-shared entropy

jay:
- Store the state of the thread-shared entropy for global init/cleanup.

- Use curl's thread support of mbedtls for all Windows builds instead of just when the threaded resolver is used via USE_THREADS_WIN32.

Prior to this change on global cleanup curl builds that have curl thread support for mbedtls freed the entropy (8b1d2298) but failed to mark that it had been freed, which caused problems on subsequent init + transfer.

Bug: https://github.com/curl/curl/discussions/11919#discussioncomment-8687105
Reported-by: awesomekosm@users.noreply.github.com

Closes #xxxxx

### DIFF
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -110,7 +110,8 @@ struct mbed_ssl_backend_data {
 };
 
 /* apply threading? */
-#if defined(USE_THREADS_POSIX) || defined(USE_THREADS_WIN32)
+#if (defined(USE_THREADS_POSIX) && defined(HAVE_PTHREAD_H)) || \
+    defined(_WIN32)
 #define THREADING_SUPPORT
 #endif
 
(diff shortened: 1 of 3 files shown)


jay:
The mbedtls threading support in curl is from like 10 years ago and was before mbedtls supported threading properly. However I notice they still do not enable threading support in their library by default. Currently the way curl works is it only uses its separate threading support (mbedtls_threadlock.c) for mbedtls if USE_THREADS_POSIX or USE_THREADS_WIN32 (as in, use the threaded resolver).

I changed it for Windows to always use it when _WIN32. Someone familiar with mbedtls will have to suggest what else should be done, like if we should just revert to relying on mbedtls to be built with thread support. But that's outside the scope of this PR.

bagder:
mbedTLS is simply not a modern TLS library and **we cannot recommend using it**. The lack of threading support by default, the lack of TLS 1.3 support (at least in curl) etc.

bagder (approved):

wyattoday:
@bagder 

I have a PR in progress that addresses both threading and TLS 1.3 with mbedTLS. It also addresses and cleans up other issues.

It amounts to a full audit of the mbedTLS support in curl. Yeah, the code needed some love.

I’ll post a draft of it in a week or 2.

PR #13072, opened 2024-03-07
digest: add check for hashing error

Karlson2k:
The digest auth functions does not check for hashing errors.
The PR fixes this problem by adding checks for error reported by hashing functions.

Actually, when hashing function fails the current code may use initialised data (the hashing result).

This PR is partially related to PR #13053 / #13070.


bagder (approved):

bagder:
Thanks!

PR #13078, opened 2024-03-07
buildconf.bat: remove outdated groff/nroff use

bagder:
Follow-up to f03c85635f35269

- don't try to generate the real hugehelp file, because it requires `curl.txt` which needs a build
- don't attempt to do anything in a c-ares subdirectory


bagder:
Hm, this assumes there is a `curl.txt` present, which requires the man page to have been built...

PR #13083, opened 2024-03-07
test1140/1173: extend wildcards to find curl.1

bagder:
... in its new build path.

Also update the test scripts to be more precise in error messages to help us understand CI errors better.

Ref: #13029

Follow-up to f03c85635f35269f1

### DIFF
--- a/docs/cmdline-opts/Makefile.am
+++ b/docs/cmdline-opts/Makefile.am
@@ -53,6 +53,3 @@ $(ASCIIPAGE): $(DPAGES) $(SUPPORT) mainpage.idx Makefile.inc gen.pl
 
 listhelp:
 	./gen.pl listhelp $(DPAGES) > $(top_builddir)/src/tool_listhelp.c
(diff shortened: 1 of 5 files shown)


dfandrich:
Leaving the old glob there with no corresponding file means the shell doesn't expand it and it ends up as a literal string and the error  `test1140.pl could not open /tmp/cm/tests/../docs/*.1 ...` occurs

bagder:
There should still be corresponding files there: `curl-config.1` and `mk-ca-bundle.1`

bagder:
... which is why the tests still run fine on configure builds.

dfandrich:
Maybe there should be, but they're not.  Run "cmake . && make -j9 && make -j9 test-ci TFLAGS='1140 1173'" in a new git clone and see for yourself.

bagder:
I believe you, I'm just hinting that the problem is still in the cmake build of the docs.

PR #13087, opened 2024-03-08
TLS, start shutdown only when peer did not already close

icing:
- refs #10290
- When curl sees a TCP close from the peer, do not start a TLS shutdown. TLS shutdown is a handshake and if the peer already closed the connection, it is not interested in participating.

### DIFF
--- a/lib/cw-out.c
+++ b/lib/cw-out.c
@@ -231,7 +231,7 @@ static CURLcode cw_out_ptr_flush(struct cw_out_ctx *ctx,
     }
     if(nwritten != wlen) {
       failf(data, "Failure writing output to destination, "
-            "passed %zu returned %zu", wlen, nwritten);
+            "passed %zu returned %zd", wlen, nwritten);
       return CURLE_WRITE_ERROR;
     }
     *pconsumed += nwritten;
--- a/lib/vtls/gtls.c
+++ b/lib/vtls/gtls.c
@@ -117,6 +117,8 @@ static ssize_t gtls_pull(void *s, void *buf, size_t blen)
                                (CURLE_AGAIN == result)? EAGAIN : EINVAL);
     nread = -1;
   }
+  else if(nread == 0)
+    connssl->peer_closed = TRUE;
   return nread;
 }
 
@@ -1489,7 +1491,7 @@ static int gtls_shutdown(struct Curl_cfilter *cf,
     bool done = FALSE;
     char buf[120];
 
-    while(!done) {
+    while(!done && !connssl->peer_closed) {
       int what = SOCKET_READABLE(Curl_conn_cf_get_socket(cf, data),
                                  SSL_SHUTDOWN_TIMEOUT);
       if(what > 0) {
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -769,6 +769,9 @@ static int ossl_bio_cf_in_read(BIO *bio, char *buf, int blen)
     if(CURLE_AGAIN == result)
       BIO_set_retry_read(bio);
   }
+  else if(nread == 0) {
+    connssl->peer_closed = TRUE;
+  }
 
   /* Before returning server replies to the SSL instance, we need
    * to have setup the x509 store or verification will fail. */
@@ -1887,7 +1890,9 @@ static void ossl_close(struct Curl_cfilter *cf, struct Curl_easy *data)
   DEBUGASSERT(backend);
 
   if(backend->handle) {
-    if(cf->next && cf->next->connected) {
+    /* Send the TLS shutdown if we are still connected *and* if
+     * the peer did not already close the connection. */
+    if(cf->next && cf->next->connected && !connssl->peer_closed) {
       char buf[1024];
       int nread, err;
       long sslerr;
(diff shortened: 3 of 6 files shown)


dorrogeray:
Hi, I have created a repository from which you can reproduce the issue that we are seeing mentioned in https://github.com/curl/curl/issues/10290#issuecomment-1980536370, where the curl does not respond to FIN ACK coming from the server by sending FIN ACK, and instead sends back ACK and tries to re-use the connection and push application data one second later: 

https://github.com/dorrogeray/curl-kinesis-rst-issue

image

I believe the curl should gracefully send back FIN ACK and remove the connection from pool (since its getting closed from the other side).

Does this make sense or am I missing something?

dorrogeray:
@icing hi, I have re-tested the same scenario with your fix, here is side by side comparison:
image
[pcap_comparison.zip](https://github.com/curl/curl/files/14555809/pcap_comparison.zip)

I think it looks much better, but I still have one question:

I can see that the FIN ACK is received by curl at 5th second, but the FIN ACK is sent back from curl only at 6th second (this is the time we try to use the same curl handler again to push more data to the kinesis stream).

Shouldn't the curl respond with the FIN ACK immediately after it receives FIN ACK (as long as its not in the process of pushing some data through at that time) rather then waiting for the next initiation from the user (the next attempt to use that connection)? Because that next initiation from the user could come after a significant amount of time (and the other side probably will not wait that long to recieve the FIN ACK anyways).

Here is what I mean, when I adjust the script to try to use curl handle after 15 seconds, and the first connection is dropped by kinessis after 5 seconds, there is 10 seconds between curl receiving FIN ACK and curl confirming FIN ACK:
image 
Pcap for this is in the attached zip file as well.

This can happen very easily when you are using workers which might have some idle time between needing to use the same connection via curl again.

icing:
@dorrogeray thanks for the traces. I believe what you see here is the  different between packets arriving at TCP level (and visible in your pcap) and curl actually **reading** the socket and seeing the FIN.

When a connection is no longer used and placed in the connection cache, its socket is no longer monitored by curl. When a new transfer then needs a connection, the cache is inspected, finds a match, checks the "liveliness" of the cached connection and *then* sees the FIN. Found to no longer healthy, it is then closed.

So, indeed, the time between the server closing a connection and curl noticing this in its cache can be very large.

dorrogeray:
@icing thanks for the explanation!

Maybe this problem could be solved by https://curl.se/libcurl/c/CURLOPT_MAXAGE_CONN.html so that when this option is set, curl would pro-actively inspect the connection and:

1. If it does not pass the "liveliness" check, and FIN ACK was received, reply with FIN ACK and gracefully close the connection
2. If it does pass the "liveliness" check, gracefully close the connection by sending FIN ACK, waiting for FIN ACK back and close the connection?

I our particular use-case with the kinesis, what we could then do is set `CURLOPT_MAXAGE_CONN` to 6 seconds. So after 5 seconds of inactivity, kinesis sends FIN ACK, and 1 second later, the proactive check from curl would "discover" it, send FIN ACK back and gracefully close the connection?

Alternatively, an option like `CURLOPT_CONN_HEALTHCHECK_INTERVAL` could be introduced to allow everyone to configure how often (if at all) curl should inspect the connections in the cache.

We are operating at scale and "hanging" TCP connections are having a major impact. For now we are forced to go with `CURLOPT_FORBID_REUSE` but then we are needlessly re-creating many TCP connections which (if they were not ending hanging) could be re-used dozens of times.

bagder:
> @bagder: there can be an argument made that we should poll the sockets in the connection cache to react more timely to closes, for example. But not that easy to add as a feature, since those connection no longer have a transfer.

Yes certainly, and one that has been brought up every now and then since the dawn of time (or at least since we created the connection pool). For "upkeep" reasons as well, so that an idle h2/h3 connection could still reply to pings etc.

bagder:
Actually, `CURLOPT_UPKEEP_INTERVAL_MS` should do this...

icing:
> Actually, `CURLOPT_UPKEEP_INTERVAL_MS` should do this...

Ah, learnt something again.

dorrogeray:
@bagder @icing Are you saying that setting `CURLOPT_UPKEEP_INTERVAL_MS` to lets say `1000` and calling https://php.watch/versions/8.2/curl_upkeep should already work in terms of detecting connections which have sent FIN ACK to curl but were not yet gracefully closed from curl side? Or is there still implementation work which needs to be done, and you are suggesting to put it behind `CURLOPT_UPKEEP_INTERVAL_MS`?

icing:
> @bagder @icing Are you saying that setting `CURLOPT_UPKEEP_INTERVAL_MS` to lets say `1000` and calling https://php.watch/versions/8.2/curl_upkeep should already work in terms of detecting connections which have sent FIN ACK to curl but were not yet gracefully closed from curl side? Or is there still implementation work which needs to be done, and you are suggesting to put it behind `CURLOPT_UPKEEP_INTERVAL_MS`?

There is still work to be done. The general mechanism is there, as @bagder said, but currently only HTTP/2 connection have an implementation using PING frames.

bagder:
Yeah, I think there is a framework to build upon, it does not do this right now.

dorrogeray:
Understood, thanks for the explanation. :+1: I would be happy help, but unfortunately this is outside of my skillet. If anyone wants to take a stab at the implementation as suggested above, I will be more than happy to help you test the solution & verify the impact at scale.

PR #7196, opened 2021-06-04
spnego_gssapi: implement TLS channel bindings for openssl

steffen-kiess:
This requires krb5 >= 1.19 because otherwise channel bindings won't be
forwarded through SPNEGO.

PR #13098, opened 2024-03-11
spnego_gssapi: implement TLS channel bindings for openssl

Foorack:
Channel Bindings are used to tie the session context to a specific TLS channel. This is to provide additional proof of valid identity, mitigating authentication relay attacks.

Major web servers have the ability to require (None/Accept/Require) GSSAPI channel binding, rendering Curl unable to connect to such websites unless support for channel bindings is implemented.

IIS calls this feature Extended Protection (EPA), which is used in Enterprise environments using Kerberos for authentication.

This change require krb5 >= 1.19, otherwise channel bindings won't be forwarded through SPNEGO.

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -5135,8 +5135,7 @@ static CURLcode ossl_get_channel_binding(struct Curl_easy *data, int sockindex,
   /* No X509_get_signature_nid support */
   (void)data; /* unused */
   (void)sockindex; /* unused */
-  *binding = NULL;
-  *len = 0;
+  (void)binding; /* unused */
   return CURLE_OK;
 #endif
 }


Foorack:
"Linux / hyper" CI failed during `install hyper` step, due to compile error in Hyper and was fixed in https://github.com/hyperium/hyper/commit/4d0c184f935adf93ed05b86ec3a16e432aa81706

lib/vtls/openssl.c
```diff
@@ -4760,6 +4760,95 @@ static ssize_t ossl_recv(struct Curl_cfilter *cf,
...
+  return CURLE_OK;
+#else
+  /* No X509_get_signature_nid support */
+  (void)data; /* unused */
+  (void)sockindex; /* unused */
+  *binding = NULL;
+  *len = 0;
+  return CURLE_OK;
+#endif
+}
```
DemiMarie, line 4850: Would it be possible to avoid APIs that perform implicit fetching?

Foorack: Hello @DemiMarie! Sorry, but what do you mean by "implicit fetching"? Do you have an alternative API in mind?

Foorack: Hi @DemiMarie :wave: Please let me know if there is any outstanding requested changes on this; either if you have any alternative API's in mind, or if I can mark this conversation as resolved.

Foorack:
Rebased to bring PR up-to-date with master due to time passed.

### Minor changes

* `ossl_ssl_backend_data` had in upstream been renamed to `ossl_ctx`, and `->handle` to `->ssl`
* changed `u_char*` to `unsigned char*` to be in line with rest of project

lib/vtls/openssl.c
DemiMarie: “Implicit fetching” means calls like `EVP_sha256()`, as opposed to explicitly looking up the algorithm in a provider.

lib/vtls/openssl.c
```diff
@@ -5157,23 +5157,39 @@ static CURLcode ossl_get_tls_server_end_point(struct Curl_easy *data,
...
+    algo_name = OBJ_nid2sn(algo_nid);
+    if(!algo_name) {
+      failf(data, "Could not find digest algorithm name for NID %d",
+            algo_nid);
       return CURLE_SSL_INVALIDCERTSTATUS;
     }
   }
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+  algo_type = EVP_MD_fetch(NULL, algo_name, NULL);
```
DemiMarie, line 5172: It’s best to cache the result of this, for performance reasons.

Foorack: Hi @DemiMarie! Thank you for continued review.
Where would it be best to cache this? I am honestly not too familiar with this level of curl internals. I am primarily working this as a continuation of the great work done by @steffe-kiess (#7196) to add support for SPNEGO GSSAPI, as it is something we critically need from command-line usage due to some web servers now enforcing this. Any pointer in right direction of caching architecture would be greatly appreciated.

I have found little material online regarding the performance impact of repeatedly calling `EVP_MD_fetch` too much. The explicit fetching will only be done once per connection, during initial connect, so for most cases this function will only be called once. This is a also a completely new feature, as curl has been unable to ever connect to servers requiring secure channel binding. Therefore wondering if caching is something which could be done in a future patch if it turns out to be too much work.

Finally, regarding Explicit Fetching, I have not found any other usage of `EVP_*_fetch` in the entire curl source code... All other places use `EVP_sha256()` and similar implicit functions. I agree it could be a good idea to support explicit fetching, but I'd ideally not want this change to be a testing ground for this migration. Rather reverting back to the implicit-only implementation and then adding OpenSSL 3.x Explicit Fetching support in a separate PR as it touches many more areas outside of this scope. Thoughts?

DemiMarie: Yup, separate PR.

Foorack: OK! Do you wanna do explicit+caching in a separate PR, or only caching?

DemiMarie: I’m not a curl developer, just someone who noticed this PR while reading about the mess that OpenSSL 3.x is causing for HAProxy.

Foorack:
Reverted Explicit Fetching implementation for now, in order to be in-line with the rest of CURL source code. Potential migration from Implicit to Explicit Fetching is outside the scope of this PR.

Also resolved merge conflicts to keep branch up-to-date.

Foorack:
@bagder We've been running this change in Production on hundreds of clients for several months, and it's been working great. What is the next step for this to be merged?

lib/vtls/openssl.c
```diff
@@ -5044,6 +5044,95 @@ static ssize_t ossl_recv(struct Curl_cfilter *cf,
...
+  if(!X509_digest(cert, algo_type, buf, &length)) {
+    failf(data, "X509_digest() failed");
+    return CURLE_SSL_INVALIDCERTSTATUS;
+  }
+
+  *binding = malloc(sizeof(prefix) - 1 + length);
+  if(!*binding)
+    return CURLE_OUT_OF_MEMORY;
+  memcpy(*binding, prefix, sizeof(prefix) - 1);
+  memcpy(*binding + sizeof(prefix) - 1, buf, length);
```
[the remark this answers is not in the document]
Foorack: @bagder I have now replaced the raw use of malloc+memcpy with `struct dynbuf`. I have also rebased on master and tested the latest build internally against both SCB and non-SCB targets and works well.

Foorack:
Some tests failed because Launchpad was temporarily down.

Foorack:
All tests pass now!

MarcelRaad (approved):
Looks good to me, thanks! It's a pity that nothing can be shared with the existing SSPI implementation as that's based on SSPI-specific types. I would appreciate more reviews from others interested in the topic (not necessarily curl maintainers).

bagder:
Thanks!

Issue #12303, opened 2023-11-10
Trouble with  --ca-native flag


nitrogene:
### I did this

Using an official 8.4.0 curl installer for windows, I have troubles using the --ca-native flag.  

Here's the current folder content:

```dos
Répertoire de C:\Users\WDAGUtilityAccount\Downloads\curl-8.4.0_6-win32-mingw\curl-8.4.0_6-win32-mingw\bin

11/10/2023  12:25 PM    <DIR>          .
11/10/2023  12:25 PM    <DIR>          ..
11/10/2023  12:15 PM           221,470 curl-ca-bundle.crt
11/10/2023  12:15 PM         4,521,544 curl.exe
11/10/2023  12:15 PM             2,326 libcurl.def
11/10/2023  12:15 PM         4,300,360 libcurl.dll
               4 fichier(s)        9,045,700 octets
               2 Rép(s)  39,957,020,672 octets libres
```

first try OK:
```dos
C:\Users\WDAGUtilityAccount\Downloads\curl-8.4.0_6-win32-mingw\curl-8.4.0_6-win32-mingw\bin>curl https://catfact.ninja/fact
{"fact":"Cats are North America\u2019s most popular pets: there are 73 million cats compared to 63 million dogs. Over 30% of households in North America own a cat.","length":149}
```

second try KO - as expected (we are renaming the crt):
```dos
C:\Users\WDAGUtilityAccount\Downloads\curl-8.4.0_6-win32-mingw\curl-8.4.0_6-win32-mingw\bin>ren curl-ca-bundle.crt curl-ca-bundle.crt.old

C:\Users\WDAGUtilityAccount\Downloads\curl-8.4.0_6-win32-mingw\curl-8.4.0_6-win32-mingw\bin>curl https://catfact.ninja/fact
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

third try KO, but should be OK (we are renaming the crt file, but are using the --ca-native flag):

```dos
C:\Users\WDAGUtilityAccount\Downloads\curl-8.4.0_6-win32-mingw\curl-8.4.0_6-win32-mingw\bin>curl https://catfact.ninja/fact --ca-native
curl: (60) SSL certificate problem: unable to get local issuer certificate
[...]

PR #13111, opened 2024-03-12
Add CURL_USE_CA_NATIVE option

talregev:
Add CURL_USE_CA_NATIVE option

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -399,6 +399,11 @@ cmake_dependent_option(CURL_USE_BEARSSL "Enable BearSSL for SSL/TLS" OFF CURL_EN
 cmake_dependent_option(CURL_USE_WOLFSSL "Enable wolfSSL for SSL/TLS" OFF CURL_ENABLE_SSL OFF)
 cmake_dependent_option(CURL_USE_GNUTLS "Enable GnuTLS for SSL/TLS" OFF CURL_ENABLE_SSL OFF)
 
+option(CURL_USE_CA_NATIVE "Use standard certificate store of operating system" OFF)
+if(CURL_USE_CA_NATIVE)
+  set(USE_CA_NATIVE ON)
+endif()
+
 set(openssl_default ON)
 if(WIN32 OR CURL_USE_SECTRANSP OR CURL_USE_SCHANNEL OR CURL_USE_MBEDTLS OR CURL_USE_WOLFSSL)
   set(openssl_default OFF)
--- a/lib/curl_config.h.cmake
+++ b/lib/curl_config.h.cmake
@@ -741,6 +741,9 @@ ${SIZEOF_TIME_T_CODE}
 /* to enable Windows SSL  */
 #cmakedefine USE_SCHANNEL 1
 
+/* Use standard certificate store of operating system */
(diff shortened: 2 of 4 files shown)


bagder:
But why? We already have the `CURLSSLOPT_NATIVE_CA`  bit for `CURLOPT_SSL_OPTIONS` that the application can set if this is needed. Why would we want to complicated the library with a new build option like this?

You need to motivate this.

talregev:
I tried to discuss with you before open this PR.
Maybe there is other way to do what I need without this flag.
The goal of this PR it that I will have an option that when I will run curl and libcurl it automatic will check the os native certificate without the need to do anything else.  That the flag --ca-native will apply implicit. 
More specific, that when I compile curl with openssl that it will check the os native certificate, on windows and any other os.
Why?
Because most of the people use curl with schannel backend because it check the os native certificate automatic.
They don't know (I didn't know) that this option can done with openssl but with extra steps. 
The expected behavior for curl is that it will check os native certificate automatic when it not find other source of validate the cartificate. I am sure there is a good reason why curl behave differently, and what I ask break that reason. That why I am offer alternative that we have both option with flag that turn off by default. 
More reason Why:
The stronger motivation is because http3. http3 cannot compile with schannel backend. it need other ssl backend that lead to multi ssl feature, that clash with http3.
The prefer way to compile curl with http3 is with openssl with nghttp3 (Yes, my previous PR). 
I want that when compile openssl on windows, it will behave as schannel backend. (the expected behavior) and check ca os native automatic. 
More reason Why:
Some of the open source project compile curl with vcpkg. (You also added to instructions how to compile curl with vcpkg, so you at least aware of it).
I would like to add http3 feature to vcpkg, and that the default backend will be openssl, because openssl can compile on most os. And I want that it will check os native certificate as schannel does on windows, or any other os that it compile. 
That will really help that transition.  

Thank you for your time reading this.

bagder:
> I tried to discuss with you before open this PR.

1. curl is a whole community, it is not me. A PR is a request to change curl and it needs to be motivated or we will just ignore it.
2. you did not refer to nor mention that discussion in this PR

>  Because most of the people use curl with schannel backend because it check the os native certificate automatic.

That is your guess, not a fact. I believe most curl+schannel users use that setup simply because that is how curl comes bundled on Windows 10 and 11 and they haven't considered or learned about any other options.

>  The prefer way to compile curl with http3 is with openssl with nghttp3 (Yes, my previous PR).

That configuration does not work properly yet and is **experimental**. We recommend using the ngtcp2 and nghttp3 setup (with GnuTLS, quictls or wolfSSL). The only non-experimental HTTP/3 setup currently in curl.

> You also added to instructions how to compile curl with vcpkg,

I did? If so, I merged someone else's instructions. I thought vcpkg did the build for you.

----

I'm curious in hearing what others have to say about this proposal, but I don't think we need this. We offer this feature already just not by default as changing the default behavior is a slippery slope that we are a little afraid of.

jay:
I don't think it's needed. Also I'm bothered by #12303 and I don't have any solution for it.

talregev:
I am reopen this PR that also vcpkg community will share their thoughts

bagder:
I don't sense a demands for this.

PR #13124, opened 2024-03-14
GHA/linux: GitHub image runner update broke the sanitizer jobs

bagder:
The GitHub image runner update from 20240304.1.0 to 20240310.1 introduces all sorts of problems in our sanitizer builds.

According to https://github.com/actions/runner-images/issues/9491:

"The issue is caused by incompatibility between llvm 14 provided in ubuntu-22.04 image and the much newer kernel configured with high-entropy ASLR"

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -139,7 +139,7 @@ jobs:
               CFLAGS="-fsanitize=memory -Wformat -Werror=format-security -Werror=array-bounds -g"
               LDFLAGS="-fsanitize=memory"
               LIBS="-ldl"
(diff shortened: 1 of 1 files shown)


bagder:
It looks like the clang (address/memory) sanitizer builds are all messed up now. Strange configure failures because clang acts oddly, and when we get configure to run, we get suspicious problems to build and run things built with this toolset.

It appears as if a recent clang update in the ubuntu image has broken these CI jobs.

bagder:
I posted a plea for help:

https://answers.launchpad.net/ubuntu/+source/llvm-toolchain-14/+question/709571

bagder:
Aha, thanks @IvanNardi for that extra detail. So the issue is not in this specific clang package then.

I found the issue already reported here: https://github.com/actions/runner-images/issues/9491

PR #13133, opened 2024-03-15
curl_sha512_256: add workaround for NetBSD bug

Karlson2k:
Based on @mkauf analysis and suggestion.

The details are [here](https://github.com/curl/curl/pull/13070#issuecomment-1996782015) and [here](https://github.com/curl/curl/pull/13070#issuecomment-1999439358).

The macro `EVP_MAX_MD_SIZE` is not available (on my machine), so I used  `SHA512_256_DIGEST_SIZE * 2` as a size of the temporal array which should be enough.

### DIFF
--- a/lib/curl_sha512_256.c
+++ b/lib/curl_sha512_256.c
@@ -153,7 +153,17 @@ Curl_sha512_256_finish(unsigned char *digest,
   CURLcode ret;
   Curl_sha512_256_ctx *const ctx = (Curl_sha512_256_ctx *)context;
 
+#ifdef __NetBSD__
+  /* Use a larger buffer to work around a bug in NetBSD:
+     https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=58039 */
+  unsigned char tmp_digest[SHA512_256_DIGEST_SIZE * 2];
+  ret = EVP_DigestFinal_ex(*ctx,
+                           tmp_digest, NULL) ? CURLE_OK : CURLE_SSL_CIPHER;
+  if(ret == CURLE_OK)
+    memcpy(digest, tmp_digest, SHA512_256_DIGEST_SIZE);
+#else  /* ! __NetBSD__ */
   ret = EVP_DigestFinal_ex(*ctx, digest, NULL) ? CURLE_OK : CURLE_SSL_CIPHER;
+#endif /* ! __NetBSD__ */
 
   EVP_MD_CTX_destroy(*ctx);
   *ctx = NULL;


Karlson2k:
@mkauf, please check this PR.

Karlson2k:
PR has been updated with fixed code formatting.

lib/curl_sha512_256.c
```diff
@@ -153,7 +153,17 @@ Curl_sha512_256_finish(unsigned char *digest,
...
 
+#ifdef __NetBSD__
+  /* Use a larger buffer to work around a bug in NetBSD:
+     https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=58039 */
+  unsigned char tmp_digest[SHA512_256_DIGEST_SIZE * 2];
+  ret = EVP_DigestFinal_ex(*ctx,
+                           tmp_digest, NULL) ? CURLE_OK : CURLE_SSL_CIPHER;
+  if(ret == CURLE_OK)
+    memcpy(digest, tmp_digest, SHA512_256_DIGEST_SIZE);
+#else  /* ! __NetBSD__ */
```
mkauf, line 164: I think it's more common to repeat the original condition here

```
#else /* __NetBSD__ */
```

lib/curl_sha512_256.c
```diff
@@ -153,7 +153,17 @@ Curl_sha512_256_finish(unsigned char *digest,
...
   ret = EVP_DigestFinal_ex(*ctx, digest, NULL) ? CURLE_OK : CURLE_SSL_CIPHER;
+#endif /* ! __NetBSD__ */
```
mkauf, line 166: also here

mkauf (commented):
Thanks! The code compiles on NetBSD 9.3 and test 1615 works.

If the NetBSD team fixes the bug, how long should curl keep the workaround?

lib/curl_sha512_256.c
Karlson2k: There are two common patterns. This pattern is more popular and already used in this file.
However, the branches are small, so I may just remove comments if you wish.

lib/curl_sha512_256.c
mkauf: OK, you can leave as is

mkauf (approved):

Karlson2k:
> Thanks! The code compiles on NetBSD 9.3 and test 1615 works.
> 
> If the NetBSD team fixes the bug, how long should curl keep the workaround?

If/When NetBSD fixed the bug, we can identify NetBSD version by `__NetBSD_Version__` (defined in <sys/param.h>) and limit the workaround only for earlier versions. If it would be fixed in ported OpenSSL, then fixed version can be detected by `OPENSSL_VERSION_NUMBER`

mkauf:
@Karlson2k : Can you please change the commit message so that it does not mention me? e.g. just remove the "@".

Last time this happened I got a lot of notification mails from GitHub... many projects copy commits and when they push, mentioned users get notifications.

Karlson2k:
> @Karlson2k : Can you please change the commit message so that it does not mention me? e.g. just remove the "@".

Done.
No change in the code, only commit message updated.

mkauf (approved):

dfandrich:
I'm not sure we need to work around this issue. It's a security issue in NetBSD and (presumably), it will be fixed fairly quickly. Even if we keep it, I wouldn't keep it around past the next curl release.

Karlson2k:
Failed tests 661 and 1541 with VS2010 seem to be unrelated.

Karlson2k:
This bug was fixed in NetBSD, but so far hasn't been pushed as update, so existing NetBSD installations still have this bug.
I think it worth to merge this PR. It is not large, not intrusive, limited to NetBSD only and ensures that curl and libcurl will not trigger segmentation failures on NetBSD.

PR #13134, opened 2024-03-15
http, separate response parsing from response action

icing:
- move code that triggers on end-of-response into separate function from parsing
- simplify some headp/headerlen usage
- add `httpversion` to SingleRequest to indicate the version of the current response


bagder (approved):

bagder:
@icing if you could just take a look at the merge conflict...

icing:
> @icing if you could just take a look at the merge conflict...

All resolved.

PR #13135, opened 2024-03-15
 Added curl_multi_waitfds libcurl API function

dkarpov1970:
Added curl_multi_waitfds libcurl API function allowing to get waitfd descriptors from a multi-handle and libtest.

### DIFF
--- a/docs/libcurl/curl_multi_waitfds.md
+++ b/docs/libcurl/curl_multi_waitfds.md
@@ -101,7 +101,7 @@ int main(void)
 
 # AVAILABILITY
 
-Added in 8.7.0
+Added in 8.8.0
 
 # RETURN VALUE
 
--- a/docs/libcurl/curl_multi_waitfds.md
+++ b/docs/libcurl/curl_multi_waitfds.md
@@ -40,8 +40,8 @@ written to.
 
 libcurl fills provided *ufds* array up to the *size*.
 If a number of descriptors used by the multi_handle is greater than the
-*size* parameter then libcurl will fill the *ufds* up to the *size*
-and will return CURLM_OUT_OF_MEMORY error.
+*size* parameter then libcurl fills the *ufds* up to the *size*
+and returns CURLM_OUT_OF_MEMORY error.
 
 If the *fd_count* argument is not a null pointer, it points to a variable
 that on returns specifies the number of descriptors used by the multi_handle to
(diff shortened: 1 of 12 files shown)


docs/libcurl/curl_multi_waitfds.md
```diff
@@ -0,0 +1,109 @@
...
+
+These structures can be used for polling on multi_handle file descriptors in a
+fashion similar to curl_multi_poll(3). The curl_multi_perform(3)
+function should be called as soon as one of them is ready to be read from or
+written to.
+
+libcurl fills provided *ufds* array up to the *size*.
+If a number of descriptors used by the multi_handle is greater than the
+*size* parameter then libcurl fills the *ufds* up to the *size*
+and returns CURLM_OUT_OF_MEMORY error.
```
bagder, line 44: Why? I think it should just be documented to return CURLM_OUT_OF_MEMORY error if the array is too small. I can't see the point in populating the array with just *some* of the data as it will not be useful.

docs/libcurl/curl_multi_waitfds.md
```diff
@@ -0,0 +1,109 @@
...
+SPDX-License-Identifier: curl
+Title: curl_multi_waitfds
+Section: 3
+Source: libcurl
+See-also:
+  - curl_multi_cleanup (3)
+  - curl_multi_init (3)
+  - curl_multi_perform (3)
+  - curl_multi_poll (3)
+  - curl_multi_wait (3)
```
bagder, line 12: I think curl_multi_fdset (3) should be there, as it is very similar to this function. Maybe remove init and cleanup calls? They don't seem very closely related to this.

I also think that you could add "curl_multi_waitfds (3)" as a see-also in the curl_multi_fdset.md page.

tests/libtest/lib2405.c
```diff
@@ -0,0 +1,309 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
+ * Copyright (C) Linus Nielsen Feltzing <linus@haxx.se>
```
bagder, line 8: I think you can add yourself as copyright holder, I don't think Linus wrote this!

docs/libcurl/curl_multi_waitfds.md
dkarpov1970: Yes, I agree. It is really pointless to use not fully filled array of descriptors.

bagder:
Thanks!

PR #13137, opened 2024-03-16
file:// : Allow getting basic directory listings

colinleroy:
Hi Daniel & curl team,

Here's a little PR that allows one to get basic directory listings over file://, like for example:

```
$ ./src/curl file:///home/colin/Documents/git-oss/curl/
winbuild
configure~
autom4te.cache
MacOSX-Framework
LICENSES
CMake
libtool
config.status
libcurl.pc.in
missing
...
```

I hope you folks will consider it useful. For me, it's a way to start diving into curl code to implement the same for smb:// (which, of course, seems to be much more complicated).


gvanem:
Useful IMHO, but not all targets have `readdir()` etc. 
So consider adding MSVC-support for this `<dirent.h>` stuff from e.g.
https://github.com/ThomasDickey/lynx-snapshots/blob/master/lib/dirent.h

into this PR first.

colinleroy:
> Useful IMHO, but not all targets have `readdir()` etc. So consider adding MSVC-support for this `<dirent.h>` stuff from e.g. https://github.com/ThomasDickey/lynx-snapshots/blob/master/lib/dirent.h
> 
> into this PR first.

Thank you for your comment, I'll do that. The linters are being mean to me, I'll handle that too.

dfandrich:
Be sure to take a look at PR#9746 and its comments.

bagder:
> The linters are being mean to me

`make checksrc` locally is a good idea...

lib/file.c
```diff
@@ -400,6 +406,64 @@ static CURLcode file_upload(struct Curl_easy *data)
...
+static CURLcode directory_list(struct FILEPROTO *file, char **data, curl_off_t *size) {
+  CURLcode result = CURLE_OK;
+  DIR *dir;
+  struct dirent *entry;
+  curl_off_t entry_len;
+
+  *data = NULL;
+  *size = 0;
+
+  if(!file) {
```
bagder, line 423: can `file` actually and legitimately be NULL here? Shouldn't this rather be an assert since it should never happen in real life?

lib/file.c
```diff
@@ -400,6 +406,64 @@ static CURLcode file_upload(struct Curl_easy *data)
...
+    result = CURLE_BAD_FUNCTION_ARGUMENT;
+    goto out;
+  }
+
+  dir = fdopendir(file->fd);
+  if(!dir) {
+    result = CURLE_BAD_FUNCTION_ARGUMENT;
```
bagder, line 430: This error is curious, since it already opened the file and it is said to be a dir, but this still fails. It seems to me that `CURLE_BAD_FUNCTION_ARGUMENT` is not the most accurate description of this error. Maybe just `CURLE_READ_ERROR` ?

lib/file.c
```diff
@@ -400,6 +406,64 @@ static CURLcode file_upload(struct Curl_easy *data)
...
+      /* Add one for \0 */
+      tmp = realloc(*data, *size + entry_len + 1);
+      if(!tmp) {
+        result = CURLE_OUT_OF_MEMORY;
+        goto out;
+      }
+
+      *data = tmp;
+      snprintf(*data + *size, entry_len + 1, "%s\n", entry->d_name);
+      *size += entry_len;
```
bagder, line 448: Would it not make sense to just pass on the file entry to the write callback as soon as you have it, rather than to create a (potentially huge) allocated buffer and send them all later?

And if you do prefer buffer approach, please use a dynbuf to avoid manual reallocs.

colinleroy: I went with a temporary buffer because it seemed much easier to calculate expected_size and do resuming. Do you think that's a good enough reason to go the buffer route?

bagder: It does provide an expected size, that is true. But resuming sounds super fragile and will break horribly if something in the directory changes in between...

colinleroy: Hmm this is true. Is it acceptable to reject resumes in case of a directory listing?

bagder: I think that's the only sensible thing to do

colinleroy:
I repushed to address your comments. Is this something like that you had in mind?

lib/file.c
```diff
@@ -651,27 +577,43 @@ static CURLcode file_do(struct Curl_easy *data, bool *done)
...
 
-    result = Curl_client_write(data, CLIENTWRITE_BODY, xfer_buf, nread);
-    if(result)
-      goto out;
+      result = Curl_client_write(data, CLIENTWRITE_BODY, xfer_buf, nread);
+      if(result)
+        goto out;
+    }
+    else {
+      DIR *dir = fdopendir(file->fd);
```
bagder, line 595: You need to add a check for `fdopendir()` to configure and CMake so that we know it exists for the platform you build this. On Windows it won't exist. For now at least, I think we can presume the other POSIX dir funtions exist if this one does.

Then you can either opt to fail on other platforms, or continue to add code support other dir-reading functions. The Windows being a primary target I guess.

macOS also needs attention as the compiler there says:

**'fdopendir' has been marked as being introduced in macOS 10.10 here, but the deployment target is macOS 10.9.0**

colinleroy: Hi @bagder,
Thanks for your feedback. I will add the configure/CMake checks this evening, and probably switch to opendir(file->path) to address the Mac OS X case. 
For Windows, I'll see if I can find someone with Windows experience to help me either add wrappers or #ifdefs.

bagder:
I googled the Windows API for this and it seems fairly simple. Something in this style lists the contents of a given directory:

~~~c
#include <windows.h>
#include <stdio.h>
#include <stdlib.h>

void listdir(const char *dir)
{
    WIN32_FIND_DATA entry;
    HANDLE h;
    char *wc = aprintf("%s\\*.*", dir);
    if(!wc)
      return; /* out of mem */

    h = FindFirstFile(wc, &entry);
    if(h == INVALID_HANDLE_VALUE)
      return;

    do {
      printf("%s\n", entry.cFileName);
    } while(FindNextFile(h, &entry));

    FindClose(h);
    free(wc);
}
~~~

Rough example, but the function calls should show.

colinleroy:
I'm unsure the two failing pipelines are failed related to the PR?

colinleroy:
Concerning Windows, is it OK if I do it in a different PR?

bagder:
> Concerning Windows, is it OK if I do it in a different PR?

Absolutely. As long as this PR does not break their builds or misbehaves for them in other ways.

lib/file.c
```diff
@@ -543,49 +551,84 @@ static CURLcode file_do(struct Curl_easy *data, bool *done)
...
+      if(Curl_pgrsUpdate(data))
+        result = CURLE_ABORTED_BY_CALLBACK;
+      else
+        result = Curl_speedcheck(data, Curl_now());
+      if(result)
+        goto out;
+    }
+  }
+  else {
+    #ifdef HAVE_OPENDIR
```
bagder, line 605: Please move the `#ifdef` (and `#else`, and `#endif`) to column zero.

lib/file.c
```diff
@@ -543,49 +551,84 @@ static CURLcode file_do(struct Curl_easy *data, bool *done)
...
+            break;
+          result = Curl_client_write(data, CLIENTWRITE_BODY, "\n", 1);
+          if(result)
+            break;
+        }
+      }
+      closedir(dir);
+    }
+    #else
+    result = CURLE_READ_ERROR;
```
bagder, line 628: maybe add a failf() here that explains the error?

lib/urldata.h
```diff
@@ -104,7 +104,7 @@ typedef unsigned int curl_prot_t;
 #define PROTO_FAMILY_SSH  (CURLPROTO_SCP|CURLPROTO_SFTP)
 
 #if !defined(CURL_DISABLE_FTP) || defined(USE_SSH) ||   \
-  !defined(CURL_DISABLE_POP3)
+  !defined(CURL_DISABLE_POP3) || !defined(CURL_DISABLE_FILE)
```
bagder, line 107: This enables `CURLOPT_DIRLISTONLY` but acts the same way no matter config. I suppose that is fine, because we can make it do a richer dir listing later. But with this change, I think the documentation for `CURLOPT_DIRLISTONLY` and the corresponding cmdline option (`--list-only`) need to now mention FILE as well

bagder (commented):
FYI: I deem this to be a "feature/change", which means that we will not merge this PR until the feature window opens again. (Normally 10 days after a release.)

colinleroy:
> FYI: I deem this to be a "feature/change", which means that we will not merge this PR until the feature window opens again. (Normally 10 days after a release.)

That seems logical :)
Your latest comments should be addressed in the last push.

colinleroy:
Hi @bagder! I think the feature freeze is over?
do I have to do something about the failed Mac pipelines? It seems like they're unrelated to the PR?
Thanks 🙂

bagder:
The freeze is over. Those CI fails should be fixed if you rebase and force-push.

I do miss a test case specifically for this feature.

colinleroy:
Hi Daniel,
PR rebased, test added (i hope it's what you expected?) and CI passed!

bagder (approved):
:+1:

bagder:
Thanks!

Issue #13139, opened 2024-03-17
Random SIGSEGV when Ares support is enabled


zinal:
### I did this

Random (but reproducible) crashes with SIGSEGV in libcurl requests when libcurl is built with Ares support.
The cause seems to be the missing initialization of `ares_options` struct in `Curl_resolver_init()`.
Posting this issue to link with the fix.

Additional details here: https://github.com/ydb-platform/ydb/issues/2856

### I expected the following

No SIGSEGV, and program working

### curl/libcurl version

libcurl 8.5.0

### operating system

Ubuntu 20.04.3 LTS
Linux ydb-gpb-test-017.ydb.yandex.net 5.4.161-26.3 #1 SMP Mon Feb 7 14:47:58 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

zinal:
Added to upstream, too: https://github.com/curl/curl/pull/13140

zinal:
Seems that the issue has been caused by multiple Conan versions used in the build. Closing as not needed.

PR #13140, opened 2024-03-17
#13139 add initialization of ares_options to Curl_resolver_init()

zinal:
See https://github.com/curl/curl/issues/13139


zinal:
Seems that the issue has been caused by multiple Conan versions used in the build. Closing as not needed.

PR #13150, opened 2024-03-19
multi_wait improvements

icing:
In `multi_wait()`, which is called all the time, `multi_getsock()` was called twice. 

The first invocation was used to count the number of sockets, a sufficiently large `struct pollfd` array was allocated, and then the second invocation was used to populate that array. This is not only an overhead, but it assumes that the second call will deliver the same number of sockets as the first. Which is not really guaranteed. Also, when parallel transfers on the same connection are active, the same socket is collected for every transfer.

- only call `multi_getsock()` once for all transfers. Grow the `struct pollfd` array when needed.
- fold repeated sockets. This is done cheaply by just comparing the current socket to the last one. It will not catch all duplicates on multiple connections.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1289,8 +1289,8 @@ static void reset_socket_fdwrite(curl_socket_t s)
 }
 #endif
 
-static CURLMcode ufds_resize(struct pollfd **pfds, unsigned int *pfds_len,
-                             unsigned int inc, bool *is_malloced)
+static CURLMcode ufds_increase(struct pollfd **pfds, unsigned int *pfds_len,
+                               unsigned int inc, bool *is_malloced)
 {
   struct pollfd *new_fds, *old_fds = *pfds;
   unsigned int new_len = *pfds_len + inc;
@@ -1388,7 +1388,7 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
         }
         else {
           if(nfds >= ufds_len) {
-            if(ufds_resize(&ufds, &ufds_len, 100, &ufds_malloc))
+            if(ufds_increase(&ufds, &ufds_len, 100, &ufds_malloc))
               return CURLM_OUT_OF_MEMORY;
           }
           DEBUGASSERT(nfds < ufds_len);
@@ -1430,7 +1430,7 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
     }
 #endif
     if(nfds >= ufds_len) {
-      if(ufds_resize(&ufds, &ufds_len, 100, &ufds_malloc))
+      if(ufds_increase(&ufds, &ufds_len, 100, &ufds_malloc))
         return CURLM_OUT_OF_MEMORY;
     }
     DEBUGASSERT(nfds < ufds_len);
@@ -1449,7 +1449,7 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
 #ifndef USE_WINSOCK
   if(use_wakeup && multi->wakeup_pair[0] != CURL_SOCKET_BAD) {
     if(nfds >= ufds_len) {
-      if(ufds_resize(&ufds, &ufds_len, 100, &ufds_malloc))
+      if(ufds_increase(&ufds, &ufds_len, 100, &ufds_malloc))
         return CURLM_OUT_OF_MEMORY;
     }
     DEBUGASSERT(nfds < ufds_len);
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -1289,6 +1289,29 @@ static void reset_socket_fdwrite(curl_socket_t s)
 }
 #endif
 
+static CURLMcode ufds_resize(struct pollfd **pfds, unsigned int *pfds_len,
+                             unsigned int inc, bool *is_malloced)
+{
+  struct pollfd *new_fds, *old_fds = *pfds;
+  unsigned int new_len = *pfds_len + inc;
+
+  new_fds = calloc(new_len, sizeof(struct pollfd));
+  if(!new_fds) {
+    if(*is_malloced)
+      free(old_fds);
+    *pfds = NULL;
+    *pfds_len = 0;
+    return CURLM_OUT_OF_MEMORY;
+  }
+  memcpy(new_fds, old_fds, (*pfds_len) * sizeof(struct pollfd));
+  if(*is_malloced)
+    free(old_fds);
+  *pfds = new_fds;
+  *pfds_len = new_len;
+  *is_malloced = TRUE;
+  return CURLM_OK;
+}
+
 #define NUM_POLLS_ON_STACK 10
 
 static CURLMcode multi_wait(struct Curl_multi *multi,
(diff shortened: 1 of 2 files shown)


icing:
@jay: I'd like your review on this because it touches on the Window parts which I tried to preserve to the best of my understanding.

bagder (approved):

lib/multi.c
```diff
@@ -1357,6 +1364,11 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
       return CURLM_INTERNAL_ERROR;
     }
 #endif
+    if(nfds >= ufds_len) {
+      if(ufds_resize(&ufds, &ufds_len, 100, &ufds_malloc))
+        return CURLM_OUT_OF_MEMORY;
+    }
+    DEBUGASSERT(nfds < ufds_len);
```
jay, line 1371: if the size could be larger then why not just use it instead
```suggestion
    if(nfds >= ufds_len) {
      size_t needed = nfds - ufds_len;
      if(ufds_resize(&ufds, &ufds_len, needed, &ufds_malloc))
        return CURLM_OUT_OF_MEMORY;
    }
```

jay: also I would call it ufds_increase instead of resize because you're passing a number to increase, either that or leave it named resize and don't pass the number to increase, pass the new size instead

ufds_resize(&ufds, &ufds_len, nfds +1, &ufds_malloc)

lib/multi.c
```diff
@@ -1261,84 +1284,68 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
...
+          }
+          DEBUGASSERT(nfds < ufds_len);
+          ufds[nfds].fd = ps.sockets[i];
+          ufds[nfds].events = events;
+          ++nfds;
+        }
+      }
 #ifdef USE_WINSOCK
+      if(mask) {
         if(WSAEventSelect(ps.sockets[i], multi->wsa_event, mask) != 0) {
```
jay, line 1337: @mback2k is WSAEventSelect supposed to be called even when mask is 0?

icing: I think `mask == 0` is used to remove a socket again. See line 1475 after the poll.

The check I added is for the case where a socket is returned, but neither POLLIN, nor POLLOUT is set. Should not happen, but needs to be handled. Calling `WSAEventSelect()` with 0 would then remove a socket and that is not correct in that situation. The socket might have been added before by another transfer.

lib/multi.c
icing: Will  change the name as suggested.

lib/multi.c
icing: Thinking about this, I am not even sure this works correctly as is for parallel transfers. Think about this case: transfers A and B on the same connection socket X.

- A wants to POLLIN on X
- B wants to POLLOUT on X

Does calling `WSAEventSelect()` on X not override the previous mask?

jay: i recall there was some issue with wsa events, where marc had added it to purposely reset the flags. 

> Does calling `WSAEventSelect()` on X not override the previous mask?

it does override so if you want to wait on read and write you have to do it together. i hope marc will have time to review this pr, he has been busy so give it a few days

fwiw i used a build of this pr yesterday with many parallel http/2 httpbin drip replies and didn't see any difference in behavior

mback2k:
Sorry, I won't be able to look at this before the weekend.

lib/multi.c
```diff
@@ -1261,84 +1284,68 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
...
-    ufds_malloc = TRUE;
-  }
   nfds = 0;
+  memset(ufds, 0, ufds_len * sizeof(struct pollfd));
+  memset(&ps, 0, sizeof(ps));
 
-  /* only do the second loop if we found descriptors in the first stage run
-     above */
-
-  if(curlfds) {
```
mback2k, line 1306: I don't really understand why the counting of fds and this check was removed in the first place. Can you please elaborate a little bit on this?

lib/multi.c
mback2k: @jay you are right regarding the resetting of flags needing to happen, but this comes later in the existing code:
https://github.com/curl/curl/blob/a41cd15bec379fcea3b2a98e98c322061304b3fd/lib/multi.c#L1428
That is why a little bit above that line I left this comment:
https://github.com/curl/curl/blob/a41cd15bec379fcea3b2a98e98c322061304b3fd/lib/multi.c#L1405-L1407

> Does calling `WSAEventSelect()` on X not override the previous mask?

Yes, it will overwrite the previous mask for the same `multi->wsa_event`. Anyway, I think your new check referenced by this review thread is fine, as `mask == 0` should not happen at this point.

mback2k: > Thinking about this, I am not even sure this works correctly as is for parallel transfers.

I am not into the multi-handle details here, would each transfer have its own multi-handle and therefore own `multi->wsa_event`? If not, you are probably right and the WSA event will need to be changed to be per-transfer.

lib/multi.c
```diff
@@ -1261,84 +1284,68 @@ static CURLMcode multi_wait(struct Curl_multi *multi,
...
             free(ufds);
           return CURLM_INTERNAL_ERROR;
         }
-#endif
       }
+#endif
     }
   }
 
+  curl_nfds = nfds; /* what curl internally used in ufds */
```
mback2k, line 1347: Sidenote: I guess the whole naming of all those nfd variables could be improved. Working on this last time and reviewing this now is really not straightforward due to those unspecific and generic variables names all over the place.

lib/multi.c
icing: Asking transfers for their pollsets twice risks the  race that the sets may be different on the second call. For example in QUIC, some actions are timer dependent with high resolution. The previous implementation would get messed up, potential writing out of bounds.

lib/multi.c
icing: I hear you. My urge to rename things is countered by the advantages of not increasing changes sizes "too much". There is often no clear line.

lib/multi.c
icing: There can be many transfers on the same multihandle and many transfers on the same connection (HTTP/2, HTTP/3). These report the connection socket with *their individual needs*.

When transfer A and B are on the same connection, A might request POLLIN and B POLLOUT. Without coalescing the two, the `mask`s will overwrite each other.

mback2k: The main question is if multi_wait will be executed in parallel for these transfers on one multi handle. Also, I am no sure what will happen if the WSA event is changed while already being waited on in another thread for example. Maybe that is undefined behavior or not supported after all?

bagder: > if multi_wait will be executed in parallel for these transfers on one multi handle

It cannot be executed in parallel for the same handle, no.

bagder:
@mback2k @jay what's your take on this now? It would be neat to merge, but I don't sense that we have your :+1: on it just yet?

calvin2021y:
this weekend I try test this patch, find it cannot merge into master.

mback2k (approved):

PR #13151, opened 2024-03-19
http2, http3: partial file error handling

icing:
Only return CURLE_PARTIAL_FILE when bytes have been received.

- should resolve spurious pytest failures when stream were reset right after response header were received


bagder (approved):

bagder:
Closed by 98f67a6145b4c (but typo in the commit message)

PR #13163, opened 2024-03-21
docs: make each libcurl man specify protocol(s)

bagder:
The mandatory header now has a mandatory list of protocols for which the manpage is relevant.

Most man pages already has a "PROTOCOLS" section, but this introduces a stricter way to specify the relevant protocols.

cd2nroff verifies that at least one protocol is mentioned (which can be `*`).

This information is not used just yet, but A) the PROTOCOLS section can now instead get generated and get a unified wording across all manpages and B) this allows us to more reliably filter/search for protocol specific manpages/options.


bagder:
Closed via b935fd4a07aa9dc0

PR #13164, opened 2024-03-21
wolfSSL: do not call the stub function wolfSSL_BIO_set_init()

fabiankeil:
Calling the function isn't necessary and causes the build to fail when wolfSSL has been compiled with NO_WOLFSSL_STUB:

     Making all in opts
       CCLD     curl
     ld: error: undefined symbol: wolfSSL_BIO_set_init
     >>> referenced by wolfssl.c:235 (vtls/wolfssl.c:235)
     >>>               libcurl_la-wolfssl.o:(wolfssl_bio_cf_create) in archive ../lib/.libs/libcurl.a
     cc: error: linker command failed with exit code 1 (use -v to see invocation)
     *** Error code 1

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -232,7 +232,6 @@ static const struct group_name_map gnm[] = {
 static int wolfssl_bio_cf_create(WOLFSSL_BIO *bio)
 {
   wolfSSL_BIO_set_shutdown(bio, 1);
-  wolfSSL_BIO_set_init(bio, 1);
   wolfSSL_BIO_set_data(bio, NULL);
   return 1;
 }


lib/vtls/wolfssl.c
```diff
@@ -232,7 +232,6 @@ static const struct group_name_map gnm[] = {
 static int wolfssl_bio_cf_create(WOLFSSL_BIO *bio)
 {
   wolfSSL_BIO_set_shutdown(bio, 1);
-  wolfSSL_BIO_set_init(bio, 1);
```
jay, line 235: Does wolfSSL need any BIO initialization then? This code was probably copied from [ossl_bio_cf_create](https://github.com/curl/curl/blob/curl-8_6_0/lib/vtls/openssl.c#L673-L682). I can't find wolfSSL_BIO_set_init documented anywhere in wolfSSL. Nor shutdown for that matter.

@icing since the openssl and wolfssl create function is only used on bio creation why is there a shutdown call?

bagder (approved):

lib/vtls/wolfssl.c
bagder: wolfSSL_BIO_set_shutdown() at least does something - it stores that it has been shutdown...

bagder:
Thanks!

PR #13165, opened 2024-03-21
transfer, function for writing single response header

icing:
Add `Curl_xfer_write_resp_hd()` and similar method in protocol handlers to allow for processing a single, 0-terminated response header. Avoids parsing and copying these lines.

http2, ngtcp2: use Curl_xfer_write_resp_hd()
- use filter scratch buffer to compile complete header line for writing


bagder:
oops, conflicting now

bagder (approved):

PR #13168, opened 2024-03-21
docs/libcurl: add TLS backend info for all TLS options

bagder:
All man pages that are listed to be for TLS now must also specify exactly what TLS backends the option works for, or use `*` if they all work.

cd2nroff makes sure this is done and that the listed backends(s) exist.

### DIFF
--- a/docs/libcurl/opts/CURLINFO_CAINFO.md
+++ b/docs/libcurl/opts/CURLINFO_CAINFO.md
@@ -11,7 +11,7 @@ See-also:
 Protocol:
   - TLS
 TLS-backend:
-  - *
+  - All
 ---
 
 # NAME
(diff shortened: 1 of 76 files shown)


docs/libcurl/opts/CURLINFO_CAINFO.md
```diff
@@ -10,6 +10,8 @@ See-also:
   - curl_easy_setopt (3)
 Protocol:
   - TLS
+TLS-backend:
+  - *
```
jay, line 14: is it your intention for it to render as `*` because I think `All` would be easier to understand

bagder: I primarily think of machine parsing this, but using `All` is indeed a good idea for the human readers.

bagder:
Using this info, I'm working on generating a support matrix for the website. Here's a current screenshot of what we can now use this info for.
Screenshot 2024-03-22 at 09-33-03 TLS backend support of libcurl options

PR #13171, opened 2024-03-22
curl: make --help adapt to the terminal width

bagder:
Instead of assuming and working with 80 colums, try figuring out what width is actually used.

Ref: #13141

### DIFF
--- a/src/tool_cb_prg.c
+++ b/src/tool_cb_prg.c
@@ -229,28 +229,25 @@ int tool_progress_cb(void *clientp,
   return 0;
 }
 
-void progressbarinit(struct ProgressData *bar,
-                     struct OperationConfig *config)
+/*
+ * get_terminal_columns() returns the number of columns in the current
+ * terminal. It will return 79 on failure. Also, the number can be very big.
+ */
+
+unsigned int get_terminal_columns(void)
 {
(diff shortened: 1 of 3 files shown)


bagder:
@jay can you double-check how this math works for Windows. I suspect we might have to tweak it there.

jay:
looks ok, here's an excerpt

~~~
 -Z, --parallel                    Perform transfers in parallel
     --parallel-immediate        Do not wait for multiplexing (with --parallel)
     --parallel-max <num>          Maximum concurrency for parallel transfers
     --pass <phrase>               Pass phrase for the private key
     --path-as-is                  Do not squash .. sequences in URL path
     --pinnedpubkey <hashes>      FILE/HASHES Public key to verify peer against
     --post301                     Do not switch to GET after a 301 redirect
     --post302                     Do not switch to GET after a 302 redirect
     --post303                     Do not switch to GET after a 303 redirect
     --preproxy [protocol://]host[:port]  Use this proxy first
 -#, --progress-bar                Display transfer progress as a bar
     --proto <protocols>           Enable/disable PROTOCOLS
     --proto-default <protocol>    Use PROTOCOL for any URL missing a scheme
~~~

PR #13188, opened 2024-03-25
urldata: remove values that are not used depending on compilation settings

MAntoniak:
Reduced size of dynamically_allocated_data structure.

Reduced number of stored values in enum dupstring and enum dupblob. This affects the reduced array placed in the UserDefined structure.

It is also possible to remove the STRING_SASL_AUTHZID and STRING_BEARER values from the dupstring. However, this requires a separate PR.

### DIFF
--- a/lib/urldata.h
+++ b/lib/urldata.h
@@ -1327,22 +1327,28 @@ struct UrlState {
   /* Dynamically allocated strings, MUST be freed before this struct is
      killed. */
   struct dynamically_allocated_data {
-    char *proxyuserpwd;
     char *uagent;
     char *accept_encoding;
     char *userpwd;
     char *rangeline;
     char *ref;
     char *host;
+#ifndef CURL_DISABLE_COOKIES
     char *cookiehost;
+#endif
+#ifndef CURL_DISABLE_RTSP
     char *rtsp_transport;
+#endif
     char *te; /* TE: request header */
 
     /* transfer credentials */
     char *user;
     char *passwd;
+#ifndef CURL_DISABLE_PROXY
+    char *proxyuserpwd;
     char *proxyuser;
     char *proxypasswd;
+#endif
   } aptr;
 
   unsigned char httpwant; /* when non-zero, a specific HTTP version requested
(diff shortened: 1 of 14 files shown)


lib/urldata.h
```diff
@@ -1417,85 +1423,123 @@ struct Curl_multi;    /* declared in multihandle.c */
...
+  STRING_KEY_TYPE,        /* format for private key (default: PEM) */
+  STRING_SSL_CAPATH,      /* CA directory name (doesn't work on windows) */
+  STRING_SSL_CAFILE,      /* certificate file to verify peer against */
+  STRING_SSL_PINNEDPUBLICKEY, /* public key file to verify peer against */
+  STRING_SSL_CIPHER_LIST, /* list of ciphers to use */
+  STRING_SSL_CIPHER13_LIST, /* list of TLS 1.3 ciphers to use */
+  STRING_SSL_CRLFILE,     /* crl file to check certificate */
+  STRING_SSL_ISSUERCERT, /* issuer cert file to check certificate */
+  STRING_SERVICE_NAME,    /* Service name */
+#ifndef CURL_DISABLE_PROXY
```
bagder, line 1438: This is in direct contradiction to the comment above the enum which says:

**This enumeration MUST not use conditional directives (#ifdefs)**

But: the comment also refers to a (misspelled) source code file that was removed in 821d108fc9adfb, which makes me suspect maybe this comment does not apply anymore?

@monnerat can you fill in some blanks maybe?

monnerat: It seems the comment is obsolete: the current test is now based on `CURLOPT_*`, not on `STRING_*` anymore.

bagder: @MAntoniak can you perhaps update that comment to reflect reality?

MAntoniak: If I understand correctly just remove it?

bagder: Yes!

bagder (approved):

bagder:
Thanks!

PR #13193, opened 2024-03-26
remember protocol handler at easy requests

icing:
- remember protocol handler at `data->req` so that members in `data->req.p.*` can be accessed more safely
- used in connection filters to not access information for "foreign" handlers


icing:
Superseeded by #13204.

PR #13196, opened 2024-03-27
bump version to match last tag

eworm-de:


bagder:
Handled differently in 97c417ae283dcf

PR #13197, opened 2024-03-27
cmake: generate misc manpages and install `mk-ca-bundle.pl`

### LINKED ISSUE #13194: Man page tests 1140 and 1173 fail on cmake
These two tests when run via cmake fail. We have a few CI builds that run a failing configuration (on Appveyor), except that they explicitly disable these two tests so we don't see the failures. In tree and out-of-tree doesn't seem to matter. Old cmake or new cmake doesn't seem to matter. It shows for me with a plain-Jane cmake build:

`cmake . && make -j9 && make -j9 test-ci TFLAGS='1140 1173'`

(it fails without the TFLAGS, too; this is just quicker). It also fails on an Ubuntu 22.04 container with the minimum packages needed to build installed.

I'm completely confused as to why nobody else has reported this because my testing shows it happens in any cmake build. Does absolutely nobody using cmake run the test suite? Or do they all (including us, it seems) just disable those tests and go on with their day?

Here's an example failure log: https://github.com/curl/curl/actions/runs/8440831621/job/23118626767?pr=13029

_Originally posted by @dfandrich in https://github.com/curl/curl/issues/13028#issuecomment-1984253530_

vszakats:
- install `mk-ca-bundle.pl` like autotools does.

- generate and install `mk-ca-bundle.1` and `curl-config.1` like autotools. This fixes tests 1140 and 1173.

  Fixes #13194

- add option `BUILD_MISC_DOCS` to control building the above two manpages. Enabled by default.

- appveyor: stop disabling tests 1140 and 1173.

Closes #13197

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ endif()
 find_package(Perl)
 
 option(BUILD_LIBCURL_DOCS "to build libcurl man pages" ON)
+option(BUILD_MISC_DOCS "to build misc man pages (e.g. curl-config and mk-ca-bundle)" ON)
 option(ENABLE_CURL_MANUAL "to build the man page for curl and enable its -M/--manual option" ON)
 
 if(ENABLE_CURL_MANUAL OR BUILD_LIBCURL_DOCS)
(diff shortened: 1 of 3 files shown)


dfandrich:
The tests pass fine for me with this!

bagder (approved):

PR #13198, opened 2024-03-27
cmdline-docs: fix make install with configure --disable-docs

arachsys:
`make -C docs/cmdline-opts install` depends on `all-am`, which in turn depends on `$(MANS)`, unconditionally defined to be `$(man_MANS)`.

As with `CLEANFILES`, only add `curl.1` to `man_MANS` when `BUILD_DOCS` is true so we don't try to build `curl.1` unnecessarily.


bagder:
thanks!

Issue #13200, opened 2024-03-27
Regression in rustls pkg-config detection


kpcyrd:
### I did this

I tried to compile the latest curl release with the rustls TLS backend.

This is the error I get during configure:

```
[...]
checking for mit-krb5-gssapi options with pkg-config... found
checking for gss.h... no
checking for gssapi/gssapi.h... yes
checking for gssapi/gssapi_generic.h... yes
checking for gssapi/gssapi_krb5.h... yes
checking if GSS-API headers declare GSS_C_NT_HOSTBASED_SERVICE... yes
checking for pkg-config... (cached) /usr/bin/pkg-config
checking for mit-krb5-gssapi options with pkg-config... found
checking if we can link against GSS-API library... yes
checking whether to enable Windows native SSL/TLS... no
checking whether to enable Secure Transport... no
checking whether to enable Amiga native SSL/TLS (AmiSSL v5)... no
checking for pkg-config... (cached) /usr/bin/pkg-config
checking for rustls options with pkg-config... found
configure: pkg-config: SSL_LIBS: "-lrustls"
configure: pkg-config: SSL_LDFLAGS: ""
configure: pkg-config: SSL_CPPFLAGS: ""
configure: detected rustls
configure: error: TLS not detected, you will not be able to use HTTPS, FTPS, NTLM and more.
Use --with-openssl, --with-gnutls, --with-wolfssl, --with-mbedtls, --with-schannel, --with-secure-transport, --with-amissl, --with-bearssl or --with-rustls to address this.
```

pkg-config is correctly setup for librustls:

```
% pacman -Ql librustls
librustls /usr/
librustls /usr/include/
librustls /usr/include/rustls.h
librustls /usr/lib/
librustls /usr/lib/librustls.so
librustls /usr/lib/librustls.so.0.12.1
librustls /usr/lib/pkgconfig/
librustls /usr/lib/pkgconfig/rustls.pc
librustls /usr/share/
librustls /usr/share/licenses/
librustls /usr/share/licenses/librustls/
librustls /usr/share/licenses/librustls/LICENSE-APACHE
librustls /usr/share/licenses/librustls/LICENSE-ISC
librustls /usr/share/licenses/librustls/LICENSE-MIT
```

The configure options used are:

```
--prefix='/usr'
--mandir='/usr/share/man'
--disable-ldap
--disable-ldaps
--disable-manual
--enable-ipv6
--enable-threaded-resolver
--with-gssapi
--with-libssh2
[...]

PR #13202, opened 2024-03-27
m4: fix rustls pkg-config codepath

Kangie:
The previous pkg-config code would successfully detect rustls but did not set all appropriate variables and call the right macros to properly configure cURL.

Closes: #13200

### DIFF
--- a/m4/curl-rustls.m4
+++ b/m4/curl-rustls.m4
@@ -100,9 +100,6 @@ if test "x$OPT_RUSTLS" != xno; then
           AC_MSG_ERROR([--with-rustls was specified but could not find rustls.]),
           -lpthread -ldl -lm)
 
-        USE_RUSTLS="yes"
-        ssl_msg="rustls"
-
         LIB_RUSTLS="$PREFIX_RUSTLS/lib$libsuff"
         if test "$PREFIX_RUSTLS" != "/usr" ; then
           SSL_LDFLAGS="-L$LIB_RUSTLS"
@@ -142,6 +139,10 @@ if test "x$OPT_RUSTLS" != xno; then
       LIBS="$SSL_LIBS $LIBS"
       USE_RUSTLS="yes"
       ssl_msg="rustls"
+      AC_DEFINE(USE_RUSTLS, 1, [if rustls is enabled])
+      AC_SUBST(USE_RUSTLS, [1])
+      RUSTLS_ENABLED=1
+      test rustls != "$DEFAULT_SSL_BACKEND" || VALID_DEFAULT_SSL_BACKEND=yes
     else
       AC_MSG_ERROR([pkg-config: Could not find rustls])
     fi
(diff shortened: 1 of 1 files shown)


cpu:
I've been able to reproduce the error from https://github.com/curl/curl/issues/13200 on my system trying to build curl with rustls 0.12.1+ using `pkg-config`. 

This patch seems to get further along, but unfortunately it's still failing (_or I'm holding something wrong :laughing:_). Prior to this branch's update I saw the error reported in #13200 ("configure: error: TLS not detected ..."), now it's:

>  one or more libs available at link-time are not available run-time

```
# In the rustls-ffi checkout
$ git rev-parse HEAD
47d7b0ff1ae214b596648914c01d2be6bd88b9d5

$ cargo cinstall --release --prefix=/tmp/librustls-0.12.1/
<snipped>
  Populating uninstalled header directory
  Installing pkg-config file
  Installing header file
  Installing static library
  Installing shared library
```

```
# In the curl checkout
$ git rev-parse HEAD
4371a199767abd1579fa5ef4b20e75f395ab76e8

$ autoreconf -fi
<snipped>
$ ./configure --with-rustls=/tmp/librustls-0.12.1
<snipped>
checking for sys/time.h... (cached) yes
checking for sys/socket.h... (cached) yes
checking for struct timeval... yes
checking run-time libs availability... failed
configure: error: one or more libs available at link-time are not available run-time. Libs used at link-time: -lrustls -lz
```

Earlier in the `configure` output I do see `rustls` being found via pkg-config:

```
configure: PKG_CONFIG_LIBDIR will be set to "/tmp/librustls-0.12.1/lib/pkgconfig"
checking for rustls options with pkg-config... found
configure: pkg-config: SSL_LIBS: "-lrustls"
configure: pkg-config: SSL_LDFLAGS: "-L/tmp/librustls-0.12.1/lib"
configure: pkg-config: SSL_CPPFLAGS: "-I/tmp/librustls-0.12.1/include"
configure: built with one SSL backend
```

If I use a static build of rustls without pkg-config support (e.g. the legacy `Makefile` approach described in `docs/RUSTLS.md`), the configure succeeds for both rustls 0.12.1 and 0.12.2. It seems like it's the pkg-config setup that's not quite 100% yet.

cpu:
> or I'm holding something wrong

Hmmm looking at the upstream Nixpkgs derivation for `curl` I'm starting to suspect this to be the case. I'll investigate further but for now consider my report of the fix not working with a grain of salt.

cpu:
>> or I'm holding something wrong
>
>Hmmm looking at the upstream Nixpkgs derivation for curl I'm starting to suspect this to be the case.

Yup, 100% the case :facepalm: I needed to set `LD_LIBRARY_PATH` correctly. Sorry about the noise!

I can confirm this fix worked for me and allowed building with pkg-config and librustls. Thanks Kangie!

bagder:
Thanks everyone, will merge!

PR #13204, opened 2024-03-27
http/2, http/3, decouple stream state from easy handle

icing:
- add convenience functions to `hash.h` for using a hash with `curl_off_t` as keys. Use this in h2+h3 connection filters to associate `data->id` with the internal stream state.
- changed implementations of all affected connection filters
- removed `h2_ctx*` and `h3_ctx*` from `struct HTTP` and thus the easy handle
- solves the problem of attaching "foreign protocol" easy handles during connection shutdown


bagder:
This introduces a second hash table implementation. Have you ruled out instead using the existing lib/hash.c functions for this? It seems to me like they could be used...

icing:
Separate hash implementation removed again. Convenience functions added.

bagder:
merge conflicts now after some other merges

icing:
`curl-for-win` fails with:

```
++ gpg --batch --keyserver-options timeout=15 --display-charset utf-8 --keyid-format 0xlong --verify-options show-primary-uid-only --verify pkg.sig pkg.bin
gpg: can't open signed data 'pkg.bin'
gpg: can't hash datafile: No such file or directory
```

Anyone has an idea what is going on here? @vszakats ?

vszakats:
Package or signature downloads from GitHub failed:
```
++ curl --disable --user-agent '' --fail --silent --show-error --globoff --remote-time --xattr --connect-timeout 15 --max-time 60 --retry 3 --max-redirs 10 --location --proto-redir =https --output pkg.bin https://github.com/rockdaboot/libpsl/releases/download/0.21.5/libpsl-0.21.5.tar.gz --output pkg.sig https://github.com/rockdaboot/libpsl/releases/download/0.21.5/libpsl-0.21.5.tar.gz.sig
curl: (22) The requested URL returned error: 500
curl: (22) The requested URL returned error: 500
curl: (22) The requested URL returned error: 500
curl: (22) The requested URL returned error: 500
```
A retry should fix this.

bagder (approved):

PR #13207, opened 2024-03-27
cmake: speed up libcurl doc builds again

vszakats:
This time limit the number of files per command to avoid exceeding limitations of certain OS/shell envs.

Such known env is Windows with the `cmd.exe` shell, which features an 8K command-line length limit to this day.

Allowlisting `UNIX` to have no limit and using a limit of 200 for all other envs to be safe. If there is a way to detect `cmd.exe` and/or we know which precise envs are sensitive to this, we can tweak these conditions further.

Even with the low limit, this patch reduces external commands by 200x, making full and CI builds much faster.

Ref: #12762 2620aa930bc73af1e4c70b10e3125b957b96ecfb (initial)
Ref: #13047 f03c85635f35269f1f45b983bf216624f541760a (revert)

Closes #13207

### DIFF
--- a/docs/libcurl/CMakeLists.txt
+++ b/docs/libcurl/CMakeLists.txt
@@ -26,21 +26,39 @@ transform_makefile_inc("Makefile.inc" "${CMAKE_CURRENT_BINARY_DIR}/Makefile.inc.
 include("${CMAKE_CURRENT_BINARY_DIR}/Makefile.inc.cmake")
 
 function(add_manual_pages _listname)
-  foreach(_file IN LISTS ${_listname})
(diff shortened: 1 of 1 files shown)


bagder:
Does this still work correctly if you update a single `.md` file and run make again?

vszakats:
It does, yes. Confirmed with a macOS and a Windows cross-build. In the latter, it rebuilds a batch of 200 files in which the modified one is present. It's instantaneous, even on an ancient machine.

bagder (approved):

jay:
I tried this change out in a Windows 10 VM by building with cmake mingw makefiles and the difference compared to the parent commit is significant. It cut build time by 50 seconds.

> Does this still work correctly if you update a single `.md` file and run make again?

yes

PR #13208, opened 2024-03-27
Fix detection of OpenSSL 1.1.1 or later in curl_sha512_256.c

pghmcfc:
Use the same OPENSSL_VERSION_NUMBER comparison as in lib/vtls/openssl.c.

As it stands, builds using an OpenSSL 1.1.0 release fail due to trying to use the unavailable symbol EVP_sha512_256.


pghmcfc:
The version number would be correct as per the current numbering documentation. However, this changed with OpenSSL 3.x - the documentation for OpenSSL 1.x was slightly different: https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_VERSION_NUMBER.html

bagder (approved):

bagder:
Thanks!

PR #13213, opened 2024-03-27
libssh2: set length to 0 if strdup failed

stoeckmann:
Internally, libssh2 would dereference the NULL pointer if length is non-zero. The callback function cannot return the error condition, so at least prevent subsequent crash.


bagder (approved):

bagder:
Thanks!

PR #13215, opened 2024-03-28
tool_getparam: output warning for leading unicode quote character

### LINKED ISSUE #13214: curl reports "fault filter abortcurl: (6) Could not resolve host: store.example.xn--com-9o0a"
### I did this

From a Google doc I copy-pasted a curl command that was:

```
curl -i 10.1.2.9 -H “host: store.example.com”
```

Note that what appear as double-quotes around `host: store.example.com` are not the normal double-quotes used in computing but used in Document editing programs like GoogleDoc or Word. But this trips up curl and it gives me the following output:

```
HTTP/1.1 404 Not Found
content-length: 18
content-type: text/plain
date: Thu, 28 Mar 2024 05:04:59 GMT
via: 1.1 google

fault filter abortcurl: (6) Could not resolve host: store.example.xn--com-9o0a
```

I spent considerable amount of time trying to troubleshoot this since the error message was very misleading

### I expected the following

curl should probably recognize the unsupported characters on the command line and generate a message while parsing those arguments instead of generating an error deep inside (which is what "fault filter abort" looks like)

### curl/libcurl version

curl 7.88.1 (x86_64-pc-linux-gnu) libcurl/7.88.1 OpenSSL/3.0.11 zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.3 libpsl/0.21.2 (+libidn2/2.3.3) libssh2/1.10.0 nghttp2/1.52.0 librtmp/2.3 OpenLDAP/2.5.13
Release-Date: 2023-02-20, security patched: 7.88.1-10+deb12u5

### operating system

Linux xxxxx 6.1.0-18-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01) x86_64 GNU/Linux

bagder:
... in the option argument.

Typically this is a mistake done when copying example command lines from online documentation using the wrong quote character.

Presumably there are also other potential quote characters that might be used, and this check is done without even knowing that unicode is used!

Reported-by: Sanjay Pujare
Fixes #13214

### DIFF
--- a/src/tool_getparam.c
+++ b/src/tool_getparam.c
@@ -1337,6 +1337,11 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
         warnf(global, "The file name argument '%s' looks like a flag.",
               nextarg);
       }
+      else if(!strncmp("\xe2\x80\x9c", nextarg, 3)) {
+        warnf(global, "The argument '%s' starts with a unicode quote where "
+              "maybe an ASCII \" was intended?",
+              nextarg);
+      }
     }
     else if((a->desc == ARG_NONE) && !toggle) {
       err = PARAM_NO_PREFIX;
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -73,7 +73,7 @@ test426 test427 test428 test429 test430 test431 test432 test433 test434 \
 test435 test436 test437 test438 test439 test440 test441 test442 test443 \
 test444 test445 test446 test447 test448 test449 test450 test451 test452 \
 test453 test454 test455 test456 test457 test458 test459 test460 test461 \
-test462 test463 test467 test468 test469 \
+test462 test463 test467 test468 test469 test470 \
 \
 test490 test491 test492 test493 test494 test495 test496 test497 test498 \
 test499 test500 test501 test502 test503 test504 test505 test506 test507 \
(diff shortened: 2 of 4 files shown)


bagder:
Example output:
~~~shell
$ curl -H “host: store.example.com”
Warning: The argument '“host:' starts with a unicode quote where maybe an 
Warning: ASCII " was intended?
curl: (6) Could not resolve host: store.example.xn--com-9o0a

emanuele6:
To test this also on windows you could use `-K -` and `<stdin>-H “host: %HOSTIP:%HTTPPORT/”</stdin>` instead of command line arguemnts I think; maybe as a separate test.

bagder:
Oh right, clever! Let me try that...

dfandrich:
Is UTF-8 even allowed in HTTP headers? If not, then warning if the first character is >=0x80 is probably the way to go. That catches funniness in non-UTF-8 8-bit locales, too. Does this work properly on EBCDIC machines?

bagder:
> Is UTF-8 even allowed in HTTP headers?

No, but there is usually nothing that prevents UTF-8 from being transported in HTTP headers.

> If not, then warning if the first character is >=0x80 is probably the way to go. That catches funniness in non-UTF-8 8-bit locales, too. 

Maybe. It would then however also catch all the "funny tricks" and things people might be using and doing quite intentionally. I believe "the wrong quote" is much more likely to be a genuine mistake - and one we have seen done multiple times. I have never seen another unicode character used by mistake.

> Does this work properly on EBCDIC machines?

I doubt that. How does UTF-8 work on EBCDIC? I don't think we have any user of EBCDIC ever contributing... I would be surprised if curl works correctly on EBCDIC systems.

We dropped support for `CURL_DOES_CONVERSIONS` back in 2022.

bagder:
Oh, and btw, this warning is for *any* option argument starting with a unicode quote character, not just for `-H`.

PR #13218, opened 2024-03-28
doc(pytest): `--repeat` -> `--count`

jamacku:
Pytest doesn't have a `--repeat` option, but it does have a `--count` option.

```
--count=COUNT         Number of times to repeat each test
```


bagder (approved):

bagder:
Thanks!

Issue #13212, opened 2024-03-27
--compressed fails with error 23


finkjsc:
### I did this

curl --compressed -o "test.txt" "https://www.troyhunt.com/rss/"

### I expected the following

Output file to be written without error, but fails with error 23: Failed writing received data to disk/application.

Oddly the file is written correctly but still the error persists.

Works without --compressed.

Most other URL's work, but many now fail with this error.

Versions before 8.7.1 worked without this error.

### curl/libcurl version

curl 8.7.1 (x86_64-w64-mingw32) libcurl/8.7.1 LibreSSL/3.8.3 zlib/1.3.1 brotli/1.1.0 zstd/1.5.5 WinIDN libpsl/0.21.5 libssh2/1.11.0 nghttp2/1.60.0 ngtcp2/1.4.0 nghttp3/1.2.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL SSPI threadsafe UnixSockets zstd

### operating system

Windows 10 Pro 22H2

dfandrich:
This may be a duplicate of #13209

icing:
Please see #13219 for a fix.

Issue #13209, opened 2024-03-27
auracle fail with libcurl 8.7.1


eworm-de:
### I did this

I pushed `curl 8.7.1-1` to the *Arch Linux* testing repository earlier today. Turns out we have a regression with an AUR helper application [auracle](https://github.com/inglor/auracle), that links against `libcurl`. Sadly output is little to no useful only:
```
% auracle clone st
error: UNKNOWN: 
```
Running `git bisect` I was pointed at 463472a2d6e3301c1468b5323b856cb67a91f579 being the first bad commit.

Investigating the code I landed in [src/aur/aur.cc line 468](https://github.com/inglor/auracle/blob/6fdcc4de11cd95ba66f5fe68a45bed6ca21a239c/src/aur/aur.cc#L468) where `msg` is assigned, and `msg->data.result` some lines below is `23`, which is `CURLE_WRITE_ERROR`. With `curl 8.6.0` the result was `0` or `CURLE_OK`.

From the commit message it is not obvious to me if the public api is supposed to change - I guess no. (Everything else should require a `soname` bump anyway.) So not sure what to blame here... Also possible that the api was misused and worked by chance only.

### I expected the following

The application should continue to work as expected, downloading its content with `libcurl`.

### curl/libcurl version

```
% pacman -Q curl
curl 8.7.1-1
```
```
% curl --version
curl 8.7.1 (x86_64-pc-linux-gnu) libcurl/8.7.1 OpenSSL/3.2.1 zlib/1.3.1 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.7 libpsl/0.21.2 libssh2/1.11.0 nghttp2/1.60.0 nghttp3/1.2.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

*Arch Linux* with testing repositories enabled

eworm-de:
CC @inglor

eworm-de:
CC @icing (for commit authorship)

icing:
[...]

PR #13219, opened 2024-03-28
brotli and others, simply pass through 0-length writes

icing:
- refs #13212 and #13209
- curl's transfer handling may write 0-length chunks at the end of the download with an EOS flag. (HTTP/2 does this commonly)
- content encoders need to pass-through such a write and not count this as error in case they are finished decoding

### DIFF
--- a/lib/content_encoding.c
+++ b/lib/content_encoding.c
@@ -300,7 +300,7 @@ static CURLcode deflate_do_write(struct Curl_easy *data,
   struct zlib_writer *zp = (struct zlib_writer *) writer;
   z_stream *z = &zp->z;     /* zlib state structure */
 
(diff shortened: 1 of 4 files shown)


eworm-de:
Works for me, and pushed a package with that change to testing:
https://gitlab.archlinux.org/archlinux/packaging/packages/curl/-/commit/ef6ee3e202fdb4fe3bf0b0803140eb2cbd15d33a

bagder (approved):

jeroen:
The wise men at Apple have now pushed libcurl 8.7.1 to the latest MacOS 13 and 14, which introduces the bug that was fixed in this PR...

bagder:
:disappointed: Unfortunate.

icing:
Indeed.

PR #13228, opened 2024-03-30
openldap: Fix support for IPv6 addresses

sergiodj:
Ref.: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1053643

When the user specified an IPv6 address to be used as an LDAP server,
curl will fail to properly enclose it in square brackets, which causes
the connection to fail because the host address cannot be
distinguished from the port:

```
$ curl -v ldap://[fd42:be5:e632:a6b3:216:3eff:feb1:5bc4]:389
...
* LDAP local: Cannot connect to ldap://fd42:be5:e632:a6b3:216:3eff:feb1:5bc4:389, Bad parameter to an ldap routine
...
```

Fix this by always enclosing the IPv6 address in square brackets.

While this could be done without caring for the IP protocol version
being used (as per RFC 3986, which indicates that IPv4 also works
enclosed in square brackets), I decided to make the code IPv6-specific
to minimize the possibility of regressions.

I'm not familiar with curl's internals, so any suggestions/corrections
are welcome.

### DIFF
--- a/lib/openldap.c
+++ b/lib/openldap.c
@@ -522,6 +522,7 @@ static CURLcode oldap_connect(struct Curl_easy *data, bool *done)
   static const int version = LDAP_VERSION3;
   int rc;
   char *hosturl;
+  char *hostname_quoted;
 #ifdef CURL_OPENLDAP_DEBUG
   static int do_trace = -1;
 #endif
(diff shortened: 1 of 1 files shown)


bagder:
How about this slightly simpler approach?
~~~diff
diff --git a/lib/openldap.c b/lib/openldap.c
index 47266f64e4..93d25b61cf 100644
--- a/lib/openldap.c
+++ b/lib/openldap.c
@@ -546,13 +546,16 @@ static CURLcode oldap_connect(struct Curl_easy *data, bool *done)
     result = oldap_parse_login_options(conn);
     if(result)
       return result;
   }
 
-  hosturl = aprintf("ldap%s://%s:%d",
-                    conn->handler->flags & PROTOPT_SSL? "s": "",
-                    conn->host.name, conn->remote_port);
+  hosturl = aprintf("%s://%s%s%s:%d",
+                    conn->handler->scheme,
+                    conn->bits.ipv6_ip? "[": "",
+                    conn->host.name,
+                    conn->bits.ipv6_ip? "[": "",
+                    conn->remote_port);
   if(!hosturl)
     return CURLE_OUT_OF_MEMORY;
 
   rc = ldap_init_fd(conn->sock[FIRSTSOCKET], li->proto, hosturl, &li->ld);
   if(rc) {

sergiodj:
Ah, I completely missed the existence of `ipv6_ip`.  That makes things much simpler, indeed.  LGTM, thanks!

PR #13233, opened 2024-03-30
Enabling IDN via macOS/iOS DNS services

MonkeybreadSoftware:
If no IDN2 library available and no Windows, we can leverage the resolving of UTF-8 encoded domains by macOS/iOS network services.

We regularly patch this into new CURL versions and I'd just like to pass it to the source code.

This is a fallback for when building curl for iOS or MacOS without LibIDN2 and still make domains work with umlauts.

### DIFF
--- a/lib/idn.c
+++ b/lib/idn.c
@@ -181,6 +181,18 @@ static CURLcode idn_decode(const char *input, char **output)
     result = CURLE_NOT_BUILT_IN;
 #elif defined(USE_WIN32_IDN)
   result = win32_idn_to_ascii(input, &decoded);
+#elif defined(HAVE_NETDB_H) && defined(__APPLE__)
+    /* let MacOS/iOS do this as part of DNS service */
+    struct hostent *p = gethostbyname(input);
+    if(p) {
+      if(p->h_name && (p->h_name[0])) {
+        /* host name is idn encoded by MacOS */
+        char *name = strdup(p->h_name);
+        if(name) {
+          decoded = name;
+        }
+      }
+    }
 #endif
   if(!result)
     *output = decoded;


lib/idn.c
```diff
@@ -181,6 +181,18 @@ static CURLcode idn_decode(const char *input, char **output)
     result = CURLE_NOT_BUILT_IN;
 #elif defined(USE_WIN32_IDN)
   result = win32_idn_to_ascii(input, &decoded);
+#elif defined(HAVE_NETDB_H) && defined(__APPLE__)
+    /* let MacOS/iOS do this as part of DNS service */
+    struct hostent *p = gethostbyname(input);
```
gvanem, line 186: Perhaps check for `HAVE_GETHOSTBYNAME_R` or call `Curl_ipv4_resolve_r()`?

MonkeybreadSoftware: Well, macOS and iOS have no gethostbyname_r as far as I know. 
Calling Curl_ipv4_resolve_r may be a possibility, so we use Curl_getaddrinfo_ex here, which uses getaddrinfo internally, skipping gethostbyname all together.

lib/idn.c
```diff
@@ -181,6 +181,18 @@ static CURLcode idn_decode(const char *input, char **output)
...
+    if(p) {
+      if(p->h_name && (p->h_name[0])) {
+        /* host name is idn encoded by MacOS */
+        char *name = strdup(p->h_name);
+        if(name) {
+          decoded = name;
+        }
+      }
+    }
```
bagder, line 195: I see three if() calls that I presume all should set an error code if they fail?

bagder (commented):
What is a little bit scary here is that since this makes a full blown name resolve, this function can take **a looong** time to completion. Meaning multiple seconds.

This only provides IDN to punycode, not the other way around. So only "half" of the IDN features libcurl uses.

This should *probably* set the IDN feature in the curl_version_info features?

lib/idn.c
bagder: Maybe worst of all: this requires that the name resolves. It can't convert just an arbitrary host name...!

MonkeybreadSoftware: Maybe I should add some context. We had clients complaining that domains with umlauts didn't work. libidn isn't preinstalled on macOS or iOS by Apple. And we didn't want to carry the libidn2 library with every app to just get occasionally the domain resolved.

bagder: I understand, but that does not change my comments.

MonkeybreadSoftware:
Thanks for the feedback. I think we may have something better to suggest in a few days.

MonkeybreadSoftware:
See new pr:
https://github.com/curl/curl/pull/13246

PR #13239, opened 2024-03-30
maketgz: put docs/RELEASE-TOOLS.md into the tarball

bagder:
Generated with scripts/release-tools.sh

The script lists the exact Debian package names and version numbers for the tools that are used to generate the tarball.

## `docs/RELEASE-TOOLS.md`

~~~
# Release tools

The following tools and their Debian package version numbers were used to
produce this release tarball.

- autoconf: 2.71-3
- automake: 1:1.16.5-1.3
- libtool: 2.4.7-7
- make: 4.3-4.1
- perl: 5.38.2-3.2
- git: 1:2.43.0-1+b1

# Reproduce the tarball

- Clone the repo and checkout the release tag
- Install the same set of tools + versions as listed above

## Do a standard build

- autoreconf -fi
- ./configure [...]
- make

## Generate the tarball

- ./maketgz [version]

### DIFF
--- a/scripts/release-tools.sh
+++ b/scripts/release-tools.sh
@@ -1,4 +1,27 @@
 #!/bin/sh
+#***************************************************************************
+#                                  _   _ ____  _
+#  Project                     ___| | | |  _ \| |
+#                             / __| | | | |_) | |
+#                            | (__| |_| |  _ <| |___
+#                             \___|\___/|_| \_\_____|
+#
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# This software is licensed as described in the file COPYING, which
+# you should have received as part of this distribution. The terms
+# are also available at https://curl.se/docs/copyright.html.
+#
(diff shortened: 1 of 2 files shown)


bagder:
The script also uses `sh`, `find`, `date`, `sed` etc but I cannot see that their versions should be relevant

scripts/release-tools.sh
```diff
@@ -0,0 +1,67 @@
...
+
+cat <<MOO
+# Release tools
+
+The following tools and their Debian package version numbers were used to
+produce this release tarball.
+
+MOO
+
+exists=`which dpkg`;
```
kalvdans, line 34:
```suggestion
exists=`which dpkg`
```

scripts/release-tools.sh
```diff
@@ -0,0 +1,67 @@
...
+if test ! -e "$exists"; then
+    echo "(unknown, could not find dpkg)"
+    exit
```
kalvdans, line 37:
```suggestion
    exit 1
```

PR #5371, opened 2020-05-10
Add IDN for Apple and iOS via System

MonkeybreadSoftware:
This is a proposal for a better way to implement #5330 with using threaded resolver.
We ask MacOS and iOS to provide us the canonical name for IDN domains by setting AI_CANONNAME and then using the canon name as new host name for the HTTP request.

PR #5330, opened 2020-05-03
add idn for MacOS and iOS

MonkeybreadSoftware:
we let do MacOS and iOS the IDN encoding via gethostbyname.
Makes a query to e.g. Müller.de work on MacOS without using libidn.
Still if libidn is present, it will be preferred.

PR #13246, opened 2024-03-31
idn: add native AppleIDN (icucore) support for macOS/iOS

MonkeybreadSoftware:
I implemented the IDN functions for macOS and iOS using Unicode
libraries coming with macOS and iOS.

Builds and runs here on macOS 14.2.1. Also verified to load and
run on older macOS version 10.13.

Build requires macOS SDK 13 or equivalent.

Set `-DUSE_APPLE_IDN=ON` CMake option to enable it.
With autotools and other build tools, set these manual options:
```
CPPFLAGS=-DUSE_APPLE_IDN
LIBS=-licucore
```

Completes TODO 1.6.

TODO: add autotools option and feature-detection.

Refs: #5330 #5371
Co-authored-by: Viktor Szakats                                                            
Closes #13246


MonkeybreadSoftware:
Please note that applications using libcurl may need to use -licucore to get icucore library added.
But the normal curl command line tool just builds, so I assume icucore is already referenced somehow.

lib/idn.c
```diff
@@ -50,6 +50,69 @@
...
+
+    int32_t len = uidna_nameToASCII_UTF8(idna, in, inlength, buffer,
+      sizeof(buffer), &info, &err);
+    uidna_close(idna);
+    if(U_FAILURE(err)) {
+      return CURLE_URL_MALFORMAT;
+    }
+    else {
+      *out = strdup(buffer);
+      fprintf(stderr, "%s: %s\n", __FUNCTION__, *out);
```
vszakats, line 77: Suggesting to replace `__FUNCTION__` with a literal. Same in the next hunk.

bagder: fprintf() is not allowed though. infof() or failf() should be used for logging/errors.

MonkeybreadSoftware: Of course. Removed. I just had them to print the input/output for debugging.
And I thought I removed them all...

vszakats:
> Please note that applications using libcurl may need to use -licucore to get icucore library added. But the normal curl command line tool just builds, so I assume icucore is already referenced somehow.

Speaking of CMake, this snippet would add that:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 656aa7c74..9f0a2b4b0 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -870,6 +870,10 @@ if(WIN32)
   endif()
 endif()
 
+if(APPLE)
+  list(APPEND CURL_LIBS "icucore")
+endif()
+
 #libpsl
 option(CURL_USE_LIBPSL "Use libPSL" ON)
 mark_as_advanced(CURL_USE_LIBPSL)
```

I wonder if we should also add an build option for this, like we have for Windows. Some may prefer to keep IDN-support disabled (also helps if there is any regression).

MonkeybreadSoftware:
I just tried it on iOS and it seems to work fine here.
And much better than my older PR. :-)

lib/idn.c
```diff
@@ -50,6 +50,80 @@
...
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 170000
+#define XCODE15 1
+#endif
+
+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 140000
+#define XCODE15 1
+#endif
+
+#if XCODE15
+#include <unicode/uidna.h>
```
vszakats, line 70: Also available in macOS SDK 13:
```
/Library/Developer/CommandLineTools/SDKs/MacOSX13.sdk/usr/include/unicode/uidna.h
/Library/Developer/CommandLineTools/SDKs/MacOSX13.sdk/usr/lib/libicucore.tbd
```

MonkeybreadSoftware: Strange, because the CI reported the header missing!?

If the header should exist, then the CI should not fail and the check for SDK versions is not needed.

MonkeybreadSoftware:
trying now with `#if __has_include(<unicode/uidna.h>)` to check if the header exists.

lib/idn.c
vszakats: That's odd, but the header and lib does exist on this outdated CommandLineTools install. It also compiles and links and returns a non-NULL address for `uidna_openUTS46(UIDNA_CHECK_BIDI, &err)`.

Do you have a link for the CI failure?

lib/idn.c
```diff
@@ -181,6 +242,8 @@ static CURLcode idn_decode(const char *input, char **output)
     result = CURLE_NOT_BUILT_IN;
 #elif defined(USE_WIN32_IDN)
   result = win32_idn_to_ascii(input, &decoded);
+#elif defined(HAVE_APPLE_IDN)
+  result = mac_idn_to_ascii(input, &decoded);
```
vszakats, line 246:
```suggestion
  result = mac_idn_to_ascii(input, &decoded);
#else
  (void)input;
```
To avoid:
```
curl/lib/idn.c:235:40: error: unused parameter 'input' [-Werror=unused-parameter]
```

lib/idn.c
```diff
@@ -50,6 +50,67 @@
...
+  if(U_FAILURE(err)) {
+    return CURLE_OUT_OF_MEMORY;
+  }
+  else {
+    UIDNAInfo info = UIDNA_INFO_INITIALIZER;
+    char buffer[256] = {0};
+    int32_t inlength = -1; /* zero terminated */
+
+    int32_t len = uidna_nameToASCII_UTF8(idna, in, inlength, buffer,
+      sizeof(buffer), &info, &err);
```
bagder, line 70:
```suggestion
    int32_t len = uidna_nameToASCII_UTF8(idna, in, -1, /* null terminated */
      buffer, sizeof(buffer), &info, &err);
```

lib/idn.c
```diff
@@ -50,6 +50,67 @@
...
+  if(U_FAILURE(err)) {
+    return CURLE_OUT_OF_MEMORY;
+  }
+  else {
+    UIDNAInfo info = UIDNA_INFO_INITIALIZER;
+    char buffer[256] = {0};
+    int32_t inlength = -1; /* zero terminated */
+
+    int32_t len = uidna_nameToUnicodeUTF8(idna, in, inlength, buffer,
+      sizeof(buffer), &info, &err);
```
bagder, line 98:
```suggestion
    int32_t len = uidna_nameToUnicodeUTF8(idna, in, -1, /* null terminated */
      buffer, sizeof(buffer), &info, &err);
```

bagder (commented):
Maybe skip the variables here?

lib/idn.c
MonkeybreadSoftware: This is fixed with __has_include() check. Compiles using the feature if you have the header, but not if it is missing.

lib/idn.c
MonkeybreadSoftware: This is no longer needed since we only set USE_IDN if we have the header, so the `#else` is not needed.

lib/idn.c
MonkeybreadSoftware: Well, if you prefer, we can skip them. I like to document the meaning of the value as I don't like magic numbers that much. Or we use strlen(in) for the parameter.

bagder: It saves having to declare a pointless variable, the "documentation" is still there. And it's not like you document all the other arguments. Finally: it's not your job to document the API function that is used.

MonkeybreadSoftware: Okay. Change made.

lib/idn.h
```diff
@@ -24,9 +24,16 @@
  *
  ***************************************************************************/
 
+/* for macOS and iOS targets */
+#if defined(__APPLE__)
+#if __has_include(<unicode/uidna.h>)
```
vszakats, line 29: Thinking of the gcc (or possibly other non-Apple) compilers targeting Apple, enclosing this in `#if defined __has_include` might be useful to detect availability of the preprocessor feature:
https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

(at the moment GCC 13 is incompatible with SDK 13.x. Haven't tried 15.x though.)

MonkeybreadSoftware: Thanks for the idea. I've added a check for whether __has_include is available.

vszakats: LGTM, thanks!

bagder (approved):

MonkeybreadSoftware:
Does it build for you without?
For me it does as I don't have it in the link lists. Maybe some other library...

Okay, found it: CoreFoundation.framework is references by lib curl and that references /usr/lib/libicucore.A.dylib internally. So you don't need to reference it explicitly.

vszakats:
@MonkeybreadSoftware: We're good then, thanks for checking!

vszakats:
Made a test with curl-for-win, and it seems the extra library is missing (with Apple clang):
https://github.com/curl/curl-for-win/actions/runs/8525243980/job/23351725118#step:3:4941

```
Undefined symbols for architecture arm64:
  "_uidna_close", referenced from:
      _curl_url_get in libcurl.a(unity_0_c.c.o)
      _Curl_idn_decode in libcurl.a(unity_0_c.c.o)
      _Curl_idn_encode in libcurl.a(unity_0_c.c.o)
      _Curl_idnconvert_hostname in libcurl.a(unity_0_c.c.o)
  "_uidna_nameToASCII_UTF8", referenced from:
      _curl_url_get in libcurl.a(unity_0_c.c.o)
      _Curl_idn_decode in libcurl.a(unity_0_c.c.o)
      _Curl_idnconvert_hostname in libcurl.a(unity_0_c.c.o)
  "_uidna_nameToUnicodeUTF8", referenced from:
      _curl_url_get in libcurl.a(unity_0_c.c.o)
      _Curl_idn_encode in libcurl.a(unity_0_c.c.o)
  "_uidna_openUTS46", referenced from:
      _curl_url_get in libcurl.a(unity_0_c.c.o)
      _Curl_idn_decode in libcurl.a(unity_0_c.c.o)
      _Curl_idn_encode in libcurl.a(unity_0_c.c.o)
      _Curl_idnconvert_hostname in libcurl.a(unity_0_c.c.o)
ld: symbol(s) not found for architecture arm64
```

Also true for standard llvm:
https://github.com/curl/curl-for-win/actions/runs/8525243980/job/23351725441#step:3:4946

MonkeybreadSoftware:
Sounds like we need a "-licucore" command for the linking.

CMakeLists.txt
```diff
@@ -292,15 +292,19 @@ if(ENABLE_IPV6 AND NOT WIN32)
         CACHE BOOL "Define if you want to enable IPv6 support" FORCE)
   endif()
 
-  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND NOT ENABLE_ARES)
-    set(use_core_foundation_and_core_services ON)
+  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    if (ENABLE_ARES)
+      list(APPEND CURL_LIBS "-licucore")
+    else()
```
vszakats, line 298: Isn't `-licucore` necessary for all `APPLE`? The next issue is that this makes it necessary to detect the presence of this OS feature in CMake/autotools, then pass this on via `HAVE_APPLE_IDN` (replacing `__has_include`). Then add this library if the feature is available and enable it in source on this condition too.

MonkeybreadSoftware: Sure, we could add it for the other case, too. Or just move it outside the inner condition.
I like the __has_include as it allows to check for a header on compile time.
Feel free to adjust the build scripts.

vszakats: Moving it around will not help for Apple targets not supporting this feature. It will fail when linking a non-existing library.

MonkeybreadSoftware: Well, if here is someone who can write CMake conditions and configure checks to check for presents of unicode/uidna.h and whether you can link "icucore". A check for file on disk won't work since the library is in the dylib cache archive. 

On the other side, why bother to check at all. All platforms from Apple support the header file and include the library as far as I know: Apple TV OS, iOS, MacOS, WatchOS and Vision Pro OS. And I verified that the library exists on macOS 10.13, so included for a couple of years already.

Of course it would be nice to have a switch to turn it off in case of someone seeing a problem. But CMake/Configure scripts is outside my expertise.

vszakats: I proposed a patch earlier that added `icucore` unconditionally to CMake. The current logic seems wrong, e.g. what does `ENABLE_ARES` have to do with enabling `icucore`?

But even that simple patch would:
- fail when building with an older (pre v13) Apple toolchain.
- likely fail to build when using a new Apple toolchain but targeting an older OS release.
- doesn't allow to disable this feature. This might have valid reasons like someone hitting a snag with the newly added logic, and still wanting to build curl. Or, someone just doesn't want IDN support for whatever reason.
- leaves this feature broken with autotools. That needs the same detection/enabler/lib-adder logic.

I can't volunteer for this now, because I'm working since Thursday pretty much day and night on open source stuff on my free time.

A way to fix this is to remove header detection from the C code, replace it with `HAVE_APPLE_IDN` (meaning: support is available) and `USE_APPLE_IDN` (we also want to use it) macros, all expected to be set manually for the time being. Then later add the CMake logic doing the automatic detecition/feature-toggle, and the same for autotools.

MonkeybreadSoftware: >  what does ENABLE_ARES have to do with enabling icucore?

Well, if you enable ARES, you don't link to CoreFoundation, which links icucore.
So we don't need to reference icucore directly.
But yes, we could change it to use constants as switches.

CMakeLists.txt
```diff
@@ -292,15 +292,18 @@ if(ENABLE_IPV6 AND NOT WIN32)
         CACHE BOOL "Define if you want to enable IPv6 support" FORCE)
   endif()
 
-  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND NOT ENABLE_ARES)
-    set(use_core_foundation_and_core_services ON)
+  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    list(APPEND CURL_LIBS "-licucore")
+    if (NOT ENABLE_ARES)
```
vszakats, line 297:
```suggestion
    if(NOT ENABLE_ARES)
```

CMakeLists.txt
vszakats: The problem remains that `icucore` has only been available since SDK 13, so this break building curl on everything before. Even with newer SDKs it breaks when explicitly targeting older ones.

There is no other way IMO than making this an option. Then implementing detection for this at the build level.

MonkeybreadSoftware: I know, that you'd like to have a switch and I may look into adding one.

How far in past do you need to support older macOS versions? We support 10.13 (from 2017) since that is what current Xcode 15 can build.

CMakeLists.txt
```diff
@@ -292,15 +292,18 @@ if(ENABLE_IPV6 AND NOT WIN32)
         CACHE BOOL "Define if you want to enable IPv6 support" FORCE)
   endif()
 
-  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND NOT ENABLE_ARES)
-    set(use_core_foundation_and_core_services ON)
+  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    list(APPEND CURL_LIBS "-licucore")
```
vszakats, line 296: In context of `icucore` shouldn't it use `if(APPLE)` instead?

MonkeybreadSoftware: You are right. It doesn't need to be in the block for "ENABLE_IPV6 AND NOT WIN32" and Darwin and not ARES.

CMakeLists.txt
vszakats: We don't have this formally defined, I build curl-for-win for 10.9.

IMO this isn't really be about trying to hardwire or guess these versions in the build or source.

This PR already is already good if the feature availability can be passed manually, and offers another option to enable it. We use these mechanics for similar cases.

As a next step (in a separate PR even) detecting availability can be automated. E.g. by setting up the icucore lib and checking for one of the APIs, using `check_function_exists()` in CMake.

MonkeybreadSoftware: For curl-for-**win** for **mac**OS 10.9?

We could detect macOS 10.13 SDK (or iOS 11.0 SDK) via #if like this:

```
#if TARGET_OS_OSX

#if MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_13
// disable feature for old MacOS SDK
#endif

#elif TARGET_OS_IPHONE

#if __IPHONE_OS_VERSION_MIN_REQUIRED  < __IPHONE_11_0
// disable feature for old iOS SDK (also Watch, VisionOS, AppleTV, etc.)
#endif

#endif
```

So it could be automatically disabled for older SDK.

We'd just need a nice way to detect whether library is there and add it. Configure script needs to get a **-licucore**, too. But not sure where.

vszakats: What's the benefit of scattering detection into these three: a hardwired logic in-source, a dynamic-header detection in-source and future-developed library detection in autotools and CMake?

MonkeybreadSoftware: Well, if I could do it all in source, I wouldn't need to wait for someone to do autotools and CMake.
I'd prefer to have it all automatic. If this is opt-in for building on macOS, I fear it may not get used.

vszakats: Pure in-source detection would be nice, but I can't see how this could be technically feasible.

Since we do need detection, our option is to either make it opt-in for now and add detection in a separate commit, or add detection now. We also need an opt-out option in either case.

This is a useful feature, so it's reasonable to assume that even the two-stage implementation will pass through quickly. My worry is autotools, which would need to be done in the same step, but needs a very different skillset than CMake. Worst case we can make an exception in the CI job responsible for keeping CMake and autotools in sync.

lib/idn.c
```diff
@@ -50,6 +50,63 @@
...
+{
+  UErrorCode err = U_ZERO_ERROR;
+  UIDNA* idna = uidna_openUTS46(UIDNA_CHECK_BIDI, &err);
+  if(U_FAILURE(err)) {
+    return CURLE_OUT_OF_MEMORY;
+  }
+  else {
+    UIDNAInfo info = UIDNA_INFO_INITIALIZER;
+    char buffer[256] = {0};
+    int32_t len = uidna_nameToASCII_UTF8(idna, in, -1, buffer,
```
vszakats, line 67: This is showing the warning (also for the next similar line):
```
./lib/idn.c:67:13: warning: unused variable 'len' [-Wunused-variable]
    int32_t len = uidna_nameToASCII_UTF8(idna, in, -1, buffer,
            ^
./lib/idn.c:93:13: warning: unused variable 'len' [-Wunused-variable]
    int32_t len = uidna_nameToUnicodeUTF8(idna, in, -1, buffer,
            ^
2 warnings generated.
```

Is `len` supposed to be used later? Or should these be cast to `(void)` to ignore the return value?

MonkeybreadSoftware: Well, we don't use the len currently:

`*out = strdup(buffer);
`

We could use a strdup variant which takes a length instead of relaying on the NULL byte.
Or just cast to void. What would you prefer?

vszakats: These APIs return the length to let us allocate a buffer dynamically. Since we use fixed ones (also for WinIDN and libidn2), we can just `(void)` them, I think.

MonkeybreadSoftware: Okay. I'll change it.

vszakats:
Pushed an update:
- add CMake support with auto-detection. Enabled by default. Requires SDK 13 or later.
- accomodate some logic to the new IDN.
- rebase on master. This adds curl-for-win macOS CI builds.
- with autotools and other builds tool AppleIDN can be enabled with:
  ```
  CPPFLAGS=-DUSE_APPLE_IDN
  LIBS=-licucore
  ```

Regarding targeting old macOS versions:

Limited local tests confirm that a build targeting 10.9 and using SDK 13 successfully detects and builds AppleIDN support, and the binary also runs and apparently works (converting an accented domain name to punycode) when run on a macOS that's actually missing the dynamically linked `/usr/lib/libicucore.A.dylib`. I don't know how this works.

If this isn't always true or my tests are wrong, we should probably disable AppleIDN by default, or allow/block-list it in CMake based on `CMAKE_OSX_DEPLOYMENT_TARGET` or something.

vszakats:
Is it correct that this PR completes TODO 1.6, and we can delete it?

MonkeybreadSoftware:
Yes. Thanks. Finally get Todo 1.6 done :-)

vszakats:
Thank you @MonkeybreadSoftware!

PR #13250, opened 2024-03-31
Dockerfile for release automation

daniel-j-h:
Hey @bagder :wave: I've seen your post on mastodon on how to reproduce the release tarballs.

I wanted to bounce this off of you as an idea
1. We can check in a Dockerfile build environment based on a specific distribution (debian bookworm here)
2. We then can have a Github Action to build and create the release tarballs in this specific environment

This is far from reproducible *builds* but maybe at least it could be a cheap way forward for reproducible *release tarballs*.

Here is the github action running in my fork

https://github.com/daniel-j-h/curl/actions/runs/8499306747/job/23279970815

What's left to do here is
- [ ] discussing if this could be a way forward
- [ ] have the github action use a new signing key (put in github secrets) and sign the tarballs
- [ ] have the github action upload release tarballs and signatures on every tag push

Refs
- https://mastodon.social/@bagder/112185825134288411
- https://curl.se/mail/lib-2024-03/0062.html


bagder:
Having a dockerfile that can do identical tarballs to the one we ship is a sensible idea as it makes it easier for anyone who feels like it to reproduce them. I don't think there is any particular need to do releases "in the cloud" as it does not actually add value or protection to the process. As long as the tarballs can be reproduced, they can be verified to be built from a known git repository state and then it does not matter where they were generated.

daniel-j-h:
> Having a dockerfile that can do identical tarballs to the one we ship is a sensible idea as it makes it easier for anyone who feels like it to reproduce them. I don't think there is any particular need to do releases "in the cloud" as it does not actually add value or protection to the process. As long as the tarballs can be reproduced, they can be verified to be built from a known git repository state and then it does not matter where they were generated.

Got it, that makes sense! I can remove the github action integration if you don't want the github action to create and upload those tarballs. The only problem I can see is if the Dockerfile is not part of your workflow or part of some sort of automation that it it's very easy to forget to keep it up to date. Do you have ideas how we could make sure we don't forget about it?

---

Then I did the following experiment: I went to the release page and downloaded the latest `curl-8.7.1.tar.gz` and compared it against a locally created `curl-8.7.1.tar.gz` from the git tag checkout and using the Dockerfile running the commands below.

```bash
docker build -t curl/curl .

docker run --rm -v (pwd):/usr/src -w /usr/src curl/curl autoreconf -fi
docker run --rm -v (pwd):/usr/src -w /usr/src curl/curl ./configure --without-ssl --without-libpsl
docker run --rm -v (pwd):/usr/src -w /usr/src curl/curl make -j8
docker run --rm -v (pwd):/usr/src -w /usr/src curl/curl ./maketgz 8.7.1
```

Here is an overview of what a recursive diff is showing me just following these simple steps:

```bash
$ diff -qr local remote

Files local/curl-8.7.1/CHANGES and remote/curl-8.7.1/CHANGES differ
Files local/curl-8.7.1/docs/curl-config.1 and remote/curl-8.7.1/docs/curl-config.1 differ
Files local/curl-8.7.1/include/curl/curlver.h and remote/curl-8.7.1/include/curl/curlver.h differ
Files local/curl-8.7.1/include/curl/Makefile.in and remote/curl-8.7.1/include/curl/Makefile.in differ
Files local/curl-8.7.1/ltmain.sh and remote/curl-8.7.1/ltmain.sh differ
Files local/curl-8.7.1/packages/Makefile.in and remote/curl-8.7.1/packages/Makefile.in differ
Files local/curl-8.7.1/packages/vms/Makefile.in and remote/curl-8.7.1/packages/vms/Makefile.in differ
Files local/curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.10/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.10/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.10/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.10/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.20/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.20/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.20/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.20/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.30/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.30/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.30/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.30/src/curl.vcxproj differ
Files local/curl-8.7.1/scripts/Makefile.in and remote/curl-8.7.1/scripts/Makefile.in differ
Files local/curl-8.7.1/src/tool_hugehelp.c and remote/curl-8.7.1/src/tool_hugehelp.c differ
Files local/curl-8.7.1/tests/data/Makefile.in and remote/curl-8.7.1/tests/data/Makefile.in differ
Files local/curl-8.7.1/tests/http/Makefile.in and remote/curl-8.7.1/tests/http/Makefile.in differ
Files local/curl-8.7.1/tests/Makefile.in and remote/curl-8.7.1/tests/Makefile.in differ
```

What I can see in there
- the `CHANGES` file differs: the one from the local built is missing individual changes; my hunch is I need git installed for this file to be assembled?
- There are date stamps in a few files e.g. in the `curl-config.1` file and in the `curlver.h` file (the maketgz script datestamps [here](https://github.com/curl/curl/blob/5418812d3441f7b6bf73ab783b521221b1cec8d9/maketgz#L80-L81). I guess we can ignore this.
- There are some makefile arguments different in `Makefile.in` (--gnu vs --foreign)
- The `src/tool_hugehelp.c` tool in the github release seems to have an additional second part to it with a compressed part that gets sent through gzip; this part is not in my local tarball

and this just looking though the first five or so differences.

I don't think I can get to the bottom of all of those differences and I for sure can not judge why they're different. I believe some could be from slightly different tools or how we invoke them. I would appreciate your help here in getting to the bottom of this.

bagder:
1. Right, we need a CI job to make sure that the Dockerfile remains functional. Maybe we should make use of it for the `distcheck.yml` job?
2. We need git installed to generate `CHANGES` in the tarball
3. You should set the env variable `SOURCE_DATE_EPOCH` to a fixed time (epoch) before you start the tests so that they use the exact same
4. We need `zlib1g-dev` installed so that configure detects it and hugehelp.c gets the compressed version bundled
5. The MS project files should not differ...

bagder:
The autoconf differences I assume is because your image does not use the exact same set that I used when made the release.

vszakats:
> 3. You should set the env variable `SOURCE_DATE_EPOCH` to a fixed time (epoch) before you start the tests so that they use the exact same

Possibly `export TZ=UTC`?

Speaking of 8.7.1, also `export LC_ALL=C`, to generate the date in English. This will not be necessary in upcoming versions after today's afdd1129166938a4ba952aa1a9503adf0ec1a7a3.

daniel-j-h:
Thanks folks! I made some progress here and found two date-related issues
1. The maketgz script does not respect the `SOURCE_DATE_EPOCH` env var but always uses today; fixing in https://github.com/curl/curl/pull/13250/commits/817f968306b9431a8a644fde9a00a138594adbc9
2. Our date computation is off somewhere resulting in the previous day before the release day to end up in a manpage

Looking at the github release for 8.7.1 it was on 2024-03-27 (~10am ish)

https://github.com/curl/curl/releases/tag/curl-8_7_1

In the release tarballs I can find a datestamp from 2024-03-**26**

```
remote/curl-8.7.1/docs/curl-config.1
2:.TH curl-config 1 "March 26 2024" curl-config
```

and other datestamps seem to be correctly set to 2024-03-**27**

```
remote/curl-8.7.1/include/curl/curlver.h
73:#define LIBCURL_TIMESTAMP "2024-03-27"
```

My hunch is that
1. either our datestamping logic for stamping the man page is off somewhere, or
2. maybe there was no full release re-creation done for the releases from a clean state on 2024-03-27 by chance but some of it got datestamped the day before

This is in no way critical I just wanted to leave it here since it makes automatically comparing the diffs tricky.

---

Then I looked into the differences in the `*.in` files, e.g.

```diff
djh@rf /t/curlpkgs> diff {local,remote}/curl-8.7.1/include/curl/Makefile.in
462c462
<       echo ' cd $(top_srcdir) && $(AUTOMAKE) --gnu include/curl/Makefile'; \
---
>       echo ' cd $(top_srcdir) && $(AUTOMAKE) --foreign include/curl/Makefile'; \
464c464
<         $(AUTOMAKE) --gnu include/curl/Makefile
---
>         $(AUTOMAKE) --foreign include/curl/Makefile
```

I checked the debian bookworm docker image; here are the versions installed
- autoconf 2.71
- automake 1.16.5
- libtoolize 2.4.7
- make 4.3
- perl 5.36.0
- git 2.39.2

Comparing to the mail https://curl.se/mail/lib-2024-03/0062.html

>For the most recent curl release, my toolset that I believe might affect the results include:
>
>- autoconf (GNU Autoconf) 2.71
>- automake (GNU automake) 1.16.5
>- libtoolize (GNU libtool) 2.4.7
>- GNU Make 4.3
>- perl v5.38.2
>- git version 2.43.0

The versions seems to match except for the perl and git binaries but those should not make a difference here.

What I can see is diffs like

```diff
<        version:        $progname $scriptversion Debian-2.4.7-5
---
>        version:        $progname $scriptversion Debian-2.4.7-7
```

so I believe even the slight differences in `libtool` versions - even if they're both version 2.4.7 make a difference:
- `2.4.7-5` (debian bookworm), vs
- `2.4.7-7` (release tarball)

---

I don't know why the Microsoft files are changing and I don't have the experience to look into this.

---

Summary: there are still some slight differences between the release tarballs and what I can re-create now with this docker build environment; some we can explain (see above) and some is still unexplained (Microsoft builds, mismatch in datestamp). The autotools generated code seems to differ even though the versions have the same major/minor/patch versions, making it tricky to do any automated comparisons.

vszakats:
> I don't know why the Microsoft files are changing and I don't have the experience to look into this.

Can you tell which files these are and how are they changing? It is their line ending perhaps?

daniel-j-h:
The Microsoft related files I can see in my `diff -qr` are

```
Files local/curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.10/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.10/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.10/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.10/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.20/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.20/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.20/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.20/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.30/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.30/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.30/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.30/src/curl.vcxproj differ
```

Here is `curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj` for my local tarball and release, respectively. Compare
- https://gist.github.com/daniel-j-h/4e323c1c42b2751ebd5a956bf8a997a9#file-local-libcurl-vcxproj-L2560
- https://gist.github.com/daniel-j-h/4e323c1c42b2751ebd5a956bf8a997a9#file-remote-libcurl-vcxproj-L2560

I don't know how these files are getting generated, if the entries there can be of different order, or if the build command differs.

vszakats:
It seems indeed that `ItemGroup` list items are not alpha-sorted.

There is a generator logic in `./Makefile.in` (and also in `projects/generate.bat` when run on Windows natively).

bagder:
`maketgz` invokes `make vc-ide`, and in `Makefile.am` there is a complicated awk script that generates the project files using the variables (lists of file names really) set in the Makefile. 

This script sorts the lists. I have not figured out why it would create output not sorted...

bagder:
> remote/curl-8.7.1/docs/curl-config.1

This file is included in the tarball by mistake. The date in this file is the date when it was generated, the day before the release. I will make a PR to remove it from the dist, it should get generated in the build.

bagder:
@daniel-j-h maybe you can submit the maketgz fix separately, to allow us to gradually and step by step fix the nits you have identified

bagder:
Ah, the image needs `gawk` installed as well! For `make vc-ide` to work.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM debian:bookworm-slim@sha256:993f5593466f84c9200e3e877ab5902dfc0e4a792f291c25c365dbe89833411f
+# autoconf 2.71
+# automake 1.16.5
+# libtoolize 2.4.7
+# make 4.3
+# perl 5.36.0
+# git 2.39.2
+
+RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
+    build-essential make autoconf automake libtool git perl zip zlib1g-dev && \
```
bagder, line 27:
```suggestion
    build-essential make autoconf automake libtool git perl zip zlib1g-dev gawk && \
```

daniel-j-h:
> Ah, the image needs `gawk` installed as well! For `make vc-ide` to work.

Good catch!! I just added it to the docker environment and re-build the release tarballs.

It looks like there are still unsorted lists in there, though.

For example compare the files in this list
1. https://gist.github.com/daniel-j-h/936a722c2c2c8b3afa3a546e4fca44ac#file-local-libcurl-vcxproj-L2389 (this branch)
2. https://gist.github.com/daniel-j-h/936a722c2c2c8b3afa3a546e4fca44ac#file-remote-libcurl-vcxproj-L2389 (release tarball)

These differences still exist even after re-building everything with `gawk` in the docker environment in the Microsoft files

```
Files local/curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.10/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.10/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.10/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.10/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.20/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.20/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.20/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.20/src/curl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.30/lib/libcurl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.30/lib/libcurl.vcxproj differ
Files local/curl-8.7.1/projects/Windows/VC14.30/src/curl.vcxproj and remote/curl-8.7.1/projects/Windows/VC14.30/src/curl.vcxproj differ
```

---

On a positive note: if we figure out the Microsoft file differences then the only differences are due to the slight mismatch in tools outlined in https://github.com/curl/curl/pull/13250#issuecomment-2032966496 - I'm wondering, are you on debian bookworm and using the packages from apt? Or did you install specific packages locally yourself from source?

There I can see three ways forward: either we track your specific setup in this Dockerfile, or we standardize on tools from a distribution such as debian bookworm, or maybe you are willing to do releases in the self-contained docker environment.

bagder:
> if we figure out the Microsoft file differences

We should figure that out and make them stable.

> are you on debian bookworm and using the packages from apt

I'm on Debian unstable/sid and I use packages from apt. but...

> maybe you are willing to do releases in the self-contained docker environment

I am prepared to use this dockerfile when releasing tarballs going forward to ease the process for people who want to reproduce them. I can probably also make the daily snapshot builds use the same thing.

bagder:
#13294 is a take to simplify the project file generation in the hope that the sorting was a reason for the diff. It was nonetheless unnecessary.

vszakats:
`date -d` needs GNU Make.

`-d` is not POSIX and e.g. BSDs don't support it (or with a different meaning, depending on flavour.)

bagder:
< date -d needs GNU Make.

I don't think this is a problem. This docker image is using Debian. And I have used Debian to do curl release for the last twenty years or so.

bagder:
@daniel-j-h this merge conflicts now after @vszakats's maketgz cleanups. Can you fix and force-push?

bagder:
When I build a release with this image all the files it makes get owned by root. Very inconvenient. Is there an easy fix to make them own by my regular user?

bagder:
Ah yes, I need to pass `-u $(id -u):$(id -u)` to docker run.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+#   docker build -t curl/curl .
+#
+# Then run commands from within the build environment, for example
+#
+#   docker run --rm -v $(pwd):/usr/src -w /usr/src curl/curl autoreconf -fi
+#   docker run --rm -v $(pwd):/usr/src -w /usr/src curl/curl ./configure --without-ssl --without-libpsl
+#   docker run --rm -v $(pwd):/usr/src -w /usr/src curl/curl make
+#   docker run --rm -v $(pwd):/usr/src -w /usr/src curl/curl ./maketgz 8.7.1
+
+FROM debian:bookworm-slim@sha256:993f5593466f84c9200e3e877ab5902dfc0e4a792f291c25c365dbe89833411f
```
bagder, line 18: Forgive a docker newbie, but where does the data for this line come from?

jzakrzewski: You can find it when you pull the image:
```
docker pull debian:bookworm-slim
bookworm-slim: Pulling from library/debian
8a1e25ce7c4f: Pull complete
Digest: sha256:ccb33c3ac5b02588fc1d9e4fc09b952e433d0c54d8618d0ee1afadf1f3cf2455
Status: Downloaded newer image for debian:bookworm-slim
docker.io/library/debian:bookworm-slim
```
But also probably using `docker manifest`.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# git 2.39.2
+
+RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
+    build-essential make autoconf automake libtool git perl zip zlib1g-dev gawk && \
+    rm -rf /var/lib/apt/lists/*
+
+# >>> from datetime import datetime
+# >>> int(datetime(2024, 3, 27, 9).timestamp())
+# 1711526400
+ENV LC_ALL=C TZ=UTC SOURCE_DATE_EPOCH=1711526400
```
vszakats, line 33:
```suggestion
ENV SOURCE_DATE_EPOCH=1711526400
```
`maketgz` should take care of `LC_ALL` and `TZ` now, do we still need them here as well?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# perl 5.36.0
+# git 2.39.2
+
+RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
+    build-essential make autoconf automake libtool git perl zip zlib1g-dev gawk && \
+    rm -rf /var/lib/apt/lists/*
+
+# >>> from datetime import datetime
+# >>> int(datetime(2024, 3, 27, 9).timestamp())
+# 1711526400
```
vszakats, line 32:
```suggestion
# >>> from datetime import datetime, timezone
# >>> int(datetime(2024, 3, 27, 23, 54, 59, tzinfo=timezone.utc).timestamp())
# 1711583699
```
Settle TZ and use a timestamp which makes it easier to spot when messing up the timezone.

Dockerfile
vszakats, line 33:
```suggestion
# >>> from datetime import datetime, timezone
# >>> int(datetime(2024, 3, 27, 23, 54, 59, tzinfo=timezone.utc).timestamp())
# 1711583699
ENV SOURCE_DATE_EPOCH=1711583699
```
Settle TZ and use a timestamp which makes it easier to spot when messing up the timezone.

Also, `maketgz` should take care of `LC_ALL` and `TZ` now, do we still need them here as well?

Dockerfile
xquery: alternately one might be able to use the  existing curl-dev-debian image as a base (in curl ghcr)

Dockerfile
bagder: I don't think we should set a fixed timestamp in the Dockerfile. I think we can have the comment there to show *how* to set the timestamp if wanted, but if we leave out the stamp there we can use this file unmodified for doing daily snapshots and the real releases

Dockerfile
bagder: what's the practical difference ?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+FROM debian:bookworm-slim@sha256:993f5593466f84c9200e3e877ab5902dfc0e4a792f291c25c365dbe89833411f
+# autoconf 2.71
+# automake 1.16.5
+# libtoolize 2.4.7
+# make 4.3
+# perl 5.36.0
+# git 2.39.2
+
+RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
```
arianvp, line 26: If we want to deterministically install the same package versions I think we need to drop the `apt-get update` here? Otherwise you'll get whatever version was latest in bookworm at the time you run the docker build command. Which is not reproducibile.. Instead update the `FROM` clause whenever we want to bump the versions for reproducibility.

Dockerfile
vszakats: Yes, it's even better.

Or derive it automatically from the git commit date for daily runs. That can also be done outside the `Dockerfile`.

Dockerfile
lrvick: 1. This still relies in some non-reproducible packages in debian.
2. You are going to still need tools that don't ship with the minimal debian container
3. Packages installed via apt-get update get purged regularly so you would need to use the debian archive
4. The debian archive is very slow, very unreliable, and very bandwidth constrained. You also need to write a lot of custom tooling to hash-lock exact package versions or apt -will- try to grab whatever the mirror hands it.
5. Signatures on packages pinned today will expire and break apt

I have gone all the way down this road, and realized Debian is just a non-starter for practical/maintainable reproducible builds. I am about to push another PR with a stagex-based Containerfile that is deterministic and full source bootstrapped.

lyxell: Can we be more concrete here? It would be fine if the tool versions differ as long as they still produce the same tarball. Which of the steps taken to produce the tarballs will be give different results when the tool versions differ?

lrvick: A different version of automake autoconf tar, zip, bzip2 etc could all end up causing different results in the future.

I just pushed a draft PR with an example that uses a "from scratch" container and an explicit set of fixed/deterministic/multi-signed dependencies:

https://github.com/curl/curl/pull/13338

Does not have zip support (until next week as we need to do a release cycle to package it) but otherwise should be good to go, and can be easily upgraded to build any deterministic binaries of curl if desired as well.

lyxell: How about something along the lines of this?
```
FROM debian:bookworm-slim@sha256:993f5593466f84c9200e3e877ab5902dfc0e4a792f291c25c365dbe89833411f

RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
    build-essential=12.9 \
    make=4.3-4.1 \
    autoconf=2.71-3 \
    automake=1:1.16.5-1.3 \
    ...
```

daniel-j-h:
Sorry for the slow responses I've been out sick these days, hope to get better by the weekend. Some quick responses below

> When I build a release with this image all the files it makes get owned by root. Very inconvenient. Is there an easy fix to make them owned by my regular user?
> Ah yes, I need to pass `-u $(id -u):$(id -u)` to docker run.

There are two sides to this
1. When you `docker run` you can pass your host user id and group id to the container
2. In the Dockerfile we can create a user id and a group id and switch to that for all subsequent commands

        RUN groupadd --gid 1000 dev \
         && useradd --uid 1000 --gid dev --shell /bin/bash --create-home dev

        USER dev:dev

https://docs.docker.com/reference/dockerfile/#user

The second step is optional and is only needed because some programs expect an actual user and group (and e.g. home dir) to exist when they see the uid and gid that you pass in. And if you want to map your host user and group to the container's user and group it's best to add `ARG` we can change when we run it end to end, because on your host user id 1000 might not exist.

Dockerfile
daniel-j-h: I got this digest by going to

https://hub.docker.com/_/debian

checking tags

https://hub.docker.com/_/debian/tags

and filtering for bookworm-slim

---

I didn't know there is already a curl docker image; I read the original mail listing tools and operating system and simply went for debian bookworm. Happy to change this.

Dockerfile
daniel-j-h: Agree with you folks, when building the user needs to supply `SOURCE_DATE_EPOCH`.

I'll get to this when I rebase and also folks in the `LC_ALL` and `TZ` changes that are now in master.

lrvick: It is pretty common to set SOURCE_DATE_EPOCH=1 to simply disable timestamps for all intents and purposes which is what we do in stagex globally so we can reproduce packages at any point in the future from any ref easily.

I suggest `ARG SOURCE_DATE_EPOCH=1` so you can override it at build time with `--build-arg SOURCE_DATE_EPOCH=1234` else it defaults to 1.

Dockerfile
lrvick: I once asked that same exact question. It led me down a several month rabbit hole I would save you from if I can.

1. sub-dependencies could still shift to latest versions where supported unless you do this for -all- of them, thus you end up needing tools to generate a huge lockfile of every single package like this: https://git.distrust.co/public/airgap/src/branch/main/config/toolchain/packages-x86_64.list
2. As soon as newer versions of dependencies come out, the ones you are pinning now will get pulled from mirrors, so you need to have scripts to automatically pull, pin, then pivot to using the debian archive like this: https://codeberg.org/distrust/toolchain/src/commit/2560e543a1871b86cb1c29b31ce9141eaec94e8c/scripts/packages-update
3. Next you will find the signing keys expire every day, and apt will refuse to install those packages. You can disable signing in apt, but then you -must- implent your own hash locking strategy. You end up having to automatically generate a hashlock file like this: https://git.distrust.co/public/airgap/src/branch/main/config/toolchain/package-hashes-x86_64.txt
4. You will then realize that the debian archive provide files at dialup speeds, if it works at all, so you will need to mirror several GB of old versions of packages yourself with Git LFS or similar.
5. Then you will realize updating all of this is a real pain, and you need to make a whole repo of make helpers and scripts to maintain it all: https://git.distrust.co/public/toolchain
6. Then you realize that all of this -works- but is awful to audit and maintain and no one wants touch projects that use it (I deployed this setup for a few orgs as it was the -only- option for reproducible builds from a mostly reproducible/signed lniux distro.)
7. You then face facts that debian only ships very old versions of packages, and in most cases single maintainers have God keys and complete trust to ship whatever they want, and many critical packages are still not reproducible so trust in the whole tree still comes down to trusting any one out of 100+ people is not compromised or malicious.

antonleviathan: @lyxell That would work temporarily, but would not remain deterministic as specific Debian package versions get purged and moved to archive. Then you have to either download and maintain the install files so that you still have them available, or switch to using archive links which are painfully slow. You also inherit a ton of dependencies from Debian that you really don't need. The PR @lrvick suggested is easier to maintain, has a smaller dependency footprint, and all of the deps that you do use are built deterministically and multi-reproduced. Just for full transparency I contribute to [StageX](https://codeberg.org/stagex/stagex) which Lance suggested in his PR.

antonleviathan: I can attest to what Lance is saying. I went through the process of making a build deterministic using Debian and it was a painful process. StageX makes it a lot easier, and the other benefit that hasn't been mentioned is that the whole StageX build-chain is verifiable, deterministic, and multi-reproduced. So what you end up with is that the tool-chain, the packages, and then your software that's built with it are all deterministic, and you get to decide if you want to also multi-reproduce what's being built using this setup.

lyxell: I can imagine that building binaries in a deterministic fashion on Debian will not work in practice. However, I imagine that building tarballs where all of the files of the tarballs will be plain text files should be feasible even if the tool versions used to build those tarballs may vary.

That is, in practice I would expect tools like `perl`, `gzip` and `tar` to produce byte-by-byte identical outputs for the commands listed in `maketgz` the last decade or so on Debian, under all versions. But this is just me guessing, I would be happy to be proven wrong.

antonleviathan: If you only run `maketgz` as it is implemented right now that would likely work as you described. Side note: `gzip` has likely had optimizations over the years. `automake` changes its mind on stuff from one version to the next so that concern is valid. The natural follow up to this is to have `make` in the curl repo use the container to actually produce deterministic binaries of curl as a baseline and that has to use stagex or something like it to achieve determinism.

vszakats: [ We do deterministic binaries in the sub-project curl-for-win: https://github.com/curl/curl-for-win

It is based on Debian versioned images, with the caveat discussed above, where `apt update`/`install` may break strict reproducibility of the tools used. Code dependencies are all built from source reproducibly. ]

nurupo:
This assumes that the docker image used is not compromised itself, either via distro or via the docker image maintainers. If possible, it would be nice to be able to generate the tarball using different distros dockerized by different groups of people, e.g. Debian and Alpine, or perhaps Fedora, and compare them. Not sure if a tarball can be made reproducible like that though.

bagder:
> The tool I'm most worried about is automake though

autoconf, automake and libtool are most likely the tools which versions are most relevant. The others most probably not at all.

I also don't want a "trust competition". Reproducible builds are primarily interesting and important in the short term: to allow users to easily verify the last few releases (say within the last year or so).

An obvious advantage of using Debian as a base for this is familiarity, trusting the brand, its processes and that they also patch packages for functionality. Going with a separate environment means leaning on someone else's process. Possibly a less familiar one.

bagder:
and also: we have removed the need for awk in maketgz (since #13311) so `gawk` can be removed again :smile:

lrvick:
> Reproducible builds are primarily interesting and important in the short term: to allow users to easily verify the last few releases (say within the last year or so).

A breaking change resulting from building from latest Debian packages could come at any time. Maybe in 5 years, maybe in a week.

My arguments here would be far stronger if we were talking about generating deterministic binaries. That said, binaries (and the archives that generate them) can stick around and see some level of use for decades. There are (sadly) many mission critical embedded devices being used today with very old versions of curl on them and we can expect versions installed today will likewise be used many years in the future.

When it comes to security auditing and forensics it becomes relevant to recreate the exact supply chains of binaries created years ago, that resulted in recent harm, to rule out or identify supply chain attacks or obscure vulnerabilities that only show up depending on the tools used to build them.

My efforts here and elsewhere are about trying to get widely used projects to trend towards releasing using methods that are always predictably deterministic and long-term-auditable. That of course does not have to be with Stagex, though with my bias hat as firmly in place, I can say it does make it much simpler to audit and maintain IMO.

If using Debian is a hard requirement, I would suggest using scripts like https://github.com/reproducible-containers/repro-sources-list.sh or the setup I wrote here in https://git.distrust.co/public/toolchain to produce a set of locked hashes of a given snapshot of debian dependencies, then download and install these exact snapshots as I describe above. This path is complex, slow, difficult to maintain, very disk heavy, and definitely a hack to force Debian to do something it was not meant do do, but it does -work- and is my go-to when projects mandate Debian out of familiarity or other reasons. 

We did have this debian-locking approach to determinism audited by Cure53 as we used it in AirgapOS, though we are now moving that project to stagex to reduce maintenance burden: https://git.distrust.co/public/airgap/src/branch/main/audits/cure53-2020.pdf

bagder:
> My arguments here would be far stronger if we were talking about generating deterministic binaries

Those who ship binaries might appreciate reproducible source tarballs, but they need to take care of the binary producing part themselves.

> always predictably deterministic and long-term-auditable

Sure, that's ideal. But I realize we may need to make priorities and take decisions. Making it dead easy to verify source packages twenty years later is not something I think it's worth spending many brain cells on. Diminishing returns and all that.

The people who are stuck on ancient versions are also probably the ones *least* likely to actually want to verify these things as they are clearly not very concerned...

> If using Debian is a hard requirement

I don't believe it is. I think we are still in a process where we asses our options and their pros and cons.

vszakats:
Considering the (continuous) efforts necessary for keeping the pre-built autotools bits deterministic (and then verifying them), I'd risk saying that a more efficient alternative is to offer a source tarball that doesn't contain pre-generated autotools files. Those would require the builder to run `autoreconf -fi` (is there any downside of that?), or use CMake, which works out of the box.

A trivial way to do this is to GPG sign the tarballs created automatically by GitHub:
https://github.com/curl/curl/archive/refs/tags/curl-8_7_1.tar.gz
https://github.com/curl/curl/archive/refs/tags/curl-8_7_1.tar.zip

This has the downside of missing pre-generated manual files, and building those locally requires Perl.

Fixing this needs a little bit more involved solution, but much less so than the dance with reproducible OS environments, IMO.

bagder:
I agree that we could remove the need for a lot of this by not generating anything at all for releases. I just don't think it would benefit our users.

1. We have users that build curl from configure, especially on older systems, because they then don't need to install additional build software (like cmake or autotools). Forcing users to install the autotools to use configure will push workloads to users and there will be follow-up squeaks and problems.

2. People have many times over the years expressed their desire to avoid having to install perl to build curl and we have worked on preserving this ability. Only very recently even.

3. The GitHub tarballs are not reproducible are they?

Sacrificing build convenience for easier tarball verification would go against what we know plenty want in favor of something virtually nobody has asked for.

vszakats:
> 3. The GitHub tarballs are not reproducible are they?

They are (for now). Last year this broke for a few days due an update they did. The fallout was large enough to have it reverted. They then guaranteed it then for 1 year, meaning: Not guaranteed, but they still are reproducible.

https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/

The next best thing is cloning a specific Git hash. Which needs Git of course, and using SHA1 hashes known for collisions. Also easy to mess-up by using the unencrypted git:// protocol. Is it possible to double-check a Git clone, against a known state?

> Sacrificing build convenience for easier tarball verification would go against what we know plenty want in favor of something virtually nobody has asked for.

I understand.

Adding that it's probably expected that cmake, or autotools might be installed already on many systems, just like (GNU) make is. Or as easy to install as the latter. But yeah, might take time till some of these becomes as ubiquitous. (Maybe worthy a survey question?)

Dockerfile
eli-schwartz: The only package versions that matter at all here are the versions of the autotools packages (autoconf, automake, libtool...).

Simply print the debian (apt) versions of this exceedingly small number of packages to a reproducible-manifest.txt inside the tarball before you tar it up. Problem solved.

@lrvick As far as rebuilding all packages from scratch goes, it appears you have reinvented the Gentoo Linux distro except NIH. Which, incidentally, also allows you to record an exact git repository hash of the Gentoo tree and guarantee all packages in your container use that and nothing else.

But all this is the mootest of moot points, because if you want to guarantee that everyone is using the exact same docker container to reproduce the tarballs, then you simply publish the container you used -- as mentioned above. This is *very* reproducible as you're uploading @bagder's actual (virtual) laptop used to make the release, and anyone can *download @bagder's (virtual) laptop and use it to create the tarball* (and check that it's the virtual laptop created in CI).

eli-schwartz:
> Adding that it's probably expected that cmake, or autotools might be installed already on many systems, just like (GNU) make is. Or as easy to install as the latter. But yeah, might take time till some of these becomes as ubiquitous. (Maybe worthy a survey question?)

The people who are appreciative of using configure on older systems because it doesn't require installing anything, may be people on operating systems that don't have a C++ compiler, so no cmake. They may have a non-GNU make, or a GNU make from two decades ago. They may not have autotools at all, or have autotools from two decades ago that can't be used to regenerate the curl configure.ac -- a major stated advantage of autotools is that you can make a dist tarball on bleeding edge distros, but still have that work on very old systems without modern autotools, as long as it has *a* posix shell make and general utilities.

Continuing to offer autotools but requiring users of autotools to generate the configure script themselves seems like a very mixed message.

Dockerfile
lrvick: A decision has already been made here so I am merely responding to some potential misconceptions above for the sake of anyone following this thread.

> Simply print the debian (apt) versions of this exceedingly small number of packages to a reproducible-manifest.txt inside the tarball before you tar it up. Problem solved.

That will require the user that wishes to verify to track down all those versions, and all their dependencies, and assemble them together again. Version numbers are not good enough, as those keys expire and are rotated often. You would at a minimum need to publish the entire installed apt tree in the container with /hashes/ of every .deb file so they can be retrieved from the Debian archive and directly installed in a reproduction container. This will be very arduous for a user still without any scripted help, but it would at least be a path. I have written a few tools to help with this mentioned elsewhere. It is a terrible reproduction path, but it does work.

> @lrvick As far as rebuilding all packages from scratch goes, it appears you have reinvented the Gentoo Linux distro except NIH.

Gentoo linux is not reproducible, quorum-signed, full-source-bootstrapped, or OCI native. It is true that we do in fact compile everything from source, but we publish our binaries and expect users to use those, so that makes us closer to traditional linux distros like Debian than Gentoo.

The only other deterministic full-source-bootstrapped Linux distro that exists to my knowledge is GNU Guix.They however do not publish reproducible container images, and do not do quorum signing. They are also glibc based and optimized for end user desktop use, which comes with dramatically higher attack surface and complexity than what is required with a goal of simply securely building software. Still, Guix would be the best distro to compare us to.

If you are not familiar with full-source-bootstrapping, Guix has a great writeup as they were the first distro to do it. Stagex was the second.

https://guix.gnu.org/en/blog/2023/the-full-source-bootstrap-building-from-source-all-the-way-down/
 
> If you want to guarantee that everyone is using the exact same docker container to reproduce the tarballs, then you simply publish the container you used 

Agreed. That is the reason to use containers, but the context of this are trying to minimize the chance of supply chain attacks. We are trying to solve for trust. 

If the binary container image is compromised by the person with the publishing API key, then anything downstream from that container image is also compromised. See the XZ attack and Ken Thompson's Reflections on Trusting Trust.

You can however pull down the "musl" and "autoconf" binary stagex container images and directly use them together. You can also verify multiple people built them for you and got the same exact hash, via their PGP signatures so you don't have to build it if you trust they are not colluding. You also can -choose- to build them yourself and add your own signature if you don't trust those folks.

With Debian OTOH, they build their published docker images in a non-deterministic way and do not sign them, so you have no evidence it is free of tampering. Very much like the XZ situation. Using Docker Debian unfortunately just moves the problem up one layer.

vszakats:
> They may not have autotools at all, or have autotools from two decades ago that can't be used to regenerate the curl configure.ac

Out of curiousity, is there any other autotools compatibility requirement for `configure.ac`, besides being v2.59 (from 2003-11-06) or newer?: https://github.com/curl/curl/blob/49f83c30e43749752072c65a43d553eb015453d1/configure.ac#L26

It seems a little arbitrary to take for granted both a compatible make tool and C compiler, but not a compatible autotools. Though I admit to have no data or even anecdotal info about this. I'd have expected autotools to have lived together with make and C compilers way back. CMake is much newer of course, and the minimum required 3.7 (2018-12-03), can be a limitation indeed. (even without this limitation, CMake really only seems to be going back a decade from now as 'ubiquitous'.)

Either way, I understand that pre-packaging autotools stuff is something expected, period.

Also those worried about the state of pre-packaged stuff (and have a compatible autotools), can always `autoreconf`.

eli-schwartz:
> It seems a little arbitrary to take for granted both a compatible make tool and C compiler, but not a compatible autotools. Though I admit to have no data or even anecdotal info about this. I'd have expected autotools to have lived together with make and C compilers way back.

Both make and C are much older than autotools -- and are required by POSIX. However, I'll admit I didn't check what the minimum autotools requirement for curl is before commenting. I wonder if it's still tested against that. :thinking:

xc-am-iface.m4 does offer support for "versions of automake older than 1.14" though it's not clear what the actual minimum is (nor, indeed, whether that is tested either).

Dockerfile
eli-schwartz: All of this is completely irrelevant, since the curl project is trying to solve reproducible curl release tarballs, not reproducible linux distros as general-purpose computing platforms.

> That will require the user that wishes to verify to track down all those versions, and all their dependencies, and assemble them together again. Version numbers are not good enough, as those keys expire and are rotated often. You would at a minimum need to publish the entire installed apt tree in the container with /hashes/ of every .deb file so they can be retrieved from the Debian archive and directly installed in a reproduction container.

No, again, version numbers of the autotools set of packages alone (and not any dependencies) are sufficient since they allow one to determine the correct version of autotools packages to install from ANY source, modulo debian-specific patches publicly recorded in the Debian VCS for that small handful of packages.

No hashes necessary, no dependencies necessary, no access to .debs necessary.

It is slightly more awkward to install a local autotools toolchain with debian patches on your choice of personally trusted computing systems than to install a prebuilt .deb, but not much harder.

Also "as those keys expire and are rotated often" does not sound like a practical objection to me. If you use old stuff signed by crypto keys which are old, then naturally you backtrace the trust source for those keys, and subsequently backdate the verification routine to respect keys that claim to have expired in the past, but from the perspective of what you're reproducing, in the future.

> If you are not familiar with full-source-bootstrapping, Guix has a great writeup as they were the first distro to do it. Stagex was the second.

Sure I'm familiar with it. I don't think that's the concern the curl project has. It will also be more interesting to *me personally* when it no longer relies on untrusted binaries such as idk, the docker daemon, guile-bootstrap, or any other potentially malicious tools used to host the computing platform that runs the full-source bootstrap.

> If the binary container image is compromised by the person with the publishing API key, then anything downstream from that container image is also compromised. See the XZ attack and Ken Thompson's Reflections on Trusting Trust.

I've seen both quite well, thanks. Assuming a non-compromised docker daemon, it really is not hard to rebuild the container image using, say, https://snapshot.debian.org/ and test that your binary container image produces the same tarball. Again, the container doesn't have to be byte-for-byte reproducible, it simply has to be some form of instructions that others can follow to yield the same... curl release tarball.

No need to trust curl devs to publish a non-compromised spinoff container.

> With Debian OTOH, they build their published docker images in a non-deterministic way and do not sign them, so you have no evidence it is free of tampering. Very much like the XZ situation. Using Docker Debian unfortunately just moves the problem up one layer.

Sounds like a market opportunity for someone to build deterministic debian docker images that are signed.

The underlying .deb packages are reproducible: https://tests.reproducible-builds.org/debian/reproducible.html

(Feel free to run your own builders to verify that!)

dfandrich:
> However, I'll admit I didn't check what the minimum autotools requirement for
> curl is before commenting. I wonder if it's still tested against that. 🤔

INTERNALS.md says these are the minimums:

 - GNU Libtool  1.4.2
 - GNU Autoconf 2.59
 - GNU Automake 1.7
 - GNU M4       1.4

But, I doubt people actually test with these very often.  The linux-old CI
environment (Debian stretch) contains these:

 - GNU Libtool  2.4.6
 - GNU Autoconf 2.69
 - GNU Automake 1.15
 - GNU M4       1.4.18

Those are all newer (mostly much newer) than the documented minimum, so it's
probably worth adding an autoconf build there to at least help ensure
compatibility with somewhat older versions.

vszakats:
Thanks for the minimums info.

Dates for these:

- GNU Libtool  1.4.2 (2001-09-11)
- GNU Autoconf 2.59 (2003-11-06)
- GNU Automake 1.7 (2002-09-25)
- GNU M4       1.4 (2007-09-21)

I wonder what might be the oldest GNU C (or other brand) compiler able to compile curl. [3.0.0](https://github.com/gcc-mirror/gcc/tree/releases/gcc-3.0.0) to [3.3.0](https://github.com/gcc-mirror/gcc/tree/releases/gcc-3.3.0) were released in the above timeframe. 2.9.5 is from 1999-07-29.

It'd be an interesting experiment to test for the autotools minimums (linux-old CI already revealed two CMake build issues). If there exists any re-usable online infrastructure for that.

bagder:
> I wonder what might be the oldest GNU C (or other brand) compiler able to compile curl. 

A C89 compliant compiler should still be able to build curl even if the compiler is from the 90s. We stick to C89 partly because of that.

And if it does not due to some mistake somewhere, it can't be very important since nobody has reported it...

bagder:
> It seems a little arbitrary to take for granted both a compatible make tool and C compiler, but not a compatible autotools.

It is not arbitrary. It is a design choice and how autotools always worked. Users everywhere have always been able to install software widely without having autotools themselves. You install autotools only when you want to develop software, while if you want to just build and install software autotools may not be installed (or up-to-date).

Thus, suddenly asking that people should install and use autotools is wrong and will cause a lot of friction.

daniel-j-h:
Hey @bagder I just made some changes here
1. rebased on master to get the latest maketgz changes
2. added a dedicated user so that the container doesn't run on root and we can have the same user permissions on the host and in the container
3. added build arguments for the timestamp and also user and group id

To build

    docker build \
      --build-arg SOURCE_DATE_EPOCH=1711526400 \
      --build-arg UID=$(id -u) \
      --build-arg GID=$(id -g) \
      -t curl/curl .

to run

    docker run --rm -it -u $(id -u):$(id -g) \
      -v (pwd):/home/dev/curl -w /home/dev/curl \
      curl/curl bash

This gets you into the container based on the image you built in the previous step. Because we created a dedicated user we work in its home directory and map the host's user and group ids to the container's user.

Like I said in https://github.com/curl/curl/pull/13250/files#discussion_r1559389357 it's not perfect but I'd still love your feedback here and if it's worth getting this in. Wanna give it a final look?

bagder:
I'm using the following script to build a test release using this Dockerfile and it works fine:

~~~shell
version="${1:-}"

if [ -z "$version" ]; then
  echo "Specify a version number!"
  exit
fi

user="$(id -u):$(id -u)"

make distclean
docker build \
       --build-arg SOURCE_DATE_EPOCH=$(date -u +%s) \
       --build-arg UID=$(id -u) \
       --build-arg GID=$(id -g) \
       -t curl/curl .

run="run --rm -it -u $(id -u):$(id -g) -v $(pwd):/usr/src -w /usr/src curl/curl"

docker $run autoreconf -fi
docker $run ./configure --without-ssl --without-libpsl
docker $run make -sj8
docker $run ./maketgz $version

bagder:
Thanks!

bagder:
The tarball generator script using this Dockerfile is in #13388

Dockerfile
lrvick: > Also "as those keys expire and are rotated often" does not sound like a practical objection to me. If you use old stuff signed by crypto keys which are old, then naturally you backtrace the trust source for those keys, and subsequently backdate the verification routine to respect keys that claim to have expired in the past, but from the perspective of what you're reproducing, in the future.

If keys are revoked because they used weak, vulnerable, algos or are already compromised, or they were by a maintainer that is no longer trusted (like the xz maintainer) then those keys could be used to sign a malicious alternative to the package you are actually requesting, with the same version number, and apt would not know the difference. This is why hash locking is an absolute necessity when using old packages where the signatures may no longer be trustworthy.

> It really is not hard to rebuild the container image using, say, https://snapshot.debian.org/

This actually is surprisingly hard. Debian shapshots are hosted by a single deployment that is most of the time crippled to dialup speeds in my experience, and the experience of virtually every user that has used projects of mine that relied on it. Also, for the above reasons about not being able to trust expired/revoked keys, you have to pull the .deb files manually without installing, hash verify them, then install from that folder as a local mirror. For these two reasons, in the many reproducible build projects I worked with based on Debian, we all ended up forced to use our own git LFS mirrors of .deb packages which was a major pain.

> It is slightly more awkward to install a local autotools toolchain with debian patches on your choice of personally trusted computing systems than to install a prebuilt .deb, but not much harder.

To be fair to this point, for your very specific use case here, you could probably get away with using generic latest debian and then build your own autotools, and then use your own autotools to build the tar. That is of course if any signed/trusted/reproducible debian base images actually existed :/

Issue #13248, opened 2024-03-31
8.7.1: rustls fails to pull from GitHub


Kangie:
### I did this

I have used GitHub as an example here, but GitLab and Bitbucket also failed in the same way.

```
(chroot) monolith /tmp # GIT_TRACE_CURL=1 GIT_CURL_VERBOSE=1 git clone https://github.com/kangie/kangie-tools.git
Cloning into 'kangie-tools'...
05:45:18.595988 http.c:843              == Info: Couldn't find host github.com in the .netrc file; using defaults
05:45:18.596557 http.c:843              == Info: Host github.com:443 was resolved.
05:45:18.596568 http.c:843              == Info: IPv6: (none)
05:45:18.596572 http.c:843              == Info: IPv4: 20.248.137.48
05:45:18.596588 http.c:843              == Info:   Trying 20.248.137.48:443...
05:45:18.647927 http.c:843              == Info: Connected to github.com (20.248.137.48) port 443
05:45:18.647980 http.c:843              == Info: ALPN: curl offers http/1.1
05:45:18.648113 http.c:843              == Info: rustls_connection wants us to write_tls.
05:45:18.648156 http.c:843              == Info: Curl_socket_check: reading would block
05:45:18.703955 http.c:843              == Info: rustls_connection wants us to read_tls.
05:45:18.704143 http.c:843              == Info: rustls_connection wants us to write_tls.
05:45:18.704183 http.c:843              == Info: rustls_connection wants us to read_tls.
05:45:18.704281 http.c:843              == Info: rustls_connection_process_new_packets: invalid peer certificate: BadSignature
05:45:18.704296 http.c:843              == Info: Closing connection
fatal: unable to access 'https://github.com/kangie/kangie-tools.git/': rustls_connection_process_new_packets: invalid peer certificate: BadSignature
(chroot) monolith /tmp # curl --version
curl 8.6.0 (x86_64-gentoo-linux-musl) libcurl/8.6.0 rustls-ffi/0.10.0/rustls/0.21.0 zlib/1.3 c-ares/1.25.0 libpsl/0.21.5
Release-Date: 2024-01-31
Protocols: dict file ftp ftps http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp smtp smtps tftp
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Largefile libz PSL SSL threadsafe UnixSockets


[...]

PR #13251, opened 2024-03-31
m4: reposition USE_RUSTLS="yes" for pkg-config 

cpu:
It's necessary to set this var to "yes" _after_ AC_DEFINE and AC_SUBST in order for a later `test` to pass so that `check_for_ca_bundle=1` ends up being set. This is in turn required for the default CA certificate bundle to be set when building w/ rustls & pkg-config.

Before the pkg-config changes [the success case configuration](https://github.com/curl/curl/blob/b564a5f5d5280387a88025c5f90260017847add4/m4/curl-rustls.m4#L55-L58) looked like:

1. AC_DEFINE...
2. AC_SUBST...
3. RUSTLS_ENABLED=1
4. USE_RUSTLS=yes

I think that while that order of operations was preserved in https://github.com/curl/curl/commit/647e86a3efe1eea7a2a456c009cfe1eb55fe48eb, in https://github.com/curl/curl/commit/9c4209837094781d5eef69ae6bcad0e86b64bf99 the order changed to:

1. USE_RUSTLS="yes"
2. AC_DEFINE...
3. AC_SUBST...
4. RUSTLS_ENABLED=1

This in turn seems to result in no default CA certificate bundle being set when using rustls & pkg-config. Moving the `USE_RUSTLS` back to the position it was before seems to resolve the issue.

Resolves https://github.com/curl/curl/issues/13248

### DIFF
--- a/m4/curl-rustls.m4
+++ b/m4/curl-rustls.m4
@@ -137,10 +137,10 @@ if test "x$OPT_RUSTLS" != xno; then
       dnl additional libs may be necessary.  Hope that we
       dnl don't need any.
       LIBS="$SSL_LIBS $LIBS"
-      USE_RUSTLS="yes"
       ssl_msg="rustls"
       AC_DEFINE(USE_RUSTLS, 1, [if rustls is enabled])
       AC_SUBST(USE_RUSTLS, [1])
+      USE_RUSTLS="yes"
       RUSTLS_ENABLED=1
       test rustls != "$DEFAULT_SSL_BACKEND" || VALID_DEFAULT_SSL_BACKEND=yes
     else


m4/curl-rustls.m4
```diff
@@ -141,6 +141,7 @@ if test "x$OPT_RUSTLS" != xno; then
       ssl_msg="rustls"
       AC_DEFINE(USE_RUSTLS, 1, [if rustls is enabled])
       AC_SUBST(USE_RUSTLS, [1])
+      USE_RUSTLS="yes"
```
cpu, line 144: Looking closer this was already being set on L140, so I'm less sure I understand why this fixes the issue.

correabuscar: maybe `AC_DEFINE(USE_RUSTLS` being after... ?

correabuscar: but if I apply only this (patch), it does fix it for me as well, and if I don't apply this at all, it breaks it - so, confirmed working:)

cpu: Before the pkg-config changes [the success case configuration](https://github.com/curl/curl/blob/b564a5f5d5280387a88025c5f90260017847add4/m4/curl-rustls.m4#L55-L58) looked like:

1. AC_DEFINE...
2. AC_SUBST...
3. RUSTLS_ENABLED=1
4. USE_RUSTLS=yes

I think that order of operations was preserved in https://github.com/curl/curl/commit/647e86a3efe1eea7a2a456c009cfe1eb55fe48eb, but in 9c4209837094781d5eef69ae6bcad0e86b64bf99 the order changed to:

 1. USE_RUSTLS="yes"
 2.  AC_DEFINE...
 3.  AC_SUBST...
 4.  RUSTLS_ENABLED=1

cpu: I updated my patch to _move_ the existing `USE_RUSTLS="yes"` instead of adding a second. The commit message & PR title are also updated to reflect that it's the order of operations here that seems to matter.

correabuscar: confirmed working for me with the repositioning too

cpu: Thank you for testing

Kangie:
Works well for me.

bagder:
Wow. Very nice catch!

I figure this happens because `AC_SUBST` clobbers the variable named like that, which I presume shows that it is a bad idea to subst a name that is also used as a local variable but boy, this was subtle.

bagder (approved):

bagder:
Thanks!

Issue #13229, opened 2024-03-30
chunked POST via callback regression in 8.7.X


bagder:
### I did this

(imported issued from https://marc.info/?l=git&m=171175679831214&w=2, authored by Jeff King (@peff))

I noticed some http-related failures in the test suite on my Debian
unstable system, which recently got an upgraded curl package. It looks
like it's related to cases where we use the remote-curl helper in
"connect" mode (i.e., protocol v2) and the http buffer is small
(requiring us to stream the data to curl). Besides just running t5551,
an easy reproduction is:
~~~
  [this works]
  $ git ls-remote https://git.kernel.org/pub/scm/git/git.git | wc -l
  1867

  [this doesn't]
  $ git -c http.postbuffer=65536 ls-remote https://git.kernel.org/pub/scm/git/git.git
  fatal: expected flush after ref listing
~~~
The error message comes from ls-remote itself, which was expecting a
FLUSH packet from the remote. Instead it gets the RESPONSE_END from
remote-curl (remember that in connect mode, remote-curl is just ferrying
bytes back and forth between ls-remote and the server).

It works with older versions of libcurl, but not 8.7.0 (or 8.7.1).
Bisecting in libcurl points to 9369c30cd (lib: Curl_read/Curl_write
clarifications, 2024-02-15).

Running with `GIT_TRACE_CURL=1` shows weirdness on the POST we send to
issue the ls-refs command. With older curl, I see this:

~~~
  => Send header: POST /pub/scm/git/git.git/git-upload-pack HTTP/1.1
  => Send header: Host: git.kernel.org
  => Send header: User-Agent: git/2.44.0.789.g252ee96bc5.dirty
  => Send header: Accept-Encoding: deflate, gzip
  => Send header: Content-Type: application/x-git-upload-pack-request
  => Send header: Accept: application/x-git-upload-pack-result
  => Send header: Git-Protocol: version=2
  => Send header: Transfer-Encoding: chunked
  => Send header:
  => Send data: 14..0014command=ls-refs...
  => Send data: 2a..002aagent=git/2.44.0.789.g252ee96bc5.dirty..
  [and so on until...]
  == Info: Signaling end of chunked upload via terminating chunk.
~~~
But with the broken version, I get:
~~~
  => Send header: POST /pub/scm/git/git.git/git-upload-pack HTTP/1.1
  => Send header: Host: git.kernel.org
  => Send header: User-Agent: git/2.44.0.789.g252ee96bc5.dirty
[...]

PR #13257, opened 2024-04-01
chunked POST forced, disable length check on read callback

icing:
- when an application forces HTTP/1.1 chunked transfer encoding by setting the corresponding header and instructs curl to use the CURLOPT_READFUNCTION, disregard any POST length information.
- this establishes backward compatibility with previous curl versions

Applications are encouraged to not force "chunked", but rather set length information for a POST. By setting -1, curl will auto-select chunked on HTTP/1.1 and work properly on other HTTP versions.

refs #13229

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -2046,8 +2046,13 @@ static CURLcode set_reader(struct Curl_easy *data, Curl_HttpReq httpreq)
       else
         result = Curl_creader_set_null(data);
     }
-    else { /* we read the bytes from the callback */
-      result = Curl_creader_set_fread(data, postsize);
+    else {
+      /* we read the bytes from the callback. In case "chunked" encoding
+       * is forced by the application, we disregard `postsize`. This is
+       * a backward compatibility decision to earlier versions where
+       * chunking disregarded this. See issue #13229. */
+      bool chunked = !!Curl_checkheaders(data, STRCONST("Transfer-Encoding"));
+      result = Curl_creader_set_fread(data, chunked? -1 : postsize);
     }
     return result;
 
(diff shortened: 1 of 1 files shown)


lib/http.c
```diff
@@ -2046,8 +2046,13 @@ static CURLcode set_reader(struct Curl_easy *data, Curl_HttpReq httpreq)
...
         result = Curl_creader_set_null(data);
     }
-    else { /* we read the bytes from the callback */
-      result = Curl_creader_set_fread(data, postsize);
+    else {
+      /* we read the bytes from the callback. In case "chunked" encoding
+       * is forced by the application, we disregard `postsize`. This is
+       * a backward compatibility decision to earlier versions where
+       * chunking disregarded this. See issue #13229. */
+      bool chunked = !!Curl_checkheaders(data, STRCONST("Transfer-Encoding"));
```
bagder, line 2054: I think should instead use `data->req.upload_chunky` ?

icing: No, that does not work. That flag is set after the installed reader is checked for known length.

bagder: Aha! But then maybe the checks should at least be done the same way? 
~~~c
  bool chunked = FALSE;
  char *ptr = Curl_checkheaders(data, STRCONST("Transfer-Encoding"));
  if(ptr) {
    /* Some kind of TE is requested, check if 'chunked' is chosen */
    chunked =
      Curl_compareheader(ptr,
                         STRCONST("Transfer-Encoding:"), STRCONST("chunked"));
  }

peff: Right. It works for Git's case because we are setting `chunked`, but I guess some other caller setting `Transfer-Encoding: gzip` or something would confuse things. My understanding is that you can pass multiple items there (e.g, `Transfer-Encoding: gzip, chunked`), and I don't think that would match correctly here. But AFAICT none of that is new (and is certainly not something Git does).

icing: Just added the check as suggested by @bagder.

We have a bit backward compatibility weirdness here. Ideally, we would either auto-reset post sizes between transfers OR require all applications to set it. But that would certainly also break some apps.

bagder (approved):

PR #13266, opened 2024-04-03
appveyor: OpenSSL 3 no longer found by CMake, revert to 1.1.1

vszakats:
OpenSSL moved directories, and bumped versions in AppVeyor CI.

Downgrading is not an ideal solution, but however trivial the solution
may be, I failed to come with anything that made CMake recognize either
OpenSSL 3.1 or 3.2.

Possibly caused by:
https://github.com/appveyor/build-images/commit/702e8cdca01f28f6a40687783f493c786cebbe2c
https://github.com/appveyor/build-images/pull/149

Closes #13266

### DIFF
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -29,7 +29,8 @@ set -eux; [ -n "${BASH:-}${ZSH_NAME:-}" ] && set -o pipefail
 # build
 
 if [ "${APPVEYOR_BUILD_WORKER_IMAGE}" = 'Visual Studio 2022' ]; then
-  openssl_root_win='C:/OpenSSL-v31-Win64'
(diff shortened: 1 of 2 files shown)


vszakats:
Root cause is the OpenSSL binary distro changed its directory layout and library names:
https://gitlab.kitware.com/cmake/cmake/-/issues/25702
https://gitlab.kitware.com/cmake/cmake/-/issues/25856

CMake hasn't fully tracked those yet.

(But, it also failed to link when pointing to those new names, so let's see if the libs also changed internally.)

vszakats:
Fix merged for CMake 3.29.2.
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9402/diffs

PR #13267, opened 2024-04-03
winbuild: fix RC path was wrong if we have modified it at the beginning。

David-Lee123:
#Utilities.
#If a path is required that contains characters such as space, quote the path.
MT         = mt.exe
RC         = rc.exe
If we modified the RC path at the beginning， the rc $(CURL_RC_FLAGS) could be failed。


bagder (approved):

bagder:
Closed via ffa8a385a024f

PR #13268, opened 2024-04-03
dist: remove the curl-config.1 from the tarball

bagder:
The markdown file is already there and the .1 file gets generated in the build.

Ref: #13250


pghmcfc:
The curl-config.1 file is generated OK for me in 8.8.0 - but not installed.

bagder:
See #13741

PR #13270, opened 2024-04-03
url: do not URL decode proxy credentials

### LINKED ISSUE #13265: libcurl: unexpected url-decoding for CURLOPT_PROXYPASSWORD & CURLOPT_PROXYUSERNAME values
### I did this

```
#include <curl/curl.h>

int main(void)
{
  if(curl) {
    CURLcode res;
    curl_easy_setopt(curl, CURLOPT_URL, "http://curl.se");
    curl_easy_setopt(curl, CURLOPT_PROXY, "http://curl.se:80");
    curl_easy_setopt(curl, CURLOPT_PROXYUSERNAME, "mrsmith%aaa");
    curl_easy_setopt(curl, CURLOPT_PROXYPASSWORD, "qwerty%4b!");
    curl_easy_cleanup(curl);
  }
  return 0;
}
```

### I expected the following

Proxy should get username "mrsmith%aaa" and password "qwerty%4b!".

But proxy gets this:

```
* Proxy auth using Basic with user 'mrsmitha'
> GET http://curl.se/ HTTP/1.1
Host: curl.se
Proxy-Authorization: Basic bXJzbWl0aKphOnF3ZXJ0eUsh
Accept: */*
Proxy-Connection: Keep-Alive
```

After base64 decoding auth string "bXJzbWl0aKphOnF3ZXJ0eUsh" is "mrsmitha:qwertyK!".
But auth string should be "mrsmith%aaa:qwerty%4b!".

According documentation, [CURLOPT_PROXYUSERNAME](https://curl.se/libcurl/c/CURLOPT_PROXYUSERNAME.html) and [CURLOPT_PROXYPASSWORD](https://curl.se/libcurl/c/CURLOPT_PROXYPASSWORD.html) values are not supposed to be url-encoded.
Only [CURLOPT_PROXYUSERPWD](https://curl.se/libcurl/c/CURLOPT_PROXYUSERPWD.html) value is supposed to be url-encoded.

### curl/libcurl version

8.7.0 and earlier

### operating system

all

bagder:
The two options CURLOPT_PROXYUSERNAME and CURLOPT_PROXYPASSWORD set the actual names as-is, not URL encoded.

Modified test 503 to use percent-encoded strings in the credential strings that should be passed on as-is.

Reported-by: Sergey Ogryzkov
Fixes #13265

### DIFF
--- a/lib/setopt.c
+++ b/lib/setopt.c
@@ -52,6 +52,8 @@
 #include "hsts.h"
 #include "tftp.h"
 #include "strdup.h"
+#include "escape.h"
+
 /* The last 3 #include files should be in this order */
 #include "curl_printf.h"
 #include "curl_memory.h"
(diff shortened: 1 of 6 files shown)


bagder:
These macOS CI job failures are mysterious. For example a lot of `--libcurl` tests fail like this:

~~~
 +  curl_easy_setopt(hnd, CURLOPT_IPRESOLVE, 1L);
~~~

... meaning that something makes curl enforce IPv4!!

bagder:
And when rerunning the jobs without any change, now suddenly only **one** of the macOS jobs fails.

PR #13277, opened 2024-04-04
GHA: enable unity mode for cmake jobs + tidy-ups

vszakats:
Unity mode is not supported by CMake v3.7.2 used in linux-old, but
enable it anyway for consistency and to kick in automatically once
migrating to a newer old Linux in the future.

Also:
- replace `CMAKE_COMPILE_WARNING_AS_ERROR` with `CURL_WERROR`.
- delete default build option `PICKY_COMPILER=ON`.

Closes #13277

### DIFF
--- a/.github/workflows/awslc.yml
+++ b/.github/workflows/awslc.yml
@@ -141,7 +141,7 @@ jobs:
 
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


vszakats:
Enabling unity mode shortens the build phase from ~12s to ~5s.

https://github.com/curl/curl/actions/runs/8552308825/job/23433137110: 12s
https://github.com/curl/curl/actions/runs/8553556418/job/23436988344: 5s

The "generate" phase is still slow for macOS (or anything other than Windows). Also package installation.

PR #13278, opened 2024-04-04
socket: add MPTCP support

CrapsDorian:
Multipath TCP (MPTCP), standardized in RFC8684 [1], is a TCP extension that enables a TCP connection to use different paths.

Multipath TCP has been used for several use cases. On smartphones, MPTCP enables seamless handovers between cellular and Wi-Fi networks while preserving established connections. This use-case is what pushed Apple to use MPTCP since 2013 in multiple applications [2]. On dual-stack hosts, Multipath TCP enables the TCP connection to automatically use the best performing path, either IPv4 or IPv6. If one path fails, MPTCP automatically uses the other path.

To benefit from MPTCP, both the client and the server have to support it. Multipath TCP is a backward-compatible TCP extension that is enabled by default on recent Linux distributions (Debian, Ubuntu, Redhat, ...). Multipath TCP is included in the Linux kernel since version 5.6 [3]. To use it on Linux, an application must explicitly enable it when creating the socket. No need to change anything else in the application.

This attached patch adds an --mptcp option which allows the creation of an MPTCP socket instead of TCP on Linux. If Multipath TCP is not supported on the system, an error will be reported. It is important to note that if the end server doesn't support MPTCP, the connection will continue after a seamless fallback to TCP.

-----------------------------------------------------------------------------------------------------------------------------------------------------------------

An HTTP test server indicating whether MPTCP (Multipath TCP) is utilized is now available at the following address: https://check.mptcp.dev/.

Attention points:
- This PR replaces #9713.
- Sorry for #13161, I accidentally opened it while the code was not ready yet, my bad.
- Regarding the documentation, we didn't find a good reference (`See-also`). We put `tcp-fastopen` because we had to put something, but we are happy to change.
- It is not clear for us if a test is worth it. According to your comments from #9713, it might not be worth it. We successfully tested it with https://check.mptcp.dev/ .
- The `transport` is changed to `TRNSPRT_MPTCP` in `Curl_cf_tcp_create()`, not to change the logic before ("happy eyeballs", etc.), and because this feature is an extension to TCP, and it should only modify TCP connections. We are happy to change that if there is a better way.
- In the declaration of `struct OperationConfig` in `src/tool_cfgable.h`, the `struct OperationConfig *next;` item has this comment next to it: `/* Always last in the struct */`, but it is no longer the last item in the structure. Is it normal?


bagder:
the macOS fails on GitHub Actions are the same weird and mysterious problems we have seen pop up in other PRs recently so not related to this

lib/cf-socket.c
```diff
@@ -244,6 +244,13 @@ void Curl_sock_assign_addr(struct Curl_sockaddr_ex *dest,
     dest->socktype = SOCK_STREAM;
     dest->protocol = IPPROTO_IP;
     break;
+  case TRNSPRT_MPTCP:
+#ifndef IPPROTO_MPTCP
```
piru, line 248: Wouldn't it be more clean to rather `result = CURLE_UNSUPPORTED_PROTOCOL;` if the define is missing?
If the idea is to hide the error quietly, then do so, rather than using hardcoded value (which may or may not be the same on all platforms).

lib/cf-socket.c
```diff
@@ -1602,6 +1609,12 @@ CURLcode Curl_cf_tcp_create(struct Curl_cfilter **pcf,
     result = CURLE_OUT_OF_MEMORY;
     goto out;
   }
+
+#ifdef __linux__
```
piru, line 1613: I think it is unwise to limit the MPTCP support for Linux only. AFAIK macOS/iOS already support it and FreeBSD is not far behind.

matttbe: It would be great to support other OS. But unfortunately, it is only easily accessible with Linux, thanks to `IPPROTO_MPTCP` that is for the moment specific to Linux

On iOS, MPTCP is available, but:
- It is easy only if you use their SDK: [doc](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/improving_network_reliability_using_multipath_tcp)
- If not, you need to use private libraries (we are not even sure the headers are available) with specific functions to create sockets that are apparently not documented, e.g., [OpenSSH for iOS](https://github.com/apple-oss-distributions/OpenSSH/blob/main/openssh/sshconnect.c#L487)

On FreeBSD, there was an ongoing implementation, but that was years ago, and not working today according to [this](http://www-cs-students.stanford.edu/~sjac/freebsd_mptcp_info.html)

There are other implementations, but on specific systems (Citrix load balancer, userspace, etc.): more details [here](http://blog.multipath-tcp.org/blog/html/2018/12/15/apple_and_multipath_tcp.html)

So it looks like we might have to restrict this to Linux only for the moment. Is it an issue? It is always possible to add a `#elif` later, no?

lib/cf-socket.c
matttbe: > If we want an explicit error to say that the feature is not supported, maybe it would be better to do that in an `#else` after the `#ifdef __linux__` we added in `Curl_cf_tcp_create()`, no?

Something like this maybe?

```diff
diff --git a/lib/cf-socket.c b/lib/cf-socket.c
index a852a5231..8bc5423d9 100644
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1610,10 +1610,14 @@ CURLcode Curl_cf_tcp_create(struct Curl_cfilter **pcf,
     goto out;
   }
 
+  if(data->set.mptcp) {
 #ifdef __linux__
-  if(data->set.mptcp)
     transport = TRNSPRT_MPTCP;
+#else
+    result = CURLE_UNSUPPORTED_PROTOCOL;
+    goto out;
 #endif
+  }
 
   cf_socket_ctx_init(ctx, ai, transport);
 

```

matttbe: FYI, I just pushed a new version have this diff (+ a rebase on top of master)

docs/libcurl/opts/CURLOPT_MPTCP.md
```diff
@@ -0,0 +1,76 @@
...
+    curl_easy_setopt(curl, CURLOPT_MPTCP, 1L);
+    curl_easy_perform(curl);
+  }
+}
+~~~
+
+# AVAILABILITY
+
+Support for MPTCP in libcurl requires Linux 5.6 or later.
+The features availability in libcurl can also depend on the version of libcurl.
```
bagder, line 68: For clarity, this could probably also mention that since this is TCP only, it has no effect for HTTP/3 (QUIC),

docs/libcurl/opts/CURLOPT_MPTCP.md
```diff
@@ -0,0 +1,76 @@
...
+simultaneously.
+MPTCP is beneficial in networks where multiple paths exist between clients
+and servers, such as mobile networks where a device may switch between WiFi
+and cellular data or in wired networks with multiple Internet Service Providers.
+
+Enabling MPTCP can improve the performance and reliability of network requests,
+particularly in environments where multiple network paths (e.g., WiFi and
+cellular) are available.
+
+Note: MPTCP support depends on the underlying operating system and network
```
bagder, line 41:
```suggestion
MPTCP support depends on the underlying operating system and network
```

docs/cmdline-opts/mptcp.md
```diff
@@ -0,0 +1,39 @@
...
+## Usage
+
+To use MPTCP for your connections, add the `--mptcp` option when using `curl'.
+
+## Requirements
+
+This feature is currently only supported on Linux starting from kernel 5.6.
+
+The server you are connecting to must also support MPTCP. If not, the connection
+seamlessly falls back to TCP.
```
bagder, line 35: For clarity, this could probably also mention that since this is TCP only, it has no effect for HTTP/3 (QUIC),

matttbe:
@bagder thank you for the review! I just applied the suggested modification and addressed your 2 other comments.

I also rebased the modifications on top of master, and squashed the last modifications. I hope that's OK to do that here. Or maybe you prefer when modifications are done in separated commits?

bagder:
> I also rebased the modifications on top of master, and squashed the last modifications

That's perfect!

bagder:
Since there is also a [multipath QUIC](https://multipath-quic.org/) proposal brewing that could perhaps materialize in a future, should we perhaps rather call this option `CURLOPT_MULTIPATH` to make it possible to support QUIC (or something else) in a distant future?

matttbe:
> Since there is also a [multipath QUIC](https://multipath-quic.org/) proposal brewing that could perhaps materialize in a future, should we perhaps rather call this option `CURLOPT_MULTIPATH` to make it possible to support QUIC (or something else) in a distant future?

Good point! 'multipath' can indeed be used in different transport protocols (MPTCP, MPQUIC, MPDCCP), but also at a lower layer (e.g. IP Route Multipath → equal cost multipath) or higher one (e.g. downloading the same file HTTP in multipath parts, using multiple IPs available on the client side: i.e. half downloaded via WiFi, the other half via the cellular network, similar to [Samsung's Download Booster](https://www.samsung.com/za/support/mobile-devices/what-is-the-download-booster-and-how-do-i-enable-it-on-my-samsung-galaxy-note4/) feature that was confusedly called "multiple paths HTTP" by some). I usually don't recommend using this generic term because it can be interpreted differently, but maybe a good documentation can help here? What do you think?

Also, it is important to note that even if MPQUIC is pushed by some of the same people who pushed MPTCP (and who even wrote #9713 :) ), its usage by apps is different compared to MPTCP: instead of simply enabling it by changing one line and letting the kernel doing the rest depending on how the latter has been configured, the app will likely have to do this configuration part to have the different paths created and decide on how to use them. Please see the Quiche [PR](https://github.com/cloudflare/quiche/pull/1310) for more details.

QUIC libs supporting MPQUIC will likely come with a default packets scheduler, and maybe a path-manager, so the app can use multiple paths without too much of an effort. But my point here is that with MPQUIC, the end user might have to pass more info to the app: which paths to use or which path manager to select, what packet scheduler technique to use, etc. Apps like Curl might have to support multiple techniques to be able to serve multiple purposes. In this case, I guess there will be other specific MPQUIC options (or env vars?). Do you then think we should have a generic `CURLOPT_MULTIPATH` option, plus some MPQUIC specific ones? Or only MPTCP and MPQUIC specific ones?

matttbe:
(I just pushed a new version: it is just a rebase on top of master because there was a conflict with a362962b72 -- I didn't rename `CURLOPT_MPTCP`, but feel free to tell us if we should do that)

icing:
> It looks like the issues with '[Linux / hyper](https://github.com/curl/curl/actions/runs/8701937864/job/23864942392?pr=13278)' job is not specific to this modification: we can see the same errors in other PRs if I'm not mistaken. I don't know if we need to do something for that.

The CI is using the master branch of hyper and they seem to have borked that.

bagder:
Yes, #13380 tracks that

bagder:
When using this option, is there a way to tell it is actually working (has the feature enabled once a connection is established)? Since it falls back transparently it will be hard to tell based on behavior.

matttbe:
> Since it falls back transparently it will be hard to tell based on behavior.

Correct.

Note that if the kernel doesn't support MPTCP, an error will be returned: there is no transparent fallback in this case.

If the kernel supports MPTCP, an easy to check if this feature is working correctly with Curl is to use the test server mentioned above, which will return the current status:

```
$ ./src/curl http://test.multipath-tcp.org:5000
You are not using MPTCP.
$ ./src/curl --mptcp http://test.multipath-tcp.org:5000
You are using MPTCP.
```

Or by checking externally that the created socket didn't fall back to TCP, e.g. with `ss -Mia`.

> When using this option, is there a way to tell it is actually working (has the feature enabled once a connection is established)?

From an application, there is a way to check that an established connection didn't fallback from MPTCP to TCP with a `getsockopt()`, e.g.

```c
#define MPTCP_INFO 1
#define MPTCP_INFO_FLAG_FALLBACK 1
bool socket_is_mptcp(int sockfd)
{
    socklen_t len = sizeof(struct mptcp_info);
    struct mptcp_info info = { 0 };

    /* We should get EOPNOTSUPP (or ENOPROTOOPT in v6) in case of fallback */
    if (getsockopt(sockfd, SOL_MPTCP, MPTCP_INFO, &info, &len) < 0) {
        if (errno != EOPNOTSUPP && errno != ENOPROTOOPT)
            perror("getsockopt(MPTCP_INFO)");
        return false;
    }
    return (info.mptcpi_flags & MPTCP_INFO_FLAG_FALLBACK) == 0;
}
```

The main downside is that it requires kernel 5.16. There are more details on the new version of our website (under construction, should be published very soon): [Check for TCP fallback](https://mux99.github.io/mptcp.dev/mptcp-info.html#check-for-tcp-fallback)

@bagder : When would you like to have such info? Is it just to check the modification works, or to add the info somewhere, only visible in verbose mode?

CrapsDorian:
oops, sorry, I accidentally pushed on the wrong branch and closed the PR, my bad

CrapsDorian:
Just for information we changed the test server url from http://test.multipath-tcp.org:5000/ to https://check.mptcp.dev/

Neustradamus:
Nice PR!

matttbe:
FYI, I looked at MPTCP support on macOS:
- The proper way is to use [Apple's Network framework](https://developer.apple.com/documentation/network) instead of using the socket API, but that's quite a bit of changes to support that (see [here](https://github.com/apple-oss-distributions/OpenSSH/blob/9b6202341ee10b42e7391229ad5c0f2eb8aea8af/openssh/sshconnect.c#L467-L599) how it looks in Apple's OpenSSH fork: once the framework is used, enabling MPTCP is just one line).
- I looked for an alternative (but undocumented) method where the socket has to be created with `AF_MULTIPATH` domain, and `connectx()` needs to be used instead of `connect()`, similar to what is currently done with TCP FastOpen:
  - The modification in the code is small, see [here](https://github.com/matttbe/curl/commit/18196adaa32e1fb0fd04541e6dd292a2566cfe28#diff-548113407958134f68604a9c5767d7c838f60a687ecdd179d07e6f6ab39b4693)
  - With that, it looks like an MPTCP connection is being created!
  - But only a single path is created... which is a bit limiting if we want to use **Multipath** TCP :)

This took a bit of time for me to get that tested, because I don't have the hardware. I don't think I could look at using the [Apple's Network framework](https://developer.apple.com/documentation/network) then.

Anyway, while looking at that, I changed the way MPTCP was handled: no longer using a new `transport` (switched from `TRNSPRT_TCP` to `TRNSPRT_MPTCP` in `Curl_cf_tcp_create()`), but by changing the behaviour in `socket_open()` method, see: [here](https://github.com/matttbe/curl/commit/dc9515ca3ac7c3c8cb6ec05e360627892af20d1f#diff-548113407958134f68604a9c5767d7c838f60a687ecdd179d07e6f6ab39b4693). If you prefer this way, I can update this PR.

bagder:
Is this new libcurl option actually necessary? Is the existing `CURLOPT_OPENSOCKETFUNCTION` option not enough for the application to create the socket with the MPTCP option set ?

matttbe:
> Is this new libcurl option actually necessary? Is the existing `CURLOPT_OPENSOCKETFUNCTION` option not enough for the application to create the socket with the MPTCP option set ?

@bagder indeed, `CURLOPT_OPENSOCKETFUNCTION` option is enough to create the socket with the MPTCP option set. Thank you for the idea.

I just updated Dorian's branch to do that (+ rebased on top of `master` and changed the versions in the doc). Please note that Dorian is no longer able to work on that, so I took over his work then.

matttbe:
(now with Windows support)

matttbe:
Sorry, it is not clear to me why there is an issue with the [windows / msvc (`x64, windows, -DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=ON -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=ON`)](https://github.com/curl/curl/actions/runs/9347941624/job/25726027545?pr=13278#logs) job: it looks like the job is still in progress, and I cannot see the end of the logs. Maybe a timeout?

docs/cmdline-opts/mptcp.md
```diff
@@ -0,0 +1,41 @@
+---
+c: Copyright (C) Dorian Craps, <dorian.craps@student.vinci.be>
+SPDX-License-Identifier: curl
+Long: mptcp
+Added: 8.9.0
+Help: Enable Multipath TCP (MPTCP)
```
bagder, line 6: Since the command line option itself is called `mptcp` maybe we can drop the `(MPTCP)` part from the help text? It seems superfluous to me.

bagder: Besides that, I think we are ready to merge

matttbe: Thank you for the review!

I just removed the non-necessary `(MPTCP)` in the help text.

I rebased the branch on top of master and force-pushed.

matttbe: (I also had to modify the same help message in `src/tool_listhelp.c`.)

bagder (approved):

bagder:
Thanks!

Neustradamus:
Good job to all!

PR #13281, opened 2024-04-04
docs/cmdline-opts: invoke managen using a relative path

bagder:
... no need to use an absolute path, that makes the build unncessarily fail if invoked using a different mount point


bagder:
macOS CI failures: #13284

Issue #13284, opened 2024-04-04
mysterious failures and IPv6 problems for macOS GitHub actions jobs


bagder:
### I did this

Over the last several days, CI jobs running on GitHub actions have turned flaky. Sometimes one of the macOS jobs fails, sometimes 10 of them fail. Rerunning the same code again may result in a different outcome.

The failures seem IPv6 related somehow. [example job](https://github.com/curl/curl/actions/runs/8528433634/job/23362051611?pr=13250), [another example](https://github.com/curl/curl/actions/runs/8528433634/job/23362059353?pr=13250)

The most obvious symptom seems to be that when it has problems, it can't "resolve" `::1`. It can be noted that for macOS we call `getaddrinfo` even on plain numerical IP addresses

## failures

When this problem happens, usually the following test cases fail: 1085 1400 1401 1402 1403 1404 1405 1406 1407 1420 1465 1467 1468 2100

Looking at test logs in a failed run , we can see that all HTTP-IPv6 tests are skipped because curl cannot verify the IPv6 server:
~~~
2024-04-02T19:43:36.1057020Z  * Added connection 0. The cache now contains 1 members
2024-04-02T19:43:36.1057900Z  * STATE: CONNECT => RESOLVING handle 0x7fc26980be08; line 1954
2024-04-02T19:43:36.1058690Z  * Could not resolve host: ::1
~~~

## test 1085

Tries to use an IPv6 address for `--interface` but the resolving of the host (`::1`) fails, which makes the test return an unexpected return code.

## test 1400-1468

These tests verify `--libcurl` - outputting generated libcurl using source code for a curl command line.

These tests mysteriously get an extra line of libcurl code added:

~~~c
curl_easy_setopt(hnd, CURLOPT_IPRESOLVE, 1L);
~~~

The only way `CURLOPT_IPRESOLVE` is set by the curl command line tool is when the `-4/--ipv4` option is used. And it is not used in any of these tests.

## test 2100

Uses DoH. When failing, the request does not ask for an IPv6 address (only IPv4) which makes the protocol check fail. Because it deems IPv6 not working?

## macOS runner image
~~~
Image: macos-12
  Version: 20240329.1
  Included Software: https://github.com/actions/runner-images/blob/macOS-12/20240329.1/images/macos/macos-12-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/macOS-12%2F20240329.1
~~~

### I expected the following

The tests should just work.

### curl/libcurl version

current git

[...]

PR #13285, opened 2024-04-04
GHA: on macOS remove $HOME/.curlrc

bagder:
Ref: https://github.com/actions/runner-images/pull/9586
Ref: #13284

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -167,6 +167,9 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - run: rm -f $HOME/.curlrc
+        name: remove $HOME/.curlrc
+
       - run: autoreconf -fi
         name: 'autoreconf'
 


jay:
it's going to be reversed actions/runner-images#9626

vszakats:
Would it be possible to ignore `.curlrc` for all tests?

The same can happen with any image anytime, and even more so on a local system with `.curlrc` customized for daily use.

bagder:
> Would it be possible to ignore .curlrc for all tests?

This certainly shows a use case for exactly that. I imagine the only sensible way to do what is to use an environment variable as a signal.

Slightly related to my "debugging config file options" ideas: https://curl.se/mail/archive-2024-04/0000.html

bagder:
Removing the file is harmless and fixes the issue short-term so I merge this now.

vszakats:
> Slightly related to my "debugging config file options" ideas: https://curl.se/mail/archive-2024-04/0000.html

Is there any downside in introducing an env to override the default config filename?

E.g. `CURL_CONFIG_FILE=~/my-curlrc` or `CURLRC=~/my-curlrc`?

This could also have other uses, and such feature isn't unprecedented with other
tools. This also allows testing the feature itself by pointing it to a test config file.

We could limit it to debug builds, in case there is any concern for production use.

dfandrich:
Regarding the possibility of ignoring .curlrc for all tests, we could just
unconditionally add -q to all curl commands before they're run in runtests.pl.
We're adding a lot of other arguments to each curl command already, so one more
isn't an issue. That is already overridden by a subsequent -K so tests that
need a config file will still work fine.

PR #13296, opened 2024-04-05
projects/Windows: drop project files for Visual Studio before 2015

bagder:
Microsoft themselves only support the 2022 version.

They have not been shipped in release tarballs for a while either.


MarcelRaad:
I'd rather remove the project files for the *newer* versions. Or rather, if we can remove the older project files, then we can probably remove all of them.

Visual Studio itself ships CMake and can even handle CMake from within the IDE since 2017. Current CMake versions, on the other hand, can only target Visual Studio 2013 and above, and targeting 2013 is deprecated.

jay:
As I think I mentioned the last time this came up I use both the 2010 (VC10) and 2013 (VC12) project files. I understand why there is interest in retiring them but for very old systems they are faster. My use case it a little different though because I am working on the project sometimes on old computers, and in those cases need a way to switch fast between different configurations. For most people they are going to use a single configuration that they generate with the winbuild (nmake) or cmake (visual studio, mingw, etc) files.

bagder:
Then let's do it the @MarcelRaad way: we drop the newer ones instead? Which then also implies not shipping them/any project files in the tarball?

bagder:
I switch out this proposal for #13311 instead!

PR #13299, opened 2024-04-05
dist: `set -eu`, fix shellcheck, make reproducible and smaller tarballs

vszakats:
- set bash `-eu` and fix fallouts.
- fix shellcheck warnings.
- set and use `SOURCE_DATE_EPOCH` for reproducibility.
  Authored-by: Daniel J. H.
  Ref: #13280
- set `TZ=UTC` and `LC_ALL=C` for reproducibility.
- make file timestamps in tarball/zip reproducible.
- make directory timestamps in zip reproducible.
- make timestamps of tarballs/zip reproducible.
- make file order in tarball/zip reproducible.
- omit extra file metadata from zip for reproducibility.
- use maximum zip compression.
- use POSIX `ustar` tarball format to avoid supply chain vulnerability:
  https://seclists.org/oss-sec/2021/q4/0
- make uid/gid in tarball reproducible.
- omit owner user/group names from tarball for reproducibility and privacy.
- omit current timestamp from .gz header for reproducibility.
- display SHA-256 hashes of produced tarballs/zip.
- fix whitespace.

`.tar.gz` also became smaller in the process: 4,462,311 -> 4,148,249 bytes (8.7.1)

Requires GNU tar, GNU date, `sha256sum`.

Ref: #13250
Closes #13299


bagder (approved):
:+1:

bagder:
The tarball is made by the `make -sj dist` command. The `dist` target is created by automake.

vszakats:
> The tarball is made by the `make -sj dist` command. The `dist` target is created by automake.

That was my fear :) I'll have to look... 

After looking:

- the tar format isn't set, so it's not necessarily `ustar`. On my macOS, it is, but the official binary for example isn't (it looks like GNU).
  It should be `ustar` (or maybe `pax`) to avoid supply chain vulnerability: https://seclists.org/oss-sec/2021/q4/0.
- file order is non-reproducible.
- local uid/gid are spilling into the tarball. Repro issue.
- local user/group names may be spilling into the tarball. (on macOS, on the official ones this is fortunately not the case.) Privacy and repro issue.

As for how to tell autotools to do the right thing, I have not explored yet. I assume it's impossible or at least painful.

bagder:
I think we can settle with shipping the content reproducible as start, and not the tarball files themselves.

bagder:
> one option is to untar and repack. Costs extra build work and logic to maintain for us

If we want the tarball images themselves reproducible I figure that's a rather easy thing to add.

vszakats:
FWIW curl-for-win uses this:
```
TZ=UTC tar --create \
        --format=ustar \
        --owner=0 --group=0 --numeric-owner \
        --files-from "${_FLS}"
```
where `${_FLS}` lines are reproducibly ordered.

Also needs to install and remap `gtar` to `tar` on macOS for example. (Possibly `bsdtar` can be setup to create the desired output, though this was simpler/safer.)

I'd vote to make the tar images reproducible, if we're tackling this part anyway.

By easy, which method would you go?

bagder:
> By easy, which method would you go?

I figure untaring and retaring with the proper options should work fine. Then we can let `make dist` work as it does and get reproducibility by our own means.

bagder:
Maybe like this?
~~~diff
diff --git a/maketgz b/maketgz
index 602f1071b..089d1bfb8 100755
--- a/maketgz
+++ b/maketgz
@@ -174,10 +174,28 @@ res=$?
 if test "$res" != 0; then
     echo "make dist failed"
     exit 2
 fi
 
+retar() {
+  tempdir=$1
+  rm -rf $tempdir
+  mkdir $tempdir
+  cd $tempdir
+  gzip -dc ../$targz | tar -xf -
+  find curl-* | sort > files
+  tar --create --format=ustar --owner=0 --group=0 --numeric-owner --files-from files -zf out.tar.gz
+  rm files
+  mv out.tar.gz ../
+  cd ..
+  rm -rf $tempdir
+}
+
+retar ".tarbuild"
+echo "replace $targz with out.tar.gz"
+mv out.tar.gz "$targz"
+
 ############################################################################
 #
 # Now make a bz2 archive from the tar.gz original
 #

vszakats:
Oh well, `make dist` fails all over the place inside autotools on macOS. Let's stick to Linux and GNU tools.

vszakats:
Source tarballs and zip should be fully reproducible now. (Assuming `SOURCE_DATE_EPOCH` is set to e.g. the tagged commit's timestamp.)

Couldn't test `maketgz` as a whole. Let me know of any issues.

maketgz
```diff
@@ -192,32 +219,38 @@ gzip -dc $targz | bzip2 --best > $bzip2
...
+# Set deterministic timestamp
+touch -c -t "$filestamp" "$targz" "$bzip2" "$xz" "$zip"
+
 echo "------------------"
 echo "maketgz report:"
 echo ""
-ls -l $targz $bzip2 $zip $xz
+ls -l "$targz" "$bzip2" "$xz" "$zip"
+
+command -v openssl >/dev/null 2>&1 && openssl dgst -sha256 "$targz" "$bzip2" "$xz" "$zip"
```
bagder, line 253: This adds an unfortunate dependency on the openssl tool which I think we can avoid. How about just just use `sha256sum`?

vszakats: I choose openssl because some form of it is available in most envs (tests/certs/scripts/*.sh already depend on it). `sha256sum` is missing by default from macOS, also possibly from BSDs.

It should be easy to install on any platform though, so why not.

maketgz
```diff
@@ -168,22 +178,39 @@ echo "produce RELEASE-TOOLS.md"
...
+  exit 2
 fi
 
+retar() {
+  tempdir=$1
+  rm -rf "$tempdir"
+  mkdir "$tempdir"
+  cd "$tempdir"
+  gzip -dc "../$targz" | tar -xf -
+  find curl-* -type f -exec touch -c -t "$filestamp" '{}' +
```
bagder, line 195: @vszakats was there a particular reason you selected to not set the date on the directories? They now "leak" their source datestamps into the tarball in a way the ordinary files were made not to.

vszakats: In the next line I'm only feeding the files to tar, so the directory entries are not actually included in the tarball. What you may be seeing is the automatically created directories _on extract_, which get the current timestamp.

If we include directories in `find` feeding `tar`, `tar` will include duplicate files. Dir entries will pickup all files in them, and explicit file entries are added on top of that.

Meaning, if we want tar to store directory entries and extract those with their stable timestamp, we will need a slightly different approach.

bagder: Aaah yeah that's it of course. Silly me... thanks.

Issue #13302, opened 2024-04-06
Change in behavior for --remote-header-name


morganwillcock:
### I did this

Commit 75d79a4486b279100209ddf8c7fdb12955fb66e9 seems to introduce a change in behavior where the filename used for the downloaded file may no longer be affected by the `--remote-header-name` option.

The easiest way to demonstrate the problem is to try downloading a file from Sourceforge:
```sh
curl -fLOJ https://sourceforge.net/projects/openofficeorg.mirror/files/4.1.15/README.md/download
```

### I expected the following

Currently the filename used will be "download".

In older versions of Curl and with commit 75d79a4486b279100209ddf8c7fdb12955fb66e9 reverted, the filename used will be "README.md". The issue was noticed due to CI workflows failing with the latest release of Curl, when downloading dependencies from Sourceforge.

This seems a significant change in behavior.  Is it completely intentional?

### curl/libcurl version

8.7.0-DEV (On commit `de7b3e89218467159a7af72d58cea8425946e97d`)

### operating system

Debian 12.5
`Linux hostname 6.1.0-18-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01) x86_64 GNU/Linux`

jay:
Please try #13303

morganwillcock:
That does fix it. Thank you.

PR #13303, opened 2024-04-06
tool_cb_hdr: allow --remote-header-name for all http responses

jay:
- Use the content-disposition filename (server-specified filename) for non-2xx responses.

For example, a server may send a content-disposition filename header with a redirect reply (3xx) but not with the final response (2xx). Without this change curl would ignore the server's specified filename and continue to use the filename extracted from the user-specified URL.

This is a partial revert of 75d79a44, which limited content-disposition and etag header processing to 2xx response codes. There's no explanation for why that was done.

Reported-by: Morgan Willcock

Fixes https://github.com/curl/curl/issues/13302
Closes #xxxx

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -107,19 +107,20 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
 
   curl_easy_getinfo(per->curl, CURLINFO_SCHEME, &scheme);
   scheme = proto_token(scheme);
+
+  /*
+   * Write etag to file when --etag-save option is given.
+   */
+
   if((scheme == proto_http || scheme == proto_https)) {
     long response = 0;
     curl_easy_getinfo(per->curl, CURLINFO_RESPONSE_CODE, &response);
 
-    if(response/100 != 2)
-      /* only care about these headers in 2xx responses */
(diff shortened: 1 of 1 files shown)


jay:
This looks like a lot of change but it's almost all whitespace change. Add `w=1` to the URL query to suppress the whitespace changes https://github.com/curl/curl/pull/13303/files?w=1

bagder:
> There's no explanation for why that was done.

It was made to make curl more careful about when to use data in response headers. It should probably allow 3xx as well as 2xx. Not convinced about 1xx, 4xx and 5xx.

jay:
> It was made to make curl more careful about when to use data in response headers. It should probably allow 3xx as well as 2xx. Not convinced about 1xx, 4xx and 5xx.

The RFC doesn't limit it though I struggle to think of why a server would want to send a filename for those responses.

bagder:
> I struggle to think of why a server would want to send a filename for those responses.

Right. Hence this precaution: it seems unlikely and surprising to get the header for the other ranges, so it seems safer to not parse it then.

jay:
I don't think it is safer just limiting.

jay:
Does anyone else have an opinion on this

jay:
Closing in favor of #13484

PR #13306, opened 2024-04-07
GHA: disable permissions where missing

vszakats:
Closes #13306

### DIFF
--- a/.github/workflows/badwords.yml
+++ b/.github/workflows/badwords.yml
@@ -15,6 +15,8 @@ on:
     branches:
     - master
 
+permissions: {}
+
 jobs:
   check:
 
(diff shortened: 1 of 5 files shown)


bagder:
Isn't that the default anyway?

vszakats:
I don't know. We've been adding these to the other workflows earlier in a2f5a4ca6f2cd1b6d10a46e8710d85ca2bc2dba9 #9928. That suggest it wasn't, or at least not always, with the exact circumstances unclear.

In case it's the default, we might want to delete them from existing places.

Otherwise I'd err for caution and use it everywhere where we don't actually need permissions.

vszakats:
Article about defaults, saying it depends on mode, which can be "permissive" and "restricated". I don't know which one applies to the curl repos. Even restricted allows 'read' in some cases:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

bagder (approved):

PR #13307, opened 2024-04-07
GHA: add shellcheck job and fix warnings, shell tidy-ups

vszakats:
Closes #13307

---

- [x] Requires #13299
- [ ] I excluded OS/400 scrips, with a FIXME. #13309
- [ ] `MacOSX-Framework` might use an overhaul. Planned for a separate commit.

Without whitespace changes: https://github.com/curl/curl/pull/13307/files?w=1


bagder (approved):

scripts/contributors.sh
```diff
@@ -29,20 +29,22 @@
...
+  start=$(git tag --sort=taggerdate | grep "^curl-" | tail -1)
+  echo "Since $start:"
 fi
 
 # We also include curl-www if possible. Override by setting CURLWWW
-if [ -z "$CURLWWW" ] ; then
-    CURLWWW=../curl-www
+if [ -z "${CURLWWW:-}" ] ; then
+  CURLWWW=../curl-www
 fi
```
bagder, line 48: Maybe the cooler way to write this in proper shell is this?

```suggestion
CURLWWW="${CURLWWW:-../curl-www}"
```

asedeno:
While I do perfer the newer command substitution syntax, the old syntax works in more places. I'm going to work around this locally by switching `curl-config` to use `#!/bin/bash` on the ancient (SunOS 5.10) system where I care, and am mostly leaving this comment so folks are aware that this does break some folks.

vszakats:
I think we can revert to the backticks for the few affected lines for maximum compatibility.

PR #13309, opened 2024-04-07
OS400: fix shellcheck warnings in scripts

vszakats:
- use `$()` instead of backticks, and re-arrange double-quotes inside.
- add missing `|| exit 1` to `cd` calls. (could be dropped by using `set -eu`.)
- add `-n` to a few `if`s.
- shorten redirections by using `{} >` (as shellcheck recommended).
- silence warnings where variables were detected as unused (SC2034).
- a couple misc updates to silence warnings.
- switch to bash shebang for `-ot` feature.
- split two lines to unbreak syntax highlighting in my editor. (`$(expr \`, `$(dirname \`)

Also enable CI checks for OS/400 shell scripts.

Ref: #13307
Closes #13309

---

~~Added one FIXME in `initscript.sh` where there is a single use of a non-POSIX 
expression `[ "${1}" -ot "${2}" ]` causing a warning. We can fix this by bumping
the shebangs to bash, or come up with a POSIX alternative.~~

/cc @monnerat


vszakats:
The rust job is failing permanently due to:
```
tests/http/test_02_download.py:353: 
[...]
assert os.path.exists(dfile)
```
Ref: https://github.com/curl/curl/actions/runs/8601526016/job/23584842387?pr=13309#step:36:360

Unrelated to this PR.

bagder (approved):

PR #13310, opened 2024-04-07
dist: do not require Perl in `maketgz`

vszakats:
Perl remains required for the tarball build process.

Follow-up to 860cd5fc2dc8e165fadd2c19a9b7c73b3ae5069d #13299
Closes #13310

### DIFF
--- a/maketgz
+++ b/maketgz
@@ -66,8 +66,7 @@ fi
 echo "removing all old *.dist files"
 find . -name "*.dist" -a ! -name Makefile.dist -exec rm {} \;
 
-numeric=$(perl -e \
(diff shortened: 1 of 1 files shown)


bagder:
I'm fine with this. But note that maketgz should be invoked after a proper configure + make run, which requires perl present - so that the tool_hugehelp.c file gets generated properly. Thus, we can't remove perl from the #13250 docker image.

vszakats:
Yes, Perl is here to stay, just this one-liner was easy replace with shell code.
(Also another one-liner in #13307 / 7c425cb502ee648c4e387b30c578840ae2e08b09)

bagder (approved):

PR #13313, opened 2024-04-08
build: remove MacOSX-Framework script

bagder:
I don't think this is much used these days.

Also remove the libcurl.plist file used (only) by this script


vszakats:
In case someone is interested in universal binaries, curl-for-win is able to create them.
It uses this script: https://github.com/curl/curl-for-win/blob/2a3eee0a552bafdd7bee70fd81438869a1bef0a2/_macuni.sh

PR #13314, opened 2024-04-08
mbedtls: call mbedtls_ssl_setup() after RNG callback is set

kailun-qin:
Since mbedTLS v3.6.0, the RNG check added in ssl_conf_check() will fail if no RNG is provided when calling mbedtls_ssl_setup().

Therefore, mbedtls_ssl_conf_rng() needs to be called before the SSL context is passed to mbedtls_ssl_setup().

Ref: https://github.com/Mbed-TLS/mbedtls/commit/b422cab052b51ec84758638d6783d6ba4fc60613


bagder (approved):

bagder:
Thanks!

PR #13323, opened 2024-04-09
stabilitze test_02_23* tests

icing:
- h2-download now always opens the output file on first write callback invocation, if it will pause the transfer or not.
- Checks on output files then does not depend on the amount of data curl has collected for the first write.


lib/libcurl.plist
```diff
@@ -0,0 +1,35 @@
...
+	<key>CFBundleSignature</key>
+	<string>????</string>
+
+	<key>CFBundleShortVersionString</key>
+	<string>libcurl 8.7.2-DEV</string>
+
+	<key>CFBundleGetInfoString</key>
+	<string>libcurl.plist 8.7.2-DEV</string>
+</dict>
+</plist>
```
bagder: I'm pretty sure you did not intend to add this =)

PR #13330, opened 2024-04-09
appveyor: make VS2010 job build-only, enable Schannel, fix compiler warnings

vszakats:
Tests were consistently flaky for a while.

Also fix compiler warnings in `CertOpenStore()` calls for old MSVC compilers:
```
C:/projects/curl/lib/vtls/schannel.c(688):
  warning C4306: 'type cast' : conversion from 'int' to 'LPCSTR' of greater size
C:/projects/curl/lib/vtls/schannel_verify.c(642):
  warning C4306: 'type cast' : conversion from 'int' to 'LPCSTR' of greater size
```
Ref: https://ci.appveyor.com/project/curlorg/curl/builds/49580310/job/ywu2y44kymgc0nif#L106

Closes #13330

### DIFF
--- a/lib/vtls/schannel_int.h
+++ b/lib/vtls/schannel_int.h
@@ -53,6 +53,16 @@
 #define CERT_ALT_NAME_IP_ADDRESS 8
 #endif
 
+#if defined(_MSC_VER) && (_MSC_VER <= 1600)
+/* Workaround for warning:
+   'type cast' : conversion from 'int' to 'LPCSTR' of greater size */
+#undef CERT_STORE_PROV_MEMORY
+#undef CERT_STORE_PROV_SYSTEM_A
+#undef CERT_STORE_PROV_SYSTEM_W
+#define CERT_STORE_PROV_MEMORY    ((LPCSTR)(size_t)2)
+#define CERT_STORE_PROV_SYSTEM_A  ((LPCSTR)(size_t)9)
+#define CERT_STORE_PROV_SYSTEM_W  ((LPCSTR)(size_t)10)
+#endif
 
 #ifndef SCH_CREDENTIALS_VERSION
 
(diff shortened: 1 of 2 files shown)


vszakats:
Enabling Schannel is VS2010 turned up two compiler warnings:
```
C:/projects/curl/lib/vtls/schannel.c(688):
  warning C4306: 'type cast' : conversion from 'int' to 'LPCSTR' of greater size
C:/projects/curl/lib/vtls/schannel_verify.c(642):
  warning C4306: 'type cast' : conversion from 'int' to 'LPCSTR' of greater size
```

MarcelRaad:
Do you remember what the test failures were about? The reason why there were VS 2010 builds in addition to the VS 2008 builds was that the former were able to run tests.

lib/vtls/schannel_int.h
```diff
@@ -53,6 +53,16 @@
 #define CERT_ALT_NAME_IP_ADDRESS 8
 #endif
 
+#if defined(_MSC_VER) && (_MSC_VER <= 1600)
```
MarcelRaad, line 56: Isn't that related to the Windows SDK version rather than the compiler version?

MarcelRaad:
Sorry for missing this before the merge. I wonder why the warnings did not appear in the VS 2008 build, which also has Schannel enabled.

lib/vtls/schannel_int.h
vszakats: It's a recurring warning, so I took the previous fix to this case:
8cfc936f5c9d1fe9a99956a916f95e137476b27c #10674 #6284

It also happened in libssh2, and got fixed with the same fix (albeit with <1700 instead of <=1600)
https://github.com/libssh2/libssh2/commit/2e3e0be816b2e71f59a6f56a0d38898ceeb5c82f
https://github.com/libssh2/libssh2/pull/925

Hard to say, but it could be a compiler regression causing this bogus warnings. It may also be a combination of compiler version and SDK. It's paired with 7.1 in this particular case.

vszakats:
@MarcelRaad No probs. This is at least 3 of this same case now, so if we have a better fix for this, we'll need to apply it all cases.

PR #13333, opened 2024-04-10
fix typos

RainRat:


lib/cfilters.c
```diff
@@ -590,7 +590,7 @@ CURLcode Curl_conn_ev_data_idle(struct Curl_easy *data)
 
 /**
  * Notify connection filters that the transfer represented by `data`
- * is donw with sending data (e.g. has uploaded everything).
+ * is down with sending data (e.g. has uploaded everything).
```
bagder, line 593:
```suggestion
 * is done with sending data (e.g. has uploaded everything).
```

lib/cfilters.h
```diff
@@ -451,7 +451,7 @@ CURLcode Curl_conn_ev_data_idle(struct Curl_easy *data);
 
 /**
  * Notify connection filters that the transfer represented by `data`
- * is donw with sending data (e.g. has uploaded everything).
+ * is down with sending data (e.g. has uploaded everything).
```
bagder, line 454:
```suggestion
 * is done with sending data (e.g. has uploaded everything).
```

bagder (commented):
Apart from these two nits, it seems entirely correct!

lib/cfilters.h
RainRat: ok, that also makes sense

bagder:
@RainRat can you please squash this commit frenzy into a single commit and force-push?

PR #13338, opened 2024-04-10
Deterministic archives via Containerfile

lrvick:
Generates deterministic dist archives with full-source-bootstrapped, deterministic, and hash-locked dependencies.

All dependencies are from stagex and a simple "make" (and a fair bit of waiting) in the stagex repo should produce the exact same digests as the dependencies in the included Containerfile.

This is IMO much more futureproof than a debian based approach, as package versions change from one day to the next so something reproducible today will not be tomorrow. Stagex was created to address use cases exactly like this one.

See: https://codeberg.org/stagex/stagex for more background

Usage:
```
docker build --progress=plain --build-arg VERSION=1.2.3 --output=rewrite-timestamp=true,./dist/ -f Containerfile .
```

Resulting archives will be found in the "dist" directory

NOTE: zip support is not yet packaged in stagex, so zip support is currently disabled until the next stagex release in a few days. Will include it and drop "DRAFT" status then.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -7,7 +7,6 @@ FROM stagex/autoconf@sha256:e42cb175aa5739b511d4ab75a6dd24797bcaa957f12082240b87
 FROM stagex/automake@sha256:749940b3c06e7e1f80f667a0870d59a00c86298e865530f6d11c85b1d05fc548 as automake
 FROM stagex/pkgconf@sha256:470052019202f89aa8a63cf227364beb2e23e3e5d3e15d413c7df48cdb891aef as pkgconf
 FROM stagex/libtool@sha256:40037dd8aba84b58e9c08390043e53cc90b4d6e2f4abfb4978674030f63fe219 as libtool
-FROM stagex/openssl@sha256:859f892877f29972d468d5077817402c7aec1ad4b1e5c25e7cba8bcab79ea7ce as openssl
 FROM stagex/file@sha256:1195ad7a437d5a2b2cfe8f5319945111d730baf62838718f5d4aa5b54240d3af as file
 FROM stagex/m4@sha256:2c8b055ce71cc5452b519b5b0540ce5e0ef9d36fc14164d5609e35dff85d2ce9 as m4
 FROM stagex/gcc@sha256:a5c3774cb42719d308f98fc9311fea3a73f638353a7ed81bc9bef39803c8dadd as gcc
@@ -25,7 +24,6 @@ COPY --from=binutils . /
 COPY --from=autoconf . /
 COPY --from=automake . /
 COPY --from=libtool . /
-COPY --from=openssl . /
 COPY --from=m4 . /
 COPY --from=file . /
 COPY --from=gcc . /
@@ -38,7 +36,7 @@ ENV SOURCE_DATE_EPOCH=1
 RUN --network=none <<-EOF
     set -eux
     autoreconf -fvi
-    ./configure --with-openssl
+    ./configure --without-ssl --without-libpsl
     ./maketgz ${VERSION}
     mkdir /rootfs
     mv curl-${VERSION}.* /rootfs
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,49 @@
+FROM stagex/busybox@sha256:2006bf09c974d842ebc62476074f01181c092022b2a1fdfa63b32d6aed5db323 as busybox
+FROM stagex/make@sha256:e9aef80a9b2bd7003c7fb040cb875d4cee94eab777de93a798ed000ea5ee5fc1 as make
+FROM stagex/musl@sha256:e52da07cf7e2824fcfc7563e5b8792edc296e8ee37dcb59397ad183e742a6dc0 as musl
+FROM stagex/perl@sha256:fc4c642ea8aaf6d9f68b2f01ea5073f781e8994ac2c8267e98b4623d5b2ef752 as perl
+FROM stagex/binutils@sha256:47c9449f655b43e3da5fb6ab1694a9d0a513bfa36cae88b27008a8d95f4df943 as binutils
+FROM stagex/autoconf@sha256:e42cb175aa5739b511d4ab75a6dd24797bcaa957f12082240b875f8ab68c1fa0 as autoconf
+FROM stagex/automake@sha256:749940b3c06e7e1f80f667a0870d59a00c86298e865530f6d11c85b1d05fc548 as automake
+FROM stagex/pkgconf@sha256:470052019202f89aa8a63cf227364beb2e23e3e5d3e15d413c7df48cdb891aef as pkgconf
+FROM stagex/libtool@sha256:40037dd8aba84b58e9c08390043e53cc90b4d6e2f4abfb4978674030f63fe219 as libtool
+FROM stagex/openssl@sha256:859f892877f29972d468d5077817402c7aec1ad4b1e5c25e7cba8bcab79ea7ce as openssl
+FROM stagex/file@sha256:1195ad7a437d5a2b2cfe8f5319945111d730baf62838718f5d4aa5b54240d3af as file
+FROM stagex/m4@sha256:2c8b055ce71cc5452b519b5b0540ce5e0ef9d36fc14164d5609e35dff85d2ce9 as m4
+FROM stagex/gcc@sha256:a5c3774cb42719d308f98fc9311fea3a73f638353a7ed81bc9bef39803c8dadd as gcc
+FROM stagex/bzip2@sha256:79b9f18b94fe2dd27709ff5960849c7b40482b64ed8a64c6f5c1992e0c08d85a as bzip2
+FROM stagex/xz@sha256:8a8843cd206bcf8322f5d1446284e05e7db282a1deeeab02ec7aeaca23a0f858 as xz
+FROM stagex/tar@sha256:45d9dd30aa7a447755d5671654b22ae7aabf43048e694f098ade81aba0878959 as tar
+
+FROM scratch as build
+ARG VERSION=unknown
+COPY --from=musl . /
+COPY --from=busybox . /
+COPY --from=make . /
+COPY --from=perl . /
+COPY --from=binutils . /
+COPY --from=autoconf . /
+COPY --from=automake . /
+COPY --from=libtool . /
+COPY --from=openssl . /
+COPY --from=m4 . /
+COPY --from=file . /
+COPY --from=gcc . /
+COPY --from=bzip2 . /
+COPY --from=xz . /
+COPY --from=tar . /
+ADD . /src
+WORKDIR /src
+ENV SOURCE_DATE_EPOCH=1
+RUN --network=none <<-EOF
+    set -eux
+    autoreconf -fvi
+    ./configure --with-openssl
+    ./maketgz ${VERSION}
+    mkdir /rootfs
+    mv curl-${VERSION}.* /rootfs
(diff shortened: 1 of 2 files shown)


Containerfile
```diff
@@ -0,0 +1,49 @@
...
+ARG VERSION=unknown
+COPY --from=musl . /
+COPY --from=busybox . /
+COPY --from=make . /
+COPY --from=perl . /
+COPY --from=binutils . /
+COPY --from=autoconf . /
+COPY --from=automake . /
+COPY --from=libtool . /
+COPY --from=openssl . /
```
bagder, line 28: I don't think openssl is needed? We can do the build step without TLS.

lrvick: It complained about selecting an ssl version when I did ./configure without --with-openssl. 

When I don't run ./configure at all it complained a generated configure file was missing.

I know how to do deterministic build containers, but I clearly don't know the exact steps I should be using to do curl distribution :)

Containerfile
```diff
@@ -0,0 +1,49 @@
...
+COPY --from=bzip2 . /
+COPY --from=xz . /
+COPY --from=tar . /
+ADD . /src
+WORKDIR /src
+ENV SOURCE_DATE_EPOCH=1
+RUN --network=none <<-EOF
+    set -eux
+    autoreconf -fvi
+    ./configure --with-openssl
```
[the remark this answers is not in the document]
lrvick: oh! will try this now

Containerfile
```diff
@@ -0,0 +1,49 @@
...
+FROM stagex/automake@sha256:749940b3c06e7e1f80f667a0870d59a00c86298e865530f6d11c85b1d05fc548 as automake
+FROM stagex/pkgconf@sha256:470052019202f89aa8a63cf227364beb2e23e3e5d3e15d413c7df48cdb891aef as pkgconf
+FROM stagex/libtool@sha256:40037dd8aba84b58e9c08390043e53cc90b4d6e2f4abfb4978674030f63fe219 as libtool
+FROM stagex/openssl@sha256:859f892877f29972d468d5077817402c7aec1ad4b1e5c25e7cba8bcab79ea7ce as openssl
+FROM stagex/file@sha256:1195ad7a437d5a2b2cfe8f5319945111d730baf62838718f5d4aa5b54240d3af as file
+FROM stagex/m4@sha256:2c8b055ce71cc5452b519b5b0540ce5e0ef9d36fc14164d5609e35dff85d2ce9 as m4
+FROM stagex/gcc@sha256:a5c3774cb42719d308f98fc9311fea3a73f638353a7ed81bc9bef39803c8dadd as gcc
+FROM stagex/bzip2@sha256:79b9f18b94fe2dd27709ff5960849c7b40482b64ed8a64c6f5c1992e0c08d85a as bzip2
+FROM stagex/xz@sha256:8a8843cd206bcf8322f5d1446284e05e7db282a1deeeab02ec7aeaca23a0f858 as xz
+FROM stagex/tar@sha256:45d9dd30aa7a447755d5671654b22ae7aabf43048e694f098ade81aba0878959 as tar
```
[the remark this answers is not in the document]
lrvick: The entire point of stagex is you don't have to trust me or anyone else so I am glad you brought this up :)

We did a bunch of work to ensure every single image we publish is bit-for-bit reproducible, and every single one with signed review and signed reproduction by signed by multiple parties (including curl itself).

If you go to the stagex repo and run "make" it will build a 256 byte assembly seed. It will then pivot to using that seed to build sha256 and a very primitive shell, and build from there up to gcc, and finally all the tools you see included above.

The repo should be trivial to audit as well as we introduce no first-party code, and merely use the OCI standard to build/publish everything from source.

You will see in your "digests.txt" file the final resulting hashes of your build will match the published containers and the exact hashes you see pinned here.

Or you can see in the signatures file that multiple maintainers have signed every package, depending on your threat model.

By comparison with the debian containers you have to trust that whoever has the dockerhub API key didn't leak it, as their images are not signed. Also quite a few packages in debian are not reproducible, and 100% of packages are only signed by one maintainer.

Even if you are totally fine with those trust problems in debian, you are going to end up with a lot of issues forcing debian to generate deterministic builds that will work a week from now. As far as I know I am the only person to have actually done this, and I don't recommend it. 

See my comment here: https://github.com/curl/curl/pull/13250#discussion_r1559411198

Containerfile
lrvick: saw your other comment. Testing now

Containerfile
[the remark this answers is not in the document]
lrvick: Worked. Pushed it up.

Containerfile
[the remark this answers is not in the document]
antonleviathan: Like @lrvick said, please *don't* trust us, just run `make`. You may want to rent a powerful VPS to do this as otherwise it will take a while, I recommend a 32 core machine with 40GB+ of ram - it should take about 1.5 days for the build with these specs. You could avoid building `rust` and cut that down significantly.

Containerfile
```diff
@@ -0,0 +1,47 @@
...
+COPY --from=xz . /
+COPY --from=tar . /
+ADD . /src
+WORKDIR /src
+ENV SOURCE_DATE_EPOCH=1
+RUN --network=none <<-EOF
+    set -eux
+    autoreconf -fvi
+    ./configure --without-ssl --without-libpsl
+    ./maketgz ${VERSION}
```
bagder, line 40: This will not make a correct release since it does not run make before maketgz.

Containerfile
```diff
@@ -0,0 +1,47 @@
...
+    mkdir /rootfs
+    mv curl-${VERSION}.* /rootfs
+    find /rootfs -exec touch -hcd "@0" "{}" +
```
bagder, line 43: This extra touch overrides maketgz which seems wrong.

Containerfile
```diff
@@ -0,0 +1,47 @@
...
+ADD . /src
+WORKDIR /src
+ENV SOURCE_DATE_EPOCH=1
+RUN --network=none <<-EOF
+    set -eux
+    autoreconf -fvi
+    ./configure --without-ssl --without-libpsl
+    ./maketgz ${VERSION}
+    mkdir /rootfs
+    mv curl-${VERSION}.* /rootfs
```
bagder, line 42: Why does it move tarballs like this?

Containerfile
```diff
@@ -0,0 +1,47 @@
...
+FROM stagex/autoconf@sha256:e42cb175aa5739b511d4ab75a6dd24797bcaa957f12082240b875f8ab68c1fa0 as autoconf
+FROM stagex/automake@sha256:749940b3c06e7e1f80f667a0870d59a00c86298e865530f6d11c85b1d05fc548 as automake
+FROM stagex/pkgconf@sha256:470052019202f89aa8a63cf227364beb2e23e3e5d3e15d413c7df48cdb891aef as pkgconf
+FROM stagex/libtool@sha256:40037dd8aba84b58e9c08390043e53cc90b4d6e2f4abfb4978674030f63fe219 as libtool
+FROM stagex/file@sha256:1195ad7a437d5a2b2cfe8f5319945111d730baf62838718f5d4aa5b54240d3af as file
+FROM stagex/m4@sha256:2c8b055ce71cc5452b519b5b0540ce5e0ef9d36fc14164d5609e35dff85d2ce9 as m4
+FROM stagex/gcc@sha256:a5c3774cb42719d308f98fc9311fea3a73f638353a7ed81bc9bef39803c8dadd as gcc
+FROM stagex/bzip2@sha256:79b9f18b94fe2dd27709ff5960849c7b40482b64ed8a64c6f5c1992e0c08d85a as bzip2
+FROM stagex/xz@sha256:8a8843cd206bcf8322f5d1446284e05e7db282a1deeeab02ec7aeaca23a0f858 as xz
+FROM stagex/tar@sha256:45d9dd30aa7a447755d5671654b22ae7aabf43048e694f098ade81aba0878959 as tar
```
bagder, line 15: zlib seems to be misssing

Containerfile
lrvick: Did not realize a full compile was required for source archives to render correctly.  Added this.

Given that a deterministic binary is now being produced, it might make sense to copy that to the final output layer as well, but I have not done this for now as the scope of interest was just source archives.

bagder: > Given that a deterministic binary is now being produced, it might make sense to copy that to the final output layer as well

Nah, We have no use for that. This is a curl release, that is a set of source tarballs only.

Containerfile
lrvick: When we run "output" in docker, it will dump whatever is in the root filesystem of the final "layer" out to the specified directory on the host machine.

We don't want to output automake, bzip, busybox, gcc etc, so we create a final "rootfs" folder that contains just the files we want to export, then pivot to a "scratch" layer and copy the files from "rootfs" that as the final layer as the new root.

If we were also building binaries, we would do the same, and copy the binaries to this final layer, so you only get what you actually want to export from the deterministic build environment.

Containerfile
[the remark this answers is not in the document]
lrvick: They could also just compile the relevant dependencies you wish to verify (and their dependencies):

```
make out/musl/index.json out/busybox/index.json out/make/index.json out/perl/index.json out/binutils/index.json out/autoconf/index.json out/automake/index.json out/libtool/index.json out/m4/index.json out/file/index.json out/zlib/index.json out/gcc/index.json out/bzip2/index.json out/xz/index.json out/tar/index.json
```

If built with Docker 25, the "digest" values in each of those index.json files will correspond to the digests in our public images pinned in this release.

Our next release will be based on Docker 26 and will contain "zip" which is required for this PR to be valid.

bagder:
BTW, the REUSE CI job failure is because the new file does not have copyright and license info. Like this:
~~~
# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
#

bagder:
Stagex seems like generally a good idea but also might be a little early to jump on from where I stand.

Debian is known, trusted and has even been around several years longer than curl, we can likely expect it to be around for a while more into the future. Stagex is new, entirely unproven and has not been tested in real world much yet.

There are glitches in the reproducibility with Debian but we can do a best effort and probably achieve something that in practice mostly likely just works. Right now, I don't think we can tell which one these that are most likely to successfully verify a 2024 curl release tarball in 2034. (As mentioned before, I don't think people will bother much to verify ancient releases.)

lrvick:
I would have said the exact same thing a couple of years ago before my earlier rabbitholes trying to force Debian to be a reproducible builds system, so fair enough.

Debian is the devil people know, and a great workstation distro. In your case you may be able to get some mileage out of it as a build distro most can't get away with as you have very few dependencies and are not shipping binaries. I would again however caution that Docker Debian is not signed and does not have reproducible digests. Trust is anchored in a Dockerhub API key, unlike the Debian on my workstation and likely yours which are at least anchored to PGP signatures (albeit from a SPOF key).

I will also shamelessly plug that the stagex produced curl package is as far as I know the only full-source-bootstrapped, quorum-signed, deterministic, and OCI-native package of curl that exists today.

It can be included into any container, regardless of base image distro
- https://codeberg.org/stagex/stagex/src/branch/main/packages/curl/Containerfile
- https://hub.docker.com/r/stagex/curl

If curl or any projects in your universe need easy-to-verify deterministic archives/builds at any point by all means let us know. Happy to do the work for free to level up supply chain security across the FOSS ecosystem.

Closing this for now though hope it is revisited in the future when we have earned a bit more community trust. Appreciate the consideration and interest.

Issue #13337, opened 2024-04-10
Write function callback is called twice after resume transfer and return CURL_WRITEFUNC_ERROR


pkropachev:
### I did this

We have the scenario when we put the transfer on pause and resume it after a while and then stop transfer with `CURL_WRITEFUNC_ERROR`. For that we register `CURLOPT_WRITEFUNCTION` callback and pause the transfer by returning `CURL_WRITEFUNC_PAUSE` from this callback. After some delay we resume transfer by calling `curl_easy_pause(..., CURLPAUSE_CONT)`. Then after our callback is called we stop transfer by returning `CURL_WRITEFUNC_ERROR`. In this scenario we faced with the problem that our callback is called twice after it had returned `CURL_WRITEFUNC_ERROR`.

We reproduced the behavior using example of code from the [comment](https://github.com/curl/curl/issues/4966#issuecomment-591544506), but slightly simplify it.

```c
#include <curl/curl.h>
#include <stdbool.h>
#include <stdio.h>
#include <unistd.h>

bool paused;
int call_counter;

static size_t http_body(void *buffer, size_t size, size_t nmemb, void *userp) {
  call_counter++;
  fprintf(stderr, "http_body called at %d time\n", call_counter);

  if (call_counter == 1) {
    paused = true;
    fprintf(stderr, "Return pause\n");
    return CURL_WRITEFUNC_PAUSE;
  }

  fprintf(stderr, "Return abort\n");
  return CURL_WRITEFUNC_ERROR;
}

int main(int argc, char **argv) {
  CURLM *curl_multi_handle;
  CURL *curl;

  const char *url = "http://httpbin.org";
  struct CURLMsg *msg;
  int running_handles;
  int msgs_left;
  call_counter = 0;
  paused = false;

  curl_global_init(CURL_GLOBAL_DEFAULT);
  curl_multi_handle = curl_multi_init();
  curl = curl_easy_init();

  curl_easy_setopt(curl, CURLOPT_URL, url);
  curl_easy_setopt(curl, CURLOPT_BUFFERSIZE, 1024);
  curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);
  curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, http_body);
  curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);

  curl_multi_add_handle(curl_multi_handle, curl);

[...]

PR #13340, opened 2024-04-10
cw-out, error and pause handling

icing:
- remember error encountered in invoking write callback and always fail afterwards without further invokes
- do not "finish" transfers that have send/received everything but still paused data waiting to be delivered to the client
- check behaviour in test_02_17 with h2-pausing client
- refs #13337

Further stability improvements:
- go not clear `data->req.keepon` at the end of send/recv. Instead leave any PAUSE bits there to hold transfers until the app unpauses
- move unpausing action of client writer into `Curl_readwrite()`, since errors encountered there are transfer error. Applications do not always check errrors on `curl_easy_unpaus()` and may keep a transfer going even though it has failed.

### DIFF
--- a/lib/cw-out.c
+++ b/lib/cw-out.c
@@ -102,6 +102,7 @@ static void cw_out_buf_free(struct cw_out_buf *cwbuf)
 struct cw_out_ctx {
   struct Curl_cwriter super;
   struct cw_out_buf *buf;
+  BIT(paused);
   BIT(errored);
 };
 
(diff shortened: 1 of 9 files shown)


icing:
@bagder, this change now might impact applications that PAUSE writing but do not UNPAUSE as such a transfer will no longer finish. This might have been unnoticed if the PAUSE happened on the last response chunk.

bagder:
I understand, but I agree with you that this seems like the sensible approach nonetheless. Leaving the transfer paused means it has not yet completed...

PR #13349, opened 2024-04-11
build: prefer `USE_IPV6` macro internally (was: `ENABLE_IPV6`)

vszakats:
Before this patch, two macros were used to guard IPv6 features in curl
sources: `ENABLE_IPV6` and `USE_IPV6`. This patch makes the source use
the latter for consistency with other similar switches.

`-DENABLE_IPV6` remains accepted for compatibility as a synonym for
`-DUSE_IPV6`, when passed to the compiler.

`ENABLE_IPV6` also remains the name of the CMake and `Makefile.vc`
options to control this feature.

Closes #13349

---

~~Ideally I would prefer to switch over all uses in C from `ENABLE_IPV6` to
`USE_IPV6`, which fits the naming used for _all_ other public feature macros.~~

~~Then leave `ENABLE_IPV6` as a compatibility macro that merely sets
`USE_IPV6`. Thoughts?~~


bagder:
> Ideally I would prefer to switch over all uses in C from ENABLE_IPV6 to USE_IPV6, which fits the naming used for all other public feature macros.

Gets a :+1: from me!

PR #13352, opened 2024-04-11
lib: merge `ENABLE_QUIC` C macro into `USE_HTTP3`

vszakats:
Before this patch `lib/curl_setup.h` defined these two macros right
next to each other, then the source code used them interchangeably.

After this patch, `USE_HTTP3` guards all HTTP/3 / QUIC features.
(Like `USE_HTTP2` does for HTTP/2.) `ENABLE_QUIC` is no longer used.

This patch doesn't change the way HTTP/3 is enabled via autotools
or CMake. Builders who enabled HTTP/3 manually by defining both of
these macros via `CPPFLAGS` can now delete `-DENABLE_QUIC`.

Closes #13352

---

I could not spot the difference between the meaning of these two macros
in context of curl's HTTP/3 support. Let me know if I missed something.

### DIFF
--- a/lib/altsvc.c
+++ b/lib/altsvc.c
@@ -303,7 +303,7 @@ struct altsvcinfo *Curl_altsvc_init(void)
 #ifdef USE_HTTP2
     | CURLALTSVC_H2
 #endif
-#ifdef ENABLE_QUIC
+#ifdef USE_HTTP3
     | CURLALTSVC_H3
 #endif
     ;
--- a/lib/config-os400.h
+++ b/lib/config-os400.h
@@ -237,7 +237,7 @@
 
 /* Define to enable HTTP3 support (experimental, requires NGTCP2, QUICHE or
    MSH3) */
-#undef ENABLE_QUIC
+#undef USE_HTTP3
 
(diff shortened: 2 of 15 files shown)


talregev:
There is a several ways to build http3 with curl.
Can you elaborate about each of one, if it will not effect them after your patch?
I am more interested if you can able to compile openssl + nghttp3 to have http3 feature inside curl after your patch?

vszakats:
My patch doesn't change how these features are enabled / selected.

OpenSSL + nghttp3 can be enabled the same way as before:
CMake: `-DCURL_USE_OPENSSL=ON -DOPENSSL_ROOT_DIR=<path> -DUSE_OPENSSL_QUIC=ON`
autotools: `--with-openssl=<path> --with-openssl-quic`

(If someone was using `ENABLE_QUIC` + `USE_HTTP3` as a raw CPPFLAGS flag to enable HTTP3, `ENABLE_QUIC` no longer needs to be passed.)

talregev:
Thank you for clearing it up!

bagder (approved):

PR #13353, opened 2024-04-11
DISTROS: mark rolling release distros

dfandrich:
These are ones that are unlikely to have back-ported curl patches.

Closes #13353

### DIFF
--- a/docs/DISTROS.md
+++ b/docs/DISTROS.md
@@ -10,7 +10,8 @@ SPDX-License-Identifier: curl
 
 Lots of organizations distribute curl packages to end users. This is a
 collection of pointers to where to learn more about curl on and with each
-distro.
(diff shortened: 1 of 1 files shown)


rsbeckerca:
Do you want info on the NonStop ports?

dfandrich:
If you think its packaging information & patches would be useful to others, sure. I suspect it's pretty specific to that platform, though, and there may not be many lessons to share with other packagers by including it.

rsbeckerca:
> If you think its packaging information & patches would be useful to others, sure. I suspect it's pretty specific to that platform, though, and there may not be many lessons to share with other packagers by including it.

I'm not sure there is any use to other packagers. But let me know if you decide differently. The offer is open.

bagder (approved):

PR #13361, opened 2024-04-13
ci: add CMake build variation, fixup libssh detection in `linux-old`

vszakats:
To test without c-ares and hit `easy_lock.h` on an old system. Use this
new build step to introduce small variations, and also test libssh2.

Also add workaround to existing job to enable libssh. (CMake's generic
auto-detection doesn't seem to work here.):
```
CMake Warning at CMakeLists.txt:908 (find_package):
  Could not find a package configuration file provided by "libssh" with any
  of the following names:

    libsshConfig.cmake
    libssh-config.cmake
```
Ref: https://github.com/curl/curl/actions/runs/8661316091/job/23750974358#step:5:69

Closes #13361

### DIFF
--- a/.github/workflows/linux-old.yml
+++ b/.github/workflows/linux-old.yml
@@ -79,19 +79,19 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - name: 'cmake build-only (libssh2)'
+      - name: 'cmake build-only (out-of-tree, libssh2)'
         run: |
           mkdir bld-1
           cd bld-1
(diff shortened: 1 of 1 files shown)


dfandrich:
I wanted at least one build to be "everything on" with the maximum number of
possible options enabled to test the widest range of dependencies. The two
builds now seems to each have a random selection of dependencies that make it
hard to tell what has been tested and what has not.

vszakats:
The only thing I changed in the existing test is to really enable `libssh`. Before this patch, the automatic detection failed, so SSH/SCP was missing.

The newly added test uses libssh2 instead of libssh and disables c-ares to compile the built-in resolver. And disables GSSAPI and zstd, because those are tested in the existing test already.

dfandrich:
I see now; I was comparing the old test to the new one.

PR #13362, opened 2024-04-13
OS400: fix shellcheck warnings in scripts (fixup)

vszakats:
Build scripts must be executed by the os/400 shell (sh), not bash which
is a PASE program: The `-ot` non-POSIX test extension works in os/400 as
well. Ref: https://github.com/libssh2/libssh2/pull/1364#issue-2241646754

Follow-up to 8a622baf9e9233241bbe93d6599c99cb46478614 #13309
Assisted-by: Patrick Monnerat
Closes #xxxxx


pheiduck:
`#!/usr/bin/env sh`
lends you some flexibility on different systems

vszakats:
All curl POSIX scripts use `/usr/bin`, including these before the patch this one is fixing up. So, for this PR, I prefer to just revert to that.

vszakats:
We migrated most of this PR to #13366

vszakats:
Most of this PR superceded by #13366.

PR #13363, opened 2024-04-14
[cmake] Error on libidn2 not found when user request it.

talregev:
Cmake: Error on libidn2 not found when user request it.


talregev:
As I see it, USE_LIBIDN2 should be off by default, because user need to provide it.
Because there is no error msg that libidn2 is found, that was lead to hidden bug in vcpkg. that user request libidn2, but didn't get it. that I fix.

CMakeLists.txt
```diff
@@ -856,13 +856,16 @@ if(CURL_DISABLE_LDAP)
...
-option(USE_LIBIDN2 "Use libidn2 for IDN support" ON)
+option(USE_LIBIDN2 "Use libidn2 for IDN support" OFF)
 if(USE_LIBIDN2)
   check_library_exists("idn2" "idn2_lookup_ul" "" HAVE_LIBIDN2)
   if(HAVE_LIBIDN2)
     set(CURL_LIBS "idn2;${CURL_LIBS}")
     check_include_file_concat("idn2.h" HAVE_IDN2_H)
   endif()
+  if(NOT HAVE_LIBIDN2 OR NOT HAVE_IDN2_H)
+    message(FATAL_ERROR "libidn2 not found")
```
vszakats, line 867:
```suggestion
    message(STATUC "libidn2 requested, but not found")
    set(USE_LIBIDN2 OFF)
```
What about just resetting `USE_LIBIDN2` to `OFF` and showing a warning instead?
This way we could leave it on by default and let it be used automatically e.g. on systems where it's present. Also retaining compatiblity with existing builds.

vszakats:
Changing the default to `OFF` will break existing builds where IDN2 was present and found automatically (e.g. Linux).

Wouldn't it be a better fix to reset `HAVE_LIBIDN2` to `OFF` when the header is not found? Optionally with a warning telling `idn2 requested, but not found'?:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -862,6 +862,10 @@ if(USE_LIBIDN2)
   if(HAVE_LIBIDN2)
     set(CURL_LIBS "idn2;${CURL_LIBS}")
     check_include_file_concat("idn2.h" HAVE_IDN2_H)
+    if(NOT HAVE_IDN2_H)
+      message(FATAL_ERROR "libidn2 requested, but not found")
+      set(HAVE_LIBIDN2 OFF)
+    endif()
   endif()
 else()
   set(HAVE_LIBIDN2 OFF)
```

talregev:
> edit: deleted change proposal. A warning only makes sense if it's OFF by default, however, I believe the current ON default is reasonable to find this on systems providing it.

Ubuntu is not providing it automatic. Even in your tests, when I put it on, the cmake tests on ubuntu are failed because it don't found it.
And this is the problem, that you don't aware that you don't have the idn2 in your system, until there will some tests. 

The biggest change is for the repo that ship curl that provide the idn2 withnin, they will need to set this flag on. Maybe it already happen that they set USE_LIBIDN2=on, and if not, we can notify them. (I am not sure they compile it with cmake) ....

I will share with you the bug I fixing:
https://github.com/microsoft/vcpkg/pull/38160

vszakats:
One sure way to check if all the desired features (and no undesired ones) got enabled is to run `curl --version` on the binary. There is also an `Enabled protocols:` line at the end of the CMake generator output.

On other systems/distros libidn2 may be there by default, or may have been installed explicitly.
curl's autotools build picks it up by default too, so it's reasonable to keep it that way also in CMake.

This workaround works to configure libidn2 (for a custom libidn2 build) without requiring a patch:
https://github.com/curl/curl-for-win/blob/46b202efb3bf8eb5b3119706bae1064db20f836d/curl-cmake.sh#L356-L360
```shell
cmake [...] -DUSE_LIBIDN2=ON 'CMAKE_C_FLAGS=-I<path-to-libidn2>/include -L<path-to-libidn2>/lib -lidn2'
```

talregev:
This workaround will not work for vcpkg, because this is more complex system, and you cannot manipulate cflags for one port.
And I don't think I can change your approach. 
I will close this PR.
Thank you for your time to consider my approach.

PR #13366, opened 2024-04-14
os400: post-shellcheck changes adjustments

monnerat:
Build scripts must be executed by the os/400 shell (sh), not bash which is a PASE program.
Shell function get_make_vars() escaping reworked to match $() subcommand construct.

Fixup to 8a622ba
Closes #13366

/cc @vszakats


vszakats (approved):

packages/OS400/initscript.sh
```diff
@@ -278,15 +279,16 @@ get_make_vars()
...
-                -e 's/@\\([A-Za-z0-9_]*\\)@/${\\1}/g'                   \
-                -e 's/[[:space:]]*=[[:space:]]*/=/'                     \
-                -e 's/=\\(.*[^[:space:]]\\)[[:space:]]*$/=\\"\\1\\"/'   \
-                -e 's/\\\$(\\([^)]*\\))/\${\\1}/g'                      \
-                < \""${1}"\")"
+                -e 's/[[:space:]][[:space:]]*/ /g'                      \
+                -e '/^[A-Za-z_][A-Za-z0-9_]* *=/!d'                     \
+                -e 's/@\([A-Za-z0-9_]*\)@/${\1}/g'                      \
+                -e 's/ *= */=/'                                         \
+                -e 's/=\(.*[^ ]\) *$/=\"\1\"/'                          \
```
vszakats, line 291:
```suggestion
                -e 's/=\(.*[^ ]\) *$/="\1"/'                            \
```
Can we save escaping the double-quotes? (it works in my limited test)

monnerat: Works on the os/400 too. Thanks for the suggestion. Force pushed.

vszakats:
Thank you @monnerat, merged now!

PR #13367, opened 2024-04-14
mprintf: check fputc error rather than matching returned character

monnerat:
OS/400 ascii fputc wrapper deviates from the posix standard by the fact that it returns the ebcdic encoding of the original ascii character. Testing for a matching value for success will then always fail.

This commit replaces the chariacter comparison by an explicit error return check.

Fixup to ef2cf58c77fc92960695d6669648f92ca4792e9c


bagder (approved):

bagder:
Thanks!

PR #13372, opened 2024-04-15
docs: clarify CURLOPT_MAXFILESIZE and CURLOPT_MAXFILESIZE_LARGE

the-blank-x:
The bounds of the size parameter were not specified, and nor was it specified how to disable the maximum file size check.

The documentation also incorrectly stated that CURLOPT_MAXFILESIZE always returns CURLE_OK and that CURLOPT_MAXFILESIZE_LARGE only returns CURLE_OK or CURLE_UNKNOWN_OPTION.

It also did not mention what the default value is, which is zero. This commit updates the documentation to make note of all these things.


bagder (approved):
Excellent!

bagder:
Thanks!

PR #13364, opened 2024-04-14
cmake: forward `USE_LIBRTMP` option to C

talregev:
Define in c USE_LIBRTMP if user request it from the cmake.

PR #13373, opened 2024-04-15
cmake: add librtmp/rtmpdump option and detection

vszakats:
Add CMake option `USE_LIBRTMP`. Disabled by default.

This library requires OpenSSL TLS-backend when linked statically.

Follow-up to #13364
Closes #13373

---

Requires:
- [x] #13364


vszakats:
Limited local tests confirm that a build targeting 10.9 and using SDK 13 successfully detects and builds AppleIDN support, and the binary also runs and apparently works (converting an accented domain name to punycode) when run on a macOS that's actually missing the dynamically linked `/usr/lib/libicucore.A.dylib`. I don't know how this works.

If this isn't always true or my tests are wrong, AppleIDN should probably be disabled be default, or allow/block-listed in CMake based on `CMAKE_OSX_DEPLOYMENT_TARGET` or something.

PR #13374, opened 2024-04-15
h3, extend download abort tests, fixes in ngtcp2

icing:
- fix flow handling in ngtcp2 to ACK data on streams we abort ourself.
- extend test_02_23* cases to also run for h3
- skip test_02_23* for OpenSSL QUIC as it gets stalled on progressing the connection


bagder:
somehow the hyper builds broke a little?

bagder:
Ah no, they are broken in other PRs too...

PR #13384, opened 2024-04-16
ci: enable ECH with curl-for-win / cmake / boringssl

vszakats:
Closes #13384

### DIFF
--- a/.github/workflows/curl-for-win.yml
+++ b/.github/workflows/curl-for-win.yml
@@ -51,7 +51,7 @@ jobs:
             "${DOCKER_IMAGE}" \
             sh -c ./_ci-linux-debian.sh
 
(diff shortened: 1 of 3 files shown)


bagder:
with OpenSSL? Then you to build one from @sftcd's repo, right?

vszakats:
Indeed, I forgot that OpenSSL needs a custom patch for ECH... too early for this then.

I'll give BoringSSL a try, but chances are that those builds broke a while ago.

vszakats:
They are:
https://github.com/curl/curl/actions/runs/8702705205/job/23867331909?pr=13384#step:3:3707

(I wish BoringSSL offered a way to disable tests without rolling a large/fragile patch.)

PR #13386, opened 2024-04-16
tls session storage

icing:
- add session with destructor callback
- remove vtls `session_free` method
- let `Curl_ssl_addsessionid()` take ownership of session object, freeing it also on failures
- change tls backend use
-  added tests for SSL session resumption in `test_17_*`.

Update: removed the SNI testing commits again for a separate PR after this is merged.


bagder:
Can I ask you to squash this set of commits into the exact commits you think should be merged? This looks like more than one but probably fewer than 15, right?

icing:
> Can I ask you to squash this set of commits into the exact commits you think should be merged? This looks like more than one but probably fewer than 15, right?

Restore the PR again to be only about ssl sessions. Will add the SNI tests in another PR after this one has landed.

bagder (approved):

PR #13388, opened 2024-04-16
dmaketgz: release tarball generator script using docker

bagder:
For easier reproducibility.

Mention using this script in RELEASE-PROCEDURE

### DIFF
--- a/scripts/dmaketgz
+++ b/scripts/dmaketgz
@@ -41,6 +41,8 @@ docker build \
 
 run="run --rm -it -u $(id -u):$(id -g) -v $(pwd):/usr/src -w /usr/src curl/curl"
 
+# shellcheck disable=SC2086
+# "Double quote to prevent globbing and word splitting" on the $run use below
 docker $run autoreconf -fi
 docker $run ./configure --without-ssl --without-libpsl
 docker $run make -sj8
--- a/scripts/dmaketgz
+++ b/scripts/dmaketgz
@@ -1,0 +1,49 @@
+#!/bin/sh
+# docker-maketgz
+#
+#***************************************************************************
+#                                  _   _ ____  _
+#  Project                     ___| | | |  _ \| |
+#                             / __| | | | |_) | |
+#                            | (__| |_| |  _ <| |___
+#                             \___|\___/|_| \_\_____|
+#
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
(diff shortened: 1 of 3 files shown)


scripts/dmaketgz
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+make distclean
+docker build \
+       --build-arg SOURCE_DATE_EPOCH="$(date -u +%s)" \
+       --build-arg UID="$(id -u)" \
+       --build-arg GID="$(id -g)" \
+       -t curl/curl .
+
+run="run --rm -it -u $(id -u):$(id -g) -v $(pwd):/usr/src -w /usr/src curl/curl"
```
daniel-j-h, line 42: Ow, I believe now that we are creating a user with a home dir, we need to run this in the home dir, too

https://github.com/curl/curl/blob/41c03b4c98dbc639a32d32486ed5146be2e73ee1/Dockerfile#L33-L36

as in

```bash
docker run --rm -it -u $(id -u):$(id -g) \
  -v $(pwd):/home/dev/curl -w /home/dev/curl \
  curl/curl bash
```

this is what I tested locally and should get you files where the host:container user:group are the same.

bagder: Does it really matter where it runs? The generated output tarballs seem fine to me...

daniel-j-h: Interesting! I expected the `dev` user not having permissions to `/usr/src/` but if it works for you it works for you!

scripts/dmaketgz
```diff
@@ -0,0 +1,52 @@
...
+
+# shellcheck disable=SC2086
+# "Double quote to prevent globbing and word splitting" on the $run use below
+docker $run autoreconf -fi
+# shellcheck disable=SC2086
+docker $run ./configure --without-ssl --without-libpsl
+# shellcheck disable=SC2086
+docker $run make -sj8
+# shellcheck disable=SC2086
+docker $run ./maketgz $version
```
vszakats, line 52:
```suggestion
docker run --rm -it -u $(id -u):$(id -g) -v $(pwd):/usr/src -w /usr/src curl/curl sh -c "
  set -e
  autoreconf -fi
  ./configure --without-ssl --without-libpsl
  make -sj8
  ./maketgz $version"
```
This make it a bit more compact and avoids the shellcheck issue.

bagder: Neat! As I was too quick to merge the first version, I took your version to  #13391

PR #13391, opened 2024-04-16
dmaketgz: compacter

bagder:
Follow-up to d28f74913c2
Proposed-by: Viktor Szakats


vszakats (approved):

vszakats:
Perhaps a `set -eu` would be nice before `version=`, to make it stop on any error.

PR #13396, opened 2024-04-17
urlapi: add CURLU_GET_EMPTY for empty queries and fragments

bagder:
By default the API inhibits empty queries and fragments extracted. Unless this new flag is set.

This also makes the behavior more consistent: without it set, zero length queries and fragments are considered not present in the URL. With the flag set, they are returned as a zero length strings if they were in fact present in the URL.

This applies when extracting the individual query and fragment components and for the full URL.


bagder:
We could possibly change this to be one by default, but I'm a little scared how such a behavior change could surprise users.

We could also possibly consider adding two separate flags for this, but I figured users are more likely to want consistent behavior in either direction for both components.

chris-morgan:
I come here having discovered this behaviour a few days ago.

It is beyond dispute that stripping empty queries and fragments changes URLs’ meaning in ways that *can* break them. (I myself *almost* decided to publish [a page](https://chrismorgan.info/no-query-strings) at the exact URL `https://chrismorgan.info/?`, and may yet do so in the future. All browsers would be fine with this, but apparently this would make the page inaccessible to curl!)

It seems to me a bad idea for a foundational library to deliberately change the meaning of URLs it receives. I don’t know if curl has precedent for that or not (e.g. does curl ever change percent-encoding?).

> We could possibly change this to be on by default, but I'm a little scared how such a behavior change could surprise users.

This confuses me. Isn’t having it *off* the behaviour change? Perhaps I’m misunderstanding something. But speaking of the command line interface: before curl 8.8.0, an empty query string would be retained, but curl 8.8.0 was a breaking change, removing a part of the URL against my will.

PR #13397, opened 2024-04-17
result of Curl_bufq_cread is not used

MonkeybreadSoftware:
Since result is not used, we can ignore it.
Curl_bufq_cread will not return an error since the the empty check is done before.

Alternatively we can use `if (result) return result;` to exit there right away and pass up the error code.

This silences a warning in my Xcode curl test project.


bagder (changes requested):
I think this is the wrong fix. I think the `cr_eob_read` function should instead be made to `return result;`.

MonkeybreadSoftware:
Okay, so let's close it here and move on to
https://github.com/curl/curl/pull/13398

PR #13398, opened 2024-04-17
result of Curl_bufq_cread was not used

MonkeybreadSoftware:
We pass back the result to the caller.


bagder (commented):
I think the `cr_eob_read` function should instead be made to `return result;`. Not return early. That looks more like what was intended.

@icing what do you think?

MonkeybreadSoftware:
Yes, otherwise we would miss the last lines in the functions.

icing:
This looks correct.

bagder:
Thanks!

PR #13401, opened 2024-04-17
ngtcp2: fix macro use

bagder:
macro "H3_STREAM_CTX" requires 2 arguments, but only 1 given

Follow-up to c6655f7029ec5c128561e3ecf1f93db3ed0432a4


bagder:
For some reason I did not see this in CI before the merge.

icing:
That code was brought in by another PR and the combination was never run in CI.

PR #13419, opened 2024-04-19
Curl_client_read: init two variables to avoid using them uninited

bagder:
Within the function we trace the values independently of the Curl_creader_read() returning error or not, which could leave them uninitialized.

Spotted by CodeSonar


icing:
Hmm, I do prefer - maybe just taste - to have a call with `out` parameters to always set values. The caller should not have to worry about that.

Which probably means the fix should be in `Curl_creader_read()` itself?

bagder:
Like this?

icing (approved):

PR #13429, opened 2024-04-19
explicitly include `MacTypes.h`

isaactorz:
The latest macos sdk release - 15.3, seems to have broken a transitive include for OSStatus types used in sectransp.c

The fix here is to explicitly include the header that declares these symbols.

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -54,6 +54,7 @@
 
 #include <limits.h>
 
+#include <MacTypes.h>
 #include <Security/Security.h>
 /* For some reason, when building for iOS, the omnibus header above does
  * not include SecureTransport.h as of iOS SDK 5.1. */
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -56,7 +56,6 @@
 
 #include <MacTypes.h>
 #include <Security/Security.h>
-#include <Security/SecCertificate.h>
 /* For some reason, when building for iOS, the omnibus header above does
  * not include SecureTransport.h as of iOS SDK 5.1. */
 #include <Security/SecureTransport.h>
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -56,6 +56,7 @@
 
 #include <MacTypes.h>
 #include <Security/Security.h>
+#include <Security/SecCertificate.h>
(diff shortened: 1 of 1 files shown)


isaactorz:
It turns out this is not the only missing include...

Will update in a sec

lib/vtls/sectransp.c
```diff
@@ -54,7 +54,9 @@
 
 #include <limits.h>
 
+#include <MacTypes.h>
 #include <Security/Security.h>
+#include <Security/SecCertificate.h>
```
nickzman, line 59: I just checked this in Xcode 15.3, and the `Security/Security.h` umbrella header already includes SecCertificate.h. Are you sure this change is necessary?

isaactorz: Ah you're right

I just tried it since after applying the first patch I was still getting `undeclared identifier` errors complaining about symbols that are declared in `SecCertificate.h`..

But I'm still getting the failure so obviously something else is up. Trying a clean build now.

isaactorz:
> It turns out this is not the only missing include...
> 
> Will update in a sec

This turned out to be an issue with our build configurations - but the `MacTypes` issue was valid.

nickzman:
Is it? I'm able to build the latest source using Xcode 15.3 with no problems, so I'm not convinced this change is needed.

bagder:
@isaactorz ?

isaactorz:
> Is it? I'm able to build the latest source using Xcode 15.3 with no problems, so I'm not convinced this change is needed.

Sorry for the delayed response.

I think I might've jumped the gun with this change - I meant valid in the sense that it's a transitive include that AFAIK isn't pulled in by some dedicated omnibus header.

We got caught off guard by lots of mac builds suddenly breaking around this and related errors and this seemed to be necessary to fix the problem - so I wanted to upstream it quickly in case we could avoid using a fork or adding custom patches.

What I ended up doing is re-evaluating why we're building macos native (and deprecated TLS), when we are already using openssl. So I ended up just building curl against that to fix the problem. We are also using a non-standard build system so this error is going to be hard to reproduce outside of our codebase.

In the end this is an extremely trivial change that'll unlikely to majorly hurt or help anyone so I am totally ok with closing this!

Thanks for the responsiveness and I'll be careful to not jump the gun if I ever end up contributing here again in the future.

jay:
Thanks for the update, closing

PR #13436, opened 2024-04-21
Enable test 1117 for hyper HTTP backend as it currently works

Alvenix:
This PR is intended to test the CICD for now, as when I tried this test with the latest hyper it worked.

Edit:
It seems the test is successful on CICD, so it may be enabled. (Sorry If I missed something)


bagder (approved):

bagder:
Thanks!

PR #13437, opened 2024-04-21
sendf: useless assignment in cr_lc_read()

bagder:
Spotted by CodeSonar


bagder:
The circleci problems: #13438

Issue #13428, opened 2024-04-19
domain name with trailing dot fails the certificate check when using gnutls


ryandesign:
When built using gnutls, curl cannot connect to `https://pyropus.ca.` possibly because the period at the end of the hostname is not being stripped for SNI purposes. This problem does not occur when curl is built using openssl. I'm not sure whether this is a curl bug or a gnutls bug.

For prior history on issues accessing this server, which intentionally uses a canonical name ending in a period, see #8290 reported by @ccazabon. There, the problem was that curl *was* stripping the period from the hostname for non-SNI purposes when it should not have been doing so.

### I did this

```
curl -I https://pyropus.ca./
```

### I expected the following

```
HTTP/1.1 200 OK
Date: Fri, 19 Apr 2024 16:21:47 GMT
Server: Apache/2.4.57 (Debian)
Accept-Ranges: bytes
Vary: Accept-Encoding
X-Frame-Options: sameorigin
Content-Type: text/html
```

### but I got this

```
curl: (60) SSL: certificate subject name (pyropus.ca) does not match target host name 'pyropus.ca.'
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

### curl/libcurl version

```
curl 8.7.1 (x86_64-apple-darwin21.6.0) libcurl/8.7.1 GnuTLS/3.7.10 zlib/1.3.1 brotli/1.1.0 zstd/1.5.6 libidn2/2.3.7 libpsl/0.21.5 nghttp2/1.61.0 ngtcp2/1.4.0 nghttp3/1.2.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Largefile libz NTLM PSL SSL threadsafe TLS-SRP UnixSockets zstd
```

### operating system

Darwin hostname-redacted 21.6.0 Darwin Kernel Version 21.6.0: Mon Feb 19 20:24:34 PST 2024; root:xnu-8020.240.18.707.4~1/RELEASE_X86_64 x86_64

bagder:
[...]

PR #13440, opened 2024-04-22
hostnames with trailing dot, testcases, fix for gnutls

icing:
- add test_01_07 for successful verification with a trailing dot
- add test_01_08 for expected failure with double dot
- fix gnutls to use the normalized SNI name for checking when available
- refs #13428

### DIFF
--- a/lib/vtls/gtls.c
+++ b/lib/vtls/gtls.c
@@ -962,6 +962,9 @@ Curl_gtls_verifyserver(struct Curl_easy *data,
   gnutls_protocol_t version = gnutls_protocol_get_version(session);
 #endif
   long * const certverifyresult = &ssl_config->certverifyresult;
+  /* if we have use SNI, verify the certificate with that. If not, e.g.
+   * the hostname is an IP address, use the hostname */
+  const char *verify_name = peer->sni? peer->sni : peer->hostname;
 
 #ifndef CURL_DISABLE_VERBOSE_STRINGS
   /* the name of the cipher suite used, e.g. ECDHE_RSA_AES_256_GCM_SHA384. */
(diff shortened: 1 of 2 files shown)


bagder:
Back to the never-ending mess that is trailing dots. AFAIK, it should not use the SNI name to verify the host name in the certificates.

RFC 2818 section 3.1 says:

> In general, HTTP/TLS requests are generated by dereferencing a URI.
   As a consequence, the hostname for the server is known to the client.
   If the hostname is available, the client MUST check it against the
   server's identity as presented in the server's Certificate message,
   in order to prevent man-in-the-middle attacks.

The **hostname** here has a trailing dot. The SNI name does not. Should the verification not get done with the hostname?

RFC 9525 section 6.1 is less specific but says:

> The inputs used by the client to construct its list of reference identifiers might be a URI that a user has typed into an interface (e.g., an HTTPS URL for a website)

Again, that is the hostname. Not the SNI name.

But... clearly these TLS libraries fail to verify the name with the dot...

icing:
Yes, the mess with trailing dots in dns names. Not sure what is the best way forward.

We set `peer->sni` if it is a DNS name, do a strlower and remove one trailing dot if present. We have then two options

1. always use `peer->hostname` and let verification fail if the TLS library does not handle a trailing dot.
2. use our `peer->sni` to work around for these libraries.

I am not sure what is better. Please advise and I adapt the PR. At least we'll have test cases then.

bagder:
I've asked GnuTLS to clarify the situation (they want the name provided without a trailing dot) in their docs: https://gitlab.com/gnutls/gnutls/-/issues/1548

bagder:
A TLS server cannot make a difference between `name` and `name.` since the latter is not legal and thus it will always assume the name has no trailing dot. I think we can (under protest) agree to passing in the SNI name in the hostname field if that is what the TLS library insists we need to do.

icing:
I am closing this. I moved the test cases over to #13386 and reverted the gnutls workaround (for now).

PR #13442, opened 2024-04-22
mbedTLS: Implemented CURLOPT_SSL_CIPHER_LIST option

jan2000:
Use a lookup list to set the cipher suites, allowing the ciphers to be set by either openssl or IANA names.

To keep the binary size of the lookup list down we compress each entry in the cipher list down to 2 + 6 bytes using the C preprocessor.


lib/vtls/mbedtls.c
```diff
@@ -358,6 +360,95 @@ set_ssl_version_min_max(struct Curl_cfilter *cf, struct Curl_easy *data)
   return result;
 }
 
+#ifdef MBEDTLS_TLS_ECJPAKE_WITH_AES_128_CCM_8
+static int
+mbed_cipher_suite_get_str(uint16_t id, char *buf, size_t buf_size,
+                          bool prefer_rfc)
+{
+  if(id == MBEDTLS_TLS_ECJPAKE_WITH_AES_128_CCM_8)
+    strcpy(buf, "TLS_ECJPAKE_WITH_AES_128_CCM_8");
```
jan2000, line 369: `strcpy` is not bound by `buf_size`, do something like `msnprintf(buf, buf_size, "%s", ...);` instead

bagder: but then you'll get a truncated (broken) name if the buffer is too small.  It seems better to check if the name fits and return an error if it does not fit.

lib/vtls/cipher_suite.c
```diff
@@ -0,0 +1,697 @@
...
+
+    /* determine the length of the part */
+    cur = nxt;
+    for(; nxt < end && *nxt != '\0' && *nxt != separator; nxt++);
+    len = nxt - cur;
+
+    /* lookup index for the part (skip empty string at 0) */
+    for(idx = 1, entry = cs_txt + 1; idx < CS_TXT_LEN; idx++) {
+      size_t elen = strlen(entry);
+      if(elen == len && strncasecmp(entry, cur, len) == 0)
```
bagder, line 565: strncasecmp is a no no

lib/vtls/mbedtls.c
jan2000: Changed to `snprintf` and clarified in the comments of the wrapped function `Curl_cipher_suite_get_str` that  the caller is responsible to provide a large enough buffer to prevent string truncation. The max string length of a cipher suite name is predictable, so providing a buffer of 64 bytes should suffice.

lib/vtls/mbedtls.c
```diff
@@ -358,6 +360,95 @@ set_ssl_version_min_max(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+}
+
+static uint16_t
+mbed_cipher_suite_walk_str(const char **str, const char **end)
+{
+  uint16_t id = Curl_cipher_suite_walk_str(str, end);
+  size_t len = *end - *str;
+
+  if(id == 0) {
+    if(strncasecmp("TLS_ECJPAKE_WITH_AES_128_CCM_8", *str, len) == 0)
```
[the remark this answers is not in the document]
jan2000: Case insensitive is wanted to align the with other backends. Other changes implemented.

lib/vtls/mbedtls.c
```diff
@@ -358,6 +360,95 @@ set_ssl_version_min_max(struct Curl_cfilter *cf, struct Curl_easy *data)
   return result;
 }
 
+#ifdef MBEDTLS_TLS_ECJPAKE_WITH_AES_128_CCM_8
```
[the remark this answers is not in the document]
jan2000: Added comment.

lib/vtls/mbedtls.c
```diff
@@ -1027,6 +1132,8 @@ static void mbedtls_close(struct Curl_cfilter *cf, struct Curl_easy *data)
 #ifdef MBEDTLS_X509_CRL_PARSE_C
   mbedtls_x509_crl_free(&backend->crl);
 #endif
+  free((void *) backend->ciphersuites);
+  backend->ciphersuites = NULL;
```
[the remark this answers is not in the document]
jan2000: Changed accordingly.

lib/vtls/cipher_suite.c
jan2000: Updated to use `strncasecompare`.

bagder:
Ah, the test case got a merge conflict now as another commit unfortunately was merged that used the same number.

jan2000:
Ok, I have rebased to latest master and updated to test 3205.

lib/vtls/cipher_suite.c
```diff
@@ -0,0 +1,699 @@
...
+    for(i = 0; i < CS_LIST_LEN; i++) {
+      if(memcmp(cs_list[i].zip, zip, sizeof(zip)) == 0)
+        return cs_list[i].id;
+    }
+  }
+
+  return 0;
+}
+
+static bool is_separator(char c)
```
vszakats, line 648:
```suggestion
static bool cs_is_separator(char c)
```
and same for callers. To avoid the chance of overlapping with another function with the same name.

jan2000: Sure can do, but is it really necessary though? As this is static function. BTW the exact same function can be found in vtls/sectransp.c and vtls/bearssl.c.

vszakats: It's common failure mode for "Unity" builds where all sources are merged together into a single C file and then compiled.

This update would fix SecureTransport + mbedTLS MultiSSL builds with CMake with Unity mode enabled.

The BearSSL / SecureTransport cases also need to be fixed. Thanks for pointing to this.

jan2000: Oh I did not know about unity builds, seems they were introduced after my last contribution. (I must say I have the same concerns as jay though about it: https://github.com/curl/curl/pull/11095#issuecomment-1541418372 )

Will fix, what is preferred: add a commit or overwrite commit + force push?

I have PR's for BearSSL and SecureTransport lined up for consideration based on this PR if this gets merged.

vszakats: Thanks, no strong perference, adding a commit is fine IMO.

For the BearSSL / SecureTransport I opened #13450 earlier.

jan2000: > no strong perference

Ok, in this case I prefer overwrite + force. Changed code as suggested.

bagder (approved):

bagder:
Thanks!

Issue #13438, opened 2024-04-21
Circle CI openssl breakage on macos


bagder:
### I did this

I made [this PR](https://github.com/curl/curl/pull/13437)

The regular x86 macos CI jobs on Circle CI all fail, with unrelated problems. It seems configure cannot find OpenSSL in those builds.

[example build](https://app.circleci.com/pipelines/github/curl/curl/8274/workflows/cbd411bd-ccd5-4a0e-9def-467a210aee5f/jobs/81537)

OpenSSL is installed with homebrew and nothing has been changed in that setup recently. We also do OpenSSL builds on macOS on GitHub actions that work, so it does not seem to be an obvious homebrew related glitch?

### I expected the following

The CI jobs to work fine.

### curl/libcurl version

current git master

### operating system

CI jobs on macOS

bagder:
Probably related, Circle CI will shut down their x86 macos machines on June 28th:

## Subject: macOS Intel support deprecation 

Hello,

On June 28th, 2024 we will end support for our Intel macOS resources. This change impacts all
customers using Intel Mac resources, across all plans.

In November 2020, Apple released the first Macs with an ARM-based M1 chip and announced that the
transition from Intel processors to their own silicon-based processors would take about two years.
Now that we’re well past that two-year timeframe, we believe we can best serve our customers by
focusing on delivering the latest Apple Silicon resources moving forward.

Any customer that has a macOS project running on an Intel resource will need to update their config
to use a new resource class by June 28th, 2024.

dfandrich:
As of commit d7332e3e4 (last January) we shouldn't have any more x86 builds on Cirrus.

bagder:
I presume you mean Circle CI, not Cirrus.

Oh! I now think my mention of `x86` in the description above is wrong and they are all indeed aarch64 builds. I'll update accordingly.

bagder:
[log](https://app.circleci.com/pipelines/github/curl/curl/8289/workflows/12f4c6e7-0c8c-46f6-9054-93d2f755e936/jobs/81814/parallel-runs/0/steps/0-103)
~~~
ld: library 'crypto' not found
~~~

It looks like maybe it gets the wrong path from pkg-config?

~~~
configure: pkg-config: SSL_LDFLAGS: "-L/opt/homebrew/Cellar/openssl@3/3.3.0"
~~~

vszakats:
[...]

PR #13445, opened 2024-04-23
tests: fix feature case in test1481

dfandrich:
This test was being skipped everywhere because the feature never
matched.


bagder (approved):

bagder:
the macOS CI failures: #13438

Issue #13439, opened 2024-04-21
Quiche CI job occasionally fails during a download abort test


jay:
For the last few days there are occasional CI failures in the [quiche job](https://github.com/curl/curl/actions/workflows/quiche-linux.yml). It is hard to diagnose them because the logs are so large due to millions of "curl_multi_poll() -> 0" output (from [h2-download.c](https://github.com/curl/curl/blob/727c946d823750f4e7f33e3136a60aa8cda7ae95/tests/http/clients/h2-download.c#L376-L457)). The output is truncated and downloading the raw archive fails repeatedly.

The most recent commit to have the failure: "[http: acknowledge a returned error code](https://github.com/curl/curl/actions/runs/8771459833)". After repeated attempts github finally gave me the whole file. I stripped out most of the excess `curl_multi_poll() -> 0` to make it readable: [log.txt](https://github.com/curl/curl/files/15053465/log.txt). Excerpt:

~~~
2024-04-19T13:23:10.4146231Z =================================== FAILURES ===================================
2024-04-19T13:23:10.4147994Z _______________ TestDownload.test_02_23a_lib_abort_paused[0-h3] ________________
2024-04-19T13:23:10.4148530Z 
2024-04-19T13:23:10.4148762Z self = <test_02_download.TestDownload object at 0x7f6229f10220>
2024-04-19T13:23:10.4149272Z env = <testenv.env.Env object at 0x7f622a029de0>
2024-04-19T13:23:10.4151755Z httpd = <testenv.httpd.Httpd object at 0x7f622a0c8cd0>
2024-04-19T13:23:10.4152554Z nghttpx = <testenv.nghttpx.NghttpxQuic object at 0x7f6229e01a50>, proto = 'h3'
2024-04-19T13:23:10.4153075Z repeat = 0
2024-04-19T13:23:10.4153232Z 
2024-04-19T13:23:10.4153536Z     @pytest.mark.parametrize("proto", ['http/1.1', 'h2', 'h3'])
2024-04-19T13:23:10.4154389Z     def test_02_23a_lib_abort_paused(self, env: Env, httpd, nghttpx, proto, repeat):
2024-04-19T13:23:10.4155785Z         if proto == 'h3' and env.curl_uses_ossl_quic():
2024-04-19T13:23:10.4156581Z             pytest.skip('OpenSSL QUIC fails here')
2024-04-19T13:23:10.4157297Z         if proto in ['h2', 'h3']:
[...]

PR #13447, opened 2024-04-23
quiche, expire all active transfers on connection close

icing:
- when a connection close is detected, all ongoing transfers need to expire bc no more POLL events are likely to happen for them.
- refs #13439


bagder:
The AppVeyor CI fails have to be independent from this.

Issue #13451, opened 2024-04-23
Change in behaviour for duplicate "Transfer-Encoding = chunked" headers


bacchanalia:
### I did this

Prior to 8.6.0 (specifically d7b6ce64ce0ad787ad2ed3ee05c94938a6b4f551), duplicate "Transfer-Encoding = chunked" headers were ignored, but now curl will attempt to nest an unencoder for each duplicate header. I believe the change was unintended in that I couldn't find it discussed either on github on in the release notes.  I'm not sure whether the new behavior is bug or if it is an inadvertent fix for a latent bug. If the new behavior is preferred, I believe it should be clearly documented.

### I expected the following

For user facing behavior to either remain the same or to be clearly documented in the change log.

### curl/libcurl version

curl 8.6.0 (x86_64-pc-linux-gnu) libcurl/8.6.0 OpenSSL/3.0.13 zlib/1.3.1 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.7 libpsl/0.21.5 libssh2/1.11.0 nghttp2/1.60.0
Release-Date: 2024-01-31
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

Linux 5.15.133 #1-NixOS SMP Sat Sep 23 09:10:03 UTC 2023 x86_64 GNU/Linux

bagder:
Duplicate chunked headers like this is probably not okay according to the spec, but I also realize that they occur in the wild and that A) browsers survive them and B) curl use to survive them. I think we should restore curl handling multiple ones.

bacchanalia:
Wow, that was fast. Thanks!

PR #13461, opened 2024-04-24
http chunked decoding, ignore duplicates

icing:
- ignore duplicate "chunked" transfer-encodings from a server to accomodate for broken implementations
- add test1482
- refs #13451


bagder:
Maybe add a separate test if there are two `Transfer-Encoding: chunked`  headers?

icing:
> Maybe add a separate test if there are two `Transfer-Encoding: chunked` headers?

Done.

bagder:
GHA problems:  #13463

bagder (approved):

PR #13462, opened 2024-04-24
CI: show more failed `config.log` on Circle CI

vszakats:
Show last 1000 lines of `config.log` if `./configure` fails. This was
already done for one job, this patch extends it to all.

Ref: #13438
Closes #13462

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,49 +36,49 @@ commands:
       - run:
           command: |
             autoreconf -fi
-            ./configure --enable-warnings --enable-werror --with-openssl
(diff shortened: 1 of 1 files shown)


vszakats:
Needs merging to master to see if this turns up something regarding #13438.

bagder (approved):

dfandrich:
This will likely turn the configure failure result code into a tail success
code, causing the build to (attempt to) continue. Better would be
"|| { tail -1000 config.log; false; }"

vszakats:
Good idea! → PR #13468

PR #13464, opened 2024-04-24
bearssl: Use common code for cipher suite lookup

jan2000:
Take advantage of the Curl_cipher_suite_walk_str() and Curl_cipher_suite_get_str() functions introduced in commit fba9afeb.

This also fixes CURLOPT_SSL_CIPHER_LIST not working at all for bearssl due to commit ff74cef5.


bagder (approved):

bagder:
Thanks!

PR #13465, opened 2024-04-24
mqtt: Read variable header indepentenly from payload

schicho:
# MQTT read variable header indepentenly from payload

Started by me in a discussion at https://github.com/curl/curl/discussions/13448

---

## current upstream implementation

The MQTT _publish_ packet sent by a broker to a subscribing client is only parsed for the fixed header indicating the packet type.

However, the packet itself still contains a variable header indicating the topic length, topic and packet ID. http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718039 (Section 3.3.2)

The whole variable header and the following payload is output by the curl tool. Yet, we know that the first two bytes are the topic length, which, given that it is significantly short, results in the first byte being `x00`. That results in curl aborting the output and warning of binary output. This makes curl unsuitable for testing any MQTT functionality on the command line.

## proposed change

Break the current implementation and provide the variable header values topic and header ID as header values hidden behind a -v flag like with HTTP.

The output will now only be the actual payload of the packet.

You may want to test this with the `-N` flag. Otherwise you won't see the output immediately.

### further help needed

- The topic name can be up to 32KB. What is the correct cURL way to accumulate that data and provide it back to the user? Is a dynbuf the correct approach? Or can we allocate it all on the stack?
- The reading from the socket and the accompanying error handling is duplicated throughout the code. What is the correct cURL approach to minimize this?

### DIFF
--- a/lib/mqtt.c
+++ b/lib/mqtt.c
@@ -714,12 +714,8 @@ static CURLcode mqtt_read_pub_varheader(struct Curl_easy *data)
   if(rest > sizeof(buffer))
     rest = sizeof(buffer);
   result = Curl_xfer_recv(data, buffer, rest, &nread);
-  if(result) {
-    if(CURLE_AGAIN == result) {
-      infof(data, "E AGAIN");
-    }
+  if(result)
     goto end;
-  }
   if(!nread) {
     infof(data, "server disconnected");
     result = CURLE_PARTIAL_FILE;
@@ -794,12 +790,8 @@ static CURLcode mqtt_read_pub_varheader(struct Curl_easy *data)
       mq->npacket -= nread;
       nprocessed = 0;
       result = Curl_xfer_recv(data, buffer, sizeof(buffer), &nread);
-      if(result) {
-        if(CURLE_AGAIN == result) {
-          infof(data, "E AGAIN");
-        }
+      if(result)
         goto end;
-      }
       if(!nread) {
         infof(data, "server disconnected");
         result = CURLE_PARTIAL_FILE;
@@ -898,6 +890,7 @@ MQTT_SUBACK_COMING:
     data->req.bytecount = 0;
     data->req.size = remlen;
     mq->npacket = remlen; /* get this many bytes */
+
     /* we are at the very start of the packet an need to read the header */
     result = mqtt_read_pub_varheader(data);
     if(result)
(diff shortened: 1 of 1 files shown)


schicho:
I have implemented a first draft for this now. Maybe I can receive some feedback on it. I am not an experienced C programmer at all, so help is very much appreciated.

lib/mqtt.c
```diff
@@ -617,6 +617,149 @@ static void mqstate(struct Curl_easy *data,
...
+  ssize_t nprocessed = 0;
+  char buffer[4*1024];
+  ssize_t ntopic;
+  size_t rest = mq->npacket;
+  if(rest > sizeof(buffer))
+    rest = sizeof(buffer);
+  result = Curl_xfer_recv(data, buffer, rest, &nread);
+  if(result) {
+    if(CURLE_AGAIN == result) {
+      infof(data, "E AGAIN");
```
bagder: I don't think it should infof() this, it will become highly annoying.

lib/mqtt.c
```diff
@@ -617,6 +617,149 @@ static void mqstate(struct Curl_easy *data,
...
+  }
+  if(!nread) {
+    infof(data, "server disconnected");
+    result = CURLE_PARTIAL_FILE;
+    goto end;
+  }
+
+  /* read topic */
+  if(nread < 2) {
+    failf(data, "packet too short. topic length missing");
```
bagder, line 727: Shouldn't a short read be okay and just continued at next read?

lib/mqtt.c
```diff
@@ -671,17 +814,26 @@ static CURLcode mqtt_read_publish(struct Curl_easy *data, bool *done)
...
+    /* we are in the middle of a packet */
     char buffer[4*1024];
     size_t rest = mq->npacket;
     if(rest > sizeof(buffer))
       rest = sizeof(buffer);
     result = Curl_xfer_recv(data, buffer, rest, &nread);
     if(result) {
       if(CURLE_AGAIN == result) {
-        infof(data, "EEEE AAAAGAIN");
+        infof(data, "E AGAIN");
```
bagder: I think this infof() should rather be removed.

bagder (commented):
I like this take and I think it is worth breaking old users for taking this step.

This also needs test cases and documentation.

lib/mqtt.c
schicho: Unfortunately, I am unfamiliar with all of curl's internal structure.

Can it happen that a call to `mqtt_read_publish` is made for only one byte, yet more data for that particular mqtt package will arrive later and handled in a subsequent call?

`mqtt_read_pub_varheader` is called only once for each `mqtt_read_publish` call, as I was under the assumption it's per mqtt package. So if for some reason the read bytes are less than 2, we cannot do anything with that information and error.

craigsprayer: Payload $

bagder:
abandoned with many test failures still, closing

PR #13469, opened 2024-04-24
tests: silence `-Wsign-conversion`, type tidy-ups

vszakats:
Silence and fix sign conversion issues and compiler warnings 
in tests. Also switch all unit tests to return `CURLcode`. Before
this patch all unit tests and about a third of libtests did this.

Also fix mask signedness and/or types in some `printf` calls.

Ref: #13467
Cherry-picked from #13489
Closes #13469

### DIFF
--- a/tests/http/clients/ws-data.c
+++ b/tests/http/clients/ws-data.c
@@ -225,8 +225,8 @@ int main(int argc, char *argv[])
     fprintf(stderr, "maxlen must be >= 0, got %ld\n", l2);
     return 2;
   }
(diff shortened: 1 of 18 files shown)


vszakats:
Tests building without any signedness warnings after this PR (and the preceding #13467):
https://github.com/curl/curl/actions/runs/8824107659/job/24225918645?pr=13469#step:34:32

vszakats:
Closing in favour of #13501 and cherry-picking from there. It has the same commits as this one.

PR #13473, opened 2024-04-25
ci: remove microsoft-prod.list

pheiduck:
This is added by default, and it is often broken, but we don't need anything from it.


vszakats:
Wouldn't this also apply to `codespell.yml`?

pheiduck:
> Wouldn't this also apply to `codespell.yml`?

Good catch, I missed that file.

vszakats (approved):

bagder:
Thanks!

PR #13476, opened 2024-04-25
Python PEP-0668 fallout

icing:
- brew's python follows PEP-0668 which no longer allows using 'pip' to install packages system-wide
- add a venv to the macos CI, see if that works for us


icing:
replaced with #13479

PR #13477, opened 2024-04-25
sendf: Curl_cwriter_write: remove comment disallowing zero length

schicho:
`Curl_client_write` calls `Curl_cwriter_write`, which already has this limitation in place in its comment.

blen is not checked in `Curl_client_write`. Stumbled upon this working on my other MQTT PR.

### DIFF
--- a/lib/sendf.c
+++ b/lib/sendf.c
@@ -173,8 +173,7 @@ void Curl_creader_set_rewind(struct Curl_easy *data, bool enable)
   data->req.rewind_read = !!enable;
 }
 
-/* Write data using an unencoding writer stack. "nbytes" is not
-   allowed to be 0. */
+/* Write data using an unencoding writer stack. */
 CURLcode Curl_cwriter_write(struct Curl_easy *data,
                              struct Curl_cwriter *writer, int type,
                              const char *buf, size_t nbytes)


icing:
A zero length write is allowed and sometimes also done. What are you trying to achieve?

schicho:
I was already surprised myself that a zero length write would be disallowed.
I am trying to achieve consistency across the comments.


`Curl_client_write` currently has no limitation on the size of `blen`
https://github.com/curl/curl/blob/master/lib/sendf.c#L70

some further lines down it calls `Curl_cwriter_write` with `blen` passed into it without earlier checks.
https://github.com/curl/curl/blob/master/lib/sendf.c#L91

`Curl_cwriter_write` is defined here and has a comment stating:
/* Write data using an unencoding writer stack. "nbytes" is not
   allowed to be 0. */
https://github.com/curl/curl/blob/master/lib/sendf.c#L178

I assumed the comment of `Curl_client_write` was missing this statement.
Do I have this backwards and instead the comment of  `Curl_cwriter_write` is incorrect?
I  would change my PR in that case.

icing:
Thanks for the fix.

icing (approved):

bagder (approved):

bagder:
Thanks!

PR #13478, opened 2024-04-25
GHA: prepare macos for ARM runners

vszakats:
GHA switched macOS runners to ARM64. Fixup hard-coded
Homebrew package dependency paths to dynamic ones.

```
Current runner version: '2.316.0'
Operating System
  macOS
  14.4.1
  23E224
Runner Image
  Image: macos-14-arm64
```
Ref: https://github.com/curl/curl/actions/runs/8834067375/job/24254892013?pr=13478#step:1:1

Closes #13478

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -66,11 +66,11 @@ jobs:
             macosx-version-min: 10.9
           - name: libssh-c-ares
             install: openssl nghttp2 libssh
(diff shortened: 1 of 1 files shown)


icing:
Like this. I think I take this as the base and add my other fixes from #13476 on top into a new PR.

vszakats:
Okay, good idea, I'm closing this one then.

PR #13482, opened 2024-04-25
build: check for access() availability

jay:
- Define HAVE_ACCESS when access() exists.

- When access() does not exist and readability cannot be checked via access(filename, R_OK) then use fopen(filename, "rb") instead.

Prior to this change there is some platform (???) curl failed to build on because access() is not available.

Bug: https://github.com/curl/curl/pull/13412#issuecomment-2065505415
Reported-by: farazrbx@users.noreply.github.com

Closes #xxxx

---

draft based on the assumption that there is some platform without access(). I set HAVE_ACCESS true in all the `config-<platform>` files, which might not be a good idea. I'm just assuming that it's available since nobody else has reported a missing access().


dfandrich:
access is POSIX, so not having it would likely mean lots of other functions
wouldn't be available. I say drop it and keep curl 146 lines smaller until
there's a demonstrable need for it.

vszakats:
Another option (with no added cost for all build/builder) is to add a `CURL_NO_ACCESS` flag which omits the 5 uses it has in curl:
```
lib/mime.c:    if(stat(filename, &sbuf) || access(filename, R_OK))
lib/vssh/libssh2.c:            else if(access(sshc->rsa, R_OK) != 0) {
lib/vssh/libssh2.c:              else if(access(sshc->rsa, R_OK) != 0) {
lib/vssh/libssh2.c:            if(sshc->rsa && access(sshc->rsa, R_OK) != 0) {
lib/vssh/libssh2.c:              if(sshc->rsa && access(sshc->rsa, R_OK) != 0) {
```

It may also be an option to omit the use in `mime.c`, and replace them in libssh2 with `open()`, or just drop? After all, open errors are detected on the actual `open()` anyway. Also safer, because it leaves no time gap between the `access()` check and `open()`, when permissions might change.

jay:
Closing in favor of #13498

PR #13483, opened 2024-04-25
TODO: add TLS channel binding

bk2204:
TLS channel binding is a security improvement that cryptographically ties authentication to a specific TLS connection, preventing TLS MITM proxies from successfully authenticating.  This would be valuable for many security-sensitive users, so add it to the TODO list.

### DIFF
--- a/docs/TODO
+++ b/docs/TODO
@@ -127,6 +127,7 @@
  13.14 Support the clienthello extension
  13.15 Select signature algorithms
  13.16 QUIC peer verification with wolfSSL
+ 13.17 TLS channel binding
 
  14. GnuTLS
  14.2 check connection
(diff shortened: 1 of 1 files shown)


danielgustafsson (approved):
No objection to adding it to the TODO

bk2204:
I noticed this had grown a conflict, so I rebased onto the latest version of the main branch and renumbered the new entry.

christophvw:
It is already supported with SCHANNEL backend.

and a pull request for openssl backend is available as well:
https://github.com/curl/curl/pull/13098

bk2204:
This proposal is not specific to SPNEGO.  I'm proposing an approach where a user can use SPNEGO _or_ a custom authentication scheme via a callback.

CuiTongyun:
@bk2204, May I know when could curl support TLS channel binding with OpenSSL. thanks.

bagder:
> @bk2204, May I know when could curl support TLS channel binding with OpenSSL. thanks.

There is no ETA. Feel free to make it happen.

PR #13484, opened 2024-04-26
tool_cb_hdr: allow etag and content-disposition for 3xx reply

jay:
- Parse etag and content-disposition headers for 3xx replies.

For example, a server may send a content-disposition filename header with a redirect reply (3xx) but not with the final response (2xx). Without this change curl would ignore the server's specified filename and continue to use the filename extracted from the user-specified URL.

Prior to this change, 75d79a4 had limited etag and content-disposition to 2xx replies only.

Reported-by: Morgan Willcock

Fixes https://github.com/curl/curl/issues/13302
Closes #xxxx

### DIFF
--- a/src/tool_cb_hdr.c
+++ b/src/tool_cb_hdr.c
@@ -118,8 +118,9 @@ size_t tool_header_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
     long response = 0;
     curl_easy_getinfo(per->curl, CURLINFO_RESPONSE_CODE, &response);
 
-    if(response/100 != 2)
-      /* only care about these headers in 2xx responses */
+    if((response/100 != 2) && (response/100 != 3))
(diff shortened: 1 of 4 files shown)


bagder (approved):

bagder:
It would be good to also get a test or two for this.

bagder:
I created two tests for this: 473 and 1487. If you first rebase your branch on a current master, the patches in this tarball should apply fine.

[tests.tar.gz](https://github.com/user-attachments/files/15811244/tests.tar.gz)

bagder:
I decided not to try to push these to your branch because I made them against a current master and they did not apply clean on your existing branch and then it quickly got complicated... If you want me to, I can branch off the work in a separate PR with your fix and my tests.

jay:
Thanks, I rebased and squashed them in, let's see what happens

bagder:
:heavy_check_mark:

PR #13486, opened 2024-04-26
Add tests for SNI and peer name checks

icing:
- connect to DNS names with trailing dot
- connect to DNS names with double trailing dot
- rustls, always give `peer->hostname` and let it figure out SNI itself
- add SNI tests for ip address and localhost
- document in code and TODO that QUIC with ngtcp2+wolfssl does not do proper peer verification of the certificate
- mbedtls, skip tests with ip address verification as not supported by the library

### DIFF
--- a/tests/http/test_17_ssl_use.py
+++ b/tests/http/test_17_ssl_use.py
@@ -101,4 +101,77 @@ class TestSSLUse:
             else:
                 assert djson['SSL_SESSION_RESUMED'] == exp_resumed, f'{i}: {djson}'
 
+    # use host name with trailing dot, verify handshake
+    @pytest.mark.parametrize("proto", ['http/1.1', 'h2', 'h3'])
+    def test_17_03_trailing_dot(self, env: Env, httpd, nghttpx, repeat, proto):
+        if env.curl_uses_lib('gnutls'):
+            pytest.skip("gnutls does not match hostnames with trailing dot")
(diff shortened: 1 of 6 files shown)


tests/http/test_17_ssl_use.py
```diff
@@ -101,4 +101,77 @@ def test_17_02_sslinfo_reconnect(self, env: Env, httpd, nghttpx, tls_max, repeat
...
+            # the SNI the server received is without trailing dot
+            assert r.json['SSL_TLS_SNI'] == env.domain1, f'{r.json}'
+
+    # use host name with double trailing dot, verify handshake
+    @pytest.mark.parametrize("proto", ['http/1.1', 'h2', 'h3'])
+    def test_17_04_double_dot(self, env: Env, httpd, nghttpx, repeat, proto):
+        if proto == 'h3' and not env.have_h3():
+            pytest.skip("h3 not supported")
+        if proto == 'h3' and env.curl_uses_lib('wolfssl'):
+            pytest.skip("wolfSSL HTTP/3 peer verification does not properly check")
```
bagder, line 127: We should file a proper issue about this wolfSSL verification problem so that it gets noticed by everyone who needs to see it, and we will not forget to fix it as easily.

bagder: #13487

bagder (approved):

PR #13481, opened 2024-04-25
lib: silence `-Wsign-conversion`, tidy-ups, fixes

vszakats:
`lib` building without signedness warnings after this PR.

Also:
- mingw declares `sin_family`/`sin6_family` as `short`.
- fix assert in `Curl_hash_add`.

Cherry-picked from #13489
Follow-up to 3829759bd042c03225ae862062560f568ba1a231 #12489
Closes #13481

PR #13472, opened 2024-04-25
lib/v*: silence `-Wsign-conversion`, tidy-ups, fixes

vszakats:
`lib/v*` building without signedness warnings after this PR.

Also:
- openssl: fix `ctx_option_t` type for openssl 1.1.x.
- openssl: add three missing negative checks.
- sectransp: fix `CURLcode`/`OSStatus` mix-up.
- gtls: add FIXMEs to `Curl_gtls_verifyserver()`.

Cherry-picked from #13489
Follow-up to 3829759bd042c03225ae862062560f568ba1a231 #12489
Closes #13472

PR #13470, opened 2024-04-25
src: silence `-Wsign-conversion`, type tidy-ups

vszakats:
`src` (curl tool) building without signedness warnings after this PR.

Also:
- add a negative check to `get_terminal_columns()`.
- tool_parsecfg: fix ParameterError mix-up.

Cherry-picked from #13489
Follow-up to 3829759bd042c03225ae862062560f568ba1a231 #12489
Closes #13470

PR #13489, opened 2024-04-27
build: enable `-Wsign-conversion` warnings and fix/silence them (OpenSSF)

vszakats:
- silence all `-Wsign-conversion` warnings by `size_t` and `curl_off_t` casts.
- build: enable (non-fatal) `-Wsign-conversion` warnings (OpenSSF).
- build: make `-Wsign-conversion` an error (OpenSSF).

Related fixes, including type changes or casts in interface boundaries:

- disable `-Wsign-conversion` warning for `FD_SET()`/`FD_ISSET()` macros. → #13896
  These trigger on some systems due to the macros' implementation.
- ~~fix signedness in some `printf` masks.~~ → #13600 #13862
- ~~tests: switch unit tests to return `CURLcode`.~~ → #13600
- ~~mingw declares `sin_family`/`sin6_family` as `short`.~~ → #13862
- ~~openssl: fix `ctx_option_t` type for openssl 1.1.x~~ → #13622
- ~~sectransp: fix `CURLcode`/`OSStatus` mix-up.~~ → #13622
- ~~tool_parsecfg: fix `ParameterError` mix-up.~~ → #13614
- ~~fix assert in `Curl_hash_add`.~~ → #13502 #13601

Closes #13489

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -493,11 +493,11 @@ CURLcode Curl_parse_interface(const char *input, size_t len,
     host_part = memchr(input, '!', len);
     if(!host_part || !*(host_part + 1))
       return CURLE_BAD_FUNCTION_ARGUMENT;
-    *iface = Curl_memdup0(input, host_part - input);
+    *iface = Curl_memdup0(input, (size_t)(host_part - input));
     if(!*iface)
       return CURLE_OUT_OF_MEMORY;
     ++host_part;
-    *host = Curl_memdup0(host_part, len - (host_part - input));
+    *host = Curl_memdup0(host_part, len - (size_t)(host_part - input));
     if(!*host) {
       free(*iface);
       *iface = NULL;
--- a/lib/curl_krb5.h
+++ b/lib/curl_krb5.h
@@ -32,8 +32,7 @@ struct Curl_sec_client_mech {
   void (*end)(void *);
   int (*check_prot)(void *, int);
   int (*encode)(void *, const void *, int, int, void **);
-  int (*decode)(void *, void *, int, enum protection_level,
-                struct connectdata *);
+  int (*decode)(void *, void *, int, int, struct connectdata *);
 };
 
 #define AUTH_OK         0
--- a/lib/curl_krb5.h
+++ b/lib/curl_krb5.h
@@ -32,7 +32,8 @@ struct Curl_sec_client_mech {
   void (*end)(void *);
   int (*check_prot)(void *, int);
   int (*encode)(void *, const void *, int, int, void **);
-  int (*decode)(void *, void *, int, int, struct connectdata *);
+  int (*decode)(void *, void *, int, enum protection_level,
+                struct connectdata *);
 };
 
 #define AUTH_OK         0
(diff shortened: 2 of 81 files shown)


vszakats:
PR now passes all CI builds and tests.

It did help finding small issues, but I haven't investigated all possible ones, only the trivial ones.
Some of the fixed ones were type disagreements, error code confusions, wrong types assumed
at API boundaries, and wrong casts.

About half of the patch is addressing `CURLcode`/`int` mixups in unit tests. They now all 
universally use `CURLcode`.

To my eyes the resulting code better expresses what's happening and more clearly describes 
component boundaries. It also makes it easier to spot (or grep) potential issues arising from 
switching between signed/unsigned size variables, IMO.

There is no exact number for the warnings fixed. There were about 450 in the first round and 
in the ballpark of 900 total after addressing all, disregarding unit tests. Relative to the size of
the codebase, this number doesn't seem to be high and suggest a clean codebase.

Of these, around 140 were silenced with the `(size_t)(ptr1 - ptr2)` pattern, which we discussed
earlier.

It's possible that warnings are still present in code not compiled in CI and/or build combinations 
triggering more.

Next steps?:
- make a list of fixes
- make a list of unrelated issues found
- split up the patch into smaller PRs, I was keeping tests, examples, lib, lib/*, src separate
  while working on this. I'd probably start with moving unit tests fixes to a separate PR,
  or merge them separately via #13469.

vszakats:
Separate PRs for the major curl components:
- tests: #13469
- curl tool: #13470
- libcurl: #13481
- libcurl backends: #13472
- examples: #13501

These are the same sub-commits as here, except two controlling the warning option, which is unique to this one.

lib/getinfo.c
```diff
@@ -361,7 +361,7 @@ static CURLcode getinfo_offt(struct Curl_easy *data, CURLINFO info,
 #ifdef DEBUGBUILD
   char *timestr = getenv("CURL_TIME");
   if(timestr) {
-    unsigned long val = strtol(timestr, NULL, 10);
+    unsigned long val = (unsigned long)strtol(timestr, NULL, 10);
```
bagder, line 364:
```suggestion
    unsigned long val = strtoul(timestr, NULL, 10);
```

lib/getinfo.c
```diff
@@ -450,7 +450,7 @@ static CURLcode getinfo_double(struct Curl_easy *data, CURLINFO info,
...
```
bagder, line 453:
```suggestion
    unsigned long val = strtoul(timestr, NULL, 10);
```

bagder (commented):
I just think that adding typecasts in hundreds of new places is not a big win. They just silence the warnings with code that is a tad bit harder to read, the same logic is still there.

The improvements done by changing variable types and prototypes are the interesting ones I think. A typecast is always a bit of a "I surrender" sign.

vszakats:
Just realized that the set of functions/macros in `lib/warnless.h` were created with a similar / overlapping goal as this PR. I'm thinking maybe after cherry-picking-off all the non-size_t (and alike) cases, it's an option to use the warnless features to silence the rest. It makes these fixes recognizable (greppable) while allowing to enable `-Wsign-conversion`.

Though it keeps me wondering what coding patterns could make these tricks unnecessary in the first place.

vszakats:
`-Wsign-conversion` is unwanted in curl.

PR #13498, opened 2024-04-29
libssh2: replace `access()` with `stat()`

vszakats:
Prefer `stat()` to verify the presence of key files.

This drops the last uses of `access()` in the codebase, which was
reported to cause issues in some cases.

Also add `access()` to the list of banned functions in checksrc.

Ref: https://github.com/curl/curl/pull/13412#issuecomment-2065505415
Ref: https://github.com/curl/curl/pull/13482#issuecomment-2078980522
Ref: #13497
Closes #13498


jay (approved):

jay:
I added a commit to your branch that adds access to the bannedfunc list so that we don't run into this again

bagder (approved):

PR #13501, opened 2024-04-29
examples: fix/silence `-Wsign-conversion`

vszakats:
Cherry-picked from #13489
Closes #13501


vszakats:
If no objections I'd go along merging this.

It has legitimate fixes and there is only one sign cast.

This one, if undesired, can be deleted or beefed up with sign checks. It's used
when converting the return value of `if_nametoindex` (int or long unsigned 
depending on platform) to the `long` needed by `curl_easy_setopt()`.

PR #13512, opened 2024-04-30
file: make _wopen() and _open() able to handle long legacy DOS paths

sergio-nsk:
Issue: #8361

Perl in `C:\Program Files\Git\usr\bin` supports long paths w/o any manifest or registry changes. Why not let curl support long paths.

BTW the solution offered by "5.13 long paths are not fully supported on Windows" in the [known bugs](https://github.com/curl/curl/blob/master/docs/KNOWN_BUGS) does not work, because there is no way how to specify `//?/` in `file://`. [RFC 8089](https://datatracker.ietf.org/doc/html/rfc8089#appendix-C) excludes such paths `\\?\` and `\\.\`. `file://localhost///?/C:/foo` or `file://localhost/\\?\C:\foo` are also invalid URLs.

### DIFF
--- a/lib/curl_multibyte.c
+++ b/lib/curl_multibyte.c
@@ -91,13 +91,38 @@ char *curlx_convert_wchar_to_UTF8(const wchar_t *str_w)
 int curlx_win32_open(const char *filename, int oflag, ...)
 {
   int pmode = 0;
+  char *filename_long = NULL;
+  va_list param;
+  int result = -1;
 
 #ifdef _UNICODE
-  int result = -1;
   wchar_t *filename_w = curlx_convert_UTF8_to_wchar(filename);
+  if(ISALPHA(filename_w[0]) && filename_w[1] == ':'
+     && wcslen(filename_w) >= MAX_PATH) {
+    /* Legacy DOS path (eg C:\foo) is too long.
+       _wopen() does not support long legacy DOS paths. */
+    filename_long = malloc(strlen(filename) + 4 + 1);
+    if(filename_long) {
+      strcpy(filename_long, "\\\\?\\");
+      strcpy(filename_long + 4, filename);
+      curlx_unicodefree(filename_w);
+      filename_w = curlx_convert_UTF8_to_wchar(filename_long);
+    }
+  }
+#else
+  if(ISALPHA(filename[0]) && filename[1] == ':'
+     && strlen(filename) >= MAX_PATH) {
+    /* Legacy DOS path (eg C:\foo) is too long.
+       _open() does not support long legacy DOS paths. */
+    filename_long = malloc(strlen(filename) + 4 + 1);
+    if(filename_long) {
+      strcpy(filename_long, "\\\\?\\");
+      strcpy(filename_long + 4, filename);
+      filename = filename_long;
+    }
+  }
 #endif
 
-  va_list param;
   va_start(param, oflag);
   if(oflag & O_CREAT)
     pmode = va_arg(param, int);
@@ -110,10 +135,11 @@ int curlx_win32_open(const char *filename, int oflag, ...)
   }
   else
     errno = EINVAL;
-  return result;
 #else
-  return (_open)(filename, oflag, pmode);
+  result = (_open)(filename, oflag, pmode);
 #endif
+  free(filename_long);
+  return result;
 }
 
 FILE *curlx_win32_fopen(const char *filename, const char *mode)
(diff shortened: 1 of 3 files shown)


jay:
prepending `\\?\` to a path disables all parsing of the path. like any forward slashes and .. will not be converted. you are assuming that any path starting with the drive letter and a colon will not need any interpretation

nmoinvaz:
@jay that modified path is only passed into _open and _wopen. Are you seeing otherwise? As far as I can tell after it is used by those functions it is no longer used anywhere else.

jay:
> that modified path is only passed into _open and _wopen

I agree but I'm saying the logic is wrong. You can't assume that if a user specifies a path with a drive letter and colon that it is not going to need some interpretation. c:\foo/bar/../baz is not \\?\c:\foo/bar/../baz it is \\?\c:\foo\baz. you can't prepend \\?\ without first interpreting. Also IIRC \\?\ is only actually effective on the unicode W-suffixed api functions and crt w functions that wrap them (ie UNICODE/_UNICODE builds)

sergio-nsk:
> prepending `\\?\` to a path disables all parsing of the path. like any forward slashes and .. will not be converted. you are assuming that any path starting with the drive letter and a colon will not need any interpretation

Right, it disables the path parsing, but it does not matter, the limitation `PATH_MAX` breaks the file opening before the parse result that can be less than `PATH_MAX`. Hence, the result is the same unable to open a file.

The simple code with the existing 259 char path:
```c
#include <fcntl.h>
#include <io.h>
#include <stdio.h>
#include <string.h>

void win_open(const char *path, const char *msg)  {
   int fh1 = _open( path, _O_RDONLY ); // C4996
   // Note: _open is deprecated; consider using _sopen_s instead
   if(fh1 == -1) {
      printf("Open %s failed on file ", msg);
      perror("");
   } else {
      printf("Open %s succeeded on input file\n", msg);
      _close(fh1);
   }
}

int main(void) {
   const char *path259_l =   "\\\\?\\C:\\Work\\solid\\third_party\\curl\\tests\\log\\verylooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongdir\\dir\\..\\verylooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongfile3206.txt";
   const char *parsed259_l = "\\\\?\\C:\\Work\\solid\\third_party\\curl\\tests\\log\\verylooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongdir\\verylooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongfile3206.txt";

   win_open(path259_l + 4, "259 legacy");
   win_open(path259_l, "259 long");

   win_open(parsed259_l + 4, "259 parsed legacy");
   win_open(parsed259_l, "259 parsed long");
}
```
The output:
```
Open 259 legacy failed on file No such file or directory
Open 259 long failed on file Invalid argument
Open 259 parsed legacy succeeded on input file
Open 259 parsed long succeeded on input file
```

sergio-nsk:
As for forward slashes, curl translates them to backward slashes before `curlx_win32_open()` call.

jay:
> Right, it disables the path parsing, but it does not matter, the limitation `PATH_MAX` breaks the file opening before the parse result that can be less than `PATH_MAX`. Hence, the result is the same - unable to open a file.

It does matter. I think it's wrong to prepend `\\?\` without properly converting the path. Your logic assumes a long path (`wcslen(filename_w) >= MAX_PATH`) you can just prepend `\\?\` and that is not necessarily true.  Also IIRC there are still places in curl where we use PATH_MAX/MAX_PATH, as I mentioned in that other issue. You can't assume that all of libcurl code will successfully pass a long path to curlx_win32_open.

sergio-nsk:
> that is not necessarily true.

> You can't assume that all of libcurl code will successfully pass a long path to curlx_win32_open.

Not necessarily true, right. I don't assume anything. I just made it better in some cases, and not worse in other cases.

bagder (changes requested):
The new test fails in several windows CI jobs.

sergio-nsk:
MinGW builds don't seem to translate / to \ file paths. Will loot for the best solution, maybe just will just replace / with \ in the changed function.

jay:
I'm proposing #13522 to address this issue with a normalization check and cover all the file functions that we wrap.

sergio-nsk:
I will check https://github.com/curl/curl/pull/13522 in WinRT on Monday. Long paths are usual in WinRT.

sergio-nsk:
Closing in favor of the other PR.

PR #13516, opened 2024-05-02
Add CURL_USE_SYSROOT_LIBZ variable to automatically link zlib in sysroot

cntrump:
For the cross-platform compilation environment, zlib is already in `sysroot`. 
In this case, is it possible to directly specify the link parameter `-lz`?

example: `cmake -G Ninja -DCURL_USE_SYSROOT_LIBZ=ON ...`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,23 +555,27 @@ endif()
 # Keep ZLIB detection after TLS detection,
 # and before calling openssl_check_symbol_exists().
 
-set(HAVE_LIBZ OFF)
-set(USE_ZLIB OFF)
-optional_dependency(ZLIB)
-if(ZLIB_FOUND)
-  set(HAVE_LIBZ ON)
-  set(USE_ZLIB ON)
+if(CURL_USE_SYSROOT_LIBZ)
+  list(APPEND CURL_LIBS "-lz")
+else()
+  set(HAVE_LIBZ OFF)
+  set(USE_ZLIB OFF)
+  optional_dependency(ZLIB)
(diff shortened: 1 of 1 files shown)


vszakats:
Is it correct that you'd like to override the zlib found automatically by CMake to one that's necessary for your cross-target?

If so, have you tried this method?:
```
-DZLIB_INCLUDE_DIR=<sysroot>/include
-DZLIB_LIBRARY=<sysroot>/lib/libz.a
```

cntrump:
> Is it correct that you'd like to override the zlib found automatically by CMake to one that's necessary for your cross-target?
> 
> If so, have you tried this method?:
> 
> ```
> -DZLIB_INCLUDE_DIR=<custom-sysroot>/include
> -DZLIB_LIBRARY=<custom-sysroot>/lib/libz.a
> ```

I encountered this problem when compiling the apple universal version, because there is only `libz.tbd` and no `libz.a`. Use `-lz` to link to `libz.tbd` correctly.

vszakats:
Would this work in your case to spec the .tbd?:
```
-DZLIB_LIBRARY=/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk/usr/lib/libz.tbd
```
(replace with your custom full path)

cntrump:
> Would this work in your case to spec the .tbd?:
> 
> ```
> -DZLIB_LIBRARY=/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk/usr/lib/libz.tbd
> ```
> 
> (replace with your custom full path)

Works, thanks !

PR #13518, opened 2024-05-02
Make sure `thread_wait_resolv` gets called by Curl_resolver_kill

### LINKED ISSUE #13509: Windows DNS resolution: Curl crash when GetAddrInfoExW callback invoked on shutdown
### I did this

We are using libcurl in our C++ project on windows 10 and noticed strange issues after upgrading our curl from 8.3.0 to 8.6.0 or later. After spending some time reproducing the issue, we managed to get an ASAN stacktrace of the exact crash:
```
=================================================================
==41196==ERROR: AddressSanitizer: heap-use-after-free on address 0x12184f8b06cc at pc 0x7ffbe4b442c4 bp 0x00c7a3dfdcf0 sp 0x00c7a3dfd480
READ of size 36 at 0x12184f8b06cc thread T-1
==41196==WARNING: Failed to use and restart external symbolizer!
++++++++++++++++++++++++++++++++++++++++deinit
    #0 0x7ffbe4b442c3 in _asan_wrap_memmove+0x193 (C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\clang_rt.asan_dynamic-x86_64.dll+0x1800342c3)
    #1 0x7ffd4bc46936 in NdrFullPointerFree+0xd36 (C:\WINDOWS\System32\RPCRT4.dll+0x180006936)
    #2 0x7ffd4bd1d87c in NdrMesSimpleTypeEncodeAll+0x6e2c (C:\WINDOWS\System32\RPCRT4.dll+0x1800dd87c)
    #3 0x7ffd4bd1aa74 in NdrMesSimpleTypeEncodeAll+0x4024 (C:\WINDOWS\System32\RPCRT4.dll+0x1800daa74)
    #4 0x7ffd4bd1dc6f in NdrClientCall3+0xef (C:\WINDOWS\System32\RPCRT4.dll+0x1800ddc6f)
    #5 0x7ffd4a348fb0 in DnsGetAdaptersInfo+0xe40 (C:\WINDOWS\SYSTEM32\DNSAPI.dll+0x180008fb0)
    #6 0x7ffd4a3486bf in DnsGetAdaptersInfo+0x54f (C:\WINDOWS\SYSTEM32\DNSAPI.dll+0x1800086bf)
    #7 0x7ffd4a34c616 in DnsQueryEx+0x9b6 (C:\WINDOWS\SYSTEM32\DNSAPI.dll+0x18000c616)
    #8 0x7ffd4a34bdc5 in DnsQueryEx+0x165 (C:\WINDOWS\SYSTEM32\DNSAPI.dll+0x18000bdc5)
    #9 0x7ffd4a65b162 in WSPStartup+0x1902 (C:\WINDOWS\system32\mswsock.dll+0x18000b162)
    #10 0x7ffd4a65af33 in WSPStartup+0x16d3 (C:\WINDOWS\system32\mswsock.dll+0x18000af33)
    #11 0x7ffd4a65a60d in WSPStartup+0xdad (C:\WINDOWS\system32\mswsock.dll+0x18000a60d)
    #12 0x7ffd4d7c9e79 in WSALookupServiceBeginW+0x1139 (C:\WINDOWS\System32\WS2_32.dll+0x180009e79)
    #13 0x7ffd4d8e0cd8 in RtlDeactivateActivationContext+0x2c8 (C:\WINDOWS\SYSTEM32\ntdll.dll+0x180070cd8)
    #14 0x7ffd4d8c31b9 in TpReleaseCleanupGroupMembers+0xad9 (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800531b9)
    #15 0x7ffd4d717343 in BaseThreadInitThunk+0x13 (C:\WINDOWS\System32\KERNEL32.DLL+0x180017343)
    #16 0x7ffd4d8c26b0 in RtlUserThreadStart+0x20 (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526b0)

0x12184f8b06cc is located 140 bytes inside of 180-byte region [0x12184f8b0640,0x12184f8b06f4)
freed by thread T0 here:
    #0 0x7ffbe4b4f585 in _asan_wrap_RtlFreeHeap+0x4d5 (C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\clang_rt.asan_dynamic-x86_64.dll+0x18003f585)
    #1 0x7ffd4a359747 in DnsApiFree+0x37 (C:\WINDOWS\SYSTEM32\DNSAPI.dll+0x180019747)
    #2 0x7ffd4a66cfff in Tcpip6_WSHGetWildcardSockaddr+0xc48f (C:\WINDOWS\system32\mswsock.dll+0x18001cfff)
    #3 0x7ffd4a65f6f6 in Tcpip6_WSHAddressToString+0xe6 (C:\WINDOWS\system32\mswsock.dll+0x18000f6f6)
    #4 0x7ffd4d7c1b41 in Ordinal487+0x1b41 (C:\WINDOWS\System32\WS2_32.dll+0x180001b41)
    #5 0x7ffd4d7d3350 in WSACreateEvent+0xf0 (C:\WINDOWS\System32\WS2_32.dll+0x180013350)
    #6 0x7ffd4d7d022b in WSCEnumProtocols+0x97b (C:\WINDOWS\System32\WS2_32.dll+0x18001022b)
    #7 0x7ffd4d7cfb33 in WSCEnumProtocols+0x283 (C:\WINDOWS\System32\WS2_32.dll+0x18000fb33)
    #8 0x7ffd4d7d03d9 in WahEnumerateHandleContexts+0xf9 (C:\WINDOWS\System32\WS2_32.dll+0x1800103d9)
    #9 0x7ffd4d7d04f5 in WSACleanup+0xe5 (C:\WINDOWS\System32\WS2_32.dll+0x1800104f5)
    #10 0x7ffbe6a5a472 in Curl_win32_cleanup C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\system_win32.c:164
    #11 0x7ffbe6a3eaba in curl_global_cleanup C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\easy.c:291
    ...
    #19 0x7ff65bae75ab in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288    #20 0x7ffd4d717343 in BaseThreadInitThunk+0x13 (C:\WINDOWS\System32\KERNEL32.DLL+0x180017343)
    #21 0x7ffd4d8c26b0 in RtlUserThreadStart+0x20 (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526b0)

previously allocated by thread T3 here:
    #0 0x7ffbe4b4ed29 in _asan_wrap_RtlAllocateHeap+0x179 (C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\clang_rt.asan_dynamic-x86_64.dll+0x18003ed29)
    #1 0x7ffd4a65aafe in WSPStartup+0x129e (C:\WINDOWS\system32\mswsock.dll+0x18000aafe)
    #2 0x7ffd4a65a2a7 in WSPStartup+0xa47 (C:\WINDOWS\system32\mswsock.dll+0x18000a2a7)
    #3 0x7ffd4d7c9089 in WSALookupServiceBeginW+0x349 (C:\WINDOWS\System32\WS2_32.dll+0x180009089)
    #4 0x7ffd4d7c8e48 in WSALookupServiceBeginW+0x108 (C:\WINDOWS\System32\WS2_32.dll+0x180008e48)
    #5 0x7ffd4d7c83bf in GetNameInfoW+0x133f (C:\WINDOWS\System32\WS2_32.dll+0x1800083bf)
    #6 0x7ffd4d7c81c8 in GetNameInfoW+0x1148 (C:\WINDOWS\System32\WS2_32.dll+0x1800081c8)
    #7 0x7ffd4d7c6a76 in GetAddrInfoW+0xf56 (C:\WINDOWS\System32\WS2_32.dll+0x180006a76)
    #8 0x7ffd4d7c4390 in GetAddrInfoExW+0x4e0 (C:\WINDOWS\System32\WS2_32.dll+0x180004390)
    #9 0x7ffbe6a4a755 in init_resolve_thread C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\asyn-thread.c:657
    #10 0x7ffbe6a49cd5 in Curl_resolver_getaddrinfo C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\asyn-thread.c:940
    #11 0x7ffbe6a61049 in Curl_resolv C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\hostip.c:814
    #12 0x7ffbe6a551c6 in resolve_fresh C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\url.c:3297
    #13 0x7ffbe6a535b1 in create_conn C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\url.c:3802
    #14 0x7ffbe6a51979 in Curl_connect C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\url.c:3873
    #15 0x7ffbe6a407b8 in multi_runsingle C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\multi.c:2002
    #16 0x7ffbe6a3fc00 in curl_multi_perform C:\vcpkg\buildtrees\curl\src\curl-8_6_0-f17d00bdb6.clean\lib\multi.c:2780
    ...
    #20 0x7ffbe6afd455 in thread_start<unsigned int (__cdecl*)(void *),1> minkernel\crts\ucrt\src\appcrt\startup\thread.cpp:97
    #21 0x7ffbe4b5ebde in _asan_default_suppressions__dll+0x122e (C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\clang_rt.asan_dynamic-x86_64.dll+0x18004ebde)
    #22 0x7ffd4d717343 in BaseThreadInitThunk+0x13 (C:\WINDOWS\System32\KERNEL32.DLL+0x180017343)
    #23 0x7ffd4d8c26b0 in RtlUserThreadStart+0x20 (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526b0)

Thread T3 created by T0 here:
    #0 0x7ffbe4b60897 in _asan_wrap_CreateThread+0x77 (C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\clang_rt.asan_dynamic-x86_64.dll+0x180050897)
    #1 0x7ffbe6afd59e in _beginthreadex minkernel\crts\ucrt\src\appcrt\startup\thread.cpp:209
    ...
    #10 0x7ff65bae75ab in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288    #11 0x7ffd4d717343 in BaseThreadInitThunk+0x13 (C:\WINDOWS\System32\KERNEL32.DLL+0x180017343)
    #12 0x7ffd4d8c26b0 in RtlUserThreadStart+0x20 (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526b0)

SUMMARY: AddressSanitizer: heap-use-after-free (C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\clang_rt.asan_dynamic-x86_64.dll+0x1800342c3) in _asan_wrap_memmove+0x193
Shadow bytes around the buggy address:
  0x043b59796080: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x043b59796090: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fa
  0x043b597960a0: fa fa fa fa fa fa fa fa 00 00 00 00 00 00 00 00
  0x043b597960b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x043b597960c0: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
=>0x043b597960d0: fd fd fd fd fd fd fd fd fd[fd]fd fd fd fd fd fa
  0x043b597960e0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x043b597960f0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x043b59796100: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x043b59796110: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x043b59796120: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
```

What happens here is that a DNS resolver thread created by `GetAddrInfoExW` did not terminate yet, however curl is already shutting down, deinitializing Winsock with `WSACleanup()` leading to an access violation.

After investigating recent changes we came to the conclusion that there is most likely a bug in the current async windows DNS resolution code around `GetAddrInfoExW` implemented in https://github.com/curl/curl/pull/12482.
We tried a few things to verify this:
 - Use Curl 8.5.0 or older => works
 - Use Curl 8.6.0 or newer => crashes
 - Use Curl 8.6.0 or newer with c-ares => works

Our test looks as follows:
- Initialize curl, create 1 thread to process all curl multi requests
- Create a second thread to push work to the first thread
- Shutdown all the new threads, join them, then deinit curl

We made sure that everything is thread safe by using proper locking mechanisms, the issue is reproducable roughly 60% of the time per run. The bug is most likely the cause of `Curl_resolver_kill` -> `Curl_resolver_cancel()` -> `destroy_async_data()` which then has this code: https://github.com/curl/curl/blob/0199104e944c1625056b4cf78b6cdb14a7b82e3e/lib/asyn-thread.c#L555-L569

Closing the completion event handle in the first branch (not done) seems like a bad idea as the next call to `Curl_GetAddrInfoExCancel()` might not finish instantly. We then would wait for `WaitForSingleObject(td->complete_ev, INFINITE);` but this will instantly return with an error as the handle is closed. This is however just a theory and not verified yet.

### I expected the following

Curl shutdown / multi close properly stops all running DNS requests and doesn't crash.
Tests for the Windows implementation would also be great.

### curl/libcurl version

`>= curl 8.6.0`

### operating system

Windows 10 22H2

pps83:
+ make sure complete_ev is set to NULL in the same places where thread_hnd is set to NULL
 + set thread_hnd to NULL, check thread_hnd before calling Curl_thread_destroy

fixes #13509

### DIFF
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -566,6 +566,7 @@ static void destroy_async_data(struct Curl_async *async)
         Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
         WaitForSingleObject(td->complete_ev, INFINITE);
         CloseHandle(td->complete_ev);
+        td->complete_ev = NULL;
       }
 #endif
       if(td->thread_hnd != curl_thread_t_null)
@@ -713,6 +714,7 @@ static CURLcode thread_wait_resolv(struct Curl_easy *data,
   if(td->complete_ev) {
     WaitForSingleObject(td->complete_ev, INFINITE);
     CloseHandle(td->complete_ev);
+    td->complete_ev = NULL;
     if(entry)
       result = getaddrinfo_complete(data);
   }
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -754,6 +754,13 @@ void Curl_resolver_kill(struct Curl_easy *data)
   /* If we're still resolving, we must wait for the threads to fully clean up,
      unfortunately.  Otherwise, we can simply cancel to clean up any resolver
      data. */
+#ifdef _WIN32
+  if(td && td->complete_ev) {
+    Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
+    (void)thread_wait_resolv(data, NULL, FALSE);
+  }
+  else
+#endif
   if(td && td->thread_hnd != curl_thread_t_null
      && (data->set.quick_exit != 1L))
     (void)thread_wait_resolv(data, NULL, FALSE);
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -554,11 +554,15 @@ static void destroy_async_data(struct Curl_async *async)
 
     if(!done) {
 #ifdef _WIN32
-      if(td->complete_ev)
+      if(td->complete_ev) {
         CloseHandle(td->complete_ev);
-      else
+        td->complete_ev = NULL;
+      }
 #endif
-      Curl_thread_destroy(td->thread_hnd);
+      if(td->thread_hnd != curl_thread_t_null) {
+        Curl_thread_destroy(td->thread_hnd);
+        td->thread_hnd = curl_thread_t_null;
+      }
     }
     else {
 #ifdef _WIN32


bagder:
Is this a duplicate of #13517 ?

pps83:
> Is this a duplicate of #13517 ?
> 
> I mean, it is clearly a different patch, I'm just a little curious why they are two separate ones as the relation between them is not totally obvious to me.

I replied to the issue request explaining why I made these two: https://github.com/curl/curl/issues/13509#issuecomment-2089522519 (also, check my previous message there)

Does Curl_resolver_kill need to simply wait, or it should cancel request?

bagder:
If it can cancel, it should cancel.

pps83:
perhaps, it should even do this:

#ifdef _WIN32
  if(td && td->complete_ev) {
    Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
    (void)thread_wait_resolv(data, NULL, FALSE);
    return;
  }
#endif

to avoid that quirky `else` inside `ifdef` block.

bagder:
Or even this?
~~~diff
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -752,15 +752,22 @@ void Curl_resolver_kill(struct Curl_easy *data)
   struct thread_data *td = data->state.async.tdata;
 
   /* If we're still resolving, we must wait for the threads to fully clean up,
      unfortunately.  Otherwise, we can simply cancel to clean up any resolver
      data. */
+#ifdef _WIN32
+  if(td && td->complete_ev) {
+    Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
+    (void)thread_wait_resolv(data, NULL, FALSE);
+  }
+#else
   if(td && td->thread_hnd != curl_thread_t_null
      && (data->set.quick_exit != 1L))
     (void)thread_wait_resolv(data, NULL, FALSE);
   else
     Curl_resolver_cancel(data);
+#endif
 }

pps83:
> Or even this?

No, that's not OK. This breaks all windows builds that do not have that async dns api (complete_ev test checks if dns api is used, otherwise regular threads).

lib/asyn-thread.c
```diff
@@ -754,6 +754,13 @@ void Curl_resolver_kill(struct Curl_easy *data)
   /* If we're still resolving, we must wait for the threads to fully clean up,
      unfortunately.  Otherwise, we can simply cancel to clean up any resolver
      data. */
+#ifdef _WIN32
+  if(td && td->complete_ev) {
+    Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
+    (void)thread_wait_resolv(data, NULL, FALSE);
+  }
+  else
+#endif
```
pps83, line 763:
```suggestion
#ifdef _WIN32
  if(td && td->complete_ev) {
    Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
    (void)thread_wait_resolv(data, NULL, FALSE);
    return;
  }
#endif

```

Ch40zz:
Hey, thanks a lot for looking into this! :)  
We tried this out today, but it sadly did not fix the issue. We're currently not sure yet why or how but we'll try to do a bit more debugging soon and maybe get a proper repro for you as well.

Heads up, the current implementation introduced another bug now where `CloseHandle()` gets twice on the same handle, leading to a 0xC0000008 "Invalid Handle" exception in the presence of debuggers. You should probably null out the handle after closing it.

pps83:
@Ch40zz can you please try this PR again, I added a [minor change](https://github.com/curl/curl/pull/13518/commits/699f92254ae7bcd49c9cfbe3fbe77b11677c980c) to see if you still get "Invalid Handle" exception

Ch40zz:
> @Ch40zz can you please try this PR again, I added a [minor change](https://github.com/curl/curl/pull/13518/commits/699f92254ae7bcd49c9cfbe3fbe77b11677c980c) to see if you still get "Invalid Handle" exception

Thanks, we'll test it today, will let you know later. Hope we can also get a repro for you or somehow debug why the issue still persists on shutdown

pps83:
> If you are using the [quick exit](https://curl.se/libcurl/c/CURLOPT_QUICK_EXIT.html) option see if disabling it makes a difference.

supposedly, this PR makes `quick_exit` irrelevant when async dns api is present on Windows.

Ch40zz:
We didn't have much time today so we couldn't get you a repro yet, but we at least ran the new commit and it indeed fixed the double `CloseHandle()` issue. However, the crash itself is not yet fixed. We'll have more time next week to actually get a repro and debug deeper to see why that is!

> > Thanks, we'll test it today, will let you know later. Hope we can also get a repro for you or somehow debug why the issue still persists on shutdown
> 
> If you are using the [quick exit](https://curl.se/libcurl/c/CURLOPT_QUICK_EXIT.html) option see if disabling it makes a difference.

We are not using quick exit but in theory this code should behave the same way independently of that option.

Ch40zz:
Quick update from us after testing for a day:
This commit is correct and fixes a bug in curl. Our initial bug is still occuring when building with ASAN, however it is not a curl issue. It also occurs when pasting the sample code for `GetAddrInfoExW()`. We suspect that either ASAN has a bug (false positive) or the API itself causes a heap-use-after-free. We will report the bug to Microsoft and hope it gets fixed soon.
So long people building with ASAN will run into the error on shutdown. We will link the bug in the main issue when it has been created.
Thanks for your help!

pps83:
> Quick update from us after testing for a day: This commit is correct and fixes a bug in curl. Our initial bug is still occuring when building with ASAN, however it is not a curl issue. It also occurs when pasting the sample code for `GetAddrInfoExW()`. We suspect that either ASAN has a bug (false positive) or the API itself causes a heap-use-after-free. We will report the bug to Microsoft and hope it gets fixed soon. So long people building with ASAN will run into the error on shutdown. We will link the bug in the main issue when it has been created. Thanks for your help!

if you can provide some minimal code that exposes the issue, I'll look into it (with asan or not). There is also a possibility that your exit sequence is not properly handled. When I added that code originally, I tested it when cable was unplugged (eg when dns requests weren't answered) and didn't have any issues at exist.

lib/asyn-thread.c
```diff
@@ -566,6 +566,7 @@ static void destroy_async_data(struct Curl_async *async)
         Curl_GetAddrInfoExCancel(&td->tsd.w8.cancel_ev);
         WaitForSingleObject(td->complete_ev, INFINITE);
         CloseHandle(td->complete_ev);
+        td->complete_ev = NULL;
```
jay, line 569: there's another closehandle a few lines above this that you could also set td->complete_ev = NULL;

pps83: I know, but I tried to mimic existing code logic, and not to null it after each CloseHandle.
In other words, where thread_hnd is set to NULL, complete_ev should also be set to NULL. Curl_thread_destroy sets does CloseHandle, without setting it to NULL.

pps83: Setting it to NULL might seem like a good idea, but then code flow (where thread_hnd/complete_ev are checked) will be entirely different. Perhaps, I should add a comment explaining why it's not set to NULL there. What do you guys think?

jay: IMO if the variables can be evaluated after destroy_async_data they should not hold the invalid values. What logic would this affect:
~~~
diff --git a/lib/asyn-thread.c b/lib/asyn-thread.c
index 4b54b0a..aa0aa95 100644
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -554,11 +554,16 @@ static void destroy_async_data(struct Curl_async *async)
 
     if(!done) {
 #ifdef _WIN32
-      if(td->complete_ev)
+      if(td->complete_ev) {
         CloseHandle(td->complete_ev);
+        td->complete_ev = NULL;
+      }
       else
 #endif
-      Curl_thread_destroy(td->thread_hnd);
+      {
+        Curl_thread_destroy(td->thread_hnd);
+        td->thread_hnd = curl_thread_t_null;
+      }
     }
     else {
 #ifdef _WIN32
~~~

pps83: In this case the change would be OK, as long as others agree. This would be somewhat unrelated change (updating thread_hnd). Do you think this should be done as part of this PR? If yes, then I'll update

pps83: > What logic would this affect

Well, it's not that clear. That entire function is quite mysterious. Most of the logic seemingly happens inside the block that should rarely execute. At the top of the block it goes: 

```
struct thread_data *td = async->tdata;
...
if(!done) {
  Curl_thread_destroy(td->thread_hnd);
}
else {
  Curl_thread_join(&td->thread_hnd);
  destroy_thread_sync_data(&td->tsd);
  free(async->tdata);
}
async->tdata = NULL;
...
```

it looks as if it should have been `if (done)` instead of `if (!done)`

pps83: the only time done=1 is when starting a thread of creating the complete_ev fails. Makes no sense to try to execute that else block that curl has now.

pps83: @Ch40zz can you please flip that done condition locally, do you still get the error with asan?

Ch40zz:
> if you can provide some minimal code that exposes the issue, I'll look into it (with asan or not). There is also a possibility that your exit sequence is not properly handled. When I added that code originally, I tested it when cable was unplugged (eg when dns requests weren't answered) and didn't have any issues at exist.

as mentioned before we have a repro in curl and a repro with pure WinAPI, both trigger the exact same ASAN issue.
compile the code with MSVC and enable ASAN to get the exact same stack trace as I posted in the issue.
I'm very positive that the issue lies somewhere else and not in the current curl code if this PR gets merged, but please verify that yourself by looking at the PoC code: https://gist.github.com/Ch40zz/f3a33139f35fd608d71db5e4085e0bee

pps83:
> I'm very positive that the issue lies somewhere else and not in the current curl code if this PR gets merged, but please verify that yourself by looking at the PoC code: https://gist.github.com/Ch40zz/f3a33139f35fd608d71db5e4085e0bee

Oh, ok, I see that the test code tries winapi directly without using curl and triggers the same asan error. Yes, in this case, most likely asan or winapi internal error (perhaps the error gets handled internally when asan isn't used)

lib/asyn-thread.c
pps83: > If you are uncomfortable with it I can submit it separately as a PR so it goes in before this and then you can rebase
I'm OK with this, I was just asking if you are ok with it

lib/asyn-thread.c
```diff
@@ -554,18 +554,23 @@ static void destroy_async_data(struct Curl_async *async)
...
         CloseHandle(td->complete_ev);
-      else
+        td->complete_ev = NULL;
+      }
 #endif
-      Curl_thread_destroy(td->thread_hnd);
+      if(td->thread_hnd != curl_thread_t_null) {
+        Curl_thread_destroy(td->thread_hnd);
+        td->thread_hnd = curl_thread_t_null;
+      }
```
pps83, line 565: That's the correct way: has to check thread_hnd != curl_thread_t_null before calling Curl_thread_destroy.
Curl_thread_destroy does CloseHandle on the thread_hnd, and calling on NULL handle isn't ok. with pthreads it does the check internally.

pps83:
@jay PR updated, should be ready to merge

lib/asyn-thread.c
jay: I think that ultimately it would be better to address this in Curl_thread_destroy and get rid of the inconsistency, for example
~~~diff
diff --git a/lib/curl_threads.c b/lib/curl_threads.c
index 222d936..93fa2da 100644
--- a/lib/curl_threads.c
+++ b/lib/curl_threads.c
@@ -131,7 +131,8 @@ curl_thread_t Curl_thread_create(unsigned int (CURL_STDCALL

 void Curl_thread_destroy(curl_thread_t hnd)
 {
-  CloseHandle(hnd);
+  if(hnd != curl_thread_t_null)
+    CloseHandle(hnd);
 }

 int Curl_thread_join(curl_thread_t *hnd)
~~~

Ch40zz:
Just a heads up that the bug ticket on MS Developer Community has been created here: https://developercommunity.visualstudio.com/t/ASAN:-heap-use-after-free-in-NdrFullPoin/10654169

jay:
> @jay PR updated, should be ready to merge. Perhaps, "fixes https://github.com/curl/curl/issues/13509" should be removed from the commit message

Thanks. I left it with the fixed tag because it did fix the Curl_resolver_kill issue..

PR #13521, opened 2024-05-02
sectransp: Use common code for cipher suite lookup

jan2000:
Take advantage of the Curl_cipher_suite_walk_str() and Curl_cipher_suite_get_str() functions introduced in commit fba9afe.


lib/vtls/sectransp.c
```diff
@@ -1453,18 +1265,57 @@ static bool is_cipher_suite_strong(SSLCipherSuite suite_num)
   return true;
 }
 
-static bool sectransp_is_separator(char c)
+static int sectransp_cipher_suite_get_str(uint16_t id, char *buf,
```
jan2000, line 1268: These wrapper functions for `Curl_cipher_suite_get_str` and `Curl_cipher_suite_walk_str` are here so there is no change in behavior with this commit. But, I guess no one will shed a tear if support for setting/getting these ciphers is removed?

lib/vtls/sectransp.c
```diff
@@ -1453,18 +1265,57 @@ static bool is_cipher_suite_strong(SSLCipherSuite suite_num)
...
-  switch(c) {
-  case ' ':
-  case '\t':
-  case ':':
-  case ',':
-  case ';':
-    return true;
-  }
-  return false;
+  /* are these fortezza suites ever supported ? */
```
nickzman, line 1271: That should be “even” supported. And I have no idea if they’re supported or not.

lib/vtls/sectransp.c
```diff
@@ -1453,18 +1265,57 @@ static bool is_cipher_suite_strong(SSLCipherSuite suite_num)
...
+}
+
+static uint16_t sectransp_cipher_suite_walk_str(const char **str,
+                                                const char **end)
+{
+  uint16_t id = Curl_cipher_suite_walk_str(str, end);
+  size_t len = *end - *str;
+
+  if(!id) {
+    /* are these fortezza suites ever supported ? */
```
nickzman, line 1300: Same change here.

nickzman (changes requested):
Thanks for your work on this. I only noticed some minor things; the code looks okay, and it doesn’t appear to introduce any obvious regressions.

lib/vtls/sectransp.c
jan2000: Well actually I meant to say "were these ever supported ?", but also in present and future tense. As in: can this be safely removed, or should this be kept in for backward compatibility?

jan2000:
@nickzman I made the changes you suggested. See https://github.com/curl/curl/compare/1ae71316953f0aa57e5c6a8ad0cb00b9a7deb317..7746df8cf110219407aeeb07e04167d50b336fe9

However, after pushing I saw CI tests failing due to unit2604. So I rebased to latest master and pushed again. Sorry if this caused any inconvenience for the review.

But now I am a bit puzzled as to how the test could be failing on unit2604 in the first place, as it was not yet in the master that this PR was initially based on. How does that work?

nickzman (approved):
LGTM. Thanks for your work on this.

bagder:
Thanks!

PR #13522, opened 2024-05-03
curl_multibyte: support Windows paths longer than MAX_PATH

jay:
- Add a helper function for the Windows file wrapper functions that will normalize a long path (or a filename in a long path) and add the prefix `\\?\` so that Windows will access the file.

Prior to this change if a filename (when normalized internally by Windows to its full path) or a path was longer than MAX_PATH (260) then Windows would not open the path, unless it was already normalized by the user and had the `\\?\` prefix prepended.

The `\\?\` prefix could not be passed to file:// so for example something like file://c:/foo/bar/filename255chars could not be opened prior to this change.

There's some code in tool_doswin that will need to be modified as well to further remove MAX_PATH (aka PATH_MAX) limitation.

Ref: https://github.com/curl/curl/issues/8361
Ref: https://github.com/curl/curl/pull/13512
Ref: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
Ref: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

Closes #xxxx

### DIFF
--- a/lib/curl_multibyte.c
+++ b/lib/curl_multibyte.c
@@ -88,12 +88,170 @@ char *curlx_convert_wchar_to_UTF8(const wchar_t *str_w)
 
 #if defined(USE_WIN32_LARGE_FILES) || defined(USE_WIN32_SMALL_FILES)
 
+/* declare GetFullPathNameW for mingw-w64 UWP builds targeting old windows */
+#if defined(CURL_WINDOWS_UWP) && defined(__MINGW32__) && \
+  (_WIN32_WINNT < _WIN32_WINNT_WIN10)
+WINBASEAPI DWORD WINAPI GetFullPathNameW(LPCWSTR, DWORD, LPWSTR, LPWSTR *);
+#endif
+
+/* Fix excessive paths (paths that exceed MAX_PATH length of 260).
+ *
+ * This is a helper function to fix paths that would exceed the MAX_PATH
+ * limitation check done by Windows APIs. It does so by normalizing the passed
+ * in filename or path 'in' to its full canonical path, and if that path is
+ * longer than MAX_PATH then setting 'out' to "\\?\" prefix + that full path.
+ *
+ * For example 'in' filename255chars in current directory C:\foo\bar is
+ * fixed as \\?\C:\foo\bar\filename255chars for 'out' which will tell Windows
+ * it is ok to access that filename even though the actual full path is longer
+ * than 260 chars.
+ *
+ * For non-Unicode builds this function may fail sometimes because only the
+ * Unicode versions of some Windows API functions can access paths longer than
+ * MAX_PATH, for example GetFullPathNameW which is used in this function. When
+ * the full path is then converted from Unicode to multibyte that fails if any
+ * directories in the path contain characters not in the current codepage.
+ */
+static bool fix_excessive_path(const TCHAR *in, TCHAR **out)
+{
+  size_t needed, count;
+  const wchar_t *in_w;
+  wchar_t *fbuf = NULL;
+
+  /* MS documented "approximate" limit for the maximum path length */
+  const size_t max_path_len = 32767;
+
+#ifndef _UNICODE
+  wchar_t *ibuf = NULL;
+  char *obuf = NULL;
+#endif
+
+  *out = NULL;
+
+  /* skip paths already normalized */
+  if(!_tcsncmp(in, _T("\\\\?\\"), 4))
+    goto cleanup;
+
+#ifndef _UNICODE
+  /* convert multibyte input to unicode */
+  needed = mbstowcs(NULL, in, 0);
+  if(needed == (size_t)-1 || needed >= max_path_len)
+    goto cleanup;
+  ++needed; /* for NUL */
+  ibuf = malloc(needed * sizeof(wchar_t));
+  if(!ibuf)
+    goto cleanup;
+  count = mbstowcs(ibuf, in, needed);
+  if(count == (size_t)-1 || count >= needed)
+    goto cleanup;
+  in_w = ibuf;
+#else
+  in_w = in;
+#endif
+
+  /* GetFullPathNameW returns the normalized full path in unicode. It converts
+     forward slashes to backslashes, processes .. to remove directory segments,
+     etc. Unlike GetFullPathNameA it can process paths that exceed MAX_PATH. */
+  needed = (size_t)GetFullPathNameW(in_w, 0, NULL, NULL);
+  if(!needed || needed > max_path_len)
+    goto cleanup;
+  /* skip paths that are not excessive and do not need modification */
+  if(needed <= MAX_PATH)
+    goto cleanup;
+  fbuf = malloc(needed * sizeof(wchar_t));
+  if(!fbuf)
+    goto cleanup;
+  count = (size_t)GetFullPathNameW(in_w, (DWORD)needed, fbuf, NULL);
+  if(!count || count >= needed)
+    goto cleanup;
+
+  /* prepend \\?\ or \\?\UNC\ to the excessively long path.
+   *
+   * c:\longpath            --->    \\?\c:\longpath
+   * \\.\c:\longpath        --->    \\?\c:\longpath
+   * \\?\c:\longpath        --->    \\?\c:\longpath  (unchanged)
+   * \\server\c$\longpath   --->    \\?\UNC\server\c$\longpath
+   *
+   * https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
+   */
+  if(!wcsncmp(fbuf, L"\\\\?\\", 4))
+    ; /* do nothing */
+  else if(!wcsncmp(fbuf, L"\\\\.\\", 4))
+    fbuf[2] = '?';
+  else if(!wcsncmp(fbuf, L"\\\\.", 3) || !wcsncmp(fbuf, L"\\\\?", 3)) {
+    /* Unexpected, not UNC. The formatting doc doesn't allow this AFAICT. */
+    goto cleanup;
+  }
+  else {
+    wchar_t *temp;
+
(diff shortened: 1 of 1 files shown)


lib/curl_multibyte.c
```diff
@@ -88,12 +88,106 @@ char *curlx_convert_wchar_to_UTF8(const wchar_t *str_w)
...
+  if(written == (size_t)-1 || written >= needed)
+    goto error;
+  in_w = ibuf;
+#else
+  in_w = in;
+#endif
+
+  /* get full unicode path of the unicode filename or path */
+  needed = (size_t)GetFullPathNameW(in_w, 0, NULL, NULL);
+  if(!needed || needed > (32767 - 4))
```
bagder, line 142: I think these magic numbers can use a comment. Where do they come from and why?

lib/curl_multibyte.c
```diff
@@ -88,12 +88,106 @@ char *curlx_convert_wchar_to_UTF8(const wchar_t *str_w)
...
+    goto error;
+  written = mbstowcs(ibuf, in, needed);
+  if(written == (size_t)-1 || written >= needed)
+    goto error;
+  in_w = ibuf;
+#else
+  in_w = in;
+#endif
+
+  /* get full unicode path of the unicode filename or path */
```
nmoinvaz, line 140: Maybe improve comment here to suggest that we are using `GetFullPathNameW` to evaluate `\..\` back-one-level path slashes and `\.\`.

lib/curl_multibyte.c
```diff
@@ -88,12 +88,106 @@ char *curlx_convert_wchar_to_UTF8(const wchar_t *str_w)
...
+  if(written == (size_t)-1 || written >= needed)
+    goto error;
+  *out = obuf;
+  obuf = NULL;
+#else
+  *out = fbuf;
+  fbuf = NULL;
+#endif
+
+error:
```
nmoinvaz, line 174: Perhaps use `cleanup:` instead of `error:`, since cases like this are not actually errors yet it says `goto error`:
```c
/* ignore paths which are not excessive and don't need modification */
```

lib/curl_multibyte.c
nmoinvaz: I think part of it is from here:
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

lib/curl_multibyte.c
```diff
@@ -88,12 +88,106 @@ char *curlx_convert_wchar_to_UTF8(const wchar_t *str_w)
...
+  written = mbstowcs(ibuf, in, needed);
+  if(written == (size_t)-1 || written >= needed)
+    goto error;
+  in_w = ibuf;
+#else
+  in_w = in;
+#endif
+
+  /* get full unicode path of the unicode filename or path */
+  needed = (size_t)GetFullPathNameW(in_w, 0, NULL, NULL);
```
nmoinvaz, line 141: Perhaps we should mention in the comment that `GetFullPathNameW` is also being used to translate path slashes such as `\.\` and `\..\` and to convert forward slashes to backslashes. 

Although the MS docs don't explicitly state it, I ran a simple test using:
```c++
#include <cstring>
#include <windows.h>
#include <iostream>

int main() {
  static const char a[] = "c:\\test\\..\\hello";
  char o[1000]{};
  GetFullPathNameA(a, sizeof(o), o, nullptr);
  std::cout << a << " -> " << o << std::endl;

  static const char b[] = "c:\\test\\../hello/./";
  char o2[1000]{};
  GetFullPathNameA(b, sizeof(o2), o2, nullptr);
  std::cout << b << " -> " << o2 << std::endl;
  return 0;
}
```
Output:
```bash
PS D:\test1> .\build\Debug\main.exe
c:\test\..\hello -> c:\hello
c:\test\../hello/./ -> c:\hello\
```

lib/curl_multibyte.c
jay: > I think these magic numbers can use a comment. Where do they come from and why?

Ok. I made it a constant and added a comment:

~~~c
  /* MS documented "approximate" limit for the maximum path length */
  const size_t max_path_len = 32767;
~~~

The 4 was extra space but it's in a separate section with the changed code so it should be easier to understand.

lib/curl_multibyte.c
jay: > Perhaps we should mention in the comment that `GetFullPathNameW` is also being used to translate path slashes such as `\.\` and `\..\` and to convert forward slashes to backslashes.

Sure yes. That is normalization and I didn't explain exactly what it was but I changed the comment in the latest commit to address this:

~~~c
  /* GetFullPathNameW returns the normalized full path in unicode. It converts
     forward slashes to backslashes, processes .. to remove directory segments,
     etc. Unlike GetFullPathNameA it can process paths that exceed MAX_PATH. */
  needed = (size_t)GetFullPathNameW(in_w, 0, NULL, NULL);
~~~

nmoinvaz: Looks great!

lib/curl_multibyte.c
jay: Anything that makes the function fail is an error so I'm going to leave it. When we skip paths, the function fails (returns false).

bagder:
Merge conflicts. Any plans to move forward on this?

jay:
I fixed the conflicts. Also I reversed and took @nmoinvaz's suggestion to use 'goto cleanup' instead of 'goto error' to eliminate any confusion in the logic.

Though I don't foresee this PR causing any problems I'm going to wait until the next feature window to land it.

nmoinvaz:
@bagder is it possible for this to be merged for the next release? Thanks.

bagder:
@nmoinvaz this PR is done by @jay and he can/will merge this when he thinks it is fine.

nmoinvaz:
Ok I wasn't sure. I just wanted to make sure it wasn't abandoned because you told us on another PR that you closed it because we didn't push hard enough for it. Thanks!

jay:
This is held up by a CI failure in a Windows UWP job that I haven't had time to reproduce, apparently there is a missing [GetFullPathNameW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew) but according to the doc it exists in UWP apps

~~~
In file included from D:/a/curl/curl/bld/src/CMakeFiles/curl.dir/Unity/unity_0_c.c:139:
D:/a/curl/curl/lib/curl_multibyte.c: In function 'fix_excessive_path':
D:/a/curl/curl/lib/curl_multibyte.c:150:20: error: implicit declaration of function 'GetFullPathNameW'; did you mean 'GetLongPathNameW'? [-Wimplicit-function-declaration]
  150 |   needed = (size_t)GetFullPathNameW(in_w, 0, NULL, NULL);
      |                    ^~~~~~~~~~~~~~~~
      |                    GetLongPathNameW
D:/a/curl/curl/lib/curl_multibyte.c:150:20: error: nested extern declaration of 'GetFullPathNameW' [-Werror=nested-externs]
ninja: build stopped: subcommand failed.
~~~

The earliest this will be in a release is 2 months from now because it's not going in the patch release coming this Wednesday

sergio-nsk:
1. curl defines `WIN32_LEAN_AND_MEAN`. Thus, `#include <fileapi.h>` should be in setup-win32.h.
2. Or Windows SDK should be upgraded.

jay:
setup-win32.h is used everywhere and we purposely use WIN32_LEAN_AND_MEAN so that all the extra stuff is not included in every unit. fileapi.h should be included because the file api is included by winbase which is included by windows.h regardless of lean and mean. not all sdks have the file api in fileapi.h. possibly this is a unity build issue?

jay:
@vszakats do you have any idea about what is causing this UWP CI job failure, before I set up a Windows environment that can build uwp and find out why the ci job is failing, or is there a better way to go into the ci or something without doing all that? AFAICT I'd have to add these packages:

https://github.com/curl/curl/blob/7eb8c048470ed2cc14dca75be9c1cdae7ac8498b/.github/workflows/windows.yml#L84-L95

so

~~~
pacman --noconfirm --ask 20 --noprogressbar --sync --needed \
  mingw-w64-x86_64-cc \
  mingw-w64-x86_64-cmake \
  openssh \
  mingw-w64-x86_64-openssl \
  mingw-w64-x86_64-libssh2 \
  mingw-w64-x86_64-libpsl \
  mingw-w64-x86_64-c-ares \
  mingw-w64-x86_64-winstorecompat-git
~~~

and then build with the flags from the uwp 'test' configuration

https://github.com/curl/curl/blob/7eb8c048470ed2cc14dca75be9c1cdae7ac8498b/.github/workflows/windows.yml#L181-L204

vszakats:
@jay To sort out if it's unity-related, perhaps the easiest is to duplicate the job and disable unity in one of them with `-DCMAKE_UNITY_BUILD=OFF`. (it doesn't look like one, but it may be worth a test.)

I've only tested UWP builds in CI. With MinGW, it can often help to peek into the headers. They are available here: https://downloads.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v12.0.0.tar.bz2
Or online:
https://github.com/mirror/mingw-w64/blob/93f3505a758fe70e56678f00e753af3bc4f640bb/mingw-w64-headers/include/fileapi.h#L67-L88

The header only defines it for non-UWP. The MS SDK defines it for UWP, suggesting a
MinGW header bug, and perhaps also an implib bug. Depending on which, it may help to
declare it manually for MinGW + UWP until the headers are fixed. If it's also missing from
its implib, this feature will have to be disabled for MinGW + UWP.

Since this function requires Win10, some sort of mechanism seems to be necessary anyway
to disable it for incompatible targets. To fix all this in one, maybe `GetFullPathNameW` could
be converted to a dynamic call and check at runtime if it's available. This could make it work
for all compatible targets, regardless of dev env.

PR #13526, opened 2024-05-03
doh: some cleanups in ECH related functions

bagder:
Reduce memcpy and alloc calls


bagder:
@sftcd you're more than welcome to check how badly I botch your work here...

sftcd:
> @sftcd you're more than welcome to check how badly I botch your work here...

LGTM (sorry for slow response)

PR #13530, opened 2024-05-04
tests: add pytest for --ciphers and --tls13-ciphers options

jan2000:


jan2000:
CI Linux: address-sanitizer leaked memory is fixed by https://github.com/curl/curl/pull/13531

icing (approved):
LGTM. The list of slips is large, but that just reflects the current variations on our backends.

tests/http/test_17_ssl_use.py
```diff
@@ -175,3 +175,68 @@ def test_17_06_localhost(self, env: Env, httpd, nghttpx, repeat, proto):
...
+                pytest.skip('boringssl does not support setting TLSv1.3 ciphers')
+            elif env.curl_uses_lib('wolfssl'):
+                extra_args = ['--ciphers', ':'.join(cipher_names)]
+            else:
+                extra_args = ['--tls13-ciphers', ':'.join(cipher_names)]
+        else:
+            # test setting TLSv1.2 ciphers
+            if env.curl_uses_lib('schannel'):
+                pytest.skip('schannel does not support setting TLSv1.2 ciphers by name')
+            elif env.curl_uses_lib('wolfssl'):
```
jan2000, line 228: Come to think of it, two remarks:

While exceptions (slips?) for ssl backends that I anticipated for are defined here, it should be mentioned that not all ssl backends are tested by pytest in CI. For example AFAIK sectranp and schannel are not.

Once mbedtls gains TLSv1.3 support, and this code is not updated, problems with setting TLSv1.3 ciphers will go on undetected. This can be anticipated by changing `pytest.skip('mbedtls does not support TLSv1.3')` to `succeed = False`, but if this is the correct/smart thing to do?

icing: That seems a good way to assure we detect the added support, yes.

icing (approved):

jan2000:
Rebased to latest master due to merge conflict. Also mbedtls TLSv1.3 support has been added to master since last discussed, so the mbedtls check has been updated accordingly.

bagder:
Thanks!

PR #13533, opened 2024-05-04
Curl_dyn_vaddf should return error instead of OK if memory fails

MonkeybreadSoftware:
I think you meant CURLE_OUT_OF_MEMORY to report if you are out of memory for vaprintf returning NULL.

Looks strange to return CURLE_OK after calling Curl_dyn_free.


danielgustafsson (approved):
LGTM

PR #13538, opened 2024-05-05
fixed typos

Servall4:


lib/setopt.c
```diff
@@ -2324,7 +2324,7 @@ CURLcode Curl_vsetopt(struct Curl_easy *data, CURLoption option, va_list param)
         Curl_hsts_cleanup(&data->hsts);
         data->hsts = data->share->hsts;
       }
-#endif   /* CURL_DISABLE_HTTP */
+#endif   /* CURL_DISABLE_HSTS */
```
danielgustafsson, line 2327: Nice catch, when pushing I will probably just remove this comment altogether instead since the `ifndef` is vertically very close and leaving these comments out on short blocks matches the style of the file.

docs/libcurl/curl_getdate.md
```diff
@@ -65,7 +65,7 @@ UTC. Supported formats include: -1200, MST, +0100.
 ## day of the week items
 
 Specifies a day of the week. Days of the week may be spelled out in full
-(using English): `Sunday', `Monday', etc or they may be abbreviated to their
+(using English): `Sunday`, `Monday`, etc or they may be abbreviated to their
```
danielgustafsson, line 68: This isn't supposed to be marked up as code, but is using the older style quoting of `quotedvalue'. This being parsed as markdown makes it render weird in Github so we should just make them 'quotedvalue' instead.

docs/libcurl/libcurl-security.md
```diff
@@ -380,7 +380,7 @@ CURLOPT_TIMEOUT(3) and/or CURLOPT_LOW_SPEED_LIMIT(3) options can
 be used to mitigate against this.
 
 A malicious server could cause libcurl to download an infinite amount of data,
-potentially causing all of memory or disk to be filled. Setting the
+potentially causing all memory or disk to be filled. Setting the
```
danielgustafsson, line 383: I would prefer to rephrase this since it's adding details where we discuss general issues. I would propose:

> potentially causing system resources to be exhausted resulting in a potential system, or application, crash.

Servall4: good point, thanks!

danielgustafsson:
Pushed the quoting and typo fixes but left out the whitespace changes.

PR #13539, opened 2024-05-05
mbedtls: support TLS 1.3

MAntoniak:
See #12876.

Question, where in the documentation can I write down the information that the required version of the mbedtls library is 3.6.0?
Should the tests be modified?

### DIFF
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -122,6 +122,10 @@ struct mbed_ssl_backend_data {
 #define mbedtls_strerror(a,b,c) b[0] = 0
 #endif
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && MBEDTLS_VERSION_NUMBER >= 0x03060000
+#define TLS13_SUPPORT
+#endif
+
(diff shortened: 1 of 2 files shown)


bagder:
> Question, where in the documentation can I write down the information that the required version of the mbedtls library is 3.6.0?

`docs/INTERNALS.md` lists minimum version of multiple dependencies. But is that really the required version? If so, should not you make the build/configure/cmake fail if we try to build with a too old version?

>  Should the tests be modified?

It seems `http/test_17_ssl_use.py` needs a poke.

Neustradamus:
To follow this important PR!

MAntoniak:
> `docs/INTERNALS.md` lists minimum version of multiple dependencies. But is that really the required version? If so, should we not make the build/configure/cmake fail if we try to build with a too old version?

I have thought through the topic of documentation and indeed there is no need to include information about the required version of mbedtls to support TLSv1.3. Version 3.6.0 is the first version which TLSv1.3 is not experimental, it is available by defaut.

bagder (approved):

bagder:
Thanks!

PR #13543, opened 2024-05-06
cookies: the max expire time is 400 days

bagder:
draft-ietf-httpbis-rfc6265bis-14 says:

 "The limit SHOULD NOT be greater than 400 days (34560000 seconds) in
 duration. The RECOMMENDED limit is 400 days in duration, but the user
 agent MAY adjust the limit. Max-Age attributes that are greater than
 the limit MUST be reduced to the limit."


bagder:
The draft is now in working group last call and I consider it likely that this wording is going to end up in a final version of the document.

The biggest flaw I think right now is that this change requires a debug build to be properly tested - since we want to set a fake time to be able to verify this correctly.

bagder:
@danielgustafsson I brought this back for consideration

danielgustafsson:
> @danielgustafsson I brought this back for consideration

Thanks! I think we should pursue this, will take it for a spin tomorrow for a review.

danielgustafsson:
It still rubs me the wrong way that we need to fudge dates with debug builds. I'm going to have one more go at creating a test which doesn't need that and if that fails we'll just go with this I reckon.

lib/cookie.c
```diff
@@ -368,6 +368,18 @@ static void strstore(char **str, const char *newstr, size_t len)
   *str = Curl_memdup0(newstr, len);
 }
 
+static time_t time_now(void)
+{
+#ifdef DEBUGBUILD
+  char *timestr = getenv("CURL_TIME");
+  if(timestr) {
+    unsigned long val = strtol(timestr, NULL, 10);
```
vszakats, line 376:
```suggestion
    long val = strtol(timestr, NULL, 10);
```
(or maybe change to `strtoul()`)

bagder:
The spec proposal is still only a draft. I don't want to turns so many cookie tests into debug-only tests. It now merge conflicts.

I let this go for now, and we will get back to this topic again later.

PR #13547, opened 2024-05-07
Deprioritize SecureTransport on macOS

MonkeybreadSoftware:
Moved SecureTransport behind OpenSSL, so we can build CURL with both and prefer using OpenSSL over SecureChannel. Our users still can opt-in to use SecureTransport, but a curl build with both will default to OpenSSL.

### DIFF
--- a/lib/vtls/vtls.c
+++ b/lib/vtls/vtls.c
@@ -1349,8 +1349,6 @@ const struct Curl_ssl *Curl_ssl =
   &Curl_ssl_multi;
 #elif defined(USE_WOLFSSL)
   &Curl_ssl_wolfssl;
-#elif defined(USE_SECTRANSP)
-  &Curl_ssl_sectransp;
 #elif defined(USE_GNUTLS)
   &Curl_ssl_gnutls;
 #elif defined(USE_MBEDTLS)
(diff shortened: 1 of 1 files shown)


danielgustafsson:
s/Channel/Transport/ =)

Neustradamus:
To follow this important PR!

jay:
> To follow this important PR!

To follow a thread click the subscribe button. Please don't post a reply to every thread you want to follow just to get subscribed.

subscribe

Neustradamus:
@jay: I know but I do not received emails if I do not publish a comment.

jay:
> I know but I do not received emails if I do not publish a comment.

If the subscribe button is not working for you please contact github support.

jay:
I don't see why we would change the order for everyone just to meet your use requirements. We have `--with-default-ssl-backend=openssl` for configure which maps to CURL_DEFAULT_SSL_BACKEND. For cmake you should be able to do -DCURL_DEFAULT_SSL_BACKEND=openssl.

For other build systems you'd probably have to pass it as a define to the compiler because there aren't separate options. For winbuild I think you'd have to pass it like `nmake CC="cl /DCURL_DEFAULT_SSL_BACKEND=openssl"` or there might be a better way I don't remember

MonkeybreadSoftware:
Thank you. I didn't know these switches and I can add them to my build scripts.

The patch here is basically to change the default order in case you don't specify this switch to deprioritize SecureTransport. SecureChannel for Windows is already lower in the list.

vszakats:
I agree with this, remembering this caused some pain (https://github.com/curl/curl-for-win/commit/fba339182fafd4dd06ae4e02a5a5002685f011c9 https://github.com/curl/curl-for-win/commit/55ff9b9790557efec943877365ded111f1838b4c) when adding macOS support to curl-for-win. Because on Windows, Schannel didn't take priority, so the macOS behaviour was unexpected. One more reason to deprioritize SecureTransport is that it's deprecated and also without TLS 1.3 support. Likely most users are better served by OpenSSL or its forks than by SecureTransport, and this will only become truer in the future.

PR #13560, opened 2024-05-08
BUG-BOUNTY.md: clarify the third party situation

bagder:
We do not pay bounties for problems in other libraries.

### DIFF
--- a/docs/BUG-BOUNTY.md
+++ b/docs/BUG-BOUNTY.md
@@ -67,6 +67,13 @@ infrastructure.
 The curl security team is the sole arbiter if a reported flaw is subject to a
 bounty or not.
 
+## Third parties
+
+The curl bug bounty does not cover flaws in third party dependencies
+(libraries) used by curl or libcurl. If the bug triggers because of curl
+behaving wrongly or abusing a third party dependency, the problem is rather in
+curl and not in the dependency and then the bounty might cover the problem.
+
 ## How are vulnerabilities graded?
 
 The grading of each reported vulnerability that makes a reward claim is


docs/BUG-BOUNTY.md
```diff
@@ -67,6 +67,13 @@ infrastructure.
 The curl security team is the sole arbiter if a reported flaw is subject to a
 bounty or not.
 
+## Third parties
+
+The curl bug bounty does not cover flaws in third party dependencies
+(libraries) used by curl or libcurl. If the bug triggers because of curl
+behaving wrongly or abusing a third party dependency, the problem is rather in
+curl and not in the dependency and then the bounty might cover the problem.
```
danielgustafsson, line 75: Maybe we should add here, or somewhere else, that any reporter who is unsure should report it anyways and we can assist in redirecting in case it's in a third-party. If in doubt, it's better to report.

bagder: I think the difficulty there is the phrasing. (And secondary *where* that phrasing would be put.)

We already have a steady stream of claimed suspected security issues that people submit that really had no place being submitted in the first place. I don't want our document to imply they did it right when they threw ridiculous claims at us.

But I also don't want to say that you need to be entirely sure when you submit an issue, because that would be silly.

danielgustafsson: No disagreement there, we don't want any more bogus reports, but we also don't want legit reports go missing. It's a hard balancing act for sure.

bagder: I'll save this for a separate edit.

PR #13562, opened 2024-05-08
h3, ngtcp2, error handling

icing:
- identify ngtcp2 and nghttp3 error codes that are fatal
- close quic connection on fatal errors
- refuse further filter operations once connection is closed

### DIFF
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -176,6 +176,8 @@ struct h3_stream_ctx {
 
 #define H3_STREAM_CTX(ctx,data)   ((struct h3_stream_ctx *)(\
             data? Curl_hash_offt_get(&(ctx)->streams, (data)->id) : NULL))
+#define H3_STREAM_CTX_ID(ctx,id)  ((struct h3_stream_ctx *)(\
+            Curl_hash_offt_get(&(ctx)->streams, (id))))
 
 static void h3_stream_ctx_free(struct h3_stream_ctx *stream)
 {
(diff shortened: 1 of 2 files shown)


lib/vquic/curl_ngtcp2.c
```diff
@@ -434,19 +481,33 @@ static int cb_recv_stream_data(ngtcp2_conn *tconn, uint32_t flags,
...
+
+    if(fatal) {
+      CURL_TRC_CF(cdata, cf, "[%" CURL_PRId64 "] fatal nghttp3 error on "
+                  "nghttp3_conn_read_stream() -> %" CURL_PRId64,
+                  stream_id, (curl_int64_t)nconsumed);
+    }
+    if(!data) {
+      CURL_TRC_CF(cdata, cf, "[%" CURL_PRId64 "] discard data for unknown "
+                  "stream", stream_id);
+      return 0;
```
tatsuhiro-t, line 498: Connection must be closed (return NGTCP2_ERR_CALLBACK_FAILURE) when nconsumed < 0 with no exception.

icing: Aye. Will change this then.

icing: That does not work. When I reset stream 0, and packets arrive, our callback is invoked for stream 0 and the call to `nghttp3_conn_read_stream()` returns -609. If we then return NGTCP2_ERR_CALLBACK_FAILURE the connection and all other ongoing transfers break.

Any ideas?

tatsuhiro-t: I think that that is due to the misuse of nghttp3_conn_shutdown_stream_read and nghttp3_conn_close_stream at https://github.com/curl/curl/blob/e1cf21b5c2572bc05504c1b3a39d1c0620123e75/lib/vquic/curl_ngtcp2.c#L236-L238

Calling these two functions do not actually shutdown stream or close stream.   They should be called after stream is actually shutdown or closed.   I am not sure when h3_data_done is called, but could be called before actual QUIC stream is closed.  Instead of calling them, you should call ngtcp2_conn_shutdown_stream.  After that, ngtcp2 stops notifying data on the closed stream.

icing: Thanks for the explanation. A misunderstanding how to use the API then. Will try to kill the ngtcp2 stream then.

icing: Should be fixed now in latest push. Thanks for the help!

PR #13574, opened 2024-05-10
lib: clear the easy handle's saved errno before transfer

jay:
- Clear data->state.os_errno before transfer.

- Explain the change in behavior in the CURLINFO_OS_ERRNO doc.

- Add to the CURLINFO_OS_ERRNO doc the list of libcurl network-related errors that may cause the errno to be saved.

data->state.os_errno is saved before libcurl returns a network-related failure such as connection failure. It is accessible to the user via CURLINFO_OS_ERRNO so they can get more information about the failure.

Prior to this change it wasn't cleared before transfer so if a user retrieved the saved errno it could be from a previous transfer. That is because an errno is not always saved for network-related errors.

Closes #xxxxx


bagder (approved):
Nice catch!

PR #13575, opened 2024-05-10
CI: ignore test 286 on Appveyor gcc 7 build

vszakats:
Disabled earlier for gcc 9 builds. gcc 7 uses the same runner and
prone to similar intermittent failures.

Follow-up to f1e05a6e6e7225fa09952abb2c935ae1abe44f45 #12106 #12040
Closes #13575

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -140,7 +140,8 @@ environment:
       ENABLE_UNICODE: 'ON'
       HTTP_ONLY: 'OFF'
       TESTING: 'ON'
-      DISABLED_TESTS: '!1086 !1139 !1451 !1501 !1177 !1477'
+      # test 286 disabled due to https://github.com/curl/curl/issues/12040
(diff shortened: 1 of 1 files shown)


dfandrich (approved):
286 is 3% flaky in this build so this is probaby a good idea. It's also 3.7% flaky in the gcc 9 build, so maybe you want to do that one at the same time.

vszakats:
> 286 is 3% flaky in this build so this is probaby a good idea. It's also 3.7% flaky in the gcc 9 build, so maybe you want to do that one at the same time.

gcc 9 should be fine after your f1e05a6e6e7225fa09952abb2c935ae1abe44f45 commit earlier. This one is just a copy of that, for gcc 7. (unless I'm misreading something!)

dfandrich:
Sorry, I misread my tool; the gcc 9 error *is* already being ignored.

PR #13589, opened 2024-05-10
lib: Add explicit_bzero to zero out sensitive buffers

danielgustafsson:
Sharing this early as a proof-of-concept to hear others thoughts on this. The CMake and Windows parts are untested so getting them onto the CI is a motivator.

This adds cross platform support for zeroing out buffers containing sensitive content without the zeroing being optimized away by the compiler due being a dead-store. In case the platform doesn't have `bzero`, `explicit_bzero` (available on Archlinux etc) or `memset_s` there is a fallback which use a plain `memset` via a volatile pointer.

A set of buffers are included here, going through and finding all potential buffers is future work.

This is influenced on how OpenSSH and PostgreSQL zeroes out sensitive buffers. It clearly won't move the security needle all that much but every little bit helps.


jay:
I would think the secrets are probably still in memory somewhere. I'd want to hear from some security experts to know whether this is really worth it for us and actually does what we expect.

CMakeLists.txt
```diff
@@ -1302,6 +1302,9 @@ check_symbol_exists(getrlimit      "${CURL_INCLUDES}" HAVE_GETRLIMIT)
 check_symbol_exists(setlocale      "${CURL_INCLUDES}" HAVE_SETLOCALE)
 check_symbol_exists(setmode        "${CURL_INCLUDES}" HAVE_SETMODE)
 check_symbol_exists(setrlimit      "${CURL_INCLUDES}" HAVE_SETRLIMIT)
+check_symbol_exists(bzero          "${CURL_INCLUDES};strings.h" HAVE_BZERO)
+check_symbol_exists(explicit_bzero "${CURL_INCLUDES};strings.h" HAVE_EXPLICIT_BZERO)
+check_symbol_exists(memset_s       "${CURL_INCLUDES};string.h" HAVE_MEMSET_S)
```
vszakats, line 1307: It's safe to enclose these check inside `if(NOT WIN32)`. Saves time.

(libssh2 does this.)

configure.ac
```diff
@@ -3754,6 +3754,9 @@ CURL_CHECK_FUNC_STRERROR_R
 CURL_CHECK_FUNC_STRICMP
 CURL_CHECK_FUNC_STRTOK_R
 CURL_CHECK_FUNC_STRTOLL
+CURL_CHECK_FUNC_BZERO
+CURL_CHECK_FUNC_EXPLICIT_BZERO
+CURL_CHECK_FUNC_MEMSET_S
```
vszakats, line 3759: Probably here too, with `"$curl_cv_native_windows" != "yes"`.

vszakats:
There exists one more variant: `explicit_memset`:
```cmake
  check_symbol_exists("explicit_memset" "string.h" HAVE_EXPLICIT_MEMSET)
```

```c
#define _libssh2_explicit_zero(buf, size) (void)explicit_memset(buf, 0, size)
```

https://man.netbsd.org/explicit_memset.3

CMakeLists.txt
vszakats: I think `${CURL_INCLUDES};` is unnecessary for these checks.

CMakeLists.txt
```diff
@@ -1302,6 +1302,9 @@ check_symbol_exists(getrlimit      "${CURL_INCLUDES}" HAVE_GETRLIMIT)
 check_symbol_exists(setlocale      "${CURL_INCLUDES}" HAVE_SETLOCALE)
 check_symbol_exists(setmode        "${CURL_INCLUDES}" HAVE_SETMODE)
 check_symbol_exists(setrlimit      "${CURL_INCLUDES}" HAVE_SETRLIMIT)
+check_symbol_exists(bzero          "${CURL_INCLUDES};strings.h" HAVE_BZERO)
+check_symbol_exists(explicit_bzero "${CURL_INCLUDES};strings.h" HAVE_EXPLICIT_BZERO)
```
vszakats, line 1306: This page:
https://man.archlinux.org/man/explicit_bzero.3.en
suggests that the header is `string.h` (without the `s`).
(this is also what libssh2 uses.)

lib/strzero.c
```diff
@@ -0,0 +1,96 @@
...
+  bzero(buffer, length);
+}
+#elif defined(HAVE_EXPLICIT_BZERO)
+Curl_explicit_bzero(void *buffer, size_t length)
+{
+  if(!buffer)
+    return;
+  explicit_bzero(buffer, length);
+}
+#elif defined(_WIN32)
```
vszakats, line 58: Would it make sense to move this to the top of the list and use this native function always, on Windows?

It's an intrinsic function and should work in all versions.

lib/strzero.c
```diff
@@ -0,0 +1,96 @@
...
+
+#include "strzero.h"
+
+/*
+ * Curl_explicit_bzero zeroes out a memory buffer in such a way to ensure that
+ * it won't be optimized away by the compiler due to being a dead-store. If no
+ * API for this exists it will fall back to plain memset using a volatile ptr.
+ */
+
+#ifdef HAVE_MEMSET_S
```
vszakats, line 35: Minor nit to keep the order of these blocks conistent with the feature checks: bzero, explicit_bzero, then memset_s.

vszakats:
In libssh2 the wrapper for all the variants is a macro. It has the advantage to inline the native Windows call (and maybe others if they have a similar implementation). Throwing it here as an idea to consider.

https://github.com/libssh2/libssh2/blob/d19b61907039554b8261a90898f2f31e1a706f38/src/misc.h#L42-L59

lib/strzero.c
```diff
@@ -0,0 +1,96 @@
...
+}
+#elif defined(_WIN32)
+#include <windows.h>
+void
+Curl_explicit_bzero(void *buffer, size_t length)
+{
+  if(!buffer)
+    return;
+  /*
+   * TODO: Windows 11 ships a new function SecureZeroMemory2 which is a safer
```
vszakats, line 66: MSDN says the new variant is safe to use in older Windows versions (suggesting this is also an intrinsic):
https://learn.microsoft.com/en-us/windows/win32/memory/winbase-securezeromemory2#remarks
Given the necessary SDK at build-time and doing the detection dance, we might build with it without breaking in older envs.
It requires a recent MSVC; mingw-w64 doesn't support it yet (as of v11).

CMakeLists.txt
danielgustafsson: Meh, of course, that's a copy-pasteo

lib/strzero.c
danielgustafsson: Fair point.

lib/strzero.c
danielgustafsson: I've removed `bzero` in the updated version, I agree that `memset`  via a `volatile` pointer offers stronger guarantees here.

lib/strzero.c
danielgustafsson: I don't know Windows development even if it jumps up and hits me in the behind (ironic given my employer, I know), so I'm not sure how to use this in a safe way for the various ways to build on Windows.

CMakeLists.txt
danielgustafsson: I think so too, but we use `${CURL_INCLUDES}` on all the other checks where it's likely not needed so I went for consistency. Maybe a cleanup patch taking all checks into consideration would be in order?

configure.ac
danielgustafsson: Wrapped this in a windows check.

danielgustafsson:
> There exists one more variant: `explicit_memset`:

Interesting, we even use it in the code for the NetBSD bug (which is active for NetBSD < 9 but explicit_memset is only available for 7.0+). Added support for it.

danielgustafsson:
> In libssh2 the wrapper for all the variants is a macro. It has the advantage to inline the native Windows call (and maybe others if they have a similar implementation). Throwing it here as an idea to consider.

Absolutely, that's a possibility for sure. This is unlikely to be used in performance critical codepaths but might still be worth doing?

danielgustafsson:
One more TODO is to move clearing of values to when they're no longer needed and not just before the `free` call.

CMakeLists.txt
vszakats: It's though because each of them needs to be carefully validated. I didn't dare touching it yet, though it's on my TODO list. For these new ones, we can be sure it's the only headers needed.

lib/strzero.c
vszakats: Fair enough! :) I don't have MSVC (or Windows) around, which makes this difficult, so I think the next best thing is to wait till someone figures it out. For now, it's so new that GitHub shows zero actual use.

vszakats:
> > In libssh2 the wrapper for all the variants is a macro. It has the advantage to inline the native Windows call (and maybe others if they have a similar implementation). Throwing it here as an idea to consider.
> 
> Absolutely, that's a possibility for sure. This is unlikely to be used in performance critical codepaths but might still be worth doing?

You're right. Let's see if this starts to hit something performance critical and deal with it then.

CMakeLists.txt
```diff
@@ -1302,6 +1302,13 @@ check_symbol_exists(getrlimit      "${CURL_INCLUDES}" HAVE_GETRLIMIT)
 check_symbol_exists(setlocale      "${CURL_INCLUDES}" HAVE_SETLOCALE)
 check_symbol_exists(setmode        "${CURL_INCLUDES}" HAVE_SETMODE)
 check_symbol_exists(setrlimit      "${CURL_INCLUDES}" HAVE_SETRLIMIT)
+# On Windows we use SecureMemoryZero which is available in all supported
+# versions of Windows
+if(NOT WIN32)
+  check_symbol_exists(memset_s       "${CURL_INCLUDES};string.h" HAVE_MEMSET_S)
+  check_symbol_exists(explicit_bzero "${CURL_INCLUDES};string.h" HAVE_EXPLICIT_BZERO)
+  check_symbol_exists(explicit_memset "${CURL_INCLUDES};string.h" HAVE_EXPLICIT_BZERO)
```
vszakats, line 1310:
```suggestion
  check_symbol_exists(memset_s "string.h" HAVE_MEMSET_S)
  if(NOT HAVE_MEMSET_S)
    check_symbol_exists(explicit_bzero "string.h" HAVE_EXPLICIT_BZERO)
    if(NOT HAVE_EXPLICIT_BZERO)
      check_symbol_exists(explicit_memset "string.h" HAVE_EXPLICIT_BZERO)
    endif()
  endif()
```
What do you think of making this a bit faster by stopping on the first hit?
Same would be useful for autotools, if there is a clean way.

danielgustafsson:
> I would think the secrets are probably still in memory somewhere. I'd want to hear from some security experts to know whether this is really worth it for us and actually does what we expect.

This is pretty standard operating procedure in code dealing with sensitive information (OpenSSL, OpenSSH, libssh2 et.al), defined as [CWE-14](https://cwe.mitre.org/data/definitions/14.html), and provides basic code hygiene. I agree that it won't push the needle all that far, but provides more defense in depth.

lib/strzero.c
```diff
@@ -0,0 +1,95 @@
...
+   * method, but is not as widely available as SecureZeroMemory which exist
+   * all the way back to Windows XP.
+   */
+  (void)SecureZeroMemory(buffer, length);
```
vszakats, line 47: According to this copy of the SDK:
https://github.com/hughbe/windows-sdk-headers/blob/89f151d343587fcd5b854a4851fcb8f7187f3ac4/Microsoft.Windows.SDK.CPP/latest/Include/um/WinBase.h#L123
a simple `#ifdef` might be enough to detect it, at least with MSVC. We'd need to test it before adding it.

danielgustafsson: Hmm, I agree that it seems like an ifdef is all we need. I've added it to the code in this rebase for a test to see if any blows up.

CMakeLists.txt
danielgustafsson: That's not a bad idea, any way we can speed up the build is good. Added in the rebase to CMake for now.

danielgustafsson:
With the window open again, any :+1: or :-1: on this? (it needs a rebase of course)

lib/vauth/digest_sspi.c
```diff
@@ -666,6 +666,8 @@ void Curl_auth_digest_cleanup(struct digestdata *digest)
 
   /* Free the copy of user/passwd used to make the identity for http_context */
   Curl_safefree(digest->user);
+  if(digest->passwd)
+    Curl_explicit_bzero(digest->passwd, strlen(digest->passwd);
```
vszakats, line 670:
```suggestion
    Curl_explicit_bzero(digest->passwd, strlen(digest->passwd));
```
This should fix AppVeyor CI builds.

bagder:
As this has fallen a bit behind I'm closing. Feel free to reopen if you want to rebase and have another go.

CMakeLists.txt
vszakats: > It's tough because each of them needs to be carefully validated. I didn't dare touching it yet, though it's on my TODO list. For these new ones, we can be sure it's the only headers needed.

This has landed since via in 8e345057761a8f796403923a96f2c8fd3edca647 #15164 .

PR #13583, opened 2024-05-10
GHA: add NetBSD, OpenBSD, FreeBSD/arm64 and OmniOS jobs

vszakats:
Add these jobs to GHA:
- NetBSD, cmake-unity, clang, OpenSSL, x86_64, with tests, w/o python,
  no parallelism (was flaky sometimes)
- OpenBSD, cmake-unity, clang, LibreSSL, x86_64, with tests,
  with python, -j8, TFTP results ignored due to #13623.
- FreeBSD, cmake-unity and autotools, clang, OpenSSL, arm64
  (Tests disabled for arm64, because they are slow. It's available for
  x86_64 with python, -j12.)
  Configuration matches our existing Cirrus CI one.
- OmniOS, autotools, gcc, OpenSSL, x86_64, with tests, -j12.

All build with websockets and examples.

Closes #13583

---

- [x] cmake debug configuration might be simplified once #13592 (or a variant) landed.
- [x] see how to make `cpp` work on OmniOS. [FIXED by symlinking `cpp` to `gcpp`.]
- [x] fix test1167 on OpenBSD. → #13634
- [x] FIXME: TFTP tests on OpenBSD: https://github.com/curl/curl/actions/runs/9042288591/job/24855020568?pr=13583#step:3:3443 → #13623 → converted to a FIXME comment and executing these test but ignoring the result.
- [x] convert FreeBSD job to cmake for more variation.

PR #13592, opened 2024-05-11
cmake: `ENABLE_DEBUG=ON` to always set `-DDEBUGBUILD`

vszakats:
Before this patch `ENABLE_DEBUG=ON` always enabled the TrackMemory
(aka `ENABLE_CURLDEBUG=ON`) feature, but required the `Debug` CMake
configration to actually enable curl debug features
(aka `-DDEBUGBUILD`).

Curl debug features do not require compiling with C debug options. This
also made enabling debug features unintuitive and complicated to use.
Due to other issues (subject to PR #13694) it also caused an error in
default (and `Release`/`MinSizeRel`/`RelWithDebInfo`) configs, when
building the `testdeps` target:
```
ld: CMakeFiles/unit1395.dir/unit1395.c.o: in function `test':
unit1395.c:(.text+0x1a0): undefined reference to `dedotdotify'
```
Ref: https://github.com/curl/curl/actions/runs/9037287098/job/24835990826#step:3:2483

Fix it by always defining `DEBUGBUILD` when setting `ENABLE_DEBUG=ON`.
Decoupling this option from the selected CMake configuration.

Note that after this patch `ENABLE_DEBUG=ON` unconditionally enables
curl debug features. These features are insecure and unsuited for
production. Make sure to omit this option when building for production
in default, `Release` (and other not-`Debug`) modes.

Also delete a workaround no longer necessary in GHA CI jobs.

Ref: 1a62b6e68c08c7e471ff22dd92932aba7e026817 (2015-03-03)
Ref: #13583
Closes #13592

---

- [x] Depends on #13643.
- [x] Rebase on #13654.
- [x] Rebase on #13668.
- [x] settle on `ENABLE_DEBUG` to always set `DEBUGBUILD`
- [x] try untangling `DEBUGBUILD` / `CURLDEBUG` / `UNITTESTS` guards. AFAIU the first two should work independently of each other, whether we are building/running tests or not, whether we are building with debug info or not. → in separate PR.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -240,7 +240,6 @@ jobs:
           fi
           [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
           [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
-          [ '${{ matrix.type }}' = 'Release' ] && [[ '${{ matrix.config }}' = *'ENABLE_DEBUG=ON'* ]] && cflags+=' -DDEBUGBUILD'
           cmake -B bld ${options} \
             "-DCMAKE_C_FLAGS=${cflags}" \
             "-DCMAKE_RC_COMPILE_OBJECT=${rcopts}" \
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -292,7 +292,7 @@ jobs:
             env: '9.5.0-x86_64'
             url: 'https://github.com/brechtsanders/winlibs_mingw/releases/download/9.5.0-10.0.0-msvcrt-r1/winlibs-x86_64-posix-seh-gcc-9.5.0-mingw-w64msvcrt-10.0.0-r1.7z'
             config: '-DENABLE_DEBUG=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=OFF'
-            type: 'Debug'
+            type: 'Release'
             tflags: '~1139 ~1177 ~1477 ~2301 ~2302 ~3027'
           - build: 'cmake'
             env: '7.3.0-x86_64'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,8 +119,7 @@ option(ENABLE_CURLDEBUG "Set to ON to build with TrackMemory feature enabled" OF
 include(PickyWarnings)
 
 if(ENABLE_DEBUG)
-  # DEBUGBUILD will be defined only for Debug builds
-  set_property(DIRECTORY APPEND PROPERTY COMPILE_DEFINITIONS $<$<CONFIG:Debug>:DEBUGBUILD>)
+  set_property(DIRECTORY APPEND PROPERTY COMPILE_DEFINITIONS DEBUGBUILD)
   set(ENABLE_CURLDEBUG ON)
 endif()
 
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -43,7 +43,7 @@ if [ "${BUILD_SYSTEM}" = 'CMake' ]; then
   [ "${PRJ_CFG}" = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
   [[ "${PRJ_GEN}" = *'Visual Studio'* ]] && options+=' -DCMAKE_VS_GLOBALS=TrackFileAccess=false'
   if [ "${PRJ_GEN}" = 'Visual Studio 9 2008' ]; then
-    [ "${PRJ_CFG}" = 'Debug' ] && [ "${DEBUG}" = 'ON' ] && [ "${SHARED}" = 'ON' ] && SKIP_RUN='Crash on startup in -DDEBUGBUILD shared builds'
+    [ "${DEBUG}" = 'ON' ] && [ "${SHARED}" = 'ON' ] && SKIP_RUN='Crash on startup in -DDEBUGBUILD shared builds'
     # Fails to run without this due to missing MSVCR90.dll / MSVCR90D.dll
     options+=' -DCURL_STATIC_CRT=ON'
   fi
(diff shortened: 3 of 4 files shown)


vszakats:
MSVC build with `ENABLE_DEBUG=ON` and `Release`; `curl --version` crashes:
https://ci.appveyor.com/project/curlorg/curl/builds/49792835/job/eup3qgdx5jgt9wk9

vszakats:
- Ended up keeping `ENABLE_CURDEBUG` because some builds set this and it's a subset of `ENABLE_DEBUG`, not exactly the same.
- A `Debug` variant of `CMake, VS2008, Release, x86, Schannel, Build-only` segfaults with and without this patch.
- New compiler warnings appeared when building the Release + `ENABLE_DEBUG=ON` combo.
- Could not yet figure out how to make the VS2008 job _verbose_. It'd help seeing the actual compiler command-lines.

vszakats:
I'm not quite sure that enabling `-DDEBUGBUILD` unconditionally on `ENABLE_DEBUG=ON` is the right way to go. Reason is that it makes it a bit _too easy_ to enable debug features, which are fatal for security. This can affect existing build configs which happen to set `ENABLE_DEBUG=ON` while doing a `Release` build. With the patch, this would produce a debug build proper, while without it was only enabling memory debugging (aka `ENABLE_CURLDEBUG=ON`), though probably that's also not desired for release builds.

Not doing anything leaves the `ENABLE_DEBUG=ON + ENABLE_TESTING=ON` builds broken by default, which is also undesired (tests are very well runnable without debug info), though it likely affects very few users outside our own CI infrastructure.

One way to solve this cleanly, is introducing a new option that has the new behaviour. This could be called `ENABLE_DEBUGBUILD=ON`. This could also _not_ set `-DCURLDEBUG` to give more direct control over these options.

vszakats:
Another option is to set `-DDEBUGBUILD` unconditionally when both `ENABLE_DEBUG` and `ENABLE_TESTING` are set.

vszakats:
Another solution is to require `Debug` builds for `BUILD_TESTING`.

jay:
I take issue with "too easy" being a bad thing. My thinking is if the user has made a choice to enable curl debug builds then why would we not honor that. Also, that's the way we already do it for autotools, winbuild and cmake. So my opinion is leave that option as is.

As far as another option, I have to wonder why does running the test suite in cmake require DEBUGBUILD? Couldn't that be the problem instead of adding DEBUGBUILD with another option?

vszakats:
What I've found so far is that running or building tests don't require `DEBUGBUILD`, but building them with `CURLDEBUG` does. The problem is that `ENABLE_DEBUG` always sets the latter, but sets the former only when building in cmake `Debug` config type. I don't know why that choice was made (1a62b6e68c08c7e471ff22dd92932aba7e026817). Perhaps because in autotools the equivalent option `--enable-debug` also enables compiler debug options, like `Debug` config does? FWIW in autotools `--enable-debug` always sets `DEBUGBUILD`.

Changing `ENABLE_DEBUG` to also always set `DEBUGBUILD` would be the easiest fix. I'm also okay with it making this option easier to use (that was actually my goal after trying setup tests with CMake in new CI jobs), but are we also okay that this changes build behavior just enough to accidentally enable `DEBUGBUILD` for the the `ENABLE_DEBUG` + default (no cmake config type) or `Release`/`MinSizeRel`/`RelWithDebInfo` config types? I we can agree that this is fine, I'm all for this solution.

Why `DEBUGBUILD` is required with `CURLDEBUG` to build certain tests? Based on CMake it seems that without `CURLDDEBUG`, unit tests are not built, so the problem doesn't surface. When bulding them, e.g. `test1395` calls `dedotdotify()` which is declared with `UNITTEST` in `lib/urlapi.c`, where `UNITTEST` is set to `static` unless `DEBUGBUILD` is set.

So perhaps the root cause is that the `UNITTEST` macro should depend on `CURLDEBUG` and not `DEBUGBUILD`? Ref 05b100aee247bb9bec8e9a1b0166496aa4248d1c.

vszakats:
Settled with the original concept to always enable `-DDEBUGBUILD` when `ENABLE_DEBUG=ON`.

vszakats:
> I take issue with "too easy" being a bad thing. My thinking is if the user has made a choice to enable curl debug builds then why would we not honor that. Also, that's the way we already do it for autotools, ~winbuild~ _(edit: winbuild has a DEBUG=yes but only to enable compiler debug flags)_ and cmake. So my opinion is leave that option as is.

@jay: This patch now makes `ENABLE_DEBUG=ON` honor this request and enables curl debug feature regardless of compiler debug flags. Before this patch it required a cmake `Debug` build. Do you agree with the change?

> As far as another option, I have to wonder why does running the test suite in cmake require DEBUGBUILD? Couldn't that be the problem instead of adding DEBUGBUILD with another option?

I chased that down, which became a tree of fixes, with its main portion in #13694.

(There is no new option to enable `DEBUGBUILD`, `ENABLE_DEBUG=ON` does that now unconditionally.)

jay:
> @jay: This patch now makes `ENABLE_DEBUG=ON` honor this request and enables curl debug feature regardless of compiler debug flags. Before this patch it required a cmake `Debug` build. Do you agree with the change?

Seems right to me since it's documented as enabling curl debug features. It's confusing that --enable-debug and ENABLE_DEBUG are not exactly the same since ENABLE_DEBUG does not actually enable compiler debugging. But then again that's not the way cmake works. I'm not sure what we can or should do about that.

vszakats:
> Seems right to me since it's documented as enabling curl debug features.

Thanks!

> It's confusing that --enable-debug and ENABLE_DEBUG are not exactly the same since ENABLE_DEBUG does not actually enable compiler debugging. But then again that's not the way cmake works. I'm not sure what we can or should do about that.

Agreed.

What we could do, is setting `ENABLE_DEBUG` to ON` by default when the `Debug` configuration is selected. That seems like a natural fit for CMake and also matches what autotools does, assuming `--enable-debug` is the counterpart of `--config Debug` in CMake.

jay:
> What we could do, is setting ENABLE_DEBUG to ON by default when the Debug configuration is selected. That seems like a natural fit for CMake and also matches what autotools does, assuming --enable-debug is the counterpart of --config/CMAKE_BUILD_TYPE = Debug in CMake.

cmake is just different from autotools and the user expectations are different. For example a user may specify a debug configuration because they want compiler debug settings but not curl's DEBUGBUILD. I think it is better to leave it as a separate option, even if that's not what autotools is doing.

PR #13603, opened 2024-05-11
hsts: explicitly skip blank lines

danielgustafsson:
Keep blank lines or lines containing only whitespace to make it all the way to the more expensive sscanf call in hsts_add.

### DIFF
--- a/lib/hsts.c
+++ b/lib/hsts.c
@@ -528,8 +528,11 @@ static CURLcode hsts_load(struct hsts *h, const char *file)
       char *lineptr = Curl_dyn_ptr(&buf);
       while(*lineptr && ISBLANK(*lineptr))
         lineptr++;
-      if(*lineptr == '#')
-        /* skip commented lines */
+      /*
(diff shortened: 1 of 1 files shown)


lib/hsts.c
```diff
@@ -528,8 +528,8 @@ static CURLcode hsts_load(struct hsts *h, const char *file)
       char *lineptr = Curl_dyn_ptr(&buf);
       while(*lineptr && ISBLANK(*lineptr))
         lineptr++;
-      if(*lineptr == '#')
-        /* skip commented lines */
+      if((*lineptr == '#') || strlen(lineptr) == 0)
+        /* skip empty or commented lines */
```
bagder, line 532: Since `Curl_get_line()` works hard to return lines with a trailing newline, should this perhaps rather check for `<= 1` ?

danielgustafsson: That's a good point, it should. Rebased the PR with this fix.

bagder (approved):

PR #13599, opened 2024-05-11
GHA: add MSYS, mingw-w64, Cygwin jobs

vszakats:
- re-implement autotools MSYS and Cygwin AppVeyor jobs in GHA.
  Now build with SSL and PSL to improve test coverage.
- re-implement MSYS2 mingw-w64 gcc 13 AppVeyor job in GHA.
  `CMake, mingw-w64, gcc 13, Debug, x64, Schannel, Static, Unicode`
- add new cmake Cygwin job (build-only).
- enable `-j14` parallelism when running tests.
- delete the 5 migrated jobs from AppVeyor CI.
- add 2 build-only mingw-w64 builds, gcc Release and clang OpenSSL.
- also enable brotli, libssh2, nghttp2 for more test coverage.

These jobs offer better performance, more flexibility and
parallelization compared to the AppVeyor ones they replace. It also
offloads AppVeyor, allowing to iterate faster. They also appear more
reliable than e.g. Azure Windows jobs, where runners are prone to fail
[^1].

Closes #13599

[^1]: `Exit code 143 returned from process: file name 'C:\Windows\system32\docker.EXE', arguments 'exec -i   6b13a669c6dfe7fb9f59414369872fd64d61c7182f880c3d39c135cb4c115c8f C:\__a\externals\node\bin\node.exe C:\__w\_temp\containerHandlerInvoker.js'.`

---

- [x] Also migrate the mingw-w64 gcc 13 AppVeyor job?
- [x] Delete the 5 AppVeyor CI jobs if these ones work as expected?
- [x] Test mingw-w64 in `Release` CMake configuration.
- [x] Depends on #13643.

PR #7762, opened 2021-09-23
add IP TOS support

zhao85:
add IP type of service support when any socket is creating

PR #13606, opened 2024-05-12
Support IP Type of Service / Traffic Class: --ip-tos

orgads:
Add `--ip-tos` option to the command line tool for setting TOS for IPv4 or Traffic Class for IPv6.


CMake/CurlTests.c
```diff
@@ -318,6 +318,29 @@ int main(void)
 }
 #endif
 
+#ifdef HAVE_SETSOCKOPT_SOL_IP
```
vszakats, line 321: It'd be useful to pre-cache this for Windows, which always seems to fail this test:
```diff
--- a/CMake/Platforms/WindowsCache.cmake
+++ b/CMake/Platforms/WindowsCache.cmake
@@ -170,6 +170,7 @@ set(HAVE_IOCTLSOCKET_CAMEL_FIONBIO 0)
 set(HAVE_IOCTLSOCKET_FIONBIO 1)
 set(HAVE_IOCTL_FIONBIO 0)
 set(HAVE_SETSOCKOPT_SO_NONBLOCK 0)
+set(HAVE_SETSOCKOPT_SOL_IP 0)
 set(HAVE_POSIX_STRERROR_R 0)
 set(HAVE_BUILTIN_AVAILABLE 0)
 set(HAVE_MSG_NOSIGNAL 0)
```

orgads: Nice, thanks!

orgads:
I don't know what to do with the remaining tests that fail. Can you please advise?

bagder:
Why is a new libcurl option needed? Can't `CURLOPT_SOCKOPTFUNCTION` already do this?

bagder:
This is a IPv4-only feature. What happens for IPv6 sockets and how is a user supposed to know when this works or not?

Can you please explain your use case for this?

src/tool_getparam.c
```diff
@@ -999,6 +1004,76 @@ static CURLcode set_trace_config(struct GlobalConfig *global,
   return result;
 }
 
+#ifdef HAVE_SETSOCKOPT_SOL_IP
```
bagder, line 1007: The command line option parsing cannot depend on this. If it should, it needs to check if *this* libcurl supports it in run-time.

orgads:
> Why is a new libcurl option needed? Can't `CURLOPT_SOCKOPTFUNCTION` already do this?

We use it with the command-line interface. And if I understand correctly, `CURLOPT_SOCKOPTFUNCTION` is for user applications, not for internal use by curl, right?

src/tool_getparam.c
orgads: The problem is that all the "user-friendly" values here are defined in `netinet/ip.h`, which is not available on all platforms. What do you suggest to do? Maybe support only numeric values on other platforms? But what for actually, if it has no effect?

orgads:
> This is a IPv4-only feature. What happens for IPv6 sockets and how is a user supposed to know when this works or not?
> 
> Can you please explain your use case for this?

We need it for applying QoS for IP packets. All the other parts of our application support this, and we've been patching curl for long time to add support. Now we'd like to upstream our patches (2 more will follow :)).

I noticed that another used proposed a similar change in #7762, but it was abandoned. I took the user-friendly names from that PR.

bagder:
> > Why is a new libcurl option needed? Can't `CURLOPT_SOCKOPTFUNCTION` already do this?
> 
> We use it with the command-line interface. And if I understand correctly, `CURLOPT_SOCKOPTFUNCTION` is for user applications, not for internal use by curl, right?

You add `CURLOPT_TYPE_OF_SERVICE` in this PR. It is not for the command line tool, that's a new option for libcurl and I wonder why you need it.

src/tool_getparam.c
bagder: We work hard to decouple the tool from the library and its specific versions and features. What if you run with a different build of the library - with the same tool version, how would the tool know what features the library supports?

orgads:
> > > Why is a new libcurl option needed? Can't `CURLOPT_SOCKOPTFUNCTION` already do this?
> > 
> > 
> > We use it with the command-line interface. And if I understand correctly, `CURLOPT_SOCKOPTFUNCTION` is for user applications, not for internal use by curl, right?
> 
> You add `CURLOPT_TYPE_OF_SERVICE` in this PR. It is not for the command line tool, that's a new option for libcurl and I wonder why you need it.

It is used in `tool_operate.c`. If I understand the relations correctly, this sets the library options according to the command-line arguments. So do you suggest using `CURLOPT_SOCKOPTFUNCTION` there? I don't see any other option that uses it there, why should this one be different?

src/tool_getparam.c
orgads: If you prefer that, I'll just remove the readable names, and let it just accept a numeric value (that's what we use anyway. The idea to use names was taken from #7762). Then I can remove the condition.

orgads: By the way, `ECH` is also disabled in command-line with `#ifdef`. Is that wrong too?

src/tool_getparam.c
```diff
@@ -999,6 +1004,76 @@ static CURLcode set_trace_config(struct GlobalConfig *global,
...
+    return IPTOS_LOWDELAY;
+  if(!strcmp(nextarg, "THROUGHPUT"))
+    return IPTOS_THROUGHPUT;
+  if(!strcmp(nextarg, "RELIABILITY"))
+    return IPTOS_RELIABILITY;
+  if(!strcmp(nextarg, "MINCOST"))
+    return IPTOS_MINCOST;
+  /* user defined tos value */
+  if(PARAM_OK == str2unummax(&res, nextarg, 0xFF))
+    return (int)res;
```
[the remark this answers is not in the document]
orgads: Done

src/tool_getparam.c
```diff
@@ -999,6 +1004,76 @@ static CURLcode set_trace_config(struct GlobalConfig *global,
...
+  if(!strcmp(nextarg, "CE"))
+    return IPTOS_ECN_CE;
+  if(!strcmp(nextarg, "LOWDELAY"))
+    return IPTOS_LOWDELAY;
+  if(!strcmp(nextarg, "THROUGHPUT"))
+    return IPTOS_THROUGHPUT;
+  if(!strcmp(nextarg, "RELIABILITY"))
+    return IPTOS_RELIABILITY;
+  if(!strcmp(nextarg, "MINCOST"))
+    return IPTOS_MINCOST;
```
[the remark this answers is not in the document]
orgads: Done

orgads:
> > So do you suggest using CURLOPT_SOCKOPTFUNCTION there?
> 
> I do. I don't see why libcurl needs another option when you can already accomplish the same thing using an existing option.

The next feature I plan to add is setting VLAN priority, which is another setsockopt (`setsockopt(sockfd, SOL_SOCKET, SO_PRIORITY, (char*)&priority, sizeof(priority))`). How can I support both options? There is only one `SOCKOPTFUNCTION` entry.

bagder:
> There is only one SOCKOPTFUNCTION entry.

So do both in the same callback?

m4/curl-functions.m4
```diff
@@ -56,7 +56,6 @@ curl_includes_arpa_inet="\
     [], [], [$curl_includes_arpa_inet])
 ])
 
-
```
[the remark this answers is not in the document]
orgads: Thanks, done.

src/tool_getparam.c
orgads: Done, I hope you like it now :)

orgads:
> > There is only one SOCKOPTFUNCTION entry.
> 
> So do both in the same callback?

Done. Is this what you meant?

m4/curl-functions.m4
```diff
@@ -3394,6 +3394,60 @@ AC_DEFUN([CURL_CHECK_FUNC_MEMRCHR], [
...
+  tst_compi_setsockopt_sol_ip="unknown"
+  tst_allow_setsockopt_sol_ip="unknown"
+  #
+  AC_MSG_CHECKING([if setsockopt SOL_IP is compilable])
+  AC_COMPILE_IFELSE([
+    AC_LANG_PROGRAM([[
+      $curl_includes_sys_socket
+      #include <netinet/ip.h>
+    ]],[[
+      if(0 != setsockopt(0, SOL_IP, IP_TOS, 0, 0))
```
bagder, line 3415: Why check for `SOL_IP` support for setsockopt when the code is using `IPPROTO_IP` in that argument?

bagder: According to some [posts](https://lists.gnu.org/archive/html/autoconf/2010-03/msg00119.html) I found, maybe the other way around is more portable?

orgads: Done.

src/tool_getparam.c
```diff
@@ -1026,6 +1031,52 @@ static const struct LongShort *single(char letter)
   return singles[letter - ' '];
 }
 
+#ifdef IPTOS_DSCP_AF11
```
orgads, line 1034: This is a bit ugly. Do you prefer to add a configure check (for `IP_TOS`? something else?), check here for `IP_TOS` or just remove these abbreviations altogether?

orgads:
Now we also have Windows 10 support :)

According to [this answer](https://stackoverflow.com/a/75552881/764870), it must be enabled in the registry to work (Set `HKLM\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\DisableUserTOSSetting` to 0 and restart the machine). Should we document it somewhere?

src/tool_getparam.c
orgads: Replaced with `IP_TOS`. Looks like (at least on FreeBSD) `netinet/in.h` has to be included before `netinet/ip.h` anyway.

Let's see if it compiles on all platforms.

orgads: Ok, no more macros. I just used the numeric values inline (they are invariant).

orgads:
The failing test in `quiche` looks like a flaky test, I doubt it's related to my changes.

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,52 @@
...
+  - tcp-nodelay
+Example:
+  - --type-of-service CS5 $URL
+---
+
+# `--type-of-service`
+
+Set IP Type of Service (TOS). (Added in 8.8.0).
+
+The values allowed for \<string\> can be a numeric value or one of the
```
bagder, line 21:
```suggestion
The values allowed for \<string\> can be a numeric value between 1 and 255 or one of the
```

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,52 @@
...
+* AF43
+* EF
+* ECT1
+* ECT0
+* CE
+* LOWCOST
+* LOWDELAY
+* THROUGHPUT
+* RELIABILITY
+* MINCOST
```
bagder, line 52:
```suggestion
* MINCOST

This option is limited to IPv4 connections only. When curl uses IPv6, this option does nothing.
```

docs/options-in-versions
```diff
@@ -257,6 +257,7 @@
 --trace-config                       8.3.0
 --trace-ids                          8.2.0
 --trace-time                         7.14.0
+--type-of-service                    8.8.0
```
bagder, line 260: It is too late for 8.8.0 but it should be fit for 8.9.0.
```suggestion
--type-of-service                    8.9.0
```

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,52 @@
...
+Multi: single
+See-also:
+  - tcp-nodelay
+Example:
+  - --type-of-service CS5 $URL
+---
+
+# `--type-of-service`
+
+Set IP Type of Service (TOS). (Added in 8.8.0).
```
bagder, line 19:
```suggestion
Set IP Type of Service (TOS). (Added in 8.9.0).
```

src/tool_operate.c
```diff
@@ -144,6 +155,29 @@ static bool is_pkcs11_uri(const char *string)
...
+  (void)val;
+  (void)curlfd;
+#ifdef IP_TOS
+  if(val->ip_tos > 0 && val->ip_tos <= 0xff) {
+    int tos = (int)val->ip_tos;
+    if(setsockopt(curlfd, SOL_IP, IP_TOS,
+      (const char *)&tos, sizeof(tos)) < 0) {
+      int error = errno;
+      fprintf(tool_stderr, "IP_TOS %d failed with errno %d: %s;\n",
+              tos, error, strerror(error));
```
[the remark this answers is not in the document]
orgads: Can you suggest a way to check the protocol type here?

docs/cmdline-opts/type-of-service.md
orgads: Done

docs/cmdline-opts/type-of-service.md
orgads: Done

docs/options-in-versions
orgads: Done

docs/cmdline-opts/type-of-service.md
orgads: Done

src/tool_operate.c
```diff
@@ -2197,6 +2231,9 @@ static CURLcode single_transfer(struct GlobalConfig *global,
           my_setopt_str(curl, CURLOPT_ECH, config->ech_config);
 #endif
 
+        my_setopt(curl, CURLOPT_SOCKOPTFUNCTION, sockopt_callback);
+        my_setopt(curl, CURLOPT_SOCKOPTDATA, &config->sockopt);
```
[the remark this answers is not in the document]
orgads: Ok

src/tool_operate.c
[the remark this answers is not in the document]
bagder: Can `getsockname` help?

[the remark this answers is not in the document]
orgads: Actually I found `getsockopt(sockfd, SOL_SOCKET, SO_DOMAIN, &family, &len)`.

But before I change that, isn't it better to show an error on such failure rather than silently ignoring the option?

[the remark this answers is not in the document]
bagder: This goes back to what I asked you from the start: users typically won't know if IPv4 or IPv6 gets used so they can't tell when this option takes effect. Do you think a user wants an error message when it doesn't? I doubt that.

Since v4/v6 may even vary between separate invokes to the same host name it could even result in an error message intermittently every nth invoke which probably will be highly confusing and hardly productive for users.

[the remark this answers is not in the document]
orgads: Done

src/tool_getparam.c
```diff
@@ -1027,6 +1029,49 @@ static const struct LongShort *single(char letter)
   return singles[letter - ' '];
 }
 
+struct TOSEntry {
+  const char *name;
+  unsigned char value;
+};
+
+static const struct TOSEntry tos_entries[]= {
```
rpigott, line 1037: Not that it really matters, but TOS/DSCP constants are also available from `#include <netinet/ip.h>` if you want to use them.

There you should also find the constants IPTOS_DSCP_VA (0x0b, RFC5865 Voice Admit) and IPTOS_DSCP_LE (0x04, RFC8622 Lower Effort) defined, which I don't see represented in this table. The full list is https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml#dscp-registry-1.

src/tool_getparam.c
```diff
@@ -1027,6 +1029,49 @@ static const struct LongShort *single(char letter)
...
+  {"AF43", 0x98},
+  {"CS0", 0x00},
+  {"CS1", 0x20},
+  {"CS2", 0x40},
+  {"CS3", 0x60},
+  {"CS4", 0x80},
+  {"CS5", 0xa0},
+  {"CS6", 0xc0},
+  {"CS7", 0xe0},
+  {"ECT0", 0x02},
```
rpigott, line 1058: Are you sure it even makes sense to allow ECT(0) and ECT(1) values?  Linux masks these bits out for SOCK_STREAM sockets anyway.

https://github.com/torvalds/linux/blob/29c73fc794c83505066ee6db893b2a83ac5fac63/net/ipv4/ip_sockglue.c#L590-L592

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,54 @@
...
+* AF31
+* AF32
+* AF33
+* AF41
+* AF42
+* AF43
+* EF
+* ECT1
+* ECT0
+* CE
```
rpigott, line 47: This value isn't present in the table in the code. I'm not sure it's worth supporting anyway, though.

rpigott (commented):
I'm also interested in this feature, so thanks for writing it up!

src/tool_getparam.c
orgads: Thank you. I initially used the implementation from #7762, but then the tests failed on some platforms that don't have `netinet/ip.h`, so I ended up just hardcoding them.

Thanks for the list, I'll update it.

docs/cmdline-opts/type-of-service.md
orgads: Added now. Thank you.

orgads:
@bagder Is this good to go, now that 8.8 was released?

bagder (commented):
We cannot merge this anyway until the feature window opens on June 1st if everything goes well.

src/tool_operate.c
```diff
@@ -2197,6 +2239,12 @@ static CURLcode single_transfer(struct GlobalConfig *global,
           my_setopt_str(curl, CURLOPT_ECH, config->ech_config);
 #endif
 
+        /* new in 8.9.0 */
+        if(config->ip_tos > 0) {
+          my_setopt(curl, CURLOPT_SOCKOPTFUNCTION, sockopt_callback);
+          my_setopt(curl, CURLOPT_SOCKOPTDATA, config);
+        }
```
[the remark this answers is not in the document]
orgads: Done

src/tool_operate.c
```diff
@@ -144,6 +152,40 @@ static bool is_pkcs11_uri(const char *string)
...
+    return CURL_SOCKOPT_OK;
+  (void)config;
+  (void)curlfd;
+#ifdef IP_TOS
+  if(config->ip_tos > 0 && get_address_family(curlfd) == AF_INET) {
+    int tos = (int)config->ip_tos;
+    if(setsockopt(curlfd, SOL_IP, IP_TOS,
+      (const char *)&tos, sizeof(tos)) < 0) {
+      int error = errno;
+      warnf(config->global, "IP_TOS %d failed with errno %d: %s;\n",
```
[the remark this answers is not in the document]
orgads: Done

src/tool_operate.c
```diff
@@ -2197,6 +2254,14 @@ static CURLcode single_transfer(struct GlobalConfig *global,
           my_setopt_str(curl, CURLOPT_ECH, config->ech_config);
 #endif
 
+        /* new in 8.9.0 */
+#if defined(IP_TOS) || defined(IPV6_TCLASS)
+        if(config->ip_tos > 0) {
+          my_setopt(curl, CURLOPT_SOCKOPTFUNCTION, sockopt_callback);
+          my_setopt(curl, CURLOPT_SOCKOPTDATA, config);
+        }
```
vszakats, line 2262: Would it make sense to display a warning or trace message if a >0 `ip_tos` was requested by the user, but the feature isn't supported?

orgads: Done

orgads:
Is the msys2 with debug flaky? I couldn't find anything in the log that looks related to my changes.

vszakats:
@orgads: Yes, the MSYS2 mingw-w64 ones sometimes **hang** in tests. I'm tracking the problem here: https://github.com/curl/curl/pull/13599#issuecomment-2119372376. Could not yet figure out _which_ test is prone to hanging though.

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,56 @@
...
+Multi: single
+See-also:
+  - tcp-nodelay
+Example:
+  - --type-of-service CS5 $URL
+---
+
+# `--type-of-service`
+
+Set IP Type of Service (TOS). (Added in 8.9.0).
```
bagder, line 19: Since you added  support for IPv6 and its "Traffic Class" field:

1. Should this option really be named like this now?
2. Should the documentation not specify the IPv6 details?

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,56 @@
...
+* ECT0
+* CE
+* LE
+* LOWCOST
+* LOWDELAY
+* THROUGHPUT
+* RELIABILITY
+* MINCOST
+
+This option is limited to IPv4 connections only. When curl uses IPv6, this option does nothing.
```
bagder, line 56: This seems to be wrong now?

docs/cmdline-opts/type-of-service.md
```diff
@@ -0,0 +1,56 @@
...
+* VOICE-ADMIT
+* ECT1
+* ECT0
+* CE
+* LE
+* LOWCOST
+* LOWDELAY
+* THROUGHPUT
+* RELIABILITY
+* MINCOST
```
bagder, line 54: Are all these values also specified/used for IPv6?

docs/cmdline-opts/type-of-service.md
orgads: 1. Do you have a better suggestion? Maybe add another `--traffic-class` which does the same?
2. These fields have the same purpose and structure (6 bits DSCP, 2 bits ECN), only the name and the location within the packet differs.

docs/cmdline-opts/type-of-service.md
orgads: Thanks! Removed the note.

docs/cmdline-opts/type-of-service.md
orgads: Yes, it's the same.

orgads:
Is there a way to retrigger only the failing job?

vszakats:
@orgads: I don't think that is made possible by GitHub. Tried to fix the obscure issue that prevented caching the failing download here: 1d63e331de4ef8ad1e2f228c6815fa6fecec6dbb.

Restarted that job manually — if it fails again, please rebase on master.

orgads:
Passed, thanks!

docs/cmdline-opts/type-of-service.md
bagder: No, this should be a single option I think. Having two different ones doing the same thing does not make anything easier for anyone. It would make more sense to come up with a generic name that works for applying to both cases. Maybe it should even include "IP" in the name to better hint what it is about.

Maybe...

- `--ip-traffic`
- `--ip-type`
- `--ip-service`

?

orgads: I'm not sure any of these suggestions is more descriptive than type-of-service.

TOS used to have a different meaning, but nowadays, it is just like traffic class, divided to 6 bits DSCP and 2 bits ECN.

Maybe we should just have `--ip-dscp` and `--ip-ecn`?

rpigott: Just some helpful info on other tools:

linux iproute2 has `tos` and its alias `dsfield` for adding/showing routes that match on the dsfield. It accepts a named DSCP value or a masked 8 bit TOS value, and the ECN bits are ignored. The option is named the same for both v4 and v6 routing tables. E.g. `ip route add to 1.2.3.4/32 dsfield AF21 counter`

linux nftables has `ip dscp` and `ip6 dscp` payload expresions that match/set the 6 bit ds field in the respective protocol family. It accepts either a named or 6 bit DSCP value. E.g. `nft add rule inet filter input ip dscp af21 meta priority set 1:4`.

`tcpdump -v` displays the tos / tclass byte as "tos" or "class" followed by the 8 bit hex value of the byte, including the ECN field.

`ping -Q` sets the tos / tclass byte just like this patch, and linux copies the entire byte in echo replies.

`--ip-tos` or even just `--tos` seems fine to me. I'm partial to `--dscp` I think. I mentioned it earlier but  linux ignores the ECN field from this sockopt for SOCK_STREAM sockets, and I suppose the quic library will manage the ECN field otherwise, so I expect that setting the the ECN field with this option will never be effective, hence `--dscp` seems reasonable.

orgads: @rpigott Thanks a lot for the detailed overview!

I'm a bit reluctant. My considerations:
* `tos`/`type-of-service`:
  * Pro: Widely used (relatively for this niche feature :)) and should be easily recognized.
  * Con: It's an obsolete name for this field, and doesn't describe its actual use.
* `traffic-class`:
  * Pro: Better description of the field, was chosen over ToS for IPv6, which is more recent.
  * Con: Less known, harder to discover.

I call for a vote.

❤ `--ip-tos`
🚀 `--tos`
🎉 `--ip-type-of-service`
👀 `--ip-traffic-class`

orgads: Aaaaaaand we have a winner!

Changed to `--ip-tos`. Ready for review.

docs/cmdline-opts/ip-tos.md
```diff
@@ -0,0 +1,54 @@
+---
+c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+SPDX-License-Identifier: curl
+Long: ip-tos
+Arg: <string>
+Help: Configure IP Type of Service (TOS) or Traffic Class
```
bagder, line 6:
```suggestion
Help: Set IP Type of Service or Traffic Class
```

We want this phrase to be as short and succinct as possible

orgads: Done

bagder (approved):

orgads:
All passed. I have one last PR, but it depends on this one, so I'll push it after this is merged.

Thanks you all for your constructive review!

vszakats (approved):

src/tool_getparam.c
```diff
@@ -1630,6 +1678,16 @@ ParameterError getparameter(const char *flag, /* f or -long-flag */
     case C_TCP_NODELAY: /* --tcp-nodelay */
       config->tcp_nodelay = toggle;
       break;
+    case C_IP_TOS: { /* --ip-tos */
+      const struct TOSEntry *entry = bsearch(
+        &nextarg, tos_entries, sizeof(tos_entries)/sizeof(*tos_entries),
+        sizeof(*tos_entries), find_tos);
```
vszakats, line 1684: Just curious, is declaring a variable in a `case` block valid C89? I was under the impression that it isn't, and curl also used explicit blocks in some other code in similar cases.

orgads: It's in a new block.

orgads: Declaring *and initializing* a variable in a case that has subsequent cases is considered undefined behavior (in all C/C++ variants), and GCC warns about it.

vszakats: It'd make sense, but the pattern often (?) seen is this:
```c
case <n>:
{
   int var = 1;
}
```
`case` being a block is not 100% self-evident: https://en.wikipedia.org/wiki/Duff%27s_device

Hence the question.

orgads: But there's a brace there. The same style is used all over this file. Examples:
```c
case C_PARALLEL_MAX: {  /* --parallel-max */
case C_LOCAL_PORT: { /* --local-port */
case C_REFERER: { /* --referer */
```

vszakats: I apologize for the noise, indeed there is, I didn't notice!

orgads:
Is anything still missing, or are we good to go?

bagder:
Thanks!

PR #13611, opened 2024-05-12
autotools: Only probe for SGI MIPS compilers on IRIX

danielgustafsson:
MIPSPro and the predecessor compiler which was part of the IDO (IRIS Development Option) were only ever shipped on the SGI IRIX operating system (with MIPSPro on 6.0+ which was released in 1994).  Limit the autoconf check to IRIX when probing for these compilers to save some cycles on other platforms.

If this is accepted then it can be expanded to cover other heavily platform-dependent options as well, but I prefer to start somewhere (with something I know, being a survivor from the IRIX days).


bagder (approved):

danielgustafsson:
Is this a feature, waiting for the next window, or a fix? I don't have strong opinions, I think there is an argument to be made for both.

bagder:
I consider it a bugfix.

PR #13616, opened 2024-05-13
urlapi: add CURLU_NO_GUESS_SCHEME

bagder:
Used for extracting:

- when used asking for a scheme, it will return CURLUE_NO_SCHEME if the stored information was a guess

- when used asking for a URL, the URL is returned without a scheme, like when previously given to the URL parser when it was asked to guess

- as soon as the scheme is set explicitly, it is no longer internally marked as guessed

The idea being:

1. allow a user to figure out if a URL's scheme was set as a result of guessing

2. extract the URL without a guessed scheme

3. this makes it work similar to how we already deal with port numbers

Extend test 1560 to verify.

### DIFF
--- a/docs/libcurl/curl_url_get.md
+++ b/docs/libcurl/curl_url_get.md
@@ -125,6 +125,22 @@ nothing following the hash sign.
 
 (Added in curl 8.8.0)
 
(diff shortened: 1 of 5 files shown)


icing:
`CURLU_CLUELESS`.

PR #13638, opened 2024-05-14
quiche, in CI skip test_02_23*

icing:
For unknown reasons, these tests fail in CI often, but run fine locally. Skip them in CI to avoid unrelated PRs to have failures.


bagder:
Ironically, it failed another test now...

icing:
Adding more information about curl client timing out and now the timeout does not happen. Classic.

bagder (approved):

PR #13644, opened 2024-05-14
ci: set semantic type as CI and include digests as CI operations

cmeister2:
Hopefully removing the last of the "chores" from Renovate.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,25 +3,44 @@
   "extends": [
     "config:best-practices"
   ],
+  "semanticCommitType": "ci",
   "packageRules": [
     {
-      "matchManagers": ["github-actions"],
+      "matchManagers": [
+        "github-actions"
+      ],
       "commitMessagePrefix": "gha: ",
-      "labels": ["CI"]
+      "labels": [
+        "CI"
+      ]
     },
     {
-      "matchUpdateTypes": ["pin", "pinDigest"],
+      "matchUpdateTypes": [
+        "pin",
+        "pinDigest",
+        "digest"
+      ],
       "commitMessagePrefix": "ci: ",
-      "labels": ["CI"]
+      "labels": [
+        "CI"
+      ]
     },
     {
-      "matchManagers": ["regex"],
+      "matchManagers": [
+        "regex"
+      ],
       "commitMessagePrefix": "ci: ",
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -64,4 +83,4 @@
       "registryUrlTemplate": "https://github.com"
     }
   ]
-}
+}
```
vszakats, line 86: Minor nit: A newline would nice here.

bagder: agreed

bagder (approved):
Apart from the little nit, this looks good

renovate.json
cmeister2: Will fix - auto formatter got a bit enthusiastic I think

cmeister2:
Fix pushed, will merge after checks complete.

jay:
usually in a pr you would squash your incremental work together into one commit, unless they are truly separate, and then rebase before landing. you can judge whether that is the case here. your commit will need an explanation and a reference most of the time. references have a separate blank line and then `Closing #xxxx` and/or `Ref: url` etc. please review [push-access-guidelines](https://github.com/curl/curl/wiki/push-access-guidelines) for more detail. it may help to go through a bit of the repo commit history with `gitk --all` + down button to quickly get an idea of what is acceptable

cmeister2:
> usually in a pr you would squash your incremental work together into one commit, unless they are truly separate, and then rebase before landing. you can judge whether that is the case here. your commit will need an explanation and a reference most of the time. references have a separate blank line and then `Closing #xxxx` and/or `Ref: url` etc. please review [push-access-guidelines](https://github.com/curl/curl/wiki/push-access-guidelines) for more detail. it may help to go through a bit of the repo commit history with `gitk --all` + down button to quickly get an idea of what is acceptable

I've rebased the nit comment. I don't think there are any URLs to reference here. If I'm doing pushes wrong I'm happy to let someone with a bit more experience handle the merge here.

cmeister2:
> > If I'm doing pushes wrong
> 
> The `Closes #[number]` line was missing in some previous commits.

Ah ok - I didn't have a corresponding issue created for this work.

bagder:
It should say `Closes`  for the PR and `Fixes` for the issue, if there is one.

PR #13656, opened 2024-05-15
Revert "ci: update nghttp2/nghttp2 to v1.62.0"

bagder:
This reverts commit 14f2c767555b7598d7783ccd9093670b84d28488.


bagder:
I was clearly not paying attention when I merged that one as it breaks the build. Probably because of this => https://github.com/nghttp2/nghttp2/issues/2194

bagder:
According to upstream this happens because the g++ version used does not support C++20 properly, so bumping nghttp2 in the CI requires us to also run a newer C++ compiler.

PR #13660, opened 2024-05-15
reuse: migrate standalone license file to dep5

vszakats:
Follow-up to 73a36021207284ad2b4340ffde34a51b0ba4d47a
Closes #xxxxx


cmeister2:
Ah, thanks @vszakats! The docs seemed to indicate the *.license file was the best way to go for uncommentable files but this seems like a much better solution.

vszakats:
@cmeister2: Thanks for confirming! (I was wondering if there was a special requirement for this particular file.)

bagder (approved):

PR #13661, opened 2024-05-15
pytest, fixes for recent python, add FTP tests

icing:
Fixes:
- in uds tests, abort also silently on os errors
- be conservative on the h3 goaway duration
- detect curl debug build and use in checks
- fix caddy version check for slight difference under linux
- set caddy default path fitting for linux
- fix deprecation warnings in valid time checks

FTP tests:
- add '--with-test-vsftpd=path' to configure
- use vsftpd default path suitable for linux
- add test_30 with plain FTP tests
- add test_31 with --ssl-reqd FTP tests
- add vsftpd to linux GHA for pytest workflows

### DIFF
--- a/.github/scripts/spellcheck.words
+++ b/.github/scripts/spellcheck.words
@@ -906,6 +906,7 @@ vnd
 VRF
 VRFY
 VSE
+vsftpd
 vsprintf
 vt
 vtls
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -383,7 +383,7 @@ jobs:
 
       - if: contains(matrix.build.install_steps, 'pytest')
         run: |
-          sudo apt-get install apache2 apache2-dev libnghttp2-dev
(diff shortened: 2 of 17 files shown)


dfandrich:
Shouldn't the ftp tests go into a new "tests/ftp/" directory? It's unexpected seeing ftp tests in a directory called "http".

icing:
> Shouldn't the ftp tests go into a new "tests/ftp/" directory? It's unexpected seeing ftp tests in a directory called "http".

That would be one idea. Or rename `http` to something else.

bagder:
I think the rustls.c fixes should be put into a separate PR to keep this one limited to the test suite.

icing:
> I think the rustls.c fixes should be put into a separate PR to keep this one limited to the test suite.

Done in #13670.

bagder:
I propose `tests/pytests` as a new name, but done in a separate PR.

Issue #13620, opened 2024-05-13
Disabling USE_UNIX_SOCKETS does not remove the use of AF_UNIX when compiling libcurl.


b5i:
### I did this

I tried to compile Curl using the curl_config.h in attachement 
[curl_config.h.zip](https://github.com/curl/curl/files/15296128/curl_config.h.zip) and with the `BUILDING_LIBCURL` and `HAVE_CONFIG_H` enabled. Changing  `AF_UNIX` to `AF_INET` in `src/socketpair.h` when defining `wakeup_create(p)` does the trick and libcurl builds successfully.

### I expected the following

The curl_config.h file contains `//#define USE_UNIX_SOCKETS 1` because ESP32 with FreeRTOS and the `xtensa-esp32-elf` toolchain don't support them. Therefore, `AF_UNIX` shouldn't be used when compiling libcurl as it would be undefined (error: `/src/socketpair.h:41:42: error: 'AF_UNIX' undeclared (first use in this function); did you mean 'AF_INET'?`). Would changing it to `AF_INET` when USE_UNIX_SOCKETS = 0 be a good workaround?

### curl/libcurl version

curl 8.8.0-DEV

### operating system

Darwin MBP-de-Antoine 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:37 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6031 arm64, compiling with `toolchain-xtensa-esp32@8.4.0+2021r2-patch5/xtensa-esp32-elf` from PlatformIO.

bagder:
Does it actually still work then? Or is your build using the `Curl_socketpair()` "emulation" function?

b5i:
> Does it actually still work then? 

Do you mean if libcurl builds? In that case yes when changing AF_UNIX to AF_INET. 
If you meant whether curl actually works when built that's what I'm testing right now with a simple project to make sure it's not broken.

> Or is your build using the `Curl_socketpair()` "emulation" function?
I guess so as HAVE_SOCKETPAIR is not defined in my curl_config.h.

b5i:
I can confirm that I'm able to do POST and GET requests when compiling with the workaround.

bagder:
Are you up for providing it as a PR?

bagder:
Fixed in #13666

PR #13666, opened 2024-05-15
lib: fixed compilation error when USE_UNIX_SOCKETS is not defined

b5i:
Fix for #13620.

### DIFF
--- a/lib/socketpair.h
+++ b/lib/socketpair.h
@@ -38,7 +38,14 @@
 #define wakeup_write     swrite
 #define wakeup_read      sread
 #define wakeup_close     sclose
-#define wakeup_create(p) Curl_socketpair(AF_UNIX, SOCK_STREAM, 0, p)
+
+#ifdef USE_UNIX_SOCKETS
+#define CURL_SOCKET_FAMILY AF_UNIX
+#else
+#define CURL_SOCKET_FAMILY AF_INET
+#endif
+
+#define wakeup_create(p) Curl_socketpair(CURL_SOCKET_FAMILY, SOCK_STREAM, 0, p)
 
 #endif /* HAVE_PIPE */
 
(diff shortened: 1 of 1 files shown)


lib/socketpair.h
```diff
@@ -38,7 +38,14 @@
 #define wakeup_write     swrite
 #define wakeup_read      sread
 #define wakeup_close     sclose
-#define wakeup_create(p) Curl_socketpair(AF_UNIX, SOCK_STREAM, 0, p)
+
+#ifdef USE_UNIX_SOCKETS
+#define CURL_SOCKET_FAMILY AF_UNIX
+#else
+#define CURL_SOCKET_FAMILY AF_INET
+#endif
```
bagder, line 46: If the native `socketpair()` function is used, it cannot be changed to AF_INET. And if the non-native implementation is used, the family is not used so might as well just be set to zero.

Also, no need to use `CURL_` prefix for a private define.

May I suggest:

```suggestion
#if defined(USE_UNIX_SOCKETS) && defined(HAVE_SOCKETPAIR)
#define SOCKET_FAMILY AF_UNIX
#elif !defined(USE_UNIX_SOCKETS) && !defined(HAVE_SOCKETPAIR)
#define SOCKET_FAMILY 0 /* not used */
#else
#error "unsupported unix domain and socketpair build combo"
#endif
```

lib/socketpair.h
```diff
@@ -38,7 +38,14 @@
...
+
+#define wakeup_create(p) Curl_socketpair(CURL_SOCKET_FAMILY, SOCK_STREAM, 0, p)
```
bagder, line 48:
```suggestion
#define wakeup_create(p) Curl_socketpair(SOCKET_FAMILY, SOCK_STREAM, 0, p)
```

bagder (changes requested):

lib/socketpair.h
b5i: Thanks! I didn't know about that

bagder (approved):

bagder:
Thanks!

Issue #13487, opened 2024-04-26
wolfSSL HTTP/3 peer verification is lacking


bagder:
### I did this

build curl with ngtcp2 + nghttp3 + wolfSSL

Issue a request that should fail the certificate check, like this:

~~~
curl --http3 --resolve example:443:151.101.129.91 https://example
~~~

The problem is likely here:

https://github.com/curl/curl/blob/c8e0cd1de8db9eba00c284ebf5fd5b3cba263e06/lib/vquic/vquic-tls.c#L325-L330

This does not really verify the peer certificate. For TCP connections this works as it is wired into the wolfSSL connect() implementation and gives a special return code on such a fail.

### I expected the following

A request to a server that fails the check should error out.

### curl/libcurl version

git master

### operating system

independent

bagder:
@icing feel free to edit/follow-up with corrections or added details as you see fit

bagder:
This appears to be a wolfSSL problem, would you agree @icing? I see nothing in the [wolfSSL docs](https://www.wolfssl.com/doxygen/group__Setup.html#gab9b75f5fb10ce88f0026c57716858074) that suggests that this would not work for QUIC connections but only for TCP. Or am I missing something?

icing:
I agree that I found no obvious way to do this with the wolfSSL API.

bagder:
Since we believe this problem is in the wolfSSL end, I'm closing this issue.

julek-wolfssl:
Its a curl bug. Fix @ https://github.com/curl/curl/pull/13680.

PR #13680, opened 2024-05-16
Curl_vquic_tls_verify_peer: use correct API for wolfSSL

julek-wolfssl:
wolfSSL_X509_check_host checks the peer name against the alt names and the common name.

Fixes https://github.com/curl/curl/issues/13487. Tested using the same reproducer as in the issue. Correct url passes check.


bagder (approved):

bagder:
Thanks!

Issue #13672, opened 2024-05-16
SSLKEYLOGFILE not working with curl-for-win build


jay:
From discussion https://github.com/curl/curl/discussions/13665 the reporter @glslang is using the official curl for windows build 8.7.1 64-bit and SSLKEYLOGFILE is not working. A blank file is created and nothing is written.
~~~
curl 8.7.1 (x86_64-w64-mingw32) libcurl/8.7.1 LibreSSL/3.9.2 zlib/1.3.1 brotli/1.1.0 zstd/1.5.6 WinIDN libpsl/0.21.5 libssh2/1.11.0 nghttp2/1.62.0 ngtcp2/1.5.0 nghttp3/1.3.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL SSPI threadsafe UnixSockets zstd
~~~

I can reproduce with the same build. curl 8.7.1 with OpenSSL then SSLKEYLOGFILE works so I suspect this is an issue with how we coded it for LibreSSL

botovq:
LibreSSL [doesn't really support][1] the keylog mechanism. The API exists since a number of things don't properly test and guard for it, but it is a noop.

[1]: https://man.openbsd.org/SSL_CTX_set_keylog_callback

glslang:
And I  guess the reason to use LibreSSL on Windows is because it makes for easier builds? Or just a preference in general due to possible security considerations?

jay:
> LibreSSL [doesn't really support](https://man.openbsd.org/SSL_CTX_set_keylog_callback) the keylog mechanism. The API exists since a number of things don't properly test and guard for it, but it is a noop.

Thanks, I missed that on review when we enabled support. I am proposing #13682 to revert to the legacy TLS 1.2 logging for LibreSSL. Is there a way we can extract TLS 1.3 secrets with LibreSSL?

> And I guess the reason to use LibreSSL on Windows is because it makes for easier builds? Or just a preference in general due to possible security considerations?

There were a number of reasons, see curl/curl-for-win#44

/cc @vszakats

jshanab:
[...]

PR #13682, opened 2024-05-17
openssl: revert keylog_callback support for LibreSSL

jay:
- Revert to the legacy TLS 1.2 key logging code for LibreSSL.

Prior to this change if the user specified a filename in the SSLKEYLOGFILE environment variable and was using LibreSSL 3.5.0+ then an empty file would be created and no keys would be logged.

This is effectively a revert of e43474b4 which changed openssl.c to use SSL_CTX_set_keylog_callback for LibreSSL 3.5.0+. Unfortunately LibreSSL added that function only as a stub that doesn't actually do anything.

Reported-by: Gonçalo Carvalho

Fixes https://github.com/curl/curl/issues/13672
Closes #xxxx

---

This means we don't have TLS 1.3 key logging for LibreSSL? The legacy TLS 1.2 logging records only CLIENT_RANDOM, but I think for TLS 1.3 we need all of the secrets to decrypt that SSL_CTX_set_keylog_callback would record (ie SERVER_HANDSHAKE_TRAFFIC_SECRET, EXPORTER_SECRET, SERVER_TRAFFIC_SECRET_0, CLIENT_HANDSHAKE_TRAFFIC_SECRET, CLIENT_TRAFFIC_SECRET_0)

### DIFF
--- a/docs/cmdline-opts/_ENVIRONMENT.md
+++ b/docs/cmdline-opts/_ENVIRONMENT.md
@@ -101,8 +101,8 @@ if Schannel is used as the TLS backend.
 If you set this environment variable to a filename, curl stores TLS secrets
 from its connections in that file when invoked to enable you to analyze the
 TLS traffic in real time using network analyzing tools such as Wireshark. This
-works with the following TLS backends: OpenSSL, libressl, BoringSSL, GnuTLS
-and wolfSSL.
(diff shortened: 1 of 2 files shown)


bagder:
Maybe `docs/cmdline-opts/_ENVIRONMENT.md` should be clarified about the libressl state as well?

vszakats (approved):

bagder (approved):

PR #13689, opened 2024-05-17
tests: test DEBUGBUILD, CURLDEBUG, shared/static combinations

vszakats:

PR #13694, opened 2024-05-17
build: untangle `UNITTESTS` and `DEBUGBUILD` macros

vszakats:
- fix `DEBUGBUILD` guards that should be `UNITTESTS`, in libcurl
  code used by unit tests.
- fix guards for libcurl functions used in unit tests only.
- sync `UNITTEST` attribute between declarations and definitions.
- drop `DEBUGBUILD` guard from test `unit2600`.
- fix guards for libcurl HSTS code used by both a unit test (`unit1660`)
  and `test0446`.
- update an existing AppVeyor CI job to test the issues fixed.

This fixes building tests with `CURLDEBUG` enabled but `DEBUGBUILD`
disabled. This can happen when building tests with CMake with
`ENABLE_DEBUG=ON` in Release config, or with `ENABLE_CURLDEBUG=ON`
and _without_ `ENABLE_DEBUG=ON`. Possibly also with autotools
when using `--enable-curldebug` without `--enable-debug`.

Test results:
- before: https://ci.appveyor.com/project/curlorg/curl/builds/49835609
- after: https://ci.appveyor.com/project/curlorg/curl/builds/49839255
  (the two failures are unrelated, subject to PR #13705)

Ref: #13592 (issue discovery)
Ref: #13689 (CI testing this PR with `DEBUGBUILD`/`CURLDEBUG` combinations)
Closes #13694

### DIFF
--- a/lib/hsts.c
+++ b/lib/hsts.c
@@ -54,7 +54,7 @@
 #define MAX_HSTS_DATELENSTR "64"
 #define UNLIMITED "unlimited"
 
-#ifdef DEBUGBUILD
+#ifdef UNITTESTS
 /* to play well with debug builds, we can *set* a fixed time this will
    return */
 time_t deltatime; /* allow for "adjustments" for unit test purposes */
--- a/lib/hsts.c
+++ b/lib/hsts.c
@@ -54,7 +54,7 @@
 #define MAX_HSTS_DATELENSTR "64"
 #define UNLIMITED "unlimited"
 
-#ifdef UNITTESTS
+#if defined(DEBUGBUILD) || defined(UNITTESTS)
 /* to play well with debug builds, we can *set* a fixed time this will
    return */
 time_t deltatime; /* allow for "adjustments" for unit test purposes */
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -39,6 +39,7 @@ if [ "${BUILD_SYSTEM}" = 'CMake' ]; then
   options=''
   [[ "${TARGET:-}" = *'ARM64'* ]] && SKIP_RUN='ARM64 architecture'
   [ "${OPENSSL}" = 'ON' ] && options+=" -DOPENSSL_ROOT_DIR=${openssl_root_win}"
+  [ -n "${CURLDEBUG:-}" ] && options+=" -DENABLE_CURLDEBUG=${CURLDEBUG}"
   [ "${PRJ_CFG}" = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
   [ "${PRJ_CFG}" = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
   [[ "${PRJ_GEN}" = *'Visual Studio'* ]] && options+=' -DCMAKE_VS_GLOBALS=TrackFileAccess=false'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -80,6 +80,9 @@ environment:
       SCHANNEL: 'ON'
       ENABLE_UNICODE: 'OFF'
       HTTP_ONLY: 'OFF'
+      DEBUG: 'OFF'
+      CURLDEBUG: 'ON'
+      TFLAGS: 'skiprun'
     - job_name: 'CMake, VS2010, Debug, x64, Schannel, Static, Build-only'
       APPVEYOR_BUILD_WORKER_IMAGE: 'Visual Studio 2015'
       BUILD_SYSTEM: CMake
--- a/lib/cfilters.c
+++ b/lib/cfilters.c
@@ -45,7 +45,7 @@
 #define ARRAYSIZE(A) (sizeof(A)/sizeof((A)[0]))
 #endif
 
-#ifdef DEBUGBUILD
+#ifdef UNITTESTS
 /* used by unit2600.c */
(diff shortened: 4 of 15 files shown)


lib/hsts.c
```diff
@@ -54,7 +54,7 @@
 #define MAX_HSTS_DATELENSTR "64"
 #define UNLIMITED "unlimited"
 
-#ifdef DEBUGBUILD
+#if defined(DEBUGBUILD) || defined(UNITTESTS)
```
bagder, line 57: Why is this needed?

The unit tests assume that `DEBUGBUILD` is defined as a prereq for running. This change seems to say that you can build with `UNITTESTS` without `DEBUGBUILD` ?

bagder:
> This fixes building tests with CURLDEBUG enabled but DEBUGBUILD disabled

I wish we could drop the distinction completely and just use one of them. I can't remember the difference, I don't think anyone uses them separately. Am I wrong?

lib/hsts.c
vszakats: Unit tests (and any other test) are _building_ fine without either `DEBUGBUILD` or `CURLDEBUG`. (That is after fixing the places where `DEBUGBUILD` was used to enable functions for unit tests.)

This is an outlier case where the `deltatime` variable of a debug feature (`CURL_TIME`) is actually required for building a unit test.

vszakats:
> > This fixes building tests with CURLDEBUG enabled but DEBUGBUILD disabled
> 
> I wish we could drop the distinction completely and just use one of them. I can't remember the difference, I don't think anyone uses them separately. Am I wrong?

They are not the same thing, but their names don't make it trivial to figure out which does what exactly:
- `DEBUGBUILD`: curl debug features, such as extra output, and envs that allow controlling internals.
- `CURLDEBUG`: memory tracking feature, aka 'TrackMemory'. It's an intrusive option that redefines a lot of internal functions, somewhat fragile.

In CMake, `DEBUGBUILD` always enabled `CURLDEBUG`, but the latter could also be enabled separately. Autotools looks somewhat similar. Both offer separate options to enable them. `Makefile.mk` also has independent switches. winbuild enables `DEBUGBUILD`, but never `CURLDEBUG`. It's probably rare indeed that someone is concsiously using them separately though.

Yet, they are separate and possible to use separately, with quite a bit of breakage around the edges.

The goal of this patch isn't really to untangle these two from each other, but to fix to use the `UNITTESTS` macro to enable internal functions and features required for building _unit tests_. This is mandatory for a successful build, unlike `DEBUGBUILD` and `CURLDEBUG` which are optional for building tests. (`DEBUGBUILD` seems required to _run_ them, but I'd have to test this.)

Merging `DEBUGBUILD` and `CURLDEBUG` is an option, but something for separate PR(s).

This PR is an attempt t fix all existing build combinations.

vszakats:
There seem to be a few suspicious cases, where `DEBUGBUILD` and `CURLDEBUG` are mixed up in the code, too.
E.g. `src/tool_operate.c` guards some debug features with `CURLDEBUG`. There is also one case in `src/tool_getparam.c` + `src/tool_cfgable.h`. On a quick glance, in `lib` almost all `CURLDEBUG` should be `DEBUGBUILD`.

I think we might consider renaming `CURLDEBUG` to `CURL_TRACK_MEMORY` or something to make its purpose clear and avoid future mis-uses.

With the list of related open PRs in the air, I'd prefer merging these first, then continue tackling more. Two PRs are blocked by this one: #13705 and #13698.

jay:
> I wish we could drop the distinction completely and just use one of them. I can't remember the difference, I don't think anyone uses them separately. Am I wrong?

As Viktor mentions above, yes they are separate because CURLDEBUG is (supposed to be) the guard for the trackmemory feature. Typically I run debug builds of curl with DEBUGBUILD but not CURLDEBUG.

> I think we might consider renaming CURLDEBUG to CURL_TRACK_MEMORY / CURL_DEBUG_MEMORY or something to make its purpose clear and avoid future mis-uses. For DEBUGBUILD, CURL_DEBUG (or anything starting with CURL_ might also work better.

I'm with you on the first part (maybe something shorter?) but not the second part. Renaming DEBUGBUILD to CURL_DEBUG is confusing to me since we've been using CURLDEBUG for memory tracking.

vszakats:
> I'm with you on the first part (maybe something shorter?) but not the second part. Renaming DEBUGBUILD to CURL_DEBUG is confusing to me since we've been using CURLDEBUG for memory tracking.

Indeed, I agree.

_If_ we want to rename them (or one of them), we can brainstorm for new names that are short, tell the purpose, and without confusion. Starting with `CURL_` would be a plus for namespacing I think. (`DEBUGBUILD` might be something used by a dependency, though if it didn't happen for 2 decades, it's not likely to happen now.)

Anyway, after tidying this up, I think it will be a little bit harder to make a mistake by copy-pasting existing code or looking up existing uses. `CURLDEBUG` is used in just a few specific places after these updates.

As for `DEBUGBUILD` and `UNITTESTS`, they remain easy to confuse. The former isn't required to build tests, the latter is.
It makes the purpose of these debug features easier to grasp, but it's debatable if this is a strong enough purpose, esp. if new code continues to get them mixed up. I'm 100% yet, but merging `UNITTESTS` into `DEBUGBUILD` _might_ allow dropping the `curlu` library, by including all these in the debug libcurl lib, which could then be used by unit tests.

jay:
I can see UNITTESTS being dependent on DEBUGBUILD but not the other way around

PR #13704, opened 2024-05-19
ssl: Avoid explicitly setting verifypeer to 1

orgads:
It is the default, just like verifyhost.

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -1731,14 +1731,11 @@ static CURLcode single_transfer(struct GlobalConfig *global,
           my_setopt_str(curl, CURLOPT_SSLKEYTYPE, config->key_type);
           my_setopt_str(curl, CURLOPT_PROXY_SSLKEYTYPE,
                         config->proxy_key_type);
+
+          /* libcurl default is strict verifyhost -> 1L, verifypeer -> 1L */
           if(config->insecure_ok) {
             my_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0L);
             my_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0L);
-          }
-          else {
-            my_setopt(curl, CURLOPT_SSL_VERIFYPEER, 1L);
-            /* libcurl default is strict verifyhost -> 2L   */
-            /* my_setopt(curl, CURLOPT_SSL_VERIFYHOST, 2L); */
           }
 
           if(config->doh_insecure_ok) {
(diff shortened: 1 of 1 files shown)


orgads:
I was wondering also about the lengthy documentation of the differences between 1 and 2 for `CURLOPT_SSL_VERIFYHOST`. It looks like the effective default is `TRUE` (I couldn't find 2 anywhere), and there is no difference for current versions between 1 and 2, so maybe the documentation should not state the difference in the main description? It can be stated later in a comment, that previous versions had a difference between 1 and 2 and describe this difference.

Same goes for `CURLOPT_DOH_SSL_VERIFYHOST`.

src/tool_operate.c
```diff
@@ -1736,9 +1736,9 @@ static CURLcode single_transfer(struct GlobalConfig *global,
             my_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0L);
           }
           else {
-            my_setopt(curl, CURLOPT_SSL_VERIFYPEER, 1L);
-            /* libcurl default is strict verifyhost -> 2L   */
-            /* my_setopt(curl, CURLOPT_SSL_VERIFYHOST, 2L); */
+            /* libcurl default is strict verifyhost -> 1L, verifypeer -> 1L */
+            /* my_setopt(curl, CURLOPT_SSL_VERIFYPEER, 1L); */
+            /* my_setopt(curl, CURLOPT_SSL_VERIFYHOST, 1L); */
```
vszakats, line 1741: Speaking of the most important security settings for TLS to actually work, wouldn't it better to keep them here for clarity and making sure we don't rely on defaults set elsewhere in the code (often residing in a separate shared libcurl library)?

The the DOH/proxy insecure checks also don't enable security explicitly, or do it only for `VERIFYPEER`. IMO correcting those to set both to `1L` would be a better direction for robustness.

orgads: [lib]curl was "secure by default" since ever, and this is extremely unlikely to change in future versions, and `VERIFYHOST` is already excluded. Digging through history, I found 0417d34533c187db1ac0fadd6765573defb69d00 by @bagder, which is quite similar to this change.

vszakats: It was done to make `--libcurl` output easier to follow, in 2010. I'd vote for being explicit when it comes to security, but if this is just me, the `my_setopt(curl, CURLOPT_PROXY_SSL_VERIFYPEER, 1L);` line shouldn't also go?

orgads: It's commented out, similar to `VERIFYHOST` (for `VERIFYHOST` I only changed the default to 1, see my previous comment).

I don't have a strong opinion on this, will let @bagder decide.

bagder:
I can agree to this change. But I would also like:
1. the code removed, not commented. The entire else clause is unused then.
2. the same logic applied to the proxy and doh equivalent options

orgads:
> I can agree to this change. But I would also like:
> 
> 1. the code removed, not commented. The entire else clause is unused then.
> 2. the same logic applied to the proxy and doh equivalent options

Done

bagder (approved):

bagder:
Thanks!

PR #13709, opened 2024-05-19
ci: modify two Windows jobs to improve coverage

vszakats:
Based on failing then fixed jobs in #13689


vszakats:
The two updated jobs should fail without applying the fixes pending in separate PRs. Will continue with applying them here, then may move these updates to the relevant PRs:
- #13694
- #13705

vszakats:
Commits migrated to the PRs they belong to.

PR #13710, opened 2024-05-19
url: remove duplicate call to Curl_conncache_remove_conn when pruning

nmoinvaz:
The connection is already removed from the connection cache in `extract_if_dead`.

### DIFF
--- a/lib/url.c
+++ b/lib/url.c
@@ -839,11 +839,6 @@ static bool extract_if_dead(struct connectdata *conn,
   return FALSE;
 }
 
-struct prunedead {
-  struct Curl_easy *data;
-  struct connectdata *extracted;
-};
-
 /*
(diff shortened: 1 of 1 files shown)


lib/url.c
```diff
@@ -884,9 +885,6 @@ static void prune_dead_connections(struct Curl_easy *data)
                                  call_extract_if_dead)) {
       /* unlocked */
 
-      /* remove connection from cache */
-      Curl_conncache_remove_conn(data, prune.extracted, TRUE);
-
```
bagder, line 889: How about leaving a comment saying that `extract_if_dead` already removed the conn from the cache? It saves the reader from having to look it up or think about that case.

nmoinvaz: I have added a comment and did some more cleanup. Let me know if any of that is unwanted.

bagder (approved):

bagder:
Thanks!

PR #13717, opened 2024-05-20
cmake: set `ENABLE_DEBUG=ON` by default for `Debug` builds

vszakats:
When `CMAKE_BUILD_TYPE` is set to `Debug`, enable curl debug features
by default. We do this to more closely match how autotools builds
work, where `--enable-debug` enables both compiler debug feature and
curl ones.

You can override this by explicitly setting `ENABLE_DEBUG=OFF` and
disable curl debug feature while leaving compiler debug enabled.

Ref: https://github.com/curl/curl/pull/13592#issuecomment-2119876734
Closes #13717

---

/cc @jay

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,9 @@ cmake_dependent_option(ENABLE_THREADED_RESOLVER "Set to ON to enable threaded DN
         ON "NOT ENABLE_ARES"
         OFF)
 
-option(ENABLE_DEBUG "Set to ON to enable curl debug features" OFF)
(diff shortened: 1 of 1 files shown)


vszakats:
Hm, looking at it, and how this may change existing configs ...this seems to make it more complex to figure out which configuration ends up enabling what.

jay:
I thought about it some more and I think we should not do it, see my reply in https://github.com/curl/curl/pull/13592#issuecomment-2121884580

vszakats:
Thanks @jay, agreed!

Closing without merge.

PR #13718, opened 2024-05-20
build: untangle `CURLDEBUG` and `DEBUGBUILD` macros

vszakats:
`CURLDEBUG` is meant to enable memory tracking, but in a bunch of cases,
it was protecting debug features that were supposed to be guarded with
`DEBUGBUILD`.

Replace these uses with `DEBUGBUILD`.

This leaves `CURLDEBUG` uses solely for its intended  purpose: to enable
the memory tracking debug feature.

Also:
- autotools: rely on `DEBUGBUILD` to enable `checksrc`.
  Instead of `CURLDEBUG`, which worked in most cases because debug
  builds enable `CURLDEBUG` by default, but it's not accurate.
- include `lib/easyif.h` instead of keeping a copy of a declaration.
- add CI test jobs for the build issues discovered.

Ref: https://github.com/curl/curl/pull/13694#issuecomment-2120311894
Closes #13718

---

This PR builds upon #13694 and also includes #13592, #13698, #13705
for testing everything related in one.

TODO:
- [x] strip the other PRs and rebase once they were merged.
- [x] add job to AppVeyor: CMake, VS2022, Debug, x64, Schannel, Shared, DEBUGBUILD=ON, CURLDEBUG=OFF, build-only (w/o tests) [this failed after applying this PR]
- [x] fix exporting debug-only `curl_easy_perform_ev()` from libcurl. This uses a workaround hard-wired to `ENABLE_CURLDEBUG` in CMake, which needs to be updated. (Couldn't spot the same trick in autotools.) Job fallout: https://ci.appveyor.com/project/curlorg/curl/builds/49852346. Needs further fixing for autotools / non-Windows.
- [x] appveyor: split `TESTING` to `TESTBUILD` and `TESTRUN`. `TESTBUILD` defaulting to `TESTRUN`.
- [x] maybe move some patches into separate PRs.
- [x] rebase on #13759.


vszakats:
There is also a mixup in autotools, which uses `CURLDEBUG` to decide whether to run `checksrc`.

PR #13719, opened 2024-05-20
Support bind to interface and IP

orgads:
Support bind to interface and IP

Introduce new notation for CURLOPT_INTERFACE / --interface:
`ifhost!<interface>!<host>`

Binding to an interface doesn't set the address, and an interface can have multiple addresses.

When binding to an address (without interface), the kernel is free to choose the route, and it can route through any device that can access the target address, not necessarily the one with the chosen address.

Moreover, it is possible for different interfaces to have the same IP address, on which case we need to provide a way to be more specific.

Factor out the parsing part of interface option, and add unit tests.

### DIFF
--- a/docs/libcurl/opts/CURLOPT_INTERFACE.md
+++ b/docs/libcurl/opts/CURLOPT_INTERFACE.md
@@ -28,15 +28,16 @@ CURLcode curl_easy_setopt(CURL *handle, CURLOPT_INTERFACE, char *interface);
 
 Pass a char pointer as parameter. This sets the *interface* name to use as
 outgoing network interface. The name can be an interface name, an IP address,
-or a hostname.
+or a hostname. If you prefer one of these, you can use the following special
+prefixes:
 
-If the parameter starts with "if!" then it is treated only as an interface
-name. If the parameter starts with "host!" it is treated as either an IP
-address or a hostname.
+* `if!<name>` - Interface name
+* `host!<name>` - IP address or hostname
+* `ifhost!<interface>!<host>` - Interface name and IP address or hostname
 
-If "if!" is specified but the parameter does not match an existing interface,
-*CURLE_INTERFACE_FAILED* is returned from the libcurl function used to perform
-the transfer.
+If `if!` or `ifhost!` is specified but the parameter does not match an existing
+interface, *CURLE_INTERFACE_FAILED* is returned from the libcurl function used
+to perform the transfer.
 
 libcurl does not support using network interface names for this option on
 Windows.
@@ -74,7 +75,9 @@ int main(void)
 
 # AVAILABILITY
 
-The "if!" and "host!" syntax was added in 7.24.0.
+The `if!` and `host!` syntax was added in 7.24.0.
+
+The `ifhost!` syntax was added in 8.9.0.
 
 # RETURN VALUE
 
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -73,6 +73,7 @@
 #include "multihandle.h"
 #include "rand.h"
 #include "share.h"
+#include "strdup.h"
 #include "version_win32.h"
 
 /* The last 3 #include files should be in this order */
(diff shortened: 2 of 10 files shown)


orgads:
Can you suggest a way to add unit tests, at least for the parsing part? I'd love to use real callbacks for the actual bind to interface (setsockopt) and bind to ip (bind) parts, but that would be cumbersome, and require additional exported symbols. Are there tests for similar features?

bagder:
Please provide a description when you make a pull request so that we can get a better understanding what is proposed without having to read the patch.

orgads:
> Please provide a description when you make a pull request so that we can get a better understanding what is proposed without having to read the patch.

Sure. Please just let me know which approach you prefer from the ones discussed in #13688 and then I'll prepare a single commit with a proper commit message and update the PR.

docs/libcurl/opts/CURLOPT_INTERFACE.md
```diff
@@ -27,16 +27,17 @@ CURLcode curl_easy_setopt(CURL *handle, CURLOPT_INTERFACE, char *interface);
 # DESCRIPTION
 
 Pass a char pointer as parameter. This sets the *interface* name to use as
-outgoing network interface. The name can be an interface name, an IP address,
-or a hostname.
+outgoing network interface. By default, the name can be an interface name,
```
bagder, line 30: No need to add "by default".

```suggestion
outgoing network interface. The name can be an interface name,
```

lib/cf-socket.c
```diff
@@ -434,23 +434,37 @@ static CURLcode bindlocal(struct Curl_easy *data, struct connectdata *conn,
...
       dev += strlen(host_prefix);
-      is_host = TRUE;
+      strcpy(host_input, dev);
+    }
+    else if(strncmp(if_host_prefix, dev, strlen(if_host_prefix)) == 0) {
+      const char *host_part;
+      dev += strlen(if_host_prefix);
+      host_part = strchr(dev, '!');
+      if(host_part) {
+        strncpy(iface, dev, host_part - dev);
```
bagder, line 458: There is no boundary check for this?

lib/cf-socket.c
```diff
@@ -434,23 +434,37 @@ static CURLcode bindlocal(struct Curl_easy *data, struct connectdata *conn,
...
-    bool is_host = FALSE;
+    if2ip_result_t if2ip_result = IF2IP_NOT_FOUND;
     static const char *if_prefix = "if!";
     static const char *host_prefix = "host!";
+    static const char *if_host_prefix = "ifhost!";
 
     if(strncmp(if_prefix, dev, strlen(if_prefix)) == 0) {
       dev += strlen(if_prefix);
-      is_interface = TRUE;
+      strcpy(iface, dev);
```
bagder, line 447: Boundary check?

lib/cf-socket.c
```diff
@@ -434,23 +434,37 @@ static CURLcode bindlocal(struct Curl_easy *data, struct connectdata *conn,
...
     }
     else if(strncmp(host_prefix, dev, strlen(host_prefix)) == 0) {
       dev += strlen(host_prefix);
-      is_host = TRUE;
+      strcpy(host_input, dev);
```
bagder, line 451: Boundary check?

lib/cf-socket.c
```diff
@@ -490,7 +513,7 @@ static CURLcode bindlocal(struct Curl_easy *data, struct connectdata *conn,
           /* Signal the caller to try another address family if available */
           return CURLE_UNSUPPORTED_PROTOCOL;
         case IF2IP_FOUND:
-          is_interface = TRUE;
+          strcpy(iface, dev);
```
bagder, line 516: boundary?

bagder:
Maybe use dynbufs instead of those local buffers to make it easier to manage without strcpys ?

lib/cf-socket.c
```diff
@@ -434,23 +434,37 @@ static CURLcode bindlocal(struct Curl_easy *data, struct connectdata *conn,
 
   if(dev && (strlen(dev)<255) ) {
```
orgads, line 435: The boundary check is done here. Isn't it enough?

docs/libcurl/opts/CURLOPT_INTERFACE.md
orgads: Done.

lib/cf-socket.c
orgads: Done

lib/cf-socket.c
orgads: Done

lib/cf-socket.c
orgads: Done

lib/cf-socket.c
orgads: Done

orgads:
> Please provide a description when you make a pull request so that we can get a better understanding what is proposed without having to read the patch.

Done

orgads:
@bagder any comments on this?

bagder:
Thanks!

Issue #13692, opened 2024-05-17
WebSocket tests 2301, 2302 flaky on Windows (mingw-w64)


vszakats:
For the first time in CI I enabled WebSockets in a native Windows job with tests,
where ~~2305 fails consistently~~, 2301 and 2302 failed once each (so far):
- ~~https://github.com/curl/curl/actions/runs/9133449443/job/25116951463#step:14:14502
  2305: https://github.com/curl/curl/actions/runs/9133449443/job/25116951463#step:14:11075~~
- https://github.com/curl/curl/actions/runs/9133227792/job/25116241601#step:14:18246
  2301: https://github.com/curl/curl/actions/runs/9133227792/job/25116241601#step:14:9886
  ~~2305: https://github.com/curl/curl/actions/runs/9133227792/job/25116241601#step:14:10002~~
- ~~https://github.com/curl/curl/actions/runs/9134361774/job/25119734300#step:14:18616
  2305: https://github.com/curl/curl/actions/runs/9134361774/job/25119734300#step:14:9896~~
- https://github.com/curl/curl/actions/runs/9139155361/job/25131144383?pr=13689
  2302: https://github.com/curl/curl/actions/runs/9139155361/job/25131144383?pr=13689#step:14:9892

Ref: https://github.com/curl/curl/issues/13599#issuecomment-2118365489

curl/libcurl version: master
operating system: Windows

Ref: https://github.com/curl/curl/blob/36fd2dd6ee874726c628e67fcf6415a2e52bfe29/.github/workflows/windows.yml#L147

edit: test 2305 fixed via dbd626ab825a28c02dd5b38cd4e5b09985ba9bad #13724.

vszakats:
This seems to an expected LF vs generated CRLF issue for 2305. For 2301 it's also that, but there may be something else at play because it doesn't happen always.

vszakats:
test 2305 fixed in dbd626ab825a28c02dd5b38cd4e5b09985ba9bad #13724.

vszakats:
This probably goes under the umbrella of a more generic problem where TFTP, WebSockets, MQTT and FTP are all kind of flaky in Windows CI machines. For now this has been mitigated by ignoring the results of most of these tests (except FTP).

PR #13724, opened 2024-05-20
tests: fix TFTP test 2305 on Windows

vszakats:
Ref: #13692
Closes #13724


vszakats:
Ops, made a mistake of calling this TFTP instead of WebSocket in the merged commit. Apologies.

PR #13737, opened 2024-05-21
cmake: fix building with both md4 and md5 in unity mode

vszakats:
Macro and static function names were colliding between
`lib/md4.c` and
`lib/md5.c`.

Fix it by namespacing these symbols.

Seen with a basic macOS build using these options:
`-DCMAKE_UNITY_BUILD=ON -DCURL_USE_SECTRANSP=ON`

Closes #13737

### DIFF
--- a/lib/md4.c
+++ b/lib/md4.c
@@ -277,14 +277,14 @@ static void MD4_Final(unsigned char *result, MD4_CTX *ctx);
  * F and G are optimized compared to their RFC 1320 definitions, with the
  * optimization for F borrowed from Colin Plumb's MD5 implementation.
  */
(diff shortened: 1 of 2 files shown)


pps83:
Finally! I think `#define I` is the piece of code I hate to most due to the buggy visual assist intellisense that keeps suggesting/replacing `i` with `I` because of that macro somewhere deep down in curl. `define S` should be fixed next :)

vszakats:
Well spotted, and there is also an `R` :)

Would you like to make a PR of this?

pps83:
> Would you like to make a PR of this?

https://github.com/curl/curl/pull/13751

PR #13738, opened 2024-05-21
Avoid using GetAddrInfoExW when impersonation is used

### LINKED ISSUE #13612: Windows DNS resolution fails when impersonation is used
### I did this

Our windows service calls `ImpersonateLoggedOnUser` before any curl calls so that the proxy calls can be authenticated using the logged on user's token and not from the system. 

DNS resolution fails when this is called because of the change in commit a6bbc87f9e9ffb46a1801dfb983e7534825ed56b to use `GetAddrInfoExW`. `GetAddrInfoExW()` keeps returning `WSATRY_AGAIN` and never succeeds even after many retries.

### I expected the following

DNS resolution works even with impersonation. 

### curl/libcurl version

libcurl 8.7.1

### operating system

Windows

pps83:
Multiple reports suggest that GetAddrInfoExW fails when impersonation is used. This PR checks if thread is impersonating and avoids using GetAddrInfoExW api.

fixes #13612

### DIFF
--- a/lib/system_win32.c
+++ b/lib/system_win32.c
@@ -267,4 +267,14 @@ HMODULE Curl_load_library(LPCTSTR filename)
 #endif
 }
 
+bool Curl_win32_impersonating(void)
+{
+  HANDLE token = NULL;
+  if(OpenThreadToken(GetCurrentThread(), TOKEN_QUERY, TRUE, &token)) {
+    CloseHandle(token);
+    return TRUE;
+  }
+  return FALSE;
+}
+
 #endif /* _WIN32 */
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -638,7 +638,8 @@ static bool init_resolve_thread(struct Curl_easy *data,
 
 #ifdef _WIN32
   if(Curl_isWindows8OrGreater && Curl_FreeAddrInfoExW &&
-     Curl_GetAddrInfoExCancel && Curl_GetAddrInfoExW) {
+     Curl_GetAddrInfoExCancel && Curl_GetAddrInfoExW &&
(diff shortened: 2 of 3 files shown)


jay:
This is almost the same as what @edmcln suggested so I will credit him as a co author unless you have any objection. Also he says he's still looking into it so I'd like to wait and see if he finds anything else before we go with this.

lib/system_win32.c
```diff
@@ -267,4 +267,17 @@ HMODULE Curl_load_library(LPCTSTR filename)
 #endif
 }
 
+bool Curl_win32_impersonating(void)
+{
+  bool ret = FALSE;
+#ifndef CURL_WINDOWS_APP
```
pps83, line 273: not sure if this code should be disabled for win app. Perhaps, I should enable it?

lib/system_win32.c
```diff
@@ -267,4 +267,17 @@ HMODULE Curl_load_library(LPCTSTR filename)
...
+  HANDLE token = NULL;
+  if(OpenThreadToken(GetCurrentThread(), TOKEN_QUERY, TRUE, &token)) {
+    CloseHandle(token);
+    ret = TRUE;
+  }
+#endif
```
pps83, line 279: Perhaps, this code should also be enabled for winapp:

```suggestion
  HANDLE token = NULL;
  if(OpenThreadToken(GetCurrentThread(), TOKEN_QUERY, TRUE, &token)) {
    CloseHandle(token);
    ret = TRUE;
  }
```

pps83:
@jay @bagder looks like the fix is [good](https://github.com/curl/curl/issues/13612#issuecomment-2127486802). Should be ok to merge

bagder:
Thanks!

PR #13739, opened 2024-05-22
winbuild: fix PE version info debug flag

jay:
- Only set PE file flag VS_FF_DEBUG if curl.exe and libcurl.dll were built with winbuild option DEBUG=yes which builds with debug info.

VS_FF_DEBUG is a PE flag (Portable Executable file flag - dll, exe, etc) that indicates the file contains or was built with debug info.

Prior to this change when winbuild was used to build curl, curl.exe and libcurl.dll always had VS_FF_DEBUG set, regardless of build option DEBUG=yes/no, due to some bad logic.

Closes #xxxx

---

The bad logic defined DEBUGBUILD=0 for resource file compilations that did not contain debug information. Those resource files then guarded the VS_FF_DEBUG flag with defined(DEBUGBUILD) which of course is defined even when DEBUGBUILD is 0. Even if that was fixed it still wouldn't make sense because winbuild doesn't define DEBUGBUILD for building curl or libcurl --only for building their respective resource files-- so I moved to defining and checking _DEBUG instead for resource file compilations.

caught while reviewing #13730

### DIFF
--- a/winbuild/MakefileBuild.vc
+++ b/winbuild/MakefileBuild.vc
@@ -428,13 +428,13 @@ CFGSET = true
 # CURL_XX macros are for the curl.exe command
 
 !IF "$(DEBUG)"=="yes"
-RC_FLAGS = /dDEBUGBUILD=1 /Fo $@ $(LIBCURL_SRC_DIR)\libcurl.rc
+RC_FLAGS = /d_DEBUG /Fo $@ $(LIBCURL_SRC_DIR)\libcurl.rc
 CURL_CC       = $(CC_DEBUG) $(RTLIB_DEBUG)
-CURL_RC_FLAGS = $(CURL_RC_FLAGS) /i../include /dDEBUGBUILD=1 /Fo $@ $(CURL_SRC_DIR)\curl.rc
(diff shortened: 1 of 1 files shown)


vszakats (approved):

vszakats:
Completely unrelated, just noticed while looking that `MakefileBuild.vc` is setting `WIN32` in a few places. Do you think that's stil necessary now that the codebase uses `_WIN32`?

jay:
> Completely unrelated – just noticed while looking – that `MakefileBuild.vc` is setting `WIN32` in a few places. Do you think that's necessary now that the codebase uses `_WIN32`?

Those defines can be removed. Do you want to make a pr for it? if not i can do it later.

vszakats:
> > Completely unrelated – just noticed while looking – that `MakefileBuild.vc` is setting `WIN32` in a few places. Do you think that's necessary now that the codebase uses `_WIN32`?
> 
> Those defines can be removed. Do you want to make a pr for it? if not i can do it later.

It's better if you do it – I have no ways to test winbuild mods.

jay:
ok #13742

PR #13744, opened 2024-05-22
build: add more supported attributes for the IAR compiler

josepho0918:
Enable `deprecated` and `format` attributes for the IAR compiler.


bagder (approved):

bagder:
Thanks!

PR #13751, opened 2024-05-22
cmake: fix building with both sha256 and sha512 in unity mode

pps83:
+ `Ch` and `Maj` are also defined in curl_sha512_256.c.
 + Avoid defining short defines like `R`, `S`

### DIFF
--- a/lib/curl_sha512_256.c
+++ b/lib/curl_sha512_256.c
@@ -476,8 +476,8 @@ MHDx_sha512_256_transform(curl_uint64_t H[SHA512_256_HASH_SIZE_WORDS],
 
   /* 'Ch' and 'Maj' macro functions are defined with widely-used optimisation.
      See FIPS PUB 180-4 formulae 4.8, 4.9. */
-#define Ch(x,y,z)     ( (z) ^ ((x) & ((y) ^ (z))) )
-#define Maj(x,y,z)    ( ((x) & (y)) ^ ((z) & ((x) ^ (y))) )
+#define Sha512_Ch(x,y,z)     ( (z) ^ ((x) & ((y) ^ (z))) )
+#define Sha512_Maj(x,y,z)    ( ((x) & (y)) ^ ((z) & ((x) ^ (y))) )
 
   /* Four 'Sigma' macro functions.
      See FIPS PUB 180-4 formulae 4.10, 4.11, 4.12, 4.13. */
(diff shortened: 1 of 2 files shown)


vszakats:
What do you think of also prefixing `Ch` and `Maj` in `curl_sha512_256.c`? Mainly for good measure and to avoid bumping into this when adding a future hash.

pps83:
> What do you think of also prefixing `Ch` and `Maj` in `curl_sha512_256.c`? Mainly for good measure/symmetry and to avoid bumping into this when adding a future hash.

updated

lib/curl_sha512_256.c
```diff
@@ -548,8 +548,8 @@ MHDx_sha512_256_transform(curl_uint64_t H[SHA512_256_HASH_SIZE_WORDS],
        'wt' calculation may change other data as well every time when
        used. */
 #define SHA2STEP64(vA,vB,vC,vD,vE,vF,vG,vH,kt,wt) do {                  \
-      (vD) += ((vH) += SIG1 ((vE)) + Ch ((vE),(vF),(vG)) + (kt) + (wt)); \
-      (vH) += SIG0 ((vA)) + Maj ((vA),(vB),(vC)); } while (0)
+      (vD) += ((vH) += SIG1 ((vE)) + Sha512_Ch((vE),(vF),(vG)) + (kt) + (wt)); \
+      (vH) += SIG0 ((vA)) + Sha512_Maj((vA),(vB),(vC)); } while (0)
```
vszakats, line 552:
```suggestion
      (vD) += ((vH) += SIG1((vE)) + Sha512_Ch((vE),(vF),(vG)) + (kt) + (wt)); \
      (vH) += SIG0((vA)) + Sha512_Maj((vA),(vB),(vC)); } while (0)
```
Still need to find at least one character to fit `checksrc`.
Maybe `SIG1(vE)`? `SIG1()` (and `SIG0()`) already put the argument into parenthesis.

pps83: can also rename `Sha512_Maj` to `S512_Maj`

vszakats: I'd vote for a consistent prefix. Perhaps just splitting the line would work instead?

vszakats (approved):

jay:
Thanks

PR #13753, opened 2024-05-22
GHA: increase timeout for Cygwin autotools build tests step

orgads:
Apparently 10 minutes are not (always) enough:
https://github.com/curl/curl/actions/runs/9197003907/job/25296439556#step:8:1936

Closes #13753


vszakats:
Can you link an example run where this step expired?

orgads:
> Can you link an example run where this step expired?

https://github.com/curl/curl/actions/runs/9197003907

vszakats:
Thanks @orgads

Issue #13752, opened 2024-05-22
content length not available in header function, again


piru:
### I did this

Use `CURLINFO_CONTENT_LENGTH_DOWNLOAD_T` in `CURLOPT_HEADERFUNCTION` function when the last header is processed (empty newline is received). Rather than getting the content length `-1` is returned. The call working in this specific scenario is undocumented behaviour that some (broken) code relies on. This issue was tracked earlier as #7804 and has surfaced again couple of versions ago (not sure when exactly).

### I expected the following

Maybe this behaviour should be maintained since for example WebKit curl backend (*) is broken now. However, this is undocumented behaviour after all, so not fixing it is perfectly fine as well. WebKit and other code depending on this should be fixed either way.

*) https://github.com/WebKit/WebKit/blob/57affb62969db66ef07975567f4876d62d86da55/Source/WebCore/platform/network/curl/CurlRequest.cpp#L320

### curl/libcurl version

curl 8.8.0

### operating system

Linux hostname 6.8.9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.8.9-1 (2024-05-15) x86_64 GNU/Linux

piru:
Ok that works at least in initial tests. I also had to defer `Curl_dyn_reset(&data->state.headerb);` to occur only after the `Curl_client_write()` call.

piru:
Just moving the `Curl_dyn_reset` later is wrong though, at least when considering the comment at https://github.com/curl/curl/blob/1c4813c769ea65c128c067004d0711433288aa05/lib/http.c#L3663

I tried creating a dynbuf copy of the hd, hdlen data and then doing `Curl_dyn_reset` as before, and using the copy for the `Curl_client_write´. It works (tests pass), but if there is recursion the order of header callbacks changes. I don't feel confident if this would be correct.

icing:
My proposal on how to solve this - test case included: #13757.

PR #13757, opened 2024-05-23
http, write last header line late

icing:
- HEADERFUNCTIONS might inspect response properties like CURLINFO_CONTENT_LENGTH_DOWNLOAD_T on seeing the last header line. If the line is being written before this is initialized, values are not available.
- write the last header line late when analyzing a HTTP response so that all information is available at the time of the writing.
- add test1485 to verify that CURLINFO_CONTENT_LENGTH_DOWNLOAD_T works on seeing the last header.
- refs #13752

### DIFF
--- a/lib/http.c
+++ b/lib/http.c
@@ -3411,7 +3411,7 @@ static CURLcode http_on_response(struct Curl_easy *data,
     conn->bundle->multiuse = BUNDLE_NO_MULTIUSE;
   }
 
-  if(k->httpcode < 100 && last_hd) {
+  if(k->httpcode < 200 && last_hd) {
     /* Intermediate responses might trigger processing of more
      * responses, write the last header to the client before
      * proceeding. */
--- a/lib/http.c
+++ b/lib/http.c
@@ -3363,8 +3363,35 @@ CURLcode Curl_bump_headersize(struct Curl_easy *data,
   return CURLE_OK;
 }
 
+static CURLcode http_write_header(struct Curl_easy *data,
+                                  const char *hd, size_t hdlen)
+{
+  CURLcode result;
+  int writetype;
+
+  /* now, only output this if the header AND body are requested:
+   */
+  Curl_debug(data, CURLINFO_HEADER_IN, (char *)hd, hdlen);
+
+  writetype = CLIENTWRITE_HEADER |
+    ((data->req.httpcode/100 == 1) ? CLIENTWRITE_1XX : 0);
+
+  result = Curl_client_write(data, writetype, hd, hdlen);
+  if(result)
+    return result;
+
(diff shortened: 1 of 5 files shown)


lib/http.c
```diff
@@ -3384,9 +3411,20 @@ static CURLcode http_on_response(struct Curl_easy *data,
     conn->bundle->multiuse = BUNDLE_NO_MULTIUSE;
   }
 
+  if(k->httpcode < 100 && last_hd) {
```
piru, line 3414: isn't `< 100` just outright error or is there some other hidden meaning in this specific case?

icing: A bug, thanks for catching. Fixed in last push.

lib/http.c
```diff
@@ -3624,7 +3666,16 @@ static CURLcode http_on_response(struct Curl_easy *data,
     k->download_done = TRUE;
 
   /* final response without error, prepare to receive the body */
-  return Curl_http_firstwrite(data);
+  result = Curl_http_firstwrite(data);
+
+out:
+  if(last_hd) {
```
bagder, line 3672: Surely this should not be done if it already has an error in `result` here?

icing: I am aiming or backward compatibility. We did write the end first and then fail. A callback now would no longer see the last header line. Maybe it does not make a difference, but...

bagder: But that first error could come as an error from the header callback can't it? Meaning that it would then be called again?

tests/data/test1485
```diff
@@ -0,0 +1,51 @@
+<testcase>
```
piru, line 1: Should we also have test cases for HTTP/2 (and maybe HTTP/3) ?

lib/http.c
icing: I think not. Line 3419 sets `last_hd = NULL;` before checking the result.

tests/data/test1485
icing: There is no 101 protocol switch for HTTP/3, though. For HTTP/2 this could be done.

piru: Hmm, but this is about having the `CURLINFO` things available by the time header callback is called for the (last) empty header though, and not 101 protocol switch.

icing: Yes, the PR fixes the issue you raised, I believe...😌

piru: Right, so the other protocol go through the same codepaths in the end.

lib/http.c
bagder: Ah yes, I missed that. Okay, but can you then add a comment explaining why this is called even if it may already have received an error so that future me won't have to ask again? :smiley:

PR #13763, opened 2024-05-24
winsock, move sendbuf update into cf-socket.c

icing:
- we updated the WINSOCK sendbuf in the transfer loop, but cf-socket.c seems a much better place.
- moving code and timestamp into the socket filter

I am not a Windows dev, hope this works as intended.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -395,7 +395,25 @@ void Curl_sndbufset(curl_socket_t sockfd)
 
   setsockopt(sockfd, SOL_SOCKET, SO_SNDBUF, (const char *)&val, sizeof(val));
 }
+
+#ifndef SIO_IDEAL_SEND_BACKLOG_QUERY
+#define SIO_IDEAL_SEND_BACKLOG_QUERY 0x4004747B
 #endif
+
+static void win_update_buffer_size(curl_socket_t sockfd)
+{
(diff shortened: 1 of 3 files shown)


icing:
> I'm seeing some strangeness when i test this, data->conn->writesockfd is -1 for the first call, and more if i step through it in the debugger
> 
> `curl -v --tls-max 1.2 https://example.com`
> 
> edit: I've got breakpoints which is slowing things down and causes the 1000 logic to trigger more often but anyway the point is it seems data->conn->writesockfd is not set before at least the first call to cf_socket_send

Thanks for testing. The move was incomplete. The sendbuf update should use the filter's socket `ctx->sock` and not the writesockfd at the connection. Pushed an update.

jay (approved):
This PR is fine as is however I've made a branch [modified__winsock-move-sendbuf-update](https://github.com/curl/curl/compare/master...jay:curl:modified__winsock-move-sendbuf-update?expand=1) and added a [change](f63c28a1) that renames the functions to be easier to understand, moves the time check into win_update_sndbuf_size and 
saves the last sndbuf_size to avoid multiple setsockopt calls every second. I can open a separate PR after you land this one or you're welcome to squash it in.

jay:
Thanks. I landed this as is and the follow-up is in #13827.

PR #13771, opened 2024-05-24
examples: fix compiling with MSVC

vszakats:
- `websocket.c`: use `Sleep()` on Windows.
  `sleep()` and `unistd.h` are not available in MSVC.

- `http2-upload.c`: use local `gettimeofday()` implementation
  when compiled with MSVC.
  (Alternate solution is to disable the trace function for MSVC.)
  Public domain code copied and adapted from libssh2:
  https://github.com/libssh2/libssh2/blob/e973493f992313b3be73f51d3f7ca6d52e288558/src/misc.c#L719-L743

- silence compiler warning for deprecated `inet_addr()`.
  Also drop duplicate winsock2 include.
  ```
  curl\docs\examples\externalsocket.c(125,32): error C2220: the following warning is treated as an error [curl\bld\docs\examples\curl-example-externalsocket.vcxproj]
  curl\docs\examples\externalsocket.c(125,32): warning C4996: 'inet_addr': Use inet_pton() or InetPton() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings [curl\bld\docs\examples\curl-example-e
  ```
  Ref: https://github.com/curl/curl/actions/runs/9227337318/job/25389073450#step:4:95

- silence an MSVC compiler warning. This is in conflict with
  `checksrc` rules, so silence the rule in favour of the
  warning-free C syntax.
  ```
  curl\docs\examples\multi-legacy.c(152,1): error C2220: the following warning is treated as an error [curl\bld\docs\examples\curl-example-multi-legacy.vcxproj]
  curl\docs\examples\multi-legacy.c(152,1): warning C4706: assignment within conditional expression [curl\bld\docs\examples\curl-example-multi-legacy.vcxproj]
  ```
  Ref: https://github.com/curl/curl/actions/runs/9227337318/job/25389073450#step:4:226

- do not use `sys/time.h` and `unistd.h` in Windows builds.
  Some of these includes look unnecessary. Subject to another
  PR.

Cherry-picked from #13766
Closes #13771


docs/examples/CMakeLists.txt
```diff
@@ -40,5 +40,8 @@ foreach(example IN LISTS check_PROGRAMS)
   if(LIB_SELECTED STREQUAL LIB_STATIC AND WIN32)
     set_property(TARGET ${example_target} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_STATICLIB")
   endif()
+  if(MSVC)
+    set_property(TARGET ${example_target} APPEND PROPERTY COMPILE_DEFINITIONS "_WINSOCK_DEPRECATED_NO_WARNINGS")  # for inet_addr()
```
MarcelRaad, line 44: Maybe disable the warning only for the one example which emits it? As the warning is mostly about old IPv6-incompatible functions, I think it's valuable even in examples.

vszakats: I agree, it's just a single source that's affected.

Once the tests pass in #13766, I'll apply the patch to this PR.

It applies this macro to all compilers (not just MSVC) just in
case some would rely on the Microsoft SDK headers (or would
replicate this behaviour). Or do you think it's better to restrict
to `_MSC_VER`?

MarcelRaad: Looks good to me as is now, thanks!

MarcelRaad (approved):

PR #13780, opened 2024-05-25
[configure] Use `$EGREP` in place of `grep -E`

asedeno:
`$EGREP` is set based on an earlier test in configure so that we can work with systems that have `egrep` and a `grep` that does not support `-E`.


bagder (approved):

bagder:
Thanks!

PR #13789, opened 2024-05-27
noproxy: patterns need to be comma separated

bagder:
or they will not parse correctly.

Mentioned in DEPRECATED since Janurary 2023 (in 7ad8a7ba9ebdedc).


Cleanfastllc (commented):
Yes

PR #13792, opened 2024-05-27
cmake: allow `ENABLE_CURLDEBUG=OFF` with `ENABLE_DEBUG=ON`

vszakats:
Before this patch, `ENABLE_CURLDEBUG` (memory tracking) was
unconditionally enabled when `ENABLE_DEBUGBUILD` was set. This made
testing some build configurations complicated. To fix it, this patch
makes `ENABLE_CURLDEBUG` to receive the value of `ENABLE_DEBUG` by
default, while allowing free override by the user.

This allows to use the config:
`ENABLE_DEBUGBUILD=ON ENABLE_CURLDEBUG=OFF`
to enable debug features, without also enabling memory tracking.

This is important because some other build methods allow to set one of
these features but not the other. This patch allows to test any
combination with CMake.

This makes it unnecessary to use the workaround of passing
`-DDEBUGBUILD` via `CMAKE_C_FLAGS`. Which has the disadvantage that our
CMake logic cannot easily detect it, e.g. for disabling symbol hiding on
Windows for `ENABLE_DEBUG`/`DEBUGBUILD` builds.

Cherry-picked from #13718
Closes #13792


vszakats:
Several new test fallouts in NetBSD, OpenBSD and Windows, though it's unlikely this PR causes them. There is no CI test using a config this PR enables.

PR #13802, opened 2024-05-27
openssl: rectify the TLS version check for QUIC

### LINKED ISSUE #13799: Regression: In 8.8.0, HTTP3 with QuicTLS errors out with "QUIC needs at least TLS version 1.3" with CURLOPT_SSLVERSION set as documented
### I did this

The `CURLOPT_SSLVERSION` option has [always been documented as](https://curl.se/libcurl/c/CURLOPT_SSLVERSION.html) specifying the _**minimum** acceptable version_ when using the `CURL_SSLVERSION_TLSv1_?` constants (as opposed to the more recently added `CURL_SSLVERSION_MAX_TLSv1_?` variants to limit the maximum allowed). A perfectly sensible choice nowadays then is to set CURLOPT_SSLVERSION to CURL_SSLVERSION_TLSv1_2, which has, thus far, meant allowing either TLSv1.2 or TLSv1.3, with HTTP3/QUIC naturally using TLSv1.3 as defined.

Unfortunately 8.8.0 shipped with a new mistaken check in [/lib/vtls/openssl.c](https://github.com/curl/curl/blob/837378309860839496d2c7429d5befa7faac538c/lib/vtls/openssl.c#L3581) that confuses the minimum TLS version for the maximum.
For anything other than HTTP3, [lib]curl 8.8.0 continues to work fine and automatically uses TLSv1.3 or v1.2 when set to `CURL_SSLVERSION_TLSv1_2`, as expected.

Example (assuming quictls/nghttp3/ngtcp2 build):
```
$ curl --tlsv1.2 --http3-only https://www.cloudflare.com/cdn-cgi/trace

curl: (35) QUIC needs at least TLS version 1.3
```

### I expected the following

Proper functioning with no error message, since we only requested a _minimum_ of TLSv1.2, and did not specify `--tls-max 1.2` or lower.
It now fails even if you spell out the maximum as v1.3: `curl --tls-max 1.3 --tlsv1.2 --http3-only https://www.cloudflare.com/cdn-cgi/trace`

HTTP3 now only works if you outright set either a minimum of 1.3, or leave `CURLOPT_SSLVERSION` set as `CURL_SSLVERSION_DEFAULT`.

### curl/libcurl version

curl 8.8.0 (x86_64-apple-darwin) libcurl/8.8.0 quictls/3.0.13 zlib/1.2.12 brotli/1.1.0 zstd/1.5.6 libidn2/2.3.7 libssh/0.10.6/openssl/zlib nghttp2/1.62.1 ngtcp2/1.5.0 nghttp3/1.3.0
Release-Date: 2024-05-22
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

Not OS dependent (but tested on Darwin 23.5.0 ie. macOS Sonoma 14.5 23F79)

bagder:
The version check wrongly complained and return error if the *minimum* version was set to something less than 1.3. QUIC is always TLS 1.3, but that means minimum 1.2 is still fine to ask for.

This also renames the local variable to make the mistake harder to make in the future.

Regression shipped in 8.8.0
    
Follow-up to 3210101088dfa3d6a125

Reported-by: fds242 on github
Fixes #13799

### DIFF
--- a/lib/vtls/gtls.c
+++ b/lib/vtls/gtls.c
@@ -376,9 +376,15 @@ set_ssl_version_min_max(struct Curl_easy *data,
   long ssl_version = conn_config->version;
   long ssl_version_max = conn_config->version_max;
 
+  if((ssl_version == CURL_SSLVERSION_DEFAULT) ||
+     (ssl_version == CURL_SSLVERSION_TLSv1))
+    ssl_version = CURL_SSLVERSION_TLSv1_0;
+  if(ssl_version_max == CURL_SSLVERSION_MAX_NONE)
+    ssl_version_max = CURL_SSLVERSION_MAX_DEFAULT;
+
   if(peer->transport == TRNSPRT_QUIC) {
-    if((ssl_version != CURL_SSLVERSION_DEFAULT) &&
-       (ssl_version < CURL_SSLVERSION_TLSv1_3)) {
+    if((ssl_version_max != CURL_SSLVERSION_MAX_DEFAULT) &&
+       (ssl_version_max < CURL_SSLVERSION_MAX_TLSv1_3)) {
       failf(data, "QUIC needs at least TLS version 1.3");
       return CURLE_SSL_CONNECT_ERROR;
      }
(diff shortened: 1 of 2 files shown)


fds242:
FWIW, the same issue might've also got added to GnuTLS in lib/vtls/gtls.c, in that it at least contains the same error message, and largely the same checking code. I'm not set up to test or at all familiar with GnuTLS support myself though.

PR #13809, opened 2024-05-28
[WIP] CI: drop cmake Debug/Release configs

vszakats:
`Debug`/`Release` is no longer required to build with or without 
curl debug facilities, or to run the tests. Dropping it simplifies the
configuration. `Debug` configs are also expected to build faster
(due no compiler optimization), but might use more disk space and
run slower when running tests, than `Release`. CMake docs don't
make it clear what the default behaviour is and suggest to select 
one explicitly. `Debug` can also be useful to test `assert()` lines.

https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html
https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#default-and-custom-configurations

It might make sense to choose one universally for all jobs. This
may save some jobs where the only distinction was Debug vs.
Release. (→ there was just one)

### DIFF
--- a/.github/workflows/non-native.yml
+++ b/.github/workflows/non-native.yml
@@ -61,16 +61,16 @@ jobs:
             cmake -B bld \
               -DCMAKE_UNITY_BUILD=ON \
               -DCURL_WERROR=ON \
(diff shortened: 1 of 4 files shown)


vszakats:
This doesn't seem useful overall.

PR #13813, opened 2024-05-28
curl: offer a default 'random' variable

bagder:
... and two new output functions: `uuid` and `hex`

The variable holds a 128 bit semi-random value. Output it like

- `{{random:hex}}` to get it all as a 32 byte hex sequence
- `{{random:uuid}}` shows it as a UUID sequence
- `{{random:url}}` shows it URL encoded
- `{{random:b64}}` shows it base64 encoded

For now, this holds fixed content, meaning that if you use it several times in the command line it will have the same value.

- is this useful and wanted at all?
- should we use a "reserved" prefix for internally provided variables?
- if so, this needs to be documented
- test cases needed

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -2711,6 +2711,8 @@ CURLcode operate(struct GlobalConfig *global, int argc, argv_item_t argv[])
   setlocale(LC_NUMERIC, "C");
 #endif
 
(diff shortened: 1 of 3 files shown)


bagder:
Saving this idea for a later.

Issue #13754, opened 2024-05-23
aws-sigv4 does not work with URLs containing "=" and other special characters


ashtuchkin:
### I did this

I'm trying to use --aws-sigv4 to download S3 objects containing "=" in their key. Example `data/asset_id=my-asset/dt=2024-05-22/data.parquet` - this is called hive partitioning scheme and is popular in data lakes.

Here's how I call it:
```shell
BUCKET="lasca-prod"
KEY="data/asset_id=my-asset/dt=2024-05-22/data.parquet"
curl --aws-sigv4 "aws:amz:us-west-2:s3" \
    --user "${AWS_ACCESS_KEY_ID}:${AWS_SECRET_ACCESS_KEY}" \
    -H "x-amz-security-token: ${AWS_SESSION_TOKEN}" \
    "https://s3.us-west-2.amazonaws.com/$BUCKET/$KEY"
```

Curl 8.8 returns the following error (formatted for readability):
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>SignatureDoesNotMatch</Code>
    <Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
    <AWSAccessKeyId>snip</AWSAccessKeyId>
    <StringToSign>AWS4-HMAC-SHA256
20240523T061547Z
20240523/us-west-2/s3/aws4_request
ce199afb9ea57628628a3d3254934e1ee480977872d06b3da40baa5ae4d2f9a7</StringToSign>
    <SignatureProvided>snip</SignatureProvided>
    <StringToSignBytes>snip</StringToSignBytes>
    <CanonicalRequest>GET
/lasca-prod/data/asset_id%3Dmy-asset/dt%3D2024-05-22/data.parquet

host:s3.us-west-2.amazonaws.com
x-amz-content-sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
x-amz-date:20240523T061547Z
x-amz-security-token:snip 

host;x-amz-content-sha256;x-amz-date;x-amz-security-token
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855</CanonicalRequest>
  ...
</Error>
```

[...]

PR #13814, opened 2024-05-28
aws-sigv4: fix canonical path encoding

outscale-mgo:
fix: #13754

I refactor canon_query, so I could use the encoding part of the function to use it in the path.

As the path doesn't encode '/', but encode '=', I had to add some conditions to know If I was doing the query or path encoding.

Also, instead of adding a `bool in_path` variable, I use `bool *found_equals` to know if the function was called for the query or path, as found_equals is used only in query_encoding.

Note: When I try replacing manually `=` with `%3D`, curl without this patch worked, so they might still be a bug out-there.

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -423,6 +423,76 @@ static int compare_func(const void *a, const void *b)
 
 #define MAX_QUERYPAIRS 64
 
+/**
+ * found_equals have a double meaning,
+ * detect if an equal have been found when called from canon_query,
+ * and mark that this function is called to compute the path,
+ * if found_equals is NULL.
+ */
+static CURLcode canon_string(const char *q, size_t len,
+                             struct dynbuf *dq, bool *found_equals)
+{
+  CURLcode result = CURLE_OK;
+
+  for(; len && !result; q++, len--) {
+    if(ISALNUM(*q))
+      result = Curl_dyn_addn(dq, q, 1);
+    else {
+      switch(*q) {
+      case '-':
+      case '.':
+      case '_':
+      case '~':
+        /* allowed as-is */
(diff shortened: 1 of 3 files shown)


lib/http_aws_sigv4.c
```diff
@@ -423,6 +423,77 @@ static int compare_func(const void *a, const void *b)
...
+ * and mark that this function is called to compute the path,
+ * if found_equals is NULL.
+ */
+static CURLcode canon_string(const char *q, size_t str_len,
+                             struct dynbuf *dq, bool *found_equals)
+{
+  CURLcode result = CURLE_OK;
+  size_t len;
+
+  for(len = str_len; len && !result; q++, len--) {
```
bagder, line 438: You can remove a variable like this:

```suggestion
static CURLcode canon_string(const char *q, size_t len,
                             struct dynbuf *dq, bool *found_equals)
{
  CURLcode result = CURLE_OK;

  for(; len && !result; q++, len--) {
```

lib/http_aws_sigv4.c
```diff
@@ -460,54 +531,11 @@ static CURLcode canon_query(struct Curl_easy *data,
...
-          char out[3]={'%'};
-          out[1] = hex[((unsigned char)*q)>>4];
-          out[2] = hex[*q & 0xf];
-          result = Curl_dyn_addn(dq, out, 3);
-          break;
-        }
-        }
-      }
-    }
+    canon_string(q, ap->len, dq, &found_equals);
```
bagder, line 538: this should receive the return code in `result` should it not?

```suggestion
    result = canon_string(q, ap->len, dq, &found_equals);
```

outscale-mgo:
done, thanks for the review.

bagder (approved):

bagder:
Thanks!

PR #13643, opened 2024-05-14
lib: fix compiler warnings (gcc)

vszakats:
Seen when setting `ENABLE_DEBUG=ON` and `-DDEBUGBUILD` for mingw-w64
gcc 13.2.0 CMake unity builds in 'Release' configurations.

```
curl/lib/curl_gethostname.c:71:5: error: 'strncpy' specified bound 1025 equals destination size [-Werror=stringop-truncation]
   71 |     strncpy(name, force_hostname, namelen);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from curl/_bld/lib/CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c:175:
In function 'hostcache_timestamp_remove',
    inlined from 'Curl_hash_clean_with_criterium' at curl/lib/hash.c:265:19,
    inlined from 'Curl_hash_clean_with_criterium' at curl/lib/hash.c:247:1,
    inlined from 'hostcache_prune' at curl/lib/hostip.c:228:3,
    inlined from 'Curl_hostcache_prune' at curl/lib/hostip.c:256:21:
curl/lib/hostip.c:205:12: error: 'now' may be used uninitialized [-Werror=maybe-uninitialized]
  205 |     time_t age = prune->now - c->timestamp;
      |            ^~~
curl/lib/hostip.c: In function 'Curl_hostcache_prune':
curl/lib/hostip.c:241:10: note: 'now' was declared here
  241 |   time_t now;
      |          ^~~
In function 'hostcache_timestamp_remove',
    inlined from 'fetch_addr' at curl/lib/hostip.c:310:8:
curl/lib/hostip.c:205:23: error: 'user.now' may be used uninitialized [-Werror=maybe-uninitialized]
  205 |     time_t age = prune->now - c->timestamp;
      |                  ~~~~~^~~~~
curl/lib/hostip.c: In function 'fetch_addr':
curl/lib/hostip.c:304:33: note: 'user' declared here
  304 |     struct hostcache_prune_data user;
      |                                 ^~~~
In file included from curl/_bld/lib/CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c:40:
curl/lib/cf-socket.c: In function 'cf_socket_send':
curl/lib/cf-socket.c:1294:10: error: 'c' may be used uninitialized [-Werror=maybe-uninitialized]
 1294 |     if(c >= ((100-ctx->wblock_percent)*256/100)) {
[...]

PR #13815, opened 2024-05-28
lib: prefer `var = time(NULL)` over `time(&var)`

vszakats:
Following up on previous occurrences showing up as gcc warnings, replace
the remaining `time(&var)` calls with `var = time(NULL)`, though these
aren't specifically causing compiler warnings. These are in the TFTP
client code (`lib/tftp.c`), except one which is in a debug branch in
`lib/http_aws_sigv4.c`.

What's unexplainable is that this patch seems to mitigate TFTP tests
often hanging or going into an infinite loop on GHA windows workflows
with MSYS2, mingw-w64 and MSVC (Cygwin is unaffected):
  https://github.com/curl/curl/pull/13599#issuecomment-2119372376
TFTP hangs did not entirely disappear though, so could be unrelated.

`time()` docs:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/time-time32-time64
https://manpages.debian.org/bookworm/manpages-dev/time.2.en.html

Follow-up to 58ca0a2f0743a586716ca357c382b29e3f08db69 #13800
Follow-up to d0728c9109629ee82b855b350a4c3f1f52ee61df #13643
Closes #13815

---

More test runs here: https://github.com/vszakats/curl/pull/1

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -232,7 +232,7 @@ jobs:
         timeout-minutes: 30
         shell: msys2 {0}
         run: |
-          export TFLAGS='-j14 !TFTP ${{ matrix.tflags }}'
+          export TFLAGS='-j14 ${{ matrix.tflags }}'
           if [ -x "$(cygpath "${SYSTEMROOT}/System32/curl.exe")" ]; then
             TFLAGS+=" -ac $(cygpath "${SYSTEMROOT}/System32/curl.exe")"
           fi
@@ -296,7 +296,7 @@ jobs:
         timeout-minutes: 40
         shell: msys2 {0}
         run: |
-          export TFLAGS='-j14 !TFTP ${{ matrix.tflags }}'
+          export TFLAGS='-j14 ${{ matrix.tflags }}'
           if [ -x "$(cygpath "${SYSTEMROOT}/System32/curl.exe")" ]; then
             TFLAGS+=" -ac $(cygpath "${SYSTEMROOT}/System32/curl.exe")"
           fi
@@ -398,7 +398,7 @@ jobs:
         timeout-minutes: 40
         shell: C:\msys64\usr\bin\bash.exe {0}
         run: |
-          export TFLAGS='-j14 !TFTP ${{ matrix.tflags }}'
+          export TFLAGS='-j14 ${{ matrix.tflags }}'
           if [ -x "$(cygpath "${SYSTEMROOT}/System32/curl.exe")" ]; then
             TFLAGS+=" -ac $(cygpath "${SYSTEMROOT}/System32/curl.exe")"
           fi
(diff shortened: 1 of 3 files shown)


vszakats:
Okay, so got hung again right away, hard to tell where:
https://github.com/curl/curl/actions/runs/9273526999/job/25513835652?pr=13815#step:6:4437

jay:
That's truly weird. I wonder if this is a 32/64 bit issue, the [msdn time doc](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/time-time32-time64?view=msvc-170) says it's supposed to be 64 bit in most cases, and so maybe the reason for the weirdness is time(&foo) is either the function is expecting a 32-bit and the variable is 64 bit or the other way around, and it's writing outside its memory or half the memory isn't written? 

try using a 64-bit variable explicitly, like
{
__time64_t now = 0xFFFFFFFFFFFFFFFFi64;
time(&now);
DEBUGASSERT(now < 0xFFFFFFFFUL);
DEBUGASSERT(sizeof(time_t) == 8);
}

dfandrich:
Truly weird is also a good description of the possibly related TFTP timeout issues I found in #12040. The problem there appeared to be a simple `time(&var)` call was being set to 0.

vszakats:
I could not spot any issues with `time_t` size, or the wrong `time()` flavour being called. This seems like a lucky coincidence. Also TFTP is still able to hang with this patch in place:
https://github.com/vszakats/curl/actions/runs/9278913609/job/25530650928#step:14:13754

It'd probably still be a good PR to merge (_without_ re-enabling TFTP tests) just to do these calls the same way across the codebase. The assigment style can also avoid the `EOVERFLOW` failure mode on Linux at least.

As a next step maybe it'd be useful to check for a -1 value returned, as Jay suggested.

jay (approved):

vszakats:
Another idea is to replace remaining direct `time()` calls with `Curl_now()`, which is using more robust alternatives on popular platforms.

@jay: Do you vote to re-enable TFTP tests with this patch, or leave them disabled? We can always disable them again in a separate PR – or try other fixes.

jay:
Actually I would leave them disabled because they don't seem to work properly for mingw AFAICS, and based on what you've said and no reports other than us I don't think it's a curl bug

jay:
> Another idea is to replace remaining direct `time()` calls with `Curl_now()`, which is using more robust alternatives on popular platforms.

Curl_now and time are not synonymous, so that would have to be approached very carefully and I don't know if it's worth the work

jay:
> Curl_now and time are not synonymous, so that would have to be approached very carefully and I don't know if it's worth the work

I reviewed the time calls in tftp and they are all based around recording the receive timestamp so they could easily be changed to Curl_now but I don't see how it would be better

dfandrich:
Curl_now will use a monotonic time source if available so timeouts won't break when certain clock updates happen.

Issue #13818, opened 2024-05-28
ECH: SIGSEGV when use HTTP/3


vvb2060:
### I did this

curl --http3 --ech true --doh-url https://1.0.0.1/dns-query https://www.cloudflare.com/cdn-cgi/trace

```
* thread #1, name = 'curl', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
  * frame #0: 0x0000007fb6b69880 libc.so`strlen_default + 16
    frame #1: 0x00000055556e682c curl`fetch_addr + 60
    frame #2: 0x00000055556e67a0 curl`Curl_fetch_addr + 88
    frame #3: 0x000000555574deec curl`Curl_ossl_ctx_init + 1484
    frame #4: 0x0000005555753564 curl`cf_ngtcp2_connect + 932
    frame #5: 0x00000055556d59f0 curl`cf_he_connect + 1328
    frame #6: 0x00000055556d4dfc curl`cf_setup_connect + 104
    frame #7: 0x00000055556d8a10 curl`cf_hc_connect + 1056
    frame #8: 0x00000055556d3b28 curl`Curl_conn_connect + 96
    frame #9: 0x000000555570c3f8 curl`multi_runsingle + 1564
    frame #10: 0x000000555570b938 curl`curl_multi_perform + 312
    frame #11: 0x00000055556c1e1c curl`main + 7808
    frame #12: 0x0000007fb6b685b8 libc.so`__libc_init + 112
```

### I expected the following

exit 0

### curl/libcurl version

curl 8.8.0

### operating system

android arm64

dfandrich:
What is the output of curl -V? Can you reproduce this on other than Android?

vvb2060:
https://github.com/vvb2060/curl-android/releases/tag/v8.8.0
```
curl 8.8.0-DEV (unknown-linux-android) libcurl/8.8.0-DEV BoringSSL zlib/1.2.11 nghttp2/1.62.1 ngtcp2/1.5.0 nghttp3/1.3.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS ECH HSTS HTTP2 HTTP3 HTTPS-proxy IPv6 Largefile libz NTLM SSL threadsafe UnixSockets
```
I only have Android build environment

[...]

PR #13822, opened 2024-05-29
h3+ech fix

sftcd:
This PR addresses a crash (in ECH code) when the experimental ECH feature is attempted with the experimental h3 feature. Note that the fix doesn't mean h3+ECH will work, it just means it won't cause a crash.

This addresses https://github.com/curl/curl/issues/13818


sftcd:
I just noticed there could be a similar issue if using wolfssl+ECH+h3. Will check that out now.

bagder:
Thanks!

Issue #13782, opened 2024-05-25
8.8.0: Test 1553 (IMAP cleanup before a connection was created) appears flaky on some Gentoo systems


Kangie:
I am unable to have test 1553 pass consistently on Gentoo in 8.8.0, however we have not had this issue with previous versions.

I raised this briefly during the 8.8.0-older-mbedtls testing, but Stefan and Daniel were unable to reproduce so I put it down to a weird failure as it'd already passed on another of my dev machines with a very similar configuration. One of the Gentoo automated build machines (tinderboxes) and a Gentoo developer have noted this issue to me, so I figured it worth raising here.

I have one workstation where this test consistently passes. I have another where it passes if I run the tests as my user or if I run them with the portage `network-sandbox` feature disabled (fine for testing, but not sustainable).

We can disable this test in the Gentoo ebuild however I'm interested in why this was not an issue for us in the past and appears inconsistent.

[build.log](https://github.com/curl/curl/files/15445469/build.log)

### See also

- https://bugs.gentoo.org/932660

### I did this

```
srcdir=/var/tmp/portage/net-misc/curl-8.8.0/work/curl-8.8.0/tests /usr/bin/perl -I/var/tmp/portage/net-misc/cu
rl-8.8.0/work/curl-8.8.0/tests /var/tmp/portage/net-misc/curl-8.8.0/work/curl-8.8.0/tests/runtests.pl -a -s -n
 -v -a -k -am -p -j64 !241 !1083
Using curl: ../src/curl
CMD (0): ../src/curl --version 1>log/curlverout.log 2>log/curlvererr.log
********* System characteristics ******** 
* curl 8.8.0 (x86_64-pc-linux-gnu) 
* libcurl/8.8.0 OpenSSL/3.2.1 zlib/1.3.1 c-ares/1.28.1 libpsl/0.21.5 nghttp2/1.62.0
* Features: alt-svc AsynchDNS HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM PSL SSL threadsafe TLS-SRP UnixS
ockets
* Disabled: 
* Host: localhost
* System: Linux localhost 6.8.8-gentoo--bouncy-mouse #1 SMP PREEMPT_DYNAMIC Wed May  8 14:03:09 AEST 2024 x86_
64 AMD Ryzen 9 5950X 16-Core Processor AuthenticAMD GNU/Linux
* OS: linux
* Jobs: 64
* Env: 
* Seed: 241509
[...]

PR #13824, opened 2024-05-29
test1553, debug sporadic failures

icing:
- sprodic failes of test1553 are reported in #13782. These seem only to happen in very specific environments, but point to a race somewhere in curl's async resolver handling.
- add debug logging to test1553 to get more output
- add debug logging to the c-ares resolve monitoring, e.g. listing sockets and timeouts returned


icing:
found the problem, will make another PR with the fix.

PR #13825, opened 2024-05-29
fix multi_wait() timeout handling

icing:
- determine the actual poll timeout *after* all sockets have been collected. Protocols and connection filters may install new timeouts during collection.
- add debug logging to test1533 where the mistake was noticed
- refs #13782


bagder (approved):
:+1: nice catch!!

PR #13830, opened 2024-05-30
Autoconf: Remove 'deeper' checks for `AC_CHECK_FUNCS`

Kangie:
The net effect of the deeper checks is to raise implicit function decls on modern compilers.

These checks appear to have been added ~20 years ago, relating to an unverifiable claim about HP-UX. Autoconf support for the platform has grown in leaps and bounds since.

It didn't cause a real problem here, but when investigating a FP this came up. No evidence has been identified that this was actually broken in the past, and there is no evidence that this is necessary now.

`-Werror=implicit-function-declarations` is enabled for both checks; without a working prototype they will both fail regardless. In the second case there will in fact never be a working prototype and therefore it will always fail unconditionally.

`AC_CHECK_FUNCS` does effectively the same thing as the removed checks, except it actually defines a dummy prototype to see if it links.

If `AC_CHECK_FUNCS` is broken on a given platform we have bigger problems than trying to build cURL. This should also be faster.

CC @thesamesam @eli-schwartz 

Bug: https://bugs.gentoo.org/932827

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3800,26 +3800,6 @@ AC_CHECK_FUNCS([\
   snprintf \
   utime \
   utimes \
-],[
-],[
-  func="$ac_func"
-  eval skipcheck=\$skipcheck_$func
-  if test "x$skipcheck" != "xyes"; then
-    AC_MSG_CHECKING([deeper for $func])
-    AC_LINK_IFELSE([
-      AC_LANG_PROGRAM([[
-      ]],[[
-        $func ();
-      ]])
-    ],[
-      AC_MSG_RESULT([yes])
-      eval "ac_cv_func_$func=yes"
-      AC_DEFINE_UNQUOTED(XC_SH_TR_CPP([HAVE_$func]), [1],
-        [Define to 1 if you have the $func function.])
-    ],[
-      AC_MSG_RESULT([but still no])
-    ])
-  fi
 ])
 
 dnl On Android, the only way to know if fseeko can be used is to see if it is


eli-schwartz:
Autoconf check:
```
configure:41712: checking for arc4random
configure:41712: x86_64-gentoo-linux-musl-gcc -o conftest -march=native -fstack-protector-all -O2 -ggdb -pipe -fdiagnostics-color=always -frecord-gcc-switches -Werror=implicit-function-decla
ration -Werror=implicit-int -Werror=int-conversion -Werror=incompatible-pointer-types -Werror-implicit-function-declaration      -Wl,-O1 -Wl,--as-needed -Wl,--defsym=__gentoo_check_ldflags__
=0     conftest.c -lnghttp2 -lpsl -lssl -lcrypto -lssl -lcrypto -lz  >&5
/usr/lib/gcc/x86_64-gentoo-linux-musl/13/../../../../x86_64-gentoo-linux-musl/bin/ld: /var/tmp/portage/net-misc/curl-8.8.0/temp/ccanlBMD.o: in function `main':
/var/tmp/portage/net-misc/curl-8.8.0/work/curl-8.8.0-abi_x86_64.amd64/conftest.c:216:(.text.startup+0x15): undefined reference to `arc4random'
configure:41712: $? = 1
configure: failed program was:
| /* confdefs.h */


[...]

| /* end confdefs.h.  */
| 
| #define arc4random innocuous_arc4random
| #ifdef __STDC__
| # include <limits.h>
| #else
| # include <assert.h>
| #endif
| #undef arc4random
| #ifdef __cplusplus
| extern "C"
| #endif
| char arc4random ();
| #if defined __stub_arc4random || defined __stub___arc4random
| choke me
| #endif
| 
| int main (void)
| {
| return arc4random ();
|  ;
|  return 0;
| }
configure:41712: result: no
```

Note how it fakes out the compiler, because what it is actually interested in is "does this symbol exist".

The "deeper" check:
```
configure:41725: checking deeper for arc4random
configure:41742: x86_64-gentoo-linux-musl-gcc -o conftest -march=native -fstack-protector-all -O2 -ggdb -pipe -fdiagnostics-color=always -frecord-gcc-switches -Werror=implicit-function-declaration -Werror=implicit-int -Werror=int-conversion -Werror=incompatible-pointer-types -Werror-implicit-function-declaration      -Wl,-O1 -Wl,--as-needed -Wl,--defsym=__gentoo_check_ldflags__=0     conftest.c -lnghttp2 -lpsl -lssl -lcrypto -lssl -lcrypto -lz  >&5
conftest.c: In function 'main':
conftest.c:204:9: error: implicit declaration of function 'arc4random' [-Werror=implicit-function-declaration]
  204 |         arc4random ();
      |         ^~~~~~~~~~
cc1: some warnings being treated as errors
configure:41742: $? = 1
configure: failed program was:
| /* confdefs.h */

[...]

| /* end confdefs.h.  */
| 
| 
| 
| int main (void)
| {
| 
|         arc4random ();
| 
|  ;
|  return 0;
| }
| 
configure:41756: result: but still no
```

-Werror=implicit-function-declarations actually comes from curl-compilers.m4 (it is also in my CFLAGS, so it appears twice). A build system that wants to check for symbols without using the "fake out the prototype" trick requires you supply the header you claim it should exist in; curl isn't providing any headers, which means the check is erroneous as it cannot be used to "check deeper".

bagder (approved):
It seems I added this in 70b80b0160 (March 2003)

bagder:
thanks!

PR #13838, opened 2024-05-31
mbedtls v3.6.0, workarounds

icing:
- add special sauce to disable unwanted peer verification by mbedtls when negotiating TLS v1.3
- add special sauce for MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET return code on *writing* TLS data. We assume the data had not been written and EAGAIN.
- return correct Curl error code when peer verification failed.
- disable test_08_05 with 50 HTTP/1.1 connections, as mbedtls reports a memory allocation failed during handshake.
- bump CI mbedtls version to 3.6.0

### DIFF
--- a/tests/http/test_17_ssl_use.py
+++ b/tests/http/test_17_ssl_use.py
@@ -77,9 +77,8 @@ class TestSSLUse:
             exp_resumed = 'Initial'  # rustls does not support sessions, TODO
         if env.curl_uses_lib('bearssl') and tls_max == '1.3':
             pytest.skip('BearSSL does not support TLSv1.3')
-        if env.curl_uses_lib('mbedtls') and tls_max == '1.3' and \
-                not env.curl_lib_version_at_least('mbedtls', '3.6.0'):
-            pytest.skip('mbedtls TLSv1.3 support requires at least 3.6.0')
+        if env.curl_uses_lib('mbedtls') and tls_max == '1.3':
+            pytest.skip('mbedtls TLSv1.3 session resume not working in 3.6.0')
 
         curl = CurlClient(env=env)
         # tell the server to close the connection after each request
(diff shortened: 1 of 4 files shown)


MAntoniak:
I am currently on vacation and. In my case, I always verify the other side. I don't remember what the implementation difference is between tls1.2 and tls 1.3 in the mbedtls library. I would have to analyze the peer verification in both cases.

bagder (approved):

tests/http/test_17_ssl_use.py
```diff
@@ -77,9 +77,8 @@ def test_17_02_sslinfo_reconnect(self, env: Env, httpd, nghttpx, tls_max, repeat
             exp_resumed = 'Initial'  # rustls does not support sessions, TODO
         if env.curl_uses_lib('bearssl') and tls_max == '1.3':
             pytest.skip('BearSSL does not support TLSv1.3')
-        if env.curl_uses_lib('mbedtls') and tls_max == '1.3' and \
-                not env.curl_lib_version_at_least('mbedtls', '3.6.0'):
-            pytest.skip('mbedtls TLSv1.3 support requires at least 3.6.0')
+        if env.curl_uses_lib('mbedtls') and tls_max == '1.3':
+            pytest.skip('mbedtls TLSv1.3 session resume not working in 3.6.0')
```
jan2000, line 81: I think this description needs clarification/alteration... The version check is removed, but the description still mentions version 3.6.0. And more iterestingly, is tls1.3 session resumption not working in mbedtls? Or does libcurl not yet support mbedtls's tls1.3 session resumption?

icing: curl supports session setting in mbedtls, but our test server reports that resumption has not happened. That is why I excluded this test here.

If that is a shortcoming in mbedtls or if we miss something special for TLSv1.3 and mbedtls here remains to be investigated.

bagder:
While this is a work-around for what we consider to be an mbedTLS bug, it has not been clearly communicated by mbedTLS that they agree with our view. This is also a fine work-around in the way it is implemented and it should not cause a problem even if mbedTLS decides to fix this bug in coming releases. Thus, I agree with this and will move forward and merge this. It will help users who want to run curl with mbedTLS.

PR #13841, opened 2024-05-31
GHA: Unify http3 workflows into one

tatsuhiro-t:
This commit unifies the following http3 workflows into http3-linux.yml:

- ngtcp2-linux.yml
- osslq-linux.yml
- quiche-linux.yml

The idea is better use of the build cache.  Previously, they independently create caches with the same key.  Some of the caches include source code and intermediate object files, which makes cache quite large.  In this commit, only built artifacts are cached, which drastically reduces the cache size.  OpenSSL v3, mod_h2 and quiche caches still include all stuff, but they are left for the later improvement.  Because the contents of the cache have been changed, the cache keys are also changed to include the word "http3".


icing:
cc @MAntoniak, in case you're interested.

MAntoniak:
> cc @MAntoniak, in case you're interested.

Hmm, not really. I think you mean #13838 :)

bagder (approved):

bagder:
Thanks!

Issue #13532, opened 2024-05-04
Compiling "threaded-ssl.c" with PThreads4W


gvanem:
### I did this

Comping `docs/examples/threaded-ssl.c` using *POSIX threads for Windows* or
[PThreads4W](https://sourceforge.net/projects/pthreads4w/), gives me this compile error (from clang-cl):
```c
threaded-ssl.c(68,24): error: operand of type 'pthread_t' (aka '__ptw32_handle_t') 
where arithmetic or pointer type is required
   68 |   ret = (unsigned long)pthread_self();
      |                        ^~~~~~~~~~~~~~
```

This is caused by `pthread_t` which is a structure on *Pthreads4W*. 

But with this patch, it compiles:
```diff
--- a/docs/examples/threaded-ssl.c 2023-01-05 13:01:10
+++ b/docs/examples/threaded-ssl.c 2024-05-04 09:18:18
@@ -65,7 +65,11 @@
 {
   unsigned long ret;

+#if defined(__PTW32_H)   /* Using 'Pthreads for Windows' */
+  ret = *(unsigned long*) (pthread_self().p);
+#else
   ret = (unsigned long)pthread_self();
+#endif
   return ret;
 }
```

And the program works.

### I expected the following

A successful compile. 

### curl/libcurl version

```
WARNING: this libcurl is Debug-enabled, do not use in production

curl 8.8.0-DEV (x86_64-pc-win32) libcurl/8.8.0-DEV (mbedTLS/3.6.0) (OpenSSL/3.3.0) Schannel (BearSSL) 
(rustls-ffi/0.13.0/rustls/0.23.4) zlib/1.3 brotli/1.1.0 zstd/1.5.5 c-ares/1.28.1 WinIDN libssh2/1.10.1_DEV nghttp2/1.61.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp 
smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli Debug HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL 
NTLM SPNEGO SSL SSPI threadsafe TLS-SRP TrackMemory UnixSockets zstd
```

### operating system

Win-10.

bagder:
I think that proposal introduces a hideous piece of code I rather not have in an example we provide showing off libcurl with.

[...]

PR #13850, opened 2024-06-01
examples/threaded-ssl: explain pthread thread ID method will vary

jay:
- Add a comment explaining that pthread_self(), which returns pthread_t, is not necessarily an integer thread ID.

pthread_t type varies depending on the pthread library. For some pthread libraries, such as PThreads4W, it is a struct and not integer thread ID.

Rather than update the example to handle every pthread implementation of thread ID we now explain that it varies depending on the library.

Bug: https://github.com/curl/curl/issues/13532
Reported-by: Gisle Vanem

Closes #xxxx

### DIFF
--- a/docs/examples/threaded-ssl.c
+++ b/docs/examples/threaded-ssl.c
@@ -65,6 +65,8 @@ static unsigned long thread_id(void)
 {
   unsigned long ret;
 
+  /* pthreads_self() returns a pthread_t that for most pthreads libraries is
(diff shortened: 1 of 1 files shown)


jay:
AFAICS CRYPTO_set_id_callback looks deprecated and some of the code in the example may not be necessary for any recent OpenSSL

jay:
Closing in favor of #13851

PR #13857, opened 2024-06-02
x509asn1: add some common ECDSA OIDs

Bo98:
`ecdsa-with-SHA256` and `ecdsa-with-SHA384` are used by GitHub.com's certificate chain

See #13845


bagder:
Do we actually need this once we merge #13858?

Bo98:
It's not necessary in order to fix #13845, but I think it still has some value in that it slightly improves the output of `curl -w '%{certs}'`.

Before (assuming #13858 already is applied):

```
Signature Algorithm:1.2.840.10045.4.3.2
```

After:

```
Signature Algorithm:ecdsa-with-SHA256
```

Which is consistent with what we already do for RSA certs:

```
Signature Algorithm:sha384WithRSAEncryption
```

and with how it is displayed for other SSL backends that do OID mapping themselves.

So I'd see #13858 as the real bugfix for #13855 and this PR as a cosmetic enhancement for ECDSA certs.

bagder:
Lovely, thanks!

PR #13868, opened 2024-06-03
tidy-up: add `curl_off_t` casts

vszakats:
Cherry-picked from #13489
Closes #13868

### DIFF
--- a/lib/c-hyper.c
+++ b/lib/c-hyper.c
@@ -670,7 +670,7 @@ static int uploadstreamed(void *userdata, hyper_context *ctx,
     }
     /* increasing the writebytecount here is a little premature but we
        don't know exactly when the body is sent */
-    data->req.writebytecount += fillcount;
+    data->req.writebytecount += (curl_off_t)fillcount;
     Curl_pgrsSetUploadCounter(data, data->req.writebytecount);
     rc = HYPER_POLL_READY;
   }
--- a/lib/file.c
+++ b/lib/file.c
@@ -370,7 +370,7 @@ static CURLcode file_upload(struct Curl_easy *data)
     /* skip bytes before resume point */
     if(data->state.resume_from) {
       if((curl_off_t)nread <= data->state.resume_from) {
-        data->state.resume_from -= nread;
+        data->state.resume_from -= (curl_off_t)nread;
         nread = 0;
         sendbuf = xfer_ulbuf;
       }
@@ -390,7 +390,7 @@ static CURLcode file_upload(struct Curl_easy *data)
       break;
     }
 
-    bytecount += nread;
+    bytecount += (curl_off_t)nread;
 
     Curl_pgrsSetUploadCounter(data, bytecount);
 
--- a/lib/formdata.c
+++ b/lib/formdata.c
@@ -318,7 +318,8 @@ CURLFORMcode FormAdd(struct curl_httppost **httppost,
       break;
     case CURLFORM_CONTENTSLENGTH:
       current_form->contentslength =
-        array_state?(size_t)array_value:(size_t)va_arg(params, long);
+        array_state?(curl_off_t)(size_t)array_value:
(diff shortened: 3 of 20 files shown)


bagder:
Do you feel that these typedefs add value to the source code? I feel the opposite. They make the code harder to read.

The question is then if they add value for the safety or correctness of the functionality so that it is worth sacrificing readability. Are any of these typecasts actually fixing a bug other than a compiler warning if you ask the compiler to warn for this?

vszakats:
I'm a 100% convinced about this one either, that's the reason for the separate PR.

Ultimately the riddle I'm trying to solve is how to reach the point in curl, that we can enable `-Wsign-conversion`, assuming this is a useful warning to have, and so far nobody commented otherwise. It's on the shortlist of warnings OpenSSF keeps recommending.

If not for these casts, I wonder how to solve these cleanly?

bagder:
I would not be opposed to document how we disagree with the use of this warning in spite of OpenSSF flogging it, on the basis that it actually does not save us from bugs to any measurable degree but instead makes the code harder to read and follow (and annoying to write) which in itself are also bad properties for source code.

Are other prominent C projects actually using this option? I question that. I question the basis and the experiences that backed up this option getting added to the OpenSSF list.

I would probably argue for this option myself if I did not see the trickling effects it causes and the amount of typecasts we need to do because of it. A good theory that is proven difficult in practice. Maybe even downright counter-productive.

@icing @dfandrich @danielgustafsson thoughts?

vszakats:
For this specific issue curl already has a dedicated wrapper function called `curlx_uztoso()`. It is already used elsewhere in the code to mitigate this same problem.

curl has a complete set of similar wrappers in `lib/warnless.c` and `lib/warnless.h` to silence other type conversions. Their names indeed make the code less readable.

Avoiding casts is the best of course, but IMO a `(curl_off_t)` cast is a better way to tackle this than the `warnless.c` way. Though just having `warnless.c` suggests that this problem was once in the past important enough to trigger a solution.

The warnless method has the advantage to make these conversion points easily greppable, which helps long term in trying to drop them all in favour of cast-less solutions.

At the end the goal would be to avoid buffer under/overruns when converting between signed and unsigned values. One style suggests that using signed everywhere is the way to go. I don't know, but the goal sounds reasonable to me.

danielgustafsson:
I think my main concern with getting into a habit of applying such casts would be that we will create a subtle bug at some point due a cast made more from muscle-memory than careful thought. Basically the workflow of: 1. write code; 2. get a compiler warning for sign conversion; 3. slap a typecast in there to silence compiler; 4. Re-compile; 5. Submit PR. What if step 3 was a legit warning for buggy code which just buggy in a different way now which looks right (with our typecast aligned glasses).

@vszakats boiled it down well in a comment here:

> At the end the goal would be to avoid buffer under/overruns when converting between signed and unsigned values.

I wonder if there is a better way, is there a clean easy-to-use internal API we can devise which will ensure over/underflow safe arithmetic?

bagder:
These new casts now, instead of causing warnings with compiler flags we don't like, cause occasional static code analyzer warnings. 72abf7c13a479edcd triggered at least three complaints from CodeSonar. (whose warnings unfortunately still are only available to me).

I do not think we should enable this warning and we should not push for these typecasts. They don't help us.

vszakats:
It's not a fight I'm ready to die for, but I find the outcome confusing: curl's codebase already uses a lot of casts, including `size_t` and `curl_off_t` the ones. Then there is the `warnless.c` hack, of which the purpose is now unclear.

The unsettling part is that this class of problem looks like a real one, not something imaginery, or one that the curl codebase is automatically protected from.

I'm happy to look at the 3 CodeSonar fallouts. Though 72abf7c13a479edcde80afa60faad3f35f672c0b is tackling the same kind of issues and the preceding 4 (?) similar commits, but this time for `lib`.

Either way, please revert if it causes too much trouble.

PR #13874, opened 2024-06-04
socketpair: add `eventfd` and use `SOCK_NONBLOCK` for `socketpair()`

panjf2000:
Currently, we use `pipe` for `wakeup_create`, which requires ***two*** file descriptors. Furthermore, given its complexity inside, `pipe` is a bit heavyweight for just a simple event wait/notify mechanism.

`eventfd` would be a more suitable solution for this kind of scenario, kernel also advocates for developers to use `eventfd` instead of `pipe` in some simple use cases:
> Applications can use an eventfd file descriptor instead of a pipe
   (see pipe(2) in all cases where a pipe is used simply to signal
    events.  The kernel overhead of an eventfd file descriptor is
    much lower than that of a pipe, and only one file descriptor is
    required (versus the two required for a pipe).

This PR adds the new backend of `eventfd` for `wakeup_create` and uses it where available, eliminating the overhead of `pipe`. Also, it optimizes the `wakeup_create` to eliminate the system calls that make file descriptors non-blocking by moving the logic of setting non-blocking flags on file descriptors to `socketpair.c` and using `SOCK_NONBLOCK` for `socketpair(2)`, `EFD_NONBLOCK` for `eventfd(2)`.

Ref:
https://man7.org/linux/man-pages/man7/pipe.7.html
https://man7.org/linux/man-pages/man2/eventfd.2.html
https://man7.org/linux/man-pages/man2/socketpair.2.html


lib/socketpair.h
```diff
@@ -26,19 +26,35 @@
...
+    defined(__x86_64__) && \
+    defined(__aarch64__) && \
+    defined(__ia64__) && \
+    defined(__ppc64__) && \
+    defined(__mips64) && \
+    defined(__sparc64__) && \
+    defined(__riscv_64e) && \
+    defined(__s390x__)
+
+/* Use eventfd only on 64-bit CPU architectures */
```
bagder, line 39: I'm not against this but I'm curious: why 64 bit only?

panjf2000: `eventfd` has a stringent policy of requiring the 8-byte buffer when calling `read(2)` and `write(2)` on it. In some rare cases, the C standard library implementation on a 32-bit system might choose to define uint64_t as a 32-bit type for various reasons (memory limitations, compatibility with older code), which makes `eventfd` broken.

Ref: https://github.com/curl/curl/actions/runs/9362009840/job/25769979500

lib/asyn-thread.c
```diff
@@ -302,6 +302,14 @@ query_complete(DWORD err, DWORD bytes, LPWSAOVERLAPPED overlapped)
   struct Curl_addrinfo *ca;
   struct Curl_addrinfo *cafirst = NULL;
   struct Curl_addrinfo *calast = NULL;
+#ifndef CURL_DISABLE_SOCKETPAIR
+#ifdef USE_EVENTFD
+  const void *buf;
+  static const uint64_t val = 1;
```
bagder, line 308: Why static? Won't having this static risk causing problems in multi-threaded situations?

lib/asyn-thread.c
```diff
@@ -454,7 +465,12 @@ static unsigned int CURL_STDCALL getaddrinfo_thread(void *arg)
   char service[12];
   int rc;
 #ifndef CURL_DISABLE_SOCKETPAIR
+#ifdef USE_EVENTFD
+  const void *buf;
+  static const uint64_t val = 1;
```
bagder, line 470: Same comment about static here.

lib/multi.c
```diff
@@ -1639,6 +1632,15 @@ CURLMcode curl_multi_wakeup(struct Curl_multi *multi)
      it has to be careful only to access parts of the
      Curl_multi struct that are constant */
 
+#if defined(ENABLE_WAKEUP) && !defined(USE_WINSOCK)
+#ifdef USE_EVENTFD
+  const void *buf;
+  static const uint64_t val = 1;
```
bagder, line 1638: and here

lib/asyn-thread.c
```diff
@@ -251,7 +251,7 @@ int init_thread_sync_data(struct thread_data *td,
 
 #ifndef CURL_DISABLE_SOCKETPAIR
   /* create socket pair or pipe */
-  if(wakeup_create(&tsd->sock_pair[0]) < 0) {
+  if(wakeup_create(tsd->sock_pair, FALSE) < 0) {
```
bagder, line 254: Is there a reason why this socketpair is not set nonblocking?

lib/asyn-thread.c
panjf2000: it's read-only, so it should be fine.

lib/asyn-thread.c
bagder: I was just looking for a reason for why `wakeup_create()` would need that second argument for non-blocking or not, and if we instead assume always non-blocking things get a little bit simpler. Right?

lib/asyn-thread.c
panjf2000: When there are multiple callers, sharing the variable should be a good thing in general, right? Given that these functions can get called a lot.

lib/asyn-thread.c
panjf2000: Yes, that would be more simpler. But I'm not sure someday someone may want to create a blocking wakeup_pair. So maybe keep this flexibility might help in the future.

lib/asyn-thread.c
bagder: Is it? If it just exists on the stack it might be much cheaper than having it static - and either way, I doubt you could even measure a performance difference if you tried. I actually don't know the exact winner here but I know that using `static` is a red flag in code in general and it makes the code stand out less if you avoid it.

lib/asyn-thread.c
bagder: I disagree. Providing flexibility today for something "maybe in the future" is a bad idea. You should write exactly and only for this current use case. In the future more things have changed anyway.

panjf2000: True, if we can make it simple for the current situation, it's worth doing it. But the current usage of the blocking pair seems like a real thing, they don't handle the `EAGAIN` error and retry. So I think blocking pair is indeed required by those callers.

lib/asyn-thread.c
panjf2000: Convinced, we will go with the stack memory.

lib/asyn-thread.c
bagder: That is a valid reason for using blocking indeed. The question is perhaps is if it matters. It writes a single byte to the freshly created descriptor that then gets closed. Can it actually do EAGAIN in practice?

But yes, I suppose there are systems or theoretical models where it can happen. You are right. Let's do it blocking for this case, as it matches how it works for the current use case when `pipe() is used.

panjf2000: So we can leave it unchanged here and move forward?

bagder: yes!

bagder (approved):

MarcelRaad:
Unfortunately, this breaks `--disable-socketpair` / `CURL_DISABLE_SOCKETPAIR`, and there's a `curl_socket_t` to `int` narrowing conversion warning in 64-bit MinGW builds:
https://curl.se/dev/log.cgi?id=20240605035856-3529577
I wonder why the latter didn't show up in the GitHub CI.

panjf2000:
How did it break? What's the error info?

panjf2000:
I'm investigating this.

bagder:
> Unfortunately, this breaks `--disable-socketpair` / `CURL_DISABLE_SOCKETPAIR`

Is this breakage perhaps specific to Windows? I built `./configure --disable-socketpair` but I can't spot any problems...

panjf2000:
> > Unfortunately, this breaks `--disable-socketpair` / `CURL_DISABLE_SOCKETPAIR`
> 
> Is this breakage perhaps specific to Windows? I built `./configure --disable-socketpair` but I can't spot any problems...

I've run `./configure --disable-socketpair --with-openssl && make && make test` on my Linux server, and it passed all tests without reporting any errors. I assume that issue is MinGW-specific.

PR #13875, opened 2024-06-04
test: fix the -Werror=maybe-uninitialized for logmsg

panjf2000:
Fixes the compiler error caused by bc21c50

Ref: https://github.com/curl/curl/actions/runs/9361428368/job/25768434305?pr=13874

### DIFF
--- a/tests/ftpserver.pl
+++ b/tests/ftpserver.pl
@@ -690,8 +690,9 @@ sub disc_handshake {
                 $size = hex($1);
             }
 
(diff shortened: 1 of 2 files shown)


dfandrich:
I fixed it just before spotting this PR. Since `read_data_block()` already logs the # bytes, I went with the simple approach of just dropping the size in the initial log message.

panjf2000:
Good to know.

Issue #10318, opened 2023-01-18
CURLOPT_TCP_KEEPCNT is absent


vrubleg:
There is `CURLOPT_TCP_KEEPIDLE` for setting `TCP_KEEPIDLE`, there is `CURLOPT_TCP_KEEPINTVL` for setting `TCP_KEEPINTVL`, but there is no `CURLOPT_TCP_KEEPCNT` for setting `TCP_KEEPCNT` that is strongly tied to the other two socket settings and all of them should be set at the same time. All these three socket settings are available on Linux, MacOS, and Windows 10+.

bagder:
Thanks, but this description sounds as if you're asking for a new feature/change. We use this tracker for bugs and issues only, we put ideas to work on in the future in the TODO document. We basically drown in good ideas so they don't do much use in our tracker.

If you really want to see this happen, start working on an implementation and submit a PR for it or join the mailing list and talk up more interest for it and see what help from others you can get!

Issue #8564, opened 2022-03-09
SSL_read won't return after disconnect network during downloading a big file on ios


TXTT2016:
### I did this

I have found just on IOS platform, SSL_read function won't return until 10mins timeout when you are downloading a big file. But it will return quickly on windows and androids.

### I expected the following

Is it possible that all platforms have the same behavior？

### curl/libcurl version

7.80.0

[curl -V output]

dfandrich:
Is this the issue you're seeing? 
https://curl.se/docs/faq.html#Why_does_not_curl_return_an_erro

TXTT2016:
> Is this the issue you're seeing? https://curl.se/docs/faq.html#Why_does_not_curl_return_an_erro

Thank you for reply !

I did not use the --speed-limit、--speed-time or --max-time, but set the keepalive-time.

just like this for all connections:

```
    /* enable TCP keep-alive for this transfer */
    VOID_CURL_METHOD(curl_easy_setopt, m_curlHandle.get(), CURLOPT_TCP_KEEPALIVE, 1L);
    /* keep-alive idle time to 60 seconds */
    VOID_CURL_METHOD(curl_easy_setopt, m_curlHandle.get(), CURLOPT_TCP_KEEPIDLE, 60L);
    /* interval time between keep-alive probes: 60 seconds */
    VOID_CURL_METHOD(curl_easy_setopt, m_curlHandle.get(), CURLOPT_TCP_KEEPINTVL, 60L);
```

Maybe Unplugging a cable is not an error situation exactly, Is this just a platform differences for the download connection ?

bagder:
That sounds like a problem with the underlying TCP and TLS stacks rather than actual curl code. 

Maybe 10 minutes is too short to detect it if the number of probes is 9, which seems to be default on some systems?

 TCP_KEEPCNT: The number of unanswered probes required to force closure of the socket

jay:
SSL_read the way we use it is supposed to be nonblocking so I'm not sure how this is a curl problem

TXTT2016:
> That sounds like a problem with the underlying TCP and TLS stacks rather than actual curl code.
> 
> Maybe 10 minutes is too short to detect it if the number of probes is 9, which seems to be default on some systems?
> 
> TCP_KEEPCNT: The number of unanswered probes required to force closure of the socket

Thanks for reply !

It seems you are right, maybe this is just a platform differences with the underlying TCP and TLS stacks.

[...]

PR #13885, opened 2024-06-05
tcpkeepalive: add `CURLOPT_TCP_KEEPCNT` for `libcurl` and `keepalive-cnt` for `curl`

panjf2000:
Customizing `TCP_KEEPIDLE` and `TCP_KEEPINTVL` for `curl` has been supported since v7.18.0, with `TCP_KEEPCNT` missing all the time. `TCP_KEEPIDLE`, `TCP_KEEPINTVL`, and `TCP_KEEPCNT` are not included in the POSIX standard for system call `setsockopts()`, yet while these three socket options are widely available on most Unix-like systems and Windows.

I reckon that `curl` should support setting `TCP_KEEPCNT` and use a default value if unspecified across all platforms based off of the following justification:

- The system default value of `TCP_KEEPCNT` varies across different platforms that support this option: the maximum number of keep-alive probes sent before dropping the connection is 8 on *BSD/macOS/AIX, 9 on Linux/AIX(old), 5/10 on Windows (5 on Windows Server 2003, Windows XP, and Windows 2000, 10 on Windows Vista and later), or even more unorthodox, undefined default value on Solaris/illumos, and so on, which brings forth the inconsistency of `curl` on various platforms when it comes to the behaviors of TCP keep-alive mechanism. Setting up a default value for `TCP_KEEPCNT` regardless of specific platforms will consolidate the behaviors of TCP keep-alive.

- The requirement of precisely tuning the TCP keep-alive behaviors for some purposes: 1) aggressive default values of TCP keep-alive can result in excessive CPU consumption and drain the battery fast if it's on mobile devices (that are tolerant with dead connections), people can alleviate it by tuning up `TCP_KEEPINTVL` and tuning down `TCP_KEEPCNT`, 2) on the contrary, for some applications that are aliveness-sensitive about the remote peers, people need to tune down both the `TCP_KEEPINTVL` and `TCP_KEEPCNT` to detect the dead peers ASAP.

Thus, the support of the full TCP keep-alive mechanism (all three options) enables users to take control of the TCP keep-alive mechanism more finely and build more superior applications based on `curl`/`liburl`.

## References

- [AIX](https://www.ibm.com/support/pages/ibm-aix-tcp-keepalive-cnt)
- [DragonflyBSD](https://man.dragonflybsd.org/?command=tcp&section=4)
- [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=tcp)
- [HP-UX](https://docstore.mik.ua/manuals/hp-ux/en/B2355-60130/TCP.7P.html)
- [illumos](https://illumos.org/man/4P/tcp)
- [Linux](https://man7.org/linux/man-pages/man7/tcp.7.html)
- [NetBSD](https://man.netbsd.org/NetBSD-8.0/tcp.4)
- [Solaris 11.3](https://docs.oracle.com/cd/E86824_01/html/E54777/tcp-7p.html)
- [Solaris 11.4](https://docs.oracle.com/cd/E88353_01/html/E37851/tcp-4p.html)
- [Windows](https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-tcp-socket-options)

### Mac OS

In earlier versions, macOS only supported setting `TCP_KEEPALIVE` (the equivalent of `TCP_KEEPIDLE` on other platforms), but since macOS 10.8 it has supported `TCP_KEEPINTVL` and `TCP_KEEPCNT`.

Check out [this mailing list](https://lists.apple.com/archives/macnetworkprog/2012/Jul/msg00005.html) and [the source code](https://github.com/apple/darwin-xnu/blob/main/bsd/netinet/tcp.h#L215-L230) for more details.

### Solaris

Solaris claimed it supported the TCP keep-alive mechanism, but `TCP_KEEPIDLE`, `TCP_KEEPINTVL`, and `TCP_KEEPCNT` were not available on Solaris until the latest version 11.4. Therefore, we need to simulate the TCP keep-alive mechanism on other platforms via `TCP_KEEPALIVE_THRESHOLD` + `TCP_KEEPALIVE_ABORT_THRESHOLD`.

### Relevant issues
#8564
#10318


panjf2000:
Hah, this [CI failure](https://github.com/curl/curl/actions/runs/9379581567/job/25824821802?pr=13885) seems to be caused by [this change](https://github.com/curl/curl/pull/13885/files#diff-582f4a1855e51a0a30a44ce62d8d65511b7007077863cff377e7af6d04304ef8)? I'm not sure, I wonder if changing the option from 1 to 9 or 10 would fix it.

bagder:
> Hah, this [CI failure](https://github.com/curl/curl/actions/runs/9379581567/job/25824821802?pr=13885) seems to be caused by [this change](https://github.com/curl/curl/pull/13885/files#diff-582f4a1855e51a0a30a44ce62d8d65511b7007077863cff377e7af6d04304ef8)? 

Not necessarily, it might also just be a flaky test/build...

dfandrich:
My question would be when users run curl on a new platform, should they expect
to see platform default behaviour or curl default behaviour? I would venture
that platform default behavior is more to be expected.  These are low-level
details that are expected to be different on different OSes; that's what makes
different OSes different, after all. Or, from another perspective, why stop at
trying to make TCP keepalive behaviour the same across platforms? Why not make
ephemeral port ranges the same, or socket buffer sizes the same, etc., etc.?

panjf2000:
> My question would be when users run curl on a new platform, should they expect
to see platform default behaviour or curl default behaviour? I would venture
that platform default behavior is more to be expected.  These are low-level
details that are expected to be different on different OSes; that's what makes
different OSes different, after all.

Not all system default behaviors are rationale, sometimes they're even lame. Take the TCP keep-alive for instance, most of the platforms use 2 hours as the default value of `TCP_KEEPIDLE`, which makes this whole mechanism useless because it would take you over 2 hours to get notified that an idle connection was dead. That's why most of the libraries that utilize TCP keep-alive don't use its default values defined by the OSs, but change those values of options accordingly to make it actually pragmatic. And that's also what `curl` has been doing: set `TCP_KEEPIDLE` and `TCP_KEEPINTVL` to 60s by default when keep-alive is enabled but these two options weren't specified.

> why stop at
trying to make TCP keepalive behaviour the same across platforms? Why not make
ephemeral port ranges the same, or socket buffer sizes the same, etc., etc.?

We should only try to consolidate the ***behaviors*** for libraries that aim to be cross-platform and portable (and I believe this has also been `curl`'s goal for years, maybe from the day it was created). We don't compulsively need to line up all system configurations under the hood when they don't have an evident impact on the ***behaviors*** of applications, like the socket buffer you referred to, different default values for socket buffers on various platforms rarely led people to spot the different behaviors during networking, moreover, most of the platforms use the dynamic socket buffers instead of static ones, which makes the consolidation of that redundant. But for TCP keep-alive options, people will effectively see the consequence due to the inconsistent default values across platforms, because it's a time-sensitive mechanism.

panjf2000:
When can we get the review started? Would you take a glance at this PR? @bagder

bagder (approved):

bagder:
Thanks!

PR #13888, opened 2024-06-05
socketpair: omit the function definition on Windows for disable-socketpair

panjf2000:
Updates #13874

Ref: https://curl.se/dev/log.cgi?id=20240605035856-3529577

### DIFF
--- a/lib/socketpair.c
+++ b/lib/socketpair.c
@@ -75,7 +75,29 @@ int Curl_pipe(curl_socket_t socks[2], bool nonblocking)
 #endif
 
 
-#if !defined(HAVE_SOCKETPAIR) && !defined(CURL_DISABLE_SOCKETPAIR)
+#ifndef CURL_DISABLE_SOCKETPAIR
+#ifdef HAVE_SOCKETPAIR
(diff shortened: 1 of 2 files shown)


lib/socketpair.c
```diff
@@ -239,6 +239,7 @@ int Curl_socketpair(int domain, int type, int protocol,
   return -1;
 }
 #else
+#ifndef _WIN32
```
MarcelRaad, line 242: Shouldn't this code block be dependent on `defined(HAVE_SOCKETPAIR) && !defined(CURL_DISABLE_SOCKETPAIR)` instead?

I think `HAVE_SOCKETPAIR` must be true as `socketpair` is used here, and `Curl_socketpair` must only be defined if `CURL_DISABLE_SOCKETPAIR` is not defined.

panjf2000: You're right! I checked out the code elsewhere, it seemed pointless to define this function with `CURL_DISABLE_SOCKETPAIR` specified.

MarcelRaad (approved):
Looks good to me and works for me, thanks! 👍

bagder:
Thanks!

PR #13892, opened 2024-06-05
GHA: try energy estimation

vszakats:


vszakats:
Measurement results are difficult to read and it didn't sparked much interest.

PR #13893, opened 2024-06-05
connection shutdown, basics and first use

icing:
This adds connection shutdown infrastructure and first use for FTP in combination with OpenSSL. FTP data connections, when not encountering an error, are now shut down in a blocking way with a 2sec timeout.

- add cfilter `Curl_cft_shutdown` callback
- keep a shutdown start timestamp and timeout at connectdata
- provide shutdown timeout default and member in `data->set.shutdowntimeout`.
- provide methods for starting, interrogating and clearing shutdown timers
- provide `Curl_conn_shutdown_blocking()` to shutdown the `sockindex` filter chain in a blocking way. Use that in FTP.
- add `Curl_conn_cf_poll()` to wait for socket events during shutdown of a connection filter chain. This gets the monitoring sockets and events via the filters "adjust_pollset()" methods. This gives correct behaviour when shutting down a TLS connection through a HTTP/2 proxy.
- Implement shutdown for socket, http/2 and openssl filters
- add shutdown forwarding to happy eyeballers and https connect ballers when applicable.
- have dummy implementation for other TLS backends until we get to implement them.

Update: added shutdown support in gnutls, bearssl, mbedtls, wolfssl and rustls.
Update: added shutdown support in schannel

### DIFF
--- a/lib/cf-h1-proxy.c
+++ b/lib/cf-h1-proxy.c
@@ -1077,6 +1077,7 @@ struct Curl_cftype Curl_cft_h1_proxy = {
   cf_h1_proxy_destroy,
   cf_h1_proxy_connect,
   cf_h1_proxy_close,
+  Curl_cf_def_shutdown,
   Curl_cf_http_proxy_get_host,
   cf_h1_proxy_adjust_pollset,
   Curl_cf_def_data_pending,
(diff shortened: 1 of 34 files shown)


icing:
Closed as replaced by #13904.

PR #13902, opened 2024-06-06
noproxy: test bad ipv6 net size first

bagder:
No need to parse anything if the size is odd.

Added some tests to this effect to test 1614.


bagder:
Good point, will amend.

PR #13903, opened 2024-06-06
build: keep `-Wsign-conversion` disabled

vszakats:
This warning is enabled by `-Wconversion`. Disable it explicitly to keep
builds free of these warnings.

Also delete the preparation steps made earlier to leave it enabled in
one of the CI jobs and provisions to leave it enabled by default for all
builds.

Ref: 2dbe75bd7f3c36837aa06fd87a442bdf3fb7faef #12492 (partial revert)
Closes #13903


vszakats:
Okay, so `-Wsign-conversion` comes by default with the `-Wconversion` option. Has to be disabled explicitly.

Issue #13556, opened 2024-05-07
FTP upload with TLS 1.3 results in 0 byte file on the server, eventually times out


blach:
### I did this

I'm using the following command line to upload the file "test.html" with a size of 18815 bytes to a server using FTP with SSL encryption:

`curl --ssl-reqd ftp://<redacted>.kasserver.com/test/ -u <redacted> --upload-file test.html`

This command **results in a 0 byte file on the server** instead of the expected file with 18815 bytes.

The transfer progress looks like this:

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 18815    0     0  100 18815      0    311  0:01:00  0:01:00 --:--:--     0
curl: (28) FTP response timeout
```

It claims that 18815 bytes are transferred immediately, but then nothing happens and the upload times out after 1 minute.

When I add `--tls-max 1.2`, the upload works immediately and the file on the server has the correct size and contents.

This is the FTP server of the popular German hoster https://all-inkl.com/

### I expected the following

I expected the file to upload successfully.

### curl/libcurl version

curl 8.7.1 (aarch64-apple-darwin23.4.0) libcurl/8.7.1 (SecureTransport) OpenSSL/3.3.0 zlib/1.2.12 brotli/1.1.0 zstd/1.5.6 libidn2/2.3.7 libssh2/1.11.0 nghttp2/1.61.0 librtmp/2.3 OpenLDAP/2.6.7
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

macOS Sonoma 14.4.1

blach:
This is the verbose output of the failing FTP upload command that eventually times out and results in a 0 byte file on the server:

```
* Connected to <redacted>.kasserver.com (85.13.<redacted>) port 21
< 220 FTP on <redacted>.kasserver.com ready
> AUTH SSL
< 234 AUTH SSL successful
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
[...]

PR #13904, opened 2024-06-07
connection shutdown, basics and first use

icing:
This adds connection shutdown infrastructure and first use for FTP. FTP data connections, when not encountering an error, are now shut down in a blocking way with a 2sec timeout.

- add cfilter `Curl_cft_shutdown` callback
- keep a shutdown start timestamp and timeout at connectdata
- provide shutdown timeout default and member in
  `data->set.shutdowntimeout`.
- provide methods for starting, interrogating and clearing shutdown timers
- provide `Curl_conn_shutdown_blocking()` to shutdown the
  `sockindex` filter chain in a blocking way. Use that in FTP.
- add `Curl_conn_cf_poll()` to wait for socket events during
  shutdown of a connection filter chain.
  This gets the monitoring sockets and events via the filters
  "adjust_pollset()" methods. This gives correct behaviour when
  shutting down a TLS connection through a HTTP/2 proxy.
- Implement shutdown for all socket filters
  - for HTTP/2 and h2 proxying to send GOAWAY
  - for TLS backends to the best of their capabilities 
  - for tcp socket filter to make a final, nonblocking receive to avoid unwanted RST states 
  - add shutdown forwarding to happy eyeballers and https connect ballers when applicable.

Test coverage of the FTP blocking shutdown of its data connection is covered in `test_30_5` and `test_31_05` ftp upload tests.

Update: removed known bugs 7.5 and 7.11 fixed by this PR. Verified by local upload tests with `vsftpd` and confirmation on #13556.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1010,6 +1010,30 @@ static void cf_socket_close(struct Curl_cfilter *cf, struct Curl_easy *data)
   cf->connected = FALSE;
 }
 
+static CURLcode cf_socket_shutdown(struct Curl_cfilter *cf,
+                                   struct Curl_easy *data,
+                                   bool *done)
+{
+  if(cf->connected) {
+    struct cf_socket_ctx *ctx = cf->ctx;
+
+    CURL_TRC_CF(data, cf, "cf_socket_shutdown(%" CURL_FORMAT_SOCKET_T
+                ")", ctx->sock);
+    /* On TPC, and when the socket looks well and non-blocking mode
+     * can be enabled, receive dangling bytes before close to avoid
+     * TCP entering RST states unnecessarily. */
+    if(ctx->sock != CURL_SOCKET_BAD &&
+       ctx->transport == TRNSPRT_TCP &&
+       (curlx_nonblock(ctx->sock, TRUE) >= 0)) {
+      unsigned char buf[1024];
+      (void)sread(ctx->sock, buf, sizeof(buf));
+    }
+    cf_socket_close(cf, data);
+  }
+  *done = TRUE;
+  return CURLE_OK;
+}
+
 static void cf_socket_destroy(struct Curl_cfilter *cf, struct Curl_easy *data)
 {
   struct cf_socket_ctx *ctx = cf->ctx;
@@ -1729,6 +1753,7 @@ struct Curl_cftype Curl_cft_tcp = {
   cf_socket_destroy,
   cf_tcp_connect,
   cf_socket_close,
+  cf_socket_shutdown,
   cf_socket_get_host,
   cf_socket_adjust_pollset,
   cf_socket_data_pending,
@@ -1872,6 +1897,7 @@ struct Curl_cftype Curl_cft_udp = {
   cf_socket_destroy,
   cf_udp_connect,
   cf_socket_close,
+  cf_socket_shutdown,
   cf_socket_get_host,
   cf_socket_adjust_pollset,
   cf_socket_data_pending,
@@ -1923,6 +1949,7 @@ struct Curl_cftype Curl_cft_unix = {
   cf_socket_destroy,
   cf_tcp_connect,
   cf_socket_close,
+  cf_socket_shutdown,
   cf_socket_get_host,
   cf_socket_adjust_pollset,
   cf_socket_data_pending,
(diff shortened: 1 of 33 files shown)


lib/cf-socket.c
```diff
@@ -1010,6 +1010,30 @@ static void cf_socket_close(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+static CURLcode cf_socket_shutdown(struct Curl_cfilter *cf,
+                                   struct Curl_easy *data,
+                                   bool *done)
+{
+  if(cf->connected) {
+    struct cf_socket_ctx *ctx = cf->ctx;
+
+    CURL_TRC_CF(data, cf, "cf_socket_shutdown(%" CURL_FORMAT_SOCKET_T
+                ")", ctx->sock);
+    /* On TPC, and when the socket looks well and non-blocking mode
```
vszakats, line 1022:
```suggestion
    /* On TCP, and when the socket looks well and non-blocking mode
```

vszakats (approved):

lib/cf-socket.c
icing: Fixed, thanks!

lib/connect.c
```diff
@@ -142,6 +142,43 @@ timediff_t Curl_timeleft(struct Curl_easy *data,
...
+    now = Curl_now();
+    nowp = &now;
+  }
+  data->conn->shutdown.start[sockindex] = *nowp;
+  data->conn->shutdown.timeout_ms = (data->set.shutdowntimeout > 0) ?
+    data->set.shutdowntimeout : DEFAULT_SHUTDOWN_TIMEOUT_MS;
+}
+
+timediff_t Curl_shutdown_timeleft(struct connectdata *conn, int sockindex,
+                                  struct curltime *nowp)
```
bagder, line 161: This function is used in two places, both of them set `nowp` to NULL. Do you really need this argument?

lib/connect.c
```diff
@@ -142,6 +142,43 @@ timediff_t Curl_timeleft(struct Curl_easy *data,
   return (ctimeleft_ms < timeleft_ms)? ctimeleft_ms : timeleft_ms;
 }
 
+void Curl_shutdown_start(struct Curl_easy *data, int sockindex,
+                         struct curltime *nowp)
```
bagder, line 146: This function is used in two places, both of them set `nowp` to NULL. Do you really need this argument?

lib/cfilters.c
```diff
@@ -168,6 +177,61 @@ void Curl_conn_close(struct Curl_easy *data, int index)
   }
 }
 
+CURLcode Curl_conn_shutdown_blocking(struct Curl_easy *data, int sockindex)
```
bagder, line 180: This risks blocking the multi interface for the entire timeout period (DEFAULT_SHUTDOWN_TIMEOUT_MS) when the FTPS data connection is closed. Do you plan to add handling of this in a non-blocking way in a follow-up change?

bagder (commented):
It looks good. Just some questions!

lib/connect.c
icing: I thought I follow the spirit of `Curl_timeleft()`. Shoudl we remove it there as well?

lib/connect.c
icing: See comment above.

lib/cfilters.c
icing: To have a non-blocking shutdown of the data connection would be nice, indeed. That would require some more radical change in the FTP code, I guess.

For now, the blocking way at least assures that the FTP data connection works. We have multiple reports that at least uploads fail regularly if we do not do this, as FTP servers have become more sensitive to a clean TLS shutdown.

The change in this PR is a bugfix at the expense of parallel performance.

lib/connect.c
bagder: In `Curl_timeleft`'s case there are several callers that use the argument so there it does save a few calls to `Curl_now()`.

lib/cfilters.c
bagder: Yeah. Let's merge this as it fixes issues, and maybe work on improving it further later on.

rmja:
Could it be that this PR breaks FTP uploads with a POSTQUOTE?
With curl 8.9.1 i now get a timeout during the call to `curl_easy_perform()` when I have specified a POSTQUOTE that renames the just uploaded file using RNFR/RNTO.
This happens on a Microsoft FTP Service (which I do not control). This previously worked fine in 8.8.0.

cc @icing

PR #13907, opened 2024-06-07
curl: support VLAN Priority: --vlan-priority

orgads:
Add --vlan-priority option to the command line tool for setting VLAN priority.


orgads:
What's the `hyper` check? It failed and there are no logs whatsoever.

icing:
> What's the `hyper` check? It failed and there are no logs whatsoever.

Looks like github had a problem. When the CI workflow is done, try restarting the failed jobs.

docs/cmdline-opts/vlan-priority.md
```diff
@@ -0,0 +1,21 @@
...
+Multi: single
+See-also:
+  - ip-tos
+Example:
+  - --vlan-priority 4 $URL
+---
+
+# `--vlan-priority`
+
+Set VLAN priority as defined in IEEE 802.1Q. (Added in 8.9.0).
```
bagder, line 19: Maybe mention that this sets a field on Ethernet level and maybe something about the limitations that brings? Ie that this only really works within a local network.

src/tool_operate.c
```diff
@@ -195,6 +196,18 @@ static int sockopt_callback(void *clientp, curl_socket_t curlfd,
...
     }
   }
+#endif
+#ifdef SO_PRIORITY
+  if(config->vlan_priority > 0) {
+    int priority = (int)config->vlan_priority;
+    if(setsockopt(curlfd, SOL_SOCKET, SO_PRIORITY,
+      (const char *)&priority, sizeof(priority)) != 0) {
+      int error = errno;
+      warnf(config->global, "SO_PRIORITY %d failed with errno %d: %s;\n",
```
bagder, line 206: The user will not know what `SO_PRIORITY` is. They used `--vlan-priority`.

orgads: Reworded.

docs/cmdline-opts/vlan-priority.md
orgads: Done

orgads:
These flaky tests are really frustrating...

bagder (approved):

src/tool_operate.c
```diff
@@ -2253,13 +2266,19 @@ static CURLcode single_transfer(struct GlobalConfig *global,
...
 #else
-          warnf(config->global,
-                "Type of service is not supported in this build.");
+          if(config->ip_tos > 0) {
+            warnf(config->global,
+                  "Type of service is not supported in this build.");
+          }
+          if(config->vlan_priority > 0) {
+            warnf(config->global,
+                  "VLAN priority is is not supported in this build.");
```
michaelgruenewald, line 2280: typo:
```suggestion
                  "VLAN priority is not supported in this build.");
```

orgads: Nice catch! Fixed.

orgads:
I moved the "unsupported" handling to the callback (in a separate commit). If you like it, you can leave it or squash. Otherwise, I'll discard this commit (probably tomorrow night).

bagder:
> I moved the "unsupported" handling to the callback (in a separate commit). If you like it, you can leave it or squash. Otherwise, I'll discard this commit (probably tomorrow night).

I prefer to detect the problem in the earlier stage rather than delaying and erroring out after having done more processing, when it is possible. For that reason prefer the previous take a little more.

Also, I think we should use `errorf()` instead of `warnf()` for this since we bail out and return an error for this case.

orgads:
> > I moved the "unsupported" handling to the callback (in a separate commit). If you like it, you can leave it or squash. Otherwise, I'll discard this commit (probably tomorrow night).
> 
> I prefer to detect the problem in the earlier stage rather than delaying and erroring out after having done more processing, when it is possible. For that reason prefer the previous take a little more.
> 
> Also, I think we should use `errorf()` instead of `warnf()` for this since we bail out and return an error for this case.

Done. Better now?

bagder (approved):

bagder:
Thanks!

Cleanfastllc:
Cool

PR #13910, opened 2024-06-07
curl: support -w '%{num_retries}

bagder:
Suggested-by: Jay Guerette
Ref: https://github.com/curl/curl/discussions/13901


jay (approved):
note two retry variables so similar in name `num_retries` and `retry_numretries` different in value might be a little confusing to someone modifying the code in the future. `retry_numretries -> retry_remaining`?

Cleanfastllc (commented):
Cool

PR #13911, opened 2024-06-07
libcurl.pc: add `Requires.private`, `Requires` for static linking

### LINKED ISSUE #864: consider utilizing `Requires.private` directives in `libcurl.pc` instead of `Libs.private`
the curl build system goes through a good amount of effort to try and figure out when some external libs need other external libs, often for static linking purposes.  for example, it probes openssl and then tries to see if openssl also needs zlib and libdl.  this doesn't always work (new libs get added and curl doesn't probe those, or the libs get rebuilt which means curl needs to be rebuilt simply to reprobe the library).

ideally the `libcurl.pc` file, instead of looking like:
```
Libs.private: -lnghttp2 -lssl -lcrypto -lssl -lcrypto -lz
```

it would look like:
```
Libs.private:
Requires.private: libnghttp2 openssl zlib
```

now (1) the data will always be up-to-date regardless of how those other libs were configure/rebuilt/updated, and (2) curl wouldn't have to do all that probing.

NB: (1) can be done now and generally w/out pain, but (2) would mean that curl's build system itself would now rely on pkg-config being available/sane at build time.  not sure if non-pkg-config builds is something curl cares about.

vszakats:
- cmake: populate for dependencies.
- autotools: populate for dependencies.
  (including mbedtls, though the script does not detect
  mbedtls through pkgconfig. mbedtls 3.6.0 now supports it.)

Skip dealing with gssapi in this patch.

Fixes #864
Closes #13911


jay:
The static libraries required for linking sometimes require linking to other libraries which is why this is hard to get right I think. Have you tested that this PR works as you expect?

vszakats:
`Require.private` and `Requires` are listing _pkgconfig_ names (aka `<name>.pc` files), not actual (static) libs. Listing the actual static libs and/or referencing further pkgconfig dependencies is up the referenced `.pc` files. This method makes it more likely to work than listing a raw list of static libs, as the existing `Libs.private` and `Libs` entries do.

Obviously I won't be able to test all these combinations myself, but this is supposed to work if the generated `libcurl.pc` file lists the necessary dependencies accurately. There is not much more curl can do to make this work, unless I'm missing something obvious.

jay:
> `Requires.private` and `Requires` are listing _pkgconfig_ names (aka `<name>.pc` files), not actual (static) libs.

Ah that makes sense.

PR #13937, opened 2024-06-13
connections graceful shutdown

icing:
Implement graceful shutdown of connections in libcurl:
- when a connection is ready to be discarded, run a non-blocking shutdown on all connection filters. If that returns errors, discard the connection. If that sucessfully shuts down all filters, discard as well. Otherwise, add it to a "shutdown" list in the transfers connection cache.
- limit the shutdown list to the same maximum length as the connection cache's live limit. When shutdown down a new connection at the limit, the oldest one is discarded.
- In multi_wait() and multi_waitfds(), collect all connection caches involved (each transfer might carry its own) into a temporary list. Let each connection cache on the list contribute sockets and POLLIN/OUT events it's connections are waiting for.
- in multi_perform() collect the connection caches the same way and let them peform their maintenance. This will make another non-blocking attempt to shutdown all connections on its shutdown list.
- when a connection cache is destroyed, discard all connections on the shutdown list. In debug builds, with the envionment variable CURL_GRACEFUL_SHUTDOWN, run a blocking shutdown (using poll) and the default 2 second timeout. This is for testing (see below) and might be exposed as an option to new libcurl API function.

**TODO**: multi_socket() does not know how to handle sockets from connection caches.

Details
- similar to `connected` add a `shutdown` bit at the connection filter instance to remember it has already shutdown and need not be invoked again when we do repeated, non-blocking attempts to shut down the complete filter chain.
- conncache structs have a `tmp_list` member that allows collecting caches onto a temporary list inside multi_wait and multi_perform handling. The list is always cleared before calls return.
- replace `Curl_discard(data, conn)` with `Curl_conncache_shutdown_conn(data, conn,...) to handle keeping connections for non-blocking shutdown handling.
- When CURL_DEBUG is in the environment and  curl is a debug build, make the conncache closing handle verbose.
- fix openssl shutdown to be way more careful about still receiving data after its side has shut down. HTTP/2 connections may have a server-side GOAWAY in transit. We need to cope with that without error.

Testing
- add test cases using `tcpdump` to check if a curl run caused TCP RST packets. These checks are used in test_02_30, test_30_06 and test_31_06.

### DIFF
--- a/lib/cf-h2-proxy.c
+++ b/lib/cf-h2-proxy.c
@@ -183,7 +183,6 @@ struct cf_h2_proxy_ctx {
   BIT(conn_closed);
   BIT(rcvd_goaway);
   BIT(sent_goaway);
-  BIT(shutdown);
   BIT(nw_out_blocked);
 };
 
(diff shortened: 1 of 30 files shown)


icing:
Closing to make an updated, squashed one.

PR #13942, opened 2024-06-13
cf-socket: optimize curlx_nonblock() and check its return error

panjf2000:
Updates:
Due to the particularity of `cf_socket_shutdown()`, we've decided not to use `sock_nonblocking` for `cf_socket_shutdown()`. Meanwhile, we currently don't check if the socket created by `cf_socket_open` is non-blocking when connecting, which can lead to unexpected behaviors (blocking connect), therefore, we can leverage `sock_nonblocking` for these cases. Check out https://github.com/curl/curl/pull/13942#issuecomment-2166047002

---

We sometimes need to check if a socket is non-blocking to determine the subsequent actions, such as the `cf_socket_shutdown()`, which sets the non-blocking flag on the socket and then drains its buffer during a shutdown. It has to use system calls on the socket anyway because there is no way to tell if the current socket is non-blocking, therefore, adding a new flag for struct `cf_socket_ctx` to indicate if the socket is non-blocking will help us eliminate some redundant system calls.

This PR mainly does two things:
1. Add the new field `sock_nonblocking` for `cf_socket_ctx` and assign proper values to it (also reference it) accordingly
2. Improve function `curlx_nonblock()`, make it handle errors properly, and avoid needless system calls

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1234,12 +1234,25 @@ static CURLcode cf_socket_open(struct Curl_cfilter *cf,
 #endif
 
 #ifndef SOCK_NONBLOCK
-  /* set socket non-blocking */
-  (void)curlx_nonblock(ctx->sock, TRUE);
+  /* Set socket non-blocking, must be a non-blocking socket for
+   * a non-blocking connect. */
+  error = curlx_nonblock(ctx->sock, TRUE);
+  if(error < 0) {
+    result = CURLE_UNSUPPORTED_PROTOCOL;
+    ctx->error = SOCKERRNO;
+    goto out;
+  }
 #else
-  if(data->set.fopensocket)
-    /* set socket non-blocking */
-    (void)curlx_nonblock(ctx->sock, TRUE);
+  if(data->set.fopensocket) {
+    /* Set socket non-blocking, must be a non-blocking socket for
+     * a non-blocking connect. */
+    error = curlx_nonblock(ctx->sock, TRUE);
+    if(error < 0) {
+      result = CURLE_UNSUPPORTED_PROTOCOL;
+      ctx->error = SOCKERRNO;
+      goto out;
+    }
+  }
 #endif
   ctx->sock_connected = (ctx->addr.socktype != SOCK_DGRAM);
 out:
--- a/lib/nonblock.c
+++ b/lib/nonblock.c
@@ -50,9 +50,18 @@ int curlx_nonblock(curl_socket_t sockfd,    /* operate on this */
   /* most recent unix versions */
   int flags;
   flags = sfcntl(sockfd, F_GETFL, 0);
+  if(flags < 0)
+    return -1;
+  /* Check if the current file status flags have already satisfied
+   * the request, if so, it's no need to call fcntl() to replicate it.
+   */
+  if(!!(flags & O_NONBLOCK) == !!nonblock)
+    return 0;
   if(nonblock)
-    return sfcntl(sockfd, F_SETFL, flags | O_NONBLOCK);
-  return sfcntl(sockfd, F_SETFL, flags & (~O_NONBLOCK));
+    flags |= O_NONBLOCK;
+  else
+    flags &= ~O_NONBLOCK;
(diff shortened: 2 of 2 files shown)


icing (commented):
I agree with you that saving syscalls in a filter that is involved heavily is a good idea. But for this change to be reliable, it would really be necessary that the socket is only manipulated by the filter itself.

Unfortunately, this is not the case in libcurl for historical reasons. If you scan the code for calls of `curlx_nonblock()`, you'll find several locations outside cf-socket.c.

The other observation I had when making the shutdown change in cf-socket is that Windows sometimes seems to block on receives *unless* the setting to nonblocking just before did indeed succeed. That is the reason for the final receive only being done then.

Due to this, I think using a flag for skipping the `curx_nonblock()` during shutdown will introduce strange edge behaviour, making it issues hard to debug.

panjf2000:
> I agree with you that saving syscalls in a filter that is involved heavily is a good idea. But for this change to be reliable, it would really be necessary that the socket is only manipulated by the filter itself.
> 
> Unfortunately, this is not the case in libcurl for historical reasons. If you scan the code for calls of `curlx_nonblock()`, you'll find several locations outside cf-socket.c.
> 
> The other observation I had when making the shutdown change in cf-socket is that Windows sometimes seems to block on receives _unless_ the setting to nonblocking just before did indeed succeed. That is the reason for the final receive only being done then.
> 
> Due to this, I think using a flag for skipping the `curx_nonblock()` during shutdown will introduce strange edge behaviour, making it issues hard to debug.

Thanks for sharing the particularity of `cf_socket_shutdown()` with non-blocking sockets. So we may not use this new flag in that function, how about only using it on those sockets managed inside `cf-socket.c`?

icing:
> Thanks for sharing the particularity of `cf_socket_shutdown()` with non-blocking sockets. So we may not use this new flag in that function, how about only using it on those sockets managed inside `cf-socket.c`?

I might be wrong, but in cf-socket.c I see the flag only set, but never used outside of the shutdown function?

panjf2000:
It's used in `cf_udp_setup_quic()` if the `curlx_nonblock()` call in `cf_socket_open()` failed for some reason. Also, one unusual thing I've noticed is that currently we never check if the socket is non-blocking after calling `cf_socket_open()`, if the `curlx_nonblock()` call in `cf_socket_open()` failed, as I mentioned before, we would get a blocking socket and start a blocking `connect()` later in `cf_tcp_connect()` and `cf_udp_connect()`, which seems like an unexpected behavior. Maybe we can start using this new flag for these cases? Check out the latest code changes.

panjf2000:
Hi @icing, would you take a glance at the latest code changes?

icing:
> Hi @icing, would you take a glance at the latest code changes?

Thanks for the ping. Looking at the diff, I like the optimization in `nonblock.c` which then also does work for the `ftp.c` use cases.

The `ctx->sock_nonblocking` I still find unnecessary and, over the socket lifetime, might be wong, as it never gets updated. It is read in a single DEBUGASSERT for QUIC. And since never updated from the start, should always be true.

I think you should remove `ctx->sock_nonblocking` and the PR is a good optimization.

panjf2000:
> > Hi @icing, would you take a glance at the latest code changes?
> 
> Thanks for the ping. Looking at the diff, I like the optimization in `nonblock.c` which then also does work for the `ftp.c` use cases.
> 
> The `ctx->sock_nonblocking` I still find unnecessary and, over the socket lifetime, might be wong, as it never gets updated. It is read in a single DEBUGASSERT for QUIC. And since never updated from the start, should always be true.
> 
> I think you should remove `ctx->sock_nonblocking` and the PR is a good optimization.

Done.

icing (approved):

panjf2000:
Hi @icing @bagder, this PR should be good to go, shall we get it merged?

icing:
I am sure @bagder will get right to it after Sweden celebrates midsummer.

panjf2000:
> I am sure @bagder will get right to it after Sweden celebrates midsummer.

OK, thanks for the notice! Have a good one!

bagder:
Thanks!

Issue #1863, opened 2017-09-05
Set up curl lib for Visual Studio 2008


vudoancs:
Hello,

I set up curl lib for VS Studio 2008 using static lib, but I got the below error and have been tried to fix it for several hours, but I cant find the solution.

`1>libcurl_a_debug.lib(idn_win32.obj) : error LNK2019: unresolved external symbol __imp__IdnToAscii@20 referenced in function _curl_win32_idn_to_ascii
1>libcurl_a_debug.lib(idn_win32.obj) : error LNK2019: unresolved external symbol __imp__IdnToUnicode@20 referenced in function _curl_win32_ascii_to_idn`

I appreciate if anyone helps me.
Thank in advance!

jzakrzewski:
You have to add ```Normaliz.lib``` to your linker (see: https://msdn.microsoft.com/en-us/library/windows/desktop/dd318149%28v=vs.85%29.aspx ).

bagder:
Three years ago @jay [wrote](https://curl.haxx.se/mail/lib-2014-12/0084.html):

> VS2008 default SDK doesn't have IdnToAscii or IdnToUnicode in normaliz.lib.

and encouraged the user to build with:

`nmake /f Makefile.vc mode=dll VC=9 ENABLE_IDN=no`

vudoancs:
@jzakrzewski 
Thank you very much!
I already added Normaliz.lib and also got the same problem.

@bagder 
Thank you for your help.

I did like this and then run. I got the message about libcurl.dll.

Can you tell me how to put libcurl.dll in the project?

bagder:
Hi!

Thanks for your interest in curl and related manners, but you seem to have asked a question or help, and this is an issue tracker that we reserve for dealing with issues/bugs really.

Please ask questions and ask for support etc on the [dedicated mailing lists](https://curl.haxx.se/mail/). You will also reach a lot more people that way!

PR #13947, opened 2024-06-14
build: enable AppleIDN / WinIDN by default

vszakats:
Enable and use WinIDN and AppleIDN by default on these platforms,
if they can be used.

Compatibility warning: This change can make default builds incompatible
with Windows XP by default. WinIDN requires `Normaliz.dll`. Windows XP
isn't shipping with it and the necessary extra package is no longer
offered by Microsoft. Thus, for Windows XP compatibility, we recommend
building with `-DUSE_WIN32_IDN=OFF`.

Discussion: #13565
Closes #13947

---

Cleaner diff without whitespace changes: https://github.com/curl/curl/pull/13947/files?w=1

There are a few ways to resolve compatibility with existing builds
and also to handle interactions with the `USE_LIBIDN2` option.
This is an initial attempt.

TODO:
- [ ] rethink defaulting logic to honor an explicitly selected libidn2?
- [x] is WinIDN useful without Unicode? [YES] Enable it only in Unicode builds by default? (Also needs reshuffling options in Azure and GHA/Windows)
Downside: it makes the defaulting logic difficult to predict.
- [x] rebase on #14680. (delete explicit WinIDN options added to GHA/Windows/cross.)
- [x] rebase on #14693.
- [x] add jobs that disable AppleIDN/WinIDN for both autotools and cmake.
- [x] configure: enable WinIDN by default for `HAVE_WIN32_WINNT >= 0x0600`, as cmake.
- [x] review and fixup AppleIDN/WinIDN and libidn2 interactions in autotools and cmake.
- [x] tackle AppleIDN.
- [x] cmake: detect WinXP target using `HAVE_WIN32_WINNT` and adjust default `USE_WIN32_IDN` setting accordingly?
- [x] review and update CI jobs where/if necessary.
- [x] fix autotools logic to only make WinIDN the default when targeting native Windows.
- [x] figure out how to make IDN tests work on Windows. [SKIP]
- [x] split off VS2008 log dump on failure in AppVeyor into separate PR. → #13957

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2844,8 +2844,11 @@ AS_HELP_STRING([--with-apple-idn],[Enable AppleIDN])
 AS_HELP_STRING([--without-apple-idn],[Disable AppleIDN]),
       [OPT_IDN=$withval])
     case "$OPT_IDN" in
-      yes)
-        dnl --with-apple-idn option used
+      no)
+        dnl --without-apple-idn option used
+        AC_MSG_RESULT([no])
+        ;;
+      *)
         AC_MSG_RESULT([yes, check])
         AC_CHECK_LIB(icucore, uidna_openUTS46,
         [
@@ -2858,9 +2861,6 @@ AS_HELP_STRING([--without-apple-idn],[Disable AppleIDN]),
             tst_links_appleidn='yes'
           )
         ])
-        ;;
-      *)
-        AC_MSG_RESULT([no])
         ;;
     esac
     ;;
--- a/configure.ac
+++ b/configure.ac
@@ -610,6 +610,23 @@ case $host_os in
   cygwin*|msys*) curl_cv_cygwin='yes';;
 esac
 
+dnl Check if native Windows targeting Vista or newer
+curl_cv_windows_vista='no'
+if test "$curl_cv_native_windows" = 'yes'; then
+  AC_COMPILE_IFELSE([
+    AC_LANG_PROGRAM([[
+      #include <windows.h>
+    ]],[[
+      #if _WIN32_WINNT < 0x0600
+      #error
+      #endif
+    ]])
+  ],[
+    curl_cv_windows_vista='yes'
+  ],[
+  ])
+fi
+
 AM_CONDITIONAL([HAVE_WINDRES],
   [test "$curl_cv_native_windows" = "yes" && test -n "${RC}"])
 
@@ -5296,19 +5313,8 @@ if test "$tst_atomic" = "yes"; then
 elif test "x$USE_THREADS_POSIX" = "x1" -a \
      "x$ac_cv_header_pthread_h" = "xyes"; then
   SUPPORT_FEATURES="$SUPPORT_FEATURES threadsafe"
-else
-  AC_COMPILE_IFELSE([
-    AC_LANG_PROGRAM([[
-      #include <windows.h>
-    ]],[[
-      #if (WINVER < 0x600) && (_WIN32_WINNT < 0x600)
-      #error
-      #endif
-    ]])
-  ],[
-    SUPPORT_FEATURES="$SUPPORT_FEATURES threadsafe"
-  ],[
-  ])
+elif test "$curl_cv_windows_vista" = 'yes'; then
+  SUPPORT_FEATURES="$SUPPORT_FEATURES threadsafe"
 fi
 
 if test "x$want_winuni" = "xyes"; then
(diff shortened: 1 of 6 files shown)


vszakats:
Tests are unprepared to handle IDN with Windows, both UNICODE and non-UNICODE modes (but differently).

vszakats:
Unexpected fallout with VS2008. Vista target, normaliz.lib specified, yet the IDN functions are not found:
```
6>unity_0.obj : error LNK2019: unresolved external symbol __imp__IdnToAscii@20 referenced in function _win32_idn_to_ascii
[116](https://ci.appveyor.com/project/curlorg/curl/builds/50021801#L116)6>unity_0.obj : error LNK2019: unresolved external symbol __imp__IdnToUnicode@20 referenced in function _win32_ascii_to_idn
[117](https://ci.appveyor.com/project/curlorg/curl/builds/50021801#L117)6>C:\projects\curl\_bld\lib\libcurl.dll : fatal error LNK1120: 2 unresolved externals
```

I still couldn't figure out how to enable verbose logging for the VS2008 jobs. Anybody with a solution for that? Or for the above issue?

vszakats:
The VS2008 issue was reported here earlier: #1863

This trace message says that `normaliz` is added to the list of libs:
```
-- ws2_32;bcrypt;wldap32;normaliz
```
Ref: https://ci.appveyor.com/project/curlorg/curl/builds/50021801/job/xpd1d5cor294gp3m#L42

It feels like something very obvious (or something very cryptic). Seeing the actual linker command would tell more of what's happening.

vszakats:
Backing off of this. It looks nearly hopeless to make IDN tests pass on Windows (without going crazies that is.) It need passing Unicode via the command-line for starters.

I might cherry pick the part moving WINVER detection up to the top.

Feel free to pick up and continue.

vszakats:
Is there any insterest in this?

This PR is now fairly trivial, with roadblocks solved (most via separate commits),
except the Windows Unicode issue/question.

It'd be a logical next step to enable it, but I wonder if default IDN is something 
anybody actually wants on Apple and Windows.

This PR required a fair bit of maintenance to keep around, with 25+ conflicting
rebases. It still falls over whenever touching GHA/{macos,windows}. If there is
no interest or seems problematic due Windows Unicode (or other reasons), I'm
happy to abandon this effort. Otherwise it's ready to merge.

configure.ac
```diff
@@ -2654,7 +2671,17 @@ AS_HELP_STRING([--with-winidn=PATH],[enable Windows native IDN])
...
-    no|default)
+    default)
+      dnl enable by default on native Windows with Vista and newer
+      if test "$curl_cv_windows_vista" = 'yes'; then
+        want_winidn="yes"
+        want_winidn_path="default"
+      else
+        want_winidn="no"
+      fi
+      AC_MSG_RESULT([no])
```
jay, line 2682:
```suggestion
      if test "$curl_cv_windows_vista" = 'yes'; then
        want_winidn="yes"
        want_winidn_path="default"
        AC_MSG_RESULT([yes])
      else
        want_winidn="no"
        AC_MSG_RESULT([no])
      fi
```

jay:
> This PR is now fairly trivial, with roadblocks solved (most via separate commits),
> except the Windows Unicode issue/question.

In Windows if WinIDN is enabled then libcurl treats a non ascii hostname as potentially a UTF-8 encoded hostname and passes it to IdnToAscii. This happens regardless of whether Unicode support was enabled. I think it predates Windows Unicode support in libcurl. libidn I think has different logic where on Windows we treat the non-ascii hostname as UTF-8 encoded only if libcurl was built with Unicode support.

> This PR required a fair bit of maintenance to keep around, with 25+ conflicting
> rebases. It still falls over whenever touching GHA/{macos,windows}.

Users may not expect it if they are enabling libidn, which may be more common for users that build with autotools. Like if libidn is enabled it should take precedence.

> winbuild has it enabled by default: [#13565 (comment)](https://github.com/curl/curl/discussions/13565#discussioncomment-9363774)

winbuild users are different, they typically build curl in a very straightforward way without tests. Though I stand by what I said then. It's totally reasonable to do it for other build systems (as long as it doesn't add a burden). If it is still problematic like tests fail or there is some maintenance requirement then skip it.

vszakats:
Thanks @jay, good points.

I'll (probably) need to rework it heavily to prioritize libidn2.

It doesn't come naturally in cmake at least, because libidn2 was ON by default,
so figuring out an explicit user intent for libidn2 needs some tricks.

jay:
I see. I was thinking of a scenario where a user specified libidn, then I would assume they want that over WinIDN. However if libidn and WinIDN are both on by default, and the user specified neither, then I guess it is a good question as to which should be preferred in Windows.

PR #13951, opened 2024-06-14
docs: Reference non deprecated curl options

yedayak:
There are a places where man pages reference deprecated CURLOPT options, where it doesn't make sense, replace them with the reccomended replacement option.


bagder (approved):

bagder:
Thanks!

PR #13952, opened 2024-06-14
examples: Add missing binaries to .gitignore

yedayak:
They were showing as changed when built. Add them sorted alphabetically, while also moving a few more entries to sorted order.


bagder (approved):

bagder:
Thanks!

PR #13963, opened 2024-06-17
Add vcpkg ci

talregev:
Add vcpkg ci. I bring you the power of vcpkg. It can build 3rd parties on many os, and can help build curl features and configuration. It can integrate with cmake. 
I also add a cache that when it will be other PR and hit the same cache, it will automatic draw the last time compilation. leaving most of the ci time for compile curl and testing.
I also add an upload logs as artifact in case of failing, to see what cause the vpckg to fail. 
vcpkg set to the latest, but it can changed. Also most of the time vpckg is stable, but it can have port fails. Usually vcpkg very responsive, especially in the very popular ports (packages) that needed for curl. 

I added a tests. tests in osx and windows are  fails, not sure why. Maybe I did something in my side. Maybe it can improve curl that way.
Note, I didn't add all the other features for curl because:
1. curl cmake needed to adapt to be more general (and vcpkg friendly).
2. libressl is conflict with openssl. by vcpkg rules it cannot compile togther because they conflict in the functions names and more. many ports (packages) that curl needed also compile with them openssl, that mean I cannot add them because of that conflicts.  

I fix one of the test. I hope I did it correctly, because on windows it complain that the last line is a dead code that cannot reach. 

I will happy to read your feedback and your thoughts.

### DIFF
--- a/.github/workflows/vcpkg-linux.yml
+++ b/.github/workflows/vcpkg-linux.yml
@@ -1,0 +1,165 @@
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+
+name: vcpkg Linux
+
+on:
+  push:
+    branches:
+      - master
+      - '*/ci'
+    paths-ignore:
+      - '**/*.md'
+      - '.azure-pipelines.yml'
+      - '.circleci/**'
+      - '.cirrus.yml'
+      - 'appveyor.*'
+      - 'packages/**'
+      - 'plan9/**'
+      - 'projects/**'
+      - 'winbuild/**'
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+      - '.azure-pipelines.yml'
+      - '.circleci/**'
+      - '.cirrus.yml'
+      - 'appveyor.*'
+      - 'packages/**'
+      - 'plan9/**'
+      - 'projects/**'
+      - 'winbuild/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+permissions: {}
+
+env:
+  OS: linux
+  BUILD_DIR: vcpkg/cache
+  VCPKG_ROOT: vcpkg/cache/vcpkg
+  VCPKG_LINK: https://github.com/microsoft/vcpkg/
+  BINARY_CACHE: vcpkg/cache/linux
+
+jobs:
+  build:
+    name: vcpkg-ubuntu-${{ matrix.PROCESSOR }}-${{ matrix.BUILD_OPTION }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - BUILD_OPTION: openssl
+          PROCESSOR: x64
+          CMAKE_OPTIONS: -DCURL_USE_LIBSSH2=ON
+        - BUILD_OPTION: libressl
+          PROCESSOR: x64
+          CMAKE_OPTIONS: -DCURL_USE_LIBSSH2=OFF
+    env:
+      PROCESSOR: ${{ matrix.PROCESSOR }}
+      BUILD_OPTION: ${{ matrix.BUILD_OPTION }}
+      CMAKE_OPTIONS: ${{ matrix.CMAKE_OPTIONS }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Cache dependencies
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{env.BINARY_CACHE }}/${{ matrix.BUILD_OPTION }}
+          key: ${{ env.OS }}-${{ matrix.BUILD_OPTION }}-${{ hashFiles('.github/workflows/vcpkg-linux.yml') }}
+          restore-keys: ${{ env.OS }}-${{ matrix.BUILD_OPTION }}-${{ hashFiles('.github/workflows/vcpkg-linux.yml') }}
+
+      - name: Install ninja
+        if: success()
+        run: |   
+          sudo apt install -y ninja-build
+          ninja --version
+          whereis ninja
+
+      - name: Init vcpkg
+        if: success()
+        run: |        
+          mkdir -p $BUILD_DIR
+          git -C $BUILD_DIR clone $VCPKG_LINK
+          $VCPKG_ROOT/bootstrap-vcpkg.sh
+  
+      - name: vcpkg build openssl
+        if: ${{ success() && matrix.BUILD_OPTION == 'openssl' }}
+        run: |
+          export CURRENT_BINARY_CACHE="$BINARY_CACHE/$BUILD_OPTION"
+          mkdir -p $CURRENT_BINARY_CACHE
+          export VCPKG_BINARY_SOURCES="clear;files,$PWD/$CURRENT_BINARY_CACHE,readwrite;"
+          $VCPKG_ROOT/vcpkg x-set-installed libssh2[core,openssl] nghttp2 openldap openssl zstd --triplet=$PROCESSOR-$OS-release
+
+      - name: vcpkg build libressl
+        if: ${{ success() && matrix.BUILD_OPTION == 'libressl' }}
+        run: |
+          export CURRENT_BINARY_CACHE="$BINARY_CACHE/$BUILD_OPTION"
+          mkdir -p $CURRENT_BINARY_CACHE
+          export VCPKG_BINARY_SOURCES="clear;files,$PWD/$CURRENT_BINARY_CACHE,readwrite;"
+          $VCPKG_ROOT/vcpkg x-set-installed libressl nghttp2 zstd --triplet=$PROCESSOR-$OS-release
+
+      - name: cmake config
+        if: success() 
+        run: |
+          export CMAKE_GENERATOR=Ninja
+          cmake . -B build \
+              -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
+              -DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed \
+              -DVCPKG_TARGET_TRIPLET=$PROCESSOR-$OS-release \
+              -DCMAKE_UNITY_BUILD=ON \
+              -DCURL_WERROR=ON \
+              -DBUILD_SHARED_LIBS=OFF \
+              -DCURL_ZSTD=ON \
+              -DUSE_NGHTTP2=ON \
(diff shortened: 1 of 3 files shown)


tests/unit/unit2604.c
```diff
@@ -104,7 +104,9 @@ UNITTEST_START
   }
 
   free((void *)list[0].cp);
-  return error == 0 ? CURLE_OK : TEST_ERR_FAILURE;
+  if(error == 0) {
+    return CURLE_OK;
+  }
```
bagder, line 109: This looks wrong. What if error is not zero, what does it return then?

bagder:
If you want to add directories and files for testing purposes, I think they should be put in `tests/` and not in the root dir.

tests/unit/unit2604.c
talregev: It going to this line and return:
return (CURLcode)unitfail; 
https://github.com/curl/curl/blob/master/tests/unit/curlcheck.h#L110

talregev: What do you think should be the value? 
Maybe this line:
https://github.com/curl/curl/blob/master/tests/unit/curlcheck.h#L110
should return TEST_ERR_FAILURE?

talregev:
> If you want to add directories and files for testing purposes, I think they should be put in `tests/` and not in the root dir.

I don't mind to add them to tests, but this is configs files that help for building the vcpkg configuration, and not tests files. 
Do you have a 3rd-parties library?

tests/unit/unit2604.c
talregev: I think I know what needed to be done. I understand my error.
I think I should call abort_test.

bagder:
> these are configs files that help building vcpkg configurations

I don't understand. Why do we want files added for building vcpkg configurations? We typically don't host files for building packages because they are usually best authored and maintained by the peeps who build and host those packages.

talregev:
> > these are configs files that help building vcpkg configurations
> 
> I don't understand. Why do we want files added for building vcpkg configurations? We typically don't host files for building packages because they are usually best authored and maintained by the peeps who build and host those packages.

It json files that tell which ports (libs) to build. It a configurations, and I use it for the ci building. So I move them to tests as you suggested.

dfandrich:
Files needed for packaging curl aren't kept in the curl source repo; the files in packages/ are configs needed for even basic builds on a few platforms. I don't think vcpkg files belong here any more than Debian control files or RPM spec files, or a hundred similar config files. Making sure curl remains able to be packaged by vcpkg using CI jobs is something vcpkg people should be doing, but IMHO it's out of scope of the curl project.

talregev:
I am not package curl via vcpkg.
I am building the libraries that needed for curl compilation via vcpkg in the ci. similar like curl do in the ci `sudo apt install libssl-dev`
Then the cmake is continue to build regular like you do cmake building in the ci. 
I am also using the cmake install command that on linux do `make install` like curl do in the ci.

This purpose of the ci is to build the latest libraries needed curl to test curl building with them, then test curl with the libraries with curl ci test.
This is also testing different cmake build options, and test curl cmake itself.

dfandrich:
Ok, that makes a bit more sense. But what do these builds add that the existing ones don't already do, that get their dependencies with means other than vcpkg? Just using vcpkg doesn't add any value that I can see.

vszakats:
I see the value of vcpkg for MSVC jobs to test with freely chosen dependencies. Serving as the MSVC equivalent of `brew` and `apt`. However for macOS and Linux, existing tools work with no issues, what does vcpkg bring there?

As for MSVC, what do you think of integrating vcpkg into the existing MSVC workflow in `windows.yml`?

Can we use `vcpkg` _without_ a whole tree of config files and empty directory committed to the repo? Does vcpkg support requesting dependencies without those files, from the command-line? It'd be a big plus if the CI logic would remain self-contained within their CI .yaml files.

Why is it necessary to modify a unit test result?

bagder:
> Serving as the MSVC equivalent of brew and apt

We use brew and apt to install a lot of dependencies for CI jobs. That's their purpose. I don't see that this newly built vcpkg in these CI jobs is actually used to install dependencies for builds. So why do we build it?

vszakats:
What I meant is that for _MSVC jobs_, there is no mechanism to install dependencies, like we used to with `brew` and `apt` (etc). For _these jobs_ (MSVC), `vcpkg` is the package manager. Without it, the only way to use dependencies is to either have them installed by default, or install them manually, which is very painful, with different methods for each dependency.

MSYS2 solves this for mingw-w64. Cygwin also has this problem solved.

So basically all platforms have a package manager of some kind, _except MSVC_, and vcpkg seems to plug that hole.

For other targets, I don't see the benefit of vcpkg, in curl's context at least.

talregev:
> I see the value of vcpkg for MSVC jobs to test with freely chosen dependencies. Serving as the MSVC equivalent of `brew` and `apt`. However for macOS and Linux, existing tools work with no issues, what does vcpkg bring there?

The main different from apt (I am not sure about brew) that vcpkg bring you the latest library version that exist in vcpkg that usually the latest version library that exist. especially on the libraries that are very popular that curl are using. 
For example in openssl. openssl on ubuntu is 3.0.x, and on vcpkg is bring you 3.3.1 (couple days after they release it).
For osx there some tests that was failed, I think it due the cases is not check in the osx ci, and now you can add easily cases and improve osx.

> As for MSVC, what do you think of integrating vcpkg into the existing MSVC workflow in windows.yml? (rather than rolling their separate .yml, with potentially redundant CI jobs)

It can be done to integrate vcpkg to windows.yml. but You have working ci, and before you commit to vcpkg, you should check it on sevral PRs and then can integrate it to one file after you see it working as you expected. Also this is work in progress, because you should also adapt cmake that will find all other libraries from vcpkg that curl needed that currently is not the case. for example zlib. So it better to have different ci that easily change / remove as you want. 

Can we use vcpkg without a tree of config files and empty directory committed to the repo? Does vcpkg support requesting dependencies without those files, from the command-line? It'd be a big plus if CI logic would remain self-contained within their CI .yaml files.

Yes I did. I refactor the ci files and remove the vcpkg folder entirely.

vszakats:
> > I see the value of vcpkg for MSVC jobs to test with freely chosen dependencies. Serving as the MSVC equivalent of `brew` and `apt`. However for macOS and Linux, existing tools work with no issues, what does vcpkg bring there?
> 
> The main different from apt (I am not sure about brew) that vcpkg bring you the latest library version that exist in vcpkg that usually the latest version library that exist. especially on the libraries that are very popular that curl are using. For example in openssl. openssl on ubuntu is 3.0.x, and on vcpkg is bring you 3.3.1 (couple days after they release it). For osx there some tests that was failed, I think it due the cases is not check in the osx ci, and now you can add easily cases and improve osx.

For `apt` we can have this effect by using debian:unstable or a rolling distro of choice. For specific deps we solve this by building the them from source. This is mostly painless on Linux; on the plus side it's standard, everyone understands it. I think it's also a plus to avoid large dependencies like vcpkg unless unavoidable. `brew` offers rolling releases as standard.

> > As for MSVC, what do you think of integrating vcpkg into the existing MSVC workflow in windows.yml? (rather than rolling their separate .yml, with potentially redundant CI jobs)
> 
> It can be done to integrate vcpkg to windows.yml. but You have working ci, and before you commit to vcpkg, you should check it on sevral PRs and then can integrate it to one file after you see it working as you expected. Also this is work in progress, because you should also adapt cmake that will find all other libraries from vcpkg that curl needed that currently is not the case. for example zlib. So it better to have different ci that easily change / remove as you want. and over time to integrate one to another.

I'd be happy with integrating as part of this PR. Not least, CI tests are CPU/time expensive and add flakyness.

What do you think of making the two no-SSL (= not so useful) MSVC jobs in `windows.yml` more useful by enabling OpenSSL / LibreSSL respectively? with your libssh2 settings. Seems like a perfect fit.

> > Can we use vcpkg without a tree of config files and empty directory committed to the repo? Does vcpkg support requesting dependencies without those files, from the command-line? It'd be a big plus if CI logic would remain self-contained within their CI .yaml files.
> 
> Yes I did. I refactor the ci files and remove the vcpkg folder entirely.

Looking good, thanks!

> On windows on openssl build option, it create this compilation warning / error:
> 
> ```
> D:\a\curl\curl\tests\unit\unit2604.c(111) : error C2220: the following warning is treated as an error
> D:\a\curl\curl\tests\unit\unit2604.c(111) : warning C4702: unreachable code
> ```
> 
> So I change the test file that it will be without unreachable code. I would like to have a feedback if I did the fix correctly, or how I can fix correctly that solve the unreachable code problem.

Fair enough. Would you mind adding this to the commit/PR message?

talregev:
> Fair enough. Would you mind adding this to the commit/PR message?
I think I will make a different PR for test fix.

Do you want me to delete the ci for linux and osx?
Will you attempted to fix the tests failing on osx vcpkg ci?  (Even if you don't want the ci itself).

I am not sure I want to refactor windows.yml. It more time for me. I can make the usecase more usefull on vcpkg-windows.yml ci.

bagder:
The Windows fails test 1516 because the shell used there converts the `/path/1516` string given as an argument to the test code into `c:/../path/1516` which the test program does not deal with.

talregev:
@bagder Also interesting test that can fix easily:

test 1013...[Compare curl --version with curl-config --protocols]
Mismatch in protocols lists:
curl:        dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
curl-config: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smtp smtps telnet tftp
 1013: postcheck FAILED

On vcpkg-osx.yml secure transport option

What is the missing smb smbs protocols?
How I can add them in cmake build for curl-config?

vszakats:
> > Fair enough. Would you mind adding this to the commit/PR message?
> 
> I think I will make a different PR for test fix.
> 
> Do you want me to delete the ci for linux and osx? Will you attempted to fix the tests failing on osx vcpkg ci? (Even if you don't want the ci itself).
>
> I am not sure I want to refactor windows.yml. It more time for me. I can make vcpkg-windows.yml more usefull ci.

I prefer your #13979 over this PR. Left a bunch of reviews with mostly small things.

bagder:
Does this PR still have merit after #13979 was merged?

talregev:
> Does this PR still have merit after #13979 was merged?

I left this one open because there is a bug in osx that the tests 1013 and 1014 failed.

vszakats:
@talregev Thanks for the heads up, I've opened #14065 for the smb/smbs protocol-, and NTLM in the feature-list fallouts.

vszakats:
(I think it's safe to close this now.)

talregev:
Thank you for fixing the tests!
you should also add a tests case that check these bugs.

PR #13965, opened 2024-06-17
hostip: refuse to resolve the .i2p TLD

marek22k:
Following the issue https://github.com/curl/curl/issues/543 and the discussion in https://github.com/curl/curl/discussions/13964 as well as the comments in https://daniel.haxx.se/blog/2024/05/17/curl-tor-dot-onion-and-socks/, this commit implements that curl no longer resolves .i2p domains.

Closes https://github.com/curl/curl/discussions/13964

(And yes, I "stole" the content itself from the commit https://github.com/curl/curl/commit/0ae0abbe72514a75c10bfc4108d9f254f594c086.)

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -694,6 +694,12 @@ enum resolve_t Curl_resolv(struct Curl_easy *data,
     failf(data, "Not resolving .onion address (RFC 7686)");
     return CURLRESOLV_ERROR;
   }
+  if(hostname_len >= 5 &&
+     (curl_strequal(&hostname[hostname_len - 4], ".i2p") ||
+      curl_strequal(&hostname[hostname_len - 5], ".i2p."))) {
+    failf(data, "Not resolving .i2p address");
+    return CURLRESOLV_ERROR;
+  }
   *entry = NULL;
 #ifndef CURL_DISABLE_DOH
   conn->bits.doh = FALSE; /* default is not */
--- a/tests/data/Makefile.inc
+++ b/tests/data/Makefile.inc
@@ -189,7 +189,7 @@ test1455 test1456 test1457 test1458 test1459 test1460 test1461 test1462 \
 test1463 test1464 test1465 test1466 test1467 test1468 test1469 test1470 \
 test1471 test1472 test1473 test1474 test1475 test1476 test1477 test1478 \
 test1479 test1480 test1481 test1482 test1483 test1484 test1485 test1486 \
-test1487 \
+test1487 test1488 test1489 \
 \
 test1500 test1501 test1502 test1503 test1504 test1505 test1506 test1507 \
 test1508 test1509 test1510 test1511 test1512 test1513 test1514 test1515 \
--- a/tests/data/test1488
+++ b/tests/data/test1488
@@ -1,0 +1,41 @@
+<testcase>
+<info>
+<keywords>
+I2P
+FAILURE
+</keywords>
+</info>
+#
+# Server-side
+<reply>
+</reply>
+
+#
+# Client-side
+<client>
+<server>
+none
+</server>
+<features>
+http
+</features>
+<name>
+Fail to resolve .i2p TLD
+</name>
+<command>
+proxy.i2p
+</command>
+</client>
+
+#
+# Verify data after the test has been "shot"
+<verify>
+# Couldn't resolve host name
+<errorcode>
+6
+</errorcode>
+<stderr mode="text">
+curl: (6) Not resolving .i2p address
+</stderr>
+</verify>
+</testcase>
--- a/tests/data/test1489
+++ b/tests/data/test1489
@@ -1,0 +1,41 @@
+<testcase>
+<info>
+<keywords>
+I2P
+FAILURE
+</keywords>
+</info>
+#
+# Server-side
+<reply>
+</reply>
+
+#
+# Client-side
+<client>
+<server>
+none
+</server>
+<features>
+http
+</features>
+<name>
+Fail to resolve .i2p TLD
+</name>
+<command>
+router.i2p
+</command>
+</client>
+
+#
+# Verify data after the test has been "shot"
+<verify>
+# Couldn't resolve host name
+<errorcode>
+6
+</errorcode>
+<stderr mode="text">
+curl: (6) Not resolving .i2p address
+</stderr>
+</verify>
+</testcase>


bagder:
I'm against this purely on the basis that there is no RFC for this, which there is for `.onion` and that was a strong reason why we merged that code.

If we are going to loosen the requirements, where do we draw the line? There are likely several other inventions and projects who want to reserve a TLD for their area and could insist on curl rejecting them. Should we reject everything that some group somewhere think we should?

eyedeekay:
Apologies, I meant to join this discussion yesterday but life had other plans. So I researched this from our(I2P's) perspective and there's some places I believe we could start a discussion.

First, this could be considered a documentation issue for both I2P and for curl. The rationale for considering this a doc issue is that the issue only occurs when using `curl` with a SOCKS proxy. When using `curl` with an HTTP proxy, the behavior is different, and DNS leaks do not occur. So for now there is a solution/workaround which will work for HTTP-over-I2P. So somewhere on I2P's docs I put a warning "Only use `curl` with HTTP proxies, see script for examples. When using `curl` with SOCKS, DNS leaks are likely to occur."

> If we are going to loosen the requirements, where do we draw the line? There are likely several other inventions and projects who want to reserve a TLD for their area and could insist on curl rejecting them. Should we reject everything that some group somewhere think we should?

Interesting questions all. At this time, I know of at least 5 or 6 such inventions where DNS leaks when using SOCKS proxies could occur, because they use something TLD-like, and because they are capable of exposing a socket or something that works like a socket. All of these might have different workarounds involving, potentially, such as running a local DNS resolver, or chaining an HTTP proxy in front of the SOCKS proxy. So, three possible answers:

Answer \#1: Don't draw the line at all. Make it configurable, and make the default `.onion`. Let the users draw the line, by giving them the tools to do so.

Answer \#2: Require that projects file some basic "paperwork" with you, both so that they can explain the need and so that you can include it in your documentation. That "paperwork" would need to justify why the project has needs comparable to those described by the RFC and should specifically reference the relevant RFC sections. I've asked @marek22k to consider a template for this paperwork if this is a thing you choose.

Answer \#3: Do nothing. Declare that all non-`.onion` inventions are unsupported, and that `curl` will probably leak DNS when using them with SOCKS proxies. Recommend that SOCKS proxy users chain an HTTP proxy in advance of the SOCKS proxy to avoid the issue. Protocols which can't be routed through HTTP proxies can't be reliably supported.

> I'm against this purely on the basis that there is no RFC for this, which there is for .onion and that was a strong reason why we merged that code.

This might be a trite argument, but `.i2p` domains may be breaking the current recommendations/rules for new global top-level domains by using an alphanumeric TLD(See page 64: https://newgtlds.icann.org/sites/default/files/guidebook-full-04jun12-en.pdf). It happened before my time, but according to second-hand accounts this actually is part of the reason that `.i2p` is not in that RFC. I don't suppose this is a good enough reason to refuse to resolve it?

dfandrich:
> "Only use curl with HTTP proxies, see script for examples.
> When using curl with SOCKS, DNS leaks are likely to occur."

DNS leaks shouldn't occur in curl when using socks5h:// proxies. Are you saying it still leaks in that case?

eyedeekay:
I think we only tested with socks5, not socks5h. If what you say is the case then this is truly a documentation issue. I will have to check and get back with you. @marek22k can you also confirm for us?

bagder:
If you use SOCKS that resolves hostnames locally, the leak through the name resolver. If you ask the proxy to resolve hostnames, libcurl will **not** leak the name. It sort of goes with how the protocols work.

curl supports proxy-resolving for both SOCKS4 and SOCKS5

bagder:
> Answer 1: Don't draw the line at all. Make it configurable

That's not really changing much. That would then just move the rejection from libcurl to curl and we would then have the same same question  there. Unless you mean that we should provide a command line option for `--ignore-tlds onion,i2p` - which then would not save a lot of people because the people we would want to save with this feature is the ones who aren't aware of the possible leak problem to begin with.

> Answer 2: Require that projects file some basic "paperwork" with you

We added the `.onion` ignore because there is an RFC that says we should. (This did not save us from an avalanche of critique from those who think that we should not follow the RFC.) I think we need to at least have some set of requirements for what TLDs should go in there. In this case of `i2p` it is not even a valid TLD, it is just a three letter combination that a self-selected group on the internet have decided is special and "theirs". Since `i2p` is not even  recognized as a TLD on the Internet, should curl then really take that step?

> Answer 3: Do nothing

We have "done nothing" for some 26 years before this was brought here. And "do nothing" is also a state that can always change as no decisions or moves are forever.

marek22k:
Sorry for the late response. As far as I understand it, it leaks in the same scenario as .onion domains. If it is not explicitly stated that the DNS resolution should also be done via the proxy - so it only leaks for `socks5`. I have not yet tested the whole thing with the HTTP proxy.

eyedeekay:
@marek22k I have tested it with the HTTP proxy as have a few others and for us it seems that curl does not leak hostnames to the DNS when an HTTP proxy is specified with `http_proxy` and/or `https_proxy` as the case may be. It also appears to me that it does not leak if `socks5h` is specified.

At least in the short term, in light of those results^, this is a documentation issue on our side. Simply telling people "You must use `socks5h`" gives us what we want, non-leaking resolution of I2P names, without introducing a new configuration option or requiring curl to implement a process for incorporating new changes that don't have a corresponding RFC. It's just a question of how do we make sure that people know this information before they start trying to use curl in a leaky way.

bagder:
As discussed, we will not pull change (now).

marek22k:
Is there any news yet? Will curl still not prevent the leaking of .i2p domains to DNS servers?

bagder:
There is still no standard/RFC saying transfer tools/libraries should.

marek22k:
I just think it's unfair to favor Tor over I2P.
And just because it's not in a standard, does that mean you can't do it?
The first I2P version was released in 2003 (according to Wikipedia). Isn't 22 years of existence enough to consider it a kind of "standard due to long existence"?

Do you have instructions on how to standardize something like this? As far as I know, an RFC takes several years and proposals are often ignored (at least there are many drafts that make sense but have never been converted into an RFC).

marek22k:
I asked about this, and it appears that the IETF did not want to adopt .i2p: `IETF refused adoption`
Therefore: I2P wanted to be standardized, but the IETF didn't want it. Should I2P users now be disadvantaged because the IETF is not doing its job?

bagder:
> I just think it's unfair to favor Tor over I2P.

Just repeating the same arguments again when nothing has changed might not change the outcome.

It is not a good idea for every new network system to invent their own random TLDs and then go around the world and insist tools need to add special code to deal with them. "But i2p is special". Yes, everyone is special so here we are.

Claiming "unfair" treatment is missing the point. It is about drawing lines and having requirements for what we support. Plus of course, curl is not about being "fair" to random projects out there. curl is about doing rock solid internet transfers for the world.

Finally: the `.onion` filtering is already questioned every once in a while by several Tor users who rather want us to *remove* it again. I don't think adding yet another filter is going to make things better or less complicated for us. The filtering we have is already *complicated*.

PR #13967, opened 2024-06-18
Fix unreachable code warning in test for windows msvc compilation

talregev:
After run vcpkg-windows ci with openssl build option, I found this compilation error:

```
D:\a\curl\curl\tests\unit\unit2604.c(111) : error C2220: the following warning is treated as an error
D:\a\curl\curl\tests\unit\unit2604.c(111) : warning C4702: unreachable code
```

I fix the test.

### DIFF
--- a/tests/unit/unit2604.c
+++ b/tests/unit/unit2604.c
@@ -104,7 +104,10 @@ UNITTEST_START
   }
 
   free((void *)list[0].cp);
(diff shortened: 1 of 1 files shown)


bagder:
I propose #13973 instead.

talregev:
Thank you for your fix. I will close this PR.

PR #13973, opened 2024-06-19
unit2604: use 'unitfail' instead of 'error' variable

bagder:
Since the framework is already returning that variable by default. Avoids a warning for unreachable code.

Fixes #13967


talregev:
Thank you for your fix. I will try to compile it on vcpkg-windows ci to test that it solve the problem.

PR #13976, opened 2024-06-19
graceful connection shutdown

icing:
When libcurl discards a connection there are two phases this may go through: "shutdown" and "closing". If a connection is aborted, the shutdown phase is skipped and it is closed right away.

The connection filters attached to the connection implement the phases in their `do_shutdown()` and `do_close()` callbacks. Filters carry now a `shutdown` flag next to `connected` to keep track of the shutdown operation.

Filters are shut down from top to bottom. If a filter is not connected, its shutdown is skipped. Notable filters that *do* something during shutdown are HTTP/2 and TLS. HTTP/2 sends the GOAWAY frame. TLS sends its close notify and expects to receive a close notify from the server.

As sends and receives may EAGAIN on the network, a shutdown is often not successful right away and needs to poll the connection's socket(s). To facilitate this, such connections are placed on a new shutdown list inside the connection cache.

Since managing this list requires the cooperation of a multi handle, only the connection cache belonging to a multi handle is used. If a connection was in another cache when being discarded, it is removed there and added to the multi's cache shutdown list. If no multi handle is available at that time, the connection is shutdown and closed in a one-time, best-effort attempt.

When a multi handle is destroyed, all connection still on the shutdown list are discarded with a final shutdown attempt and close. In curl debug builds, the environment variable `CURL_GRACEFUL_SHUTDOWN` can be set to make this graceful with a timeout in milliseconds given by the variable.

The shutdown list is limited to the max number of connections configured for a multi cache. Set via CURLMOPT_MAX_TOTAL_CONNECTIONS. When the limit is reached, the oldest connection on the shutdown list is discarded.

### Operation
- In multi_wait() and multi_waitfds(), iterate over all shutdown connections in the multi's cache. Each such connection contributes sockets and POLLIN/OUT events.
- in multi_perform() call perform on the multi's cache to process the shutdown list.
- for event based multis (multi->socket_cb set), add the sockets and their poll events via the callback. When `multi_socket()` is invoked for a socket not known by an active transfer, forward this to the multi's cache for processing. On closing a connection, remove its socket(s) via the callback.

### TLS
TLS connection filters MUST NOT send close nofity messages in their `do_close()` implementation. The reason is that a TLS close notify signals a success. When a connection is aborted and skips its shutdown phase, the server needs to see a missing close notify to detect something has gone wrong.

### FTP
A graceful shutdown of FTP's data connection is performed implicitly before regarding the upload/download as complete and continuing on the control connection. For FTP without TLS, there is just the socket close happening. But with TLS, the sent/received close notify signals that the transfer is complete and healthy. Servers like `vsftpd` verify that and reject uploads without a TLS close notify.

### Tests
- added test_19_* for shutdown related tests
- test_19_01 and test_19_02 test for TCP RST packets which happen without a graceful shutdown and should no longer appear otherwise.
- add test_19_03 for handling shutdowns by the server
- add test_19_04 for handling shutdowns by curl
- add test_19_05 for event based shutdowny by server
- add test_30_06/07 and test_31_06/07 for shutdown checks on FTP up- and downloads.

## TODO

This PR handles graceful shutdown during `multi` operations without changes visible in the API. Enabling a graceful shutdown when a cache is destroyed is available to debug builds via environment variables only.

The maximum number of connections in shutdown is the maximum number configured or a connection cache. This is only available for caches owned by a multi handle, where `CURLMOPT_MAX_TOTAL_CONNECTIONS` can be used. 

There is a `data->set.shutdowntimeout`, but there is no `CURLOPT_SHUTDOWN_TIMEOUT_MS` to set it. Instead, the default, internal `DEFAULT_SHUTDOWN_TIMEOUT_MS` of 2 seconds it used then.

### Non-Features

For share caches, the discarded connections move into the multi's cache (if there is a multi, otherwise they are shut down and closed right away). If the multi is an "easy multi", the easy cleans up the multi when done and that in turns shutdown+closes all pending shutdowns. So with serial easy handles, as used in `curl`, there is no graceful shutdown going on in the background. I presume we can live with that.

### DIFF
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -29,13 +29,17 @@
 
 #include "urldata.h"
 #include "url.h"
+#include "cfilters.h"
 #include "progress.h"
 #include "multiif.h"
 #include "sendf.h"
 #include "conncache.h"
+#include "http_negotiate.h"
+#include "http_ntlm.h"
 #include "share.h"
 #include "sigpipe.h"
 #include "connect.h"
+#include "select.h"
 #include "strcase.h"
 
 /* The last 3 #include files should be in this order */
@@ -44,6 +48,26 @@
 #include "memdebug.h"
 
 #define HASHKEY_SIZE 128
+
+static void conncache_discard_conn(struct conncache *connc,
+                                   struct Curl_easy *last_data,
+                                   struct connectdata *conn,
+                                   bool aborted);
+static void connc_disconnect(struct Curl_easy *data,
+                             struct connectdata *conn,
+                             struct conncache *connc,
+                             bool do_shutdown);
+static void connc_run_conn_shutdown(struct Curl_easy *data,
+                                    struct connectdata *conn,
+                                    bool *done);
+static void connc_run_conn_shutdown_handler(struct Curl_easy *data,
+                                            struct connectdata *conn);
+static CURLcode conncache_update_shutdown_ev(struct Curl_multi *multi,
+                                             struct Curl_easy *data,
+                                             struct connectdata *conn);
+#ifdef DEBUGBUILD
+static void conncache_shutdown_all(struct conncache *connc, int timeout_ms);
+#endif
 
 static CURLcode bundle_create(struct connectbundle **bundlep)
 {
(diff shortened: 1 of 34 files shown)


lib/conncache.c
```diff
@@ -100,25 +124,38 @@ static void free_bundle_hash_entry(void *freethis)
   bundle_destroy(b);
 }
 
-int Curl_conncache_init(struct conncache *connc, size_t size)
+static curl_off_t next_cache_id;
```
bagder, line 127: Without thinking through properly but is there not a risk that this misbehaves in multi-threaded contexts? Should it not be a field in the multi handle instead of being a global static?

lib/conncache.c
```diff
@@ -45,6 +49,26 @@
...
+static void connc_run_conn_shutdown(struct Curl_easy *data,
+                                    struct connectdata *conn,
+                                    bool *done);
+static void connc_run_conn_shutdown_handler(struct Curl_easy *data,
+                                            struct connectdata *conn);
+static CURLcode conncache_update_shutdown_ev(struct Curl_multi *multi,
+                                             struct Curl_easy *data,
+                                             struct connectdata *conn);
+#ifdef DEBUGBUILD
+static void conncache_shutdown_all(struct conncache *connc, int timeout_ms);
```
bagder, line 69: Both `connc_` and `conncache_` are used as prefixes for connection cache functions. Is the former one for the pool of closing connections? It's not obvious to me based on the name.

lib/conncache.c
icing: The problem are the shares with their conncache. Initially, they managed a shutdown list as well, but that got messy in the event based handling. So, we could remove the cache id again, I guess, and avoid this.

lib/conncache.c
icing: Yeah, the static functions should all start with `connc_` for unity. Will change.

lib/conncache.c
icing: cache id removed again.

lib/conncache.c
icing: all static functions now start with `connc_`.

bagder (approved):

bagder:
Thanks!

PR #13977, opened 2024-06-20
curl verbose option, repeated use

icing:
- make mentioning `-vv` on the curl command line increase the verbosity of the trace output
- related discussion https://github.com/curl/curl/discussions/13810
- add trace group names for components: `network`, `protocol`, `proxy`. Usable in `--trace-config`.

Extract from curl's manpage on `--verbose:`

...
Since curl 8.10, mentioning this option several times in the same argument
increases the level of the trace output. However, as before,
a single `-v`, `--verbose` or `--no-verbose` reverts any additions by
previous `-vv` again. This means that `-vv -v` is equivalent to `-v`. This
avoids unwanted verbosity when the option is mentioned in the command line
*and* curl config files.

Using it twice, e.g. `-vv`, will output time (`--trace-time`) and transfer
ids (`--trace-ids`), as well as enable tracing for all protocols 
(`--trace-config protocol`).

Adding a third verbose will output transfer content (`--trace-ascii %`) and
enable tracing of more components (`--trace-config read,write,ssl`).

A forth time will add tracing of all network components 
(`--trace-config network`).
...

### DIFF
--- a/docs/cmdline-opts/verbose.md
+++ b/docs/cmdline-opts/verbose.md
@@ -28,8 +28,28 @@ and a line starting with * means additional info provided by curl.
 If you only want HTTP headers in the output, --include or --dump-header might
 be more suitable options.
 
-If you think this option still does not give you enough details, consider using
---trace or --trace-ascii instead.
+Since curl 8.10, mentioning this option several times in the same argument
+increases the level of the trace output. However, as before,
+a single `-v`, `--verbose` or `--no-verbose` reverts any additions by
+previous `-vv` again. This means that `-vv -v` is equivalent to `-v`. This
+avoids unwanted verbosity when the option is mentioned in the command line
+*and* curl config files.
+
+Using it twice, e.g. `-vv`, outputs time (`--trace-time`) and transfer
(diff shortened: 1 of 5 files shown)


jay:
You said in https://github.com/curl/curl/discussions/13810#discussioncomment-9579638

> I agree that this would change behaviour for people having -v in their .curlrc and adding it on the command line.

IMO we shouldn't do it that way, it's too unexpected and not practical that users will either remember or modify scripts which they may not control to adjust. A verbosity level could be taken from curlrc and then from the command line and go with whichever is greater.

bagder:
I think this is an interesting take.

I'm thinking maybe we could limit the unexpected side-effects if we make sure to only treat consecutive ones that are adjacent to each other for "leveling", and not ones that are separated by anything. So `-vv` and `-vvv` are higher level logging requests, but `-v -v` is not. That way, having multiple config files setting `-v` in each will not be affected.

icing:
Nice idea. I like it.

icing:
Implemented that a single `-v`, `--verbose` or `--no-verbose` undoes all previous effects of a `-vv...`again.

bagder (approved):

PR #13979, opened 2024-06-21
GHA: use vcpkg to install packages for MSVC jobs

talregev:
You already have vcpkg in your msvc jobs in your ci. You didn't aware of it.
I refactor the vcpkg and add more libraries that can compile with curl.
I added cache that can be restore, and not building every library from scratch every PR, it will use the cache. (save ~7min on openssl and schannel build).

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -493,8 +493,8 @@ jobs:
             "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake" \
             "-DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed" \
             '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
-            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
             -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
+            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
             -DCMAKE_UNITY_BUILD=ON \
             -DCURL_WERROR=ON \
             -DBUILD_EXAMPLES=ON \
@@ -511,7 +511,7 @@ jobs:
           export VCPKG_ROOT="$GITHUB_WORKSPACE/vcpkg"
           mkdir -p bld
           7z a -t7z -r -mx=9 logs.7z \
-            "$VCPKG_ROOT/buildtrees/*.log" \
+            "$VCPKG_ROOT"/buildtrees/*.log \
             bld/.ninja_log \
             bld/build.ninja \
             bld/install_manifest.txt \
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -488,8 +488,8 @@ jobs:
             system='Windows'
           fi
           export VCPKG_ROOT="$PWD/vcpkg"
-          cmake . -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=$system" \
+          cmake -B bld ${options} \
+            "-DCMAKE_SYSTEM_NAME=${system}" \
             "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake" \
             "-DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed" \
             '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -413,52 +413,129 @@ jobs:
           cmake --build bld --config '${{ matrix.type }}' --target test-ci
 
   msvc:
-    name: 'msvc (${{ matrix.arch }}, ${{ matrix.plat }}, ${{ matrix.config }})'
+    name: 'msvc (${{ matrix.arch }}, ${{ matrix.plat }}, ${{ matrix.name }})'
     runs-on: windows-latest
     timeout-minutes: 30
     strategy:
       matrix:
         include:
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=ON  -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=ON ', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=OFF', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON          -DENABLE_UNICODE=OFF', type: 'Debug' }
+          # name should be unique and without special chars. It creates a folder with that name for the cache.
+        - name: 'schannel'
+          install: 'zstd zlib nghttp2 libssh2[core,zlib]'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON  -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON'
+        - name: 'openssl'
+          install: 'zstd zlib nghttp2 nghttp3 openssl libssh2'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_OPENSSL_QUIC=ON'
+        - name: 'no ssl'
+          install: 'zstd zlib'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_ZSTD=ON -DHTTP_ONLY=ON'
       fail-fast: false
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: 'restore cache dependencies'
+        uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4
+        with:
+          path: 'cache/${{ matrix.name }}'
+          key: ${{ matrix.plat }}-${{ matrix.name }}
+          restore-keys: ${{ matrix.plat }}-${{ matrix.name }}
+
+      - name: 'vcpkg init'
+        timeout-minutes: 5      
+        if: success()
+        shell: bash
+        run: |
+          git clone --quiet --depth 1 https://github.com/microsoft/vcpkg/
+          vcpkg/bootstrap-vcpkg.sh
+
+      - name: 'vcpkg build'
+        timeout-minutes: 10
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$PWD/vcpkg"
+          export CURRENT_BINARY_CACHE='cache/${{ matrix.name }}'
+          mkdir -p "$CURRENT_BINARY_CACHE"
+          export VCPKG_BINARY_SOURCES="clear;files,$GITHUB_WORKSPACE/$CURRENT_BINARY_CACHE,readwrite;"
+          vcpkg/vcpkg x-set-installed ${{ matrix.install }} --triplet=${{ matrix.arch }}-${{ matrix.plat }}
+
+      - name: 'save cache dependencies'
+        id: cache-save
+        uses: actions/cache/save@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4
+        with:
+          path: 'cache/${{ matrix.name }}'
+          key: ${{ matrix.plat }}-${{ matrix.name }}-${{ hashFiles('vcpkg/installed/vcpkg/info/*', 'vcpkg/installed/vcpkg/updates/*') }}
+
       - name: 'cmake configure'
         timeout-minutes: 5
         shell: bash
         run: |
-          archgen=${{ matrix.arch }}; [ "${archgen}" = 'x86' ] && archgen='Win32'
           if [ '${{ matrix.plat }}' = 'uwp' ]; then
             system='WindowsStore'
             options='-DCMAKE_SYSTEM_VERSION=10.0'
           else
             system='Windows'
           fi
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
-          cmake -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
-            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
+          export VCPKG_ROOT="$PWD/vcpkg"
+          cmake . -B bld ${options} \
+            -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
+            -DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed \
+            -DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }} \
+            -DCMAKE_BUILD_TYPE=${{ matrix.type }} \
+            -DCMAKE_SYSTEM_NAME=$system \
             -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
-            '-DCMAKE_UNITY_BUILD=${{ matrix.unity }}' \
-            "-DCMAKE_C_FLAGS=${cflags}" \
-            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
             -DCMAKE_UNITY_BUILD=ON \
             -DCURL_WERROR=ON \
             -DBUILD_EXAMPLES=ON \
             -DENABLE_WEBSOCKETS=ON \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG= \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE= \
             ${{ matrix.config }} || { cat bld/CMakeFiles/CMake*.yaml; false; }
+
+      - name: 'prepare logs on failure'
+        timeout-minutes: 5
+        if: failure()
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$GITHUB_WORKSPACE/vcpkg"
+          mkdir -p bld
+          7z a -t7z -r -mx=9 logs.7z \
+            $VCPKG_ROOT/buildtrees/*.log \
+            bld/.ninja_log \
+            bld/build.ninja \
+            bld/install_manifest.txt \
+            bld/vcpkg-manifest-install.log \
+            bld/CMakeFiles/CMake*.yaml
+
+      - name: 'upload logs on failure'
+        timeout-minutes: 10
+        if: failure()
+        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4
+        with:
+          name: windows_logs_${{matrix.name}}_${{github.event.pull_request.head.sha}}
+          path: logs.7z
 
       - name: 'cmake build'
         timeout-minutes: 5
         shell: bash
         run: |
           cmake --build bld --config '${{ matrix.type }}' --parallel 5
-          [[ '${{ matrix.config }}' != *'BUILD_SHARED_LIBS=OFF'* ]] && cp -f -p bld/lib/*.dll bld/src/
+          export VCPKG_ROOT="$PWD/vcpkg"
+          export VCPKG_TRIPLED_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
+          export debug_lib="$VCPKG_TRIPLED_DIR/debug/bin"
+          export release_lib="$VCPKG_TRIPLED_DIR/bin"
+          export PATH="$PWD/bld/lib/:$release_lib:$debug_lib:$PATH"
           bld/src/curl.exe --disable --version
 
       - name: 'cmake build tests'
@@ -474,5 +551,9 @@ jobs:
         shell: bash
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets ${{ matrix.tflags }}'
-          ls bld/lib/*.dll >/dev/null 2>&1 && cp -f -p bld/lib/*.dll bld/tests/libtest/
+          export VCPKG_ROOT="$PWD/vcpkg"
+          export VCPKG_TRIPLED_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
+          export debug_lib="$VCPKG_TRIPLED_DIR/debug/bin"
+          export release_lib="$VCPKG_TRIPLED_DIR/bin"
+          export PATH="$PWD/bld/lib/:$release_lib:$debug_lib:$PATH"
           cmake --build bld --config '${{ matrix.type }}' --target test-ci
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -453,7 +453,7 @@ jobs:
           restore-keys: ${{ matrix.plat }}-${{ matrix.name }}
 
       - name: 'vcpkg init'
-        timeout-minutes: 5      
+        timeout-minutes: 5
         if: success()
         shell: bash
         run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -489,11 +489,11 @@ jobs:
           fi
           export VCPKG_ROOT="$PWD/vcpkg"
           cmake . -B bld ${options} \
-            -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
-            -DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed \
-            -DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }} \
-            -DCMAKE_BUILD_TYPE=${{ matrix.type }} \
-            -DCMAKE_SYSTEM_NAME=$system \
+            "-DCMAKE_SYSTEM_NAME=$system" \
+            "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake" \
+            "-DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed" \
+            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
+            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
             -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
             -DCMAKE_UNITY_BUILD=ON \
             -DCURL_WERROR=ON \
@@ -511,7 +511,7 @@ jobs:
           export VCPKG_ROOT="$GITHUB_WORKSPACE/vcpkg"
           mkdir -p bld
           7z a -t7z -r -mx=9 logs.7z \
-            $VCPKG_ROOT/buildtrees/*.log \
+            "$VCPKG_ROOT/buildtrees/*.log" \
             bld/.ninja_log \
             bld/build.ninja \
             bld/install_manifest.txt \
@@ -532,9 +532,9 @@ jobs:
         run: |
           cmake --build bld --config '${{ matrix.type }}' --parallel 5
           export VCPKG_ROOT="$PWD/vcpkg"
-          export VCPKG_TRIPLED_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
-          export debug_lib="$VCPKG_TRIPLED_DIR/debug/bin"
-          export release_lib="$VCPKG_TRIPLED_DIR/bin"
+          export VCPKG_TRIPLET_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
+          export debug_lib="$VCPKG_TRIPLET_DIR/debug/bin"
+          export release_lib="$VCPKG_TRIPLET_DIR/bin"
           export PATH="$PWD/bld/lib/:$release_lib:$debug_lib:$PATH"
           bld/src/curl.exe --disable --version
 
@@ -552,8 +552,8 @@ jobs:
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets ${{ matrix.tflags }}'
           export VCPKG_ROOT="$PWD/vcpkg"
-          export VCPKG_TRIPLED_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
-          export debug_lib="$VCPKG_TRIPLED_DIR/debug/bin"
-          export release_lib="$VCPKG_TRIPLED_DIR/bin"
+          export VCPKG_TRIPLET_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
+          export debug_lib="$VCPKG_TRIPLET_DIR/debug/bin"
+          export release_lib="$VCPKG_TRIPLET_DIR/bin"
           export PATH="$PWD/bld/lib/:$release_lib:$debug_lib:$PATH"
           cmake --build bld --config '${{ matrix.type }}' --target test-ci


.github/workflows/windows.yml
```diff
@@ -413,66 +413,134 @@ jobs:
...
+    name: 'msvc vcpkg (${{ matrix.name }})'
     runs-on: windows-latest
     timeout-minutes: 30
     strategy:
       matrix:
         include:
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=ON  -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=ON ', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=OFF', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON          -DENABLE_UNICODE=OFF', type: 'Debug' }
+        # name should be unique and without specail chars. It create a folder with that name for cache the libreries.
```
vszakats, line 422:
```suggestion
        # name should be unique and without special chars. It create a folder with that name for cache the libreries.
```

.github/workflows/windows.yml
```diff
@@ -413,66 +413,134 @@ jobs:
           cmake --build bld --config '${{ matrix.type }}' --target test-ci
 
   msvc:
-    name: 'msvc (${{ matrix.arch }}, ${{ matrix.plat }}, ${{ matrix.config }})'
+    name: 'msvc vcpkg (${{ matrix.name }})'
```
vszakats, line 416:
```suggestion
    name: 'msvc (${{ matrix.name }})'
```

.github/workflows/windows.yml
```diff
@@ -413,66 +413,134 @@ jobs:
...
+          type: Debug          
+          tflags: ~1516 ~2301 ~2302 ~2303 ~2307
+          config: -DBUILD_SHARED_LIBS=OFF -DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DENABLE_UNICODE=ON -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_OPENSSL_QUIC=ON
+        - name: no ssl
+          install: nghttp2 zstd zlib
+          arch: x64 
+          plat: windows
+          type: Debug
+          tflags: ~1516 ~2301 ~2302 ~2303 ~2307
+          config: -DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON  -DENABLE_UNICODE=OFF
```
vszakats, line 443: I suggest using single quotes for all the above strings. It helps syntax highlighters and linters.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,136 @@ jobs:
...
     timeout-minutes: 30
     strategy:
       matrix:
         include:
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=ON  -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=ON ', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=OFF', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON          -DENABLE_UNICODE=OFF', type: 'Debug' }
+        # name should be unique and without specail chars. It create a folder with that name for cache the libreries.
+        - name: schannel
+          install: nghttp2 zstd libssh2[openssl]
```
vszakats, line 424: Perhaps use libssh2-schannel instead? It seems like a waste to install openssl just for the sake of libssh2. Also mixing crypto backends is best to avoid.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,136 @@ jobs:
...
+          arch: x64
+          plat: windows
+          type: Debug          
+          tflags: ~1516 ~2301 ~2302 ~2303 ~2307
+          config: -DBUILD_SHARED_LIBS=OFF -DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=ON -DCURL_USE_LIBSSH2=ON
+        - name: openssl
+          install: nghttp2 nghttp3 zstd libssh2[openssl]
```
vszakats, line 431: Same as above.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,136 @@ jobs:
...
+          install: nghttp2 zstd zlib
+          arch: x64 
+          plat: windows
+          type: Debug
+          tflags: ~1516 ~2301 ~2302 ~2303 ~2307
+          config: -DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON  -DENABLE_UNICODE=OFF
       fail-fast: false
     steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - uses: actions/checkout@v4
```
vszakats, line 446: Please restore the original line. Otherwise it triggers a bot and we'll need to restore it in a separate PR.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,136 @@ jobs:
...
+
+      - name: Cache dependencies
+        uses: actions/cache/restore@v4
+        with:
+          path: "cache/${{ matrix.name }}"
```
vszakats, line 451: Single quotes

.github/workflows/windows.yml
```diff
@@ -413,66 +413,136 @@ jobs:
...
+          key: ${{ matrix.plat }}-${{ matrix.name }}
+          restore-keys: ${{ matrix.plat }}-${{ matrix.name }}
+
+      - name: Init vcpkg
+        if: success()
+        shell: bash
+        run: |
+          git clone https://github.com/microsoft/vcpkg/
```
vszakats, line 459: Add `--depth 1` to make this much faster?

.github/workflows/windows.yml
```diff
@@ -413,66 +413,136 @@ jobs:
...
+          export VCPKG_BINARY_SOURCES="clear;files,$GITHUB_WORKSPACE/$CURRENT_BINARY_CACHE,readwrite;"          
+          vcpkg/vcpkg x-set-installed ${{ matrix.install }} --triplet=${{ matrix.arch }}-${{ matrix.plat }}
+          echo vcpkg/installed/vcpkg/status:
+          ls vcpkg/installed/vcpkg/status
+
+      - name: Save cache dependencies
+        id: cache-save
+        uses: actions/cache/save@v4
+        with:
+          path: "cache/${{ matrix.name }}"
```
vszakats, line 477: Single quotes please

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
-          cmake -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
-            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
-            -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
-            '-DCMAKE_UNITY_BUILD=${{ matrix.unity }}' \
-            "-DCMAKE_C_FLAGS=${cflags}" \
-            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
```
vszakats, line 449: Can we keep the UWP logic?
Can we also keep the quotes, and `-DCMAKE_VS_GLOBALS=TrackFileAccess=false`.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
+          cmake . -B bld \
+            -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
+            -DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed \
+            -DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }} \
+            -DCMAKE_BUILD_TYPE=${{ matrix.type }} \
             -DCMAKE_UNITY_BUILD=ON \
             -DCURL_WERROR=ON \
             -DBUILD_EXAMPLES=ON \
             -DENABLE_WEBSOCKETS=ON \
-            ${{ matrix.config }} || { cat bld/CMakeFiles/CMake*.yaml; false; }
```
vszakats, line 454: Let's keep the .yaml dump logic in failure feature.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
+        run: |
+          export VCPKG_ROOT=$PWD/vcpkg
+          mkdir -p bld
+          7z a -t7z -r -mx=9 logs.7z \
+              $VCPKG_ROOT/buildtrees/*.log \
+              bld/.ninja_log \
+              bld/build.ninja \
+              bld/install_manifest.txt \
+              bld/vcpkg-manifest-install.log \
+              bld/CMakeFiles/CMake*.yaml
```
vszakats, line 510: Indentation

talregev:
I am closing this PR because I want to debug the cache, and I don't want to always activate curl ci. I will do it on my own ci.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - name: Cache dependencies
+        uses: actions/cache/restore@v4
+        with:
+          path: "cache/${{ matrix.name }}"
+          key: ${{ matrix.plat }}-${{ matrix.name }}
+          restore-keys: ${{ matrix.plat }}-${{ matrix.name }}
+
+      - name: Init vcpkg
```
vszakats, line 455: `vcpkg init`

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
+          path: logs.7z
 
       - name: 'cmake build'
         timeout-minutes: 5
         shell: bash
         run: |
-          cmake --build bld --config '${{ matrix.type }}' --parallel 5
-          [[ '${{ matrix.config }}' != *'BUILD_SHARED_LIBS=OFF'* ]] && cp -f -p bld/lib/*.dll bld/src/
+          export VCPKG_ROOT="$PWD/vcpkg"
+          cmake --build bld --config ${{ matrix.type }} --parallel 5
```
vszakats, line 524: Please keep existing quotes.

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
+          export PATH="$PWD/bld/lib/:$release_lib:$debug_lib:$PATH"
           bld/src/curl.exe --disable --version
 
       - name: 'cmake build tests'
         if: ${{ matrix.tflags != 'skipall' }}
         timeout-minutes: 10
         shell: bash
         run: |
-          cmake --build bld --config '${{ matrix.type }}' --parallel 5 --target testdeps
+          cmake --build bld --config ${{ matrix.type }} --parallel 5 --target testdeps
```
vszakats, line 536: Same here

.github/workflows/windows.yml
```diff
@@ -413,66 +413,137 @@ jobs:
...
 
       - name: 'cmake run tests'
         if: ${{ matrix.tflags != 'skipall' && matrix.tflags != 'skiprun' }}
         timeout-minutes: 30
         shell: bash
         run: |
-          export TFLAGS='-j14 !TFTP !MQTT !WebSockets ${{ matrix.tflags }}'
-          ls bld/lib/*.dll >/dev/null 2>&1 && cp -f -p bld/lib/*.dll bld/tests/libtest/
-          cmake --build bld --config '${{ matrix.type }}' --target test-ci
+          export TFLAGS="-j14 ${{ matrix.tflags }}"
```
vszakats, line 543: Why deleting existing global test exceptions?
Also let's keep quotes. It makes the diff easier to review if we omit style changes.

vszakats (changes requested):

.github/workflows/windows.yml
talregev: on vcpkg thereisn't libssh2 with schannel. only openssl.

talregev: I can remove the libssh2 from schannel build.

.github/workflows/windows.yml
talregev: here I install openssl anyway. for openssl tests.

.github/workflows/windows.yml
talregev: I remove them because it doing a problem if you try to switch to cmd shell. Sometimes I do that to test thing.
I can convert all the martrix.xxx to environment variables.

.github/workflows/windows.yml
talregev: What does -DCMAKE_VS_GLOBALS=TrackFileAccess=false do?
I will restore the uwp logic.

.github/workflows/windows.yml
talregev: When I remove these exception, I add more tests on the ci that we pass.

talregev:
on this line I must use double quotes. The ci will fail on one quote
https://github.com/curl/curl/pull/13979/files#diff-5ce5c9ff86f58b1f87b35b5227b2e84cb69f022d6741e1854f3e1e181091773cR468

.github/workflows/windows.yml
vszakats: Right!

.github/workflows/windows.yml
vszakats: That's a bummer. Agreeing to remove libssh2 then. It's tested in the other job anyway.

talregev:
On environment variable I must use  double quotes.
when I do like this:
export 'tal=$tal'
echo $tal
output: $tal

.github/workflows/windows.yml
vszakats: Keeping changes minimal and serving the CI config (and its shell) should get priority IMO.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+        shell: bash
+        run: |
+          git clone https://github.com/microsoft/vcpkg/ --depth 1
+          vcpkg/bootstrap-vcpkg.sh
+
+      - name: 'vcpkg build'
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$PWD/vcpkg"
+          export CURRENT_BINARY_CACHE="cache/${{ matrix.name }}"
```
vszakats, line 466: Single quotes is enough here. `${{ }}` is replaced before execution by GHA.

.github/workflows/windows.yml
vszakats: It disables internal administration done by MSBUILD, which makes the IDE (or subsequent builds?) work faster (as I understand).
This tracking work makes the CI job run slower and we don't use IDE, nor repeated builds, so it can be saved in this context.

vszakats:
Aside question: Few years ago I remember vcpkg was installing packages from source. Is this still true? On a cursory look the install command finished quickly that hinted this might no longer be the case. If so, do we need caching?


> On environment variables I must use double quotes. when I use single quotes it leave the $ symbol, and not give the value of the variable:
> 
> ```shell
> export 'tal=$tal'
> echo $tal
> ```
> 
> `output: $tal`

That's correct. (I wasn't suggesting single quotes in bash string with variables in them)

On the other hand `${{ }}` is replaced by GHA before execution, so those can be single-quoted to exclude the chance of an unwanted evaluation.

talregev:
> Aside question: Few years ago I remember vcpkg was installing packages from source. Is this still true? On a cursory look the install command finished quickly that hinted this might no longer be the case. If so, do we need caching?
> 
> > On environment variables I must use double quotes. when I use single quotes it leave the $ symbol, and not give the value of the variable:
> > ```shell
> > export 'tal=$tal'
> > echo $tal
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `output: $tal`
> 
> That's correct. (I wasn't suggesting single quotes in bash string with variables in them)
> 
> On the other hand `${{ }}` is replaced by GHA before execution, so those can be single-quoted to exclude the chance of an unwanted evaluation.

vcpkg still installing packages from source
Because of my cache implementation on the ci, it use the cache and install the binaries faster.
We need the cache for save ci time.

.github/workflows/windows.yml
```diff
@@ -473,6 +546,10 @@ jobs:
         timeout-minutes: 30
         shell: bash
         run: |
-          export TFLAGS='-j14 !TFTP !MQTT !WebSockets ${{ matrix.tflags }}'
-          ls bld/lib/*.dll >/dev/null 2>&1 && cp -f -p bld/lib/*.dll bld/tests/libtest/
+          export TFLAGS='-j14 ${{ matrix.tflags }}'
```
vszakats, line 549:
```suggestion
          export TFLAGS='-j14 !TFTP !MQTT !WebSockets ${{ matrix.tflags }}'
```

talregev: Why you don't want the other tests will run on the ci?

vszakats: They cause the tests to hang or fail, quite often. These exception were added after weeks of observing job runs and stabilizing them.

talregev: ok, Thank you for the explanations!

.github/workflows/windows.yml
talregev: ok.

.github/workflows/windows.yml
talregev: Are you sure the libssh2 that you use in other places (like linux mingw osx) not coming with openssl?
I think you don't aware of it.

.github/workflows/windows.yml
talregev: done.

.github/workflows/windows.yml
talregev: ok, Thank you for the explanations!

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
           fi
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
-          cmake -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
-            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
+          export VCPKG_ROOT="$PWD/vcpkg"
+          cmake . -B bld $options \
```
vszakats, line 489:
```suggestion
          cmake . -B bld ${options} \
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
             -DENABLE_WEBSOCKETS=ON \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG= \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE= \
             ${{ matrix.config }} || { cat bld/CMakeFiles/CMake*.yaml; false; }
 
+      - name: Prepare logs on failure
+        if: failure()
+        shell: bash
+        run: |
+          export VCPKG_ROOT=$PWD/vcpkg
```
vszakats, line 508:
```suggestion
          export VCPKG_ROOT="$PWD/vcpkg"
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+          install: 'nghttp2 zstd zlib'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516'
+          config: '-DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON  -DENABLE_UNICODE=OFF'
       fail-fast: false
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
```
vszakats, line 447:
```suggestion
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
         run: |
-          archgen=${{ matrix.arch }}; [ "${archgen}" = 'x86' ] && archgen='Win32'
           if [ '${{ matrix.plat }}' = 'uwp' ]; then
             system='WindowsStore'
             options='-DCMAKE_SYSTEM_VERSION=10.0'
           else
             system='Windows'
           fi
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
```
vszakats, line 440: Can we keep these two lines for consistency with the rest of the workflow?

We may optimize this globally in a separate commit. (It was done this way to avoid a cmake warning about unused variables, which doesn't seem to happen here for now.)

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: 'retore cache dependencies'
+        uses: actions/cache/restore@v4
+        with:
+          path: 'cache/${{ matrix.name }}'
+          key: ${{ matrix.plat }}-${{ matrix.name }}
+          restore-keys: ${{ matrix.plat }}-${{ matrix.name }}
+
+      - name: 'vcpkg init'
```
vszakats, line 455:
```suggestion
      - name: 'vcpkg init'
      - timeout-minutes: 5
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+        if: success()
+        shell: bash
+        run: |
+          git clone https://github.com/microsoft/vcpkg/ --depth 1
+          vcpkg/bootstrap-vcpkg.sh
+
+      - name: 'vcpkg build'
```
vszakats, line 462:
```suggestion
      - name: 'vcpkg build'
      - timeout-minutes: 5
```
Or any value you find fitting here. (Seems to take 30 seconds in current runs?)

vszakats (changes requested):

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG= \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE= \
             ${{ matrix.config }} || { cat bld/CMakeFiles/CMake*.yaml; false; }
 
+      - name: 'prepare logs on failure'
+        if: failure()
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$PWD/vcpkg"
+          mkdir -p bld
```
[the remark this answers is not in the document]
talregev: Not always. It can fail only the vcpkg build without the cmake generation.

.github/workflows/windows.yml
talregev: I preferer to do this on this PR. 
We already refactoring.
This is cleaner way to do things.

.github/workflows/windows.yml
talregev: 5 min it wrong. 
It can be 10min.
On the worst case that library is updated, it need to compile it from source, and it take time.
It can be more then 10 when we add different libs. I will change this to 10 min.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=OFF', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON          -DENABLE_UNICODE=OFF', type: 'Debug' }
+        # name should be unique and without special chars. It create a folder with that name for cache the libreries.
+        - name: 'schannel'
+          install: 'nghttp2 zstd zlib'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516'
+          config: '-DBUILD_SHARED_LIBS=OFF -DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=ON'
```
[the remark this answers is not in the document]
talregev: I didn't not understand your suggestion.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516'
+          config: '-DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON  -DENABLE_UNICODE=OFF'
       fail-fast: false
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: 'retore cache dependencies'
+        uses: actions/cache/restore@v4
```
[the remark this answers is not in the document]
talregev: It should clear that we do restore.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+          mkdir -p "$CURRENT_BINARY_CACHE"
+          export VCPKG_BINARY_SOURCES="clear;files,$GITHUB_WORKSPACE/$CURRENT_BINARY_CACHE,readwrite;"
+          vcpkg/vcpkg x-set-installed ${{ matrix.install }} --triplet=${{ matrix.arch }}-${{ matrix.plat }}
+
+      - name: 'save cache dependencies'
+        id: cache-save
+        uses: actions/cache/save@v4
+        with:
+          path: 'cache/${{ matrix.name }}'
+          key: ${{ matrix.plat }}-${{ matrix.name }}-${{ hashFiles('vcpkg/installed/vcpkg/info/*', 'vcpkg/installed/vcpkg/updates/*') }}
```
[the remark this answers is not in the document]
talregev: Yes we do, is the way of the vcpkg is working, it can update without change of the ci, that why we need to save it if it change.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
             options='-DCMAKE_SYSTEM_VERSION=10.0'
           else
             system='Windows'
           fi
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
-          cmake -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
```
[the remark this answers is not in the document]
talregev: No. vcpkg build system is handling this.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,125 @@ jobs:
...
+          type: 'Debug'
+          tflags: '~1516'
+          config: '-DBUILD_SHARED_LIBS=OFF -DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DENABLE_UNICODE=ON -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_OPENSSL_QUIC=ON'
+        - name: 'no ssl'
+          install: 'nghttp2 zstd zlib'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516'
+          config: '-DENABLE_DEBUG=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON  -DENABLE_UNICODE=OFF'
```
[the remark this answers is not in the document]
talregev: Thank you. I didn't understand your other suggestion. 
I think it refer to channel. Can you make it again?

.github/workflows/windows.yml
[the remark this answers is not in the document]
talregev: I figure it out.

.github/workflows/windows.yml
vszakats: 10 mins sounds fine; we can adjust it as needed. Point is to avoid a runaway (or hanging) step.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=ON  -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=ON ', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=OFF', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON          -DENABLE_UNICODE=OFF', type: 'Debug' }
+          # name should be unique and without special chars. It creates a folder with that name for the cache.
+        - name: 'schannel'
+          install: 'nghttp2 zstd zlib'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516'
```
vszakats, line 428:
```suggestion
          tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
```
Also for the two other lines below. These are flaky tests, and they will bite.

.github/workflows/windows.yml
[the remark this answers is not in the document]
vszakats: Yeah, but then all the bld files referenced from 7z would be missing also. Is there any reason to have an empty `bld` directory there?

Local tests say that 7z fails with 127 anyway. This makes it seems that a ` || true` is what's needed and the `mkdir` is redundant.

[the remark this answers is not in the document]
talregev: We have 2 cases.
First case: `vcpkg build`. In this case there is $VCPKG_ROOT/buildtrees/*.log.
If you will put  || true it will fail on create the log file, and upload nothing. 
On my case it will create empty dir, will upload   $VCPKG_ROOT/buildtrees/*.log and all other files, it will give warning that it not find them, it create the log.7z and upload it.

In the second case: cmake configure
The folder already exsit. the mkdir will not create folder and not fail. (because of the -p argument).
It will find the files and upload them.

.github/workflows/windows.yml
talregev: This tests is websocket tests. It will bite when you remove !websocket. As I understand you are not going to do that.

talregev: I will add them anyway.

.github/workflows/windows.yml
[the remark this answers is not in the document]
vszakats: Ok, got it and could also reproduce locally.

Is the `-p` option needed though? We're only creating a single-level subdir here.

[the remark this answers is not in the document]
talregev: Yes, because the folder can be exist. if you make a dir with exist folder without `-p` argument it fail. you can test it local on the bash:
```bash
mkdir tal
mkdir tal
mkdir: cannot create directory ‘tal’: File exists
```

.github/workflows/windows.yml
talregev: @vszakats libssh2 come only with openssl.
https://github.com/libssh2/libssh2/blob/master/CMakeLists.txt

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
           cmake --build bld --config '${{ matrix.type }}' --target test-ci
 
   msvc:
-    name: 'msvc (${{ matrix.arch }}, ${{ matrix.plat }}, ${{ matrix.config }})'
+    name: 'msvc (${{ matrix.name }})'
```
vszakats, line 416:
```suggestion
    name: 'msvc (${{ matrix.arch }}, ${{ matrix.name }})'
```

.github/workflows/windows.yml
vszakats: libssh2: not exactly, it also supports WinCNG (which is the native Win32 crypto API), mbedTLS and libgcrypt.

I think it's enough to test libssh2 with openssl here. An additional Schannel + libssh2/openssl build doesn't add any further extra coverage for the build or tests, like a libssh2/WinCNG build would.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_UNICODE=OFF', type: 'Debug' }
-          - { arch: 'x64', plat: 'windows', tflags: '~1516 ~2301 ~2302 ~2303 ~2307', config: '-DENABLE_DEBUG=ON -DCURL_USE_SCHANNEL=OFF -DHTTP_ONLY=ON          -DENABLE_UNICODE=OFF', type: 'Debug' }
+          # name should be unique and without special chars. It creates a folder with that name for the cache.
+        - name: 'schannel'
+          install: 'zstd zlib nghttp2'
+          arch: 'x64'
+          plat: 'windows'
+          type: 'Debug'
+          tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF'
```
vszakats, line 429:
```suggestion
          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON  -DCURL_ZSTD=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF'
```
Tiny whitespace nit for alignment.

talregev: ok.

vszakats (approved):

.github/workflows/windows.yml
talregev: Thank you for sharing with me this information!

talregev: Actually I check on vcpkg, there is an option with WinCNG backend. I will add it to Schannel.

talregev: I added.

vszakats: Nice, thanks!

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
+
+      - name: 'restore cache dependencies'
+        uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4
+        with:
+          path: 'cache/${{ matrix.name }}'
+          key: ${{ matrix.plat }}-${{ matrix.name }}
+          restore-keys: ${{ matrix.plat }}-${{ matrix.name }}
+
+      - name: 'vcpkg init'
+        timeout-minutes: 5
```
vszakats, line 456:
```suggestion
        timeout-minutes: 5
```
Delete line ending spaces here.

(I wish GitHub would make these visible.)

talregev-imagry: Can you make a script that check that?

talregev: Can you make a scrip for that?

vszakats: `checksrc` could probably do that. With the right amount of Perl-fu.

bagder: Nah, checksrc checks C code, having it checking yaml seems like an overstep and a job for a separate script.

vszakats: Yep, it's indeed usually not C code that is hit. I have a script doing it locally, but written in a foreign language:
https://github.com/vszakats/hb/blob/83173b807b4eceda0a46704aa6544e92240cc21c/bin/commit.hb#L579-L656

talregev: we can have python script that sit inside script job that check yaml format.

vszakats: We could reasonably go for _all_ files (not just YAML) with some exclusions (e.g. test definitions, certs), and allowing tabs in some make files.

Sticking with Perl would make sense to avoid adding an extra build dependency. Also some of the necessary logic is already implemented in `checksrc`. That could be reused for a whitespace checker.

vszakats: The checker landed in 1ccdad64efecbf3167d6f60737037157011c1468 #14031.

talregev: I will rebase it to here.
This PR is count as a bugfix? that it can be merge to curl in feature freeze time?
Because I don't change the curl at all ....

vszakats: Yes, I thnk it's fine to merge CI changes outside the feature window.

What new test cases are you thinking of?

talregev: > What new test cases are you thinking of?

When I added here the other libraries, your tests cases was activate and check more cases on msvc jobs. that what I meant.
But since you ask, I am planning to add more working libraries from vcpkg. and add wup case.
But this will be next PR. as soon this one will be merge.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
           fi
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
-          cmake -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
-            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
+          export VCPKG_ROOT="$PWD/vcpkg"
+          cmake . -B bld ${options} \
```
vszakats, line 491:
```suggestion
          cmake -B bld ${options} \
```

vszakats: The dot is the default and not necessary in this case.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
+            -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
+            -DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed \
+            -DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }} \
+            -DCMAKE_BUILD_TYPE=${{ matrix.type }} \
+            -DCMAKE_SYSTEM_NAME=$system \
```
vszakats, line 496:
```suggestion
            "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake" \
            "-DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed" \
            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
            "-DCMAKE_SYSTEM_NAME=${system}" \
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
 
+      - name: 'prepare logs on failure'
+        timeout-minutes: 5
+        if: failure()
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$GITHUB_WORKSPACE/vcpkg"
+          mkdir -p bld
+          7z a -t7z -r -mx=9 logs.7z \
+            $VCPKG_ROOT/buildtrees/*.log \
```
vszakats, line 514:
```suggestion
            "$VCPKG_ROOT"/buildtrees/*.log \
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
       - name: 'cmake build'
         timeout-minutes: 5
         shell: bash
         run: |
           cmake --build bld --config '${{ matrix.type }}' --parallel 5
-          [[ '${{ matrix.config }}' != *'BUILD_SHARED_LIBS=OFF'* ]] && cp -f -p bld/lib/*.dll bld/src/
+          export VCPKG_ROOT="$PWD/vcpkg"
+          export VCPKG_TRIPLED_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
+          export debug_lib="$VCPKG_TRIPLED_DIR/debug/bin"
+          export release_lib="$VCPKG_TRIPLED_DIR/bin"
```
vszakats, line 537:
```suggestion
          export VCPKG_TRIPLET_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
          export debug_lib="$VCPKG_TRIPLET_DIR/debug/bin"
          export release_lib="$VCPKG_TRIPLET_DIR/bin"
```

.github/workflows/windows.yml
```diff
@@ -474,5 +551,9 @@ jobs:
         shell: bash
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets ${{ matrix.tflags }}'
-          ls bld/lib/*.dll >/dev/null 2>&1 && cp -f -p bld/lib/*.dll bld/tests/libtest/
+          export VCPKG_ROOT="$PWD/vcpkg"
+          export VCPKG_TRIPLED_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
+          export debug_lib="$VCPKG_TRIPLED_DIR/debug/bin"
+          export release_lib="$VCPKG_TRIPLED_DIR/bin"
```
vszakats, line 557:
```suggestion
          export VCPKG_TRIPLET_DIR="$VCPKG_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
          export debug_lib="$VCPKG_TRIPLET_DIR/debug/bin"
          export release_lib="$VCPKG_TRIPLET_DIR/bin"
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG= \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE= \
             ${{ matrix.config }} || { cat bld/CMakeFiles/CMake*.yaml; false; }
 
+      - name: 'prepare logs on failure'
+        timeout-minutes: 5
+        if: failure()
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$GITHUB_WORKSPACE/vcpkg"
```
vszakats, line 511:
```suggestion
          export VCPKG_ROOT="$PWD/vcpkg"
```
Is there a particular reason to prefer `$GITHUB_WORKSPACE` here, over `$PWD` that is used elsewhere?

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
+            -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake \
+            -DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed \
+            -DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }} \
+            -DCMAKE_BUILD_TYPE=${{ matrix.type }} \
+            -DCMAKE_SYSTEM_NAME=$system \
             -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
-            '-DCMAKE_UNITY_BUILD=${{ matrix.unity }}' \
-            "-DCMAKE_C_FLAGS=${cflags}" \
-            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
             -DCMAKE_UNITY_BUILD=ON \
```
vszakats, line 498:
```suggestion
            "-DCMAKE_C_FLAGS=${cflags}" \
            '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
            -DCMAKE_UNITY_BUILD=ON \
```

.github/workflows/windows.yml
talregev: even if it default, it more clear to do that. but if you insist, I will remove.

.github/workflows/windows.yml
talregev: Yes. I try before with $PWD, but `7z` don't accept that folder, and needed a windows type folder
aka 'D:/....'

.github/workflows/windows.yml
talregev: No. This is wrong. All the flags should be handle inside vcpkg build system.

.github/workflows/windows.yml
vszakats: Please do, I deleted those purposefully, hence missing from all other calls too.

.github/workflows/windows.yml
vszakats: Makes sense, thanks for the explanation.

.github/workflows/windows.yml
vszakats: Edited to exclude it. I'm thinking of CPPFLAGS, which vcpkg has no knowledge of. Let's get back to this when there is a specific flag we want to pass. Ideally this won't happen.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
             system='Windows'
           fi
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
+          export VCPKG_ROOT="$PWD/vcpkg"
           cmake -B bld ${options} \
-            "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
+            "-DCMAKE_SYSTEM_NAME=$system" \
```
vszakats, line 492:
```suggestion
            "-DCMAKE_SYSTEM_NAME=${system}" \
```

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
 
+      - name: 'prepare logs on failure'
+        timeout-minutes: 5
+        if: failure()
+        shell: bash
+        run: |
+          export VCPKG_ROOT="$GITHUB_WORKSPACE/vcpkg"
+          mkdir -p bld
+          7z a -t7z -r -mx=9 logs.7z \
+            "$VCPKG_ROOT/buildtrees/*.log" \
```
vszakats, line 514:
```suggestion
            "$VCPKG_ROOT"/buildtrees/*.log \
```
`*` needs to stay outside of double-quotes to expand

.github/workflows/windows.yml
talregev: No. I insist. if you really want to change cppflags, you can do it with environment variables:
https://stackoverflow.com/questions/5541946/cflags-ccflags-cxxflags-what-exactly-do-these-variables-control
https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html

Also all needed flags should be handle inside the cmake. also vcpkg build system give you the flags you need for the system. Let's no go there.

vszakats (changes requested):

.github/workflows/windows.yml
talregev: It windows style, you should give all the path protected with double quotes.

.github/workflows/windows.yml
vszakats: ? ...I deleted that line.

As for the env, it'd be nice though I never managed to make it work. We also need to pass to C, not C++.

talregev: l don't want to do it even with environment variables.
What I am saying that if there is a problem with cppflags, we should solve it with the cmake and correct it.

.github/workflows/windows.yml
vszakats: This is bash code. Let's stick to bash syntax!

talregev: I think the 7z do the expansion as well.

talregev: I can test it on my own ci.

vszakats: It may be able to do the expansion, but when using bash, let's do it the bash way.

talregev: It doing the expansion, but doing it flat. Better as you suggest.

.github/workflows/windows.yml
talregev: Thank you. I resolve this thread.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
-          [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
-          [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
+          export VCPKG_ROOT="$PWD/vcpkg"
           cmake -B bld ${options} \
             "-DCMAKE_SYSTEM_NAME=${system}" \
-            -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            "-DCMAKE_GENERATOR_PLATFORM=${archgen}" \
+            "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake" \
+            "-DVCPKG_INSTALLED_DIR=$VCPKG_ROOT/installed" \
             '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
```
vszakats, line 495:
```suggestion
            '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
            -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
```
Move this line back to its original posision for a clean diff.

.github/workflows/windows.yml
```diff
@@ -413,52 +413,129 @@ jobs:
...
-            -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
-            '-DCMAKE_UNITY_BUILD=${{ matrix.unity }}' \
-            "-DCMAKE_C_FLAGS=${cflags}" \
             '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
+            -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
```
vszakats, line 497:
```suggestion
```
See above.

.github/workflows/windows.yml
talregev: It doing the same, I will do as you suggested that it will not confuse.

.github/workflows/windows.yml
talregev: done.

.github/workflows/windows.yml
talregev: done.

.github/workflows/windows.yml
talregev: done.

talregev:
@vszakats Can you approve the PR?

vszakats (approved):

talregev:
@vszakats Thank you!

talregev:
@bagder Can you take a look on this PR?

bagder (commented):
Sorry, I cannot add anything here. Windows and cmake are two weak points of mine. @vszakats is an appropriate authority here!

vszakats:
Thank you @talregev, merged now!

PR #13984, opened 2024-06-23
CURLOPT_RESOLVE: support "*" for port number

bagder:
Allow resolves for a hostname independent of port number, to a fixed address. Resolves to a fixed port number have priority.

Also works if the hostname itself is using "*" to catch all hostnames.

A key to this change is to no longer use the port number from the DNS struct stored in the cache when connecting since it is not the correct one when a wildcard port is used for CURLOPT_RESOLVE.

Removes TODO item 1.8

Verified by test 1488 and 1489

### DIFF
--- a/docs/TODO
+++ b/docs/TODO
@@ -22,7 +22,6 @@
  1.3 struct lifreq
  1.4 alt-svc sharing
  1.5 get rid of PATH_MAX
- 1.8 CURLOPT_RESOLVE for any port number
  1.9 Cache negative name resolves
  1.10 auto-detect proxy
  1.11 minimize dependencies with dynamically loaded modules
@@ -246,14 +245,6 @@
  Currently the libssh2 SSH based code uses it, but to remove PATH_MAX from
  there we need libssh2 to properly tell us when we pass in a too small buffer
  and its current API (as of libssh2 1.2.7) does not.
-
-1.8 CURLOPT_RESOLVE for any port number
-
- This option allows applications to set a replacement IP address for a given
- host + port pair. Consider making support for providing a replacement address
- for the hostname on all port numbers.
-
- See https://github.com/curl/curl/issues/1264
 
 1.9 Cache negative name resolves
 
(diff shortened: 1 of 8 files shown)


bagder:
@icing I'm not too happy with how I get the port number in `Curl_cf_tcp_create`, maybe you have a better idea there? It needs to get the secondary port number for FTP and otherwise pass on the primary.

icing:
> @icing I'm not too happy with how I get the port number in `Curl_cf_tcp_create`, maybe you have a better idea there? It needs to get the secondary port number for FTP and otherwise pass on the primary.

Yeah, this looks a bit ugly. As I understand it, the Curl_resolve() always gave a `Curl_dns_entry` that contained a `Curl_addrinfo` that carried the ip address and port. And that port was always the port that was asked for when resolving.

Now we have a wildport(?) `Curl_dns_entry` with port -1 and the information about the port we want to resolve is lost when that is passed on. That's why you added the extra port parameter to overwrite that later when the actual socket is being created for `addr`.

I think it would be better if `hostip.c`'s `fetch_addr()` returns a copy of a wildport dns entry that carries the correct port in its address infos.

bagder:
> I think it would be better if hostip.c's fetch_addr() returns a copy of a wildport dns entry that carries the correct port in its address infos.

Right, I was pondering on that as well. I'll try that, even if that means having to copy the buffer there instead of just referencing it.

bagder:
As was [pointed out to me](https://woodland.cafe/@untitaker/112665800860371534), this can already be accomplished with `--connect-to`:

    curl --connect-to ::127.0.0.1: example.com

Maybe I could still support this syntax on the cmdline, and convert it to this `--connect-to` version under the hood. Also saves us from having to change libcurl.

bagder:
Or maybe I just document it better...

bagder:
#13989

bagder:
Let's ditch this PR since we can accomplish this already!

PR #13987, opened 2024-06-23
tool_getparam: make --show-headers the same as --include

bagder:
Simply a name alias that better explains what the option does.


fogti:
maybe it would be a good idea to include in the option name that it particularly concerns response headers (from the name alone I would've expected that it shows both request headers and response headers)

bagder:
> maybe it would be a good idea to include in the option name that it particularly concerns response headers

Can you think of such a name that does not also make it veeeeery long?

PR #13993, opened 2024-06-23
curl: add --skip-existing

bagder:
With this option, the entire download is skipped if the selected target filename already exists when the operation is about to begin. Also works fine with "globs".

- [X] code
- [X] documentation
- [X] test cases (993, 994, 995)

Ref: #11012


yedayak:
This seems to empty the output file, instead of just doing nothing:
```
$ echo hello > foo.txt
$ cat foo.txt
hello
$ ./src/curl https://google.com -v --skip-existing -o foo.txt
Note: skips transfer, "foo.txt" exists locally
$ cat foo.txt
$
```

bagder:
Oh. I'll check it out, that certainly looks wrong.

bagder:
> This seems to empty the output file, instead of just doing nothing:

Yeps, it did wrong - fixed now. I updated test 993 to verify the content of existing file after the transfer to make sure it does not do this anymore.

I also added test 995 to verify the case where the file does not exist before the transfer start.

PR #14008, opened 2024-06-24
wolfssl: assume key_file equal to clientcert in the absence of key_file

### LINKED ISSUE #14007: wolfssl backend behavior differs from openssl when CURLOPT_SSLCERT is used but CURLOPT_SSLKEY wasn't
### I did this

I tried running `curl --cert <path> <url_requiring_client_cert>` which works with openssl backend but fails with wolfssl.

The reason being that openssl backend will assume key_file is the same as cert_file (code [here](https://github.com/curl/curl/blob/master/lib/vtls/openssl.c#L1510)) which wolfssl makes no such assumption (code [here](https://github.com/curl/curl/blob/master/lib/vtls/wolfssl.c#L762)).

### I expected the following

I expect wolfssl backend to assume key_file is the same as cert_file, which means both ssl backend show behave the same.

### curl/libcurl version

curl 8.8.0

### operating system

All

alexsn:
When user sets CURLOPT_SSLCERT but leaves CURLOPT_SSLKEY unset assume the path passed in CURLOPT_SSLCERT holds the ssl key which is what we do in openssl implementation.

Closed #14007

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -759,7 +759,8 @@ wolfssl_connect_step1(struct Curl_cfilter *cf, struct Curl_easy *data)
 
 #ifndef NO_FILESYSTEM
   /* Load the client certificate, and private key */
-  if(ssl_config->primary.clientcert && ssl_config->key) {
+  if(ssl_config->primary.clientcert) {
+    char *key_file = ssl_config->key;
     int file_type = do_file_type(ssl_config->cert_type);
 
     if(file_type == WOLFSSL_FILETYPE_PEM) {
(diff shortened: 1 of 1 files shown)


bagder:
~~~
./vtls/wolfssl.c:763:10: warning: space after declarative asterisk (ASTERISKSPACE)
     char* key_file = ssl_config->key;
          ^
./vtls/wolfssl.c:763:9: warning: no space before asterisk (ASTERISKNOSPACE)
     char* key_file = ssl_config->key;
         ^

alexsn:
Fixed spacing lint

bagder (approved):

bagder:
Thanks!

PR #14018, opened 2024-06-25
wolfssl: add support for ssl cert blob / ssl key blob options

alexsn:
wolfSSL supports setting certificates/private keys from memory blobs which allow us to implement both CURLOPT_SSLCERT_BLOB and CURLOPT_SSLKEY_BLOB options.

### DIFF
--- a/docs/libcurl/opts/CURLOPT_SSLCERT_BLOB.md
+++ b/docs/libcurl/opts/CURLOPT_SSLCERT_BLOB.md
@@ -15,6 +15,7 @@ TLS-backend:
   - Secure Transport
   - Schannel
   - mbedTLS
+  - wolfSSL
 Added-in: 7.71.0
 ---
 
--- a/docs/libcurl/opts/CURLOPT_SSLKEY_BLOB.md
+++ b/docs/libcurl/opts/CURLOPT_SSLKEY_BLOB.md
@@ -11,6 +11,7 @@ Protocol:
   - TLS
 TLS-backend:
   - OpenSSL
+  - wolfSSL
 Added-in: 7.71.0
 ---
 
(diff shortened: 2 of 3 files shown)


docs/libcurl/opts/CURLOPT_SSLCERT_BLOB.md
```diff
@@ -78,7 +79,7 @@ int main(void)
 # AVAILABILITY
 
 Added in 7.71.0. This option is supported by the OpenSSL, Secure Transport,
-Schannel and mbedTLS (since 7.78.0) backends.
+Schannel, mbedTLS (since 7.78.0) and wolfSSL (since 8.9.0) backends.
```
bagder, line 82: Please use the exact phrase `(added in [version])` for the version information, to allow our automation to work. It can selectively hide this information in the output. The mbedTLS details should be marked up like that as well, the existing way is just an omission.

docs/libcurl/opts/CURLOPT_SSLKEY_BLOB.md
```diff
@@ -79,7 +80,8 @@ int main(void)
 
 # AVAILABILITY
 
-Added in 7.71.0. This option is supported by the OpenSSL backends.
+Added in 7.71.0. This option is supported by the OpenSSL and
+wolfSSL (since 8.9.0) backends.
```
bagder, line 84: `Added in...`

bagder:
We qualify this as a *change* (not bugfix), which means it needs to wait for the feature window to open before we can merge it. These options will therefore not be supported with wolfSSL in 8.9.0 but most likely in 8.10.0.

bagder (approved):

alexsn:
> We qualify this as a _change_ (not bugfix), which means it needs to wait for the feature window to open before we can merge it. These options will therefore not be supported with wolfSSL in 8.9.0 but most likely in 8.10.0.

Should I update this pull request to say 8.10.0 in the docs?

bagder:
> Should I update this pull request to say 8.10.0 in the docs?

Yes please!

bagder:
@alexsn can you please fix the merge conflicts in the docs and force-push?

PR #14019, opened 2024-06-25
system_win32: add missing curl.h include

MarcelRaad:
It's required for `CURLcode`.

### DIFF
--- a/lib/system_win32.h
+++ b/lib/system_win32.h
@@ -28,6 +28,8 @@
 
 #ifdef _WIN32
 
+#include <curl/curl.h>
+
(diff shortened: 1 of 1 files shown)


MarcelRaad:
Yes, it shows up in both of my autobuild runners:
https://curl.se/dev/log.cgi?id=20240626040432-3047891#prob1
Probably because of the `--disable-socketpair` as socketpair.h is the only include before system_win32.h in https://github.com/curl/curl/blob/72abf7c13a479edcde80afa60faad3f35f672c0b/lib/asyn-thread.c#L57.

bagder (approved):

PR #14026, opened 2024-06-26
OpenSSL QUIC, require at least v3.3.0

icing:
- when checking for QUIC support in OpenSSL, also check for it being at least 3.3.0
- remove workarounds for features buggy or missing in 3.2


bagder (approved):

talregev:
this mean openssl + nghttp3 will not be experimental anymore?

talregev:
Please update the cmake as well.

bagder:
> this mean openssl + nghttp3 will not be experimental anymore?

No. This means that it requires OpenSSL 3.3 to build and that OpenSSL 3.2 is no longer enough to enable QUIC with the OpenSSL QUIC backend.

PR #14034, opened 2024-06-27
conncache: done always evaluates to false

bagder:
Follow-up to c9b95c0bb30f88bf00e1ac

Spotted by CodeSonar

### DIFF
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -689,7 +689,7 @@ static void connc_discard_conn(struct conncache *connc,
    * not what we want. */
   if(aborted)
     done = TRUE;
-  else if(!done) {
+  else {
     /* Attempt to shutdown the connection right away. */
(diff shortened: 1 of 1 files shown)


lib/conncache.c
```diff
@@ -689,7 +689,7 @@ static void connc_discard_conn(struct conncache *connc,
    * not what we want. */
   if(aborted)
     done = TRUE;
-  else if(!done) {
+  else {
```
icing, line 692: I think this should be `if(!done) {` since we do not run the shutdown for aborted connections.

icing (changes requested):

lib/conncache.c
bagder: fair!

icing (approved):

PR #14037, opened 2024-06-27
synopsis.yml: fixup is triggered by `.md` files

### LINKED ISSUE #14032: CI: synopsis workflow no longer running
### I did this

Noticed that the SYNOPSIS workflow stopped running:
https://github.com/curl/curl/actions/workflows/synopsis.yml

It would trigger on `.3` file updates, but we replaced them
with `.md` files a while back.

`.github/scripts/verify-synopsis.pl` is unused too as a result.

### I expected the following

The workflow to run.

### curl/libcurl version

master

### operating system

N/A

pheiduck:
Fixes #14032
Closes #14037

### DIFF
--- a/.github/workflows/synopsis.yml
+++ b/.github/workflows/synopsis.yml
@@ -11,11 +11,15 @@ on:
       - '*/ci'
     paths:
       - 'docs/libcurl/curl_*.md'
+      - 'scripts/*.pl'
+      - '.github/workflows/*.yml'
   pull_request:
     branches:
       - master
     paths:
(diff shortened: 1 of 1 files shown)


bagder:
Am I blind or does this job not run here?

Should the `paths:` also include the script and the yaml files?

vszakats:
> Am I blind or does this job not run here?
> 
> Should the `paths:` also include the script and the yaml files?

Yes, otherwise it only triggers when a .md file is updated.

bagder (approved):

bagder:
Thanks!

PR #14041, opened 2024-06-27
GHA: add --enable-werror to the quiche job

bagder:


.github/workflows/http3-linux.yml
```diff
@@ -294,6 +294,7 @@ jobs:
               LDFLAGS="-Wl,-rpath,/home/runner/quiche/target/release"
               --with-openssl=/home/runner/quiche/quiche/deps/boringssl/src
               --enable-debug
+              --enable-werror
```
vszakats, line 297:
```suggestion
              --enable-warnings --enable-werror --enable-debug
              --with-openssl=/home/runner/quiche/quiche/deps/boringssl/src
```
Would it make sense to also level the warnings to the other jobs?

vszakats: Well, `--enable-warnings` might be the default — autotools defaults confuse me every single time. In that case the patch above may just sync the options visually with the other jobs, for clarity!

bagder: The default is complicated because it is not a clean on/off; there's logic.

But the main point here is perhaps that --enable-debug enables warnings.

vszakats: Yeah, I still think it'd be more readable to use the same line for all jobs here. Even if functionally the same.

PR #14056, opened 2024-06-29
Fix UDP_GRO struct cmsghdr data type

tatsuhiro-t:
The data type for UDP_GRO in struct cmsghdr is int.  Limit the usage of UDP_GRO to linux only because it is not portable.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -1899,7 +1899,8 @@ static CURLcode cf_udp_setup_quic(struct Curl_cfilter *cf,
 #endif
   }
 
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Reference:

- https://github.com/torvalds/linux/commit/436864095a95fcc611c20c44a111985fa9848730
- https://github.com/torvalds/linux/blob/de0a9f4486337d0eabacc23bd67ff73146eacdc0/tools/testing/selftests/net/udpgso_bench_rx.c#L239

tatsuhiro-t:
The previous code works on little endian machines because CMSG_SPACE(sizeof(uint16_t)) == CMSG_SPACE(sizeof(int)).

bagder (approved):

PR #14059, opened 2024-06-29
curl: support embedding a CA bundle

vszakats:
Add the ability to embed a CA bundle into the curl binary. It is used
when no other runtime or build-time option set one.

This helps curl-for-win macOS and Linux builds to run standalone, and
also helps Windows builds to avoid picking up the CA bundle from an
arbitrary (possibly world-writable) location (though this behaviour is
not currently disablable).

Usage:
- cmake: `-DCURL_CA_EMBED=/path/to/curl-ca-bundle.crt`
- autotools: `--with-ca-embed=/path/to/curl-ca-bundle.crt`
- Makefile.mk: `CURL_CA_EMBED=/path/to/curl-ca-bundle.crt`

Also add new command-line option `--dump-ca-embed` to dump the embedded
CA bundle to standard output.

Closes #14059

---

TODO:
- [x] rebase on #14182
- [x] skip embedding if no TLS backend support it (both autotools and cmake)

Ideas:
- ~~gzipped embedding? (230KB → 130KB)~~
  with reusable unzipper function inside `src/tool_something.c`
- move hugehelp to use that too? (ATM the unzip code is part of the generated source)
- move up cacert embedding to libcurl?
- strip CR bytes before embedding for better reproducibility?
- ~~generate the dummy `tool_ca_embed.c` also in CMake/Makefile.mk to keep binary output in sync? (though the dummy code is likely discarded by the linker in autotools builds.)~~
- build-time option to disable `FindWin32CACert()` feature?


src/tool_operate.c
```diff
@@ -1658,6 +1662,23 @@ static CURLcode single_transfer(struct GlobalConfig *global,
...
+          curl_easy_setopt(curl, CURLOPT_CAINFO_BLOB, &blob);
+        }
+        if(!config->proxy_cacert && !config->proxy_capath) {
+          struct curl_blob blob;
+          blob.data = (void *)curl_ca_embed;
+          blob.len = strlen((const char *)curl_ca_embed);
+          blob.flags = CURL_BLOB_NOCOPY;
+          curl_easy_setopt(curl, CURLOPT_PROXY_CAINFO_BLOB, &blob);
+        }
+#endif
```
bagder, line 1680: You should check the return codes from the setopt() calls.

As [our TLS option table shows](https://curl.se/libcurl/c/tls-options.html), these options are not supported by all TLS backends so there's a risk someone builds with this but then it fails at run-time because libcurl cannot comply!

bagder (commented):
It might make sense to display this as a feature in curl -V so that a user can tell if it has a cacert bundle built-in or not.

src/tool_operate.c
vszakats: Good idea to catch it here indeed.

src/tool_help.c
```diff
@@ -215,6 +215,9 @@ void tool_version_info(void)
     printf("Features:");
     for(builtin = feature_names; *builtin; ++builtin)
       printf(" %s", *builtin);
+#ifdef CURL_CA_EMBED
+    printf(" CAcert");
+#endif
```
vszakats, line 220: Is there a succint way to print it at its sorted position?

This is almost there except one corner case where it's shown first:
```c
    for(builtin = feature_names; *builtin; ++builtin) {
#ifdef CURL_CA_EMBED
      const char *ca_embed_feat = "CAcert";
      if(struplocompare(ca_embed_feat, *builtin) < 0 || !*(builtin+1)) {
        printf(" %s", ca_embed_feat);
      }
#endif
      printf(" %s", *builtin);
    }
```

bagder: What corner case?

If that that can't be fixed easy enough, I suppose simply qsorting the list is another option.

bagder: If we ever want to add a second feature in the tool to that list we would probably prefer to sort it anyway.

vszakats: The corner case is if there is a single feature in the list and that's starting with an A or B. Output ending up as `CAcert alt-svc` for example.

vszakats:
Do we want to fall back to the old malloc-less feature display in case `malloc()` fails? (this would omit the tool-only optional features, that is `CAcert` for now):
```diff
--- a/src/tool_help.c
+++ b/src/tool_help.c
@@ -214,6 +214,7 @@ void tool_version_info(void)
   if(feature_names[0]) {
     const char **feat_ext;
     size_t feat_ext_count = feature_count;
+    printf("Features:");
 #ifdef CURL_CA_EMBED
     ++feat_ext_count;
 #endif
@@ -228,12 +229,15 @@ void tool_version_info(void)
       feat_ext[feat_ext_count] = NULL;
       qsort((void *)feat_ext, feat_ext_count, sizeof(*feat_ext),
             struplocompare4sort);
-      printf("Features:");
       for(builtin = feat_ext; *builtin; ++builtin)
         printf(" %s", *builtin);
-      puts(""); /* newline */
       free((void *)feat_ext);
     }
+    else {
+      for(builtin = feature_names; *builtin; ++builtin)
+        printf(" %s", *builtin);
+    }
+    puts(""); /* newline */
   }
   if(strcmp(CURL_VERSION, curlinfo->version)) {
     printf("WARNING: curl and libcurl versions do not match. "
```

bagder:
> Do we want to fall back to the old malloc-less feature display in case `malloc()` fails?

When malloc fails (which is basically not possible on most modern systems) we should bail out. Not try something else. IMHO.

vszakats:
Thanks, I agree, leaving it as-is.

On another question: Do we want a `--dump-embedded-cacert` option (which would write the embedded CA bundle to stdout and quit)? Any preference for the option name?

It'd be useful for transparency, i.e. to see the trusted root without peeking into the binary with a capable viewer (or `strings`).

src/tool_operate.c
```diff
@@ -1658,6 +1662,31 @@ static CURLcode single_transfer(struct GlobalConfig *global,
             break;
         }
 
+#ifdef CURL_CA_EMBED
+        if(!config->cacert && !config->capath) {
+          struct curl_blob blob;
+          blob.data = (void *)curl_ca_embed;
+          blob.len = strlen((const char *)curl_ca_embed);
+          blob.flags = CURL_BLOB_NOCOPY;
+          result = curl_easy_setopt(curl, CURLOPT_CAINFO_BLOB, &blob);
```
yedayak, line 1671: Maybe we should log here in verbose mode? Currently the log from this is `successfully imported CA certificate blob` from `lib/vtls/openssl.c` and `lib/vtls/wolfssl.c`, but it will be good if it's more clear we're using a bundled CA (and not from somewhere else)

vszakats: Agreed this would be a nice touch. But how to make a verbose output from `src`? I could not figure it out.

vszakats: Ah, perhaps `notef()`?

yedayak: > Ah, perhaps `notef()`?

Hmm, Yeah that seems correct

vszakats: Pushed a commit with that call.

src/CMakeLists.txt
```diff
@@ -54,6 +54,22 @@ endif()
 transform_makefile_inc("Makefile.inc" "${CMAKE_CURRENT_BINARY_DIR}/Makefile.inc.cmake")
 include(${CMAKE_CURRENT_BINARY_DIR}/Makefile.inc.cmake)
 
+if(CURL_CA_EMBED_SET)
+  if(PERL_FOUND)
+    add_definitions("-DCURL_CA_EMBED")
+    add_custom_command(
+      OUTPUT tool_ca_embed.c
+      COMMAND "${PERL_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/mk-file-embed.pl" --var curl_ca_embed < "${CURL_CA_EMBED}" > tool_ca_embed.c
```
yedayak, line 62: `tool_ca_embed.c` should be in `src/.gitignore`, like `tool_hugehelp.c`

vszakats: Good point! Added.

bagder:
> Do we want a `--dump-embedded-cacert` option (which would write the embedded CA bundle to stdout and quit)? Any preference for the option name?

I think such a command line option makes a lot of sense, yes. I don't have any better name suggestion!

vszakats:
I went with `--dump-ca-embed`. This rhymes with the rest of new symbols added. Not too long.

vszakats:
Patch that allows disabling the Windows-specific `PATH`-search for a CA bundle:
```diff
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -621,7 +621,7 @@ CURLcode FindWin32CACert(struct OperationConfig *config,
 {
   CURLcode result = CURLE_OK;
 
-#ifdef CURL_WINDOWS_APP
+#if defined(CURL_WINDOWS_APP) || defined(CURL_WINDOWS_SKIP_SEARCHPATH_CACERT)
   (void)config;
   (void)backend;
   (void)bundle_file;
```
This option becomes useful when combined with embedded bundles, as it prevents curl from picking one up from a "random" location on disk, including world-writable ones. It still allows to specify one manually either via the documented envs or command-line options, overriding the embedded default.

Does this make sense? Does it raise any concern?

vszakats:
This option is now tested via curl-for-win. Also in this PR, after this commit and re-running the workflow: https://github.com/curl/curl-for-win/commit/b4945dd81600e053e7a4f9c676d6011b8f8fda92

bagder:
> gzipped embedding? (230KB → 130KB)

A downside with compressing this:

Since the TLS libraries cannot use the blob compressed, this would then have to use 230KB allocated memory to store the cacert bundle in, in addition to the 130KB compressed read-only memory already used for the compressed image. Thus spending more total memory in run-time than doing it completely uncompressed will need...

vszakats:
I agree it wouldn't be optimal. Also the size saving is not that much. And reduces transparency while making everything more complicated. Let's pass on this.

gvanem:
Compiling with `-DCURL_CA_EMBED` and a `curl.exe  --dump-ca-embed` wirks fine. But
compiling w/o `-DCURL_CA_EMBED` and the same command prints nothing. 

Is this intentional? Why not a `"--dump-ca-embed option was disabled at build-time"` (as for `--manual`)

vszakats:
Thanks for your feedback.

My thinking is that an empty output is a clear sign that there is no embedded CA. Unlike `--manual`, I expect this output to be more often used in scripts or redirected, instead of being read by humans on the terminal. For those uses an stderr output _might_ be more of a burden than useful (though an errorlevel might help?). All in all, no strong opinion here.

This feature is brand new and I guess time/feedback/contributions will hint what would yet be useful to tweak around it.

PR #14060, opened 2024-06-29
mk-ca-bundle.pl: delay 'curl -V' execution until it is needed

sertonix:
Avoid an `Can't exec "curl"` message when curl is not actually needed.


bagder (approved):
Makes sense!

bagder:
Thanks!

PR #3619, opened 2019-02-26
Secure Transport: no more "darwinssl"

bagder:
Everyone calls it Secure Transport, now we do too.

(After previous discusssions with @nickzman and @danielgustafsson.)

PR #14065, opened 2024-06-30
cmake: fix feature and protocol lists for SecureTransport

vszakats:
NTLM was missing from the features list, and SMB/SMBS from
the protocols list in SecureTransport builds.

Follow-up to 76a9c3c4be10b3d4d379d5b23ca76806bbae536a #3619

Reported-by: Tal Regev
Bug: https://github.com/curl/curl/pull/13963#issuecomment-2178791390
Closes #14065


talregev:
Thank you for the fix!

PR #14070, opened 2024-07-01
docs: misc language polish

bagder:
- CURLINFO_FILETIME*: improve language
- add '32bit' and '64bit' as bad words, use 32-bit and 64-bit instead
- mksymbolsmanpage.pl: avoid "will"

### DIFF
--- a/.github/scripts/badwords.txt
+++ b/.github/scripts/badwords.txt
@@ -58,3 +58,9 @@ doesn't:does not
 won't:will not
 couldn't:could not
 \bwill\b:rewrite to present tense
+32bit=32-bit
+64bit=64-bit
+32 bit\b=32-bit
+64 bit\b=64-bit
+64-bits:64 bits or 64-bit
+32-bits:32 bits or 32-bit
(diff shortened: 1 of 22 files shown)


vszakats:
It might be subjective, but the hyphened version '32-bit' and '64-bit' comes out less ambiguous (more readable), at least in technical contexts, IMO. Someone correct me if this is wrong, but I remember the rule to be '32/64 bits' in plural and without the hyphen, and '32/64-bit' singular with it, to be the "canonical" (correct?) forms.

curl codebase stats:
```
$ git grep -E '(32|64)-bit' | wc -l
      80
$ git grep -E '(32|64) bit' | wc -l
      66
```

Plural version:
```
$ git grep -E '(32|64) bits' | wc -l
      11
```

Then there is the plural/hyphened combo in `lib/curl_ntlm_core.c`, shouldn't that read better as '64 bits' (or '64-bit')?:
```
/* Expand the 56-bit key to 64-bits */
```

bagder:
For a short moment I considered cleaning up the code comments to adhere to the badwords script as well, but it became just a little too much. Contractions and use of `will` everywhere.

PR #14076, opened 2024-07-01
help: Add flags to output, ssh category

yedayak:
Add --output, --remove-on-error, --output-dir and --created-dirs to the output help category
Add --hostpubmd5, --hostpubsha256, --insecure (-k) and --pubkey to the ssh help category


vszakats:
Perhaps the new categories should also be added to the corresponding `.md` manuals' `Category:` line in `docs/cmdline-opts/`?

yedayak:
> Perhaps the new categories should also be added to the corresponding `.md` manuals' `Category:` line in `docs/cmdline-opts/`?

Sure, did that and force pushed

bagder (approved):

vszakats (approved):

bagder:
Maybe `--output-dir` should go into **output** as well?

How about `--create-dirs` ?

yedayak:
> Maybe `--output-dir` should go into **output** as well?
> 
> How about `--create-dirs` ?

Agreed, added those two

yedayak:
I added two more commits that add options to the ssh category, and mention the exact command needed to generate the `tool_listhelp.c`

scripts/managen
```diff
@@ -918,7 +918,7 @@ sub listhelp {
  * This source file is generated with the following command in an autotools
  * build:
  *
- * "make listhelp"
+ * "make -C docs/cmdline-opts/ listhelp"
```
vszakats, line 921:
```suggestion
 * "make -C docs/cmdline-opts listhelp"
```
For consistency with other similar lines, e.g. in `src/tool_help.h`.

src/tool_listhelp.c
```diff
@@ -29,7 +29,7 @@
  * This source file is generated with the following command in an autotools
  * build:
  *
- * "make listhelp"
+ * "make -C docs/cmdline-opts/ listhelp"
```
vszakats, line 32: Also here.

bagder: I can generate the source fine with `make listhelp` exactly as already documented.

yedayak: > I can generate the source fine with `make listhelp` exactly as already documented.
> 
> I invoke that command in the `src/` directory.

OK, that makes sense, I didn't know `src/` had it's own targets. Will remove this commit.

bagder (approved):

bagder:
Thanks!

PR #14078, opened 2024-07-01
 cmdline-docs: fix `--proxy-ca-native` example + tidy-ups

vszakats:
Also:
- fix an indentation.
- fix capitalized option in comment.

Closes #14078

### DIFF
--- a/docs/cmdline-opts/proxy-ca-native.md
+++ b/docs/cmdline-opts/proxy-ca-native.md
@@ -12,7 +12,7 @@ See-also:
   - capath
   - insecure
 Example:
-  - --ca-native $URL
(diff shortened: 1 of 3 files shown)


docs/cmdline-opts/ca-native.md
```diff
@@ -12,7 +12,7 @@ See-also:
   - capath
   - insecure
 Example:
-  - --ca-native $URL
+  - --ca-native
```
bagder, line 15: Why remove the URL? Without it, the example is incomplete

vszakats: Yep, just realized this, it's an example. Thanks for noticing!

PR #14085, opened 2024-07-02
GHA: Windows job exclusions tweaks

vszakats:
- disable SMTP tests in MSYS2/mingw-w64 and MSVC jobs.
  On the suspicion of sometimes hanging:
  https://github.com/curl/curl/actions/runs/9346162475/job/25720437944?pr=13855#step:14:2838
  https://github.com/curl/curl/actions/runs/9758011305/job/26931678639?pr=14084#step:14:2834
  https://github.com/curl/curl/actions/runs/9774468536/job/26982805294#step:11:4731

- run TFTP, MQTT, WebSockets tests in MSYS2/msys jobs again.

- switch hanging old-mingw-w64 7.3.0 job to Release (from Debug).
  Guessing here, 9.5.0 is more solid, and one difference is
  Debug/Release mode. Let's match 7.3.0 with that and see how it changes
  hangs and flakiness.
  The other difference is Unicode ON in 7.3.0. Flaky 6.3.0 was also
  Debug, with Unicode OFF:
  217878bade884202ee5fb2e80186c5fd130392e8 #13566.
  (Unicode unlikely to play a role here IMO.)
  If 7.3.0 keeps hanging / remains flaky I'll consider disabling its
  test runs.

- opt-out from vcpkg telemetry.

Ref: https://github.com/curl/curl/pull/13599#issuecomment-2119372376
Closes #14085

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -466,7 +466,7 @@ jobs:
         shell: bash
         run: |
           git clone --quiet --depth 1 https://github.com/microsoft/vcpkg/
-          vcpkg/bootstrap-vcpkg.sh
+          vcpkg/bootstrap-vcpkg.sh -disableMetrics
 
       - name: 'vcpkg build'
         timeout-minutes: 10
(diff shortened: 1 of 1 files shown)


vszakats:
Looks like it was enough to claim just once that 9.5.0 is "solid", to make it hang!:
https://github.com/curl/curl/actions/runs/9775422731/job/26985795679

vszakats:
Well, it hung right away after SMTP and test 3008:
https://github.com/curl/curl/actions/runs/9788066896/job/27025540999?pr=14097#step:14:2878

Seems like a mission impossible.

PR #14087, opened 2024-07-03
doc: delete the redundant backslash in keepalive-cnt.md

panjf2000:


bagder:
It is not redundant. The `managen` script triggers a warning otherwise. This, because markdown viewers often get "fooled" by spurious less-than or greater-than embedded in texts.

panjf2000:
> It is not redundant. The `managen` script triggers a warning otherwise. This, because markdown viewers often get "fooled" by spurious less-than or greater-than embedded in texts.

Got it!

Issue #13700, opened 2024-05-19
GCC-14 build fails on MacOSX using vcpkg


garthgoodson:
### I did this

Compiled curl using vcpkg on Mac OSX using gcc-14 (14.1) installed from brew.
macOS: 14.3-arm64
CLT: 15.3.0.0.1.1708646388

I got the following error:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/System/Library/Frameworks/Security.framework/Headers/SecBase.h:28,
                 from /Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/System/Library/Frameworks/Security.framework/Headers/Security.h:27,
                 from /src//external/vcpkg/buildtrees/curl/src/curl-8_7_1-d8a6f70132.clean/lib/vtls/sectransp.c:57:
/Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFUserNotification.h:126:1: error: 
attributes should be specified before the declarator in a function definition
  126 | CF_INLINE CFOptionFlags CFUserNotificationCheckBoxChecked(CFIndex i) API_AVAILABLE(macos(10.0)) API_UNAVAILABLE(ios, watchos, tvos) {return ((
CFOptionFlags)(1UL << (8 + i)));}
      | ^~~~~~~~~

### I expected the following

vcpkg to compile the package

### curl/libcurl version

curl-8_7_1-d8a6f70132

### operating system

Darwin MacBook-Pro.local 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:59 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6030 arm64

vszakats:
An Apple system header is failing to compile with gcc. It's been happening since macOS SDK 13, also with gcc 13.2. The solution is to use SDK 12, or Apple clang. Or Apple making their SDK headers gcc-compatible.

garthgoodson:
My system builds with gcc 13.2.

vszakats:
The problem is detected in Apple's headers, not in curl code. It may be a new incompatibility then (and the old ones fixed?).

Can you try building without Secure Transport to see if that changes the results?

vszakats:
Found the root cause:

It's caused by a gcc patch-set applied by Homebrew to gcc, which
enabled the `availability` attribute. This breaks the build because
[...]

PR #14091, opened 2024-07-03
TODO, remove item about 'SSL_peak'

icing:
GnuTLS todo item about using an equivalent of `SSL_peak()`, which nicely escaped the word checks, is no longer relevant.

We do not use `SSL_peek()` anymore since connection filters were introduced.

### DIFF
--- a/docs/TODO
+++ b/docs/TODO
@@ -127,9 +127,6 @@
  13.16 Share the CA cache
  13.17 Add missing features to TLS backends
 
- 14. GnuTLS
- 14.2 check connection
-
  15. Schannel
  15.1 Extend support for client certificate authentication
  15.2 Extend support for the --ciphers option
@@ -926,13 +923,6 @@
  The feature matrix at https://curl.se/libcurl/c/tls-options.html shows which
  features are supported by which TLS backends, and thus also where there are
  feature gaps.
-
-14. GnuTLS
-
-14.2 check connection
-
- Add a way to check if the connection seems to be alive, to correspond to the
- SSL_peak() way we use with OpenSSL.
 
 15. Schannel
 


bagder:
Those CI failures with cmake on macOS look strange!

icing:
Someone is messing with us.

vszakats:
There was an image update, though the Xcode version remained the same, so how that broke I couldn't readily spot.
But, for sure gcc isn't something Apple desires to remain compatible with (see: #13700, #10356, https://github.com/curl/curl-for-win/commit/e2db3c475f5981352e6e6a79854a255805b28deb), so perhaps it's time to give up testing it on macOS?

icing:
After re-running the failed macOS jobs, it worked. Not sure if this just hit the old images or something otherwise as broken.

vszakats:
Found the root cause for the macOS fallout:

Turns out the Homebrew build of gcc 12.4.0 silently enabled
the `availability` attribute, over 12.3.0:

test-availability.c
```c
#if __has_attribute(availability)
static yes;
#endif
```

test.sh
```shell
$ 12.3.0/bin/gcc-12 -E test-availability.c 
# 0 "test-availability.c"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "test-availability.c"

$ 12.4.0/bin/gcc-12 -E test-availability.c 
# 0 "test-availability.c"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "test-availability.c"

static yes;
```

This breaks the build because the way the Apple SDK uses attributes (when available) are incompatible with how gcc accepts them. (possibly tracked here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108796, more info: https://github.com/llvm/llvm-project/issues/81767)

The commit implementing the `availability` macro:
https://github.com/iains/gcc-12-branch/commit/fd5530b7cb0012bf4faeddd45e13054a1dfa6783

This is part of a Darwin _compatibility_ pack that is applied to the Homebrew builds. Relevant commit:
https://github.com/Homebrew/homebrew-core/commit/b904223d9893f62bec2a8f7483bf5992747fc6c7#diff-89dd0b4176eca7fcc24b591943509bf8a8d6ea904d71e5dfcd6b78fed62fc574R44-R48

I have a patch pending for this, but wanted to get to the root cause before making a PR.

[ This was the most painful investigation probably ever that I had done, and this is just 1 of the 15 or so issues that it turned up. ]

PR #14092, opened 2024-07-03
connection filters, turn 2 methods into query

icing:
Two connection filter callback methods fit the query pattern. Remove the methods and define query constants for them.

- eliminate connection filters 'data_pending()' callback. Add query method `CF_QUERY_INPUT_PENDING`.
- eliminate connection filters 'is_alive()' callback. Add  query method `CF_QUERY_IS_ALIVE`.
- add helper method `Curl_conn_cf_input_pending()`, `Curl_conn_cf_is_alive()` to query a specific filter chain
- change filter implementations

Remove `struct connectdata *conn` parameter from `is_alive()` methods and protocol callback `conncheck()`, since `data` needs to be attached to the connection anyway.

### DIFF
--- a/lib/cf-h1-proxy.c
+++ b/lib/cf-h1-proxy.c
@@ -378,7 +378,7 @@ static CURLcode recv_CONNECT_resp(struct Curl_cfilter *cf,
   error = SELECT_OK;
   *done = FALSE;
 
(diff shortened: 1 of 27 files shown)


icing:
Aborting this PR. Will be replaced by another one.

PR #14097, opened 2024-07-04
GHA/macos: improve, fix gcc/llvm, add new test matrix

vszakats:
This PR began as an attempt to drop GCC support, after repeated reports
on fallouts when trying to use it on macOS.

Then it transformed into a 3-week project turning up the issues causing
the fallouts, ending up including llvm and all available Xcode / macOS
SDK, macOS runner image, build tools and compiler vendors and versions.
Accumulating 400 sub-commits.

I developed and tested all fixes under this PR, then merged them as
separate patches.

This PR retained CI jobs updates, extensively reworking and extending
them: [1]

At first it seemed GCC and the Apple SDK is "naturally" growing more
incompatible, as Apple added further non-standard features to their
headers. This is partly true, but reality is more complicated.

Besides some issues local to curl, there were bugs in Apple SDK
headers, Homebrew GCC builds, feature missing in the old llvm version
pre-installed on GitHub CI runner images, and subtle incompatibilities
between GCC and llvm/clang when handling language extensions.

Resulting compiler errors seldom pointed to a useful direction, and
internet search was silent about these issues too. Thus, I had to peel
them off layer by layer, using trial and error, and by recognizing
patterns of failures accross 150-200 builds combinations. Exposing
configure logs, and curl_config.h in the CI logs helped too.

1. GCC header compatibility layer ("hack" as GCC calls it)

The toughest issue is GCC's built-in compatibility layer:
  https://github.com/gcc-mirror/gcc/tree/master/fixincludes

This patch layer is further patched by a "Darwin compatibility" project
applied on top by Homebrew GCC via:
  https://github.com/iains/gcc-12-branch
  https://github.com/iains/gcc-13-branch
  https://github.com/iains/gcc-14-branch

The hack layer is designed in a way that breaks more builds than it
fixes, esp. in context of GHA runners. The idea is to build GCC
specifically for the SDK for the target macOS version. The problem with
this approach is that the Xcode + SDK installed on the local/CI machine
often does not match with the SDK used on while building GCC on
Homebrew's build machines. In these cases the GCC compatibility layer
turns into an "uncompatibility" layer and consistently breaks builds.
curl cannot offer a fix for this, because the solution (I found) is to
patch the toolchain on the local machine. I implemented this for our CI
builds and curl-for-win. In other case the user must do this patching
manually, or choose a compatible GCC + Xcode/SDK combination.

An upstream fix doesn't seem trivial either, because the issue is
ingrained in the compatibility layer's design. Offering an `-fapplesdk`
(or recognizing `-target`) option and/or fixing them within the compiler
would seem like a more robust option, and also how mainline llvm solves
this.

Here's a table summarizing the GCC + SDK combinations and curl build
failures: [2]

More info: https://github.com/curl/curl/issues/10356#issuecomment-2222734103

db135f8d7207b20d531e7e2100a49f3e16bdcfab #14119 macos: add workaround for gcc, non-c-ares, IPv6, compile error
Ref: https://github.com/curl/curl-for-win/commit/e2db3c475f5981352e6e6a79854a255805b28deb
Ref: https://github.com/curl/curl-for-win/commit/f5c58d7fef78e972be33ca2355dcb42ba56622a6

2. Homebrew GCC's `availability` extension

A recent minor Homebrew GCC upgrade caused major breakage. The "Darwin
compatibility" patch applied to GCC implemented the `availability`
compiler attribute in GCC. Apple SDK detected this and enabled using
them, but as it turns out GCC accepts compiler attributes with slightly
different rules than llvm/clang, and how the Apple SDK uses them,
breaking builds.

Affected Homebrew GCC versions are: 12.4.0, 13.3.0 and 14.1.0.

Possibly tracked here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108796
More info: https://github.com/llvm/llvm-project/issues/81767

Commit implementing the `availability` macro:
gcc-12: https://github.com/iains/gcc-12-branch/commit/fd5530b7cb0012bf4faeddd45e13054a1dfa6783
gcc-13: https://github.com/iains/gcc-13-branch/commit/cb7e4eca68cfc4763474e2eb0935a844458842a8
gcc-14: https://github.com/iains/gcc-14-branch/commit/ff62a108865a6403f5017380d7018250c1d3306f

That applied to Homebrew GCC (12.4.0):
https://github.com/Homebrew/homebrew-core/commit/b904223d9893f62bec2a8f7483bf5992747fc6c7#diff-89dd0b4176eca7fcc24b591943509bf8a8d6ea904d71e5dfcd6b78fed62fc574R44-R48

Ref: #13700
More info: https://github.com/curl/curl/pull/14091#issuecomment-2222703468

e91fcbac7d86292858718a0bfebad57978761af4 #14155 macos: undo `availability` macro enabled by Homebrew gcc

3. Proprietary Apple SDK macros

Apple SDK expects certain macros predefined by the compiler. Missing
them may causes odd issues. Mainline llvm is keeping up with Apple
clang, but it needs a fresh version, while the one installed on GitHub
runners is old (v15). I patched these in `lib/curl_setup.h`.

baa3270846b2a7307cbd0dd5c02c4e5f00e388dd #14134 build: fix llvm 16 or older + Xcode 15 or newer, and gcc

4. Apple SDK header bug

Without certain predefined macros, SDK headers can take a codepath where
it mis-defines its own `TARGET_OS_OSX` macro, which make it break its
own headers later. I patched it in `lib/curl_setup.h`.

ff784af461175584c73e7e2b65af00b1a5a6f67f #14159 build: fix llvm 17 and older + macOS SDK 14.4 and newer

5. `TargetConditionals.h` requires `sys/types.h`

Fixed in curl. It caused feature-detection failurs with autotools, and
could break builds in certain configurations.

e1f6192939c9c5ab2310b60bedf4c07d635193f6 #14130 configure: fix `SystemConfiguration` detection

6. Differences between autotools and CMake compiler options

Fixed it by syncing compiler warning options.

59cadacfcc1d39472245979cdbd614c7a9af6f0d #14128 build: sync warning options between autotools, cmake & compilers

7. Differences between autotools and CMake dependency detection

Fixed it by improving detection of libidn2, with some more fixes
pending for the next feature window.

f43adc2c4978f7f82a359e89186e58a31d17b0ad #14137 cmake: detect `libidn2` also via `pkg-config`
Ref: #14136 cmake: detect `nghttp2` via `pkg-config`, enable by default

8. libidn2 detection bug with CMake

Fixed the root cause and also the trigger in the CI config.

764fbabf6ed4c1d36c0ab2033ac8df52d9923cd7 #14175 cmake: fix builds with detected libidn2 lib but undetected header

9. Suppressed compiler warnings inside Apple-specific curl code

Fixed these warnings, which allowed to stop silencing them.

b05dc7eb3592305de9fa85640767f3dd2a8d4c93 #14122 sectransp: fix `HAVE_BUILTIN_AVAILABLE` checks to not emit warnings
5fa534b0dacdc120aaab0766047e0ecac37be4b3 #14162 sectransp: fix clang compiler warnings, stop silencing them

10. CMake mis-detecting a CA bundle path on macOS

d2ef6255f4a040397d2f40ab7cbf65a203201cd9 #14182 cmake: sync CA bundle/path detection with autotools

11. Failure to build tests with LibreSSL or wolfSSL with CMake

Fixed by dropping unnecessary includes, makign test builds dependent
on dependency headers.

3765d75ce47b66289f946382b649d0e99389dc77 #14172 cmake: fix building `unit1600` due to missing `ssl/openssl.h`

12. curl tests with CMake

curl's CMake was missing bits for running the C preprocessor accurately.
It made tests 1119 and 1167 fail. I implemented the missing bits.

efc2c5184d008fe2e5910fd03263e1ab0331d4e6 #14124 tests: include current directory when running test Perl commands
c09db8b51b88ee6ad55bd637dcb4b47678e30906 #14129 cmake: create `configurehelp.pm` like autotools does
67cc1e3400b77536a3ca529c986247e1ef985e6e #14125 test1119: adapt for `.md` input

13. GCC missing `__builtin_available()` support

curl source code assumes this is available to enable certain codepaths.
It's also intermixed with monotonic timer support.

14. Monotonic timer support with GCC

Detected by GCC, while it probably shouldn't be. llvm/clang detects it
depending on target OS version. I've been playing with this, but so far
without a conclusion or fix.

15. Runtime/test failures with GCC

I couldn't find the reason for most of this. A bunch of RTSP tests fail
with GCC. SecureTransport + HTTP/2 is failing a bunch of tests. With
OpenSSL it fails two of those. SecureTransport builds also fail one DoH
test.

16. Runtime/test failure in llvm/clang

AppleIDN support received a fix with two more remaining.

fd0250869f7837e4a48d7e6f54cc0801ad3820e8 #14179 #14176 IDN: fix ß with AppleIDN

17. Other issues found and fixed while working on this:

2c15aa5765900d4351e733671a1c8c3785beee1a        GHA/macos: delete misplaced `CFLAGS`, drop redundant CMake option
80fb7c0bef209735ab352bf4afa31193a7bc65f1 #14126 configure: limit `SystemConfiguration` test to non-c-ares, IPv6 builds
cfd6f43d6ca7e57670b422bab7bbf10221a2cf3e #14127 build: tidy up `__builtin_available` feature checks (Apple)
bae555359979016999a9425a2d489f219a78abdd #14174 runtests: show name and keywords for failed tests in summary
09cdf7e5315711dea4ce7dcf5d99a4d41e7f658b #14178 cmake: delete unused `HAVE_LIBSSH2`, `HAVE_LIBSOCKET` macros
d3595c74fab829f07ef44da1b7fc2f5668767020 #14186 configure: CA bundle/path detection fixes
58772b0e082eda333e0a5fc8fb0bc7f17a3cd99c #14187 runtests: set `SOURCE_DATE_EPOCH` to fix failing around midnight
18f1cd7a77c4beecfd977d43f55634eb593ac99e #14183 tests: sync feature names with `curl -V`
4c22d97be786ed801e050da6872dd3143d6d0a59 #14181 build: use `#error` instead of invalid syntax

Pending merges:

- #14185 runtests: fold test details for GitHub CI runs
- #14197 cmake: grab-bag of tidy-ups
- #14196 configure: limit `__builtin_available` test to Darwin

Summary:

In general GCC doesn't seem to be a good fit with curl and macOS for
now. These "lucky" combinations (GitHub Actions runner) will build out
of the box now: macos-14 + Xcode 15.0.1 + gcc-11, gcc-12, gcc-14. The
rest builds with the ugly workaround in place, but all this still leaves
some runtime issues.

More info and links in the commit messages and source code.

[1]: This PR:
- add info about target OS version requirements per feature, with OS
  names and release years.
- stop using `-Wno-deprecated-declarations` to suppress warnings.
- use `LDFLAGS=-w` to suppress 'object file was built for newer macOS
  version than being linked' warnings.
  (there were tens of thousands of them in some jobs)
- allow overriding Xcode version in all jobs.
- improve job names.
- abbreviate CMake as CM, autotools as AM for more compact job names.
- shorten job names by using `!` instead of `no-` and `non-`.
- bump parellel tests to 10 (from 5).
- drop using `--enable-maintainer-mode` `./configure` option.
- add gcc-12 no-ssl, autotools job with tests, ignore failing test
  results. (It's not yet clear why gcc-12 builds have different runtime
  results than clang/llvm ones.)
- add comments with OS names and release years next to version numbers,
  e.g. 10.15  # Catalina (2019)
- fix broken gcc-12 SecureTransport build.
- show compiler, Xcode, SDK, gcc hack SDK versions, Homebrew
  preinstalled packages and C compiler predefined macros for each job.
  Useful for debugging all the strange problems these builds might have.
- merge brew bundle and install steps.
- move step names to the top.
- dump configure log for both cmake and autotools also for successful
  builds. Useful for debugging.
- dump curl_config.h in short (sorted #defines) and full form.
- add support for the mainline llvm compiler.
- set sysroot for gcc and llvm.
- add timeout for cmake jobs.
- add new job matrix: combinations
  It supports building all possible compiler, runner image, Xcode/SDK
  combinations, with cmake and autotools, target OS versions and with or
  without SecureTransport. It's quick. GHA limits the maximum number of
  matrix jobs at 256.
  I used this as a test-rig to fix the macOS build fallouts with gcc and
  llvm.
  I settled with 16 jobs, trying to maximize fallout coverage.
- implement hack to make Homebrew gcc work with all available SDKs.
- add handy mini-table about Xcode / SDK versions, OS names, years for
  each GHA images, with the defaults.
- add tests for cmake jobs.
- make cmake config hack to link GnuTLS less intrusive.
- stop ignoring test 1452, seems fine now.
- fix to enable libpsl in autotools builds.
- enable libpsl in cmake builds.
- add an llvm job with tests (both autotools and cmake).
- delete similar macOS jobs from Circle CI. GHA is now arm64 too.

[2]: Homebrew GCC vs GHA runner images vs curl builds:
```
macOS      Xcode   gcc         gcc SDK hacks      Xcode SDK   SDK major Build Compile
           (*def)  (Homebrew)  (CommandLineTools)             versions        error
--------  -------- ----------  ------------------ ----------  --------- ----- ---------------------
macos-12   13.1    GCC 11.4.0  MacOSX12           MacOSX12.0
macos-12   13.2.1  GCC 11.4.0  MacOSX12           MacOSX12.1
macos-12   13.3.1  GCC 11.4.0  MacOSX12           MacOSX12.3
macos-12   13.4.1  GCC 11.4.0  MacOSX12           MacOSX12.3
macos-12   14.0.1  GCC 11.4.0  MacOSX12           MacOSX12.3
macos-12   14.1    GCC 11.4.0  MacOSX12           MacOSX13.0  MISMATCH  FAIL  /Applications/Xcode_14.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/os/object.h:275:1: error: expected ';' before 'extern'
macos-12  *14.2    GCC 11.4.0  MacOSX12           MacOSX13.1  MISMATCH  FAIL  /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/os/object.h:275:1: error: expected ';' before 'extern'
macos-13   14.1    GCC 11.4.0  MacOSX13           MacOSX13.0
macos-13   14.2    GCC 11.4.0  MacOSX13           MacOSX13.1
macos-13   14.3.1  GCC 11.4.0  MacOSX13           MacOSX13.3
macos-13  *15.0.1  GCC 11.4.0  MacOSX13           MacOSX14.0  MISMATCH  FAIL  /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:103:1: error: unknown type name 'dispatch_queue_t'
macos-13   15.1    GCC 11.4.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Applications/Xcode_15.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:103:1: error: unknown type name 'dispatch_queue_t'
macos-13   15.2    GCC 11.4.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:103:1: error: unknown type name 'dispatch_queue_t'
macos-14   14.3.1  GCC 11.4.0  MacOSX14           MacOSX13.3  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-14  *15.0.1  GCC 11.4.0  MacOSX14           MacOSX14.0
macos-14   15.1    GCC 11.4.0  MacOSX14           MacOSX14.2
macos-14   15.2    GCC 11.4.0  MacOSX14           MacOSX14.2
macos-14   15.3    GCC 11.4.0  MacOSX14           MacOSX14.4
macos-14   15.4    GCC 11.4.0  MacOSX14           MacOSX14.5
macos-14   16.0    GCC 11.4.0  MacOSX14           MacOSX15.0  MISMATCH  FAIL  /opt/homebrew/Cellar/gcc@11/11.4.0/lib/gcc/11/gcc/aarch64-apple-darwin23/11/include-fixed/stdio.h:83:8: error: unknown type name 'FILE'
macos-12   13.1    GCC 12.4.0  MacOSX12           MacOSX12.0
macos-12   13.2.1  GCC 12.4.0  MacOSX12           MacOSX12.1
macos-12   13.3.1  GCC 12.4.0  MacOSX12           MacOSX12.3
macos-12   13.4.1  GCC 12.4.0  MacOSX12           MacOSX12.3
macos-12   14.0.1  GCC 12.4.0  MacOSX12           MacOSX12.3
macos-12   14.1    GCC 12.4.0  MacOSX12           MacOSX13.0  MISMATCH  FAIL  /Applications/Xcode_14.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/os/object.h:275:1: error: expected ';' before 'extern'
macos-12  *14.2    GCC 12.4.0  MacOSX12           MacOSX13.1  MISMATCH  FAIL  /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/os/object.h:275:1: error: expected ';' before 'extern'
macos-13   14.1    GCC 12.4.0  MacOSX13           MacOSX13.0
macos-13   14.2    GCC 12.4.0  MacOSX13           MacOSX13.1
macos-13   14.3.1  GCC 12.4.0  MacOSX13           MacOSX13.3
macos-13  *15.0.1  GCC 12.4.0  MacOSX13           MacOSX14.0  MISMATCH  FAIL  /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:103:1: error: unknown type name 'dispatch_queue_t'
macos-13   15.1    GCC 12.4.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Applications/Xcode_15.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:103:1: error: unknown type name 'dispatch_queue_t'
macos-13   15.2    GCC 12.4.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:103:1: error: unknown type name 'dispatch_queue_t'
macos-14   14.3.1  GCC 12.4.0  MacOSX14           MacOSX13.3  MISMATCH
macos-14  *15.0.1  GCC 12.4.0  MacOSX14           MacOSX14.0
macos-14   15.1    GCC 12.4.0  MacOSX14           MacOSX14.2
macos-14   15.2    GCC 12.4.0  MacOSX14           MacOSX14.2
macos-14   15.3    GCC 12.4.0  MacOSX14           MacOSX14.4
macos-14   15.4    GCC 12.4.0  MacOSX14           MacOSX14.5
macos-14   16.0    GCC 12.4.0  MacOSX14           MacOSX15.0  MISMATCH  FAIL  /opt/homebrew/Cellar/gcc@12/12.4.0/lib/gcc/12/gcc/aarch64-apple-darwin23/12/include-fixed/stdio.h:83:8: error: unknown type name 'FILE'
macos-12   13.1    GCC 13.3.0  MacOSX12           MacOSX12.0
macos-12   13.2.1  GCC 13.3.0  MacOSX12           MacOSX12.1
macos-12   13.3.1  GCC 13.3.0  MacOSX12           MacOSX12.3
macos-12   13.4.1  GCC 13.3.0  MacOSX12           MacOSX12.3
macos-12   14.0.1  GCC 13.3.0  MacOSX12           MacOSX12.3
macos-12   14.1    GCC 13.3.0  MacOSX12           MacOSX13.0  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-12  *14.2    GCC 13.3.0  MacOSX12           MacOSX13.1  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-13   14.1    GCC 13.3.0  MacOSX13           MacOSX13.0
macos-13   14.2    GCC 13.3.0  MacOSX13           MacOSX13.1
macos-13   14.3.1  GCC 13.3.0  MacOSX13           MacOSX13.3
macos-13  *15.0.1  GCC 13.3.0  MacOSX13           MacOSX14.0  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-13   15.1    GCC 13.3.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-13   15.2    GCC 13.3.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-14   14.3.1  GCC 13.3.0  MacOSX14           MacOSX13.3  MISMATCH  FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-14  *15.0.1  GCC 13.3.0  MacOSX14           MacOSX14.0            FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-14   15.1    GCC 13.3.0  MacOSX14           MacOSX14.2            FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-14   15.2    GCC 13.3.0  MacOSX14           MacOSX14.2            FAIL  /Users/runner/work/curl/curl/bld/lib/curl_config.h:792:19: error: two or more data types in declaration specifiers
macos-14   15.3    GCC 13.3.0  MacOSX14           MacOSX14.4
macos-14   15.4    GCC 13.3.0  MacOSX14           MacOSX14.5
macos-14   16.0    GCC 13.3.0  MacOSX14           MacOSX15.0  MISMATCH  FAIL  /opt/homebrew/Cellar/gcc@13/13.3.0/lib/gcc/13/gcc/aarch64-apple-darwin23/13/include-fixed/stdio.h:83:8: error: unknown type name 'FILE'
macos-12   13.1    GCC 14.1.0  MacOSX12           MacOSX12.0
macos-12   13.2.1  GCC 14.1.0  MacOSX12           MacOSX12.1
macos-12   13.3.1  GCC 14.1.0  MacOSX12           MacOSX12.3
macos-12   13.4.1  GCC 14.1.0  MacOSX12           MacOSX12.3
macos-12   14.0.1  GCC 14.1.0  MacOSX12           MacOSX12.3
macos-12   14.1    GCC 14.1.0  MacOSX12           MacOSX13.0  MISMATCH  FAIL  /Applications/Xcode_14.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/os/object.h:275:1: error: expected ';' before 'extern'
macos-12  *14.2    GCC 14.1.0  MacOSX12           MacOSX13.1  MISMATCH  FAIL  /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/os/object.h:275:1: error: expected ';' before 'extern'
macos-13   14.1    GCC 14.1.0  MacOSX13           MacOSX13.0
macos-13   14.2    GCC 14.1.0  MacOSX13           MacOSX13.1
macos-13   14.3.1  GCC 14.1.0  MacOSX13           MacOSX13.3
macos-13  *15.0.1  GCC 14.1.0  MacOSX13           MacOSX14.0  MISMATCH  FAIL  /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:70:1: error: type defaults to 'int' in declaration of 'DISPATCH_DECL_FACTORY_CLASS_SWIFT' [-Wimplicit-int]
macos-13   15.1    GCC 14.1.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Applications/Xcode_15.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:70:1: error: type defaults to 'int' in declaration of 'DISPATCH_DECL_FACTORY_CLASS_SWIFT' [-Wimplicit-int]
macos-13   15.2    GCC 14.1.0  MacOSX13           MacOSX14.2  MISMATCH  FAIL  /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dispatch/queue.h:70:1: error: type defaults to 'int' in declaration of 'DISPATCH_DECL_FACTORY_CLASS_SWIFT' [-Wimplicit-int]
macos-14   14.3.1  GCC 14.1.0  MacOSX14           MacOSX13.3  MISMATCH
macos-14  *15.0.1  GCC 14.1.0  MacOSX14           MacOSX14.0
macos-14   15.1    GCC 14.1.0  MacOSX14           MacOSX14.2
macos-14   15.2    GCC 14.1.0  MacOSX14           MacOSX14.2
macos-14   15.3    GCC 14.1.0  MacOSX14           MacOSX14.4
macos-14   15.4    GCC 14.1.0  MacOSX14           MacOSX14.5
macos-14   16.0    GCC 14.1.0  MacOSX14           MacOSX15.0  MISMATCH  FAIL  /opt/homebrew/Cellar/gcc/14.1.0_1/lib/gcc/current/gcc/aarch64-apple-darwin23/14/include-fixed/stdio.h:83:8: error: unknown type name 'FILE'
```
Source: https://github.com/curl/curl/actions/runs/9883956647/job/27299564218

Closes #14097


bagder (approved):
I have to agree, keeping them around is now only causing us grief

vszakats:
What's strange is that I cannot find any meaningful reference about this fallout on the internet, though something like this should hit elsewhere too.

On the other hand I also cannot spot anything (obvious) in curl that could trigger such error. In theory an innocent typedef or keyword redefine could do this.

Also `lib/macos.c` is doing one single SDK call, and I wonder if there is any other way to implement it:
31f631a142d855f069242f3e0c643beec25d1b51 #7121
1a5e41d97c8f9c8ca3c5f3ec11f4f3865533098a #7128 #7129

In particular, #7128 also reported a GCC fallout, and that one was fixed by replacing one SDK header with another.
That report calls out `lib/macos.c` as fixing "a obscure, rare edge case".

vszakats:
Also noting that such innocent typedef / keyword-redefine may also occure in an dependency header.

Like for example in this wolfSSL case with `byte`: d68a121266671c806b5065c2fdce52d292bf7830

Though the fallout is unrelated to the TLS backend:
https://github.com/curl/curl/actions/runs/9791701214

As for the runner images:
working: https://github.com/actions/runner-images/blob/macos-14-arm64/20240625.2/images/macos/macos-14-arm64-Readme.md
failing: https://github.com/actions/runner-images/blob/macos-14-arm64/20240701.9/images/macos/macos-14-arm64-Readme.md

Notably Xcode 15 is introduced and GCC 12.4.0 (over 12.3.0):
```diff
--- macos-14-arm64-Readme-0625.md
+++ macos-14-arm64-Readme-0701.md
@@ -6,7 +6,7 @@
 # macOS 14
 - OS Version: macOS 14.5 (23F79)
 - Kernel Version: Darwin 23.5.0
-- Image Version: 20240625.2
+- Image Version: 20240701.9
 
 ## Installed Software
 
@@ -16,11 +16,11 @@
 - Clang/LLVM 14.0.3
 - Clang/LLVM (Homebrew) 15.0.7 - available on `$(brew --prefix llvm@15)/bin/clang`
 - GCC 11 (Homebrew GCC 11.4.0) - available by `gcc-11` alias
-- GCC 12 (Homebrew GCC 12.3.0) - available by `gcc-12` alias
+- GCC 12 (Homebrew GCC 12.4.0) - available by `gcc-12` alias
 - GCC 13 (Homebrew GCC 13.3.0) - available by `gcc-13` alias
 - GCC 14 (Homebrew GCC 14.1.0_1) - available by `gcc-14` alias
 - GNU Fortran 11 (Homebrew GCC 11.4.0) - available by `gfortran-11` alias
-- GNU Fortran 12 (Homebrew GCC 12.3.0) - available by `gfortran-12` alias
+- GNU Fortran 12 (Homebrew GCC 12.4.0) - available by `gfortran-12` alias
 - GNU Fortran 13 (Homebrew GCC 13.3.0) - available by `gfortran-13` alias
 - GNU Fortran 14 (Homebrew GCC 14.1.0_1) - available by `gfortran-14` alias
 - Kotlin 2.0.0-release-341
@@ -34,7 +34,7 @@
 - Bundler 2.5.14
 - Carthage 0.39.1
 - CocoaPods 1.15.2
-- Homebrew 4.3.7
+- Homebrew 4.3.8
 - NPM 10.7.0
 - NuGet 6.3.1.1
 - Pip3 24.0 (python 3.12)
@@ -51,7 +51,7 @@
 - 7-Zip 17.05
 - aria2 1.37.0
 - azcopy 10.25.1
-- bazel 7.2.0
+- bazel 7.2.1
 - bazelisk 1.20.0
 - bsdtar 3.5.3 - available by 'tar' alias
 - Curl 8.6.0
@@ -69,19 +69,19 @@
 - zstd 1.5.6
 
 ### Tools
-- AWS CLI 2.17.1
-- AWS SAM CLI 1.119.0
+- AWS CLI 2.17.5
+- AWS SAM CLI 1.120.0
 - AWS Session Manager CLI 1.2.633.0
 - Azure CLI 2.61.0
 - Azure CLI (azure-devops) 1.0.1
 - Bicep CLI 0.28.1
 - Cmake 3.29.6
-- CodeQL Action Bundle 2.17.5
+- CodeQL Action Bundle 2.17.6
 - Fastlane 2.221.1
 - SwiftFormat 0.54.0
 - Xcbeautify 2.4.0
 - Xcode Command Line Tools 15.3.0.0.1.1708646388
-- Xcodes 1.4.1
+- Xcodes 1.5.0
 
 ### Linters
 
@@ -139,13 +139,13 @@
 
 #### PowerShell Modules
 - Az: 12.0.0
-- Pester: 5.6.0
+- Pester: 5.6.1
 - PSScriptAnalyzer: 1.22.0
 
 ### Xcode
 | Version          | Build    | Path                           |
 | ---------------- | -------- | ------------------------------ |
-| 16.0 (beta)      | 16A5171c | /Applications/Xcode_16.0.app   |
+| 16.0 (beta)      | 16A5171r | /Applications/Xcode_16.0.app   |
 | 15.4             | 15F31d   | /Applications/Xcode_15.4.app   |
 | 15.3             | 15E204a  | /Applications/Xcode_15.3.app   |
 | 15.2             | 15C500b  | /Applications/Xcode_15.2.app   |
@@ -198,17 +198,17 @@
 | Simulator - watchOS 10.4                                | watchsimulator10.4                            | 15.3          |
 | Simulator - watchOS 10.5                                | watchsimulator10.5                            | 15.4          |
 | Simulator - watchOS 11.0                                | watchsimulator11.0                            | 16.0          |
-| visionOS 1.0                                            | xros1.0                                       | 15.2          |
 | Simulator - visionOS 1.0                                | xrsimulator1.0                                | 15.2          |
-| Simulator - visionOS 1.1                                | xrsimulator1.1                                | 15.3          |
+| visionOS 1.0                                            | xros1.0                                       | 15.2          |
 | visionOS 1.1                                            | xros1.1                                       | 15.3          |
+| Simulator - visionOS 1.1                                | xrsimulator1.1                                | 15.3          |
 | Simulator - visionOS 1.2                                | xrsimulator1.2                                | 15.4          |
 | visionOS 1.2                                            | xros1.2                                       | 15.4          |
 | visionOS 2.0                                            | xros2.0                                       | 16.0          |
 | Simulator - visionOS 2.0                                | xrsimulator2.0                                | 16.0          |
 | Asset Runtime SDK for macOS hosts targeting watchOS 9.4 | assetruntime.host.macosx.target.watchos9.4    | 14.3.1        |
-| Asset Runtime SDK for macOS hosts targeting iOS 16.4    | assetruntime.host.macosx.target.iphoneos16.4  | 14.3.1        |
 | Asset Runtime SDK for macOS hosts targeting tvOS 16.4   | assetruntime.host.macosx.target.appletvos16.4 | 14.3.1        |
+| Asset Runtime SDK for macOS hosts targeting iOS 16.4    | assetruntime.host.macosx.target.iphoneos16.4  | 14.3.1        |
 | DriverKit 22.4                                          | driverkit22.4                                 | 14.3.1        |
 | DriverKit 23.0                                          | driverkit23.0                                 | 15.0.1        |
 | DriverKit 23.2                                          | driverkit23.2                                 | 15.1, 15.2    |
```

vszakats:
Another interesting pattern is that it only affects CMake builds!

autotools jobs with the new image compile without issues:
https://github.com/curl/curl/actions/runs/9791701214/job/27036027532#step:10:254

vszakats:
It's not Unity-related:
https://github.com/curl/curl/actions/runs/9797232157/job/27053429493?pr=14097#step:7:219

vszakats:
Also not cmake related, it's just that we don't have an autotools with gcc. (Despite calling `gcc`, but that's an alias for `clang`.)

vszakats:
Disabling `CURL_MACOS_CALL_COPYPROXIES` for gcc causes these:
```
TESTFAIL: These test cases failed: 567 568 569 570 571 572 577 689 1156 1539 3100
```
Ref: https://github.com/curl/curl/actions/runs/9797774554/job/27055046618?pr=14097#step:13:3647

vszakats:
The best fix I could figure out is this ugly hack:
```diff
--- a/lib/macos.c
+++ b/lib/macos.c
@@ -30,7 +30,14 @@
 
 #include "macos.h"
 
+#if defined(__clang__)
 #include <SystemConfiguration/SCDynamicStoreCopySpecific.h>
+#else
+#include <CoreFoundation/CFDictionary.h>
+typedef const struct CF_BRIDGED_TYPE(id) __SCDynamicStore * SCDynamicStoreRef;
+CFDictionaryRef __nullable
+SCDynamicStoreCopyProxies(SCDynamicStoreRef __nullable store);
+#endif
 
 CURLcode Curl_macos_init(void)
 {
```

It works by avoiding the system headers that are incompatible at
various points depending on gcc and SDK version.

Another option is to find an alternative hack in place of what is
currently used in `lib/macos.c`.

Last but not least c-ares builds seem to be immune to this, because
they don't use the hack in `lib/macos.c`. Also non-IPv6 builds, that
doesn't seem much useful.

SecureTransport builds also require system headers and these 
remain broken.

Also it turns out that a bunch of tests fail in macOS + gcc builds.
I haven't looked into it:
RTSP(all), 1156, 1539

lib/macos.c
```diff
@@ -30,7 +30,14 @@
...
 #include "macos.h"
 
+#if defined(__clang__)
 #include <SystemConfiguration/SCDynamicStoreCopySpecific.h>
+#else
+#include <CoreFoundation/CFDictionary.h>
+typedef const struct CF_BRIDGED_TYPE(id) __SCDynamicStore * SCDynamicStoreRef;
+CFDictionaryRef __nullable
+SCDynamicStoreCopyProxies(SCDynamicStoreRef __nullable store);
+#endif
```
bagder, line 40: If you decide to go with this, I think a clarifying comment about why this is done could be valuable.

vszakats:
Here's a CI run with gcc 11, 12, 13, 14:
https://github.com/curl/curl/actions/runs/9800272567/job/27061944370?pr=14097

- 11.4.0 works
- 12.4.0 fails at `CF_INLINE`
- 13.3.0 fails at `__API_AVAILABLE2` and thousands of other places
- 141.0 fails at `CF_INLINE`

Is there anyone who can make some sense of this?

vszakats:
An even more bizarre run, with gcc 11, 12, 13, 14 on macos-12, 13, 14(latest) runners with both OpenSSL and SecureTransport.

- Without the `lib/macos.c` patch: https://github.com/curl/curl/actions/runs/9800720682/job/27062989452
  - gcc 11 + macos-14 compatible, the rest not.

- With the patch: https://github.com/curl/curl/actions/runs/9800770005/job/27063103792
  - gcc 11 + macos-14 compatible.
  - gcc 13 is incompatible all around.
  - all incompatible with SecureTransport (except gcc + mac-14).

There is a separate issue with gcc 11 + macos-14 + SecureTransport and _autotools_, where `SystemConfiguration` framework is not linked, and tests fail to link due to unresolved `SCDynamicStoreCopyProxies` symbol. This is caused by broken detection in `CURL_DARWIN_SYSTEMCONFIGURATION`: https://github.com/curl/curl/actions/runs/9800981762/job/27063595813#step:9:6351

TL;DR:
1. gcc and Xcode don't mix well in recent years.
2. the proxy hack added to `lib/macos.c` is a source of many problems.

vszakats:
So, while this patch for the hack restores some gcc compatibility, the compatibility situation between GCC and macOS SDK is fuzzy enough to make such builds fragile and overall impractical.

1. SecureTransport builds are affected. It is easy to fix by not using this backend. It's on its way out anyway.
2. For other builds these are the options:
  - Drop the hack `lib/macos.c` calling into `SCDynamicStoreCopyProxies` altogether.
    Is there a replcement for it? Will there be other hacks in the future when curl needs to call into an Apple framework?
    Can this call perhaps be made dynamically?
  - Drop the hack for GCC builds.
  - Apply the ugly patch, which fixes it for now with no loss of functionality.
  - Drop GCC support for macOS builds.

What do we prefer?

vszakats:
I got stuck with these failures, both checks are using the C preprocessor:
```
TESTFAIL: These test cases failed: 1119 1167
```
It happens when building curl _out-of-tree_. Possibly only on macOS builds. (We have out-of-tree Linux jobs in CI, which work as expected).

https://github.com/curl/curl/actions/runs/9824255601/job/27123041962#step:17:2380
```
test 1119...[Verify that symbols-in-versions and headers are in sync]

 1119: stdout FAILED:
--- log/3/check-expected	2024-07-07 03:29:22
+++ log/3/check-generated	2024-07-07 03:29:22
@@ -1 +0,0 @@
-OK[LF]
== Contents of files in the log/3/ dir after test 1119
=== Start of file check-expected
 OK[LF]
=== End of file check-expected
=== Start of file server.cmd
 Testnum 1119
=== End of file server.cmd
=== Start of file stderr1119
 In file included from ../../tests/../include/curl/stdcheaders.h:27:
 In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/types.h:187:
 In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/_types/_fd_def.h:32:
 In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/Availability.h:168:
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/AvailabilityInternal.h:38:30: error: missing '(' after '__has_builtin'
     #if defined(__has_builtin) && __has_builtin(__is_target_os)
                              ^
[...]
```

This replicates locally too.

vszakats:
Re 1119 1167, the problem is that the Perl invocation isn't including the current directory where the generated `configurehelp.pm` is. This isn't so much of a problem, unless there is a non-standard/default preprocessor, like on macOS.

vszakats:
Still chasing these sub-issues:
- whether our use of underscore-less `MAC_OS_X_VERSION_MAX_ALLOWED`, `MAC_OS_X_VERSION_MIN_REQUIRED` are indeed correct and working in all combinations.
- if my `-Wpointer-bool-conversion` warning fix is actually correct.
- after fixing a list of fallouts, I'm back to the original one related to `CF_INLINE`. This seems to be the remaining blocking issues across the combinations. Ref: https://github.com/curl/curl/actions/runs/9860040784/job/27225310422

Some hacks necessary to unbreak gcc are blackbelt ugly ones few people would consider practical unless absolutely wanting to use a non-Apple compiler. It had to be done to peel away the layers of breakage though.

vszakats:
I consider this done. Commit message update to the write-up of the whole story.

macOS GCC builds fail some tests, in particular HTTP/2, esp with SecureTransport.
One DoH test and ~10 RTSP tests. I wonder if those are caused by curl crashes.
For now the problematic tests results are ignored.

vszakats:
Managed to mess up the merge, the commit message got cut off by `git am`.
Also losing the `Closing` line. Very frustrating. Sorry about this.

Merged in 1e75edd372868048c9f805ac4ca6d2cb5a88ff5a.

vszakats:
@bagder Shall I revert and re-commit with an amended commit message?

vszakats:
OK, did a revert and a re-issue.

Issue #13873, opened 2024-06-04
`--ciphers` may not properly set `--tls13-ciphers`


MasterInQuestion:
The current behavior (["--ciphers"](https://curl.se/docs/manpage.html#--ciphers), "--tls13-ciphers" each independent) may feel tricky.
    Would it be more appropriate to readapt the design?
    (or maybe better explain the rationale..?)

    E.g.
    `--tls13-ciphers "TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256"`
    `--ciphers "TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256"`
    .
    `--tls13-ciphers "TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256" --ciphers "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"`
    `--ciphers "TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256:TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"`

    See also:
    "--ciphers" may include other, unspecified ciphers in TLS handshake
    [ https://github.com/curl/curl/issues/7198](https://github.com/curl/curl/issues/7198)
    .
    Clarify setting TLS 1.3 ciphers using different backends
    [ https://github.com/curl/curl/issues/3938](https://github.com/curl/curl/issues/3938)
    [ https://github.com/curl/curl/issues/3077#issuecomment-426561378](https://github.com/curl/curl/issues/3077#issuecomment-426561378)
    [ https://github.com/curl/curl/issues/2435](https://github.com/curl/curl/issues/2435)

    Test site: https://browserleaks.com/tls

MasterInQuestion:
What's the expectation for the "--ciphers" alike..?
    [ https://github.com/curl/curl/issues/1316#issuecomment-285779700](https://github.com/curl/curl/issues/1316#issuecomment-285779700)

    Mere pass-through?
    https://www.openssl.org/docs/manmaster/man1/openssl-ciphers.html#CIPHER-LIST-FORMAT

jan2000:
The situation with the cipher options is not as straightforward as one might expect, because handling of these options differ for each SSL backends that libcurl is build against. For some backends (e.g. mbedtls) the `--cipher` option is parsed by libcurl itself, for others (e.g. openssl, wolfssl) it is passed through directly to the SSL backend.

[...]

PR #14106, opened 2024-07-05
openssl 1.3 ciphers

icing:
Allow TLS 1.3 ciphers in CURLOPT_SSL_CIPHER_LIST and --ciphers for OpenSSL iff:

- no CURLOPT_TLS13_CIPHERS or --tls13-ciphers is used
- the minimum and maximum TLS versions are either default or at least 1.3

When 1.3 ciphers are set this way, the minimum TLS version is raised to 1.3, e.g. a 1.2 handshake will not longer work.

refs #13873

### DIFF
--- a/docs/CIPHERS.md
+++ b/docs/CIPHERS.md
@@ -22,12 +22,12 @@ suites by using the respective regular cipher option.
 
 With curl 8.10 you may specify TLS 1.3 ciphers also via `CURLOPT_SSL_CIPHER_LIST`
 and `--ciphers` if they are *only* 1.3 ones. Mixing 1.2 and 1.3 ciphers in
-the same argument does not work. 
+the same argument does not work.
 
 Note that when 1.3 ciphers are configured
 using these options, they implicitly raise the minimum TLS version to 1.3.
 It the minimum TLS version is configured to be lower, setting 1.3 ciphers
-with these options will fail.
+with these options fails.
 
 The names of the known ciphers differ depending on which TLS backend that
 libcurl was built to use. This is an attempt to list known cipher names.
--- a/docs/CIPHERS.md
+++ b/docs/CIPHERS.md
@@ -20,6 +20,15 @@ and
 . If you are using a different SSL backend you can try setting TLS 1.3 cipher
 suites by using the respective regular cipher option.
 
+With curl 8.10 you may specify TLS 1.3 ciphers also via `CURLOPT_SSL_CIPHER_LIST`
+and `--ciphers` if they are *only* 1.3 ones. Mixing 1.2 and 1.3 ciphers in
+the same argument does not work. 
+
+Note that when 1.3 ciphers are configured
+using these options, they implicitly raise the minimum TLS version to 1.3.
+It the minimum TLS version is configured to be lower, setting 1.3 ciphers
+with these options will fail.
+
 The names of the known ciphers differ depending on which TLS backend that
 libcurl was built to use. This is an attempt to list known cipher names.
 
--- a/docs/cmdline-opts/tls13-ciphers.md
+++ b/docs/cmdline-opts/tls13-ciphers.md
@@ -27,3 +27,11 @@ https://curl.se/docs/ssl-ciphers.html
 This option is currently used only when curl is built to use OpenSSL 1.1.1 or
 later, or Schannel. If you are using a different SSL backend you can try
 setting TLS 1.3 cipher suites by using the --ciphers option.
+
+With curl 8.10 or newer, TLS 1.3 ciphers for OpenSSL may be configured with
+`--ciphers` also under the following conditions:
+- only 1.3 ciphers are mentioned
+- the minimum and maximum TLS versions are unspecified or at least 1.3
+
+Setting 1.3 ciphers via `--ciphers` implies that the minimum TLS version
+to negotiate is at least 1.3.
--- a/lib/vquic/curl_quiche.c
+++ b/lib/vquic/curl_quiche.c
@@ -1120,7 +1120,7 @@ out:
     nwritten = -1;
   }
   CURL_TRC_CF(data, cf, "[%" CURL_PRIu64 "] cf_send(len=%zu) -> %zd, %d",
-              stream? stream->id : (uint64_t)~0, len, nwritten, *err);
+              stream? stream->id : (curl_uint64_t)~0, len, nwritten, *err);
   return nwritten;
 }
 
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -2689,9 +2689,6 @@ static CURLcode
 ossl_set_ssl_version_min_max(struct Curl_cfilter *cf, SSL_CTX *ctx)
 {
   struct ssl_primary_config *conn_config = Curl_ssl_cf_get_primary_config(cf);
-  /* first, TLS min version... */
-  long curl_ssl_version_min = conn_config->version;
-  long curl_ssl_version_max;
 
   /* convert curl min SSL version option to OpenSSL constant */
 #if (defined(OPENSSL_IS_BORINGSSL)  || \
@@ -2703,7 +2700,9 @@ ossl_set_ssl_version_min_max(struct Curl_cfilter *cf, SSL_CTX *ctx)
   long ossl_ssl_version_min = 0;
   long ossl_ssl_version_max = 0;
 #endif
-  switch(curl_ssl_version_min) {
+  switch(conn_config->version) {
+  case CURL_SSLVERSION_DEFAULT: /* no restriction, let lib use default */
+    break;
   case CURL_SSLVERSION_TLSv1: /* TLS 1.x */
   case CURL_SSLVERSION_TLSv1_0:
     ossl_ssl_version_min = TLS1_VERSION;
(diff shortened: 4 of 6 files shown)


bagder (approved):

jan2000:
I see this already approved... could you hold of merging this for a little while? As I have some reservations/thoughts about this, but I do not have the time to properly outline them right now.

bagder:
It is also labeled **feature-window** so it will not be merged until earliest when the feature window opens again: August 3, 2024.

jan2000:
While I do agree that it is unfortunate that there exists a `--tls13-ciphers` option, I see some problems with trying to incorporate its function into the `--ciphers` option in an effort to abandon it.

With this changeset one can set either tls1.3 ciphers or tls1.2 ciphers with the `--ciphers` option. From this it will be logical for one to assume that both tls1.3 and tls1.2 ciphers can be set set with the `--ciphers` option. However, when setting any tls1.2 ciphers with this option the default tls1.3 ciphers will be used instead, and any specified tls1.3 ciphers are silently ignored. Even more counter-intuitive is that the right way to set the newer tls1.3 ciphers in this case is to use the now "abandonded" `--tls13-ciphers` option.

Also, this changeset only works for OpenSSL/LibreSSL, because it relies on the behaviour of its cipher string parsing. Making it harder to offer an as consistent as possible interface across the various SSL backends:
- Schannel: Setting the tls1.3 ciphers with the `--ciphers` option is not possible, thus still requiring the `--tls13-ciphers` option.
- BoringSSL: This will result in an error that the ciphers are not recognized.
- GnuTLS and rustls: The option will not be taken into account at all, thus unexpectedly allowing lower than tls1.3 connections.
- WolfSSL and mbedTLS: It will work as expected if only tls1.3 ciphers are specified through the  `--cipher` option, but here lies another consistency problem. When specifying tls1.2 ciphers through the `--cipher` option all other ciphers are disabled, thus also the default 1.3 ciphers. This is in contrast to OpenSSL where the default tls1.3 ciphers are left untouched. With this changeset this discrepant behaviour can now no longer be easily fixed.

Now, in order to provide an as consistent as possible interface across all SSL backends, and because we can not fully get rid of the `--tls13-ciphers` option (without BC breaking changes), I would suggest an alternative.

Instead of abandoning the `--tls13-ciphers` option and relying on the cipher string parsing of OpenSSL, I would suggest to fully embrace the `--tls13-ciphers` option and set the tls_min version based on just the presence of this option. Leaving the `--ciphers` option for tls1.2 only, so that it can be fully deprecated maybe some day.

So as an alternative I would suggest the following:
- All SSL backends: If `--tls13-ciphers` is set, `--ciphers` is not set, tls_min is not set and tls_max is not lower than 1.3 (not set or 1.3), then set tls_min to 1.3.
- wolfSSL: FIX tls_min / tls_max
- wolfSSL and mbedTLS: Set the ciphers to a combination of `[none if tls_max < 1.3, else --tls13-ciphers or default13_ciphers] and [none if tls_min > 1.2, else --ciphers or default12_ciphers]`.

And for clarity:
- Where setting tls1.3 ciphers is not supported: WARN if `--tls13-ciphers` option is set.
- Where setting tls1.2 ciphers is not supported: WARN if `--ciphers` option is set.
- All: WARN if `--tls13-ciphers` is set and tls_max is lower than 1.3.
- All: WARN if `--tls13-ciphers` is set, `--ciphers` is not set and tls_min is set to lower than 1.3.
- All: WARN if `--ciphers` is set and tls_min is 1.3 or higher.

These changes would result in the following:
```
$ curl --tls13-ciphers TLS_AES_128_GCM_SHA256 $url
            status  TLSv1.3 ciphers         TLSv1.2 ciphers
openssl:            TLS_AES_128_GCM_SHA256  -
boringssl:  WARN    default13               -
schannel:           TLS_AES_128_GCM_SHA256  -
mbed/wolf:          TLS_AES_128_GCM_SHA256  -
gnutls:     WARN    default13               -
sectr/bear: ERROR   -                       -

$ curl --ciphers ECDHE-RSA-AES128-GCM-SHA256 $url
            status  TLSv1.3 ciphers         TLSv1.2 ciphers
openssl:            default13               ECDHE-RSA-AES128-GCM-SHA256
boringssl:          default13               ECDHE-RSA-AES128-GCM-SHA256
schannel:   ERROR   -                       -
mbed/wolf:          default13               ECDHE-RSA-AES128-GCM-SHA256
gnutls:     WARN    default13               default12
sectr/bear:         -                       ECDHE-RSA-AES128-GCM-SHA256

$ curl --tls13-ciphers TLS_AES_128_GCM_SHA256 --ciphers ECDHE-RSA-AES128-GCM-SHA256 $url
            status  TLSv1.3 ciphers         TLSv1.2 ciphers
openssl:            TLS_AES_128_GCM_SHA256  ECDHE-RSA-AES128-GCM-SHA256
boringssl:  WARN    default13               ECDHE-RSA-AES128-GCM-SHA256
schannel:   ERROR   -                       -
mbed/wolf:          TLS_AES_128_GCM_SHA256  ECDHE-RSA-AES128-GCM-SHA256
gnutls:     WARN    default13               default12
sectr/bear: WARN    -                       ECDHE-RSA-AES128-GCM-SHA256
```

MasterInQuestion:
Whatsoever go ahead..? Couldn't be worse seemingly.

jan2000:
> I am not proficient enough with Schannel to implement the analog behaviour as this PR does for OpenSSL. If someone knows how to do it, please go ahead.

> BoringSSL does not allow the settings of any tls1.3 ciphers, AFAIK. It seems to be destined to always be an outlier in this.

> Again, GnuTLS does not seem to allow settings of ciphers currently. The PR does not change that.

Indeed, but the thing is that this PR introduces setting tls1.3 ciphers and implicitly raising tls_min to 1.3 based on the `--chipers` option that is not implementable for these backends, increasing inconsistency between the SSL backends.

With my suggestion no specific changes are needed for these backends. And while BoringSSL and GnuTLS do not allow for setting tls1.3 ciphers, the implicit raising of tls_min to 1.3 will at least be achieved.

> > * All SSL backends: If `--tls13-ciphers` is set, `--ciphers` is not set, tls_min is not set and tls_max is not lower than 1.3 (not set or 1.3), then set tls_min to 1.3.
> 
> Isn't this what the PR effectively does for OpenSSL?

No. The idea you introduced with this PR of implicitly raising tls_min to 1.3 is good. My suggestion is a variation on your idea, but with crucial differences.

With this PR raising of tls_min is dependent on the parsing of the contents of of the `--cipher` option. My suggestion is to make it instead dependent on the `--tls13-ciphers` option. And just on its presence, not its contents. This way it can be implemented for all SSL backends, regardless whether the backend can parse the option or not.

Also with my suggestion `--ciphers` is kept solely for tls1.2 (1.1, 1.0) ciphers and `--tls13-ciphers` solely for tls1.3 ciphers. This is easier to document and clearer for the end-user. The intent that one should not use `--ciphers` to set tls1.3 ciphers is crucial in order to fix the discrepant behaviour of WolfSSL and mbedTLS.

> I am all for aligning different backends as much as possible. This PR is about OpenSSL behaviour and you just extended its scope, then saying it does not honor it.
>
> My question: is anything in this PR preventing or opposing the vision you have for cipher treatment against all backends?

Yes, the problem with this PR is that it introduces behaviour for OpenSSL only that can not be implemented on the other backends.

With my suggestion the sitation will be straightforward:
- To set tls1.2 (1.1, 1.0) ciphers one uses the `--ciphers` option.
- To set tls1.3 ciphers one uses the `--tls13-ciphers` option.
- If `--tls13-ciphers` is set and `--ciphers` is not, then `--tlsv1.3` will be implicitly set (regardless whether the the SSL backend supports `--ciphers` and/or `--tls13-ciphers`).

This PR conflicts with my suggestion as:
- It introduces setting tls1.3 ciphers through the `--ciphers` option
- The condition for implicitly setting `--tlsv1.3` is different and only implementable for OpenSSL.

MasterInQuestion:
Your suggestion is pretty much no different from what currently observed in #13873...
    Plus an unreasonable bad assertion on TLS 1.3 upgrade:
    Specifying TLS 1.3 ciphers may not imply wish to insist TLS 1.3+.

icing:
> > > * All SSL backends: If `--tls13-ciphers` is set, `--ciphers` is not set, tls_min is not set and tls_max is not lower than 1.3 (not set or 1.3), then set tls_min to 1.3.

This is a change in current behaviour and would break scripts using `curl --tls13-ciphers` that currently *can* negotiate a v1.2 handshake. As I described in https://github.com/curl/curl/issues/13873#issuecomment-2210432007, this is what I wanted to avoid.

Why do you think this is an acceptable change? Maybe I missed something?

MasterInQuestion:
Probably just bewildered by the mess...

    Eventually I think the only appropriate solution would be interpreting the names and polyfill.
    [ Mentioned in: [ https://github.com/curl/curl/issues/13873#issuecomment-2154249293](https://github.com/curl/curl/issues/13873#issuecomment-2154249293) ]

icing:
Rereading everything again, I don't know if we are getting anywhere here with tweaking `--ciphers` in order to have a common handling for all (or at least most) TLS backends. We need another approach.

I agree with @jan2000 's proposal to issue a warning if options are set that cannot be honoured or are conflicting. I am now convinced that we should not mess with `tls_min` and `tls_max` by cipher options alone. We have the other tls options for that.

Having a standardised way across all TLS backends seems near impossible, as the TLS libraries vary quite a lot. Also we do *not* want to prohibit support for new ciphers because curl does not recognise the name, but the TLS lib would.

Therefore I withdraw this PR.

MasterInQuestion:
Permissive parsing: that not recognized, arranged best-effort.

    Not necessarily all-the-way standardized, mere polyfill: to make things less surprising.

PR #14107, opened 2024-07-05
Make use of REUSE 3.2 and REUSE.toml

mxmehl:
REUSE received a significant update of its specification. This version allows for a new way to communication licensing and copyright of files that cannot be annotated directly: [REUSE.toml](https://reuse.software/spec-3.2/#reusetoml).

This PR migrates the deprecated way of doing so ([dep5](https://reuse.software/faq/#dep5-history)) to this improved way. It would also allow for adding multiple REUSE.toml files, e.g. one separate in `/docs/` and `/tests/`, but I figured this "central" approach might have it advantages for the curl project.

Two FAQ items on REUSE.toml if you're interested, one about [bulk licensing](https://reuse.software/faq/#bulk-license), the other about [additional licensing info](https://reuse.software/faq/#aggregate-info).

Here is the [full changelog of REUSE 3.2](https://github.com/fsfe/reuse-website/blob/main/CHANGELOG.md#32---2024-07-03).


mxmehl:
I noticed that `scripts/copyright.pl` also makes use of the dep5 file.

I wonder whether the file is actually still needed. If you wanted to learn whether you're compliant, `reuse lint` is the way to go. If you need to know which file has which license/copyright, `reuse lint -j` may help.

REUSE.toml
```diff
@@ -0,0 +1,126 @@
+# SPDX-License-Identifier: curl
+# SPDX-FileCopyrightText: Daniel Stenberg, <daniel@haxx.se>, et al.
+
+version = 1
+SPDX-PackageName = "curl"
+SPDX-PackageSupplier = "Daniel Stenberg <daniel@haxx.se>"
+SPDX-PackageDownloadLocation = "https://curl.se"
```
vszakats, line 7:
```suggestion
SPDX-PackageDownloadLocation = "https://curl.se/"
```
minor nit

REUSE.toml
```diff
@@ -0,0 +1,126 @@
...
+SPDX-FileCopyrightText = "Daniel Stenberg, <daniel@haxx.se>, et al."
+SPDX-License-Identifier = "curl"
+
+[[annotations]]
+path = "packages/vms/macro32_exactcase.patch"
+SPDX-FileCopyrightText = "Daniel Stenberg, <daniel@haxx.se>, et al."
+SPDX-License-Identifier = "curl"
+
+[[annotations]]
+path = "packages/vms/readme"
```
vszakats, line 99:
```suggestion
path = [
  "packages/vms/build_vms.com",
  "packages/vms/curl_release_note_start.txt",
  "packages/vms/curlmsg.sdl",
  "packages/vms/macro32_exactcase.patch",
  "packages/vms/readme",
]
```
would it be a good opportunity to merge these entries?

perhaps more?

REUSE.toml
mxmehl: Fine with me, it was just copied from 

https://github.com/curl/curl/blob/2abfc759b98852cd916ebfb6207d296762b1171a/.reuse/dep5#L4

REUSE.toml
mxmehl: Good idea, I will come up with a suggestion

mxmehl: Done within bcbab9d

REUSE.toml
mxmehl: Fixed

REUSE.toml
```diff
@@ -0,0 +1,90 @@
...
+
+# checksrc control files
+[[annotations]]
+path = [
+  "docs/examples/.checksrc",
+  "lib/.checksrc",
+  "src/.checksrc",
+  "tests/libtest/.checksrc",
+
+]
```
bagder, line 88: I think this could be simplified with all files in a single list as they have the same copyright and same license.

docs/CONTRIBUTE.md
```diff
@@ -316,7 +316,7 @@ fine.
 This means that all files need to have their license and copyright information
 clearly stated. Ideally by having the standard curl source code header, with
 the SPDX-License-Identifier included. If the header does not work, you can use a
-smaller header or add the information for a specific file to the `.reuse/dep5`
+smaller header or add the information for a specific file to the `REUSE.toml`
```
bagder, line 319: I suppose there is no way to *not* add a new file to the root directory? The previous version was cleverly "hiding" there due to the leading dot in the name.

REUSE.toml
```diff
@@ -0,0 +1,90 @@
+# SPDX-License-Identifier: curl
+# SPDX-FileCopyrightText: Daniel Stenberg, <daniel@haxx.se>, et al.
+
+version = 1
+SPDX-PackageName = "curl"
+SPDX-PackageSupplier = "Daniel Stenberg <daniel@haxx.se>"
```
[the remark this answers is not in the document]
mxmehl: You are right, that's not quite helpful and probably rather useful if you get an SBOM for a package you receive from a vendor.

I'd be fine with removing this line. It's actually an artifact of the conversion script.

docs/CONTRIBUTE.md
mxmehl: Unfortunately that is not possible by design. A REUSE.toml can only describe files that are next to it or in child directories.

REUSE.toml
mxmehl: That's possible of course. Why didn't we/you do this with the old dep5 file? I figured it there to somehow get a better overview of these files.

Feel free to make a decision, I don't have a preference.

bagder: I have thought about making it so in the old file as well but never got around to it but since it is now getting an overhaul anyway I think it is worth doing that. It simplifies the file a lot and removes repetition outside of the file lists.

docs/CONTRIBUTE.md
bagder: I was afraid of that. It is a common "decease": adding config files to the source root. It does not scale and is user hostile. But I realize we need to live with this now.

I propose that we then add a comment at the top of this file that explains its purpose and why it is there so that users finding the file and checking it out have a chance to be less confused about it.

REUSE.toml
mxmehl: OK, did that in my recent commits!

REUSE.toml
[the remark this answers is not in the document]
mxmehl: Line has been removed

docs/CONTRIBUTE.md
mxmehl: Understood. I have amended the REUSE.toml file accordingly.

The REUSE project team is aware of this downside but all other options were worse after all, especially when aiming to make the specification still maintainable and logical for humans and tools. It follows the info-as-close-to-the-file-as-possible principle.

vszakats: It would be nice if we could name it `.reuse.toml` (or `.REUSE.toml`).

bagder:
> I noticed that `scripts/copyright.pl` also makes use of the dep5 file. I wonder whether the file is actually still needed.

I think it has outlived its purpose now. You can probably remove it in this PR.

bagder (approved):

bagder:
Thanks!

PR #14112, opened 2024-07-05
Update GIT-INFO.md for perl version requirements.

areitz:
[Per ChatGPT](https://chatgpt.com/share/5e1b4cd8-a10e-40d4-8049-abf1bed2a31e), the `<:crlf` discipline was added to `open()` in Perl 5.8.0. I tried running `scripts/managen` on a machine that has Perl 5.6.1 and got an error:

```
Unknown open() mode ':<crlf' at ../../scripts/managen line 777.
```

While I was in there, I also made the list of requirements a valid Markdown list, so it displays properly when viewed in GitHub.

## Before:

Screenshot 2024-07-05 at 6 45 37 PM

## After:

Screenshot 2024-07-05 at 6 45 23 PM

I made these separate commits, so you can not take either change if you wish.

Thanks!

### DIFF
--- a/GIT-INFO.md
+++ b/GIT-INFO.md
@@ -30,11 +30,11 @@ In environments that don't support configure (i.e. Windows), do this:
 For `autoreconf` and `configure` (not `buildconf.bat`) to work, you need the
 following software installed:
 
- o autoconf 2.57  (or later)
- o automake 1.7   (or later)
- o libtool  1.4.2 (or later)
- o GNU m4 (required by autoconf)
- o perl 5.8.0 (or later)
+ - autoconf 2.57  (or later)
+ - automake 1.7   (or later)
+ - libtool  1.4.2 (or later)
+ - GNU m4 (required by autoconf)
+ - perl 5.8.0 (or later)
 
 If you don't have perl and don't want to install it, you can rename the source
 file `src/tool_hugehelp.c.cvs` to `src/tool_hugehelp.c` and avoid having to
--- a/GIT-INFO.md
+++ b/GIT-INFO.md
@@ -27,12 +27,6 @@ In environments that don't support configure (i.e. Windows), do this:
 
 ## REQUIREMENTS
 
-See [docs/INTERNALS.md][0] for information on the requirements that you'll
-need or `autoreconf` and `configure` (not `buildconf.bat`) to work.
-
-If you don't have perl and don't want to install it, you can rename the source
-file `src/tool_hugehelp.c.cvs` to `src/tool_hugehelp.c` and avoid having to
-generate this file. This will give you a stubbed version of the file that
-doesn't contain actual content.
+See [docs/INTERNALS.md][0] for requirement details.
 
 [0]: docs/INTERNALS.md
--- a/GIT-INFO.md
+++ b/GIT-INFO.md
@@ -34,7 +34,7 @@ following software installed:
  - automake 1.7   (or later)
  - libtool  1.4.2 (or later)
  - GNU m4 (required by autoconf)
- - perl 5.8.0 (or later)
+ - perl     5.8.0 (or later)
 
 If you don't have perl and don't want to install it, you can rename the source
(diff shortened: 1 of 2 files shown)


vszakats:
FWIW, another GPT gives this answer to the same questions:
"The :crlf discipline for the open() function in Perl was added in Perl 5.10.0, which was released on December 18, 2007. This feature allows Perl to automatically handle line ending conversions when reading or writing files, making it easier to work with text files that may have been created on different platforms with different line ending conventions."

To "are you sure?", it replies with a different version:
"I apologize for the confusion in my previous responses. The :crlf discipline for the open() function in Perl was actually added in Perl 5.14.0, not Perl 5.10.0 as previously mentioned. Perl 5.14.0 was released on May 14, 2011. This feature allows Perl to handle line ending conversions when reading or writing files, simplifying the process of working with text files across different platforms. Thank you for pointing out the error, and I appreciate your understanding."

Is this somewhere documented in a canonical source?

dfandrich:
The canonical location for version information is docs/INTERNALS.md. I didn't
realize there were some versions duplicated in GIT-INFO.md and I don't think we
should bother keeping both. GIT-INFO.md should probably defer to
docs/INTERNALS.md for versions if it's going to mention specific requirements
at all.

INTERNALS.md says perl 5.6 is (was?) fine (released in 2000) so moving to 5.8
(released 2002) should be fine. Our "Old Linux" CI build uses 5.24 so there's
no problem there.

areitz:
I did some digging, but didn't study the `perl` source code too carefully. It seems like this is the commit that adds the `<:crlf` syntax:

https://github.com/Perl/perl5/commit/95c70f2077ea14f8e2426dbcac9828b259a86bf2

Looking at the release timeline, that is after Perl 5.6.0 but before 5.8.0:

https://en.wikipedia.org/wiki/Perl_5_version_history

I think this predates `git`, so I'm not sure how I'd track down if this change was backported to the 5.6.x train or stayed in the 5.8.x train. But the release notes for 5.8.0 say:

> New PerlIO implementation

And I believe this feature is part of `PerlIO`, so I'd believe that it first came in Perl 5.8.0.

areitz:
> The canonical location for version information is docs/INTERNALS.md. I didn't
realize there were some versions duplicated in GIT-INFO.md and I don't think we
should bother keeping both. GIT-INFO.md should probably defer to
docs/INTERNALS.md for versions if it's going to mention specific requirements
at all.

Thanks! I made that update, but I'm wondering if `GIT-INFO.md` is needed at all?

dfandrich:
GIT-INFO.md is not shipped in the tar balls, so it should be restricted to 
containing information of use to users who clone the git repo. That means 
autoconf info should be there since they won't get the pre-generated autoconf 
files, but I think that file should just point to INTERNALS.md for people who 
want the versions information. perl is needed by everyone, not just users clone 
from git, so I don't think it really bears mentioning there.

I know this isn't really about this PR, but if GIT-INFO is being updated I 
think it should be updated One Final Time so we don't need to keep two docs in 
sync. Let's see if anyone else agrees.

GIT-INFO.md
```diff
@@ -27,16 +27,12 @@ In environments that don't support configure (i.e. Windows), do this:
...
-For `autoreconf` and `configure` (not `buildconf.bat`) to work, you need the
-following software installed:
-
- o autoconf 2.57  (or later)
- o automake 1.7   (or later)
- o libtool  1.4.2 (or later)
- o GNU m4 (required by autoconf)
- o perl
+See [docs/INTERNALS.md][0] for information on the requirements that you'll
+need or `autoreconf` and `configure` (not `buildconf.bat`) to work.
```
bagder, line 31:
```suggestion
See [docs/INTERNALS.md][0] for requirement details.
```

... and I think the next paragraph can be completely removed.

areitz: Thanks! How does this look now?

bagder (approved):

bagder:
Thanks!

PR #14114, opened 2024-07-07
wolfssl: use larger error buffer when formatting errors

alexsn:
Currently we're using WOLFSSL_MAX_ERROR_SZ to define the error buffer size, this value is user defined which means it can be overwritten with -DWOLFSSL_MAX_ERROR_SZ=512 when building wolfssl and this overwrite is not exported to the users of wolfssl.

Instead of relying on WOLFSSL_MAX_ERROR_SZ we'll just use a 256 bytes error buffer and use wolfSSL_ERR_error_string_n to fill it thus dropping the dependency on WOLFSSL_MAX_ERROR_SZ altogether.

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -1009,6 +1009,23 @@ wolfssl_connect_step1(struct Curl_cfilter *cf, struct Curl_easy *data)
 }
 
 
+static char *wolfssl_strerror(unsigned long error, char *buf, size_t size)
+{
+  DEBUGASSERT(size);
+  *buf = '\0';
+
+  wolfSSL_ERR_error_string_n(error, buf, size);
+
+  if(!*buf) {
+    const char *msg = error ? "Unknown error" : "No error";
+    strncpy(buf, msg, size - 1);
+    buf[size - 1] = '\0';
+  }
+
+  return buf;
+}
+
+
 static CURLcode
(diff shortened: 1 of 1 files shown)


lib/vtls/wolfssl.c
```diff
@@ -1008,6 +1008,45 @@ wolfssl_connect_step1(struct Curl_cfilter *cf, struct Curl_easy *data)
...
+
+  len = wolfssl_version(buf, size);
+  DEBUGASSERT(len < (size - 2));
+  if(len < (size - 2)) {
+    buf += len;
+    size -= (len + 2);
+    *buf++ = ':';
+    *buf++ = ' ';
+    *buf = '\0';
+  }
```
bagder, line 1037: Adding the library name plus version number in the error string seems... unorthodox. I don't think we should do it here as we don't do it for other backends.

alexsn: > Adding the library name plus version number in the error string seems... unorthodox. I don't think we should do it here as we don't do it for other backends.

This is done in [openssl](https://github.com/curl/curl/blob/master/lib/vtls/openssl.c#L928) backend so I wanted to remain consistent.

alexsn: actually I think openssl implementation is buggy as `buf` is incremented after adding a prefix and that's what we return so the version is never actually printed.

bagder (approved):

bagder:
Thanks!

PR #14116, opened 2024-07-07
GHA/windows: usability improvements

vszakats:
- move `curl --version` into separate step.

- move configure log to separate step. Run on success, too.

- add step with `curl_config.h` dump (full and brief/sorted).

- make `autoreconf` a separate step.

- add each job configuration a short name.

- shorten job names.
  Dedupe/drop redundant info, introduce abbreviations:
  AM = autotools, CM = CMake, U = Unicode, R = Release, not -> `!`, etc.
  Instead of mentioning `debug`, mentioned when it's not.

- simplify `PATH` forming for MSVC jobs.
  It's sufficient to add the release binary directory of vcpkg, the debug one
  is redundant.
  Follow-up to e26cbe20cbedbea0ca743dd33880517309315cb2 #13979

- other minor tidy-ups.

Closes #14116


vszakats:
before:
Screen Shot 2024-07-07 before

after:
Screen Shot 2024-07-07 after

dfandrich:
"A" and "C" are pretty obscure without more context. "AM" and "CM" would be
marginally better, but they're still far from obvious to most developers.

vszakats:
Not pefect for sure. There is little visible space for these descriptions sadly.
I tried AT and CM earlier but they made the rest of text misaligned. I'll try
AM and CM, these seem to fix this issue, thanks for the idea!

PR #14118, opened 2024-07-07
Make variable example use it with expand

yedayak:
I used double quotes since it seemed required for powershell, so this example works in both (ba)sh and powershell as well as cmd.exe.


bagder:
Thanks!

PR #14135, opened 2024-07-09
Add Initial Early Data support

ad-chaos:
We had proposed a design in #14013, we've implemented it for OpenSSL initially for discussion, it can be implemented similarly for other SSL backends. 

Since early data is sent during the connection itself we defer the connection until `ossl_send` in order to send early data.
`CURLOPT_SSL_EARLYDATA` can be set to `1L` in order to enable sending early data.

Currently if early data was rejected, we do nothing. How would you suggest we handle this? Fallback to a normal send again or error out? Resending the request would require some co-operation from `multi_runsingle`, erroring out is easy.

All tests pass, the ones that fail are for missing documentation for `CURLOPT_SSL_EARLY_DATA`.


icing:
First off, this looks good, going in the right direction!

> To answer your question: on error you MUST resend the data as if "early_data" was never used.

Yes,  indeed. How this works in this design would be as follows:

1. first `ossl_send()` writes via `SSL_write_early_data()`, then continues the handshake. This may not complete on the first attempt and need to return.
2. Returning before handshake is complete must return CURLE_AGAIN and remember how much data was already written.
3. next `ossl_send()` calls continue the handshake, once that succeeds, check the EarlyData indicator from the server.
   a. Accepted: Return CURLE_OK with the amount of data send in the initial `SSL_write_early_data()`.
   b. Rejected: forget the earlier amount and `SSL_write_ex()` the complete `mem` content. Return what that resulted in

You may want to check in 3 that this is being called with at least the `len` that has been written into early data.

Another suggestion: if you defined a flag like `SSLSUPP_EARLYDATA`, you can set that in the `Curl_ssl_openssl` and avoid the additional `early_data` callback member. That way, you would not need to change the other TLS backends.

ad-chaos:
Thanks for the comments!

> This may not complete on the first attempt and need to return

Openssl says that unless `SSL_MODE_ENABLE_PARTIAL_WRITE` is enabled, it will not return success if the complete chunk was not written. We're assuming that it instead raises `SSL_WANT_READ` and in that case currently curl sets `written` to `-1` and tries the whole chunk again.

> SSL_MODE_ENABLE_PARTIAL_WRITE
Allow SSL_write_ex(..., n, &r) to return with 0 < r < n (i.e. report success when just a single record has been written). This works in a similar way for SSL_write(). When not set (the default), SSL_write_ex() or SSL_write() will only report success once the complete chunk was written. Once SSL_write_ex() or SSL_write() returns successful, r bytes have been written and the next call to SSL_write_ex() or SSL_write() must only send the n-r bytes left, imitating the behaviour of write().

Given this information, maybe we don't have to keep track of written bytes?

Thus, if we succeed in writing early data, we complete the connection then check early data status, if early data was rejected we resend the message on this newly formed connection.

icing:
> Thanks for the comments!
> 
> > This may not complete on the first attempt and need to return
> 
> Openssl says that unless `SSL_MODE_ENABLE_PARTIAL_WRITE` is enabled, it will not return success if the complete chunk was not written. We're assuming that it instead raises `SSL_WANT_READ` and in that case currently curl sets `written` to `-1` and tries the whole chunk again.
> 

You may early_write the whole buf, yes, but the handshake is not complete. So you do not know if you can skip the buf content *until* the server completes the handshake and acknowledges or rejects the early data.

> Thus, if we succeed in writing early data, we complete the connection then check early data status, if early data was rejected we resend the message on this newly formed connection.

In order to resend, you would need to have the `buf` with the same bytes as before. To achieve that, the method should return `CURLE_AGAIN` until the handshake is done, right?

ad-chaos:
> In order to resend, you would need to have the buf with the same bytes as before. To achieve that, the method should return CURLE_AGAIN until the handshake is done, right?

yes, so we'll set `written` to `-1` and return `CURLE_AGAIN`

bagder:
Can we add this feature as a new flag to `CURLOPT_SSL_OPTIONS`  instead of creating a new option?

ad-chaos:
> Can we add this feature as a new flag to CURLOPT_SSL_OPTIONS instead of creating a new option?

Done

Also removed the unnecessary checking before setting earlydata since only the backend implementation file will ever read it

ad-chaos:
What we are doing now is deferring connect until `ossl_send` and on the first successful send we complete the handshake. This doesn't maximally utilise sending early data since the application level request might be chunked and sent over multiple calls to `ossl_send`, protocols like http/2 can call `ossl_send` after connecting on the tls socket in `MSTATE_CONNECTING` to send the preface and settings and in `MSTATE_DO` send the actual request.

If we were to send the entire request as early data, we'll need to defer the completion of the connection until we've sent the entire request, maybe in something like `Curl_req_done`. That would also require an addition of a new method like `transfer_done` to cfilter and vtls. If the server rejects early data in this case, we'll have to keep track of all the request 'chunks' and retry them afterwards. Should we go down this route? Or keep the early data limited to whatever is in the first `ossl_send`?

As an admittedly incredibly terrible hack, the latest commit skips `cf_h2_connect` if early data is enabled and thus anything that would be sent during the `MSTATE_CONNECTING` phase is sent during the `MSTATE_DO` phase, thus maximally utilising sending of early data.

icing:
> What we are doing now is deferring connect until `ossl_send` and on the first successful send we complete the handshake. This doesn't maximally utilise sending early data since the application level request might be chunked and sent over multiple calls to `ossl_send`, protocols like http/2 can call `ossl_send` after connecting on the tls socket in `MSTATE_CONNECTING` to send the preface and settings and in `MSTATE_DO` send the actual request.
> 
> If we were to send the entire request as early data, we'll need to defer the completion of the connection until we've sent the entire request, maybe in something like `Curl_req_done`. That would also require an addition of a new method like `transfer_done` to cfilter and vtls. If the server rejects early data in this case, we'll have to keep track of all the request 'chunks' and retry them afterwards. Should we go down this route? Or keep the early data limited to whatever is in the first `ossl_send`?
> 
> As an admittedly incredibly terrible hack, the latest commit skips `cf_h2_connect` if early data is enabled and thus anything that would be sent during the `MSTATE_CONNECTING` phase is sent during the `MSTATE_DO` phase, thus maximally utilising sending of early data.

Yeah, this is tricky. I think you should scope the work to remain in the openssl filter only. As you noticed, the optimal use of early data will require all filters "above" to send as much as they can in the first call. Meaning HTTP/2 would need to send the initial frames together with the first stream frames.

tl;dr

To make this work in the best way, the complete `MSTATE_CONNECTING` would need to be eliminated and all the connect work must happen lazily in the `MSTATE_DO` phase. That is a lot of work to get it right.

Personally, I would restrict this PR to do the right thing for HTTP/1.1 request without any proxy involved.

ad-chaos:
This PR should now be ready for review, early data now works for http/1.1 on backends that support sending early data, we changed a couple things regarding session resumption on some of them:

- Ticket based session resumption was broken on wolfssl because it tried to get a session on the first connection during `step3` but the server can send a new session ticket anytime post handshake. We refactored that code to now use `wolfSSL_CTX_sess_set_new_cb` like openssl.
- mbedtls raises an error when it recieves a new session ticket during `mbed_recv` we store the session in the internal cache on that error
- gnutls gives us a zero length session ticket on tls 1.2 connections, so we've disabled them in those cases.

docs/libcurl/opts/CURLOPT_SSL_OPTIONS.md
```diff
@@ -86,6 +86,13 @@ certificate that supports client authentication in the OS certificate store it
 could be a privacy violation and unexpected.
 (Added in 7.77.0)
 
+## CURLSSLOPT_EARLYDATA
+
+Tell libcurl to try sending application data as TLS1.3 early data. This option
+is only supported for OpenSSL, WolfSSL and GnuTLS. This option only works when
+using HTTP/1.1 without a proxy otherwise its a no-op.
```
bagder, line 93: What happens if this flag is set and the TLS negotiation ends up with a lower version than 1.2? It would be good to mention that here.

Also, since QUIC uses TLS 1.3 internally, it might be good to explicitly mention if that is also covered by this flag or not.

docs/libcurl/symbols-in-versions
```diff
@@ -1053,6 +1053,7 @@ CURLSSLOPT_NATIVE_CA            7.71.0
 CURLSSLOPT_NO_PARTIALCHAIN      7.68.0
 CURLSSLOPT_NO_REVOKE            7.44.0
 CURLSSLOPT_REVOKE_BEST_EFFORT   7.70.0
+CURLSSLOPT_EARLYDATA            8.09.0
```
bagder, line 1056: 8.10.0 now presumably

lib/vtls/gtls.c
```diff
@@ -344,8 +344,8 @@ static CURLcode handshake(struct Curl_cfilter *cf,
       return CURLE_SSL_CONNECT_ERROR;
     }
 
-    /* Reset our connect state machine */
-    connssl->connecting_state = ssl_connect_1;
+    /* Advance our connect state machine */
+    connssl->connecting_state = ssl_connect_3;
```
bagder, line 348: Is this not odd that you can just change this unconditionally even if early data is not requested?

lib/vtls/gtls.c
```diff
@@ -845,10 +845,13 @@ static CURLcode gtls_client_init(struct Curl_cfilter *cf,
...
-  /* Disable TLS session tickets */
-  init_flags |= GNUTLS_NO_TICKETS;
-#endif
+  /* Disable TLS session tickets for tlsv1.2
+   * Its incredibly strange, enabling it makes gnutls report and store zero
+   * length session tickets on tls1.2 connections
+   * */
+  if(config->version == CURL_SSLVERSION_TLSv1_2) {
+    init_flags |= GNUTLS_NO_TICKETS;
+  }
```
bagder, line 854: This change removes the `#if defined(GNUTLS_NO_TICKETS)`. Why? This breaks builds with older GnuTLS versions.

lib/vtls/openssl.c
```diff
@@ -4861,13 +4922,42 @@ static ssize_t ossl_send(struct Curl_cfilter *cf,
   struct ssl_connect_data *connssl = cf->ctx;
   struct ossl_ctx *octx = (struct ossl_ctx *)connssl->backend;
 
+  size_t written;
```
bagder, line 4925: As you change type for this variable, can't it be a `ssize_t` now then since that is what is returned - to save some typecasts?

docs/libcurl/opts/CURLOPT_SSL_OPTIONS.md
ad-chaos: > What happens if this flag is set and the TLS negotiation ends up with a lower version than 1.3

in this case, or when the server sends a `HelloRetryRequest`, openssl and others either raise an error or let us know that early data was rejected. We resend the request again in those cases post-handshake, since early data was not sent.

Should we mention that this option only works on a best effort basis, and in situations when it wasn't possible to send early data we fallback to resending the request post handshake?

We didn't implement early data for QUIC

lib/vtls/gtls.c
ad-chaos: This is more of a refactor so that the code looks more like what is happening in all the other backends
`ssl_connect_1` -> Setting up context, and other bookkeeping
`ssl_connect_2` -> Actual Handshake
`ssl_connect_3` -> Verifying peers

Previously `ssl_connect_3` was not utilised, `handshake()` completed the handshake and in `gtls_connect_common` the verification part followed unconditionally.

lib/vtls/gtls.c
ad-chaos: Apologies, that was not intended

lib/vtls/gtls.c
```diff
@@ -1700,20 +1710,37 @@ gtls_connect_common(struct Curl_cfilter *cf,
       result = rc;
       goto out;
     }
+    backend = (struct gtls_ssl_backend_data *)
+                  connssl->backend;
```
bagder, line 1714: funny indent level?

lib/vtls/gtls.c
```diff
@@ -1700,20 +1710,37 @@ gtls_connect_common(struct Curl_cfilter *cf,
...
+    }
   }
 
   /* Finish connecting once the handshake is done */
-  if(ssl_connect_1 == connssl->connecting_state) {
-    struct gtls_ssl_backend_data *backend =
-      (struct gtls_ssl_backend_data *)connssl->backend;
-    gnutls_session_t session;
+  if(ssl_connect_3 == connssl->connecting_state) {
+    backend = (struct gtls_ssl_backend_data *)connssl->backend;
```
bagder, line 1743: isn't this just assigning it again exactly as on line 1713?

lib/vtls/gtls.c
```diff
@@ -1700,20 +1710,37 @@ gtls_connect_common(struct Curl_cfilter *cf,
...
     DEBUGASSERT(backend);
```
bagder, line 1744: This assert probably needs to be moved to immediately after the first assigning of this variable.

docs/libcurl/opts/CURLOPT_SSL_OPTIONS.md
```diff
@@ -86,6 +86,16 @@ certificate that supports client authentication in the OS certificate store it
...
+## CURLSSLOPT_EARLYDATA
+
+Tell libcurl to try sending application data as TLS1.3 early data. This option
+is only supported for OpenSSL, WolfSSL, MbedTLS and GnuTLS. This option works
+on a best effort basis, in cases when it wasn't possible to send early
+data the request is resent normally post-handshake.
+This option does not work when using QUIC or HTTP/2, it only works when using
+HTTP/1.1 without a proxy, otherwise its a no-op.
+(Added in 8.10.0)
+
```
bagder, line 98: How is this limited to HTTP/1 only? I don't see any changes in the HTTP code, so why does it limit itself to just version 1 and not to other versions or other protocols?

include/curl/curl.h
```diff
@@ -940,6 +940,10 @@ typedef enum {
    a client certificate for authentication. (Schannel) */
 #define CURLSSLOPT_AUTO_CLIENT_CERT (1<<5)
 
+/* - CURLSSLOPT_EARLYDATA tells libcurl to try sending application data using
+   tls 1.3 early data */
```
bagder, line 944:
```suggestion
/* If possible, send data using TLS 1.3 early data */
```

bagder (commented):
How do you test this? Have you thought about how to make curl tests for this?

docs/libcurl/opts/CURLOPT_SSL_OPTIONS.md
ad-chaos: It kind of works with http/2 but it doesn't maximally utilise sending early data, it sends the initial http/2 frames as early data and the rest of them are sent post handhshake. https://github.com/curl/curl/pull/14135#issuecomment-2230520071, maximally utilising sending early data would require significant refactoring of `multi_runsingle` https://github.com/curl/curl/pull/14135#issuecomment-2230536121

http/3 doesn't use the `Curl_cft_ssl` cfilter hence it isn't supported by this PR

The actual behaviour is that only the first tls record sent over the connection (as part of any protocol) is sent as early data, the rest is sent post handshake.

ad-chaos:
> How do you test this? Have you thought about how to make curl tests for this?

We tested the implementation using
```c
#include <curl/curl.h>
#include <stdio.h>
#include <stdlib.h>

void doit(CURLSH* share, const char *url)  {
  CURL* curl = curl_easy_init();

  curl_easy_setopt(curl, CURLOPT_HTTPGET, 1L);
  curl_easy_setopt(curl, CURLOPT_SSL_OPTIONS, CURLSSLOPT_EARLYDATA);
  curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_3);
  curl_easy_setopt(curl, CURLOPT_SHARE, share);

  curl_easy_perform(curl);
  curl_easy_cleanup(curl);
}

int test(const char* URL) {
  CURLSH* share = curl_share_init();
  curl_share_setopt(share, CURLSHOPT_SHARE, CURL_LOCK_DATA_SSL_SESSION);

  doit(share, URL);
  doit(share, URL);
  return 0;
}

int main(void) {
  test("https://google.com");
}
```

If it succeeds to send early data then it'll print something like `* OpenSSL SSL_write_early_data: Early data was accepted`, else it'll print `* OpenSSL SSL_write_early_data: Early data was rejected`.

What would be a good way to test it? Would inspecting curl verbose output be a reasonably strategy? How do we go about doing that?

icing (changes requested):
My summary on reviewing this PR and how I think we should go forward:

1. I think we should implement Early Data on *one* TLS backend first, before changing all the others. This would make it easier to get it "air tight" and allow us to focus the technical discussions on one source. What you did on the other TLS backends remains good work. We just save effort by doing one first.
2. You correctly foresaw the problem of getting a filter `recv()` call before the handshake is done. Looking at OpenSSL, which seems the most advanced here, you simulate a 0-length early data send and return, but do not progress the handshake. I am not sure if a subsequent `ossl_send()` does the right thing.
3. In openssl.c:5085, on having successfully written the early data bytes (meaning OpenSSL has sent them off) you proceed the `ossl_connect()` as should be, but the DEBUGASSERT() on line 5089 does not need to hold. The `ossl_connect()` can have retries. If you remove the ASSERT, the logic does not hold, as the `ossl_send()` would return success before the server answer is back.

The handling of all possible states here is quite complex. There is `SSL_write_early_data()` that may fail to send everything and need retrying. And even if it succeeds, it means only that the ClientHello plus data has been flushed down the filter chain. It might be on its way to the server or it might be in the early data buffer of a TLS filter "below" this one. The response from the sever, which may be negative, may take several calls to arrive and be processed.

I think it works better when we enhance `ossl_connect()` and have that as a "guard" at the beginnings of `ossl_send()/ossl_recv()`. Only when that succeeds can these functions progress as before. Something like:

```c
CURLcode ossl_connect_send(struct Curl_cfilter *cf,
                         struct Curl_easy *data,
                         const void *buf, size_t blen,
                         size_t *pconsumed);
```
which `SSL_write_early_data()` the `buf` until that succeeds, then progresses the connect, returning CURLE_AGAIN until finished and on CURLE_OK returns in `pconsumed` how many of the early data bytes have been accepted by the server.

Analog a

```c
CURLcode ossl_connect_recv(struct Curl_cfilter *cf,
                         struct Curl_easy *data);
```
for guarding `ossl_recv()`. This requires the early data to be buffered internally, so the send/recv calls can be interleaved.

Does that make sense?

I do not know how much time you can spend on this. If you want me to, I can make a proposal in a separate PR.

ad-chaos:
Superseded by #15211

PR #14136, opened 2024-07-09
cmake: detect `nghttp2` via `pkg-config`, enable by default

vszakats:
- also detect nghttp2 via `pkg-config` to match nghttp3 detection
  and autotools.

- enable nghttp2 by default to match autotools.

Cherry-picked from #14097
Closes #14136


CMake/FindNGHTTP2.cmake
```diff
@@ -21,21 +21,41 @@
 # SPDX-License-Identifier: curl
 #
 ###########################################################################
-include(FindPackageHandleStandardArgs)
 
-find_path(NGHTTP2_INCLUDE_DIR "nghttp2/nghttp2.h")
+if(UNIX)
+  find_package(PkgConfig QUIET)
```
talregev, line 26: not only unix, also windows with vcpkg

vszakats: Thanks, I was wondering about it.

But, I'll handle it in a separate commit, because at the moment almost all such checks are limited to `UNIX`, none of them should be. (or at least nghttp3 shouldn't from where I copied this logic.)

vszakats: MSYS2/mingw-w64 builds might be supporting this too.

vszakats: Or... I might tackle this here and handle the rest separately to avoid a merge queue.

vszakats: Done

vszakats:
While testing a similar change in smaller scale in libssh2, I bumped into two issues:

1. with the pattern used in `CMake/Find*.cmake`, `pkg-config` output is used as a _hint_ for the native `find_*()` functions. This fails when the package returned by `pkg-config` has a dependency, with their header path(s) appended to the returned header list. After passing this list through the regular find (as a hint), these additional paths are lost and dependency headers are not found. (seen it with Libgcrypt.)

2. `pkg-config` tends to find stuff it shouldn't in a cross-build it find the host-native package. FWIW this is a problem also with `find_library()` and `find_path()` according to past notes (for `libssh` specifically). But, adding `pkg-config` to an existing dependency might break downstream builds. Could it help to exclude `pkg-config` for `NOT CMAKE_CROSSCOMPILING`?

I wonder if there is a canonical CMake pattern that avoids these pitfalls.

PR #14139, opened 2024-07-09
Improve code appearance format

mohammadshg84:


bagder (commented):
I don't think this is necessary. The code looks fine already.

PR #14147, opened 2024-07-10
Small memory leak fix in hostip.c

iCMDdev:
Hi!

First time looking at the curl source code, and I think I found a (most likely insignificant) memory leak.
Either way, I added a `free(ca);` to prevent memory leaks if `Curl_inet_pton(AF_INET6, "::1", ipv6)` fails in `*get_localhost6(int port, const char *name)`.

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -536,8 +536,10 @@ static struct Curl_addrinfo *get_localhost6(int port, const char *name)
   sa6.sin6_port = htons(port16);
   sa6.sin6_flowinfo = 0;
   sa6.sin6_scope_id = 0;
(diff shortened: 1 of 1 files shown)


bagder:
While this is "technically correct", in that **if** `Curl_inet_pton()` returns failure this would leak. But given this input, the function will not fail so it will never return anything but 1 for this code flow. 

I rather suggest that we instead make the code reflect this reality, perhaps like this:

~~~c
    (void)Curl_inet_pton(AF_INET6, "::1", ipv6);

bagder (changes requested):
See comment

bagder (approved):

bagder:
Thanks!

iCMDdev:
No problem, thanks for the commit!

PR #14149, opened 2024-07-10
Wrap 2 Example URLs that Overrun Styling in Docs

martinpeck:
Within MANUAL.md there are two example urls where it is impossible to read the end of the sample because the text overruns the styling. Both appear in the `POST (HTTP)` section of the document, and illustration of the issue is shown below (I have selected the text in question so that the hidden text is visible):

## Example 1
image

## Example 2
image

The issue appears to occur regardless of zoom.

To address this I have simply forced these examples to wrap onto two lines, using the same indentation style as other examples within this document.


bagder (approved):

bagder:
Thanks!

PR #14151, opened 2024-07-10
ECH: Fix leak and zero-length crash

sftcd:
This PR fixes a leak and a crash that can happen when curl encounters bad HTTPS RR values in DNS. We're starting to do better testing of that kind of thing and e.g. have published bad HTTPS RR values at dodgy.test.defo.ie. There could be some more fixes to come as we make those tests more systematic and add test cases but we figured it would be better to create this small PR now in any case.

The problems here were found by @niallor as part of  work he's doing on better handling of multiple HTTPS RR values in curl.


bagder (approved):

bagder:
The manpage tests can fail in CI if the test runs on the other side of the date shift from the build, as it then expects another date in the generated output... I need to fix this somehow.

sftcd:
If me doing something fixes that man-page CI thing, just say...

bagder:
Thanks!

PR #14155, opened 2024-07-11
macos: undo `availability` macro enabled by Homebrew gcc

### LINKED ISSUE #13700: GCC-14 build fails on MacOSX using vcpkg

vszakats:
Homebrew gcc builds starting with 12.4.0, 13.3.0 and 14.1.0 enabled
the `availability` attribute.

This broke builds because the way the Apple SDK uses attributes (when
available) are incompatible with how gcc accepts them. Causing these
errors:
```
  error: attributes should be specified before the declarator in a function definition
  error: expected ',' or '}' before
```

Upstream commits implementing the `availability` macro:
gcc-12: https://github.com/iains/gcc-12-branch/commit/fd5530b7cb0012bf4faeddd45e13054a1dfa6783
gcc-13: https://github.com/iains/gcc-13-branch/commit/cb7e4eca68cfc4763474e2eb0935a844458842a8
gcc-14: https://github.com/iains/gcc-14-branch/commit/ff62a108865a6403f5017380d7018250c1d3306f

The project above is a Darwin gcc compatibility pack, that is applied
to Homebrew gcc builds.

This patch works by redefining the `availability` macro to an invalid
value, making `__has_attribute(availability)` checks fail, stopping
Apple SDK from inserting the incompatible attributes.

It also replaces the previous, local workaround for `lib/macos.c`.

Example with gcc 12.4.0 with macOS SDK 14.0 (Xcode 15.0.1):
```
In file included from <path-to-sdk>/MacOSX14.0.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CoreFoundation.h:54,
                 from <path-to-sdk>/MacOSX14.0.sdk/System/Library/Frameworks/SystemConfiguration.framework/Headers/SCDynamicStoreCopySpecific.h:30,
                 from /Users/runner/work/curl/curl/lib/macos.c:33,
                 from /Users/runner/work/curl/curl/build/lib/CMakeFiles/libcurl_shared.dir/Unity/unity_0_c.c:244:
<path-to-sdk>/MacOSX14.0.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFUserNotification.h:126:1: error: attributes should be specified before the declarator in a function definition
  126 | CF_INLINE CFOptionFlags CFUserNotificationCheckBoxChecked(CFIndex i) API_AVAILABLE(macos(10.0)) API_UNAVAILABLE(ios, watchos, tvos) {return ((CFOptionFlags)(1UL << (8 + i)));}
      | ^~~~~~~~~
```
Ref: https://github.com/curl/curl/actions/runs/9787982387/job/27025351601?pr=14096#step:7:18

The gcc vs. llvm/clang incompatibility possibly tracked here upstream:
  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108796
More info:
  https://github.com/llvm/llvm-project/issues/81767
  https://github.com/gcc-mirror/gcc/commit/8433baadec88e5f31fa141b6d78094e91256079d
  https://discourse.llvm.org/t/changing-attribute-ast-printing-location-for-gcc-compatibility/73215
  https://reviews.llvm.org/D159362

Follow-up to db135f8d7207b20d531e7e2100a49f3e16bdcfab #14119
Ref: https://github.com/curl/curl/pull/14091#issuecomment-2222703468
Fixes #13700
Cherry-picked from #14097
Closes #14155

### DIFF
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -54,7 +54,7 @@
   !defined(__clang__) &&                 \
   defined(__GNUC__) && __GNUC__ >= 12 && \
   defined(__has_attribute)
-#define availability curl_invalid
+#define availability curl_pp_attribute_disabled
 #endif
 
 /*
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -40,6 +40,23 @@
 #include <_mingw.h>
 #endif
 
+/* Workaround for Homebrew gcc 12.4.0, 13.3.0, 14.1.0 and newer (as of 14.1.0)
+   that started advertising the `availability` attribute, which then gets used
+   by Apple SDK, but, in a way incompatible with gcc, resulting in a misc
+   errors inside SDK headers, e.g.:
+     error: attributes should be specified before the declarator in a function
+            definition
+     error: expected ',' or '}' before
+   Followed by missing declarations.
+   Fix it by overriding the built-in feature-check macro used by the headers
+   to enable the problematic attributes. This makes the feature check fail. */
+#if defined(__APPLE__) &&                \
+  !defined(__clang__) &&                 \
+  defined(__GNUC__) && __GNUC__ >= 12 && \
+  defined(__has_attribute)
+#define availability curl_invalid
+#endif
+
 /*
  * Disable Visual Studio warnings:
  * 4127 "conditional expression is constant"
--- a/lib/macos.c
+++ b/lib/macos.c
@@ -30,17 +30,7 @@
 
 #include "macos.h"
 
-/* Certain Apple SDK v13.0+ headers are incompatible with the GCC compiler.
-   As a workaround, use a minimal header and define the function we need,
-   to avoid hitting those incompatibilities when compiling with GCC. */
-#if defined(__clang__)
 #include <SystemConfiguration/SCDynamicStoreCopySpecific.h>
-#else
-#include <CoreFoundation/CFDictionary.h>
-typedef const struct CF_BRIDGED_TYPE(id) __SCDynamicStore * SCDynamicStoreRef;
-CFDictionaryRef __nullable
-SCDynamicStoreCopyProxies(SCDynamicStoreRef __nullable store);
-#endif
 
 CURLcode Curl_macos_init(void)
 {


saschasc:
@vszakats I have the same issue with grpc. So there is no way in fixing that behaviour in Homebrew `gcc`?

vszakats:
@saschasc An obvious fix would be to reverse the breaking patches in `https://github.com/iains/gcc-*-branch/` (aka "hack" layers) and Homebrew GCC applying the fixed versions.

Perhaps the best would be to report this to either Homebrew or the hack-layer maintainer(s) to learn about the motivations for these patches and find the best way to fix it.

saschasc:
It seems that they have been removed in the newest gcc 14.2 version.

See here https://github.com/iains/gcc-14-branch/commit/b50a0534fd2b8abfa298501f90cc019e797cbf59

vszakats:
Are you sure it is reverting the problematic `availability` patch? The linked patch seems to delete parts of the hack-layer patch that is now being dealt with its local implementation of `availability`. If so it means the problem is not resolved, and instead the project is doubling down on this feature. (Though all this is quite confusing, so I might be mis-reading it.)

`availability` support is still present in the latest master:
https://github.com/iains/gcc-14-branch/blob/848f27a4af59f534ab4c611d18257f66e2459a1d/gcc/config/darwin-protos.h#L91

vszakats:
@saschasc Thanks for raising this upstream. If clang-like positioning for `__attribute__` would land in GCC, it'd be helpful indeed. It'd allow us to drop this individual hack from curl and start using these attributes as designed, with compatible GCC builds.

I guess the next elephant in the room is the issue of the GCC hack-layer being bound to a specific SDK version _when building GCC itself_, and the hack-layer breaking build when this doesn't match the actual SDK available (or selected) upon use. But, that's a different topic.

saschasc:
A new Homebrew GCC version 14.2 should be released soon (see here https://github.com/Homebrew/homebrew-core/pull/194797). Perhaps you would like to look at/test this again? I don't know what consequences the current libcurl hack has by simply ignoring the attributes.

Regarding the next elephant in the room: As I understand it, Iains is the only maintainer for gcc on macOS. If he had the resources, he would like to solve this differently. But without consideration from Apple this will be a difficult undertaking and clang and gcc have differences. But communication is very good and Iains was very helpful.

vszakats:
> A new Homebrew GCC version 14.2 should be released soon (see here [Homebrew/homebrew-core#194797](https://github.com/Homebrew/homebrew-core/pull/194797)). Perhaps you would like to look at/test this again? I don't know what consequences the current libcurl hack has by simply ignoring the attributes.

The curl hack just deactivates the "availability" attribute, making the SDK headers think that it's unsupported by the compiler. It makes things behave like they did before the hack-layer enabled this attribute in these recent updates.

The updated patch looks promising. Once it hits the CI runner, I can test it by disabling the curl hack for 14.2. (I cannot test locally as Monterey is out of support and can't compile packages from source due the resources it needs.)
 
If it works, we can then disable the hack for 14.3 and newer. (We cannot pick up Homebrew package revisions from C headers, and feature detection for this feels overkill.)

> Regarding the next elephant in the room: As I understand it, Iains is the only maintainer for gcc on macOS. If he had the resources, he would like to solve this differently. But without consideration from Apple this will be a difficult undertaking and clang and gcc have differences. But communication is very good and Iains was very helpful.

It's a potentially large-scale fix. Understandable.

After the 14.2 fix SDKs will match indeed, but will mismatch again, when the GHA runner decides to upgrade Xcode or set a new default SDK. It's a constantly changing situation, where it sometimes matches, but most of the time doesn't. It can happen outside GHA too: Xcode and default SDK is updated on different schedules than Homebrew gcc bottle is rebuilt and deployed/updated.

vszakats:
The Homebrew update of gcc 14.2.0_1 did not yet hit the curl CI, but hit curl-for-win's.
It broke the curl build there, as it seems the fix converts the existing curl workaround
into something breaking. It means that the curl workaround fixes 12.4.0, 13.3.0, 14.1.0,
14.2.0 but breaks 14.2.0_1. Opened #15508 as an attempt to fix it.

The problem is that there is no way to distinguish between 14.2.0 and 14.2.0_1 in 
source (that I know of), so there will remain one broken gcc version/build, even after
merging the fix above. This broken version will be 14.2.0 which seems to be the
least bad option.

Also it seems there are no (not yet?) similar Homebrew updates for 12.4.0, 13.3.0, so
the workaround needs to be kept for those major versions.

PR #14160, opened 2024-07-11
tidy-up: adjust casing of project names

vszakats:
Mostly TLS/SSH project name.

Closes #14160


vszakats:
Would that be OK, too?:
```diff
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -1745,7 +1745,7 @@ static void *wolfssl_get_internals(struct ssl_connect_data *connssl,
 }
 
 const struct Curl_ssl Curl_ssl_wolfssl = {
-  { CURLSSLBACKEND_WOLFSSL, "WolfSSL" }, /* info */
+  { CURLSSLBACKEND_WOLFSSL, "wolfSSL" }, /* info */
 
 #ifdef KEEP_PEER_CERT
   SSLSUPP_PINNEDPUBKEY |
```

bagder (commented):
Maybe you can also remove `libSSH`, `boringssl` and `libressl` from `.github/scripts/spellcheck.words` to help us keep the correct spellings better?

bagder:
> Would this be OK?:

It would for `curl_global_sslset()` at least, as that makes case intensive checks for the backend names. Can't think of where it is actually used more?

vszakats:
> `curl_global_sslset`

That's the only place I could find too. I went with lowercase to match the other backend names in this field.

PR #12331, opened 2023-11-15
build: add more picky warnings and fix them

vszakats:
Enable more picky compiler warnings. I've found these options in the
nghttp3 project when implementing the CMake quick picky warning
functionality for it [1].

`-Wunused-macros` was too noisy to keep around, but fixed a few issues
it revealed while testing.

- autotools: reflect the more precisely-versioned clang warnings.
  Follow-up to 033f8e2a08eb1d3102f08c4d8c8e85470f8b460e #12324
- autotools: sync between clang and gcc the way we set `no-multichar`.
- autotools: avoid setting `-Wstrict-aliasing=3` twice.
- autotools: disable `-Wmissing-noreturn` for MSYS gcc targets [2].
  It triggers in libtool-generated stub code.

- lib/timeval: delete a redundant `!MSDOS` guard from a `WIN32` branch.

- lib/curl_setup.h: delete duplicate declaration for `fileno`.
  Added in initial commit ae1912cb0d494b48d514d937826c9fe83ec96c4d
  (1999-12-29). This suggests this may not be needed anymore, but if
  it does, we may restore this for those specific (non-Windows) systems.
- lib: delete unused macro `FTP_BUFFER_ALLOCSIZE` since
  c1d6fe2aaa5a26e49a69a4f2495b3cc7a24d9394.
- lib: delete unused macro `isxdigit_ascii` since
  f65f750742068f579f4ee6d8539ed9d5f0afcb85.
- lib/mqtt: delete unused macro `MQTT_HEADER_LEN`.
- lib/multi: delete unused macro `SH_READ`/`SH_WRITE`.
- lib/hostip: add `noreturn` function attribute via new `CURL_NORETURN`
  macro.
- lib/mprintf: delete duplicate declaration for `Curl_dyn_vprintf`.
- lib/rand: fix `-Wunreachable-code` and related fallouts [3].
- lib/setopt: fix `-Wunreachable-code-break`.
- lib/system_win32 and lib/timeval: fix double declarations for
  `Curl_freq` and `Curl_isVistaOrGreater` in CMake UNITY mode [4].
- lib/warnless: fix double declarations in CMake UNITY mode [5].
  This was due to force-disabling the header guard of `warnless.h` to
  to reapply it to source code coming after `warnless.c` in UNITY
  builds. This reapplied declarations too, causing the warnings.
  Solved by adding a header guard for the lines that actually need
  to be reapplied.
- lib/vauth/digest: fix `-Wunreachable-code-break` [6].
[...]

PR #1722, opened 2017-08-02
darwin: silence compiler warnings

bagder:
With a clang pragma and three type fixes

Previous attempt to silencing compiler warnings in darwinssl.c was found in #1705

PR #1706, opened 2017-07-27
travis: add a normal macos build using darwinssl

bagder:

PR #14162, opened 2024-07-11
sectransp: fix clang compiler warnings, stop silencing them

vszakats:
Fix `-Wpointer-bool-conversion` warnings with the method suggested by
both Apple clang and mainline llvm. This was already tried and dropped
in #1705 (in year 2017), but the issue reported there no longer
replicates.

Verified with Apple clang 14, llvm 15, llvm 18 and gcc 11, 14 that the
generated objects are bit by bit identical before and after this patch.

Also:
- stop silencing `-Wtautological-pointer-compare`. This warning don't
  seem to be appearing anymore (with or without this patch), at least
  with the tested compilers and SDKs (clang 13.1.6-16.0.0beta, llvm 15,
  18, gcc 11, 14) and minimum macOS target of 10.8. Older targets fail
  to build curl with SecureTransport.

- silence `-Wunreachable-code` for clang only. Previously I applied it
  also to GCC, by mistake.
  Ref: https://github.com/curl/curl/pull/12331/commits/8d7172d20a48ebc6c1b1d94a76e2c5fb19dd9bfa

Apple clang `-Wpointer-bool-conversion`:
```
curl/lib/vtls/sectransp.c:1103:6: error: address of function 'SSLCreateContext' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]
  if(SSLCreateContext) {  /* use the newer API if available */
  ~~ ^~~~~~~~~~~~~~~~
curl/lib/vtls/sectransp.c:1103:6: note: prefix with the address-of operator to silence this warning
  if(SSLCreateContext) {  /* use the newer API if available */
     ^
     &
```
Ref: https://github.com/curl/curl/actions/runs/9819538439/job/27113201384#step:8:382

llvm `-Wpointer-bool-conversion`:
```
curl/lib/vtls/sectransp.c:2663:8: error: address of function 'SSLCreateContext' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]
    if(SSLCreateContext)
    ~~ ^~~~~~~~~~~~~~~~
curl/lib/vtls/sectransp.c:2663:8: note: prefix with the address-of operator to silence this warning
    if(SSLCreateContext)
       ^
       &
```
Ref: https://github.com/curl/curl/actions/runs/9819538439/job/27113200291#step:8:417

gcc still needs `-Waddress` suppressed to avoid these:
```
curl/lib/vtls/n/sectransp.c: In function 'getsubject':
curl/lib/vtls/n/sectransp.c:379:6: warning: the address of 'SecCertificateCopyLongDescription' will always evaluate as 'true' [-Waddress]
  379 |   if(&SecCertificateCopyLongDescription)
      |      ^
[...]
```

Follow-up to 59cadacfcc1d39472245979cdbd614c7a9af6f0d #14128
Follow-up to af271ce9b9717ba289417e9cbb7f278c2a12f959 #1722
Follow-up to 2b7ce3f56dfede107113c6de7d0ca457109d3eda #1706
Cherry-picked from #14097
Closes #14162

---

- [x] run CI test and stop silencing `-Wtautological-pointer-compare` too. https://github.com/curl/curl/actions/runs/9899018157

### DIFF
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -44,7 +44,6 @@
 #ifdef __GNUC__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Waddress"
-#pragma GCC diagnostic ignored "-Wunreachable-code"
 #endif
 
 #include <limits.h>
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -41,11 +41,6 @@
 #include "strerror.h"
 #include "cipher_suite.h"
 
-#ifdef __clang__
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wtautological-pointer-compare"
-#endif /* __clang__ */
-
 #ifdef __GNUC__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Waddress"
@@ -2919,10 +2914,6 @@ const struct Curl_ssl Curl_ssl_sectransp = {
 
 #ifdef __GNUC__
 #pragma GCC diagnostic pop
-#endif
-
-#ifdef __clang__
-#pragma clang diagnostic pop
 #endif
 
 #endif /* USE_SECTRANSP */
--- a/lib/vtls/sectransp.c
+++ b/lib/vtls/sectransp.c
@@ -41,6 +41,11 @@
 #include "strerror.h"
 #include "cipher_suite.h"
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunreachable-code"
+#endif /* __clang__ */
+
 #ifdef __GNUC__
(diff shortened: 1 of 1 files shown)


vszakats:
> In #1705, @nickzman suggested that these feature checks (which are causing an assortment of compiler warnings, some unfixable) could be improved by using `__builtin_available()`. But discarded the idea for compatibility reasons, because it requires Xcode 9, at beta at the time, and released later in 2017. Since that thread, `sectransp.c` started using `__builtin_available()`. Also 7 years passed, and probably very few people rely on Xcode 8 from 2016 (or earlier) anymore.
> 
> I wonder if it'd be a good time to switch to `__builtin_available()`?

Well, storno, `__builtin_available()` isn't supported by GCC. It's might be one the reasons GCC tests regress compared to clang/llvm ones, though that's investigation yet to be done.

PR #14163, opened 2024-07-11
GHA/windows: build uwp tests with MSVC

talregev:
build uwp tests

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -544,7 +544,7 @@ jobs:
           arch: 'x64'
           plat: 'uwp'
           type: 'Debug'
-          tflags: 'skipall'
+          tflags: 'skiprun'
           config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON                      -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_OPENSSL_QUIC=ON'
         - name: 'libressl'
           install: 'brotli zlib zstd libpsl nghttp2 libressl libssh2[core,zlib]'


vszakats:
We have an UWP build for tests, with mingw. Is there anything new that an MSVC variant brings here?

talregev:
> We have an UWP build for tests, with mingw. Is there anything new that an MSVC variant brings here?

You can say that mingw also compile openssl, so why we need to compile with msvc. When cmake generate for msvc it generate with solution files and compile with visual studio c++ compiler and not gcc.
I don't know what happened in uwp with vcpkg vs uwp on mingw

vszakats:
Yeah, but in case of tests, the code compiled is totally the same, except maybe 2 lines. We also cannot run tests with UWP, so the value is already low. OpenSSL is on the other end of this spectrum.

talregev:
I think it not add time overall, because with compile the uwp is shorter then other tests that in parallel.

Let me know if you want me to close this PR.

talregev:
> Yeah, but in case of tests, the code compiled is totally the same, except maybe 2 lines. We also cannot run tests with UWP, so the value is already low. OpenSSL is on the other end of this spectrum.

code is the same with different compiler maybe.

talregev:
I think this explains the difference 
https://stackoverflow.com/questions/12909139/can-windows-store-applications-be-built-with-mingw

vszakats:
> https://stackoverflow.com/questions/12909139/can-windows-store-applications-be-built-with-mingw

The accepted answer is saying that mingw-w64 is capable of creating UWP apps. Which is what we use already.
The advantage of mingw-w64 that it supports cross builds, allowing testing without a Windows env.

Building the tests once is sufficient to ensure the code isn't regressing IMO. Perhaps even that is a waste: There was no sign anybody ever attempted to build tests for UWP, let alone run them, they have been broken till I fixed them a month ago: 998b17ea7f0a4aecaa39c15d0e6f9ece110a41e1.

talregev:
Thank you for your time to read my comments.
As I understand when compile uwp with mingw and vcpkg we are using different compilers: gcc and msvc.
Even if it the same code, different compiler are respond and give different warning / errors. 
I can give you 2 examples: 
1. when I compile tests in msvc one test is not compile due the warning that it give. Even it was the same code that was compile with mingw. 
2. You still compile the same code with gcc and clang compiler, even if the same code, as it can give different warning and errors as I describe above. 

So my conclusion is that compile uwp with msvc it not give you overread  of time testing overall, and also can detected potentials errors when compile it with msvc.

vszakats:
I still can't see the value in this. Judging by the reports and feedback, it doesn't seem there is a whole lot of people interested to compile and run tests (or curl even) under UWP. As I mentioned most of curl didn't compile for UWP for years.

We already compile curl with both mingw and msvc extensively, so what may be interesting here, is the tiny, UWP-specific branches. They are in lib3026 and 3 more files. The patterns used there are also used in lib/tool, that we build already, and some stub code.

This suggests that the existing coverage is sufficient.

talregev:
ok. Thank you for your time for review. I will close this PR.

Issue #14165, opened 2024-07-12
Email payload is broken when set CURLOPT_CRLF = 1


vuonganh1993:
### I did this

My program sends emails by curl with CURLOPT_CRLF = 1. Everything worked fine with 7.76.0. Recently, we upgraded to curl 8.8.0 and the email is always broken if the payload is greater than 64Kb. I found out that the payload is missing some data.
![image](https://github.com/user-attachments/assets/8b67e643-b030-4c30-9020-55279ea28986)

To fix the issue, I need to set CURLOPT_CRLF = 0 and add "\r\n" to the end of every line in the payload. it works fine again.

### curl/libcurl version

curl 8.8.0

### operating system

Windows Enterprise 11

bagder:
please tell us a way how we can reproduce this!

vuonganh1993:
Hi @bagder ,
Attach the sample code and the sample email file.
For testing, I place the email file at "C:\test_email.txt", and then read the content as the input stream.

[test_email.txt](https://github.com/user-attachments/files/16192501/test_email.txt)
[send_email_code.txt](https://github.com/user-attachments/files/16192537/send_email_code.txt)

icing:
I can verify that there is a bug with input larger than 64.

vuonganh1993:
Thank @icing

icing:
I propose #14169 as fix for this, including a modified test case to trigger the bug.

If you can build curl yourself it would be nice if you could verify this. (There is a Windows nightly build out there, I think, but could not find a link. Someone can help out here?)

vszakats:
Daily Windows builds can be found here, scroll down to Artifacts:
https://github.com/curl/curl-for-win/actions/workflows/daily.yml

(don't mind the red, it affects mac builds and pushed a(nother) fix just today.)

PR #14169, opened 2024-07-12
fix crlf conversion of input

icing:
When CRLF line end conversion was enabled (--crlf), input after the last newline in the upload buffer was not sent, if the buffer contained a newline.

- refs #14165


tests/data/test941
```diff
@@ -32,6 +32,657 @@ converted
 with
 the
 --crlf option
+0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
```
bagder, line 35: Maybe instead use a `%repeat[]%` instruction here?

icing: Yeah, I was looking for something like that. Will fix.

icing: That was a little bit more work than expected. Had to enhance the test runner.

bagder (approved):

PR #14182, opened 2024-07-14
cmake: sync CA bundle/path detection with autotools

vszakats:
- skip the entire CA logic if no selected TLS backend support CA
  certs/bundles.
  Follow-up to 082bb41311a832ae1b83bb8fe1dfdefcf4e68ea5 #2545

- sync bundle path detection logic with `./configure`.

- fix to not auto-detect CA bundle/path on Windows.

- fix to reflect that BearSSL has CA bundle support.

- show the detected bundle path (as with the cert bundle).

- tidy up CMake syntax, fix typos in comments.

Closes #14182

---

Easier to parse without whitespace changes:
https://github.com/curl/curl/pull/14182/files?w=1


vszakats:
As per the Linux comparison script autotools _sometimes_ detects both bundle and path.

bagder:
> As per the Linux comparison script autotools _sometimes_ detects both bundle and path.

I don't understand what this means.

vszakats:
There were difference in CA bundle/path auto-detection behaviour between autotools and cmake in macOS tests. autotools was detecting the bundle _file_, while cmake was detecting file and path. But our automated tests reports that autotools detecting both, in that test at least.

I'm making more tests to figure out what I missed.

bagder:
> But our automated tests reports that autotools detecting both, in that test at least.

Right, this is the logic setting it up:

https://github.com/curl/curl/blob/4c22d97be786ed801e050da6872dd3143d6d0a59/acinclude.m4#L1298-L1305

vszakats:
> > But our automated tests reports that autotools detecting both, in that test at least.
> 
> Right, this is the logic setting up the path check:
> 
> https://github.com/curl/curl/blob/4c22d97be786ed801e050da6872dd3143d6d0a59/acinclude.m4#L1298-L1305

I think the difference came down to this line:
```shell
if test -d "$a" && ls "$a"/[[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f]].0 >/dev/null 2>/dev/null; then
```

On macOS the directory exists, but empty. CMake checks its presence, but not its content.

bagder:
> On macOS the directory exists, but empty. CMake checks its presence, but not its content.

Aha, a pretty small but significant difference then!

PR #14183, opened 2024-07-14
tests: sync feature names with `curl -V`

vszakats:
Some feature names used in tests had minor differences compared to
the well-known ones from `curl -V`. This patch syncs them to make test
results easier to grok.

Closes #14183

---

Example:
https://github.com/curl/curl/actions/runs/9927383366/job/27422261107#step:15:3740


bagder (approved):
Works for me!

PR #14185, opened 2024-07-14
runtests: add `-qs` (quiet success) option

vszakats:
Add `-qs` quiet success option. It omits the two log lines for tests
that run successfully. It makes logs ~3500 lines shorter in most cases.

Apply this option to GHA/non-native CI jobs.

Example runs:
- no failure: https://github.com/curl/curl/actions/runs/10021979612
- with failure: https://github.com/curl/curl/actions/runs/10021785141

Closes #14185

### DIFF
--- a/.github/workflows/non-native.yml
+++ b/.github/workflows/non-native.yml
@@ -69,7 +69,7 @@ jobs:
             bld/src/curl --disable --version
             if [ '${{ matrix.arch }}' = 'x86_64' ]; then  # Slow on emulated CPU
               cmake --build bld --config Debug --parallel 3 --target testdeps
-              export TFLAGS='-j0'  # flakies: ~389 ~392 ~TFTP and more
+              export TFLAGS='-j0 -qs'  # flakies: ~389 ~392 ~TFTP and more
               cmake --build bld --config Debug --target test-ci
             fi
 
@@ -104,7 +104,7 @@ jobs:
             bld/src/curl --disable --version
             if [ '${{ matrix.arch }}' = 'x86_64' ]; then  # Slow on emulated CPU
               cmake --build bld --config Debug --parallel 3 --target testdeps
-              export TFLAGS='-j8 ~3017 ~TFTP'  # FIXME: TFTP requests executed twice? Related: `curl: (69) TFTP: Access Violation`?
+              export TFLAGS='-j8 -qs ~3017 ~TFTP'  # FIXME: TFTP requests executed twice? Related: `curl: (69) TFTP: Access Violation`?
               cmake --build bld --config Debug --target test-ci
             fi
 
@@ -152,7 +152,7 @@ jobs:
               make -j3 -C tests
               # The OpenSSH server instance for the testsuite cannot be started on FreeBSD,
               # therefore the SFTP and SCP tests are disabled right away from the beginning.
-              make test-ci V=1 TFLAGS='-j12 !SFTP !SCP'
+              make test-ci V=1 TFLAGS='-j12 -qs !SFTP !SCP'
             fi
 
       - name: 'cmake'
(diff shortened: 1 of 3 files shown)


vszakats:
On sluggish systems (and/or huge logs), this patch makes it quicker to scroll down to the end of the test log (and back to the top of the page), for jobs with tests.
On the other hand it makes the unfolded test details load slower while scrolling (it loads _while_ scrolling, in my browser at least).
Finding a specific test of interest by number continues to work well by using the search field.

If nobody is against, I'd merge this. If it cause issues or we don't like it, we can revert or refine it.

dfandrich:
I'm not a big fan of this. Personally, the test results are usually the only thing I go to the logs to see and having them hidden by default is annoying. They do appear when searching for a specific test in the fold, which is nice, but it takes a while during which time there's no feedback that anything is happening. I'm also not sure what the benefit is of having the test results hidden. They're usually towards the end of a huge compile log, so only people scrolling through those would reach the fold and I'm guessing there aren't many people doing that. Those who want to get to the test summary quickly don't benefit because rather than scrolling they'll just hit the End button to get there immediately. It was interesting to see what GitHub does with them, but I don't think it's worth implementing. With the possible exception of the "Run Time Stats" section which is big and unwieldy (and a section I seldom look at personally :-) and does get in the way between the test results and the summary.

 Also, the "Run Time Stats" section shows up even if there are no run time stats available, which is misleading.

vszakats:
Thanks for your feedback!

For failing jobs, GitHub automatically unfolds the 'run' step and move to its end, and that being slow, it makes navigating the page awkward. E.g. when trying to look into a preceding step. This became an issue when repeating this cycle many-many times a day.

The actual failures are indeed interesting, but I almost always have to search for them and often they are in the middle of the output, intermixed with the successful ones (3-4 thousand lines).

Would it be possible to collapse / hide the successful lines (two lines per test) and only include the failing ones in the output? 

(As slight aside it'd also help (e.g. with diffing) if tests results came out sorted by test number.)

vszakats:
Committed a second take to improve this:

Instead of folding _all_ test run details into a group, I added an option (`-qs`) to not emit any output for tests that finished successfully. It reduces logs without adding the many, less interesting, lines in the first place, keeping all failures there, as before. Since only the failures are logged, it's easier to scroll through them and find the details, without resorting to search in typical scenarios with a couple of failures.

Also kept the fold for the timing results.

It looks like this in action:
clean: https://github.com/curl/curl/actions/runs/10021979612
with failure: https://github.com/curl/curl/actions/runs/10021785141

dfandrich:
While I'm not against such an option, it should never be turned on in CI
builds. It's important to be able to see which tests have actually run so you
can tell what has actually been tested in a platform. Test Clutch also uses
that information to gather test coverage data. There's already so much
scrolling necessary to get to all the different sections of the logs that
eliminating this information information doesn't eliminate much. This is the
same reason we don't run the compiles with V=0, because successful compilation
provides important information to the user as well.

vszakats:
I cannot confirm that scrolling isn't a problem. In my experience it very much is. Taking quite a long time to load, and then the rest. Most jobs have a separate step for running tests (meaning the others don't play a role by default), and it's excessively long, and auto-opened, auto-scrolled to the end, if there is a failure. Finding the actual essence is a problem, in my experience.

I've proposed two solutions, if none of them is acceptable for CI use, can you propose something that can make these logs usable in a browser?

vszakats:
For Test Clutch, can we store the necessary details in a file, and dump it in a separate step, like it's done with `configure.log` in the `macos` and `windows` workflows? Such separate step would be unfoldable manually in the browser, and seen as part of the raw log by Test Clutch.

dfandrich:
I guess I don't know what the problem is you're trying to solve with this PR. I personally haven't seen major issues with loading or scrolling and I don't know what you're seeing. Can you give more details and some example links?

Dumping the test results twice, once in summary and once in full, would confuse Test Clutch as well as many developers, I suspect. If you have to do something, folding seems the least disruptive of the options.

vszakats:
Timing details commits moved to #14284.

PR #14186, opened 2024-07-14
configure: CA bundle/path detection fixes

vszakats:
- fix to not auto-detect CA bundle/path on Windows.

- two checks missed BearSSL, but they were only run for supported
  TLS backends anyway. Delete these redundant checks.

- fix typos in a comment nearby.

Follow-up to 082bb41311a832ae1b83bb8fe1dfdefcf4e68ea5 #2545
Closes #14186

---

Nicer without whitespaces:
https://github.com/curl/curl/pull/14186/files?w=1

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -1259,8 +1259,8 @@ AS_HELP_STRING([--without-ca-path], [Don't use a default CA path]),
     capath="$want_capath"
     ca="no"
   else
-    dnl first try autodetecting a CA bundle , then a CA path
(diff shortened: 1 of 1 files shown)


bagder (approved):

vszakats:
On a second look it seems these checks are not necessary at all since 082bb41311a832ae1b83bb8fe1dfdefcf4e68ea5.
The code is not touched unless a TLS backend supports it.

PR #14188, opened 2024-07-15
GHA/windows: verify 1448 2046 2047 in IDN Unicode jobs

vszakats:
These tests pass with Unicode and fail without.

Follow-up to cb22cfca69bded45bf7f9c72c8e6764990490f11 #14077
Closes #14188

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -654,7 +654,10 @@ jobs:
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP ~FTP ${{ matrix.tflags }}'
           if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
-            TFLAGS+=' ~165 ~1448 ~2046 ~2047'
+            TFLAGS+=' ~165'
+            if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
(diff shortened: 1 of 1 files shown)


vszakats:
Based on these tests from #13947:
https://github.com/curl/curl/actions/runs/9530286814/job/26269989972

1448 2046 2047 fail with IDN + !Unicode. With Unicode they pass.

.github/workflows/windows.yml
```diff
@@ -654,7 +654,10 @@ jobs:
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP ~FTP ${{ matrix.tflags }}'
           if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
-            TFLAGS+=' ~165 ~1448 ~2046 ~2047'
+            TFLAGS+=' ~165'
```
talregev, line 657: I think it's better to remove the if, and put the relevant ignore tests in all places that have unicode in matrix configuration.

MarcelRaad (approved):

.github/workflows/windows.yml
vszakats: For failures with a specific or unique combinations, the per-job exception remains an option. It also works for those with no analysis, just to silence them.

But, once we discover a connection between option and failure, the rule based method is useful to document it and can help fixing the problem.

PR #14189, opened 2024-07-15
getinfo: CURLINFO_POSTTRANSFER_TIME_T

alexsn:
Returns the time, in microseconds, from the start until the last byte is sent by libcurl (i.e. the request is sent off).

Note that I've used `timerid->TIMER_POSTRANSFER` as it was defined but unused.


alexsn:
See discussion [here](https://github.com/curl/curl/discussions/14184), note that I'm missing documentation which I'll add once this PR is accepted.

alexsn:
@bagder are you ok with this change? If so I'll update the docs to reflect this PR.

bagder:
This PR still breaks a lot of tests.

alexsn:
@bagder - Most tests are now passing. I still see a failure in 1541 on 'Linux / AM hyper' where `CURLINFO_POSTTRANSFER_TIME_T` is zero after `curl_easy_perform` is done (though this test runs on all platforms without errors).

Is it possible for `req_set_upload_done ` to not get called?

alexsn:
@bagder -merge window is about to close, could I please get a response on the above?

bagder:
> Is it possible for req_set_upload_done to not get called?

I just tested your PR locally with a hyper build and yes, that's exactly what (not) happens. `req_set_upload_done` is not called... /cc @icing

icing:
> req_set_upload_done

Made #14539 to add the missing call in hyper.

bagder:
@alexsn please rebase on top of master and force-push. Should work now!

icing:
I would exclude test500 from hyper. The time of the sending is a guess in hyper.

alexsn:
> I would exclude test500 from hyper. The time of the sending is a guess in hyper.

You want me to do it in this PR?

icing:
> > I would exclude test500 from hyper. The time of the sending is a guess in hyper.
> 
> You want me to do it in this PR?

That would be best.

bagder:
Thanks!

PR #14195, opened 2024-07-16
curl dns_entry related improvements

icing:
Replace `Curl_resolv_unlock()` with `Curl_resolv_unlink()`:

- replace `inuse` member with `refcount` in Curl_dns_entry
- pass `Curl_dns_entry **` to unlink, so it gets always cleared
- solve potential (but unlikley) UAF in FTP's handling of looked up Curl_dns_entry. Esp. do not use addr information after unlinking an entry.
  In reality, the unlink will not free memory, as the dns entry is still referenced by the hostcache. But this is not safe and relying on no other code pruning the cache in the meantime.
- pass `permanent` flag when adding a dns entry instead of fixing `timestamp` afterwards.

url.c, fold several static `*resolve_*` functions into one.

### DIFF
--- a/lib/hostip.c
+++ b/lib/hostip.c
@@ -115,7 +115,7 @@
  * CURLRES_* defines based on the config*.h and curl_setup.h defines.
  */
 
-static void freednsentry(void *freethis);
+static void hostcache_unlink_entry(void *entry);
 
 #ifndef CURL_DISABLE_VERBOSE_STRINGS
 static void show_resolve_info(struct Curl_easy *data,
@@ -178,7 +178,7 @@ create_hostcache_id(const char *name,
 struct hostcache_prune_data {
   time_t now;
   time_t oldest; /* oldest time in cache not pruned. */
-  int cache_timeout;
+  int max_age_sec;
 };
 
 /*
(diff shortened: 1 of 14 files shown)


lib/hostip.c
```diff
@@ -496,11 +497,15 @@ Curl_cache_addr(struct Curl_easy *data,
...
-  dns->inuse = 1;   /* the cache has the first reference */
+  dns->refcount = 1; /* the cache has the first reference */
   dns->addr = addr; /* this is the address(es) */
-  time(&dns->timestamp);
-  if(dns->timestamp == 0)
-    dns->timestamp = 1;   /* zero indicates permanent CURLOPT_RESOLVE entry */
+  if(permanent)
+    dns->timestamp = 0; /* an entry that never goes stale */
+  else {
+    time(&dns->timestamp);
```
bagder, line 505:
```suggestion
    dns->timestamp = time(NULL);
```

For consistency with how we use it elsewhere, but I also recall that some system somewhere caused us grief otherwise.

icing: Changed.

bagder:
You think this should merge before 8.9.0?

icing:
> You think this should merge before 8.9.0?

I think it can wait.

PR #14200, opened 2024-07-16
GHA/windows: enable libidn2 in !ssl MSVC job

talregev:
Add MSVC idn2 to !ssl job

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -561,12 +561,12 @@ jobs:
           tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
           config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_HTTPSRR=ON -DUSE_ECH=ON'
         - name: '!ssl'
-          install: 'brotli zlib zstd libpsl'
+          install: 'brotli zlib zstd libpsl pkgconf libidn2'
           arch: 'x64'
           plat: 'windows'
           type: 'Debug'
           tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
-          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DHTTP_ONLY=ON'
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -561,12 +561,12 @@ jobs:
...
           config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_HTTPSRR=ON -DUSE_ECH=ON'
         - name: '!ssl'
-          install: 'brotli zlib zstd libpsl'
+          install: 'brotli zlib zstd libpsl pkgconf libidn2'
           arch: 'x64'
           plat: 'windows'
           type: 'Debug'
-          tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
-          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DHTTP_ONLY=ON'
+          tflags: '~1516 ~2301 ~2302 ~2303 ~2307 ~165 ~1448 ~2046 ~2047'
```
vszakats, line 568: What do you think about doing this in the test run step?:
```shell
         if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* || \
                '${{ matrix.config }}' = *'-DUSE_LIBIDN2_IDN=ON'* ]]; then
```

talregev: ok :)

talregev: But the tests ~1448 ~2046 ~2047 are failed with -DENABLE_UNICODE=OFF as well. Then I will need to remove the second if, and add them anyway.

.github/workflows/windows.yml
```diff
@@ -660,11 +660,9 @@ jobs:
...
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP ~FTP ${{ matrix.tflags }}'
-          if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
-            TFLAGS+=' ~165'
-            if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
-              TFLAGS+=' ~1448 ~2046 ~2047'
-            fi
+          if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* || \
+                '${{ matrix.config }}' = *'-DUSE_LIBIDN2=ON'* ]]; then
+            TFLAGS+=' ~165 ~1448 ~2046 ~2047'
```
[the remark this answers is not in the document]
talregev: Where we tested it?
We should add a test for that ...

.github/workflows/windows.yml
vszakats: See my other comment.

libidn2 seems to break IDN differently than WinIDN, in particular 165 fails with this:
```
There was no content at all in the file log/8/server.input.
```

.github/workflows/windows.yml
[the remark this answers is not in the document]
talregev: We should also test that -DUSE_WIN32_IDN=ON and -DENABLE_UNICODE=OFF are failing for regression when the bug will fixed.

vszakats (approved):

vszakats:
Thank you!

PR #14202, opened 2024-07-16
GHA/windows: add MSVC c-ares to openssl job

talregev:
Add MSVC c-ares to openssl job

This job will fail with warning/error:

```
D:\a\curl\curl\lib\asyn-ares.c(332,13): error C2220: the following warning is treated as an error [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
D:\a\curl\curl\lib\asyn-ares.c(332,13): warning C4996: 'ares_getsock': Use ARES_OPT_EVENT_THREAD or ARES_OPT_SOCK_STATE_CB instead [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
```

---
edit:

ignore the warning:
- warning C4996

ignore tests:
- FAIL 558: 'libtest memory tracking operational' TrackMemory
- FAIL 1330: 'unit tests memory tracking operational' unittest, TrackMemory


talregev:
@vszakats I add an ignore for the warning. 
Will curl accept this solution?

vszakats:
Only as a last resort, e.g. for compiler bugs or things out of our control.

Have you verified that this one doesn't have a better fix?

talregev:
> Only as a last resort, e.g. for compiler bugs or things out of our control.
> 
> Have you verified that this one doesn't have a better fix?

As I understand curl, you already use call back in cares.
Maybe this is an old code that maintain?

vszakats:
Yes, the warning suggests to use a symbol to replace a previous one, that's possibly deprecated? (I haven't checked closely)

talregev:
> Yes, the warning suggests to use a symbol to replace a previous one, that's possibly deprecated? (I haven't checked closely)

It deprecated

vszakats:
In that case, using the new macro name for c-ares versions supporting it is the preferred way to fix it.

talregev:
> In that case, using the new macro name for c-ares versions supporting it is the preferred way to fix it.

I know that this is cares is deprecated. But I don't know if it really code for old version or is still production code.

vszakats:
I'm not sure I understand, are you asking if curl actually needs that code with the deprecated c-ares, or if it's just left there by accident?

talregev:
> I'm not sure I understand, are you asking if curl actually needs that code with the deprecated c-ares, or if it's just left there by accident?

Yes.

vszakats:
All curl code in the repo is production / active. As dependencies evolve, curl needs to adapt to them, like in this case. The practice is to add the new way of doing things and use it when building for c-ares versions that support it, and limit the old code for the older versions.

talregev:
I am not sure how to remove / replace the deprecated code.

vszakats:
I can't offer a "magic" solution. My experience is that the burden of fixing things falls onto the person adding new tests. This can be challenging and/or time-consuming.

If you you need help fixing it or want to defer it to sombody else, perhaps opening a Issue is the next best course of action.

It may also be useful to reasearch the history of the deprecation in the c-ares project. There is often useful information to find. Once you know what to look for, perhaps the issue has already been fixed in another project, that you can re-use in the curl project.

PR #14204, opened 2024-07-16
multi: fix timing metrics on connect failure

jmaggard10:
On platforms without AsynchDNS, if name resolution fails for a transaction, CURLINFO_TOTAL_TIME (and all other timing metrics) return 0, which is rather misleading.

This commit adds a final Curl_pgrsUpdate() call before returning to update the timing metrics.

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -2643,6 +2643,7 @@ statemachine_end:
         else if(data->mstate == MSTATE_CONNECT) {
           /* Curl_connect() failed */
           (void)Curl_posttransfer(data);
+          (void)Curl_pgrsUpdate(data);
         }
 
         multistate(data, MSTATE_COMPLETED);
--- a/tests/data/test1268
+++ b/tests/data/test1268
@@ -23,7 +23,7 @@ UnixSockets
 filename argument looks like a flag
 </name>
 <command>
---stderr %LOGDIR/moo%TESTNUMBER --unix-socket -k hej://moo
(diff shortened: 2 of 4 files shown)


bagder (approved):

dfandrich:
Analysis of PR #14204:

[Test 1268 failed](https://github.com/curl/curl/actions/runs/9965217539),
which has NOT been flaky recently, so **there could be a real issue in the PR**.
Note that this test has failed in 70 different CI jobs (the link just goes to one of them).

[Test 1471 failed](https://github.com/curl/curl/actions/runs/9965217539),
which has NOT been flaky recently, so **there could be a real issue in the PR**.
Note that this test has failed in 70 different CI jobs (the link just goes to one of them).

[Test 1472 failed](https://github.com/curl/curl/actions/runs/9965217539),
which has NOT been flaky recently, so **there could be a real issue in the PR**.
Note that this test has failed in 70 different CI jobs (the link just goes to one of them).

dfandrich:
Several tests aren't expecting to see progress output in their stderr, so those should be updated to add --no-progress-meter or similar.

bagder (changes requested):
Test fails need attention

jmaggard10:
Tests fixed and passing now.

bagder (approved):

bagder:
Hm, I'm having second thoughts about this. It might upset a few people as it now (uslessly) outputs that progress bar in failure cases is never did before...

bagder:
I propose #14239 instead. It should get the same effect but without the progress callback/meter update.

jmaggard10:
Sounds good to me, thanks.

PR #14208, opened 2024-07-17
scorecard upload tests

icing:
- add upload tests to scorecard, invoke with
  > python3 tests/http/scorecard.py -u h1|h2|h3
- add a reverse proxy setup from Caddy to httpd for upload tests since Caddy does not have other PUT/POST handling
- add caddy tests in test_08 for POST/PUT
- increase read buffer in mod_curltest for larger reads


icing:
@vszakats care to merge this? I can use this for measuring other PRs more easily then.

bagder:
I'll do it!

PR #14209, opened 2024-07-17
http/3: resume upload on ack if we have more data to send

### LINKED ISSUE #14198: HTTP/3 file uploads is significantly slower comparing to HTTP/2 or an in house HTTP/3 client
### I did this

I'm settings to a http3 transfer by using CURLOPT_HTTP_VERSION and capturing request stats, as can be seen below the H3 request send takes a lot longer to complete.

### I expected the following

I expect the file upload using H3 to be at least as fast as H2, currently it's about 3 - 4 times slower:
### Request stats
```
// Request stats using H2, notice the delta between last_byte_flushed_ms and first_byte_flushed_ms which is 314ms
{
  "dns_resolution_time_ms": 4,
  "request_queue_time_ms": 0,
  "request_header_size": 700,
  "time_to_last_byte_ms": 2718,
  "time_to_first_byte_ms": 2718,
  "total_request_time_ms": 2718,
  "status_code": 200,
  "request_body_size": 1022505,
  "last_byte_flushed_ms": 497,
  "first_byte_flushed_ms": 183,
  "connect_time_ms": 58,
  "response_body_size": 159,
  "http_method": "POST",
  "new_connection": true,
  "server_address": "2a03:2880:f042:112:face:b00c:0:2",
  "handshake_time_ms": 120,
  "response_header_size": 145,
  "http_version": "2.0"
}

// Same request using H3, notice the time it takes to send the request is 1028ms
{
  "dns_resolution_time_ms": 4,
  "request_queue_time_ms": 0,
  "request_header_size": 697,
  "time_to_last_byte_ms": 3102,
  "time_to_first_byte_ms": 3027,
  "total_request_time_ms": 3102,
  "status_code": 200,
  "request_body_size": 1022505,
  "last_byte_flushed_ms": 1112,
  "first_byte_flushed_ms": 84,
  "connect_time_ms": 17,
  "response_body_size": 159,
  "host_name": "rupload.facebook.com",
  "http_method": "POST",
  "new_connection": true,
  "server_address": "2a03:2880:f042:112:face:b00c:0:2",
  "handshake_time_ms": 60,
  "response_header_size": 162,
  "http_version": "3.0"
}
```

### Analysis:
I've taken packet capture from libcurl transfer and it looks like libcurl sends 64k of data then blocks until it all frames are acked which suspends the upload for about 40ms-50ms on every 64k chunk
Screenshot 2024-07-16 at 16 32 53

When comparing this to a in house HTTP/3 client there are no hangs and the upload continues smoothly throughout the transfer
Screenshot 2024-07-16 at 16 36 25

Checking [curl_ngtcp2.c](https://github.com/curl/curl/blob/master/lib/vquic/curl_ngtcp2.c) it looks like there's a buffering issue where the obvious suspect is `sendbuf_len_in_flight` which is where the 64k part comes from and there are multiple locations waiting for this value to drop to zero (like [here](https://github.com/curl/curl/blob/master/lib/vquic/curl_ngtcp2.c#L1276) and [here](https://github.com/curl/curl/blob/master/lib/vquic/curl_ngtcp2.c#L1589)) before resuming.

I've tried setting CURLOPT_UPLOAD_BUFFERSIZE to a larger value (512kb) which does help (makes it about 2x faster) though it's still about 2x slower then H2.

### curl/libcurl version

libcurl 8.8.0 with wolfssl / ngtcp2 / nghttp3 and nghttp2.

### operating system

Linux <hostname> 5.12.0-0 #1 SMP Mon Jun 26 17:52:47 PDT 2023 x86_64 x86_64 x86_64 GNU/Linux

alexsn:
Currently we're waiting for sendbuf_len_in_flight to hit zero before resuming upload which means we're blocking and waiting for _all_ acks to arrive before sending more data. This causes significant delays especially when ack delay is used on the server side.

The fix addresses several issues in h3 over ngtcp2:
  - On ack we now call nghttp3_conn_resume_stream() when we have more data to send
  - upload_left was incorrectly computed on CF_CTRL_DATA_DONE_SEND as we need to subtract the ammount of data we have in flight.
  - Remove upload_blocked_len as we Curl_bufq_write call will do the right thing when called from cf_ngtcp2_send.

Closes #14198

### DIFF
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -162,7 +162,6 @@ struct h3_stream_ctx {
   struct bufq sendbuf;   /* h3 request body */
   struct h1_req_parser h1; /* h1 request parsing */
   size_t sendbuf_len_in_flight; /* sendbuf amount "in flight" */
-  size_t upload_blocked_len; /* the amount written last and EGAINed */
   curl_uint64_t error3; /* HTTP/3 stream error code */
   curl_off_t upload_left; /* number of request bytes left to upload */
   int status_code; /* HTTP status code */
@@ -1272,8 +1271,8 @@ static int cb_h3_acked_req_body(nghttp3_conn *conn, int64_t stream_id,
   Curl_bufq_skip(&stream->sendbuf, skiplen);
   stream->sendbuf_len_in_flight -= skiplen;
 
-  /* Everything ACKed, we resume upload processing */
-  if(!stream->sendbuf_len_in_flight) {
+  /* Resume upload processing if we have more data to send */
+  if(stream->sendbuf_len_in_flight < Curl_bufq_len(&stream->sendbuf)) {
     int rv = nghttp3_conn_resume_stream(conn, stream_id);
     if(rv && rv != NGHTTP3_ERR_STREAM_NOT_FOUND) {
       return NGTCP2_ERR_CALLBACK_FAILURE;
@@ -1528,21 +1527,6 @@ static ssize_t cf_ngtcp2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
     sent = -1;
     goto out;
   }
-  else if(stream->upload_blocked_len) {
-    /* the data in `buf` has already been submitted or added to the
-     * buffers, but have been EAGAINed on the last invocation. */
-    DEBUGASSERT(len >= stream->upload_blocked_len);
-    if(len < stream->upload_blocked_len) {
-      /* Did we get called again with a smaller `len`? This should not
-       * happen. We are not prepared to handle that. */
-      failf(data, "HTTP/3 send again with decreased length");
-      *err = CURLE_HTTP3;
-      sent = -1;
-      goto out;
-    }
-    sent = (ssize_t)stream->upload_blocked_len;
-    stream->upload_blocked_len = 0;
-  }
   else if(stream->closed) {
     if(stream->resp_hds_complete) {
       /* Server decided to close the stream after having sent us a final
@@ -1583,18 +1567,6 @@ static ssize_t cf_ngtcp2_send(struct Curl_cfilter *cf, struct Curl_easy *data,
   result = cf_progress_egress(cf, data, &pktx);
   if(result) {
     *err = result;
-    sent = -1;
-  }
-
-  if(stream && sent > 0 && stream->sendbuf_len_in_flight) {
-    /* We have unacknowledged DATA and cannot report success to our
-     * caller. Instead we EAGAIN and remember how much we have already
-     * "written" into our various internal connection buffers. */
-    stream->upload_blocked_len = sent;
-    CURL_TRC_CF(data, cf, "[%" CURL_PRId64 "] cf_send(len=%zu), "
-                "%zu bytes in flight -> EGAIN", stream->id, len,
-                stream->sendbuf_len_in_flight);
-    *err = CURLE_AGAIN;
     sent = -1;
   }
 
@@ -1965,7 +1937,8 @@ static CURLcode cf_ngtcp2_data_event(struct Curl_cfilter *cf,
     struct h3_stream_ctx *stream = H3_STREAM_CTX(ctx, data);
     if(stream && !stream->send_closed) {
       stream->send_closed = TRUE;
-      stream->upload_left = Curl_bufq_len(&stream->sendbuf);
+      stream->upload_left = Curl_bufq_len(&stream->sendbuf) -
+        stream->sendbuf_len_in_flight;
       (void)nghttp3_conn_resume_stream(ctx->h3conn, stream->id);
     }
     break;
--- a/lib/vquic/curl_osslq.c
+++ b/lib/vquic/curl_osslq.c
@@ -560,7 +560,6 @@ struct h3_stream_ctx {
   struct bufq recvbuf;   /* h3 response body */
   struct h1_req_parser h1; /* h1 request parsing */
   size_t sendbuf_len_in_flight; /* sendbuf amount "in flight" */
-  size_t upload_blocked_len; /* the amount written last and EGAINed */
   size_t recv_buf_nonflow; /* buffered bytes, not counting for flow control */
   curl_uint64_t error3; /* HTTP/3 stream error code */
   curl_off_t upload_left; /* number of request bytes left to upload */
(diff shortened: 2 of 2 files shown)


icing:
Thanks for working on this. I'll have a look tomorrow at this. This needs some careful inspection, because the problem with  this "blocked" behaviour is the sending of the last chunk of upload data. If that reports success before everything is ACKed, the curl transfer handling will think all is done and stop sending.

alexsn:
> Thanks for working on this. I'll have a look tomorrow at this. This needs some careful inspection, because the problem with this "blocked" behaviour is the sending of the last chunk of upload data. If that reports success before everything is ACKed, the curl transfer handling will think all is done and stop sending.

We don't report success / failure here, we report how much data was queued for xfer. When upload stops we'll receive `CF_CTRL_DATA_DONE_SEND` in which case `-> upload_left ` will be updated to reflect how much data we have queued but not in flight. Does this make sense or did I miss something?

icing:
This change exposes the following behaviour. When uploading a large file, it will produce:

```
...
[0-0] * [HTTP/3] [0] cf_send(len=146) -> 146, 0
[0-0] } [146 bytes data]
[0-0] * upload completely sent off: 10485760 bytes
[0-0] * [HTTP/3] [0] unblock quic flow
[0-0] * [HTTP/3] recvd 1 packets with 44 bytes -> 0
[0-0] * [HTTP/3] [0] read req body -> 5 vecs EOF with 65682 (buffered=80115, left=0)
[0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
[0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
[0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
[0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
[0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
[0-0] * [HTTP/3] vquic_send(len=10426, gso=1444) -> 0, sent=10426
[0-0] * [HTTP/3] [0] cf_recv(blen=102400) -> -1, 81
[0-0] * [HTTP/3] data idle
...
```

curl reports the upload to be complete while there are still 64KB of unsent data in the buffers (plus 15KB unacked). The actual sending then happens later when it wants to read the response. In general, we'd like to report the "completely sent off" when everything was actually sent.

Now, QUIC is different. For other HTTP versions, data may linger in the TCP kernel buffer, so claiming it has been sent is not totally accurate either. But I think it would be better to `CURL_EAGAIN` sends for data that has not been placed on the network.

icing:
This PR definitely goes the right direction for performance. So my critic here is more along the lines of making it even better! Using #14208, I see the following:

```
> python3 tests/http/scorecard.py -u --caddy h3
Uploads
Branch      Size  single(1x1) [cpu/rss]       serial(50x1) [cpu/rss]       parallel(50x50) [cpu/rss]
master       1MB       6 MB/s [94.8%/8MB]          10 MB/s [99.5%/10MB]        139 MB/s [98.2%/15MB]
master      10MB       9 MB/s [99.9%/8MB]          12 MB/s [99.9%/10MB]        189 MB/s [99.9%/17MB]
master     100MB      14 MB/s [100.0%/8MB]         14 MB/s [99.9%/10MB]        182 MB/s [99.9%/19MB]

pr14209      1MB      16 MB/s [67.7%/5MB]         129 MB/s [92.2%/10MB]        146 MB/s [98.6%/14MB]
pr14209     10MB      91 MB/s [85.9%/7MB]         185 MB/s [98.9%/11MB]        191 MB/s [99.9%/17MB]
pr14209    100MB     176 MB/s [99.5%/9MB]         187 MB/s [99.8%/11MB]        188 MB/s [99.8%/19MB]

```
Which is great!

alexsn:
> This change exposes the following behaviour. When uploading a large file, it will produce:
> 
> ```
> ...
> [0-0] * [HTTP/3] [0] cf_send(len=146) -> 146, 0
> [0-0] } [146 bytes data]
> [0-0] * upload completely sent off: 10485760 bytes
> [0-0] * [HTTP/3] [0] unblock quic flow
> [0-0] * [HTTP/3] recvd 1 packets with 44 bytes -> 0
> [0-0] * [HTTP/3] [0] read req body -> 5 vecs EOF with 65682 (buffered=80115, left=0)
> [0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
> [0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
> [0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
> [0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
> [0-0] * [HTTP/3] vquic_send(len=14440, gso=1444) -> 0, sent=14440
> [0-0] * [HTTP/3] vquic_send(len=10426, gso=1444) -> 0, sent=10426
> [0-0] * [HTTP/3] [0] cf_recv(blen=102400) -> -1, 81
> [0-0] * [HTTP/3] data idle
> ...
> ```
> 
> curl reports the upload to be complete while there are still 64KB of unsent data in the buffers (plus 15KB unacked). The actual sending then happens later when it wants to read the response. In general, we'd like to report the "completely sent off" when everything was actually sent.
> 
> Now, QUIC is different. For other HTTP versions, data may linger in the TCP kernel buffer, so claiming it has been sent is not totally accurate either. But I think it would be better to `CURL_EAGAIN` sends for data that has not been placed on the network.

My thought here is once the data reached `stream->sendbuf` is the same as data reaching kernel sockbuf. This data will be handed for xfer once cb_h3_read_req_body is called (we unconditionally call nghttp3_conn_resume_stream if Curl_bufq_write wrote something to sendbuf in cf_ngtcp2_send) so I don't think there's any dependency on response reading.

alexsn:
Also data placed on the network is also not exactly right either as we can do retransmissions so we can say we're done sending once everything is acked. Now this is fine _but_ I think we can start receiving a response before that (say the server uses ack delay) which is weird (started receiving the response before we're done sending the request).

The question here is what is the expectation of `Curl_cft_send`, does success mean data is queued (which is what we get from kernel sockets and with this PR) or something else? The blocking thing is us trying to hack an async call (queued / completed) into a sync function.

icing:
After a night's sleep, I think this PR is fine. 

I plan to add a `EOS` flag (end of stream) to the connection filters `do_send()` method. This will help protocol filters like HTTP/2 and 3 to send the stream's EOF flag at the right time and they may also EAGAIN such sends if their buffers are not fully flushed.

icing (approved):

alexsn:
Just noticed `curl_osslq.c` will have the same issue, should I update this PR with the same changes applied there?

icing:
> Just noticed `curl_osslq.c` will have the same issue, should I update this PR with the same changes applied there?

Please, go ahead!

alexsn:
> > Just noticed `curl_osslq.c` will have the same issue, should I update this PR with the same changes applied there?
> 
> Please, go ahead!

Done

bagder:
Thanks!

Issue #13288, opened 2024-04-04
websocket.c example: curl_ws_recv returns "Socket not ready for send/recv" (errno 81) 


untitaker:
### I did this

1. copied [this example](https://github.com/curl/curl/blob/2cac6a49b3d8b57d77f08f94b023ced211c70d55/docs/examples/websocket.c) into `foo.c`
2. edited the websocket URL in `foo.c` to be a valid websocket server. `wss://ws.postman-echo.com/raw` works, `ws://localhost:8080` works as well, with [websocat](https://github.com/vi/websocat) being invoked as `websocat -s 8080`
3. cloned curl master, moved `foo.c` into root folder
4. `cmake -DENABLE_WEBSOCKETS=on`
5. `make`
6. `cc -L lib/ foo.c -lcurl`
7. `LD_LIBRARY_PATH=lib/ ./a.out`

### I expected the following

some kind of ping-pong demo, or a hanging connection, depending on which websocket server is being used.

instead the output is:

```
ws: curl_ws_recv returned 81, received 0
```

in the real-world application that is built based on that example code, I get (just posting this for searchability):

```
code 81: "Socket not ready for send/recv"
```

### curl/libcurl version

git sha 376cd67035ed2b3cc1504d9ce8e440b1c32099a2

I thought this was related to #11443, so I checked curl 7 as well. On tag `curl-7_88_1` (046209e56), `foo.c` hangs instead, though `websocket.c` does not exist on that commit at all anyway

### operating system

Ubuntu 23.10

untitaker:
Dug around a little, and I suspect that this is intended behavior due to https://github.com/curl/curl/pull/10625, and it's just the example that needs updating, similar to https://github.com/curl/curl/issues/10760

MonkeybreadSoftware:
Shouldn't you simply update the example to do `curl_ws_recv` again when you get `CURLE_AGAIN`. Better with a short pause to avoid the CPU running on 100%.

```
	CURLcode result = curl_ws_recv(curl, buffer, sizeof(buffer), &rlen, &meta);
	while (result == CURLE_AGAIN)
	{
		Curl_wait_ms(1);
		result = curl_ws_recv(curl, buffer, sizeof(buffer), &rlen, &meta);
	}

```

untitaker:
it would be an improvement but there should be an actual select or callback-based API to figure out when to poll again. busy-polling as you do it will make the example correct but not fit for real-world use

MonkeybreadSoftware:
[...]

PR #14214, opened 2024-07-17
docs/examples/websocket: fix "socket not ready" on recv

jay:
- Overhaul the recv_any function to block waiting for a complete websocket message from the server.

- Have recv_pong and recv_header call recv_any to retrieve a complete message, rather than duplicate the efforts in each function.

- Fix sleep() macro for Windows to treat the parameter as seconds instead of milliseconds.

- Improve documentation of curl_ws_recv/send to better explain CURLE_AGAIN behavior.

- Use echo.websocket.org instead of example.com as the example websocket server since it is a server the example actually works on.

recv_any now calls curl_ws_recv repeatedly, waiting until the socket is readable and a full websocket message is read from the server.

Prior to this change the example did not handle the non-blocking behavior of curl_ws_recv. When curl_ws_recv returned CURLE_AGAIN the example would fail with "socket not ready".

Fixes https://github.com/curl/curl/issues/13288
Closes #xxxx

---

This is the second take. The first take is in [jay:curl:fix_websocket_example__alternate_take](https://github.com/curl/curl/compare/master...jay:curl:fix_websocket_example__alternate_take). The first take is more thorough because I wrote a blocking function that can return incomplete non-control messages when they are disrupted by a control message. However after I wrote it I thought it was too complicated for the example.

This only fixes the example for recv. Neither take covers partial sends. It's unclear to me what to do in that case. For example let's say curl_ws_send is called to send a PING with a payload of 6 bytes, but when it returns only 5 of the 6 bytes of payload were sent. What is the caller supposed to do in that case, has the PING been sent with just the 5 bytes or is it waiting to send the remaining byte and curl_ws_send has to be called again with offset?

### DIFF
--- a/docs/examples/websocket.c
+++ b/docs/examples/websocket.c
@@ -25,16 +25,67 @@
  * WebSocket using CONNECT_ONLY
  * </DESC>
  */
+#ifdef _WIN32
+#include <winsock2.h>
+#include <windows.h>
+#endif
+
 #include <stdio.h>
 #include <string.h>
+
 #ifdef _WIN32
-#include <windows.h>
-#define sleep(s) Sleep((DWORD)(s))
+#define sleep(s) Sleep((DWORD)((s)*1000))
 #else
 #include <unistd.h>
 #endif
 
 #include <curl/curl.h>
+
+/* Auxiliary function that waits on the socket.
+   If 'for_recv' is true it waits until data is received, if false it waits
+   until data can be sent on the socket.
+   If 'timeout_ms' is negative then it waits indefinitely.
+   */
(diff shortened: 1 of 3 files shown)


jay:
I've taken this example as far as I can or want to. I feel like it's _too_ thorough. OTOH I think working with raw websockets (the connect-only option) is error prone and I'm not sure that we should have an example for it or otherwise encourage it. The main issue is the user needs to manage the receives and possibly? piece them together, instead of using normal websocket behavior /writefunction. I'm sure there is room for improvement here and I don't want to write this off but I struggled to produce an example I thought was correct and could handle mixed control messages. Maybe, in any case, they should be handled through a user callback. I'm closing this but I'm going to leave the branch up in case anyone needs it for reference.

PR #14217, opened 2024-07-17
curldown: make 'added-in:' a mandatory header field

bagder:
- generate AVAILABILITY manpage sections automatically - for consistent wording

- allows us to double-check against other documumentation (symbols-in-versions etc)

- enables proper automation/scripting based on this data

- lots of them were wrong or missing in the manpages

- several of them repeated (sometimes mismatching) backend support info

Add test 1488 to verify "added-in" version numbers against symbols-in-versions.


bagder:
An added bonus is also that we can extract the `added-in` information easier from the source data, like for generating a table on the website.

PR #14218, opened 2024-07-18
ssl: Fix compilation with OpenSSL 1.x with md4 disabled

orgads:
If OpenSSL 1.x is used, and it is configured with md4 disabled, OPENSSL_NO_MD4 is defined in opensslconf.h, but this header was not included before checking for this define.

Later in md4.c, openssl/md4.h is included, and it includes that header indirectly, leading to inconsistency within md4.c.

Since the md4.h branch was taken, wincrypt.h (or others) is not included, and later below the USE_WIN32_CRYPTO branch is taken, but the types are not defined.

### DIFF
--- a/lib/md4.c
+++ b/lib/md4.c
@@ -37,6 +37,9 @@
 #if (OPENSSL_VERSION_NUMBER >= 0x30000000L) && !defined(USE_AMISSL)
 /* OpenSSL 3.0.0 marks the MD4 functions as deprecated */
 #define OPENSSL_NO_MD4
+#else
+/* Cover also OPENSSL_NO_MD4 configured in openssl */
+#include <openssl/opensslconf.h>
 #endif
 #endif /* USE_OPENSSL */
 


orgads:
Is the failing test known to be flaky? What exactly failed there?

vszakats:
Yes, _all_ native tests are flaky.

orgads:
> Yes, _all_ native tests are flaky.

So should I rerun it? How?

bagder (approved):

MarcelRaad:
> So should I rerun it? How?

You can click on the job that failed, and then on "re-run failed jobs" in the menu at the top. I just did that, and enabled debug logging as there was no output at all for the tests, only the failure mark.

MarcelRaad (approved):
Nit: this is done in an `#else` branch for other backends.

bagder (changes requested):
Ah yes, using an `#else` there as mentioned by @MarcelRaad makes sense.

orgads:
Done

bagder (approved):

bagder:
Thanks!

PR #14220, opened 2024-07-18
send eos flag

icing:
Adds a `bool eos` flag to send methods to indicate that the data is the last chunk the invovled transfer wants to send to the server.

This will help protocol filters like HTTP/2 and 3 to forward the stream's EOF flag and also allow to EAGAIN such calls when buffers are not yet fully flushed.


bagder:
too eager (did not spot the feature window tag), reverted and opening this again. It caused a problem with wolfssh builds: https://github.com/curl/curl/commit/be93299f10ef0b2bf7fe5c82140120073831867a

PR #14232, opened 2024-07-19
GHA/windows: enable libssh in !ssl MSVC job

talregev:
GHA/windows: add MSVC libssh to openssl job

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -530,12 +530,16 @@ jobs:
           tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
           config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_HTTPSRR=ON -DUSE_ECH=ON'
         - name: '!ssl'
-          install: 'brotli zlib zstd libpsl pkgconf libidn2'
+          install: 'brotli zlib zstd libpsl pkgconf libidn2 libssh'
           arch: 'x64'
           plat: 'windows'
           type: 'Debug'
           tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
-          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DHTTP_ONLY=ON -DUSE_LIBIDN2=ON'
+          # WARNING: libssh uses hard-coded world-writable paths (/etc/..., ~/.ssh/) to
+          #          read its configuration from, making it vulnerable to attacks on
+          #          Windows. Do not use this component till there is a fix for these.
+          # https://github.com/curl/curl-for-win/blob/3951808deb04df9489ee17430f236ed54436f81a/libssh.sh#L6-L8
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DHTTP_ONLY=ON -DUSE_LIBIDN2=ON -DCURL_USE_LIBSSH2=OFF -DCURL_USE_LIBSSH=ON'
         - name: 'wolfssl'
           install: 'brotli zlib zstd libpsl nghttp2 wolfssl libssh2 pkgconf gsasl'
           arch: 'x64'


vszakats:
Can we add libssh to an existing job?

talregev:
> Can we add libssh to an existing job?

Yes. which one you suggest?

vszakats:
What about the !ssl one? Or openssl uwp if libssh supports it.

Notice however that libssh has an pretty huge vulnerability on Windows, so this is a dependency nobody should use in a serious or any setting. Someone with a Gitlab account shall probably report it.

talregev:
> What about the !ssl one? Or openssl uwp if libssh supports it.
> 
> Notice however that libssh has an pretty huge vulnerability on Windows, so this is a dependency nobody should use in a serious or any setting. Someone with a Gitlab account shall probably report it: https://github.com/curl/curl-for-win/blob/3951808deb04df9489ee17430f236ed54436f81a/libssh.sh#L6-L8 (the issue was there last time I checked a few weeks ago)

So maybe we should wait for it? 
I just added it to activate more curl tests ...

vszakats:
In the `!ssl` job it would? But, SSH tests are not running on Windows at all, so it remains a build-only test:
```
TESTINFO: "failed starting SSH server" 51 times (582, 583, 600, 601, 602, 603, 604, 605, 606 and 42 more)
```

Building it still useful, but perhaps with a comment telling about this caveat?

/off
This is seems to be the problem (couldn't figure out how to make permalinks with GitLab, sorry):
https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/config.c#L588 `ssh_config_make_absolute()`
is loading the config from a hard-coded `/etc/ssh/`, `~/.ssh` and other unixy things. On Windows this is `C:/etc/ssh`, which is world-writable / unprotected.
Other unixy parts:
https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/misc.c#L1175 `ssh_path_expand_escape()`
https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/options.c#L1815 `ssh_options_parse_config()`
This latter function is called from `ssh_connect()` with a default (`NULL`) filename on all platforms.
A `GLOBAL_CLIENT_CONFIG` settings is configurable via CMake, and defaults to `/etc/ssh/ssh_config` on all platforms. On Windows this is effectively `<current drive>:/etc/ssh/ssh_config`, world-writable. This also cannot be configured securely, because there is no universally secure path on Windows.

talregev:
where we should add this comment?

.github/workflows/windows.yml
```diff
@@ -530,12 +530,12 @@ jobs:
...
           config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DUSE_HTTPSRR=ON -DUSE_ECH=ON'
         - name: '!ssl'
-          install: 'brotli zlib zstd libpsl'
+          install: 'brotli zlib zstd libpsl libssh'
           arch: 'x64'
           plat: 'windows'
           type: 'Debug'
           tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
-          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DHTTP_ONLY=ON'
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DHTTP_ONLY=ON -DCURL_USE_LIBSSH2=OFF -DCURL_USE_LIBSSH=ON'
```
vszakats, line 538: Before the `config:` line perhaps?

talregev: Done.

talregev:
> In the `!ssl` job it would? But, SSH tests are not running on Windows at all, so it remains a build-only test:
> 
> ```
> TESTINFO: "failed starting SSH server" 51 times (582, 583, 600, 601, 602, 603, 604, 605, 606 and 42 more)
> ```
> 
> Building it is still useful, but perhaps with a comment telling about this caveat?
> 
> /off This is seems to be the problem (couldn't figure out how to make permalinks with GitLab, sorry): https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/config.c#L588 `ssh_config_make_absolute()` is loading the config from a hard-coded `/etc/ssh/`, `~/.ssh`, and other unixy things. On Windows this is `C:/etc/ssh`, which is world-writable / unprotected. Other unixy parts: https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/misc.c#L1175 `ssh_path_expand_escape()` https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/options.c#L1815 `ssh_options_parse_config()` This latter function is called from `ssh_connect()` with a default (`NULL`) filename on all platforms. A `GLOBAL_CLIENT_CONFIG` setting is configurable via CMake, and defaults to `/etc/ssh/ssh_config` on all platforms. On Windows this is effectively `<current drive>:/etc/ssh/ssh_config`, world-writable. This also cannot be configured securely, because there is no universally secure path on Windows. It's almost the same issue as this two-decade old one with OpenSSL: [openssl/openssl#24528](https://github.com/openssl/openssl/issues/24528), but contrary to OpenSSL, libssh doesn't want to load DLLs via these configs.

How it done on libssh2?
Why not to read the config from .ssh in the user setting folder in windows?
`cd %USERPROFILE%\.ssh`

vszakats:
libssh2 doesn't have an on-disk configuration. It also only reads filenames passed to it explicitly. The same class of issues can only affect the caller. In curl's case `lib/vssh/libssh2.c`. That isn't perfect (ref #13571, and lack of `USERPROFILE`) either, but it's not a libssh2 issue.

vszakats:
>Why not to read the config from .ssh in the user setting folder in windows? `cd %USERPROFILE%\.ssh` or `cd Program Files\etc\ssh` Program files in windows is protected folder, access to every user by reading. Can access from environment variable: `cd %ProgramFiles%`

If `%USERPROFILE%` and `%ProgramFiles%` are detected at runtime, it's indeed one possible good fix. Another option is to offer an API to set these paths. Or both.

OpenSSL for example has been hard-coding `/etc/ssl` for a long time, then (after CVE-2019-5443) changed this to a _built-time_ hard-coded `%ProgramFiles%` (that is still wrong), with a buggy override and no way to disable. Lately added an option to disable some (all?) of the functionality depending on hard-coded paths. And is now shifting to a convoluted registry-based setting which requires an installer be run by an admin, with the existing bad solution still in place, for "compatibility". It's puzzling.

vszakats (approved):

talregev:
%USERPROFILE% and %ProgramFiles% are detected at runtime.

vszakats:
Can you rebase on master?

talregev:
> Can you rebase on master?

Done.

vszakats:
Thanks @talregev!

PR #14236, opened 2024-07-19
GHA/windows: timeout earlier with hung tests

vszakats:
Finishing tests takes on average 10 or less minutes depending on
platform. Reduce job step timeouts to reflect that. It helps
concluding hung/failed tests earlier, which allows to retry them
earlier.

This makes it more difficult to tell from a job if it hung or not,
because we lose the long runtime as a telltale sign. Let's see how it
works out and adjust as necessary.

Also fix a comment while here.

Closes #14236

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -173,7 +173,7 @@ jobs:
 
       - name: 'cmake run tests'
         if: ${{ matrix.build == 'cmake' && matrix.tflags != 'skipall' && matrix.tflags != 'skiprun' }}
-        timeout-minutes: 40
+        timeout-minutes: 15
         run: |
           export TFLAGS='-j8 ${{ matrix.tflags }}'
           if [ -x "$(cygpath "${SYSTEMROOT}/System32/curl.exe")" ]; then
@@ -355,7 +355,7 @@ jobs:
 
       - name: 'cmake run tests'
         if: ${{ matrix.build == 'cmake' && matrix.tflags != 'skipall' && matrix.tflags != 'skiprun' }}
-        timeout-minutes: 40
+        timeout-minutes: 10
         run: |
           export TFLAGS='-j14 ${{ matrix.tflags }}'
           if [ '${{ matrix.sys }}' != 'msys' ]; then
@@ -471,7 +471,7 @@ jobs:
 
       - name: 'cmake run tests'
         if: ${{ matrix.tflags != 'skipall' && matrix.tflags != 'skiprun' }}
-        timeout-minutes: 40
+        timeout-minutes: 10
         run: |
           PATH="$(cygpath "${USERPROFILE}")/my-cache/${{ matrix.dir }}/bin:/c/msys64/usr/bin:$PATH"
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets ~FTP ${{ matrix.tflags }}'
@@ -620,7 +620,7 @@ jobs:
 
       - name: 'cmake run tests'
         if: ${{ matrix.tflags != 'skipall' && matrix.tflags != 'skiprun' }}
-        timeout-minutes: 30
+        timeout-minutes: 12
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP ~FTP ${{ matrix.tflags }}'
           if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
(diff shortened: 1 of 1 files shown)


vszakats:
This doesn't help with a kind of hang where the "run" step seemingly doesn't start,
so GHA doesn't apply the step timeout:
https://github.com/curl/curl/actions/runs/10014744139/job/27685013931?pr=14236

The GHA "log collector" also has issues in this case:
```
2024-07-19T21:49:26.7946580Z Requested labels: windows-latest
2024-07-19T21:49:26.7946898Z Job defined at: curl/curl/.github/workflows/windows.yml@refs/pull/14236/merge
2024-07-19T21:49:26.7947004Z Waiting for a runner to pick up this job...
2024-07-19T21:49:29.8631448Z Job is waiting for a hosted runner to come online.
2024-07-19T21:49:33.4883837Z Job is about to start running on the hosted runner: GitHub Actions 120 (hosted)
```

talregev:
> Finishing tests usually takes on average 10 or less minutes depending on platform. Reduce job step timeouts to reflect that. It helps concluding hung/failed tests earlier, which allows to retry them earlier.
> 
> Also fix a comment while here.
> 
> Closes #14236

This because there are slow vms.
Not all the vms are equal. Sometimes it run on slow vm, then you try debug the tests.
In my opinion, give larger timeout to let the slow vm chance to finish the tests too.

vszakats:
If you have example from the run archive for slow VMs, can you link to some?

I can't remember seeing them. We also need to work with these daily, and I'm willing a sacrifice some legitimate runs to timeout, rather than wait 30 minutes extra for every hung tests, which happen on almost every run. That said if the slow VMs are common we can certainly adjust accordingly.

talregev:
> If you have example from the run archive for slow VMs, can you link to some?
> 
> I can't remember seeing them. We also need to work with these daily, and I'm willing a sacrifice some legitimate runs to timeout, rather than wait 30 minutes extra for hung tests, which happen on almost every run. That said if the slow VMs are common we can certainly adjust accordingly.

You can compare a vm that run the tests 30 min and more, to other vm with similar job that run the tests 10min or less here on curl.
Then you can see that other operation like checkout, take longer. then you can conclude that the slow machine is not connected to the tests.

vszakats:
I think we should see actual examples, before we optimize for exceptions.

talregev:
> I think we should see actual examples, before we optimize for exceptions.

For example for the recent run you have:
The fail vm (slow vm) that cannot finish the test in 30 min:
https://github.com/curl/curl/actions/runs/10023621685/job/27704728068
image

Other similar job that test are take 2min:
https://github.com/curl/curl/actions/runs/10023621685/job/27704727967
image


Notice to cache compiler phase that is not connected to the tests, but in the fail (slow vm) is much longer.
And if you notice this, you can find the pattern in other job, when the slow vm is failing due not enough time for finish the tests.

vszakats:
The failing job finished the build step more than 2x faster than the successful one. Do you think it's realistic that the test run step takes more than 13x _longer_ on the same machine? To me this seems like a hung test, which would have caught by the _job_ timeout had the the build step took a bit longer. With the quick buld step, it got caught by the step timeout.

talregev:
ok, we can try your way, but give in mind that you still have a slow machine while you debug hung tests.

vszakats:
In your example, based on the build times, the hang happens on the _fast_ machine, which confirms that hangs do not always happen on a slow one. On the slow one, tests finished in 2m49s.

This tells that the proposed new timeouts would fit them with a fair margin.

Correlating runner perf vs pronness to hang could be interesting, but in practice it will be difficult to do. These machines are shared VMs so perf isn't static throughout a session, but depend on the neighbours running at the same time, plus possibly a bunch of other things we cannot see or measure. Meaning, these different runs may well be "lucky" and "less-lucky" sessions on the _same_ hardware, rather than fast and slow machines.

Based on my observation, machine perf is relatively consistent within a certain band. It also tends to improve by time.

talregev:
ok. Now I see it. I agree.

vszakats:
One remaining concern:
With high timeouts it's easy to identify a hung job by the time it took. 40–45m = hang. No need to peek into the log.
Lower timeouts will make it less obvious. Something to test in practice I guess.

talregev:
Why you don't want to leave the timeout like current state (before this PR)? If you already monitor and re-run the ci ....

vszakats:
The reason is the the PR message. It allows to see the failure earlier when hung. Which helps iterating and/or retrying faster. For now one needs to wait 45 minutes to know that the run failed.

(_Retrying_ is also held back by the Cygwin job that is stable but takes a longer time than others. For this I was planning to move them to a separate `cygwin.yml` workflow.)

PR #14244, opened 2024-07-20
GHA/non-native: improve, migrate x86_64 FreeBSD with tests from Cirrus CI

vszakats:
- run tests via `make test-ci` instead of `make check` with autotools.
- add `x86_64` job for FreeBSD, with tests.
  It matches the existing Cirrus CI job, with these differences:
  - finishes 3x faster (thanks to parallel tests enabled).
  - librtmp is not enabled because it's slated for removal by FreeBSD.
    (already past the removal deadline, thought the package still
    installs.)
  - DICT and TELNET servers fail to start. Couldn't figure out why.
    It means skipping test 1450 and 1452.
  - it runs more tests, e.g. websockets and ip6-localhost.
  - no `pkg update -f`.
  - it misses the `CRYPTOGRAPHY_DONT_BUILD_RUST=1`, `pkg delete curl`,
    `chmod 777`, `sudo -u nobody` and `sysctl net.inet.tcp.blackhole`
    tricks. The latter is the default in these runners, the others did
    not affect results.
- set `-j0` for tests in the NetBSD job. Flaky otherwise.

Closes #14244


vszakats:
```
Message from librtmp-2.4.20190330_1:
--
===>   NOTICE:
This port is deprecated; you may wish to reconsider installing it:
Depends on legacy functionality of OpenSSL and superseded by multimedia/ffmpeg.
It is scheduled to be removed on or after 2024-04-30.
```

vszakats:
The only two problems seen compared to cirrus were these two:
```
TESTINFO: "failed starting neg TELNET server" 1 time (1452)
TESTINFO: "failed starting DICT server" 1 time (1450)
```
https://github.com/curl/curl/actions/runs/10020191246/job/27697561609#step:3:10382

vszakats:
Is there any objection to migrate our FreeBSD Intel CI job from Cirrus CI to GHA?

PR #14253, opened 2024-07-22
http2+transfer+gnutls, upload eos/shutdown handling

icing:
http2, upload eos handling

- replace the counting of upload lengths with the new eos send flag
- improve frequency of stream draining to happen less on  events where it is not needed
- this PR is based on #14220

test case improvement:
- let client 'upload-pausing' handle http versions

transfer: force send/recv while shutdown of a direction is in progress.

gnutls: fix handling when sending the close notify blocks.

With these changes, the pytest cases work when simulating a 90% chance of socket write block.


bagder:
This PR now has merge conflicts.

PR #14257, opened 2024-07-22
connection shutdown, multi_socket handling

icing:
- implement the socket hash user/reader/writer processing also for connections that are being shut down by the connection cache.
- split out handling of current vs. last pollset socket event handling into a function available in other code parts
- add `shutdown_poll` pollset to `connectdata` struct so that changes in the pollset can be recorded during shutdown. (The internal handle cannot keep it since it might be used for many connections)


bagder:
This fixes #14252, right?

bagder (approved):

PR #14266, opened 2024-07-23
vtls: stop offering alpn http/1.1 for http2-prior-knowledge

jay:
- For HTTPS if http2-prior-knowledge is set then only offer h2 (HTTP/2) alpn to the server for protocol negotiation.

Prior to this change both HTTP/2 ("h2") and HTTP/1.1 ("http/1.1") were offered for ALPN when http2-prior-knowledge was set.

CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE (tool: --http2-prior-knowledge) is meant to send non-TLS HTTP requests HTTP/2 when it is known the server supports them. However when HTTPS is used then it attempts to first negotiate the connection with ALPN. In that case the user likely does not want to offer http/1.1 to the server as an acceptable protocol.

Reported-by: kit-ty-kate@users.noreply.github.com

Fixes https://github.com/curl/curl/issues/9963
Closes #xxxxx

### DIFF
--- a/docs/cmdline-opts/http2-prior-knowledge.md
+++ b/docs/cmdline-opts/http2-prior-knowledge.md
@@ -23,3 +23,7 @@ Issue a non-TLS HTTP requests using HTTP/2 directly without HTTP/1.1 Upgrade.
 It requires prior knowledge that the server supports HTTP/2 straight away.
 HTTPS requests still do HTTP/2 the standard way with negotiated protocol
 version in the TLS handshake.
+
+Since 8.10.0 if this option is set for an HTTPS request then the application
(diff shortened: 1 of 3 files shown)


bagder:
A matching mention in the docs for `CURLOPT_HTTP_VERSION` would be great!

jay:
ok how about this

> Since 8.10.0 if this option is set for an HTTPS request then the application layer protocol version (ALPN) offered to the server is only HTTP/2. Prior to that both HTTP/1.1 and HTTP/2 were offered.

icing (approved):

PR #14273, opened 2024-07-25
Support __ss_family use on NonStop platforms.

rsbeckerca:
The definition of sockaddr_storage incorrectly specifies the ss_family field as __ss_family. This fix conditionally allows builds to succeed on all NonStop platforms.


bagder (approved):

bagder:
Thanks!

PR #14274, opened 2024-07-25
curl.h: lower happy eyeballs timeout from 200ms to 40ms

bagder:
- because if IPv6 is somehow not working correctly or quickly, we want the IPv4 attempt to kick in and cover for the issue

- this is for example notable on Windows where failing to connect to a non-listening port can take way longer than 200 milliseconds even on localhost

A downside with more aggressive shorter timeout is that libcurl is likely to do parallel connects more frequently. (thus using more resources in the client side but possibly also in the server side)

### DIFF
--- a/docs/cmdline-opts/happy-eyeballs-timeout-ms.md
+++ b/docs/cmdline-opts/happy-eyeballs-timeout-ms.md
@@ -25,4 +25,6 @@ first connection to be established is the one that is used.
 The range of suggested useful values is limited. Happy Eyeballs RFC 6555 says
 "It is RECOMMENDED that connection attempts be paced 150-250 ms apart to
 balance human factors against network load." libcurl currently defaults to
-200 ms. Firefox and Chrome currently default to 300 ms.
+40 ms.
+
+(Up until curl 8.9.0, curl's default was 200ms.)
--- a/docs/libcurl/opts/CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS.md
+++ b/docs/libcurl/opts/CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS.md
@@ -34,14 +34,14 @@ milliseconds. If the IPv6 address cannot be connected to within that time then
 a connection attempt is made to the IPv4 address in parallel. The first
 connection to be established is the one that is used.
 
-The range of suggested useful values for *timeout* is limited. Happy
-Eyeballs RFC 6555 says "It is RECOMMENDED that connection attempts be paced
-150-250 ms apart to balance human factors against network load." libcurl
-currently defaults to 200 ms. Firefox and Chrome currently default to 300 ms.
+The range of suggested useful values for *timeout* is limited. Happy Eyeballs
+RFC 6555 says "It is RECOMMENDED that connection attempts be paced 150-250 ms
+apart to balance human factors against network load." libcurl currently
+defaults to 40 ms.
 
 # DEFAULT
 
-CURL_HET_DEFAULT (currently defined as 200L)
+CURL_HET_DEFAULT (defined as 40L)
 
 # %PROTOCOLS%
 
@@ -64,6 +64,10 @@ int main(void)
 ~~~
 
 # %AVAILABILITY%
+
+# HISTORY
+
+Up until curl 8.9.0, libcurl's default was 200ms.
 
 # RETURN VALUE
 
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -941,9 +941,9 @@ typedef enum {
 #define CURLSSLOPT_AUTO_CLIENT_CERT (1<<5)
 
 /* The default connection attempt delay in milliseconds for happy eyeballs.
-   CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS.3 and happy-eyeballs-timeout-ms.d document
-   this value, keep them in sync. */
-#define CURL_HET_DEFAULT 200L
+   CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS.md and happy-eyeballs-timeout-ms.md
+   document this value, keep them in sync. */
+#define CURL_HET_DEFAULT 40L
 
 /* The default connection upkeep interval in milliseconds. */
 #define CURL_UPKEEP_INTERVAL_DEFAULT 60000L


bagder:
Thoughts, @icing ?

dfandrich:
It would be nice to have some hard data to choose the timeout value. In reading
some recent random reports on the web, I see statements on user latency surveys
like "average latency is usually under 30 ms", "median latencies ranged from 10
ms to 27 ms", "Latency by connection type 10-42 ms [for landline connections]"
"9-19 ms latency within metropolitan area".  And some of those were ambiguous
about whether these were one-way or round-trip latencies.

Even if we assume that they're round trip, a 20 ms timeout in curl would then
imply that more than half of people would receive a timeout and therefore a
second parallel connection on basically every connection attempt, EVEN IF they
have working IPv6. 20 ms is simply not enough time to send a connection request
and receive the response. And the statements above were mostly for home
connections, so mobile (a huge source of traffic, and one where avoiding
unnecessary traffic is important) and satellite ones would presumably be even
worse.

Happy Eyeballs isn't supposed to double the connection traffic, so I think in
2024 the timeout should remain considerably higher than 20 ms.

jay:
> * because if IPv6 is somehow not working correctly or quickly, we want the IPv4 attempt to kick in and cover for the issue

Yeah but it already does that? 200ms is lower iirc than what others use and it's worked well so far, I don't see this as a reason to lower it further

> * this is for example notable on Windows where failing to connect to a non-listening port can take way longer than 200 milliseconds even on localhost

Is there an issue you are thinking of and are you sure it is not just an outlier?

dfandrich:
FWIW, for some reason I thought the PR was lowering the timeout to 20 ms, not 40 ms, which is why my arguments are all talking about 20 ms. I still think 40 ms is too low, but I'm not as strongly opposed as I would be at 20 ms.

vszakats:
Considering this is a workaround for Windows, would it make sense to limit this default to Windows only? It'd avoid applying the downside to all platforms just because Windows works oddly.

bagder:
> Is there an issue you are thinking of and are you sure it is not just an outlier?

On Windows, try running `curl localhost:6789` or against any other port where nothing is listening and see how long time it takes for curl to fail.

jay:
I'm not seeing much of a difference with localhost and I don't have a public IPv6 address on this computer so I can't test curl.se. Here's the localhost results. I used a utility called [pbench](https://github.com/chemodax/pbench) (which I built [modified](https://github.com/jay/pbench/tree/modified) from source, I don't know about the installer but I looked at the source and it is ok)

pbench --count 5 curl -v -4 localhost:6789
~~~
== Performance of curl -v -4 localhost:6789

The command was run 5 times.

                 average     minimum     maximum
   Real time:    1.042 s     1.035 s     1.046 s
    CPU time:    0.012 s     0.000 s     0.015 s
   User time:    0.006 s     0.000 s     0.015 s
 Kernel time:    0.006 s     0.000 s     0.015 s
     Read IO:        1 (0 KB)  1 (0 KB)  1 (0 KB)
    Write IO:        0 (0 KB)  0 (0 KB)  0 (0 KB)
    Other IO:      134 (2 KB)  132 (2 KB)  135 (2 KB)
Peak WS size:     6059 KB  6040 KB  6076 KB

  Total time:    5.215 s
~~~

pbench --count 5 curl -v --happy-eyeballs-timeout-ms 40 localhost:6789
~~~
== Performance of curl -v --happy-eyeballs-timeout-ms 40 localhost:6789

The command was run 5 times.

                 average     minimum     maximum
   Real time:    1.081 s     1.075 s     1.096 s
    CPU time:    0.012 s     0.000 s     0.031 s
   User time:    0.003 s     0.000 s     0.015 s
 Kernel time:    0.009 s     0.000 s     0.015 s
     Read IO:        1 (0 KB)  1 (0 KB)  1 (0 KB)
    Write IO:        0 (0 KB)  0 (0 KB)  0 (0 KB)
    Other IO:      159 (2 KB)  155 (2 KB)  177 (2 KB)
Peak WS size:     6051 KB  6040 KB  6064 KB

  Total time:    5.411 s
~~~

pbench --count 5 curl -v --happy-eyeballs-timeout-ms 200 localhost:6789
~~~
== Performance of curl -v --happy-eyeballs-timeout-ms 200 localhost:6789

The command was run 5 times.

                 average     minimum     maximum
   Real time:    1.240 s     1.236 s     1.247 s
    CPU time:    0.009 s     0.000 s     0.015 s
   User time:    0.006 s     0.000 s     0.015 s
 Kernel time:    0.003 s     0.000 s     0.015 s
     Read IO:        1 (0 KB)  1 (0 KB)  1 (0 KB)
    Write IO:        0 (0 KB)  0 (0 KB)  0 (0 KB)
    Other IO:      155 (2 KB)  155 (2 KB)  158 (2 KB)
Peak WS size:     6048 KB  6036 KB  6060 KB

  Total time:    6.208 s
~~~

I always see time of at least one second plus the eyeballs timeout, for example
40ms: `curl: (7) Failed to connect to localhost port 6789 after 1049 ms: Could not connect to server`
200ms: `curl: (7) Failed to connect to localhost port 6789 after 1208 ms: Could not connect to server`

bagder:
> I'm not seeing much of a difference with localhost

On Windows it takes a *long* time to find out that the connection doesn't work. Easily more than 200 milliseconds, so even when there is nothing listening we spend the full 200 milliseconds before the IPv4 attempt can start.

Thus, if you have a working IPv4 but non-working IPv6 for the server, curl is doing happy eyeballs is annoyingly enough by default *much* slower than an old-fashioned IPv4-only attempt. Shortening the happy eyeballs timeout somewhat limits the damage.

Ideally of course, we can figure out if we can make these connect attempts fail faster on Windows. For example, on my Linux machine trying `curl localhost:6789` returns a failure in about 18 milliseconds.

bagder:
I'm retracting this for now. It can be worked-around by users for now. Ideally we can improve error-detection on Windows.

dfandrich:
For posterity, the user workaround involves explicitly setting [CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS.html)

PR #14276, opened 2024-07-26
GHA/windows: enable MulitSSL in an MSVC job

talregev:
Ref: 98da147b18fa4963efb425df8400ca40e7526646 #14305
Closes #14276

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -493,12 +493,12 @@ jobs:
       matrix:
         include:
         - name: 'schannel U'
-          install: 'brotli zlib zstd libpsl nghttp2 libssh2[core,zlib] pkgconf gsasl'
+          install: 'brotli zlib zstd libpsl nghttp2 libssh2[core,zlib] pkgconf gsasl openssl'
           arch: 'x64'
           plat: 'windows'
           type: 'Debug'
           tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
-          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON  -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DUSE_WIN32_IDN=ON -DCURL_USE_GSASL=ON'
+          config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON  -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DUSE_NGHTTP2=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DUSE_WIN32_IDN=ON -DCURL_USE_GSASL=ON -DCURL_USE_OPENSSL=ON -DCURL_DEFAULT_SSL_BACKEND=schannel'
         - name: 'openssl'
           install: 'brotli zlib zstd libpsl nghttp2 nghttp3 openssl libssh2 pkgconf gsasl'
           arch: 'x64'


talregev:
> Can you make this part of an existing job?
> 
> These jobs are already flaky and each one added increases the chance for a PR to go red.

This PR is just meant to show the compilation problem. When it solve, we can think where to put the multissl.

lib/vtls/vtls.c
```diff
@@ -414,7 +414,14 @@ int Curl_ssl_init(void)
...
 
 #if defined(CURL_WITH_MULTI_SSL)
+# ifdef _MSC_VER
+#   pragma warning(push)
+#   pragma warning(disable : 4132)
+# endif
 static const struct Curl_ssl Curl_ssl_multi;
+# ifdef _MSC_VER
+#   pragma warning(pop)
+# endif
```
vszakats, line 424: Is this an MSVC bug? Can we do a standard C initialization here to avoid silencing warnings?

talregev: I don't think it msvc bug. But I cannot add standard C initialization because on the case where we don't have multi ssl, and it create a  multi redefinition, in line 1326

talregev:
@vszakats I add another suggestion.

talregev:
https://stackoverflow.com/questions/40738351/forward-declaration-as-struct-vs-class

FYI

lib/vtls/vtls.c
vszakats: Have you tried moving the `Curl_ssl_cleanup` function below the initialization (e.g. before the function `multissl_version`), and dropping the declaration from line 417?

talregev:
https://learn.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements-2017?view=msvc-170#uninitialized-const-variables

There is a problem with forward deceleration and const. Maybe it is msvc bug?

lib/vtls/vtls.c
talregev: Yes. and it work. See my  new code.

bagder:
> const object should be initialized

Global variables are guaranteed to be initialized to zero according to the standard. So what exactly does this mean?

talregev:
> > const object should be initialized
> 
> Global variables are guaranteed to be initialized to zero according to the standard. So what exactly does this mean?

On the original code on line 417, it forward declaration, not a definition. It should not do initialization on this line.
For me it seems that you cannot do forward declaration with const in msvc, even if it legal in c.  That why I think it msvc bug.

talregev:
https://learn.microsoft.com/th-th/cpp/cpp/declarations-and-definitions-cpp?view=msvc-170#definitions

talregev:
https://stackoverflow.com/a/936589/550384

vszakats (changes requested):
Looking good! As discussed earlier, I'd prefer integrating this into an existing job, rather than adding a new one to avoid adding more chance for flakiness.

How about enabling Schannel in the existing 'openssl' job?

talregev:
I know you don't want to add many tests. but my plan is to have multissl to install as much ssl backends for curl together.
I already succeeded with vcpkg to have 4:  
```
C:\src\vcpkg\installed\x64-windows\tools\curl>curl --version
curl 8.9.0-DEV (Windows) libcurl/8.9.0-DEV GnuTLS/3.8.4 (mbedTLS/2.28.8) (OpenSSL/3.3.1) (Schannel) zlib/1.3.1
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI threadsafe TLS-SRP Unicode UnixSockets
```

vszakats:
Tests will only run on the default TLS backend, so running with extra backends will not increase test coverage. Having a single extra backend tests the MultiSSL codepath, this is covered in this PR. Adding more backends would only add _build tests_ for the extra backends. We have this covered for many now in separate jobs, and new ones can be appended to an existing job.

I think the situation with flaky MSVC jobs is already very bad, breaking almost all runs. I don't know why that it, but adding more flaky jobs don't seem to help us because of this.

All in all, I think we'd better off we using the existing `openssl` job to test MultiSSL, and untested backends can be appended there too.

talregev:
Can we just build the tests without running them on multissl job?
If I add openssl to Schannel, openssl will be the default.

vszakats:
Why would that be better than adding it to an existing job?

talregev:
My goal is to add on the same build, as much backend together. If I add more backends in openssl test, it will be not be a default anymore, because curl choose the default automatic according the first in the alphabet order. 
As you can see this code is not compile with msvc on curl ci and then we found warning / errors on the code.
Also I want to find these error and warning, and try to fix them if possible.

talregev:
I am not sure why the ci fail

vszakats:
> My goal is to add on the same build, as much backend together. If I add more backends in openssl test, it will be not be a default anymore, because curl choose the default automatic according the first in the alphabet order. As you can see this code is not compile with msvc on curl ci and then we found warning / errors on the code. Also I want to find these error and warning, and try to fix them if possible.

You can manually select the default TLS backend using `-DCURL_DEFAULT_SSL_BACKEND=` if it becomes something else than OpenSSL.

The MultiSSL compiler warning came up because of enabling MultiSSL at all. Adding more backends will not uncover more such issues. Either way I've nowhere said not to enable more backends, but to add new to ones to an existing job.

Also notice that MultiSSL isn't really where the puck is going in curl, e.g. because it lacks support for HTTP/3 and the original reasons for having it are now obsolete.

talregev:
ok. I will Enable Schannel in openssl job.
Thank you for your review, and time for listening my ideas.

talregev:
we can wait until the PR #14305 will be merge, and I will rebase this PR from the master after the merge.

vszakats (dismissed):

vszakats:
We need to find another job for MultiSSL. The openssl one has HTTP/3 enabled which is incompatible with MultiSSL. Hence the CI failure.

How about the 'Schannel U' job? It needs the `-DCURL_DEFAULT_SSL_BACKEND=schannel` option for keeping Schannel the default.

talregev:
Done.

vszakats (approved):

MarcelRaad (approved):

vszakats:
Thanks @talregev!

PR #14281, opened 2024-07-26
os400: workaround an IBM ASCII run-time library bug

monnerat:
IBM-provided ASCII function puts() does not output an expected trailing newline: emulate the correct behavior using other functions.


bagder (approved):

bagder:
Thanks!

PR #14283, opened 2024-07-26
debug test or event based shutdown issues

icing:
- build libcurl with debug enabled
- set CURLOPT_VERBOPSE and CURLOPT_DEBUGFUNCTION, CURLOPT_DEBUDATA on each transfer
- shutdown handling will copy over the function+data in the internal shutdown handle
- DEBUGFUNCTION needs to work when the transfer has been freed!

produce trace log, ideally also set `curl_global_trace("all") for extra info


dfandrich:
Not sure if this is just a temporary PR for CI only or not but in case not, some of the constant names in the description have typos.

icing:
Replaced with #14296

PR #14289, opened 2024-07-27
os400: build cli manual.

monnerat:
This is my first attempt to use PASE perl in the build process.

Unfortunately perl does not deal properly with all the various possible OS/400 file encodings , thus I now unconditionally transcode all source files to UTF-8: this might be somewhat lengthy upon first build.

The man page is not built because OS/400 has no man command. However the text version is stored both in DB2 and IFS, and the cli built-in manual is enabled.

See the commit comments for more info.

Please note we're still far from a working test environment. Mainly because forks and execs.

### DIFF
--- a/packages/Makefile.am
+++ b/packages/Makefile.am
@@ -33,6 +33,7 @@ EXTRA_DIST = README.md \
   OS400/curl.inc.in \
   OS400/initscript.sh \
   OS400/config400.default \
+  OS400/make-docs.sh \
   OS400/make-include.sh \
   OS400/make-lib.sh \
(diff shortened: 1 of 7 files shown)


bagder (approved):

monnerat:
Thanks for approval @bagder . However let me fix shellcheck warnings before commit.
Turning this into a draft and I'll tell you when reaady.

monnerat:
@bagder : shellcheck problems resolved. IMO it's ready for commit.

bagder:
Thanks!

Issue #14280, opened 2024-07-26
socket_function callbacks for unrelated sockets that are never removed


JulesRapanga:
### I did this

We use libcurl in our software via the multi API driven with libev. After updating from 8.7.1 to 8.9.0 our program would start using 100% cpu after a couple of minutes. On boot we make HTTP requests a several hosts and (when idle) make a single request to one host every minute. It seems that when we make the second request to this host we get callbacks to the function provided to `CURLMOPT_SOCKETFUNCTION` telling us to monitor several sockets that are connected to different hosts. Subsequently `libev` constantly tells us there is activity on these sockets and we call `curl_multi_socket_action` but to no effect or error. After the request finishes we are told to stop monitoring the socket that was actually used for the request but not for the spurious ones, so we continue to spin.

### curl/libcurl version

curl 8.9.0

### operating system

Ubuntu Focal

icing:
This seems a side effect of the connection shutdown changes we made in 8.9. I assume that the shutdown triggers the socket monitoring, but then never causes the sockets to be removed again. Or do you seen "unkown" sockets to appear there?

JulesRapanga:
They're not unknown, they're all ones that have been used for previous requests.

JulesRapanga:
Is there a particular commit I could try backing out to see if that fixes it?

icing:
That would be difficult. It was a larger operation. I'll have a look at it next week, see what I can find. Can you build a libcurl with which we can test changes?

JulesRapanga:
Yeah we can build whatever patches you provide and say what it says

CountMurphy:
Seeing similar behavior when running rtorrent.

icing:
I made #14283 with extra DEBUG output on connection shutdown and its event handling. If you could build a libcurl with `--enable-debug` and set a CURLOPT_DEBUGFUNCTION on the transfers, we should get some traces to show us what is going wrong, hopefully.

JulesRapanga:
@icing 
[libcurl.log](https://github.com/user-attachments/files/16396694/libcurl.log)

The sockets `23, 49, 50, 265, 281` were spuriously passed to the socket callback at `20:45:15`

icing:
Thanks for the log! I see that we are on the right track. The sockets are, correctly, added for shutting down the connections in the multi's connection cache when they are considered to be "too old". But the cache's is never called in invoke the events on these sockets.

[...]

PR #14296, opened 2024-07-29
fix connection shutdown for event based processing

icing:
connections being shutdown would register sockets for events, but then never remove these sockets again. Nor would the shutdown effectively been performed. See #14280

- If a socket event involves a transfer, check if that is the connection cache internal handle and run its multi_perform() instead (the internal handle is used for all shutdowns).
- When a timer triggers for a transfer, check also if it is about the connection cache internal handle.
- During processing shutdowns in the connection cache, assess the shutdown timeouts. Register a Curl_expire() of the lowest value for the cache's internal handle.
- do connection shutdown under the same `sigpipe` handling as other transfers

### DIFF
--- a/lib/multi.c
+++ b/lib/multi.c
@@ -2739,9 +2739,13 @@ CURLMcode curl_multi_perform(struct Curl_multi *multi, int *running_handles)
         sigpipe_ignore(data, &pipe_st);
         nosig = data->set.no_signal;
       }
-      result = multi_runsingle(multi, &now, data);
-      if(result)
-        returncode = result;
+
+      if(data != multi->conn_cache.closure_handle) {
+        /* connection cache handle is processed below */
+        result = multi_runsingle(multi, &now, data);
+        if(result)
+          returncode = result;
+      }
 
       data = datanext; /* operate on next handle */
     } while(data);
@@ -3192,6 +3196,7 @@ static CURLMcode multi_socket(struct Curl_multi *multi,
   bool first = FALSE;
   bool nosig = FALSE;
   SIGPIPE_VARIABLE(pipe_st);
+  bool run_conn_cache = FALSE;
 
   if(checkall) {
     /* *perform() deals with running_handles on its own */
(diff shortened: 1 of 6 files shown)


lib/multi.c
```diff
@@ -3303,6 +3314,9 @@ static CURLMcode multi_socket(struct Curl_multi *multi,
   if(first)
     sigpipe_restore(&pipe_st);
 
+  if(run_conn_cache)
+    Curl_conncache_multi_perform(multi);
```
bagder, line 3318: I think this should be done *before* `sigpipe_restore()`, shouldn't it? I figure there is a risk that it would use TLS on a dead connection.

icing: Fixed in latest push.

bagder (approved):

Cogitri:
Hej,

I believe this change has caused libcurl to leak SIGPIPE in some situations - at least our test suite exited with SIGPIPE at random after upgrading to libcurl 8.9.1.  Reverting this commit causes it to work again. This might very well be an error in our test suite - should we handle SIGPIPE ourselves? If this is intended behaviour, I would suggest a mention in the release notes.

I have to admit I don't really know enough of the inner workings of cURL or OpenSSL to judge what's right though - sorry for not providing more info here. If there's anything I can help with, just @ me! Thanks for the work on this :)

dfandrich:
What OS and TLS back-end are you using and what version? Are you setting CURLOPT_NOSIGNAL?

imaami:
> Hej,
> 
> I believe this change has caused libcurl to leak SIGPIPE in some situations - at least our test suite exited with SIGPIPE at random after upgrading to libcurl 8.9.1. Reverting this commit causes it to work again. This might very well be an error in our test suite - should we handle SIGPIPE ourselves? If this is intended behaviour, I would suggest a mention in the release notes.
> 
> I have to admit I don't really know enough of the inner workings of cURL or OpenSSL to judge what's right though - sorry for not providing more info here. If there's anything I can help with, just @ me! If you prefer, I can also open an issue about this. Thanks for the work on this :)

I am seeing what looks a lot like this with transmission-daemon in Termux (aarch64). Since upgrading libcurl to 8.9.1 yesterday, transmission-daemon will simply die after maybe 15-30 seconds with the exit code 141 (i.e. 128 + SIGPIPE).

Cogitri:
Opened the issue, I’ll check for the last curl call tomorrow. Thanks for the quick reply!

@dfandrich I don‘t think I‘m setting that flag. I‘m using OpenSSL 3.x

PR #14304, opened 2024-07-29
tool_operate: more defensive socket code

bagder:
- use 'struct sockaddr' to getsockname() and its sa_family member

- use 'curl_socklen_t' instead of 'socklen_t'

- check for AF_INET6 to exist instead assuming it does

Reported-by: Harry Sintonen

### DIFF
--- a/src/tool_operate.c
+++ b/src/tool_operate.c
@@ -154,14 +154,10 @@ static bool is_pkcs11_uri(const char *string)
 #ifdef IP_TOS
 static int get_address_family(curl_socket_t sockfd)
 {
-  struct sockaddr_storage addr;
-  socklen_t addrlen = sizeof(addr);
+  struct sockaddr addr;
+  curl_socklen_t addrlen = sizeof(addr);
   if(getsockname(sockfd, (struct sockaddr *)&addr, &addrlen) == 0)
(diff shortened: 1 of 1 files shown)


dfandrich:
I'm guessing the removal of the __TANDEM fix was an accident.

bagder:
> I'm guessing the removal of the __TANDEM fix was an accident.

Actually not. This uses `sockaddr` and not `sockaddr_storage` so it uses `sa_family` and not `ss_family` which I presume has different conditions. If this works in CI I'll ask @rsbeckerca to verify.

bagder:
@rsbeckerca can you check if this version builds fine for you on NonStop? It seems like a generally more portable approach.

rsbeckerca:
> @rsbeckerca can you check if this version builds fine for you on NonStop? It seems like a generally more portable approach.

Yes. $DAYJOB is kicking my ____ I will get to this shortly.

bagder:
I'll go ahead and merge this in the mean time as this helps other platforms and I suspect it works on NonStop as well.

rsbeckerca:
Build on NonStop works. Starting test suite.

rsbeckerca:
@bagder LGTM

PR #14306, opened 2024-07-30
wolfssl: avoid calling get_cached_x509_store if store is uncachable

alexsn:
There's no need for get_cached_x509_store call if the return value won't be used for caching anyway.

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -571,7 +571,7 @@ CURLcode Curl_wssl_setup_x509_store(struct Curl_cfilter *cf,
   bool cache_criteria_met;
 
   /* Consider the X509 store cacheable if it comes exclusively from a CAfile,
-     or no source is provided and we are falling back to OpenSSL's built-in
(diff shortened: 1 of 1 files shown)


bagder:
@icing, @dfandrich and me discussed this a bit. It is not clear that a user prefers to disable CA caching just because they use the callback, because it is a quite significant performance blow and there are numerous use cases for the callback that does not alter the CA store at all. Or that alter it but is fine with it remaining in memory.

So, we now stress the caching details in the CTX callback documentation: 674e1025043b8e41d

Don't you think that is enough? It is a minor behavior difference, but I think for the good for almost everyone.

alexsn:
> Don't you think that is enough? It is a minor behavior difference, but I think for the good for almost everyone.

Yeah, I think it's fine. I'll update the PR to avoid calling `get_cached_x509_store` if `cache_criteria_met == FALSE`

bagder (approved):

bagder:
Thanks!

PR #14313, opened 2024-07-30
Support QUIC sockets

moritzbuhl:
Hi, this is still very early as the quic kernel module is still under discussion on the LKML.
This is unfinished and needs a lot more modification.
I would appreciate some initial feedback now before I put more work in.
`./src/curl --http3-only https://cloudflare-quic.com/` works with the code, I did not test any further yet.

Attempt to support QUIC sockets from https://github.com/lxin/quic and add a H3 filter based on curl_ngtcp2.c.

Currently links against libquic instead of using the available TLS code.
Does not use the send and recv functions from vquic.c, i.e. only uses sendmsg.
Code needs simplification and clean-up.
Tests are missing.


icing:
Interesting. What is the protocol exposed on such a socket? The decrypted QUIC packets? The one would need something like `ngtcp2`, but without the crypto parts? Or does the kernel want to manage retransmissions, ACKS and congestion windows? What then is talked on the socket?

vszakats:
Suggestion to exclude this check for Windows:
```diff
--- a/CMake/Platforms/WindowsCache.cmake
+++ b/CMake/Platforms/WindowsCache.cmake
@@ -138,6 +138,8 @@ set(HAVE_SYS_UTIME_H 1)
 set(HAVE_TERMIOS_H 0)
 set(HAVE_TERMIO_H 0)
 set(HAVE_UTIME_H 0)  # mingw-w64 has it (wrapper to sys/utime.h)
+set(HAVE_NETINET_QUIC_H 0)
+set(HAVE_LINUX_QUIC_H 0)
 
 set(HAVE_DIRENT_H 0)
 set(HAVE_OPENDIR 0)
```

moritzbuhl:
@icing I think the best documentation on how it work is at https://github.com/lxin/quic/blob/main/libquic/quic.man
But in short, reads and writes to a socket are wrapped in stream or datagram frames by the kernel module. 
A
To tell the kernel which id the data is for, a cmsghdr with a `stream_info` struct is used.

bagder:
@moritzbuhl running `make checksrc` locally is a good first step to make sure you (mostly) follow our code style

vszakats:
Reminder to add the new CMake `Find*` module to `Makefile.am` to avoid repeating a dist snafu like #14320.

moritzbuhl:
libquic (also at https://github.com/lxin/quic/tree/main/libquic) is just some abstraction on the necessary gnutls and setsockopt/send/recv calls to establish a quic connection.
It should be replaced with Curl_vquic_tls calls.

I will add some info on the kernel module to the HTTP3 docs.

For now, here are some build instructions fro the kernel module from the README:
```
# Packages Required: (kernel_version >= 5.14)
# - make autoconf automake libtool pkg-config
# - gnutls-devel / gnutls-dev
# - kernel-devel / linux-headers-$(uname -r)

git clone https://github.com/lxin/quic.git
cd quic/
./autogen.sh
./configure --prefix=/usr
make
make install
```

For curl I do:
`cmake .. -DUSE_LINUX_QUIC=on -DUSE_GNUTLS=on -DENABLE_DEBUG=on`

moritzbuhl:
[curl_linux.c L130-690](https://github.com/curl/curl/pull/14313/files#diff-dc1410f98f7babc52a2a9c005d2fd57b42cc022e06f2adee7b26317947530c58R130-R690) are now similar to what the `libngtcp2_crypto_...` libraries are doing, just explicitly for the handshake.
The quictls frok-family code is still completely untested, the gnutls code should work and the wolfssl code is still missing.
Once the other crypto libraries work I will update the build instructions.

CMakeLists.txt
```diff
@@ -886,7 +886,27 @@ if(USE_OPENSSL_QUIC)
...
+  include_directories(${NGHTTP3_INCLUDE_DIRS})
+  list(APPEND CURL_LIBS ${NGHTTP3_LIBRARIES})
+  list(APPEND LIBCURL_PC_REQUIRES_PRIVATE "libnghttp3")
+endif()
+
+if(USE_MBEDTLS OR
+   USE_BEARSSL OR
+   USE_SECTRANSP)
+  message(WARNING "A selected TLS library does not support TLS 1.3.")
+endif()
```
vszakats, line 907: This snippet has been delete in master, it's safe to delete from here as well.

moritzbuhl:
I think with a few more ifdefs and the wolfssl/openssl/ssl.h defines it would be possible to reuse the ssl crypto code.
I decided against it because I didn't see that elsewhere in the code base.

moritzbuhl:
Testcase 2502 still fails (probably) because of a QUIC `FRAME_ENCODING_ERROR (0x07)`. We are still debugging this.

I tried removing the tracking of `upload_left` and the extra buffering of sent data that ngtcp2 does because the user space QUIC implementations do the ACK handling. To my understanding nghttp3 also integrates into this with `nghttp3_conn_add_ack_offset` and with the kernel module the only situation of receiving an ACK in user space is a `QUIC_EVENT_STREAM_UPDATE` event. However, to my understanding committing the HTTP3 send data to the QUIC socket ensures there is still a reference to that data until it is either ACKd or the connection dies beforehand. Maybe this could cause problems with resumption? I don't really know HTTP (and curl) well enough to tell.

moritzbuhl:
https://github.com/lxin/quic/commit/88de4be2426b2dcce4c1b1c31202c09ad24ebeaa should fix testcase 2502.

bagder:
Since this PR is not working, is not actively being worked on, merge conflicts with master and it seems to be waiting for work to get accepted/merged into the kernel, it is now closed.

We can reopen it again at once when there is new work or activity. Or just create a new one, whichever is the easiest.

PR #14316, opened 2024-07-30
unittests and fixes for gtime2str

icing:
x506asn1.c: fix issues in GTime2str() and add unit test cases to verify correct behaviour.

### DIFF
--- a/lib/vtls/x509asn1.c
+++ b/lib/vtls/x509asn1.c
@@ -518,12 +518,13 @@ static CURLcode GTime2str(struct dynbuf *store,
   fracl = 0; /* no fractional seconds detected so far */
   if(fracp < end && (*fracp == '.' || *fracp == ',')) {
     /* Have fractional seconds, e.g. "[.,]\d+". How many? */
-    tzp = fracp++; /* should be a digit char or BAD ARGUMENT */
+    fracp++; /* should be a digit char or BAD ARGUMENT */
+    tzp = fracp;
     while(tzp < end && ISDIGIT(*tzp))
       tzp++;
     if(tzp == fracp) /* never looped, no digit after [.,] */
       return CURLE_BAD_FUNCTION_ARGUMENT;
-    fracl = tzp - fracp - 1; /* number of fractional sec digits */
+    fracl = tzp - fracp; /* number of fractional sec digits */
     DEBUGASSERT(fracl > 0);
     /* Strip trailing zeroes in fractional seconds.
      * May reduce fracl to 0 if only '0's are present. */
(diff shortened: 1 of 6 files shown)


tests/unit/unit1656.c
```diff
@@ -0,0 +1,133 @@
...
+  { "19032113434017.01-01:10", "1903-21-13 43:40:17.01 UTC-01:10", CURLE_OK },
+};
+
+static bool do_test(struct test_spec *spec, size_t i, struct dynbuf *dbuf)
+{
+  CURLcode result;
+  const char *in = spec->input;
+
+  Curl_dyn_reset(dbuf);
+  result = Curl_x509_GTime2str(dbuf, in, in + strlen(in));
```
dubek, line 86: Consider adding a test were the string doesn't end with `\0` (that is, `end` points to a digit or a dot or something like that).

lib/vtls/x509asn1.c
```diff
@@ -552,6 +559,15 @@ static CURLcode GTime2str(struct dynbuf *store,
                        sep, (int)tzl, tzp);
 }
 
+#ifdef UNITTESTS
+/* used by unit1489.c */
```
dubek, line 563:
```suggestion
/* used by unit1656.c */
```

tests/unit/unit1656.c
bagder: Right, add a length-field in the struct for each case or something so that you can make a date string end in the middle and it should stop parsing there.

lib/vtls/x509asn1.c
icing: Right, needed to use another number and now there are relics. Thanks for catching this.

tests/unit/unit1656.c
icing: Which begs the question: what would be the expected test result on this?

icing: @bagder and it should fail any 0 between being and end?

bagder: Well, if the string until the end if a valid string then that is fine and that's the date, even if it continues looking like a date beyond the end. If it is not a valid date before the end, then it's a failure.

bagder (approved):
Are you fine with me merging this as-is so we can have it in the 8.9.1 release?

icing:
> Are you fine with me merging this as-is so we can have it in the 8.9.1 release?

Yes, please. Let's improve this afterwards.

lib/vtls/x509asn1.c
bagder: maybe commit this correction at least before merge?

bagder:
I just committed the comment fix, I intend to merge this once all/most CI jobs are green.

dubek:
If you think it's good idea, I think I can work on a PR which adds fuzz entrypoints for GTime2str and Curl_extract_certinfo in curl-fuzzer project (not this week, though). It's a bit tricky because it looks like we build curl with openssl over there (which means no Curl_extract_certinfo ...). But if some distros are changing to build curl with gnutls, maybe we should have some fuzzing power spent over there.

bagder:
@dubek it is a really good idea, and as you say that requires a build that uses one of the TLS backends for which we use the x509asn1 parser code. But I see no downside in doing that.

bagder:
I will merge this PR now then so that we can have this fix in the release tomorrow. We can still polish things further in new PRs to merge later.

PR #14319, opened 2024-07-30
Adds OpenSSF Scorecard documentation to `Security.md`.

vpetersson:
As per our conversation today @bagder - adding the much deserved OpenSSF badge :)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -17,7 +17,7 @@ We treat security issues with confidentiality until controlled and disclosed res
 
 ## OpenSSF Scorecard
 
-curl has earned Gold status on the OpenSSF Scorecard, reflecting its adherence to
+curl has earned Gold status on the OpenSSF Best Practices, reflecting its adherence to
 rigorous security and best practice standards. This achievement highlights curl's
 comprehensive documentation, secure development processes, effective change control
 mechanisms, and strong maintenance routines. Meeting these criteria demonstrates curl's
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,3 +14,14 @@ If you have found or just suspect a security problem somewhere in curl or
 libcurl, report it on [HackerOne](https://hackerone.com/curl).
 
 We treat security issues with confidentiality until controlled and disclosed responsibly.
+
+## OpenSSF Scorecard
+
+curl has earned Gold status on the OpenSSF Scorecard, reflecting its adherence to
+rigorous security and best practice standards. This achievement highlights curl's
+comprehensive documentation, secure development processes, effective change control
+mechanisms, and strong maintenance routines. Meeting these criteria demonstrates curl's
+commitment to security and reliability, ensuring the project's sustainability and
+trustworthiness. This recognition by OpenSSF underscores curl's role as a leader in
+open-source software practices. More information can be found on
+their [OpenSSF page](https://www.bestpractices.dev/projects/63).
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
 SPDX-License-Identifier: curl
 -->
 
+[![OpenSSF Best Practices](https://www.bestpractices.dev/projects/63/badge)](https://www.bestpractices.dev/projects/63)
+
 # [![curl logo](https://curl.se/logo/curl-logo.svg)](https://curl.se/)
 
 Curl is a command-line tool for transferring data specified with URL
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@ Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
 SPDX-License-Identifier: curl
 -->
 
-[![OpenSSF Best Practices](https://www.bestpractices.dev/projects/63/badge)](https://www.bestpractices.dev/projects/63)
-
 # [![curl logo](https://curl.se/logo/curl-logo.svg)](https://curl.se/)
 
 Curl is a command-line tool for transferring data specified with URL


vpetersson:
Linting error appears to be unrelated to the change.

bagder:
We removed the plethora of badges back in 7399fa5b05ce. I'm not sure we are ready to restart this.

I think just providing the info in words and a prominent link might be sufficient.

danielgustafsson:
> We removed the plethora of badges back in [7399fa5](https://github.com/curl/curl/commit/7399fa5b05ce897277b8713c659c939d068570a4). I'm not sure we are ready to restart this.

No thanks.

> I think just providing the info in words and a prominent link might be sufficient.

:+1:

vszakats:
Well, we still explicitly suppress `-Wsign-conversion` warnings :(

bagder:
> Well, we still explicitly suppress `-Wsign-conversion` warnings :(

That's not a requirement for getting a gold for best-practices.

vszakats:
The OpenSSF "Best Practices Working Group" still lists it in their guide:
https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html

Maybe there is a separate guide about those best practices that can be safely ignored.

bagder:
The badge is about the OpenSSF Best Practices Badge Program: https://www.bestpractices.dev/en/projects/63

bagder:
I would be deeply upset if anyone would insist on -Wsign-conversion for real, because it is not intelligent.

vszakats:
Oh, thanks @bagder!

So maybe those compiler folks who decided to enable it via `-Wconversion` are also unintelligent, as well the working group.

bagder:
Warnings are helpful, but they are toggles for a reason. I'm sure this option works great in some environments and I' don't understand what you're after here but you cannot deny that enforcing this option on a code base like curl is invasive to a degree that does not improve the source code.

That's why I'm saying anything like **thou must use this option** is a bad idea.

vszakats:
I'm not after anything, just spent a couple of full weeks trying to fix that in curl, only to throw away most of it (even some of the good parts unfortunately, in `lib`). To this day I'm confused as to why such a bad advice can make into a prominent best-practice guide, and remain there. And wonder what makes one codebase fit for that advice and unfit for others. Such discussion would have been enlightening, to at least learn something out of this.

At the end nobody learned anything (except that you think I'm unintelligent, which makes it for a rather fantastic ending of a long day).

bagder:
Viktor, I don't think you are unintelligent and I never said so.

Read what I wrote again; I said that it is wrong to **insist on -Wsign-conversion for real, because it is not intelligent**.

First, we need to separate persons from actions. I never said anyone *person* was stupid, but such an *act* would be silly. Then, I have not seen you insist on this so this does not even point to an action done by you.

Your PR trying to enable this option was a great effort. Unfortunately it could not be merged in the end. It's sad when we spend time and effort on something which does not end up getting used, but it's not entirely pointless because we learned a thing or two.

> To this day I'm confused as to why such a bad advice can make it into a prominent best-practice guide

Me too, but as I was partially involved in discussing details in that guide with OpenSSF I know that it was made by just a random bunch of people. It was not done by carefully asking around old open source C projects or even long-time C programmers. Actually, simply given a question about this I would probably also easily have said that the option is a good idea. Before we tried hard to actually enforce it and saw what kind of workarounds we would have to do in order to satisfy it.

vszakats:
@bagder: Thanks for clarifying, it's somewhat of a relief. Any kind of feedback, especially positive is rather hard to come by in volunteer work, and there're those that are difficult to assess or easy to mis-read. Yesterday I was also laughted at by a vcpkg maintainer over here in curl, and bashed by another person; for work taking 2-shift days and sometimes weeks. It truly makes me second-guessing my own intelligence.

Back to the warning, besides the controversial best-practice promoted by OpenSSF, what also keeps me wondering is that this warning _comes by default_ with the otherwise useful and widely used `-Wconversion` option [^1]. Meaning, it's enabled by default in possibly quite a lot of projects. Unless they also manually suppress it like curl does (which is of course possible, I haven't made statistics, after realizing/remembering the default status too late into the task).

Thirdly, in curl there is a long-time warning suppression effort called `warnless`. I may be totally mistaken by it's purpose or goal, but judging by the name and what it does, it seems like something similar to what my dropped/rejected effort was trying to do. If so, it either could be extended onto other cases, or dropped, if it's just unwanted noise or obsolete. Either way, it'd be useful (IMO) to touch this briefly to learn what its goal is and how does it (or doesn't) relate to sign conversions.

[^1]: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wconversion

bagder:
> Meaning, it's enabled by default in possibly quite a lot of projects

Maybe it is. Maybe it isn't. I'm not aware of anyone ever have tried to investigate what warning options that are in effective use in widely portable C programs. Also: I'm not sure it would change our situation even if it turned out to be used a lot.

> there is a long-time warning suppression effort called warnless. I may be totally mistaken by it's purpose or goal, but judging by the name and what it does, it seems like something similar to what my dropped/rejected effort was trying to do.

Yes I believe so. That effort was spear-headed by long-time absent Yang Tse.

> If so, it either could be extended onto other cases, or dropped, if it's just unwanted noise or obsolete.

Like everything else in the source I'm sure it can be enhanced or change into something else, sure. I suspect *dropping* it completely would just introduce a lot of typecasts instead and I don't think that's an improvement.

> it'd be useful (IMO) to touch this briefly to learn what its goal is and how does it (or doesn't) relate to sign conversions.

warnless.h provides functions and macros for type conversions (that should not cause warnings).

vpetersson:
> We removed the plethora of badges back in [7399fa5](https://github.com/curl/curl/commit/7399fa5b05ce897277b8713c659c939d068570a4). I'm not sure we are ready to restart this.
> 
> I think just providing the info in words and a prominent link might be sufficient.

Fair fair! Didn't know about the badge removal.

I've added it in text to `SECURITY.md` instead which might be more appropriate.

SECURITY.md
```diff
@@ -14,3 +14,14 @@ If you have found or just suspect a security problem somewhere in curl or
 libcurl, report it on [HackerOne](https://hackerone.com/curl).
 
 We treat security issues with confidentiality until controlled and disclosed responsibly.
+
+## OpenSSF Scorecard
+
+curl has earned Gold status on the OpenSSF Scorecard, reflecting its adherence to
```
bagder, line 20: The [OpenSSF scorecard](https://github.com/ossf/scorecard) is different from the [OpenSSF best practices](https://www.bestpractices.de). The link you provide here is for the best practices badge. Worth getting the terminology right I think!

vpetersson: My bad! Yes, Scorecard is something different. Got it confused in my head. Pushed a fix.

bagder (approved):

bagder:
Thanks!

PR #14285, opened 2024-07-27
cmake: detect `nettle` when building with GnuTLS

vszakats:
`nettle` is a direct dependency of curl, when building with GnuTLS.
Add a new `Find` module to detect it.

Also:
- GHA/macos: drop `nettle` hack no longer necessary.
- add `nettle` to `libcurl.pc`.
- also add `nettle` to `libcurl.pc` in autotools builds.

Follow-up to 781242ffa44a9f9b95b6da5ac5a1bf6372ec6257 #11967
Closes #14285

PR #14320, opened 2024-07-31
dist: add missing `FindNettle.cmake`

vszakats:
Follow-up to 669ce42275635dc1f881dab3dfc9a55c9ab49b21 #14285
Reported-by: Christoph Reiter
Bug: https://github.com/curl/curl/pull/14285#issuecomment-2259880050
Closes #14320

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -43,6 +43,7 @@ CMAKE_DIST =                                    \
  CMake/FindNGHTTP2.cmake                        \
  CMake/FindNGHTTP3.cmake                        \
  CMake/FindNGTCP2.cmake                         \
+ CMake/FindNettle.cmake                         \
  CMake/FindQUICHE.cmake                         \
  CMake/FindWolfSSL.cmake                        \
  CMake/FindZstd.cmake                           \


vszakats:
I wonder how this could be caught in CI. There is a test, but I guess we'd need separate ones to touch each `Find*` module, which is then just as easy to forget about as adding it to `/Makefile.am`.

bagder:
Unless we can write a parser that figures out what modules we use and then check that they are all present...

talregev:
You can create test that curl compile with other library/program (curl is a 3rd party) and check this things.

talregev:
Similar to test, but test have different cmake that don't depends on curl cmake.

vszakats:
Here's an attempt to test this in CI (and locally with the `-DENABLE_DIST_TEST=ON` option): #14323

bagder:
> You can create test that curl compile with other library/program (curl is a 3rd party) and check this things.

sure, but then we would need to make builds (from a dist tarball) for all dependencies that use a cmake module

talregev:
Maybe the bug can reproduce without the distro tar ball, with simple make install, and other cmake will try to find curl and it dependencies.

talregev:
I think we talking here about more errors that can found with cmake builds and make them automatic.

PR #14321, opened 2024-07-31
libcurl.pc: add `Cflags.private`

vszakats:
Ref: https://manpages.debian.org/bookworm/pkgconf/pc.5.en.html#PROPERTY_KEYWORDS
Ref: https://www.msys2.org/docs/pkgconfig/#cflagsprivate-static-libraries

Closes #14321

---

/cc @lazka


vszakats:
@lazka: Can you review this, if it's any good?

vszakats:
Could replace downstream patch:
https://github.com/msys2/MINGW-packages/blob/731e4faa16e3ce35e64700f833a29a82d4460036/mingw-w64-curl/0004-more-static-fixes.patch

lazka:
lgtm, thanks!

PR #14323, opened 2024-07-31
cmake: distcheck for files in CMake subdir

vszakats:
- add CMake option to verify if the `CMake/*.cmake`, `CMake/*.in` files
  are listed as distributable in autotools' `EXTRA_DIST`. The check can
  be enabled with `-DENABLE_DIST_TEST=ON` CMake option.

- add CI job to that effect.

Ref: #14320
Closes #14323

---

- [x] rebase on #14320

`IN_LIST` is CMake 3.3: https://cmake.org/cmake/help/latest/command/if.html#in-list

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -93,6 +93,8 @@ DIST_SUBDIRS = $(SUBDIRS) tests packages scripts include docs
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libcurl.pc
 
+### USE SIMPLE LIST ASSIGMENTS ABOVE THIS LINE ### DO NOT DELETE
+
 # List of files required to generate VC IDE .dsp, .vcproj and .vcxproj files
 include lib/Makefile.inc
 include src/Makefile.inc
--- a/.github/workflows/distcheck.yml
+++ b/.github/workflows/distcheck.yml
@@ -129,3 +129,13 @@ jobs:
           cmake -B build -DCURL_WERROR=ON
           make -C build -j5
         name: 'verify out-of-tree cmake build'
+
+  verify-cmake-test-dist:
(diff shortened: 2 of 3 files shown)


vszakats:
Fail check successful:
```
CMake Error at CMake/CMakeLists.txt:41 (message):
  Source file missing from the CMake/Makefile.inc CMAKE_DIST list:
  FindNettle.cmake
```
https://github.com/curl/curl/actions/runs/10178081847/job/28150986631?pr=14323

Makefile.am
```diff
@@ -26,32 +26,7 @@ AUTOMAKE_OPTIONS = foreign
...
- CMake/FindQUICHE.cmake                         \
- CMake/FindWolfSSL.cmake                        \
- CMake/FindZstd.cmake                           \
- CMake/Macros.cmake                             \
- CMake/OtherTests.cmake                         \
- CMake/PickyWarnings.cmake                      \
- CMake/Platforms/WindowsCache.cmake             \
- CMake/Utilities.cmake                          \
- CMakeLists.txt
+CMAKE_DIST = CMakeLists.txt
```
bagder, line 29: We do use `.inc` extension for other make include files, maybe that would make sense here too?

bagder: Ah no sorry, it's not an include file. Strike this!

vszakats: There are other problems. I'm adjusting back not to add an extra automake directory (I don't know what I'm doing and it doesn't work).

vszakats:
Cut it back a little. This adds an extra `.inc` file to the root, and this is disturbing.

vszakats:
Now without the extra file, with a small, hopefully non-offensive hack.

PR #14329, opened 2024-07-31
GHA/windows: Remove VCPKG bin path in MSVC jobs

talregev:
Remove vcpkg bin path from PATH

- The path is wrong, because we compile on debug, and we are using the release bin path.
- The path is not needed, cmake curl copy the needed dlls to the compilation cmake folder where the curl exe is found.


vszakats (approved):

vszakats:
I suggest starting the title with `GHA/windows: ` (or `GHA: ` if the context makes it clear it's Windows) and try keeping the length below 50-ish chars.

talregev:
I hope this is better. Thank you for your suggestion.

vszakats:
Thank you @talregev, merged!

PR #14331, opened 2024-08-01
CHANGES: rename to CHANGES.md, no longer generated

bagder:
- scripts/log2changes.pl was not included in release tarballs, which broke reproducible builds

- since log2changes uses git to generate the contents, it makes it difficult to generate the same contents later (it would need to be fixed)

- the CHANGES file has outlived its purpose. the main changes are in the RELEASE-NOTES, the rest are better tracked directly using git or on GitHub


icing:
Since CHANGES is such a standard location to look for, should we include one that say "look at RELEASE-NOTES"?

bagder:
> Since CHANGES is such a standard location to look for, should we include one that say "look at RELEASE-NOTES"?

A reasonable note to add, yes.

bagder:
@vszakats The curl-for-win jobs fail here because it seems they access `./CHANGES`  which I am now renaming to `CHANGES.md` but I am also making it less relevant. Maybe we should just skip that now in curl-for-win?

vszakats:
I did few minutes after you opening the PR, but forgot it's also in the distro pack, thanks for the ping!

Should be fine now after https://github.com/curl/curl-for-win/commit/fa22468c855195a97bc2c3699f78a74ab574b1cc.

bagder:
Thanks!

CHANGES.md
```diff
@@ -0,0 +1,12 @@
...
+
+SPDX-License-Identifier: curl
+-->
+
+In a release tarball, check the RELEASES-NOTES file for what was done in the
+most recent release. In a git check-out, that file mentions changes that have
+been done since the previous release.
+
+See the online [changelog](https://curl.se/changes.html) for the edited and
+human readable version of what has changed in different curl releases.
```
vszakats, line 12: Wouldn't it be useful to add the link above also to `RELEASE-NOTES`?

bagder: Yeah probably. I fix.

vszakats: Sorry, what I meant to suggest was to add the <https://curl.se/changes.html> link into the `RELEASE-NOTES` text.

It's a useful link, and e.g. in curl-for-win I stop distributing `CHANGES.md`, so the link will not be exposed, but in `RELEASE-NOTES` it would be!

(This update is also good regardless)

bagder: Ah ok. Yes, we can add a link to the RELEASE-NOTES, but I think I'll do that in a separate PR since I also need to make sure some other scripts handle that proper.

bagder: The GHA job for checking links did not like the markdown link to a local file anyway so I will revert that.

PR #14332, opened 2024-08-01
src: fix linking with nghttp2 when USE_EXPLICIT_LIB_DEPS isn't set

chouquette:
This PR fixes linking error when attempting to build libcurl with `--with-nghttp2`

While other dependencies are explicitly provided, `LIB_H2` isn't, which causes the build to fail with
```
/opt/toolchains/bin/../lib/gcc/x86_64-unknown-linux-gnu/11.4.0/../../../../x86_64-unknown-linux-gnu/bin/ld.bfd: ../lib/.libs/libcurl.so: undefined reference to `nghttp2_session_consume'
```
and a bunch of other missing symbols from `libnghttp2`

Full disclosure, I'm not 100% sure this is the correct fix. For starters, I'm surprised this was unnoticed given this was introduced many years ago.
I had this issue while attempting to bump a toolchain from a quite old gcc to a more recent one. I suspect the linker behavior changed and it's not defaulting to implicitly linking with the entries in `DT_NEEDED`, but this is rather blurry to me.

Thanks in advance for your review!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2816,6 +2816,7 @@ if test X"$want_nghttp2" != Xno; then
   LDFLAGS="$LDFLAGS $LD_H2"
   CPPFLAGS="$CPPFLAGS $CPP_H2"
   LIBS="$LIB_H2 $LIBS"
+  AC_SUBST(LIB_H2)
 
   # use nghttp2_session_get_stream_local_window_size to require nghttp2
   # >= 1.15.0
--- a/configure.ac
+++ b/configure.ac
@@ -2816,7 +2816,6 @@ if test X"$want_nghttp2" != Xno; then
   LDFLAGS="$LDFLAGS $LD_H2"
   CPPFLAGS="$CPPFLAGS $CPP_H2"
   LIBS="$LIB_H2 $LIBS"
-  AC_SUBST(LIB_H2)
 
   # use nghttp2_session_get_stream_local_window_size to require nghttp2
   # >= 1.15.0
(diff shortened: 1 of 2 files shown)


chouquette:
I probably should also have included this line which is relevant:
```
/opt/toolchains/bin/../lib/gcc/x86_64-unknown-linux-gnu/11.4.0/../../../../x86_64-unknown-linux-gnu/bin/ld.bfd: warning: libnghttp2.so.14, needed by ../lib/.libs/libcurl.so, not fou
nd (try using -rpath or -rpath-link)
```
And indeed including -rpath-link fixes the issue, which explains why this isn't more widespread as I'm using a custom rpath in my build.

It still feels like it makes sense to include `-lnghttp2` to the link line, but I'll let you judge!

bagder:
> For starters, I'm surprised this was unnoticed given this was introduced many years ago.

Well, why do you need explicit libs set to begin with? It looks like you build on Linux and if so, it seems you do something unusual?

chouquette:
The honest answer is that I don't know. This flag value ultimately comes from libtool's `link_all_deplibs` but I can't say I understand how it is set.

The rationale here is mostly that some dependencies are explicitly provided, so it felt sensible to do the same for `libnghttp2`

chouquette:
>  It looks like you build on Linux

yes

>  it seems you do something unusual?

Besides providing an explicit rpath to all our built DSOs, not much else

bagder:
The entire `USE_EXPLICIT_LIB_DEPS` has been removed in master. Can you rebuild a clean build from master and see if you still need to change anything and if so, what?

PR #14333, opened 2024-08-01
GHA: scan git repository and detect unvetted binary files

bagder:
The only binary-looking files that are accepted in the git repository need to match the sha256sum file .github/scripts/binarycheck.sums.

This is done to make sure that no one has planted any hidden (encrypted) potentially dangerous payload in the tree.

### DIFF
--- a/.github/scripts/binarycheck.pl
+++ b/.github/scripts/binarycheck.pl
@@ -1,0 +1,111 @@
+#!/usr/bin/env perl
+#***************************************************************************
+#                                  _   _ ____  _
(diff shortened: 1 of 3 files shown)


dfandrich:
A laudable goal, but there are a lot of places where base64 encoded binaries
can and do live right now. Someone who wants an arbitrary binary can easily
bypass this check that way.

bagder:
Sure, but it's about raising reasonable "guard rails" here - not fooling anyone into saying that we can automatically detect any malicious payload. We don't accept large base encoded chunks either, but we reject those by manual review.

PR #14350, opened 2024-08-02
verify-release is a new script to "verify" a release tarball

bagder:
Verify in this case means making sure that rebuilding the tarball generates the **exact same** output.

The verify-release script requires 754acd1a9dc6, so it does not work on past curl release tarballs.


vszakats:
Do we still need the old `maketgz reproducibility test` job after this?

bagder:
> Do we need the old `maketgz reproducibility test` step after this?

Ah no. This new one is a more complete test that succeeds the old one.

Issue #14352, opened 2024-08-02
GnuTLS certificates can be bigger than `MAX_X509_CERT`


sergiodj:
### I did this

We've received the following bug in Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076800

We are linking curl against GnuTLS, and the following error started happening:

```
# curl -w '%{size_header}' https://feeds2.feedburner.com/OreillyConferencesOSCON
curl: (100) Failed extracting certificate chain
```

### I expected the following

When using curl linked against OpenSSL, I see:

```
$ curl -w '%{size_header}' https://feeds2.feedburner.com/OreillyConferencesOSCON
<?xml version="1.0" encoding="UTF-8"?>
<rss xmlns:georss="http://www.georss.org/georss" xmlns:twitter="http://api.twitter.com" version="2.0" xmlns:atom="http://www.w3.org/2005/A
tom">
  <channel>
    <title>Twitter / oscon</title> 
    <link>http://twitter.com/oscon</link>
    <atom:link rel="self" href="http://api.twitter.com/1/statuses/user_timeline.rss?screen_name=oscon" type="application/rss+xml"/>
    <description>Twitter updates from O'Reilly OSCON / oscon.</description>
    <language>en-us</language>
    <ttl>40</ttl>
  <item>
    <title>oscon: Easily Invoke Common Protocols with Twisted - Spin up Python-friendly services with 0 lines of code http://t.co/29oTkk0i
sW</title>
...
  </channel>
</rss>
1494
```

The output above has been truncated for readability.

### curl/libcurl version

curl 8.9.1

### operating system

Debian unstable

sergiodj:
After some investigation, I noticed that the problem is happening on `Curl_extract_certinfo`.  Specifically, the following excerpts:

https://github.com/curl/curl/blob/45246ebca159f3df1eda172956c0b938240978d6/lib/vtls/x509asn1.c#L1096

https://github.com/curl/curl/blob/45246ebca159f3df1eda172956c0b938240978d6/lib/vtls/x509asn1.c#L1235-L1252

The certificate we get from the website has a size that's much bigger than `MAX_X509_STR`.

icing:
Can reproduce. Analyzing the cause.

icing:
I propose #14354 as fix for this. Could you verify?

PR #14353, opened 2024-08-02
x509asn1: initialize dynbuf with MAX_X509_CERT

sergiodj:
Ref.: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076800

When using GnuTLS we may see certificates whose sizes are bigger than 10000, so use `MAX_X509_CERT` as the maximum `dynbuf` size (instead of `MAX_X509_STR`).

Closes: #14352


sergiodj:
#14354 is more complete; closing this one.

PR #14354, opened 2024-08-02
x509 certificate information, raise limit

icing:
Raise the limit for certificate information from 10 thousand to 100 thousand bytes. Certificates can be larger than 10k.

Change the infof() debug output to add '...' at the end when the max limit it can handle is exceeded.

Refs #14352


sergiodj:
I'd filed #14353 earlier (right after I filed the issue), but I forgot to link both.

PR #14357, opened 2024-08-02
cmake: drop no-op `tests/data/CMakeLists.txt`

vszakats:
It's a no-op since
d162fca69a22d199c0369c5ce3d66a9531752ba2 #9333 (2022-08-18).

Also revert 476499c75c83ffa3d65da398b5759a5f2fdb980d that is
no longer necessary: move `Makefile.inc` back into `Makefile.am`.

Closes #14357


bagder:
I *really* wished you had not done this. This adds non-trivial work for like 20 pending PRs...

vszakats:
Can't really do any good over here! I'm sorry.

bagder:
No need to be fatalistic. Just me reminding you and others that central files we update in just about every commit probably need some syncing and discussion before we rename to reduce friction.

PR #14360, opened 2024-08-02
tidy-up: OS names

vszakats:
Use these words and casing more consistently across text, comments and
one curl tool output:
AIX, ALPN, ANSI, BSD, Cygwin, Darwin, FreeBSD, GitHub, HP-UX, Linux,
macOS, MS-DOS, MSYS, MinGW, NTLM, POSIX, Solaris, UNIX, Unix, Unicode,
WINE, WebDAV, Win32, winbind, WinIDN, Windows, Windows CE, Winsock.

Mostly OS names and a few more.

Also a couple of other minor text fixups.

Closes #14360

---

- [x] `win32` → `Win32`/`Windows`
- [x] `Win32` → `Windows` (in tests and few more places)


vszakats:
I used `macOS` when talking about `macOS 10.7 and later` even though 10.7 wasn't called macOS yet. Existing usage wasn't consistent. Feel free to change it. I left 'OS X' as-is in `HISTORY.md`.

Windows is often referred to as `win32` and `Win32`, which is stricly speaking not the same, but practically almost always is. I left these as-is anyway in almost all cases to avoid messing it up too. Also some internal flags and API use `win32`.

PR #14370, opened 2024-08-04
aws_sigv4: Fix ordering for headers with same prefix in the canonical request

austinmoore-:
# Problem

AWS SigV4 signing requires headers to be lexicographically ordered by name. The current implementation uses `strcmp` on the full header, leading to incorrect ordering when header names have identical prefixes.

I found two scenarios where curl users could be affected:
1. S3 object uploads with [user-defined metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html) (prefixed by `x-amz-meta-`).
2. [SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html#specifying-s3-c-encryption) operations using `x-amz-server-side-encryption-customer-key` and `x-amz-server-side-encryption-customer-key-md5` headers.

Example of the issue:
```bash
curl -X PUT --verbose --silent \
    --aws-sigv4 "aws:amz:us-east-1:s3" --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY" \
    -H "x-amz-meta-test-two: Test2" \
    -H "x-amz-meta-test: Test" \
    --data-binary @testfile \
    "https://examplebucket.s3.amazonaws.com/testfile"
```

This results in `x-amz-meta-test-two` incorrectly preceding `x-amz-meta-test` in the canonical request, causing S3 to return a `SignatureDoesNotMatch` error.

# Fix

This PR resolves the issue by:
1. Using `strncmp` with the length of the shorter header name for comparison for the canonical request (actual request headers are left in the specified order).
2. If the compared substrings match, the shorter header name is compared as "less than" the longer one.

The fix has been verified with a few real S3 requests, along with the added libcurl test.

### DIFF
--- a/lib/http_aws_sigv4.c
+++ b/lib/http_aws_sigv4.c
@@ -129,6 +129,37 @@ static void trim_headers(struct curl_slist *head)
 /* string been x-PROVIDER-date:TIMESTAMP, I need +1 for ':' */
 #define DATE_FULL_HDR_LEN (DATE_HDR_KEY_LEN + TIMESTAMP_SIZE + 1)
 
+/* alphabetically compare two headers by their name, expecting
+   headers to use ':' at this point */
+static int compare_header_names(const char *a, const char *b)
+{
+  const char *colon_a;
+  const char *colon_b;
+  size_t len_a;
+  size_t len_b;
+  size_t min_len;
+  int cmp;
+
+  colon_a = strchr(a, ':');
+  colon_b = strchr(b, ':');
+
+  DEBUGASSERT(colon_a);
+  DEBUGASSERT(colon_b);
+
+  len_a = colon_a ? (size_t)(colon_a - a) : strlen(a);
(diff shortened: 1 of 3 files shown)


bagder (approved):

bagder:
Thanks!

jdelvare:
I took a look at the test case and I don't think it actually tests what the commit fixed.
In the example given in the commit message, the 2 headers are in lowercase, so one is indeed a prefix of the other, which is the case which needs special care. However in the test case, one header uses uppercase letters, the other doesn't, so the shorter one isn't considered to be a prefix of the longer one (sort order is case-sensitive according to a comment in the code).
So what the test case actually tests is whether the ordering is still case-sensitive as it is supposed to be, and already was before the fix. In order to also test whether the fix works, you would have to add a third header, of which one of the other two would be a prefix. For example:
`x-amz-meta-test-three: test3`

PR #14377, opened 2024-08-04
Point out DOH server IP pinning

duxsco:
The IP address in the example is taken out of [RFC 5737](https://datatracker.ietf.org/doc/html/rfc5737).


bagder (approved):

bagder:
Thanks!

PR #14379, opened 2024-08-04
configure: fail if PSL is not disabled but not found

### LINKED ISSUE #14373: configure no longer errors when libpsl is not found and not explicitly disabled
### I did this

```
./configure --without-ssl
```

### I expected the following

```
configure: error: libpsl was not found
```

### curl/libcurl version

8.9.1

### operating system

macOS 12

bagder:
Regression since 9b3f67e (shipped in 8.7.0)
Reported-by: Ryan Carsten Schmidt
Fixes #14373

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -449,7 +449,7 @@ jobs:
       - name: 'install autotools'
         if: ${{ matrix.build == 'autotools' }}
         run: |
-          echo autoconf automake libtool openssl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
+          echo autoconf automake libtool openssl libpsl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
           while [[ $? == 0 ]]; do for i in 1 2 3; do brew update && brew bundle install --no-lock --file /tmp/Brewfile && break 2 || { echo Error: wait to try again; sleep 10; } done; false Too many retries; done
 
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -449,7 +449,7 @@ jobs:
       - name: 'install autotools'
         if: ${{ matrix.build == 'autotools' }}
         run: |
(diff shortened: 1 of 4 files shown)


bagder:
It is quite hard to find the configure line in these failing builds. We should probably do something to make it easier. Most failures seem to be legitimate: they don't find libpsl.

bagder:
@vszakats I'm getting lost in these workflow files. Can you help me figure out where we need to add libpsl to make the configure runs work?

.github/workflows/macos.yml
```diff
@@ -449,7 +449,7 @@ jobs:
       - name: 'install autotools'
         if: ${{ matrix.build == 'autotools' }}
         run: |
-          echo autoconf automake libtool openssl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
+          echo autoconf automake libtool openssl libpsl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
```
vszakats, line 452:
```suggestion
          echo autoconf automake libtool openssl pkg-config libpsl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
```
Perhaps this, based on how it's done in the autotools and cmake jobs.

vszakats: Nope, works without `pkg-config`:
https://github.com/curl/curl/actions/runs/10237324011/job/28320483945

vszakats, line 452:
```suggestion
          echo autoconf automake libtool openssl | xargs -Ix -n1 echo brew '"x"' > /tmp/Brewfile
```

It works with autotools now, but cmake doesn't pick it up and it's intentional.
The 'combinations' jobs are meant to be minimal and match between autotools and cmake. So I think the correct solution is to use `--without-libpsl` here.

.github/workflows/macos.yml
```diff
@@ -586,6 +586,7 @@ jobs:
               --disable-dependency-tracking \
               --disable-docs --disable-manual \
               --without-nghttp2 --without-libidn2 \
+              --with-libpsl=$(brew --prefix libpsl) \
```
vszakats, line 589:
```suggestion
              --without-libpsl \
```

PR #14385, opened 2024-08-04
wolfssl: Add CURLOPT_TLS13_CIPHERS support

jan2000:
Bring setting ciphers with WolfSSL in line with other SSL backends, to make the curl interface more consistent across the backends.

Now the tls1.3 ciphers are set with the --tls13-ciphers option, when not set the default tls1.3 ciphers are used. The tls1.2 (1.1, 1.0) ciphers are set with the --ciphers option, when not set the default tls1.2 ciphers are used. The ciphers available for the connection are now a union of the tls1.3 and tls1.2 ciphers.

This changes the behaviour for WolfSSL when --ciphers is set, but --tls13-ciphers is not set. Now the ciphers set with --ciphers are combined with the default tls1.3 ciphers, whereas before solely the ciphers of --ciphers were used.

Thus before when no tls1.3 ciphers were specified in --ciphers, tls1.3 was completely disabled. This might not be what the user expected, especially as this does not happen with OpenSSL.

### DIFF
--- a/lib/vtls/wolfssl.c
+++ b/lib/vtls/wolfssl.c
@@ -623,6 +623,36 @@ CURLcode Curl_wssl_setup_x509_store(struct Curl_cfilter *cf,
   return result;
 }
 
+#ifdef WOLFSSL_TLS13
+static size_t
+wssl_get_default_ciphers(bool tls13, char *buf, size_t size)
+{
+  size_t len = 0;
+  char *term = buf;
+  int i;
+  char *str;
+  size_t n;
+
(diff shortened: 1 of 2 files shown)


icing (approved):

bagder:
Can you check the merge conflict here? (Maybe due to the #14384 merge?)

jan2000:
Yes I was afraid that would conflict. That is why the doc updates are also in a seperate PR. I have rebased.

bagder:
Thanks!

PR #14390, opened 2024-08-04
sigpipe: init the struct so that first apply ignores

### LINKED ISSUE #14344: Starting with 8.9.1, SIGPIPE leaks in some cases
### I did this

After https://github.com/curl/curl/pull/14296, curl has started to leak SIGPIPE in some cases. I‘ve noticed this since the test suite I wrote for our in-house curl-wrapper now sometimes exits which SIGPIPE (I‘d say about 50% of the time). I‘ll try to post the last curl invocation tomorrow.

### I expected the following

No SIGPIPEs to leak to the program (or at least no un-announced change to the curl‘s behavior)

### curl/libcurl version

curl 8.9.1

### operating system

Ubuntu 20.04 (using the Conan package manager, thus using OpenSSL 3.x)

bagder:
Initializes 'no_signal' to TRUE, so that a call to sigpipe_apply() after init ignores the signal (unless CURLOPT_NOSIGNAL) is set.

I have read the existing code multiple times now and I think it gets the initial state reversed this missing to ignore.

Regression from 17e6f06ea37136c36d27

Reported-by: Rasmus Thomsen
Fixes #14344

### DIFF
--- a/lib/sigpipe.h
+++ b/lib/sigpipe.h
@@ -39,6 +39,7 @@ struct sigpipe_ignore {
 static void sigpipe_init(struct sigpipe_ignore *ig)
 {
   memset(ig, 0, sizeof(*ig));
+  ig->no_signal = TRUE;
 }
 
 /*


djdisodo:
tested (debian sid riscv64) and transmission-daemon works

Cogitri:
Works for us as well, thanks! 👍

PR #14394, opened 2024-08-05
Use NGHTTP3 prefix instead of NGTCP2 for errors in h3 callbacks.

moritzbuhl:
To my understanding the functions are nghttp3 callbacks and should therefore return `NGHTTP3_ERR_CALLBACK_FAILURE`.
However, this is not critical as the nghttp3 documentation states for all callbacks: (https://nghttp2.org/nghttp3/types.html#c.nghttp3_stop_sending)

>The implementation of this callback must return 0 if it succeeds. Returning [NGHTTP3_ERR_CALLBACK_FAILURE](https://nghttp2.org/nghttp3/macros.html#c.NGHTTP3_ERR_CALLBACK_FAILURE) will return to the caller immediately. Any values other than 0 is treated as [NGHTTP3_ERR_CALLBACK_FAILURE](https://nghttp2.org/nghttp3/macros.html#c.NGHTTP3_ERR_CALLBACK_FAILURE).

For `nghttp3_strerror`, a nghttp3 function sets rc, so I expect a nghttp3 function to transform the error correctly.
The values for NGHTTP3_ERR_... and NGTCP2_ERR_... are not the same, so I expect this to be more relevant.

### DIFF
--- a/lib/vquic/curl_ngtcp2.c
+++ b/lib/vquic/curl_ngtcp2.c
@@ -1049,7 +1049,7 @@ static int cb_h3_stop_sending(nghttp3_conn *conn, int64_t stream_id,
   rv = ngtcp2_conn_shutdown_stream_read(ctx->qconn, 0, stream_id,
                                         app_error_code);
   if(rv && rv != NGTCP2_ERR_STREAM_NOT_FOUND) {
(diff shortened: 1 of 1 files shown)


bagder (approved):
Nice catch!

bagder:
Thanks!

PR #14387, opened 2024-08-04
KNOWN_BUGS: macidn and winidn test problems

bagder:
Closes #14176

PR #14398, opened 2024-08-05
GHA/windows: ignore test 165 in Schannel MSVC job

talregev:
Schannel MSVC job: Ignore test 165 when -DUSE_WIN32_IDN=ON

After this ci is fail on Schannel
https://github.com/curl/curl/pull/14387

https://github.com/curl/curl/actions/runs/10251838878/job/28360790340?pr=14383
```
  TESTDONE: 1054 tests out of 1056 reported OK: 99%
  FAIL 165: 'HTTP over proxy with IDN host name' HTTP, HTTP GET, HTTP proxy, IDN
  TESTFAIL: These test cases failed: 165
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -481,6 +481,7 @@ jobs:
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP !FTP ${{ matrix.tflags }}'
           if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
+            TFLAGS+=' ~165'
             if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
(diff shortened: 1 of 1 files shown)


bagder:
I removed this the other day. Test 165 is now in `DISABLED` for `WinIDN` so should be ignored for these runs by default. Clearly something makes that fail here, but that's what we need to fix.

talregev:
> I removed this the other day. Test 165 is now in `DISABLED` for `WinIDN` so should be ignored for these runs by default. Clearly something makes that fail here, but that's what we need to fix.

I understand. Do you want me to close this PR?

bagder:
Either that, or to figure out why 165 was not ignored and fix that.

vszakats:
To add: disabling AppleIDN tests fails too in GHA/macos.

talregev:
I see that you already add these on disable test file.
I am not sure who parse it and how.

talregev:
I don't mind to close this PR. Let me know if you know / can solve it.

PR #14400, opened 2024-08-05
mbedtls: add more informative logging

jmaggard10:
After TLS handshare, indicate which TLS version was negotiated in addition to the cipher in the handshake completed log message.

Also, dump certificate info prior to bailing on cert validation failure. When cert validation fails, it's often helpful to be able to see the cert info. We also need to split the cert dump into multiple log messages to avoid log truncation on certs with lots of data (such as many SANs).


jmaggard10:
This needs a rework, as it only works in some circumstances.

PR #14401, opened 2024-08-05
configure: detect AppleIDN

bagder:
Follow-up to add22feeef07858307


bagder:
The failed Windows CI job is a separate issue:  #14403

PR #14406, opened 2024-08-05
curl: warn on unsupported SSL options including --ciphers, --tls13-ciphers

jan2000:
In curl warn when trying to use an option that is not supported by the SSL backend.

In libcurl added SSLSUPP_CIPHER_LIST flag to be able to differentiate SSL backends that support CURLOPT_SSL_CIPHER_LIST.


jan2000:
@icing This is partly of what is mentioned in #14106

bagder (approved):

bagder:
Thanks!

Issue #14365, opened 2024-08-03
cURL [8.9.1] - Windows - Fails to build via CMake with mingw installed


danoli3:
### I did this

Built **cURL 8.9.1** with openFrameworks bash script build system using CMakeList for build system.
<img width="926" alt="image" src="https://github.com/user-attachments/assets/4a486c40-034e-4c2a-b752-c95715eef93d">

```
winnt.h(137,1): error C1189: #error:  No supported target architecture. [D:\a\apothecary\apothecary\apothecary\build\curl\build_vs_64\lib\libcurl_object.vcxproj]
```

- Target **Visual Studio 2022** 
- mingw64 installed on system : CMake Build Fails 
- Prior cURL versions were successful even with mingw installed.

See Operating System for GitHub Action for what mingw installed in setup.

```
uses: msys2/setup-msys2@v2
      with:
        update: true
        install: >-
          mingw-w64-x86_64-binutils
          mingw-w64-x86_64-clang
          mingw-w64-x86_64-gcc
          mingw-w64-x86_64-gcc-libs
          mingw-w64-x86_64-cmake
          mingw-w64-x86_64-gdb
          mingw-w64-x86_64-make
```

### CMake script:

```
 OPENSSL_ROOT="$LIBS_ROOT/openssl/"
        OPENSSL_INCLUDE_DIR="$LIBS_ROOT/openssl/include"
        OPENSSL_LIBRARY="$LIBS_ROOT/zlib/lib/$TYPE/$PLATFORM/openssl.lib"
        OPENSSL_LIBRARY="$LIBS_ROOT/zlib/lib/$TYPE/$PLATFORM/libcrypto.lib"
        ZLIB_ROOT="$LIBS_ROOT/zlib/"
        ZLIB_INCLUDE_DIR="$LIBS_ROOT/zlib/include"
        ZLIB_LIBRARY="$LIBS_ROOT/zlib/lib/$TYPE/$PLATFORM/zlib.lib"
        export PKG_CONFIG_PATH="/usr/local/lib/pkgconfig;${PKG_CONFIG_PATH};${OF_LIBS_OPENSSL}/lib/$TYPE/$PLATFORM;${ZLIB_ROOT}/lib/$TYPE/$PLATFORM"
        DEFS="-DLIBRARY_SUFFIX=${ARCH} \
            -DCMAKE_BUILD_TYPE=Release \
            -DCMAKE_C_STANDARD=${C_STANDARD} \
[...]

PR #14408, opened 2024-08-06
cmake: limit libidn2 `pkg-config` detection to `UNIX`

### LINKED ISSUE #14405: Windows CMake build fails with OpenSSL unresolved external symbols
### I did this

I compiled libcurl 8.9.1 on Windows 11 using cmake with the OpenSSL backend (version 3.3.1, though I also observed the issue with 3.2.0). 

Build command:
`cmake.exe .. -G "Visual Studio 17 2022" -A x64  -D CMAKE_INSTALL_PREFIX="C:/Users/micasnyd/dev/clamav-deps/x64"  -D BUILD_LIBCURL_DOCS=OFF  -D BUILD_SHARED_LIBS=ON  -D CURL_USE_OPENSSL=ON  -D OPENSSL_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D LIB_EAY_RELEASE="C:/Users/micasnyd/dev/clamav-deps/x64/lib/libcrypto.lib"  -D SSL_EAY_RELEASE="C:/Users/micasnyd/dev/clamav-deps/x64/lib/libssl.lib"  -D ZLIB_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D ZLIB_LIBRARY_RELEASE="C:/Users/micasnyd/dev/clamav-deps/x64/lib/zlibstatic.lib"  -D LIBSSH2_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D LIBSSH2_LIBRARY="C:/Users/micasnyd/dev/clamav-deps/x64/lib/libssh2.lib"  -D USE_NGHTTP2=ON  -D NGHTTP2_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D NGHTTP2_LIBRARY="C:/Users/micasnyd/dev/clamav-deps/x64/lib/nghttp2.lib"  -D BUILD_CURL_EXE=OFF`

I encountered a build failure due to unresolved external symbols. 
```
cmake.exe .. -G "Visual Studio 17 2022" -A x64  -D CMAKE_INSTALL_PREFIX="C:/Users/micasnyd/dev/clamav-deps/x64"  -D BUILD_LIBCURL_DOCS=OFF  -D BUILD_SHARED_LIBS=ON  -D CURL_USE_OPENSSL=ON  -D OPENSSL_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D LIB_EAY_RELEASE="C:/Users/micasnyd/dev/clamav-deps/x64/lib/libcrypto.lib"  -D SSL_EAY_RELEASE="C:/Users/micasnyd/dev/clamav-deps/x64/lib/libssl.lib"  -D ZLIB_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D ZLIB_LIBRARY_RELEASE="C:/Users/micasnyd/dev/clamav-deps/x64/lib/zlibstatic.lib"  -D LIBSSH2_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D LIBSSH2_LIBRARY="C:/Users/micasnyd/dev/clamav-deps/x64/lib/libssh2.lib"  -D USE_NGHTTP2=ON  -D NGHTTP2_INCLUDE_DIR="C:/Users/micasnyd/dev/clamav-deps/x64/include"  -D NGHTTP2_LIBRARY="C:/Users/micasnyd/dev/clamav-deps/x64/lib/nghttp2.lib"  -D BUILD_CURL_EXE=OFF
-- Using CMake version 3.29.0
-- The C compiler identification is MSVC 19.39.33523.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.39.33519/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- curl version=[8.9.0]
-- Found Perl: C:/Strawberry/perl/bin/perl.exe (found version "5.38.2")
-- Found OpenSSL: C:/Users/micasnyd/dev/clamav-deps/x64/lib/libcrypto.lib (found version "3.3.1")
-- Looking for OPENSSL_IS_BORINGSSL
-- Looking for OPENSSL_IS_BORINGSSL - not found
-- Looking for OPENSSL_IS_AWSLC
-- Looking for OPENSSL_IS_AWSLC - not found
-- Found ZLIB: C:/Users/micasnyd/dev/clamav-deps/x64/lib/zlibstatic.lib (found version "1.3.1")
-- Looking for SSL_set0_wbio
-- Looking for SSL_set0_wbio - found
-- Looking for SSL_CTX_set_srp_username
-- Looking for SSL_CTX_set_srp_username - found
-- Found NGHTTP2: C:/Users/micasnyd/dev/clamav-deps/x64/lib/nghttp2.lib
-- Looking for idn2_lookup_ul in idn2
-- Looking for idn2_lookup_ul in idn2 - not found
-- Checking for module 'libidn2'
--   Found libidn2, version 2.1.1
-- Could NOT find LibPSL (missing: LIBPSL_LIBRARY LIBPSL_INCLUDE_DIR)
-- Found LibSSH2: C:/Users/micasnyd/dev/clamav-deps/x64/lib/libssh2.lib (found version "1.11.0")
-- Performing Test HAVE_WIN32_WINNT
-- Performing Test HAVE_WIN32_WINNT - Success
-- Found _WIN32_WINNT=0x0a00
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Check size of size_t
-- Check size of size_t - done
-- Check size of ssize_t
-- Check size of ssize_t - failed
-- Check size of long long
-- Check size of long long - done
-- Check size of long
-- Check size of long - done
-- Check size of int
-- Check size of int - done
-- Check size of __int64
-- Check size of __int64 - done
-- Check size of time_t
-- Check size of time_t - done
-- Check size of ADDRESS_FAMILY
-- Check size of ADDRESS_FAMILY - done
-- Check size of off_t
-- Check size of off_t - done
-- Check size of curl_off_t
-- Check size of curl_off_t - done
-- Check size of curl_socket_t
-- Check size of curl_socket_t - done
-- Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
-- Features: alt-svc AsynchDNS HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM SSL threadsafe TLS-SRP UnixSockets
-- Enabled SSL backends: OpenSSL
-- Configuring done (40.6s)
-- Generating done (4.6s)
-- Build files have been written to: C:/Users/micasnyd/dev/work/x64/curl-8.9.0/build
C:\Users\micasnyd\dev\work\x64\curl-8.9.0>cd build
C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build>CALL cmake.exe --build . --config Release
MSBuild version 17.9.8+b34f75857 for .NET Framework
1>Checking Build System
Generating curl-config.1
Building Custom Rule C:/Users/micasnyd/dev/work/x64/curl-8.9.0/docs/CMakeLists.txt
Generating mk-ca-bundle.1
Building Custom Rule C:/Users/micasnyd/dev/work/x64/curl-8.9.0/docs/CMakeLists.txt
Building Custom Rule C:/Users/micasnyd/dev/work/x64/curl-8.9.0/lib/CMakeLists.txt
altsvc.c
amigaos.c
asyn-ares.c
asyn-thread.c
base64.c
bufq.c
bufref.c
c-hyper.c
cf-h1-proxy.c
cf-h2-proxy.c
cf-haproxy.c
cf-https-connect.c
cf-socket.c
cfilters.c
conncache.c
connect.c
content_encoding.c
cookie.c
curl_addrinfo.c
curl_des.c
curl_endian.c
curl_fnmatch.c
curl_get_line.c
curl_gethostname.c
curl_gssapi.c
curl_memrchr.c
curl_multibyte.c
curl_ntlm_core.c
curl_path.c
curl_range.c
curl_rtmp.c
curl_sasl.c
curl_sha512_256.c
curl_sspi.c
curl_threads.c
curl_trc.c
cw-out.c
dict.c
doh.c
dynbuf.c
dynhds.c
easy.c
easygetopt.c
easyoptions.c
escape.c
file.c
fileinfo.c
fopen.c
formdata.c
ftp.c
ftplistparser.c
getenv.c
getinfo.c
gopher.c
hash.c
headers.c
hmac.c
hostasyn.c
hostip.c
hostip4.c
hostip6.c
hostsyn.c
hsts.c
http.c
http1.c
http2.c
http_aws_sigv4.c
http_chunks.c
http_digest.c
http_negotiate.c
http_ntlm.c
http_proxy.c
idn.c
if2ip.c
imap.c
inet_ntop.c
inet_pton.c
krb5.c
ldap.c
llist.c
macos.c
md4.c
md5.c
memdebug.c
mime.c
mprintf.c
mqtt.c
multi.c
netrc.c
nonblock.c
noproxy.c
openldap.c
parsedate.c
pingpong.c
pop3.c
progress.c
psl.c
rand.c
rename.c
request.c
rtsp.c
select.c
sendf.c
setopt.c
sha256.c
share.c
slist.c
smb.c
smtp.c
socketpair.c
socks.c
socks_gssapi.c
socks_sspi.c
speedcheck.c
splay.c
strcase.c
strdup.c
strerror.c
strtok.c
strtoofft.c
system_win32.c
telnet.c
tftp.c
timediff.c
timeval.c
transfer.c
url.c
urlapi.c
version.c
version_win32.c
warnless.c
ws.c
cleartext.c
cram.c
digest.c
digest_sspi.c
gsasl.c
krb5_gssapi.c
krb5_sspi.c
ntlm.c
ntlm_sspi.c
oauth2.c
spnego_gssapi.c
spnego_sspi.c
vauth.c
bearssl.c
cipher_suite.c
gtls.c
hostcheck.c
keylog.c
mbedtls.c
mbedtls_threadlock.c
openssl.c
rustls.c
schannel.c
schannel_verify.c
sectransp.c
vtls.c
wolfssl.c
x509asn1.c
curl_msh3.c
curl_ngtcp2.c
curl_osslq.c
curl_quiche.c
vquic.c
vquic-tls.c
libssh.c
libssh2.c
wolfssh.c
libcurl_object.vcxproj -> C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\libcurl_object.dir\Release\libcurl_object.lib
Building Custom Rule C:/Users/micasnyd/dev/work/x64/curl-8.9.0/lib/CMakeLists.txt
dllmain.c
Creating library C:/Users/micasnyd/dev/work/x64/curl-8.9.0/build/lib/Release/libcurl_imp.lib and object C:/Users/micasnyd/dev/work/x64/curl-8.9.0/build/lib/Release/libcurl_imp.exp
openssl.obj : error LNK2019: unresolved external symbol EVP_PKEY_id referenced in function Curl_ossl_certchain [C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\libcurl_shared.vcxproj]
openssl.obj : error LNK2019: unresolved external symbol EVP_PKEY_bits referenced in function Curl_oss_check_peer_cert [C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\libcurl_shared.vcxproj]
openssl.obj : error LNK2019: unresolved external symbol EVP_PKEY_security_bits referenced in function Curl_oss_check_peer_cert [C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\libcurl_shared.vcxproj]
openssl.obj : error LNK2019: unresolved external symbol SSL_get_peer_certificate referenced in function Curl_oss_check_peer_cert [C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\libcurl_shared.vcxproj]
C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\Release\libcurl.dll : fatal error LNK1120: 4 unresolved externals [C:\Users\micasnyd\dev\work\x64\curl-8.9.0\build\lib\libcurl_shared.vcxproj]
libcurl-8.9.0 x64 build failed!
Command:
cd build
CALL cmake.exe --build . --config Release
Exit code: 1
"make" script failed for x64 build
```

I think the unresolved symbols in question were renamed in OpenSSL 3.x and the original names were converted to macros for backwards compatibility. So I suspect there's a header issue of some sort, and the macros aren't being included. 

I checked with previous versions:
- 8.9.0: fails with this same issue
- 8.8.0: build succeeds

### I expected the following

build should complete successfully

### curl/libcurl version

curl 8.9.0

### operating system

Windows 11 x64

vszakats:
libidn2 is detected by default, which triggers a `pkg-config` detectio
attempt by default. This in turn may pick up libidn2 inadvertently from
the disk, and append the libidn2 header directory to the include path.
This header directory might contain incompatible system and/or component
headers, causing confusion and failed builds.

Some of these side-effects may be the result of an unknowningly
configured (or misconfigured) `pkg-config`. In another reported case,
it was hit by the `pkg-config` from Strawberry Perl. Until we
investigate the reasons and come up with a technique to avoid these
issues, limit `pkg-config` detection to UNIX platforms, like we already
do in `Find*` modules.

Notice that `-DCURL_USE_LIBSSH=ON`, `-DCURL_USE_GSASL=ON`, and
`-DCURL_USE_LIBUV=ON` options continue to have the above side-effects,
though these options are disabled by default.

Follow-up to f43adc2c4978f7f82a359e89186e58a31d17b0ad #14137
Reported-by: Micah Snyder
Fixes #14405
Closes #14408


MarcelRaad:
Does UNIX include MSYS? The documentation isn't clear about that, it only mentions CYGWIN. The only bad case I can think of is https://github.com/curl/curl/issues/14365, when targeting MSVC, so I think it would be good to have the pkg-config detection enabled for MSYS.

vszakats:
I've made PR to make these CMake flags transparent: #14417
→ Yes, MSYS is UNIX: https://github.com/curl/curl/actions/runs/10265251756/job/28401088397?pr=14417#step:16:43

Using `UNIX` here just applies the pattern already used in our `Find*` modules.
Once we know more about the nature of side-effects, we can update all of these
`pkg-config` uses with conditions giving the best result in most cases.
Including vcpkg, which also supports `pkg-config`.

#14365 looks like an explictly misconfigured env, though still awaiting feedback.
#14405 doesn't say about an explicit configuration, also awaiting feedback.

An earlier failed attempt to tackle this #14140. With no answers and adding to
the confusion. My concern is also cross-builds, which broke a CI elsewhere.

MarcelRaad (approved):

vszakats:
#14365 resolved by deleting the local `PKG_CONFIG_PATH` setting.

dg0yt:
Hm, vcpkg builds curl with static linkage on windows. I guesst this will break a working configuration. libidn2 needs libunistring, and pkg-config is how the transitive usage requirement is carried forward.

dg0yt:
> An earlier failed attempt to tackle this: #14140. With no answers and adding to
> the confusion. My concern is also cross-builds, which broke a CI elsewhere.

Which answers were missing? 
I mean I often do cross-builds when updating the vcpkg port of CURL, also targeting mingw, also with static linkage. If pkg-config comes with the wrong personality for the target, or gets the wrong environment for the target, then it is not CURL'S responsibility to fix it. This doesn't scale. Garbage in, garbage out. 

> Does UNIX include MSYS?

This means: the limited environment which doesn't under C:/ or C:\.
UNIX doesn't include MinGW.

vszakats:
It's good that vcpkg is cleared, and AFAIK it's the only platform besides UNIX which is known to tolerate (or support) `pkg-config` out of the box. Excluding the cases of explicit misconfigurations, there is at least one known case which doesn't play well with it: cross-builds run on platforms which have native `pkg-config` configured by default. A common example is Windows builds from Linux or macOS.

It doesn't seem good behaviour to make these build-cases suddenly broken, and make them require some tweak to make it work again. What's still unclear is how to define/detect the affected build cases. It's also unclear what is the best way to opt-in-to / opt-out-from `pkg-config` if defaults don't behave as expected.

dg0yt:
> It's also unclear what is the best way to opt-in-to / opt-out-from `pkg-config` if defaults don't behave as expected.

I had a clear opinion/suggestion for CMake builds, but I failed to convince you to that fixing problems in CMake builds should be done with CMake features alone.

> A common example is Windows builds from Linux or macOS.

This is a normal cross-build scenario. Even with autotools, you would be expected to setup either the enviroment or to use a wrapper which does it for you. I don't see anything special.

vszakats:
> > It's also unclear what is the best way to opt-in-to / opt-out-from `pkg-config` if defaults don't behave as expected.
> 
> I had a clear opinion/suggestion for CMake builds, but I failed to convince you to that fixing problems in CMake builds should be done with CMake features alone.

That solution [broke](https://github.com/curl/curl/pull/14140#issuecomment-2257819252) on CMake internals.

IMO a pkg-config-specific solution for a pkg-config-specific problem,
as long as it solves the opt-in/opt-out issue. A CMake knob would be
better, if there is any.

> > A common example is Windows builds from Linux or macOS.
> 
> This is a normal cross-build scenario. Even with autotools, you would be expected to setup either the enviroment or to use a wrapper which does it for you. I don't see anything special.

The envs, userbase, expectations, build options and the histories are quite
different for CMake and autotools.

`pkg-config` is also GCC-family- and Unix-biased, while CMake supports
MSVC and Windows, which don't fit into the `pkg-config` concept by default.

Perhaps that's the reason why `pkg-config` support in CMake feels like an
afterthought and neither CMake itself, nor most (?) projects have a clear idea
or patterns how to use it accurately. While in autotools, it's a natural
underpinning for dependency detection.

Either way there is/was great efforts put into syncing cmake/autotools, but
IMO it's not helpful to bring on the big hammer and break stuff for CMake on
accounts that autotools always worked a certain way.

Hence my seeking actual answers to these questions. If there are none or
the questions are wrong, anyone is free to propose alternate PRs, till we
try figuring it out step by step.

dg0yt:
The FindGnuTLS thing is a cmake bug and won't go away quickly, even when the fix is simple. 
It is mitigated by the `CMAKE_USE_GNUTLS` gate, but I understand that more users want to keep that gate open today. (But these user probably wouldn't use `CMAKE_DISABLE_FIND_PACKAGE_PkgConfig` anyways? IDK.)

There isn't just a CMake side to the pkg-config problems. There are also many pc files with problems, at least for transitive usage requirements. 

> While in autotools, it's a natural underpinning for dependency detection.

I'm still patching many up-to-date autotools projects to use pkg-config instead of hardcoded library searches. Indicating that projects don't care about static linkage, or are unaware of different binary names, or that they lack contributions.

I'm glad to see curl better fill `Requires` now. The vpckg port has an extra script to unroll link libraries, and I would love to drop it. OTOH we still need to unroll the the dependencies as long as there is `curl-config`, and as long as not all dependecies come with pkg-config.

vszakats:
> I'm glad to see curl better fill `Requires` now. The vpckg port has an extra script to unroll link libraries, and I would love to drop it. OTOH we still need to unroll the the dependencies as long as there is `curl-config`, and as long as not all dependecies come with pkg-config.

/OFF

Can you be more specific?:
- is there something in curl `Requires` that still needs fixing?
- is there something we can do with `curl-config` to improve it?

Sadly `Requires` also broke some stuff over at Debian, but I think it should be okay by now.

`pkg-config` is clearly fragile everywhere, but it just reinforces that we need to figure
out how to toggle it and avoid it in problematic cases by default.

bagder:
> `pkg-config` is clearly fragile everywhere

I don't think any system for this can be introduced without being fragile since it requires that the info provided is correctly setup and filled in, and usually the ones providing the info isn't themselves using it so it is bound to end up wrong every now and then.

I'm sure libcurl's pkg-config fits that pattern as well.

dg0yt:
> * is there something in curl `Requires` that still needs fixing

I didn't find time to validate it so far so the port just doesn't use it. (Curl release must be adopted quickly for security concerns, and at that time your (positive) work has the consequence that updating the patches in the port is more difficult than usual. Between releases time goes into housekeeping for other peoples' contributions. And that time isn't available for upstreaming useful stuff such as exporting CURL components to CMake config. Which is necessary to make CURL config a full replacement for `FindCURL.cmake`.)

> * is there something we can do with `curl-config` to improve it?

Drop it for using pkg-config? AFAIU this is what pkg-config was invented for. (I know that some `<foo>-config` delegate to pkg-config today, but I don't know now if `curl-config` is already in the same boat. And that approach might have it is own traps.)

vszakats:
#14405 turned out to be an env with an explicit `PKG_CONFIG_LIBDIR` setting. Causing OpenSSL header/lib mixup after `pkg-config` detection (libidn2). There could be other factors leading to this outcome, but deleting that setting fixed the build.

One more related update: #13947 plans to switch system IDN on by default on Windows and Apple. The consequence is that libidn2 will no longer be detected by default for these targets, also avoiding the issue fixed by this PR in default builds.

vszakats:
Update regarding #14405: Even though the env had a manual `PKG_CONFIG_LIBDIR`, an other contributing factor appears to be Strawberry Perl. It's now a build dependency for OpenSSL, and it comes with a Perl and MS-DOS batch implementation of `pkg-config` (usually in `C:/Strawberry/perl/bin`), libidn2, a copy of OpenSSL and also the mingw toolchain. With its `c/include` gets into the include path, it can cause confusion.

This may change the situation a little bit because this could be present in envs and in PATH.

Perhaps this condition could replace `if(UNIX)`: `if(NOT MSVC OR VCPKG_TOOLCHAIN)`, plus require cross-compilations to setup `PKG_CONFIG_PATH`/`PKG_CONFIG_LIBDIR`/`PKG_CONFIG_SYSROOT_DIR` accurately (or to _empty_ to disable `pkg-config`.) I can't see a situation where MSVC mixes well with `pkg-config` outside of vcpkg.

Let me know what you think.

dg0yt:
> Perhaps this condition could replace `if(UNIX)`: `if(NOT MSVC OR VCPKG_TOOLCHAIN)`, plus require cross-compilations to setup `PKG_CONFIG_PATH`/`PKG_CONFIG_LIBDIR`/`PKG_CONFIG_SYSROOT_DIR` accurately (or to _empty_ to disable `pkg-config`.) I can't see a situation where MSVC mixes well with `pkg-config` outside of vcpkg.

No. All those variables (incl. `UNIX`) bind effects tightly to variables which merely have a correlation with the observations, but are not the root cause for failure or success. And they take control completely out of user's hands. Given the dislike for `CMAKE_DISABLE_FIND_PACKAGE_PkgConfig`, `UNIX` should be replaced with a cache variable like `CURL_USE_PKGCONFIG`. Use the correlations for the variable's default.

In the meantime I did https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9709: Guard all use of `pkg_check_modules` in CMake. `CMAKE_DISABLE_FIND_PACKAGE_PkgConfig` should work in the future.

> Strawberry Perl. It's now a build dependency for OpenSSL, and it comes with a Perl (also via MS-DOS batch wrapper) implementation of `pkg-config` (usually in `C:/Strawberry/perl/bin`), libidn2, a copy of OpenSSL and other packages, and also a mingw toolchain. With its `c/include` getting into the include path, it can cause confusion

A known pain. Note that you don't get completely rid of it with `UNIX`. There are people who target Android (or embedded Linux) from Windows.

vszakats:
> > Perhaps this condition could replace `if(UNIX)`: `if(NOT MSVC OR VCPKG_TOOLCHAIN)`, plus require cross-compilations to setup `PKG_CONFIG_PATH`/`PKG_CONFIG_LIBDIR`/`PKG_CONFIG_SYSROOT_DIR` accurately (or to _empty_ to disable `pkg-config`.) I can't see a situation where MSVC mixes well with `pkg-config` outside of vcpkg.
> 
> No. All those variables (incl. `UNIX`) bind effects tightly to variables which merely have a correlation with the observations, but are not the root cause for failure or success. And they take control completely out of user's hands. Given the dislike for `CMAKE_DISABLE_FIND_PACKAGE_PkgConfig`, `UNIX` should be replaced with a cache variable like `CURL_USE_PKGCONFIG`. Use the correlations for the variable's default.

I agree `CURL_USE_PKGCONFIG` would be nicer, but is it a legit option within `Find*` modules? They are supposed to be distributed in curl binary distributions and executed outside curl's own CMake build process, where `CURL_USE_PKGCONFIG` is not defined.
 
> In the meantime I did https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9709: Guard all use of `pkg_check_modules` in CMake. `CMAKE_DISABLE_FIND_PACKAGE_PkgConfig` should work in the future.

Nice!

In the meantime we learned (Thanks @micahsnyder!) that `-DPKG_CONFIG_EXECUTABLE={""|none|OFF}` also works to disable `pkg-config`.

> A known pain. Note that you don't get completely rid of it with `UNIX`. There are people who target Android (or embedded Linux) from Windows.

I'm aware. It's a condition that was used in curl for a decade. This doesn't in itself make it accurate, but there were apparently no reports about issues, so either everyone managed to work it around or it did not cause grave problems. But, omitting the condition triggered two bug reports in a week.

dg0yt:
> I agree `CURL_USE_PKGCONFIG` would be nicer, but is it a legit option within `Find*` modules? They are supposed to be distributed in curl binary distributions and executed outside curl's own CMake build process, where `CURL_USE_PKGCONFIG` is not defined.

Why shouldn't it be legit? The find modules should be seen as private implementation details. CURL ships CMake config. If it has to ship and use extra find modules, they should be scoped to finding transitive depedencies, and they shouldn't be allowed to find something signitificantly different then used during the build. Implementation-wise, CURL CMake config would set `CURL_USE_PKGCONFIG` before calling `find_dependendency`, like it has to set `CMAKE_MODULE_PATH`. And restore before existing the scope (unless using another cache variable).

vszakats:
OK, so we must replicate the `CURL_USE_PKGCONFIG` initialization inside `curl-config.cmake`, once we start shipping any local Find module.

Do you think there anything special to do in the restore-before-existing-scope step there? (I know very little about variable caching)

PR #14407, opened 2024-08-05
Improve cipher options documentation

jan2000:

PR #14423, opened 2024-08-06
manpage: ensure a maximum width for the text version

bagder:
... using the new script 'maxline' to which we specify the maximum number of columns we allow any single line to be, or it will cause an error.

Bonus: shorten the long line in the --ipfs-gateway section.

### DIFF
--- a/docs/cmdline-opts/Makefile.am
+++ b/docs/cmdline-opts/Makefile.am
@@ -37,6 +37,10 @@ GN_1 =
 GN_ = $(GN_0)
 
 MANAGEN=$(top_srcdir)/scripts/managen
+MAXLINE=$(top_srcdir)/scripts/maxline
+
+# Maximum number of columns accepted in the ASCII version of the manpage
+MAXCOLS=100
 INCDIR=$(top_srcdir)/include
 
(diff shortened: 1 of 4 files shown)


jan2000:
So, I was already wondering about the line limit yesterday, because I have this long line in PR #14407:

```
For OpenSSL, wolfSSL, mbedTLS, Secure Transport and BearSSL a valid example
of a cipher list is:
**ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305**
```

(Which I really want to keep this way because these four ciphers are the equivalent of the two ciphers in the TLS1.3 example.)

What to do with this then?

bagder:
Either we come up with a shorter example, or we figure out a way how to split it into multiple lines.

bagder:
Also, the current limit is set to 100 columns but that's a bit arbitrary. We can tweak it if we think we should.

The most important thing is that code in src/tool_help.c has an array size that is larger than the longest line, so I want this to make sure this is true.

PR #14426, opened 2024-08-06
docs: Mention trace-config in verbose help

yedayak:


bagder (approved):

bagder:
I would say this change is no longer necessary after 06c5829dab52d. What do you think?

Issue #14176, opened 2024-07-13
AppleIDN: fails tests 165, 1034, 1035


vszakats:
### I did this

https://github.com/curl/curl/actions/runs/9919852332/job/27406122714
```
FAIL 165: [HTTP over proxy with IDN host name] HTTP HTTP GET HTTP proxy IDN
FAIL 1034: [HTTP over proxy with malformatted IDN host name] HTTP HTTP GET HTTP proxy IDN FAILURE config file
FAIL 1035: [HTTP over proxy with too long IDN host name] HTTP HTTP GET HTTP proxy IDN FAILURE

TESTFAIL: These test cases failed: 165 1034 1035
```

For 165, it seems to be caused by AppleIDN converting `ß` to `ss` instead of punycode `ß` expected by the test (and generated by libidn2 with `en_US.UTF-8`).

/cc @MonkeybreadSoftware 

### I expected the following

Tests passing

### curl/libcurl version

master 3765d75ce47b66289f946382b649d0e99389dc77

### operating system

Apple

MonkeybreadSoftware:
Yes, seems like these tests or the library may need updating. I'll check

I found a FAQ entry from DENIC:

> Most applications support the characters which have been permitted since the Internet standard 2003 became effective. The situation is different for "ß", which can only be used since the IDN standard was revised in August 2010. Up to that date, "ß" was always converted into "ss" (normalized).

https://www.denic.de/en/faqs/faqs-about-idns-ss#code-365

MonkeybreadSoftware:
to solve the große problem:

uidna_openUTS46 is called with flags, currently UIDNA_CHECK_BIDI.
Change that to 

UIDNA_CHECK_BIDI|UIDNA_NONTRANSITIONAL_TO_UNICODE|UIDNA_NONTRANSITIONAL_TO_ASCII 

for both directions, so it does the ß differently.

MonkeybreadSoftware:
For test 1034, it seems like Apple's unicode function here doesn't do any UTF-8 input validation. The invalid character is just passed through. Not sure whether you like to do an UTF-8 validation yourself in the code. 
Also no idea whether passing on the invalid UTF-8 to a DNS server later causes trouble.

For 1035, there is an RFC defining 63 characters, but the unicode functions don't enforce this. I thought domains could be longer, but well if you like to enforce this, you could add a check. 

In both cases I wonder if CURL should enforce this at all. If the name has invalid characters or is too long, the DNS resolver will not find anything.

[...]

PR #14431, opened 2024-08-07
idn: align AppleIDN behaviour with libidn2

Bo98:
* idn: support non-UTF-8 input under AppleIDN

The libcurl documentation states:

> If libcurl is built with IDN support, the server name part of the URL can use an "international name" by using the current encoding (according to locale) or UTF-8 (when WinIDN is used; or a Windows Unicode build using libidn2).
>
> https://curl.se/libcurl/c/CURLOPT_URL.html

AppleIDN was following the Windows behaviour of only using UTF-8. We can however use `iconv` here, which is shipped with macOS, to convert to UTF-8 and align the behaviour with libidn2.

This also in turn fixes test 1034 as the input is validated as a part of character encoding conversion.

---

* idn: more strictly check AppleIDN errors

`uidna_nameToASCII_UTF8` returns _two_ error values. One largely covers fatal errors and other bad inputs, which we were currently checking. However there was a second error value that validates whether the output IDNA is valid to the spec.

Checking this better aligns the behaviour with that of libidn2 and fixes test 1035.

---

Ref: #14176


bagder:
Thanks!

PR #14436, opened 2024-08-07
tracing, allow CURL_DEBUG override

icing:
On debug builds, allow environment variable CURL_DEBUG to override any setting done via '-v' or '--no-verbose'.


bagder:
Add to `docs/libcurl/libcurl-env-dbg.md` ?

icing:
> Add to `docs/libcurl/libcurl-env-dbg.md` ?

Added description and example.

bagder (approved):

PR #14444, opened 2024-08-07
mbedtls: add more informative logging

jmaggard10:
After TLS handshare, indicate which TLS version was negotiated in addition to the cipher in the handshake completed log message.

Also use the verify callback for certificate logging and collection. This allows things to work even when
MBEDTLS_SSL_KEEP_PEER_CERTIFICATE is disabled in the mbedtls library.

And lastly, catch certificate validation errors later so we can give the user more informative error messages that indicate what the failure was from certificate validation.

These changes also happen to fix an issue where setting CURLOPT_SSL_VERIFYHOST to 0 did not take effect on MbedTLS 3.6 when TLS 1.3 was enabled.

Tested on both current LTS versions (2.28 and 3.6).

### DIFF
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -115,6 +115,11 @@ struct mbed_ssl_backend_data {
   BIT(sent_shutdown);
 };
 
+struct verify_cb_data {
+  struct Curl_easy *data;
+  struct Curl_cfilter *cf;
+};
+
 /* apply threading? */
 #if (defined(USE_THREADS_POSIX) && defined(HAVE_PTHREAD_H)) || \
     defined(_WIN32)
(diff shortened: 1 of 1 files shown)


lib/vtls/mbedtls.c
```diff
@@ -802,15 +882,7 @@ mbed_connect_step1(struct Curl_cfilter *cf, struct Curl_easy *data)
...
-#ifdef TLS13_SUPPORT
-  if(!verifypeer) {
-    /* Default verify behaviour changed in mbedtls v3.6.0 with TLS v1.3.
-     * On 1.3 connections, the handshake fails by default without trust
-     * anchors. We override this questionable change by installing our
-     * own verify callback that clears all errors. */
-    mbedtls_ssl_conf_verify(&backend->config, mbed_no_verify, cf);
-  }
-#endif
+  mbedtls_ssl_conf_verify(&backend->config, mbed_verify, &cb_data);
```
bagder, line 885: This change seems unrelated and possibly even wrong?

lib/vtls/mbedtls.c
```diff
@@ -1084,18 +1100,15 @@ mbed_connect_step2(struct Curl_cfilter *cf, struct Curl_easy *data)
...
   infof(data, "mbedTLS: Handshake complete");
 #endif
   ret = mbedtls_ssl_get_verify_result(&backend->ssl);
 
-  if(!conn_config->verifyhost)
-    /* Ignore hostname errors if verifyhost is disabled */
-    ret &= ~MBEDTLS_X509_BADCERT_CN_MISMATCH;
-
-  if(ret && conn_config->verifypeer) {
+  if(ret) {
```
bagder, line 1111: What about the previous check for `if(!conn_config->verifyhost)`. Can you just remove it? Also seems to be a change that's not about logging.

lib/vtls/mbedtls.c
jmaggard10: This is related and correct. The `verifypeer` check is now done inside the callback function. If `verifypeer` is false then `mbed_verify()` will clear the flags the same way that `mbed_no_verify` did.

We log the cert info inside the callback function now; otherwise we can't log it if cert validation fails.

lib/vtls/mbedtls.c
jmaggard10: We don't need the `verifyhost` check here any more. `verifyhost` and `verifypeer` are both handled inside the callback function now.

bagder:
About 44 TLS related tests fail.

jmaggard10:
> About 44 TLS related tests fail.

@bagder , should all be fixed now. Did a bunch of testing with ASan as well on both mbedtls 2.28 and 3.6 and both look good.

bagder:
Thanks!

PR #14450, opened 2024-08-07
cmake: more small tidy-ups and fixes

vszakats:
- tidy up two `MATCHES` expression by avoiding macros expansion and
  adding quotes. Then convert then to `STREQUAL` to match other places
  in the code doing the same checks.

- fix setting `_ALL_SOURCE` for AIX to match what autotools does.

- delete stray `_ALL_SOURCE` reference from `lib/config_riscos.h`

- simplify/fix two `STREQUAL ""` checks.
  The one in the `openssl_check_symbol_exists()` macro succeeded
  regardless of the value. The other could return TRUE when
  `CMAKE_OSX_SYSROOT` was undefined.

- delete code for CMake versions (<3.7) we no longer support.

- prefer `LIST(APPEND ...)` to extend `CURL_LIBS`.

- use `CURL_LIBS` to add the `network` lib for Haiku.
  Before this patch it was done via raw C flags. I could not test this.

- move `_WIN32_WINNT`-related code next to each other.
  It also moves detection to the top, allowing more code to use
  the result.

- merge two `WIN32` blocks.

- rename internal variables to underscore + lowercase.

- unwrap a line, indent, whitespace.

Closes #14450


dfandrich:
The AIX 7.1 headers do things like defining _XOPEN_SOURCE to 700,
_POSIX_C_SOURCE to 200809L, setting _LARGE_FILE_API and a few more defines when
it sees _ALL_SOURCE. So, it looks similar to what glibc does for
_DEFAULT_SOURCE.

vszakats:
Thanks Dan.

Do you know what purpose the `#ifndef / #undef` combo may have?
After staring at it for a while I cannot figure it out. It seems, just defining `_ALL_SOURCE` should be enough for AIX, or is it not?

vszakats:
Ah OK, I think I get it, autotools converts the odd construct to this, when defined:
```
/* Define to 1 if OS is AIX. */
#ifndef _ALL_SOURCE
#  define _ALL_SOURCE 1
#endif
```

I still cannot test it, but the CMake method of passing `-D_ALL_SOURCE` via the command-line would likely work without such trick. (Unless, AIX also defines it to a different value without checking if it's already set).

Probably making it `-D_ALL_SOURCE=1` would be useful to sync it more with autotools.

dfandrich:
All the AIX references to `_ALL_SOURCE` are either `#ifdef` or `defined()` so it doesn't need to be set to a value like 1. AIX seems to set `_ALL_SOURCE` itself if none of `_XOPEN_SOURCE`, `_POSIX_SOURCE` or `ANSI_C_SOURCE` is already set. 

You should get yourself a [Compile Farm](https://portal.cfarm.net/machines/list/) account to help with this kind of testing. They have two different AIX versions available as well as other obscure machines.

vszakats:
Thanks, they answer the questions and I've closed #14461 as not necessary.

Also for the AIX tip!

PR #14456, opened 2024-08-08
cmake: show pkg-config basic env if any

vszakats:
Ref: #14405
Closes #14456


vszakats:
In CI, they are empty by default.

vszakats:
Instead of trying to debug `pkg-config` envs by default, I proposed #14504, giving better control of using `pkg-config`.

Closing without merge.

PR #14458, opened 2024-08-08
websocket, introduce blocking sends

icing:
When using `curl_ws_send()`, perform a blocking send of the data if the call is done from within a curl callback. A partial write of the data could subsequently mess up the ws framing, as a callback has a hard time handling waits and repeated calls.

Outside a callback, `curl_ws_send()` is always non-blocking *except* when generating a 0-length frame that could not completely be sent. This is no way to report this partial success.

Document this behaviour in the function and add this as item to the TODO list. Also, add `ws` as protocol trace feature.

Fixes WebSockets tests with CURL_DBG_SOCK_WBLOCK=90 set.

### DIFF
--- a/docs/TODO
+++ b/docs/TODO
@@ -481,6 +481,8 @@
 
  - curl_multi_remove_handle for any of the above. See section 2.3.
 
+ - Calling curl_ws_send() from a callback
+
 2.2 Better support for same name resolves
 
  If a name resolve has been initiated for name NN and a second easy handle
--- a/docs/libcurl/curl_global_trace.md
+++ b/docs/libcurl/curl_global_trace.md
@@ -109,6 +109,10 @@ Traces reading of upload data from the application in order to send it to the se
 
 Traces writing of download data, received from the server, to the application.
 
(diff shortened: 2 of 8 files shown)


dfandrich:
Should this blocking behaviour be documented somewhere along with the other parts of curl that still blocking?

icing:
A good question. I did not really like to introduce this, but it seems necessary with the current design. Maybe @bagder has an idea how to address this.

But if we go for this, it needs to be documented in `curl_ws_send()`, yes.

dfandrich:
There are a number of other blocking calls mentioned in TODO and KNOWN_BUGS. If
this is a temporary measure that might be fixed better in the future, then it
should probably go there, too.

icing:
Improved handling to make curl_ws_send() non-blocking in almost all cases when invoked outside a callback. Updated documentation.

polarina:
What is expected of clients who wish to resume a `curl_ws_send` call after a partial write? The semantics when partial writes are reported are unclear.

PR #14459, opened 2024-08-08
tidy-up: delete unused `m4/xc-translit.m4`

vszakats:
Unused since it was added in 49a8fe51420cf68c4e420668eb354f0cce6ee4f0.
Also missing from the source tarball.

Closes #14459


bagder (approved):
Nice catch!

PR #14463, opened 2024-08-08
dist: add CI job to detect files missing from distro

vszakats:
Also:
- delete previous, cmake-specific solution.
- move a CI script under `.github`.

Follow-up to a118a6ecddb0322a6da07815aabf9e36cd5f44bc #14323
Closes #14463

---

- [x] rebase on #14459.
- [x] rebase on #14467.


vszakats:
It detects the problem fixed in #14459:
```diff
--- /tmp/tmp.G7I2Upv0ud	2024-08-08 20:13:08.385252775 +0000
+++ /tmp/tmp.EdTxmiptJ0	2024-08-08 20:13:08.389252796 +0000
@@ -1418,6 +1418,7 @@
 m4/xc-am-iface.m4
 m4/xc-cc-check.m4
 m4/xc-lt-iface.m4
+m4/xc-translit.m4
 m4/xc-val-flgs.m4
 m4/zz40-xc-ovr.m4
 m4/zz50-xc-ovr.m4
Error: Process completed with exit code 1.
```
Ref: https://github.com/curl/curl/actions/runs/10308837478/job/28537182373?pr=14463#step:3:1595

.github/scripts/distfiles.sh
```diff
@@ -0,0 +1,65 @@
...
+^LICENSES/.+
+^docs/examples/adddocsref\.pl
+^docs/THANKS-filter
+^lib/optiontable\.pl
+^projects/Windows/.+
+^scripts/(ciconfig|cijobs|cmp-config|release-notes|singleuse)\.pl
+^scripts/(contributors|contrithanks|installcheck)\.sh
+^scripts/delta
+^src/tool_hugehelp\.c\.cvs
+^tests/CI\.md"
```
bagder, line 49: Don't you think this list might be annoying to keep up-to-date going forward?

.github/scripts/distfiles.sh
```diff
@@ -0,0 +1,65 @@
...
+^\..+
+^appveyor\..+
+^buildconf
+^renovate\.json
+^REUSE\.toml
+^(GIT-INFO|README|SECURITY)\.md
+^LICENSES/.+
+^docs/examples/adddocsref\.pl
+^docs/THANKS-filter
+^lib/optiontable\.pl
```
bagder, line 43: Since `optiontable.pl` is used in a lib/Makfile.am target and test 1276, I think we should rather include it in the dist.

.github/scripts/distfiles.sh
vszakats: Fair question, I think the list of exceptions is short, changes less often and is touched by few contributors (mainly you), when dealing with a few specific areas. Compared to the rest of files which get added more often by more people, related to any area. History shows it's easy to miss them [^1], and prone to cause fallouts.

To easy on the exceptions, we may aname/place them in such way that they are easy to exclude with more generic rules. E.g. by renaming `appveyor.*` to `.appveyor.*` or naming ci-specific scripts with `ci*.*`. There may be more.

[^1]: from the last 10 months:
b1153820f499cb511f04435f40ee463322816d8a
a4ad7dc5a3e21d573a02d8c0202e2bb26066f302
58946eed20dbf6f606c19ead5e34acb7aed817d8
5d3016adf7e855f03a644fb05308129806c98457
0a4419ae4c45fccaadc62dbdaf4725909c44dcfa
d0a851feadb28bb808ae9fe4e2c53a1636ec4c00
da8c1d15782c8161b455a7ee90197c16ae5edb90
fac90bd5c1cf81a3629979e2a7ec0757a916e213

.github/scripts/distfiles.sh
vszakats: Updated!

.github/scripts/distfiles.sh
vszakats: Would it make sense to move (some of) these inside `.github/scripts`?:
```
scripts/(ciconfig|cijobs|cmp-config|release-notes|singleuse)\.pl
```

bagder: Except for `cmp-config`, they were all written for and I still occasionally use them outside of CI jobs so I think not.

And note: I don't deny that we have done mistakes. The question is only what the best way is to make us do less errors in the future without causing friction for everyone at the same time.

vszakats: Moved `cmp-config` there and dropped the exception.

Also deleted an earlier attempt that was limited to `./CMake/*.{cmake,in}` files.

vszakats: AFAICS the only extra friction is that non-distributable files now may need to be added to the exceptions.
In return, it guarantees that distributable ones are included in the tarball.

bagder: Yeah, I can't figure out a better approach either.

vszakats:
Made the exception lists less regexpy, with the intent of making
them easier to edit, read and grep.

vszakats:
Actually, if don't care about files present in tar and not in git, we might drop the taronly exception list, by filtering for the '+' differences. I'd say there is some value (from a security angle) in checking what extra files make into the tarball, but not mandatory to achieve the goal of this PR.

```diff
--- a/.github/scripts/distfiles.sh
+++ b/.github/scripts/distfiles.sh
@@ -13,23 +13,7 @@ gitfiles="$(mktemp)"
 
 taronly="/
 ^
-Makefile.in
-^aclocal.m4
-^compile
-^configure
-^config.*
-^depcomp
-^install-sh
-^ltmain.sh
-^missing
-^docs/libcurl/libcurl-symbols.md
-^docs/RELEASE-TOOLS.md
-^docs/tarball-commit.txt
-^lib/curl_config.h.in
-^m4/libtool.m4
-^m4/lt*.m4
-^src/tool_ca_embed.c
-^src/tool_hugehelp.c"
+Makefile.in"
 
 gitonly=".git*
 ^.*
@@ -64,7 +48,14 @@ git ls-files \
   | grep -v -E "($(printf '%s' "${gitonly}" | tr $'\n' '|' | sed -e 's|\.|\\.|g' -e 's|\*|.+|g'))$" \
   | sort > "${gitfiles}"
 
-diff -u "${tarfiles}" "${gitfiles}"
+dif="$(diff -u "${tarfiles}" "${gitfiles}" | tail -n +3 || true)"
+
+echo 'Only in tarball:'
+echo "${dif}" | grep '^-'
+echo
+
+echo 'Missing from tarball:'
+echo "${dif}" | grep '^+'
 res=$?
 
 rm -rf "${tarfiles:?}" "${gitfiles:?}"
```

bagder:
> I'd say there is some value (from a security angle) in checking what extra files make into the tarball, but not mandatory to achieve the goal of this PR.

Right. And the **reproducible-releases** CI job already verifies that a release from tarball contents becomes identical as a release from git, which is a teller that what is listed as going into the tarball is what goes in and nothing else.

PR #14469, opened 2024-08-09
configure: Replace nonportable grep -o with awk

dfandrich:
This stops an error shown at the end of configure on systems with POSIX
grep and fixes the warning it's trying to convey.

Closes #14469

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5212,7 +5212,8 @@ AC_MSG_NOTICE([Configured to build curl/libcurl:
   Features:         ${SUPPORT_FEATURES}
 ])
 
-non13=`echo "$TLSCHOICE" | $EGREP -io 'bearssl|secure-transport'`;
+# grep -o would simplify this, but is nonportable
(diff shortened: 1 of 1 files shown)


dfandrich:
This is only the second use of awk in the curl build path. Since awk is not installed by default everywhere any longer, someone may wish to convert these to something else to make building ever so slightly simpler for people, at the risk of complicating the resulting code.

dfandrich:
Actually, it seems that autoconf itself uses awk internally in many places making it a hard dependency, so there's no real point in trying to remove curl's explicit use of it.

bagder (approved):

dfandrich:
Analysis of PR #14469:

[Test 1379 failed](https://circleci.com/gh/curl/curl/102340), which has NOT been flaky recently, so **there could be a real issue in the PR**.  Note that this CI job has had a number of other flaky tests recently (2, to be exact) so it may be that this failure is rather a systemic issue with this job and not with this specific PR.

dfandrich:
Test failure is the usual FTP server glitch. In any case, it's completely unrelated to the PR.

dfandrich:
autoconf ate my brackets. Trying again.

PR #14471, opened 2024-08-09
docs: add description of effect of --location-trusted on cookie

XYenon:


docs/cmdline-opts/location-trusted.md
```diff
@@ -2,7 +2,7 @@
 c: Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
 SPDX-License-Identifier: curl
 Long: location-trusted
-Help: As --location, but send auth to other hosts
+Help: As --location, but send auth and cookie to other hosts
```
bagder, line 5: In the name of trying to keep this is brief as possible (for a better `curl -h` experience):

```suggestion
Help: As --location, but send secrets to other hosts
```

Also of course because *normally* curl will send cookies to the other host as well. It is only if you set cookies with `-H Cookie:` or with `-b name=moo` this exception goes for.

docs/cmdline-opts/location-trusted.md
```diff
@@ -11,11 +11,12 @@ See-also:
...
 # `--location-trusted`
 
-Like --location, but allows sending the name + password to all hosts that the
-site may redirect to. This may or may not introduce a security breach if the
-site redirects you to a site to which you send your authentication info (which
-is clear-text in the case of HTTP Basic authentication).
+Like --location, but allows sending the name + password and cookie to all hosts
+that the site may redirect to. This may or may not introduce a security breach
+if the site redirects you to a site to which you send your authentication info
+(which is clear-text in the case of HTTP Basic authentication).
```
bagder, line 22:
```suggestion
Instructs curl to like --location follow HTTP redirects, but permits it to
send credentials and other secrets along to other hosts than the initial one.

This may or may not introduce a security breach if the site redirects you to a
site to which you send this sensitive data to. Another host means that one or
more of host name, protocol scheme or port number changed.

This option also allows curl to pass long cookies set explicitly with
--coookie or with --header.
```

docs/cmdline-opts/location.md
```diff
@@ -22,9 +22,10 @@ location (indicated with a Location: header and a 3XX response code), this
 option makes curl redo the request on the new place. If used together with
 --include or --head, headers from all requested pages are shown.
 
-When authentication is used, curl only sends its credentials to the initial
-host. If a redirect takes curl to a different host, it does not get the
-user+password pass on. See also --location-trusted on how to change this.
+When authentication or cookie is used, curl only sends its credentials to the
+initial host. If a redirect takes curl to a different host, it does not get the
+user+password or cookie pass on. See also --location-trusted on how to change
+this.
```
bagder, line 28: This needs to clarify the cookie use. Because normally cookies are only sent to the correct host names so this extra precaution is only for when you "force" cookies explicitly.

bagder (commented):
Thanks, this is indeed details we should clarify!

docs/cmdline-opts/location-trusted.md
XYenon: It looks like this only effects on `-H Cookie:` but not `-b`.

bagder: Yes, but my suggestion is the same anyway.

bagder (changes requested):
As mentioned

docs/cmdline-opts/location.md
XYenon: What do you think about this description?

bagder (approved):

bagder:
Please also fix the "badwords" error:
~~~
docs/cmdline-opts/location-trusted.md:24:9: error: found bad word "host name\b"
   24 | more of host name, protocol scheme or port number changed.
      |         ^~~~~~~~~~~
 maybe use "hostname" instead?

bagder (changes requested):
see previous comment

docs/cmdline-opts/location-trusted.md
```diff
@@ -11,11 +11,17 @@ See-also:
...
-is clear-text in the case of HTTP Basic authentication).
+Instructs curl to like --location follow HTTP redirects, but permits it to
+send credentials and other secrets along to other hosts than the initial one.
+
+This may or may not introduce a security breach if the site redirects you to a
+site to which you send this sensitive data to. Another host means that one or
+more of host name, protocol scheme or port number changed.
+
+This option also allows curl to pass long cookies set explicitly with
+--coookie or with --header.
```
XYenon, line 27: By the way, should I remove `--cookie` here?

bagder: Ah yes, I think so.

bagder (approved):

bagder:
Thanks!

docs/cmdline-opts/location.md
```diff
@@ -22,9 +22,10 @@ location (indicated with a Location: header and a 3XX response code), this
 option makes curl redo the request on the new place. If used together with
 --include or --head, headers from all requested pages are shown.
 
-When authentication is used, curl only sends its credentials to the initial
-host. If a redirect takes curl to a different host, it does not get the
-user+password pass on. See also --location-trusted on how to change this.
+When authentication is used, or send cookie with `-H Cookie:`, curl only sends
+its credentials to the initial host. If a redirect takes curl to a different
+host, it does not get the credentials pass on. See also--location-trusted on
```
Joshix-1, line 27: `also--location-trusted` seems to be missing a space? Or can the `also` be removed entirely?

bagder: Thanks, fixing

PR #14473, opened 2024-08-09
Introduce transfer multi-id, use for dependency tracking

icing:
Introduce `data->mid`, a unique identifier inside a multi

`data->id` is unique in *most* situations, but not in all. If a libcurl application uses more than one connection cache, they will overlap. This is a rare situations, but libcurl apps do crazy things.
However, for informative things, like tracing, `data->id` is superior, since it assigns new ids in curl's serial `curl_easy_perform()` use. Also, `data->id` is not cleared when the transfer is removed from the multi handle.

Introduce `data->mid` which is a unique identifer inside one multi instance, assigned on `multi_add_handle()` and cleared on `multi_remove_handle()`.

Use the `mid` in DoH operations and also in h2/h3 stream hashes.

### DIFF
--- a/lib/doh.c
+++ b/lib/doh.c
@@ -216,10 +216,13 @@ static int doh_done(struct Curl_easy *doh, CURLcode result)
 {
   struct Curl_easy *data;
 
-  data = Curl_multi_get_handle(doh->multi, doh->set.dohfor_id);
-  DEBUGF(infof(doh, "doh_done: xfer for %" CURL_FORMAT_CURL_OFF_T
-               " is %p", doh->set.dohfor_id, (void *)data));
-  if(data) {
+  data = Curl_multi_get_handle(doh->multi, doh->set.dohfor_mid);
+  if(!data) {
+    DEBUGF(infof(doh, "doh_done: xfer for mid=%" CURL_FORMAT_CURL_OFF_T
+                 " not found", doh->set.dohfor_mid));
+    DEBUGASSERT(0);
+  }
+  else {
     struct dohdata *dohp = data->req.doh;
     /* one of the DoH request done for the 'data' transfer is now complete! */
     dohp->pending--;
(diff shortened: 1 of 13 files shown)


lib/doh.c
```diff
@@ -214,19 +214,23 @@ static void local_print_buf(struct Curl_easy *data,
...
-  struct dohdata *dohp = data->req.doh;
-  /* so one of the DoH request done for the 'data' transfer is now complete! */
-  dohp->pending--;
-  infof(doh, "a DoH request is completed, %u to go", dohp->pending);
-  if(result)
-    infof(doh, "DoH request %s", curl_easy_strerror(result));
+  struct Curl_easy *data;
+
+  data = Curl_multi_get_handle(doh->multi, doh->set.dohfor_id);
+  if(data) {
```
bagder, line 220: If `data` is NULL here, it seems like a good place to add some (debug) logging right? Because it is totally unexpected.

lib/multi.c
```diff
@@ -3937,3 +3939,24 @@ static void multi_xfer_bufs_free(struct Curl_multi *multi)
...
+  struct Curl_easy *data = multi->easyp;
+  struct Curl_llist_element *e;
+
+  if(id >= 0) {
+    for(data = multi->easyp; data; data = data->next) {
+      if(data->id == id)
+        return data;
+    }
+    /* may be in msgsent queue? */
+    for(e = multi->msgsent.head; e; e = e->next) {
```
bagder, line 3955: It might also be in the `&multi->pending` list. See `MSTATE_CONNECT` handling.

bagder (commented):
Like this!

lib/doh.c
icing: Added.

lib/multi.c
icing: Added.

icing:
I think we need to keep a "multi-unique" id at easy handles and use that for lookups. The current id generation is unique per connection cache and where there are several caches in play, the ids will overlap.

For lookups at the multi handle, we need an id that is unique just inside that handle.

PR #14476, opened 2024-08-09
build: tidy up internal macro names for `libcurl.pc`

vszakats:
Rename internal macros to match their `libcurl.pc` metadata counterpart.
Also apply these to the `curl-config.in` template.

- `CPPFLAG_CURL_STATICLIB` -> `LIBCURL_PC_CFLAGS`
- `LIBCURL_LIBS` -> `LIBCURL_PC_LIBS_PRIVATE`
- `LIBCURL_NO_SHARED` -> `LIBCURL_PC_LIBS`

Closes #14476


vszakats:
There is a slightly confusing case within `curl_config.in`:
It had the name `CPPFLAG_CURL_STATICLIB` used 1. as a macro 2. as a local variable in lowercase 3. as a local variable in uppercase.

PR #14480, opened 2024-08-09
wolfssl: fixed CURLOPT_SSLVERSION

jan2000:
Before, setting CURLOPT_SSLVERSION with wolfSSL restricted the the tls proto to just the specified version. Now it properly supports a range. So it can set the min and max tls proto (max requires wolfSSL 4.2.0).

Bump the absolute minimum required version of wolfSSL to 3.4.6 (released 2015) because it is needed for the wolfSSL_CTX_SetMinVersion() function.


bagder (approved):

bagder:
Thanks!

PR #14483, opened 2024-08-10
cmake: allow `pkg-config` in more envs

vszakats:
Before this patch, `pkg-config` was used for `UNIX` builds only (with
a few exceptions like wolfSSL, libssh, gsasl, libuv). This patch extends
`pkg-config` use to all envs except: `MSVC` without vcpkg. Meaning MSVC
with vcpkg will now use it. Also mingw on Windows.

Also apply the new condition to options where `pkg-config` was used
unconditionally (= for all targets). These are:
`-DCURL_USE_WOLFSSL=ON`, `-DCURL_USE_LIBSSH=ON`,
`-DCURL_USE_GSASL=ON` and `-DCURL_USE_LIBUV=ON`

This patch may still cause regressions for cross-builds (e.g. mingw
cross-build from Unix) and potentially other cases. If that happens, we
recommend using some of these methods to explicitly disable `pkg-config`
when using CMake:
- CMake option: `-DPKG_CONFIG_EXECUTABLE=`
  (or `-DPKG_CONFIG_EXECUTABLE=nonexistent` or similar)
  This is similar to the (curl-specific) `PKG_CONFIG` env for autotools.
- export env: `PKG_CONFIG_LIBDIR=`
  (or `PKG_CONFIG_PATH`, `PKG_CONFIG_SYSROOT_DIR`,
  or the CMake-specific `PKG_CONFIG`)

We may improve control over this in a future patch, also allowing opting
in MSVC (without vcpkg).

Ref: #14405
Ref: #14408
Ref: #14140
Closes #14483

---

- [x] apply to all other case where `if(UNIX)` is used.
- [x] apply to `CURL_USE_WOLFSSL`, `CURL_USE_LIBSSH`, `CURL_USE_GSASL`, `CURL_USE_LIBUV`.
- [ ] add initial `pkg-config` support for deps which don't have it yet.
- [ ] replace this condition with an overridable `CURL_USE_PKGCONFIG` setting.

Without whitespace: https://github.com/curl/curl/pull/14483/files?w=1


talregev:
Why not vcpkg / msvc?

vszakats:
It _is_ MSVC, but only when combined with vcpkg. In other case, `pkg-config` is reported to break builds: #14405 #14365. Also outside vcpkg, `pkg-config` means mingw-specific options and libs.

Or, is there any raw MSVC env out there that supports `pkg-config`?

talregev:
> Or, is there any raw/other MSVC env out there that supports `pkg-config`?

I don't familiar with more.
Thank you for the detailed answer.

MarcelRaad (approved):

PR #14494, opened 2024-08-11
GHA/windows: Add GnuTLS MSVC job

talregev:
Add GnuTLS MSVC job


talregev:
@vszakats Can you take a look?

CMakeLists.txt
```diff
@@ -580,7 +580,14 @@ if(CURL_USE_WOLFSSL)
 endif()
 
 if(CURL_USE_GNUTLS)
-  find_package(GnuTLS REQUIRED)
+  if(NOT MSVC OR VCPKG_TOOLCHAIN)
+    find_package(PkgConfig QUIET)
+    pkg_search_module(GNUTLS "gnutls")
+    set(GNUTLS_LIBRARIES ${GNUTLS_LINK_LIBRARIES})
```
vszakats, line 586:
```suggestion
    if(GNUTLS_FOUND)
      set(GNUTLS_LIBRARIES ${GNUTLS_LINK_LIBRARIES})
    endif()
```

vszakats:
To not increase the number of flaky jobs, what about migrating `libssh` to an existing job and deleting the `!ssl` one? Build without SSL are tested widely in CI and it's also not a high priority configuration nowadays.

talregev:
> To not increase the number of flaky jobs, what about migrating `libssh` to an existing job and deleting the `!ssl` one? Build without SSL are tested widely in CI and it's also not a high priority configuration nowadays.

We can do it in different PR.

CMakeLists.txt
talregev: done.

vszakats:
Can you make such PR? We can merged this one when that's merged.

talregev:
I can create such PR.
Can you tell me where you want the libssh?

vszakats:
What about moving it inside this new GnuTLS job? Can be done within this PR as a bonus.

talregev:
ok 😅

.github/workflows/windows.yml
```diff
@@ -404,6 +404,13 @@ jobs:
           type: 'Debug'
           tflags: '~1516'
           config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=OFF -DCURL_USE_SCHANNEL=OFF -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DCURL_USE_LIBPSL=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_WOLFSSL=ON -DCURL_USE_GSASL=ON'
+        - name: 'gnutls'
+          install: 'brotli zlib zstd libpsl nghttp2 shiftmedia-libgnutls libssh2 pkgconf gsasl'
```
dg0yt, line 408: FTR there is no pristine GnuTLS for MSVC. There is the fork from ShiftMedia and its vcpkg port. Quality unknown.

CMakeLists.txt
```diff
@@ -580,7 +580,16 @@ if(CURL_USE_WOLFSSL)
 endif()
 
 if(CURL_USE_GNUTLS)
-  find_package(GnuTLS REQUIRED)
+  if(NOT MSVC OR VCPKG_TOOLCHAIN)
```
dg0yt, line 583: Reminder: vcpkg doesn't want packages to switch behaviour on vcpkg indicators. There is no need to block choice for conan, spack, `FetchContent`, whatever.

.github/workflows/windows.yml
talregev: I think the problem is coming from ngtcp2. when I adding only shiftmedia-libgnutls it not create this.

talregev: Also because I adding the shiftmedia-libgnutls to curl tests, we can estimate the quality of this port.

talregev:
@vszakats I remove the !ssl job.

.github/workflows/windows.yml
vszakats: `ngtcp2`'s crypto lib links to the Release GnuTLS DLL instead of the Debug one. Isn't this closer to the root cause?

CMakeLists.txt
vszakats: So should we restore the `UNIX` condition here? Otherwise it's known to cause breakage for people trying to build curl!

vszakats:
Thanks for merging the jobs @talregev!

.github/workflows/windows.yml
vszakats: OK, answer here: https://github.com/curl/curl/pull/14495#issuecomment-2284542160

dg0yt: This thread wasn't about the root cause.

talregev:
@vszakats I solve the conflicts.

talregev:
@vszakats better to have this PR: https://github.com/curl/curl/pull/14495

vszakats:
Thank you, merged via #14495!

PR #14499, opened 2024-08-12
unique identifier inside a multi, data->mid

icing:
`data->id` is unique in *most* situations, but not in all. If a libcurl application uses more than one connection cache, they will overlap. This is an uncommon situation, but some libcurl apps may do crazy things.
For informative purposed, like tracing, `data->id` is good. It assigns new ids in curl's serial curl_easy_perform() use, for example.

Introduce `data->mid` which is a unique identifier inside one multi instance, assigned on multi_add_handle() and cleared on multi_remove_handle(). For tracing, this is not the best, as curl_easy_perform() will give every transfer the multi id 0. However, when associating data *inside the same multi* with a transfer, it gives a unique id for lookups.

Use the `mid` in DoH operations to track relationships between a transfer and its DoH handles. Also, use `mid` to track  h2/h3 streams for transfers.

### DIFF
--- a/lib/doh.c
+++ b/lib/doh.c
@@ -214,19 +214,28 @@ static void local_print_buf(struct Curl_easy *data,
 /* called from multi.c when this DoH transfer is complete */
 static int doh_done(struct Curl_easy *doh, CURLcode result)
 {
-  struct Curl_easy *data = doh->set.dohfor;
-  struct dohdata *dohp = data->req.doh;
-  /* so one of the DoH request done for the 'data' transfer is now complete! */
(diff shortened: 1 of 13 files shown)


bagder (approved):

bagder:
If you rebase this on master it does not build anymore because of ba235ab26. Can you adjust?

icing:
> If you rebase this on master it does not build anymore because of [ba235ab](https://github.com/curl/curl/commit/ba235ab269080dc66e35835c829f7ac4290dbc1d). Can you adjust?

Done.

PR #14503, opened 2024-08-12
hash: provide asserts to verify API use

bagder:
- converted the Curl_hash_count() macro to a function
- Discourage accessing struct fields directly
- Document the API


bagder:
Needs #14505

PR #14495, opened 2024-08-11
GHA/windows: add GnuTLS to MSVC jobs

vszakats:
- enable GnuTLS in MultiSSL job.

- add new GnuTLS job with HTTP/3 enabled.
  Replacing the `!ssl` job.
  Use `Release` mode to work around a vcpkg package issue, ending up
  depending on both the debug and release versions of the gnutls DLL,
  one of them miss to copy next to the binary.

- cmake: add `pkg-config` detection for GnuTLS.

Co-authored-by: Tal Regev
Fixes #14494

PR #14519, opened 2024-08-12
IDN: fix/extend/migrate test exclusion rules

### LINKED ISSUE #14517: Curl crash with SSL connection
### I did this

Hi Team,
I am using curl 8.7.1
language - C
I am getting below crash when I send high volume of requests to one of our SSL servers. Non-ssl works fine. Please advise!

(dbx) where
.(??, ??, ??, ??, ??, ??, ??, ??, ??, ??, ??, ??, ??, ??, ??) at 0x100000000
look_str_cb@AF19_12() at 0x10063efac
int_cb_LHASH_DOALL_ARG() at 0x100544f58
doall_util_fn@AF15_10() at 0x1005243e8
engine_table_doall() at 0x100544fa8
ENGINE_pkey_asn1_find_str() at 0x10063e8dc
EVP_PKEY_asn1_find_str() at 0x1005a7b88
get_optional_pkey_id() at 0x1007c8c28
IPRA.$ssl_cipher_get_disabled() at 0x1007c9858
ssl_create_cipher_list() at 0x1007caf6c
SSL_CTX_set_cipher_list() at 0x1007c2644
ossl_connect_step1() at 0x1007b9508
ossl_connect_common() at 0x1007bc4f8
Curl_ssl_connect_nonblocking() at 0x1007b0b7c
Curl_http_connect() at 0x10085eb04
IPRA.$multi_runsingle() at 0x100825eec
curl_multi_perform() at 0x100827278
easy_perform() at 0x1007aea10
internal error: unexpected value 120 at line 5214 in file stabstring.c
internal error: unexpected value 120 at line 5214 in file stabstring.c
internal error: unexpected value 120 at line 5214 in file stabstring.c

### I expected the following

_No response_

### curl/libcurl version

curl 8.7.1

### operating system

AIX 7.2

vszakats:
- make sure to exclude failing tests when libidn2 is detected by
  default.

- ignore test 1560 results. Seen to fail with libidn2.
  I'm not sure why this test was not executed earlier:
  https://github.com/curl/curl/actions/runs/10354610889/job/28660309355#step:13:3647

- runtests: recognize `libidn2` as a feature.

- move IDN test exclusions from GHA/windows to `tests/data/DISABLED`.

- GHA/windows: drop default `-DUSE_LIBIDN2=ON` cmake config.

Cherry-picked from #14495
Closes #14519

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -480,12 +480,8 @@ jobs:
         timeout-minutes: 12
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP !FTP ${{ matrix.tflags }}'
-          if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
-            if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
-              TFLAGS+=' ~1448 ~2046 ~2047'
-            fi
-          elif [[ '${{ matrix.config }}' = *'-DUSE_LIBIDN2=ON'* ]]; then
-            TFLAGS+=' ~165 ~1448 ~2046 ~2047'
+          if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
+            TFLAGS+=' ~165 ~1448 ~1560 ~2046 ~2047'  # IDN tests failing without Unicode. 165 is for libidn2, for WinIDN it is disabled globally
           fi
           PATH="$PWD/bld/lib:$PATH"
           cmake --build bld --config '${{ matrix.type }}' --target test-ci
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -480,8 +480,5 @@ jobs:
         timeout-minutes: 12
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP !FTP ${{ matrix.tflags }}'
-          if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
-            TFLAGS+=' ~165 ~1448 ~1560 ~2046 ~2047'  # IDN tests failing without Unicode. 165 is for libidn2, for WinIDN it is disabled globally
-          fi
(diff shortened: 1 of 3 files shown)


jay:
nit: commit message has wrong pr num

vszakats:
@jay: Thanks, will overwrite it with the PR message before merge.

.github/workflows/windows.yml
```diff
@@ -480,12 +480,8 @@ jobs:
...
         run: |
           export TFLAGS='-j14 !TFTP !MQTT !WebSockets !SMTP !FTP ${{ matrix.tflags }}'
-          if [[ '${{ matrix.config }}' = *'-DUSE_WIN32_IDN=ON'* ]]; then
-            if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
-              TFLAGS+=' ~1448 ~2046 ~2047'
-            fi
-          elif [[ '${{ matrix.config }}' = *'-DUSE_LIBIDN2=ON'* ]]; then
-            TFLAGS+=' ~165 ~1448 ~2046 ~2047'
+          if [[ '${{ matrix.config }}' != *'-DENABLE_UNICODE=ON'* ]]; then
+            TFLAGS+=' ~165 ~1448 ~1560 ~2046 ~2047'  # IDN tests failing without Unicode. 165 is for libidn2, for WinIDN it is disabled globally
```
bagder, line 484: Is it not better to maintain these ignores more globally in DISABLED?

I think it makes it easier to get an overview of what is ignored, plus if these are true problems with combinations then surely others who run the tests also want the same ignores?

vszakats: Agreed, it'd be better. I did not know that the global disables till just recently.

I'll give it a try.

vszakats: After testing: It's OK, but different than putting tests on ignore. With `DISABLED` the test disappears from the log without a trace. I'd be useful to see it there to verify that it's actually disabled. (In this case it wasn't, until adding `libidn2` to the `$feature` map in `runtests.pl`.)

Also tflags-ignore _runs_ the tests, making it easy to inspect what's wrong with it.

`DISABLED` seems suited for test with no intent to fix. I think these aren't like that, though fixing some of them is non-trivial (Unicode.)

PR #14530, opened 2024-08-13
cmake: use `pkg-config` as a hint for detecting libpsl

vszakats:
Also:
- sync up internally with a reference Find module (nghttp2).
- use standard header/lib output variable names.

Follow-up to 9b3f67e267d1fa8d7867655d133bdbf8830a0ab3 #12947
Closes #14530


vszakats:
Closing in favour of #14545 and db39c668a8e33e064b9eb20892cd027f46302f77 #14542.

PR #14532, opened 2024-08-13
mime client reader, avoid infinite loop

icing:
Curl_mime_read() may go into an infinite loop when called with buffer lengths <= 4. Some encoders, like base64, are not prepared for that.

In the client reader for mime data, skip such small reads. The upload buffers will get flushed eventually and larger reads will happen again.

Improves robustness of test652 which triggered the loop on blocked sends.

Refs #14527.


bagder:
scanbuild seems to have an opinion here!

icing:
Rebased and removed the init line.

PR #14544, opened 2024-08-14
Curl_sendrecv() improvements

icing:
Renames Curl_readwrite() to Curl_sendrecv() to reflect that his is mainly about talking to the server, not reads or writes to the client. Add a `nowp` parameter since the single caller already has this.

Add `Curl_has_expired()` that checks if `data->state.expiretime` has been reached. This member carries always the earliest time when expiration timers are added/removed.

When `Curl_sendrecv()` sees the transfer has expired, it always tries to send and/or receive where applicable without checking the socket POLL states.

For non-expired transfers that have also no `data->state.select_bits` set, check if the last poll information of the transfer mentioned any sockets at all. If not, probably as the transfer is paused or otherwise hindered from progressing, do not send/recv.

However if the transfer's poll mentioned any sockets, do try send/recv where applicable. This replaces the former POLL check on the involved sockets.

It seems better to DO send/recv and get an EAGAIN than polling the sockets, because the relationship between the POLL state and the transfer's ability is not clear for all protocols and proxy combinations. It is more robust to just try it.

In order to assess a transfer's last poll information, always use `data->last_poll` when collecting this. The event based handling already did so.

### DIFF
--- a/lib/easy.c
+++ b/lib/easy.c
@@ -1121,7 +1121,7 @@ CURLcode curl_easy_pause(struct Curl_easy *data, int action)
                   (data->mstate == MSTATE_PERFORMING ||
                    data->mstate == MSTATE_RATELIMITING));
   /* Unpausing writes is detected on the next run in
-   * transfer.c:Curl_readwrite(). This is because this may result
+   * transfer.c:Curl_sendrecv(). This is because this may result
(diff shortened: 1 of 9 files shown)


icing:
Replaced by #14561

PR #14343, opened 2024-08-01
cmake: add math library when compile wolfSSL and ngtcp2

talregev:
Add math library when compile wolfSSL and ngtcp2 in non-Windows platforms.

When compile wolfSSL with ngtcp2 without this fix on linux:

```
FAILED: src/curl 
[...]

PR #14549, opened 2024-08-15
cmake: Link to math lib without full path

talregev:
Link to math lib without full path

According to @dg0yt comment:
https://github.com/curl/curl/pull/14343#issuecomment-2272712340


dg0yt:
Reminder: I also wrote that this this the wrong spot for the task.

From the pkg-config files, the picure is that libm is a transitive usage requirement of WolfSSL. CURL has a Find module for WolfSSL. It is the responsibility of that module to populate `WolfSSL_LIBRARIES` with everything which needs to be linked for using WolfSSL.

CMakeLists.txt
```diff
@@ -788,6 +788,9 @@ if(USE_NGTCP2)
       find_package(NGTCP2 REQUIRED wolfSSL)
       if(NOT WIN32)
         find_library(MATH_LIBRARY m)
+          if(MATH_LIBRARY)
+            set(MATH_LIBRARY m)
+          endif()
```
vszakats, line 793:
```suggestion
        if(MATH_LIBRARY)
          set(MATH_LIBRARY "m")
        endif()
```

vszakats:
> Reminder: I also wrote that this this the wrong spot for the task.
> 
> From the pkg-config files, the picure is that libm is a transitive usage requirement of WolfSSL. CURL has a Find module for WolfSSL. It is the responsibility of that module to populate `WolfSSL_LIBRARIES` with everything which needs to be linked for using WolfSSL.

Agreed.

talregev:
> > Reminder: I also wrote that this this the wrong spot for the task.
> > From the pkg-config files, the picure is that libm is a transitive usage requirement of WolfSSL. CURL has a Find module for WolfSSL. It is the responsibility of that module to populate `WolfSSL_LIBRARIES` with everything which needs to be linked for using WolfSSL.
> 
> Agreed.

@vszakats Do you still want to continue with this PR?

CMakeLists.txt
```diff
@@ -788,6 +788,9 @@ if(USE_NGTCP2)
       find_package(NGTCP2 REQUIRED wolfSSL)
       if(NOT WIN32)
         find_library(MATH_LIBRARY m)
```
vszakats, line 790:
```suggestion
        find_library(MATH_LIBRARY "m")
```

vszakats:
I think it improves the previous patch, but to avoid adding tech debt, I'd prefer to follow @dg0yt's suggestion.

Would you want to make an attempt to move this logic inside `CMake/FindWolfSSL.cmake`?

CMakeLists.txt
```diff
@@ -788,6 +788,9 @@ if(USE_NGTCP2)
...
+        if(MATH_LIBRARY)
+          set(MATH_LIBRARY "m")
```
vszakats, line 792:
```suggestion
        find_library(_wolfssl_math_library "m")
        if(_wolfssl_math_library)
          set(_wolfssl_math_library "m")
```

talregev:
I am not near computer, and I think your suggestions is important.
I don't mind that you will take over.
I will put this PR on draft.

vszakats:
OK, I'm making an attempt in #14576.

talregev:
> OK, I'm making an attempt in #14576.

Thank you for your hard work! and improve cmake curl, and curl itself!

PR #14551, opened 2024-08-15
tests: move the disabling of 500 for hyper from CI to DISABLED

bagder:
Follow-up to 136504195ae63


bagder:
I can reproduce the torture test fail on master.

PR #14555, opened 2024-08-15
cmake: migrate dependency detections to Find modules

vszakats:
For: libgsasl, libidn2, libssh, libuv.

The new Find modules retain using `pkg-config` natively, not as a "hint"
for the CMake-native detection. Of the pre-existing Find modules, only
FindNettle, and FindGSS (with customized code) work this way. Align
detection code for the new modules and add version detection for the
CMake-native paths.

Also, add CMake-native detection for `libgsasl`.

The remaining outlier in `CMakeLists.txt` is GnuTLS, which has
a CMake built-in Find module, but which lacks `pkg-config` support,
required for vcpkg. It remains unchanged.

Another part-outlier is `libssh`, which keeps requiring the trick
`find_package(libssh CONFIG QUIET)` for reasons I could not yet figure
out.

---

- [x] decide on Find module naming scheme.
- [x] disable `pkg-config` if custom config is set via `<NAME>_INCLUDE_DIR` / `<NAME>_LIBRARY`? (Separate PR?)
- [x] decide if double-checking of idn2 libs and header is really necessary. [→ PROBABLY NOT. DELETED]
- [x] robustify version detection with `if(EXISTS)`, more flexible regexp and synced var names. → #14572
- [ ] once there is more experience with this method and it turns out superior, consider changing the other Find modules. If this turns out worse, consider moving back to the "hint" way with these.


vszakats:
Version detection already works with krb5, or did you mean something else?:
https://github.com/curl/curl/actions/runs/10408884975/job/28827262755#step:6:62

talregev:
> Version detection already works with krb5, or did you mean something else?: https://github.com/curl/curl/actions/runs/10408884975/job/28827262755#step:6:62

I meant that it will be in here.
https://github.com/curl/curl/actions/runs/10408884975/job/28827262755#step:6:62

talregev:
> Version detection already works with krb5, or did you mean something else?: https://github.com/curl/curl/actions/runs/10408884975/job/28827262755#step:6:62

That a good a nice step. I like it!

vszakats:
> https://github.com/curl/curl/actions/runs/10408884975/job/28827262755#step:10:10

Ah, got it, to the `-V` output. These seem pretty old and steady components and their versions have little impact on functionality, perhaps that's why it's not there? It's out of scope for this PR, but you may open an Issue/Discussion about it.

talregev:
> > https://github.com/curl/curl/actions/runs/10408884975/job/28827262755#step:10:10
> 
> Ah, got it, to the `-V` output. These seem pretty old and steady components and their versions have little impact on functionality, perhaps that's why it's not there? It's out of scope for this PR, but you may open an Issue/Discussion about it.

I think is very important, because now when I got curl from ubuntu, I don't know which library was compile to get gsasl-api feature, and what the version it there. If users know, they can easly report to ubuntu (or other distro) or get decision if to use that curl. 
I already create an issue about it, and it on known issue here on curl.
I agree that we need to deal with that in separate PR.
Also I was thinking to add find_krb5, and find_hemidal and move the logic to that files from main cmake.
We can also do it in separate PR.

vszakats:
> Also I was thinking to add find_krb5, and find_hemidal and move the logic to that files from main cmake.
> We can also do it in separate PR.

Do you mean to split FindGSS into FindHeimdal and Findkrb5, or?

talregev:
> > Also I was thinking to add find_krb5, and find_hemidal and move the logic to that files from main cmake.
> > We can also do it in separate PR.
> 
> Do you mean to split FindGSS into FindHeimdal and Findkrb5, or?

If it already happen in  FindGSS that is good, and not need to change.

talregev:
I see that file now, and that is a good direction.

In the future, if we see that the logic is large in FindGSS, we can think to split. Maybe it will be more easy to handle? 
What happen if curl detect both hemidal and krb5. What it will choose?

vszakats:
> > > Also I was thinking to add find_krb5, and find_hemidal and move the logic to that files from main cmake.
> > > We can also do it in separate PR.
> > 
> > 
> > Do you mean to split FindGSS into FindHeimdal and Findkrb5, or?
> 
> If it already happen in FindGSS that is good, and not need to change.

It does. `pkg-config` detection wasn't working up until recently.
They are now both tested in GHA for macos/linux and Heimdal on NetBSD.

They are certainly not over-tested, so more tests and reports are welcome.

PR #14563, opened 2024-08-15
internals/SPLAY.md: internal API documentation

bagder:


bagder:
I will update this once #14562 lands, as it adds some minor tweaks.

PR #14581, opened 2024-08-18
appveyor: drop uploading artifacts

vszakats:
Uploading artifacts sometimes results in this error:
```
Uploading artifacts...
[1/1] _bld\src\curl.exe (2,022,912 bytes)...100%
Error uploading artifact to the storage: Remote server returned 503: Service Temporarily Unavailable
```
Ref: https://ci.appveyor.com/project/curlorg/curl/builds/50424126/job/e4envval6xkicv1i#L123

The artifacts are also probably not very useful to upload for every run.
Let me know if they are or if there is a better solution to the
intermittent failures. Also note that they were missing external DLL
dependencies.

Leave the logic there commented, to make it easy to enable as needed for
debugging or testing artifacts locally.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -211,8 +211,8 @@ skip_commits:
     - 'packages/**/*'
     - 'plan9/**/*'
 
-artifacts:
-  - path: '**/curl.exe'
-    name: curl
-  - path: '**/*curl*.dll'
-    name: libcurl dll
+#artifacts:
+#  - path: '**/curl.exe'
+#    name: curl
+#  - path: '**/*curl*.dll'
+#    name: libcurl dll


bagder:
What are these uploaded artifacts?

vszakats:
It's curl.exe and libcurl.dll.

bagder:
Ok. I doubt they are used very often. If ever.

vszakats:
I just tried to use them while testing another PR. Turns out artifacts with OpenSSL don't work because of missing OpenSSL DLLs. Meaning, its usefulness is also limited to Schannel builds. All that is fixable, but don't seem to be worth the hassle.

bagder:
We also have rather few developers actually using Windows for which said artifacts could be useful. We often fix Windows related curl problems without being able to test-run anything natively locally.

bagder (approved):

PR #14589, opened 2024-08-19
tests: improve test_17_07_ssl_ciphers

jan2000:
Change TLS proto version on the test server to test setting combinations of --tls13-ciphers and --ciphers.

Overloading of autouse fixtures does not seem to work. For the test httpd server to be reloaded with a clean config in test_18_methods, to not be affected by the config changes in test_17_ssl_use, the two class scope fixtures of test_18_methods are now combined.

Added return value to set_extra_config and clear_extra_configs to indicate changes, making it possible to prevent unnecessarily reloading of the test httpd server.

### DIFF
--- a/tests/http/test_18_methods.py
+++ b/tests/http/test_18_methods.py
@@ -44,10 +44,7 @@ class TestMethods:
         if env.have_h3():
             nghttpx.start_if_needed()
         httpd.clear_extra_configs()
-        httpd.reload()
-
-    @pytest.fixture(autouse=True, scope='class')
-    def _class_scope(self, env, httpd):
+        httpd.reload_if_config_changed()
         indir = httpd.docs_dir
         env.make_data_file(indir=indir, fname="data-10k", fsize=10*1024)
         env.make_data_file(indir=indir, fname="data-100k", fsize=100*1024)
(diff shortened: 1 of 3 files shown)


jan2000:
@icing especially notice the part about the double `_class_scope`. These are also in other pytest classes. Maybe I misunderstand it, but it seems these need to be fixed as well.

tests/http/testenv/httpd.py
```diff
@@ -104,13 +104,21 @@ def exists(self):
         return os.path.exists(self._cmd)
 
     def set_extra_config(self, domain: str, lines: Optional[Union[str, List[str]]]):
+        if domain not in self._extra_configs:
```
icing, line 107: The name `set_extra_confg` was intended to replace. If you need an `add_extra_config` it would be better to make that a new method.

tests/http/test_18_methods.py
```diff
@@ -43,11 +43,8 @@ class TestMethods:
     def _class_scope(self, env, httpd, nghttpx):
         if env.have_h3():
             nghttpx.start_if_needed()
-        httpd.clear_extra_configs()
-        httpd.reload()
-
-    @pytest.fixture(autouse=True, scope='class')
-    def _class_scope(self, env, httpd):
+        if httpd.clear_extra_configs():
+            httpd.reload()
```
icing, line 47: good catch!

icing (changes requested):

jan2000:
The previous method of preventing unnecessarily reloading of the test httpd server did not properly prevent bleeding of config into the next test. Changed to other method.

icing (approved):

PR #14590, opened 2024-08-19
tests: add test_17_09_ssl_min_max

jan2000:
Test setting all combinations of --tlsv1.x and --tls-max.

Depends on #14589.

### DIFF
--- a/tests/http/testenv/httpd.py
+++ b/tests/http/testenv/httpd.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #***************************************************************************
 #                                  _   _ ____  _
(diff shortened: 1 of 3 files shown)


tests/http/testenv/httpd.py
```diff
@@ -104,13 +104,21 @@ def exists(self):
         return os.path.exists(self._cmd)
 
     def set_extra_config(self, domain: str, lines: Optional[Union[str, List[str]]]):
+        if domain not in self._extra_configs:
```
icing, line 107: As in #14589, the `set_extra` is really mean to replace, not append. Besides purity of naming, remember that test cases *should* be able to run standalone and not rely on settings from previously run tests.

icing (changes requested):
Apart from the `set_extra` changes, the additional tests are a good idea.

icing (approved):

PR #14591, opened 2024-08-19
mbedtls: no longer use MBEDTLS_SSL_VERIFY_OPTIONAL

jan2000:
With mbedTLS if the minimum version of TLS is set to 1.3, MBEDTLS_SSL_VERIFY_OPTIONAL is not available in client mode. See: https://github.com/Mbed-TLS/mbedtls/blob/2ca6c285/library/ssl_tls.c#L1357
Also, there might be plans to remove it completely in future mbedTLS versions.

So, switch to always using MBEDTLS_SSL_VERIFY_REQUIRED. If verifypeer or verifyhost are disabled the corresponding error flags are cleared in the verify callback function. That is also where verification errors are logged.

Depends on #14588

### DIFF
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -541,12 +541,20 @@ static int mbed_verify_cb(void *ptr, mbedtls_x509_crt *crt,
       mbed_extract_certinfo(data, crt);
   }
 
-  /* we clear any faults the mbedtls' own verification found if needed.
-   * See <https://github.com/Mbed-TLS/mbedtls/issues/9210> */
   if(!conn_config->verifypeer)
     *flags = 0;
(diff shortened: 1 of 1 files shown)


bagder:
Note: "This branch has conflicts that must be resolved". Can you take a look?

jan2000:
Hmm.. I am not quite sure what the best way is to do continuation PRs (this PR depends on that PR) with github. When I rebased locally there were no conflicts. Its a shame that github does not (at least that I known) provide such functionality.

vszakats:
There was a merge queue feature announced earlier: https://github.blog/changelog/2023-07-12-pull-request-merge-queue-is-now-generally-available/
(that's all I know about it)

bagder (approved):

PR #14594, opened 2024-08-19
try fix for 1299 1613 with MSYS2

vszakats:


tests/data/test1299
```diff
@@ -36,6 +36,11 @@ Send "OPTIONS *" with --request-target
 <command>
 --request-target '*' -X OPTIONS http://%HOSTIP:%HTTPPORT/ -H "Testno: %TESTNUMBER"
 </command>
+<setenv>
+# Needed for MSYS2 to not treat the '*' as a POSIX path list
+# that has to be converted to Windows paths
+MSYS2_ARG_CONV_EXCL=*
```
bagder, line 42: Are there any cases where this functionality is desired for tests? Maybe we could just disable this once and for all in runtests.pl?

vszakats: Yes, I think it could be global. But, in this case the trick didn't work, and I wonder what controls the expansion. (assuming it's indeed glob-related.)

I also wonder what is the catch-all setting that disables all MSYS2 and mingw-runtime command-line weirdness.

There was also this in earlier versions of `appveyor.yml`:
```
MSYS2_ARG_CONV_EXCL: '/*'
```
https://www.msys2.org/wiki/Porting/

dfandrich:
Analysis of PR #14594:

[Test 1299 failed](https://github.com/curl/curl/actions/runs/10449311303), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1613 failed](https://github.com/curl/curl/actions/runs/10449311303), which has NOT been flaky recently, so **there could be a real issue in the PR**.

vszakats:
I think without replicating it in a local Windows env, this is tricky to figure out. Closing for now.

PR #14598, opened 2024-08-19
build: use -Wno-format-overflow

### LINKED ISSUE #14168: http: gcc 14 compiler warning on mac arm64 (cmake, unity)
### I did this

Built curl-for-win with gcc for mac (arm64), and this compiler warning happened:
```
curl/lib/http.c: In function 'Curl_add_custom_headers':
curl/lib/http.c:1528:42: warning: '%s' directive argument is null [-Wformat-overflow=]
 1528 |             result = Curl_dyn_addf(req, "%s\r\n", compare);
      |                                          ^~
```
https://github.com/curl/curl-for-win/actions/runs/9906570816/job/27368394282#step:3:4688

### I expected the following

No warnings.

### curl/libcurl version

master f94aa3d97f1064e29446fbb29a205493b251bac1

### operating system

macOS 14 arm64 GHA runner

bagder:
-Wformat-overflow is not a warning that we want enabled as it does not help us. It can only bring us false positives since it warns on bad uses of sprintf and vsprintf ("that might overflow the destination buffer"). Two functions we explicitly ban in curl code.

The only way this flag triggers warnings in curl code is false positives for functions we have marked with the CURL_PRINTF() macro.

Further: it seems -Wformat-trunaction option might in turn also enable -Wformat-overflow, so if this second option is used, we need to explicitly set -Wno-format-overflow - not just skip setting -Wformat-overflow.

Reported-by: Viktor Szakats
Fixes #14168

### DIFF
--- a/CMake/PickyWarnings.cmake
+++ b/CMake/PickyWarnings.cmake
@@ -206,7 +206,7 @@ if(PICKY_COMPILER)
         list(APPEND WPICKY_ENABLE
           -Walloc-zero                     #             gcc  7.0
           -Wduplicated-branches            #             gcc  7.0
-          -Wformat-overflow=2              #             gcc  7.0
+          -Wno-format-overflow             #             gcc  7.0
           -Wformat-truncation=2            #             gcc  7.0
           -Wimplicit-fallthrough           # clang  4.0  gcc  7.0
           -Wrestrict                       #             gcc  7.0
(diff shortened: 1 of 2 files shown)


vszakats:
> Further: it seems -Wformat-trunaction option might in turn also enable -Wformat-overflow, so if this second option is used, we need to explicitly set -Wno-format-overflow - not just skip setting -Wformat-overflow.

Which gcc version have you seen doing this?

Tried to reproduce with godbolt.org with `-Wformat-truncation=2`, but couldn't get the warning (without explicitly adding `-Wformat-overflow`):
https://godbolt.org/z/EjEsn36dY

bagder:
This was on my local machine using: `gcc (Debian 14.2.0-2) 14.2.0`

vszakats (approved):

PR #14600, opened 2024-08-19
curl: find curlrc in XDG_CONFIG_HOME without leading dot

### LINKED ISSUE #12129: curl tool won't search for curlrc (no leading dot) in XDG_CONFIG_HOME

bagder:
If XDG_CONFIG_HOME is set, look for XDG_CONFIG_HOME/curlrc - without a leading dot in the filename.

Fixes #12129
Reported-by: Jat Satiro

### DIFF
--- a/src/tool_findfile.c
+++ b/src/tool_findfile.c
@@ -51,7 +51,7 @@ struct finder {
    in the findfile() function */
 static const struct finder conf_list[] = {
   { "CURL_HOME", NULL, FALSE },
-  { "XDG_CONFIG_HOME", NULL, FALSE }, /* index == 1, used in the code */
+  { "XDG_CONFIG_HOME", NULL, TRUE },
   { "HOME", NULL, FALSE },
 #ifdef _WIN32
   { "USERPROFILE", NULL, FALSE },
@@ -102,7 +102,6 @@ static char *checkhome(const char *home, const char *fname, bool dotscore)
 char *findfile(const char *fname, int dotscore)
 {
   int i;
-  bool xdg = FALSE;
   DEBUGASSERT(fname && fname[0]);
   DEBUGASSERT((dotscore != 1) || (fname[0] == '.'));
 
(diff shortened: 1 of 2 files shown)


dfandrich:
Analysis of PR #14600:

[Test 433 failed](https://circleci.com/gh/curl/curl/103994), which has NOT been flaky recently, so **there could be a real issue in the PR**.  Note that this test has failed in 6 different CI jobs (the link just goes to one of them).

Issue #14580, opened 2024-08-17
cmakelint warnings


dfandrich:
### I did this

I tried running cmakelint on the curl sources with the command `cmakelint --spaces=2 --linelength=132 CMake/* $(find . -type f -name CMake\*) ` and it showed [these warnings](https://github.com/user-attachments/files/16646644/cmakelint.txt).

I don't have an opinion on the veracity of these warnings (@vszakats probably would), but if they look useful, this would be easy to add to a CI job.

### I expected the following

cmakelint to be warning-free.

### curl/libcurl version

8.10.0-DEV

### operating system

Linux

vszakats:
Nice!

vszakats:
Pushed some fixes to this existing PR #14579

1. I was just wondering about the naming of Find* modules. It's a bit all over the place. Also CMake built-in ones are.
`cmakelint` suggests full uppercase component names after `Find`. I'm fine with this, but as this `cmakelint` Issue [notes](https://github.com/cmake-lint/cmake-lint/issues/24), CMake itself [doesn't suggest](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names) such convention and I also cannot see it confirmed "in the wild".

I'd probably just do capital and then use whatever the component is named, and leave most existing uppercased ones as-is, changing only these:
```
FindLibPSL  -> FindLibpsl
FindLibSSH2 -> FindLibssh2
FindQUICHE  -> FindQuiche
Findrustls  -> FindRustls
```

Either way it will need a separate PR just for that. But first what scheme should we follow here, if any?

2. As for the "weird indentation" issues: It caught two that I was aware of (fixed now!), but the rest I'm conflicted with. `cmakelint` wants to indent multiline `if(` expression to the 2-char levels and calls it weird when aligned with the preceding `if(`. This recommendation IMO reduce readability for these expressions.

Found these these two solution:
- avoid multiline expressions. By unfolding or splitting to nested `if()`s. This is also weird, and may not be applicable to all expressions without reducing readability.
- suppress the warning with `--filter=-whitespace/indent`. This looks harsh, but this option doesn't seem to catch real indentation issues anyway. (tried by unindenting, overindenting instructions under a `function` and `if`, neighter triggerring warnings.

dfandrich:
I noticed that multiline indentation issue as well. That might be worth
bringing upstream as a bug, since indenting while ignoring the parenthesis
is definitely a step down in readability.

[...]

PR #14601, opened 2024-08-19
cmake: rename Find modules

vszakats:
- `FindCARES` -> `FindCares`
- `FindLibPSL` -> `FindLibpsl`
- `FindLibSSH2` -> `FindLibssh2`
- `FindQUICHE` -> `FindQuiche`
- `Findrustls` -> `FindRustls`

Our convention for naming Find modules (the part after the `Find`
prefix, also called as 'package name') is:

Start with uppercase. Follow with lowercase, unless there is a clear
preference for a stylized name. E.g. the project itself uses it that way
with a matching `<Name>Config.cmake` file, or we use it that way
elsewhere, or the name is an acronym.

Ref: #14580


dg0yt:
I noticed this renaming while refreshing the vcpkg patches. I'm unhappy from a CMake POV:  
The package name (`Cares`) no longer matches the variable prefix (`CARES`). 

I know that CMake isn't consistent with this in its Find modules, and it is not uncommon to see upper-case prefixes for mixed-case package names. But:
- For Config mode, exact match is cmake-provided behavior for `<Pkg>_FOUND` and `<Pkg>_DIR`.
- Even for Module mode, it is reasonable to expect "case-preserved" (and spelled literally in docs as) `<PackageName>_FOUND`. "Case-preserved" match also applies to `<PackageName>_ROOT`,  `CMAKE_REQUIRE_FIND_PACKAGE_<PackageName>`,  `CMAKE_DISABLE_FIND_PACKAGE_<PackageName>`.

(FTR some CMake find modules explicitly set both variants of `<Pkg>_FOUND` for historical reasons, others as an unexpected side effect of `select_library_configurations`.)

CMake can't break compatibility. But for CURL, it is a step back IMO.

Surprisingly, this PR was triggered by a cmakelint report with this (debatable) recommendation:
~~~
Find modules should use uppercase names; consider using FindBROTLI.cmake [convention/filename]
~~~

dg0yt:
Another reflection from editing the patches: `pkg_check_modules` is case-preserving with regard to variable prefix, too.

vszakats:
@dg0yt Thanks for your feedback. We've been publicly contemplating on this for a while (see also #14580), with no feedback and absolutely zero guidance or hint from any public sources, including the authors of CMake (that I could find). We also had a bug with BearSSL package name, ref #14599, plus more related in other PRs.

Can you elaborate what alternate/superior action are you suggesting?

Also correct me if I'm wrong, but as far I understand the Find modules used and maintained within curl are private at this point. They are used from curl's CMakeLists.txt only and not installed for public use.

I understand "internal" changes may cause issues when curl's CMake is used in less-common (?) ways. But is there any consequence of this affecting people using the well-documented / common ways (.e.g <UPPERCASENAME>_INCLUDE_DIR` and `<UPPERCASENAME>_LIBRARY` to control detection of these dependencies under curl?

If there is a widely-relied-on internal variable that's lost by these changes, it's possible to restore those for compatibility, but for this we'll need to know what these are. For these or any other fallouts I appreciate an isolate example to better undertstand what's happening and reprdocue it locally.

---

FWIW I also found the "case-preserved" suggestion practically useless, as there is no way to find out what case to preserve. Sometimes what to preserve seems arbitrary or wrong ("LibSSH2"). Also skewed by forcing a capital letter in most cases. All the while most packages refer to themselves in lowercase. CMake behaviour also depends on filesystem case-sensitivity, which fact is missing from the docs (that I could find). While CMake itself demonstrates no consistency or observable rule at all.

According to days of experimenting with this, I concluded these must match, and a match with any other variable name is not a requirement (save compatibility with those who relied on curl CMake "internal" stuff), nor done in practice (via #14599):

Package names must match case-sensitively to work on all platforms:
- `find_package(<NAME> ...)` in `CMakeLists.txt`.
- `CMake/Find<NAME>.cmake` filenames.
- `find_package_handle_standard_args(<NAME> ...` in Find modules.
- `message(STATUS "Found <NAME> ...` in Find modules.
  (to match the message shown by `find_package_handle_standard_args()`)

The actual variable names can remain uppercase regarless of the above, and this is what's used in most cases by most projects, including CMake itself.

All that said, I repeat here again that it's also an option to rename everything to uppercase, like `cmakelint` wants and call it a day. It might make thing more consistent at the end of the day, because understanding what must match with what looks close to a mission impossible.

vszakats:
> Disclaimer: The vcpkg port doesn't benefit, no matter how the name is chosen. Find modules just aren't good enough. Admittedly, the patch looks would less ugly with consistent case. Now it is like:
> 
> ```diff
> -  find_package(Cares REQUIRED)
> +  find_package(CARES NAMES c-ares CONFIG REQUIRED)
> ```

How did this patch look before? Also how would the patched version find `FindCares.cmake` on a case-sensitive filesystem? It's also a good example on why there is no good general answer, c-ares uses a dash in it's public config files. (On a slightly unrelated note, why can't our own Find modules do the mapping to `c-ares` and make this package be found in Config-mode, too? Why does this need to be patched?)

What do you mean by "Find modules just aren't good enough"? curl Find modules or Find modules in general? Can we do something in this project? In earlier posts you repeatedly suggested to move dependency detection to Find modules. Now it turns out Find modules are not good? What do you suggest doing?

> But I'm concerned about modules being moved from consistent to inconsistent.

In what way they were consistent before? Also how to maintain consistency with an inconsistent state when adding new Find modules? There have been 6 of them added anew and this question pops up for each of them.

We can endlessly discuss in generic terms, but it's not really helping. What would help is actual suggestion for unambiguous, consistent naming. So far nothing tangible came up.

> Being inconsistent is a burden when tracing problems.

I agree, but then what _is_ consistent in your opinion?

> Basically, this is a way to customize for static linkage, configuration details or strange platforms, without patching. (Is it called dependency injection?)
> 
> The strongest argument about case-preserved consistency is `<Pkg>_FOUND` - in particular with CMake's capability to search both modules and config. (And the user can choose to switch the order.)
> 
> In CMake, some mixed-case Find modules were always case-preserved. Others were adjusted and documented with case-preserved variables. Some case-preserved mixed-case examples: https://cmake.org/cmake/help/latest/module/FindIntl.html https://cmake.org/cmake/help/latest/module/FindIconv.html https://cmake.org/cmake/help/latest/module/FindPython.html

What do you mean by "case-preserved consistency"?

FWIW the newly added Find modules all adhere to the above examples: FindLibgsasl, FindLibidn2, FindLibssh, FindLibuv, FindRustls. The renames in this PR aim to do this for a couple of pre-existing ones, for consistency, with the best (mixed case) convention I could figure out (as written in the PR message).

dg0yt:
I tried to make it clear that I'm concerned about more cases of *inconsistency between mixed-case package names and upper-case variable names*. 
`find_package_handle_standard_args` is documented to set both `<PackageName>_FOUND` and `<PACKAGENAME>_FOUND` "for compatibility" (in the fine print of the deprecated `FOUND_VAR` option). But the primary documented pattern around packages is `<PackageName>_VARIABLE` aka "case-preserved".

Counting 13 questions in single post, I'm a little bit concerned that this is going to explode. Should I focus on selected ones?

vszakats:
Feel free to choose any that helps figuring out a better path than the current one.

Perhaps this one could help the most, from one post earlier:
Can you elaborate what alternate/superior action are you suggesting?

---

FWIW, curl doesn't use the `FOUND_VAR` option anymore. I interperetd the [manual](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html) as saying that `FOUND_VAR` is deprecated (hence dropped now), and regardless of being set or not, CMake guarantees to return both `<PackageName>_FOUND` and `<PACKAGENAME>_FOUND` for compatibility, quote: "Result variables of both names are always set for compatibility."

curl's CMake code always preferred the uppercase `FOUND` flavour (with 2 exceptions: Zstd, WolfSSL). After recent changes it uses uppercase `PACKAGENAME_FOUND` exclusively. Next release is about to add "native" pkg-config support (following your suggestion in another thread) for libgsasl, libidn2, libssh, libuv, nettle. Meaning pkg-config results are use directly instead of just `HINTS`. These modules use `pkg_check_modules(PACKAGENAME ...)` first, and only set uppercase `PACKAGENAME_FOUND` if found via `pkg-config`. Meaning it's best to rely on uppercase. All result variables are documented within each find module and now synced between each other:

```cmake
# Result variables:
#
# PACKAGENAME_FOUND         System has <packagename>
# PACKAGENAME_INCLUDE_DIRS  The <packagename> include directories
# PACKAGENAME_LIBRARIES     The <packagename> library names (in Find modules with pkg-config support, this is a pathless name, for others its an absolute path)
# PACKAGENAME_LIBRARY_DIRS  The <packagename> library directories (in Find modules with native pkg-config support)
# PACKAGENAME_CFLAGS        Required compiler flags (in Find modules with native pkg-config support)
# PACKAGENAME_VERSION       Version of <packagename>
```

(I experimented with calling `find_package_handle_standard_args` after a native pkg-config hit, but [turned it down](https://github.com/curl/curl/pull/14555/commits/a2536908b03d888a765df4988e7b7c7d2fcf3fc7). Besides some downside I noted that the only difference it made is defining `PackageName_FOUND`, which I deemed unnecessary, since nothing uses that in curl.)

PR #14603, opened 2024-08-19
Ccache tidy

icing:
Collected work on improving the connection cache, its bundles and other things around it.

Will open a new, squashed PR with detailed description when done.


icing:
Replaced by #14662.

PR #14606, opened 2024-08-19
Fix KEEPALIVE_FACTOR being set to 1000 for windows builds using setsockopt

### LINKED ISSUE #14368: option CURLOPT_TCP_KEEPALIVE doesn't work with new version
### I did this

I use curl to request to a php web api.
The PHP server will take a long time to answer the request ,like 600 seconds.So,I need to keep alive to wait the reply.
Old version 7.65.3 works well,but new version failed .I tried new two version 8.9.0,8.9.1,complied in win 10 with vs2019.
main code like this:

`//http is the same,this url sleep ,eg,300s ,then reply
curl_easy_setopt(curl, CURLOPT_URL, "https://www.some.com");
curl_easy_setopt(curl, CURLOPT_TCP_KEEPALIVE, 1L);
// 
curl_easy_setopt(curl, CURLOPT_TCP_KEEPIDLE,60L);
//
curl_easy_setopt(curl, CURLOPT_TCP_KEEPINTVL, 30L);	
//
curl_easy_setopt(curl, CURLOPT_TIMEOUT, 600L);//600s timeout

CURLcode res = curl_easy_perform(curl);	
if(res == CURLE_OK) {
	//7.65.3 will come here
} else {
	//timeout
	//new version will come here 8.9.0,8.9.1
}`

### I expected the following

keep alive

### curl/libcurl version

version 7.65.3 works
version 8.9.0 or 8.9.1 does'nt work

### operating system

win10,with openssl 3.3.1

samjessup:
Fixes #14368

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -147,7 +147,9 @@ static void nosigpipe(struct Curl_easy *data,
 #endif
 
 #if defined(USE_WINSOCK) && \
-	defined(TCP_KEEPIDLE)&& defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
+    defined(TCP_KEEPIDLE) && defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
+/*  Win 10, v 1709 (10.0.16299) and later can use SetSockOpt TCP_KEEP____
+ *  so should use seconds */
 #define CURL_WINSOCK_KEEP_SSO
 #define KEEPALIVE_FACTOR(x)
 #elif defined(USE_WINSOCK) || \


lib/cf-socket.c
```diff
@@ -146,7 +146,10 @@ static void nosigpipe(struct Curl_easy *data,
 #define nosigpipe(x,y) Curl_nop_stmt
 #endif
 
-#if defined(USE_WINSOCK) || \
+#if defined(USE_WINSOCK) && defined(TCP_KEEPIDLE) && defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
+#define WINSOCK_SUPPORTS_KEEP_IDLE_INTVL_CNT
+#endif
+#if (defined(USE_WINSOCK) && !defined(WINSOCK_SUPPORTS_KEEP_IDLE_INTVL_CNT)) || \
```
bagder: How about instead doing it with an `#elif`? I think it makes it a little clearer. Like:

~~~c
#if defined(USE_WINSOCK) && defined(TCP_KEEPIDLE) && defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
#define WINSOCK_SUPPORTS_KEEP_IDLE_INTVL_CNT
#elif defined(USE_WINSOCK) || \
...

lib/cf-socket.c
```diff
@@ -146,7 +146,12 @@ static void nosigpipe(struct Curl_easy *data,
 #define nosigpipe(x,y) Curl_nop_stmt
 #endif
 
-#if defined(USE_WINSOCK) || \
+#if defined(USE_WINSOCK)
+#if defined(TCP_KEEPIDLE)&& defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
```
vszakats, line 150:
```suggestion
#if defined(TCP_KEEPIDLE) && defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
```
nit

lib/cf-socket.c
```diff
@@ -146,7 +146,12 @@ static void nosigpipe(struct Curl_easy *data,
...
+#define WINSOCK_KEEP_SSO
```
vszakats, line 151:
```suggestion
#define CURL_WINSOCK_KEEP_SSO
```
Suggestion to use something namespaced to curl to avoid collisions.

lib/cf-socket.c
```diff
@@ -183,7 +188,7 @@ tcpkeepalive(struct Curl_easy *data,
   else {
 #if defined(SIO_KEEPALIVE_VALS) /* Windows */
 /* Windows 10, version 1709 (10.0.16299) and later versions */
-#if defined(TCP_KEEPIDLE) && defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
+#if defined(WINSOCK_KEEP_SSO)
```
vszakats, line 191:
```suggestion
#if defined(CURL_WINSOCK_KEEP_SSO)
```

bagder (approved):

lib/cf-socket.c
samjessup: Yes elif looks better.

bagder:
Thanks!

PR #5372, opened 2020-05-11
Fix build with pre-release libssh in Ubuntu 18.04, add libssh to CMake

Lekensteyn:
Ubuntu 18.04 is the only known distro version that ships with a pre-release version of libssh. Since they are unlikely to upgrade their libssh version, add a workaround to prevent build failures due to deprecation errors. This change is a dependency for #5370.

CMake support was added so I could easily test the libssh build.

cc @ansasaki

PR #14612, opened 2024-08-20
ssh: deduplicate SSH backend includes (and fix libssh cmake unity build)

vszakats:
For libssh, it fixes a "unity" build issue where libssh deprecation
warnings were not suppressed before this patch, because the suppression
macro was only set before just one of the two `libssh.h` includes.
If the other was compiled first in unity mode, the warnings appeared.

Seen in local curl-for-win build (`CW_CONFIG=test-x64-libssh-quictls`)
with libssh 0.11.0. (Also in a GHA/macos cmake job upcoming in #14614)

Use this opportunity to drop duplicate SSH header includes from the SSH
modules. It's enough to include them via the common `ssh.h` header.

### DIFF
--- a/lib/vssh/libssh.c
+++ b/lib/vssh/libssh.c
@@ -31,11 +31,6 @@
 
 #include <limits.h>
 
-/* in 0.10.0 or later, ignore deprecated warnings */
-#define SSH_SUPPRESS_DEPRECATED
-#include <libssh/libssh.h>
-#include <libssh/sftp.h>
-
 #ifdef HAVE_NETINET_IN_H
 #include <netinet/in.h>
 #endif
--- a/lib/vssh/libssh2.c
+++ b/lib/vssh/libssh2.c
@@ -30,9 +30,6 @@
 
 #include <limits.h>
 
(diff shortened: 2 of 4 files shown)


vszakats:
Original snippet via a158a0916691164b40b29e301d7cf589882598c0 #5372
```cmake
# libssh
option(CMAKE_USE_LIBSSH "Use libSSH" OFF)
mark_as_advanced(CMAKE_USE_LIBSSH)
if(NOT HAVE_LIBSSH2 AND CMAKE_USE_LIBSSH)
  find_package(libssh CONFIG)
  if(libssh_FOUND)
    message(STATUS "Found libssh ${libssh_VERSION}")
    # Use imported target for include and library paths.
    list(APPEND CURL_LIBS ssh)
    set(USE_LIBSSH ON)
    set(HAVE_LIBSSH_LIBSSH_H 1)
  endif()
endif()
```

vszakats:
Came up in CI after adding a libssh cmake macos job:
https://github.com/curl/curl/actions/runs/10473657245/job/29006036486#step:9:1

PR #14613, opened 2024-08-20
tidy-up: spelling 'built-in'

vszakats:


vszakats:
Merged without the `docs/SSLCERTS` change, in favour #14616.

PR #14614, opened 2024-08-20
cmake: try dropping libssh hack 1

vszakats:
Trying to fix a few issues in one:
- curl-for-win still requires a bad hack to configure libssh.
- libssh uses a hacky detection method ` find_package(libssh CONFIG QUIET)`
  no other dependency uses.
- after #14555, GHA/windows gnutls vcpkg job no longer
  auto-detects libssh:
  https://github.com/curl/curl/actions/runs/10470138955/job/28994650338


vszakats:
Original detection snippet via a158a0916691164b40b29e301d7cf589882598c0 #5372
```cmake
  find_package(libssh CONFIG)
  if(libssh_FOUND)
    message(STATUS "Found libssh ${libssh_VERSION}")
    # Use imported target for include and library paths.
    list(APPEND CURL_LIBS ssh)
    set(USE_LIBSSH ON)
    set(HAVE_LIBSSH_LIBSSH_H 1)
  endif()
```

PR #14624, opened 2024-08-21
unit1300: fix checksrc longline warnings

jay:
Closes #xxxx

---

I saw the warnings when I ran checksrc.bat

PR #14625, opened 2024-08-21
GHA: add a checksrc job

bagder:
This job unconditionally runs checksrc on ALL .c and .h files present in git.

checksrc.pl: fixed to look for ".checksrc" in the same directory from where it loads the file to check

### DIFF
--- a/projects/wolfssl_options.h
+++ b/projects/wolfssl_options.h
@@ -75,7 +75,7 @@ create the thread local storage and that could be a problem for LoadLibrary.
 Regarding the options that were added via C_EXTRA_FLAGS:
 
 FP_MAX_BITS=16384
-https://www.yassl.com/forums/topic423-cacertorgs-ca-cert-verify-failed-but-withdisablefastmath-it-works.html
+https://www.wolfssl.com/documentation/manuals/wolfssl/chapter02.html
 "Since root.crt uses a 4096-bit RSA key, you'll need to increase the fastmath
 buffer size.  You can do this using the define:
 FP_MAX_BITS and setting it to 8192."
--- a/.github/workflows/checksrc.yml
+++ b/.github/workflows/checksrc.yml
@@ -1,0 +1,29 @@
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+
+name: checksrc
+
+on:
+  # Trigger the workflow on push or pull requests, but only for the
+  # master branch
+  push:
+    branches:
+      - master
+      - '*/ci'
+  pull_request:
+    branches:
+      - master
+
+permissions: {}
+
+jobs:
+  all:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: check
+        run: git ls-files "*.[ch]" | xargs -n1 ./scripts/checksrc.pl
--- a/packages/vms/curl_crtl_init.c
+++ b/packages/vms/curl_crtl_init.c
@@ -89,89 +89,87 @@ int   decc$feature_set_value (int index, int mode, int value);
 #endif
 
 int   SYS$TRNLNM(
-    const unsigned long * attr,
-    const struct dsc$descriptor_s * table_dsc,
-    struct dsc$descriptor_s * name_dsc,
-    const unsigned char * acmode,
-    const struct itmlst_3 * item_list);
+    const unsigned long *attr,
+    const struct dsc$descriptor_s *table_dsc,
+    struct dsc$descriptor_s *name_dsc,
+    const unsigned char *acmode,
+    const struct itmlst_3 *item_list);
 int   SYS$CRELNM(
-    const unsigned long * attr,
-    const struct dsc$descriptor_s * table_dsc,
-    const struct dsc$descriptor_s * name_dsc,
-    const unsigned char * acmode,
-    const struct itmlst_3 * item_list);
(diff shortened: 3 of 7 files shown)


bagder:
Because of #14624

bagder:
It looks like old-mingw is somehow not understanding the CURL_TEMP_PRINTF() macro in curl/mprintf.h correctly.

bagder:
@vszakats what does "old-mingw" mean in this CI context? It looks like I might need to add some additional conditions.

projects/wolfssl_options.h
```diff
@@ -75,7 +75,8 @@ create the thread local storage and that could be a problem for LoadLibrary.
 Regarding the options that were added via C_EXTRA_FLAGS:
 
 FP_MAX_BITS=16384
-https://www.yassl.com/forums/topic423-cacertorgs-ca-cert-verify-failed-but-withdisablefastmath-it-works.html
+https://www.yassl.com/forums/topic423-cacertorgs-ca-cert-verify-failed-but-\
+ withdisablefastmath-it-works.html
```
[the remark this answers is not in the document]
bagder: I don't think that is a good idea in general. We have managed without it so far. In this case, there is a better and shorter URL we can use instead.

tests/http/clients/h2-download.c
```diff
@@ -84,11 +84,12 @@ static int debug_cb(CURL *handle, curl_infotype type,
...
     if(!curl_easy_getinfo(handle, CURLINFO_CONN_ID, &conn_id) &&
-        conn_id >= 0) {
-      snprintf(idsbuf, sizeof(idsbuf), TRC_IDS_FORMAT_IDS_2, xfer_id, conn_id);
+       conn_id >= 0) {
+      curl_msnprintf(idsbuf, sizeof(idsbuf), TRC_IDS_FORMAT_IDS_2, xfer_id,
+                     conn_id);
     }
     else {
-      snprintf(idsbuf, sizeof(idsbuf), TRC_IDS_FORMAT_IDS_1, xfer_id);
+      curl_msnprintf(idsbuf, sizeof(idsbuf), TRC_IDS_FORMAT_IDS_1, xfer_id);
```
vszakats, line 92: Turned these into `snprintfs` in 232302f88a152a1d1722da9f69c383a766528918 [#14382](https://github.com/curl/curl/pull/14382/commits) after a series of whack-a-mole with format failure. Couldn't figure out a better way. (From a build perspective these resemble 'examples', where `snprintf()` is also used as-is.)

vszakats:
> @vszakats what does "old-mingw" mean in this CI context? It looks like I might need to add some additional conditions.

It's a shortcut for `old-mingw-w64`, meaning old, standalone builds of mingw-w64. (not the original "old-mingw")

tests/http/clients/h2-download.c
bagder: If we can detect when to.

bagder:
> It's a shortcut for old-mingw-w64, meaning old, standalone builds of mingw-w64. (not the original "old-mingw")

I don't know how that differs from the non-old versions! Is there a particular cut-off mingw version I can check for or similar?

tests/http/clients/h2-download.c
bagder: Since this warning only happens with old mingw compilers, it feels like it could just as well be a quirk in those old versions and chasing that down sounds like work with very little benefit. We can just as well disable the printf format attributes for those compilers.

bagder: ... except that I seem to fail at that too.

bagder:
It looks like the code in tests/http/clients are built with some other curl headers and not the ones from git, as no changes done in this PR seem to "take".

Only for the cmake builds.

vszakats:
I'll look into this, thanks!

dfandrich:
Analysis of PR #14625:

[Test 291 failed](https://circleci.com/gh/curl/curl/104255), which has NOT been flaky recently, so **there could be a real issue in the PR**.

include/curl/mprintf.h
```diff
@@ -33,9 +33,10 @@ extern "C" {
 #endif
 
 #if (defined(__GNUC__) || defined(__clang__) ||                         \
-  defined(__IAR_SYSTEMS_ICC__)) &&                                      \
+     defined(__IAR_SYSTEMS_ICC__)) &&                                   \
   defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) &&         \
-  !defined(CURL_NO_FMT_CHECKS)
+  !defined(CURL_NO_FMT_CHECKS) &&                                       \
+  (!defined(__MINGW32_MAJOR_VERSION_) || (__MINGW32_MAJOR_VERSION >= 11))
```
vszakats, line 39: Opened #14640 with a fix proposal.

tests/http/clients/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ foreach(_test_name IN LISTS check_PROGRAMS)
   add_executable(${_test_target} EXCLUDE_FROM_ALL "${_test_name}.c")
   add_dependencies(testdeps ${_test_target})
   target_include_directories(${_test_target} PRIVATE
+    "${CURL_SOURCE_DIR}/include"  # for "curl/curl.h" etc
```
vszakats, line 34: The include directory is added globally here:
https://github.com/curl/curl/blob/9724526429bb5c41fc7b923c616e4c466faf5f7a/CMakeLists.txt#L113

bagder: that was clearly just a desperate attempt, will clean up once the format PR has landed.

PR #14628, opened 2024-08-21
configure: fix indentation

vszakats:
Also:
- move `then`s and an `in` inline.
- whitespace.

---

w/o whitespace: https://github.com/curl/curl/pull/14628/files?w=1


vszakats:
Linux rustls pytest stuck in a loop: https://github.com/curl/curl/actions/runs/10497365140/job/29080063715?pr=14628#step:40:704

PR #14629, opened 2024-08-21
setopt: allow CURLOPT_INTERFACE to be set to NULL

bagder:
Ref: https://github.com/curl/curl/discussions/14299#discussioncomment-10393909 Regression from 3060557af702dd591 (shipped in 8.9.0)


uninsane:
rebuilt `curlftpfs` with this patch applied to curl, and confirmed that this fixes the linked issue.

PR #14630, opened 2024-08-21
cmake: simplify version detection regex

vszakats:
Fixes syntax highlighting in some editors.
Idea taken from CMakeLists.txt's curl version detection.


vszakats:
This doesn't look necessary. If there is ever a case where there is a false positive due to for example catching a version macro that is commented C++ style, this might have a usefulness, but until then, closing without merge.

PR #14631, opened 2024-08-21
lib: delete stray undefs for `vsnprintf`, `vsprintf`

vszakats:
Stop #undefining macros that were not redefined or used in the code.

### DIFF
--- a/lib/curlx.h
+++ b/lib/curlx.h
@@ -98,7 +98,6 @@
 # undef msnprintf
 # undef vprintf
 # undef vfprintf
-# undef vsprintf
 # undef mvsnprintf
 # undef aprintf
 # undef vaprintf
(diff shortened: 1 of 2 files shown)


lib/curlx.h
```diff
@@ -112,6 +112,7 @@
 # define mvsnprintf curlx_mvsnprintf
 # define aprintf curlx_maprintf
 # define vaprintf curlx_mvaprintf
+# define vsprintf curlx_mvsprintf
```
bagder, line 115: No need to add this define. This function should not be used in libcurl code.

vszakats: OK, then `# undef vsprintf` can be safely deleted few lines above, isn't it?

bagder: Yes, it seems to serve no purpose.

PR #14626, opened 2024-08-21
cmake: honor custom `CMAKE_UNITY_BUILD_BATCH_SIZE`

vszakats:
This value tells how many sources files to bundle in a single "unity"
compilation unit.

The CMake default is 8 sources, curl's CMake set this to 0, meaning
to bundle all sources into a single unit.

This patch makes it possible to override the 0 value, and potentially
optimize the build process further by better utilizing multiple cores
in conjunction with `make -jN`.

The number of sources in lib is 172 at the time of writing this. For
a 12-core CPU, this can give a job for them all:
`-DCMAKE_UNITY_BUILD_BATCH_SIZE=15`

(Compile time may be affected by a bunch of other factors.)

PR #14632, opened 2024-08-21
src: fix potential macro confusion in cmake unity builds

vszakats:
Sources used `lib/curlx.h` with both `ENABLE_CURLX_PRINTF` set and unset
before including it.

In a cmake "unity" batch where the first included source had it unset,
the next sources did not get the macros requested with
`ENABLE_CURLX_PRINTF` because `lib/curl.x` had already been included
without them.

Fix it by by making the macros enabled permanently and globally for
internal sources, and dropping `ENABLE_CURLX_PRINTF`.

This came up while testing unity builds with smaller batches. The full,
default unity build where all `src` is bundled up in a single unit, was
not affected.

Fixes:
```
$ cmake -B build -DCMAKE_UNITY_BUILD=ON -DCMAKE_UNITY_BUILD_BATCH_SIZE=15
$ make -C build
...
curl/src/tool_getparam.c: In function ‘getparameter’:
curl/src/tool_getparam.c:2409:11: error: implicit declaration of function ‘msnprintf’; did you mean ‘vsnprintf’? [-Wimplicit-function-declaration]
 2409 |           msnprintf(buffer, sizeof(buffer), "%" CURL_FORMAT_CURL_OFF_T "-",
      |           ^~~~~~~~~
      |           vsnprintf
curl/src/tool_getparam.c:2409:11: warning: nested extern declaration of ‘msnprintf’ [-Wnested-externs]
[...]
```

Reported-by: Daniel Stenberg
Bug: https://github.com/curl/curl/pull/14626#issuecomment-2301663491

### DIFF
--- a/lib/curlx.h
+++ b/lib/curlx.h
@@ -88,10 +88,9 @@
 #define curlx_mvprintf curl_mvprintf
 #define curlx_mvfprintf curl_mvfprintf
 
-#ifdef ENABLE_CURLX_PRINTF
-/* If this define is set, we define all "standard" printf() functions to use
-   the curlx_* version instead. It makes the source code transparent and
-   easier to understand/patch. Undefine them first. */
+/* We define all "standard" printf() functions to use the curlx_* version
+   instead. It makes the source code transparent and easier to
+   understand/patch. Undefine them first. */
 # undef printf
 # undef fprintf
 # undef sprintf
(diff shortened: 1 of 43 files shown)


vszakats:
Is there a reason for any curl code to ever need `curlx.h` _without_ remapping these standard functions to curl ones?

Because if there isn't, we might just delete the `ENABLE_CURLX_PRINTF` completely and have these enabled always.

vszakats:
@bagder It builds with the `ENABLE_CURLX_PRINTF` macro deleted completely and the mapping enabled permanently. Shall we go with it?

bagder:
> @bagder It builds with the `ENABLE_CURLX_PRINTF` macro deleted completely and the mapping enabled permanently. Shall we go with it?

Yes, let's do that!

PR #14634, opened 2024-08-21
test1521: verify setting options better

bagder:
Previously this test allowed several error values when setting options. This made this test miss #14629.
    
Now, errors are generally not accepted for setopts:
    
 - numerical setopts accept CURLE_BAD_FUNCTION_ARGUMENT for funny input
 - the first setopt to an option accepts CURLE_NOT_BUILT_IN or CURLE_UNKNOWN_OPTION for when they are disabled/not built-in
 - there is an allowlist concept for some return code for some variables, managed at the top of the mk-lib1521.pl script

In curl.h: remove the OBSOLETE named values from the setopt list.

### DIFF
--- a/include/curl/curl.h
+++ b/include/curl/curl.h
@@ -1250,8 +1250,7 @@ typedef enum {
   /* send linked-list of post-transfer QUOTE commands */
   CURLOPT(CURLOPT_POSTQUOTE, CURLOPTTYPE_SLISTPOINT, 39),
 
-   /* OBSOLETE, do not use! */
-  CURLOPT(CURLOPT_OBSOLETE40, CURLOPTTYPE_OBJECTPOINT, 40),
+  /* 40 is not used */
 
   /* talk a lot */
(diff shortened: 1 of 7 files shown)


include/curl/curl.h
```diff
@@ -1352,9 +1351,7 @@ typedef enum {
   /* Max amount of cached alive connections */
   CURLOPT(CURLOPT_MAXCONNECTS, CURLOPTTYPE_LONG, 71),
 
-  /* OBSOLETE, do not use! */
-  CURLOPT(CURLOPT_OBSOLETE72, CURLOPTTYPE_LONG, 72),
-
+  /* 72 = OBSOLETE */
```
dg0yt, line 1354: Noticed in vpckg CI run on CURL HEAD:

`CURLOPT_OBSOLETE72` is still used in the header for

https://github.com/curl/curl/blob/3fc81be44e970215f6ca18ab331ac2050abda58b/include/curl/curl.h#L723

and `CURLOPT_CLOSEPOLICY` is used downstream in curlpp and curlcpp.

PR #14635, opened 2024-08-21
docs: Clarify OpenSSF Best Practices vs Scorecard

cpswan:
#14319 introduced a section to `SECURITY.md` titled **OpenSSF Scorecard** that actually documents OpenSSF Best Practices ([Scorecard](https://scorecard.dev/) is a different OpenSSF project, that incorporates Best Practices, but is distinct in its objectives and how it achieves them).

This change clarifies the terminology, and also removes any implication that Gold Best Practices is an award rather than a self certification programme.

As curl was a leader in implementing Best Practices some folk may be more familiar with the earlier Core Infrastructure Initiative (CII) naming, so a reference to that has been added.

I'll start a discussion on whether curl should be running the Scorecard workflow to generate a score there, which would perhaps go on to justify another section in `SECURITY.md` for that.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -19,7 +19,7 @@ We treat security issues with confidentiality until controlled and disclosed res
 
 curl has achieved Gold status on the Open Source Security Foundation (OpenSSF)
 [Best Practices](https://bestpractices.dev/) (formerly Core Infrastructure
-Initiative [CII] Best Practices), reflecting its adherence to rigorous
+Initiative Best Practices), reflecting its adherence to rigorous
 security and best practice standards. This achievement highlights curl's
 comprehensive documentation, secure development processes, effective change
 control mechanisms, and strong maintenance routines. Meeting these criteria
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -15,13 +15,15 @@ libcurl, report it on [HackerOne](https://hackerone.com/curl).
 
 We treat security issues with confidentiality until controlled and disclosed responsibly.
 
-## OpenSSF Scorecard
+## OpenSSF Best Practices
 
-curl has earned Gold status on the OpenSSF Best Practices, reflecting its adherence to
-rigorous security and best practice standards. This achievement highlights curl's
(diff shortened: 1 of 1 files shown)


vpetersson:
You beat me to it, @cpswan :)

bagder (approved):

bagder:
We just need to allow "CII" as a word for the spell-checker, or consider removing it from the text.

SECURITY.md
```diff
@@ -15,13 +15,15 @@ libcurl, report it on [HackerOne](https://hackerone.com/curl).
...
-mechanisms, and strong maintenance routines. Meeting these criteria demonstrates curl's
-commitment to security and reliability, ensuring the project's sustainability and
-trustworthiness. This recognition by OpenSSF underscores curl's role as a leader in
-open-source software practices. More information can be found on
-their [OpenSSF page](https://www.bestpractices.dev/projects/63).
+## OpenSSF Best Practices
+
+curl has achieved Gold status on the Open Source Security Foundation (OpenSSF)
+[Best Practices](https://bestpractices.dev/) (formerly Core Infrastructure
+Initiative [CII] Best Practices), reflecting its adherence to rigorous
```
bagder, line 22:
```suggestion
Initiative Best Practices), reflecting its adherence to rigorous
```

Simplifies it a little and saves us from having to whitelist `CII` as a word for the spellchecker.

cpswan: The only reason I put CII in there was because (weirdly) OpenSSF Scorecard still refers to 'CII Best Practices' rather than 'OpenSSF Best Practices', so the CII term is still undead.

I'm cool with the proposed change though, and I totally understand why putting CII into the spell check might be a bad idea because it's an easy typo for CI.

bagder: That's exactly my reasoning - and I really don't want to introduce a special "allow-list" for this document alone. A small sacrifice but not too bad I think.

bagder:
Thanks!

PR #14636, opened 2024-08-21
cmake: tidy up option descriptions

vszakats:
- capitalize.
- impartive voice.
- stick to `Build ...`, `Use ...`, `Enable ...`, `Disable ...` with
  a couple of exceptions.
- drop period.
- fix wolfSSH casing.


vszakats:
Final patch included a websockets → WebSockets casing fix.

PR #14637, opened 2024-08-21
CURLOPT_XFERINFOFUNCTION: clarify the callback return codes

### LINKED ISSUE #14627: XFERINFOFUNCTION: Description of return values is confusing
### Specify which documentation you found a problem with

The documentation for [CURLOPT_XFERINFOFUNCTION](https://curl.se/libcurl/c/CURLOPT_XFERINFOFUNCTION.html) and [CURLOPT_PROGRESSFUNCTION](https://curl.se/libcurl/c/CURLOPT_PROGRESSFUNCTION.html) says:

> If your callback function returns CURL_PROGRESSFUNC_CONTINUE it makes libcurl to continue executing the default progress function.
>
> Returning any other non-zero value from this callback makes libcurl abort the transfer and return [CURLE_ABORTED_BY_CALLBACK](https://curl.se/libcurl/c/libcurl-errors.html#CURLEABORTEDBYCALLBACK).

### The problem

It is only implied, but not made explicit, that the usual return value is 0.  It is also not clear which actual values are valid to cause an abort, because the value for CURL_PROGRESSFUNC_CONTINUE is not documented.  I currently use `min(-42, CURL_PROGRESSFUNC_CONTINUE - 42);` but that seems rather clumsy.

I suggest:

> If your callback function returns CURL_PROGRESSFUNC_CONTINUE (&lt;actual value here>) it makes libcurl execute the default progress function and than continue the transfer.  If it returns CURL_PROGRESSFUNC_OK (0) the default progress function is not executed and the transfer is continued.  If it returns CURL_PROGRESSFUNC_ABORT (-1) or any other value the default progress function is not executed, the transfer is aborted and transfer function will return [CURLE_ABORTED_BY_CALLBACK](https://curl.se/libcurl/c/libcurl-errors.html#CURLEABORTEDBYCALLBACK).

This assumes of course that the constants CURL_PROGRESSFUNC_OK and CURL_PROGRESSFUNC_ABORT are actually defined in the headers.

bagder:
also done in CURLOPT_PROGRESSFUNCTION.md

Fixes #14627
Reported-by: Benjamin Riefenstahl Mecom

### DIFF
--- a/docs/libcurl/opts/CURLOPT_PROGRESSFUNCTION.md
+++ b/docs/libcurl/opts/CURLOPT_PROGRESSFUNCTION.md
@@ -60,11 +60,13 @@ if you only download data, the upload size remains 0). Many times the callback
 is called one or more times first, before it knows the data sizes so a program
 must be made to handle that.
 
+Return zero from the callback if everything is fine.
+
 If your callback function returns CURL_PROGRESSFUNC_CONTINUE it causes libcurl
 to continue executing the default progress function.
 
-Returning any other non-zero value from this callback makes libcurl abort the
-transfer and return *CURLE_ABORTED_BY_CALLBACK*.
+Return 1 from this callback to make libcurl abort the transfer and return
+*CURLE_ABORTED_BY_CALLBACK*.
 
 If you transfer data with the multi interface, this function is not called
 during periods of idleness unless you call the appropriate libcurl function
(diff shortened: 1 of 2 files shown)


dfandrich:
Analysis of PR #14637:

[Test 1062 failed](https://circleci.com/gh/curl/curl/104384), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14640, opened 2024-08-21
build: make `CURL_FORMAT_CURL_OFF_T[U]` work with mingw-w64 <=7.0.0

vszakats:
Add tweak for mingw-w64 when building tests/http/client programs to
avoid a bogus `-Wformat` warning when using mingw-w64 v7.0.0 or older.
The warning is bogus because these programs use curl's `printf()`
implementation that is guaranteed to support that format spec.

Add this for both CMake and autotools. (But only CMake is CI tested with
an old toolchain.)

Apply the workaround to `docs/examples`, and fix an example to use
curl's `printf()` with `CURL_FORMAT_CURL_OFF_T`.

Reintroduce curl `printf()` calls into `tests/http/client`, via #14625.
Also restore large number masks to a printf, changed earlier in #14382.

Follow-up to 232302f88a152a1d1722da9f69c383a766528918 #14382
Ref: https://github.com/curl/curl/pull/14625#issuecomment-2302361737

---

- [x] add the same for autotools.
- [ ] fix it in `curl/system.h` instead?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -607,6 +607,19 @@ case $host_os in
     ;;
 esac
 
+have_mingw='no'
+case $host_os in
+  mingw*)
+    have_mingw='yes'
+    ;;
+esac
+
+AM_CONDITIONAL([HAVE_MINGW],
+  [test "$have_mingw" = "yes"])
+
+AM_CONDITIONAL([HAVE_WINDOWS],
+  [test "$curl_cv_native_windows" = "yes"])
+
 AM_CONDITIONAL([HAVE_WINDRES],
   [test "$curl_cv_native_windows" = "yes" && test -n "${RC}"])
 
--- a/configure.ac
+++ b/configure.ac
@@ -615,7 +615,7 @@ case $host_os in
 esac
 
 AM_CONDITIONAL([HAVE_MINGW],
-  [test "$curl_cv_native_windows" = "yes" -a "$have_mingw" = "yes"])
+  [test "$curl_cv_native_windows" = 'yes' -a "$have_mingw" = 'yes'])
 
 AM_CONDITIONAL([HAVE_WINDRES],
   [test "$curl_cv_native_windows" = "yes" && test -n "${RC}"])
--- a/configure.ac
+++ b/configure.ac
@@ -615,10 +615,7 @@ case $host_os in
 esac
 
 AM_CONDITIONAL([HAVE_MINGW],
-  [test "$have_mingw" = "yes"])
-
-AM_CONDITIONAL([HAVE_WINDOWS],
-  [test "$curl_cv_native_windows" = "yes"])
+  [test "$curl_cv_native_windows" = "yes" -a "$have_mingw" = "yes"])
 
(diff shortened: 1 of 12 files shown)


vszakats:
Here's an untested alternative plan:
```diff
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -178,10 +178,16 @@
 #  define CURL_TYPEOF_CURL_SOCKLEN_T int
 
 #elif defined(__MINGW32__)
+#  include <_mingw.h>
 #  include <inttypes.h>
 #  define CURL_TYPEOF_CURL_OFF_T     long long
+#if defined(__MINGW64_VERSION_MAJOR) && __MINGW64_VERSION_MAJOR <= 7
+#  define CURL_FORMAT_CURL_OFF_T     "lld"
+#  define CURL_FORMAT_CURL_OFF_TU    "llu"
+#else
 #  define CURL_FORMAT_CURL_OFF_T     PRId64
 #  define CURL_FORMAT_CURL_OFF_TU    PRIu64
+#endif
 #  define CURL_SUFFIX_CURL_OFF_T     LL
 #  define CURL_SUFFIX_CURL_OFF_TU    ULL
 #  define CURL_TYPEOF_CURL_SOCKLEN_T int
--- a/lib/curl_setup.h
+++ b/lib/curl_setup.h
@@ -317,17 +317,6 @@
 #define CURL_CONC_MACROS_(A,B) A ## B
 #define CURL_CONC_MACROS(A,B) CURL_CONC_MACROS_(A,B)
 
-/* curl uses its own printf() function internally. It understands the GNU
- * format. Use this format, so that is matches the GNU format attribute we
- * use with the MinGW compiler, allowing it to verify them at compile-time.
- */
-#ifdef  __MINGW32__
-#  undef CURL_FORMAT_CURL_OFF_T
-#  undef CURL_FORMAT_CURL_OFF_TU
-#  define CURL_FORMAT_CURL_OFF_T   "lld"
-#  define CURL_FORMAT_CURL_OFF_TU  "llu"
-#endif
-
 /* based on logic in "curl/mprintf.h" */
 
 #if (defined(__GNUC__) || defined(__clang__) ||                         \
```

configure.ac
```diff
@@ -607,6 +607,16 @@ case $host_os in
...
 
+have_mingw='no'
+case $host_os in
+  mingw*)
+    have_mingw='yes'
+    ;;
+esac
+
+AM_CONDITIONAL([HAVE_MINGW],
+  [test "$curl_cv_native_windows" = 'yes' -a "$have_mingw" = 'yes'])
```
bagder, line 618: This sets the variable for newer mingw versions as well, right? I mean you want to set a compiler flag for 7.0.0 and earlier but it does set it for all versions, doesn't it?

vszakats: Yes, but, AFAICT this isn't causing an issue, because newer versions set this variable to 1 by default anyway.

bagder: Understood. I just wanted to make sure it was done like that on purpose. Thanks!

bagder:
Sure!

PR #14643, opened 2024-08-22
printf: fix mingw-w64 format checks, turn C99 format strings into macro

vszakats:
Change mingw-w64 printf format checks to use `__MINGW_PRINTF_FORMAT`
instead of `gnu_printf`, to sync the format checker with the format
string macros published via `curl/system.h`. (Also disable format checks
for mingw-w64 older than 3.0.0 (2013-09-20) and classic-mingw, which
do not support this macro.)

This fixes bogus format checker `-Wformat` warnings in 3rd party code
using curl format strings with the curl printf functions, when using
mingw-w64 7.0.0 (2019-11-10) and older (with GCC, MSVCRT).   
                                                             
It also allows to delete two workaounds for this within curl itself:
- setting `-D__USE_MINGW_ANSI_STDIO=1` for mingw-w64 via cmake and
  configure for `docs/examples` and `tests/http/clients`.
  Ref: c730c8549b5b67e7668ca5d2cd82c3cc183e125d #14640
- overriding `curl/system.h` `CURL_FORMAT_CURL_OFF_T[U]` format strings
  to GNU-style for the curl source code.
  Ref: 3829759bd042c03225ae862062560f568ba1a231 #12489

But, `__MINGW_PRINTF_FORMAT` creates a new problem, where the C99 `%z`
(`size_t`) format string is no longer accepted by the format checker
when using mingw-w64 7.0.0 (2019-11-10) and earlier, which default to
`ms_printf` format string style.

Fix this by adding new, internal, `size_t` format string macros, and
replace `%z` uses across the codebase with them.

The macro adapts to mingw-w64 and other build environments
automatically.

After this patch the `curl/system.h` format string macros, the format
checker style set in `curl/mprintf.h` (and `lib/curl_setup.h`), and the
format strings used across the source code always match.

Also:
- `lib/ws.c`: add missing space to a string.
- fix to use standard printf in the example `ftpgetinfo`.

Verification script:
```bash
: check if format string translations were accurate:
curl -LO https://github.com/curl/curl/pull/14643.diff
grep '^-' 14643.diff > 14643-out.diff
grep '^+' 14643.diff > 14643-in.diff
grep -E -o '%[0-9]*z[xud]' 14643-out.diff > 14643-out-fmt.diff
grep -E -o '%[0-9]*" CURL_FORMAT_[XS]?SIZE_T' 14643-in.diff > 14643-in-fmt.diff
paste 14643-in-fmt.diff 14643-out-fmt.diff | sort | uniq -c > 14643-FORMATS.txt
: check if strings remained intact after reflow:
grep -E '^(\+|@)' 14643.diff | grep -E -o '("[^"]+"|^\+\+\+.+|^@@)' | tr '\n' '\t' | sed -E -e 's/\"\t\"//g' -e 's/\+\+\+ b/--- a/g'       | tr '\t' '\n' > 14643-in.diff
grep -E '^(-|@)'  14643.diff | grep -E -o '("[^"]+"|^---.+|^@@)'    | tr '\n' '\t' | sed -E -e 's/\"\t\"//g' -e 's/(%[0-9]*)z(x|u|d)/\1/g' | tr '\t' '\n' > 14643-out.diff
diff -u 14643-out.diff 14643-in.diff > 14643-STRINGS.diff
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -597,16 +597,6 @@ case $host_os in
     ;;
 esac
 
-have_mingw='no'
-case $host_os in
-  mingw*)
-    have_mingw='yes'
-    ;;
-esac
-
-AM_CONDITIONAL([HAVE_MINGW],
-  [test "$curl_cv_native_windows" = 'yes' -a "$have_mingw" = 'yes'])
-
 AM_CONDITIONAL([HAVE_WINDRES],
   [test "$curl_cv_native_windows" = "yes" && test -n "${RC}"])
 
--- a/docs/examples/CMakeLists.txt
+++ b/docs/examples/CMakeLists.txt
@@ -34,10 +34,6 @@ foreach(_target IN LISTS check_PROGRAMS)
   if(LIB_SELECTED STREQUAL LIB_STATIC AND WIN32)
     set_property(TARGET ${_target_name} APPEND PROPERTY COMPILE_DEFINITIONS "CURL_STATICLIB")
   endif()
-  if(MINGW)
-    # For mingw-w64 7.0.0 and earlier to avoid '-Wformat='
-    set_property(TARGET ${_target_name} APPEND PROPERTY COMPILE_DEFINITIONS "__USE_MINGW_ANSI_STDIO=1")
-  endif()
   set_target_properties(${_target_name} PROPERTIES
     OUTPUT_NAME "${_target}" UNITY_BUILD OFF
     PROJECT_LABEL "Example ${_target}")
--- a/docs/examples/Makefile.am
+++ b/docs/examples/Makefile.am
@@ -46,11 +46,6 @@ if USE_CPPFLAG_CURL_STATICLIB
 AM_CPPFLAGS += -DCURL_STATICLIB
 endif
 
-if HAVE_MINGW
-# For mingw-w64 7.0.0 and earlier to avoid '-Wformat='
-AM_CPPFLAGS += -D__USE_MINGW_ANSI_STDIO=1
-endif
-
 # Prevent LIBS from being used for all link targets
 LIBS = $(BLANK_AT_MAKETIME)
 
--- a/docs/examples/ftpgetinfo.c
+++ b/docs/examples/ftpgetinfo.c
@@ -75,8 +75,8 @@ int main(void)
       res = curl_easy_getinfo(curl, CURLINFO_CONTENT_LENGTH_DOWNLOAD_T,
                               &filesize);
       if((CURLE_OK == res) && (filesize>0))
-        curl_mprintf("filesize %s: %" CURL_FORMAT_CURL_OFF_T " bytes\n",
-                     filename, filesize);
+        printf("filesize %s: %" CURL_FORMAT_CURL_OFF_T " bytes\n",
+               filename, filesize);
     }
     else {
       /* we failed */
--- a/include/curl/mprintf.h
+++ b/include/curl/mprintf.h
@@ -37,7 +37,7 @@ extern "C" {
   defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) &&         \
   !defined(CURL_NO_FMT_CHECKS)
 #if defined(__MINGW32__)
-#if defined(__MINGW_PRINTF_FORMAT)  /* mingw-w64 4.0.0+. Needs stdio.h. */
+#if defined(__MINGW_PRINTF_FORMAT)  /* mingw-w64 3.0.0+. Needs stdio.h. */
 #define CURL_TEMP_PRINTF(fmt, arg) \
   __attribute__((__format__(__MINGW_PRINTF_FORMAT, fmt, arg)))
 #else
--- a/include/curl/mprintf.h
+++ b/include/curl/mprintf.h
@@ -39,7 +39,7 @@ extern "C" {
 #if defined(__MINGW32__) && !defined(__clang__)
 #if defined(__MINGW_PRINTF_FORMAT)  /* mingw-w64 3.0.0+. Needs stdio.h. */
 #define CURL_TEMP_PRINTF(fmt, arg) \
-  __attribute__((__format__(__MINGW_PRINTF_FORMAT, fmt, arg)))
+  __attribute__((format(__MINGW_PRINTF_FORMAT, fmt, arg)))
 #else
 #define CURL_TEMP_PRINTF(fmt, arg)
 #endif
--- a/include/curl/mprintf.h
+++ b/include/curl/mprintf.h
@@ -36,7 +36,7 @@ extern "C" {
   defined(__IAR_SYSTEMS_ICC__)) &&                                      \
   defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) &&         \
   !defined(CURL_NO_FMT_CHECKS)
-#if defined(__MINGW32__)
+#if defined(__MINGW32__) && !defined(__clang__)
 #if defined(__MINGW_PRINTF_FORMAT)  /* mingw-w64 3.0.0+. Needs stdio.h. */
 #define CURL_TEMP_PRINTF(fmt, arg) \
   __attribute__((__format__(__MINGW_PRINTF_FORMAT, fmt, arg)))
--- a/include/curl/mprintf.h
+++ b/include/curl/mprintf.h
@@ -36,9 +36,13 @@ extern "C" {
   defined(__IAR_SYSTEMS_ICC__)) &&                                      \
   defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) &&         \
   !defined(CURL_NO_FMT_CHECKS)
-#if defined(__MINGW32__) && !defined(__clang__)
+#if defined(__MINGW32__)
+#if defined(__MINGW_PRINTF_FORMAT)  /* mingw-w64 4.0.0+. Needs stdio.h. */
 #define CURL_TEMP_PRINTF(fmt, arg) \
-  __attribute__((format(gnu_printf, fmt, arg)))
+  __attribute__((__format__(__MINGW_PRINTF_FORMAT, fmt, arg)))
+#else
+#define CURL_TEMP_PRINTF(fmt, arg)
+#endif
 #else
 #define CURL_TEMP_PRINTF(fmt, arg) \
   __attribute__((format(printf, fmt, arg)))
--- a/include/curl/system.h
+++ b/include/curl/system.h
@@ -178,10 +178,16 @@
 #  define CURL_TYPEOF_CURL_SOCKLEN_T int
 
 #elif defined(__MINGW32__)
+#  include <_mingw.h>
 #  include <inttypes.h>
 #  define CURL_TYPEOF_CURL_OFF_T     long long
+#if defined(__MINGW64_VERSION_MAJOR) && __MINGW64_VERSION_MAJOR <= 7
+#  define CURL_FORMAT_CURL_OFF_T     "lld"
+#  define CURL_FORMAT_CURL_OFF_TU    "llu"
+#else
 #  define CURL_FORMAT_CURL_OFF_T     PRId64
 #  define CURL_FORMAT_CURL_OFF_TU    PRIu64
+#endif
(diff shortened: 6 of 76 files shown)


vszakats:
OK, so at first try this revealed (via new `-Wformat` warnings) all the code that's using
`CURL_FORMAT_CURL_OFF_T[U]` with CRT `printf()` functions.

/cc @bagder

bagder:
Clearly the format warning on those mingw compiler versions is not functional for how we do things.

vszakats:
Another approach to try is this:
```diff
--- a/include/curl/mprintf.h
+++ b/include/curl/mprintf.h
@@ -36,9 +36,9 @@ extern "C" {
   defined(__IAR_SYSTEMS_ICC__)) &&                                      \
   defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) &&         \
   !defined(CURL_NO_FMT_CHECKS)
-#if defined(__MINGW32__) && !defined(__clang__)
+#if defined(__MINGW32__)
 #define CURL_TEMP_PRINTF(fmt, arg) \
-  __attribute__((format(gnu_printf, fmt, arg)))
+  __attribute__((format(__MINGW_PRINTF_FORMAT, fmt, arg)))
 #else
 #define CURL_TEMP_PRINTF(fmt, arg) \
   __attribute__((format(printf, fmt, arg)))
```

I'll try it once fixed the stray standard printfs in http/client.

bagder:
The `CURL_FORMAT_CURL_` defines are meant to work with any printf(). We typically discourage applications from using the curl provided printf functions.

vszakats:
> The `CURL_FORMAT_` defines are meant to work with any printf(). We typically discourage applications from using the curl provided printf functions.

Ah OK, got it. It throws out the original idea here, but the format-spec way _might_ be a way to go that's compatible with it. Basically we need to make sure the formats and format-checker are in sync with each other and with standard printf, while curl printf accepts any format flavour anyway.

Also perhaps introducing curl printf into example was a mistake in the previous PR, and should be reverted.

vszakats:
The initial idea checks out at this point of the PR. Now turning onto the other solution.

vszakats:
With the clean approach we bump into another problem: curl internally uses the `%z` format (for `size_t*` and `curl_socket_t`), which doesn't fit the `__MINGW_PRINTF_FORMAT` spec with mingw-w64 v7 (2019-11-10) and earlier:
https://github.com/curl/curl/actions/runs/10506559918/job/29106453350?pr=14643

(This might have been the reason why I didn't go with that after testing it in #12489)

One clean (IMO), but laborous fix is to swap `%z` with a macro that's set to work for all mingw versions. Also `CURL_FORMAT_SOCKET_T`.

The upside is that this only touches internals, not the public headers.

vszakats:
@bagder: After a series of tests, the only way I see to replace
the build hack in place now, is to swap `%zd` and `%zu` within
the source to the newly added `CURL_FORMAT_[S]SIZE_T`.

This brings the format-check, the format specifiers, and the 
source code in sync with each other, regardless of mingw 
version. It also gives a tool to handle it with any other compiler
that has issues with C99 format specifiers.

(To selectively disable format checks for old mingw-w64
versions, we'd need to include `_mingw.h` to the public curl
headers. I wouldn't want to take that risk. This header may
alter the compiler env, and in a public header this is an issue.)

Do you think I should go ahead with this?

vszakats:
Script to verify that format-string replacements were done correctly:
```shell
curl -LO https://github.com/curl/curl/pull/14643.diff
grep '^-' 14643.diff > 14643-out.diff
grep '^+' 14643.diff > 14643-in.diff
grep -E -o '%[0-9]*z(x|u|d)' 14643-out.diff > 14643-out-fmt.diff
grep -E -o '%[0-9]*" CURL_FORMAT_(X|S)?SIZE_T' 14643-in.diff > 14643-in-fmt.diff
paste 14643-in-fmt.diff 14643-out-fmt.diff | sort > 14643-match.txt
```
The 493 changes checked out OK.

vszakats:
Script to verify if the strings remained the same after this patch:
```shell
curl -LO https://github.com/curl/curl/pull/14643.diff
grep -E '^(\+|@)' 14643.diff | grep -E -o '("[^"]+"|^\+\+\+.+|^@@)' | tr $'\n' $'\t' | sed -E -e 's/\"\t\"//g' -e 's/\+\+\+ b/--- a/g'       | tr $'\t' $'\n' > in
grep -E '^(-|@)'  14643.diff | grep -E -o '("[^"]+"|^---.+|^@@)'    | tr $'\n' $'\t' | sed -E -e 's/\"\t\"//g' -e 's/(%[0-9]*)z(x|u|d)/\1/g' | tr $'\t' $'\n' > out
diff -u out in > 14643-strings.diff
```

They did, except two places where I added missing spaces to existing strings (plus code changes outside of strings):
```diff
--- out	2024-08-27 11:33:52.629836670 +0200
+++ in	2024-08-27 11:33:52.618705457 +0200
@@ -1,9 +1,7 @@
 --- a/configure.ac
 @@
-"$curl_cv_native_windows$have_mingw"
 --- a/docs/examples/CMakeLists.txt
 @@
-"__USE_MINGW_ANSI_STDIO=1"
 --- a/docs/examples/Makefile.am
 @@
 --- a/docs/examples/ftpgetinfo.c
@@ -85,10 +83,10 @@
 "Included max number of cookies (%) in request!"
 --- a/lib/curl_setup.h
 @@
-"lldllucurl/mprintf.h"
 @@
+"curl/mprintf.h"
 @@
-"zd"
+"xudzxzuzd"
 --- a/lib/cw-out.c
 @@
 "cw_out, wrote % %s bytes -> %"
@@ -170,7 +168,7 @@
 @@
 "MEM %s:%d malloc(%) = %p\n"
 @@
-"MEM %s:%d calloc(%,%) = %p\n"
+"MEM %s:%d calloc(%, %) = %p\n"
 @@
 "MEM %s:%d strdup(%p) (%) = %p\n"
 @@
@@ -485,7 +483,7 @@
 @@
 "WS: decode ending with % frame bytes remaining"
 @@
-"WS: starting new frame with % bytes from last oneremaining to be sent"
+"WS: starting new frame with % bytes from last one remaining to be sent"
 @@
 "WS, using chunk size %"
 @@
@@ -531,7 +529,6 @@
 "Bad variable name length (%), skipping"
 --- a/tests/http/clients/CMakeLists.txt
 @@
-"__USE_MINGW_ANSI_STDIO=1"
 --- a/tests/http/clients/Makefile.am
 @@
 --- a/tests/libtest/lib1156.c
```

jan2000:
> As discussed, I replaced all `%z` format strings with new `CURL_FORMAT_[X|S]SIZE_T` macros.

Might I ask where this is discussed? Because I fail to see the pros of this change vs the cons. This seem like a step backward for code legibility. A workaround for an older runtime over the whole code base, while there already is a less impactful workaround in place.

vszakats:
Format masks/checks are a long time issue (a not a fun one at that). Despite spending quite a bit of time finding a workaround for making it work everywhere on all platforms (in particular all mingw-w64 versions), I have found none that doesn't have other downsides (like including `_mingw.h` in a public curl header).

This and referenced PRs (and #14703 made after this PR) have/had plenty of tries in sub commits with just as much dead-ends. 

Ultimately the problem is using C99 format strings in a C89 project.

Certainly the clean solution here is "annoying" (just like the existing solution for 64-bit formats).

If you know a better one that makes pass all cases and still offers format checks, I guess everyone would be happy to use it as a better alternative.

bagder:
> The root issue is using C99 format strings in a C89 project.

I would perhaps go a little further and say:

The root issue is using C99 format strings in a C89 project - **and asking the compiler to verify the printf formatting**.

bagder:
We could also just switch off the printf checking and save ourselves from **a lot** of trouble...

jan2000:
> Ultimately the problem is using C99 format strings in a C89 project.

Is this the problem though? If the C99 format is parsed by C89 code this seems perfectly fine to me...

> **and asking the compiler to verify the printf formatting.**

This indeed seems more like the problem, to verify with compilers that are incompatible with the format strings used.

> We could also just switch off the printf checking and save ourselves from a lot of trouble...

I would agree this would be a far better choice (for incompatible compilers) than to decrease the code legibility all over the place.

bagder:
@vszakats we could probably switch if off for *all* mingw compilers without that being too complicated

vszakats:
Switching it off all mingw is not complicated, but in that case we'll lose checks for that platform completely.

It basically means that formats will no longer be checked on Windows at all, and also that Windows-specific strings will no longer be checked at all. Returning to pre-PR #12489 where these fallout were fixed after adding format check for the first time for these.

I don't think that's a very useful way to go.

vszakats:
> > Ultimately the problem is using C99 format strings in a C89 project.
> 
> Is this the problem though? If the C99 format is parsed by C89 code this seems perfectly fine to me...

C99 format is or isn't parsed perfectly depending on Windows compiler vendor, compiler version, CRT, format check configuration. Windows is a weird place.

bagder:
> It basically means that formats will no longer be checked on Windows at all

Right, but that's not a terribly big deal IMHO. We have not checked the format for long (meaning that we can survive pretty well without them) and most flags can still be checked fine on other platforms.

> I don't think that's a useful way to go.

I'm beginning to think so. Right now that seems like a lower price than this format change.

jan2000:
> C99 format is or isn't parsed perfectly depending on Windows compiler vendor, compiler version, CRT, format check configuration. Windows is a weird place.

The C99 format is parsed in curl by the code in mprintf.c, right? Then it does not depend at all on those things.

> It basically means that formats will no longer be checked on Windows at all, and also that Windows-specific strings (or Windows-specific uses of strings) will no longer be checked at all.

Does mingw-w64 >= v8.0.0 have a compatible format checker? Maybe it can be kept on with those versions without having to do a format change.

vszakats:
> > C99 format is or isn't parsed perfectly depending on Windows compiler vendor, compiler version, CRT, format check configuration. Windows is a weird place.
> 
> The C99 format is parsed in curl by the code in mprintf.c, right? Then it does not depend at all on those things.

Correct, the actual parsing doesn't depend on it. However the compiler's checker can only check for the standard it can  understand. It doesn't know it's curl's printf, which eats everthing.

> > It basically means that formats will no longer be checked on Windows at all, and also that Windows-specific strings (or Windows-specific uses of strings) will no longer be checked at all.
> 
> Does mingw-w64 >= v8.0.0 have a compatible format checker? Maybe it can be kept on with those versions without having to do a format change.

This is what I tried first in this PR. It works. But to check for the mingw-w64 version, curl's public header must include mingw-w64's internal header `_mingw.h` which sets its version macro. Pulling that in implicitly to all code using libcurl seems ...risky. `_mingw.h` has side effects like setting a default Windows target version. That's actually [the reason why](https://github.com/curl/curl/pull/12217) we include it to curl's own code. For other code it might be undesired.

(Notice that mingw-w64 GCC might not be the sole target affected by issues around this. mingw-w64 just happens to be the first "problematic" target where we (I) enabled format checks. We haven't tried with the more popular MSVC.)

vszakats:
I gave #14703 another go and managed to make it work (though CI haven't fully finished yet).

The idea is to override the format check in public `curl/mprintf.h` with curl internal ones _while compiling curl itself_.

This makes it possible to sync public format checks with public format strings in `curl/system.h`.
Which fixes format checks for 3rd-party code. Also examples and tests/http/clients, allowing to drop build-level hacks.

Internal format checks continue to use GNU-style, which supports `%z` with all mingw-w64 versions.

The downside is that this is still compiler-specific hack (albeit the best looking one so far). It relies on the mingw-w64 feature that allows controlling format check style, adapting to C99 code internally, while sticking to whatever the compiler default is externally.

As for this PR, I agree these format macros are long and not pleasing to type or read (true for existing 64-bit ones too).
@icing already added nicer aliases for the latter: `CURL_PRId64` / `CURL_PRIu64`. I wonder if reception would be more positive if using a friendlier name for `size_t`, such as `CURL_zd` / `CURL_zu`? Food for thought in case we ever want to go full portable.

vszakats:
Closing this one in favour of #14703.

PR #14647, opened 2024-08-22
src: delete `curlx_m*printf()` aliases

vszakats:
A couple of seemingly random calls used them.
They were all mapped to `curl_m*printf()`.


bagder (approved):
Ack.

We once introduced them back in the days when we still planned to eventually remove the public functions from the API. A plan we have since given up.

PR #14654, opened 2024-08-23
CI: consolidate workflows for source and docs check

dfandrich:
A bunch of tiny jobs that run various source or documentation checks are
consolidated into two workflow files: checksrc.yml and checkdocs.yml.
This reduces the proliferation of new files containing one-line checks
and brings those that operate similarly together for better reasoning
about them.

Various calls to find, xargs and git ls-files are changed where possible
to use NUL line terminators in pipes to avoid issues with oddly-named
files that might find their way into the repo.

TODO: amend .github/labeler.yml once PR#14649 is submitted

Closes #14654


.github/workflows/checkdocs.yml
```diff
@@ -0,0 +1,146 @@
+# Copyright (C) Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# SPDX-License-Identifier: curl
+
+# This workflow contains tests that operate on documentation files only. Some
+# checks modify the source so they cannot be combined into a single job.
+
+name: Documentation Checks
```
bagder, line 8: I propose a shorter version. Maybe just drop the superfluous `Checks` part? Maybe call it just `Docs`?

image

.github/workflows/checksrc.yml
```diff
@@ -2,28 +2,93 @@
 #
 # SPDX-License-Identifier: curl
 
-name: checksrc
+# This workflow contains checks at the source code level only.
+
+name: Source Checks
```
bagder, line 7: This can then become just `Source`.

bagder (commented):
Nice!

.github/workflows/checksrc.yml
dfandrich: Done and done.

vszakats:
@dfandrich Would you like me to merge #14650 with yamllint, and you consolidate, or me waiting for merging this, and I consolidate?

dfandrich:
Since this is ready to go, I'll push it shortly so you can put the job in one of these.

PR #14662, opened 2024-08-23
Connection Pools

icing:
## Connection Pools

Rename "connection cache/conncache" to "Connection Pools/cpool" in source. This is a better match for what they do and the general "cpool" var/function prefix works well.

### Functional Changes

The pool now handles very long hostnames correctly. 

`conn->bundle->multiuse` is no more. It was used to detect if a destination (host+port) supports HTTP/2 and can "multiplex" transfers. With HTTP/3, this is no longer the right thing to do. A destination may talk HTTP/1.1 on port 443 *and* support HTTP/3 there as well.

### Implementation Changes

The following changes have been made:

* 'struct connectdata', e.g. connections, keep new members named `destination` and ' destination_len' that fully specifies interface+port+hostname of where the connection is going to. This is used in the pool for "bundling" of connections with the same destination. There is no limit on the length any more.
* Locking: all locks are done inside conncache.c when calling into the pool and released on return. This eliminates hazards of the callers keeping track.
* 'struct connectbundle' is now internal to the pool. It is no longer referenced by a connection.
* 'bundle->multiuse' no longer exists. HTTP/2 and 3 and TLS filters no longer need to set it. Instead, the multi checks on leaving MSTATE_CONNECT or MSTATE_CONNECTING if the connection is now multiplexed and new, e.g. not conn->bits.reuse. In that case the processing of pending handles is triggered.
* The pool's init is provided with a callback to invoke on all connections being discarded. This allows the cleanups in `Curl_disconnect` to run, wherever it is decided to retire a connection.
* Several pool operations can now be fully done with one call. Pruning dead connections, upkeep and checks on pool limits can now directly discard connections and need no longer return those to the caller for doing that (as we have now the callback described above).
* Finding a connection for reuse is now done via `Curl_cpool_find()` and the caller provides callbacks to evaluate the connection candidates.
* The 'Curl_cpool_check_limits()' now directly uses the max values that may be set in the transfer's multi. No need to pass them around. Curl_multi_max_host_connections() and Curl_multi_max_total_connections() are gone.
* Add method 'Curl_node_llist()' to get the llist a node is in. Used in cpool to verify connection are indeed in the list (or not in any list) as they need to.

### Follow Ups

I left the conncache.[ch] as is for now and also did not touch the documentation. If we update that outside the feature window, we can do this in a separate PR.

Multi-thread safety is not achieved by this PR, but since more details on how pools operate are now "internal" it is a better starting point to go for this in the future. To have safe multi-thread use, one would need to split the INUSE connections from the idle ones, for example. With only the idle in a pool that can be shared. Etc. etc.

**Update**: added support for what formerly `BUNDLE_NO_MULTIUSE` tried to achieve. When the destination search for a reusable connection sees a single-use, busy connection and no multiplexed ones, it seems better to not wait for any pending connection. The server probably will not suddenly negotiate multiplexing on the next one.

### DIFF
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -47,202 +47,242 @@
 #include "curl_memory.h"
 #include "memdebug.h"
 
-#define HASHKEY_SIZE 128
 
-static void connc_discard_conn(struct conncache *connc,
-                               struct Curl_easy *last_data,
+#define CPOOL_IS_LOCKED(c)    ((c) && (c)->locked)
+
+#define CPOOL_LOCK(c)                                                   \
+  do {                                                                  \
+    if((c)) {                                                           \
+      if(CURL_SHARE_KEEP_CONNECT((c)->share))                           \
+        Curl_share_lock(((c)->idata), CURL_LOCK_DATA_CONNECT,           \
+                        CURL_LOCK_ACCESS_SINGLE);                       \
+      DEBUGASSERT(!(c)->locked);                                        \
+      (c)->locked = TRUE;                                               \
+    }                                                                   \
+  } while(0)
+
+#define CPOOL_UNLOCK(c)                                                 \
+  do {                                                                  \
+    if((c)) {                                                           \
+      DEBUGASSERT((c)->locked);                                         \
+      (c)->locked = FALSE;                                              \
+      if(CURL_SHARE_KEEP_CONNECT((c)->share))                           \
+        Curl_share_unlock((c)->idata, CURL_LOCK_DATA_CONNECT);          \
+    }                                                                   \
+  } while(0)
+
+
+/* A list of connections to the same destinationn. */
+struct cpool_bundle {
+  struct Curl_llist conns; /* connections in the bundle */
+  size_t dest_len; /* total length of destination, including NUL */
+  char *dest[1]; /* destination of bundle, allocated to keep dest_len bytes */
+};
+
+
+static void cpool_discard_conn(struct cpool *cpool,
+                               struct Curl_easy *data,
                                struct connectdata *conn,
                                bool aborted);
-static void connc_disconnect(struct Curl_easy *data,
-                             struct connectdata *conn,
-                             struct conncache *connc,
-                             bool do_shutdown);
-static void connc_run_conn_shutdown(struct Curl_easy *data,
+static void cpool_close_and_destroy(struct cpool *cpool,
+                                    struct connectdata *conn,
+                                    struct Curl_easy *data,
+                                    bool do_shutdown);
+static void cpool_run_conn_shutdown(struct Curl_easy *data,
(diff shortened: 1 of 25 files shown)


calvin2021y:
Can we has a options, to force alway wait on connection for new request?

for example if set `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE`, or `CURL_HTTP_VERSION_2_0`, `CURL_HTTP_VERSION_2TLS`.   I know the server must support multiplexing.

icing:
> Can we has a options, to force alway wait on connection for new request?
> 
> for example if set `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE`, or `CURL_HTTP_VERSION_2_0`, `CURL_HTTP_VERSION_2TLS`. I know the server must support multiplexing.
> 
> or some options like `CURL_HTTP_VERSION_2ONLY` for tls, and `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE`.

Do you think it does not wait? I believe the implementation does always wait if there is a connect ongoing, pipewait for the transfer set and there is no existing connection that is single use. It goes like this (modulo bugs I wrote):

* if there is no connection, start a new one
* if there is a multiplexed connection, add to that one, if possible
* if there is an unconnected connection and pipewait is set
  * if there is no other single-use connection, wait

`http:` connections that want HTTP/2 and send "Upgrade:" requests are a bit of a bastard. They are single-use after connect, but may later, when the server responds with "101 Switch", turn to multiplexed ones. For those, we currently do not wait when checking between connect and the 101  response.

calvin2021y:
> if there is an unconnected connection and pipewait is set
>   ---  if there is no other single-use connection, wait
>   ---  otherwise, make a new connection

if client start a request with `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE`  and `CURLOPT_PIPEWAIT`,   I expect this result.

I guess this also should apply to `happy eyeballs`.

bagder (approved):

bagder:
> if client start a request with CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE and CURLOPT_PIPEWAIT, I expect this result.

Yes?

> I guess this also should apply to happy eyeballs.

How?

icing:
> > if there is an unconnected connection and pipewait is set
> > ---  if there is no other single-use connection, wait
> > ---  otherwise, make a new connection
> 
> if client start a request with `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE` and `CURLOPT_PIPEWAIT`, I expect this result.
> 
> I guess this also should apply to `happy eyeballs`.

"Happy Eyeballs" (on IP and HTTP versions) is independent of this. I try to add tests for the prior knowledge.

icing:
New tests proposed in #14694.

PR #14665, opened 2024-08-23
CI: add a job to run cmakelint

### LINKED ISSUE #14580: cmakelint warnings

dfandrich:
There are currently many issues that are raised, so just ignore the
error code for now until the issues are addressed or particular checks
are disabled.

Ref: 211cbcb4f6a050514450b7c0ddf33513e6c5f0d6 #14601
Ref: dcf5a5383ca3c46a32c66a00973df1af9c27712a #14583
Fixes #14580
Closes #14665

### DIFF
--- a/.github/workflows/checksrc.yml
+++ b/.github/workflows/checksrc.yml
@@ -13,11 +13,9 @@ name: Source
       - '*/ci'
     paths-ignore:
       - '**/*.md'
-      - '**/CMakeLists.txt'
       - '.azure-pipelines.yml'
       - '.circleci/**'
       - 'appveyor.*'
-      - 'CMake/**'
       - 'plan9/**'
       - 'tests/data/**'
       - 'winbuild/**'
@@ -26,11 +24,9 @@ name: Source
       - master
     paths-ignore:
       - '**/*.md'
-      - '**/CMakeLists.txt'
       - '.azure-pipelines.yml'
       - '.circleci/**'
       - 'appveyor.*'
-      - 'CMake/**'
       - 'plan9/**'
       - 'tests/data/**'
       - 'winbuild/**'
@@ -47,7 +43,7 @@ jobs:
       - name: check
         run: git ls-files -z "*.[ch]" | xargs -0 -n1 ./scripts/checksrc.pl
 
-  codespell:
+  codespell-cmakelint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
@@ -57,10 +53,14 @@ jobs:
         run: |
           sudo rm -f /etc/apt/sources.list.d/microsoft-prod.list
           sudo apt-get update
-          sudo apt-get install codespell
+          sudo apt-get install codespell python3-pip
+          python3 -m pip install cmakelint==1.4.3
 
       - name: spellcheck
         run: codespell --skip src/tool_hugehelp.c -I .github/scripts/codespell-ignore.txt include src lib
+
+      - name: cmakelint
+        run: scripts/cmakelint.sh
 
   reuse:
     runs-on: ubuntu-latest
--- a/scripts/Makefile.am
+++ b/scripts/Makefile.am
@@ -24,7 +24,7 @@
 
 EXTRA_DIST = coverage.sh completion.pl firefox-db2pem.sh checksrc.pl    \
  mk-ca-bundle.pl schemetable.c cd2nroff nroff2cd cdall cd2cd managen \
- dmaketgz release-tools.sh verify-release
+ dmaketgz release-tools.sh verify-release cmakelint.sh
 
 ZSH_FUNCTIONS_DIR = @ZSH_FUNCTIONS_DIR@
 FISH_FUNCTIONS_DIR = @FISH_FUNCTIONS_DIR@
--- a/scripts/cmakelint.sh
+++ b/scripts/cmakelint.sh
@@ -1,0 +1,50 @@
+#!/bin/sh
+#***************************************************************************
+#                                  _   _ ____  _
+#  Project                     ___| | | |  _ \| |
+#                             / __| | | | |_) | |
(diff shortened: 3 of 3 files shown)


.github/workflows/checksrc.yml
```diff
@@ -57,11 +53,15 @@ jobs:
...
           sudo apt-get update
-          sudo apt-get install codespell
+          sudo apt-get install codespell python3-pip
+          python3 -m pip install cmakelint
 
       - name: spellcheck
         run: codespell --skip src/tool_hugehelp.c -I .github/scripts/codespell-ignore.txt include src lib
 
+      - name: cmakelint
+        run: find . -type f \( -name CMake\* -o -name \*.cmake \) \! -name \*.h.cmake -print0 | xargs -0 cmakelint --spaces=2 --linelength=132 || echo Ignoring errors
```
vszakats, line 63: What do you think of moving out this line to a script and using `git ls-files` as a filelist source? It would allow running the script locally (in a dirty sandbox).

Also, these options silence all three issues detected:
```
--filter=-whitespace/indent,-convention/filename,-package/stdargs
```

- `whitespace/indent` has bogus warnings, and it also failed to catch obvious indentation errors, so we're probably fine without it till fixed.
- `convention/filename`, `-package/stdargs` both insist on using an uppercase package name. As discussed earlier, this is an option, but nobody seems to be following this convention. After thinking about it, I went with the mixed-case option. Unless we override this and rename most Find modules to uppercase, these warnings are not useful.

For these reasons I think it's safe to suppress them, and possibly even allow `cmakelint` to fail for other cases. E.g. to pick up any overly long lines.

dfandrich:
Analysis of PR #14665 at [837e2e4de](https://github.com/curl/curl/commit/837e2e4de3b6f2fd17c6522a3e0900aa0be118a1):

[Test 352 failed](https://circleci.com/gh/curl/curl/104796), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 210 failed](https://circleci.com/gh/curl/curl/104800), which has NOT been flaky recently, so **there could be a real issue in the PR**.

.github/workflows/checksrc.yml
dfandrich: I wasn't sure how many of those issues you were planning on fixing and how many were just limitations in the tool. I've added the filter you suggested and created a script so it can be run locally, but using find instead of git to get the file list so that it will also work from tarballs.

vszakats: Thanks! The find method picks up things non-repo, too (entirely my faul, but there is a lot of non-compliand stuff in there). I'd imagine this script is primarily useful for people who plan to push code to the Git repo, so assuming a sandbox (and thus `git ls-files` should be fine?

CMakeLists.txt
```diff
@@ -86,7 +86,8 @@ message(STATUS "CMake platform flags:${_flags}")
 set(_flags)
 
 if(CMAKE_CROSSCOMPILING)
-  message(STATUS "Cross-compiling: ${CMAKE_HOST_SYSTEM_NAME}/${CMAKE_HOST_SYSTEM_PROCESSOR} -> ${CMAKE_SYSTEM_NAME}/${CMAKE_SYSTEM_PROCESSOR}")
+  message(STATUS "Cross-compiling: ${CMAKE_HOST_SYSTEM_NAME}/${CMAKE_HOST_SYSTEM_PROCESSOR} -> \
+${CMAKE_SYSTEM_NAME}/${CMAKE_SYSTEM_PROCESSOR}")
```
vszakats, line 90:
```suggestion
  message(STATUS "Cross-compiling: "
    "${CMAKE_HOST_SYSTEM_NAME}/${CMAKE_HOST_SYSTEM_PROCESSOR} -> "
    "${CMAKE_SYSTEM_NAME}/${CMAKE_SYSTEM_PROCESSOR}")
```

.github/workflows/checksrc.yml
dfandrich: I'd be surprised if I ran cmakelint.sh on my code and it silently skipped checking my cmakefiles just because they weren't committed yet. I'd want it to check everything, regardless of their commit status and regardless of whether or not I was working in a private repository that wasn't git. The fact that it finds non-compliant stuff in local files sounds like a feature to me ☺

vszakats: Maybe, but adding new CMake files happens very rarely, and checking for lint issues before committing the first iteration is a tiny fraction in the life of a CMake source file. Before pushing, it's already committed and appears in `git ls-files`, ready to be checked.

`find .` also picks up generated CMake files for all out-of-tree build subdirs :( I almost always have some of those while working on patches.

How about this version which works both in non-Git and Git:
```shell
# Run cmakelint on the curl source code. cmakelint can be installed from PyPi
# with the command "python3 -m pip install cmakelint".
{
  if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
    git ls-files
  else
    find . -type f
  fi
} | grep -E '(/CMake|\.cmake$)' | grep -v -E '(\.h\.cmake|\.in)$' | xargs \
cmakelint \
  --spaces=2 --linelength=132 \
  --filter=-whitespace/indent,-convention/filename,-package/stdargs \
  "$@"
```

dfandrich: As discussed, the script now takes any filename arguments first, then falls back to git then find, similar to your suggestion.

vszakats: Nice! Thank you, Dan!

vszakats (approved):

PR #14669, opened 2024-08-24
tests: tweak use of impacket in smbserver

dfandrich:
The missing message only uses a single line now and has a consistent
prefix making it easier for log parsers to skip. Remove a precheck test
looking for impacket in test1451 because it's incomplete, unnecessary
and in the wrong layer; the test harness will automatically discover
that the smbserver won't start and the test was only checking a single
dependency.  Also update the tests README to no longer mention Python 2.
The servers were switched to Python 3 several years ago and no attempt
is being made to maintain Python 2 compatibility

Closes #14669

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -54,31 +54,21 @@ SPDX-License-Identifier: curl
   - nghttpx (for HTTP/2 and HTTP/3 tests)
   - An available `en_US.UTF-8` locale
 
-### Installation of python-impacket
+### Installation of impacket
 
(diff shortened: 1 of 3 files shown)


dfandrich:
Analysis of PR #14669 at [6f352264](https://github.com/curl/curl/commit/6f35226415d70f0b21be760a13d9c880d965147b):

[Test 352 failed](https://circleci.com/gh/curl/curl/104856), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14678, opened 2024-08-25
GHA/configure-vs-cmake: add Windows build, fix issues

vszakats:
- configure: disable pthreads by default on Windows.
- configure: disable detecting `fseeko()` on Windows.
  (It exists in mingw-w64 2.0.0 and newer, but it's permanently ignored
  in CMake, as this function is never necessary on Windows.)
- extend existing exceptions with their Windows variants.
- `lib/formdata.c`: prioritize `_fseeki64()` over `fseeko()`.
  To reduce the difference between Windows builds, which now all use
  `_fseeki64()`.
- cmake: perm-enable `HAVE_DIRENT_H` and `HAVE_OPENDIR` for mingw-w64,
  to match configure.
  Follow-up to bfe54b0e88239da542493321e795cd71c14af9cc #13137
  This in theory could make the dir listing feature work in mingw-w64
  build, but in my tests (on WINE) it failed at the preceding `open()`
  call.
- cmake: perm-enable `HAVE_STRINGS_H` and `HAVE_UTIME_H` for mingw-w64,
  to match configure. (They are wrappers and make no difference in the build.)

Also:
- configure: sync `USE_MANUAL` macro with cmake, by only setting it for
  `src`. Drop checker exception.
- CI: use `--disable-dependency-tracking` in existing jobs.
- CI: install packages before git checkout, in existing jobs.

---

- [x] fix pthread vs win32 thread detection difference. cmake uses native always.
- [x] smoothen/ignore the fseeko detection. or an alternative.
- [ ] cmake: add option to use pthreads (=or to not use native threads) with Windows?


dfandrich:
Analysis of PR #14678 at [a88c48ab](https://github.com/curl/curl/commit/a88c48abbc714f7b3fa183c893b30039750a94f8):

[Test 284 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 283 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 285 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 286 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 332 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 334 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1007 failed](https://github.com/curl/curl/actions/runs/10546738592/job/29218612574#check-step-39), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14685, opened 2024-08-26
Skip EOS receive when download done

icing:
When we downloaded all we wanted,
and we did not want a response body,
and no Trailer: has been announced,
and the receive gives EAGAIN,
do not hang around unnecessarily.

Some servers are buggy in HEAD processing and
fail to send the HTTP/2 EOS. Since we do not
need any more data, end the request right there.
This will cause us to send a RST_STREAM to the
server.

refs #14670

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -307,6 +307,12 @@ static CURLcode sendrecv_dl(struct Curl_easy *data,
                                 is_multiplex, &result);
     if(nread < 0) {
       if(CURLE_AGAIN == result) {
+        if(data->req.download_done && data->req.no_body &&
+           !data->req.resp_trailer) {
+          DEBUGF(infof(data, "EAGAIN, download done, no trailer announced, "
+                 "not waiting for EOS"));
+          data->req.keepon &= ~KEEP_RECV;
+        }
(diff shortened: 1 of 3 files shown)


lib/transfer.c
```diff
@@ -307,6 +307,12 @@ static CURLcode sendrecv_dl(struct Curl_easy *data,
                                 is_multiplex, &result);
     if(nread < 0) {
       if(CURLE_AGAIN == result) {
+        if(data->req.download_done && data->req.no_body &&
+           !data->req.resp_trailer) {
+          DEBUGF(infof(data, "EAGAIN, download done, no trailer announced, "
+                 "not waiting for EOS"));
+          data->req.keepon &= ~KEEP_RECV;
```
bagder, line 314: Should this condition also mark the connection for close? Or can we expect us to handle later HTTP/2 frames arriving over it for this transfer to be correctly just ignored?

icing: Good catch. I think I'll modify that a little.

icing: Changed the implementation to use the same code path for a simulated as for a real EOS.

bagder (approved):

PR #14690, opened 2024-08-26
cmake: delete unused `NEED_LBER_H`, `HAVE_LDAP_H`

vszakats:
Both are used by `./configure` internally, not by curl C code.
CMake adds `NEED_LBER_H` to `CURL_TEST_DEFINES`, which is used by
`curl_internal_test()`, but there is no LDAP-related test made with that
call.

Thus, stop detecting and publishing these from CMake.

`NEED_LBER_H` added in 7320e53d9e17b22cacea77a89ecaa8348513f0e1.
`HAVE_LDAP_H` added in 4c5307b45655ba75ab066564afdc0c111a8b9291 (initial CMake commit).


vszakats:
`NEED_LBER_H` was added in 7320e53d9e17b22cacea77a89ecaa8348513f0e1.
Unless I'm missing something, this macro also wasn't used by this initial commit.

PR #14695, opened 2024-08-26
tests: delete `libhostname.so` and `chkhostname`

vszakats:
Before this patch, `libhostname.so` and `chkhostname` were a test
facility for overriding `gethostname()` in non-debug builds on
Linux and other Unix platforms supporting `LD_PRELOAD`.

`gethostname()` has a single use with SMTP.

The alternative way to override `gethostname()` is building in debug
mode, which allows to do this via the `CURL_GETHOSTNAME` env, on all
platforms.

Drop the `LD_PRELOAD` solution in favour of the above.

Also:
- delete inactive NTLM code with a `gethostname()` call made from it.
- streamline NTLM code by dropping a `printf()` and a macro.
- tests: stop setting `CURL_GETHOSTNAME` where unnecessary.


lib/vauth/ntlm.c
```diff
@@ -516,20 +516,8 @@ CURLcode Curl_auth_create_ntlm_type3_message(struct Curl_easy *data,
...
-  if(Curl_gethostname(host, sizeof(host))) {
-    infof(data, "gethostname() failed, continuing without");
-    hostlen = 0;
-  }
-  else {
-    hostlen = strlen(host);
-  }
-#else
   (void)msnprintf(host, sizeof(host), "%s", NTLM_HOSTNAME);
   hostlen = sizeof(NTLM_HOSTNAME)-1;
```
bagder, line 520: You can even go with 

~~~c
const char *host = NTLM_HOSTNAME;
~~~

in the declaration and avoid the msnprintf() completely.

vszakats: Nicer indeed! Would this bring it too far?:
```diff
--- a/lib/vauth/ntlm.c
+++ b/lib/vauth/ntlm.c
@@ -59,10 +59,6 @@
 /* "NTLMSSP" signature is always in ASCII regardless of the platform */
 #define NTLMSSP_SIGNATURE "\x4e\x54\x4c\x4d\x53\x53\x50"
 
-/* The fixed hostname we provide, in order to not leak our real local host
-   name. Copy the name used by Firefox. */
-#define NTLM_HOSTNAME "WORKSTATION"
-
 #if DEBUG_ME
 # define DEBUG_OUT(x) x
 static void ntlm_print_flags(FILE *handle, unsigned long flags)
@@ -490,7 +486,9 @@ CURLcode Curl_auth_create_ntlm_type3_message(struct Curl_easy *data,
   unsigned char *ptr_ntresp = &ntresp[0];
   unsigned char *ntlmv2resp = NULL;
   bool unicode = (ntlm->flags & NTLMFLAG_NEGOTIATE_UNICODE) ? TRUE : FALSE;
-  char host[HOSTNAME_MAX + 1] = "";
+  /* The fixed hostname we provide, in order to not leak our real local host
+     name. Copy the name used by Firefox. */
+  static const char host[] = "WORKSTATION";
   const char *user;
   const char *domain = "";
   size_t hostoff = 0;
@@ -515,9 +513,7 @@ CURLcode Curl_auth_create_ntlm_type3_message(struct Curl_easy *data,
     user = userp;
 
   userlen = strlen(user);
-
-  (void)msnprintf(host, sizeof(host), "%s", NTLM_HOSTNAME);
-  hostlen = sizeof(NTLM_HOSTNAME)-1;
+  hostlen = sizeof(host) - 1;
 
   if(ntlm->flags & NTLMFLAG_NEGOTIATE_NTLM2_KEY) {
     unsigned char ntbuffer[0x18];
```

bagder: Nah, that works fine!

bagder (approved):

PR #14701, opened 2024-08-27
Remove NULL pointer dereference.

moritzbuhl:
If data is NULL, then it does not make sense to pass it to failf.

With Commit 7208ff6534 this became if (!data), before it additionally checked if some http struct is available. Nonetheless, the code was wrong before too.


bagder (approved):

bagder:
Thanks!

PR #14706, opened 2024-08-28
test1013.pl: require case match for features, order match for protos, fix issue

vszakats:
Update the script for test 1013 and 1014 to require:

- case-sensitive match for the curl feature list.
  (Continue to allow case-difference for protocols. They've always been
  in uppercase within curl config.)

- matching order for the protocol list.
  (Continue to allow any order for features. autotools builds on
  platforms without `sort -f` need it. E.g. Old Linux CI)

Also:

- fix casing of the `gsasl` feature in `configure`, to match `curl -V`
  and cmake.

- delete obsolete comment.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4938,7 +4938,7 @@ if test "x$curl_psl_msg" = "xenabled"; then
 fi
 
 if test "x$curl_gsasl_msg" = "xenabled"; then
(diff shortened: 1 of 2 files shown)


vszakats:
GHA/windows issue unrelated, ref #14710.

Issue #14677, opened 2024-08-25
curl POP3 STAT shows 2 minute delay


ralfjunker:
### I did this

`curl --request STAT --ssl-reqd -v --insecure -u user:password pop3s://pop.gmx.net`

### I expected the following

I expect curl to finish immediately, but see a 2 minute delay instead.

This follows up on https://github.com/curl/curl/issues/14201#issuecomment-2296896857, but the actual cause might be unrelated.

Here is the [Windows error log](https://github.com/user-attachments/files/16739488/STAT.err.8.9.1.txt).

### curl/libcurl version

Windows:

curl 8.9.1 (x86_64-w64-mingw32) libcurl/8.9.1 LibreSSL/3.9.2 zlib/1.3.1 brotli/1.1.0 zstd/1.5.6 WinIDN libpsl/0.21.5 libssh2/1.11.0 nghttp2/1.62.1 ngtcp2/1.6.0 nghttp3/1.4.0
Release-Date: 2024-07-31
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL SSPI threadsafe UnixSockets zstd

Debian Bookworm:

curl 7.88.1 (x86_64-pc-linux-gnu) libcurl/7.88.1 OpenSSL/3.0.13 zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.3 libpsl/0.21.2 (+libidn2/2.3.3) libssh2/1.10.0 nghttp2/1.52.0 librtmp/2.3 OpenLDAP/2.5.13
Release-Date: 2023-02-20, security patched: 7.88.1-10+deb12u6
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

Windows 11, Debian Bookworm.

[...]

PR #14707, opened 2024-08-28
POP3, fix multi-line responses

icing:
Some POP3 commands are multi-line, e.g. have responses terminated by a last line with '.', but some are not. Define the known command properties and fix response handling.

Refs #14677, where the 'STAT' command is hanging as it waits for a multi-line response that never comes.

Add test case for STAT.

### DIFF
--- a/lib/pop3.c
+++ b/lib/pop3.c
@@ -83,6 +83,10 @@
 #include "curl_memory.h"
 #include "memdebug.h"
 
+#ifndef ARRAYSIZE
+#define ARRAYSIZE(A) (sizeof(A)/sizeof((A)[0]))
+#endif
+
 /* Local API functions */
 static CURLcode pop3_regular_transfer(struct Curl_easy *data, bool *done);
 static CURLcode pop3_do(struct Curl_easy *data, bool *done);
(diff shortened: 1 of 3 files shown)


lib/pop3.c
```diff
@@ -199,6 +203,47 @@ static void pop3_to_pop3s(struct connectdata *conn)
...
 #endif
 
+struct pop3_cmd {
+  const char *name;
+  unsigned short nlen;
+  BIT(multiline); /* response is multi-line with last '.' line */
+  BIT(multiline_with_args); /* is multi-line when command has args */
+};
+
+static struct pop3_cmd pop3cmds[] = {
```
bagder, line 213:
```suggestion
static const struct pop3_cmd pop3cmds[] = {
```

icing: Done.

dfandrich:
Analysis of PR #14707 at [5bbeca0f](https://github.com/curl/curl/commit/5bbeca0fc0fa5d1ae9455b6604c91d09f68ba516):

[Test 310 failed](https://github.com/curl/curl/actions/runs/10595935131/job/29362791134#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

dfandrich:
NB, there are a few more POP3 commands out there.  RFC6856 defines UTF8, 
RFC2595 defines STLS, RFC5034 defines AUTH. There are also a number of
nonstandard extensions that appear to be available on servers in the wild,
defined in https://datatracker.ietf.org/doc/html/draft-hansen-pop3-xtndext-00,
namely "XTND XMIT" and "XTND XLST"

bagder:
> there are a few more POP3 commands out there

That's exactly why `--request` does not typically offer this functionality. There is simply nothing that prevents someone from having a new command in their server and curl cannot magically be made to work with those commands. I think of it as a weakness in the protocol.

However, I think this fix from @icing is small and simple enough to be worth it as it extends the functionality a little bit at least, and it fixes this one use case that a user thought was an okay use case.

Perhaps also a little bit because POP3 is not really a protocol on the rise so maybe the need for doing many others is not that huge.

icing:
We can think about what the default behaviour for unknown commands should be. I made it that single-line is the default. "XNTD" we can add as a multi-line. Or we change the defaults.

In either case, I think we should at least be able to handle the original RFC commands correctly.

icing:
Latest push:

    add more known POP3 commands, change default

    Add more command definitions that we know of and change
    the default for unknown commands to multi-line, the behaviour
    curl had for all commands before this change.

I believe this PR is now backward compatible with the addition of handling known, single-line commands correctly.

bagder (approved):

icing:
For usability it would be nice to dump the single-line command response as the body. Otherwise it would only be available via a `-D -` argument to curl, which then dumps all silly pop3 replies.

A matter of style points.

bagder:
> For usability it would be nice to dump the single-line command response as the body

Maybe. That would be an even bigger "breakage" of established procedures though since it truly is a "header" that it returns.

PR #14709, opened 2024-08-28
remove crlf_conversions counter

icing:
Since the introduction of client writers, we check the body length in the PROTOCOL phase and do FTP lineend conversions later in the CONTENT_DECODING phase. This means we no longer need to count the conversions for length checks.


bagder:
The #14713 issues seems very timely for this. It would be cool if we could somehow verify that this change fixes that issue.

PR #14710, opened 2024-08-28
GHA/windows: vcpkg GnuTLS started breaking CI, temp drop it

vszakats:
Starting today vcpkg wants to rebuild GnuTLS but fails:
```
error: building shiftmedia-libgnutls:x64-windows failed with: BUILD_FAILED
```
Ref: https://github.com/curl/curl/actions/runs/10594890318/job/29359499149#step:5:144

Temporary solution:
- drop it from the MultiSSL job.
- replace with mbedTLS job. This job still tests libssh and I could
  not find a better place for it right away.

GnuTLS to be restored once it builds again. Possibly when this hash
reaches the GHA `windows-latest` runner:
https://github.com/microsoft/vcpkg/commit/f5ec6f30ff70f04f841436a0f36600bdbabfcfbf

Also:
- switch to Debug for the mbedTLS job. Should also work now with
  GnuTLS, once it's back:
  Ref: https://github.com/microsoft/vcpkg/pull/40473

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -420,12 +420,12 @@ jobs:
       matrix:
         include:
           - name: 'schannel MultiSSL U'
-            install: 'brotli zlib zstd libpsl nghttp2 libssh2[core,zlib] pkgconf gsasl openssl mbedtls shiftmedia-libgnutls'
+            install: 'brotli zlib zstd libpsl nghttp2 libssh2[core,zlib] pkgconf gsasl openssl mbedtls'
             arch: 'x64'
             plat: 'windows'
             type: 'Debug'
             tflags: '~1516 ~2301 ~2302 ~2303 ~2307'
-            config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON  -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DCURL_USE_MBEDTLS=ON -DCURL_USE_GNUTLS=ON -DCURL_DEFAULT_SSL_BACKEND=schannel -DCURL_USE_GSASL=ON -DUSE_WIN32_IDN=ON'
+            config: '-DENABLE_DEBUG=ON -DENABLE_UNICODE=ON  -DCURL_USE_SCHANNEL=ON  -DCURL_BROTLI=ON -DCURL_ZSTD=ON -DBUILD_SHARED_LIBS=OFF -DCURL_USE_LIBSSH2=ON -DCURL_USE_OPENSSL=ON -DCURL_USE_MBEDTLS=ON -DCURL_DEFAULT_SSL_BACKEND=schannel -DCURL_USE_GSASL=ON -DUSE_WIN32_IDN=ON'
           - name: 'openssl'
             install: 'brotli zlib zstd libpsl nghttp2 nghttp3 openssl libssh2 pkgconf gsasl c-ares'
             arch: 'x64'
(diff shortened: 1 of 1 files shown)


talregev:
@vszakats not sure why it fail. 
Do you mind If I test it first on the latest vcpkg?

talregev:
@vszakats See:
https://github.com/curl/curl/pull/14711

vszakats:
@talregev: That's a drastic solution to solve a specific problem like this.

Do you know why this suddenly fails? Is it safe to use the 3rd-party GnuTLS package? I understood @dg0yt words earlier, that it may not be?

talregev:
> @talregev: That's a drastic solution to solve a specific problem like this.
> 
> Do you know why this suddenly fails? Is it safe to use the 3rd-party GnuTLS package? I understood @dg0yt words earlier, that it may not be?

It hard to debug and understand why it fail without upload the logs of vcpkg that explain why.
On the latest it not fail anymore, and I think it because one of the dependency. GnuTLS package for windows didn't change.

dfandrich:
Analysis of PR #14710 at [b804712e](https://github.com/curl/curl/commit/b804712ecc8a160eed26c65e54da564b9da5efff):

[Test 2405 failed](https://github.com/curl/curl/actions/runs/10596381350/job/29364230605#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

vszakats:
> > @talregev: That's a drastic solution to solve a specific problem like this.
> > Do you know why this suddenly fails? Is it safe to use the 3rd-party GnuTLS package? I understood @dg0yt words earlier, that it may not be?
> 
> It hard to debug and understand why it fail without upload the logs of vcpkg that explain why. On the latest it not fail anymore, and I think it because one of the dependency. GnuTLS package for windows didn't change.

I understand that, but it's not curl CI's task to debug this problem.

If vcpkg doesn't have its own CI to catch these, it's also possible to fork the curl repo and run any debugging in the fork, where all CI jobs would run just like here, with any necessary debugging added there.

talregev:
> If vcpkg doesn't have its own CI to catch these, it's also possible to fork the curl repo and run any debugging in the fork, where all CI jobs would run just like here, with any extra tracing added there.

There is a ci on vcpkg side that catch things like that. That  why I surprised that gnutls was broken here.

dg0yt:
> Do you know why this suddenly fails?

Maybe an update of GHA runner images. (Note: Microsoft uses custom images in vcpkg CI.)

talregev:
Actually the gnutls was updated to solve the debug / release issue:
https://github.com/microsoft/vcpkg/pull/40473/

talregev:
For me point of view, we remove one of very few http3 tests that I was hoping to grow up.

vszakats:
@talregev Ok, switched that job to Debug.

Is that correct that once windows-latest upgrades to vcpkg https://github.com/microsoft/vcpkg/commit/f5ec6f30ff70f04f841436a0f36600bdbabfcfbf ([Ref](https://github.com/curl/curl/pull/14711/files#diff-5ce5c9ff86f58b1f87b35b5227b2e84cb69f022d6741e1854f3e1e181091773cR490)) or newer, this issue will fix itself?

That's also the most recent (but unrelated?) commit from today, are you aware of the specific commit that fixed this?

talregev:
> Is that correct that once windows-latest upgrades to vcpkg https://github.com/microsoft/vcpkg/commit/f5ec6f30ff70f04f841436a0f36600bdbabfcfbf ([Ref](https://github.com/curl/curl/pull/14711/files#diff-5ce5c9ff86f58b1f87b35b5227b2e84cb69f022d6741e1854f3e1e181091773cR490)) or newer, this issue will fix itself?

True

> That's also the most recent (but unrelated?) commit from today, are you aware of the specific commit that fixed this?

I don't know why it fail, so I don't know which commit solve it.


I also planning to add http3 to libressl. Merge to vcpkg will be today. We can see the effect in a week or so.
I also planning to add msh3 with http3 support.

dg0yt:
> Actually the gnutls was updated to solve the debug / release issue: [microsoft/vcpkg#40473](https://github.com/microsoft/vcpkg/pull/40473)

https://github.com/microsoft/vcpkg/pull/40473 fixes downstream debug usage of shiftmedia-libgnutls with single-config generators. It does not fix failures to build shiftmedia-libgnutls.

PR #14090, opened 2024-07-03
GHA: improve vcpkg cache, add BoringSSL ECH and LibreSSL MSVC jobs

vszakats:
- cache on a per-package basis.
  Replace manual caching with a built-in solution. It shares cached
  package builds between jobs, e.g. libssh2 only builds once
  per platform (instead of once per job). Individual packages are built
  as needed (not the whole per-job tree). It also fixes the duplicate
  cache entry issues.
  Ref: https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-actions-cache
  Follow-up to e26cbe20cbedbea0ca743dd33880517309315cb2 #13979
  Follow-up to cb22cfca69bded45bf7f9c72c8e6764990490f11 #14077

- add BoringSSL job with ECH enabled. The first such job in the curl CI.

- add LibreSSL job.

- use vcpkg pre-installed on the runner image, instead of rolling our
  own. This is quicker, simpler and more robust.
  Follow-up to e26cbe20cbedbea0ca743dd33880517309315cb2 #13979

- show pre-installed vcpkg and ports version.

- drop `gsasl` dependency till it reaches the pre-installed vcpkg ports.

- re-add `find .` to see the binaries generated.

- simplify setting up `PATH`.

- exclude failing tests for any job enabling WinIDN.

- drop collecting and uploading log archives. We already dump CMake
  logs, and our build doesn't use Ninja. Rest of files weren't generated
  by the curl build. We don't aim to debug vcpkg package builds.

Closes #14090

---

TODO:
- [x] rebase on #14077.
- [x] squash into meaningful sub-commits.
- [ ] delete cache entries after merge.

PR #14711, opened 2024-08-28
GHA/windows: Change to specific hash vcpkg

talregev:
Change to specific hash vcpkg

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -484,14 +484,31 @@ jobs:
             core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 
       - name: 'vcpkg versions'
-        timeout-minutes: 1
+        timeout-minutes: 5
         run: |
+          git -C "$VCPKG_INSTALLATION_ROOT" pull
+          git -C "$VCPKG_INSTALLATION_ROOT" checkout f5ec6f30ff70f04f841436a0f36600bdbabfcfbf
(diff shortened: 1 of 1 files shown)


vszakats:
As discussed earlier, I don't think it's good for the CI to depend on the tip of a live package manager repo.
This may fix this specific case, but it will introduce constant variance in every CI run.

talregev:
> As discussed earlier, I don't think it's good for the CI to depend on the tip of a live package manager repo. This may fix this specific case, but it will introduce constant variance in every CI run.

It will not living on the tip of a live package manager repo.
I will change it to the custom hash.

talregev:
> It also add 9 minutes to each job run?

Not really. it take only less than 2 min on each run. I downgrade the timeout to be 3 to be safe.

vszakats:
This has been discussed not long ago. I still don't think that curl's CI is the correct place to debug vcpkg issues for dependencies.

Ref: f99c08dba40307c07341013ff5f71fa8e3464ffc #14090.

talregev:
> This has been discussed not long ago. I still don't think that curl's CI is the correct place to debug vcpkg issues in dependencies.
> 
> Ref: [f99c08d](https://github.com/curl/curl/commit/f99c08dba40307c07341013ff5f71fa8e3464ffc) #14090.

ok. I can remove the logs.
Should I keep this PR?

vszakats:
I think it's also not an improvement to roll our own vcpkg repo hash. It introduces a regular maintenance task and who is to tell which hash is good and in sync with the changing `windows-latest` image? Plus the overhead. I don't this solution helps us.

talregev:
ok.

PR #14712, opened 2024-08-28
libcurl.pc: reflect when a specific OpenSSL version is required

vszakats:
- require 3.0.0+ when built against such version.
  libcurl built against OpenSSL 3.x will not work with earlier OpenSSL
  versions.

- require 3.3.0 if OpenSSL is used together with QUIC.
  Example output:
  ```
  Requires.private: zlib,libbrotlidec,libzstd,openssl >= 3.3.0,libpsl,libnghttp2,libnghttp3
  ```

First time this `pkg-config` feature is used in `libcurl.pc`.

Field notes for introducing the terrible `|` → ` ` hack to `configure`:

The `Requires` list is assembled space-separated within `configure`
and those spaces replaced with commas before passing the value
to `libcurl.pc`.

But, version-dependent items can actually contain spaces:
`openssl >= 3.0.0`.

I did the next logical thing and tried assembling the list with commas.
Turns out this gets autotools confused, picking up commas within strings
as its own separators (?) and generating corrupt shell code.

As a workaround, I used a replacement character to represent space and
substitute that before passing the value to `libcurl.pc`.

I wondered if spaces are required around the `>=` operator. There is no
spec and all examples come with them. Tests revealed the spaces are
actually required. (as of `pkg-config` 0.29.2)


bagder:
> it seems unlikely that libcurl built against OpenSSL 3.x would work with earlier OpenSSL versions.

It can't. OpenSSL 3 has a different API and SONAME etc.

vszakats:
While checking up on this now, I noticed that LibreSSL also offers an `openssl.pc` with the version set to 3.9.2.

This makes LibreSSL 3.x fit the criteria for OpenSSL 3.x and 3.3, though they are incompatible. This incompatibility exists regardless of this patch though. It might help if LibreSSL offered an alternate `libressl.pc`, and maybe publish the `OPENSSL_VERSION_NUMBER` in their `openssl.pc`. (BoringSSL doesn't offer a `.pc` file, and quictls is largely missing from distros.)

bagder:
I'm a little bit curious why this is suddenly needed. OpenSSL has bumped their SONAMEs regularly over the years and has not been backwards compatible. As libraries tend to do. It seems a bit random to me to do this just for this library for only these versions.

vszakats:
This is not a must have. It's intended to avoid selecting a wrong OpenSSL for dependents at detection time, with a reasonably clear message. Also on platform without SONAME. Instead of possibly failing at a later point with a link error. 

But, like with many build matters it's somewhat difficult to predict if this actually improves things or just adds a layer of confusion.

If this is useful, it can be added for 1.1.1 or 1.1.0 or other major breaking point, also for other libs that may make use of this.

bagder:
> Also on platform without SONAME

Let's not make another weakness in Windows confuse and worsen the situation for everyone else. Remember that virtually nobody using Windows use pkg-config.

I'm not vetoing this. I'm just saying that this risks adding friction to a lot more users on non-Windows (who are using this) than help the few people on Windows for which this is primarily targeted.

vszakats:
We're also trying to unleash pkg-config on Windows (step-by-step), but I wasn't thinking of Windows specifically for this PR. The idea came about by noticing the extra attention for OpenSSL 3 in configure, and bumping into this pkg-config feature. I'm also not sure if this is useful in practice.

If anyone has experience with this, let us know!

Meanwhile some code searching may help to see if anyone else is using this feature at all.

bagder:
In my opinion, this opens up a can of worms.

First: it does not cover all OpenSSL versions, just a (to a user rather) arbitrary subset.
Then: what about all the other dependencies that require specific versions?

Keeping that accurate and working feels like a big bite of new work I'm not sure we need to sign up for without a (strong) user demand first.

PR #14715, opened 2024-08-29
openssl: add a option CURLSHOPT_CAINFO_BLOB

AkiSakurai:
- share the certificate blob by storing the OpenSSL x509 struct in curl_hash, using x509_name as the key
- Remove the unused parameter from Curl_ssl_supports
- avoid redefining memory functions when not building libcurl, but including the soruce in libtest
- Increase the conccurent connections limit in sws
- add a test for new option with concurrency
- add a ci test for thread sanitizer


bagder:
What's the benefit of adding this to the share interface? A blob is typically a piece of memory and you can just as well just set that same blob to two easy handles, can't you?

AkiSakurai:
The blob must be parsed repeatedly for every easy handle. Consider that modern protocols like HTTP/3 and TLS 1.3 have changed significantly to just reduce few round trips. Wouldn't it be great if we didn’t slow down?

Of course, one might argue that the purpose of cURL is to provide easy access to different protocols with the same interface. If higher performance is desired, one should opt for a dedicated implementation instead of using the libcurl easy interface.

bagder:
> The blob must be parsed repeatedly for every easy handle

Isn't `CURLOPT_CA_CACHE_TIMEOUT` an existing rather efficient medicine against that?

> Wouldn't it be great if we didn’t slow down?

Yes, but we also need to consider *how*.

AkiSakurai:
> Isn't `CURLOPT_CA_CACHE_TIMEOUT` an existing rather efficient medicine against that?

The cache is stored in curl_multi, which means it works only when one is using the same curl handle or the same multi handle, which cannot be shared across multiple threads.

bagder:
It sounds to me like you are the advocating having the **CA cache** shared, and not (just) the CA blob. Is there a reason you're restricting this to only blob users, when using an external file is probably a much larger use case?

bagder:
> since the current caching mechanism is restricted to external files.

I don't think it *has to* though.

PR #14734, opened 2024-08-29
build: add `poll()` detection for cross-builds

### LINKED ISSUE #14714: HAVE_POLL_FINE is always false when cross-compiling with CMake
### I did this

I was having a weird issue when using the multi interface where the `curl_multi_poll()` call occasionally returns `CURLM_UNRECOVERABLE_POLL`.

The thing is, this issue was only reproducible in cross-compiled binaries. After a bit of investigation around `Curl_poll()`, I found that this is caused by the lack of the `HAVE_POLL_FINE` define in these builds.

Looking at [CMake/OtherTests.cmake](https://github.com/curl/curl/blob/a5682d9cb9da3212775bb227488a343d32552fd6/CMake/OtherTests.cmake#L79-L110) I see that there is indeed a check for `HAVE_POLL_FINE` inside a `if (NOT CMAKE_CROSSCOMPILING)` condition.

Removing the `NOT CMAKE_CROSSCOMPILING` check and replacing `check_c_source_runs()` with `check_c_source_compiles()` fixes this.

### I expected the following

Properly detection of `HAVE_POLL_FINE`.

### curl/libcurl version

libcurl/8.10.0-DEV BearSSL nghttp2/1.61.0

### operating system

Linux rhel 3.10.0-1160.119.1.el7.tuxcare.els2.x86_64 #1 SMP Mon Jul 15 12:09:18 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

vszakats:
For cross-builds rely on `_POSIX_C_SOURCE` to decide if `poll()` is
supported, rather than just assuming it isn't.

This may still miss to detect `poll()` support, as seen for example with
Linux MUSL cross-builds.

Also:

- GHA/curl-for-win: enable RISC-V 64 cross-target for Linux MUSL.
  (to test this case with cmake, with a false-negative.)
  The first RISC-V 64 build in curl's CI.

- GHA/curl-for-win: add arm64/intel64 job for Linux glibc.
  (to test this case with cmake, and succeed.)

- cmake: delete unnecessary `#include <sys/time.h>` from non-cross-build
  `poll()` detection snippet.
  Follow-up tp cc8b8137659e1733fdd3810c19ff5ec8db438509 #14718

Fixes #14714

---

w/o whitespace: https://github.com/curl/curl/pull/14734/files?w=1

### DIFF
--- a/.github/workflows/curl-for-win.yml
+++ b/.github/workflows/curl-for-win.yml
@@ -57,7 +57,7 @@ jobs:
         run: |
           git clone --depth 1 https://github.com/curl/curl-for-win
           mv curl-for-win/* .
-          export CW_CONFIG='-main-werror-linux-musl-x64'
+          export CW_CONFIG='-main-werror-linux-musl-r64-x64'
           export CW_REVISION='${{ github.sha }}'
           . ./_versions.sh
           docker trust inspect --pretty "${DOCKER_IMAGE}"
--- a/.github/workflows/curl-for-win.yml
+++ b/.github/workflows/curl-for-win.yml
@@ -45,6 +45,32 @@ env:
   DOCKER_CONTENT_TRUST: '1'
 
 jobs:
+  linux-glibc-llvm:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+        with:
+          path: 'curl'
+          fetch-depth: 8
+      - name: 'build'
+        run: |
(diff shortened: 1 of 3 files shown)


vszakats:
Untested, and also not tested in CI.

dfandrich:
Analysis of PR #14734 at [5020d789](https://github.com/curl/curl/commit/5020d7897c29377a72fd48cab223027ec6182637):

[Test 1540 failed](https://github.com/curl/curl/actions/runs/10624142007/job/29451901343#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

bagder (approved):

PR #14744, opened 2024-08-30
configure: fix to detect `RANDOM_FILE` in more builds

vszakats:
Before this patch, `./configure` detected `RANDOM_FILE` only for builds
with the OpenSSL backend enabled. It was initially detected regardless
of backend, then [1] made it OpenSSL-specific because `RANDOM_FILE` was
solely used with OpenSSL. Many years later it got [2] a new use in
a generic random function though, and this use only expanded since.

Make `RANDOM_FILE` detected again regardless of TLS-backend, also
syncing behaviour with cmake.

In most builds the TLS backend provides the random source, so affected
builds were limited to those without one, specifically:
- mbedTLS, when using a build without built-in random support.
- Rustls.

(It's also an option to limit detection to OpenSSL/mbedTLS/Rustls, but
it risks making the codebase  fragile.)

Also:
- exclude detection for Windows. Windows builds don't use the
  `RANDOM_FILE` macro.
- GHA/configure-vs-cmake: add Linux job to test this.

Example log demonstrating the difference between autotools and cmake:
https://github.com/curl/curl/actions/runs/10636096088/job/29487155759#step:5:31

[1] df3ca59116bff161c0fa44b1af7915dc8c1da20e
[2] 365c5ba39591fab2e60bf4f0e67d9dcf79ecc506

---

- [x] move cmake fixes (one revealed by the CI job added here) to separate PR. → #14745
- [x] rebase on #14745 to fix CI.
- [ ] move `rand: exclude arc4random for LibreSSL only` to separate PR.
- [ ] move Rustls to require arc4random to separate PR? (makes Linux rustls CI fail)

### DIFF
--- a/.github/workflows/configure-vs-cmake.yml
+++ b/.github/workflows/configure-vs-cmake.yml
@@ -69,6 +69,30 @@ jobs:
       - name: compare generated curl_config.h files
         run: ./.github/scripts/cmp-config.pl lib/curl_config.h build/lib/curl_config.h
 
+  check-macos-securetransport:
+    runs-on: macos-latest
+    steps:
+      - name: install packages
+        run: |
+          while [[ $? == 0 ]]; do for i in 1 2 3; do brew update && brew install libtool autoconf automake && break 2 || { echo Error: wait to try again; sleep 10; } done; false Too many retries; done
+
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: run configure --with-secure-transport
+        run: |
+          autoreconf -fi
+          ./configure --with-secure-transport --without-libpsl
+
+      - name: run cmake
+        run: |
+          cmake -B build -DCURL_USE_LIBPSL=OFF \
+            "-DCMAKE_C_COMPILER_TARGET=$(uname -m | sed 's/arm64/aarch64/')-apple-darwin$(uname -r)" \
+            -DCURL_USE_SECTRANSP=ON -DCURL_USE_OPENSSL=OFF \
+            -DCURL_USE_LIBSSH2=OFF
+
+      - name: compare generated curl_config.h files
+        run: ./.github/scripts/cmp-config.pl lib/curl_config.h build/lib/curl_config.h
+
   check-windows:
     runs-on: ubuntu-latest
     env:
--- a/.github/workflows/configure-vs-cmake.yml
+++ b/.github/workflows/configure-vs-cmake.yml
@@ -46,6 +46,23 @@ jobs:
       - name: compare generated curl_config.h files
         run: ./.github/scripts/cmp-config.pl lib/curl_config.h build/lib/curl_config.h
 
+  check-linux-nossl:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: run configure --without-ssl
+        run: |
+          autoreconf -fi
+          ./configure --without-ssl --without-libpsl
+
+      - name: run cmake
+        run: |
+          cmake -B build -DCURL_USE_OPENSSL=OFF -DCURL_USE_LIBPSL=OFF
+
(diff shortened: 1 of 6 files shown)


bagder:
We can easily just drop mbedtls support without its own random, then we only have rustls left.

I rather think we should restrict the detection and use of `RANDOM_FILE`.

bagder:
Most systems that build with rustls probably have `arc4random` and then RANDOM_FILE is not used there either.

dfandrich:
Analysis of PR #14744 at [261cdd8f](https://github.com/curl/curl/commit/261cdd8fb92ad0696e06a490262954e946914312):

[Test 808 failed](https://github.com/curl/curl/actions/runs/10637505784/job/29491533713#check-step-15), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 806 failed](https://github.com/curl/curl/actions/runs/10637505784/job/29491533713#check-step-15), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 414 failed](https://github.com/curl/curl/actions/runs/10637505784/job/29491529848#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1561 failed](https://github.com/curl/curl/actions/runs/10637505784/job/29491529848#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

vszakats:
Agreed. Dropping `RANDOM_FILE` completely would be ideal.

I pushed a commit to require arc4random for Rustls.

Though `arc4random` on Windows (and other, esp. non-BSD platforms) can be a pain. (If this becomes an issue perhaps the OpenBSD chacha-based implementation can be imported under lib/vtls.)

mbedTLS isn't that simple if we intend to catch this at configure (I'm not too keen on doing this). It's also an option to require `arc4random` if there isn't one built-in that libcurl supports [^1]. 

Presumable most builds have random support, so what do you think of this simple solution?:
```diff
diff --git a/lib/vtls/mbedtls.c b/lib/vtls/mbedtls.c
index 1c00cbe1d..43f389e03 100644
--- a/lib/vtls/mbedtls.c
+++ b/lib/vtls/mbedtls.c
@@ -1403,6 +1403,8 @@ static CURLcode mbedtls_random(struct Curl_easy *data,
   mbedtls_havege_random(&hs, entropy, length);
   mbedtls_havege_free(&hs);
   return CURLE_OK;
+#elif !defined(HAVE_ARC4RANDOM)
+#error "mbedTLS without built-in random requires the arc4random function"
 #else
   return CURLE_NOT_BUILT_IN;
 #endif
```

[^1]: As for mbedTLS, there is a new-style (?) function `psa_generate_random()` via `psa/crypto.h` (which CAN return failure), that's not support in libcurl. The 2nd option supported by libcurl `MBEDTLS_HAVEGE_C`, was obsoleted possibly in 3.x. The 1st option's current header `mbedtls/ctr_drbg.h` is [about to be dropped](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.1) in 4.0. Hopefully the function remains.

vszakats:
With the Rustls arc4random requirement, a CI job fails.

What about restricting this PR to fix `RANDOM_FILE` detection, and attempt the other things in separate PRs?

bagder:
> Though arc4random on Windows (and other, esp. non-BSD platforms) can be a pain

On windows we already use Curl_win32_random so it doesn't need arc4random.

> With the Rustls arc4random requirement, a CI job fails.

Presumably arc4random is not present in glibc in that ubuntu version or something. I personally would still be fine of setting that as a requirement if we by that can remove every single last use of RANDOM_FILE.

> What about restricting this PR to fix RANDOM_FILE detection, and attempt the other things in separate PRs?

It just feels odd to fix RANDOM_FILE detection first if we are about to remove use of it...

vszakats:
Yeah, but I have quite of things open, and implementing feature checks for mbedTLS and dealing with Rustls fallouts gets a bits too much now :( I'd pass, and rather pass on fixing this issue too.

vszakats:
Closing in favour of #14749.

PR #14746, opened 2024-08-31
autotools: settle with option name: `--enable-windows-unicode`

vszakats:
Bring the option name style in sync with cmake and with other configure
options aiming to enable something unrelated to an optional package.

(I initially named this new option `--with-windows-unicode` within this
release cycle.)

Follow-up to 9e4a2187e763dd80a1296b07fd3e073f46c4dc8f #14478


vszakats:
This feels more aligned with other configure options and also rhyms with cmake's `ENABLE_UNICODE`.

/cc @bagder: Made an update to RELEASE-NOTES too to feature the new name.

(if this seems too late or not a good idea, let me know!)

PR #14748, opened 2024-08-31
curl.h: make CURLOPT_WRITEINFO and CURLOPT_CLOSEPOLICY compile

### LINKED ISSUE #14747: Removal of `CURLOPT_OBSOLETE72` causes downstream breakage
Noticed in vpckg CI run on CURL HEAD:

`CURLOPT_OBSOLETE72` is still used in the header for

https://github.com/curl/curl/blob/3fc81be44e970215f6ca18ab331ac2050abda58b/include/curl/curl.h#L723

and `CURLOPT_CLOSEPOLICY` is used downstream in curlpp and curlcpp. (Edit: The problem really seems to be limited to these two C++ wrappers.)

_Originally posted by @dg0yt in https://github.com/curl/curl/pull/14634#discussion_r1739717025_

bagder:
The symbols have not been in use for 17+ years and they did not do anything for several years before that, but apparently there are still code using them.

Fixes #14747


dg0yt:
I couldn't test it until today. It seems that this is not source compatible with the original implementation when seen by C++. Both curlpp and curlcpp still fail to build from source.

MSVC reports:
~~~
FAILED: CMakeFiles/curlpp.dir/src/curlpp/Options.cpp.obj 
C:\PROGRA~1\MICROS~1\2022\ENTERP~1\VC\Tools\MSVC\1438~1.331\bin\Hostx64\x64\cl.exe   /TP -Dcurlpp_EXPORTS -ID:\b\curlpp\src\95a4f05eb9-7d2abb29f5.clean\include -ID:\installed\x64-uwp\include /DWIN32 /D_WINDOWS /D_UNICODE /DUNICODE /DWINAPI_FAMILY=WINAPI_FAMILY_APP /D__WRL_NO_DEFAULT_LIB__ /nologo /Z7 /MP /GS /Gd /Gm- /W3 /WX- /Zc:wchar_t /Zc:inline /Zc:forScope /fp:precise /Oy- /EHsc  /utf-8  /D_DEBUG /MDd /Od /RTC1 /showIncludes /FoCMakeFiles\curlpp.dir\src\curlpp\Options.cpp.obj /FdCMakeFiles\curlpp.dir\ /FS -c D:\b\curlpp\src\95a4f05eb9-7d2abb29f5.clean\src\curlpp\Options.cpp
D:\b\curlpp\src\95a4f05eb9-7d2abb29f5.clean\include\curlpp/Options.hpp(281): error C2440: 'specialization': cannot convert from 'int' to 'CURLoption'
D:\b\curlpp\src\95a4f05eb9-7d2abb29f5.clean\include\curlpp/Options.hpp(281): note: Conversion to enumeration type requires an explicit cast (static_cast, C-style cast or parenthesized function-style cast)
~~~

Apple-clang reports:
~~~
FAILED: CMakeFiles/curlpp.dir/src/curlpp/Easy.cpp.o 
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -DCURL_STATICLIB -I/dev/vcpkg/buildtrees/curlpp/src/95a4f05eb9-7d2abb29f5.clean/include -isystem /dev/vcpkg/installed/x64-osx/include -fPIC -g -std=c++11 -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.7 -MD -MT CMakeFiles/curlpp.dir/src/curlpp/Easy.cpp.o -MF CMakeFiles/curlpp.dir/src/curlpp/Easy.cpp.o.d -o CMakeFiles/curlpp.dir/src/curlpp/Easy.cpp.o -c /dev/vcpkg/buildtrees/curlpp/src/95a4f05eb9-7d2abb29f5.clean/src/curlpp/Easy.cpp
In file included from /dev/vcpkg/buildtrees/curlpp/src/95a4f05eb9-7d2abb29f5.clean/src/curlpp/Easy.cpp:26:
/dev/vcpkg/buildtrees/curlpp/src/95a4f05eb9-7d2abb29f5.clean/include/curlpp/Options.hpp:281:48: error: value of type 'int' is not implicitly convertible to 'CURLoption'
        typedef curlpp::OptionTrait<curl_closepolicy, CURLOPT_CLOSEPOLICY> ClosePolicy;
                                                      ^~~~~~~~~~~~~~~~~~~
/dev/vcpkg/installed/x64-osx/include/curl/curl.h:723:29: note: expanded from macro 'CURLOPT_CLOSEPOLICY'
#define CURLOPT_CLOSEPOLICY CURLOPT_OBSOLETE72
                            ^~~~~~~~~~~~~~~~~~
/dev/vcpkg/installed/x64-osx/include/curl/curl.h:724:28: note: expanded from macro 'CURLOPT_OBSOLETE72'
#define CURLOPT_OBSOLETE72 9999
                           ^~~~
~~~

dfandrich:
libcurl is C code, not C++. There's a limit to how far it's worth libcurl
trying to get a nearly two decade old removed feature compiling in a program
written in a different language. It's time to remove this option from the
program.

dg0yt:
> libcurl is C code, not C++.

@dfandrich That's *why* I mention language in my notification. The failing packages are C++ wrappers. Both projects were already notified but there is no reaction so far. (Linked in https://github.com/curl/curl/issues/14747.)

I don't really care about which solution is chosen, incl. removal. I just neutrally report pre-release packaging results. @badger chose this solution when I noticed removal the first time. IMO it is fair to annotate this with the fact that it doesn't solve the original problem. (And my original patch no longer works either.)

Thanks for the release countdown on mastodon @bagder.

bagder:
That problem looks like an application problem and not a curl problem.

dg0yt:
I see that curlpp/curlcpp fail to build from source with what is in `curl.h` now.

Both curlpp and curlcpp use the `CURLoption` enum values in C++ template parameters. The current curl change removed the `CURLOPT_CLOSEPOLICY` enum value and added a preprocessor macro of this name, expanding to `int` 9999.

I can't see why it is wrong to use curl enum values the way these C++ wrapper libs do.
I can't assess quickly if a cast from `int` to `CURLoption` would be a safe thing. Given that 9999 is not in `CURLoption`, I smell UB.

I can confirm that disabling those lines in curlpp/curlcpp fixes the build of the C++ wrapper libs, and it should be safe for those application which don't access the deprecated option. It will break application which use the option (even if it has no effect), and I would count this forwarding curl's (C++) source-incompatible change.

bagder:
If you consider this an issue, then please submit an issue. I agree that it is one, but I can't be bothered to since fixing it seems more work than just patching away the silliness from those source codes using this.

Since 9999 is within the enum range (and it was so even back in 2007), I don't believe it is a UB.

dg0yt:
I think I did enough reporting. Project which actually use it shall take action. 

Packaging stats:  
https://repology.org/project/curlpp/versions 
https://repology.org/project/curlcpp/versions

PR #14750, opened 2024-08-31
rand: exclude arc4random for LibreSSL (3.8.x and earlier) only

vszakats:
Use `arc4random` with OpenSSL-family again, except with LibreSSL, the
only fork that's incompatible with it.

LibreSSL 3.9.0 fixed the issue that required this workaround, so also
unlock `arc4random` for builds using that or newer.

Follow-up to 7925ba431b9a099daee1fa21d36c21887f787ad5 #12274

---

- [ ] rebase on #14749.

### DIFF
--- a/lib/rand.c
+++ b/lib/rand.c
@@ -147,7 +147,8 @@ static CURLcode randit(struct Curl_easy *data, unsigned int *rnd,
   }
 #endif
 
-#if defined(HAVE_ARC4RANDOM) && !defined(USE_OPENSSL)
+#if defined(HAVE_ARC4RANDOM) && \
+  !(defined(USE_OPENSSL) && defined(LIBRESSL_VERSION_NUMBER))
   if(!seeded) {
     *rnd = (unsigned int)arc4random();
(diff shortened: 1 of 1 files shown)


bagder:
I'm aiming for a larger removal in #14749 so maybe you can just skip this?

vszakats:
I was planning to rebase on that. This guard remains necessary, but only for LibreSSL. It's a workaround for LibreSSL exporting arc4random, causing a false-positive detection, then missing declaration when referencing it from `rand.c`.

The general issue is tracked here, and partly fixed in 3.9.x:
https://github.com/libressl/portable/issues/928

Probably can be restricted now to `LIBRESSL_VERSION_NUMBER < 0x3090000fL`.

vszakats:
Yep, confirmed that LibreSSL 3.9.0 fixed this. Patch updated.

vszakats:
#14749 no longer tries `arc4random` for LibreSSL (or any TLS backend besides Rustls), making this PR safe to close without further action.

PR #14751, opened 2024-09-01
openssl: fix the data race when sharing an SSL session between threads

AkiSakurai:
The SSL_Session object is mutated during connection inside openssl, and it might not be thread-safe. Besides, according to documentation of openssl:

```
SSL_SESSION objects keep internal link information about the session
cache list, when being inserted into one SSL_CTX object's session
cache. One SSL_SESSION object, regardless of its reference count,
must therefore only be used with one SSL_CTX object (and the SSL
objects created from this SSL_CTX object).
```
If I understand correctly, it is not safe to share it even in a single thread.

Instead, serialize the SSL_SESSION before adding it to the cache, and deserialize it after retrieving it from the cache, so that  concurrent write to the same object is infeasible.

Also
 - add a ci test for thread sanitizer
 - add a test for sharing ssl sessions concurrently
 - avoid redefining memory functions when not building libcurl, but including the source in libtest
 - increase the concurrent connections limit in sws

Notice that there are fix for a global data race for openssl which is not yet release. The fix is cherry pick for the ci test with thread sanitizer.
https://github.com/openssl/openssl/commit/d8def79838cd0d5e7c21d217aa26edb5229f0ab4

### DIFF
--- a/lib/curl_memory.h
+++ b/lib/curl_memory.h
@@ -138,7 +138,7 @@ extern curl_calloc_callback Curl_ccalloc;
 extern curl_wcsdup_callback Curl_cwcsdup;
 #endif
 
-#if !defined(CURLDEBUG) && defined(BUILDING_LIBCURL)
+#ifndef CURLDEBUG
 
 /*
  * libcurl's 'memory tracking' system defines strdup, malloc, calloc,
--- a/lib/curl_memory.h
+++ b/lib/curl_memory.h
@@ -138,7 +138,7 @@ extern curl_calloc_callback Curl_ccalloc;
 extern curl_wcsdup_callback Curl_cwcsdup;
 #endif
 
-#ifndef CURLDEBUG
+#if !defined(CURLDEBUG) && defined(BUILDING_LIBCURL)
 
 /*
  * libcurl's 'memory tracking' system defines strdup, malloc, calloc,
--- a/lib/curl_threads.c
+++ b/lib/curl_threads.c
@@ -35,7 +35,9 @@
 #endif
 
 #include "curl_threads.h"
+#ifdef BUILDING_LIBCURL
 #include "curl_memory.h"
+#endif
 /* The last #include file should be: */
 #include "memdebug.h"
 
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -2010,7 +2010,7 @@ static void ossl_session_free(void *sessionid, size_t idsize)
 {
   /* free the ID */
   (void)idsize;
-  SSL_SESSION_free(sessionid);
+  free(sessionid);
 }
 
 /*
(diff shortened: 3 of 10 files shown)


lib/curl_memory.h
```diff
@@ -138,7 +138,7 @@ extern curl_calloc_callback Curl_ccalloc;
 extern curl_wcsdup_callback Curl_cwcsdup;
 #endif
 
-#ifndef CURLDEBUG
+#if !defined(CURLDEBUG) && defined(BUILDING_LIBCURL)
```
vszakats, line 141: AFAICS the existing solution to solve this within `src` was this (e.g. in `lib/base64.c`):
```c
#ifdef BUILDING_LIBCURL
#include "curl_memory.h"
#endif
```

With this update, all those guards become redundant.
Would it make sense to delete them?

Or, stay with the existing method in this PR and update
`lib/curl_threads.c` with the same guard. Then replace
them all with this neater solution in a separate PR?

lib/curl_threads.c
```diff
@@ -35,7 +35,9 @@
 #endif
 
 #include "curl_threads.h"
+#ifdef BUILDING_LIBCURL
```
bagder, line 38: Why is this necessary? the header itself has an `#ifdef CURLDEBUG`, should that not be enough?

AkiSakurai: Because I added curl_threads.c as a source of libtest, the memory wrappers in curl_memory.h are non-public symbols that libtest can't link against in shared library mode. CURLDEBUG is defined in libtest as well.

tests/libtest/lib3207.c
```diff
@@ -0,0 +1,231 @@
...
+#endif
+test_thread(void *ptr)
+{
+  struct Ctx *ctx = (struct Ctx *)ptr;
+  CURLcode res = CURLE_OK;
+
+  int i;
+
+  /* Loop the transfer and cleanup the handle properly every lap. This will
+    still reuse ssl session since the pool is in the shared object! */
```
vszakats, line 94:
```suggestion
     still reuse ssl session since the pool is in the shared object! */
```
very small nit, indent off by one

lib/vtls/openssl.c
```diff
@@ -2010,7 +2010,7 @@ static void ossl_session_free(void *sessionid, size_t idsize)
 {
   /* free the ID */
   (void)idsize;
-  SSL_SESSION_free(sessionid);
+  free(sessionid);
```
bagder, line 2013: Really? This looks like a rather drastic change that if necessary should rather be done in its own PR. Can you explain why/how you found/decided this?

bagder: Eh sorry strike that. I see it now.

icing:
I believe this PR makes a necessary change. Well done.

bagder:
Thanks!

PR #14764, opened 2024-09-02
lib: use FMT_ as prefix instead of CURL_FORMAT_

bagder:
For printf format defines used internally. Makes the code slighly easier to read.


dfandrich:
Analysis of PR #14764 at [413ec571](https://github.com/curl/curl/commit/413ec5712ff596ad92501cbfddd85c0f68528fce):

[Test 1062 failed](https://circleci.com/gh/curl/curl/105804), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14766, opened 2024-09-03
cf-socket, fix pollset for listening

icing:
When FTP does an active data connection, the socket connection filter is instantiated with a listening socket. When the filter adjusts its pollset, it needs to POLLIN, not OUT.


bagder (approved):

bagder:
ah, it got a little conflict ...

icing:
> ah, it got a little conflict because of [4ff0461](https://github.com/curl/curl/commit/4ff04615a095ed1e7c489e3bbd4600b52e68d9a7)...

That's what one gets for being nice to @vszakats ...

icing:
rebased.

PR #14535, opened 2024-08-13
rustls: add support for setting TLS version and ciphers

jan2000:
Add support for CURLOPT_SSLVERSION, CURLOPT_TLS13_CIPHERS and CURLOPT_SSL_CIPHER_LIST.

PR #14770, opened 2024-09-03
docs: document the (weak) random value situation in rustls builds

bagder:


icing (approved):

docs/RUSTLS.md
```diff
@@ -29,3 +29,18 @@ Now configure and build curl with Rustls:
     % autoreconf -fi
     % ./configure --with-rustls=${HOME}/rustls-ffi-built
     % make
+
+## Randomness
+
+Every TLS libcurl curl supports - *except* rustls - provides a function for
+curl to extract cryptographicly safe random numbers with.
```
vszakats, line 36:
```suggestion
curl to extract cryptographically safe random numbers with.
```

docs/RUSTLS.md
```diff
@@ -29,3 +29,18 @@ Now configure and build curl with Rustls:
...
+Every TLS libcurl curl supports - *except* rustls - provides a function for
+curl to extract cryptographically safe random numbers with.
+
+When you build curl with rustls, curl uses its own internal attempts to get a
+decent random value:
+
+1. Windows specific APIs
+2. arc4random
+
+If neither of those are present, then curl using rustls falls back to **weak
```
vszakats, line 44:
```suggestion
Every TLS libcurl curl supports - *except* Rustls - provides a function for
curl to extract cryptographically safe random numbers with.

When you build curl with Rustls, curl uses its own internal attempts to get a
decent random value:

1. Windows specific APIs
2. arc4random

If neither of those are present, then curl using Rustls falls back to **weak
```
Keeping up with c5cb8e7c7e8a9603df2a8dc1de2efe664e950046 #14605

vszakats (approved):

ctz:
We will address this.

cpu:
Would you be open to tagging a Rustls dev when you bump into these sorts of issues? I hate to make more work for the `curl` folks but this sort of thing can fly under the radar otherwise. I spotted it by change when looking at `git log` while curious about another topic. The `curl` repo is tricky to set a global watch on because y'all are too productive :laughing:

bagder:
@cpu sure, if you have any preference for who/how to tag you, I could certainly make an effort to!

cpu:
> any preference for who/how to tag you

You can definitely tag me liberally. I'll let ctz decide himself if he wants the notifications :-) 

Beyond bugs/issues I'm also happy to give non-blocking reviews on Rustls-related PRs like https://github.com/curl/curl/pull/14535  It's helpful to see where our APIs might be awkward/underdocumented.

ctz:
Please do tag me too.

cpu:
We landed a capability for this upstream in the yet-to-be-released `rustls-ffi` 0.14. I will pick up a fix for the `curl` side along with an update for the relevant breaking changes once we cut a release. Thanks!

PR #14775, opened 2024-09-04
KNOWN_BUGS: cleanup

bagder:
- Move all Windows unicode issues into the single "bug", as they are all at least semi-related

- Removed ancient issues that are not relevant anymore

### DIFF
--- a/docs/KNOWN_BUGS
+++ b/docs/KNOWN_BUGS
@@ -12,8 +12,6 @@ check the changelog of the current development status, as one or more of these
 problems may have been fixed or changed somewhat since this was written.
 
  1. HTTP
- 1.2 hyper is slow
- 1.5 Expect-100 meets 417
 
  2. TLS
  2.1 IMAPS connection fails with Rustls error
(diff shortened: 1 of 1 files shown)


dfandrich:
Analysis of PR #14775 at [54845341](https://github.com/curl/curl/commit/54845341f0aec5950b579b6797653f5e24b13db1):

[Test 678 failed](https://github.com/curl/curl/actions/runs/10697352526/job/29654355338#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 671 failed](https://github.com/curl/curl/actions/runs/10697352526/job/29654355338#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1904 failed](https://github.com/curl/curl/actions/runs/10697352537/job/29654374894#check-step-23), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 3028 failed](https://github.com/curl/curl/actions/runs/10697352537/job/29654374894#check-step-23), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14777, opened 2024-09-04
schannel: avoid malloc for CAinfo_blob_digest

bagder:
...it can just as well be part of the struct always.


dfandrich:
Analysis of PR #14777 at [55108fd0](https://github.com/curl/curl/commit/55108fd004eec6d4ece3d9f7a69856f6350fcf18):

[Test 1379 failed](https://circleci.com/gh/curl/curl/105925), which has NOT been flaky recently, so **there could be a real issue in the PR**.

Issue #6313, opened 2020-12-12
Building curl via cmake ExternalProject_Add does not set CURL_CA_PATH


crazyguitar:
Hi,

I discovered that CURL_CA_BUNDLE and CURL_CA_PATH does not set properly if I used ExternalProject_Add to build curl as my dependency.

### I did this

The following snippets are how I reproduced this issue. 

```cpp
// a.cc
#include <curl/curl.h>
#include <string>
#include <iostream>

static size_t callback(void *ptr, size_t size, size_t nmemb, std::string* data)
{
  data->append((char *)ptr, size * nmemb);
  return size * nmemb;
}

int main(int argc, char *argv[])
{
  auto curl = curl_easy_init();
  std::string resp;

  curl_easy_setopt(curl, CURLOPT_URL, "https://github.com");
  curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, callback);
  curl_easy_setopt(curl, CURLOPT_WRITEDATA, &resp);

  auto rc = curl_easy_perform(curl);
  std::cout << curl_easy_strerror(rc) << "\n";
  std::cout << resp << "\n";
  curl_easy_cleanup(curl);
}
```

```cmake
# CMakeLists.txt

cmake_minimum_required(VERSION 3.10)
project(test)
set(target test)
file(GLOB SOURCE *.cc)
find_package(OpenSSL REQUIRED)
find_package(ZLIB)

link_directories("${PROJECT_BINARY_DIR}/lib")

add_executable(${target} ${SOURCE})
add_dependencies(${target} curl)
target_include_directories(${target} PRIVATE "${PROJECT_BINARY_DIR}/include")
target_link_libraries(${target} PRIVATE curl pthread ZLIB::ZLIB OpenSSL::SSL OpenSSL::Crypto)

include (ExternalProject)
ExternalProject_Add(curl
  GIT_REPOSITORY "https://github.com/curl/curl"
  GIT_TAG "master"
  GIT_CONFIG advice.detachedHead=false
  PREFIX "${CMAKE_BINARY_DIR}/curl"
  CMAKE_CACHE_ARGS
    "-DBUILD_SHARED_LIBS:BOOL=OFF"
    "-DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_BINARY_DIR}"
)
```

### I expected the following

After running my test program, I expected a response from ``github.com``. However, I received the following error:

Build and run commands:

```bash
$ ls 
a.cc  CMakeLists.txt
[...]

PR #1461, opened 2017-05-01
cmake: Add CURL_CA_BUNDLE/CURL_CA_FALLBACK/CURL_CA_PATH

webmaster128:

PR #14778, opened 2024-09-04
cmake: fix cached `SET` vars with `CURL_CA_BUNDLE`/`CURL_CA_PATH` set to `none`

vszakats:
~~Follow-up to 6a9489dc4512bf92e121d7bc08924e2922ababc0 #1461~~
~~Ref: #6313~~

---

edit: I cannot reproduce the referenced issue. Also the suggested fix doesn't make a difference in the generated `config.h` and `CMakeCache.txt` when running the repro setup.

edit2: The options seems to behave unexpectedly, but it may be my lack of understanding of caching and the ideas around how this should work. I expected the `CURL_CA_BUNDLE` and `CURL_CA_PATH` settings to "stick" between cmake invocations. They don't. On each run they spring back to "auto". Meanwhile other options (e.g. `ENABLE_UNIX_SOCKETS`) retain their previously set values.

edit3: It seems a 3rd variable would be necessary to cache the builder's intent in all cases. For now, e.g. `none` cannot stick around because its unsets `CURL_CA_BUNDLE`, which triggers a reset even after guarding the option initialization with a `NOT DEFINED`. (or store this information in the `_SET` var maybe?)


vszakats:
Untested.

vszakats:
Abandoning this effort for now.

PR #14780, opened 2024-09-04
openssl: certinfo errors now fail correctly

bagder:
If there is a (memory) error when creating the certinfo data, the code would previously continue which could lead to a partial/broken response. Now, the first error aborts and cleans up the entire thing.

A certinfo "collection" error is however still not considered an error big enough to stop the handshake.

Bonus: made two functions static (and removed the Curl_ prefix) that were not used outside of openssl.c

Bonus 2: removed the unused function `Curl_ossl_set_client_cert`

### DIFF
--- a/lib/vtls/openssl.c
+++ b/lib/vtls/openssl.c
@@ -314,29 +314,36 @@ typedef unsigned long sslerr_t;
 #define USE_PRE_1_1_API (OPENSSL_VERSION_NUMBER < 0x10100000L)
 #endif /* !LIBRESSL_VERSION_NUMBER */
 
-#define push_certinfo(_label, _num)             \
-do {                              \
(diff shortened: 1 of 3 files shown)


dfandrich:
Analysis of PR #14780 at [7fd43d88](https://github.com/curl/curl/commit/7fd43d88ac0ae98abb76a61fef5b785e8a22b56c):

[Test 1120 failed](https://circleci.com/gh/curl/curl/105935), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1540 failed](https://github.com/curl/curl/actions/runs/10698718500/job/29658742808#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14784, opened 2024-09-04
smb: convert superflous assign into assert

bagder:
Since the value is already supposed to be correct, verify that with and assert instead of doing an assign that is not needed.

Bonus: removing unnecessary clearing of data


dfandrich:
Analysis of PR #14784 at [9a4f34a6](https://github.com/curl/curl/commit/9a4f34a608b9d2a016469e6f0642bb7defc6a8b5):

[Test 3100 failed](https://github.com/curl/curl/actions/runs/10701768036/job/29668452162#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14789, opened 2024-09-05
build: tidy up deprecation suppression, enable warnings for clang

vszakats:
Suppress deprecation warnings the closest to the deprecated code, using
`CURL_IGNORE_DEPRECATION()`. Then drop build-specific suppressions, and
file-wide ones. The latter is not compatible with Unity mode. Also
replace manual suppressions with a macro to apply to all compilers with
deprecation warning support. Also enable deprecation warnings for clang.

- curl/curl.h: enable deprecation warnings for clang.

- docs/examples: stop setting `CURL_DISABLE_DEPRECATION` with autotools.
  Suppression moved to C-level earlier. Syncs with cmake.
  Follow-up to 5fc61a37c1b177cefbc11dc9f5eef7b2a14538da #14123

- tests/http/clients: stop setting `CURL_DISABLE_DEPRECATION` in
  autotools. If it becomes necessary in the future, it can be done in
  C via the macro. Syncs with cmake.

- lib1545: stop setting `CURL_DISABLE_DEPRECATION` in autotools.
  Drop guard from test source.
  Follow-up to 0f103600731f7e08637017ef4df67a194e0d6711 #12444

- libtest, unit: replace `CURL_DISABLE_DEPRECATION` with
  `CURL_IGNORE_DEPRECATION()`.

- docs/examples: replace pragmas with `CURL_IGNORE_DEPRECATION()`.

---

w/o whitespace: https://github.com/curl/curl/pull/14789/files?w=1


vszakats:
Tried to enable deprecation warnings for MSVC, but it fell over because `CURL_DEPRECATED()` is also used in enums which is apparently unsupported.

PR #14794, opened 2024-09-05
asyn-thread: stop using GetAddrInfoExW on Windows

jay:
- For the threaded resolver backend on Windows, revert back to exclusively use the threaded resolver with libcurl-owned threading instead of GetAddrInfoExW with Windows-owned threading.

Winsock (the Windows sockets library) has a bug where it does not wait for all of the name resolver threads it is managing to terminate before returning from WSACleanup. The threads continue to run and may cause a crash.

This commit is effectively a revert of several commits that encompass all GetAddrInfoExW code in libcurl. A manual review of merge conflicts was used to resolve minor changes that had modified the code for aesthetic or build reasons in other commits.

Prior to this change if libcurl was built with the threaded resolver backend for Windows, and Windows 8 or later was the operating system at runtime, and the caller was not impersonating another user, then libcurl would use GetAddrInfoExW to handle asynchronous name lookups.

GetAddrInfoExW support was added in a6bbc87f, which preceded 8.6.0, and prior to that the threaded resolver backend used libcurl-owned threading exclusively on Windows.

Reported-by: Ionuț-Francisc Oancea
Reported-by: Razvan Pricope

Ref: https://developercommunity.visualstudio.com/t/ASAN:-heap-use-after-free-in-NdrFullPoin/10654169

Fixes https://github.com/curl/curl/issues/13509#issuecomment-2225338110
Closes #xxxx

---

Revert "asyn-thread: avoid using GetAddrInfoExW with impersonation"

This reverts commit 0caadc1f24d20514eed2bf6e5ef0adc140f122c3.

Conflicts:
	lib/system_win32.c

--

Revert "asyn-thread: fix curl_global_cleanup crash in Windows"

This reverts commit 428579f5d136fd473e97fe089c42ffee55b72a8f.

--

Revert "system_win32: fix a function pointer assignment warning"

This reverts commit 26f002e02ef1142a432c8dc087bd27de71ce38bf.

--

Revert "asyn-thread: use GetAddrInfoExW on >= Windows 8"

This reverts commit a6bbc87f9e9ffb46a1801dfb983e7534825ed56b.

Conflicts:
	lib/asyn-thread.c
	lib/system_win32.c

--

### DIFF
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -54,7 +54,6 @@
 #  define RESOLVER_ENOMEM  ENOMEM
 #endif
 
-#include "system_win32.h"
 #include "urldata.h"
 #include "sendf.h"
 #include "hostip.h"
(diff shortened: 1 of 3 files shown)


bagder (approved):
Unfortunate but necessary.

PR #14798, opened 2024-09-05
ftp, move listen handling to socket filter

icing:
Move the listen/accept handling of the FTP active data connection into the socket filter and monitor 'connected' status of that as with passive connections - more or less.

The advantage is that the socket filter now reports being connected only when the server has actually called and accept() has been done. This enables to bootstrap the filter chain on the data connection just like any other. A require SSL filter can then be added right at the start and does not need to be patched in later.

Still, the active connection keeps on needing special handling in ftp.c as the control connection needs to be monitored while waiting as the server might send error responses this way. So, things did not turn out quite as squeaky clean as hoped for, but still seems better to do that way.

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -2151,7 +2151,6 @@ static CURLcode cf_tcp_accept_connect(struct Curl_cfilter *cf,
   ctx->accepted = TRUE;
   socket_close(data, cf->conn, TRUE, ctx->sock);
   ctx->sock = s_accepted;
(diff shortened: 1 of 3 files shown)


icing:
@bagder: I propose to merge this only after the next release.

bagder:
@icing if you take a look at the conflict, we should probably merge this!

icing:
> @icing if you take a look at the conflict, we should probably merge this!

Done.

PR #14799, opened 2024-09-05
tests: embed copyright and license info in each test

bagder:
... instead of a wildcard match in REUSE.toml.


bagder:
For unclear reasons the reuse tool insists the data is missing in these files...

~~~
The following files have no copyright and licensing information:
* tests/data/test1160
* tests/data/test1425
* tests/data/test1426
* tests/data/test1531
* tests/data/test2080
* tests/data/test262
* tests/data/test35
* tests/data/test463
* tests/data/test545

bagder:
@mxmehl any hints on how to figure out what makes it misbehave?

mxmehl:
> @mxmehl any hints on how to figure out what makes it misbehave?

Yes, I ran `reuse --debug lint` and found the following line:

`reuse.project - INFO - 'tests/data/test1160' was detected as a binary file; not searching its contents for REUSE information.`

The same seems to happen with the other files.

The reason is that the REUSE tool runs [binaryornot](https://pypi.org/project/binaryornot/) prior to searching for licensing/copyright information. At least when we introduced it, it was a significant performance boost.

I am not sure how you could mark one file explicitly as not-binary. If you wanted to get rid of entries in REUSE.toml, one idea could be to add `.license` files for these files, so e.g. `tests/data/test1160.license` with the REUSE information.

@carmenbianca may have a better idea though.

bagder:
> I ran `reuse --debug lint`

Ah, the --debug option was the trick. I'll try that next time.

> one idea could be to add .license files for these files

I understand. I think I instead prefer to hold off this change. If we can't put the info inline into these files (without exceptions) then I think we should rather sit still in the boat and stick to the existing approach.

Thanks for your help!

carmenbianca:
You could keep the entry in `REUSE.toml`. Because `precedence` is set to `closest`, it will take the information from the REUSE.toml file for the files that are detected as being binary. Instead of a wildcard for all test files, you could specifically flag the ones that are detected as binary files, but that doesn't seem ideal.

Using REUSE.toml is perfectly acceptable for test files, though. It's not a bad approach.

bagder:
> You could keep the entry in REUSE.toml

I understand. I just rather it be consistent per file type. This is a file type that can have the information embedded and we have almost two thousand files like this. I think it gets weird to have an exception that seemingly randomly a few of them need a wildcard in REUSE.toml. That's why I then rather keep the wildcard.

(It is however a strange restriction in the reuse tool and it is weird that it does not properly convey this reason in its report.)

carmenbianca:
It's a tricky problem… If you try to get REUSE information out of binary files, performance is affected, and you may get false positives and garbage data. If you don't search binary files, you get these false negatives.

Including a reason in the report is also tricky. It'd need a great deal of refactoring first, but that isn't the real problem. The real problem is that this is one of many many many corner cases that may require a useful tip to the user. It'd be a great undertaking to engineer the code to be able to spot all these corner cases, and a similar undertaking to present such information to the user without spamming them.

'Output more helpful data' is something that is slowly being worked on, though. `--debug` helps a little. `--json` also helps a bunch, but wouldn't have helped here.

PR #14803, opened 2024-09-06
cmake: fix to show features/protocols with `CURL_DISABLE_INSTALL=ON`

vszakats:
Follow-up to aace27b0965c10394544d1dacc9c2cb2fe0de3d3 #12287

---

w/o whitespace: https://github.com/curl/curl/pull/14803/files?w=1


vszakats:
On the other hand we ship `mk-ca-bundle.1`. I'm dropping the patch touching the `mk-ca-bundle.pl` bits.

PR #14809, opened 2024-09-06
schannel: don't say "failed to decrypt" when returning EAGAIN

### LINKED ISSUE #14807: schannel: failed to decrypt data, need more data  error in window11 with libcurl
### I did this

when try to download a file using curl through proxy and below command:
curl -fvk --proxy 120.0.0.1:3501 -o Test.bin <URL_to_download> -v -k -l 
 download is successful but : "schannel: failed to decrypt data, need more data " error is coming
 
TLS handshake was successful

i tried running with --tls-max 1.2 flag and --ssl-no-revoke flag but still the issue is coming

its happening only when i tried to download the file from windows.

but how to confirm that this is issue with windows and TLS compatability. 

### I expected the following

expected to completed the download without errors

### curl/libcurl version

curl/8.2.1

### operating system

window11

bagder:
It is misleading and wrong.

Reported-by: Annpurna
Fixes #14807

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -2327,7 +2327,8 @@ schannel_recv(struct Curl_cfilter *cf, struct Curl_easy *data,
       backend->encdata_is_incomplete = true;
       if(!*err)
         *err = CURLE_AGAIN;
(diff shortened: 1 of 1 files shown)


bagder:
#14810 is a better take, closing

PR #14810, opened 2024-09-06
schannel, rework info messages and tracing

icing:
Turn most infof() and DEBUGF() traces into CURL_TRC_CF() statements that do not appear on normal verbose output.

refs #14807

### DIFF
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -131,11 +131,6 @@
 #define CERT_THUMBPRINT_STR_LEN 40
 #define CERT_THUMBPRINT_DATA_LEN 20
 
-/* Uncomment to force verbose output
- * #define infof(x, y, ...) printf(y, __VA_ARGS__)
- * #define failf(x, y, ...) printf(y, __VA_ARGS__)
- */
-
 #ifndef CALG_SHA_256
 #  define CALG_SHA_256 0x0000800c
 #endif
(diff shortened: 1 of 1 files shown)


lib/vtls/schannel.c
```diff
@@ -2318,40 +2296,38 @@ schannel_recv(struct Curl_cfilter *cf, struct Curl_easy *data,
...
+                    "server close notification received (close_notify)");
         goto cleanup;
       }
     }
     else if(sspi_status == SEC_E_INCOMPLETE_MESSAGE) {
       backend->encdata_is_incomplete = true;
       if(!*err)
         *err = CURLE_AGAIN;
-      infof(data, "schannel: failed to decrypt data, need more data");
+      CURL_TRC_CF(data, cf, "incomplete message, EAGAIN");
```
bagder, line 2308: The condition above indicates that this might happen for errors other than EAGAIN, and then this message gets misleading. How about now showing the message at all for the EAGAIN case since that looks like a more "normal" code path and not really something to mention in the verbose.

icing: Will remove it.

bagder (approved):

jay:
I disagree. Most of these are normal for verbose output such as renegotiation failed or disabled revocation checks. The issue with tracing is it outputs a lot of data in the curl tool, even if you only select ssl, like I just tried tracing a 200MB file with schannel ssl tracing and it dumped a slew of data on the screen It makes it harder for me to identify what is actually going on with schannel as opposed to verbose.

I agree some are appropriate for trace output like the encrypted/decrypted data messages, or connect stage messages would work well there. example (taken from this pr)

12:39:39.333003 [0-0] == Info: [SSL] encrypted data length: 3668
12:39:39.335864 [0-0] == Info: [SSL] encrypted cached: offset 3668 length 103424
12:39:39.339958 [0-0] == Info: [SSL] decrypted data length: 1383
12:39:39.342663 [0-0] == Info: [SSL] decrypted data added: 1383
12:39:39.345412 [0-0] == Info: [SSL] decrypted cached: offset 2766 length 102400

The "failed to decrypt data, need more data" in that other issue is correct. It failed to decrypt data because the message is incomplete. I'd prefer to keep that in trace. 

I am going to propose an alternate pr for this dialed back some

icing:
Please do. I tried to filter a bit between fail/informative/debugging classifications, but you have more experience here.

bagder:
> The "failed to decrypt data, need more data" in that other issue is correct. It failed to decrypt data because the message is incomplete. I'd prefer to keep that in trace.

Why? We don't log every EAGAIN situation internally since it is a quite expected condition and it makes it sound like it is a problem - no to mention that it can get repeated quite a number of times without it being any problem. I recommend you set it in within a DEBUGF() or something.

jay:
Rather than classify the extra-verbose messages as trace I separated them into a separate macro that can be enabled at build time. This way they do not add a lot of noise to debug builds unless built for it, since the messages are useful only when debugging schannel_recv. Also I changed the eagain messages to extra verbose. #14826

icing:
Replaced by #14826

PR #14812, opened 2024-09-06
build: drop unused `NROFF` reference

vszakats:


bagder:
`src/Makefile.mk` seems to have some NROFF leftovers as well

vszakats:
> `src/Makefile.mk` seems to have some NROFF leftovers as well

Yep, seen them! I'll open a PR, lightly tested.

I wonder if anyone is using `Makefile.mk`? It's for MS-DOS and Amiga, are those covered by `./.configure` (or CMake)?

bagder:
> I wonder if anyone is using Makefile.mk? It's for MS-DOS and Amiga, are those covered by ./.configure (or CMake)?

We don't have any CI jobs that use them. We have no real way of knowing if anyone uses them.

vszakats:
Updated `Makefile.mk` in #14813.

PR #14815, opened 2024-09-06
autotools: add support for 'unity' builds, enable in CI

vszakats:
Implement the "unity" builds as known from CMake, but for autotools.
It's limited to `lib` and `src` (CMake also supports it in `tests`).

Enable with: `--enable-unity` (disabled by default)

Unity builds speed up builds significantly. Cygwin and Windows builds in
particular, but the effect is noticeable on most systems. It also allows
discovering unity issues with autotools, benefitting also CMake when
building the same combination. In CI it makes turnaround times quicker.

This closes build performance with CMake. autotools still lags behind
because it builds shared and static libcurl in two, separate passes.
CMake does it in one. Manpage compilation isn't batched, it is in CMake.
After unity and test bundle support the slowest parts of the build are
the configuration phase (which is effectively a tedious, non-parallel,
compilation and/or linking of 300+ tiny programs. The next bottleneck
is compiling individual examples and finally test servers (only slow
with autotools).

The autotools implementation is slightly less efficient than CMake,
because 3 sources are permanently excluded while in CMake this isn't
necessary and solved more efficiently while building libtests. There is
also no 'unity' support for tests, making them a less efficient also.

Enable it in CI for most `configure` jobs. Except in GHA/dist (though
it works fine there too), to use the default config there. Also skip for
the Linux AWC-LC job where it made builds time a few seconds longer
(reason undiscovered.)

Autotools test suite builds compared between master → `--enable-unity`:
- GHA/Linux: 32s → 12s
  https://github.com/curl/curl/actions/runs/10705668823/job/29681617374
  https://github.com/curl/curl/actions/runs/10742978889/job/29796766297
- GHA/macOS: 37s → 10s
  https://github.com/curl/curl/actions/runs/10705668813/job/29681632885
  https://github.com/curl/curl/actions/runs/10742978699/job/29796768875
- GHA/FreeBSD: 15m25 → 10m58 (full workflow time, ~qemu)
  https://github.com/curl/curl/actions/runs/10705668811/job/29681607915
  https://github.com/curl/curl/actions/runs/10742978937/job/29796766115
- GHA/Cygwin: 3m32 → 1m21
  https://github.com/curl/curl/actions/runs/10705668809/job/29681609965
  https://github.com/curl/curl/actions/runs/10742978645/job/29796756933
- GHA/MSYS2: 2m42 → 50s
  https://github.com/curl/curl/actions/runs/10705668808/job/29681621166
  https://github.com/curl/curl/actions/runs/10742978662/job/29799739289
- GHA/mingw-w64: 5m32 → 1m23
  https://github.com/curl/curl/actions/runs/10705668808/job/29681628787
  https://github.com/curl/curl/actions/runs/10742978662/job/29799741568
- Azure MSYS2 mingw64 openssl: 38m55 → 19m30 (all build)
  https://dev.azure.com/daniel0244/curl/_build/results?buildId=25546&view=logs&j=b58b8c59-0f61-52e9-0f9e-fad562a1e77f&t=0f9230a7-3b10-53ca-9938-700ece377c5e
  https://dev.azure.com/daniel0244/curl/_build/results?buildId=25521&view=logs&jobId=81bb90a1-709c-5261-a92d-4869708fd1b6&j=b58b8c59-0f61-52e9-0f9e-fad562a1e77f&t=0f9230a7-3b10-53ca-9938-700ece377c5e
- Azure Ubuntu default: 2m15 → 1m36 (all build)
  https://dev.azure.com/daniel0244/curl/_build/results?buildId=25546&view=logs&j=9d58b9ac-e1e6-53b6-f83a-1f9f1d912522&t=a6b38d83-e7cf-5a9b-c762-a178412717b7
  https://dev.azure.com/daniel0244/curl/_build/results?buildId=25521&view=logs&jobId=81bb90a1-709c-5261-a92d-4869708fd1b6&j=9d58b9ac-e1e6-53b6-f83a-1f9f1d912522&t=a6b38d83-e7cf-5a9b-c762-a178412717b7

---

- [x] rebase on #14932.
- [x] rebase on #14933.
- [x] disable by default before merge.
- [x] enable in CI jobs.
- [x] re-test without test bundles and fix issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ commands:
       - run:
           command: |
             autoreconf -fi
-            ./configure --disable-dependency-tracking --enable-warnings --enable-werror --with-openssl \
+            ./configure --disable-dependency-tracking --enable-unity --enable-warnings --enable-werror --with-openssl \
               || { tail -1000 config.log; false; }
 
   configure-openssl-no-verbose:
@@ -41,7 +41,7 @@ commands:
       - run:
           command: |
             autoreconf -fi
-            ./configure --disable-dependency-tracking --disable-verbose --enable-werror --with-openssl \
+            ./configure --disable-dependency-tracking --enable-unity --disable-verbose --enable-werror --with-openssl \
               || { tail -1000 config.log; false; }
 
   configure-no-proxy:
@@ -49,7 +49,7 @@ commands:
       - run:
           command: |
             autoreconf -fi
-            ./configure --disable-dependency-tracking --disable-proxy --enable-werror --with-openssl \
+            ./configure --disable-dependency-tracking --enable-unity --disable-proxy --enable-werror --with-openssl \
               || { tail -1000 config.log; false; }
 
   install-cares:
@@ -110,7 +110,7 @@ commands:
       - run:
           command: |
             autoreconf -fi
-            ./configure --disable-dependency-tracking --enable-warnings --enable-werror --with-openssl --enable-ares \
+            ./configure --disable-dependency-tracking --enable-unity --enable-warnings --enable-werror --with-openssl --enable-ares \
               || { tail -1000 config.log; false; }
 
   configure-wolfssh:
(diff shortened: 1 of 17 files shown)


vszakats:
This new "unity" option for autotools is disabled by default, except if the env `CURL_CI` is
set, which enables it by default. This env already had some use in GHA, this PR sets it for
most workflows globally. In a few workflows it's enabled with `--enable-unity` and in two
workflows: distro and AWC-LC it's intentionally not enabled.

It does work fine in distro, but it's probably best to use the default there.

AWC-LC is a strange case where the build is already faster than elsewhere for some reason,
and unity makes the build a few seconds slower. (Perhaps something obvious, but I don't
know why this is, and I could not see similar result in other jobs.)

This PR is ready for merge.

vszakats:
Well, the `CURL_CI` trick didn't save much editing, while it reduced
transparency and local reproducibility, so went back to just enabling
`--enable-unity` manually where needed. There is no new "special
case" now, `--enable-unity` is just disabled by default, always.

PR #14820, opened 2024-09-07
maketgz: fix RELEASE-TOOLS.md for daily tarballs

dfandrich:
The daily snapshots have no associated git tag, so provide a commit has
instead in these cases. Fix the dpkg detection since the shell would
exit immediately without showing an error message if it weren't found.
Also, use "which" instead of "command" since the latter is bash-specific
and this script specifies a plain Bourne shell.

Closes #14820

### DIFF
--- a/scripts/maketgz
+++ b/scripts/maketgz
@@ -156,13 +156,13 @@ else
   automake --include-deps Makefile >/dev/null
 fi
 
-echo "produce RELEASE-TOOLS.md"
-./scripts/release-tools.sh "$timestamp" "$version" > docs/RELEASE-TOOLS.md.dist
-
 if test -n "$commit"; then
   echo "produce docs/tarball-commit.txt"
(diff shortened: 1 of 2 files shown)


dfandrich:
shellcheck tells me `command` is more portable than `which`, so I switched that back.

dfandrich:
Analysis of PR #14820 at [70efbcba](https://github.com/curl/curl/commit/70efbcba906152ce40adecb3bd2cbd343b27892d):

[Test 721 failed](https://github.com/curl/curl/actions/runs/10747864013/job/29810846754#check-step-15), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 2055 failed](https://github.com/curl/curl/actions/runs/10747864013/job/29810846754#check-step-15), which has NOT been flaky recently, so **there could be a real issue in the PR**.

bagder (approved):

PR #14821, opened 2024-09-07
dist: drop `.in` files from `EXTRA_DIST`

vszakats:
Some of the `.in` files were listed in `EXTRA_DIST`. Delete them.

`.in` files (passed to `AC_CONFIG_FILES`) are added automatically
to the distro by autotools.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -82,9 +82,9 @@ PLAN9_DIST = plan9/include/mkfile \
  plan9/src/mkfile.inc             \
  plan9/src/mkfile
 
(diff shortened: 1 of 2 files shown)


Kangie:
LGTM; seems reasonable.

vszakats:
[ Unrelated to this particular PR, but here's an example for a hung GHA job that's not obeying `timeout-minutes=14` set for the test step: https://github.com/curl/curl/actions/runs/10773165711/job/29872633099?pr=14821#step:13:5063 It also ends up being greyed-crossed at the end, and hiding the log already output. ]

PR #14822, opened 2024-09-07
build: `buildinfo.txt` improvements

vszakats:
- cmake: drop `configure.os`.
  This also includes OS version, but thus far it's not important enough
  to include it.
- autotools: drop redundant, autotools-only `{target|host}.vendor`.
  (it's part of the triplet in `{target|host}`.)
- swap order to `*.cpu` → `*.os` to match triplet-order.
- cmake: drop redundant `target`.
  It's manually filled and only in a (so far) few CI jobs. Let's revisit
  when this becomes useful.
- move `buildinfo.txt` to build root.
- dist: add `buildinfo.txt` to `DISTCLEANFILES`.
- autotools: detect human readable compiler version.
- autotools: replace `XXYY` `compiler.version` with "X.Y"-style.
  (also to match cmake.)
- autotools: use distinct `compiler_id` for Apple clang: `APPLECLANG`.
  To match cmake and also because the the "X.Y"-style version number
  is the Apple version, while `XXYY` was a value roughly translated to
  mainline llvm/clang version.
- show buildinfo at the end of the configure stage, when run in CI, or
  when `CURL_BUILDINFO` or `CURL_CI` env is set.

Follow-up to 1fdea1684602a1ae2870c67b5f3e8fd34f63da95 #14802
Assisted-by: Dan Fandrich
Ref: https://github.com/curl/curl/pull/14802#issuecomment-2334942991

---

/cc @dfandrich


dfandrich (commented):
From a Test Clutch perspective, this looks fine—it's not using any of the fields that have been removed. I'm thinking now that maybe it makes sense to include {host,target}.{os,cpu} in the autoconf output, even if it's redundant, since those two fields are both common to AC and CM and there easier for others to use.

Also, please coordinate with me before submitting this since the change of semantics in compiler.version in AC will need some work in Test Clutch to avoid storing inconsistent data.

dfandrich:
Actually, Test Clutch now automatically figures out which format of compiler version it is, so submit at your leisure.

vszakats:
Thanks Dan! I'm also fine including the fined grained cpu/os fields always. Makes parsing a little bit easier.

I'm thinking to make `buildtest.txt` visible at the end of the configure process, maybe on condition of running in CI. It'd help seeing this info early in the build, also in jobs without tests. This in turn helps seeing if detection is working as expected, and helps navigating codepaths in both configure and CMake. It could be put under a GHA fold if seen too verbose. It is partially redundant, but has the advantage of having all this info in one block, and regardless of build tool. Or is that overkill?

dfandrich:
That's probably a good idea, especially on CI where you can't poke around as easily to figure out the environment.

vszakats:
Added a `buildinfo.sysroot` field. It's not perfect because configure relies on the caller setting this (as in GHA/macos), but cmake does detect it on macOS. It shows the Apple SDK version, and sometimes Xcode version. Probably not important enough to include in the testclutch matrix, but can be a useful datapoint for builds in general.

PR #14826, opened 2024-09-08
schannel: reclassify extra-verbose schannel_recv messages

jay:
- Create a new macro SCH_DEV() to manage verbose debug messages that are only useful for debugging Schannel recv decryption.

schannel_recv contains a lot of useful debug messages to help debug the function, however in practice they are not otherwise useful and showing them in debug builds adds a lot of noise.

To show these messages curl must now be built with CURL_SCHANNEL_DEV_DEBUG defined.

Prior to this change many, but not all, extra-verbose messages were wrapped in DEBUGF() so they were only shown in debug builds.

Ref: https://github.com/curl/curl/issues/14807

Closes #xxxxx

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -201,6 +201,7 @@ jobs:
           - { build: 'cmake'    , sys: 'ucrt64' , env: 'ucrt-x86_64' , tflags: 'skiprun'                 , config: '-DENABLE_DEBUG=OFF -DBUILD_SHARED_LIBS=ON  -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=ON -DENABLE_CURLDEBUG=ON', type: 'Release', name: 'schannel R TrackMemory' }
           - { build: 'cmake'    , sys: 'clang64', env: 'clang-x86_64', tflags: 'skiprun'                 , config: '-DENABLE_DEBUG=ON  -DBUILD_SHARED_LIBS=OFF -DCURL_USE_OPENSSL=ON  -DENABLE_UNICODE=OFF', type: 'Release', name: 'openssl' }
           - { build: 'cmake'    , sys: 'mingw64', env: 'x86_64'      , tflags: 'skiprun'                 , config: '-DENABLE_DEBUG=OFF -DBUILD_SHARED_LIBS=ON  -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=ON', type: 'Release', test: 'uwp', name: 'schannel R' }
+          - { build: 'cmake'    , sys: 'mingw64', env: 'x86_64'      , tflags: 'skiprun'                 , config: '-DENABLE_DEBUG=ON  -DBUILD_SHARED_LIBS=ON  -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=ON -DCMAKE_VERBOSE_MAKEFILE=ON', type: 'Debug', cflags: '-DCURL_SCHANNEL_DEV_DEBUG', name: 'schannel dev debug' }
           - { build: 'cmake'    , sys: 'mingw32', env: 'i686'        , tflags: 'skiprun'                 , config: '-DENABLE_DEBUG=OFF -DBUILD_SHARED_LIBS=ON  -DCURL_USE_SCHANNEL=ON -DENABLE_UNICODE=ON', type: 'Release', name: 'schannel R' }
       fail-fast: false
     steps:
@@ -334,7 +335,7 @@ jobs:
           [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
           [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
           cmake -B bld -G Ninja ${options} \
-            "-DCMAKE_C_FLAGS=${cflags}" \
+            "-DCMAKE_C_FLAGS=${{ matrix.cflags }} ${cflags}" \
             "-DCMAKE_RC_COMPILE_OBJECT=${rcopts}" \
             '-DCMAKE_BUILD_TYPE=${{ matrix.type }}' \
             -DCMAKE_UNITY_BUILD=ON -DCURL_TEST_BUNDLES=ON \
--- a/lib/vtls/schannel.c
+++ b/lib/vtls/schannel.c
@@ -59,6 +59,17 @@
 #include "curl_memory.h"
 #include "memdebug.h"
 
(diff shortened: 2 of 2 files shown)


dfandrich:
Analysis of PR #14826 at [03d9f837](https://github.com/curl/curl/commit/03d9f8379caee559bb7b2e2bd1ab7829710f0f1d):

[Test 1540 failed](https://github.com/curl/curl/actions/runs/10757847053/job/29832379664#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 677 failed](https://github.com/curl/curl/actions/runs/10757847080/job/29832382383#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

jay:
> [Test 1540 failed](https://github.com/curl/curl/actions/runs/10757847053/job/29832379664#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

This test failed in the Omni OS ci job and I don't see anything unexpected in the output except the return code is strange `lib1540 returned 2006, when expecting 0`.  As far as I can't be returned (see [lib1540.c](https://github.com/curl/curl/blob/03d9f8379caee559bb7b2e2bd1ab7829710f0f1d/tests/libtest/lib1540.c#L85-L118) + [first.c](https://github.com/curl/curl/blob/03d9f8379caee559bb7b2e2bd1ab7829710f0f1d/tests/libtest/first.c#L178-L189)) unless curl_easy_perform returns it so I think more likely there must be something strange happening in runtests

> [Test 677 failed](https://github.com/curl/curl/actions/runs/10757847080/job/29832382383#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

another unknown error code `lib677.exe returned 143, when expecting 0` despite expected output, this one in CI job `msvc, CM x64-windows libressl`

jay:
@vszakats do you have advice for how I should append a definition to cmake CFLAGS for a CI job, I tried `-DCMAKE_C_FLAGS_INIT:STRING=-DCURL_SCHANNEL_DEV_DEBUG` but CMAKE_C_FLAGS_INIT was overridden by a CMAKE_C_FLAGS which is used for all jobs.

https://github.com/curl/curl/blob/12b269fe597992703ac3c0c1284ca129ccbf9366/.github/workflows/windows.yml#L176

So then I tried `-DCMAKE_C_FLAGS:STRING=-DCURL_SCHANNEL_DEV_DEBUG` but it seems cmake doesn't work that way (it only uses the last CMAKE_C_FLAGS as far as I can tell), which means it will add `-DCURL_SCHANNEL_DEV_DEBUG` overriding the CMAKE_C_FLAGS used for all jobs (shown above with `-DTESTING_CI_BEHAVIOR` which is what is overridden in this experiment)

vszakats:
> @vszakats do you have advice for how I should append a definition to cmake CFLAGS for a CI job, I tried `-DCMAKE_C_FLAGS_INIT:STRING=-DCURL_SCHANNEL_DEV_DEBUG` but CMAKE_C_FLAGS_INIT was overridden by a CMAKE_C_FLAGS which is used for all jobs.
> 
> https://github.com/curl/curl/blob/12b269fe597992703ac3c0c1284ca129ccbf9366/.github/workflows/windows.yml#L176
> 
> So then I tried `-DCMAKE_C_FLAGS:STRING=-DCURL_SCHANNEL_DEV_DEBUG` but it seems cmake doesn't work that way (it only uses the last CMAKE_C_FLAGS as far as I can tell), which means it will add `-DCURL_SCHANNEL_DEV_DEBUG` overriding the CMAKE_C_FLAGS used for all jobs (shown above with `-DTESTING_CI_BEHAVIOR` which is what is overridden in this experiment)
> 
> https://github.com/curl/curl/blob/12b269fe597992703ac3c0c1284ca129ccbf9366/.github/workflows/windows.yml#L62

Yes, only the last one counts. Would this help?:
```diff
diff --git a/.github/workflows/windows.yml b/.github/workflows/windows.yml
index 9ce3c7650..500e671f2 100644
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -171,6 +171,7 @@ jobs:
           fi
           [ '${{ matrix.type }}' = 'Debug' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG='
           [ '${{ matrix.type }}' = 'Release' ] && options+=' -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE='
+          cflags+=' -DCURL_SCHANNEL_DEV_DEBUG`
           cmake -B bld ${options} \
             "-DCMAKE_C_FLAGS=${cflags}" \
             "-DCMAKE_RC_COMPILE_OBJECT=${rcopts}" \
@@ -389,6 +390,7 @@ jobs:
         if: ${{ matrix.build == 'cmake' }}
         run: |
           cmake -B bld \
+            -DCMAKE_C_FLAGS=-DCURL_SCHANNEL_DEV_DEBUG \
             -DCMAKE_SYSTEM_NAME=Windows \
             -DCMAKE_C_COMPILER_TARGET=${TRIPLET} \
             -DCMAKE_C_COMPILER=${TRIPLET}-gcc \
@@ -502,6 +504,7 @@ jobs:
             options+=" -DOPENSSL_ROOT_DIR=$VCPKG_INSTALLATION_ROOT/installed/${{ matrix.arch }}-${{ matrix.plat }}"
           fi
           cmake -B bld ${options} \
+            -DCMAKE_C_FLAGS=-DCURL_SCHANNEL_DEV_DEBUG \
             "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake" \
             "-DVCPKG_INSTALLED_DIR=$VCPKG_INSTALLATION_ROOT/installed" \
             '-DVCPKG_TARGET_TRIPLET=${{ matrix.arch }}-${{ matrix.plat }}' \
```

jay:
@vszakats i want the define only for job 'schannel dev debug' so i changed the 'cmake configure' step to add the cflags from that particular job like this

~~~diff
-            "-DCMAKE_C_FLAGS=${cflags}" \
+            "-DCMAKE_C_FLAGS=${{ matrix.cflags }} ${cflags}" \
~~~

seems to work ok let me know if you see any problem with it

vszakats:
It looks fine to me.

Maybe moving the new job one line up, and moving the ENABLE_UNICODE option before the added cmake option would make the new job line up better with existing ones, but it's just a formatting nit.

bagder:
@jay where are we on this?

jay:
Some Windows CI failures I can't explain and probably unrelated. I rebased on master and force pushed. Assuming @icing is OK with it then it can replace #14810 and I'll land it

jay:
@icing

icing (approved):
Sorry, missed the mention during covid. Looks good to me, thanks!

PR #14827, opened 2024-09-08
ipfs: add options to disable

### LINKED ISSUE #14824: IPFS should be under enable- / disable- control
### I did this

Build 8.9.1 from source.

### I expected the following

Expect all protocols to be under enable/disable control.  IPFS is the ONLY protocol not under option control.

### curl/libcurl version

curl 8.4.0 and on.  Currently a problem in 8.9.1

### operating system

Debian testing (13+)

vszakats:
- CPPFLAGS: `-DCURL_DISABLE_IPFS`
- configure: `--disable-ipfs`
- cmake: `-DCURL_DISABLE_IPFS=ON`

Fixes #14824


docs/CURL-DISABLE.md
```diff
@@ -140,6 +140,10 @@ Disable the built-in progress meter
 
 Disable support for proxies
 
+## `CURL_DISABLE_IPFS`
+
+Disable the IPFS/IPNS protocols.
```
bagder, line 145: I think it is worth mentioning that this affects the curl tool only, as libcurl does not know anything about IPFS.

PR #14828, opened 2024-09-08
cmake: disable default OpenSSL if BearSSL, GnuTLS or Rustls is enabled

vszakats:
Disable OpenSSL by default if any of these alterntive TLS backends were
explicitly selected.

Following the logic already in place for Schannel, Secure Transport,
mbedTLS and wolfSSL.


bagder:
What about multissl builds?

vszakats:
This solely affects whether to enable OpenSSL _by default_,
when an alternate backend is explicitly selected.

For MultiSSL builds, OpenSSL needs to be explicitly enabled.

This was the case for most TLS backends. Some were missed
from the codition controlling this. This patch syncs them up.

MarcelRaad (approved):

PR #14830, opened 2024-09-09
lib: remove the final strncpy() calls

bagder:
wolfssl: use strcpy() as the target buffer is > 40 bytes

gethostname: return failure if the buffer is too small instead

### DIFF
--- a/lib/.checksrc
+++ b/lib/.checksrc
@@ -1,1 +1,2 @@
 enable STRERROR
+enable STRNCPY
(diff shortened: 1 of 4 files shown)


dfandrich:
Analysis of PR #14830 at [a03bfd17](https://github.com/curl/curl/commit/a03bfd1740db51b8765dab75d722ef7f7784ef24):

[Test 142 failed](https://circleci.com/gh/curl/curl/106404), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 353 failed](https://circleci.com/gh/curl/curl/106404), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14835, opened 2024-09-09
tests, tweak server checks, lock file handling and timers

icing:
- when server are killed by a test case, do not wait for the server lock file to go away. These tests are mostly about client timeouts and the server will hang until killed.
- when killing a server successfully, check for a remaining lock file, log its existence and remove it.
- lower the delay timings on SLOWDOWN by half
- add SLOWDOWNDATA server command to only slow down the FTP data bytes, not the control ones.
- lower some timeout values

In addition, with Windows failures in test445, make `sws.c`'s CONNECT handling non-blocking with a 1 second timeout.

Update: in addition, now always run the verifyserver checks before a testrun. This was avoided on non-torture runs and in https cases and I saw several Windows CI failures that just could not connect to the server.


tests/data/test1086
```diff
@@ -17,7 +17,7 @@ timing-dependent
 # Overload some standard FTP responses to make them shorter and faster
```
dfandrich, line 17: Now that you have added SLOWDOWN, this comment and the REPLY lines below that implement it can be removed to simplify the test case.

tests/data/test1112
```diff
@@ -17,7 +17,7 @@ timing-dependent
 # Overload some standard FTP responses to make them shorter and faster
```
dfandrich, line 17: Same here.

tests/data/test1117
```diff
@@ -32,7 +32,7 @@ partial body
 </data1>
 
 <servercmd>
-writedelay: 1000
+writedelay: 100
```
dfandrich, line 35: 100 msec might not be enough to guarantee a context switch to the client in that time on a really loaded system. I believe this delay will only happen a single time so it needs to be long enough to make a difference.

tests/data/test303
```diff
@@ -37,7 +37,7 @@ https
 HTTPS with 8 secs timeout
 </name>
 <command>
--k https://%HOSTIP:%HTTPSPORT/want/%TESTNUMBER -m 8
+-k https://%HOSTIP:%HTTPSPORT/want/%TESTNUMBER -m 3
```
dfandrich, line 40: This was last increased (to 8 sec) in commit e2fc294 with the comment *added an additional second to allow test to pass on heavily loaded servers*. Are you sure it's safe to drop it this low? Whatever it is, the name on line 37 needs to be changed to match.

tests/ftpserver.pl
```diff
@@ -2867,9 +2868,14 @@ sub customize {
             logmsg "FTPD: read POSTFETCH header data\n";
             $postfetch = $1;
         }
+        elsif($_ =~ /SLOWDOWNDATA/) {
+            $ctrldelay=0;
+            $datadelay=0.005;
+            logmsg "FTPD: send response data with 0.01 sec delay per byte\n";
```
dfandrich, line 2874: This message doesn't match the actual delay.

tests/ftpserver.pl
```diff
@@ -2867,9 +2868,14 @@ sub customize {
...
+        }
         elsif($_ =~ /SLOWDOWN/) {
-            $ctrldelay=1;
-            $datadelay=1;
+            $ctrldelay=0.005;
+            $datadelay=0.005;
             logmsg "FTPD: send response with 0.01 sec delay between each byte\n";
```
dfandrich, line 2879: This message doesn't match the actual delay.

tests/runner.pm
```diff
@@ -1066,7 +1066,10 @@ sub singletest_clean {
         }
     }
 
-    waitlockunlock($serverlogslocktimeout);
+    my @killtestservers = getpart("client", "killserver");
+    if(!@killtestservers) { # we are not killing it anyway
```
dfandrich, line 1070: What about the case where a test uses multiple servers but is only killing one?

tests/runner.pm
```diff
@@ -1091,14 +1094,19 @@ sub singletest_clean {
...
 
-    my @killtestservers = getpart("client", "killserver");
     if(@killtestservers) {
         foreach my $server (@killtestservers) {
             chomp $server;
             if(stopserver($server)) {
                 logmsg " $testnum: killserver FAILED\n";
                 return 1; # normal error if asked to fail on unexpected alive
             }
+            else {
```
dfandrich, line 1104: "else" isn't needed because the previous case returns.

tests/server/sws.c
```diff
@@ -1319,6 +1319,14 @@ static curl_socket_t connect_to(const char *ipaddr, unsigned short port)
   }
 #endif
 
+  if(0 != curlx_nonblock(serverfd, TRUE)) {
```
dfandrich, line 1322: Please add a comment why this is being set to nonblocking. It's not obvious and it would be easy for someone to see the select() loop below and just switch the connect back to blocking.

tests/server/sws.c
```diff
@@ -1360,12 +1368,40 @@ static curl_socket_t connect_to(const char *ipaddr, unsigned short port)
...
+        rc = select((int)serverfd + 1, NULL, &output, NULL, &timeout);
+        if(rc < 0 && SOCKERRNO != EINTR)
+          goto error;
+        else if(rc > 0) {
+          curl_socklen_t errSize = sizeof(error);
+          if(0 != getsockopt(serverfd, SOL_SOCKET, SO_ERROR,
+                             (void *)&error, &errSize))
+            error = SOCKERRNO;
+          if((0 == error) || (EISCONN == error))
+            goto success;
```
dfandrich, line 1387: The socket is left in nonblocking mode. Will there be any knock-on effects in the rest of the code, which assumed blocking mode?

tests/data/test1086
icing: will do

tests/data/test1112
icing: will do

tests/ftpserver.pl
icing: will adjust

tests/ftpserver.pl
icing: will adjust

tests/server/sws.c
icing: will do.

tests/server/sws.c
icing: Good point. I will add code setting it to blocking again after the successful connect.

tests/data/test1117
icing: I'll revert this again. The saving does not seem to justify the loss of reliability.

tests/data/test303
icing: Thanks for the info. I'll revert this.

tests/runner.pm
icing: There is indeed a single test case that does that 1561 (and there might be future ones).

I'll submit a change in ordering here:
* execute the `postcommanddelay` first
* kill all specified servers, removing their lock files when necessary
* wait for any remaining servers to release their locks
* record the test timing

bagder:
When I test for example 303 on my system, I see that it always wait the five seconds when killing the servers in the end which seems like a terrible waste. Since the servers *can* handle being terminated anyway, I suggest we shorten the wait from 5 seconds to .5 seconds =>

~~~diff
diff --git a/tests/processhelp.pm b/tests/processhelp.pm
index 48f385040..99402b520 100644
--- a/tests/processhelp.pm
+++ b/tests/processhelp.pm
@@ -309,14 +309,14 @@ sub killpid {
                 }
             }
         }
     }
 
-    # Allow all signalled processes five seconds to gracefully die.
+    # Allow all signalled processes half a second to gracefully die.
     if(@signalled) {
-        my $twentieths = 5 * 20;
-        while($twentieths--) {
+        my $loops = 10;
+        while($loops--) {
             for(my $i = scalar(@signalled) - 1; $i >= 0; $i--) {
                 my $pid = $signalled[$i];
                 if(!pidexists($pid)) {
                     print("RUN: Process with pid $pid gracefully died\n")
                         if($verbose);

icing:
> When I test for example 303 on my system, I see that it always waits the five seconds when killing the servers in the end which seems like a terrible waste. Since the servers _can_ handle being terminated anyway, I suggest we shorten the wait from 5 seconds to .5 seconds =>

No need, really. With my last commit here, we reap the zombies that cause the wait.

bagder:
Ah right, that really does the trick.

Test 303 goes from 20 seconds to 11 for me with this PR.

charles2910:
Quick feedback from Debian here:

We were going to investigate this flakiness on the FTP tests this weekend (it was really showing up for us all the time), but we saw this MR and I gave a try and apparently it solved the problems for us (so thanks @icing!) - both gitlab CI and locally. Tomorrow we will upload the new patch version and we'll get back to you guys with the final answer about the tests on Debian's infrastructure.

PR #14840, opened 2024-09-09
rustls: fixed minor logic bug in default cipher selection

jan2000:
Follow-up to 1e03d4bc

### DIFF
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -508,7 +508,7 @@ add_ciphers:
       struct rustls_str s;
       entry = rustls_default_ciphersuites_get_entry(j);
       s = rustls_supported_ciphersuite_get_name(entry);
-      if(s.len < 5 || strncmp(s.data, "TLS13", 5) == 0)
+      if(s.len >= 5 && strncmp(s.data, "TLS13", 5) == 0)
         continue;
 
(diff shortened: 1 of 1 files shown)


ctz (commented):
lgtm. i would be minded to to extract some small helper function like `bool suite_is_tls13(const struct rustls_supported_ciphersuite *cs)` to clarify this and  the check above on L453

jan2000:
If rustls provided such a function... Otherwise I do not think abstracting away a simple string comparison that is only done twice in a single function is beneficial. The intent is to keep the code as short, simple and clear as possible.

Right now it is directly visible in the `cr_get_selected_ciphers()` function that there is no authoritative method of rustls used to determine if a cipher is a TLS 1.3 cipher, but that it is just done by simple string comparison.

cpu:
> there is no authoritative method of rustls used to determine if a cipher is a TLS 1.3 cipher,

I think it makes sense to add a `rustls-ffi` binding for [SupportedCipherSuite.version](https://docs.rs/rustls/latest/rustls/enum.SupportedCipherSuite.html#method.version) that could meet this need without interrogating the friendly name. I'll take a look at getting that into our next release.

bagder (approved):

bagder:
Thanks!

PR #14841, opened 2024-09-10
src: tidy-up conditions for CA bundle search

vszakats:
- delete redundant Schannel check.

- move `feature_ssl` check one level up from `FindWin32CACert()`.

- check `feature_ssl` early to skip a bunch of CA bundle search logic
  for no-ssl configurations.

---

w/o whitespace: https://github.com/curl/curl/pull/14841/files?w=1

### DIFF
--- a/src/tool_doswin.c
+++ b/src/tool_doswin.c
@@ -624,25 +624,21 @@ CURLcode FindWin32CACert(struct OperationConfig *config,
   (void)config;
   (void)bundle_file;
 #else
-  /* Search and set cert file only if libcurl supports SSL. */
-  if(feature_ssl) {
+  DWORD res_len;
+  TCHAR buf[PATH_MAX];
+  TCHAR *ptr = NULL;
 
-    DWORD res_len;
-    TCHAR buf[PATH_MAX];
-    TCHAR *ptr = NULL;
+  buf[0] = TEXT('\0');
 
-    buf[0] = TEXT('\0');
-
-    res_len = SearchPath(NULL, bundle_file, NULL, PATH_MAX, buf, &ptr);
-    if(res_len > 0) {
-      char *mstr = curlx_convert_tchar_to_UTF8(buf);
-      Curl_safefree(config->cacert);
-      if(mstr)
-        config->cacert = strdup(mstr);
-      curlx_unicodefree(mstr);
-      if(!config->cacert)
-        result = CURLE_OUT_OF_MEMORY;
-    }
+  res_len = SearchPath(NULL, bundle_file, NULL, PATH_MAX, buf, &ptr);
+  if(res_len > 0) {
+    char *mstr = curlx_convert_tchar_to_UTF8(buf);
+    Curl_safefree(config->cacert);
+    if(mstr)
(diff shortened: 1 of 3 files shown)


vszakats:
There is similar code in `lib/url.c` that limits CA bundle handling to Schannel _and_ SecureTransport:
https://github.com/curl/curl/blob/813995bb17da4a41c32235feeab156bf1d5bc059/lib/url.c#L427-L435

Is there a reason for these to be different?

bagder:
> Is there a reason for these to be different than the condition in src/tool_operate.c?

Those are the two TLS backends that don't use a CA file but the native CA store by default.

vszakats:
In that case wouldn't it be better to exclude SecureTransport in the command-line tool as well from the CA bundle detection?

(There is no issue per se, but it's unnecessary work AFAIU.)

bagder:
Yes it would!

src/tool_operate.c
```diff
@@ -3042,11 +3042,13 @@ static CURLcode transfer_per_config(struct GlobalConfig *global,
      * no environment-specified filename is found then check for CA bundle
      * default filename curl-ca-bundle.crt in the user's PATH.
      *
-     * If Schannel is the selected SSL backend then these locations are
-     * ignored. We allow setting CA location for schannel only when explicitly
-     * specified by the user via CURLOPT_CAINFO / --cacert.
+     * If Schannel or SecureTransport is the selected SSL backend then these
+     * locations are ignored. We allow setting CA location for Schannel and
+     * SecureTransport when explicitly specified by the user via
```
bagder, line 3047: nit: We usually refer to it as **Secure Transport** in documentation, using two words

vszakats: Yes, I'm going to fix it in the tidy-up draft PR #14692. (also in url.c where this copy came from)

vszakats: @jay asked the same question when the `url.c` patch was submitted:
https://github.com/curl/curl/pull/12704#issuecomment-1891541873

If we don't mind making this compatibility break, it'd be more consistent to disregard these envvars by default for Secure Transport build. Apple's, Homebrew, and macport's (if I'm reading [it](https://github.com/macports/macports-ports/blob/master/net/curl/Portfile) correctly) curl are not using Secure Transport by default.

vszakats: All in all it doesn't matter given that Secure Transport is going to be deleted next May.

src/tool_doswin.c
```diff
@@ -616,40 +616,29 @@ char **__crt0_glob_function(char *arg)
  */
```
jay, line 617:
```suggestion
/ * Search and set the CA cert file for Windows.
 *
 * Do not call this function if Schannel is the selected SSL backend. We allow
 * setting CA location for Schannel only when explicitly specified by the user
 * via CURLOPT_CAINFO / --cacert.
 */
```
```

jay (approved):
I think this is fine however for posterity I suggest add a comment above the function that it should not be called for Schannel since it is not obvious now that the check is out of the function.

Issue #14843, opened 2024-09-10
curl_easy_perform() in curl 8.9.1 fails with OPERATION_TIMEOUT


rmja:
### I did this

I am uploading a file to a Microsoft FTP Service using libcurl from C#. This has worked fine, but after updating to libcurl 8.9.1 it seems as if `curl_easy_perform()` now returns `OPERATION_TIMEOUT`. The call is going through a socks proxy (needed for me to access the FTP server due to firewall restrictions).

I hope that you can use my description below, even though it is from C#. It is the the verbose output obtained through the `CURLoption.DEBUGFUNCTION` handler.
Note that the file is actually uploaded - but, if I specify a `POSTQUOTE` to be run after the upload, then it is not executed.

> CURLINFO_TEXT: 09.56.39.337   Trying 52.29.67.170:1080...
CURLINFO_TEXT: 09.56.39.363 Connected to 52.29.67.170 (52.29.67.170) port 1080
CURLINFO_TEXT: 09.56.39.454 Host kamanaftps01.westeurope.cloudapp.azure.com:21 was resolved.
CURLINFO_TEXT: 09.56.39.457 IPv6: (none)
CURLINFO_TEXT: 09.56.39.459 IPv4: 52.174.49.89
CURLINFO_TEXT: 09.56.39.461 SOCKS5 connect to 52.174.49.89:21 (locally resolved)
CURLINFO_TEXT: 09.56.39.495 SOCKS5 request granted.
CURLINFO_TEXT: 09.56.39.497 Connected to 52.29.67.170 (52.29.67.170) port 1080
CURLINFO_HEADER_IN: 09.56.39.504 220 Microsoft FTP Service
CURLINFO_HEADER_OUT: 09.56.39.506 AUTH SSL
CURLINFO_HEADER_IN: 09.56.39.537 234 AUTH command ok. Expecting TLS Negotiation.
CURLINFO_TEXT: 09.56.39.540 TLSv1.2 (OUT), TLS handshake, Client hello (1):
CURLINFO_SSL_DATA_OUT: 09.56.39.542 CURLINFO_TEXT: 09.56.39.573 TLSv1.2 (IN), TLS handshake, Server hello (2):
CURLINFO_SSL_DATA_IN: 09.56.39.575 CURLINFO_TEXT: 09.56.39.577 TLSv1.2 (IN), TLS handshake, Certificate (11):
[...]

PR #14848, opened 2024-09-10
shutdown send, proceed normally on timeout

icing:
When ending an FTP upload, we shut down the connection gracefully, since the server should be notified we had send all bytes. Mostly, this is a NOP without TLS involved. With TLS, close-notify messages should be exchanged.

As reported in #14843, not all servers seem to do that. Since it is the server's responsiblity to check it has received everything, we just log the timeout and proceed as if everything is fine.

In the receive direction, we still fail the transfer if the server does not shut down its direction properly.

### DIFF
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -653,12 +653,14 @@ static CURLcode InitiateTransfer(struct Curl_easy *data)
     /* set the SO_SNDBUF for the secondary socket for those who need it */
     Curl_sndbuf_init(conn->sock[SECONDARYSOCKET]);
 
-    Curl_xfer_setup2(data, CURL_XFER_SEND, -1, TRUE);
+    /* FTP uload, shutdown DATA, ignore shutdown errors, as we rely
+     * on the server response on the CONTROL connection. */
+    Curl_xfer_setup2(data, CURL_XFER_SEND, -1, TRUE, TRUE);
   }
   else {
-    /* FTP download: */
+    /* FTP download, shutdown, do not ignore errors */
     Curl_xfer_setup2(data, CURL_XFER_RECV,
-                     conn->proto.ftpc.retr_size_saved, TRUE);
+                     conn->proto.ftpc.retr_size_saved, TRUE, FALSE);
   }
 
   conn->proto.ftpc.pp.pending_resp = TRUE; /* expect server response */
--- a/lib/ftp.c
+++ b/lib/ftp.c
@@ -653,7 +653,7 @@ static CURLcode InitiateTransfer(struct Curl_easy *data)
     /* set the SO_SNDBUF for the secondary socket for those who need it */
     Curl_sndbuf_init(conn->sock[SECONDARYSOCKET]);
 
-    /* FTP uload, shutdown DATA, ignore shutdown errors, as we rely
+    /* FTP upload, shutdown DATA, ignore shutdown errors, as we rely
      * on the server response on the CONTROL connection. */
     Curl_xfer_setup2(data, CURL_XFER_SEND, -1, TRUE, TRUE);
   }
--- a/lib/cfilters.c
+++ b/lib/cfilters.c
@@ -210,7 +210,8 @@ CURLcode Curl_conn_shutdown(struct Curl_easy *data, int sockindex, bool *done)
   else {
     timeout_ms = Curl_shutdown_timeleft(data->conn, sockindex, &now);
     if(timeout_ms < 0) {
-      failf(data, "SSL shutdown timeout");
+      /* info message, since this might be regarded as acceptable */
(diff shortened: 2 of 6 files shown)


dfandrich:
Analysis of PR #14848 at [c560c97e](https://github.com/curl/curl/commit/c560c97e3ec07b216f2bcd7eaba6356450c7315f):

[Test 2055 failed](https://github.com/curl/curl/actions/runs/10792168932/job/29931277067#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

jay:
IMO the servers are at fault and by changing don't we run into a regression with the bugs your shutdown procedure fixed, where some servers that did not actually receive/process all the data unless there was a proper shutdown? What would you think of custom shutdown as an option like CURLOPT_SSL_SHUTDOWN, CURLSSL_FTPS_TIMEOUTS_ALLOWED or something

icing:
> IMO the servers are at fault and by changing don't we run into a regression with the bugs your shutdown procedure fixed, where some servers that did not actually receive/process all the data unless there was a proper shutdown? What would you think of custom shutdown as an option like CURLOPT_SSL_SHUTDOWN, CURLSSL_FTPS_TIMEOUTS_ALLOWED or something

Yes, the server is wrong in not sending a close notify back. Otoh, for an upload (and we only change behaviour for that here) it seems the server responsibility to check that it got all the data. If the server stores our upload as success or not, it should tell curl on the control connection, I guess. 

If we fail the transfer on the timed out shutdown, we abort the complete transfer and we do not learn what the server thinks. That does not seem helpful to the curl user.

jay:
> Yes, the server is wrong in not sending a close notify back. Otoh, for an upload (and we only change behaviour for that here) it seems the server responsibility to check that it got all the data. If the server stores our upload as success or not, it should tell curl on the control connection, I guess.
> 
> If we fail the transfer on the timed out shutdown, we abort the complete transfer and we do not learn what the server thinks. That does not seem helpful to the curl user.

If the shutdown does not complete then don't we risk an RST being sent back to the server if there is some delayed after effect, for example see an old discussion we had about this https://github.com/curl/curl/issues/6149#issuecomment-791930762 though specific to TLS 1.3. Some people made the point there that if the close notify _if libcurl initiated it_ may not have a server close notify reply. Like let's say the server just doesn't send a close notify reply because it doesn't have to in that case, and we read 0 from the connection, then that would be the shutdown. Like in other words don't some of these servers still have to shutdown even without close notify, otherwise the timeout risks something coming in after it and then an errant RST. This is tricky because on download we don't want to open up to truncation attacks by ignoring close notify if it's initiated by the server and there's no other known termination point (ie content length etc)

icing:
@jay, you make good arguments. I believe we should make this change only for FTP uploads for the following reason:

An FTP server is not obliged to close the upload DATA connection immediately once it received everything. It would be nice if it did, but the protocol does not say it is at fault for delaying this. In the reported case, this seems to be exactly what is happening.

Now, if FTP DATA connections were really standalone things in curl, we would just place it on the connection pool shutdown list and await the close concurrently. And an error there would not affect the transfer. As things are, we start the shutdown before continuing on  the CONTROL connection  and have to deal with the server being lazy.

I suggest therefore to only ignore the timeout on an FTP DATA case. Would that work for you?

icing:
I added the change as described above. FTP uploads, and only those, now make shutdown errors being logged, but otherwise ignored. Instead the server answer on the CONTROL connection will determine the success of the `STORE` command.

For all other protocols and situations, the shutdown behaviour is the same as before.

jay:
> FTP uploads, and only those, now make shutdown errors being logged, but otherwise ignored. Instead the server answer on the CONTROL connection will determine the success of the `STORE` command.

so long as there is some way to ensure the file was received fully that sounds fine. in the other issue i came to the conclusion that a close notify must be required and re-reading it yesterday it seems like it would be, however from what you are saying it sounds like in this case due to the mechanics of ftp that's just not possible and the server may keep the data connection as long as it would like

lib/ftp.c
```diff
@@ -651,12 +651,14 @@ static CURLcode InitiateTransfer(struct Curl_easy *data)
     /* set the SO_SNDBUF for the secondary socket for those who need it */
     Curl_sndbuf_init(conn->sock[SECONDARYSOCKET]);
 
-    Curl_xfer_setup2(data, CURL_XFER_SEND, -1, TRUE);
+    /* FTP uload, shutdown DATA, ignore shutdown errors, as we rely
```
bagder, line 654:
```suggestion
    /* FTP upload, shutdown DATA, ignore shutdown errors, as we rely
```

icing: thanks, fixed in recent push.

PR #14850, opened 2024-09-10
runtests: accept 'quictls' as OpenSSL compatible

bagder:
And set the "OpenSSL" feature.

Follow-up to 86d4a4124e275f355320d


dfandrich:
Analysis of PR #14850 at [d27705c3](https://github.com/curl/curl/commit/d27705c302eb006144b37bada41681d784ba6755):

[Test 1244 failed](https://github.com/curl/curl/actions/runs/10792753983/job/29933137013#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14853, opened 2024-09-10
GHA: add `valgrind` to the job titles using it, and tidy-ups

vszakats:
There is a 4-5x difference in test run times. Make the reason more
obvious by adding valgrind to the job names.

Also:
- tidy up job names.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -66,73 +66,73 @@ jobs:
       fail-fast: false
       matrix:
         build:
-          - name: BearSSL valgrind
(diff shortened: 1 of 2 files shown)


dfandrich:
Analysis of PR #14853 at [992019a9](https://github.com/curl/curl/commit/992019a992b27c00a79d937f56ca77232fc78fa7):

[Test 474 failed](https://github.com/curl/curl/actions/runs/10797901645/job/29950193445#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14855, opened 2024-09-10
url.c: removed whitespace from the end of function 'conn_maxage'

gbrlmarn:


bagder:
Thank you, but this seems like a rather pointless fix by itself. We can fix that together with a real fix later on.

gbrlmarn:
Sure, no problem. Thank you!

PR #14858, opened 2024-09-11
Fix AF_INET6 use outside of USE_IPV6

piru:
This fixes the build for systems where AF_INET6 doesn't exist.


icing (approved):
That looks correct. Thanks for fixing this.

bagder:
Thanks!

PR #14488, opened 2024-08-11
GHA/windows: enable HTTPS tests with stunnel

AkiSakurai:
- install stunnel

PR #5721, opened 2020-07-24
CI/azure: install libssh2 for use with msys2-based builds

mback2k:
This enables building and running the SFTP and SCP tests.

PR #14763, opened 2024-09-02
cmake, `Makefile.mk`: use `-isystem` for dep headers, silence BearSSL issues

vszakats:
Patch started out for working around compiler warnings in BearSSL latest
tarball release v0.6 (2018-08-14) and Apple clang 14 with CMake.

Then turned into patching CMake and `Makefile.mk` builds to use
`-isystem` instead `-I` when adding header directories for
dependencies. This avoids compiler warnings in dependency headers,
syncing behaviour with autotools.

Also:
- `Makefile.mk`: add support for BearSSL.
- delete warning suppression for mbedTLS headers. No longer necessary
  after this patch.
  Follow-up to 434db995a7566a76f6a16391897816ddd1011adc #12720

Silenced BearSSL warnings:
```
In file included from curl/lib/vtls/bearssl.c:28:
In file included from bearssl/inc/bearssl.h:127:
bearssl/inc/bearssl_hash.h:727:5: warning: 'BR_DOXYGEN_IGNORE' is not defined, evaluates to 0 [-Wundef]
    ^
bearssl/inc/bearssl_hash.h:745:5: warning: 'BR_DOXYGEN_IGNORE' is not defined, evaluates to 0 [-Wundef]
    ^
In file included from curl/lib/vtls/bearssl.c:28:
In file included from bearssl/inc/bearssl.h:136:
bearssl/inc/bearssl_ssl.h:1253:20: warning: implicit conversion loses integer precision: 'unsigned int' to 'uint16_t' (aka 'unsigned short') [-Wimplicit-int-conversion]
        cc->version_min = version_min;
                        ~ ^~~~~~~~~~~
bearssl/inc/bearssl_ssl.h:1254:20: warning: implicit conversion loses integer precision: 'unsigned int' to 'uint16_t' (aka 'unsigned short') [-Wimplicit-int-conversion]
        cc->version_max = version_max;
                        ~ ^~~~~~~~~~~
bearssl/inc/bearssl_ssl.h:1327:28: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint16_t' (aka 'unsigned short') [-Wimplicit-int-conversion]
        ctx->protocol_names_num = num;
                                ~ ^~~
5 warnings generated.
```

(These warnings were fixed in BearSSL Git master in 2019 via
2893441f2efd4603ddd6d7f49011bdda096a4a87 and
ecdf89770ee82dfea6186fb4369cff3d06cd852e.)

[...]

PR #14859, opened 2024-09-11
CI: move Azure jobs to GHA, fix fallouts, sshserver, runtests tweaks

vszakats:
CI:

- GHA/windows: enable OpenSSH server, SysInternals `handle`, `impacket`.
  Skip `impacket` on MSYS2 due to install failure.
  Skip OpenSSH server for old/standalone mingw-w64 (building curl
  without SSH there.)
- GHA/windows: make test tool installs a separate step.
- GHA/cygwin: enable OpenSSH server.
  Skip `impacket`: it's compiling for 7 minutes then breaks.
  Skip `stunnel` due to sluggish test run performance.
  (This update is unrelated to Azure jobs.)
- GHA/linux: migrate Linux jobs from Azure CI.
- GHA/linux: migrate scanbuild job from Azure CI.
- GHA/linux: enable libssh2 in a job. Also enable valgrind.
- CI/windows: enable SSPI in two jobs.
- CI/windows: disable zlib in one more job.
- CI/windows: improve `if` condition checking GnuTLS.
- CI/windows: ignore SFTP/SCP tests as necessary.
  - universally ignore SCP tests, they fail everywhere.
  - ignore test 612.
  - ignore test 613 616 618 with MSYS2 mingw-w64.
  - ignore test 614 with libssh.
  - ignore all SFTP with MSYS2 native.
  - ignore all SFTP with vcpkg with `libssh2[core,zlib]`.
  - ignore a couple of SFTP tests with MSYS2 mingw-w64.
  (This matches settings on Azure CI.)
- GHA/windows: ignore failing 1451 'Basic SMB request' test for
  old mingw-w64 7.3.0 (but not for 9.5.0!):
  ```
  2024-09-11 21:45:59,738 ERROR smbComNegotiate: b'NT LM 0.12\x00' is not in list
  [...]
  curl: (7) Could not connect to server
  [...]
  FAIL 1451: 'Basic SMB request' SMB
  ```
  Ref: https://github.com/curl/curl/actions/runs/10816280746/job/30007130770#step:13:3546
- CI/Azure: delete, now moved to GHA.

sshserver:

- sshserver: fix permissions for SSH host key on Windows, allowing sshd
  to launch.
- sshserver: fix initializing config paths with Cygwin/MSYS2 sshd.
- sshserver: fix initializing config paths in parallel builds.
- sshserver: delete redundant `DenyUsers`. This also opens the way
  to allow multiple usernames.
- sshserver: fix `AllowUsers` for GHA/windows, by allowing the
  domainless username again.
  Follow-up to 3ee7c676ec8ef32e3a9cb4d0fb2929a8335d58db #5721
- sshserver: fix `Deprecated option` warnings in `sshd_config`:
  ```
  D:/a/curl/curl/bld/tests/log/2/server/curl_sshd_config line 7: Deprecated option AuthorizedKeysFile2
  D:/a/curl/curl/bld/tests/log/2/server/curl_sshd_config line 25: Deprecated option KeyRegenerationInterval
  D:/a/curl/curl/bld/tests/log/2/server/curl_sshd_config line 35: Deprecated option RhostsRSAAuthentication
  D:/a/curl/curl/bld/tests/log/2/server/curl_sshd_config line 36: Deprecated option RSAAuthentication
  D:/a/curl/curl/bld/tests/log/2/server/curl_sshd_config line 37: Deprecated option ServerKeyBits
  D:/a/curl/curl/bld/tests/log/2/server/curl_sshd_config line 41: Deprecated option UseLogin
  ```
- sshserver: fix `Deprecated option` warnings in `ssh_config` with
  Cygwin/MSYS2.
- sshserver: fix dumping config files due to the filenames missing their
  full paths.
- sshserver: add workaround to make `logmsg` messages visible.
  Before this patch they only went to a file and never shown.

runtests:

- runtests: log details when these Windows commands are called:
  `handle`, `taskkill`, `tasklist`.
- runtests: add documentation links to Windows tools:
  `handle`, `taskkill`, `tasklist`, `icacls`.
- runtests: add `-t` (kill whole tree) option to `taskkill` in
  `servers.pm`, syncing it with the other `taskkill` call.
  Follow-up to bc72a78a11764558639131d51fa8e7c81ee1e113 #14488
- runtests: show warning if Sysinternals `handle` tool is missing.
- runtests: drop Windows XP Home compatibility `tskill` call.
  The call was made on all Windows versions. It's possibly overkill to
  do this, because XP Home is probably rarely used for running curl
  tests these days. In case it's needed, it'd be better to put it under
  an explicit option.
- runtests: show Perl version and path.

Fix/silence fallouts:

- unit2603: fix building with disabled HTTP support.
- unit2604: silence `-Woverlength-strings` warnings in C89 mode.
- test437, test1614: fix to pass with no-IPv6 builds.

---

- [x] rebase on #14919.
- [x] delete newly added redundant Linux torture test.
- [x] fix SSH test not running
- [x] run pytest. → [SKIP for now]
- [x] fix/silence test fallouts (see in comments).
- [x] delete migrated Linux jobs from Azure CI
- [x] fix SSH deprecation warnings.
- [x] sshserver.pl logmsg() output doesn't appear in the log in GHA/windows. [→ WORKAROUND]
- [x] ssh configs turn up empty in the logs in GHA/windows: [→ FIXED]
  ```
  === Start of file curl_sftp_config
  === End of file curl_sftp_config
  [...]
  === Start of file curl_sshd_config
  === End of file curl_sshd_config
  ```
- [x] SSH: `runneradmin@127.0.0.1: Permission denied (publickey).` [FIXED with an ugly hack]
- [x] cygwin: icacls issues: https://github.com/curl/curl/actions/runs/10849527999/job/30108922645?pr=14859#step:14:1605
- [x] cygwin: pathhelp issues: https://github.com/curl/curl/actions/runs/10849527999/job/30108922645?pr=14859#step:14:1820
  `Use of uninitialized value $path in pattern match (m//) at ../../tests/pathhelp.pm line 428.`
- [x] cygwin: misses to create key filenames.


vszakats:
There are few legit-looking test failures after unlocking more tests:

mingw-w64 7.3.0:
```
FAIL 1451: 'Basic SMB request' SMB
```
https://github.com/curl/curl/actions/runs/10816280746/job/30007130770#step:13:3546

Linux no-ssl no-http, single-use check:
```
Curl_memrchr in curl_memrchr
```
https://github.com/curl/curl/actions/runs/10816280747/job/30007145114#step:36:21

Linux OpenSSL3 no-ipv6:
```
[...]
-h1 127.0.0.1 46731 h1 [ffff::1] 8181 TIMESTAMP 0 0[CR][LF]
+h1 127.0.0.1 46731 h1 ffff::1 8181 TIMESTAMP 0 0[CR][LF]
[...]
fe80::ab47:4396:55c9:8474 in fe80::ab47:4396:55c9:8474/64 should match
[...]
TESTFAIL: These test cases failed: 437 1614
```
https://github.com/curl/curl/actions/runs/10816280747/job/30007144504#step:41:1390
https://github.com/curl/curl/actions/runs/10816280747/job/30007144504#step:41:3649

dfandrich:
Analysis of PR #14859 at [756c2638](https://github.com/curl/curl/commit/756c26383ba270a15b60ef7b5e8e1c18cc96967e):

[Test 437 failed](https://github.com/curl/curl/actions/runs/10816280747/job/30007144504#check-step-41), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1614 failed](https://github.com/curl/curl/actions/runs/10816280747/job/30007144504#check-step-41), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 1451 failed](https://github.com/curl/curl/actions/runs/10816280746/job/30007130770#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

vszakats:
After fixing private key permissions (I'm still unsure why this wasn't an issue on Azure), SSH still wasn't used on Windows.

It looks like the reason is an issue with usernames, because deleting `AllowUsers`, `DenyUsers` from `sshd_config` magically fixed it (after 2 days of trying).

Hard to say more because `logmsg` output (from `sshserver.pl`) doesn't appear in GHA logs for some reason (locally it doesn't, but that's not on Windows), which would be the obvious way to trace this.

Can be casing mismatch, `USERDOMAIN` appended, MSYS2-username vs Windows username, or something more mundane. (And again, Azure jobs aren't affected by this, under their custom Docker image.)

vszakats:
unit2604 fails to compile in C89 mode:
```
unit2604.c: In function ‘test’:
unit2604.c:56:5: error: string length ‘542’ is greater than the length ‘509’ ISO C90 compilers are required to support [-Woverlength-strings]
   56 |     { SA540 " c", SA540, "c", "/", CURLE_OK},
      |     ^
unit2604.c:56:5: error: string length ‘540’ is greater than the length ‘509’ ISO C90 compilers are required to support [-Woverlength-strings]
unit2604.c:57:5: error: string length ‘545’ is greater than the length ‘509’ ISO C90 compilers are required to support [-Woverlength-strings]
   57 |     { "\" " SA540 "\" c", " " SA540, "c", "/", CURLE_OK},
      |     ^
unit2604.c:57:5: error: string length ‘541’ is greater than the length ‘509’ ISO C90 compilers are required to support [-Woverlength-strings]
make[1]: *** [Makefile:1224: unit2604.o] Error 1
```
Ref: https://github.com/curl/curl/actions/runs/10840223352/job/30082133462?pr=14859

vszakats:
Interesting fallout with the vcpkg/mbedtls build after disabling zlib:
```
C:\vcpkg\installed\x64-windows\include\psa\crypto_struct.h(254,13): error C2220: the following warning is treated as an error [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
  
C:\vcpkg\installed\x64-windows\include\psa\crypto_struct.h(254,13): warning C4200: nonstandard extension used: zero-sized array in struct/union [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
```
Ref: https://github.com/curl/curl/actions/runs/10842694205/job/30088703078?pr=14859

vszakats:
@dfandrich: Do you have an idea why these two issues may happen in GHA/windows:

In `sshserver.pl` `logmsg()` output doesn't appear in the log. I tried setting `$verbose` to 1, and also patching some lines to print info when `os_is_win()`, but it didn't make them appear.

sftp/sshd config logs are dumped to the logs, but they appear empty. Even though `curl_sshd_config` is most likely not empty on disk:
```
=== Start of file curl_sftp_config
=== End of file curl_sftp_config
[...]
=== Start of file curl_sshd_config
=== End of file curl_sshd_config
```
https://github.com/curl/curl/actions/runs/10823673269/job/30029575000

vszakats:
I fixed the sftp/sshd config dump, and added a workaround to be able to see sshd `logmsg` output.

vszakats:
What remains interesting an outside of the scope of this PR,
it the fallouts seen with the now enabled SCP and SFTP tests.
In particular the total failure with vcpkg `libssh2[core,zlib]`,
and several individual ones depending on libssh, and/or the 
runner environment.

This is the first time SCP/SFTP tests are run with Cygwin, MSYS2,
and MSVC, including Cygwin/MSYS2 built-in openssh server,
and so on.

vszakats:
@curl/developers Any concerns merging this before the release?

This PR touches the CI, contains `sshserver.pl` fixes, and fixes rare build issues in unit tests.
No effect on curl/libcurl itself.

bagder (approved):
I'm fine with merge before the release.

vszakats:
There is an interesting phenomenon popping up now the second time while working on this:

I added `CURL_ZLIB=OFF` to a vcpkg job to test such build, because Azure CI had it.

Initially this job was with mbedTLS and disabling ZLIB triggered a compiler warning _in mbedTLS headers_. This felt odd.

After swapping this job back to GnuTLS + nghttp3 in master and rebasing to it, a different compiler warning, never seen before, popped up within nghttp3 headers!

I wonder how ZLIB headers or ZLIB-enabled curl code can have such an effect on external headers, and if this is an indication of an issue lurking in the builds.

They don't reproduce locally with a different compiler (I don't have MSVC or Windows).

(There is PR #14763 pending to silence warnings in 3rd-party headers, which fixes this, but I wonder nevertheless. I also did not want to merge that PR before release, though it's probably safe.)

mback2k:
Are you sure this isn't reducing the test scope compared to the jobs I originally created on Azure? Is SSH and everything stunnel-based still fully tested?

I would like to take a deeper look.

vszakats:
> Are you sure this isn't reducing the test scope compared to the jobs I originally created on Azure? Is SSH and everything stunnel-based still fully tested?
> 
> I would like to take a deeper look.

stunnel was added a few weeks ago. This PR adds the rest.

I didn't go for a 1:1 match for the Windows jobs, as we already have plenty of them, but added the unique options that were missing. For Linux there was already some overlap, and added the unique ones here.

I specifically did not add another libssh job (for mingw-w64) because it's so terribly insecure on Windows, that I figured the existing MSVC one offers enough coverage.

I also skipped the 32-bit jobs. It seems easier to clone existing GHA jobs into 32-bit versions, though it'd probably better done as a separate PR after merging the long pending PR queue, to avoid conflicting with many of them.

mback2k:
Okay, thanks for the all the effort and work on this topic. So, I guess the main thing will be dropping 32-bit and msys1-based tests for now, right?

In case you weren't aware and may want to take a look, you can find the Dockerfiles of the existing Azure container images here for reference: https://github.com/mback2k/curl-docker-winbuildenv

> * Skip `impacket` on MSYS2 due to install failure.

Maybe I can help with this. What was the exact issue in which build variant?

vszakats:
> Okay, thanks for the all the effort and work on this topic. So, I guess the main thing will be dropping 32-bit and msys1-based tests for now, right?

AFAICS, MSYS1 has not been running in Azure for a while, only MSYS2. This is a run from today:
https://dev.azure.com/daniel0244/curl/_build/results?buildId=25771&view=logs&jobId=ce68c027-2470-56de-cd4a-94aa5f1cbf84&j=ce68c027-2470-56de-cd4a-94aa5f1cbf84

But, correct, MSYS1 is not planned for GHA jobs. And yes, dropping 32-bit for now. We should probably add a couple of them just in case after this settles.

> In case you weren't aware and may want to take a look, you can find the Dockerfiles of the existing Azure container images here for reference: https://github.com/mback2k/curl-docker-winbuildenv

Thanks, I did use it a resource for this PR!

> > * Skip `impacket` on MSYS2 due to install failure.
> 
> Maybe I can help with this. What was the exact issue in which build variant?

Happens with all MSYS2 and Cygwin, the problem is that `pip install impacket` tries to build a ginormous dependency tree, with things like a rustup environment, ffi and Jinja2. At the 7-10 minutes mark, builds fail in both envs, with different errors. This also happens with `--prefer-binary`, and also when available python components are preinstalled via the package manager. (Though I may not have exploited all the tricks or may have made errors.)

You can find the links to the GHA logs with details in this commit:
https://github.com/curl/curl/pull/14859/commits/28a7ddc193c3392b74118256c443cbdb7f851d13

I feel this isn't so high priority to chase down for two reasons:
1. impacket is used in a single test (1451). Though I'm not sure, it may be 2 (that was the test count bump I saw with impacket), or even 3 (grepping around the tests/data directory).
2. impacket installs successfully in quite many jobs already. On Windows it's used with 2 old/standalone mingw-w64 jobs and all MSVC ones.

In working (Windows) jobs it's a 10-20s install task for single test taking 0.5s, so it's a bit wasteful.

But, if you have an idea how to bring it smoothly to MSYS2/Cygwin, we can certainly do that.

vszakats:
After several hours of testing, it turns out _something_ is sneaking in the MSVC compiler option `/external:W4` when ZLIB is _not_ used in this vcpkg job. With ZLIB selected, `/external:W0` is used instead. This option does what the name implies: It cranks up compiler warnings in external headers to the maximum!

This explains the mbedTLS fallout and it's entirely not something curl's code, or ZLIB code is causing. It doesn't even matter is ZLIB code is included or not.

It's curl's CMake scripts doing it, and it also doesn't seem like CMake is doing it.

Diff good vs bad MSVC compiler options:
```
--- zlib-good
+++ nozlib-bad
@@ -2,6 +2,7 @@
 /c
 /ID:\a\curl\curl\include
 /I"C:\vcpkg\installed\x64-windows\debug\..\include"
+/I"C:\vcpkg\installed\x64-windows\include"
 /ID:\a\curl\curl\bld\lib
 /ID:\a\curl\curl\lib
 /I"C:\vcpkg\installed\x64-windows\debug\..\include"
@@ -35,9 +36,8 @@
 /Zc:inline
 /Fo"libcurl_object.dir\Debug\\"
 /Fd"libcurl_object.dir\Debug\libcurl_object.pdb"
-/external:W0
+/external:W4
 /Gd
 /TC
 /errorReport:queue
-/external:I "C:/vcpkg/installed/x64-windows/include"
 D:\a\curl\curl\bld\lib\CMakeFiles\libcurl_object.dir\Unity\unity_0_c.c
```

Builds logs:
nozlib-bad: https://github.com/curl/curl/actions/runs/10874520291/job/30171891878
zlib-good: https://github.com/curl/curl/actions/runs/10874520291/job/30171892022

/cc @talregev Have you seen this, do you have an idea why is it happening?

vszakats:
Since GHA/windows already built curl without zlib (in old/manual mingw-w64 jobs), I solved the MSVC ZLIB fallout by leaving ZLIB enabled in those jobs.

mback2k:
> In working (Windows) jobs it's a 10-20s install task for a single test taking 0.5s, so it's a bit wasteful.
> 
> But, if you have an idea how to bring it smoothly to MSYS2/Cygwin, we can certainly do that.

You could try to pre-install the corresponding MSYS2/MINGW pre-built packages of all relevant dependencies of impacket:

MINGW32: https://github.com/mback2k/curl-docker-winbuildenv/blob/9988880eedcc67c2d5b8dfb17c0e3cc386d7e331/Dockerfile.msys2-mingw32#L9
MINGW64: https://github.com/mback2k/curl-docker-winbuildenv/blob/9988880eedcc67c2d5b8dfb17c0e3cc386d7e331/Dockerfile.msys2-mingw64#L9

E.g.: https://packages.msys2.org/base/mingw-w64-python-pyopenssl

And for Cygwin pre-built ones should also be available: https://cygwin.com/packages/summary/python39-openssl.html

I think impacket pulls in rustls if a modern version of pyopenssl and/or cryptography is manually being installed. By installing the packaged versions of those dependencies I was able to avoid rustls and the whole buildchain for that in the past.

vszakats:
Tried to that for MSYS2 based on your Docker commands, and failed. (You can find it in the commit history).

For Cygwin I missed the OpenSSL package you found. But, it's the kind with a specific Python version in its package name, which makes it fragile, and cumbersome to maintain. FreeBSD does that and someone needs to jump in to fix it the instant Python gets upgraded upstream (and at different times for ARM64 and Intel!), to unbreak CI. I'm not sure it's worth it.

mback2k:
Okay, I could also work on that impacket part (and maybe make it version-independent) once you have landed this PR.

talregev:
> /cc @talregev Have you seen this, do you have an idea why is it happening?

Maybe @dg0yt have an idea.

vszakats:
@mback2k I wonder if there is a better way to test SMB. impacket is a huge dependency, since it seems to be a general tool for testing all sorts of protocols. curl needs it for SMB only. Just counted on NetBSD, it's a 100MB+ install with 25 dependencies: https://github.com/curl/curl/pull/14929#issuecomment-2352326081. (It installs there instantly at least, but sadly also via a versioned package name.)

mback2k:
@vszakats I will look into it, but for now I guess it will be easier to try to install pre-compiled versions of the crypto dependencies.

> It still looks weird to install Flask + werkzeug + jinja2 + markdown processor (= complete web dev env), and also rustup environment, an extra OpenSSL copy (?), and compile that all from source in some envs, to have a headless SMB server. Not NodeJS-bad, but close.

True, but at least the crypto dependencies and compilation steps (e.g. rust) can probably be trimmed down this way.

dg0yt:
> _something_ is sneaking in the MSVC compiler option `/external:W4` when ZLIB is _not_ used in this vcpkg job. With ZLIB selected, `/external:W0` is used instead

Some CMake package could sneak in public compile options. If it disappears with zlib enabled, I wonder if something sneaks in a vendored copy or alternative instead. Running CMake with `--trace-expand` could provide insight.

vszakats:
@mback2k FYI I enabled impacket on NetBSD and macOS (it was installed but not used there) for exxtra coverage. I think we have plenty. WSL ...sounds complex! Would it work with parallel tests and other types of servers?

@dg0yt Tried `--trace-expand`: https://github.com/curl/curl/actions/runs/10889517591
There is no clear sign of `/external:W4` before it's used in the libcurl compile command. It might be hidden in an ungreppable form, or maybe visible for someone knowing where to look. Diffing the logs (also after `sort -u`) also didn't turn up leads.

vszakats:
`libssh2[core,zlib]` (using WinCNG) vcpkg builds fail to connect to the SSH server with `LIBSSH2_ERROR_KEY_EXCHANGE_FAILURE=-8`:
```
* Failure establishing ssh session: -8, Unable to exchange encryption keys
```
Ref: https://github.com/curl/curl/actions/runs/10902293795/job/30253865808#step:14:3622

PR #14861, opened 2024-09-11
multi: do not call the socket callback when cleanup is called

bagder:
There is no point since the handle is getting killed.


bagder:
It's too crude as this then fails to remove previously added events...

PR #14862, opened 2024-09-11
multi: check that the multi handle is valid in curl_multi_assign

bagder:


dfandrich:
Analysis of PR #14862 at [60ac76d6](https://github.com/curl/curl/commit/60ac76d67bf32dfb020cd155fc27fe1f03ac404f):

[Test 1540 failed](https://github.com/curl/curl/actions/runs/10811380868/job/29990667750#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14876, opened 2024-09-12
GHA/macOS: add an -e test

bagder:
Adds -e to runtests for using the event-based API in the tool.


vszakats (approved):

bagder:
We got a report over IRC about an event-based problem only happening on macOS, and then I noticed we did not run any event-based tests on macOS in the CI. Hence this PR.

Issue #14870, opened 2024-09-11
8.10 breaks --upload-file '-' with http2


nekopsykose:
### I did this

i have a pastebin script that does

```sh
file="${1:--}"
curl -sS --http2 -u "user:$secret" --globoff --upload-file "$file" "$domain"
```

if i unwrap that, and pass:

```
echo 123 | curl --http2 -u "user:$secret" --upload-file "-" "$domain"
```

it hangs forever. on the server side (nginx), i can see

```
[2024-09-11T17:53:38+00:00]-ipaddr-my.domain "499-PUT /upload HTTP/2.0" curl/8.10.0
```

which seems to indicate 'client closed connection early' with the 499 since it's an nginx-specific code

if i do

```
echo 123 > file
... --upload-file "./file"
```

it works as expected. i can attach a --trace if it's helpful, but i'm not sure if that leaks the credentials..

### I expected the following

for --upload-file to work like in 8.9.1

### curl/libcurl version

curl 8.10.0 (x86_64-chimera-linux-musl) libcurl/8.10.0 OpenSSL/3.3.2 zlib/1.3.1.zlib-ng zstd/1.5.6 c-ares/1.33.1 libidn2/2.3.7 libpsl/0.21.5 libssh2/1.11.0 nghttp2/1.63.0 nghttp3/1.5.0
Release-Date: 2024-09-11
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Largefile libz NTLM PSL SSL threadsafe TLS-SRP UnixSockets zstd

### operating system

chimera linux (`Linux phlox 6.10.9-0-generic #1 SMP PREEMPT_DYNAMIC Sun Sep  8 20:55:32 UTC 2024 x86_64`)

nekopsykose:
just as a quick test- i can reproduce this in an arch linux container too (with curl 8.10 from core-testing), and 8.9 works there too, just to rule out that it's a magic musl libc issue

nekopsykose:
bisected to 35bf766280a77cde3055e0f4e249ab02a0dcd275

dfandrich:
[...]

PR #14877, opened 2024-09-12
upload data from stdin, fix eos forwarding

icing:
When uploading data from stdin ('-T -'), and the EOS was only detected on a 0-length read, the EOS was not forwarded to the filters. This led HTTP/2 to hang on not forwarding this to the server.

Refs #14870

Added test_07_14 to reproduce and verify.


nekopsykose:
confirmed, fixes it for me too :)

icing:
> confirmed, fixes it for me too :)

Hope you tested the current version of this PR. The previous one I borked.😌

nekopsykose:
yeah, the one right as i posted that :D

PR #14881, opened 2024-09-12
tidy-up: rename `CURL_WINDOWS_APP` to `CURL_WINDOWS_UWP`

vszakats:
Rename internal macro to make its purpose more obvious.

After this patch `grep -i uwp` shows all the code related to UWP.

Ref: https://curl.se/mail/lib-2024-09/0014.html

### DIFF
--- a/lib/asyn-thread.c
+++ b/lib/asyn-thread.c
@@ -286,7 +286,7 @@ static CURLcode getaddrinfo_complete(struct Curl_easy *data)
  * and wait on it.
  */
 static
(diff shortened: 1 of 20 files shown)


MarcelRaad (approved):
The double "Windows" in there looks a bit strange to me, but I agree that having "UWP" in the identifier makes sense.

vszakats:
@MarcelRaad Oh indeed, it's twice Windows. But, `WIN` offers a tiny grepping advantage, and gives context. There is also precedent for it: `_WIN32_WCE`!

PR #14889, opened 2024-09-12
vtls/rustls: 0.14.0 update and assoc. improvements

cpu:
:wave: This branch updates the experimental `rustls` vtls backend for `curl` to use the just-released [rustls-ffi 0.14.0](https://github.com/rustls/rustls-ffi/releases/tag/v0.14.0), and the latest release of the main rust crate, [rustls 0.23.13](https://github.com/rustls/rustls/releases/tag/v%2F0.23.13).

Along the way I made a handful of other small improvements and so I recommend reviewing commit-by-commit. My goal was to have each commit build independent of the others, and to avoid introducing new functionality at the same time as updating breaking API changes. I've done my best to match local style but I'm still quite green at contributing to `curl`: feedback very welcome.

In addition to the 0.14.0 update, this branch resolves the "weak RNG" situation described in https://github.com/curl/curl/pull/14770 and replaces the ciphersuite name string comparison logic from https://github.com/curl/curl/pull/14840 with a more robust solution based on protocol version constants.

### DIFF
--- a/lib/rand.c
+++ b/lib/rand.c
@@ -100,9 +100,9 @@ CURLcode Curl_win32_random(unsigned char *entropy, size_t length)
 }
 #endif
 
-#if !defined(USE_SSL) || defined(USE_RUSTLS)
+#if !defined(USE_SSL)
 /* ---- possibly non-cryptographic version following ---- */
-CURLcode Curl_weak_random(struct Curl_easy *data,
+static CURLcode weak_random(struct Curl_easy *data,
                           unsigned char *entropy,
                           size_t length) /* always 4, size of int */
 {
@@ -151,7 +151,7 @@ CURLcode Curl_weak_random(struct Curl_easy *data,
 #ifdef USE_SSL
 #define _random(x,y,z) Curl_ssl_random(x,y,z)
 #else
-#define _random(x,y,z) Curl_weak_random(x,y,z)
+#define _random(x,y,z) weak_random(x,y,z)
 #endif
 
 static CURLcode randit(struct Curl_easy *data, unsigned int *rnd,
--- a/lib/vtls/rustls.c
+++ b/lib/vtls/rustls.c
@@ -436,7 +436,7 @@ cr_get_selected_ciphers(struct Curl_easy *data,
                         size_t *selected_size)
 {
   size_t supported_len = *selected_size;
-  size_t default_len = rustls_default_ciphersuites_len();
+  size_t default_len = rustls_default_crypto_provider_ciphersuites_len();
   const struct rustls_supported_ciphersuite *entry;
   const char *ciphers = ciphers12;
   size_t count = 0, default13_count = 0, i, j;
@@ -448,7 +448,7 @@ cr_get_selected_ciphers(struct Curl_easy *data,
     /* Add default TLSv1.3 ciphers to selection */
     for(j = 0; j < default_len; j++) {
       struct rustls_str s;
-      entry = rustls_default_ciphersuites_get_entry(j);
+      entry = rustls_default_crypto_provider_ciphersuites_get(j);
       s = rustls_supported_ciphersuite_get_name(entry);
       if(s.len < 5 || strncmp(s.data, "TLS13", 5) != 0)
         continue;
@@ -471,7 +471,7 @@ add_ciphers:
     /* Check if cipher is supported */
     if(id) {
       for(i = 0; i < supported_len; i++) {
-        entry = rustls_all_ciphersuites_get_entry(i);
+        entry = rustls_default_crypto_provider_ciphersuites_get(i);
         if(rustls_supported_ciphersuite_get_suite(entry) == id)
           break;
       }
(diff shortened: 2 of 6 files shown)


cpu:
(cc @ctz in case you were interested in giving this a :mag:)

lib/rand.c
```diff
@@ -100,7 +100,7 @@ CURLcode Curl_win32_random(unsigned char *entropy, size_t length)
 }
 #endif
 
-#if !defined(USE_SSL) || defined(USE_RUSTLS)
```
cpu, line 103: I notice some CI runs are now failing the "single use function check" ([exemplar](https://github.com/cpu/curl/actions/runs/10836519408/job/30070467574)).

I could use some advice on how best to fix that one. It seems like the `!defined(USE_SSL)` build configuration would still want this function to exist, so it shouldn't be deleted outright (?). Is the intended fix to in-line this somewhere?

lib/vtls/rustls.c
```diff
@@ -505,10 +505,9 @@ cr_get_selected_ciphers(struct Curl_easy *data,
   if(!ciphers12) {
     /* Add default TLSv1.2 ciphers to selection */
     for(j = 0; j < default_len; j++) {
-      struct rustls_str s;
       entry = rustls_default_crypto_provider_ciphersuites_get(j);
-      s = rustls_supported_ciphersuite_get_name(entry);
-      if(s.len >= 5 && strncmp(s.data, "TLS13", 5) == 0)
+      if(rustls_supported_ciphersuite_protocol_version(entry) ==
+          RUSTLS_TLS_VERSION_TLSV1_3)
```
ctz, line 510: there's another one of these further up in this function.

ctz: (but, the inverse)

cpu: Oops, good catch. Thank you. I was over-focused on the skip logic that was updated in 14840 and missed this. Fixed :ballot_box_with_check:

lib/rand.c
bagder: Hm, weird. I'll have a look.

bagder: Ah, not so weird really. The function is now *not* static and uses a `Curl_` prefix, but there is nothing outside the file that uses it so it should rather be made static (and not use the global prefix). Like this:

~~~diff
diff --git a/lib/rand.c b/lib/rand.c
index 3e4f500c0..edc293d6f 100644
--- a/lib/rand.c
+++ b/lib/rand.c
@@ -100,13 +100,13 @@ CURLcode Curl_win32_random(unsigned char *entropy, size_t length)
 }
 #endif
 
 #if !defined(USE_SSL)
 /* ---- possibly non-cryptographic version following ---- */
-CURLcode Curl_weak_random(struct Curl_easy *data,
-                          unsigned char *entropy,
-                          size_t length) /* always 4, size of int */
+static CURLcode weak_random(struct Curl_easy *data,
+                            unsigned char *entropy,
+                            size_t length) /* always 4, size of int */
 {
   unsigned int r;
   DEBUGASSERT(length == sizeof(int));
 
   /* Trying cryptographically secure functions first */
@@ -149,11 +149,11 @@ CURLcode Curl_weak_random(struct Curl_easy *data,
 #endif
 
 #ifdef USE_SSL
 #define _random(x,y,z) Curl_ssl_random(x,y,z)
 #else
-#define _random(x,y,z) Curl_weak_random(x,y,z)
+#define _random(x,y,z) weak_random(x,y,z)
 #endif
 
 static CURLcode randit(struct Curl_easy *data, unsigned int *rnd,
                        bool env_override)
 {

cpu: :bulb: Aha! That makes sense, thank you. I've applied your suggested diff.

lib/vtls/rustls.c
```diff
@@ -1024,6 +1056,16 @@ static size_t cr_version(char *buffer, size_t size)
   return msnprintf(buffer, size, "%.*s", (int)ver.len, ver.data);
 }
 
+static CURLcode
+cr_random(struct Curl_easy *data,
+                   unsigned char *entropy, size_t length) {
```
bagder, line 1061:
```suggestion
cr_random(struct Curl_easy *data, unsigned char *entropy, size_t length)
{
```

lib/rand.c
```diff
@@ -100,9 +100,9 @@ CURLcode Curl_win32_random(unsigned char *entropy, size_t length)
...
+#if !defined(USE_SSL)
 /* ---- possibly non-cryptographic version following ---- */
-CURLcode Curl_weak_random(struct Curl_easy *data,
+static CURLcode Curl_weak_random(struct Curl_easy *data,
```
bagder, line 105: This fixes the build:

```suggestion
static CURLcode weak_random(struct Curl_easy *data,
```

lib/vtls/rustls.c
cpu: Applied. Thanks

lib/rand.c
cpu: :facepalm: Ah of course. I was trying to fix this the wrong way. Adjusting.

cpu: Backed out my silly forward decl and updated with this fix. Thanks!

bagder (approved):
:+1:

cpu:
Thanks for the reviews :-)

Issue #14898, opened 2024-09-13
curl_multi is broken in libcurl 8.10.0 - PHPs curl_multi_exec still_running argument is now always non-zero, causing endless loop in reading servers response


vityank:
### I did this

```
<?php


// start testing
  echo "*** Testing curl_exec() : basic functionality ***\n";

  $url = "http://127.0.0.1:32550/get.inc?test=get";
  //$url = "http:/127.0.0.1/index.php";
  echo "Preparing handles array...\n";
  $chs = array(
    0 => curl_init(),
    1 => curl_init(),
    2 => curl_init(),
  );

  echo "Starting ob...\n";
  ob_start(); // start output buffering

  $options = array(
    CURLOPT_RETURNTRANSFER => 1,
    CURLOPT_URL => $url,
  );

  echo "Setting options...\n";
  curl_setopt_array($chs[0], $options); //set the options
  curl_setopt_array($chs[1], $options); //set the options
  curl_setopt_array($chs[2], $options); //set the options

  echo "About to call multi init...\n";
  $mh = curl_multi_init();

  // add handlers
  echo "Adding handlers...\n";
  curl_multi_add_handle($mh, $chs[0]);
  curl_multi_add_handle($mh, $chs[1]);
  curl_multi_add_handle($mh, $chs[2]);

  $running=null;
  //execute the handles
  echo "Entering loop...\n";
  do {
    $vst = curl_multi_exec($mh, $running);
    echo "cURL Code is: $vst\n";
  } while ($running > 0);

  echo "Pulling content...\n";
  $curl_content = '';
  $curl_content .= curl_multi_getcontent($chs[0]);
  $curl_content .= curl_multi_getcontent($chs[1]);
  $curl_content .= curl_multi_getcontent($chs[2]);

  //close the handles
  echo "Cleaning up...\n";
  curl_multi_remove_handle($mh, $chs[0]);
  curl_multi_remove_handle($mh, $chs[1]);
  curl_multi_remove_handle($mh, $chs[2]);
  echo "Closing multi handle...\n";
  curl_multi_close($mh);

  var_dump( $curl_content );
```

### I expected the following

*** Testing curl_exec() : basic functionality ***
Preparing handles array...
Starting ob...
Setting options...
About to call multi init...
Adding handlers...
Entering loop...
cURL Code is: 0
cURL Code is: 0
cURL Code is: 0
cURL Code is: 0
cURL Code is: 0
cURL Code is: 0
cURL Code is: 0
[...]

PR #14901, opened 2024-09-13
fix sendrecv() without interim poll

icing:
There was a "clever" optimization that skipped sendrecv() handling when the transfer's pollset was empty. This happens for paused transfers, for example.

Unforunately, if the libcurl application never calls curl_multi_poll(), the pollset is and will aways remain empty, prevent the transfer from progressing.

Remove this "optimization" and always try send/receive where applicable.

refs #14898


bagder (approved):

bagder:
You're the best @icing!

PR #14905, opened 2024-09-13
cmake/FindNGTCP2: use library path as hint for finding the crypto module

vszakats:
It allows finding the ngtcp2 crypto interface library automatically when
using a custom `NGTCP2_LIBRARY`.

Before this patch the library location had to be added via
`CMAKE_LIBRARY_PATH` or by other means.

Also add empty lines for readability / uniformity.

Fixes https://github.com/curl/curl-for-win/blob/8b8909e1206de1dcca356a8dd33eb1e4ffeea7fd/curl.sh#L289

---

TODO:
- [x] test it with curl-for-win after deleting `CMAKE_LIBRARY_PATH` option from curl.sh. (tested OK locally with non-curl-for-win, but a custom native build.)
- [x] test when this directory is empty or undefined.

### DIFF
--- a/CMake/FindNGTCP2.cmake
+++ b/CMake/FindNGTCP2.cmake
@@ -83,12 +83,15 @@ if(NGTCP2_FIND_COMPONENTS)
   endforeach()
 
   if(_ngtcp2_crypto_backend)
+    get_filename_component(_ngtcp2_library_dir "${NGTCP2_LIBRARY}" DIRECTORY)
+
     string(TOLOWER "ngtcp2_crypto_${_ngtcp2_crypto_backend}" _crypto_library)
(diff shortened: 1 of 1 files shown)


dfandrich:
Analysis of PR #14905 at [72bc43c5](https://github.com/curl/curl/commit/72bc43c54abeb8463ca7bed1a1d28107b33471b2):

[Test 3023 failed](https://github.com/curl/curl/actions/runs/10856363499/job/30130888744#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 3024 failed](https://github.com/curl/curl/actions/runs/10856363499/job/30130888744#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14906, opened 2024-09-14
cmake: separate target for examples, optimize CI, fix fallouts

vszakats:
- Move `docs/examples` builds under a separate target.

- Make `BUILD_EXAMPLES` default to `ON`. It means to generate the rules
  for `docs/examples` by default, but not build them. To build them,
  an explicit `make curl-examples` (or ninja, etc) command is necessary.
  This syncs behaviour with autotools, and also how both cmake and
  autotools are building tests.

- GHA: update cmake jobs to use the new way of building examples.

- GHA: move examples build step at the end of the job, after building
  and running tests. This allows to have build and test run results
  faster, and leave the seldom-changing examples build to the end.
  Building examples is the slowest build step with no practical way to
  make them fast.

- appveyor: enable building examples in two old-MSVC jobs.

- examples: fix examples to build cleanly with old MSVC versions.

- GHA/non-native: move example build log under a GHA foldable section.

- GHA/windows: move building examples into separate step for Linux cross
  jobs.

Follow-up to dfdd978f7c60224dffe2aac25b436dc0a5cd0186 #13491

### DIFF
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -113,11 +113,6 @@ jobs:
           find . -name '*.exe' -o -name '*.dll'
           bld/src/curl.exe --disable --version
 
-      - name: 'autotools build examples'
-        if: ${{ matrix.build == 'automake' }}
-        timeout-minutes: 5
-        run: make -C bld -j5 V=1 examples
-
       - name: 'autotools build tests'
         if: ${{ matrix.build == 'automake' && matrix.tflags != 'skipall' }}
         timeout-minutes: 15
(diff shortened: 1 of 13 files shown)


dfandrich:
Analysis of PR #14906 at [8d031188](https://github.com/curl/curl/commit/8d03118899a658cd5ee33d16f2dca41278472219):

[Test 3023 failed](https://github.com/curl/curl/actions/runs/10858001821/job/30135376138#check-step-13), which has NOT been flaky recently, so **there could be a real issue in the PR**.

vszakats:
Yes, test 3023 started failing yesterday with mingw-w64 7.3.0.
Also 1661 'bufref unit tests' on Linux and macOS.

Both unrelated to the PRs where they appeared.

PR #14907, opened 2024-09-14
transfer: remove redundant variable use: select_bits

bagder:
Follow-up to a33bcc9b53b9d6ea9f879624e8

Pointed out by CodeSonar

### DIFF
--- a/lib/transfer.c
+++ b/lib/transfer.c
@@ -424,7 +424,6 @@ CURLcode Curl_sendrecv(struct Curl_easy *data, struct curltime *nowp)
   struct SingleRequest *k = &data->req;
   CURLcode result = CURLE_OK;
   int didwhat = 0;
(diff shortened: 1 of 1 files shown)


icing (approved):
Looks reasonable. Maybe when this has settled, we blast `data->state.select_bits` as well. So 8.11.0 does not get too boring.

dfandrich:
Analysis of PR #14907 at [519c8b3b](https://github.com/curl/curl/commit/519c8b3b322ca09080a473413936ea80284091a3):

[Test 1540 failed](https://github.com/curl/curl/actions/runs/10861028598/job/30142412941#check-step-3), which has NOT been flaky recently, so **there could be a real issue in the PR**.

bagder:
We can't have boring releases.

PR #14910, opened 2024-09-14
GHA/macos: replace make with ninja for cmake builds

vszakats:
Ninja performs better than GNU Make. Also the generation step is a bit
faster. The cost is installing ninja where missing from the image, and
there is no pre-existing package install step. Otherwise the
installation is fast because the tool is tiny.

(It would probably make sense to include ninja on the runner image; it's
a near zero-cost optimization opportunity for all CI jobs using cmake.)

Speed-ups:

Build step: 3s → 1s.
Build tests step: 37s → 22s.

before: https://github.com/curl/curl/actions/runs/10865247916/job/30151507582?pr=14911
after: https://github.com/curl/curl/actions/runs/10865301046/job/30151610055?pr=14910

---

This can be applied to all cmake CI jobs.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -365,8 +365,8 @@ jobs:
       - name: 'build-cert'
         if: contains(matrix.build.generate, '-DCURL_USE_SECTRANSP=ON')
         run: |
-          ninja -C bld/tests/certs clean-certs
-          ninja -C bld/tests/certs build-certs -j 1
+          ninja -C bld clean-certs
+          ninja -C bld build-certs -j 1
 
       - name: 'cmake build'
         run: ninja -C bld --verbose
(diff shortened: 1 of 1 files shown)


vszakats:
It makes no sense for combinations builds, because ninja is not on the runner image, and needs a brew install just to install ninja, which takes 40 seconds, to save 3-4s in the build step.

Will rework this to use ninja for the full (non-combination) cmake builds.

Once ninja comes with the runner image, this optimization will make sense.

vszakats:
Rebased on this, this CI run containing all optimizations so far:
https://github.com/curl/curl/actions/runs/10867081371/job/30155444456?pr=14859

Builds curl in ZERO seconds and all tests in 2s.

Down from 2s and 3s respectively without this (ninja) patch:
https://github.com/curl/curl/actions/runs/10863902225/job/30148585629?pr=14772

PR #14912, opened 2024-09-14
lib, src, tests: added space around ternary expressions

gbrlmarn:
Added space around ternary expressions to be in accordance with code style

PR #14921, opened 2024-09-15
checksrc: check for spaces around '?', '>' and '<'

bagder:
Inspired by #14912

### DIFF
--- a/tests/server/util.c
+++ b/tests/server/util.c
@@ -562,15 +562,22 @@ static BOOL WINAPI ctrl_event_handler(DWORD dwCtrlType)
   logmsg("ctrl_event_handler: %lu", dwCtrlType);
   switch(dwCtrlType) {
 #ifdef SIGINT
-    case CTRL_C_EVENT: signum = SIGINT; break;
+  case CTRL_C_EVENT:
+    signum = SIGINT;
+    break;
 #endif
 #ifdef SIGTERM
-    case CTRL_CLOSE_EVENT: signum = SIGTERM; break;
+  case CTRL_CLOSE_EVENT:
+    signum = SIGTERM;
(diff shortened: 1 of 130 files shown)


dfandrich:
Analysis of PR #14921 at [07ffe6ff](https://github.com/curl/curl/commit/07ffe6ffb51fb37575141e7cabbece2f09225c53):

[Test 1185 failed](https://circleci.com/gh/curl/curl/107004), which has NOT been flaky recently, so **there could be a real issue in the PR**.  Note that this test has failed in 94 different CI jobs (the link just goes to one of them).

tests/server/util.c
```diff
@@ -671,7 +671,7 @@ static SIGHANDLER_T set_signal(int signum, SIGHANDLER_T handler,
   sa.sa_handler = handler;
   sigemptyset(&sa.sa_mask);
   sigaddset(&sa.sa_mask, signum);
-  sa.sa_flags = restartable? SA_RESTART: 0;
+  sa.sa_flags = restartable ? SA_RESTART: 0;
```
MarcelRaad, line 674: These asymmetric ones look strange to me now, where there are spaces on both sides of `?` but only on one side of `:`. Looks good to me other than that.

bagder: It's not intentional, but checksrc does not check the colons (yet).

bagder:
Verifying spaces around the colons turns out to be quite difficult without catching a few false positives: they are used for switches, goto labels and bit-size indicators in struct declarations where we do not want to insist on use of space. So I'll not add that check to checksrc in this PR but I used an attempt to find a bunch of occurrences fixed here.

bagder:
Buckle up!

PR #14928, opened 2024-09-15
GHA/non-native: install Perl for FreeBSD cmake jobs

vszakats:
It was implicitly installed for autotools jobs. Install it explicitly
for cmake ones.


vszakats:
I was thinking to leave perl off, to have a job testing the no-perl situation, but that's probably best done on a platform with more tests.

PR #14922, opened 2024-09-15
GHA/windows: add MSVC vcpkg MSH3 job

talregev:
Add msvc msh3 job
Fix compilation warning / error on test: Unreachable code.
Test 1014 failing. Should fix.

Http2 don't compile with msh3. Is it by design? 

Compilation error:
```
D:\a\curl\curl\lib\vquic\curl_msh3.c(251,13): error C2084: function 'void drain_stream(Curl_cfilter *,Curl_easy *,h2_stream_ctx *)' already has a body [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
      D:\a\curl\curl\lib\http2.c(352,13):
      see previous definition of 'drain_stream'
  
D:\a\curl\curl\lib\vquic\curl_msh3.c(599,7): error C2198: 'void drain_stream(Curl_cfilter *,Curl_easy *,h2_stream_ctx *)': too few arguments for call [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
  
D:\a\curl\curl\lib\vquic\curl_msh3.c(725,7): error C2198: 'void drain_stream(Curl_cfilter *,Curl_easy *,h2_stream_ctx *)': too few arguments for call [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
  
D:\a\curl\curl\lib\vquic\curl_msh3.c(7[29](https://github.com/curl/curl/actions/runs/10873342646/job/30169351365?pr=14922#step:10:29),7): error C2198: 'void drain_stream(Curl_cfilter *,Curl_easy *,h2_stream_ctx *)': too few arguments for call [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
  
D:\a\curl\curl\lib\vquic\curl_msh3.c(752,7): error C2198: 'void drain_stream(Curl_cfilter *,Curl_easy *,h2_stream_ctx *)': too few arguments for call [D:\a\curl\curl\bld\lib\libcurl_object.vcxproj]
  (compiling source file 'CMakeFiles/libcurl_object.dir/Unity/unity_0_c.c')
  
[...]

PR #14932, opened 2024-09-16
lib: fix unity builds with BearSSL, MSH3, Quiche, OmniOS

vszakats:
- fix MSH3 static symbol clash.
- fix Quiche static symbol clash.
- fix local macro clash with BearSSL header.
- fix local macro clash with OmniOS system header.
  ```
  In file included from ../../lib/urldata.h:197,
                     from ../../lib/altsvc.c:32,
                     from libcurlall.c:2:
    ../../lib/cf-socket.h:55:25: error: expected ':', ',', ';', '}' or '__attribute__' before '.' token
       55 | #define sa_addr _sa_ex_u.addr
          |                         ^
    In file included from ../../lib/urldata.h:197,
                     from ../../lib/altsvc.c:32,
                     from libcurlall.c:2:
    ../../lib/cf-socket.h:55:25: error: expected ':', ',', ';', '}' or '__attribute__' before '.' token
       55 | #define sa_addr _sa_ex_u.addr
          |                         ^
  ```
  Ref: https://github.com/curl/curl/actions/runs/10738314933/job/29781644299?pr=14772#step:3:6115

Discovered while adding support for "unity" builds for autotools.

Required-by: #14922
Cherry-picked from #14815

---

w/o whitespace: https://github.com/curl/curl/pull/14932/files?w=1

### DIFF
--- a/lib/cf-socket.c
+++ b/lib/cf-socket.c
@@ -336,7 +336,7 @@ void Curl_sock_assign_addr(struct Curl_sockaddr_ex *dest,
 
   if(dest->addrlen > sizeof(struct Curl_sockaddr_storage))
     dest->addrlen = sizeof(struct Curl_sockaddr_storage);
-  memcpy(&dest->sa_addr, ai->ai_addr, dest->addrlen);
+  memcpy(&dest->curl_sa_addr, ai->ai_addr, dest->addrlen);
 }
 
 static CURLcode socket_open(struct Curl_easy *data,
@@ -372,7 +372,7 @@ static CURLcode socket_open(struct Curl_easy *data,
 
 #if defined(USE_IPV6) && defined(HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID)
   if(data->conn->scope_id && (addr->family == AF_INET6)) {
-    struct sockaddr_in6 * const sa6 = (void *)&addr->sa_addr;
+    struct sockaddr_in6 * const sa6 = (void *)&addr->curl_sa_addr;
     sa6->sin6_scope_id = data->conn->scope_id;
   }
 #endif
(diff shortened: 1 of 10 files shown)


vszakats:
Proposing it for 8.10.1.

bagder:
> fix local macro clash with OmniOS system header. 

`sa_addr` has been defined like this since 2011 and I'm pretty sure it has been built for OmniOS many times since. How come this becomes a problem in a unity build?

vszakats:
> > fix local macro clash with OmniOS system header.
> 
> `sa_addr` has been defined like this since 2011 and I'm pretty sure it has been built for OmniOS many times since. How come this becomes a problem in a unity build?

In "unity" all code ends up in a single source, making it catch more symbol clashes and mixups.

IOW it raises the risk for clashes when using non-namespaced local symbols because the namespace is filled with _all_ the symbols used by libcurl, not just those pulled into that specific source file. Then, depending on the order of define/declare and use, it causes a redefinition issue, or messes up the user code that comes after the definition/declaration.

I think what happens here is that `sa_addr` gets defined in libcurl (via altsvc → urldata.h → cf-socket.h) _before_ including an OmniOS header that happens to use this symbol, and breaks it. With non-unity this doesn't happen because the scope of the definition is more limited, and defined _after_ including system headers.

`sa_addr` seems to be used in `net/if.h` in OmniOS:
https://github.com/omniosorg/illumos-omnios/blob/1a0b3ccac61c4543fa90d8e90de1a882e5491521/usr/src/uts/common/net/if.h#L415

dfandrich:
Analysis of PR #14932 at [f4037249](https://github.com/curl/curl/commit/f4037249fbea9fb4dfbe491014097f0988d9349c):

[Test 301 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 306 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 300 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 304 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 309 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 325 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 303 failed](https://github.com/curl/curl/actions/runs/10885164004/job/30202422939#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

PR #14936, opened 2024-09-16
WebSockets: make support official (non-experimental)

bagder:
Inverts the configure/cmake options to instead provide options that disable WebSockets and have them (ws + wss) enabled by default.


dfandrich:
Analysis of PR #14936 at [f097fa1e](https://github.com/curl/curl/commit/f097fa1ec359393f2967bde4ae05610deb5f2270):

[Test 1013 failed](https://circleci.com/gh/curl/curl/107108), which has NOT been flaky recently, so **there could be a real issue in the PR**.  Note that this test has failed in 33 different CI jobs (the link just goes to one of them).

[Test 1597 failed](https://github.com/curl/curl/actions/runs/10886149882/job/30205483516#check-step-41), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 300 failed](https://github.com/curl/curl/actions/runs/10886149877/job/30205484253#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 301 failed](https://github.com/curl/curl/actions/runs/10886149877/job/30205484253#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 304 failed](https://github.com/curl/curl/actions/runs/10886149877/job/30205484253#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 306 failed](https://github.com/curl/curl/actions/runs/10886149877/job/30205484253#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 309 failed](https://github.com/curl/curl/actions/runs/10886149877/job/30205484253#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

[Test 325 failed](https://github.com/curl/curl/actions/runs/10886149877/job/30205484253#check-step-14), which has NOT been flaky recently, so **there could be a real issue in the PR**.

dg0yt:
> Inverts the configure/cmake options to instead provide options that disable WebSockets

Not in the release notes. :disappointed:
Also not reported as unused by CMake due to https://github.com/curl/curl/blob/b1ef0e1a01c0bb6ee5367bd9c186a603bde3615a/CMakeLists.txt#L49-L64
Another thing to patch out?

vszakats:
> Another thing to patch out?

It's useful to know how curl was built when debugging, though not supercritical to have this in `buildinfo.txt`. I'll make a PR to delete it and avoid this side-effect.

Is there a better method to read the cmake command-line?

dg0yt:
> Is there a better method to read the cmake command-line?

It doesn't matter. You cannot assume to see *the complete actual configuration* in the command line of a single invocation. Input variables = cache variables. I can re-run cmake with no parameters, or I can change a single parameter. And then there are presets etc.

What you have is CMakeCache.txt ;-)

vszakats:
> > Is there a better method to read the cmake command-line?
> 
> It doesn't matter. You cannot assume to see _the complete actual configuration_ in the command line of a single invocation. Input variables = cache variables. I can re-run cmake with no parameters, or I can change a single parameter. And then there are presets etc.

In many cases the command-line tells enough to be useful.

> What you have is CMakeCache.txt ;-)

Yes, there is also `curl_dumpvars()`, but for `buildinfo.txt` they're overkill. It was just meant to match `./configure`'s `$ac_configure_args`.

dg0yt:
> In many cases the command-line tells enough to be useful.

... assuming a single invocation of CMake for configuration purposes.

PR #14944, opened 2024-09-17
test, https server check

icing:
When a https test server is responsive, we do not need to check the stunnel-proxied http server. It already took part in the first check.

When we tear down a https stunnel server, also tear down the http server behind it. We do not know in which state this is when the stunnel does not properly respond.


icing:
replaced with the squashed #14960

PR #14947, opened 2024-09-17
GHA/windows: revert enabling SSPI option

vszakats:
Because it reduces test coverage significantly:
```
TESTINFO: "curl has SSPI support" 109 times (64, 65, 67, 68, 69, 70, 72, 81, 88 and 100 more)
```

Follow-up to e53523fef07894991c69d907a7c7794c7ada4ff4 #14859


mback2k:
But doesn’t Schannel require SSPI or is this about other SSL backend builds?

vszakats:
This is about two OpenSSL jobs. (and an Schannel one, where it was enabled by default, so it didn't make a difference, just redundant.)

